Issue #2071, opened 2019-12-31
Region size separation when importing data with release3.0


lhy1024:
After a week of testing, it was basically determined that region size will separation without #1823 when speed is default.

### Cluster
tidb * 1 + tikv * 6 + pd * 3

### Config
```
config set high-space-ratio 0.05
config set low-space-ratio 0.25  
config set location-labels ""
```
### Branch 
#1991 #1816 cherry pick #1886+#1890
#1992 #1823 cherry pick #1886+#1890
tikv #tikv/6240

## Result 
<p align="center">
  <img src="https://user-images.githubusercontent.com/19542290/71617826-d0e7f380-2bf7-11ea-82b5-332076c38069.png"><br> 
  Figure 1: balance-test region size pr1991
</p>
<p align="center">
  <img src="https://user-images.githubusercontent.com/19542290/71617832-dba28880-2bf7-11ea-8cb9-014226db1984.png"><br> 
  Figure 2: balance-test operator pr1991
</p>

In the three sets of tests, the balance-region always fails to converge, and the largest region size  is twice the smallest. After turning off go-ycsb, it slowly converged after 20 hours. We think that black list block region balance.

<p align="center">
  <img src="https://user-images.githubusercontent.com/19542290/71617861-0987cd00-2bf8-11ea-9d0d-b94cf4b682f4.png"><br> 
  Figure 3: balance-test operator pr1991
</p>

There is no dispersion and volatility of Region size at slow speeds, indicating that # 1992 can adapt at low speeds.

nolouch:
move to #2002

Issue #2002, opened 2019-12-10
The introduced dispersion and volatility of Region size


lhy1024:
After a week of testing, it was basically determined that #1823 caused the region size to diverge and fluctuate in the second function.

### Cluster
tidb * 1 + tikv * 6 + pd * 3

### Config
```
stores set limit 128
config set region-schedule-limit 65536
config set max-snapshot-count 1024
config set max-pending-peer-count 1024
config set store-balance-rate 65536
config set high-space-ratio 0.05
config set low-space-ratio 0.25  
config set location-labels ""
```
### Branch 
#1991 #1816 cherry pick #1886+#1890
#1992 #1823 cherry pick #1886+#1890

## Result 

<p align="center">
  <img src="https://user-images.githubusercontent.com/19542290/70498369-1e20fa00-1b51-11ea-909f-103239ad5fb7.png"><br> 
  Figure 1: balance-test48 pr1991
</p>
<p align="center">
  <img src="https://user-images.githubusercontent.com/19542290/70498377-2416db00-1b51-11ea-80f9-19e185ff707d.png"><br> 
  Figure 2: balance-test58 pr1992
</p>

### coefficient of variation

#1991 |  The Proportion coefficient of variation below 0.075 | The Proportion coefficient of variation below 0.05 | The average of coefficient of variation
-- | -- | -- | --
test48 | 100.0% | 87.15% | 0.039
test61 | 99.75% | 87.59% | 0.035
test62 | 100.0% | 84.05% | 0.042
test63 | 100.0% | 85.54% | 0.043
test64 | 100.0% | 98.98% | 0.030
average | 99.95% | 88.66% | 0.038

#1992 | The Proportion coefficient of variation below 0.075 | The Proportion coefficient of variation below 0.05 | The average of coefficient of variation
-- | -- | -- | --
test58 | 20.12% | 0.02% | 0.113
test67 | 27.27% | 0.06% | 0.100
test68 | 28.29% | 0.01% | 0.093
test69 | 37.14% | 0.17% | 0.102
test70 | 29.15% | 0.11% | 0.089
average | 28.39% | 0.07% | 0.099

lhy1024:
@disksing

nolouch:
I think we imporve it after 5.0. close fistly.

PR #2090, opened 2020-01-09
scheduler: add operator influence when sort store

lhy1024:
Signed-off-by: lhy1024 <admin@liudos.us>

### What problem does this PR solve? 

After #1823 , we changed balance region scheduler, but we didn't consider operator influence when sort store.

### What is changed and how it works?

Add operator influence when sort store

- [x] Test.

### Check List 

Tests 

 - Unit test

PR #1795, opened 2019-10-11
scheduler: adjust max score for balance region 

lhy1024:
### What problem does this PR solve?

There are redundant schedulings occuring near the 40% inflection point, but not in other intervals. 
There are two reasons for this phenomenon:
1. we use piecewise function
2. the second slope to be locked at around 5000 because of huge max_score. The slope is too larger than the first slope which is only 1.
The PR will adjust max score to solve the second problem.

### What is changed and how it works?

The calculation method of max_score is as follows: 
Get the maximum capacity in the cluster, and then let max_score = f(max_capacity). If capacity of the new store is lower than max_score, it will not be adjusted to avoid excessive score shake of the new  store.Otherwise, we will set the max_score of all the store to the new one.
There are several plans for this function:
1. max_score = max_capacity * 10
2. max_score = max_capacity + 10t
3. max_score = min(max_capacity * 10, max_capacity + 10t)
4. max_score = max_capacity

The PR uses the fourth function.

- [x] add unit test
- [x] contrast test
- [ ] add a bigger store
### Check List 

Tests 

 - Unit test

PR #2093, opened 2020-01-09
scheduler: add trend monitor to avoid redundant scheduler

lhy1024:

it will try to solve #2071 and #2002 at the same time.

image
It is common when the store used size increases, and at this time the region size is decreasing, and it is constantly moving out. We should avoid this situation, so the `trend monitor` is added, and scheduling is allowed only when the two trends are the same.

- [ ] test
- [ ] unit test and refactor

### Check List 

Tests 

 - Unit test


lhy1024:
> Is this better than #2090?

It is a supplement to #2090 and #1795, and is used to deal with scenarios where the scheduling is too fast, or the op influence cannot limit the scheduling.

PR #2096, opened 2020-01-15
schedule, config: use leading dollar sign to mark exclusive labels

disksing:
We need maintain exclusive labels list in code. Not extensible.

- labels start with '$' are recognized as exclusive.
- use different format check for label keys and label values

### Check List 

Tests 
 - Unit test
 - Manual test (test setting and deleting labels)


disksing:
@lhy1024 When the store contains such labels, the label must be explicitly specified in the rule to fit. For example, '$engine' is exclusive label because tiflash stores(with label `$engine=tiflash`) are only used for tiflash, we don't want all other rules to use tiflash stores.

PR #2098, opened 2020-01-15
core, placement: store placement rule in separeted kvs

disksing:

Current implementation of placement rules stores all rules in a single etcd KV. If we have thouthands of rules, the single kv is updated many times and cause etcd out of space (2G by default).

Relates to (and partially fix) https://github.com/pingcap/pd/issues/2094

- store each rule separetly

### Check List 

Tests 
 - Unit test

`Use a script to create ~10000 rules, etcd space usage is about 120MiB.`

Code changes
 - Has persistent data change

Side effects
 - Breaking backward compatibility (incompatible with old storage location)


server/core/storage.go
```diff
@@ -222,29 +222,42 @@ func (s *Storage) LoadConfig(cfg interface{}) (bool, error) {
...
-	}
-	if value == "" {
-		return false, nil
-	}
-	err = json.Unmarshal([]byte(value), rules)
-	if err != nil {
-		return false, errors.WithStack(err)
+func (s *Storage) LoadRules(f func(v string) error) (bool, error) {
+	nextKey := path.Join(rulesPath, "\x00")
+	endKey := rulesPath + "0" // upper bound of all rules.
```
lhy1024: I think we need some comment about `\x00` and `"0"`

server/core/storage.go
```diff
@@ -222,29 +222,42 @@ func (s *Storage) LoadConfig(cfg interface{}) (bool, error) {
...
-	if err != nil {
-		return false, err
-	}
-	if value == "" {
-		return false, nil
-	}
-	err = json.Unmarshal([]byte(value), rules)
-	if err != nil {
-		return false, errors.WithStack(err)
+func (s *Storage) LoadRules(f func(v string) error) (bool, error) {
```
lhy1024: Do we need some unit test for updated `LoadRules`?

server/core/storage.go
disksing: good idea

server/core/storage.go
disksing: It can be covered by this one: https://github.com/pingcap/pd/blob/master/server/schedule/placement/rule_manager_test.go#L73

nolouch:
PTAL @lhy1024

PR #2099, opened 2020-01-16
scheduler: refactor hotspot scheduler

Luffbee:

The hotspot scheduler has duplicate code and its logic is not very clear.

Refactored the `hotScheduler`. Mainly three part:
+ Rewrite `pendingOpInfos` related code (introduced by PR #2046  and rename it to `regionPendings`. (the 1st commit)
+ Move duplicate code in `balanceHot{Read, Write}Regions()` into `balanceBy{Peer, Leader}()`. (the 2ed commit)
+ Use `balanceSolver` to replace `balanceBy{Peer, Leader}()`. (the 4th & 5th commit).

The reason for introducing `balanceSolver` is that the logic of `balanceBy{Peer, Leader}()` are similar, which can be described generally as `balanceSolver.solve()`.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -14,7 +14,6 @@
 package schedulers
 
 import (
-	"container/list"
 	"math"
 	"math/rand"
 	"sync"
@@ -50,19 +49,21 @@ func init() {
 }
 
 const (
+	// HotRegionName is balance hot region scheduler name.
+	HotRegionName = "balance-hot-region-scheduler"
+
+	HotRegionType      = "hot-region"       // HotRegionType is balance hot region scheduler type.
+	HotReadRegionType  = "hot-read-region"  // HotReadRegionType is hot read region scheduler type.
+	HotWriteRegionType = "hot-write-region" // HotWriteRegionType is hot write region scheduler type.
+
 	hotRegionLimitFactor    = 0.75
 	storeHotPeersDefaultLen = 100
 	hotRegionScheduleFactor = 0.95
-	// HotRegionName is balance hot region scheduler name.
-	HotRegionName = "balance-hot-region-scheduler"
-	// HotRegionType is balance hot region scheduler type.
-	HotRegionType = "hot-region"
-	// HotReadRegionType is hot read region scheduler type.
-	HotReadRegionType = "hot-read-region"
-	// HotWriteRegionType is hot write region scheduler type.
-	HotWriteRegionType               = "hot-write-region"
-	minFlowBytes                     = 128 * 1024
-	maxZombieDur       time.Duration = statistics.StoreHeartBeatReportInterval * time.Second
+
+	minFlowBytes               = 128 * 1024
+	maxZombieDur time.Duration = statistics.StoreHeartBeatReportInterval * time.Second
+
+	minRegionScheduleInterval time.Duration = statistics.StoreHeartBeatReportInterval * time.Second
 )
 
 // BalanceType : the perspective of balance
@@ -71,6 +72,13 @@ type BalanceType int
 const (
 	hotWrite BalanceType = iota
 	hotRead
+)
+
+type opType int
+
+const (
+	byPeer opType = iota
+	byLeader
 )
 
 type storeStatistics struct {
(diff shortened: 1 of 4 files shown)


server/schedulers/hot_region.go
```diff
@@ -241,28 +249,28 @@ func (h *hotScheduler) updateStatsByPendingOpInfo(storeStatsMap map[uint64]float
...
-		var n *list.Element
-		for e := pendingOpInfos.Front(); e != nil; e = n {
-			n = e.Next()
-			opInfo := e.Value.(*pendingInfluence)
-			if opInfo.isDone() {
-				if time.Now().After(opInfo.op.GetCreateTime().Add(statistics.StoreHeartBeatReportInterval * time.Second)) {
+	for regionID, pendings := range h.regionPendings {
+		empty := true
+		for ty, pending := range pendings {
+			if pending != nil && pending.isDone() {
```
nolouch: `pendings = {nil,p2}`, and  p2 is not  done, empty will be true.

server/schedulers/hot_region.go
```diff
@@ -93,73 +104,52 @@ type hotScheduler struct {
...
-		types:          []BalanceType{hotRead},
-		r:              rand.New(rand.NewSource(time.Now().UnixNano())),
-		readScores:     NewScoreInfos(),
-		writeScores:    NewScoreInfos(),
-		readPendings:   map[*pendingInfluence]struct{}{},
-		writePendings:  map[*pendingInfluence]struct{}{},
-		pendingOpInfos: pendingOpInfos,
-	}
+	ret := newHotScheduler(opController)
+	ret.name = ""
```
disksing: why reset name?

Luffbee: The original code didn't set `name`, so it should be empty string.

server/schedulers/hot_region.go
```diff
@@ -279,335 +287,386 @@ func (h *hotScheduler) analyzeStoreLoad(cluster opt.Cluster, balanceType Balance
...
+	}
+
+	if pendings, ok := bs.sche.regionPendings[region.GetID()]; ok {
+		if bs.opTy == transferLeader {
+			return false
 		}
-		if len(candidateStoreIDs) == 0 {
-			continue
+		if pendings[movePeer] != nil ||
+			(pendings[transferLeader] != nil && !pendings[transferLeader].isDone()) {
```
rleungx: Do we need metrics?

server/schedulers/hot_region.go
Luffbee: Fixed and added a test.

server/schedulers/hot_region.go
```diff
@@ -279,335 +287,386 @@ func (h *hotScheduler) analyzeStoreLoad(cluster opt.Cluster, balanceType Balance
...
+
+	return ret
 }
 
-// balanceByPeer balances the peer distribution of hot regions.
-func (h *hotScheduler) balanceByPeer(cluster opt.Cluster, storesStat statistics.StoreHotPeersStat, balanceType BalanceType) (*core.RegionInfo, *metapb.Peer, *metapb.Peer, Influence) {
-	if !h.allowBalanceRegion(cluster) {
-		return nil, nil, nil, Influence{}
+func (h *hotScheduler) balanceHotReadRegions(cluster opt.Cluster) []*operator.Operator {
+	// prefer to balance by leader
```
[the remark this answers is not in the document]
Luffbee: I don't think this should be added in this PR.

server/schedulers/hot_region.go
```diff
@@ -279,335 +287,386 @@ func (h *hotScheduler) analyzeStoreLoad(cluster opt.Cluster, balanceType Balance
...
+		id = bs.sche.selectSrcStoreByScore(bs.storesStat, bs.rwTy)
+	case transferLeader:
+		if bs.rwTy == write {
+			id = bs.sche.selectSrcStoreByHot(bs.storesStat)
 		} else {
-			scoreGuard = filter.NewDistinctScoreFilter(h.GetName(), cluster.GetLocationLabels(), cluster.GetRegionStores(srcRegion), srcStore)
+			id = bs.sche.selectSrcStoreByScore(bs.storesStat, bs.rwTy)
 		}
+	}
+	if id != 0 && bs.cluster.GetStore(id) == nil {
```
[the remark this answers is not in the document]
Luffbee: The original code do not have such metric, so I don't think this should be added in this PR.

server/schedulers/hot_region.go
Luffbee: Maybe add it in another PR.

PR #2105, opened 2020-01-19
dashboard: do not build dashboard-ui on windows

HunDunDM:
Signed-off-by: hundundm <hundundm@gmail.com>

Cannot compile pd-server in Windows.

Determine the OS in the Makefile. If it is Windows, skip the process of building Dashboard-UI and use empty_assets_handler.go instead.

### Check List 

Tests 

    - Just `make` in Windows

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,9 @@ tools: pd-tso-bench pd-recover pd-analysis pd-heartbeat-bench
 pd-server: export GO111MODULE=on
 ifeq ("$(WITH_RACE)", "1")
 pd-server:
+ifneq ($(OS),Windows_NT)
 	./scripts/embed-dashboard-ui.sh
+endif
 	CGO_ENABLED=1 go build -race -gcflags '$(GCFLAGS)' -ldflags '$(LDFLAGS)' -o bin/pd-server cmd/pd-server/main.go
 else ifeq ("$(PD_WEB)", "1")
 pd-server: retool-setup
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -53,7 +53,9 @@ tools: pd-tso-bench pd-recover pd-analysis pd-heartbeat-bench
 pd-server: export GO111MODULE=on
 ifeq ("$(WITH_RACE)", "1")
 pd-server:
+ifneq ($(OS),Windows_NT)
```
lhy1024: why can not build it on windows?

HunDunDM: It's just because the sh script may not be executed successfully in Windows. If necessary later, we may need to implement the powershell version of this script.

pkg/dashboard/uiserver/empty_assets_handler_windows.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package uiserver
+
+import (
+	assetfs "github.com/elazarl/go-bindata-assetfs"
+)
+
+func assetFS() *assetfs.AssetFS {
```
breeswish: Why it will not cause name conflict in non windows?

HunDunDM: Sorry for the late reply.

[Package build](https://golang.org/pkg/go/build/)

If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns:
```
*_GOOS
*_GOARCH
*_GOOS_GOARCH
```
(example: source_windows_amd64.go) where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (in addition to any explicit constraints in the file).

PR #2113, opened 2020-02-10
mod: fix mod using Semantic Import Versioning for release-3.0

shafreeck:
Signed-off-by: Shafreeck Sea <shafreeck@gmail.com

Update the module name according to the rules of Semantic Import Versioning. It will take effect on the next release, after that, users can get the latest v3.0+ version using `go get`

Fix the module name to make `go get pkg@v3.x.x` work

* Rename `github.com/pingcap/pd` to `github.com/pingcap/pd/v3`
* Update `Makefile` to use the new package name.

### Check List 

Tests 


nolouch:
What version should we use on master? v5?

shafreeck:
> What version should we use on master? v5?

I think it should be v4, we can bump it to v5 when creating the release-v5.0 branch

disksing:
Will it break projects that depends on pd client? (e.g. tidb)

disksing:
Will it break projects that depend on pd client? (e.g. tidb)

shafreeck:
> Will it break projects that depend on pd client? (e.g. tidb)

The project using go mod will not be broken.

nolouch:
maybe we need to wait for `v3.0.10` before merge.

PR #2130, opened 2020-02-14
api: add deploy path and binary version in member

rleungx:
We need to know the deploy path of pd binary for tidb dashboard.

This PR adds binary deploy path for member API

### Check List 

Tests 

 - Manual test

Details:
```
➜  ~ curl http://127.0.0.1:2379/pd/api/v1/members
{
  "header": {
    "cluster_id": 6793225628982604172
  },
  "members": [
    {
      "name": "pd2",
      "member_id": 5454843464880847909,
      "peer_urls": [
        "http://127.0.0.1:3380"
      ],
      "client_urls": [
        "http://127.0.0.1:3379"
      ],
      "deploy_path": "/path/to/bin"
    },
    {
      "name": "pd1",
      "member_id": 13195394291058371180,
      "peer_urls": [
        "http://127.0.0.1:2380"
      ],
      "client_urls": [
        "http://127.0.0.1:2379"
      ],
      "deploy_path": "/path/to/bin"
    }
  ],
  "leader": {
    "name": "pd1",
    "member_id": 13195394291058371180,
    "peer_urls": [
      "http://127.0.0.1:2380"
    ],
    "client_urls": [
      "http://127.0.0.1:2379"
    ]
  },
  "etcd_leader": {
    "name": "pd1",
    "member_id": 13195394291058371180,
    "peer_urls": [
      "http://127.0.0.1:2380"
    ],
    "client_urls": [
      "http://127.0.0.1:2379"
    ],
    "deploy_path": "/path/to/bin"
  }
}
```

### DIFF
--- a/server/member/leader.go
+++ b/server/member/leader.go
@@ -19,6 +19,7 @@ import (
 	"math/rand"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -352,7 +353,8 @@ func (m *Member) GetMemberDeployPath(id uint64) (string, error) {
 func (m *Member) SetMemberDeployPath(id uint64) error {
 	key := m.getMemberBinaryDeployPath(id)
 	txn := kv.NewSlowLogTxn(m.client)
-	deployPath, err := os.Executable()
+	execPath, err := os.Executable()
+	deployPath := filepath.Dir(execPath)
 	if err != nil {
 		return errors.WithStack(err)
 	}
--- a/server/member/leader.go
+++ b/server/member/leader.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math/rand"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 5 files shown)


server/member/leader.go
```diff
@@ -330,6 +331,41 @@ func (m *Member) GetMemberLeaderPriority(id uint64) (int, error) {
...
+		return "", errors.New("no value")
+	}
+	return string(res.Kvs[0].Value), nil
+}
+
+// SetMemberDeployPath saves a member's binary deploy path.
+func (m *Member) SetMemberDeployPath(id uint64) error {
+	key := m.getMemberBinaryDeployPath(id)
+	txn := kv.NewSlowLogTxn(m.client)
+	absPath, err := filepath.Abs("")
```
breeswish: This might be CWD instead of deploy path when using binary deployment. How about https://stackoverflow.com/questions/18537257/how-to-get-the-directory-of-the-currently-running-file ?

disksing: I think better use `os.Executable()`.

server/member/leader.go
```diff
@@ -330,6 +331,41 @@ func (m *Member) GetMemberLeaderPriority(id uint64) (int, error) {
...
+		return "", errors.New("no value")
+	}
+	return string(res.Kvs[0].Value), nil
+}
+
+// SetMemberDeployPath saves a member's binary deploy path.
+func (m *Member) SetMemberDeployPath(id uint64) error {
+	key := m.getMemberBinaryDeployPath(id)
+	txn := kv.NewSlowLogTxn(m.client)
+	deployPath, err := os.Executable()
```
disksing: need `filepath.Dir` to get the executable's dir path?

rleungx: Oh, you're right.

breeswish:
Good job! Is it possible to access PD binary versions as well?

rleungx:
> Good job! Is it possible to access PD binary versions as well?

Yes, we can. Does TiDB dashboard also need it?

breeswish:
@rleungx Yes. Although it is a very rare case, but the purpose is exactly to catch those rare cases so that user can be warned when it really happens.

Issue #2136, opened 2020-02-17
docker build is broken


aylei:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
```
docker build . -t aylei/pd
```

2. What did you expect to see?
`docker build` success.

3. What did you see instead?
`docker build` failed with the following error

```
./scripts/embed-dashboard-ui.sh: line 32: curl: command not found
make: *** [Makefile:56: pd-server] Error 127
```

4. What version of PD are you using (`pd-server -V`)?
https://github.com/pingcap/pd/commit/17447e5673294912cd3cd7322ea73aa2327cd3d5

PR #2137, opened 2020-02-17
build: fix docker build

aylei:
#2136 

Install `curl`, `gcc` and `g++` in builder to satisfy the dashboard script and go-sqlite compilation.

### Check List 

Tests 


nolouch:
@Luffbee Good point. I fire an issue firstly.

Issue #2139, opened 2020-02-18
scheduler: hotspot: multidimensional load balance tracking issue


Luffbee:
This is the tracking issue of multidimensional load balance scheduler.

We had design a framework for multidimensional load balancing. Based on this framework, we add `key rate` dimension beside the original `byte rate` and `count` dimensions. 

The implementation is splitted into several PRs:
- [x] Multidimensional hot peer cache, PR: https://github.com/pingcap/pd/pull/2140
- [x] Solution-based key-rate-aware balance solver, PR: https://github.com/pingcap/pd/pull/2141

PR #2141, opened 2020-02-18
scheduler: hotspot: solution-based key-rate-aware balance solver 

Luffbee:

The second step of #2139 . 

Introduce solution based balance solver.

### Check List 

Tests 

 - Unit test


nolouch:
PTAL @disksing @rleungx

server/schedulers/hot_region.go
```diff
@@ -490,29 +512,61 @@ func (bs *balanceSolver) allowBalance() bool {
...
+		return byteSort[i].GetByteRate() > byteSort[j].GetByteRate()
+	})
+	keySort := make([]*statistics.HotPeerStat, len(ret))
+	copy(keySort, ret)
+	sort.Slice(keySort, func(i, j int) bool {
+		return keySort[i].GetKeyRate() > keySort[j].GetKeyRate()
 	})
+
+	union := make(map[*statistics.HotPeerStat]struct{}, maxPeerNum)
+	for len(union) < maxPeerNum {
```
disksing: why need the for loop?

server/schedulers/hot_region.go
```diff
@@ -606,184 +660,258 @@ func (bs *balanceSolver) getDstCandidateIDs() map[uint64]struct{} {
...
-		candidateLoadDetail[id] = bs.stLoadDetail[id]
-	}
-	switch bs.opTy {
-	case movePeer:
-		return selectDstStoreByByteRate(candidateLoadDetail, bs.srcPeerStat.GetByteRate(), bs.stLoadDetail[bs.srcStoreID])
-	case transferLeader:
-		if bs.rwTy == write {
-			return selectDstStoreByCount(candidateLoadDetail, bs.srcPeerStat.GetByteRate(), bs.stLoadDetail[bs.srcStoreID])
+// calcProgressiveRank checks `bs.cur` is a progressive solution.
+func (bs *balanceSolver) calcProgressiveRank() int64 {
```
disksing: add some comments about how or/and why ranks are determined?

server/schedulers/hot_region.go
```diff
@@ -357,23 +345,18 @@ func (h *hotScheduler) balanceHotReadRegions(cluster opt.Cluster) []*operator.Op
...
-
 func (h *hotScheduler) balanceHotWriteRegions(cluster opt.Cluster) []*operator.Operator {
-	for i := 0; i < balanceHotRetryLimit; i++ {
-		// prefer to balance by peer
-		peerSolver := newBalanceSolver(h, cluster, write, movePeer)
-		ops := peerSolver.solve()
-		if len(ops) > 0 {
-			return ops
-		}
+	// prefer to balance by peer
```
rleungx: After removing the retry loop, will it affect the speed of hot spot scheduling?

server/schedulers/hot_region.go
```diff
@@ -606,184 +660,258 @@ func (bs *balanceSolver) getDstCandidateIDs() map[uint64]struct{} {
...
+	peer := bs.cur.srcPeerStat
+	rank := int64(0)
+	if bs.rwTy == write && bs.opTy == transferLeader {
+		if srcLd.Count > dstLd.Count &&
+			srcLd.KeyRate >= dstLd.KeyRate+peer.GetKeyRate() {
+			rank = -1
+		}
+	} else {
+		keyDecRatio := (dstLd.KeyRate + peer.GetKeyRate()) / (srcLd.KeyRate + 1)
+		keyHot := peer.GetKeyRate() >= 10
```
rleungx: how about using a constant?

server/schedulers/hot_region.go
```diff
@@ -387,28 +370,54 @@ type balanceSolver struct {
...
+	for _, detail := range bs.stLoadDetail {
+		bs.maxSrc = maxLoad(bs.maxSrc, detail.LoadPred.min())
+		bs.minDst = minLoad(bs.minDst, detail.LoadPred.max())
+		maxCur = maxLoad(maxCur, &detail.LoadPred.Current)
+	}
+
+	bs.rankStep = &storeLoad{
+		ByteRate: maxCur.ByteRate / 20,
+		KeyRate:  maxCur.KeyRate / 20,
+		Count:    maxCur.Count / 10,
```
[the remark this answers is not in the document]
Luffbee: These are parameters chosen by instinct. I don't know how to comment. Any suggestions?

server/schedulers/hot_region.go
Luffbee: To prevent return too many peers, which will make `balanceSolver.solve()` too slow.

server/schedulers/hot_region.go
```diff
@@ -903,3 +1041,27 @@ func (ty opType) String() string {
...
+		switch opTy {
+		case movePeer:
+			return writePeer
+		case transferLeader:
+			return writeLeader
+		}
+	case read:
+		return readLeader
+	}
+	return resourceTypeLen
```
rleungx: Will it cause panic?

server/schedulers/hot_region.go
Luffbee: I will add comments about meaning of the "rank". As the implementation of this function is desired to change in the future,  I think "how or why ranks are determined" is too  detailed as function comment. I will add "how or why"  comment in the implementation.

server/schedulers/hot_region.go
Luffbee: I believe it won't. The the statistics data is not changed for each iteration, and new `balanceSolver` will try almost all possible solutions.

server/schedulers/hot_region.go
Luffbee: It won't be used in other place, so I don't think it is necessary.

server/schedulers/hot_region.go
Luffbee: It will only cause panic when the arguments are invalid. I think this is acceptable.

server/schedulers/hot_region.go
rleungx: The constant is easy to maintain and the meaning is much more clear.

server/schedulers/hot_region.go
rleungx: It's better to panic in this function rather than use it?

server/schedulers/hot_region.go
```diff
@@ -455,28 +464,41 @@ func (bs *balanceSolver) solve() []*operator.Operator {
...
 		return nil
 	}
-	bs.srcStoreID = bs.selectSrcStoreID()
-	if bs.srcStoreID == 0 {
-		return nil
-	}
+	bs.cur = &solution{}
+	var best *solution = nil
+	var ops []*operator.Operator = nil
+	var infls []Influence = nil
```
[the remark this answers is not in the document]
Luffbee: Removed.

server/schedulers/hot_region.go
Luffbee: Comments added.

server/schedulers/hot_region.go
Luffbee: Add two constants: `minHotByteRate` and `minHotKeyRate`.

server/schedulers/hot_region.go
Luffbee: Fixed.

server/schedulers/hot_region.go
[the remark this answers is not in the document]
disksing: em, how about define 3 load weight constants as 0.5, 0.5, 0.1?

server/schedulers/hot_region.go
disksing: Got.

server/schedulers/hot_region.go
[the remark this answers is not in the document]
Luffbee: Added three constants: `{byteRate,keyRate,count}RankStepRatio`.

server/schedulers/hot_region.go
```diff
@@ -606,184 +672,264 @@ func (bs *balanceSolver) getDstCandidateIDs() map[uint64]struct{} {
...
-			return true
+// smaller is better
+func (bs *balanceSolver) compareSrcStore(st1, st2 uint64) int {
+	if st1 != st2 {
+		// compare source store
+		var lpCmp storeLPCmp
+		if bs.rwTy == write && bs.opTy == transferLeader {
+			lpCmp = sliceLPCmp(
+				minLPCmp(negLoadCmp(sliceLoadCmp(
+					stLdRankCmp(stLdCount, stepRank(bs.maxSrc.Count, bs.rankStep.Count)),
```
nolouch: What the meaning of `rankStep`? it seems always the same in the rank compare.

server/schedulers/hot_region.go
```diff
@@ -606,184 +672,264 @@ func (bs *balanceSolver) getDstCandidateIDs() map[uint64]struct{} {
...
+	rank := int64(0)
+	if bs.rwTy == write && bs.opTy == transferLeader {
+		// In this condition, CPU usage is the matter.
+		// Only consider about count and key rate.
+		if srcLd.Count > dstLd.Count &&
+			srcLd.KeyRate >= dstLd.KeyRate+peer.GetKeyRate() {
+			rank = -1
+		}
+	} else {
+		keyDecRatio := (dstLd.KeyRate + peer.GetKeyRate()) / (srcLd.KeyRate + 1)
```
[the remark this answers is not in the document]
Luffbee: The `+ 1` is used to prevent `NaN` (it won't panic). In the origin code, it is `0.95 * srcLd.KeyRate >= dstLd.KeyRate + peer.GetKeyRate()`.

server/schedulers/hot_region.go
Luffbee: It is used to discretization. Map `float64` to `int64`. For all values in the interval `[rank0 + i * step, rank0 + (i+1) * step)`,  the rank the `i`.

server/cluster/coordinator.go
```diff
@@ -420,6 +420,8 @@ func (c *coordinator) collectHotSpotMetrics() {
 		infl := pendings[storeID]
 		// TODO: add to tidb-ansible after merging pending influence into operator influence.
```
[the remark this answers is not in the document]
Luffbee: I think this is not related to the multi-dim load balance issue.

server/schedulers/hot_region.go
```diff
@@ -490,29 +524,61 @@ func (bs *balanceSolver) allowBalance() bool {
...
-	}
-	return id
+	return ret
 }
 
-func (bs *balanceSolver) getPeerList() []*statistics.HotPeerStat {
-	ret := bs.stLoadDetail[bs.srcStoreID].HotPeers
-	bs.sche.r.Shuffle(len(ret), func(i, j int) {
-		ret[i], ret[j] = ret[j], ret[i]
+func (bs *balanceSolver) getHotPeers() []*statistics.HotPeerStat {
```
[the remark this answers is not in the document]
Luffbee: I renamed this function to `filterHotPeers`, and added comments in this function.

server/schedulers/hot_region.go
```diff
@@ -606,184 +672,264 @@ func (bs *balanceSolver) getDstCandidateIDs() map[uint64]struct{} {
...
+			srcLd.KeyRate >= dstLd.KeyRate+peer.GetKeyRate() {
+			rank = -1
+		}
+	} else {
+		keyDecRatio := (dstLd.KeyRate + peer.GetKeyRate()) / (srcLd.KeyRate + 1)
+		keyHot := peer.GetKeyRate() >= minHotKeyRate
+		byteDecRatio := (dstLd.ByteRate + peer.GetByteRate()) / (srcLd.ByteRate + 1)
+		byteHot := peer.GetByteRate() > minHotByteRate
+		switch {
+		case byteHot && byteDecRatio <= 0.95 && keyHot && keyDecRatio <= 0.95:
```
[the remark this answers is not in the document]
Luffbee: Add constants: `{great, minor}DecRatio`.

server/schedulers/hot_region.go
nolouch: Got it. If I understand correctly, the ranking `i` (stepRank return) can only be 0 or 1.  Can this distinguish those stories very well?

Luffbee: No, rank  `i = int64((value - rank0) / step)` can be any `int64`.

nolouch: Oh, Sorry. I missed the `step ratio`, can u add some comments about step ratio?

Luffbee: Comments added.

PR #2143, opened 2020-02-18
cluster: intervally persistent store meta 

### LINKED ISSUE #2135: Persistent store last heartbeat ts to etcd

When PD leader changes after a store goes down, it will start to make up replicas for the store immediately (without waiting 30 minutes). This is because store's last heartbeat ts is not persistent to etcd.

save store's last heartbeat time to etcd.

disksing:
Fix #2135

~DNM Note: It should be merged after https://github.com/pingcap/kvproto/pull/559~

- add lastHeartbeat in metapb.Store
- maintain lastSaveHeartbeat in StoreInfo and save meta to etcd every 5 minutes
- adjust tests

### Check List 

Tests 
 - Unit test
    - start pd+tikv, use API to watch lastHeartbeatTS, it upates every 10s
    - close tikv and restart pd, lastHeartbeatTS rollback 0-4 minutes (instead of become 1970-1-1)

Related changes
 - Need to be included in the release notes


server/core/store.go
```diff
@@ -27,6 +27,9 @@ import (
 	"go.uber.org/zap"
 )
 
+// Interval to save store meta (including heartbeat ts) to etcd.
+const storePersistInterval = 5 * time.Minute
```
rleungx: Do we need to let users know the exact downtime may not as they expected under this case?

disksing: yes, I can update the document.

PR #2150, opened 2020-02-25
filter, scheduler: support filter stores for special use

disksing:
In the future, k8s will temporarily add some TiKV according to balance excessive workload. Since these newly added nodes will be removed later, they should only be visible to hotRegionScheduler.

- add `specialUseFilter`
- consider `specialUseFilter` in schedulers
- unit test
- manual test

### Check List 

Tests 
 - Unit test

<details>

```
- init with store 1, 2, 3, 4. 4 has label `specialUse = hotRegion`
- make sure no region has peer on store4
- use `go-ycsb` to import data, there are a few regions transfer peer to store4. Check pd's log, the operator is `move-hot-write-leader` or `move-hot-write-region`.
- use `go-ycsb` to read a segment, there are a few regions transfer peer to store4. Check pd's log, the operator is `transfer-hot-read-leader`.
- `pd-ctl store delete 4`. Regions are moved out of 4 and 4 turned to tombstone state.
```

</details>


disksing:
@rleungx could you take a look?

server/schedulers/balance_leader.go
```diff
@@ -90,7 +90,10 @@ func newBalanceLeaderScheduler(opController *schedule.OperatorController, conf *
 	for _, opt := range opts {
 		opt(s)
 	}
-	s.filters = []filter.Filter{filter.StoreStateFilter{ActionScope: s.GetName(), TransferLeader: true}}
+	s.filters = []filter.Filter{
+		filter.StoreStateFilter{ActionScope: s.GetName(), TransferLeader: true},
+		filter.NewSpecialUseFilter(AdjacentRegionName),
```
rleungx: AdjacentRegionName？

PR #2153, opened 2020-02-25
config: rewrite and persist config file

rleungx:
After we enable the dynamic config, we will rewrite the config file and persist it once the config has been updated. Related to tikv/tikv#6684

The PR adds a new function named `RewriteFile` to do it.

### Check List 

Tests 

 - Manual test


server/config/config.go
```diff
@@ -991,6 +993,25 @@ func (c *Config) GetZapLogProperties() *log.ZapProperties {
...
+	filePath := c.GetConfigFile()
+	if filePath == "" {
+		return nil
+	}
+	var buf bytes.Buffer
+	if err := toml.NewEncoder(&buf).Encode(*new); err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filePath, buf.Bytes(), 0644)
```
overvenus: Better to state what `0644` means.

server/config/config.go
```diff
@@ -991,6 +993,25 @@ func (c *Config) GetZapLogProperties() *log.ZapProperties {
...
+}
+
+// RewriteFile rewrites the config file after updating the config.
+func (c *Config) RewriteFile(new *Config) error {
+	filePath := c.GetConfigFile()
+	if filePath == "" {
+		return nil
+	}
+	var buf bytes.Buffer
+	if err := toml.NewEncoder(&buf).Encode(*new); err != nil {
```
overvenus: Looks like it also writes out hidden configurations.

server/config_manager/config_manager.go
```diff
@@ -573,12 +573,17 @@ func getUpdateValue(item, updateItem interface{}) (interface{}, error) {
 			return nil, errors.Errorf("unexpected type: %T\n", t1)
 		}
 	case reflect.Slice:
-		strSlice := strings.Split(updateItem.(string), ",")
-		slice := make([]interface{}, 0)
-		for _, str := range strSlice {
-			slice = append(slice, str)
+		// TODO: handle slice better
```
disksing: can you explain how it is not good enough now?

server/config_manager/config_manager.go
```diff
@@ -573,12 +573,17 @@ func getUpdateValue(item, updateItem interface{}) (interface{}, error) {
...
+		if item, ok := updateItem.(string); ok {
+			strSlice := strings.Split(item, ",")
+			slice := make([]interface{}, 0)
```
disksing: `var slice []interface{}` or `make([]interface{}, 0, len(strSlice))`

server/config/config.go
rleungx: Yes.

server/config/config.go
```diff
@@ -991,6 +993,25 @@ func (c *Config) GetZapLogProperties() *log.ZapProperties {
...
+		return err
+	}
+	// 0644 represent the permission of file which is "-rw-r--r--" here.
+	return ioutil.WriteFile(filePath, buf.Bytes(), 0644)
```
overvenus: To avoid corrupted content during lost power, we should:
1. write the content to temp file
2. sync the temp file
3. rename the temp file to the config file.

server/config/config.go
```diff
@@ -1008,8 +1007,21 @@ func (c *Config) RewriteFile(new *Config) error {
 	if err := toml.NewEncoder(&buf).Encode(*new); err != nil {
 		return err
 	}
-	// 0644 represent the permission of file which is "-rw-r--r--" here.
-	return ioutil.WriteFile(filePath, buf.Bytes(), 0644)
+	dir := filepath.Dir(filePath)
+	tmpfile := filepath.Join(dir, "tmp_pd.toml")
+
+	f, err := os.Create(tmpfile)
```
overvenus: The default permission is 0666, should we change it to 0644?

rleungx: I think it's unnecessary?

tszCat:
will the comments in the old config file get persisted in the updated config file?

rleungx:
This feature has been removed, it won't rewrite the config file now

tszCat:
> This feature has been removed, it won't rewrite the config file now

Oh😱...
But I still wonder that is there any solutions to rewrite config file(TOML) while its comments get persisted too?

rleungx:
AFAIK, If you use BurntSushi/toml, there are issues that have some solution to do it, but they are not the official way, otherwise, you need to implement it by yourself.

tszCat:
> AFAIK, If you use BurntSushi/toml, there are issues that have some solution to do it, but they are not the official way, otherwise, you need to implement it by yourself.


got it, thx. 
I've looked up issuses on encoding comments on BurntSushi's repo, but unfortunately I didn't get solution I expected. Maybe I should implement it by myself.

PR #2162, opened 2020-02-27
*: add test for online reload new  TLS certificates

nolouch:

ref https://github.com/pingcap/tidb/issues/14666
If the key/cert expired, we need to reload it online.  the method same as [etcd](https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/security.md#notes-for-tls-authentication), it is already supported, this PR is to add a test.

add test verify it.

### Check List 

Tests 

 - Unit test


client/base_client.go
```diff
@@ -260,7 +260,7 @@ func (c *baseClient) getOrCreateGRPCConn(addr string) (*grpc.ClientConn, error)
 	tlsCfg, err := grpcutil.SecurityConfig{
 		CAPath:   c.security.CAPath,
 		CertPath: c.security.CertPath,
-		KeyPath:  c.security.CertPath,
+		KeyPath:  c.security.KeyPath,
```
disksing: 👀

tests/client/client_tsl_test.go
```diff
@@ -0,0 +1,246 @@
+// Copyright 2018 PingCAP, Inc.
```
disksing: 2020？

client/base_client.go
nolouch: That's a bug  :(

PR #2175, opened 2020-03-02
placement: add ruleList to accelerate rule search

### LINKED ISSUE #2163: Improve performance of finding placement rules of a region
current implementation needs to iterate all placement rules. It should be too slow when there are thouthands of rules.

Use some preprocessing to make it faster.

disksing:

Fix #2163

Pre-calculate rules for each range when any rule is updated.

### Check List 

Tests 
 - Unit test
 - Manual test (create 10000 rules then `curl pd/api/v1/config/rule/key/{key}`)

Related changes

### DIFF
--- a/server/schedule/placement/rule_list.go
+++ b/server/schedule/placement/rule_list.go
@@ -92,7 +92,7 @@ func buildRuleList(rules map[[2]string]*Rule) ruleList {
 		return bytes.Compare(points[i].key, points[j].key) < 0
 	})
 
-	// deterline rules for each range.
+	// determine rules for each range.
 	var rl ruleList
 	var sr sortedRules
 	for i, p := range points {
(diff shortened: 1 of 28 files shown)


server/schedule/placement/rule_list.go
```diff
@@ -0,0 +1,150 @@
...
+	startKey   []byte
+	rules      []*Rule
+	applyRules []*Rule
+}
+
+type ruleList struct {
+	ranges []rangeRules // ranges[i] contains rules apply to (ranges[i].startKey, ranges[i+1].startKey).
+}
+
+func buildRuleList(rules map[[2]string]*Rule) ruleList {
```
rleungx: How about directly calling this function as `updateRuleList`?

disksing: good idea

disksing:
ping /cc @rleungx

Issue #1832, opened 2019-10-23
Use cobra to unify the flags processing


shafreeck:
## PCP Issue

- difficult: easy
- score: 50
- min score required: 0
- Recommended Skills: Go

## Description

By now, `pd-ctl` uses `pflag` to process the flag for itself and uses `cobra` to process flags for its subcommands. In fact, `pd-ctl` itself is a command too. We could use `cobra` to unify all the processes.

Requirements:
* Remove `pflag` library
* Use cobra to process all flags and commands
* Distinct global and local flags
* Have meaningful usage message

Reference:
* github.com/spf13/pflag
* github.com/spf13/cobra

PR #2187, opened 2020-03-03
remove package pflag in pd-ctl

loxp:
UCP #1832 

### Check List 

Tests 

Code changes

Side effects

Related changes

### DIFF
--- a/tools/pd-ctl/pdctl/ctl.go
+++ b/tools/pd-ctl/pdctl/ctl.go
@@ -118,7 +118,7 @@ func getMainCmd(args []string) *cobra.Command {
 	return rootCmd
 }
 
-// Start main command
+// MainStart start main command
 func MainStart(args []string) {
 	rootCmd := getMainCmd(args)
 	if len(commandFlags.CAPath) != 0 {
@@ -133,7 +133,7 @@ func MainStart(args []string) {
 	}
 }
 
-// Start interact command
+// Start start interact command
 func Start(args []string) {
 	rootCmd := getRootCmd(args)
 	if len(commandFlags.CAPath) != 0 {
--- a/tools/pd-ctl/pdctl/ctl.go
+++ b/tools/pd-ctl/pdctl/ctl.go
@@ -94,7 +94,7 @@ func getBasicCmd() *cobra.Command {
 	return rootCmd
 }
 
-func getRootCmd(args []string) *cobra.Command {
+func getInteractCmd(args []string) *cobra.Command {
 	rootCmd := getBasicCmd()
 
 	rootCmd.SetArgs(args)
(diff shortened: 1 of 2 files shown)


tools/pd-ctl/pdctl/ctl.go
```diff
@@ -14,10 +14,14 @@
 package pdctl
 
 import (
-	"os"
-
+	"fmt"
```
nolouch: Suggest import as group ordering, see[ more](https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering).

tools/pd-ctl/pdctl/ctl.go
```diff
@@ -68,11 +87,55 @@ func Start(args []string) {
...
+
+	rootCmd.SetArgs(args)
 	rootCmd.ParseFlags(args)
 	rootCmd.SetOutput(os.Stdout)
 
+	return rootCmd
+}
+
+// MainStart start main command
+func MainStart(args []string) {
```
rleungx: Looks like these two functions are almost the same except the `rootCmd`, can we combine them?

tools/pd-ctl/pdctl/ctl.go
```diff
@@ -31,23 +35,38 @@ type CommandFlags struct {
...
+	}
+}
+
+func getBasicCmd() *cobra.Command {
 	rootCmd := &cobra.Command{
 		Use:   "pd-ctl",
 		Short: "Placement Driver control",
 	}
+
 	rootCmd.PersistentFlags().StringVarP(&commandFlags.URL, "pd", "u", "http://127.0.0.1:2379", "pd address")
```
rleungx: It's better to change all the first letters to the upper case.

loxp: Do you mean change all first letters of functions in ctl.go to upper case?

rleungx: oh, sorry for misleading. What I mean is the help message.

tools/pd-ctl/pdctl/ctl.go
loxp: I refactor the code, PTAL.

tools/pd-ctl/pdctl/ctl.go
loxp: In fact, only the MainStart() needs to export to package main.

loxp:
@rleungx BTW, I think the `detach` flag is unused in pd-ctl. Can we remove it?

tools/pd-ctl/pdctl/ctl.go
loxp: > It's better to change all the first letters to the upper case.

OK, change the first letter in message to upper case now.

nolouch:
@loxp   `detach` flag is reserved for compatibility, previously `pd-ctl` default mode was interactive.

Issue #2181, opened 2020-03-02
UCP: Use golangci-lint in the officially recommended way


HunDunDM:
## Description

PD uses [golangci-lint](https://github.com/golangci/golangci-lint) as a linter in the Makefile. Currently, golangci-lint is installed in retool, which is not an officially recommended installation method. We need to change to the [officially recommended way](https://github.com/golangci/golangci-lint#binary) and add the `.golangci.yml` configuration file.

## Score

* 50

## Mentor

* @nolouch 
* @hundundm 

## Recommended Skills

* Golang
* make

PR #2195, opened 2020-03-04
tools: use golangci-lint in the officially recommended way

Poytr1:
UCP #2181
Signed-off-by: Poytr1 <pooytr1@gmail.com>

Resolve: https://github.com/pingcap/pd/issues/2181
Currently, golangci-lint is installed in retool, which is not an officially recommended installation method. We need to change to the officially recommended way and add the .golangci.yml configuration file.

- Delete `golangci-lint` in the `retool-install.sh` as well as `tool.json`
- Add a new task in Makefile called `golangci-lint-setup` to download `golangci-lint` binary into *$GOPATH/bin/golangci-lint* as https://github.com/golangci/golangci-lint#install suggests.
- Add a `.golangci.yml` file which has same config with the command line option.
- Update the command to call `golangci-lint` using the new config file.

### Check List 

Tests 

Run `make check`, then the `golangci-lint` will be downloaded and ran with the config file `.golangci.yml`.

Code changes

 - Has configuration change

Side effects
N/A

Related changes

 - Need to update the documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ GOCHECKER := awk '{ print } END { if (NR > 0) { exit 1 } }'
 RETOOL := ./scripts/retool
 OVERALLS := overalls
 
+TOOL_BIN_PATH := $(shell pwd)/.tools/bin
+
+PATH := $(TOOL_BIN_PATH):$(PATH)
+
 FAILPOINT_ENABLE  := $$(find $$PWD/ -type d | grep -vE "(\.git|\.retools)" | xargs ./scripts/retool do failpoint-ctl enable)
 FAILPOINT_DISABLE := $$(find $$PWD/ -type d | grep -vE "(\.git|\.retools)" | xargs ./scripts/retool do failpoint-ctl disable)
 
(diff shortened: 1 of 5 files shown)


Poytr1:
Looks like in the ci environment, the `gopath` wasn't set, should I use a path inside the project like `.tools/` to store the binary?

nolouch:
@Poytr1 I think it's ok to use a path like `.tools/bin`

Makefile
```diff
@@ -106,7 +106,10 @@ retool-setup:
 	@which retool >/dev/null 2>&1 || go get github.com/twitchtv/retool
 	@./scripts/retool sync
 
-check: retool-setup check-all check-plugin
+golangci-lint-setup:
+	@which golangci-lint >/dev/null 2>&1 || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.7
```
HunDunDM: I think it can be placed in the `./bin` directory. This directory can also be used as `GOBIN` for tools.go

Poytr1: Done. This pr has conflicts with #2196, it would be better to merge #2196 firstly then I will resolve  these conflicts.

loxp: I think it's better to merge this PR first. If merge #2196 first, it must manage golangci-lint with tools.go, which import several dependencies, and have to remove them after your PR is merged.

PR #2220, opened 2020-03-10
api: fix register handler leak

rleungx:
After #2204. we enable dynamic config by default. When we use HTTP API, the `lazyHandler` will be called. The handler will call `RegisterConfigHandlerFromEndpoint` which creates a goroutine to handle the stop logic. In this case, once we call API multiple times, goroutines will be leaked.

This PR fixes this by using `sync.Once` to prevent calling `RegisterConfigHandlerFromEndpoint` multiple time.

### Check List 

Tests 

 - Manual test 

1. start a cluster
2. call API multiple times
3. check `http://127.0.0.1:2379/debug/pprof/goroutine?debug=1` for the number of goroutines

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -166,10 +166,14 @@ type lazyHandler struct {
 	engine  *negroni.Negroni
 }
 
+var once sync.Once
+
 func (lazy *lazyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	for _, f := range lazy.options {
-		f()
-	}
+	once.Do(func() {
+		for _, f := range lazy.options {
+			f()
+		}
+	})
 
 	lazy.engine.ServeHTTP(w, r)
 }


server/server.go
```diff
@@ -166,10 +166,14 @@ type lazyHandler struct {
 	engine  *negroni.Negroni
 }
 
+var once sync.Once
+
 func (lazy *lazyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	for _, f := range lazy.options {
-		f()
-	}
+	once.Do(func() {
```
HunDunDM: If multiple server instances are started during the test, they will only be executed once.

nolouch: cc @rleungx

disksing:
Need to cherry-pick to 3.1?

rleungx:
> Need to cherry-pick to 3.1?

@disksing 3.1 doesn't contain this feature. 

@nolouch And since we have two different ways to handle the etcd race problem. After discussing with @HunDunDM, he is working on unifying these two ways and fix this problem as well. Close this PR.

Issue #2168, opened 2020-02-28
UCP: hot region schedulr supports store weight 


nolouch:
## Description
Currently, PD's schedulers cannot handle the heterogeneous problem, we use <leader_weight>, <region_weight>  to identify the processing power of the machine. but the hot region scheduler cannot consider it very well, we need to let the hot region to aware `weight`.

## Score 

* 

## Recommended Skills
* Golang

mantuliu:
Sorry, I have been relatively busy recently, and I hope others can continue this optimization.

nolouch:
@mantuliu  thank you.

PR #2239, opened 2020-03-13
 *: Parameterize hotspot scheduling and adjust hot cache

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve? 
Improve stability and let hot region configurable.

### What is changed and how it works?

- fix the hot cache problem
- parameterize the schedule
- adjust the pendings

### Check List 

Tests 

 - Unit test
 - Manual test
   Command:  sysbench oltap_read_only (table size=2000000, threads=300)
   Before:
    image
  After:
   
  image

PR #2240, opened 2020-03-13
hot region schedule supports store weight

mantuliu:
UCP #2168 

Add a method for calculate the storeWeight
```
func calcStoreWeight(regionWeight, leaderWeight float64) float64 {
	storeWeight := (regionWeight + leaderWeight) / 2
	return math.Max(storeWeight, minWeight)
}
```
byteRate and KeyRate are recalculated using:
```
ByteRate: byteRate / storeWeight,
KeyRate:  keyRate / storeWeight,
```

### Check List 

Tests 

 - Unit test

- Manual test
Command: sysbench oltp_read_only (table size=10000000, threads=300)
before
1before
2before

after set store 1 hot-weight=1 store 2 hot-weight=1 store 7 hot-weight =2
1after
2after

Side effects

 - Increased code complexity


nolouch:
PTAL @rleungx @disksing

server/schedulers/hot_region.go
```diff
@@ -288,12 +297,17 @@ func summaryStoresLoad(
 			}
 		}
 
+		//calculate the storeWeight
+		storeWeight := calcStoreWeight(
+			cluster.GetStore(id).GetRegionWeight(),
```
nolouch: This weight may need to be separated from the cluster weight, because this is related to load, and the previous weight is more related to size. you can wait for https://github.com/pingcap/pd/pull/2239.

server/schedulers/hot_region.go
```diff
@@ -755,16 +773,24 @@ func (bs *balanceSolver) betterThan(old *solution) bool {
...
+			case bs.cur.srcPeerStat.GetKeyRate()/storeWeight > old.srcPeerStat.GetKeyRate()/oldStoreWeight:
 				return true
-			case bs.cur.srcPeerStat.GetKeyRate() < old.srcPeerStat.GetKeyRate():
+			case bs.cur.srcPeerStat.GetKeyRate()/storeWeight < old.srcPeerStat.GetKeyRate()/oldStoreWeight:
 				return false
 			}
 		} else {
-			byteRkCmp := rankCmp(bs.cur.srcPeerStat.GetByteRate(), old.srcPeerStat.GetByteRate(), stepRank(0, 100))
-			keyRkCmp := rankCmp(bs.cur.srcPeerStat.GetKeyRate(), old.srcPeerStat.GetKeyRate(), stepRank(0, 10))
+			byteRkCmp := rankCmp(bs.cur.srcPeerStat.GetByteRate()/storeWeight, old.srcPeerStat.GetByteRate()/oldStoreWeight, stepRank(0, 100))
```
nolouch: maybe do it in `stepRank`? `compareSrcStore` and `compareDstStore` also need it.

server/schedulers/hot_region.go
mantuliu: OK.I will do more test and wait for #2239.

server/schedulers/hot_region.go
mantuliu: I didn't implement this in the stepRank method, because stepRank might become something like the following:
```
func stepRank(rk0 float64, step float64) func(float64) int64 {
   return func(rate storeWeight float64) int64 {
      return int64((rate/storeWeight - rk0) / step)
   }
}
```
This makes all the methods related to stepRank need to be changed,such as stLdByteRate,stLdKeyRate,sliceLoadCmp,sliceLPCmp. I personally think that such changes will increase the complexity of these functions. But I still tried to optimize this logic and added a getRateWithStoreWeight function.

what do you think?^_^

mantuliu:
> Thanks for your contribution, I will take the time to take a look and test.

Thank you,I added a new parameter(hot-weight) to measure the ability of the store to handle hot regions. I will test it in an actual cluster.

nolouch:
Hi, @mantuliu  thanks for your contribution.
We invite you to join our [sig-scheduling](https://github.com/pingcap/community/tree/master/special-interest-groups/sig-scheduling), you can communicate with us more conveniently in [`slack`](https://pingcap.com/tidbslack/) channel.

mantuliu:
@nolouch Thank you. I have joined.

server/schedulers/hot_test.go
```diff
@@ -579,6 +634,109 @@ func (s *testHotReadRegionSchedulerSuite) TestByteRateOnly(c *C) {
...
+	tc.AddRegionStoreWithWeight(2, 2, 2.0)
+	tc.AddRegionStoreWithWeight(3, 2, 2.0)
+	tc.AddRegionStoreWithWeight(4, 2, 0.5)
+	tc.AddRegionStoreWithWeight(5, 0, 1.0)
+
+	//| store_id | read_bytes_rate |
+	//|----------|-----------------|
+	//|    1     |     7.5MB       |
+	//|    2     |     9.8MB       |
+	//|    3     |       9MB      |
```
rleungx:
```suggestion
	//|    3     |       9MB       |
```

nolouch:
Hi @mantuliu . 
Recently, we have adjusted some hotspot scheduling strategies, which may be a little different from your previous version. Would you resolve the conflict firstly？thanks :)

mantuliu:
> Hi @mantuliu .
> Recently, we have adjusted some hotspot scheduling strategies, which may be a little different from your previous version. Would you resolve the conflict firstly？thanks :)

OK,I will resolve the conflict and make a test.

PR #2250, opened 2020-03-17
*: add a new method `GetAll` to the config manager

qw4990:
Wait for https://github.com/pingcap/kvproto/pull/579.
Add a new method `GetAll` to the config manager to let the `TiDB` get all configs from `PD`.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/pingcap/check v0.0.0-20191216031241-8a5a85928f12
 	github.com/pingcap/errcode v0.0.0-20180921232412-a1a7271709d9
 	github.com/pingcap/failpoint v0.0.0-20191029060244-12f4ac2fd11d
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -62,3 +62,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
+
+replace github.com/pingcap/kvproto => /Users/zhangyuanjia/Workspace/go/src/github.com/pingcap/kvproto
```
qw4990: This will be removed after merging https://github.com/pingcap/kvproto/pull/579.

rleungx: You need to replace it with your own repo, otherwise, the CI can not pass.

PR #2251, opened 2020-03-17
config: let the cert-allowed-cn same with other components

nolouch:

TiDB, TiKV Supports multiple CNs, but we depend on etcd, let the config be consisted firstly.

changed config items.

### Check List 

Tests 

 - Manual test

### DIFF
--- a/pkg/grpcutil/grpcutil.go
+++ b/pkg/grpcutil/grpcutil.go
@@ -33,7 +33,8 @@ type SecurityConfig struct {
 	// KeyPath is the path of file that contains X509 key in PEM format.
 	KeyPath        string `toml:"key-path" json:"key-path"`
 	ClientCertAuth bool   `toml:"client-cert-auth" json:"client-cert-auth"`
-	CertAllowedCN  string `toml:"cert-allowed-cn" json:"cert-allowed-cn"`
+	// CertAllowedCN is a CN which must be provided by a client
+	CertAllowedCN []string `toml:"cert-allowed-cn" json:"cert-allowed-cn"`
 }
 
 // ToTLSConfig generatres tls config.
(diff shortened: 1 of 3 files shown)


pkg/grpcutil/grpcutil.go
```diff
@@ -55,6 +60,20 @@ func (s SecurityConfig) ToTLSConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
+// GetOneAllowedCN only gets the first one CN.
+func (s SecurityConfig) GetOneAllowedCN() (string, error) {
+	allowedCN := ""
+	i := len(s.CertAllowedCN)
+	switch {
+	case i > 1:
+		return "", errors.New("Currently only supports one CN")
```
rleungx: Do we need to return one and give the error?

nolouch: I think no need.

pkg/grpcutil/grpcutil.go
```diff
@@ -55,6 +60,20 @@ func (s SecurityConfig) ToTLSConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
+// GetOneAllowedCN only gets the first one CN.
+func (s SecurityConfig) GetOneAllowedCN() (string, error) {
+	allowedCN := ""
```
HunDunDM:
```go
	switch len(s.CertAllowedCN) {
	case 1:
		return s.CertAllowedCN[0], nil
	case 0:
		return "", nil
	default:
		return "", errors.New("Currently only supports one CN")
	}
```
Would it be more readable?

nolouch:
will break the compatibility between https://github.com/pingcap/pd/pull/2241 and  https://github.com/pingcap/pd/pull/2251.

PR #2261, opened 2020-03-20
dashboard: control dashboard running node

rleungx:
We start dashboard services in all PDs now and actually we only need one. It can reduce the waste of resources.

This PR adds a mechanism to control the dashboard running node by using `dashbaord-address` parameters. It has three kinds of values:
- `auto`: select the sever whose member ID is the minimum to run dashboard.
- `none`: stop all dashboard services
- specified address like `127.0.0.1:2379`: select the given address to run dashboard.

### Check List 

Tests 

 - Unit test


pkg/dashboard/manager.go
```diff
@@ -0,0 +1,230 @@
...
+		m.stopService()
+	}
+
+	return known
+}
+
+func (m *Manager) isOnline(addr string) bool {
+	rc := m.srv.GetRaftCluster()
+	if rc == nil || !rc.IsRunning() {
+		return true
```
breeswish: why it is online in this case?

pkg/dashboard/manager.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	members, err := cluster.GetMembers(m.srv.GetClient())
+	if err != nil || len(members) == 0 {
+		log.Error("failed to get members")
+		return
+	}
+
+	// get new dashboard address
+	var addr string
+	switch len(members) {
```
breeswish: in what case there will be 0 members?

pkg/dashboard/manager.go
```diff
@@ -0,0 +1,230 @@
...
+	case 0:
+		return
+	case 1:
+		addr = members[0].GetClientUrls()[0]
+	default:
+		addr = members[0].GetClientUrls()[0]
+		leaderID := m.srv.GetMemberInfo().MemberId
+		sort.Slice(members, func(i, j int) bool { return members[i].GetMemberId() < members[j].GetMemberId() })
```
breeswish: why do we need to sort the member id?

pkg/dashboard/manager.go
```diff
@@ -0,0 +1,230 @@
...
+				addr = member.GetClientUrls()[0]
+				break
+			}
+		}
+	}
+
+	// set new dashboard address
+	rc.SetDashboardAddress(addr)
+	if m.srv.GetConfig().EnableDynamicConfig {
+		if err := m.srv.UpdateConfigManager("pd-server.dashboard-address", addr); err != nil {
```
breeswish: should it be a global config?

pkg/dashboard/servers.go
```diff
@@ -44,28 +42,19 @@ var (
...
-			})
-			srv.AddCloseCallback(func() {
-				if err := s.Stop(context.Background()); err != nil {
-					log.Error("Stop dashboard server error", zap.Error(err))
-				} else {
-					log.Info("Dashboard server is stopped")
-				}
-			})
+			m := NewManager(srv, s, redirector)
+			srv.AddStartCallback(m.start)
```
breeswish: is it possible that start can be called multiple times? @HunDunDM

HunDunDM: This life cycle is consistent with PD Server. I think it cannot be called multiple times.

pkg/dashboard/manager.go
HunDunDM: @rleungx Is it possible to change this configuration at runtime?

pkg/dashboard/manager.go
HunDunDM: We want to select the follower node with the smallest member id.

pkg/dashboard/manager.go
HunDunDM: At this time, it is impossible to determine whether it is online, so it cannot be handled as offline.

pkg/dashboard/manager.go
HunDunDM: No change during runtime

pkg/dashboard/manager.go
HunDunDM: It cannot be 0 here, I have deleted it.

pkg/dashboard/manager.go
```diff
@@ -0,0 +1,215 @@
...
+	switch len(members) {
+	case 1:
+		addr = members[0].GetClientUrls()[0]
+	default:
+		addr = members[0].GetClientUrls()[0]
+		leaderID := m.srv.GetMemberInfo().MemberId
+		sort.Slice(members, func(i, j int) bool { return members[i].GetMemberId() < members[j].GetMemberId() })
+		for _, member := range members {
+			if member.MemberId != leaderID {
+				addr = member.GetClientUrls()[0]
```
nolouch: what if the join a new member or leader changed?

HunDunDM: Nothing happens unless the member that is the dashboardAddress is deleted.

Issue #2229, opened 2020-03-11
UCP: show all component ID when using ids


rleungx:
## Description

We can use `component ids <component>` to show all IDs of a kind of component. e.g.
```
>> component ids tikv
[
    "127.0.0.1:20160",
    "127.0.0.1:20161",
]
```
If we don't give a parameter, we should show IDs of all kinds of components. e.g.
```
>> component ids
pd:
...
tikv:
...
```

## Score

- 50

## Recommended Skills

- Golang

PR #2267, opened 2020-03-22
show all component ID when using ids

wq1019:
UCP: #2229

Signed-off-by: wq1019 <2013855675@qq.com>

@rleungx 

If IDs parameter is empty, all component IDs will be returned.

1. When the IDs parameter is ignored, CTL will automatically assemble the URL into`/component/ids/all`and continue to request.
2. Updated `/component/ids/{component}` API, When the parameter passed in is`all`, all component IDS will be returned.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/tools/pd-ctl/pdctl/command/component_command.go
+++ b/tools/pd-ctl/pdctl/command/component_command.go
@@ -129,7 +129,7 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		if argsLen > 0 {
 			cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
 		} else {
-			cmd.Println("Failed to get all components: %s", err)
+			cmd.Println("Failed to get all components", err)
 		}
 		return
 	}
--- a/tools/pd-ctl/pdctl/command/component_command.go
+++ b/tools/pd-ctl/pdctl/command/component_command.go
@@ -129,7 +129,7 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		if argsLen > 0 {
 			cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
 		} else {
-			cmd.Println("Failed to get all components", err)
+			cmd.Printf("Failed to get all components: %s\n", err)
 		}
 		return
 	}
--- a/tools/pd-ctl/pdctl/command/component_command.go
+++ b/tools/pd-ctl/pdctl/command/component_command.go
@@ -129,7 +129,7 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		if argsLen > 0 {
 			cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
 		} else {
-			cmd.Printf("Failed to get all components: %s\n", err)
+			cmd.Println("Failed to get all components", err)
 		}
 		return
 	}
--- a/tools/pd-ctl/pdctl/command/component_command.go
+++ b/tools/pd-ctl/pdctl/command/component_command.go
@@ -75,7 +75,7 @@ func NewSetComponentConfigCommand() *cobra.Command {
 func NewGetComponentIDCommand() *cobra.Command {
 	sc := &cobra.Command{
 		Use:   "ids [component]",
-		Short: "get all component IDs or given component names (e.g. tikv)",
+		Short: "get component IDs for all components or with a given component name (e.g. tikv)",
 		Run:   getComponentIDCommandFunc,
 	}
 	return sc
@@ -118,9 +118,7 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		cmd.Usage()
 		return
 	}
-	var (
-		ids = "all"
-	)
+	ids := "all"
 	if argsLen == 1 {
 		ids = args[0]
 	}
@@ -131,7 +129,7 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		if argsLen > 0 {
 			cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
 		} else {
-			cmd.Printf("Failed to get all components: %s\n", err)
+			cmd.Println("Failed to get all components: %s", err)
 		}
 		return
 	}
--- a/tools/pd-ctl/pdctl/command/component_command.go
+++ b/tools/pd-ctl/pdctl/command/component_command.go
@@ -119,18 +119,16 @@ func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
 		return
 	}
 	var (
-		hasIdsVal = false
-		ids       = "all"
+		ids = "all"
 	)
 	if argsLen == 1 {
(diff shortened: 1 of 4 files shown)


tools/pd-ctl/pdctl/command/component_command.go
```diff
@@ -112,15 +112,29 @@ func deleteComponentConfigCommandFunc(cmd *cobra.Command, args []string) {
...
-	if len(args) != 1 {
+	argsLen := len(args)
+	// args too long
+	if argsLen > 1 {
 		cmd.Usage()
 		return
 	}
+	var (
+		hasIdsVal = false
+		ids       = "all"
```
rleungx: How about printing all ids when there is no argument?

wq1019: In `server/api/router.go`
```suggestion
				if varName == "all" {
			rd.JSON(w, http.StatusOK, svr.GetConfigManager().GetAllComponentIDs())
			} else {
				rd.JSON(w, http.StatusOK, svr.GetConfigManager().GetComponentIDs(varName))
			}
```

rleungx: What I mean is that we may not need to use "all" here. If `vars["component"]` is empty, we can directly call `GetAllComponentIDs`.

wq1019: Thank you for reminding me. I'll change it later.

tools/pd-ctl/pdctl/command/component_command.go
```diff
@@ -112,15 +112,29 @@ func deleteComponentConfigCommandFunc(cmd *cobra.Command, args []string) {
...
+		hasIdsVal = true
+		ids = args[0]
+	}
+	prefix := path.Join(componentConfigPrefix, "ids", ids)
 
-	prefix := path.Join(componentConfigPrefix, "ids", args[0])
 	r, err := doRequest(cmd, prefix, http.MethodGet)
 	if err != nil {
-		cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
+		if hasIdsVal {
```
rleungx: Ditto, we can directly use the length of args to do it.

tools/pd-ctl/pdctl/command/component_command.go
wq1019: If don't set a flag, other route captures will occur, such as:
```suggestion
	componentRouter.Handle("/{component_id}", gwmux).Methods("GET")
```
Do you have any good ideas?

rleungx: `ids` subcommand will be captured by `/component/ids/{component}` rather than `/{component_id}`

wq1019: However, after local debugging, I found that when my`RequestURI` = `/pd/api/v1/component/ids `, it is actually handled by handleComponentGet function.

rleungx: Oh, sorry. Just keep it.

tools/pd-ctl/pdctl/command/component_command.go
wq1019: Is there anything else I can do for this pr? 📲

rleungx: This comment has not solved yet?

wq1019: Sorry, it has been updated.

tools/pd-ctl/pdctl/command/component_command.go
```diff
@@ -112,15 +112,27 @@ func deleteComponentConfigCommandFunc(cmd *cobra.Command, args []string) {
...
+		ids = args[0]
+	}
+	prefix := path.Join(componentConfigPrefix, "ids", ids)
 
-	prefix := path.Join(componentConfigPrefix, "ids", args[0])
 	r, err := doRequest(cmd, prefix, http.MethodGet)
 	if err != nil {
-		cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
+		if argsLen > 0 {
+			cmd.Printf("Failed to get component %s's id: %s\n", args[0], err)
```
rleungx: Oh, you can change to `cmd.Println`.

tools/pd-ctl/pdctl/command/component_command.go
```diff
@@ -112,15 +112,27 @@ func deleteComponentConfigCommandFunc(cmd *cobra.Command, args []string) {
...
 
 func getComponentIDCommandFunc(cmd *cobra.Command, args []string) {
-	if len(args) != 1 {
+	argsLen := len(args)
+	// args too long
+	if argsLen > 1 {
 		cmd.Usage()
 		return
 	}
+	var (
```
rleungx: ids := "all"?

tools/pd-ctl/pdctl/command/component_command.go
```diff
@@ -74,8 +74,8 @@ func NewSetComponentConfigCommand() *cobra.Command {
 // NewGetComponentIDCommand returns a id subcommand of componentCmd.
 func NewGetComponentIDCommand() *cobra.Command {
 	sc := &cobra.Command{
-		Use:   "ids <component>",
-		Short: "get all component IDs with a given component (e.g. tikv)",
+		Use:   "ids [component]",
+		Short: "get all component IDs or given component names (e.g. tikv)",
```
rleungx: How about
```suggestion
		Short: "get component IDs for all components or with a given component name (e.g. tikv)",
```

tools/pd-ctl/pdctl/command/component_command.go
wq1019: `cmd.Println` cannot be used because 2 parameters need to be formatted.

tools/pd-ctl/pdctl/command/component_command.go
wq1019: good idea. modified.

tools/pd-ctl/pdctl/command/component_command.go
wq1019: modified.

tools/pd-ctl/pdctl/command/component_command.go
rleungx: Yes. Line 132 can be changed.

wq1019:
@rleungx Help me. 
I only modified the `cmd` usage message. However, but `Jenkins` always fails to build. The reason for the failure was: 
```
github.com/pingcap/pd/v4/tests/server/tso
```
but I did not modify the code of this module.

rleungx:
> @rleungx Help me.
> I only modified the `cmd` usage message. However, but `Jenkins` always fails to build. The reason for the failure was:
> 
> ```
> github.com/pingcap/pd/v4/tests/server/tso
> ```
> 
> but I did not modify the code of this module.

There is a known problem and it's not related to this PR

tools/pd-ctl/pdctl/command/component_command.go
wq1019: modified.

PR #2277, opened 2020-03-23
config: add replicate-mode configuration

disksing:
Fix: https://github.com/pingcap/pd/issues/2269

Add config structure and adjust function.

### Check List 

Tests 
 - Unit test

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -213,6 +213,9 @@ const (
 	defaultDisableErrorVerbose = true
 
 	defaultEnableDynamicConfig = true
+
+	defaultDRWaitStoreTimeout = time.Minute
+	defaultDRWaitSyncTimeout  = time.Minute
 )
 
 var (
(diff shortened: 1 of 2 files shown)


server/config/config.go
```diff
@@ -1115,3 +1120,36 @@ func (c DashboardConfig) ToTiDBTLSConfig() (*tls.Config, error) {
...
+	}
+	c.DRAutoSync.adjust(meta.Child("dr-autosync"))
+}
+
+// DRAutoSyncReplicateConfig is the configuration for auto sync mode between 2 data centers.
+type DRAutoSyncReplicateConfig struct {
+	LabelKey         string            `toml:"label-key" json:"label-key"`
+	Primary          string            `toml:"primary" json:"primary"`
+	DR               string            `toml:"dr" json:"dr"`
+	MasterReplicas   int               `toml:"primary-replicas" json:"primary-replicas"`
```
nolouch:
```suggestion
	PrimaryReplicas   int               `toml:"primary-replicas" json:"primary-replicas"`
```

server/config/config.go
```diff
@@ -1115,3 +1120,36 @@ func (c DashboardConfig) ToTiDBTLSConfig() (*tls.Config, error) {
...
+	DRReplicas       int               `toml:"dr-replicas" json:"dr-replicas"`
+	WaitStoreTimeout typeutil.Duration `toml:"wait-store-timeout" json:"wait-store-timeout"`
+	WaitSyncTimeout  typeutil.Duration `toml:"wait-sync-timeout" json:"wait-sync-timeout"`
+}
+
+func (c *DRAutoSyncReplicateConfig) adjust(meta *configMetaData) {
+	if !meta.IsDefined("wait-store-timeout") {
+		c.WaitStoreTimeout = typeutil.Duration{Duration: time.Minute}
```
HunDunDM: Need to define constants like other functions?

server/config/config.go
disksing: ok

server/config/config.go
disksing: ok

PR #2282, opened 2020-03-26
config: fix cannot add new config item

rleungx:
After a component is registered, if there are some new configs when we upgrading, it cannot be registered anymore. When we updating these configs, it will print out "cannot find the config item".

This PR fixes it by trying to update config item for each `Create` operation.

### Release note 
Fix the problem that cannot add new config item in upgrading when enabling dynamic config

### Check List 

Tests 

 - Unit test


yiwu-arbug:
oops, I was wanting to click rebase button, but merged the PR by mistake. Hope it is fine..

PR #2286, opened 2020-03-27
statistic: add moving average for stable hot 

lhy1024:
It will denoise and make statistics more stable. Below is a contrast between before and after.

image

We use average overtime to solve the problem that interval is different.
And add another moving average to filter out high-frequency noise.

### Release note 

Add moving average to make flow statistics more stable.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/statistics/avg_over_time.go
+++ b/server/statistics/avg_over_time.go
@@ -116,9 +116,5 @@ func (t *TimeMedian) Add(delta float64, interval time.Duration) {
 
 // Set sets the given average.
 func (t *TimeMedian) Set(avg float64) {
-	t.aot.Set(avg)
-	if t.aot.intervalSum >= t.aotInterval {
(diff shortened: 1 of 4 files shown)


server/statistics/avg_over_time.go
```diff
@@ -48,38 +53,72 @@ func NewAvgOverTime(interval time.Duration) *AvgOverTime {
...
+	}
+	t.aot.Add(delta, interval)
+	if t.aot.intervalSum >= t.aotInterval {
+		t.mf.Add(t.aot.Get())
+		t.aot.Clear()
+	}
+}
+
+// Set sets the given average.
+func (t *TimeMedian) Set(avg float64) {
```
nolouch: Set `medianFilter` directly? Is it only for the test?

nolouch:
@lhy1024  ci failed.

PR #2291, opened 2020-03-27
store: use fixed size and space ratio to decide space threshold

rleungx:
We use `low-space-ratio` and `high-space-ratio` to control the calculation of score. But in some cases which the capacity of the store is large enough. These ratios may not reasonable.

This PR adds a fixed size for low and high space threshold to help choose a better value compared with the original way.

### Release note 
Use fixed size and space ratio to decide the space threshold.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -27,10 +27,12 @@ import (
 	"go.uber.org/zap"
 )
 
-// Interval to save store meta (including heartbeat ts) to etcd.
-const storePersistInterval = 5 * time.Minute
-const lowSpaceThreshold = 100 * (1 << 10)  // 100 GB
-const highSpaceThreshold = 300 * (1 << 10) // 300 GB
+const (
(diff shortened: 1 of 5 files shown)


server/core/store.go
```diff
@@ -319,17 +322,21 @@ func (s *StoreInfo) StorageSize() uint64 {
...
+		min = space
 	}
-	return float64(s.GetAvailable()) / float64(s.GetCapacity())
+	return min
 }
 
 // IsLowSpace checks if the store is lack of space.
 func (s *StoreInfo) IsLowSpace(lowSpaceRatio float64) bool {
-	return s.GetStoreStats() != nil && s.AvailableRatio() < 1-lowSpaceRatio
+	available := float64(s.GetAvailable()) / (1 << 20)
```
lhy1024: how about replacing `1<<20` with `MB`

PR #2298, opened 2020-03-28
schedulers/hot: let leader can be schedule even if the peer is not balanced

nolouch:
…lanced

In this scenario: 9 tikv, running with a workload  `oltp_insert` (6 tables)
- the count is not suitable for calculating limit because there are many hot regions (index) and only 6 very hot regions (row data)
- there are too many hot regions, so the hot leader may have no chance to be scheduled.
image
image

- remove the auto limit
- randomly schedule peer or leader.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -76,6 +76,11 @@ const (
 	maxHotScheduleInterval = 20 * time.Second
 )
 
+var (
+	randomUpperBound = 3
+	randomLowBound   = 0
+)
+
 type hotScheduler struct {
 	name string
 	*BaseScheduler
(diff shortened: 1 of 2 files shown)


server/schedulers/hot_region.go
```diff
@@ -400,14 +404,19 @@ func (h *hotScheduler) balanceHotReadRegions(cluster opt.Cluster) []*operator.Op
 
 func (h *hotScheduler) balanceHotWriteRegions(cluster opt.Cluster) []*operator.Operator {
 	// prefer to balance by peer
-	peerSolver := newBalanceSolver(h, cluster, write, movePeer)
-	ops := peerSolver.solve()
-	if len(ops) > 0 {
-		return ops
+	seed := h.r.Intn(randomUpperBound-randomLowBound) + randomLowBound
```
rleungx: Why we need these two bound since it cannot be changed?

server/schedulers/hot_region.go
```diff
@@ -153,12 +157,12 @@ func (h *hotScheduler) IsScheduleAllowed(cluster opt.Cluster) bool {
 }
 
 func (h *hotScheduler) allowBalanceLeader(cluster opt.Cluster) bool {
-	return h.OpController.OperatorCount(operator.OpHotRegion) < minUint64(h.leaderLimit, cluster.GetHotRegionScheduleLimit()) &&
+	return h.OpController.OperatorCount(operator.OpHotRegion) < cluster.GetHotRegionScheduleLimit() &&
```
lhy1024: why to remove limit?

nolouch: The `count` is not suitable for calculating the limit for now and we already have a pending influence.

PR #2300, opened 2020-03-29
Addressing latency issue for pd-clients

mgkanani:
https://github.com/pingcap/pd/issues/1847

Avoided slice re-allocation and copying cost
changed from histogram to summary for distribution information

### Check List 

Tests 

Code changes
- No functionality change, improvement in performance
- For benchmark, added Prometheus code to print metrics data at the end of a run
 
Side effects

- In case of many clients created within single Tidb instance can lead to slightly higher memory consumption

Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -114,7 +114,7 @@ const (
 	defaultPDTimeout      = 3 * time.Second
 	dialTimeout           = 3 * time.Second
 	updateLeaderTimeout   = time.Second // Use a shorter timeout to recover faster from network isolation.
-	maxMergeTSORequests   = 10000
+	maxMergeTSORequests   = 10000       // should be higher if client is sending requests in burst
 	maxInitClusterRetries = 100
 )
 
@@ -197,7 +197,9 @@ func (c *client) tsLoop() {
 	loopCtx, loopCancel := context.WithCancel(c.ctx)
 	defer loopCancel()
 
-	var requests []*tsoRequest
+	defaultSize := maxMergeTSORequests + 1
+	requests := make([]*tsoRequest, defaultSize)
+
 	var opts []opentracing.StartSpanOption
 	var stream pdpb.PD_TsoClient
 	var cancel context.CancelFunc
@@ -231,10 +233,10 @@ func (c *client) tsLoop() {
 
 		select {
 		case first := <-c.tsoRequests:
-			requests = append(requests, first)
-			pending := len(c.tsoRequests)
-			for i := 0; i < pending; i++ {
-				requests = append(requests, <-c.tsoRequests)
+			pendingPlus1 := len(c.tsoRequests) + 1
+			requests[0] = first
+			for i := 1; i < pendingPlus1; i++ {
+				requests[i] = <-c.tsoRequests
 			}
 			done := make(chan struct{})
 			dl := deadline{
@@ -248,10 +250,9 @@ func (c *client) tsLoop() {
 				cancel()
 				return
 			}
-			opts = extractSpanReference(requests, opts[:0])
-			err = c.processTSORequests(stream, requests, opts)
+			opts = extractSpanReference(requests[:pendingPlus1], opts[:0])
+			err = c.processTSORequests(stream, requests[:pendingPlus1], opts)
 			close(done)
-			requests = requests[:0]
 		case <-loopCtx.Done():
 			cancel()
 			return
@@ -377,7 +378,9 @@ func (c *client) leaderClient() pdpb.PDClient {
 var tsoReqPool = sync.Pool{
 	New: func() interface{} {
 		return &tsoRequest{
-			done: make(chan error, 1),
+			done:     make(chan error, 1),
+			physical: 0,
+			logical:  0,
 		}
 	},
 }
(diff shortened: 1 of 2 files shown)


nolouch:
Thanks for your contribution. :)
I have a real environment and reproduced it, could you test in it with your branch?
Let me contact you on slack.

client/client.go
```diff
@@ -380,10 +383,10 @@ func (c *client) GetTSAsync(ctx context.Context) TSFuture {
 		ctx = opentracing.ContextWithSpan(ctx, span)
 	}
 	req := tsoReqPool.Get().(*tsoRequest)
-	req.start = time.Now()
+	//req.start = time.Now()
```
shafreeck, line 386: Mistake comment?

client/client.go
```diff
@@ -380,10 +383,10 @@ func (c *client) GetTSAsync(ctx context.Context) TSFuture {
...
 	req.ctx = ctx
-	req.physical = 0
-	req.logical = 0
+	//req.physical = 0
+	//req.logical = 0
```
shafreeck, line 389: Remove the unused lines

mgkanani: Addressed this.

mgkanani:
Performance Comparison(may vary system to system):
For both request count/sec = 0.9M +- 10%
@nolouch , thank you for the environment.

Master Branch:
`
Total:
count:105642065, max:76, min:0, >1ms:0, >2ms:0, >5ms:9420055 (8.917%), >10ms:96057544 (90.927%), >30ms:164413 (0.156%)
`

With this PR's changes:
`
Total:
count:104146574, max:69, min:0, >1ms:2160 (0.002%), >2ms:32434 (0.031%), >5ms:8137420 (7.813%), >10ms:95903141 (92.085%), >30ms:71371 (0.069%)
`

nolouch:
Hi @mgkanani , the environment running a TiDB cluster with a heavy workload, does your PR benefit for that situation?

mgkanani:
Hi @nolouch ,@tiancaiamao,

There is substantial difference between benchmarks `pd-tso-bench` and `sysbench(with tidb-pd-tikv cluster)`:

- With pd-tso-bench: number of requests to number of clients ratio is higher. So batching size plays a role in it.  E.g. we used 10k clients and number of requests per second was ~1M. So per client, on avg ~100(=1M/10k) batching possible. 
- Same changes with sysbench: 2000 clients generates ~2k request so lesser chance for batching. With this experiment, batch size hardly goes beyond 256 and avg is 2-4. I tried profiling and client.tsLoop function executed for ~0.05%.

If sysbench is near real time then there is no point of setting `maxMergeTSORequests` way higher ~10k.
If possible, `pd_client_request_handle_tso_batch_size` metrics data from real-server with high load may help in setting the value. Based on the data, I will make relevant changes to [maxMergeTSORequests size](https://github.com/pingcap/pd/pull/2300/files#diff-1cd41a2611da49ce63630d5a51bd82c5R105) and [default Slice size](https://github.com/pingcap/pd/pull/2300/files#diff-1cd41a2611da49ce63630d5a51bd82c5R192)

mgkanani:
> Hi @mgkanani , the environment running a TiDB cluster with a heavy workload, does your PR benefit for that situation?

Yes, if there are many pending requests which lead to higher the batch-size.

client/client.go
```diff
@@ -240,10 +247,8 @@ func (c *client) tsLoop() {
 				cancel()
 				return
 			}
-			opts = extractSpanReference(requests, opts[:0])
-			err = c.processTSORequests(stream, requests, opts)
+			err = c.processTSORequests(stream, requests[:pendingPlus1], opts)
```
mgkanani, line 250: I made a mistake here.. should be `opts[:pendingPlus1]`. Will revert fix it.

client/client.go
```diff
@@ -264,15 +269,6 @@ func (c *client) tsLoop() {
 	}
 }
 
-func extractSpanReference(requests []*tsoRequest, opts []opentracing.StartSpanOption) []opentracing.StartSpanOption {
-	for _, req := range requests {
-		if span := opentracing.SpanFromContext(req.ctx); span != nil {
-			opts = append(opts, opentracing.ChildOf(span.Context()))
```
mgkanani, line 270: Can there be scenario where for some requests span is nil?
In those scenario, current code may not work hence need to update it.

client/client.go
```diff
@@ -231,10 +233,10 @@ func (c *client) tsLoop() {
 
 		select {
 		case first := <-c.tsoRequests:
-			requests = append(requests, first)
-			pending := len(c.tsoRequests)
-			for i := 0; i < pending; i++ {
-				requests = append(requests, <-c.tsoRequests)
+			pendingPlus1 := len(c.tsoRequests) + 1
```
disksing, line 236: `pendingPlus1` can be greater than `defaultSize`.

mgkanani: max value for `len(c.tsoRequests)` can be `maxMergeTSORequests`.
Below for loop will be iterating `pendingPlus1 - 1` times as starting index is 1 instead of 0.
Max index value for `i` can be `pendingPlus1 - 1` which can be maximum `maxMergeTSORequests`.
`defaultSize` is set to `maxMergeTSORequests + 1` so it will allow array index to be `maxMergeTSORequests`.
So, `pendingPlus1` will never be greater than `defaultSize`.
https://play.golang.org/p/1kEQsGbFdnN will help in testing above scenario.

disksing: cool, I think I didn't realize the channel is a buffered channel. 😅

disksing:
Please take another look @nolouch

PR #2297, opened 2020-03-28
scheduelrs: add load expectations for hot scheudler

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve? 
To prevent excessive redundant scheduling
image        
### What is changed and how it works?
add the avg load as an expectation.

### Check List 

Tests 
 - Unit test
 - Manual test
the scheduler operator is reduced and QPS is stable.
image

PR #2308, opened 2020-04-01
schedulers: add src and dst tolerance ratio for hot scheduler

lhy1024:

After #2297, we introduce hot tolerance. After some tests, we found there need to be more accurate config

This PR split tolerance to two configs, src and dst.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/schedulers/hot_region_config.go
+++ b/server/schedulers/hot_region_config.go
@@ -42,8 +42,8 @@ func initHotRegionScheduleConfig() *hotRegionSchedulerConfig {
 		GreatDecRatio:         0.95,
 		MinorDecRatio:         0.99,
 		MaxPeerNum:            1000,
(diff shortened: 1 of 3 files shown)


server/schedulers/hot_region_config.go
```diff
@@ -42,7 +42,8 @@ func initHotRegionScheduleConfig() *hotRegionSchedulerConfig {
 		GreatDecRatio:         0.95,
 		MinorDecRatio:         0.99,
 		MaxPeerNum:            1000,
-		ToleranceRatio:        1.02, // Tolerate 2% difference
+		SrcToleranceRatio:     1.05, // Tolerate 2% difference
+		DstToleranceRatio:     1.00, // Tolerate 2% difference
```
nolouch, line 46: comments

PR #2322, opened 2020-04-02
replicate: implement recover progress

### LINKED ISSUE #2273: Synchronous replication: recover progress
In order to switch from `sync-recover` to `sync` mode. PD needs to make sure all region's states become `integrity-over-label` and show the progress in real time.

disksing:
Fix #2273

- save region replicate status to cache
- calculate recover progress based on region status (a simple approach, need improve, see: https://github.com/pingcap/pd/issues/2327)
- update test

### Check List 

Tests 
 - Unit test

Code changes
 - Has persistent data change

### DIFF
--- a/server/replicate/replicate.go
+++ b/server/replicate/replicate.go
@@ -172,6 +172,7 @@ const (
 
 // Run starts the background job.
 func (m *ModeManager) Run(quit chan struct{}) {
+	// Wait for a while when just start, in case tikv do not connect in time.
 	select {
 	case <-time.After(idleTimeout):
 	case <-quit:
--- a/server/replicate/replicate.go
+++ b/server/replicate/replicate.go
@@ -19,7 +19,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pingcap/kvproto/pkg/replicate_mode"
 	pb "github.com/pingcap/kvproto/pkg/replicate_mode"
 	"github.com/pingcap/log"
 	"github.com/pingcap/pd/v4/server/config"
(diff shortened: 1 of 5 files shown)


server/replicate/replicate.go
```diff
@@ -227,13 +231,35 @@ func (m *ModeManager) checkCanSync() bool {
...
+			return
+		}
+
+		total += len(regions)
+		for _, r := range regions {
+			if !bytes.Equal(key, r.GetStartKey()) {
+				log.Warn("found region gap", zap.ByteString("key", key), zap.ByteString("region-start-key", r.GetStartKey()))
+				total++
+			}
+			if r.GetReplicateStatus().GetRecoverId() == m.drAutosync.RecoverID &&
```
nolouch, line 251: StateID?

disksing: Update it in later PR. The kvproto is not updated here.

server/replicate/replicate.go
```diff
@@ -227,13 +231,35 @@ func (m *ModeManager) checkCanSync() bool {
...
+		if len(regions) == 0 {
+			log.Warn("scan empty regions", zap.ByteString("start-key", key))
+			total++ // make sure it won't complete
+			return
+		}
+
+		total += len(regions)
+		for _, r := range regions {
+			if !bytes.Equal(key, r.GetStartKey()) {
+				log.Warn("found region gap", zap.ByteString("key", key), zap.ByteString("region-start-key", r.GetStartKey()))
```
rleungx, line 248: How about also printing the region ID?

disksing: ok

server/replicate/replicate.go
AndrewDi: @disksing Do we need to check region state, invalid region should not include in `total` count?

Issue #2323, opened 2020-04-03
dashboard.(*Manager).needResetAddress panic


mahjonp:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?

run tipocket/bank with 
random_kill,kill_pd_leader_5min,partition_one,subcritical_skews,big_skews,shuffle-leader-scheduler,shuffle-region-scheduler,random-merge-scheduler

[tipocket-scbank-k6zv5.log](https://github.com/pingcap/pd/files/4424671/tipocket-scbank-k6zv5.log)

2. What did you expect to see?

```
NONE
```

3. What did you see instead?

```
2020/04/02 17:20:26 control.go:497: [fatal] 1 panics occurred in ns: tipocket-scbank pod tipocket-scbank-pd-1. Content: [[2020/04/02 17:19:32.465 +00:00] [FATAL] [log.go:292] [panic] [recover={}] [stack="github.com/pingcap/log.Fatal\n\t/go/pkg/mod/github.com/pingcap/log@v0.0.0-20200117041106-d28c14d3b1cd/global.go:59\ngithub.com/pingcap/pd/v4/pkg/logutil.LogPanic\n\t/go/src/github.com/pingcap/pd/pkg/logutil/log.go:292\nruntime.gopanic\n\t/usr/local/go/src/runtime/panic.go:679\nruntime.goPanicIndex\n\t/usr/local/go/src/runtime/panic.go:75\ngithub.com/pingcap/pd/v4/pkg/dashboard.(*Manager).needResetAddress\n\t/go/src/github.com/pingcap/pd/pkg/dashboard/manager.go:163\ngithub.com/pingcap/pd/v4/pkg/dashboard.(*Manager).checkAddress\n\t/go/src/github.com/pingcap/pd/pkg/dashboard/manager.go:141\ngithub.com/pingcap/pd/v4/pkg/dashboard.(*Manager).serviceLoop\n\t/go/src/github.com/pingcap/pd/pkg/dashboard/manager.go:92"]
```

4. What version of PD are you using (`pd-server -V`)?

```
/ # /pd-server -V
Release Version: v4.0.0-beta.2-32-g2b542f409
Git Commit Hash: 2b542f409047891764e4043abb370ecc8d5c8bbf
Git Branch: HEAD
UTC Build Time:  2020-04-02 01:19:25
```

PR #2326, opened 2020-04-03
dashboard: check ClientUrls length

HunDunDM:

#2323 

Check the length of ClientUrls of all members. If it is 0, it is considered that members are not ready.

### Check List 

Tests 

 - Manual test

### DIFF
--- a/pkg/dashboard/manager.go
+++ b/pkg/dashboard/manager.go
@@ -115,7 +115,20 @@ func (m *Manager) updateInfo() {
 
 	var err error
 	if m.members, err = cluster.GetMembers(m.srv.GetClient()); err != nil {
-		log.Error("failed to get members")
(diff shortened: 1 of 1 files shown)


rleungx:
The panic happened inside `needResetAddress` actually

pkg/dashboard/manager.go
```diff
@@ -115,7 +115,20 @@ func (m *Manager) updateInfo() {
 
 	var err error
 	if m.members, err = cluster.GetMembers(m.srv.GetClient()); err != nil {
-		log.Error("failed to get members")
+		log.Warn("failed to get members")
```
rleungx, line 118: How about printing the error?

HunDunDM: ok

Issue #1846, opened 2019-10-25
Improve the heartbeat processing


nolouch:
## Description

PD collects cluster information through heartbeats, It can be said that the processing power of the heartbeat is the ceiling of the entire cluster. It is very important to improve the performance of heartbeat processing. There is a benchmark tool `pd-heartbeat-bench` in PD project, now the benchmark result like:
```
Summary:
  Total:	9.7948 secs.
  Slowest:	0.6721 secs.
  Fastest:	0.0000 secs.
  Average:	0.0002 secs.
  Stddev:	0.0057 secs.
  Requests/sec:	102097.3460

Response time histogram:
  0.0000 [1]	|
  0.0672 [999313]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1344 [339]	|
  0.2016 [187]	|
  0.2688 [72]	|
  0.3361 [41]	|
  0.4033 [18]	|
  0.4705 [19]	|
  0.5377 [15]	|
  0.6049 [13]	|
  0.6721 [2]	|

Latency distribution:
  10% in 0.0000 secs.
  25% in 0.0000 secs.
  50% in 0.0000 secs.
  75% in 0.0000 secs.
  90% in 0.0000 secs.
  95% in 0.0000 secs.
  99% in 0.0000 secs.
  99.9% in 0.0329 secs.
```
Goals:
- Improve the heartbeat processing, expect above 300k Requests/sec with `region-update-ratio=0.1`
- Don't change the behavior of log in heartbeat path

Tasks:

- [x] [Reduce the unnecessary clone overhead in heartbeat process](https://github.com/pingcap/pd/issues/2391)
- [x] [Reduce the unnecessary update for the new region in a heartbeat](https://github.com/pingcap/pd/issues/2392)
- [ ] Optimize the storage process

## Difficulty

- Medium

## Score

- 3488

## Mentor(s)

- @nolouch 

## Recommended Skills

- Go

wangrzneu:
What is the test environment where you got the test result?

PR #2331, opened 2020-04-06
Improve the heartbeat processing.

mantuliu:
UCP #1846 

 Reduce the number of method calls:GetOverlaps

Try not to remove the region in the tree structure

When the store clone,add a method called ShallowClone,it won't clone the meta

### Check List 

Tests 

 - Unit test

--------- Bench heartbeat (Round 2) ----------
store 1 finish heartbeat, cost time: 4.460517773s
store 20 finish heartbeat, cost time: 4.534937041s
store 3 finish heartbeat, cost time: 4.597973534s
store 13 finish heartbeat, cost time: 4.715421933s
store 7 finish heartbeat, cost time: 4.722201601s
store 15 finish heartbeat, cost time: 4.742612709s
store 14 finish heartbeat, cost time: 4.753740889s
store 2 finish heartbeat, cost time: 4.759623904s
store 11 finish heartbeat, cost time: 4.768188915s
store 18 finish heartbeat, cost time: 4.769229559s
store 12 finish heartbeat, cost time: 4.802837829s
store 10 finish heartbeat, cost time: 4.809404506s
store 9 finish heartbeat, cost time: 4.826523713s
store 6 finish heartbeat, cost time: 4.826793564s
store 4 finish heartbeat, cost time: 4.827608051s
store 8 finish heartbeat, cost time: 4.832837682s
store 19 finish heartbeat, cost time: 4.855967181s
store 5 finish heartbeat, cost time: 4.898822867s
store 17 finish heartbeat, cost time: 4.925156153s
store 16 finish heartbeat, cost time: 4.930070917s

Summary:
  Total:        4.9301 secs.
  Slowest:      0.8004 secs.
  Fastest:      0.0000 secs.
  Average:      0.0001 secs.
  Stddev:       0.0025 secs.
  Requests/sec: 202838.9156


nolouch:
Hi @mantuliu, thanks for your contribution.
Could you show the performance improvement in more detail?

mantuliu:
> Hi @mantuliu, thanks for your contribution.
> Could you show the performance improvement in more detail?

HI @nolouch I answered on slack.Thasks,PTAL.

nolouch:
cool!

server/core/region.go
```diff
@@ -608,7 +608,51 @@ func (r *RegionsInfo) AddRegion(region *RegionInfo) []*RegionInfo {
...
+		for _, item := range overlaps {
+			r.RemoveRegion(r.GetRegion(item.GetID()))
+		}
+		if origin := r.regions.Get(region.GetID()); origin != nil {
+			r.RemoveRegion(origin)
+		}
+		r.tree.update(region)
+	} else {
+		for _, item := range overlaps {
+			if item.GetID() == region.GetID() {
```
nolouch, line 628: Why pass this? the other meta information may changed.

nolouch: Does it need to update the tree?

server/core/region.go
```diff
@@ -893,6 +939,47 @@ func DiffRegionPeersInfo(origin *RegionInfo, other *RegionInfo) string {
 	return strings.Join(ret, ",")
 }
 
+// isEqualPeers return the peers is equal or not
+func isEqualPeers(origin []*metapb.Peer, other []*metapb.Peer) bool {
```
nolouch, line 943: Why need this? I think it's easy to ignore update this function if we add some items in this meta proto.

server/core/store.go
```diff
@@ -640,7 +663,7 @@ func (s *StoresInfo) SetRegionSize(storeID uint64, regionSize int64) {
 // UpdateStoreStatus updates the information of the store.
 func (s *StoresInfo) UpdateStoreStatus(storeID uint64, leaderCount int, regionCount int, pendingPeerCount int, leaderSize int64, regionSize int64) {
 	if store, ok := s.stores[storeID]; ok {
-		newStore := store.Clone(SetLeaderCount(leaderCount),
+		newStore := store.ShallowClone(SetLeaderCount(leaderCount),
```
nolouch, line 666: Can you briefly list some optimization points made in this PR and their benefits? like optimization of precheck can improve 50%, optimization of store clone can improve xx%

server/core/region.go
mantuliu: Because the overlaps will contain the old value of the current region, here is not to delete the old region, and take a workaround to delete the regions in all overlaps that are not corresponding to the current region. In general, this overlaps only There will be the old value of the current region. At this time, this region is not deleted from the tree, but the method of updating the region in the regionItem is used: code such as: oldRegionItem.region = region

server/cluster/cluster.go
```diff
@@ -539,35 +539,51 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+
+		c.core.ReplaceOrAddRegion(region, overlaps, oldRegionItem)
 		if c.storage != nil {
 			for _, item := range overlaps {
-				if err := c.storage.DeleteRegion(item.GetMeta()); err != nil {
-					log.Error("failed to delete region from storage",
-						zap.Uint64("region-id", item.GetID()),
-						zap.Stringer("region-meta", core.RegionToHexMeta(item.GetMeta())),
-						zap.Error(err))
+				if region.GetID() != item.GetID() {
```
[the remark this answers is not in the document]
mantuliu: In the original logic, the old value of the current region is deleted first, so there is no current region in overlaps, but in the current code, in order to improve performance, getOverlaps operations are performed as little as possible. So under normal circumstances, there is current region in overlaps, so I don't go to delete this value at the storage level, if it needs to delete or update the region, leave this operation to saveKv.

server/core/region.go
mantuliu: The update will not be missed, because the store.update (region) method will be called in the addRegionToSubTree method, which will determine that this region exists in the tree, and will update this region:
```
func (rst *regionSubTree) update(region *RegionInfo) {
	if r := rst.find(region); r != nil {
		rst.totalSize += region.approximateSize - r.region.approximateSize
		rst.totalKeys += region.approximateKeys - r.region.approximateKeys
		r.region = region
		return
	}
```
The following code is mainly to prevent missed deletion, for example: when the leader transfer peer occurs, we must delete this region in the subtree and then add it to the subtree:
```
if origin := r.regions.Get(region.GetID()); origin != nil {
			remove := false
			if origin.leader.GetId() != region.leader.GetId() {
				remove = true
			} else if !isEqualPeers(origin.GetVoters(), region.GetVoters()) {
				remove = true
			} else if !isEqualPeers(origin.GetLearners(), region.GetLearners()) {
				remove = true
			} else if !isEqualPeers(origin.GetPendingPeers(), region.GetPendingPeers()) {
				remove = true
			}
			if remove {
				r.removeRegionFromSubTree(origin)
			}
		}
```

server/core/store.go
mantuliu: OK,I wil give the list tomorrow.

server/cluster/cluster.go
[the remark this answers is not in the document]
nolouch: got it, can you and a comments?

server/core/region.go
nolouch: Yes, I mean this logic not easy to maintain.  If we add a new option in `metapb.Peer`, how do you know should we update the funciton `isEualPeers`?

server/core/region.go
nolouch: so should we add a line:
```
r.tree.update(region)
```

server/core/region.go
```diff
@@ -608,7 +608,51 @@ func (r *RegionsInfo) AddRegion(region *RegionInfo) []*RegionInfo {
...
 
 	r.regions.Put(region)
+	r.addRegionToSubTree(region)
+	return overlaps
+}
+
+// ReplaceOrAddRegion adds or replaces RegionInfo to regionTree and regionMap, also update leaders and followers by region peers
+func (r *RegionsInfo) ReplaceOrAddRegion(region *RegionInfo, overlaps []*RegionInfo, oldRegionItem *regionItem) {
+	// Add to tree and regions.
+	if !isEqualRegion(region, oldRegionItem) {
```
nolouch, line 618: even if the `region id` is the same, the leader may be changed, so maybe we need to remove the origin from the subtree.

server/core/region.go
mantuliu: I think  the line of 'oldRegionItem.region = region'  has an updated effect. Using this method (oldRegionItem.region = region) instead of another one(r.tree.update(region)) is also a key point for performance improvement.

server/core/region.go
mantuliu:
```
if origin.leader.GetId() != region.leader.GetId() {
				remove = true
			} else if !isEqualPeers(origin.GetVoters(), region.GetVoters()) {
				remove = true
			} else if !isEqualPeers(origin.GetLearners(), region.GetLearners()) {
				remove = true
			} else if !isEqualPeers(origin.GetPendingPeers(), region.GetPendingPeers()) {
				remove = true
			}
			if remove {
				r.removeRegionFromSubTree(origin)
			}
```
This code can realize the problem of subtree update after leader migration.

server/core/region.go
mantuliu: I reply on slack.

server/core/store.go
mantuliu: Because the log and storage will affect the test results, during the comparison process, the log and storage were commented out by me. The test conditions are: 1 million region 20store 0.1 region-update-ratio, 10heartbeat-rounds. The average heartbeat is the average of the last 9 times.

---|---|---|---|---
1|   master branch |     |   133051 | 0%    
2|  not remove regin in the tree when old region is in the tree  |  region.go 633 : oldRegionItem.region = region   |  138364  |     4%
3|  not remove regin in the subtree when old region's peer is not change  is in the subtree  |  region.go 635-647 : ```if remove {r.removeRegionFromSubTree(origin)}```   |  175809  |     27%
4|PreCheckPutRegion add the 'LOOSE' check.|  cluster.go 447: origin, _, _, err := c.core.PreCheckPutRegion(region, core.LOOSE)   | 182601   |     3.86%
5|use ShallowClone when UpdateStoreStatus.|   store.go 666: newStore := store.ShallowClone(SetLeaderCount(leaderCount),  | 240027   |     31%
6| reduce the number of updates when update related stores.   |  cluster.go 575:		var storeMap map[uint64]uint64  |  244630  |     1.86%

server/core/region.go
nolouch: got

server/core/region.go
nolouch: The code is in `else` branch.

server/core/store.go
nolouch: thanks! 👍

server/core/region_test.go
```diff
@@ -131,6 +132,318 @@ func (*testRegionKey) TestRegionKey(c *C) {
...
+		StartKey: []byte(fmt.Sprintf("%20d", 184)),
+		EndKey:   []byte(fmt.Sprintf("%20d", 211)),
+	}, peer1)
+	region.learners = append(region.learners, region.voters[1])
+	region.pendingPeers = append(region.pendingPeers, region.voters[2])
+	overlaps, item := regions.GetOverlaps(region)
+	regions.ReplaceOrAddRegion(region, overlaps, item)
+	c.Assert(len(overlaps), Equals, 4)
+	c.Assert(item.region.GetID(), Equals, uint64(19))
+	c.Assert(regions.leaders[1].length(), Equals, 19)
```
rleungx, line 232: How about abstract these lines into a function to avoid repetition?

server/core/region_test.go
```diff
@@ -131,6 +132,318 @@ func (*testRegionKey) TestRegionKey(c *C) {
...
+	c.Assert(isEqualPeers(others3[:], peers[:]), Equals, false)
+	c.Assert(isEqualPeers(peers[:], others3[:]), Equals, false)
+
+	var others4 [4]*metapb.Peer
+	others4[2] = &metapb.Peer{StoreId: uint64(1), Id: uint64(1)}
+	others4[3] = &metapb.Peer{StoreId: uint64(2), Id: uint64(2)}
+	others4[1] = &metapb.Peer{StoreId: uint64(3), Id: uint64(3)}
+	c.Assert(isEqualPeers(peers[:], others4[:]), Equals, true)
+	c.Assert(isEqualPeers(others4[:], peers[:]), Equals, true)
+
```
rleungx, line 444: Please remove the empty line

server/core/store.go
```diff
@@ -91,6 +91,29 @@ func (s *StoreInfo) Clone(opts ...StoreCreateOption) *StoreInfo {
 	return store
 }
 
+// ShallowClone creates a copy of current StoreInfo,but not clone 'meta'.
```
rleungx, line 94:
```suggestion
// ShallowClone creates a copy of current StoreInfo, but not clone 'meta'.
```

server/core/region.go
rleungx: I agree that this part of the code is hard to maintain.

server/core/region_test.go
mantuliu: great idea.

server/core/region.go
```diff
@@ -651,16 +689,65 @@ func (r *RegionsInfo) AddRegion(region *RegionInfo) []*RegionInfo {
...
 		store.update(region)
 	}
+}
 
-	return overlaps
+// shouldRemoveFromSubTree return true when the region leader changed, peer transferred,
+// new peer was created, learners changed, pendingPeers changed, and so on.
+func (r *RegionsInfo) shouldRemoveFromSubTree(region *RegionInfo, origin *RegionInfo) bool {
+	checkPeersChange := func(origin []*metapb.Peer, other []*metapb.Peer) bool {
+		for _, a := range origin {
```
rleungx, line 698: The code between line 698 and line 716 has some overlap logic?

mantuliu: Do some change, ok?
```
	checkPeersChange := func(origin []*metapb.Peer, other []*metapb.Peer) bool {
		checkPeersDiff := func(origin []*metapb.Peer, other []*metapb.Peer) bool {
			for _, a := range origin {
				if a == nil {
					continue
				}
				both := false
				for _, b := range other {
					if b == nil {
						continue
					}
					if a.GetStoreId() == b.GetStoreId() && a.GetId() == b.GetId() {
						both = true
						break
					}
				}
				if !both {
					return true
				}
			}
			return false
		}
		if checkPeersDiff(origin, other) || checkPeersDiff(other, origin) {
			return true
		}
		return false
	}
```

server/core/region.go
```diff
@@ -608,7 +608,45 @@ func (r *RegionsInfo) AddRegion(region *RegionInfo) []*RegionInfo {
...
+			r.RemoveRegion(r.GetRegion(item.GetID()))
+		}
+		// If old region exists. remove too.
+		if origin := r.regions.Get(region.GetID()); origin != nil {
+			r.RemoveRegion(origin)
+		}
+		// Add the new region to the regionTree.
+		r.tree.update(region)
+	} else {
+		// When else, remove the regions of the overlaps except the old region.
```
nolouch, line 630: Can you add a test about this logic?
```
There is a situation:
Region:   |____________2______________|
Region:                 |______2______|(new)
````
the old region 2 split to two reigons , and new region 2(has same ID) coming from heartbeat.
In the situation, will remove the old region cause by `overlaps` contains the old region, but the update(line 639) is in the `oldRegionItem`(that already remove from the tree).

BTW, Does the main optimize is the `removeRegionFromSubTree`?

server/core/store.go
nolouch: Would you like split the PR to multiple PR, It's easy to review, reduce bugs, and even speed up your PR to merge. and also we can more easy to trace why do the change.

mantuliu: OK.PTAL.The first one : #2389

PR #2339, opened 2020-04-10
replication_mode: estimated recover progress

### LINKED ISSUE #2327: Synchronous replication: optimize recover progress
The first implementation scans all regions to check if recover process is completed. Need to improve.

disksing:
Fix #2327

- save start key, not to scan from first region every time
- use sample to estimate progress, not to scan all regions

### Check List 

Tests 
 - Unit test


server/replication/replication_mode.go
```diff
@@ -270,32 +281,74 @@ func (m *ModeManager) checkCanSync() bool {
...
-			key = r.GetEndKey()
+			m.drSampleRecoverCount = 0
+			key := m.drRecoverKey
+			for _, r := range sampleRegions {
+				if m.checkRegionRecover(r, key) {
+					m.drSampleRecoverCount++
+				}
+				key = r.GetEndKey()
+			}
+			m.drSampleTotalRegion = len(sampleRegions)
```
nolouch, line 320: Only protect by `RLock`?

server/replication/replication_mode.go
```diff
@@ -270,32 +281,74 @@ func (m *ModeManager) checkCanSync() bool {
...
+			m.drTotalRegion = m.cluster.GetRegionCount()
+			return
 		}
 	}
-	return
+}
+
+func (m *ModeManager) estimateProgress() float32 {
+	if len(m.drRecoverKey) == 0 && m.drRecoverCount > 0 {
```
nolouch, line 328: need lock?

server/replication/replication_mode.go
disksing: > they are accessed without locks as they are only used by background job.
https://github.com/pingcap/pd/pull/2339/files#diff-8d79db5b0a294471ed4a80860ded1629R45

server/replication/replication_mode.go
```diff
@@ -276,38 +287,80 @@ func (m *ModeManager) checkCanSync() bool {
...
+func (m *ModeManager) checkRegionRecover(region *core.RegionInfo, startKey []byte) bool {
+	if !bytes.Equal(startKey, region.GetStartKey()) {
+		log.Warn("found region gap", zap.ByteString("key", startKey), zap.ByteString("region-start-key", region.GetStartKey()), zap.Uint64("region-id", region.GetID()))
+		return false
+	}
+	return region.GetReplicationStatus().GetStateId() == m.drAutoSync.StateID &&
+		region.GetReplicationStatus().GetState() == pb.RegionReplicationState_INTEGRITY_OVER_LABEL
+}
+
+func (m *ModeManager) updateRecoverProgress(progress float32) {
```
rleungx, line 360: Do we need metrics?

disksing: good idea

PR #2341, opened 2020-04-10
dashboard: make it possible to not include dashboard when compiling

HunDunDM:

`CGO_ENABLED` must be 1 when including Dashboard. We need `CGO_ENABLED` to be 0 on some occasions.

Set `DASHBOARD=0` in `make pd-server` or use `make pd-server-basic` to compile pd-server without dashboard. It would use `CGO_ENABLED=0` when compiling.

### Check List 

Tests 

* Manual test (add detailed scripts or steps below)
  * `make pd-server`
  * `make pd-server-basic`
  * `DASHBOARD=0 make pd-server`


disksing:
'make it possible to ...' would be a better PR title.

breeswish:
what is the scenario for "We need CGO_ENABLED to be 0 on some occasions."?

Makefile
```diff
@@ -32,12 +32,25 @@ DEADLOCK_DISABLE := $$(\
...
 
 ifneq ($(SWAGGER), 0)
 	BUILD_TAGS += swagger_server
 endif
 
+ifeq ($(DASHBOARD), 0)
+	BUILD_TAGS += without_dashboard
+else
+	BUILD_CGO_ENABLED := 1
+endif
```
disksing, line 47: how about
```
ifneq ($(DASHBOARD), 0)
  BUILD_TAGS += with_dashboard
  BUILD_CGO_ENABLED := 1
endif
```

HunDunDM: Now, our test code is without tags, in order to ensure that the dashboard will be started in this case.

disksing: can we add `DASHBOARD=1` when run `go test`?

disksing:
> what is the scenario for "We need CGO_ENABLED to be 0 on some occasions."?

If possible, we try to avoid introducing cgo in the release binary to prevent potential dynamic library dependency problems.

Makefile
HunDunDM: It is useless to add this when running `go test`, this is only useful in the `Makefile`. `go test` needs to add `-tags "with_dashboard"`. I think the current method has changed a little, and the default startup also meets the current requirements.

disksing: ok

breeswish:
> > what is the scenario for "We need CGO_ENABLED to be 0 on some occasions."?
> 
> If possible, we try to avoid introducing cgo in the release binary to prevent potential dynamic library dependency problems.

Then we will have a cost of dashbaord feature is not available. Is this a real use case (do not want the Dashboard feature totally in order to use a self compiled PD binary)?

disksing:
> > > what is the scenario for "We need CGO_ENABLED to be 0 on some occasions."?
> > 
> > 
> > If possible, we try to avoid introducing cgo in the release binary to prevent potential dynamic library dependency problems.
> 
> Then we will have a cost of dashbaord feature is not available. Is this a real use case (do not want the Dashboard feature totally in order to use a self compiled PD binary)?

It is real use case.

PR #2343, opened 2020-04-10
replication_mod: add total and current metrics

AndrewDi:
As when we use 2dc replication, we may need more metrics to monitor replication progress.

add `total` and `current` metric through httpstatus api.

### Check List 

Tests 
 - No code

Code changes

 - Has configuration change
 - Has HTTP API interfaces change (Don't forget to [update API document](https://github.com/pingcap/pd/blob/master/docs/development.md#updating-api-documentation))
 - Has persistent data change


AndrewDi:
@disksing PTAL

PR #2344, opened 2020-04-11
replication: support update replication mode online 

disksing:
Fix: https://github.com/pingcap/pd/issues/2317

- persist replication mode config
- notify replicationModeManager when config is updated
- handle special case when mode changed from majority to dr_auto_sync
- add API to update config
- add pd-ctl command to update config
- add test

### Check List 

Tests 
 - Unit test

Code changes
 - Has HTTP API interfaces change (Don't forget to [update API document](https://github.com/pingcap/pd/blob/master/docs/development.md#updating-api-documentation))


tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -249,6 +268,15 @@ func showClusterVersionCommandFunc(cmd *cobra.Command, args []string) {
 	cmd.Println(r)
 }
 
+func showReplicationModeCommandFunc(cmd *cobra.Command, args []string) {
+	r, err := doRequest(cmd, replicationModePrefix, http.MethodGet)
+	if err != nil {
+		cmd.Printf("Failed to get replication mode config:%s\n", err)
```
rleungx, line 274:
```suggestion
		cmd.Printf("Failed to get replication mode config: %s\n", err)
```

server/api/config.go
```diff
@@ -307,3 +307,34 @@ func (h *confHandler) SetClusterVersion(w http.ResponseWriter, r *http.Request)
...
+}
+
+// @Tags config
+// @Summary Set replication mode config.
+// @Accept json
+// @Param body body object string "json params"
+// @Produce json
+// @Success 200 {string} string
+// @Failure 500 {string} string "PD server failed to proceed the request."
+// @Failure 503 {string} string "PD server has no leader."
```
rleungx, line 327: the code will not return 503 error?

server/api/router.go
```diff
@@ -86,6 +86,8 @@ func createRouter(ctx context.Context, prefix string, svr *server.Server) *mux.R
 	apiRouter.HandleFunc("/config/label-property", confHandler.SetLabelProperty).Methods("POST")
 	apiRouter.HandleFunc("/config/cluster-version", confHandler.GetClusterVersion).Methods("GET")
 	apiRouter.HandleFunc("/config/cluster-version", confHandler.SetClusterVersion).Methods("POST")
+	apiRouter.HandleFunc("/config/replication-mode", confHandler.GetReplicationMode).Methods("GET")
+	apiRouter.HandleFunc("/config/replication-mode", confHandler.SetReplicationMode).Methods("POST")
```
rleungx, line 90: Do we also need to support changing it through /config API?

disksing: I guess it is ok to not support. Just like `cluster-version`, `label-property`, etc.

rleungx: Does this config need SQL support? If it needs, I think it's better to support changing through /config. For `cluster-version` and `label-property`, PR #2354 is trying to support it.

PR #2345, opened 2020-04-11
replication_mode: persist status to a file

disksing:
Fix https://github.com/pingcap/pd/issues/2274
In a slightly different approach: the status is saved to a file. The reason is when majority PD servers are down, we need to first recover from minority nodes before we can know status.

before change status, PD leader sends a HTTP request to all members and each one save request body to a file (filename is `DR_STATE`).
### Check List 

Tests 

Code changes
 - Has persistent data change


server/server.go
```diff
@@ -1365,3 +1366,32 @@ func (s *Server) reloadConfigFromKV() error {
...
+
+// ReplicateFileToAllMembers is used to synchronize state among all members.
+// Each member will write `data` to a local file named `name`.
+func (s *Server) ReplicateFileToAllMembers(ctx context.Context, name string, data []byte) error {
+	resp, err := s.GetMembers(ctx, nil)
+	if err != nil {
+		return err
+	}
+	for _, member := range resp.Members {
+		url := member.GetClientUrls()[0] + filepath.Join("/pd/api/v1/admin/persist-file", name)
```
rleungx, line 1378: Do we need to check the length of `GetClientUrls`?

rleungx:
when do we read this file?

server/server.go
disksing: ok

disksing:
> when do we read this file?

@rleungx When the primary dc is disconnected with dr dc, we can recover service using data in dr dc. The problem is before recover, we need to make sure the replication mode status is `sync` before the network isolation (if the status is `async`, it may lead to data lose). So this is when we need to read to file: before recover using minority replica, we read the file to determine the status before network isolation.

PR #2348, opened 2020-04-13
safepoint: support to set safepoints for services

shafreeck:
Signed-off-by: Shafreeck Sea <shafreeck@gmail.com>

When a TiKV cluster has multiple components rely on the GC safe point, they should have agreement of value of safe point which determines what data would be deleted.

For example, TiCDC captures the data change event after the safe point, if the safe point advances too quickly, TiCDC can not handle the case.

This PR relies on https://github.com/pingcap/kvproto/pull/603

- [x] TODO: Merge https://github.com/pingcap/kvproto/pull/603 and update go.mod

* Add `UpdateServiceGCSafePoint` to update the safe point for service and return the minimum value across all the services

### Check List 

Tests 

 - [x] Unit test
 - [x] Manual test (add detailed scripts or steps below)

Code changes

 - Has persistent data change


server/core/storage.go
```diff
@@ -362,27 +362,44 @@ func (s *Storage) Close() error {
 }
 
 // SaveGCSafePoint saves new GC safe point to storage.
-func (s *Storage) SaveGCSafePoint(safePoint uint64) error {
-	key := path.Join(gcPath, "safe_point")
+func (s *Storage) SaveGCSafePoint(serviceID string, safePoint uint64, ttl int64) error {
```
disksing, line 365: can it guarantee backward compatibility?

server/core/storage.go
```diff
@@ -362,27 +362,44 @@ func (s *Storage) Close() error {
...
 		return 0, nil
 	}
-	safePoint, err := strconv.ParseUint(value, 16, 64)
+	min := uint64(0)
+	for _, val := range safePoints {
+		if min < safePoint {
+			min = safePoint
+		}
+	}
+	return min, nil
```
disksing, line 384: when all safepoint exceeds TTL, it returns 0?

server/core/storage.go
shafreeck: Yes, the gRPC API is compatible

server/core/storage.go
shafreeck: I think so, TiDB would not set a TTL by default, so there is always a safe point.

server/core/storage.go
nolouch: Will tidb use a serviceID? if so,  the key `xxxx/safepoint` still exists after upgrade and it does not be pushed any more.

disksing: I was not doubt about API capatibility but the old data. After upgrade PD from old version to new version. will the old saved safePoint get lost?

server/server.go
```diff
@@ -141,6 +141,9 @@ type Server struct {
 	// Add callback functions at different stages
 	startCallbacks []func()
 	closeCallbacks []func()
+
+	// The minimum safe point of all services
```
rleungx, line 145: When do we need to use it, maybe we don't need this variable?

server/core/storage.go
```diff
@@ -385,6 +386,57 @@ func (s *Storage) LoadGCSafePoint() (uint64, error) {
...
+			s.Remove(key)
+			continue
+		}
+		if ssp.SafePoint < min.SafePoint {
+			min = ssp
+		}
+	}
+
+	return min, nil
+
```
rleungx, line 437: remove the empty line

shafreeck:
@disksing @nolouch  The feature is reimplemented and please review it again.

The original API and data are left unchanged. A new API called `UpdateServiceGCSafePoint` is added to coordinate the GC safe point value, it updates the safe point for service and returns the minimum value in the cluster. For TiDB, it can call `UpdateGCSafePoint` with the minimum value to trigger the distributed GC job or do the job locally.

server/core/storage.go
```diff
@@ -385,6 +386,56 @@ func (s *Storage) LoadGCSafePoint() (uint64, error) {
...
+	min := &ServiceSafePoint{SafePoint: math.MaxUint64}
+	now := time.Now().Unix()
+	for i, key := range keys {
+		ssp := &ServiceSafePoint{}
+		if err := json.Unmarshal([]byte(values[i]), ssp); err != nil {
+			return nil, err
+		}
+		if ssp.ExpiredAt < now {
+			s.Remove(key)
```
disksing, line 428: What if the expire value has been updated to a value that is not expired?

shafreeck: Good catch! Maybe a transaction is necessary here...

disksing: It seems that etcd API can accept an option to limit the operation to specific revision. Maybe it can help.

shafreeck: This is a data race problem, the ETCD has transaction APIs to handle this but it is not wrapped by PD. It would change a lot to add transaction support. In fact, this method would not be called concurrently, maybe I should add a comment now and then fix this if necessary later?

shafreeck: @disksing PTAL, a lock is used to resolve this.

shafreeck:
@nolouch @rleungx PTAL

PR #2349, opened 2020-04-13
*: reimplement dynamic config

rleungx:
This PR removes the original implementation and adds new APIs to register and get addresses of other components. 

### Check List 

Tests 

 - Unit test


pkg/component/manager.go
```diff
@@ -0,0 +1,103 @@
...
+		addresses = append(addresses, ca...)
+	}
+	return addresses
+}
+
+// GetAllComponentAddrs returns all components' addresses.
+func (c *Manager) GetAllComponentAddrs() map[string][]string {
+	c.RLock()
+	defer c.RUnlock()
+	return c.Addresses
```
HunDunDM, line 55: Need to clone, otherwise there will be data race problems.

pkg/component/manager.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Register is used for registering a component with an address to PD.
+func (c *Manager) Register(component, addr string) error {
+	c.Lock()
+	defer c.Unlock()
+
+	str := strings.Split(addr, ":")
+	if len(str) != 0 {
+		ip := net.ParseIP(str[0])
```
HunDunDM, line 77: address is not necessarily ip? And should it come with http or https?

pkg/component/manager.go
```diff
@@ -0,0 +1,103 @@
...
+	for component, ca := range c.Addresses {
+		if contains(ca, addr) {
+			return component
+		}
+	}
+	return ""
+}
+
+// Register is used for registering a component with an address to PD.
+func (c *Manager) Register(component, addr string) error {
```
HunDunDM, line 71: How to unregister?

tests/dashboard/service_test.go
```diff
@@ -156,21 +144,6 @@ func (s *serverTestSuite) TestDashboard(c *C) {
 	s.checkServiceIsStarted(c, servers, leader)
 	c.Assert(leader.GetServer().GetScheduleOption().GetDashboardAddress(), Equals, dashboardAddress2)
 
-	// Changing dashboard address
```
HunDunDM, line 159: Do we have other ways to run this test?

rleungx: Any suggestion?

server/api/component_test.go
```diff
@@ -0,0 +1,121 @@
...
+	var output1 []string
+	err = readJSON(addr1, &output)
+	c.Assert(strings.Contains(err.Error(), "404"), IsTrue)
+	c.Assert(len(output1), Equals, 0)
+
+	// register 2 c1 and 1 c2
+	reqs := []map[string]string{
+		{"component": "c1", "addr": "127.0.0.1:1"},
+		{"component": "c1", "addr": "127.0.0.1:2"},
+		{"component": "c2", "addr": "127.0.0.1:3"},
```
overvenus, line 63: Could you add a test case to test domain address?

```suggestion
		{"component": "c2", "addr": "127.0.0.1:3"},
		{"component": "c3", "addr": "example.com"},
```

pkg/component/manager_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+func Test(t *testing.T) {
+	TestingT(t)
+}
+
+var _ = Suite(&testManagerSuite{})
+
+type testManagerSuite struct{}
+
+func (s *testManagerSuite) TestManager(c *C) {
```
HunDunDM, line 31: Test the address that is not IP?

server/cluster/cluster.go
```diff
@@ -621,6 +615,7 @@ func (c *RaftCluster) updateStoreStatusLocked(id uint64) {
 	c.core.UpdateStoreStatus(id, leaderCount, regionCount, pendingPeerCount, leaderRegionSize, regionSize)
 }
 
+//nolint
```
HunDunDM, line 618: why?

HunDunDM: We can limit the scope of `nolint`.

tests/dashboard/service_test.go
HunDunDM: I will find a way to make up for this test later.

PR #2353, opened 2020-04-15
schedulers: fix hot region scheduler panic

disksing:

Fix https://github.com/pingcap/pd/issues/2352

push counters to a temp slice, then push to the operator later.

### Check List 

Tests 
- (no test because I'm not sure how to reproduce, but the problem is obvious...)


disksing:
PTAL @rleungx @lhy1024

PR #2354, opened 2020-04-15
api: support input with a prefix when using config

rleungx:
Currently, the input of JSON only supports like `{"schedule-region-limit":20}` when using config API, but TiDB will send a JSON like `{"schedule.schedule-region-limit":20}` when using SQL to change the config. This PR can be reviewed after #2349 is merged.

This PR support using the key like "xxx.xxx-xxx-xxx" when sending a JSON as an input.

### Check List 

Tests 

 - Unit test


server/api/config.go
```diff
@@ -75,69 +78,206 @@ func (h *confHandler) GetDefault(w http.ResponseWriter, r *http.Request) {
...
-		h.rd.JSON(w, http.StatusInternalServerError, err.Error())
-		return
+
+	h.rd.JSON(w, http.StatusOK, nil)
+}
+
+func (h *confHandler) updateConfig(cfg *config.Config, key string, value interface{}) error {
+	kp := strings.Split(key, ".")
+	switch kp[0] {
+	case "schedule":
```
HunDunDM, line 122: Do we need to ensure that the length is equal to 2?

server/api/config.go
```diff
@@ -75,69 +78,206 @@ func (h *confHandler) GetDefault(w http.ResponseWriter, r *http.Request) {
...
+		return h.updateSchedule(cfg, kp[len(kp)-1], value)
+	case "replication":
+		return h.updateReplication(cfg, kp[len(kp)-1], value)
+	case "replication-mode":
+		return h.updateReplicationModeConfig(cfg, kp[1:], value)
```
HunDunDM, line 127: Need to ensure that len is greater than 1, otherwise panic.

PR #2355, opened 2020-04-15
dashboard: reverse proxy support tls

HunDunDM:

* reverse proxy support tls

### Check List 

Tests 


cmd/pd-server/main.go
```diff
@@ -127,12 +127,6 @@ func main() {
 	log.Info("Got signal to exit", zap.String("signal", sig.String()))
 
 	svr.Close()
-	switch sig {
-	case syscall.SIGTERM:
```
rleungx, line 131: It seems `SIGQUIT` will also exit with code 0

HunDunDM: Yes, tikv / tidb is currently the same behavior. This change is to allow them to have uniform behavior in tiup. In addition, as long as the signal is processed correctly, the return value should only be related to whether it was successfully closed.

HunDunDM: According to this [issue](https://github.com/pingcap/pd/issues/392), after receiving the SIGINT, the exit code of 1 is reasonable behavior, not an error. I rolled back this part of the code

PR #2363, opened 2020-04-20
api: fix the certificate problem in API client

nolouch:

Fix https://github.com/pingcap/pd/issues/2362

add TLS supports for the client.

### Release note 
- Fix the certificate problem in API client

### Check List 

Tests 


server/api/util.go
```diff
@@ -20,17 +20,14 @@ import (
...
-var dialClient = &http.Client{
-	Transport: &http.Transport{
-		DisableKeepAlives: true,
-	},
-}
-
 var (
+
+	// dialClient used to dial http request.
+	dialClient        = cluster.DialClient
```
rleungx, line 30: The client in cluster package has a timeout. Does the client in API needs it?

nolouch: I checked it and it only used in a redirect for the scheduler handler, I think it's ok to timeout.

rleungx: If we use API to get all regions and the region number is large, will it break the process of getting regions?

nolouch: the client in the user, not in PD.

rleungx: got

PR #2368, opened 2020-04-22
*: unify the http client

nolouch:

there are many HTTP clients in PD. It's not easy to maintain, we want to supports TLS if the server is configured.

- init HTTP client in `server`
- replace `middleware` client
- replace `cluster` client
- replace `server/api` client
### Release note 
no need to write

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/api/util.go
+++ b/server/api/util.go
@@ -24,13 +24,6 @@ import (
 )
 
 var (
-
-	// testDialClient used to dial http request. only used for test.
-	testDialClient = &http.Client{
-		Transport: &http.Transport{
-			DisableKeepAlives: true,
-		},
-	}
 	errNoImplement    = errors.New("no implement")
 	errOptionNotExist = func(name string) error { return errors.Errorf("the option %s does not exist", name) }
 )
--- a/server/api/util.go
+++ b/server/api/util.go
@@ -20,14 +20,17 @@ import (
 	"net/http"
 	"net/url"
 
-	"github.com/pingcap/pd/v4/server/cluster"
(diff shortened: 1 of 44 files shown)


server/server.go
```diff
@@ -310,6 +313,13 @@ func (s *Server) startEtcd(ctx context.Context) error {
 		}
 	}
 	s.client = client
+	s.httpClient = &http.Client{
```
rleungx, line 316: The client doesn't have a timeout which is different from the original behavior

nolouch: the new timeout is set up by `NewRequestWithContext`. you can check it.

nolouch: like https://github.com/pingcap/pd/pull/2368/files#diff-02ca8f1595ce93b75591586acd227047R1651-R1659

rleungx: Does https://github.com/pingcap/pd/pull/2368/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR1182 also need a context?

server/api/util.go
```diff
@@ -20,14 +20,17 @@ import (
...
-	"github.com/pingcap/pd/v4/server/cluster"
 	"github.com/pkg/errors"
 )
 
 var (
 
-	// dialClient used to dial http request.
-	dialClient        = cluster.DialClient
+	// testDialClient used to dial http request. only used for test.
+	testDialClient = &http.Client{
```
HunDunDM, line 29: Shouldn't variables used only in tests appear here?

PR #2226, opened 2020-03-11
Support rate control for removing peer

wangrzneu:
UCP #2178

### What problem does this PR solve? 
Support rate control for removing peer
https://github.com/pingcap/pd/issues/2178

### What is changed and how it works?
- add a region-remove type store limiter
- save the StoreLimit in a map, and use the type as the key

### Check List 

Tests 

 - Unit test

Related changes

 - Need to update the documentation
 - Need to be included in the release notes

PR #2372, opened 2020-04-23
schedule: change store limit to unlimited when offline a store

rleungx:
After #2226, the speed of both add and remove a store may be slower because of the rate limit of removing peers. 

This PR changes the rate of removing peers to unlimited when a store is offline.

### Release note 
Change rate limit of removing peers to unlimited when offline a store

### Check List 

Tests 

 - Manual test 

When removing a store, the limit of removing peers for the offline store will be `556780641449999900` which represents unlimited.
```
» store limit region-remove
{
  "1": {
    "rate": 556780641449999900,
    "mode": "auto"
  },
  "104": {
    "rate": 15,
    "mode": "auto"
  },
  "116": {
    "rate": 15,
    "mode": "auto"
  }
```

For operator:
Screen Shot 2020-04-23 at 1 04 02 PM
It's no longer 15/min and depends on how many stores in the cluster.

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -16,7 +16,6 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"math"
 	"net/http"
 	"sync"
 	"time"
@@ -943,8 +942,8 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
 		zap.String("store-address", newStore.GetAddress()))
 	err := c.putStoreLocked(newStore)
 	if err == nil {
-		// set the store limit to unlimit
-		c.coordinator.opController.SetStoreLimit(store.GetID(), float64(math.MaxInt64), storelimit.Auto, storelimit.RegionRemove)
+		// set the remove peer limit of the store to unlimited
+		c.coordinator.opController.SetStoreLimit(store.GetID(), storelimit.Unlimited, storelimit.Manual, storelimit.RegionRemove)
 	}
 	return err
 }
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -16,6 +16,7 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"math"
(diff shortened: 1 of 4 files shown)


server/cluster/cluster.go
```diff
@@ -938,7 +939,12 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
 	log.Warn("store has been offline",
 		zap.Uint64("store-id", newStore.GetID()),
 		zap.String("store-address", newStore.GetAddress()))
-	return c.putStoreLocked(newStore)
+	err := c.putStoreLocked(newStore)
+	if err == nil {
+		// set the store limit to unlimit
+		c.coordinator.opController.SetStoreLimit(store.GetID(), float64(math.MaxInt64), storelimit.Auto, storelimit.RegionRemove)
```
shafreeck, line 945: Set the limit mode to manual to avoid the value being changed according to the cluster load.

rleungx: But here is not set by manually, if we use manual, will it be confused? Or do you have a better idea? (I know it will conflict with the cluster load mechanism)

shafreeck: We can regard it as manual as long as it is not adjusted by PD automatically. The unlimited value can regard as the default value of an offline store?

lhy1024, line 945: do we need to add a test to check it?

rleungx:
PTAL @shafreeck @lhy1024

tests/server/cluster/cluster_test.go
```diff
@@ -856,3 +857,87 @@ func getClusterConfig(c *C, clusterID uint64, grpcPDClient pdpb.PDClient) *metap
 
 	return resp.GetCluster()
 }
+
+func (s *clusterTestSuite) TestOfflineStoreLimit(c *C) {
```
lhy1024, line 861: ci failed

PR #2380, opened 2020-04-24
go-tools: add go-deadlock dep

HunDunDM:

We used `go-deadlock` in the test but did not add it as a dependency on `go.mod`.

Add it as a dependency on `go.mod`.

### Check List 

Tests 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ cache:
   directories:
     - $HOME/.cache/go-build
     - $GOPATH/pkg/mod
-    - $GOPATH/src/github.com/pingcap/pd/.retools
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ cache:
     - $HOME/.cache/go-build
     - $GOPATH/pkg/mod
     - $GOPATH/src/github.com/pingcap/pd/.retools
```
rleungx, line 7: Do we still need it?

HunDunDM: yes

lhy1024: it will be better to add comment here

PR #2389, opened 2020-04-28
Improve the heartbeat processing by store ShallowClone.

mantuliu:
When the store clone,add a method called ShallowClone,it won't clone the meta
Improve the  heartbeat processing 
https://github.com/pingcap/pd/issues/2391

When the store clone,add a method called ShallowClone,it won't clone the meta.

---|---|---|---|---

Tests 

 - Unit test

### DIFF
--- a/server/core/store_test.go
+++ b/server/core/store_test.go
@@ -95,36 +95,6 @@ func (s *testConcurrencySuite) TestCloneStore(c *C) {
 	wg.Wait()
 }
 
-func (s *testConcurrencySuite) TestShallowCloneStore(c *C) {
-	meta := &metapb.Store{Id: 1, Address: "mock://tikv-1", Labels: []*metapb.StoreLabel{{Key: "zone", Value: "z1"}, {Key: "host", Value: "h1"}}}
-	store := NewStoreInfo(meta)
-	start := time.Now()
-	wg := sync.WaitGroup{}
-	wg.Add(2)
-	go func() {
-		defer wg.Done()
-		for {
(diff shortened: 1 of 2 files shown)


server/core/store_test.go
```diff
@@ -95,6 +95,36 @@ func (s *testConcurrencySuite) TestCloneStore(c *C) {
 	wg.Wait()
 }
 
+func (s *testConcurrencySuite) TestShallowCloneStore(c *C) {
```
rleungx, line 98: What is the purpose of this test?

mantuliu: I refer to the TestCloneStore test just to test the new ShallowClone method.

rleungx: `TestCloneStore` is used to test race problems for some create options like `SetStoreState`, you can take a look at https://github.com/pingcap/pd/pull/1802. I think there is no need to test it again? Or you can add a bench test for `ShallowClone` and `Clone` when executing `UpdateStoreStatus`, what do you think?

nolouch:
thanks! please pick up this issue: https://github.com/pingcap/pd/issues/2391

server/core/store_test.go
mantuliu: Thanks.I remove the method of TestShallowCloneStore. I didn’t add bench test, because in performance, method ShallowClone is better than method Clone.  And This has been verified by the pd-heartbeat-bench test,  what do you think?

mantuliu:
> thanks! please pick up this issue: #2391 for this PR.

done.thanks

PR #2390, opened 2020-04-28
scheduler: add hot peer filter types

rleungx:
Currently, the hot region scheduling does not have a good global view.

This PR adds hot peer filter types to let the threshold of hot peer filter be more reasonable.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -78,6 +78,8 @@ const (
 	divisor                   = float64(1000) * 2
 	hotWriteRegionMinFlowRate = 16 * 1024
 	hotReadRegionMinFlowRate  = 128 * 1024
+	hotWriteRegionMinKeyRate  = 256
+	hotReadRegionMinKeyRate   = 512
 )
 
 // schedulePeerPr the probability of schedule the hot peer.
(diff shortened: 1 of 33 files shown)


server/statistics/hot_cache.go
```diff
@@ -15,38 +15,133 @@ package statistics
...
+	hotReadRegionMinFlowRate   = 128 * 1024
+)
+
 // HotCache is a cache hold hot regions.
 type HotCache struct {
-	writeFlow *hotPeerCache
-	readFlow  *hotPeerCache
+	writeFlow     *hotPeerCache
+	readFlow      *hotPeerCache
+	hotWriteCache cache.Cache
```
nolouch, line 39: I think we can use the old `hotPeerCache` directly.

nolouch:
PTAL @lhy1024

server/schedulers/hot_region.go
```diff
@@ -74,6 +74,12 @@ const (
 
 	minHotScheduleInterval = time.Second
 	maxHotScheduleInterval = 20 * time.Second
+
+	divisor                   = float64(1000) * 2
```
lhy1024, line 78: do we need to config `divisor`?

rleungx: We can make it configurable if it is needed in the future.

rleungx:
PTAL @lhy1024 @nolouch

server/schedulers/hot_region.go
```diff
@@ -194,36 +200,67 @@ func (h *hotScheduler) prepareForBalance(cluster opt.Cluster) {
...
+		hotRegionThreshold := getHotRegionThreshold(storesStat, write)
 		h.stLoadInfos[writeLeader] = summaryStoresLoad(
 			storeByte,
 			storeKey,
 			h.pendingSums[writeLeader],
 			regionWrite,
 			minHotDegree,
-			write, core.LeaderKind)
+			hotRegionThreshold,
+			write, core.LeaderKind, mixed)
```
lhy1024, line 226: why mixed? It seemed to return false directly.

rleungx:
PTAL @nolouch @lhy1024

server/schedulers/hot_region.go
nolouch, line 78: how about make it configurable.

server/schedulers/hot_region.go
```diff
@@ -1123,6 +1181,14 @@ func (rw rwType) String() string {
 	}
 }
 
+type hotPeerFilterType int
+
+const (
+	high hotPeerFilterType = iota
```
nolouch, line 1187: be a better name?

rleungx: Any suggestions?

nolouch: let me change it in the next PR.

PR #2393, opened 2020-04-29
replication_mode: fix config issues

### LINKED ISSUE #2386: updating label key should increase replication version
Otherwise, there is no way to tell if the region is consistency over the old label key or the new label key.

### LINKED ISSUE #2387: cannot set config `dr-auto-sync.primary-replicas`
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
```
❯ ./bin/pd-ctl config set replication-mode dr-auto-sync primary-replicas 1
Failed! [400] {
  "code": "input",
  "msg": "json: cannot unmarshal string into Go struct field DRAutoSyncReplicationConfig.dr-auto-sync.primary-replicas of type int",
  "data": {
    "Value": "string",
    "Type": {},
    "Offset": 39,
    "Struct": "DRAutoSyncReplicationConfig",
    "Field": "dr-auto-sync.primary-replicas"
  }
}


```

set the primary-replicas successfully.

3. What did you see instead?

meet the error.

,master

### LINKED ISSUE #2388: `dr_auty_sync` can easily be misconfigured
The correct config is:
```
  "replication-mode": "dr_auto_sync",
  "dr-auto-sync": {
    "label-key": "zone",
    "primary": "z1",
    "dr": "z2",
    "primary-replicas": 2,
    "dr-replicas": 1,
    "wait-store-timeout": "1m0s",
    "wait-sync-timeout": "1m0s"
  }
```
as we can see, it uses both `dr_auto_sync` and `dr-auto-sync`. It is not good.

disksing:
Fix #2386 
Fix #2387 
Fix #2388 

- change `dr_auto_sync` to `dr-auto-sync` (and auto convert `_` to `-`)
- validate config, if wrong return error
- support pd-ctl set numbers (e.g. primary-replicas)
- switch status to sync_recover when the label key is updated

### Check List 

Tests 
 - Unit test

### DIFF
--- a/server/replication/replication_mode.go
+++ b/server/replication/replication_mode.go
@@ -97,8 +97,10 @@ func NewReplicationModeManager(config config.ReplicationModeConfig, storage *cor
 func (m *ModeManager) UpdateConfig(config config.ReplicationModeConfig) error {
 	m.Lock()
 	defer m.Unlock()
-	// Handle 'majority -> dr-auto-sync' as special case. (sync_recover mode)
-	if m.config.ReplicationMode == modeMajority && config.ReplicationMode == modeDRAutoSync {
+	// If mode change from 'majority' to 'dr-auto-sync', or the label key is
+	// updated, switch to 'sync_recover' state.
+	if (m.config.ReplicationMode == modeMajority && config.ReplicationMode == modeDRAutoSync) ||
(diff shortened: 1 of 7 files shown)


server/replication/replication_mode.go
```diff
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pingcap/kvproto/pkg/replication_modepb"
```
HunDunDM, line 24: Duplicate import

PR #2394, opened 2020-04-30
config: make max reset ts gap readable

### LINKED ISSUE #2365: Config item `max-reset-ts-gap` is not readable
1. What did you do?
Use pd-ctl `config show all`

```
  "pd-server": {
    "use-region-storage": "true",
    "max-reset-ts-gap": "24h"
    "key-type": "table",
    "runtime-services": "",
    "metric-storage": "",
    "dashboard-address": "http://127.0.0.1:2379"
  }
```

3. What did you see instead?
```
  "pd-server": {
    "use-region-storage": "true",
    "max-reset-ts-gap": 86400000000000,
    "key-type": "table",
    "runtime-services": "",
    "metric-storage": "",
    "dashboard-address": "http://127.0.0.1:2379"
  }
```

master (d0193e980ec3dd7ee2a3ff94b586bbc6c0ce2de6)

rleungx:
Fix #2365.

This PR uses `typeutil.Duration` instead of `time.Duration`. 

### Release note 
Fix the issue that the config item of the max gap to reset ts is not readable

### Check List 

Tests 

 - Unit test

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -912,7 +912,7 @@ type PDServerConfig struct {
 	// UseRegionStorage enables the independent region storage.
 	UseRegionStorage bool `toml:"use-region-storage" json:"use-region-storage,string"`
 	// MaxResetTSGap is the max gap to reset the tso.
-	MaxResetTSGap time.Duration `toml:"max-reset-ts-gap" json:"max-reset-ts-gap"`
+	MaxResetTSGap typeutil.Duration `toml:"max-gap-reset-ts" json:"max-gap-reset-ts"`
 	// KeyType is option to specify the type of keys.
 	// There are some types supported: ["table", "raw", "txn"], default: "table"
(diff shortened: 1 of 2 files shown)


server/config/config.go
```diff
@@ -912,7 +912,7 @@ type PDServerConfig struct {
 	// UseRegionStorage enables the independent region storage.
 	UseRegionStorage bool `toml:"use-region-storage" json:"use-region-storage,string"`
 	// MaxResetTSGap is the max gap to reset the tso.
-	MaxResetTSGap time.Duration `toml:"max-reset-ts-gap" json:"max-reset-ts-gap"`
+	MaxResetTSGap typeutil.Duration `toml:"max-gap-reset-ts" json:"max-gap-reset-ts"`
```
overvenus, line 915: Does it break compatibility?

Issue #2395, opened 2020-04-30
UCP: Reduce the unnecessary updateStoreStatusLocked call counts in heartbeat process


mantuliu:
## Score
- 200

## Mentor
- @nolouch 

## Recommended Skills
- go

## Learning Materials
Ref #1846

nolouch:
thanks!

you06:
@mantuliu try again please~

nolouch:
Is this issue already finished? @mantuliu

mantuliu:
@nolouch already finished.

PR #2396, opened 2020-04-30
Reduce the unnecessary updateStoreStatusLocked call counts.

mantuliu:
UCP #2395 

Reduce the call counts of updateStoreStatusLocked.

---|---|---|---|---

Tests 

 - Unit test

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -569,13 +569,17 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		}
 
 		// Update related stores.
+		storeMap := make(map[uint64]uint64)
+		for _, p := range region.GetPeers() {
+			storeMap[p.GetStoreId()] = p.GetStoreId()
+		}
 		if origin != nil {
 			for _, p := range origin.GetPeers() {
-				c.updateStoreStatusLocked(p.GetStoreId())
+				storeMap[p.GetStoreId()] = p.GetStoreId()
 			}
 		}
-		for _, p := range region.GetPeers() {
-			c.updateStoreStatusLocked(p.GetStoreId())
+		for key := range storeMap {
+			c.updateStoreStatusLocked(key)
 		}
 		regionEventCounter.WithLabelValues("update_cache").Inc()
 	}
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -569,13 +569,13 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		}
 
 		// Update related stores.
-		storeMap := make(map[uint64]uint64)
+		storeMap := make(map[uint64]struct{})
 		for _, p := range region.GetPeers() {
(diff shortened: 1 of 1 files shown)


server/cluster/cluster.go
```diff
@@ -569,13 +569,17 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		}
 
 		// Update related stores.
+		storeMap := make(map[uint64]uint64)
```
rleungx, line 572: How about `storeMap := make(map[uint64]struct{})`?

mantuliu: Done. Thanks.

mantuliu:
> Hi, @mantuliu I test with
> 
> ```
>  ./bin/pd-heartbeat-bench -region-update-ratio=0.1 --heartbeat-rounds=10
> ```
> 
> only got
> 
> ```
> Summary:
>   Total:        15.1577 secs.
>   Slowest:      0.3920 secs.
>   Fastest:      0.0000 secs.
>   Average:      0.0003 secs.
>   Stddev:       0.0082 secs.
>   Requests/sec: 65974.2556
> ```
> 
> In my test environment, I do not see about 7.8% imporve vs master, the QPS roughtly same. Can you tell me about your test method?
> BTW, the benchmark only updates region with [increase the version](https://github.com/pingcap/pd/blob/master/tools/pd-heartbeat-bench/main.go#L163-L165), do not change the peer. it's may affect the result.

Hi @nolouch My test like this:
```
var (
	pdAddr            = flag.String("pd", "127.0.0.1:2379", "pd address")
	storeCount        = flag.Int("store", 20, "store count")
	regionCount       = flag.Uint64("region", 1000000, "region count")
	keyLen            = flag.Int("keylen", 56, "key length")
	replica           = flag.Int("replica", 3, "replica count")
	regionUpdateRatio = flag.Float64("region-update-ratio", 0.1, "the ratio of the region need to update")
	sample            = flag.Bool("sample", false, "sample per second")
	heartbeatRounds   = flag.Int("heartbeat-rounds", 10, "the total rounds of hearbeat")
)
```
And I used the pd-heartbeat-bench of the master branch.
Calculation formula: 7.8% = (132076 - 124109)/101652.
My computer hardware like this:
```
macOS 10.15.3
cpu:2.3GHz 8 intel Core i9
memory: 32G 2667 MHz DDR4
storage: Macintosh HD 1T
```

mantuliu:
The result like this:
Summary:
  Total:        7.6203 secs.
  Slowest:      0.7093 secs.
  Fastest:      0.0000 secs.
  Average:      0.0001 secs.
  Stddev:       0.0054 secs.
  Requests/sec: 131231.3358

mantuliu:
> Is this result the same for each round? From my test, there are some errors every time.

Not all rounds are the same in my test environment. What kind of error is it?

rleungx:
I mean the different results from my testing, sometimes the master branch is better.

rleungx:
But from the code perspective, this is indeed an improvement, LGTM.

PR #2398, opened 2020-05-02
Improve the heartbeat processing by not remove region from subtrees.

mantuliu:
UCP https://github.com/pingcap/pd/issues/2392

When removing the region, try not to remove the region from the subTree,

---|---|---|---|---
3|try not to remove the region from the subTree.|   region.go 580-584: shouldRemoveFromSubTree,  | 147363   |     22.9%

Tests 

 - Unit test

### DIFF
--- a/server/core/region.go
+++ b/server/core/region.go
@@ -579,7 +579,10 @@ func (r *RegionsInfo) GetRegion(regionID uint64) *RegionInfo {
 // SetRegion sets the RegionInfo with regionID
 func (r *RegionsInfo) SetRegion(region *RegionInfo) []*RegionInfo {
 	if origin := r.regions.Get(region.GetID()); origin != nil {
-		r.RemoveRegion(origin)
+		r.removeRegionFromTreeAndMap(origin)
+		if r.shouldRemoveFromSubTree(region, origin) {
+			r.removeRegionFromSubTree(origin)
+		}
 	}
 	return r.AddRegion(region)
 }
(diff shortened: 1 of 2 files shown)


nolouch:
PTAL @rleungx @lhy1024

server/core/region.go
```diff
@@ -670,6 +685,50 @@ func (r *RegionsInfo) RemoveRegion(region *RegionInfo) {
 	}
 }
 
+// shouldRemoveFromSubTree return true when the region leader changed, peer transferred,
+// new peer was created, learners changed, pendingPeers changed, and so on.
+func (r *RegionsInfo) shouldRemoveFromSubTree(region *RegionInfo, origin *RegionInfo) bool {
+	checkPeersChange := func(origin []*metapb.Peer, other []*metapb.Peer) bool {
+		checkPeersDiff := func(origin []*metapb.Peer, other []*metapb.Peer) bool {
```
rleungx, line 692: Do we have a better way (e.g. maybe sort them and compare them?) to do this check so that we can use something like `checkPeersDiff(origin, other)` rather than `checkPeersDiff(origin, other) || checkPeersDiff(other, origin)` which seems like do the same thing?

server/core/region.go
```diff
@@ -670,6 +685,50 @@ func (r *RegionsInfo) RemoveRegion(region *RegionInfo) {
...
+		}
+		if checkPeersDiff(origin, other) || checkPeersDiff(other, origin) {
+			return true
+		}
+		return false
+	}
+
+	remove := false
+	if origin.leader.GetId() != region.leader.GetId() {
+		remove = true
```
lhy1024, line 721: how about return directly

rleungx:
@mantuliu You can reopen this PR if you have already picked up the issue.

mantuliu:
@rleungx @nolouch In the interface I see, there is no reopen-button.

mantuliu:
I think I may not have permission to reopen.

PR #2400, opened 2020-05-07
Improve the heartbeat processing by not remove region from subtrees.

mantuliu:
UCP https://github.com/pingcap/pd/issues/2392

When removing the region, try not to remove the region from the subTree,

---|---|---|---|---
3|try not to remove the region from the subTree.|   region.go 580-584: shouldRemoveFromSubTree,  | 147363   |     22.9%

Tests 

 - Unit test


mantuliu:
@rleungx > Do we have a better way (e.g. maybe sort them and compare them?) to do this check so that we can use something like `checkPeersDiff(origin, other)` rather than `checkPeersDiff(origin, other) || checkPeersDiff(other, origin)` which seems like do the same thing?

OK.Let me try.

mantuliu:
@lhy1024 > how about return directly

OK.

mantuliu:
#2398

server/core/region.go
```diff
@@ -716,17 +716,19 @@ func (r *RegionsInfo) shouldRemoveFromSubTree(region *RegionInfo, origin *Region
...
-		remove = true
-	} else if checkPeersChange(origin.GetVoters(), region.GetVoters()) {
-		remove = true
-	} else if checkPeersChange(origin.GetLearners(), region.GetLearners()) {
-		remove = true
-	} else if checkPeersChange(origin.GetPendingPeers(), region.GetPendingPeers()) {
-		remove = true
-	}
-	return remove
+		return true
```
lhy1024, line 720:
```suggestion
		return origin.leader.GetId() != region.leader.GetId() || checkPeersChange(origin.GetVoters(), region.GetVoters()) || checkPeersChange(origin.GetLearners(), region.GetLearners())||checkPeersChange(origin.GetPendingPeers(), region.GetPendingPeers()) 
```

lhy1024: I meant something  below

mantuliu: I think the present way of writing if separately is clearer and easier to understand. What do you think?^_^

mantuliu:
@lhy1024 @rleungx PTAL. Thanks.

server/core/region.go
```diff
@@ -670,6 +686,42 @@ func (r *RegionsInfo) RemoveRegion(region *RegionInfo) {
...
+		for index, peer := range origin {
+			if peer.GetStoreId() == other[index].GetStoreId() && peer.GetId() == other[index].GetId() {
+				continue
+			}
+			return true
+		}
+		return false
+	}
+
+	return origin.leader.GetId() != region.leader.GetId() ||
```
rleungx, line 719: When the size or key number of a region changes, does it also need to be removed from subtree?

mantuliu: This situation does not need to be removed from the subtree. Because there is processing logic about size and keys number in the (rst *regionSubTree) update method of region.go;

```
func (rst *regionSubTree) update(region *RegionInfo) {
	if r := rst.find(region); r != nil {
		rst.totalSize += region.approximateSize - r.region.approximateSize
		rst.totalKeys += region.approximateKeys - r.region.approximateKeys
		r.region = region
		return
	}
	rst.totalSize += region.approximateSize
	rst.totalKeys += region.approximateKeys
	rst.regionTree.update(region)
}
```

nolouch: ptal @rleungx

lhy1024:
lgtm ptal @rleungx

server/core/region.go
rleungx: The size and key number that I mentioned is not a good example. But you can see when we handle the region's heartbeat, there are several cases that we need to update the cache like flow changes and so on. The region will be removed from the subtree in an original way, but it won't now and some function like `GetStoreRegions` may be affected?

PR #2407, opened 2020-05-09
*: make the help message standard

rleungx:
The format of our help message is not unified.

After referring to some Linux commands like `cat`, `ls`, etc. , they use lower case and without periods in the help message. This PR makes changes to all binaries in PD repo.

### Check List 

Tests 

 - No code

### Release note 
- No release note


tools/pd-tso-bench/main.go
```diff
@@ -33,9 +33,9 @@ var (
 	pdAddrs     = flag.String("pd", "127.0.0.1:2379", "pd address")
 	concurrency = flag.Int("C", 1000, "concurrency")
```
lhy1024, line 34: how about lower case

rleungx: This PR will only modify the help message. We can raise another PR to do it if it's necessary.

PR #2424, opened 2020-05-15
Improve the heartbeat processing by not remove region from the tree

mantuliu:
UCP: #2431 

When removing the region, try not to remove the region from the region tree,

---|---|---|---|---
4|try not to remove the region from the subTree.|   region.go 580-584: shouldRemoveFromSubTree,  | 151024   |     18.6%
5|try not to remove the region from the tree.|   region.go 610-629: AddRegion,  | 168702   |     17.4%

Tests 

 - Unit test

### DIFF
--- a/server/core/region.go
+++ b/server/core/region.go
@@ -580,7 +580,7 @@ func (r *RegionsInfo) GetRegion(regionID uint64) *RegionInfo {
 // SetRegion sets the RegionInfo with regionID
 func (r *RegionsInfo) SetRegion(region *RegionInfo) []*RegionInfo {
 	if origin := r.regions.Get(region.GetID()); origin != nil {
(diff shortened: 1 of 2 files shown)


server/core/region.go
```diff
@@ -605,12 +607,27 @@ func (r *RegionsInfo) GetOverlaps(region *RegionInfo) []*RegionInfo {
 
 // AddRegion adds RegionInfo to regionTree and regionMap, also update leaders and followers by region peers
 func (r *RegionsInfo) AddRegion(region *RegionInfo) []*RegionInfo {
-	// Add to tree and regions.
-	overlaps := r.tree.update(region)
-	for _, item := range overlaps {
-		r.RemoveRegion(r.GetRegion(item.GetID()))
+	var overlaps []*RegionInfo
+	treeNeedAdd := true
```
nolouch, line 611: Could you add a comment about the variable.

mantuliu: Done, thanks.

PR #2425, opened 2020-05-16
Improve the heartbeat processing by reduce the number of getOverlaps.

mantuliu:
UCP: #2430 

When call the method of PreCheckPutRegion, reduce the number of calls to the method getOverlaps.

---|---|---|---|---
2|use ShallowClone when UpdateStoreStatus.|   store.go 666: newStore := store.ShallowClone(),  | 124109   |     22%
4|try not to remove the region from the subTree.|   region.go 580-584: shouldRemoveFromSubTree,  | 151024   |     18.6%
5|try not to remove the region from the tree.|   region.go 610-629: AddRegion,  | 168702   |     17.4%
6|reduce the number of calls to the method getOverlaps.|   basic_cluster.go 292-297: PreCheckPutRegion,  | 203152   |     33.9%

Tests 

 - Unit test

### DIFF
--- a/server/core/basic_cluster.go
+++ b/server/core/basic_cluster.go
@@ -14,6 +14,7 @@
 package core
 
 import (
+	"bytes"
 	"sync"
 
 	"github.com/pingcap/kvproto/pkg/metapb"
@@ -288,13 +289,15 @@ func (bc *BasicCluster) TakeStore(storeID uint64) *StoreInfo {
 // PreCheckPutRegion checks if the region is valid to put.
 func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, error) {
 	bc.RLock()
-	for _, item := range bc.Regions.GetOverlaps(region) {
-		if region.GetRegionEpoch().GetVersion() < item.GetRegionEpoch().GetVersion() {
-			bc.RUnlock()
-			return nil, ErrRegionIsStale(region.GetMeta(), item.GetMeta())
+	origin := bc.Regions.GetRegion(region.GetID())
+	if origin == nil || !bytes.Equal(origin.GetStartKey(), region.GetStartKey()) || !bytes.Equal(origin.GetEndKey(), region.GetEndKey()) {
+		for _, item := range bc.Regions.GetOverlaps(region) {
+			if region.GetRegionEpoch().GetVersion() < item.GetRegionEpoch().GetVersion() {
+				bc.RUnlock()
+				return nil, ErrRegionIsStale(region.GetMeta(), item.GetMeta())
+			}
 		}
 	}
-	origin := bc.Regions.GetRegion(region.GetID())
 	bc.RUnlock()
 	if origin == nil {
 		return nil, nil


server/core/basic_cluster.go
```diff
@@ -288,13 +289,15 @@ func (bc *BasicCluster) TakeStore(storeID uint64) *StoreInfo {
 // PreCheckPutRegion checks if the region is valid to put.
 func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, error) {
 	bc.RLock()
-	for _, item := range bc.Regions.GetOverlaps(region) {
-		if region.GetRegionEpoch().GetVersion() < item.GetRegionEpoch().GetVersion() {
-			bc.RUnlock()
-			return nil, ErrRegionIsStale(region.GetMeta(), item.GetMeta())
+	origin := bc.Regions.GetRegion(region.GetID())
+	if origin == nil || !bytes.Equal(origin.GetStartKey(), region.GetStartKey()) || !bytes.Equal(origin.GetEndKey(), region.GetEndKey()) {
```
rleungx, line 293: Does it improve performance when there are lots of merge or split operations (will have a lot of overlaps) since we need to compare it one more time than before?

mantuliu: In response to this problem, I conducted 3 tests:
Test 1: master branch

Test 2: Update the 'if condition' like this:
```
If origin == nil ||! bytes.Equal (origin.GetStartKey (), region.GetStartKey ()) ||! bytes.Equal (origin.GetEndKey (), region.GetEndKey ()) || true {
```

Test 3: Update the 'if conditions' like this:
```
If origin == nil ||! bytes.Equal (origin.GetStartKey (), region.GetStartKey ()) ||! bytes.Equal (origin.GetEndKey (), region.GetEndKey ()) || region.GetID ()% 100 == 0 {
```
The performance of Test 2 and Test 1 is almost the same, so it is proved that adding a comparison condition in the PR has no effect on performance

In the Test 3,  there war 1 million regions, 10,000 of them must be checked by getOverlaps. This test simulates the merge or split operation of a large number of regions in the actual project. The heartbeat processing ability of Test 3 is better than Test 1 and Test 2 a lot of.

Therefore, the effectiveness of this PR can be proved.

mantuliu:
@nolouch I test the master branch, and failed too. The message like this:
```
PANIC: client_test.go:479: testClientSuite.TestGetRegionByID

wait start
... Panic: runtime error: invalid memory address or nil pointer dereference (PC=0x4031F81)

/usr/local/go/src/runtime/panic.go:679
  in gopanic
/usr/local/go/src/runtime/panic.go:199
  in panicmem
/usr/local/go/src/runtime/signal_unix.go:394
  in sigpanic
client_test.go:500
  in testClientSuite.TestGetRegionByID.func1
/Users/liuguangliang/go/src/github.com/pingcap/pd/pkg/testutil/testutil.go:39
  in WaitUntil
client_test.go:497
  in testClientSuite.TestGetRegionByID
/usr/local/go/src/reflect/value.go:321
  in Value.Call
/Users/liuguangliang/go/pkg/mod/github.com/pingcap/check@v0.0.0-20191216031241-8a5a85928f12/check.go:850
  in suiteRunner.forkTest.func1
/Users/liuguangliang/go/pkg/mod/github.com/pingcap/check@v0.0.0-20191216031241-8a5a85928f12/check.go:739
  in suiteRunner.forkCall.func1
/usr/local/go/src/runtime/asm_amd64.s:1357
```

PR #2426, opened 2020-05-18
*: add engine filter to exclude special stores

nolouch:

Fix https://github.com/pingcap/pd/issues/2423
`AvaiableStore()` always returns flash nodes, so the tikv nodes cannot be used.

### Release note 
- Fix the issue that `presplit` not work within tiflash cluster.


server/schedule/filter/filters.go
```diff
@@ -543,6 +543,41 @@ func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
 	return placement.CompareRegionFit(f.oldFit, newFit) <= 0
 }
 
+type engineFilter struct {
+	scope      string
+	constraint placement.LabelConstraint
+}
+
+// NewEngineFilter creates a filter that filters store for special engine.
+func NewEngineFilter(scope string, allowUses ...string) Filter {
```
HunDunDM, line 552: Can constraint.Values be specified directly in the constructor?

nolouch: I have adds some comments.

server/schedule/filter/filters.go
```diff
@@ -543,12 +543,51 @@ func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
 	return placement.CompareRegionFit(f.oldFit, newFit) <= 0
 }
 
+type engineFilter struct {
+	scope      string
+	constraint placement.LabelConstraint
+}
+
+// NewEngineFilter creates a filter that filters out default engine stores.
+// By default, all stores that are not marked with a special engine will be filtered out.
```
lhy1024, line 552: all tikv node will be filtered out?

server/schedule/filter/filters.go
```diff
@@ -543,12 +543,51 @@ func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
...
+// Specify the special engine label if you want to include the special stores.
+func NewEngineFilter(scope string, allowSpeicalEngines ...string) Filter {
```
rleungx, line 554: There are too many inverse logic and it's hard to understand.

server/schedule/filter/filters.go
nolouch: yes.

server/schedule/filter/filters.go
nolouch: any suggestion?

nolouch:
It's about dashboard, merge it firstly.

PR #2453, opened 2020-05-25
schedule: separate region limit from merge limit

### LINKED ISSUE #2432: Separate the effect of `balance-region-limit` on balance region and merge region
the problem is if we add a store, but we also have many empty regions that need to be merge. and the limit is:
```
region-schedule-limit: 4
merge-schedule-limit: 8 
```
`Region Merge` will have more chance to schedule, `Balance Region` only have a small chance.

Now, we suggest that `region-schedule-limit` large than `merge-schedule-limit`.  But this is not a good solution.
 
Let them do not affect each other.

rleungx:
Close #2432 

This PR checks if it is a merge operator in `buildSteps` and doesn't append `OpRegion` if it is a merge operator. Also, this PR removes the unnecessary `OpKind`: `OpBalance`.

### Check List 

Tests 

 - Unit test

### Release note 
- Separate region schedule limit from merge schedule limit


server/schedulers/balance_test.go
```diff
@@ -314,22 +314,22 @@ func (s *testBalanceLeaderSchedulerSuite) TestBalanceFilter(c *C) {
 	s.tc.AddLeaderStore(4, 16)
 	s.tc.AddLeaderRegion(1, 4, 1, 2, 3)
 
-	testutil.CheckTransferLeader(c, s.schedule()[0], operator.OpBalance, 4, 1)
+	testutil.CheckTransferLeader(c, s.schedule()[0], operator.OpKind(0), 4, 1)
```
lhy1024, line 317: opRegion?

rleungx: It doesn't matter since the original `OpBalance` doesn't work.

server/schedule/operator/builder.go
```diff
@@ -279,7 +279,9 @@ func (b *Builder) buildSteps(kind OpKind) (OpKind, error) {
 		}
 		if plan.add != nil {
 			b.execAddPeer(plan.add)
-			kind |= OpRegion
+			if kind&OpMerge == 0 {
```
nolouch, line 282: Can we change the condition of the `AllowBalance` rather than change this?
Merge may move peers or only transfer leaders.

PR #2462, opened 2020-05-27
Fix the isse that scatter cannot worker after truncate with tifalsh

nolouch:
truncate a table  will do some steps:
- create the new table and pre-split regions 
- tiflash pulls the schema asynchronously to update the rules
- scatter the regions

If the rule is updated but the scheduler does not finish to fit region, it will cause scatter failed.


server/replication/replication_mode.go
```diff
@@ -329,9 +329,16 @@ func (m *ModeManager) tickDR() {
 
 	drTickCounter.Inc()
 
-	canSync := m.checkCanSync()
+	totalPrimary, totalDr := m.config.DRAutoSync.PrimaryReplicas, m.config.DRAutoSync.DRReplicas
```
disksing, line 332: seems not belong to this issue.

server/schedule/operator_controller.go
```diff
@@ -147,7 +147,26 @@ func (oc *OperatorController) Dispatch(region *core.RegionInfo, source string) {
 	}
 }
 
-func (oc *OperatorController) checkStaleOperator(op *operator.Operator, region *core.RegionInfo) bool {
+func (oc *OperatorController) checkStaleOperator(op *operator.Operator, step operator.OpStep, region *core.RegionInfo) bool {
```
disksing, line 150: not belong neither

server/schedule/placement/fit.go
```diff
@@ -43,6 +43,23 @@ func (f *RegionFit) IsSatisfied() bool {
...
 }
 
+// IsSatisfiedLoose returns if the rules are properly satisfied excluded learner.
+func (f *RegionFit) IsSatisfiedLoose() bool {
+	if len(f.RuleFits) == 0 {
+		return false
+	}
+	for _, r := range f.RuleFits {
+		// excluded learner.
+		if r.Rule.Role == Learner {
```
disksing, line 53: maybe check by engine type will be better?

PR #2463, opened 2020-05-27
schedulers: fix cache item in hot peer cache and add more test

### LINKED ISSUE #2468: Cache item maybe missed in hot peer cache
If the region is balanced from a store to another store or the leader is transferred, the region and store info will be expired in `hotPeerCache`.

We statistic the flow by comparing the current time with the previous time, which will cause the existence of zero in the previous time after balance or transfer. In order to avoid sudden fluctuations, we chose to use the region info of other stores as the last time. 

But `tmpItem` will be affected by the expired store. if there are store1, store2, and store3, and move peer to store4, the new replica cannot get cached from all unexpired stores.

lhy1024:

Fix #2468

Prepare the cache item in advance

### Check List 

Tests 

 - Unit test

### Release note 
- No release note


server/statistics/hot_peer_cache.go
```diff
@@ -116,13 +116,17 @@ func (f *hotPeerCache) CheckRegionFlow(region *core.RegionInfo, storesStats *Sto
...
+	var cacheItem *HotPeerStat
+	for _, storeID := range storeIDs {
+		cacheItem = f.getOldHotPeerStat(region.GetID(), storeID)
+		if cacheItem != nil {
+			break
+		}
+	}
+
 	for _, storeID := range storeIDs {
 		isExpired := f.isRegionExpired(region, storeID)
```
rleungx, line 128: isExpired can be moved into line 133

server/statistics/hot_peer_cache.go
```diff
@@ -116,13 +116,17 @@ func (f *hotPeerCache) CheckRegionFlow(region *core.RegionInfo, storesStats *Sto
 	byteRate := totalBytes / float64(interval)
 	keyRate := totalKeys / float64(interval)
 
-	var tmpItem *HotPeerStat
+	var cacheItem *HotPeerStat
+	for _, storeID := range storeIDs {
+		cacheItem = f.getOldHotPeerStat(region.GetID(), storeID)
```
rleungx, line 121: Is it different from the original logic?

nolouch:
Could you add a test?

server/statistics/hot_peer_cache.go
lhy1024: I think we only need to get a cached item whether expired. cc @nolouch

lhy1024:
PTAL @rleungx @nolouch

server/statistics/hot_peer_cache.go
```diff
@@ -116,9 +114,13 @@ func (f *hotPeerCache) CheckRegionFlow(region *core.RegionInfo, storesStats *Sto
...
 	keyRate := totalKeys / float64(interval)
 
+	// old region is in the front and new region is in the back
+	// which ensures it will hit the cache if moving peer or transfer leader occurs with the same replica number
+
 	var tmpItem *HotPeerStat
+	storeIDs := f.getAllStoreIDs(region)
 	for _, storeID := range storeIDs {
-		isExpired := f.isRegionExpired(region, storeID)
+		isExpired := f.isRegionExpired(region, storeID) // transfer leader or remove peer
```
rleungx, line 123: I realized that it seems we can directly update the cache instead of returning the item which needs to be updated?

lhy1024: I will open another PR to merge them.

server/statistics/hot_peer_cache_test.go
```diff
@@ -59,6 +56,151 @@ func (t *testHotPeerCache) TestStoreTimeUnsync(c *C) {
...
+func checkNeedDelete(c *C, ret []*HotPeerStat, storeID uint64) {
+	for _, item := range ret {
+		if item.StoreID == storeID {
+			c.Assert(item.needDelete, IsTrue)
+			return
+		}
+	}
+}
+
+const interval = uint64(60)
```
rleungx, line 138: This variable can be moved into `buildRegion` since other places won't use it.

server/statistics/hot_peer_cache_test.go
```diff
@@ -121,34 +126,43 @@ func checkHit(c *C, cache *hotPeerCache, region *core.RegionInfo, kind FlowKind,
...
+			c.Assert(item.needDelete, IsTrue)
+			return
+		}
+	}
+}
+
 const interval = uint64(60)
 
-func schedule(operator operator, region *core.RegionInfo, kind FlowKind) *core.RegionInfo {
+func schedule(operator operator, region *core.RegionInfo, kind FlowKind) (srcStore uint64, _ *core.RegionInfo) {
```
rleungx, line 140:
```suggestion
func schedule(operator operator, region *core.RegionInfo, kind FlowKind) (uint64, *core.RegionInfo) {
```

lhy1024: I want to indicate the meaning of this return var, I think it is better than putting it in a comment

server/statistics/hot_peer_cache.go
nolouch: why directly update the cache? we need to check it first for reducing heartbeat lock overload.

nolouch:
cc @lhy1024

lhy1024:
PTAL @rleungx

server/statistics/hot_peer_cache_test.go
HunDunDM: > I want to indicate the meaning of this return var, I think it is better than putting it in a comment

I agree

PR #2467, opened 2020-05-27
cluster: add suspect regions cache

### LINKED ISSUE #2422: Region split during scheduling process is not handled properly
Consider we balance a region. The operator is _add learner + promote learner + remove peer_. If the region splits at the first 2 steps, there is only 1 region (which has the original region id) can continue removing the extra peer. Others will have an extra peer, until the replica-checker or rule-checker checks the region.

disksing:

Fix #2422

- `raftCluster` maintains a cache, namely `suspectRegions`. It contains all regions that potentially broken and need to fix
- when region splits, add all sub regions to the cache (we can include more cases later, for example, operator timeout)
- when coordinator patrols all regions, it first checks suspect regions

### Check List 

Tests 
 - Unit test

Related changes
- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:

### Release note 
- No release note

### DIFF
--- a/server/cluster/cluster_worker_test.go
+++ b/server/cluster/cluster_worker_test.go
@@ -19,6 +19,9 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 
 	// Register schedulers
+	"github.com/pingcap/pd/v4/pkg/mock/mockid"
+	"github.com/pingcap/pd/v4/server/core"
+	"github.com/pingcap/pd/v4/server/kv"
 	_ "github.com/pingcap/pd/v4/server/schedulers"
 )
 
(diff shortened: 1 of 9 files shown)


server/cluster/cluster_worker_test.go
```diff
@@ -19,6 +19,9 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 
 	// Register schedulers
```
rleungx, line 21: The comment should be moved

disksing: ok

disksing:
PTAL @rleungx @nolouch @lhy1024

server/cluster/coordinator.go
```diff
@@ -103,6 +103,27 @@ func (c *coordinator) patrolRegions() {
 			return
 		}
 
+		// Check suspect regions first.
+		for _, id := range c.cluster.GetSuspectRegions() {
```
lhy1024, line 107: it seems to be similar to the block below, can we abstract it into a function to avoid redundancy

disksing: em, I think they are not quite alike. In this loop, the region may not exist yet, and we need to remove the id from suspect list in some cases.

PR #2471, opened 2020-05-28
server: Fix bug for region_score with store's used_size is zero

### LINKED ISSUE #2470: If store's used_size is zero, can not get store info from PD
1. What did you do?
Under certain conditions, TiFlash send heartbeat with its used_size = 0.

Can get store info by querying `/pd/api/v1/store/${store-id}` and `/pd/api/v1/stores`.

3. What did you see instead?
Return a string `json: unsupported value: NaN` with status code = 500.

master

JaySon-Huang:
Signed-off-by: JaySon-Huang <tshent@qq.com>

Fix #2470 

If store's used size is zero, just set `amplification = 1` to avoid divide by zero error.

### Check List 

Tests 

Related changes

### Release note 

- Fix bug that can not get store information when store's used size is zero

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -304,7 +304,7 @@ func (s *StoreInfo) RegionScore(highSpaceRatio, lowSpaceRatio float64, delta int
 	available := float64(s.GetAvailable()) / mb
 	used := float64(s.GetUsedSize()) / mb
 
-	if s.GetRegionSize() == 0 {
+	if s.GetRegionSize() == 0 || used == 0 {
 		amplification = 1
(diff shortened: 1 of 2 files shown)


rleungx:
@JaySon-Huang Can you tell more about why the used size is 0?

JaySon-Huang:
@rleungx It is detected by a special version of TiFlash, which report zero bytes used with non-zero regions.
I think that even in this situation, PD should not set region score to `NaN`. Or we can get that store information from API anymore, which may make many other components not work.

PR #2481, opened 2020-05-29
rule: prevent misconfiguration of placement rules in some situations

disksing:
Fix https://github.com/pingcap/pd/issues/2476

when update placement rules, return error if after update:
- there is no rule lest
- there is gap between rules (region may match no rules)

### Check List 

Tests 
 - Unit test

Related changes
- Need to cherry-pick to the release branch: release-4.0

### Release note 
- Prevent misconfiguration of placement rules in some situations


server/schedule/placement/rule_manager.go
```diff
@@ -169,8 +173,13 @@ func (m *RuleManager) SetRule(rule *Rule) error {
...
 
-	if err = m.store.SaveRule(rule.StoreKey(), rule); err != nil {
-		// restore
+	ruleList, err := buildRuleList(m.rules)
+	if err == nil {
+		err = m.store.SaveRule(rule.StoreKey(), rule)
+	}
+
+	// restore
+	if err != nil {
```
rleungx, line 182: else?

disksing: cannot use `else`, `err` may become not nil in if

rleungx: got it.

PR #2517, opened 2020-06-08
config: validate dashboard-address

### LINKED ISSUE #2503: Setting dashboard-address in the wrong format can cause unexpected behavior
Please answer these questions before submitting your issue. Thanks!

1. What did you do?

If possible, provide a recipe for reproducing the error.

* Start a cluster
* Use `pd-ctl` to set the dashboard-address config in the wrong format
* Use `pd-ctl` to set the dashboard-address config in the correct format

The wrong setting will not have effect, the correct setting will take effect.

3. What did you see instead?

The dashboard is not working properly.

```bash
Release Version: v4.0.0
Edition: Community
Git Commit Hash: 56d4c3d2237f5bf6fb11a794731ed1d95c8020c2
Git Branch: HEAD
UTC Build Time:  2020-06-02 07:56:57
```

HunDunDM:

* fix #2503 

* When setting the `pd-server` config, validate the `dashboard-address` and reject the incorrect setting.

### Check List

Tests

- Unit test

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -965,6 +965,20 @@ func (c *PDServerConfig) Clone() *PDServerConfig {
 	}
 }
 
+// Validate is used to validate if some pd-server configurations are right.
+func (c *PDServerConfig) Validate() error {
+	switch c.DashboardAddress {
+	case "auto":
+	case "none":
+	default:
+		if err := ValidateURLWithScheme(c.DashboardAddress); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // StoreLabel is the config item of LabelPropertyConfig.
 type StoreLabel struct {
 	Key   string `toml:"key" json:"key"`
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -948,10 +948,10 @@ func (c *PDServerConfig) adjust(meta *configMetaData) error {
 	if !meta.IsDefined("dashboard-address") {
 		c.DashboardAddress = defaultDashboardAddress
 	}
-	return nil
+	return c.Validate()
 }
 
-// Clone retruns a cloned PD server config.
+// Clone returns a cloned PD server config.
 func (c *PDServerConfig) Clone() *PDServerConfig {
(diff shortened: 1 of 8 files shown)


server/config/config.go
```diff
@@ -965,6 +965,20 @@ func (c *PDServerConfig) Clone() *PDServerConfig {
 	}
 }
 
+// Validate is used to validate if some pd-server configurations are right.
+func (c *PDServerConfig) Validate() error {
+	switch c.DashboardAddress {
+	case "auto":
```
lhy1024, line 971: what mean "auto" and "none"

server/config/config_test.go
```diff
@@ -237,6 +237,71 @@ func newTestScheduleOption() (*PersistOptions, error) {
...
+	tests := []struct {
+		cfgData          string
+		hasErr           bool
+		dashboardAddress string
+	}{
+		{
+			`
+[pd-server]
+dashboard-address = "http://127.0.0.1:2379"
+`,
```
lhy1024, line 250: format code, plz

server/config/config.go
HunDunDM: A specific address represents a specific instance of the running dashboard. `none` represents no instance that has dashboard running, which means that the dashboard is stopped. "auto" means that PD can later decide on an instance of the running dashboard, and after PD is run, it will be replaced with a specific address.

HunDunDM: [Chinese Docs](https://docs.pingcap.com/zh/tidb/v4.0/dashboard-ops-deploy#%E5%88%87%E6%8D%A2%E5%85%B6%E4%BB%96-pd-%E5%AE%9E%E4%BE%8B%E6%8F%90%E4%BE%9B-tidb-dashboard-%E6%9C%8D%E5%8A%A1)

lhy1024: do we need to comment on them to code?

HunDunDM: There is a comment at the definition: `// There are some values supported: "auto", "none", or a specific address, default: "auto"`

PR #2519, opened 2020-06-08
*: cleanup go.mod and Makefile

HunDunDM:

* cleanup go.mod and Makefile
* update `golangci-lint` and fix some issues of lint

### Check List

Tests

- No code


HunDunDM:
According to [the cmd of staticcheck](https://github.com/dominikh/go-tools/blob/master/cmd/staticcheck/staticcheck.go), we can see that `stylecheck` is also used. After adding `stylecheck` to the golangci-lint configuration file, staticcheck can be removed safely.

PR #2530, opened 2020-06-11
operator: fast-fail if the leader changed to the RemovePeer

nolouch:

Fix https://github.com/pingcap/pd/issues/2493
If the leader changed cause by TiKV(maybe too busy or network problem), the operator cannot step to the final state, such as
```
PD:
try to remove peer 24
TiKV:
reject remove peer 24 because it is a leader.
``` 
Then the operator needs 10mins to timeout, which may cause the rule sync learner became slow.

### Check List

Tests

- Unit test

### Release note

- Let the operator fast fail if the leader changed to the RemovePeer

### DIFF
--- a/server/schedule/operator/step.go
+++ b/server/schedule/operator/step.go
@@ -162,7 +162,7 @@ func (al AddLearner) CheckSafety(region *core.RegionInfo) error {
 	if peer.GetId() != al.PeerID {
 		return errors.Errorf("peer id is not matched, expect %d, got %d", al.PeerID, peer.GetId())
 	}
-	if peer.IsLearner {
+	if !peer.IsLearner {
 		return errors.New("peer already is a voter")
 	}
 	return nil
@@ -424,7 +424,7 @@ func (al AddLightLearner) CheckSafety(region *core.RegionInfo) error {
 	if peer.GetId() != al.PeerID {
 		return errors.Errorf("peer id is not matched, expect %d, got %d", al.PeerID, peer.GetId())
 	}
-	if peer.IsLearner {
+	if !peer.IsLearner {
 		return errors.New("peer already is a voter")
 	}
 	return nil
--- a/server/schedule/operator/step.go
+++ b/server/schedule/operator/step.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/pingcap/pd/v4/server/core"
 	"github.com/pingcap/pd/v4/server/schedule/storelimit"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
@@ -30,6 +31,7 @@ type OpStep interface {
 	fmt.Stringer
 	ConfVerChanged(region *core.RegionInfo) bool
 	IsFinish(region *core.RegionInfo) bool
+	CheckSafety(region *core.RegionInfo) error
 	Influence(opInfluence OpInfluence, region *core.RegionInfo)
 }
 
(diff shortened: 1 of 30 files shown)


server/schedule/operator_controller.go
```diff
@@ -147,7 +147,26 @@ func (oc *OperatorController) Dispatch(region *core.RegionInfo, source string) {
 	}
 }
 
-func (oc *OperatorController) checkStaleOperator(op *operator.Operator, region *core.RegionInfo) bool {
+func (oc *OperatorController) checkStaleOperator(op *operator.Operator, step operator.OpStep, region *core.RegionInfo) bool {
+	switch s := step.(type) {
+	case operator.RemovePeer:
+		if s.FromStore == region.GetLeader().GetStoreId() {
+			if oc.removeOperatorWithoutBury(op) {
```
rleungx, line 154: How about using a new function for line 154 to line 167

nolouch:
PTAL @rleungx @HunDunDM

server/schedule/operator_controller.go
```diff
@@ -147,7 +147,21 @@ func (oc *OperatorController) Dispatch(region *core.RegionInfo, source string) {
 	}
 }
 
-func (oc *OperatorController) checkStaleOperator(op *operator.Operator, region *core.RegionInfo) bool {
+func (oc *OperatorController) checkStaleOperator(op *operator.Operator, step operator.OpStep, region *core.RegionInfo) bool {
+	switch s := step.(type) {
+	case operator.RemovePeer:
+		if s.FromStore == region.GetLeader().GetStoreId() {
```
disksing, line 153: I think it's too tricky to make special judgments on removePeer and leader here. Maybe we can solve it by recording the expected region leader at each step and check if the leader is the same. (like `op.ConfVerChanged`)

server/schedule/operator_controller.go
```diff
@@ -147,7 +147,21 @@ func (oc *OperatorController) Dispatch(region *core.RegionInfo, source string) {
...
+			if oc.RemoveOperator(op,
+				zap.String("reason", "stale operator, cannot remove leader per"),
```
rleungx, line 155: typo?

server/schedule/operator_controller.go
```diff
@@ -552,6 +562,17 @@ func (oc *OperatorController) buryOperator(op *operator.Operator) {
...
 			zap.Reflect("operator", op))
 		operatorCounter.WithLabelValues(op.Desc(), "timeout").Inc()
+	case operator.CANCELED:
+		fileds := []zap.Field{
+			zap.Uint64("region-id", op.RegionID()),
+			zap.Duration("takes", op.RunningTime()),
+			zap.Reflect("operator", op),
+		}
+		fileds = append(fileds, extraFileds...)
+		log.Info("operator caceled",
```
lhy1024, line 572:
```suggestion
		log.Info("operator cancelled",
```

server/schedule/operator/step.go
```diff
@@ -130,6 +153,21 @@ func (al AddLearner) IsFinish(region *core.RegionInfo) bool {
...
+func (al AddLearner) CheckSafety(region *core.RegionInfo) error {
+	peer := region.GetStorePeer(al.ToStore)
+	if peer == nil {
+		return nil
+	}
+	if peer.GetId() != al.PeerID {
+		return errors.Errorf("peer id is not matched, expect %d, got %d", al.PeerID, peer.GetId())
+	}
+	if peer.IsLearner {
+		return errors.New("peer already is a voter")
```
rleungx, line 166: learner?

server/schedule/operator/step.go
```diff
@@ -101,6 +115,15 @@ func (ap AddPeer) Influence(opInfluence OpInfluence, region *core.RegionInfo) {
 	to.AdjustStepCost(storelimit.RegionAdd, regionSize)
 }
 
+// CheckSafety checks if the step meets the safety.
+func (ap AddPeer) CheckSafety(region *core.RegionInfo) error {
+	peer := region.GetStorePeer(ap.ToStore)
+	if peer != nil && peer.GetId() != ap.PeerID {
+		return errors.Errorf("peer id is not matched, expect %d, got %d", ap.PeerID, peer.GetId())
```
rleungx, line 122: How about
```suggestion
		return errors.Errorf("peer %d has already existed in store %d, the operator is trying to add peer %d on the same store", peer.GetId(), ToStore, ap.PeerID)
```

server/schedule/operator/step.go
```diff
@@ -52,6 +54,18 @@ func (tl TransferLeader) IsFinish(region *core.RegionInfo) bool {
 	return region.GetLeader().GetStoreId() == tl.ToStore
 }
 
+// CheckSafety checks if the step meets the safety.
```
disksing, line 57: meets safy problem?

nolouch:
PTAL @disksing @rleungx

PR #2531, opened 2020-06-11
schedule: scatter regions to stores with the same engine

### LINKED ISSUE #2485: Region Scatter supports the learner peer

some user may use tiflash + tikv, and use `pre-split` to pre -balance the regions.
because the region scatters only consider voters, so let the learner peer may not under the balance, which causes some tiflashes IO became very busy. 

region scatter can consider the learner and the label of the engine.

sticnarf:

Fixes #2485 

This PR stores scattering information by engine in `engineContext`. Then, we do scattering only within stores with the same engine. I think it should help tiflash scatter its regions.

### Check List

Tests

- Unit test

Side effects

- - The scattering is more complex now. But I think it shouldn't be a big problem because it's rare to do this operation.

Related changes

- Need to cherry-pick to the release branch (?)

### Release note

Support scattering regions in stores with special engines (like TiFlash).


nolouch:
The test case relative to [here](https://github.com/pingcap/pd/blob/master/server/schedulers/scheduler_test.go#L187-L255), maybe we can move it to `schedule`. :)

sticnarf:
@nolouch PTAL

I am new to PD and not sure if it's a correct approach.

sticnarf:
Sorry, a test is still failing. Let me debug it more...

server/schedule/filter/filters.go
```diff
@@ -485,19 +485,19 @@ type engineFilter struct {
...
+		constraint: placement.LabelConstraint{Key: "engine", Op: "in", Values: allowedEngines},
 	}
+}
+
+// NewSpecialEngineFilter creates a filter that filters out special engine stores.
+func NewSpecialEngineFilter(scope string) Filter {
 	return &engineFilter{
 		scope:      scope,
-		constraint: placement.LabelConstraint{Key: "engine", Op: "notIn", Values: values},
+		constraint: placement.LabelConstraint{Key: "engine", Op: "notIn", Values: allSpeicalEngines},
```
nolouch, line 500: It is necessary to distinguish the type of engineFilter in order to be distinguished in metrics.

server/schedule/region_scatterer.go
```diff
@@ -102,30 +112,56 @@ func (r *RegionScatterer) Scatter(region *core.RegionInfo) (*operator.Operator,
...
+	var ordinaryPeers []*metapb.Peer
+	specialPeers := make(map[string][]*metapb.Peer)
+	// Group peers by the engine of their stores
 	for _, peer := range region.GetPeers() {
-		if len(stores) == 0 {
-			// Reset selected stores if we have no available stores.
-			r.selected.reset()
-			stores = r.collectAvailableStores(region)
+		store := r.cluster.GetStore(peer.GetStoreId())
+		if specialFilter.Target(r.cluster, store) {
```
nolouch, line 121: Should the targe of the `specialFilter` is special stores rather than ordinary store? or named to ordinaryFilter?

sticnarf: OK. All names related to "filter" are confusing - -
Let me change it to follow the rule that the name of the filter is what it keeps.

nolouch:
PTAL @lhy1024

PR #2543, opened 2020-06-16
scheduler: add opinfluence for balance leader scheduler

### LINKED ISSUE #2571: scheduler: no opinfluence for transferLeaderOut
## Bug Report

When we pick store for transfer leader in `transferLeaderOut`, there is no `opinfluence`.

Suppose：
There is a leader in store1 and followers in store2 and store3. And the score of store2 and store3 is the same.
When there has been some leader transfered to store2, we still pick it without `opinfluence`.

lhy1024:

We miss op influence when sort target store in balance leader scheduler. To fix #2571 

add it

### Check List

Tests

- Unit test

### Release note
- No release note

### DIFF
--- a/server/schedulers/balance_test.go
+++ b/server/schedulers/balance_test.go
@@ -333,7 +333,7 @@ func (s *testBalanceLeaderSchedulerSuite) TestTransferLeaderOut(c *C) {
 				regions[op.RegionID()] = struct{}{}
 				tr := op.Step(0).(operator.TransferLeader)
 				c.Assert(tr.FromStore, Equals, uint64(4))
-				targets[tr.ToStore] --
+				targets[tr.ToStore]--
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


server/schedulers/balance_test.go
```diff
@@ -304,6 +305,34 @@ func (s *testBalanceLeaderSchedulerSuite) TestScheduleWithOpInfluence(c *C) {
...
+	s.tc.AddLeaderStore(1, 7)
+	s.tc.AddLeaderStore(2, 8)
+	s.tc.AddLeaderStore(3, 9)
+	s.tc.AddLeaderStore(4, 12)
+	s.opt.TolerantSizeRatio = 0.1
+	for i := uint64(1); i <= 7; i++ {
+		s.tc.AddLeaderRegion(i, 4, 1, 2, 3)
+	}
+
+	// balance leader: 4->1, 4->1, 4->2
```
disksing, line 320: can we check if the target ids are correct?

lhy1024: There is no function to get target store id in `op`?

disksing: `op.Step(0).ToStore`

lhy1024: Thanks

PR #2573, opened 2020-06-23
filter, schedulers: split DistinctScoreFilter to LocatoinSafeguard and LocationImprover

disksing:
Enhance distinctScoreFilter to allow it can be used as score safe guard or improve placement.

- add 2 different policies: `safeGuard` and `improve` (the improve policy will be used by replicaChecker later)
- add tests

### Check List
Tests
- Unit test

### Release note
- No release note

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -279,17 +279,33 @@ func (f *storageThresholdFilter) Target(opt opt.Options, store *core.StoreInfo) 
 	return !store.IsLowSpace(opt.GetLowSpaceRatio())
 }
 
+const (
+	locationSafeguard = "safeguard"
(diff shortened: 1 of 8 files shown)


server/schedule/filter/filters.go
```diff
@@ -279,17 +279,33 @@ func (f *storageThresholdFilter) Target(opt opt.Options, store *core.StoreInfo)
 	return !store.IsLowSpace(opt.GetLowSpaceRatio())
 }
 
+const (
+	locationSafeguard = "safeguard"
+	locationImprove   = "improve"
```
lhy1024, line 284: Would you like to comment on them?

disksing: good idea

PR #2578, opened 2020-06-24
filter: refactor StoreStateFilter

disksing:

make `StoreStateFilter` more easy to maintain.

- extract `conditionFunc`s
- table driven check conditions
- support `AllowTemporaryStates` (aims to be used by `replicaChecker` later)

The rules are defined by following table:
```
// The condition table.
// Y: the condition is temporary (expected to become false soon).
// N: the condition is expected to be true for a long time.
// X means when the condition is true, the store CANNOT be selected.
//
// Condition    Down Offline Tomb Block Disconn Busy RmLimit AddLimit Snap Pending Reject
// IsTemporary  N    N       N    N     Y       Y    Y       Y        Y    Y       N
//
// LeaderSource X            X    X     X
// RegionSource                                 X    X                X
// LeaderTarget X    X       X    X     X       X                                  X
// RegionTarget X    X       X          X       X            X        X    X
```

The behaivour should be the same as before except for:
- `RegionSource` allows store to be in `Down` or `Tombstone` states. (It should be reasonable because we don't need a peer be online in order to remove it)

### Check List

Tests
- Unit test

### Release note

- No release note

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -330,6 +330,8 @@ type StoreStateFilter struct {
 	TransferLeader bool
 	// Set true if the schedule involves any move region operation.
 	MoveRegion bool
+	// Set true if allows temporary states.
+	AllowTemporaryStates bool
 }
 
 // Scope returns the scheduler or the checker which the filter acts on.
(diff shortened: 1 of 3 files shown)


server/schedule/filter/filters.go
```diff
@@ -358,6 +358,8 @@ type StoreStateFilter struct {
 	TransferLeader bool
 	// Set true if the schedule involves any move region operation.
 	MoveRegion bool
+	// Set true if allows temporary states.
+	AllowTemporaryStates bool
```
rleungx, line 362: When will it be true?

disksing: next version of `replicaChecker` will use it.

rleungx: got it

server/schedule/filter/filters.go
```diff
@@ -370,64 +372,119 @@ func (f StoreStateFilter) Type() string {
...
+// N: the condition is expected to be true for a long time.
+// X means when the condition is true, the store CANNOT be selected.
+//
+// Condition    Down Offline Tomb Block Disconn Busy RmLimit AddLimit Snap Pending Reject
+// IsTemporary  N    N       N    N     Y       Y    Y       Y        Y    Y       N
+//
+// LeaderSource X            X    X     X
+// RegionSource                                 X    X                X
+// LeaderTarget X    X       X    X     X       X                                  X
+// RegionTarget X    X       X          X       X            X        X    X
```
lhy1024, line 438: if store is blocked, it still can be selected as region target?

disksing: yes, block may be a bad name. It only means block leader transfer. I may have some other PR to change the name.

disksing: To block state is marked only by `evict-leader-scheduler` and `grant-leader-scheduler` AFAIK.

PR #2592, opened 2020-06-30
config: improve scheduler config reload behavior

disksing:

Stale scheduler may appear after PD leader change, caused by the reload behavior that it merges in-memory config with persist config.

Drop in-memory state and merge default schedulers with persist configuration.

### Check List

Tests
- Unit test
- Manually test

```
start 3pd + 1tikv
add evict-leader-scheduler-1
remove evict-leader-scheduler-1
kill pd1
show schedulers
make sure there is no evict-leader-scheduler-1
```

### Release note
- Fix the problem of stale scheduler after leader change

### DIFF
--- a/server/cluster/coordinator_test.go
+++ b/server/cluster/coordinator_test.go
@@ -672,8 +672,11 @@ func (s *testCoordinatorSuite) TestPersistScheduler(c *C) {
 	_, err = schedule.CreateScheduler(schedulers.AdjacentRegionType, oc, storage, schedule.ConfigJSONDecoder([]byte("null")))
 	c.Assert(err, IsNil)
 	// suppose we add a new default enable scheduler
-	newOpt.AddSchedulerCfg(schedulers.AdjacentRegionType, []string{})
-	c.Assert(newOpt.GetSchedulers(), HasLen, 5)
+	config.DefaultSchedulers = append(config.DefaultSchedulers, config.SchedulerConfig{Type: "adjacent-region"})
+	defer func() {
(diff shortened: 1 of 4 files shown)


server/cluster/coordinator_test.go
```diff
@@ -672,8 +672,11 @@ func (s *testCoordinatorSuite) TestPersistScheduler(c *C) {
 	_, err = schedule.CreateScheduler(schedulers.AdjacentRegionType, oc, storage, schedule.ConfigJSONDecoder([]byte("null")))
```
rleungx, line 672: I find another problem that when we use `ConfigJSONDecoder` here, the scheduler name will be empty and the path will only be `scheduler-config` instead of `scheduler-config/xxx`

server/config/persist_options.go
```diff
@@ -505,33 +499,16 @@ func (o *PersistOptions) Reload(storage *core.Storage) error {
...
-		needRestore := true
-		for _, s := range scheduleCfg.Schedulers {
-			if s.Type == ps.Type && reflect.DeepEqual(s.Args, ps.Args) {
-				needRestore = false
-				break
-			}
-		}
-		if needRestore {
-			restoredSchedulers = append(restoredSchedulers, ps)
+func (o *PersistOptions) adjustScheduleCfg(scheduleCfg *ScheduleConfig) {
```
lhy1024, line 502: do we need some test here?

disksing: It is covered. I migrated some old tests to adapt to the new behaviour.

PR #2596, opened 2020-07-01
misc: fix the deploy path not being updated

### LINKED ISSUE #2595: Deploy path of the store should be updated
### What did you do?

Move TiKV deploy directory to another place

The deploy path in the output of pd-ctl store command should be updated.

### What did you see instead?

It is not updated.

master(af5511a359a6cb08f8fdbb2ac7e51429fa0b1314)

rleungx:

Closes #2595.

This PR updates the existed store's deploy path when `Putstore` is called after restarting.

### Check List

Tests

- Unit test

Related changes

### Release note

- Fix the issue that the deploy path of the store is not updated if the deploy directory is moved

### DIFF
--- a/server/core/store_option.go
+++ b/server/core/store_option.go
@@ -64,6 +64,15 @@ func SetStoreVersion(githash, version string) StoreCreateOption {
 	}
 }
 
+// SetStoreDeployPath sets the deploy path for the store.
+func SetStoreDeployPath(deployPath string) StoreCreateOption {
+	return func(store *StoreInfo) {
+		meta := proto.Clone(store.meta).(*metapb.Store)
+		meta.DeployPath = deployPath
+		store.meta = meta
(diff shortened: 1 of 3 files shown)


server/core/store_option.go
```diff
@@ -64,6 +64,15 @@ func SetStoreVersion(githash, version string) StoreCreateOption {
 	}
 }
 
+// SetStoreDeployPath sets the deploy path for the store.
+func SetStoreDeployPath(deployPath string) StoreCreateOption {
+	return func(store *StoreInfo) {
+		meta := proto.Clone(store.meta).(*metapb.Store)
```
HunDunDM, line 70: Why `Clone` is needed here, it seems that other functions do not do this.

HunDunDM:
LGTM

Issue #2227, opened 2020-03-11
UCP: Improve the output of HTTP API POST method


rleungx:
## Description

Currently, we have many POST methods in our HTTP API, you can find them in `server/api/router.go`.
We need to improve the output of these methods e.g.
```
curl -X POST -d '{"region-schedule-limit":1}' http://127.0.0.1:33381/pd/api/v1/config
```
or 
```
curl -X POST -d '{"region-schedule-limit":1}' http://127.0.0.1:33381/pd/api/v1/config/schedule
```
If it is successful, it should return `success`, but actually it returns:
```
null
```
All the APIs should be changed.

## Score

- 50

## Recommended Skills

- Golang

disksing:
I think the `Success!` messages are generated from pd-ctl side.

rleungx:
Right, but I think this may confuse people who use the API. I have encountered many times.

disksing:
Do you suggest to remove all `Success!` from pd-ctl side too?

rleungx:
What do you think? @disksing

shafreeck:
For `Get` requests, the command should return the queried result, for `POST` requests, the command could choose to return the posted result. If all our APIs have this behavior, the `Success!` text may be removed.

disksing:
Maybe we can use some fixed predefined message structure, such as `{" code ": 200," message ":" request success "}`

ti-srebot:
Dear contributor, the second Performance challenge program has been closed. But you can still contribute to this issue and we appreciate your contribution.

PR #2597, opened 2020-07-01
api: improve the output of HTTP API POST method

JmPotato:

Issue #2227  

Replace all HTTP API POST/DELETE method‘s output with more specified messages.

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/pingcap/pd/blob/master/docs/development.md#updating-api-documentation))

### DIFF
--- a/server/api/admin.go
+++ b/server/api/admin.go
@@ -62,7 +62,7 @@ func (h *adminHandler) HandleDropCacheRegion(w http.ResponseWriter, r *http.Requ
 // @Accept json
 // @Param body body object true "json params"
 // @Produce json
-// @Success 200 {string} string "Reset ts success."
+// @Success 200 {string} string "Reset ts successfully."
 // @Failure 400 {string} string "The input is invalid."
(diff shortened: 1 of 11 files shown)


pkg/apiutil/apiutil.go
```diff
@@ -117,13 +117,32 @@ func ReadJSONRespondError(rd *render.Render, w http.ResponseWriter, body io.Read
 func ErrorResp(rd *render.Render, w http.ResponseWriter, err error) {
 	if err == nil {
 		log.Error("nil is given to errorResp")
-		rd.JSON(w, http.StatusInternalServerError, "nil error")
+		rd.JSON(w, http.StatusInternalServerError, WrapOutput(http.StatusInternalServerError, "nil error"))
```
disksing, line 120: Maybe more easy to use (don't have to wrapOutput for every response) if we can somehow wrap the `render.Render`?

JmPotato:
@disksing @rleungx PTAL

PR #2598, opened 2020-07-01
*: rename store block to pauseLeaderTransfer

disksing:

The term `block` is ambiguous.

Some renames.

### Check List

Tests
- Unit test

Side effects
  - Have API error message change.

### Release note
- No release note

### DIFF
--- a/server/core/basic_cluster.go
+++ b/server/core/basic_cluster.go
@@ -128,18 +128,18 @@ func (bc *BasicCluster) GetAdjacentRegions(region *RegionInfo) (*RegionInfo, *Re
 	return bc.Regions.GetAdjacentRegions(region)
 }
 
-// BlockStore stops balancer from selecting the store.
-func (bc *BasicCluster) BlockStore(storeID uint64) error {
+// PauseLeaderTransfer stops balancer from selecting the store.
+func (bc *BasicCluster) PauseLeaderTransfer(storeID uint64) error {
 	bc.Lock()
 	defer bc.Unlock()
-	return bc.Stores.BlockStore(storeID)
+	return bc.Stores.PauseLeaderTransfer(storeID)
 }
 
-// UnblockStore allows balancer to select the store.
-func (bc *BasicCluster) UnblockStore(storeID uint64) {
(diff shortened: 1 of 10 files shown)


server/core/basic_cluster.go
```diff
@@ -128,18 +128,18 @@ func (bc *BasicCluster) GetAdjacentRegions(region *RegionInfo) (*RegionInfo, *Re
 	return bc.Regions.GetAdjacentRegions(region)
 }
 
-// BlockStore stops balancer from selecting the store.
-func (bc *BasicCluster) BlockStore(storeID uint64) error {
+// PauseLeaderTransfer stops balancer from selecting the store.
```
rleungx, line 131: Does the comment need to be changed?

server/core/store.go
```diff
@@ -566,28 +566,28 @@ func (s *StoresInfo) SetStore(store *StoreInfo) {
 	s.stores[store.GetID()] = store
 }
 
-// BlockStore blocks a StoreInfo with storeID.
-func (s *StoresInfo) BlockStore(storeID uint64) errcode.ErrorCode {
-	op := errcode.Op("store.block")
+// PauseLeaderTransfer pauses a StoreInfo with storeID.
```
rleungx, line 569: ditto

server/core/store.go
```diff
@@ -566,28 +566,29 @@ func (s *StoresInfo) SetStore(store *StoreInfo) {
...
 }
 
-// UnblockStore unblocks a StoreInfo with storeID.
-func (s *StoresInfo) UnblockStore(storeID uint64) {
+// ResumeLeaderTransfer cleans a store's pause state. The store can be selected
+// as source or target of TransferLeader again.
+func (s *StoresInfo) ResumeLeaderTransfer(storeID uint64) {
 	store, ok := s.stores[storeID]
 	if !ok {
 		log.Fatal("store is unblocked, but it is not found",
```
lhy1024, line 588: comment need be updated

HunDunDM: careful!

disksing: good point.

Issue #2524, opened 2020-06-09
Increase the probability of balancing new regions


nolouch:
### Describe your feature request related problem
There is an issue about the new region may have a low probability to be balanced，such as:
```
                             8
                         7   7
                     6   6   6  
                 5   5   5   5
             4   4   4   4   4
         3   3   3   3   3   3
     2   2   2   2   2   2   2 
 1   1   1   1   1   1   1   1 
t1  t2  t3  t4  t5  t6  t7  t8 
```
such as the above drawing, the y-axis represents the existing regions at that moment. region 8 only appears at time t8, the probability of R8 being selected in the entire sample space is very small.  
this issue will lead to a scenario: import a new table into a big cluster,  the region of this table will be unevenly distributed, which will bring some hot store issues.
### Describe alternatives you've considered
Empirically, We hope that the distribution of data is relatively random. 

### Teachability, Documentation, Adoption, Migration Strategy
- Go
- Stochastic Processes / Probability theory

lhy1024:
We put new regions with TTL to the fresh set and pick them with more priority. Its disadvantage is that it is difficult for us to give the probability between the fresh set and the other one.

lhy1024:
The second proposal:

Suppose that we expect all regions have the same probability of being scheduled in the past `x` period. And that the total number of regions at the current time is `m`. 

When placing a new region with a TTL, which will live x period, the probability of being scheduled `( ttl + 1) /m`

Its disadvantage is that when the difference between `x` and `m` is too large, there is not a big difference in their probability

dimstars:
I tested a solution:
The total online time of the system `t1` minus the survival time of each region `t2` is used to represent the scheduling weight `w`, and the sum of all region weights `w` is `sumt`. Each region is selected with a probability of `w/sumt`.

The figure below shows the comparison between the three modifications and the probability of the original scheme after 1000 cycles (a new region will be generated in each cycle). The strategies they choose for each cycle are as follows:
Original scheme: all regions are in no particular order with equal probability.
w: The `w/sumt` scheme above.
w^2: In order to further reflect the urgency of choosing a new region, `w^2` is used instead of `w` for calculation.
w^10: Calculate `w^10` instead of `w`.
![image](https://user-images.githubusercontent.com/42988940/85983004-59194b00-ba19-11ea-9f53-6f18150957e1.png)

It can be seen that the "w^10" scheme can greatly shorten the selection probability of most regions in the middle

dimstars:
[...]

PR #2599, opened 2020-07-01
region: Add accelerate regions schedule API

Yisaer:
Add one new region HTTP API.

close https://github.com/pingcap/pd/issues/2593

Add one new region related http api to accelerate regions schedule by given key range.

### Check List

Tests

- Unit test

Related changes

### Release note

- Support Region HTTP API to promote regions schedule by given key range

### DIFF
--- a/server/api/region.go
+++ b/server/api/region.go
@@ -15,6 +15,8 @@ package api
 
 import (
 	"container/heap"
+	"encoding/hex"
+	"fmt"
 	"net/http"
 	"net/url"
 	"sort"
@@ -24,6 +26,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/kvproto/pkg/replication_modepb"
+	"github.com/pingcap/pd/v4/pkg/apiutil"
 	"github.com/pingcap/pd/v4/server"
 	"github.com/pingcap/pd/v4/server/core"
 	"github.com/unrolled/render"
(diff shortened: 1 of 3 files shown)


server/api/region.go
```diff
@@ -557,6 +558,43 @@ func (h *regionsHandler) GetTopSize(w http.ResponseWriter, r *http.Request) {
...
+// @Tags region
+// @Summary Promote regions with the given key range.
+// @Param startkey query string true "Region Start key"
+// @Param endkey query string true "Region End key"
+// @Param limit query integer false "Limit count" default(16)
+// @Produce json
+// @Success 200 {object} string
+// @Failure 400 {string} string "The input is invalid."
+// @Router /regions/keys/promote [get]
+func (h *regionsHandler) PromoteRegionsInRange(w http.ResponseWriter, r *http.Request) {
```
disksing, line 570: I think `promote` is kind ambiguous, especially we already use the term for the meaning of _promote a learner peer to voter_.

disksing: Could it be better to accept hex formatted keys?

Yisaer: @disksing It's ok to me to change the API/function name here and do you know the API examples to know about the better format to accept hex formatted keys?

rleungx:
I think https://github.com/pingcap/pd/issues/2524 wants to solve a similar problem. Maybe we can find a common way to do it?

disksing:
> I think #2524 wants to solve a similar problem. Maybe we can find a common way to do it?

IMO not very alike.

disksing:
Basically LGTM. But I wonder if we need to support a large number of regions in the range. One optimization method is to directly save the range, and then scan the range with higher priority first when scanning.

server/api/region.go
```diff
@@ -557,6 +559,71 @@ func (h *regionsHandler) GetTopSize(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	endKey, err := parseKey("end_key", input)
+	if err != nil {
+		h.rd.JSON(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	limit := defaultRegionLimit
```
disksing, line 601: 16 seems to small

server/api/router.go
```diff
@@ -148,6 +148,7 @@ func createRouter(ctx context.Context, prefix string, svr *server.Server) *mux.R
 	clusterRouter.HandleFunc("/regions/check/hist-size", regionsHandler.GetSizeHistogram).Methods("GET")
 	clusterRouter.HandleFunc("/regions/check/hist-keys", regionsHandler.GetKeysHistogram).Methods("GET")
 	clusterRouter.HandleFunc("/regions/sibling/{id}", regionsHandler.GetRegionSiblings).Methods("GET")
+	clusterRouter.HandleFunc("/regions/keys/accelerate-schedule", regionsHandler.AccelerateRegionsScheduleInRange).Methods("POST")
```
disksing, line 151: could `regions/accelerate-schedule` be bettter?

server/api/region.go
Yisaer: @disksing  Do you have any advice? What about 256?

disksing: Sounds good to me

server/api/router.go
Yisaer: updated.

server/api/region.go
```diff
@@ -557,6 +559,70 @@ func (h *regionsHandler) GetTopSize(w http.ResponseWriter, r *http.Request) {
...
+		limit = maxRegionLimit
+	}
+
+	regions := rc.ScanRegions([]byte(startKey), []byte(endKey), limit)
+	if len(regions) > 0 {
+		regionsIDList := make([]uint64, 0, len(regions))
+		for _, region := range regions {
+			regionsIDList = append(regionsIDList, region.GetID())
+		}
+		rc.AddSuspectRegions(regionsIDList...)
```
lhy1024, line 620: What will happen if repeatedly adding the same regions?

Yisaer: The duplicated operators would be discarded due to `checkAddOperator (due to same priority)` in `AddWaitingOperator` if the operators added in the first time haven't been successes yet.  Is that true? @disksing

server/api/region.go
rleungx: I think maybe you can refer to `parseKey` in `region_command.go`?

server/api/region_test.go
```diff
@@ -258,6 +258,18 @@ func (s *testRegionSuite) TestTopSize(c *C) {
...
+func (s *testRegionSuite) TestAccelerateRegionsScheduleInRange(c *C) {
+	r1 := newTestRegionInfo(557, 13, []byte("a1"), []byte("a2"))
+	r2 := newTestRegionInfo(558, 14, []byte("a2"), []byte("a3"))
+	r3 := newTestRegionInfo(559, 15, []byte("a3"), []byte("a4"))
+	mustRegionHeartbeat(c, s.svr, r1)
+	mustRegionHeartbeat(c, s.svr, r2)
+	mustRegionHeartbeat(c, s.svr, r3)
+
+	err := postJSON(testDialClient, fmt.Sprintf("%s/regions/accelerate-schedule", s.urlPrefix), []byte(`{"start_key":"a1", "end_key": "a4"}`))
+	c.Assert(err, IsNil)
```
rleungx, line 270: I think testing with a range `{"start_key":"a1", "end_key": "a3"}` and checking if only two regions in SuspectRegions will be better?

server/api/region.go
Yisaer: @disksing  Do you think we need to add a `format` parameter and support multi key format here?

server/api/region_test.go
Yisaer: good idea, updated.

server/api/region.go
disksing: IMO, hex format is enough because it is our preferred format in the future. Of course, supporting more formats won't make it worse.

Yisaer: Thanks for clarifying.

Yisaer:
@rleungx @disksing  Ready to be reviewed.

server/api/region.go
```diff
@@ -557,6 +560,74 @@ func (h *regionsHandler) GetTopSize(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	h.rd.JSON(w, http.StatusOK, nil)
```
rleungx, line 628: How about using Text and add a message for it?

Yisaer: updated.

server/api/region.go
```diff
@@ -557,6 +560,73 @@ func (h *regionsHandler) GetTopSize(w http.ResponseWriter, r *http.Request) {
...
+
+	regions := rc.ScanRegions([]byte(startKey), []byte(endKey), limit)
+	if len(regions) > 0 {
+		regionsIDList := make([]uint64, 0, len(regions))
+		for _, region := range regions {
+			regionsIDList = append(regionsIDList, region.GetID())
+		}
+		rc.AddSuspectRegions(regionsIDList...)
+	}
+	h.rd.JSON(w, http.StatusOK, fmt.Sprintf("accelerate regions scheduling in a given range[%s,%s)", rawStartKey, rawEndKey))
```
rleungx, line 627:
```suggestion
	h.rd.Text(w, http.StatusOK, fmt.Sprintf("Accelerate regions scheduling in a given range [%s,%s)", rawStartKey, rawEndKey))
```

server/api/region.go
disksing: I think true.

Yisaer:
@sre-bot /merge

lhy1024:
LGTM

rleungx:
LGTM

Issue #2526, opened 2020-06-09
Report QPS about each type of command in RegionHeartbeat


nolouch:
## Development Task
Ref: https://github.com/pingcap/pd/issues/2520
Report QPS about each type of command in RegionHeartbeat. for the multi-dimensions hotspot scheduling.

PR #2604, opened 2020-07-01
Report QPS about each type of command in RegionHeartbeat 

lhy1024:

To solve #2526

### Check List

Tests

- Unit test

Related changes

- PR to update [`pingcap/kvproto`](https://github.com/pingcap/kvproto/pull/638)

### Release note

- No release note


server/core/region.go
```diff
@@ -403,6 +403,14 @@ func (r *RegionInfo) GetReplicationStatus() *replication_modepb.RegionReplicatio
 	return r.replicationStatus
 }
 
+// GetReadQPS returns the region's qps by type.
+func (r *RegionInfo) GetReadQPS() uint64 {
+	if r.queryStats == nil {
+		return 0
+	}
+	return r.queryStats.Get + r.queryStats.Coprocessor + r.queryStats.Scan
```
disksing, line 411: do we need to assign different weights for different query types? As that `coprocessor` or `scan` requests can be heavier than point gets.

server/core/region.go
```diff
@@ -403,6 +403,22 @@ func (r *RegionInfo) GetReplicationStatus() *replication_modepb.RegionReplicatio
...
+}
+
+// GetQPSWrite returns the region's write QPS.
+func (r *RegionInfo) GetQPSWrite() uint64 {
+	if r.queryStats == nil {
+		return 0
+	}
+	return r.queryStats.Put
```
disksing, line 419: Should `Delete` be considered as write too?

lhy1024:
PTAL @HunDunDM

server/core/region.go
ZenoTan: Or we can have a better compare function when ranking topK?

lhy1024: Do you have any good suggestions

PR #2611, opened 2020-07-06
Increase the probability of balancing new regions

dimstars:
Signed-off-by: dimstars <liangjinrongcm@gmail.com>

Solve issue #2524 

Maintains the latest region in a queue. It is supported to give priority to these new regions with certain probability when balancing regions.

### Check List
Tests

- Integration test

### Release note

- No release note

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -777,6 +777,31 @@ func (c *RaftCluster) GetStoreRegions(storeID uint64) []*core.RegionInfo {
 	return c.core.GetStoreRegions(storeID)
 }
 
+// RandNewRegion returns a random region in new region set.
+func (c *RaftCluster) RandNewRegion(storeID uint64, ranges []core.KeyRange, optPending core.RegionOption, optOther core.RegionOption, optAll core.RegionOption) *core.RegionInfo {
+	return c.core.RandNewRegion(storeID, ranges, optPending, optOther, optAll)
+}
+
+// RemoveNewRegion removes region from NewRegions.
+func (c *RaftCluster) RemoveNewRegion(regionID uint64) {
+	c.core.RemoveNewRegion(regionID)
+}
+
+// DisableNewRegion prevents the region from being selected again.
+func (c *RaftCluster) DisableNewRegion(regionID uint64) {
+	c.core.DisableNewRegion(regionID)
+}
+
+// EnableNewRegion makes that the region can be selected again.
+func (c *RaftCluster) EnableNewRegion(regionID uint64) {
+	c.core.EnableNewRegion(regionID)
+}
+
+// GetNewRegions gets all new regions.
+func (c *RaftCluster) GetNewRegions() []*core.RegionInfo {
+	return c.core.GetNewRegions()
+}
+
 // RandLeaderRegion returns a random region that has leader on the store.
 func (c *RaftCluster) RandLeaderRegion(storeID uint64, ranges []core.KeyRange, opts ...core.RegionOption) *core.RegionInfo {
 	return c.core.RandLeaderRegion(storeID, ranges, opts...)
--- a/server/core/basic_cluster.go
+++ b/server/core/basic_cluster.go
@@ -15,7 +15,9 @@ package core
 
 import (
 	"bytes"
+	"math/rand"
 	"sync"
+	"time"
 
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
@@ -25,19 +27,52 @@ import (
 	"go.uber.org/zap"
 )
 
+/* SelectConfig provides parameters for selecting regions.
+ *
+ * NewRegionFirst : Select new regions first when balancing regions if true.
+ * NewProbability : The probability of new regions being selected.
+ * MaxRegionCount : The largest size of the new regions set.
+ * TimeThreshold  : Regions with a survival time less than this value are new.
+ */
+type SelectConfig struct {
+	NewRegionFirst bool
+	NewProbability float64
+	MaxRegionCount int
+	TimeThreshold  uint64
+}
+
 // BasicCluster provides basic data member and interface for a tikv cluster.
 type BasicCluster struct {
 	sync.RWMutex
-	Stores  *StoresInfo
-	Regions *RegionsInfo
+	Stores     *StoresInfo
+	Regions    *RegionsInfo
+	NewRegions *regionCache
+	SelectConf *SelectConfig
 }
 
 // NewBasicCluster creates a BasicCluster.
 func NewBasicCluster() *BasicCluster {
 	return &BasicCluster{
-		Stores:  NewStoresInfo(),
-		Regions: NewRegionsInfo(),
+		Stores:     NewStoresInfo(),
+		Regions:    NewRegionsInfo(),
+		NewRegions: newRegionCache(),
+		SelectConf: NewSelectConfig(),
 	}
+}
+
+// NewSelectConfig creates a SelectConfig.
+func NewSelectConfig() *SelectConfig {
+	return &SelectConfig{
+		NewRegionFirst: true,
+		NewProbability: 0.5,
+		MaxRegionCount: 1000,
+		TimeThreshold:  60 * 10,
+	}
+}
+
+// SetSelectConfig sets the SelectConfig for selecting regions.
+func (bc *BasicCluster) SetSelectConfig(selectConf *SelectConfig) {
+	bc.SelectConf = selectConf
 }
 
 // GetStores returns all Stores in the cluster.
@@ -73,6 +108,13 @@ func (bc *BasicCluster) GetRegions() []*RegionInfo {
 	bc.RLock()
 	defer bc.RUnlock()
 	return bc.Regions.GetRegions()
+}
+
+// GetNewRegions gets all RegionInfo from NewRegions.
+func (bc *BasicCluster) GetNewRegions() []*RegionInfo {
+	bc.RLock()
+	defer bc.RUnlock()
+	return bc.NewRegions.getRegions()
 }
 
 // GetMetaRegions gets a set of metapb.Region from regionMap.
@@ -160,6 +202,49 @@ func (bc *BasicCluster) UpdateStoreStatus(storeID uint64, leaderCount int, regio
 }
 
 const randomRegionMaxRetry = 10
+
+// RandNewRegion returns a random region in new region set.
+func (bc *BasicCluster) RandNewRegion(storeID uint64, ranges []KeyRange, optPending RegionOption, optOther RegionOption, optAll RegionOption) *RegionInfo {
+	bc.Lock()
+	if bc.SelectConf.NewRegionFirst {
+		rand.Seed(time.Now().UnixNano())
+		p := float64(rand.Intn(100)+1) / 100.0
+		if p <= bc.SelectConf.NewProbability {
+			region := bc.NewRegions.randomRegion(storeID, ranges, optPending, optOther, optAll, bc.SelectConf.TimeThreshold)
+			bc.Unlock()
+			return region
+		}
+	}
+	bc.Unlock()
+	return nil
+}
+
+// RemoveNewRegion removes region from NewRegions.
+func (bc *BasicCluster) RemoveNewRegion(regionID uint64) {
+	bc.Lock()
+	if bc.SelectConf.NewRegionFirst {
+		bc.NewRegions.remove(regionID)
+	}
+	bc.Unlock()
+}
+
+// DisableNewRegion prevents the region from being selected again.
+func (bc *BasicCluster) DisableNewRegion(regionID uint64) {
+	bc.Lock()
+	if bc.SelectConf.NewRegionFirst {
+		bc.NewRegions.disable(regionID)
+	}
+	bc.Unlock()
+}
+
+// EnableNewRegion makes that the region can be selected again.
+func (bc *BasicCluster) EnableNewRegion(regionID uint64) {
+	bc.Lock()
+	if bc.SelectConf.NewRegionFirst {
+		bc.NewRegions.enable(regionID)
+	}
+	bc.Unlock()
+}
 
 // RandFollowerRegion returns a random region that has a follower on the store.
 func (bc *BasicCluster) RandFollowerRegion(storeID uint64, ranges []KeyRange, opts ...RegionOption) *RegionInfo {
@@ -323,6 +408,14 @@ func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, erro
 func (bc *BasicCluster) PutRegion(region *RegionInfo) []*RegionInfo {
 	bc.Lock()
 	defer bc.Unlock()
+	if bc.SelectConf.NewRegionFirst && region.GetApproximateSize() > 1 {
+		if bc.Regions.GetRegion(region.GetID()) == nil || bc.NewRegions.getRegion(region.GetID()) != nil {
+			bc.NewRegions.update(region)
+			if bc.NewRegions.length() > bc.SelectConf.MaxRegionCount {
+				bc.NewRegions.pop()
+			}
+		}
+	}
 	return bc.Regions.SetRegion(region)
 }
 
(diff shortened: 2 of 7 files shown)


nolouch:
Can you make him be an optional policy, and can you share your test report?

dimstars:
It's better to make it optional. How do you hope to control it? Add an interface to "server/api" or something else?

nolouch:
PTAL @lhy1024 @rleungx

lhy1024:
Would you like to show test report?

server/cluster/cluster.go
```diff
@@ -723,6 +723,13 @@ func (c *RaftCluster) GetStoreRegions(storeID uint64) []*core.RegionInfo {
 	return c.core.GetStoreRegions(storeID)
 }
 
+
+// new******************************
```
lhy1024, line 727: please fix ci

rleungx:
@dimstars Thanks for your contribution! Please sign off CLA first.

dimstars:
I'm not sure when the region will be dispatched successfully. After a new region is selected, where in the code should it be removed from NewRegions? Or can we remove it from NewRegions as soon as operation is generated?

rleungx:
> I'm not sure when the region will be dispatched successfully. After a new region is selected, where in the code should it be removed from NewRegions? Or can we remove it from NewRegions as soon as operation is generated?

The dispatch logic is in server/schedule/operator_controller.go, you can take a look at `Dispatch` function.

dimstars:
So far I realized a scheme based on the new region set, just like that proposed by @lhy1024 :
1. When a new region with TTL is added to the cluster, it is also placed in the NewRegions set.
2. Set a probability p to select region in NewRegions set first.
3. If the rand function choose the old set, or if the region is not successfully selected from the new set, the selection is done in the original way.
4. After the new region is successfully dispatched, it is removed from NewRegions set.

I don't know exactly how to maintain this new set. Should I keep the NewRegions set at a fixed size or periodically remove old regions with TTL as a threshold? @rleungx 
As for the probability p, I don't think there's a fixed value. So it's better to let the administrator set it up.

dimstars:
I'm going to make these methods optional. In the future, there may be new and better proposals.
For parameters that are difficult to determine, provide an interface to set temporarily.

server/core/basic_cluster.go
```diff
@@ -317,6 +382,9 @@ func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, erro
 func (bc *BasicCluster) PutRegion(region *RegionInfo) []*RegionInfo {
 	bc.Lock()
 	defer bc.Unlock()
+	if bc.SelectConf.NewRegionFirst {
+		bc.NewRegions.SetRegion(region)
```
rleungx, line 386: When the cluster is restarted, it will put all regions into `NewRegions`

dimstars:
First, let me describe the original logic of balancing regions:
1. Call function `balanceRegionScheduler.Schedule`
2. Call `RandPendingRegion`, `RandFollowerRegion`, `RandLeaderRegion` and `RandLearnerRegion` in sequence according to the priority order, until a qualified region is selected.
3. The above four functions are almost consistent in logic. They first select 10(`randomRegionMaxRetry`=10) regions randomly within a given range and then return the first region which satisfies the condition.

dimstars:
Now, add new code without affecting the logic of the original code.
1. Make the strategy of region selection as an option, and the default is to use the original scheme: randomly select regions that meet the conditions.
2. Add a new Region set. Set a certain probability to select the region in the new set first (default p=0.5).
3. When a new region with a timestamp is added to the cluster, place it in the NewRegions set.
4. When a region is removed from the cluster, it is also removed from NewRegions.
5. If the random result of the RAND function does not select the new set, or if the region is not successfully found from the new set, then the region is selected in the original way.
6. After successfully dispatching the new region, delete the region from NewRegions.
7. Set a time threshold (such as one hour), and periodically remove regions with a survival time greater than this threshold from NewRegions.

dimstars:
Further, the region priority in the NewRegions set can be weighted according to the duration of survival. Preference is given to newer regions.
1. The time threshold `t1` minus the survival time `t2` represents the scheduling weight `w`.
2. The sum of all weights `w` is `sum`.
3. The probability being selected of each region is `w/sum`.
4. You can use the nonlinear increasing function `f`, replacing `w` with `f(w)`.

dimstars:
At present, the following disputes exist:
> 2\. Add a new Region set. Set a certain probability to select the region in the new set first (default p=0.5).
1. Is it necessary to maintain such a set?




> 7\. Set a time threshold (such as one hour), and periodically remove regions with a survival time greater than this threshold from NewRegions.
2. Updating the set regularly is expensive.Are there any other ways to maintain it?




> 3\. The above four functions are almost consistent in logic. They first select 10(`randomRegionMaxRetry`=10) regions randomly within a given range and then return the first region which satisfies the condition.
3. Should we select regions according to the probability before or after selecting the 10 regions? (If before that, the data quantity is all the regions in the new set within the range.If after that, the region may not be very new.)

server/cluster/cluster.go
```diff
@@ -504,6 +504,7 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		)
 		saveKV, saveCache, isNew = true, true, true
 	} else {
+		region.SetTimestamp(origin.GetTimestamp())
```
nolouch, line 507: region should be read-only after create from heartbeat, please see the detail comments about `RegionInfo`.

server/core/basic_cluster.go
```diff
@@ -317,9 +393,43 @@ func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, erro
 func (bc *BasicCluster) PutRegion(region *RegionInfo) []*RegionInfo {
 	bc.Lock()
 	defer bc.Unlock()
+	if bc.SelectConf.NewRegionFirst && region.GetTimestamp() >= uint64(time.Now().Unix())-bc.SelectConf.TimeThreshold {
```
nolouch, line 396: It's too intrusive to maintain such additional information.

dimstars: So just add a new set and a switch, and get rid of everything else？

server/cluster/cluster.go
dimstars: I'll modify this.

server/core/basic_cluster.go
```diff
@@ -317,9 +393,43 @@ func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, erro
...
+		bc.NewRegions.SetRegion(region)
+		bc.RemoveOldRegion()
+		bc.CalculateWeights()
```
ZenoTan, line 399: If PutRegion is called frequently, every time we need to scan a lot of regions, which will cost much.

server/core/basic_cluster.go
ZenoTan: My opinion is to try to integrate it into original logic well, without apparent decrease of performance(e.g. scan regions frequently).

dimstars: I see. I will try to reduce the cost of the new scheme and get close to the desired effect.

server/core/basic_cluster.go
dimstars: I had planned to scan regions periodically, but I did not complete it. Maybe scan is not a good idea. I will maintain this new set in other ways.

server/core/basic_cluster.go
```diff
@@ -24,21 +26,46 @@ import (
...
 }
 
 // NewBasicCluster creates a BasicCluster.
 func NewBasicCluster() *BasicCluster {
 	return &BasicCluster{
-		Stores:  NewStoresInfo(),
-		Regions: NewRegionsInfo(),
+		Stores:     NewStoresInfo(),
+		Regions:    NewRegionsInfo(),
+		NewRegions: newRegionQueue(),
```
ZenoTan, line 50: Can we make new region as an abstract structure?

dimstars: In that case we need two files. A file declares the abstract interface(put it in pd/ pkg/queue?). The other implements the region's queue, just like the regionTree. Would that be better?

ZenoTan: They can be put under `server/statictics`. You can see how hot region is managed as an example.

ZenoTan: Hot cache is like new region set. It is of limited size, and keep what schedulers care about.

dimstars: Ok, I'll try.

dimstars: I can't do like this because `import cycle not allowed`. Package `core` has been imported by files in `statistics`, which cannot be imported by files in `core` circularly.

server/core/new_region_cache.go
```diff
@@ -0,0 +1,187 @@
...
+func (queue *regionQueue) findNode(regionID uint64) *regionQueueNode {
+	temp := queue.start
+	for temp != nil {
+		if temp.region.GetID() == regionID {
+			return temp
+		}
+		temp = temp.next
+	}
+	return nil
+}
```
ZenoTan, line 60: This can be done faster using map?

dimstars: Map is a key-value store. It should be faster than scaning a linked list.  It's better to use map here. I will modify it.

server/core/basic_cluster.go
```diff
@@ -24,21 +26,44 @@ import (
...
 }
 
 // NewBasicCluster creates a BasicCluster.
 func NewBasicCluster() *BasicCluster {
 	return &BasicCluster{
-		Stores:  NewStoresInfo(),
-		Regions: NewRegionsInfo(),
+		Stores:     NewStoresInfo(),
+		Regions:    NewRegionsInfo(),
+		NewRegions: newRegionCache(1000),
```
ZenoTan, line 49: This 1000 can be put into config.

ZenoTan:
LGTM

dimstars:
### Final design

1. Maintain the most recently joined regions with a set;
2. When a region joins/moves out of the cluster, it is added/deleted into the collection;
3. The capacity of the collection can be set (the default is 1000), and the oldest region can be removed when the capacity is exceeded.
4. Make "Preferred New Region" an optional option (not on by default);
5. With this option on, set a probability `p` (default `p =0.5`) to select the new region;
6. The selection strategy of new region is: random selection (equal probability of each region);
7. If a new region is selected and the dispatch is successful, the region will be removed from the new set;
8. If the option is not turned on, or the random function does not select a new region, or there is no qualified new region, then a region is randomly selected in the original way.

### Specific implementation

1. In order to store the new region in an orderly way, the new region is stored in a bidirectional linked list, FIFO in default;
2. To find regions quickly, the RegionID was mapped to a linked list node using a `map`.
3. When region is inserted, the node is first searched through map, and then the node is directly updated when it is found; otherwise, the region is added to the end of the list and the map is updated at the same time.
4. When region is removed, the node is first searched through map, and then the node is directly removed and the map is updated at the same time.
5. When selecting a region, map's random traversal mechanism is used to return the first region.

server/schedulers/balance_test.go
```diff
@@ -1430,3 +1429,98 @@ func (s *testScatterRangeLeaderSuite) TestBalanceWhenRegionNotHeartbeat(c *C) {
 		schedule.ApplyOperator(tc, ops[0])
 	}
 }
+
+func (s *testBalanceRegionSchedulerSuite) TestBalanceNewRegion(c *C) {
```
ZenoTan, line 1433: Maybe some tests in simulator can be performed to see the result. @nolouch

server/core/basic_cluster.go
```diff
@@ -24,21 +26,46 @@ import (
 	"go.uber.org/zap"
 )
 
+// SelectConfig provides parameters for selecting regions.
+type SelectConfig struct {
+	NewRegionFirst bool
+	NewProbability float64
+	MxaRegionCount int
```
ZenoTan, line 33: Typo here
```suggestion
	MaxRegionCount int
```

server/core/basic_cluster.go
```diff
@@ -317,6 +374,12 @@ func (bc *BasicCluster) PreCheckPutRegion(region *RegionInfo) (*RegionInfo, erro
 func (bc *BasicCluster) PutRegion(region *RegionInfo) []*RegionInfo {
 	bc.Lock()
 	defer bc.Unlock()
+	if bc.SelectConf.NewRegionFirst {
```
nolouch, line 377: Could you make this function as a `Hook`, and extract this logic looks like  an plugin.

server/core/region_cache.go
```diff
@@ -159,9 +159,20 @@ func (cache *regionCache) pop() *RegionInfo {
 }
 
 // randomRegion returns a random region from regionCache.
-func (cache *regionCache) randomRegion() *RegionInfo {
+func (cache *regionCache) randomRegion(storeID uint64, ranges []KeyRange, n int) []*RegionInfo {
```
ZenoTan, line 162: Is current test result normal?

ZenoTan:
Make `DefaultStoreLimit` to 2000 for convenience of bench test. thx.

server/config/config.go
```diff
@@ -241,7 +241,7 @@ var (
 	defaultRuntimeServices = []string{}
 	defaultLocationLabels  = []string{}
 	// DefaultStoreLimit is the default store limit of add peer and remove peer.
-	DefaultStoreLimit StoreLimit = StoreLimit{AddPeer: 15, RemovePeer: 15}
```
rleungx, line 244: Why change it to 2000?

server/core/basic_cluster.go
```diff
@@ -25,21 +27,48 @@ import (
...
+	return &SelectConfig{
+		NewRegionFirst: true,
+		NewProbability: 0.5,
+		MaxRegionCount: 1000,
+		TimeThreshold:  60 * 10,
 	}
 }
 
+// SetSelectConfig sets the SelectConfig for selecting regions.
+func (bc *BasicCluster) SetSelectConfig(selectConf *SelectConfig) {
```
rleungx, line 68: Is it only used in tests?

server/core/basic_cluster.go
```diff
@@ -161,6 +197,49 @@ func (bc *BasicCluster) UpdateStoreStatus(storeID uint64, leaderCount int, regio
 
 const randomRegionMaxRetry = 10
 
+// RandNewRegion returns a random region in new region set.
+func (bc *BasicCluster) RandNewRegion(storeID uint64, ranges []KeyRange, optPending RegionOption, optOther RegionOption, optAll RegionOption) *RegionInfo {
+	bc.RLock()
+	if bc.SelectConf.NewRegionFirst {
```
rleungx, line 203: I think we can move this condition outside of this function

server/core/basic_cluster.go
```diff
@@ -25,21 +27,48 @@ import (
 	"go.uber.org/zap"
 )
 
+// SelectConfig provides parameters for selecting regions.
+type SelectConfig struct {
```
rleungx, line 31: Can you add some comments to explain these configurations?

server/core/basic_cluster.go
```diff
@@ -25,21 +27,48 @@ import (
...
 
 // NewBasicCluster creates a BasicCluster.
 func NewBasicCluster() *BasicCluster {
 	return &BasicCluster{
-		Stores:  NewStoresInfo(),
-		Regions: NewRegionsInfo(),
+		Stores:     NewStoresInfo(),
+		Regions:    NewRegionsInfo(),
+		NewRegions: newRegionCache(),
+		SelectConf: NewSelectConfig(),
```
rleungx, line 53: IMO, it's kind of weird to add a config here.

server/config/config.go
ZenoTan: Just for current bench test, he will change it back.

server/core/basic_cluster.go
dimstars: Yes. This is also due to the potential need for dynamic modification.

dimstars: Yes. This is also due to the potential need for dynamic modification.

server/core/basic_cluster.go
dimstars: This optimization is for balancing region. Perhaps config can be managed by balance_region scheduler. When scheduling, config is passed to cluster as a parameter. Is this modification feasible?

server/core/basic_cluster.go
dimstars: I'm sorry. Because of a lot of "benchmark" information recently, I've been ignoring some emails until now.

server/core/basic_cluster.go
dimstars: That leaves the scheduler to do the judging. It does reduce the number of function calls.

server/core/basic_cluster.go
dimstars: 好的，我将在下一次提交时添加它。

JmPotato:

PR #2612, opened 2020-07-06
config: fix configuration reload

disksing:
Fix https://github.com/pingcap/pd/issues/2609

The bug was introduced by https://github.com/pingcap/pd/pull/2592
We cannot use the configuration in the storage directly for backward compatibility reason. When we introduce new configurations, storage data won't contain it. In this case, we need to initialize it using default value.

### Check List
Tests
- Unit test

Related changes

### Release note
- No release note

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -769,7 +769,9 @@ func (c *ScheduleConfig) adjust(meta *configMetaData) error {
 	adjustFloat64(&c.LowSpaceRatio, defaultLowSpaceRatio)
 	adjustFloat64(&c.HighSpaceRatio, defaultHighSpaceRatio)
 
-	// make a copy to avoid changing DefaultSchedules unexpectly
+	// make a copy to avoid changing DefaultSchedules unexpectly.
+	// When reloading from storage, the config is passed to json.Unmarshal.
+	// Without clone, the DefaultSchedulers could be overwritten.
 	adjustSchedulers(&c.Schedulers, append(DefaultSchedulers[:0:0], DefaultSchedulers...))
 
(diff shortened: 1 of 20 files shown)


server/config/config.go
```diff
@@ -768,7 +768,9 @@ func (c *ScheduleConfig) adjust(meta *configMetaData) error {
 	}
 	adjustFloat64(&c.LowSpaceRatio, defaultLowSpaceRatio)
 	adjustFloat64(&c.HighSpaceRatio, defaultHighSpaceRatio)
-	adjustSchedulers(&c.Schedulers, DefaultSchedulers)
+
+	// make a copy to avoid changing DefaultSchedules unexpectly
```
rleungx, line 772:
```suggestion
	// make a copy to avoid changing DefaultSchedulers unexpectedly
```

server/config/config.go
```diff
@@ -768,7 +768,9 @@ func (c *ScheduleConfig) adjust(meta *configMetaData) error {
...
+	adjustSchedulers(&c.Schedulers, append(DefaultSchedulers[:0:0], DefaultSchedulers...))
```
rleungx, line 773: Would you like to add a test for this case or explain it more?

disksing: Good idea. I'll add more description.

HunDunDM: I think clone and related comments should be inside `adjustSchedulers`.

HunDunDM:
Can we put `o.adjustScheduleCfg(&cfg.Schedule)` in `if isExist`?

HunDunDM:
LGTM

lhy1024:
LGTM

PR #2614, opened 2020-07-07
tso: make follower fail fast for tso requests

disksing:

It takes too long (20s) for a follower PD to return an error to client. It may cause a longer unavailable time when PD leader changes.
Note: The client will not wait for 20s, but will timeout after 3 seconds.

Check lease before alloc ts.

### Check List
Tests
- Unit test

Related changes

### Release note
- Fix the issue that the TSO request may take too long when PD leader switches

### DIFF
--- a/tests/server/tso/tso_test.go
+++ b/tests/server/tso/tso_test.go
@@ -53,46 +53,6 @@ func (s *testTsoSuite) TearDownSuite(c *C) {
 	s.cancel()
 }
 
-func (s *testTsoSuite) TestRequestFollower(c *C) {
-	cluster, err := tests.NewTestCluster(s.ctx, 2)
-	c.Assert(err, IsNil)
-	defer cluster.Destroy()
-
-	err = cluster.RunInitialServers()
-	c.Assert(err, IsNil)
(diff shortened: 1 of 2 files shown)


tests/server/tso/tso_test.go
```diff
@@ -170,6 +170,46 @@ func (s *testTsoSuite) TestTsoCount0(c *C) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	tsoClient, err := grpcPDClient.Tso(ctx)
+	c.Assert(err, IsNil)
+	defer tsoClient.CloseSend()
+
+	start := time.Now()
+	err = tsoClient.Send(req)
+	c.Assert(err, IsNil)
+	_, err = tsoClient.Recv()
```
JmPotato, line 205: I think if we could check the response to make sure the tso returned is zero, it will make this test case more rigorous.

disksing: IMO not very useful because clients should always check `err` before using the response data. It does not really matter when err is not nil.

AndreMouche:
LGTM

PR #2615, opened 2020-07-08
cluster: change return value for getSchedulers

### LINKED ISSUE #2630: Return value of getSchedulers is not safe
In `server/cluster/coordinator.go`, we have a function named `getSchedulers`.
```
func (c *coordinator) getSchedulers() map[string]*scheduleController {
	c.RLock()
	defer c.RUnlock()
        return c.schedulers
}
```
But here we directly return a map that may not safe if we do some operations concurrently on this map outside.

rleungx:

Close #2630.

This PR changes the return value of `getSchedulers` from `map` to `[]string` and adds a new function named `GetSchedulerHandlers`.

### Check List

Tests

- Unit test

### Release note

- Fix the issue that `getSchedulers` may cause a data race


HunDunDM:
LGTM

PR #2616, opened 2020-07-08
client: check the timeout for creating stream

rleungx:

This PR uses an extra goroutine to check if creating stream causes too much time and will call `cancel` once it reaches the timeout setting of the client.

### Check List

Tests

- TODO


lhy1024:
LGTM

disksing:
@rleungx Did you send PR to update TiDB dependency?

rleungx:
> @rleungx Did you send PR to update TiDB dependency?

I will update it.

PR #2617, opened 2020-07-09
Add leader safeguard filter for balance-leader-scheduler

Yisaer:

close https://github.com/pingcap/pd/issues/2607

Add filter to check whether the target store break the placement rules in balance-leader-schelder in transferLeaderOut case and transferLeaderIn case.

### Check List

Tests

- Unit test

### Release note

- Fix the bug that balance-leader-scheduler won't consider placement rule when it enabled.


server/schedule/filter/filters.go
```diff
@@ -565,6 +565,46 @@ func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
...
+func (f *ruleLeaderFitFilter) Type() string {
+	return "rule-fit-filter"
+}
+
+func (f *ruleLeaderFitFilter) Source(opt opt.Options, store *core.StoreInfo) bool {
+	return true
+}
+
+func (f *ruleLeaderFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
+	oldLeaderPeerReplica := *f.region.GetLeader()
```
rleungx, line 601: Why do we need `*f.region.GetLeader()` instead of `f.region.GetLeader()`?

server/schedule/filter/filters.go
```diff
@@ -672,6 +722,32 @@ func (f *specialUseFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
 	return !f.constraint.MatchStore(store)
 }
 
+type alwaysAllowedFilter struct {
```
rleungx, line 725: IMO, we can remove this filter for simplicity.

Yisaer: updated.

server/schedule/filter/filters.go
```diff
@@ -565,6 +565,46 @@ func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
...
+		oldLeaderStoreID: oldLeaderStoreID,
+	}
+}
+
+func (f *ruleLeaderFitFilter) Scope() string {
+	return f.scope
+}
+
+func (f *ruleLeaderFitFilter) Type() string {
+	return "rule-fit-filter"
```
rleungx, line 593: For better metrics, the type should be unique.

server/schedulers/balance_rule_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/pingcap/pd/v4/pkg/mock/mockcluster"
+	"github.com/pingcap/pd/v4/pkg/mock/mockoption"
+	"github.com/pingcap/pd/v4/server/core"
+	"github.com/pingcap/pd/v4/server/kv"
+	"github.com/pingcap/pd/v4/server/schedule"
+	"github.com/pingcap/pd/v4/server/schedule/operator"
+	"github.com/pingcap/pd/v4/server/schedule/placement"
+)
+
+var _ = Suite(&testBalanceLeaderSchedulerWithRuleEnabledSuite{})
```
rleungx, line 29: How about moving this test into `server/schedulers/schedulers_test.go`?

server/schedulers/balance_leader.go
```diff
@@ -240,7 +244,18 @@ func (l *balanceLeaderScheduler) transferLeaderIn(cluster opt.Cluster, target *c
...
+	targets := []*core.StoreInfo{
+		target,
+	}
+	finalFilters := l.filters
+	if leaderFilter := filter.NewPlacementLeaderSafeguard(l.GetName(), cluster, region, source); leaderFilter != nil {
+		finalFilters = append(l.filters, leaderFilter)
+	}
+	targets = filter.SelectTargetStores(targets, finalFilters, cluster)
+	if len(targets) < 1 {
+		return nil
```
lhy1024, line 256: do we need to add a new inc in `schedulerCounter.WithLabelValues`

Yisaer: good catch, updated.

PR #2625, opened 2020-07-10
cooridinator: fix the issue caused adeadlock when deleting scheduler 

nolouch:

The PD service cannot serve after remove the scheduler.

The main reason:

goroutine 1:
* To delete the schedule, call `func (c *RaftCluster) RemoveScheduler(name string)
* Hold RaftCluster write lock
* Wait for coordinator write lock

goroutine  2:
* Background thread, collect statistics continuously, call func (c *coordinator) collectHotSpotMetrics()
* Hold the coordinator read lock
* Call s.Scheduler.(hasHotStatus).GetHotWriteStatus() to wait for hotScheduler read lock

goroutine3:
* Scheduling process, scheduling, call func (h *hotScheduler) Schedule(cluster opt.Cluster)
* Hold hotScheduler write lock
* Call cluster.GetStoresStats() to wait for RaftCluster read lock

### Check List

### Release note

- fix the issue that caused deadlock when deleting scheduler

### DIFF
--- a/server/cluster/coordinator.go
+++ b/server/cluster/coordinator.go
@@ -401,12 +401,13 @@ func (c *coordinator) resetSchedulerMetrics() {
 
 func (c *coordinator) collectHotSpotMetrics() {
 	c.RLock()
-	defer c.RUnlock()
 	// Collects hot write region metrics.
 	s, ok := c.schedulers[schedulers.HotRegionName]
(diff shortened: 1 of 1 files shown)


disksing:
Better to add some comments in case someone accidentally writes similar code in the future.

PR #2626, opened 2020-07-10
rule: let default rule fit the region at last

Yisaer:

Given region and store with following status:
```sh
// Stores:     1    2    3    4
// Leaders:    1    0    0    0
// Region1:    L    F    F    F
```
Then post a rule like following:
```json
  {
    "group_id": "test",
    "id": "test-leader",
    "index": 1,
    "start_key": "",
    "end_key": "",
    "role": "leader",
    "count": 1,
    "label_constraints": [
      {
        "key": "role",
        "op": "in",
        "values": [
          "leader"
        ]
      }
    ],
    "location_labels": [
      "host"
    ]
  }
```
Then label store 1 as "role": "leader"

### The expected RegionFit is:

For rule "test", the peers are peer-1 in store 1
For rule "pd", the peers are [peer-2,peer-3,peer-4] in store2/3/4

### What we got is:

For rule "test", the peers are empty.
For rule "pd", the peers are [peer-1,peer-2,peer-3] in store 1/2/3

One possible solution is to let default rule fit the region at last.

### Check List

Tests

- Unit test

### DIFF
--- a/server/schedule/placement/fit.go
+++ b/server/schedule/placement/fit.go
@@ -123,13 +123,25 @@ func FitRegion(stores core.StoreSetInformer, region *core.RegionInfo, rules []*R
 	if len(rules) == 0 {
 		return &regionFit
 	}
-	for _, rule := range rules {
+	f := func(rule *Rule) {
 		rf := fitRule(peers, rule)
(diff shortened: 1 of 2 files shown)


disksing:
I've composed another fix so closing it.

PR #2639, opened 2020-07-14
placement: improve regoin fit

disksing:
When we fit region's peers into rules, we check rules one by one, and for each one, we pick the best option (if there are multiple best options we simply pick the first one). The problem is that the current choice will affect subsequent rules, so the current optimal option is not necessarily the global optimal option.

For example, a region has 4 peers: `[host=h1, host=h2, host=h3, host=h4]`, and it matches 2 rules: `3 voters` + `1 voter on h2`. The best choice is `[host=h1, host=h3, host=h4]` goes to the first rule and `[host=h2]` goes to the second rule. Without consider globally, the first rule may pick `[host=h1, host=h2, host=h3]` and left `[host=h4]` which is not suitable for the second rule.

The problem is originally brought up from https://github.com/pingcap/pd/pull/2626

Recursive all options to choose the global best.

### Check List
Tests
- Unit test

### Release note
- Fix the issue that when enabling placement rules, sometimes region replicas cannot schedule to optimal


server/schedule/placement/fit.go
```diff
@@ -115,48 +114,134 @@ func compareRuleFit(a, b *RuleFit) int {
...
+}
+
 // FitRegion tries to fit peers of a region to the rules.
-func FitRegion(stores core.StoreSetInformer, region *core.RegionInfo, rules []*Rule) *RegionFit {
-	peers := prepareFitPeers(stores, region)
+func FitRegion(stores StoreSet, region *core.RegionInfo, rules []*Rule) *RegionFit {
+	w := newFitWorker(stores, region, rules)
+	w.run()
+	return &w.bestFit
+}
```
Yisaer, line 128: I suggest adding unit test for `FitRegion` so that we can add test cases to test `FitRegion` and compare regionFit result as a whole system.

disksing: The unit test does use `FitRegion` (but the test name is not correct now, I'll update it)

server/schedule/placement/fit_test.go
```diff
@@ -38,131 +38,120 @@ func (s *testFitSuite) makeStores() map[uint64]*core.StoreInfo {
...
+	// only support k=v type constraint
+	for _, c := range strings.Split(splits[2], ",") {
+		if c == "" {
+			break
+		}
+		kv := strings.Split(c, "=")
+		rule.LabelConstraints = append(rule.LabelConstraints, LabelConstraint{
+			Key:    kv[0],
+			Op:     "in",
+			Values: []string{kv[1]},
```
Yisaer, line 87: It's ok for me to only support `in` operator here, but could you support multi value for the `in` operator instead of only one value? I have some unit tests need this feature.

disksing: good idea.

server/schedule/placement/fit.go
```diff
@@ -115,48 +114,134 @@ func compareRuleFit(a, b *RuleFit) int {
...
+		// Reset previous result after position index.
+		for i := index + 1; i < len(w.rules); i++ {
+			w.bestFit.RuleFits[i] = nil
+		}
+		w.fitRule(index + 1)
+		w.updateOrphanPeers(index + 1)
+		return true
+	case 0:
+		if w.fitRule(index + 1) {
+			w.bestFit.RuleFits[index] = rf
```
JmPotato, line 227: Why do we need to update the `w.bestFit.RuleFits[index]` in the condition of `index + 1` has a better rule to fit?

JmPotato: Oh I see…When two rules have the same effect, the one which can make the next index have a better rule will become the current best one.

disksing: true

server/schedule/placement/fit_test.go
disksing: @Yisaer addressed

Yisaer:
I test a case with 

> "1111,1221,1331,1241", []string{"3/voter/host=host1+host2+host3+host4/rack,host", "1/voter/host=host3/"

The result is that "1111,1221,1331" fit the first rule and no peer fits the second rule. Is the result expected?

disksing:
> I test a case with
> 
> > "1111,1221,1331,1241", []string{"3/voter/host=host1+host2+host3+host4/rack,host", "1/voter/host=host3/"
> 
> The result is that "1111,1221,1331" fit the first rule and no peer fits the second rule. Is the result expected?

It is expedted. for rule1, `1111,1221,1331` is better than `1111,1221,1241`.

It is the same situation with my test case `{"1111,2211,3111,3112", []string{"3/voter//zone", "1/voter/rack=rack2/"}, "1111,2211,3111//3112"}`.

PR #2647, opened 2020-07-14
checker: extract replica_strategy

disksing:
refactor

- extract common utilities of `replicaChecker` and `ruleChecker` to `ReplicaStrategy`
- use `ReplicaStrategy` in both checkers
- remove the selector which is not used any more

### Check List

Tests
- Unit test

### Release note
- No release note


server/schedule/checker/replica_strategy.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// SelectStoreToAdd returns the store to add a replica to a region.
+// `coLocationStores` are the stores used to compare location with target
+// store.
+// `extraFilters` is used to set up more filters based on the context that
+// calling this method.
+//
+// For example, to select a target store to replace a region's peer, we can use
+// the peer list without the peer to be removed as `coLocationStores`.
```
JmPotato, line 42: > the peer list without the peer to be removed as `coLocationStores`.

What's the meaning of this sentence?

disksing: For example, region contains 3 peers `[p1, p2, p3]`. In order to replace `p3`, we will call `SelectStoreToAdd` and pass `[p1, p2]` in.

JmPotato: I see...So this sentence like this maybe better?

> the peer list with the peer removed as `coLocationStores`.

disksing: good idea. thanks.

PR #2659, opened 2020-07-16
schedule: improve the leader distribution after region scatter

nolouch:

Fix https://github.com/pingcap/pd/issues/2655

- add leader store counter
- picker the minimal store
### Check List

Tests

- Unit test
the new result with this PR.
```
MySQL [test]> select count(s.region_id) cnt, s.index_name, p.store_id from INFORMATION_SCHEMA.TIKV_REGION_STATUS s join INFORMATION_SCHEMA.tikv_region_peers p on s.region_id = p.region_id where s.table_name = 'ss' group by index_name, p.store_id order by index_name,cnt desc;
+-----+------------+----------+
| cnt | index_name | store_id |
+-----+------------+----------+
| 172 | NULL       |        6 |
| 172 | NULL       |       11 |
| 172 | NULL       |        1 |
| 170 | NULL       |       93 |
| 170 | NULL       |  3670905 |
| 170 | NULL       |       10 |
| 170 | NULL       |        4 |
| 170 | NULL       |        5 |
| 170 | NULL       |        8 |
|   1 | idx1       |       11 |
|   1 | idx1       |        1 |
|   1 | idx1       |        6 |
+-----+------------+----------+
12 rows in set (0.43 sec)

MySQL [test]> select count(s.region_id) cnt, s.index_name, p.store_id from INFORMATION_SCHEMA.TIKV_REGION_STATUS s join INFORMATION_SCHEMA.tikv_region_peers p on s.region_id = p.region_id where s.table_name = 'ss' and p.is_leader = 1 group by index_name, p.store_id order by index_name,cnt desc;
+-----+------------+----------+
| cnt | index_name | store_id |
+-----+------------+----------+
|  59 | NULL       |        1 |
|  58 | NULL       |        6 |
|  57 | NULL       |       93 |
|  57 | NULL       |       10 |
|  57 | NULL       |        8 |
|  57 | NULL       |        4 |
|  56 | NULL       |        5 |
|  56 | NULL       |  3670905 |
|  55 | NULL       |       11 |
|   1 | idx1       |        6 |
+-----+------------+----------+
10 rows in set (0.47 sec)

```

### Release note
-  improve the leader distribution after region scatter


server/schedule/region_scatterer.go
```diff
@@ -29,6 +30,29 @@ import (
 
 const regionScatterName = "region-scatter"
 
+type selectedLeaderStores struct {
+	mu     sync.Mutex
+	stores map[uint64]uint64
```
Yisaer, line 35: plz add the comment to explain meaning of the key and value in this map.

server/schedule/region_scatterer.go
```diff
@@ -216,3 +245,19 @@ func (r *RegionScatterer) collectAvailableStores(region *core.RegionInfo, contex
 	}
 	return targets
 }
+
+func (r *RegionScatterer) collectAvailableLeaderStores(peers map[uint64]*metapb.Peer, context engineContext) uint64 {
```
Yisaer, line 249: What about `searchLeastLeaderStore`, `collectLeaderStores` is misleading.

server/schedule/region_scatterer.go
```diff
@@ -153,6 +179,8 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo) *operator.Opera
 	}
 
 	scatterWithSameEngine(ordinaryPeers, r.ordinaryEngine)
+	// FIXME: target leader only consider the ordinary engine.
+	targetLeader := r.collectAvailableLeaderStores(targetPeers, r.ordinaryEngine)
```
Yisaer, line 183: Will the leader store be collected again after `collectAvailableLeaderStores` in `scatterWithSameEngine` function?

server/schedule/region_scatterer.go
```diff
@@ -29,6 +30,29 @@ import (
...
+type selectedLeaderStores struct {
+	mu     sync.Mutex
+	stores map[uint64]uint64
+}
+
+func (s *selectedLeaderStores) put(id uint64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.stores[id] = s.stores[id] + 1
+}
```
Yisaer, line 42: If the leader is transferred manually, will the old store's count minus 1 and new store's count plus 1?

server/schedule/region_scatterer_test.go
```diff
@@ -129,16 +143,16 @@ func (s *testScatterRegionSuite) scatterSpecial(c *C, numOrdinaryStores, numSpec
 		GroupID: "pd", ID: "learner", Role: placement.Learner, Count: 3,
 		LabelConstraints: []placement.LabelConstraint{{Key: "engine", Op: placement.In, Values: []string{"tiflash"}}}}), IsNil)
 
-	ordinarySeq := newSequencer(numOrdinaryStores)
-	specialSeq := newSequencerWithMinID(numOrdinaryStores+1, numOrdinaryStores+numSpecialStores)
+	//ordinarySeq := newSequencer(numOrdinaryStores)
+	// specialSeq := newSequencerWithMinID(numOrdinaryStores+1, numOrdinaryStores+numSpecialStores)
```
rleungx, line 147: No need anymore?

server/schedule/region_scatterer.go
nolouch: now, only consider the dorinaryEngine.

server/schedule/region_scatterer.go
nolouch: Yes, you are right. This scheduler assumes that the leader and region will not change significantly after scheduling. Maybe we need to discuss optimization in another issue.

server/schedule/region_scatterer.go
```diff
@@ -153,6 +179,8 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo) *operator.Opera
 	}
 
 	scatterWithSameEngine(ordinaryPeers, r.ordinaryEngine)
+	// FIXME: target leader only consider the ordinary engine.
```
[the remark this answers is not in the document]
nolouch: I add more comments.

server/schedule/region_scatterer.go
Yisaer: I think we can directly to make a mechanism (syncer or tracker) to record the correct leader count distrubution.

nolouch: In fact, it will be more complicated to do this, such as `truncate table` we need to remove the regions from the tracker, and what if `recover a table`? The current approach is at least feasible in general, there will be no more operator after cluster in balanced, especially in big cluster.

nolouch: But the tracker you mentioned is a good idea, and I even want to use it to report why there is this operator produce.
cc @Yisaer

Issue #2528, opened 2020-06-10
proposal: add new replication config to isolate replicas explicitly and forcibly


disksing:
### Describe your feature request related problem
https://github.com/pingcap/pd/issues/2418

### Describe the feature you'd like
Add a new config `isolation-label` (type: string).
When `isolation-label` is set, PD checkers do not consider `location-labels` any more. Instead, PD makes sure replicas of a region be isolated by the specified label.
For example, with `isolation-label = zone`, PD ensure that all replicas be placed in different zones. Even if a zone is down, PD will not try to make up replicas in other zone.

JmPotato:
So the `isolation-label` will be strictly enforced even the replicas may stay in a state where less than three? Do we already have a mechanism now to check whether a satisfied store is back online in this situation or we have to do some extra work?

disksing:
Yes and yes.

JmPotato:
@disksing My current thought is to add a config named `isolation-label` and do some work in `ReplicaChecker`'s selector. `selectBestPeerToAddReplica` will return `nil` if no target meets this isolation rule. In this way both this new strict rule is added and regularly checking will be guaranteed.

disksing:
> @disksing My current thought is to add a config named `isolation-label` and do some work in `ReplicaChecker`'s selector. `selectBestPeerToAddReplica` will return `nil` if no target meets this isolation rule. In this way both this new strict rule is added and regularly checking will be guaranteed.

Good idea. Do you want to do it? I'm refactoring the `replicaChecker`, you may want to help to review https://github.com/pingcap/pd/pull/2610 and fix this issue afterwards.

JmPotato:
> > @disksing My current thought is to add a config named `isolation-label` and do some work in `ReplicaChecker`'s selector. `selectBestPeerToAddReplica` will return `nil` if no target meets this isolation rule. In this way both this new strict rule is added and regularly checking will be guaranteed.
> 
> Good idea. Do you want to do it? I'm refactoring the `replicaChecker`, you may want to help to review #2610 and fix this issue afterwards.

Yeah. Glad to help!

JmPotato:
@disksing What if `isolation-label` is conflicted with a placement rule? Should `isolation-label` only be guaranteed in `replica_checker` than `rule_checker`? IMO, a placement rule has a higher priority. Maybe  `rule_checker` shouldn't consider this `isolation-label`.

disksing:
[...]

Issue #2418, opened 2020-05-13
replica_checker: replace down replica may violate location labels


disksing:
when a store is down, PD selects the _best_ target store to replace the down replica. The problem is, the best target may not as good as the original one, so the isolation level will be decreased.

example: consider a cluster deployed to 3 DCs, and each DC has 2 stores:
dc1: store1, store2
dc2: store3, store4
dc3: store5, store6

suppose we have a region that replicates over 3 DCs, its peers are located in store1, store3, store5. When store1 is down, the best replacement will be store2. However, if store2 is almost full or busy, PD will select store4 or store6 as the _best_ one, then after the operator is finished, the region does not replicate over 3 DCs any more.

how to fix: we may introduce a new option to control the behaivour: allow decrease isolation level or don't allow it.

MimeLyc:
refresh

PR #2662, opened 2020-07-16
config, replica_strategy, filter: add new replication config isolation-level to isolate replicas explicitly and forcibly

JmPotato:

Add a new replication config named `isolation-level` according to #2528 #2418 

With the new config `isolation-level`, `SelectStoreToAdd` in `replica_strategy.go` will filter out stores to isolate replicas explicitly and forcibly. For example, a region has 3 replicas in z1, z2 and z3 individually. With IsolationLevel = zone, if the region on z1 is down, we need to filter out z2 and z3 because these two zones already have one of the region's replicas on them. We need to choose a store on z1 or z4 to place the new replica to meet the `isolation-level`.

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has configuration change
- Has `replica_strategy` change
- Has `filter` change

### Release note

Add a new replication config named `isolation-level` to isolate replicas explicitly and forcibly.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -954,7 +954,7 @@ type ReplicationConfig struct {
 	// StrictlyMatchLabel strictly checks if the label of TiKV is matched with LocationLabels.
 	StrictlyMatchLabel bool `toml:"strictly-match-label" json:"strictly-match-label,string"`
 
-	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not used any more.
+	// When PlacementRules feature is enabled. MaxReplicas, LocationLabels and IsolationLabels are not used any more.
 	EnablePlacementRules bool `toml:"enable-placement-rules" json:"enable-placement-rules,string"`
 
 	// IsolationLevel is used to isolate replicas explicitly and forcibly if it's not empty.
@@ -982,22 +982,19 @@ func (c *ReplicationConfig) clone() *ReplicationConfig {
 
 // Validate is used to validate if some replication configurations are right.
 func (c *ReplicationConfig) Validate() error {
-	validIsolationLevel := false
-	if len(c.IsolationLevel) <= 0 {
-		validIsolationLevel = true
-	}
+	foundIsolationLevel := false
 	for _, label := range c.LocationLabels {
 		err := ValidateLabels([]*metapb.StoreLabel{{Key: label}})
 		if err != nil {
 			return err
 		}
 		// IsolationLevel should be empty or one of LocationLabels
-		if !validIsolationLevel && label == c.IsolationLevel {
-			validIsolationLevel = true
+		if !foundIsolationLevel && label == c.IsolationLevel {
+			foundIsolationLevel = true
 		}
 	}
-	if !validIsolationLevel {
-		return errors.New("isolation-level must be one of location-labels")
+	if len(c.IsolationLevel) > 0 && !foundIsolationLevel {
+		return errors.New("isolation-level must be one of location-labels or empty")
 	}
 	return nil
 }
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -954,8 +954,14 @@ type ReplicationConfig struct {
 	// StrictlyMatchLabel strictly checks if the label of TiKV is matched with LocationLabels.
 	StrictlyMatchLabel bool `toml:"strictly-match-label" json:"strictly-match-label,string"`
 
-	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not uesd any more.
+	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not used any more.
 	EnablePlacementRules bool `toml:"enable-placement-rules" json:"enable-placement-rules,string"`
+
+	// IsolationLabel is used to isolate replicas explicitly and forcibly.
+	// For example, with isolation-label = zone, PD ensure that all replicas be placed in different zones.
+	// Even if a zone is down, PD will not try to make up replicas in other zone
+	// because other zones already have replicas on it.
+	IsolationLabel string `toml:"isolation-label" json:"isolation-label,string"`
 }
 
 func (c *ReplicationConfig) clone() *ReplicationConfig {
@@ -966,13 +972,27 @@ func (c *ReplicationConfig) clone() *ReplicationConfig {
 		LocationLabels:       locationLabels,
 		StrictlyMatchLabel:   c.StrictlyMatchLabel,
 		EnablePlacementRules: c.EnablePlacementRules,
+		IsolationLabel:       c.IsolationLabel,
 	}
 }
 
 // Validate is used to validate if some replication configurations are right.
 func (c *ReplicationConfig) Validate() error {
+	validIsolationLabel := false
 	for _, label := range c.LocationLabels {
 		err := ValidateLabels([]*metapb.StoreLabel{{Key: label}})
+		if err != nil {
+			return err
+		}
+		if !validIsolationLabel && label == c.IsolationLabel {
+			validIsolationLabel = true
+		}
+	}
+	if len(c.IsolationLabel) > 0 {
+		if !validIsolationLabel {
+			return errors.New("invalid isolation-label")
+		}
+		err := ValidateLabels([]*metapb.StoreLabel{{Key: c.IsolationLabel}})
 		if err != nil {
 			return err
 		}
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -957,11 +957,15 @@ type ReplicationConfig struct {
 	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not used any more.
 	EnablePlacementRules bool `toml:"enable-placement-rules" json:"enable-placement-rules,string"`
 
-	// IsolationLabel is used to isolate replicas explicitly and forcibly.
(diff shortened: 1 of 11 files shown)


server/config/config.go
```diff
@@ -966,16 +972,30 @@ func (c *ReplicationConfig) clone() *ReplicationConfig {
...
 		}
+		if !validIsolationLabel && label == c.IsolationLabel {
+			validIsolationLabel = true
+		}
+	}
+	if len(c.IsolationLabel) > 0 {
+		if !validIsolationLabel {
+			return errors.New("invalid isolation-label")
+		}
+		err := ValidateLabels([]*metapb.StoreLabel{{Key: c.IsolationLabel}})
```
rleungx, line 995: Do we need this? Maybe it has already validated by the above loop? Another question is what if we don't set the LocationLabels and only set IsolationLabel? It seems that it will return an error. Does it meet your expectation?

JmPotato: You're right. Above loop guarantees this.

server/config/config.go
```diff
@@ -954,8 +954,14 @@ type ReplicationConfig struct {
...
 
-	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not uesd any more.
+	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not used any more.
 	EnablePlacementRules bool `toml:"enable-placement-rules" json:"enable-placement-rules,string"`
+
+	// IsolationLabel is used to isolate replicas explicitly and forcibly.
+	// For example, with isolation-label = zone, PD ensure that all replicas be placed in different zones.
+	// Even if a zone is down, PD will not try to make up replicas in other zone
+	// because other zones already have replicas on it.
+	IsolationLabel string `toml:"isolation-label" json:"isolation-label"`
```
Yisaer, line 964: What's the default value of the `IsolationLabel`?  Does it always be enabled or have a switch to control ?

Yisaer: Also I think we can list the valid option for this configuration.

Yisaer: Maybe `isolation-level` is better?

server/schedule/filter/filters_test.go
```diff
@@ -135,6 +135,73 @@ func (s *testFiltersSuite) TestStoreStateFilter(c *C) {
...
+		{
+			core.NewRegionInfo(&metapb.Region{Peers: []*metapb.Peer{
+				{Id: 1, StoreId: 1},
+				{Id: 2, StoreId: 4},
+				{Id: 3, StoreId: 6},
+			}}, &metapb.Peer{StoreId: 1, Id: 1}),
+			"host",
+			[]bool{true, true, true, true, true, true, true},
+			[]bool{false, false, true, false, true, false, true},
+		},
```
Yisaer, line 193: What about add an invalid isolationLabel input case

server/config/config.go
JmPotato: `IsolationLabel` should be one of the `LocationLabels`. If it's empty, we will think the strict isolation level is off. But if we rename it to `isolation-level`, maybe its value should be the index of `LocationLabels`? For example, we may have a configuration like this:

```
LocationLabels: ["zone", "rack", "host"]

IsolationLabel: "zone"
// Or IsolationLevel: 0
```

What's your opinion?

server/schedule/filter/filters_test.go
JmPotato: Actually `isolationLabel ` will be validated in `ReplicationConfig`'s `Validate()`, so the filter will definitely get a correct one.

server/config/config.go
Yisaer: I tend to make the configuration in following format which is obvious for the users even they have little knowledge about the pd isolation scheduling strategy.

```sh
LocationLabels: ["zone", "rack", "host"]

IsolationLevel: "zone"
```

JmPotato: Nice idea.

server/config/config.go
```diff
@@ -954,8 +954,18 @@ type ReplicationConfig struct {
 	// StrictlyMatchLabel strictly checks if the label of TiKV is matched with LocationLabels.
 	StrictlyMatchLabel bool `toml:"strictly-match-label" json:"strictly-match-label,string"`
 
-	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not uesd any more.
+	// When PlacementRules feature is enabled. MaxReplicas and LocationLabels are not used any more.
```
disksing, line 957: MaxReplicas, LocationLabels and IsolationLabels are not used any more

server/config/config.go
```diff
@@ -966,16 +976,28 @@ func (c *ReplicationConfig) clone() *ReplicationConfig {
...
 		if err != nil {
 			return err
 		}
+		// IsolationLevel should be empty or one of LocationLabels
+		if !validIsolationLevel && label == c.IsolationLevel {
+			validIsolationLevel = true
+		}
+	}
+	if !validIsolationLevel {
+		return errors.New("isolation-level must be one of location-labels")
```
disksing, line 1000: I think this way could be better.
```
foundIsolationLevel := false
for ... {
    ....
    if label == c.IsolationLevel {
        foundIsolationLabel = true
    }
}
if c.IsolationLevel != "" && !foundIsolationLabel {
    return err
}
```

server/schedule/checker/replica_strategy.go
```diff
@@ -53,6 +54,7 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 
 	filters := []filter.Filter{
 		filter.NewExcludedFilter(s.checkerName, nil, s.region.GetStoreIds()),
+		filter.NewIsolationFilter(s.checkerName, s.isolationLevel, s.locationLabels, s.cluster.GetRegionStores(s.region)),
```
disksing, line 57: should use `coLocationStores` instead of `s.cluster.GetRegonStores(s.region))` here.

server/schedule/filter/filters.go
```diff
@@ -601,3 +601,74 @@ const (
 
 var allSpecialUses = []string{SpecialUseHotRegion, SpecialUseReserved}
 var allSpeicalEngines = []string{EngineTiFlash}
+
+type isolationFilter struct {
+	scope         string
+	constraintSet [][]*placement.LabelConstraint
```
disksing, line 607: I think no need to introduce `placement.LabelConstraint` here. We can just use `[][]string`.

server/schedule/checker/replica_strategy.go
```diff
@@ -50,13 +51,16 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
...
-
 	filters := []filter.Filter{
 		filter.NewExcludedFilter(s.checkerName, nil, s.region.GetStoreIds()),
 		filter.NewStorageThresholdFilter(s.checkerName),
 		filter.NewSpecialUseFilter(s.checkerName),
 		filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true, AllowTemporaryStates: true},
 	}
+	// IsolationLevel should only be available when PlacementRulesEnabled is false
+	if s.cluster.IsPlacementRulesEnabled() {
+		filters = append(filters, filter.NewIsolationFilter(s.checkerName, s.isolationLevel, s.locationLabels, coLocationStores))
```
disksing, line 62: maybe pass in the `filter` using the `extraFilters` is better

server/schedule/filter/filters.go
```diff
@@ -601,3 +601,77 @@ const (
...
+func (f *isolationFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
+	// No isolation constraint to fit
+	if len(f.constraintSet) <= 0 {
+		return true
+	}
+	for _, constrainList := range f.constraintSet {
+		match := true
+		for idx, constraint := range constrainList {
+			// Check every constraint in constrainList
+			labelConstraint := &placement.LabelConstraint{
```
disksing, line 665: I meant that you can compare label values directly without introducing `LabelConstraint` here.

JmPotato: You mean just use `GetLabelValue` to check label and value manually?

JmPotato: I got it. It's indeed much simpler than using `LabelConstraint` in this case...

server/schedule/filter/filters.go
```diff
@@ -601,3 +601,72 @@ const (
...
+// With isolationLevel = zone, if the region on z1 is down, we need to filter out z2 and z3
+// because these two zones already have one of the region's replicas on them.
+// We need to choose a store on z1 or z4 to place the new replica to meet the isolationLevel explicitly and forcibly.
+func NewIsolationFilter(scope, isolationLevel string, locationLabels []string, regionStores []*core.StoreInfo) Filter {
+	isolationFilter := &isolationFilter{
+		scope:          scope,
+		locationLabels: locationLabels,
+		constraintSet:  make([][]string, 0),
+	}
+	if len(isolationLevel) <= 0 || len(locationLabels) <= 0 {
```
rleungx, line 622: In this case, I think we don't need to create a filter?

JmPotato: You are right. This part may should be moved out to where we create a filter.

rleungx: Can we remove it now?

PR #2664, opened 2020-07-16
Support check regions after rule updated.

Yisaer:
close https://github.com/pingcap/pd/issues/2594

For the cache package, make TTL as private and create a type struct TTLString.

After each rule created or updated, the regions of the new version key range and the old version key ranges would be stored by TTLString, and during patrol regions, coordinator would pop one group of key-range and scan the regions in these key-range and finally throw these regions into suspect regions (waiting for next round to check suspect regions in patrol regions processing)

### Check List

Tests

- Unit test


server/cluster/cluster.go
```diff
@@ -444,6 +446,32 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
 	c.suspectRegions.Remove(id)
 }
 
+// AddSuspectKeyRanges add the key ranges with the its ruleID as the key
+// The distribution for the keyRanges is like following:
+// [4][]byte: new version start key/new version end key/old version start key/old version end key
+func (c *RaftCluster) AddSuspectKeyRanges(storeID string, keyRanges [4][]byte) {
```
disksing, line 452: can `[][2][]byte` instead of `[4][]byte` be better?

server/api/rule.go
```diff
@@ -179,10 +179,22 @@ func (h *ruleHandler) Set(w http.ResponseWriter, r *http.Request) {
...
+		rule.StartKey,
+		rule.EndKey,
+		rule.StartKey,
+		rule.EndKey,
+	}
+	if oldRule != nil {
+		suspectKeyRanges[2] = oldRule.StartKey
+		suspectKeyRanges[3] = oldRule.EndKey
+	}
+	cluster.AddSuspectKeyRanges(rule.StoreKey(), suspectKeyRanges)
```
disksing, line 197: should do it when deleting a rule too

server/cluster/coordinator.go
```diff
@@ -157,6 +162,33 @@ func (c *coordinator) patrolRegions() {
...
+	if !success {
+		return
+	}
+	regionMap := map[uint64]struct{}{}
+	f := func(regions []*core.RegionInfo) {
+		for _, region := range regions {
+			regionMap[region.GetID()] = struct{}{}
+		}
+	}
+	newKeyRegions := c.cluster.ScanRegions(keyRangeGroup[0], keyRangeGroup[1], patrolScanRegionLimit)
```
disksing, line 179: The range to be very large. Use `patrolScanRegionLimit` may only scan a small part of the range.

server/cluster/cluster.go
Yisaer: good idea

server/api/rule.go
Yisaer: nice catch.

server/cluster/coordinator.go
```diff
@@ -157,6 +162,33 @@ func (c *coordinator) patrolRegions() {
...
+	f(newKeyRegions)
+	if !bytes.Equal(keyRangeGroup[0], keyRangeGroup[2]) || !bytes.Equal(keyRangeGroup[1], keyRangeGroup[3]) {
+		oldKeyRegions := c.cluster.ScanRegions(keyRangeGroup[2], keyRangeGroup[3], patrolScanRegionLimit)
+		f(oldKeyRegions)
+	}
+	var regionIDList []uint64
```
rleungx, line 185: How about initializing it with the capacity?

pkg/cache/ttl.go
```diff
@@ -153,28 +153,84 @@ func (c *TTL) doGC() {
 
 // TTLUint64 is simple TTL saves only uint64s.
 type TTLUint64 struct {
-	*TTL
+	*ttl
 }
 
 // NewIDTTL creates a new TTLUint64 cache.
 func NewIDTTL(ctx context.Context, gcInterval, ttl time.Duration) *TTLUint64 {
```
rleungx, line 160: Maybe we can use a better name for it.

pkg/cache/ttl.go
```diff
@@ -153,28 +153,84 @@ func (c *TTL) doGC() {
...
+	}
+}
+
+// Put put the string key with the value
+func (c *TTLString) Put(key string, value interface{}) {
+	c.ttl.put(key, value)
+}
+
+// PopOneKeyValue pop one key/value that is not expired
+func (c *TTLString) PopOneKeyValue() (string, interface{}) {
```
rleungx, line 223: How about moving this function to `ttl` itself?

server/cluster/coordinator.go
Yisaer: What about 1024 or 2048 for the limit?

pkg/cache/ttl.go
Yisaer: updated.

server/cluster/coordinator.go
Yisaer: updated.

server/cluster/coordinator.go
disksing: how about scan some regions and update the range so we can scan rest regions next scan?

pkg/cache/ttl.go
```diff
@@ -100,23 +100,38 @@ func (c *TTL) GetKeys() []uint64 {
...
-func (c *TTL) Remove(key uint64) {
+func (c *ttlCache) remove(key interface{}) {
 	c.Lock()
 	defer c.Unlock()
 
 	delete(c.items, key)
 }
 
+// pop one key/value that is not expired. If boolean is false, it means didnt find the valid one.
+func (c *ttlCache) pop() (interface{}, interface{}, bool) {
```
rleungx, line 111: How about adding some tests for this method?

server/api/rule_test.go
```diff
@@ -30,6 +33,8 @@ type testRuleSuite struct {
 	svr       *server.Server
 	cleanup   cleanUpFunc
 	urlPrefix string
+	// make sure different test case push and pop value into suspect KeyRange in order
+	mu sync.Mutex
```
rleungx, line 37: I think we don't need this mutex.

server/cluster/coordinator.go
Yisaer: good idea, updated.

pkg/cache/ttl.go
Yisaer: updated.

server/api/rule_test.go
Yisaer: updaetd.

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
 	c.suspectRegions.Remove(id)
 }
 
+// AddSuspectKeyRanges add the key ranges with the its ruleID as the key
+// The instance of each keyRange is like following format:
+// [2][]byte: start key/ned key
+func (c *RaftCluster) AddSuspectKeyRanges(key string, keyRanges [][2][]byte) {
+	c.Lock()
+	defer c.Unlock()
+	c.suspectKeyRanges.Put(key, keyRanges)
```
disksing, line 455: I wonder if it is necessary to put 2 ranges at the same time. Can we store `[2][]byte` instead?

Yisaer: I use `[][2][]byte` due to they have the same key, generated by rule' StoreKey and the value here is newVersionStartKey/newVersionEndKey/oldVersionStartKey/oldVersionEndKey.

Yisaer: updated.

server/cluster/coordinator.go
```diff
@@ -157,6 +162,46 @@ func (c *coordinator) patrolRegions() {
...
+	limit := 1024
+	regions := c.cluster.ScanRegions(keyRange[0], keyRange[1], limit)
+	if len(regions) < 1 {
+		return
+	}
+	// if the size equals limit which means there probably existed the rest regions.
+	// so we search the rest key range and put these key range into suspect keyRanges again.
+	if len(regions) == limit {
+		// search the biggest start Key in given regions
+		var biggestStartKey []byte
```
disksing, line 182: I think you can use the last region directly.
Also, no need to check `if len(regions) == limit`, compare last region's endKey with `keyRange[1]` is better because `len(regions) == limit` does not necessarily mean there are more regions.

server/cluster/coordinator.go
```diff
@@ -157,6 +162,46 @@ func (c *coordinator) patrolRegions() {
...
+// The regions of new version key range and old version key range would be placed into
+// the suspect regions map
+func (c *coordinator) checkSuspectKeyRanges() {
+	_, keyRange, success := c.cluster.PopOneSuspectKeyRange()
+	if !success {
+		return
+	}
+	limit := 1024
+	regions := c.cluster.ScanRegions(keyRange[0], keyRange[1], limit)
+	if len(regions) < 1 {
```
disksing, line 175: if len(regions) == 0

server/cluster/coordinator.go
```diff
@@ -157,6 +162,46 @@ func (c *coordinator) patrolRegions() {
...
+				biggestStartKey = region.GetStartKey()
+			}
+		}
+		restKeyRange := [2][]byte{
+			biggestStartKey,
+			keyRange[1],
+		}
+		c.cluster.AddSuspectKeyRanges(keyutil.BuildKeyRangeKey(biggestStartKey, keyRange[1]), restKeyRange)
+	}
+	regionMap := map[uint64]struct{}{}
```
disksing, line 194: why need the map?

server/cluster/coordinator.go
```diff
@@ -157,6 +162,46 @@ func (c *coordinator) patrolRegions() {
...
+	for _, region := range regions {
+		regionMap[region.GetID()] = struct{}{}
+	}
+	regionIDList := make([]uint64, 0, 2048)
```
disksing, line 198: why use 2018 while the limit is 1024?

server/cluster/coordinator.go
Yisaer: updated.

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
 	c.suspectRegions.Remove(id)
 }
 
+// AddSuspectKeyRanges add the key ranges with the its ruleID as the key
+// The instance of each keyRange is like following format:
+// [2][]byte: start key/ned key
```
disksing, line 451: end key

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
 	c.suspectRegions.Remove(id)
 }
 
+// AddSuspectKeyRanges add the key ranges with the its ruleID as the key
```
disksing, line 449: adds

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
...
+// The instance of each keyRange is like following format:
+// [2][]byte: start key/ned key
+func (c *RaftCluster) AddSuspectKeyRanges(key string, keyRanges [2][]byte) {
```
disksing, line 452: keyRanges -> keyRange?

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
...
+func (c *RaftCluster) PopOneSuspectKeyRange() (string, [2][]byte, bool) {
+	c.Lock()
+	defer c.Unlock()
+	key, value, success := c.suspectKeyRanges.Pop()
+	if !success {
+		return "", [2][]byte{}, false
+	}
+	v, ok := value.([2][]byte)
+	if !ok {
+		return "", [2][]byte{}, false
```
disksing, line 470: looks impossible

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
...
+// AddSuspectKeyRanges add the key ranges with the its ruleID as the key
+// The instance of each keyRange is like following format:
+// [2][]byte: start key/ned key
+func (c *RaftCluster) AddSuspectKeyRanges(key string, keyRanges [2][]byte) {
+	c.Lock()
+	defer c.Unlock()
+	c.suspectKeyRanges.Put(key, keyRanges)
+}
+
+// PopOneSuspectKeyRange get one suspect keyRange group
```
disksing, line 458: gets

server/cluster/cluster.go
```diff
@@ -444,6 +446,39 @@ func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
...
+		return "", [2][]byte{}, false
+	}
+	v, ok := value.([2][]byte)
+	if !ok {
+		return "", [2][]byte{}, false
+	}
+	return key, v, true
+}
+
+// ClearSuspectKeyRanges clear the suspect keyranges, only for unit test
```
disksing, line 475: clears

server/cluster/util.go
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+package cluster
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// BuildKeyRangeKey build a key for the give startKey and endKey
```
disksing, line 21: builds

server/cluster/coordinator.go
```diff
@@ -157,6 +162,37 @@ func (c *coordinator) patrolRegions() {
...
+	_, keyRange, success := c.cluster.PopOneSuspectKeyRange()
+	if !success {
+		return
+	}
+	limit := 1024
+	regions := c.cluster.ScanRegions(keyRange[0], keyRange[1], limit)
+	if len(regions) == 0 {
+		return
+	}
+	regionIDList := make([]uint64, 0, limit)
```
disksing, line 178: em, seems `len(regions)` is better than `limit`.

server/cluster/coordinator.go
```diff
@@ -157,6 +162,37 @@ func (c *coordinator) patrolRegions() {
...
+	for _, region := range regions {
+		regionIDList = append(regionIDList, region.GetID())
+	}
+
+	// if the last region's endkey is smaller the keyRange[1] which means there existed the remaining regions between
+	// keyRange[0] and keyRange[1] after scan regions, so we put the endkey and keyRange[1] into Suspect KeyRanges
+	lastRegion := regions[len(regions)-1]
+	if bytes.Compare(lastRegion.GetEndKey(), keyRange[1]) < 0 {
```
disksing, line 186: should consider that `nil` means +inf

server/cluster/cluster.go
Yisaer: Yes, but I still suggest remain this logic as we can't ensure the value type.

server/cluster/coordinator.go
Yisaer: updated.

pkg/cache/ttl.go
```diff
@@ -100,23 +100,38 @@ func (c *TTL) GetKeys() []uint64 {
...
 // Remove eliminates an item from cache.
-func (c *TTL) Remove(key uint64) {
+func (c *ttlCache) remove(key interface{}) {
 	c.Lock()
 	defer c.Unlock()
 
 	delete(c.items, key)
 }
 
+// pop one key/value that is not expired. If boolean is false, it means didnt find the valid one.
```
rleungx, line 110:
```suggestion
// pop one key/value that is not expired. If boolean is false, it means didn't find the valid one.
```

server/cluster/util.go
```diff
@@ -0,0 +1,24 @@
...
+
+package cluster
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// BuildKeyRangeKey builds a key for the give startKey and endKey
+func BuildKeyRangeKey(startKey, endKey []byte) string {
```
rleungx, line 22: Why we need two same `BuildKeyRangeKey` functions?

PR #2667, opened 2020-07-20
server/core: check region term before updating cache

### LINKED ISSUE #2657: Consider check the term when update the leader

There are sometimes the PD updates invalid cache information because TiKV reported old heartbeat information, we already consider check the epoch before update the region cache, but the leader changed do not changed the epoch, we need to check the term to prevent the invalid update:

image

- check the term before updating the region cache to prevent the invalid updates.

- Go
- Relative Code: https://github.com/pingcap/pd/blob/master/server/core/basic_cluster.go#L292-L314

howardlau1999:
Signed-off-by: howardlau1999 <howardlau1999@hotmail.com>

Close #2657

Collect the `term` field in the heartbeat message into `RegionInfo` and check if the term is greater than the one in cache to avoid invalid cache updates.

### Check List

Tests

- Unit test

Code changes

- Has `PreCheckPutRegion` changed
- Add field `term` in struct `RegionInfo`

### Release note

Avoid invalid cache updates after the leader of a region has changed.


disksing:
When did TiKV support sending `term` in heartbeat message? Do we need to consider backward compatibility problem?

howardlau1999:
TiKV support sending `term` in heartbeat message after v3.0. It may cause problems if TiKV older than v3.0 and newer than v3.0 are running at the same time. Heartbeat messages from old versions do not contain `term` which will default to 0 so the check will always fail if the leader transfers from new versions to old versions TiKV. This will cause the region to constantly fail to be updated.

If all TiKVs do not contain `term`, then this PR has no effect because the `term` will default to zero all the time.

server/core/region.go
```diff
@@ -31,6 +31,7 @@ import (
 // RegionInfo records detail region info.
 // Read-Only once created.
 type RegionInfo struct {
+	term              uint64
```
nolouch, line 34: we need to add it to the `Clone` function.

tests/server/cluster/cluster_test.go
```diff
@@ -1022,3 +1022,72 @@ func (s *clusterTestSuite) TestUpgradeStoreLimit(c *C) {
...
+		Term:            5,
+		ApproximateSize: 10,
+	}
+
+	region := core.RegionFromHeartbeat(regionReq)
+	err = rc.HandleRegionHeartbeat(region)
+	c.Assert(err, IsNil)
+
+	// Transfer leader
+	regionReq.Term = 6
```
nolouch, line 1081: Best to do this like `region.Clone(IncTerm)`. as the comment says, the `RegionInfo` is Read-Only once created.

howardlau1999: Didn't modify `RegionInfo` directly here. Just modified the `RegionHeartbeatRequest` used it to create new `RegionInfo`.

Issue #2621, opened 2020-07-09
Introduce auto-scale API 


rleungx:
## Development Task
To implement an auto-scale cluster, we need to support the auto-scale API. The input could be the configuration set of resources (CPU, memory, etc.). The output could be one of the configurations in the sets.

PR #2673, opened 2020-07-22
autoscaling: introduce auto scaling API

rleungx:

Solve #2621.

This PR is trying to implement the basic framework of auto scaling API.

### Check List

Tests

- Unit test


pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,107 @@
...
+type Strategy struct {
+	Rules     []*Rule     `json:"rules"`
+	Resources []*Resource `json:"resources"`
+}
+
+// Rule is a set of constraints for a kind of component.
+type Rule struct {
+	Component               string       `json:"component"`
+	CPURule                 *CPURule     `json:"cpu_rule"`
+	StorageRule             *StorageRule `json:"storage_rule"`
```
Yisaer, line 36: Do we need to add `omitempty` tag here?

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,107 @@
...
+	ScaleOutIntervalSeconds uint64       `json:"scale_out_interval_seconds"`
+	ScaleInIntervalSeconds  uint64       `json:"scale_in_interval_seconds"`
```
Yisaer, line 38: What about set these attributes as pointer and omitempty, If the user didn't set the interval seconds, we could set a default value instead of 0.

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,107 @@
...
+type CPURule struct {
+	MaxThreshold  float64  `json:"max_threshold"`
+	MinThreshold  float64  `json:"min_threshold"`
+	MaxCount      uint64   `json:"max_count"`
+	ResourceTypes []string `json:"resource_types"`
+}
+
+// StorageRule is the constraints about storage.
+type StorageRule struct {
+	MaxThreshold  float64  `json:"max_threshold"`
```
Yisaer, line 51: I think we should use min instead of max.

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,107 @@
...
+	MaxCount      uint64   `json:"max_count"`
+	ResourceTypes []string `json:"resource_types"`
+}
+
+// Resource represents a kind of resource set including CPU, memory, storage.
+type Resource struct {
+	ResourceType string `json:"resource_type"`
+	CPU          uint64 `json:"cpu"`
+	Memory       uint64 `json:"memory"`
+	Storage      uint64 `json:"storage"`
```
Yisaer, line 61: What's the basic unit, byte or kilobyte?

pkg/autoscaling/handler.go
rleungx: From the semantics of the variable, it's not appropriate to use a pointer here. I think we can use another way to adjust the default value.

pkg/autoscaling/handler.go
rleungx: IMO, byte is good.

pkg/autoscaling/handler.go
rleungx: Can you explain more why we need to replace max with min?

pkg/autoscaling/handler.go
rleungx: I think we need to keep at least one rule. Maybe StorageRule can be omitted but CPURule can not?

pkg/autoscaling/handler.go
howardlau1999: What does storage threshold mean? The ratio of used space to total space or available space to total space?

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,107 @@
...
+type StorageRule struct {
+	MaxThreshold  float64  `json:"max_threshold"`
+	MaxCount      uint64   `json:"max_count"`
+	ResourceTypes []string `json:"resource_types"`
+}
+
+// Resource represents a kind of resource set including CPU, memory, storage.
+type Resource struct {
+	ResourceType string `json:"resource_type"`
+	CPU          uint64 `json:"cpu"`
```
howardlau1999, line 59: Is the basic unit here milli-core or core?

pkg/autoscaling/handler.go
howardlau1999: But according to the design doc and the example json in the test, the basic unit here is gigabyte.

rleungx: Oh, yes. But I think byte is more appropriate? If we reach an agreement, I will modify the design doc.

pkg/autoscaling/handler.go
rleungx: What do you think? @Yisaer

pkg/autoscaling/handler.go
rleungx: The ratio of used space to total space. And I will add more comments if this API has no more problems.

pkg/autoscaling/handler.go
rleungx: IMO, the core is better since cloud TiDB may not have the concept of milli-core?

pkg/autoscaling/handler.go
howardlau1999: https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/calculate/storage.go#L75
TiDB operator is using available size to determine whether under storage pressure

pkg/autoscaling/handler.go
Yisaer: I suggest use milli-core here as  Operator supports it.

pkg/autoscaling/handler.go
Yisaer: I think we can make both rule as `omitempty` here while we could verify rule count shouldn't be less than 1 in code logic.

pkg/autoscaling/handler.go
Yisaer: > The ratio of used space to total space. And I will add more comments if this API has no more problems.

We can't use used space and total space because we can't guarantee the target storage have its own volume. Maybe there exist other storage systems share the same volume with us.

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	Component    string               `json:"component"`
+	Count        uint64               `json:"count"`
+	ResourceType string               `json:"resource_type"`
+	Labels       []*metapb.StoreLabel `json:"labels"`
+}
+
+// ResourceRequirement is a mechanism to control the resource.
+type ResourceRequirement struct {
+	Component   string `json:"component"`
+	Requirement uint64 `json:"requirement"`
```
Yisaer, line 76: I think we can `Requirement` here is misleading as we are going to use it to calculate the cpu expectation.

pkg/autoscaling/handler.go
```diff
@@ -0,0 +1,115 @@
...
+
+// StorageRule is the constraints about storage.
+type StorageRule struct {
+	MinThreshold  float64  `json:"min_threshold"`
+	MaxCount      uint64   `json:"max_count"`
+	ResourceTypes []string `json:"resource_types"`
+}
+
+// Resource represents a kind of resource set including CPU, memory, storage.
+type Resource struct {
```
Yisaer, line 58: Please add the comment to point out the basic unit for the attribute in this struct

pkg/autoscaling/service.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/unrolled/render"
+	"github.com/urfave/negroni"
+)
+
+const autoScalingPrefix = "/autoscaling"
+
+var (
+	autoscalingServiceGroup = server.ServiceGroup{
+		Name:       "autoscaling",
+		Version:    "v1",
```
nolouch, line 31: How about `v1alpha`?

rleungx: ok

pkg/autoscaling/service.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/unrolled/render"
+	"github.com/urfave/negroni"
+)
+
+const autoScalingPrefix = "/autoscaling"
+
+var (
+	autoscalingServiceGroup = server.ServiceGroup{
+		Name:       "autoscaling",
+		Version:    "v1alpha",
```
howardlau1999, line 31: is it `v1alpha1`?

rleungx: I think `v1alpha` is enough?

howardlau1999: agree

PR #2681, opened 2020-07-27
Add some useful comments for hot region scheduler.

Yisaer:
Hot region scheduler is lack of comment and hard to understand, this request add some useful comments for it.

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -95,11 +95,13 @@ type hotScheduler struct {
 	r           *rand.Rand
 
 	// states across multiple `Schedule` calls
-	pendings       [resourceTypeLen]map[*pendingInfluence]struct{}
+	pendings [resourceTypeLen]map[*pendingInfluence]struct{}
+	// regionPendings stores regionID -> [opType]Operator
 	regionPendings map[uint64][2]*operator.Operator
 
 	// temporary states but exported to API or metrics
 	stLoadInfos [resourceTypeLen]map[uint64]*storeLoadDetail
+	// pendingSums indicates the [resourceTypeLen] storeID -> pending Influence
 	pendingSums [resourceTypeLen]map[uint64]Influence
 	// config of hot scheduler
 	conf *hotRegionSchedulerConfig
@@ -190,6 +192,8 @@ func (h *hotScheduler) dispatch(typ rwType, cluster opt.Cluster) []*operator.Ope
 	return nil
 }
 
+// prepareForBalance calculate the summary of pending Influence for each store and prepare the load detail for
+// each store
 func (h *hotScheduler) prepareForBalance(cluster opt.Cluster) {
 	h.summaryPendingInfluence()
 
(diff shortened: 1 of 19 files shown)


server/statistics/region_stat_informer.go
```diff
@@ -18,7 +18,9 @@ import "github.com/pingcap/pd/v4/server/core"
 // RegionStatInformer provides access to a shared informer of statistics.
 type RegionStatInformer interface {
 	IsRegionHot(region *core.RegionInfo) bool
+	// RegionReadStats return the storeID -> write stat of peers on this store
```
rleungx, line 21:
```suggestion
	// RegionWriteStats return the storeID -> write stat of peers on this store
```

server/schedulers/hot_region.go
```diff
@@ -95,11 +95,13 @@ type hotScheduler struct {
 	r           *rand.Rand
 
 	// states across multiple `Schedule` calls
-	pendings       [resourceTypeLen]map[*pendingInfluence]struct{}
+	pendings [resourceTypeLen]map[*pendingInfluence]struct{}
+	// regionPendings stores regionID -> [opType]Operator
```
rleungx, line 99: I think a better comment is to explain how these variables work.

server/schedulers/hot_region.go
```diff
@@ -394,6 +407,7 @@ func filterHotPeers(
 	for _, peer := range peers {
 		if (kind == core.LeaderKind && !peer.IsLeader()) ||
 			peer.HotDegree < minHotDegree ||
+			// hotPeerFilterTy here is always mix currently, isHotPeerFiltered will directly return false here.
```
lhy1024, line 410:
```suggestion
			// when hotPeerFilterTy here is always mix currently, isHotPeerFiltered will directly return false here.
```

server/schedulers/hot_region.go
Yisaer: updated.

server/schedulers/utils.go
```diff
@@ -197,6 +197,8 @@ func newPendingInfluence(op *operator.Operator, from, to uint64, infl Influence)
 	}
 }
 
+// summaryPendingInfluence calculate the summary pending Influence for each store and return storeID-> Influence
+// It make each rate(key/byte/count) become (1+w) times to the origin value while f is the function to provide w(weight)
```
rleungx, line 201:
```suggestion
// It makes each rate(key/byte/count) become (1+w) times to the origin value while f is the function to provide w(weight)
```

And there is no count rate.

server/schedulers/utils.go
```diff
@@ -197,6 +197,8 @@ func newPendingInfluence(op *operator.Operator, from, to uint64, infl Influence)
 	}
 }
 
+// summaryPendingInfluence calculate the summary pending Influence for each store and return storeID-> Influence
```
rleungx, line 200:
```suggestion
// summaryPendingInfluence calculate the summary pending Influence for each store and return storeID -> Influence
```

server/schedulers/hot_region.go
```diff
@@ -842,6 +869,8 @@ func (bs *balanceSolver) calcProgressiveRank() {
 			}
 			return a - b
 		}
+		// we use DecRatio(Decline Ration) to expect that the dst store's (key/byte) rate should still be less
```
rleungx, line 872:
```suggestion
		// we use DecRatio(Decline Ratio) to expect that the dst store's (key/byte) rate should still be less
```

disksing:
It improves understandability a lot!

PR #2685, opened 2020-07-28
Add leader filter in hot region scheduler

Yisaer:
During `filterDstStores` in hot region scheduler, it didn't consider the placement rule enabled in `transferLeader` case.

This requests add `PlacementLeaderSafeguard` filter in the `filterDstStores`

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -77,9 +77,6 @@ func Target(opt opt.Options, store *core.StoreInfo, filters []Filter) bool {
 	storeAddress := store.GetAddress()
 	storeID := fmt.Sprintf("%d", store.GetID())
 	for _, filter := range filters {
-		if filter == nil {
-			continue
-		}
 		if !filter.Target(opt, store) {
 			filterCounter.WithLabelValues("filter-target", storeAddress, storeID, filter.Scope(), filter.Type()).Inc()
 			return false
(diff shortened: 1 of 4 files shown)


server/schedule/filter/filters.go
```diff
@@ -77,6 +77,9 @@ func Target(opt opt.Options, store *core.StoreInfo, filters []Filter) bool {
 	storeAddress := store.GetAddress()
 	storeID := fmt.Sprintf("%d", store.GetID())
 	for _, filter := range filters {
+		if filter == nil {
```
rleungx, line 80: When will this case happen?

Yisaer: `NewPlacementLeaderSafeguard` will return nil if placementRule is not enabled.

rleungx: But it will check if `leaderFilter != nil` before appending.

HunDunDM, line 80: When will `filter` be `nil`? I did not find this.

Yisaer: see `NewPlacementLeaderSafeguard`, it would return nil if placementRules is not enabled.

rleungx:
How about adding a unit test?

Yisaer:
> How about adding a unit test?

updated.

PR #2686, opened 2020-07-28
build: update script file

HunDunDM:

* Prevent the version information of pd-server from being dirty.
* `build.ps1` failed to run

* Synchronize the core modifications of `embed-dashboard-ui.sh` to `embed-dashboard-ui.ps1`
* Download the dashboard embed asset file to the cache directory.
* Run `go mod tidy` at the end of `make swagger-spec` to restore `go.mod`.

### Check List

Tests

  - `git clone` to get a clean code directory
  - run `make pd-server`
  - run `./build.ps1` in Windows

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,9 +96,7 @@ PD_SERVER_DEP :=
 ifneq ($(SWAGGER), 0)
 	PD_SERVER_DEP += swagger-spec
 endif
-ifneq ($(DASHBOARD), 0)
-	PD_SERVER_DEP += dashboard-ui
-endif
+PD_SERVER_DEP += dashboard-ui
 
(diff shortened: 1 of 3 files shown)


HunDunDM:
> I suspect whether the PowerShell really works, since the Unix Shell version also provides a compiling feature. There are other necessary parts in the building process missing a PowerShell edition, like `describe-dashboard` scripts. Personally I would prefer to drop the PowerShell support and encourage people to use WSL instead, which works like a charm. The rest LGTM

@lhy1024 helped me test it under windows and it can compile successfully. I think the significance of this script is to provide minimal support for windows. This does not conflict with our recommendation to develop on WSL.

Issue #2680, opened 2020-07-26
Placement Rules: support atomic helper api


xhebox:
### Describe your feature request related problem

Hi, I am the mentee of the TiDB project `Define the Placement Rules in Sql`.

The current PD api makes it very hard to support this feature. Since rules will not be modified frequently, it is fine to have some extra helper APIs, such that we can avoid many problems.

For example, one statement may contain multiple rule definitions. The only way is to send HTTP request one rule by one rule. What if TiDB is down in the middle? Of course, it is possible to delete posted rules somehow later(a little bit hard). But posted rule may take effect before they are deleted.

Another example, it is good to use PD as the only source for placement rules storage. TiDB is distributed, thus there may be problems on synchronization/consistency. To update a rule, we need to get/modify/repost the old rule from PD. There is no guarantee for atomicity.

In another view, it is far better to send one HTTP request than sending many, many more requests.

I've checked that `RuleManager` is using an `RWMutex`. So it should not be hard to add new APIs.

### Describe the feature you'd like

+ An API for updating rules(overwrite modified part only)
+ An API for batch operations, uploading multiple add/delete/update in one HTTP request. So it ensures atomicity, and reduce the number of HTTP requests.

### PR track

#2688 #2699 

These api is not completely safe. See https://github.com/xhebox/pd/blob/delete/server/schedule/placement/rule_manager.go#L294-L299.

nolouch:
Thanks, @xhebox. you are right. maybe you can help us to improve it if you are interested.

xhebox:
@nolouch Is there anything special that I need to take care of?

nolouch:
Maybe we can make it can be able to process multiple rules at once firstly. cc @disksing  @Yisaer

xhebox:
Sounds good to me. My basic idea is to a generalized version of `SetRule`: `SetRules`. Before every rule is set in the map, the old value is pushed into a slice as a backup. In case there is an error, just apply the old values. Otherwise, we are done.

nolouch:
@xhebox Do you consider handling failure situations, such as PD leader has down?

disksing:
> Sounds good to me. My basic idea is a generalized version of `SetRule`: `SetRules`. Before every rule is set in the map, the old value is pushed into a slice as a backup. In case there is an error, just apply the old values(the slice). Otherwise, we are done.

Good idea.

xhebox:
[...]

PR #2688, opened 2020-07-29
placement: api for adding rules in one request

xhebox:

Related to #2680.

Now one can post an array of rules to `/config/rules`, to set all rules in one request. In case there is an error, all modifications are cancelled.

### Check List

Tests

- Unit test
- Integration test

Code changes

### Release note

- atomic POST api `/config/rules` for setting all rules in one request, parameter is an array of rules

### DIFF
--- a/server/schedule/placement/rule_manager.go
+++ b/server/schedule/placement/rule_manager.go
@@ -275,12 +275,13 @@ func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *Regio
 	return FitRegion(stores, region, rules)
 }
 
-func (m *RuleManager) setRule(rule *Rule) *Rule {
+func (m *RuleManager) swapRule(rule *Rule) *Rule {
 	old := m.rules[rule.Key()]
 	m.rules[rule.Key()] = rule
 	return old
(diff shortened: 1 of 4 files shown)


disksing:
Thanks, @xhebox , very high-quality code. Not only correct but also elegant.

server/schedule/placement/rule_manager.go
```diff
@@ -273,3 +274,53 @@ func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *Regio
...
+		}
+	}
+
+	m.Lock()
+	defer m.Unlock()
+
+	oldRules := make(map[[2]string]*Rule)
+
+	for _, rule := range rules {
+		oldRules[rule.Key()] = m.swapRule(rule)
```
nolouch, line 299: May `oldRules` be nils? will it build a `rulelist` successfully if `oldRules` contains nil pointer? we need to add a test for it.

xhebox: Yes. The first case, actually, should already have `oldRules` full of nils. But its ID is duplicated with `SetRule`.

PR #2658, opened 2020-07-16
Define error code

ZenoTan:
### What problem does this PR solve?
Assign error code to each type of log errors in pd. Closes https://github.com/pingcap/pd/issues/2449.

### What is changed and how it works?
PD error code is from 9800 to 10000, and we reserve double code space for future use.
Log errors are classified by types, such as storage, io, interal... Each error code has its name by its class plus some details.
For example, StorageEtcdLoadErr means it is related to storage, uses etcd, happens when users try to load something.

### Check List

Tests

- No code

Code changes

- New pkg for error code definitions.

Side effects

Related changes

Issue #2449, opened 2020-05-25
Define some error codes to classify the existed errors


rleungx:
We need to define some error codes to make each error has its own code. It can help improve usability since we can give the reason why this error happened, and how to make a walk around according to the error codes.

PR #2695, opened 2020-07-29
RFC log format

ZenoTan:

Change log format according to RFC. Solve #2449 

Rewrite `log.Fatal` or `log.Error`. Append rfc format error information to log. Use pingcap/errors.
RFC example is **[error="[PD:format:ErrFormatParseHistoryIndex] parse history index error"]**.
RFC code must have **[Registry:Class:Error]**.
Log output is like:
```
[2020/07/30 13:18:25.243 +08:00] [FATAL] [main.go:95] ["join meet error"] [error="join self is forbidden"] [error="[PD:storage:ErrStorageEtcdLoad] load config from storage etcd error"] [stack="github.com/pingcap/log.Fatal\n\t/Users/zeyuantan/go/pkg/mod/github.com/pingcap/log@v0.0.0-20200511115504-543df19646ad/global.go:59\nmain.main\n\t/Users/zeyuantan/Temp/my/err/cmd/pd-server/main.go:95\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"]
```

### Check List

Tests

- No code

Code changes

- Has log format change

Side effects

Related changes

#2658


client/base_client.go
```diff
@@ -15,6 +15,7 @@ package pd
 
 import (
 	"context"
+	errs "github.com/pingcap/pd/v4/pkg/errors"
```
disksing, line 18: em, why not name the package `github.com/pingcap/pd/v4/pkg/errs` directly?

ZenoTan: Originally, I used "errors" in #2658 . Then when I changed log code, I found github.com/pkg/errors was also used. So... = =

HunDunDM: Need to remove `pkg/errors`?

ZenoTan: `a`

ZenoTan: Some places use `pkg/errors` like `error.WithStack`.

Issue #2694, opened 2020-07-29
Use the batch API to save rules


disksing:
## Development Task

Now that we support updating rules in batch (https://github.com/pingcap/pd/pull/2688). `pd-ctl placement save` can benefit from it.

PR #2698, opened 2020-07-29
pd-ctl: make use of new batch ruledef api

xhebox:

#2694, as title, save rules using the new api.

### Check List

Tests

- Manual test

Test save rules with `[{count=3}, {count=0}, {count=0}, {count=3}]`, and `[{count=3}, {count=3}, {count=0}]`.

### Release note

- No release note

### DIFF
--- a/tools/pd-ctl/pdctl/command/config_command.go
+++ b/tools/pd-ctl/pdctl/command/config_command.go
@@ -16,7 +16,6 @@ package command
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"path"
--- a/tools/pd-ctl/pdctl/command/config_command.go
+++ b/tools/pd-ctl/pdctl/command/config_command.go
@@ -521,10 +521,10 @@ func putPlacementRulesFunc(cmd *cobra.Command, args []string) {
 	}
 
 	n := len(validRules)
-	for k := 0; k < n ; k++ {
+	for k := 0; k < n; k++ {
 		r := validRules[k]
 		if r.Count == 0 {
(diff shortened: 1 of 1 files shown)


disksing:
It seems that we need to find a way to make delete and set atomic too. 😢

xhebox:
If you want so... It can be done in the next PR of deletion. In that PR, I can introduce a `DELETE` method `/config/rules`. Maybe together with a real `/config/batch` API, supporting `delete` and `add` by a slice of new struct `placement.Batch` or so.

tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -507,31 +507,30 @@ func putPlacementRulesFunc(cmd *cobra.Command, args []string) {
...
-			b, _ := json.Marshal(r)
-			_, err = doRequest(cmd, rulePrefix, http.MethodPost, WithBody("application/json", bytes.NewBuffer(b)))
-			if err != nil {
-				fmt.Printf("failed to save rule %s/%s: %v\n", r.GroupID, r.ID, err)
-				return
-			}
-			fmt.Printf("saved rule %s/%s\n", r.GroupID, r.ID)
+
+	validOpts := opts[:0]
+	for _, Op := range opts {
```
JmPotato, line 518: Maybe using `op` instead of `Op` will be more in line with the Go's official standard?

nolouch:
LGTM

xhebox:
Integration is excepted to fail, there is one PR waiting to merge on TiDB. The failed test seems related `/config/label-property`. Do I need to do anything on this PR?

xhebox:
@ti-srebot /run-all-tests

PR #2699, opened 2020-07-29
placement: support batch deletions, with insertions

xhebox:

Related to #2680.

Now one can post an array of operations(insert/delete) to `/config/rules/batch` in one request. In case there is an error, all modifications are cancelled.

Two notable points:
1. Operations should be independent. Their IDs should be different.
2. I noticed that this PR and the previous PR is not completely safe. See https://github.com/xhebox/pd/blob/delete/server/schedule/placement/rule_manager.go#L294-L299.

### Check List

Tests

- Unit test
- Integration test

Code changes

### Release note

- atomic POST api `/config/rules/batcch` for batch(insert/delete) rules in one request

### DIFF
--- a/server/schedule/placement/rule_manager.go
+++ b/server/schedule/placement/rule_manager.go
@@ -385,9 +385,9 @@ const (
 // Batch is for batching placement rule actions. The action type is
 // distinguished by the field `Action`.
 type Batch struct {
-	*Rule               // infomation of the placement rule
+	*Rule               // information of the placement rule to add/delete
 	Action  BatchAction `json:"action"`   // the operation type
-	MatchID bool        `json:"match_id"` // only take effect if action is deletion
+	MatchID bool        `json:"match_id"` // only take effect if action is deletion, will match the prefix of id
 }
 
 // Batch execute a series of actions at once.
--- a/server/schedule/placement/rule_manager.go
+++ b/server/schedule/placement/rule_manager.go
@@ -372,13 +372,13 @@ func (m *RuleManager) delRule(t *Batch, oldRules map[[2]string]*Rule) {
 	}
 }
 
-// The operation type for Batch
+// BatchAction indicates the operation type
 type BatchAction byte
 
 const (
-	// Add a placement rule, only need to specify the field *Rule
+	// BatchAdd a placement rule, only need to specify the field *Rule
 	BatchAdd BatchAction = iota + 1
-	// Del a placement rule, only need to specify the field `GroupID`, `ID`, `MatchID`
+	// BatchDel a placement rule, only need to specify the field `GroupID`, `ID`, `MatchID`
 	BatchDel
 )
 
--- a/server/schedule/placement/rule_manager.go
+++ b/server/schedule/placement/rule_manager.go
@@ -349,48 +349,46 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
 	return nil
 }
 
-func (m *RuleManager) emptyRule(group, id string) ([2]string, *Rule) {
+func (m *RuleManager) delRule(group, id string, oldRules map[[2]string]*Rule) {
 	key := [2]string{group, id}
 	old, ok := m.rules[key]
 	if ok {
 		delete(m.rules, key)
 	}
-	return key, old
+	oldRules[key] = old
 }
 
-func (m *RuleManager) delRule(t *Batch, oldRules map[[2]string]*Rule) {
-	if !t.MatchID {
-		k, r := m.emptyRule(t.GroupID, t.ID)
-		oldRules[k] = r
+func (m *RuleManager) delRuleByPrefix(t *Batch, oldRules map[[2]string]*Rule) {
+	if !t.DeleteByIDPrefix {
+		m.delRule(t.GroupID, t.ID, oldRules)
 	} else {
 		for key := range m.rules {
 			if key[0] == t.GroupID && strings.HasPrefix(key[1], t.ID) {
-				k, r := m.emptyRule(key[0], key[1])
-				oldRules[k] = r
+				m.delRule(key[0], key[1], oldRules)
 			}
 		}
 	}
 }
 
 // BatchAction indicates the operation type
-type BatchAction byte
+type BatchAction string
 
 const (
 	// BatchAdd a placement rule, only need to specify the field *Rule
-	BatchAdd BatchAction = iota + 1
+	BatchAdd BatchAction = "insert"
 	// BatchDel a placement rule, only need to specify the field `GroupID`, `ID`, `MatchID`
-	BatchDel
+	BatchDel BatchAction = "delete"
 )
 
 // Batch is for batching placement rule actions. The action type is
 // distinguished by the field `Action`.
 type Batch struct {
-	*Rule               // information of the placement rule to add/delete
-	Action  BatchAction `json:"action"`   // the operation type
-	MatchID bool        `json:"match_id"` // only take effect if action is deletion, will match the prefix of id
+	*Rule                        // information of the placement rule to add/delete
+	Action           BatchAction `json:"action"`       // the operation type
+	DeleteByIDPrefix bool        `json:"delete_by_id"` // if action == delete, delete by the prefix of id
 }
 
-// Batch execute a series of actions at once.
+// Batch executes a series of actions at once.
 func (m *RuleManager) Batch(todo []Batch) error {
 	for _, t := range todo {
 		switch t.Action {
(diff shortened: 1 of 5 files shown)


server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
+}
 
-	if err != nil {
-		for key, rule := range oldRules {
-			if rule == nil {
-				delete(m.rules, key)
-			} else {
-				m.rules[key] = rule
+// BatchAction indicates the operation type
+type BatchAction byte
```
disksing, line 376: seems better to use `string`.

server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
-				break
+	if err := m.tryBuildSave(oldRules); err != nil {
+		return err
+	}
+
+	log.Info("placement rule updated", zap.String("rules", fmt.Sprint(rules)))
+	return nil
+}
+
+func (m *RuleManager) emptyRule(group, id string) ([2]string, *Rule) {
```
disksing, line 352: I think we can make it more like `addRule` by changing it to:
```
func (m *RuleManager) delRule(group, id string, oldRules map[[2]string]*Rule) {
```

server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
+	// BatchDel a placement rule, only need to specify the field `GroupID`, `ID`, `MatchID`
+	BatchDel
+)
+
+// Batch is for batching placement rule actions. The action type is
+// distinguished by the field `Action`.
+type Batch struct {
+	*Rule               // information of the placement rule to add/delete
+	Action  BatchAction `json:"action"`   // the operation type
+	MatchID bool        `json:"match_id"` // only take effect if action is deletion, will match the prefix of id
```
disksing, line 390: IMO it is better to strict match id by default. So I'm proposing to change it to something like `DeleteByPrefix`.

server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
+}
+
+// Batch execute a series of actions at once.
```
disksing, line 393: executes

server/schedule/placement/rule_manager.go
```diff
@@ -275,10 +276,51 @@ func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *Regio
...
+				r := Rule{
+					GroupID: key[0],
+					ID:      key[1],
+				}
+				err = m.store.DeleteRule(r.StoreKey())
+			}
+			if err != nil {
+				// TODO: it is not completely safe
+				// 1. in case that half of rules applied, error.. we have to cancel persisted rules
+				// but that may fail too, causing memory/disk inconsistency
```
disksing, line 296: It is admirable for you to notice this issue! I think we can add some note in the API annotation.

server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
+		}
+	}
+
+	if err := m.tryBuildSave(oldRules); err != nil {
 		return err
 	}
 
-	m.ruleList = ruleList
-	log.Info("placement rule updated", zap.String("rules", fmt.Sprint(rules)))
+	log.Info("placement rules updated", zap.String("rules", fmt.Sprint(todo)))
```
disksing, line 423: rules -> batch?

server/schedule/placement/rule_manager.go
```diff
@@ -296,31 +338,88 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
+const (
+	// BatchAdd a placement rule, only need to specify the field *Rule
+	BatchAdd BatchAction = iota + 1
+	// BatchDel a placement rule, only need to specify the field `GroupID`, `ID`, `MatchID`
+	BatchDel
+)
+
+// Batch is for batching placement rule actions. The action type is
+// distinguished by the field `Action`.
+type Batch struct {
```
Yisaer, line 385: I have to say `Batch` here is a bit of misleading as one Batch only contains one rule.

server/schedule/placement/rule_manager.go
xhebox: I see. It is 'match id' by default. This name is confusing. Fixed.

server/schedule/placement/rule_manager.go
```diff
@@ -412,14 +410,14 @@ func (m *RuleManager) Batch(todo []Batch) error {
 		case BatchAdd:
 			m.addRule(t.Rule, oldRules)
 		case BatchDel:
-			m.delRule(&t, oldRules)
+			m.delRuleByPrefix(&t, oldRules)
```
disksing, line 413: The name is `delRuleByPrefix` but it may not actually delete by the prefix. It is confusing.
How about `delRule` -> `delRuleByID` and `delRuleByPrefix` -> `delRule`.

server/schedule/placement/rule_manager.go
```diff
@@ -349,48 +349,46 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
 }
 
 // BatchAction indicates the operation type
-type BatchAction byte
+type BatchAction string
 
 const (
 	// BatchAdd a placement rule, only need to specify the field *Rule
-	BatchAdd BatchAction = iota + 1
+	BatchAdd BatchAction = "insert"
```
disksing, line 378: Please make it consistent. Either use `add` or `insert`, not use them both.

server/schedule/placement/rule_manager.go
```diff
@@ -349,48 +349,46 @@ func (m *RuleManager) SetRules(rules []*Rule) error {
...
 
 // Batch is for batching placement rule actions. The action type is
 // distinguished by the field `Action`.
 type Batch struct {
-	*Rule               // information of the placement rule to add/delete
-	Action  BatchAction `json:"action"`   // the operation type
-	MatchID bool        `json:"match_id"` // only take effect if action is deletion, will match the prefix of id
+	*Rule                        // information of the placement rule to add/delete
+	Action           BatchAction `json:"action"`       // the operation type
+	DeleteByIDPrefix bool        `json:"delete_by_id"` // if action == delete, delete by the prefix of id
```
disksing, line 388: `delete_by_id` -> `delete_by_id_prefix`?

server/schedule/placement/rule_manager.go
xhebox: Nice idea.

server/schedule/placement/rule_manager.go
xhebox: Any good idea? `RuleOp`?

disksing:
> It seems currently one batch request couldn't guarantee that the inserting and deleting rules can be executed atomically. I think we can leave an issue to solve how to provide atomically batch rules api (if needed).

why cannot ensure atomic?

Yisaer:
I think the contributor have explained on the following. @disksing 

```sh
				// TODO: it is not completely safe
				// 1. in case that half of rules applied, error.. we have to cancel persisted rules
				// but that may fail too, causing memory/disk inconsistency
				// either rely a transaction API, or clients to request again until success
				// 2. in case that PD is suddenly down in the loop, inconsistency again
				// now we can only rely clients to request again
```

server/schedule/placement/rule_manager.go
Yisaer: `RuleOp` is ok to me. one `Batch`contains multiple `RuleOp` seems more reasonable.

server/schedule/placement/rule_manager.go
```diff
@@ -275,10 +276,51 @@ func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *Regio
...
+				err = m.store.DeleteRule(r.StoreKey())
+			}
+			if err != nil {
+				// TODO: it is not completely safe
+				// 1. in case that half of rules applied, error.. we have to cancel persisted rules
+				// but that may fail too, causing memory/disk inconsistency
+				// either rely a transaction API, or clients to request again until success
+				// 2. in case that PD is suddenly down in the loop, inconsistency again
+				// now we can only rely clients to request again
+				break
```
Yisaer, line 300: If a transaction API is needed, please create issue to track this (ignore me if already existed).

xhebox: This is unexpected. Anyway, I think it is enough to use my old issue to track it.

nolouch:
LGTM

Issue #2704, opened 2020-08-03
Refine the log according to RFC


rleungx:
## Development Task
According to https://github.com/pingcap/tidb/blob/master/docs/design/2020-05-08-standardize-error-codes-and-messages.md, we need to refine our error/fatal level log, but we still need to follow https://github.com/tikv/rfcs/blob/master/text/2018-12-19-unified-log-format.md. 

Here are some packages we need to refine:

- [x] client 
- [x] apiutil
- [x] adapter
- [x] etcdutil
- [x] metricutil
- [x] server #2713
- [x] member
- [x] cluster
- [x] placement
- [x] schedulers #2705
- [x] tso 
- [x] syncer
- [x] main
- [x] api with schedulers

lhy1024:
@lhy1024 pick schedulers

JmPotato:
@JmPotato pick up tso

howardlau1999:
@howardlau1999 deals with cluster

Yisaer:
@Yisaer pickup etcdutil and metricutil

ZenoTan:
@ZenoTan pick up server

PR #2707, opened 2020-08-03
*: refine member/tso/adapter logs

rleungx:

Part of #2704.

This PR refines logs in `member/tso/adapter` packages

### Check List

Tests

- Unit test

Related changes


pkg/errs/errs.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 PingCAP, Inc.
```
lhy1024, line 1: This file does not seem to include previous error types

rleungx: > This file does not seem to include previous error types

Yes. I think maybe define error by the package is a better choice?

ZenoTan: Maybe we should deal with this error file together at last.

rleungx: Of course

PR #2723, opened 2020-08-05
autoscaling: update auto scaling API

rleungx:

Related to #2621. After #2673, there are still some problems which are existed operator or TiDB cloud implementation.

This PR updates the API to make it easier.

### Check List

Tests

- Unit test

Code changes

- Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/pingcap/pd/blob/master/docs/development.md#updating-api-documentation))

### DIFF
--- a/tests/autoscaling/autoscaling_test.go
+++ b/tests/autoscaling/autoscaling_test.go
@@ -53,13 +53,13 @@ func (s *apiTestSuite) TestAPI(c *C) {
 
 	var jsonStr = []byte(`
 {
+    "max_cpu_quota":40000,
     "rules":[
         {
             "component":"tikv",
             "cpu_rule":{
                 "max_threshold":0.8,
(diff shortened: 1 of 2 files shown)


tests/autoscaling/autoscaling_test.go
```diff
@@ -89,35 +88,20 @@ func (s *apiTestSuite) TestAPI(c *C) {
...
-            "count":2
-        }
-    ],
     "resources":[
         {
             "resource_type":"resource_a",
             "cpu":1,
             "memory":8,
-            "storage":1000
+            "storage":1000,
```
howardlau1999, line 96: The unit here is not the same as the definition

rleungx: It is only a test for testing the function of serialization.

nolouch:
LGTM

PR #2725, opened 2020-08-05
Fix balance-leader won't work when placement rule enabled

Yisaer:
When placement rule enabled, balance-leader won't work.

The region copy logic in ruleLeaderFilter's Target function is wrong. This request fix this bug.

### Check List

Tests

- Unit test

Related changes

### Release note

- Fix the bug that balance-leader won't work when placement rule enabled (imported from v4.0.3)

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -467,7 +467,7 @@ func (f *ruleLeaderFitFilter) Target(opt opt.Options, store *core.StoreInfo) boo
 	var targetPeer *metapb.Peer
 	targetPeer = f.region.GetStorePeer(store.GetID())
 	if targetPeer == nil {
-		log.Warn("ruleLeaderFitFilter could't find peer on target Store", zap.Uint64("newStoreID", store.GetID()))
+		log.Warn("ruleLeaderFitFilter could't find peer on target Store", zap.Uint64("target-store", store.GetID()))
 		return false
 	}
 	region := f.region.Clone(core.WithLeader(targetPeer))
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -16,7 +16,6 @@ package filter
 import (
 	"fmt"
 
-	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	"github.com/pingcap/pd/v4/pkg/slice"
 	"github.com/pingcap/pd/v4/server/core"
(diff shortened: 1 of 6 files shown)


server/schedule/filter/filters.go
```diff
@@ -461,9 +464,18 @@ func (f *ruleLeaderFitFilter) Source(opt opt.Options, store *core.StoreInfo) boo
...
-	region := f.region.Clone(core.WithReplacePeerStore(f.oldLeaderStoreID, store.GetID()), core.WithLeader(&oldLeaderPeerReplica))
+	var targetPeer *metapb.Peer
+	for _, peer := range f.region.GetPeers() {
+		if peer.StoreId == store.GetID() {
+			targetPeer = peer
+			break
+		}
+	}
+	if targetPeer == nil {
+		log.Info("ruleLeaderFitFilter could't find peer on target Store", zap.Uint64("newStoreID", store.GetID()))
```
rleungx, line 475: Will the log be noisy?

server/schedule/filter/filters.go
```diff
@@ -461,9 +464,18 @@ func (f *ruleLeaderFitFilter) Source(opt opt.Options, store *core.StoreInfo) boo
 }
 
 func (f *ruleLeaderFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
-	oldLeaderPeerReplica := *f.region.GetLeader()
-	oldLeaderPeerReplica.StoreId = store.GetID()
-	region := f.region.Clone(core.WithReplacePeerStore(f.oldLeaderStoreID, store.GetID()), core.WithLeader(&oldLeaderPeerReplica))
+	var targetPeer *metapb.Peer
+	for _, peer := range f.region.GetPeers() {
```
disksing, line 468: f.region.GetStorePeer()

server/schedule/filter/filters.go
disksing: I think normally it should not happen. Maybe `log.Warn` here.

server/schedule/filter/filters.go
Yisaer: updated.

server/schedule/filter/filters.go
Yisaer: updated.

server/schedule/filter/filters.go
```diff
@@ -461,9 +464,13 @@ func (f *ruleLeaderFitFilter) Source(opt opt.Options, store *core.StoreInfo) boo
 }
 
 func (f *ruleLeaderFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
-	oldLeaderPeerReplica := *f.region.GetLeader()
-	oldLeaderPeerReplica.StoreId = store.GetID()
-	region := f.region.Clone(core.WithReplacePeerStore(f.oldLeaderStoreID, store.GetID()), core.WithLeader(&oldLeaderPeerReplica))
+	var targetPeer *metapb.Peer
+	targetPeer = f.region.GetStorePeer(store.GetID())
```
disksing, line 468:
```suggestion
	targetPeer := f.region.GetStorePeer(store.GetID())
```

server/schedule/filter/filters.go
```diff
@@ -467,7 +467,7 @@ func (f *ruleLeaderFitFilter) Target(opt opt.Options, store *core.StoreInfo) boo
 	var targetPeer *metapb.Peer
 	targetPeer = f.region.GetStorePeer(store.GetID())
 	if targetPeer == nil {
-		log.Warn("ruleLeaderFitFilter could't find peer on target Store", zap.Uint64("newStoreID", store.GetID()))
+		log.Warn("ruleLeaderFitFilter could't find peer on target Store", zap.Uint64("target-store", store.GetID()))
```
rleungx, line 470:
```suggestion
		log.Warn("ruleLeaderFitFilter couldn't find peer on target Store", zap.Uint64("target-store", store.GetID()))
```

PR #2727, opened 2020-08-06
pd-analysis: Fix mv peer log parse bug

ZenoTan:

When parse log by pd-analysis, run
`./bin/pd-analysis -input temp_log -operator balance-region -style transfer-counter`
get
`[2020/08/06 12:00:39.900 +08:00] [FATAL] [main.go:73] ["strconv.ParseInt: parsing \"\": invalid syntax"] [stack="main.main\n\t/Users/zeyuantan/Temp/my/pd/tools/pd-analysis/main.go:73\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"]`

Move peer operator log is `store [1] to [2]` instead of `store 1 to 2`, fix parsing format.

### Check List

Tests

- Unit test

Code changes

- If operator needs move peer, fix store parse format.

Side effects

Related changes

### Release note

- Fix pd-analysis parse log bug.

### DIFF
--- a/tools/pd-analysis/analysis/parse_log.go
+++ b/tools/pd-analysis/analysis/parse_log.go
@@ -25,6 +25,8 @@ import (
 )
 
 var supportOperators = []string{"balance-region", "balance-leader", "transfer-hot-read-leader", "move-hot-read-region", "transfer-hot-write-leader", "move-hot-write-region"}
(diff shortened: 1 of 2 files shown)


ZenoTan:
@lhy1024 PTAL

disksing:
I think it is better to unify log output.
Or make `[` and `]` optional in the regex pattern (`\[?[0-9]+\]?`)

PR #2728, opened 2020-08-06
kv, core: unify behaviour of different storages and bug fixes

disksing:
- There are different implementations of kv storage while their behaviours are different somehow
For example, when the key does not exist, etcd_kv returns `"", nil` but leveldb_kv returns `"", ErrNotFound`

- `LoadRange` has a potential bug (when one path is the prefix of another path (no problem for PD now))
For example, if there are keys `[foo/a, foo/b, foo-a/a, foo-a/b]` in storage, scan `["foo/", "foo0")` returns all the 4 keys.

- unit test coverage is low

- unify behaviour when key does not exist
- fix `LoadRange` problem and unify all usages
- test all storages in unit tests
- Fix the bug that the `adjacent-region-scheduler` may have no name (causing its config stores in `scheduler_config` instead of `scheduler_config/adjacent-region)

### Check List

Tests
- Unit test

### Release note
- No release note


server/core/storage.go
```diff
@@ -237,18 +237,13 @@ func (s *Storage) DeleteRule(ruleKey string) error {
...
-	endKey := rulesPath + "0"
+	nextKey := rulesPath + "/"
+	endKey := clientv3.GetPrefixRangeEnd(nextKey)
 	for {
 		keys, values, err := s.LoadRange(nextKey, endKey, minKVRangeLimit)
 		if err != nil {
 			return false, err
 		}
-		if len(keys) == 0 {
-			return false, nil
```
rleungx, line 250: It seems that the first return value is omitted?

disksing: yes, seems not very useful. I think we can remove it.

PR #2729, opened 2020-08-06
Update kvproto in dependencies

HunDunDM:

* Update `kvproto` to develop `JointConsensus`

* [x] Update `kvproto`: wait for [kvproto/pull/658](https://github.com/pingcap/kvproto/pull/658)
* Compatible with the change from `IsLearner` to `Role`.

### Check List

Tests

- Unit test


server/versioninfo/feature.go
```diff
@@ -37,13 +36,22 @@ const (
 	// BatchSplit can speed up the region split.
 	// and PD will response the BatchSplit request.
 	BatchSplit
+	Version3_0
```
rleungx, line 39: Why do we need 3_0/4_0?

HunDunDM: This part of the code is split into another PR.

pkg/metautil/peer.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package metautil
+
+import (
+	"github.com/pingcap/kvproto/pkg/metapb"
+)
+
+// IsLearner judges whether the Peer's Role is Learner.
+func IsLearner(peer *metapb.Peer) bool {
```
rleungx, line 21: Why do we need a `metautil` package, can we put it into `core/region.go`?

HunDunDM: I'll put it into `core/meta.go`.

PR #2734, opened 2020-08-07
placement: support RuleGroup properties

disksing:

Tracking issue: https://github.com/pingcap/pd/issues/2721

Add RuleGroup and consider it when calculating rule to fit a region. 
(Before complete the feature fully, it won't take any affect)

### Check List

Tests
- Unit test

### Release note
- No release note

### DIFF
--- a/server/schedule/placement/rule.go
+++ b/server/schedule/placement/rule.go
@@ -39,7 +39,7 @@ func validateRole(s PeerRoleType) bool {
 
 // Rule is the placement rule that can be checked against a region. When
 // applying rules (apply means schedule regions to match selected rules), the
-// apply order is defined by the tuple [GroupID, Index, ID].
+// apply order is defined by the tuple [GroupIndex, GroupID, Index, ID].
 type Rule struct {
 	GroupID          string            `json:"group_id"`                    // mark the source that add the rule
 	ID               string            `json:"id"`                          // unique ID within a group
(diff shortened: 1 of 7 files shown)


server/schedule/placement/rule.go
```diff
@@ -54,26 +54,46 @@ type Rule struct {
...
+func (r *Rule) groupIndex() int {
+	if r.group != nil {
+		return r.group.Index
+	}
+	return 0
+}
+
+// RuleGroup defines properties of a rule group.
+type RuleGroup struct {
+	ID       string `json:"id,omitempty"`
```
JmPotato, line 85: Does this field conflict with `Rule. GroupID`?

disksing: Rules and RuleGroups will be maintained and stored separately. ID is used to associate them together. (consider it as a foreign key)

JmPotato: I got it.

server/schedule/placement/rule_manager.go
```diff
@@ -365,3 +367,14 @@ func (m *RuleManager) Batch(todo []RuleOp) error {
 	log.Info("placement rules updated", zap.String("batch", fmt.Sprint(todo)))
 	return nil
 }
+
+func (m *RuleManager) rebuildRuleList() (ruleList, error) {
+	for _, r := range m.rules {
+		r.group = m.groups[r.GroupID]
```
JmPotato, line 373: When will the `RuleManager.Group` (i.e `m.groups`) be initialized?

disksing: RuleManager.Initialize(): load from etcd // not included in this PR

JmPotato: Ok.

PR #2736, opened 2020-08-07
Add priority schedule after scatter check failed.

Yisaer:
Before region scattering, scatter will check the region status, if the region is not valid, scatter will refuse to scatter the region and wait coordinator to fix it.

This request let scatter put the invalid region into high priority schedule queue instead of discarding it directly.

ref https://github.com/pingcap/pd/issues/2733

Add a new interface called HighPrioritySchedule to handle the regions for high priority scheduling need.

### Check List

Tests

- Unit test

Related changes


tests/server/cluster/cluster_work_test.go
```diff
@@ -158,7 +158,7 @@ func (s *clusterWorkerTestSuite) TestSuspectRegions(c *C) {
 	c.Assert(err, IsNil)
 	ids := []uint64{regions[0].GetMeta().GetId(), res.Ids[0].NewRegionId, res.Ids[1].NewRegionId}
 	sort.Slice(ids, func(i, j int) bool { return ids[i] < ids[j] })
-	suspects := rc.GetSuspectRegions()
+	suspects := rc.GetHighPriorityScheduleRegions()
```
rleungx, line 161: why the name here is still suspects?

server/cluster/cluster.go
```diff
@@ -423,24 +423,24 @@ func (c *RaftCluster) SetStorage(s *core.Storage) {
...
+// GetHighPriorityScheduleRegions gets all suspect regions.
+func (c *RaftCluster) GetHighPriorityScheduleRegions() []uint64 {
 	c.RLock()
 	defer c.RUnlock()
 	return c.suspectRegions.GetAllID()
 }
 
-// RemoveSuspectRegion removes region from suspect list.
-func (c *RaftCluster) RemoveSuspectRegion(id uint64) {
+// CancelRegionHighPrioritySchedule removes region from suspect list.
```
rleungx, line 442: This function name is inconsistency with the other two.

server/cluster/cluster.go
```diff
@@ -423,24 +423,27 @@ func (c *RaftCluster) SetStorage(s *core.Storage) {
...
 
-// AddSuspectRegions adds regions to suspect list.
-func (c *RaftCluster) AddSuspectRegions(ids ...uint64) {
+// AddHighPriorityScheduleRegions adds regions to be scheduled with high priority.
+// it implements HighPriority.AddHighPriorityScheduleRegions
+func (c *RaftCluster) AddHighPriorityScheduleRegions(ids ...uint64) {
 	c.Lock()
 	defer c.Unlock()
 	for _, id := range ids {
 		c.suspectRegions.Put(id, nil)
```
rleungx, line 432: Does it need to be changed?

Yisaer: 1. changing name here is to implement `AddHighPriorityScheduleRegions`
2. `AddHighPriorityScheduleRegions` is more clear than `AddSuspectRegions` to describe what this method actually do.

server/schedule/priorityscheduling/priority_schedule.go
```diff
@@ -0,0 +1,23 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package priorityscheduling
```
disksing, line 14: package name not easy to read. maybe no need to use a new package.

server/cluster/cluster.go
rleungx: What I mean is inside of this function, we still call `c.suspentRegions`.

disksing:
What is the difference between the **real** HighPriorityScheduleRegions and the mock one?
IMO they are basically the same. So it is better to implement it some where, and use it in both cluster and mock cluster. No need to define the interface either.

PR #2738, opened 2020-08-10
autoscaling: implement the basic framework

### LINKED ISSUE #2676: Migrate the existed calculation logic from operator to PD
## Development Task
Since `tidb-operator` has already implemented some of auto-scaling calculation logic, we can refer it and implement the logic in PD. Details can be found in [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/calculate/calculate.go)

rleungx:

Resolves #2676.

This PR implements the basic framework for auto scaling.

### Check List

Tests

- Will add tests later.

Related changes


pkg/autoscaling/calculation.go
```diff
@@ -0,0 +1,210 @@
...
+	return resources
+}
+
+// TODO: implement calculateScaleOutPlan details
+func calculateScaleOutPlan(rc *cluster.RaftCluster, strategy *Strategy, component ComponentType, scaleOutQuota float64, currentQuota uint64) ([]*Plan, uint64) {
+	groups := getScaledGroupsByComponent(component)
+	group := findBestGroupToScaleOut(rc, strategy, scaleOutQuota, groups, component)
+
+	resCPU := float64(getCPUByResourceType(strategy, group.ResourceType))
+	scaleOutCount := uint64(math.Ceil(scaleOutQuota / resCPU))
```
Yisaer, line 134: I think currently we can simplify the logic as if the auto-scaling instances count changing is always 1 which means If the usage rate > max threshold, we will add one instance.

pkg/autoscaling/calculation.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func calculateScaleInPlan(rc *cluster.RaftCluster, strategy *Strategy, component ComponentType, scaleInQuota float64) []*Plan {
+	groups := getScaledGroupsByComponent(component)
+	if len(groups) == 0 {
+		return nil
+	}
+	group := findBestGroupToScaleIn(rc, strategy, scaleInQuota, groups)
+	resCPU := float64(getCPUByResourceType(strategy, group.ResourceType))
+	scaleInCount := uint64(math.Floor(scaleInQuota / resCPU))
```
Yisaer, line 160: ditto.

pkg/autoscaling/calculation.go
howardlau1999: Is it better to make step configurable?

pkg/autoscaling/calculation.go
```diff
@@ -0,0 +1,210 @@
...
+	totalCPUTime := float64(currentQuota) * MetricsTimeDuration.Seconds()
+	usage := totalCPUUseTime / totalCPUTime
+	maxThreshold, minThreshold := getCPUThresholdByComponent(strategy, component)
+
+	if usage > maxThreshold {
+		scaleOutQuota := totalCPUUseTime/maxThreshold - totalCPUTime
+		return calculateScaleOutPlan(rc, strategy, component, scaleOutQuota, currentQuota)
+	}
+	if usage < minThreshold {
+		scaleInQuota := totalCPUUseTime/minThreshold - totalCPUTime
```
howardlau1999, line 69: The result here will be a negative number

rleungx:
Add a TODO. PTAL @howardlau1999 @Yisaer

howardlau1999:
LGTM

pkg/autoscaling/calculation.go
Yisaer: Making step configurable is ok to me but it will make the logic complex which against our initial goal. We can support it later.

pkg/autoscaling/calculation.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"github.com/pingcap/kvproto/pkg/metapb"
+	"github.com/pingcap/pd/v4/server/cluster"
+)
+
+// MetricsTimeDuration is used to get the metrics of a certain time period.
+// TODO: adjust the value or make it configurable.
+var MetricsTimeDuration = 5 * time.Second
+
+func calculate(rc *cluster.RaftCluster, strategy *Strategy) []*Plan {
```
lhy1024, line 28: where will it be used

pkg/autoscaling/calculation.go
```diff
@@ -0,0 +1,211 @@
...
+		plans = append(plans, tidbPlans...)
+		afterQuota += tidbAfterQuota
+	}
+	if exceedMaxCPUQuota(strategy, afterQuota) {
+		return nil
+	}
+	return plans
+}
+
+func getPlans(rc *cluster.RaftCluster, strategy *Strategy, component ComponentType) ([]*Plan, uint64) {
```
lhy1024, line 45: what is `uint64` in return value? the number of scale-out?

pkg/autoscaling/calculation.go
lhy1024: Making step configurable is ok to me too

pkg/autoscaling/calculation.go
rleungx: It is used in `HTTPHandler`

pkg/autoscaling/calculation.go
rleungx: The total CPU quota of a component if we plan to scale-out.

pkg/autoscaling/calculation.go
HunDunDM: Making step configurable is ok to me too

PR #2739, opened 2020-08-10
api: Support include tombstone option for store.GetAllLimit

Yisaer:
Close https://github.com/pingcap/pd/issues/2714

Add `includeTombstone` to dedicate whether the server need to return the store limit information include tombstone stores.

### Check List

Tests

- Unit test

Related changes


server/api/store.go
```diff
@@ -464,12 +464,35 @@ func (h *storesHandler) SetAllLimit(w http.ResponseWriter, r *http.Request) {
...
 // @Tags store
 // @Summary Get limit of all stores in the cluster.
+// @Param includeTombstone query bool false "include Tombstone" default(true)
 // @Produce json
 // @Success 200 {object} string
 // @Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /stores/limit [get]
 func (h *storesHandler) GetAllLimit(w http.ResponseWriter, r *http.Request) {
 	limits := h.GetScheduleConfig().StoreLimit
+	includeTombstone := true
```
rleungx, line 474: I think the default value should be false?

server/api/store.go
```diff
@@ -464,12 +464,35 @@ func (h *storesHandler) SetAllLimit(w http.ResponseWriter, r *http.Request) {
...
+	var err error
+	if includeStr := r.URL.Query().Get("includeTombstone"); includeStr != "" {
```
rleungx, line 476: How about
```suggestion
	if includeStr := r.URL.Query().Get("include_tombstone"); includeStr != "" {
```

server/api/store.go
lhy1024: I also think false is better

Issue #1775, opened 2019-09-27
Simulator case to import data cannot end normally


lhy1024:
Simulator case to import data cannot end normally now. When I run it, it can not reach balance.

It need to fix and add config.

PR #2741, opened 2020-08-11
pd-simulator: update import-data case

ZenoTan:

Import data case is a case in pd simulator. Originally, it has 40 regions, and it imports many regions afterwards. Then import regions >> previous regions. The real scenario is, import regions << previous regions. So now it imports regions 10% of previous size.

Changed import data case.

### Check List

Tests

- No code


tools/pd-simulator/simulator/cases/import_data.go
```diff
@@ -40,8 +40,8 @@ func newImportData() *Case {
 		})
 	}
 
-	storeIDs := rand.Perm(3)
-	for i := 0; i < 40; i++ {
+	for i := 0; i < 10000; i++ {
```
lhy1024, line 43: can it be configured?

ZenoTan: Fixed.

ZenoTan:
> Will it be helpful to the previous issue?

Which one...

ZenoTan:
PTAL @nolouch

lhy1024:
> > Will it be helpful to the previous issue?
> 
> Which one...

#1775

tools/pd-simulator/simulator/cases/import_data.go
```diff
@@ -120,21 +112,28 @@ func newImportData() *Case {
...
 		}
 		simutil.Logger.Info("import data information",
 			zap.String("table-leader", tableLeaderLog),
 			zap.String("table-peer", tablePeerLog),
 			zap.String("total-leader", totalLeaderLog),
 			zap.String("total-peer", totalPeerLog))
+		checkCount++
+		dev := 0.0
+		for _, p := range regionProps {
+			dev += (p - 10) * (p - 10) / 100
```
lhy1024, line 132: will it still be 10 after region num is configured?

ZenoTan: It means if balanced, store each get 10 % of data, no matter what the total is.

ZenoTan:
> > > Will it be helpful to the previous issue?
> > 
> > 
> > Which one...
> 
> #1775

In fact, after changing to the new scenario, because import data(3 out of 10 peers) makes it less balanced, we don't need to set a stop time, but we should observe the deviation. If schedulers work properly, balance should not be greatly affected by the import.

lhy1024:
> > > > Will it be helpful to the previous issue?
> > > 
> > > 
> > > Which one...
> > 
> > 
> > #1775
> 
> In fact, after changing to the new scenario, because import data(3 out of 10 peers) makes it less balanced, we don't need to set a stop time, but we should observe the deviation. If schedulers work properly, balance should not be greatly affected by the import.

This behavior seems to be different compared to other simulators

tools/pd-simulator/simulator/cases/import_data.go
lhy1024: OK

ZenoTan:
> > > > > Will it be helpful to the previous issue?
> > > > 
> > > > 
> > > > Which one...
> > > 
> > > 
> > > #1775
> > 
> > 
> > In fact, after changing to the new scenario, because import data(3 out of 10 peers) makes it less balanced, we don't need to set a stop time, but we should observe the deviation. If schedulers work properly, balance should not be greatly affected by the import.
> 
> This behavior seems to be different compared to other simulators

Yes. It is because the new scenario is an unbalanced import. Indeed, it is trivial to add a stop condition.

ZenoTan:
PTAL @nolouch @lhy1024

ZenoTan:
> LGTM. we also need to manual test.

It will stop at tick 1009 if region number is 10000.

PR #2751, opened 2020-08-12
tests: fix the vet warning in go1.15

JmPotato:

In go1.15, the vet tool now warns about conversions of the form `string(x)` where `x` has an integer type other than rune or byte. Because of this, pd will not pass tests with go1.15. This pr fix this wrong usage.

More details: https://golang.org/doc/go1.15#vet

`string(id)` -> `fmt.Sprint(id)`

### Check List

Tests

- Unit test

### Release note

Fix a bug that pd will not pass the vet check in go1.15.

### DIFF
--- a/tests/pdctl/member/member_test.go
+++ b/tests/pdctl/member/member_test.go
@@ -16,6 +16,7 @@ package member_test
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


disksing:
Do we need to update go version in CI?

JmPotato:
> Do we need to update go version in CI?

Not necessarily need. It's a bug that caused by incorrect usage of `string()` essentially.

PR #2761, opened 2020-08-13
placement: forbid mutiple leader replicas, and require at least one leader

xhebox:

This PR comes from the discussion between me and @djshow832. As the PD community pointed out that it is invalid to define multiple leaders over the same range. Now it is forbidden.

First, `adjustRule` will filter out rules like `{role=leader, count=3}`. Then, when building the rule list, a loop is used to avoid cases like two rules `{role=leader, count=1}, {role=leader, count=1}` over the same range.

**EDIT**: Suggested by @djshow832 , now it also checks if there is still one leader after add/del. In another word, it is only valid iff `(leader+voter) >= 1 && leader <= 1`.

### Check List

Tests

- Unit test
- Integration test

### Release note

- Forbid adding multiple leaders over the same range.

### DIFF
--- a/server/schedule/placement/rule_list.go
+++ b/server/schedule/placement/rule_list.go
@@ -148,6 +148,8 @@ func buildRuleList(rules map[[2]string]*Rule) (ruleList, error) {
 					strings.ToUpper(hex.EncodeToString(endKey))))
 			}
 
+			rr = prepareRulesForApply(rr) // clone internally
+
 			err := checkApplyRules(rr)
 			if err != nil {
 				return ruleList{}, errs.ErrBuildRuleList.FastGenByArgs(fmt.Sprintf("%s for range {%s, %s}",
@@ -162,7 +164,7 @@ func buildRuleList(rules map[[2]string]*Rule) (ruleList, error) {
 			rl.ranges = append(rl.ranges, rangeRules{
 				startKey:   p.key,
 				rules:      rr,
-				applyRules: prepareRulesForApply(rr), // clone internally
+				applyRules: rr,
 			})
 		}
 	}
--- a/server/schedule/placement/rule_list.go
+++ b/server/schedule/placement/rule_list.go
@@ -64,6 +64,27 @@ func (sr *sortedRules) deleteRule(rule *Rule) {
 	}
 }
 
+func checkRulesRaft(rules []*Rule) string {
+	// check raft constraint
+	// one and only one leader
+	leaderCount := 0
+	voterCount := 0
+	for _, rule := range rules {
+		if rule.Role == Leader {
+			leaderCount += rule.Count
+		} else if rule.Role == Voter {
+			voterCount += rule.Count
+		}
+		if leaderCount > 1 {
+			return "multiple leader replicas"
+		}
+	}
+	if (leaderCount + voterCount) < 1 {
+		return "needs at least one leader or voter"
+	}
+	return ""
+}
+
 type rangeRules struct {
 	startKey []byte
 	// rules indicates all the rules match the given range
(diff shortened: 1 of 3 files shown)


server/schedule/placement/rule_list.go
```diff
@@ -108,17 +108,39 @@ func buildRuleList(rules map[[2]string]*Rule) (ruleList, error) {
...
+					return ruleList{}, errs.ErrBuildRuleList.FastGenByArgs(fmt.Sprintf("multiple leader replicas for range {%s, %s}",
+						strings.ToUpper(hex.EncodeToString(p.key)),
+						strings.ToUpper(hex.EncodeToString(endKey))))
+				}
+			}
+			if (leaderCount + voterCount) < 1 {
+				return ruleList{}, errs.ErrBuildRuleList.FastGenByArgs(fmt.Sprintf("needs at least one leader or voter for range {%s, %s}",
+					strings.ToUpper(hex.EncodeToString(p.key)),
+					strings.ToUpper(hex.EncodeToString(endKey))))
+			}
```
Yisaer, line 143: I suggest moving check out of build function and use an independent function to maintain checking logic.

xhebox: Something like the new commit?

nolouch:
PTAL @disksing @Yisaer

server/schedule/placement/rule_list.go
```diff
@@ -108,17 +129,27 @@ func buildRuleList(rules map[[2]string]*Rule) (ruleList, error) {
 			sr.deleteRule(p.rule)
 		}
 		if i == len(points)-1 || !bytes.Equal(p.key, points[i+1].key) {
+			var endKey []byte
+			if i != len(points)-1 {
+				endKey = points[i+1].key
+			}
+
```
Yisaer, line 136: Why move these code?

server/schedule/placement/rule_list.go
```diff
@@ -63,6 +63,27 @@ func (sr *sortedRules) deleteRule(rule *Rule) {
 	}
 }
 
+func checkRulesRaft(rules []*Rule) string {
```
Yisaer, line 66: Could you add an unit test for this single function?

server/schedule/placement/rule_list.go
xhebox: Just for reusing. It is not a big deal to write it twice, though. Do I need to change?

server/schedule/placement/rule_list.go
xhebox: here it is.

nolouch:
Hi， @xhebox  please sign off the new DCO，thanks :)

server/schedule/placement/rule_list.go
```diff
@@ -64,6 +64,27 @@ func (sr *sortedRules) deleteRule(rule *Rule) {
...
```
disksing, line 67: how about returning error instead?

server/schedule/placement/rule_list.go
```diff
@@ -113,17 +134,27 @@ func buildRuleList(rules map[[2]string]*Rule) (ruleList, error) {
...
-				var endKey []byte
-				if i != len(points)-1 {
-					endKey = points[i+1].key
-				}
 				return ruleList{}, errs.ErrBuildRuleList.FastGenByArgs(fmt.Sprintf("no rule for range {%s, %s}",
 					strings.ToUpper(hex.EncodeToString(p.key)),
 					strings.ToUpper(hex.EncodeToString(endKey))))
 			}
+
+			err := checkRulesRaft(rr)
```
disksing, line 150: I think should check `applyRules` instead

xhebox: how about now?

disksing: em, I meant that you need to check rules after calling `prepareRulesForApply`

https://github.com/tikv/pd/blob/ddb5665c465a1f65d5541b317b2bfb94ab33d97d/server/schedule/placement/rule_list.go#L165)

The reason is that rules can be overridden. It is acceptable if there are 2 rules both specify a leader while one rule is overridden by the other.

xhebox: Got it, done.

nolouch:
ci failed.

nolouch (commented):
LGTM

server/schedule/placement/rule_list.go
disksing: I think it's ok to update like this.

disksing (approved):

ti-srebot (approved):

disksing:
port in use problem. I'll just ignore and merge.

PR #2762, opened 2020-08-13
autoscaling: implement get total CPU use time and quota

howardlau1999:

Implemented the TODOs of `getTotalCPUUseTime` and `getTotalCPUQuota` in #2738 

Moved the `datasource` package to `autoscaling` to prevent cycle imports and implemented the logic of querying the total CPU use time and total CPU quota.

### Check List

Tests

- Unit test

### Release note

- No release note


pkg/autoscaling/types.go
```diff
@@ -111,3 +111,7 @@ type instance struct {
 	id      uint64
 	address string
 }
+
+func (inst instance) String() string {
```
rleungx, line 115: `Address` is better?

pkg/autoscaling/calculation.go
```diff
@@ -103,14 +127,46 @@ func getTiDBInstances() []instance {
 	return []instance{}
 }
 
-// TODO: get total CPU use time through Prometheus.
-func getTotalCPUUseTime(component ComponentType, instances []instance, duration time.Duration) float64 {
-	return 1.0
+func instancesToStrings(instances []instance) []string {
```
rleungx, line 130: How about `getAddresses`?

howardlau1999: We need to convert `instance` to string for querying somehow and using the address for the conversion is a temporary solution

pkg/autoscaling/calculation.go
```diff
@@ -69,9 +74,30 @@ func getPlans(rc *cluster.RaftCluster, strategy *Strategy, component ComponentTy
...
-	totalCPUUseTime := getTotalCPUUseTime(component, instances, MetricsTimeDuration)
-	currentQuota := getTotalCPUQuota(component, instances)
-	totalCPUTime := float64(currentQuota) * MetricsTimeDuration.Seconds()
+	client, err := promClient.NewClient(promClient.Config{
+		Address: cfg.MetricStorage,
+	})
+	if err != nil {
+		log.Error("error initializing Prometheus client", zap.String("address", cfg.MetricStorage), zap.Error(err))
+		return nil, 0
+	}
```
Yisaer, line 83: I think we can only init promClient once.

pkg/autoscaling/calculation.go
```diff
@@ -103,14 +129,73 @@ func getTiDBInstances() []instance {
...
-// TODO: get total CPU quota through Prometheus.
-func getTotalCPUQuota(component ComponentType, instances []instance) uint64 {
-	return 1
+// TODO: suppport other metrics storage
+// get total CPU use time (in seconds) through Prometheus.
+func getTotalCPUUseTime(querier Querier, component ComponentType, instances []instance, timestamp time.Time, duration time.Duration) (float64, error) {
+	result, err := querier.Query(NewQueryOptions(component, CPUUsage, getInstanceNames(instances), timestamp, duration))
+	if err != nil {
+		return 0.0, err
+	}
```
Yisaer, line 173: It seems we only extract the instance name and use it to link the store/tidb and prometheus targets. I think this would have some risk as we can deploy 2 tidbcluster with the same name in different namespace and make them as one united tidbcluster.

I think we also need to extract the namespace information.

pkg/autoscaling/prometheus.go
```diff
@@ -164,3 +176,35 @@ func buildCPUUsagePromQL(options *QueryOptions) (string, error) {
 	query := fmt.Sprintf(pattern, options.duration.String())
 	return query, nil
 }
+
+// this function assumes that addr is already a valid resolvable address
+// returns in format "podname_namespace"
+func getInstanceNameFromAddress(addr string) string {
```
Yisaer, line 182: As the address here is special (made by tidb-cloud's own monitor rule), I suggest check the format here, and if it not match the format we want, just throwing an error instead of throwing the origin address.

howardlau1999: Is the format `pod-name.peer-svc.namespace.svc`?

howardlau1999: updated

pkg/autoscaling/prometheus.go
```diff
@@ -164,3 +179,35 @@ func buildCPUUsagePromQL(options *QueryOptions) (string, error) {
...
+	// Just to make sure it is not an IP address
+	ip := net.ParseIP(hostname)
+	if ip != nil {
+		// Hostname is an IP address, return the whole address
+		return "", errors.Errorf("address %s is an ip address", addr)
+	}
+
+	parts := strings.Split(hostname, ".")
+	if len(parts) < 4 {
+		return "", errors.Errorf("address %s has less than 4 parts", addr)
```
rleungx, line 203: This log is poor readability. I prefer to output the format we expect and the actual result

pkg/autoscaling/calculation.go
```diff
@@ -40,14 +44,35 @@ var (
...
-func calculate(rc *cluster.RaftCluster, strategy *Strategy) []*Plan {
+var (
+	querierInstance Querier
+	once            sync.Once
+)
+
+func getQuerier(cfg *config.PDServerConfig) Querier {
+	once.Do(func() {
+		var err error
+		client, err := promClient.NewClient(promClient.Config{
```
rleungx, line 55: What if we modify the `MetricStorage`?

pkg/autoscaling/prometheus.go
```diff
@@ -11,28 +11,30 @@
...
 	promModel "github.com/prometheus/common/model"
-	types "github.com/tikv/pd/pkg/autoscaling"
 	"go.uber.org/zap"
 )
 
 const (
-	tikvSumCPUUsageMetricsPattern = `sum(increase(tikv_thread_cpu_seconds_total[%s])) by (instance)`
-	tidbSumCPUUsageMetricsPattern = `sum(increase(process_cpu_seconds_total{job="tidb"}[%s])) by (instance)`
+	tikvSumCPUUsageMetricsPattern = `sum(increase(tikv_thread_cpu_seconds_total[%s])) by (instance,kubernetes_namespace)`
+	tidbSumCPUUsageMetricsPattern = `sum(increase(process_cpu_seconds_total{job="tidb"}[%s])) by (instance,kubernetes_namespace)`
```
rleungx, line 33:
```suggestion
	tikvSumCPUUsageMetricsPattern = `sum(increase(tikv_thread_cpu_seconds_total[%s])) by (instance, kubernetes_namespace)`
	tidbSumCPUUsageMetricsPattern = `sum(increase(process_cpu_seconds_total{job="tidb"}[%s])) by (instance, kubernetes_namespace)`
```

pkg/autoscaling/calculation.go
howardlau1999: It seems that we don't have a notify mechanism for config reloading. Maybe it would be simpler to just create a new client for each request?

rleungx: I think so.

rleungx (commented):
LGTM

Yisaer (approved):
LGTM plz solve the conflict

howardlau1999:
@rleungx @Yisaer conflict has been solved

rleungx (approved):

PR #2766, opened 2020-08-13
placement: Add some detail comment and unit test

Yisaer:
Add some comments for some complex function and add unit test for the `buildRuleList` function

### Check List

Tests

- Unit test

### DIFF
--- a/server/schedule/placement/rule.go
+++ b/server/schedule/placement/rule.go
@@ -115,7 +115,7 @@ func sortRules(rules []*Rule) {
 	sort.Slice(rules, func(i, j int) bool { return compareRule(rules[i], rules[j]) < 0 })
 }
 
-// prepareRulesForApply search the target the rules from the given rules.
+// prepareRulesForApply search the target rules from the given rules.
 // it will filter the rules depends on the interval index override in the same group or the
 // group-index override between different groups
 // For example, given rules:
--- a/server/schedule/placement/rule.go
+++ b/server/schedule/placement/rule.go
@@ -115,7 +115,18 @@ func sortRules(rules []*Rule) {
 	sort.Slice(rules, func(i, j int) bool { return compareRule(rules[i], rules[j]) < 0 })
 }
 
-// Sort Rules, trim concealed rules.
+// prepareRulesForApply search the target the rules from the given rules.
+// it will filter the rules depends on the interval index override in the same group or the
+// group-index override between different groups
+// For example, given rules:
+// ruleA: group_id: 4, id: 2, override: true
+// ruleB: group_id: 4, id: 1, override: true
+// ruleC: group_id: 3
+// ruleD: group_id: 2
+// RuleGroupA: id:4, override: false
+// RuleGroupB: id:3, override: true
+// RuleGroupC: id:2, override: false
+// Finally only ruleA and ruleC will be selected.
 func prepareRulesForApply(rules []*Rule) []*Rule {
 	var res []*Rule
 	var i, j int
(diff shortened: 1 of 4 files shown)


server/schedule/placement/rule_manager.go
```diff
@@ -428,13 +431,15 @@ func (m *RuleManager) GetRuleGroups() []*RuleGroup {
 }
 
 // SetRuleGroup updates a RuleGroup.
+// TODO update ruleGroup after set rule/rules
```
disksing, line 434: em, why need to update ruleGroup?

Yisaer: For example, If a new rule with a new groupID is created(set), do we need to update groups in rule_manager?

disksing: If the group does not exist, it will use default configuration so I suppose it is unnecessary to persist group config.

Yisaer: got it.

server/schedule/placement/rule_test.go
```diff
@@ -107,3 +108,84 @@ func (s *testRuleSuite) TestGroupProperties(c *C) {
...
+
+// TODO: fulfill unit test case to cover BuildRuleList
+func (s *testRuleSuite) TestBuildRuleList(c *C) {
+	defaultRule := &Rule{
+		GroupID:  "pd",
+		ID:       "default",
+		Index:    0,
+		Override: false,
+		StartKey: []byte{},
+		EndKey:   []byte{},
```
disksing, line 120: default values can be omitted

server/schedule/placement/rule.go
```diff
@@ -115,7 +115,18 @@ func sortRules(rules []*Rule) {
 	sort.Slice(rules, func(i, j int) bool { return compareRule(rules[i], rules[j]) < 0 })
 }
 
-// Sort Rules, trim concealed rules.
+// prepareRulesForApply search the target the rules from the given rules.
```
JmPotato, line 118:
```suggestion
// prepareRulesForApply search the target rules from the given rules.
```

JmPotato: Is this what you mean?

server/schedule/placement/rule_list.go
```diff
@@ -42,6 +42,7 @@ type sortedRules struct {
 	rules []*Rule
 }
 
+// insertRule insert rule into soretedRules and still keep order
```
JmPotato, line 45:
```suggestion
// insertRule inserts a rule into soretedRules while keeping the order unchanged
```

server/schedule/placement/rule_list.go
```diff
@@ -64,15 +65,19 @@ func (sr *sortedRules) deleteRule(rule *Rule) {
 }
 
 type rangeRules struct {
-	startKey   []byte
-	rules      []*Rule
+	startKey []byte
+	// rules indicates all the rules match the given range
+	rules []*Rule
+	// applyRules indicates the selected rules(filter by prepareRulesForApply) from the given rules
```
JmPotato, line 71:
```suggestion
	// applyRules indicates the selected rules(filtered by prepareRulesForApply) from the given rules
```

server/schedule/placement/rule_list.go
```diff
@@ -64,15 +65,19 @@ func (sr *sortedRules) deleteRule(rule *Rule) {
...
 	applyRules []*Rule
 }
 
 type ruleList struct {
-	ranges []rangeRules // ranges[i] contains rules apply to (ranges[i].startKey, ranges[i+1].startKey).
+	ranges []rangeRules // ranges[i] contains rules apply to [ranges[i].startKey, ranges[i+1].startKey).
 }
 
+// buildRuleList build the applied ruleList for the give rules
```
JmPotato, line 79:
```suggestion
// buildRuleList builds the applied ruleList for the give rules
```

PR #2771, opened 2020-08-15
*: replace some sync.(RW)Mutex with atomic.Value

JmPotato:

In some simple Read/Write cases, using `atomic.Value` is considered more efficient and simpler than `sync.(RW)Mutex`. To be more specifically, here is an example:

```go
type A struct {
    mu sync.Mutex
    field int
}
```

If the methods and usage of struct `A` only involve reading and writing `field` concurrently. Replacing it with `atomic.Value` can benefit from:

* A CPU-level atomic operation
* Scalable CPU performance(More CPUs, more efficiency we can get)
* Less runtime resource usage(`sync.Mutex/RWMutex` is more complex)

Therefore, I try to replace some of `sync.(RW)Mutex` fit this case with `atomic.Value` to get a theoretical better performance.

Replace some `sync.(RW)Mutex` with `atomic.Value`.

### Check List

Tests

- Unit test
- Integration test

### Release note

Replace some `sync.(RW)Mutex` with `atomic.Value` to get a better performance.


Yisaer:
> Therefore, I try to replace some of sync.(RW)Mutex fit this case with atomic.Value to get a theoretical better performance.

What about adding a bench test?

JmPotato:
> > Therefore, I try to replace some of sync.(RW)Mutex fit this case with atomic.Value to get a theoretical better performance.
> 
> What about adding a bench test?

Here is a simple [benchmark](https://github.com/talkgo/night/issues/364#issuecomment-487283558) to compare the performance among `Mutex`, `atomic` and `chan`. You can also check this [discussion](https://stackoverflow.com/questions/47445344/is-there-a-difference-in-go-between-a-counter-using-atomic-operations-and-one-us) to get more information.

disksing:
IMO it is bad practice to use this pattern in non-strictly performance related places. Because `atomic.Value` drops type information and makes it harder to read and easier to make mistakes.

Yisaer:
> > > Therefore, I try to replace some of sync.(RW)Mutex fit this case with atomic.Value to get a theoretical better performance.
> > 
> > 
> > What about adding a bench test?
> 
> Here is a simple [benchmark](https://github.com/talkgo/night/issues/364#issuecomment-487283558) to compare the performance among `Mutex`, `atomic` and `chan`. You can also check this [discussion](https://stackoverflow.com/questions/47445344/is-there-a-difference-in-go-between-a-counter-using-atomic-operations-and-one-us) to get more information.

It's ok to me to replace mutex by atomic, As your request wants to improve performance in this case, I think you need to add the go bench test to prove that in your case.

Issue #2783, opened 2020-08-18
RegionInfo.Clone and FitRegion consume too much cpu in ruleLeaderFitFilter


Yisaer:
### What did you do?
The user previously used pd v4.0.4 and found that heartbeat region event qps become lower. After downgrading to v4.0.0, the qps become normal. The cpu profile have been captured for both running pd with v4.0.4 and v4.0.0 version. We found that ruleLeaderFitFilter consume too much cpu resources in v4.0.4 comparing v4.0.0.

The placementRule is enabled and only have default rule. 

### What did you see instead?

RegionInfo.Clone and FitRegion consume too much cpu in ruleLeaderFitFilter.

![image](https://user-images.githubusercontent.com/13427348/90470059-0b81aa80-e14d-11ea-83c4-3bb5e936213c.png)

Here is the cpu profile for v4.0.4

[profile(1).gz](https://github.com/tikv/pd/files/5088196/profile.1.gz)

Here is the cpu profile for v4.0.0

[profile-400.gz](https://github.com/tikv/pd/files/5088200/profile-400.gz)

### What version of PD are you using (`pd-server -V`)?
v4.0.4

nolouch:
![image](https://user-images.githubusercontent.com/6428910/90496555-8f9d5780-e178-11ea-9a9f-395b47862618.png)
It always has many `no-target-stores` events in 4.0.4 version.

PR #2787, opened 2020-08-18
scheduler: limit retry limit for balance-*-scheduler when placement rule is enabled.

Yisaer:

ref https://github.com/tikv/pd/issues/2783

When placement rule is enabled, FitRegion will be called during balance-*-scheduler and the numbers of calling `FitRegion` function would be `maxScheduleRetries` * `store count` * `retry limit` * const number(about 3). Then FitRegion would be consumed too much cpu resources which affects overall performance.

Add new property `fitRegionCallingCount` in scheduler, when placement rule is enabled, the fitRegionCallingCount is limited by  the fitRegionCallingLimit.

### Check List

Tests

- Unit test

Related changes

### DIFF
--- a/server/schedulers/balance_leader.go
+++ b/server/schedulers/balance_leader.go
@@ -35,6 +35,10 @@ const (
 	BalanceLeaderType = "balance-leader"
 	// balanceLeaderRetryLimit is the limit to retry schedule for selected source store and target store.
 	balanceLeaderRetryLimit = 10
+	// fitRegionCalling is the limit of the count of calling FitRegion in one scheduler Schedule
+	// when placement rule is enabled.
+	// TODO: set a proper value
+	fitRegionCallingLimit = 10
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


server/schedulers/balance_leader.go
```diff
@@ -35,6 +35,10 @@ const (
 	BalanceLeaderType = "balance-leader"
 	// balanceLeaderRetryLimit is the limit to retry schedule for selected source store and target store.
 	balanceLeaderRetryLimit = 10
+	// fitRegionCalling is the limit of the count of calling FitRegion in one scheduler Schedule
+	// when placement rule is enabled.
+	// TODO: set a proper value
+	fitRegionCallingLimit = 10
```
nolouch, line 41: The value may cause us cannot find a suitable store.

disksing: right.

lhy1024: Should it be proportional to the number of stores

lhy1024:
why close it?

PR #2794, opened 2020-08-18
filter: replace region.Clone in filter

Yisaer:

ref https://github.com/tikv/pd/issues/2783

region.Clone consumes too much cpu resources.

replace region.Clone with a self-maintained clone function as FitRegion only need partial information of a region.

BenchmarkCloneRegionTest is using new clone method and BenchmarkCloneRegionTest2 is using region.Clone
```sh
$ go test -bench=. -run=none
goos: darwin
goarch: amd64
pkg: github.com/tikv/pd/server/schedule/filter
BenchmarkCloneRegionTest-4       4356267               267 ns/op
BenchmarkCloneRegionTest2-4       316604              3292 ns/op
PASS
ok      github.com/tikv/pd/server/schedule/filter       2.555s
```

### Check List

Tests

- Unit test

Related changes

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -426,7 +426,9 @@ func (f *ruleFitFilter) Source(opt opt.Options, store *core.StoreInfo) bool {
 }
 
 func (f *ruleFitFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
-	region := cloneRegion(f.region, core.WithReplacePeerStore(f.oldStore, store.GetID()))
+	region := createRegionForRuleFit(f.region.GetStartKey(), f.region.GetEndKey(),
+		f.region.GetPeers(), f.region.GetLeader(),
+		core.WithReplacePeerStore(f.oldStore, store.GetID()))
 	newFit := f.fitter.FitRegion(region)
 	return placement.CompareRegionFit(f.oldFit, newFit) <= 0
 }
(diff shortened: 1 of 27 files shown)


server/schedule/filter/filters.go
```diff
@@ -672,3 +673,28 @@ func (f *isolationFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
 	}
 	return true
 }
+
+// cloneRegion is used to replace region.Clone in filter which call the FitRegion,
+// as FitRegion only needs partial information(startKey, endKey, peers, leader peer)
+func cloneRegion(region *core.RegionInfo, opts ...core.RegionCreateOption) *core.RegionInfo {
```
nolouch, line 679: Do not clone the `downPeers`, `pendingPeers`?  maybe rename to a suitable name?

disksing: How about `createRegionForRuleFit(peers []*metapb.Peer, leader *metapb.Peer) *core.RegionInfo`?

Yisaer: @nolouch  currently FitRegion didn't consider the downPeers and pending peers, after discussed with @disksing , I create a tracking issue https://github.com/tikv/pd/issues/2796

@disksing  good advice, updated.

server/schedule/filter/filters.go
```diff
@@ -672,3 +677,29 @@ func (f *isolationFilter) Target(opt opt.Options, store *core.StoreInfo) bool {
...
+// createRegionForRuleFit is used to create a clone region with RegionCreateOptions which is only used for
+// FitRegion in filter
+func createRegionForRuleFit(startKey, endKey []byte,
+	peers []*metapb.Peer, leader *metapb.Peer, opts ...core.RegionCreateOption) *core.RegionInfo {
+	copyLeader := &metapb.Peer{
+		Id:      leader.Id,
+		StoreId: leader.StoreId,
+		Role:    leader.Role,
+	}
+	copyPeers := make([]*metapb.Peer, 0, len(peers))
```
disksing, line 690: Seems we don't have to copy leader. And if the peer list is untouched, we don't need to copy it either.

Yisaer: I think for both `ruleFitFilter` and `ruleLeaderFitFilter`, they will both touch peers and leaders. I think it is necessary to copy peers and leaders either.

disksing: It makes sense.

disksing (approved):

nolouch (approved):
LGTM. 
BTW, How about the benchmark result?

ti-srebot (approved):

Yisaer:
> LGTM.
> BTW, How about the benchmark result?

I updated the benchmark result in the description.

lhy1024 (approved):

PR #2810, opened 2020-08-20
autoscaling: Implement TidbInformer

Yisaer:
Implement TidbInformer

### Check List

Tests

- Unit test

Related changes


Yisaer:
@howardlau1999 PTAL

pkg/autoscaling/calculation.go
```diff
@@ -41,6 +45,12 @@ var (
 	MaxScaleInStep uint64 = 1
 )
 
+func setInformer(rc *cluster.RaftCluster) {
+	if informer == nil {
+		informer = newTidbInformer(rc.GetEtcdClient())
+	}
```
howardlau1999, line 51: when multiple requests arrive at the same time, is it acceptable to create multiple informer instances here?

howardlau1999:
The rest LGTM

pkg/autoscaling/calculation.go
Yisaer: good catch, updated.

pkg/autoscaling/calculation.go
```diff
@@ -288,7 +288,11 @@ func getScaledTiKVGroups(informer core.StoreSetInformer, healthyInstances []inst
 func getScaledTiDBGroups(informer tidbInformer, healthyInstances []instance) []*Plan {
 	planMap := make(map[string]map[string]struct{}, len(healthyInstances))
 	for _, instance := range healthyInstances {
-		tidb := informer.GetTiDB(instance.address)
+		tidb, err := informer.GetTiDB(instance.address)
+		if err != nil {
+			// TODO: error handling
```
rleungx, line 293: What should we do to handle this kind of error?

pkg/autoscaling/types.go
```diff
@@ -138,3 +148,38 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+}
+
+func (informer *tidbInformerImpl) GetTiDB(address string) (*TiDBInfo, error) {
+	key := fmt.Sprintf("/topology/tidb/%s/info", address)
+	resp, err := etcdutil.EtcdKVGet(informer.etcdClient, key)
+	if err != nil {
+		return nil, err
+	}
+	if resp.Count < 1 {
+		err := fmt.Errorf("resp loaded for tidb[%s] is empty", address)
```
rleungx, line 169: There is no need to define a new error. We can directly output what we wanna show in the log.

pkg/autoscaling/calculation.go
Yisaer: I think we could package the error message and tell the caller(deploy layer) the error reason in response.

pkg/autoscaling/types.go
Yisaer: updated.

pkg/autoscaling/calculation.go
```diff
@@ -117,9 +125,21 @@ func filterTiKVInstances(informer core.StoreSetInformer) []instance {
 	return instances
 }
 
-// TODO: get TiDB instances, we can directly visit prometheus 'up{job="tidb"}' metrics to know the healthy tidb instances
-func getTiDBInstances() []instance {
-	return []instance{}
+func getTiDBInstances(rc *cluster.RaftCluster) []instance {
+	informerInit.Do(func() {
```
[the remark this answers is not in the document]
Yisaer: removed.

pkg/autoscaling/types.go
```diff
@@ -134,7 +144,66 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+func (informer *tidbInformerImpl) GetTiDBs() ([]*TiDBInfo, error) {
+	prefix := "/topology/tidb/"
+	resps, err := etcdutil.EtcdKVGet(informer.etcdClient, prefix, clientv3.WithPrefix())
+	if err != nil {
+		return nil, err
+	}
+	tidbs := make([]*TiDBInfo, 0, resps.Count)
+	for _, resp := range resps.Kvs {
+		key := string(resp.Key)
+		if tidbInfoPattern.MatchString(key) {
```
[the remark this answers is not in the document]
Yisaer: Using `regexp.MatchString` would be needed to handle error in every loop. I tend to init `tidbInfoPattern` at first.

pkg/autoscaling/types.go
```diff
@@ -138,3 +140,46 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+
+// GetTiDB get TiDB info which registered in PD by address
+func GetTiDB(etcdClient *clientv3.Client, address string) (*TiDBInfo, error) {
+	key := fmt.Sprintf("/topology/tidb/%s/info", address)
+	resp, err := etcdutil.EtcdKVGet(etcdClient, key)
+	if err != nil {
+		return nil, err
+	}
+	if resp.Count < 1 {
+		err := fmt.Errorf("resp loaded for tidb[%s] is empty", address)
```
rleungx, line 152:
```suggestion
		err := fmt.Errorf("resp loaded for tidb [%s] is empty", address)
```

pkg/autoscaling/types.go
```diff
@@ -138,3 +140,46 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+	if err != nil {
+		return nil, err
+	}
+	tidb.Address = address
+	return tidb, nil
+}
+
+// GetTiDBs list TiDB register in PD
+func GetTiDBs(etcdClient *clientv3.Client) ([]*TiDBInfo, error) {
+	tidbInfoPattern, err := regexp.Compile("/topology/tidb/.+/info")
```
rleungx, line 166: prefer to use a const for "/topology/tidb/.+/info"

Yisaer: updated.

rleungx (approved):
LGTM

pkg/autoscaling/types.go
```diff
@@ -138,3 +140,54 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+	prefix := "/topology/tidb/"
+	resps, err := etcdutil.EtcdKVGet(etcdClient, prefix, clientv3.WithPrefix())
+	if err != nil {
+		return nil, err
+	}
+	tidbs := make([]*TiDBInfo, 0, resps.Count)
+	for _, resp := range resps.Kvs {
+		key := string(resp.Key)
+		if tidbInfoPattern.MatchString(key) {
+			address := key[len("/topology/tidb/") : len(key)-5]
```
howardlau1999, line 183:
```suggestion
			address := key[len(prefix) : len(key)-len("/info")]
```

howardlau1999:
The rest LGTM

pkg/autoscaling/types.go
```diff
@@ -138,3 +140,54 @@ func (t *TiDBInfo) getLabelValue(key string) string {
...
+	tidbInfoPatternStr = "/topology/tidb/.+/info"
+)
+
+// GetTiDBs list TiDB register in PD
+func GetTiDBs(etcdClient *clientv3.Client) ([]*TiDBInfo, error) {
+	tidbInfoPattern, err := regexp.Compile(tidbInfoPatternStr)
+	if err != nil {
+		return nil, err
+	}
+	prefix := "/topology/tidb/"
```
howardlau1999, line 174: How about making it a constant

pkg/autoscaling/types.go
howardlau1999: Or you can add a capture group in the regex to directly extract the address

pkg/autoscaling/types.go
Yisaer: updated.

howardlau1999:
LGTM

Issue #2808, opened 2020-08-20
Support to get region info from PD followers


JmPotato:
### Describe your feature request related problem

For now, PD client only can get region info from PD leader as the following codes show in [client/client.go](https://github.com/tikv/pd/blob/master/client/client.go#L477).

```go
resp, err := c.leaderClient().GetRegion(ctx, &pdpb.GetRegionRequest{
	Header:    c.requestHeader(),
	RegionKey: key,
})
```

But if the PD leader is busy, in bad network or even down, the region info getting will become slow. The corresponding implementation of RPC server is in here [pd/grpc_service.go](https://github.com/tikv/pd/blob/master/server/grpc_service.go#L417).

### Describe the feature you'd like

To improve PD cluster's availability, we can support to get region info from PD followers. The region info has been synchronized between leaders and followers with etcd or region_syncer through current implementation. All you have to do is update the PD client and allow it can get region info from any of PD servers.

### Describe alternatives you've considered

To make this feature more meaningful, maybe you can make the target PD server change dynamically according to response time.

### Teachability, Documentation, Adoption, Migration Strategy

* Go
* (g)RPC knowledge

PR #2812, opened 2020-08-21
client: support to get region info from PD followers

RinChanNOWWW:

https://github.com/tikv/pd/issues/2808
Support to get region info from PD followers.

The client now gets region info from the server which responds earliest.

### Check List

Tests

- Unit test

Side effects

- Generate more goroutines

### Release note

* Support to get region info from PD followers.


rleungx:
What if the follower doesn't have the latest Region information?

client/client.go
```diff
@@ -465,6 +465,54 @@ func (c *client) parseRegionResponse(res *pdpb.GetRegionResponse) *Region {
 	return r
 }
 
+func (c *client) waitForRegion(ctx context.Context, key []byte) (*pdpb.GetRegionResponse, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
JmPotato, line 469: The old `ctx` should also be used to detect the `<-ctx.Done()`, otherwise the upper timeout will be meaningless.

client/client.go
```diff
@@ -465,6 +465,54 @@ func (c *client) parseRegionResponse(res *pdpb.GetRegionResponse) *Region {
...
+
+	ch := make(chan *pdpb.GetRegionResponse, len(c.connMu.clientConns))
+
```
JmPotato, line 472: Maybe we can use `defer close(ch)` here rather than just close channel when received a non-nil `resp`?

client/client.go
```diff
@@ -465,6 +465,54 @@ func (c *client) parseRegionResponse(res *pdpb.GetRegionResponse) *Region {
...
+	return resp, nil
+}
+
+func (c *client) getRegionResponse(ctx context.Context, cc pdpb.PDClient, key []byte, ch chan<- *pdpb.GetRegionResponse) {
+	resp, err := cc.GetRegion(ctx, &pdpb.GetRegionRequest{
+		Header:    c.requestHeader(),
+		RegionKey: key,
+	})
+	if err != nil {
+		log.Warn("get region error")
```
JmPotato, line 505: We need more log info to see what the `err` is and where the `resp` from.

client/client.go
```diff
@@ -465,6 +465,54 @@ func (c *client) parseRegionResponse(res *pdpb.GetRegionResponse) *Region {
...
+	}
+
+	select {
+	case <-ctx.Done():
+	case ch <- resp:
+		return
+	}
+}
+
 func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
```
nolouch, line 516: I think it is better to provide an independent interface, don't change the behavior of the current interface.

RinChanNOWWW: Understood

client/client.go
zz-jason: Seems this comment has not been addressed. PTAL again.

client/client.go
zz-jason: please use a different variable name for the new `ctx` to improve the code readability.

client/client.go
```diff
@@ -54,6 +54,8 @@ type Client interface {
 	// Also it may return nil if PD finds no Region for the key temporarily,
 	// client should retry later.
 	GetRegion(ctx context.Context, key []byte) (*Region, error)
+	// GetRegionRandomly gets a region from the server which responds earliest
+	GetRegionRandomly(ctx context.Context, key []byte) (*Region, error)
```
nolouch, line 58: How about `GetRegionFromMember`,  we only need GetRegion from the specified member or specified members.

RinChanNOWWW: > How about `GetRegionFromMember`, we only need GetRegion from the specified member or specified members.

But this method is to get region from the member which responds first, could we name it `GetRegionFromMember` ?

nolouch: Yes, I mean the function only get the region from the specified PD, rather than all PDs, and pick the first response.

RinChanNOWWW: OK

client/client.go
```diff
@@ -488,6 +490,73 @@ func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
...
 }
 
+func (c *client) waitForRegion(ctx context.Context, key []byte) (*pdpb.GetRegionResponse, error) {
+	childContext, cancel := context.WithCancel(ctx)
+
+	ch := make(chan *pdpb.GetRegionResponse, len(c.connMu.clientConns))
+
+	c.connMu.RLock()
+	for _, conn := range c.connMu.clientConns {
+		go c.getRegionResponse(childContext, pdpb.NewPDClient(conn), key, ch)
```
nolouch, line 500: we can try it one by one, do not request all PDs, which will make the request multiply increase.

client/client.go
```diff
@@ -488,6 +490,73 @@ func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
 	return c.parseRegionResponse(resp), nil
 }
 
+func (c *client) waitForRegion(ctx context.Context, key []byte) (*pdpb.GetRegionResponse, error) {
+	childContext, cancel := context.WithCancel(ctx)
+
+	ch := make(chan *pdpb.GetRegionResponse, len(c.connMu.clientConns))
+
+	c.connMu.RLock()
+	for _, conn := range c.connMu.clientConns {
```
nolouch, line 499: if the leader does not change, the connection with another member will not exist.

nolouch (commented):
With this [PR](https://github.com/pingcap/tidb/pull/19212), we can label the TiDB. that's means, each TiDB can request the region from the PD with the same label, and the same label meaning TiDB and PD deploy in the same data center,  which can reduce a cross-regional network request in cross-region [deployed cluster](https://docs.pingcap.com/zh/tidb/stable/three-data-centers-in-two-cities-deployment).  so, `GetRegionFromMember` should accept arguments with a slice PD member URL. such as :
```
GetRegionFromMember(ctx context.Context, key []byte, membersURL []string) (*Region, error)
```

Yisaer:
Is this request under working now? @RinChanNOWWW

RinChanNOWWW:
> Is this request under working now? @RinChanNOWWW

yes

client/client.go
```diff
@@ -54,6 +54,8 @@ type Client interface {
 	// Also it may return nil if PD finds no Region for the key temporarily,
 	// client should retry later.
 	GetRegion(ctx context.Context, key []byte) (*Region, error)
+	// GetRegionFromMember gets a region from certain members
```
JmPotato, line 57:
```suggestion
	// GetRegionFromMember gets a region from certain members.
```

client/client.go
```diff
@@ -488,6 +490,73 @@ func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
 	return c.parseRegionResponse(resp), nil
 }
 
+func (c *client) waitForRegion(ctx context.Context, key []byte, memberURLs []string) (*pdpb.GetRegionResponse, error) {
+	childContext, cancel := context.WithCancel(ctx)
+
```
JmPotato, line 495: Could you please trim these unnecessary empty spaces?

client/client.go
```diff
@@ -488,6 +490,73 @@ func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
...
+	ch := make(chan *pdpb.GetRegionResponse, len(memberURLs))
+
+	for _, memberURL := range memberURLs {
+		go c.getRegionResponse(childContext, memberURL, key, ch)
```
JmPotato, line 499: As @nolouch said, you could try to get region from these members one by one and return the first valid response. Doing this concurrently will cause read amplification.

nolouch (commented):
LGTM

client/client.go
```diff
@@ -524,6 +527,44 @@ func (c *client) GetRegion(ctx context.Context, key []byte) (*Region, error) {
...
+func (c *client) GetRegionFromMember(ctx context.Context, key []byte, memberURLs []string) (*Region, error) {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span = opentracing.StartSpan("pdclient.GetRegionFromMember", opentracing.ChildOf(span.Context()))
+		defer span.Finish()
+	}
+	start := time.Now()
+	defer func() { cmdDurationGetRegion.Observe(time.Since(start).Seconds()) }()
+
+	var resp *pdpb.GetRegionResponse
+	for _, url := range memberURLs {
```
rleungx, line 539: The time of getting regions is decided by the sequence of the member list?

rleungx:
friendly ping~

RinChanNOWWW:
> friendly ping~

I didn't recognize that my comment was pending

client/client.go
rleungx: Is it going to replace GetRegion? @JmPotato

JmPotato: No. We plan to use it with the location labels of TiDB after to provide lower latency, but for now, this is just a new method added.

rleungx: if it is decided by the sequence of the member list, it's may have some latency when the front of this list can not be  reached

RinChanNOWWW: Yes. But the member list is determined by the user.

rleungx (approved):
Consider this API is not used to replace the original GetRegion API, LGTM.

ti-srebot (approved):

PR #2816, opened 2020-08-21
placement: refactor rule config management

disksing:

To update rule configurations, formally we implemented an `update, tryCommit, recover` process. During the update phase, we record all old rules/groups that are replaced, and at the end, if any error occurs, we need to recover. This is hard to maintain.

Introduce `ruleConfigPatch` to cache pending updates and only merge to config after all checks passed.

### Check List

Tests
- Unit test

### Release note
- No release note


nolouch:
Would you like to review this PR? @xhebox

xhebox (commented):
https://github.com/tikv/pd/blob/67b93772040603f62a7e7b92840fc556b2ca92c9/server/schedule/placement/rule_manager.go#L188, would it better to use `beginPatch` and `tryCommitPatch` instead of `Batch`?

It is switched to `Batch` for consistency. But now it seems that it is pretty easy to replace `Batch` with `ruleConfigPatch`. `Batch` now is just another simple wrapper around `ruleConfigPatch`.

disksing:
@xhebox good idea!

xhebox:
LGTM

server/schedule/placement/config.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+// ruleConfig contains rule and rule group configurations.
+type ruleConfig struct {
+	rules  map[[2]string]*Rule
+	groups map[string]*RuleGroup
```
Yisaer, line 24: Please add the comment to explaining the meaning of the key, like:
// rules {groupID,ruleID} -> Rule

server/schedule/placement/config.go
```diff
@@ -0,0 +1,160 @@
...
+
+package placement
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+// ruleConfig contains rule and rule group configurations.
+type ruleConfig struct {
```
Yisaer, line 22: How about adding unit test for ruleConfig methods or adding a TODO to track it ?

Yisaer (commented):
Rest LGTM

server/schedule/placement/config.go
disksing: Good idea.

server/schedule/placement/config.go
disksing: I thought about it but after refactoring, I found out the test coverage is pretty good. Most branches are already covered by old tests. So I guess it's ok not to add new tests.

server/schedule/placement/config.go
```diff
@@ -0,0 +1,160 @@
...
+func (c *ruleConfig) beginPatch() *ruleConfigPatch {
+	return &ruleConfigPatch{
+		c:   c,
+		mut: newRuleConfig(),
+	}
+}
+
+// A helper data structure to update ruleConfig.
+type ruleConfigPatch struct {
+	c   *ruleConfig
```
Yisaer, line 78: Does `c` mean the existed rules before committing the patching? Please add a comment to explain what `c` works for?

server/schedule/placement/rule_list.go
```diff
@@ -98,15 +98,16 @@ type ruleList struct {
 	ranges []rangeRules // ranges[i] contains rules apply to [ranges[i].startKey, ranges[i+1].startKey).
 }
 
+type ruleContainer interface {
+	iterateRules(func(*Rule))
+}
```
Yisaer, line 103: what about `rulesIterator`.

Yisaer (commented):
Rest LGTM

server/schedule/placement/config.go
disksing: ok

server/schedule/placement/rule_list.go
disksing: em, I think `rulesIterator` usually has a `Next()` functions and returns an element every time `Next` is called. But here it works differently.

Yisaer (approved):
LGTM

server/schedule/placement/rule_list.go
Yisaer: got it.

lhy1024 (approved):

ti-srebot (approved):

PR #2822, opened 2020-08-23
bug fix: calculate limit rate bases on minute when mode is auto

1150310621:
Signed-off-by: 1150310621 <2604947873@qq.com>

when store limit mode is auto, tidb calculate limit rate on seconds and add it to schedule config mistakely.

return the calculated rate without divide by the store balance base time

### Release note

- Fix the unit of store limit rate when the mode is auto


rleungx (approved):
LGTM

nolouch (approved):
LGTM

ti-srebot (approved):

PR #2831, opened 2020-08-25
election, member: adjust some election methods

JmPotato:

Adjust some election methods.

* Make `member.Leadership` non-public.
* Make `WatchLeader` a more general method because the TSO allocator will also use the same logic to watch the allocator changing.
* Adjust some functions' signatures to make them more sense in using.
* Add and update some comments.

### Check List

Tests

- Unit test
- Integration test

### Release note

* Adjust some election methods.


server/member/member.go
```diff
@@ -115,21 +112,53 @@ func (m *Member) GetLeader() *pdpb.Member {
...
+func (m *Member) KeepLeadership(ctx context.Context) {
+	m.leadership.Keep(ctx)
+}
+
+// CheckLeadership is used to check whether a leadership of
+// PD member is still avalibale.
+func (m *Member) CheckLeadership() bool {
+	return m.leadership.Check()
+}
+
```
Yisaer, line 161: I think we can hide the `LeaderShip` in these methods here.

JmPotato: Done.

Yisaer (approved):
LGTM

PR #2840, opened 2020-08-26
autoscaling: get resource type from instance label

howardlau1999:
Signed-off-by: Howard Lau <howardlau1999@hotmail.com>

PD does not have resource type information for each group when making auto-scaling decisions.

We assume that every instance added by auto-scaling has a label `resource-type` and we associate the group name with `resource-type` to provide the missing information.

### Check List

Tests

- Unit test

### Release note

- No release note

### DIFF
--- a/pkg/autoscaling/calculation.go
+++ b/pkg/autoscaling/calculation.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	promClient "github.com/prometheus/client_golang/api"
@@ -204,7 +205,11 @@ func getResourcesByComponent(strategy *Strategy, component ComponentType) []*Res
 }
 
 func calculateScaleOutPlan(rc *cluster.RaftCluster, strategy *Strategy, component ComponentType, scaleOutQuota float64, currentQuota uint64, instances []instance) []*Plan {
-	groups := getScaledGroupsByComponent(rc, component, instances)
+	groups, err := getScaledGroupsByComponent(rc, component, instances)
+	if err != nil {
+		// TODO: error handling
+		return nil
+	}
 	group := findBestGroupToScaleOut(rc, strategy, scaleOutQuota, groups, component)
 
 	resCPU := float64(getCPUByResourceType(strategy, group.ResourceType))
@@ -231,7 +236,11 @@ func calculateScaleOutPlan(rc *cluster.RaftCluster, strategy *Strategy, componen
 }
 
 func calculateScaleInPlan(rc *cluster.RaftCluster, strategy *Strategy, component ComponentType, scaleInQuota float64, instances []instance) []*Plan {
-	groups := getScaledGroupsByComponent(rc, component, instances)
+	groups, err := getScaledGroupsByComponent(rc, component, instances)
+	if err != nil {
+		// TODO: error handling
+		return nil
+	}
 	if len(groups) == 0 {
 		return nil
 	}
@@ -269,19 +278,19 @@ func getCountByResourceType(strategy *Strategy, resourceType string) uint64 {
 	return 0
 }
 
-func getScaledGroupsByComponent(rc *cluster.RaftCluster, component ComponentType, healthyInstances []instance) []*Plan {
+func getScaledGroupsByComponent(rc *cluster.RaftCluster, component ComponentType, healthyInstances []instance) ([]*Plan, error) {
 	switch component {
 	case TiKV:
 		return getScaledTiKVGroups(rc, healthyInstances)
 	case TiDB:
 		return getScaledTiDBGroups(rc.GetEtcdClient(), healthyInstances)
 	default:
-		return nil
+		return nil, errors.Errorf("unknown component type %s", component.String())
 	}
 }
 
 // TODO: error handling while processing labels
-func getScaledTiKVGroups(informer core.StoreSetInformer, healthyInstances []instance) []*Plan {
+func getScaledTiKVGroups(informer core.StoreSetInformer, healthyInstances []instance) ([]*Plan, error) {
 	planMap := make(map[string]map[string]struct{}, len(healthyInstances))
 	resourceTypeMap := make(map[string]string)
 	for _, instance := range healthyInstances {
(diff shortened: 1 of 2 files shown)


howardlau1999:
@Yisaer PTAL

pkg/autoscaling/calculation.go
```diff
@@ -307,35 +324,50 @@ func getScaledTiDBGroups(etcdClient *clientv3.Client, healthyInstances []instanc
...
+			Component:    componentType.String(),
+			Count:        uint64(len(groupInstances)),
+			ResourceType: resourceTypeMap[groupName],
 			Labels: []*metapb.StoreLabel{
 				{
 					Key:   groupLabelKey,
-					Value: groupLabel,
+					Value: groupName,
 				},
 			},
```
Yisaer, line 372: I think we can directly place the ResourceType into Labels as we need to ensure the deploying layer should place the labels we want.

pkg/autoscaling/calculation.go
```diff
@@ -281,21 +282,37 @@ func getScaledGroupsByComponent(rc *cluster.RaftCluster, component ComponentType
...
+		if !isAutoScaledGroup(groupName) {
+			continue
+		}
+
+		buildPlanMap(planMap, groupName, instance.address)
+
+		if _, ok := resourceTypeMap[groupName]; !ok {
+			resourceType := store.GetLabelValue(resourceTypeLabelKey)
+			if resourceType == "" {
+				log.Warn("store is in auto-scaled group but has no resource type label", zap.Uint64("store-id", instance.id))
```
Yisaer, line 304: If the precondition is not meet (as we didn't find resourceType label here), I think we should throw an error here and directly quit the calculation.

pkg/autoscaling/calculation.go
```diff
@@ -307,35 +324,50 @@ func getScaledTiDBGroups(etcdClient *clientv3.Client, healthyInstances []instanc
...
+		groupName := tidb.getLabelValue(groupLabelKey)
+		if !isAutoScaledGroup(groupName) {
+			continue
+		}
+
+		buildPlanMap(planMap, groupName, instance.address)
+		resourceType := tidb.getLabelValue(resourceTypeLabelKey)
+		if _, ok := resourceTypeMap[groupName]; !ok {
+			if resourceType == "" {
+				log.Warn("tidb is in auto-scaled group but has no resource type label", zap.String("tidb-address", instance.address))
```
Yisaer, line 337: ditto.

pkg/autoscaling/calculation.go
```diff
@@ -281,21 +282,37 @@ func getScaledGroupsByComponent(rc *cluster.RaftCluster, component ComponentType
 
 func getScaledTiKVGroups(informer core.StoreSetInformer, healthyInstances []instance) []*Plan {
 	planMap := make(map[string]map[string]struct{}, len(healthyInstances))
+	resourceTypeMap := make(map[string]string)
 	for _, instance := range healthyInstances {
 		store := informer.GetStore(instance.id)
 		if store == nil {
 			log.Warn("inconsistency between health instances and store status, exit auto-scaling calculation",
 				zap.Uint64("store-id", instance.id))
 			return nil
```
Yisaer, line 291: We should throw the error here.

pkg/autoscaling/calculation.go
```diff
@@ -281,21 +282,37 @@ func getScaledGroupsByComponent(rc *cluster.RaftCluster, component ComponentType
 
 func getScaledTiKVGroups(informer core.StoreSetInformer, healthyInstances []instance) []*Plan {
```
Yisaer, line 283: We can add a TODO here to track the error handling.

Yisaer (commented):
Rest LGTM

pkg/autoscaling/calculation.go
howardlau1999: throw what kind of error?

Yisaer: I think the built-in error is enough currently. `getScaledTiKVGroups`  and `getScaledTiDBGroups` will return `[]*Plan` and `error` instead. we could handle the error later.

pkg/autoscaling/calculation.go
```diff
@@ -307,35 +326,55 @@ func getScaledTiDBGroups(etcdClient *clientv3.Client, healthyInstances []instanc
...
 	var plans []*Plan
-	for groupLabel, groupInstances := range planMap {
+	for groupName, groupInstances := range planMap {
+		resourceType := resourceTypeMap[groupName]
 		plans = append(plans, &Plan{
-			Component: componentType.String(),
-			Count:     uint64(len(groupInstances)),
+			Component:    componentType.String(),
+			Count:        uint64(len(groupInstances)),
+			ResourceType: resourceType,
```
rleungx, line 369: Do we still need it since we have already recorded the resource type in the label?

howardlau1999: The API definition requires this field

pkg/autoscaling/calculation.go
howardlau1999: Updated. How the API returns the error may be discussed later.

pkg/autoscaling/calculation.go
rleungx: Yes, but if it's duplicated, we can change the API.

howardlau1999:
Then I think we can remove this field in the API

获取 Outlook for Android<https://aka.ms/ghei36>

________________________________

howardlau1999:
For now it's better to keep both

获取 Outlook for Android<https://aka.ms/ghei36>

________________________________

pkg/autoscaling/calculation.go
```diff
@@ -268,74 +278,112 @@ func getCountByResourceType(strategy *Strategy, resourceType string) uint64 {
...
 		}
-		v := store.GetLabelValue(groupLabelKey)
-		buildPlanMap(planMap, v, instance.address)
 	}
-	return buildPlans(planMap, TiKV)
+	return buildPlans(planMap, resourceTypeMap, TiKV), nil
 }
 
-func getScaledTiDBGroups(etcdClient *clientv3.Client, healthyInstances []instance) []*Plan {
+// TODO: error handling while processing labels
```
Yisaer, line 323: I think we can remove this TODO now.

Yisaer (approved):
LGTM

rleungx (approved):
LGTM

ti-srebot (approved):

PR #2842, opened 2020-08-26
tso: remove the use of atomic.Value

JmPotato:

After refactoring `tso` pkg, the `leadership` won't have data race problems any more. So there's no need to use `atomic.Value`. Considering the readability, this pr changes it back to a normal struct field.

Remove the use of `atomic.Value` from Global TSO Allocator.

### Check List

Tests

- Unit test
- Integration test

### Release note

* Remove the use of `atomic.Value` from Global TSO Allocator.


lhy1024 (approved):

ti-srebot (approved):

Yisaer (approved):
LGTM

PR #2846, opened 2020-08-26
placement: add bundle APIs to allow updating group and rule configurations with a single request

disksing:

With placement rule API, we can update Rule config and RuleGroup config separately. But a common requirement is to update the Group configuration and the configuration of all Rules in the Group at the same time.

Some previous discussion can be found in https://github.com/tikv/pd/issues/2680

Introduce a new series of APIs:

- `GET /config/placement-rule` read all rules by groups

return:
```
[
  {
    "group_id": "foo",
    "index": 0,
    "override": false,
    "rules":  [ /* rules in group foo */ ]
  },
  {
    "group_id": "bar",
    "index": 1,
    "override": true,
    "rules":  [ /* rules in group bar */ ]
  },
  ... other groups
]
```

- `GET /config/placement-rule/{group}` read rules from a group

returns

```
{
  "group_id": "foo",
  "index": 0,
  "override": false,
  "rules":  [ /* rules in group foo */ ]
},
```

- `POST /config/placement-rule/{group}` update a group

sends

```
{
  "group_id": "foo",
  "index": 1,
  "override": true,
  "rules":  [ /* rules  */ ]
},
```
It performs in a `replace` semantic which means it works just like first remove all old rules then set the received rules and finally the rules in the group will be the same as received config.

- `POST /config/placement-rule` update full configuration

sends

```
[
  {
    "group_id": "foo",
    "index": 0,
    "override": false,
    "rules":  [ /* rules in group foo */ ]
  },
  {
    "group_id": "bar",
    "index": 1,
    "override": true,
    "rules":  [ /* rules in group bar */ ]
  },
  ... other groups
]
```

It also performs in `replace` semantic too, which means all existing configurations will be dropped and replaced by received configurations.

- `DELETE /config/placement-rules/{group}` remove a group

It removes both configuration of the group and also all rules belong to the group. 

- `DELETE /config/placement-rules/{pattern}?regexp` remove groups by parttern

It removes groups that have ID match the pattern.

### Check List

Tests
- Unit test

Code changes

### Release note
- Add placement APIs to allow updating group and rule configurations with a single request

### DIFF
--- a/server/api/rule_test.go
+++ b/server/api/rule_test.go
@@ -615,3 +615,80 @@ func (s *testRuleSuite) TestBatch(c *C) {
 		}
 	}
 }
+
+func (s *testRuleSuite) TestBundle(c *C) {
+	// GetAll
+	b1 := placement.GroupBundle{
+		ID: "pd",
+		Rules: []*placement.Rule{
+			{GroupID: "pd", ID: "default", Role: "voter", Count: 3},
+		},
+	}
+	var bundles []placement.GroupBundle
+	err := readJSON(testDialClient, s.urlPrefix+"/placement-rule", &bundles)
+	c.Assert(err, IsNil)
+	c.Assert(bundles, HasLen, 1)
+	compareBundle(c, bundles[0], b1)
+
+	// Set
+	b2 := placement.GroupBundle{
+		ID:       "foo",
+		Index:    42,
+		Override: true,
+		Rules: []*placement.Rule{
+			{GroupID: "foo", ID: "bar", Index: 1, Override: true, Role: "voter", Count: 1},
+		},
+	}
+	data, err := json.Marshal(b2)
+	c.Assert(err, IsNil)
+	err = postJSON(testDialClient, s.urlPrefix+"/placement-rule/foo", data)
+	c.Assert(err, IsNil)
+
+	// Get
+	var bundle placement.GroupBundle
+	err = readJSON(testDialClient, s.urlPrefix+"/placement-rule/foo", &bundle)
+	c.Assert(err, IsNil)
+	compareBundle(c, bundle, b2)
+
+	// GetAll again
+	err = readJSON(testDialClient, s.urlPrefix+"/placement-rule", &bundles)
+	c.Assert(err, IsNil)
+	c.Assert(bundles, HasLen, 2)
+	compareBundle(c, bundles[0], b1)
+	compareBundle(c, bundles[1], b2)
+
+	// Delete
+	_, err = doDelete(testDialClient, s.urlPrefix+"/placement-rule/pd")
+	c.Assert(err, IsNil)
+
+	// GetAll again
+	err = readJSON(testDialClient, s.urlPrefix+"/placement-rule", &bundles)
+	c.Assert(err, IsNil)
+	c.Assert(bundles, HasLen, 1)
+	compareBundle(c, bundles[0], b2)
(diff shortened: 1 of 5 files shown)


server/schedule/placement/rule_manager.go
```diff
@@ -437,5 +442,148 @@ func (m *RuleManager) DeleteRuleGroup(id string) error {
...
+}
+
+// SetAllGroupBundles resets full configuration. All old configurations are dropped.
+func (m *RuleManager) SetAllGroupBundles(groups []GroupBundle) error {
+	m.Lock()
+	defer m.Unlock()
+	p := m.ruleConfig.beginPatch()
+	for k := range m.ruleConfig.rules {
+		p.deleteRule(k[0], k[1])
+	}
```
xhebox, line 506: I think there can be an api to delete all old rules/groups for ruleConfig. Why iterate over all rules, if it can be done by reslicing.

server/schedule/placement/rule_manager.go
```diff
@@ -437,5 +442,148 @@ func (m *RuleManager) DeleteRuleGroup(id string) error {
...
+
+// GetGroupBundle returns a group and all rules belong to it.
+func (m *RuleManager) GetGroupBundle(id string) (b GroupBundle) {
+	m.RLock()
+	defer m.RUnlock()
+	b.ID = id
+	if g := m.ruleConfig.groups[id]; g != nil {
+		b.Index, b.Override = g.Index, g.Override
+	}
+	for _, r := range m.ruleConfig.rules {
```
xhebox, line 490: Is there a need to filter rules, if the specified group is empty/non-exist? A similar problem raises on `SetGroupBundle`, and `DeleteGroupBundle`: If it is not a regexp request, a check can be dome by a simple map operation.

server/api/rule_test.go
```diff
@@ -615,3 +616,92 @@ func (s *testRuleSuite) TestBatch(c *C) {
...
+	b1 := placement.GroupBundle{
+		ID: "pd",
+		Rules: []*placement.Rule{
+			{GroupID: "pd", ID: "default", Role: "voter", Count: 3},
+		},
+	}
+	var bundles []placement.GroupBundle
+	err := readJSON(testDialClient, s.urlPrefix+"/placement-rule", &bundles)
+	c.Assert(err, IsNil)
+	c.Assert(bundles, HasLen, 1)
```
xhebox, line 631: Looks like the test failed due to other inserted rules.

xhebox (commented):
Rest LGTM

server/api/rule_test.go
disksing: Thanks, I'll take a look.

server/schedule/placement/rule_manager.go
disksing: Good point.

server/schedule/placement/rule_manager.go
disksing: We cannot simply clear the map because we need to record all delete operations in the patch and later we need to apply the operations to persistent storage.

xhebox: Oh, yes. Persistence needs the records.

Yisaer (approved):
LGTM

xhebox (approved):
LGTM

ti-srebot (approved):

PR #2847, opened 2020-08-26
metrics: add store pick in hot peer schedule direction metrics

lhy1024:

add store pick

image

image

image

### Check List

Tests

- Unit test


rleungx (approved):
LGTM

Issue #2732, opened 2020-08-07
request scattering region considering table scattering distribution


Yisaer:
### Describe your feature request related problem
Currently, the region scatter only consider the cluster-level region scatter distribution.  When there are multi sessions executing `split table`, it will cause the unbalance of the distribution(eg: leader) with single table's perspective.

Here's steps to reproduce (version v4.0.4):
1. enable `tidb_wait_split_region_finish` and `tidb_scatter_region` for each session.
2. execute `create table t1 (a int, b int,index idx1(a)) shard_row_id_bits = 10 pre_split_regions=8;` and `create table t2 (a int, b int,index idx1(a)) shard_row_id_bits = 10 pre_split_regions=8;` at same time by these 2 sessions.
3. check each table's leader distribution.

```sh
mysql> select count(s.region_id) cnt, s.index_name, p.store_id from INFORMATION_SCHEMA.TIKV_REGION_STATUS s join INFORMATION_SCHEMA.tikv_region_peers p on s.region_id = p.region_id where s.table_name = 't1' and p.is_leader = 1 group by index_name, p.store_id order by index_name,cnt desc;
+-----+------------+----------+
| cnt | index_name | store_id |
+-----+------------+----------+
|  42 | NULL       |        4 |
|  41 | NULL       |      105 |
|  40 | NULL       |      104 |
|  37 | NULL       |       88 |
|  35 | NULL       |      125 |
|  34 | NULL       |        1 |
|  27 | NULL       |        5 |
|   1 | idx1       |       88 |
+-----+------------+----------+
8 rows in set (0.06 sec)

mysql> select count(s.region_id) cnt, s.index_name, p.store_id from INFORMATION_SCHEMA.TIKV_REGION_STATUS s join INFORMATION_SCHEMA.tikv_region_peers p on s.region_id = p.region_id where s.table_name = 't2' and p.is_leader = 1 group by index_name, p.store_id order by index_name,cnt desc;
+-----+------------+----------+
| cnt | index_name | store_id |
+-----+------------+----------+
|  48 | NULL       |        5 |
|  40 | NULL       |      125 |
|  37 | NULL       |        1 |
|  35 | NULL       |      104 |
|  33 | NULL       |      105 |
|  32 | NULL       |       88 |
|  31 | NULL       |        4 |
|   1 | idx1       |        1 |
+-----+------------+----------+
8 rows in set (0.05 sec)
```

### Describe the feature you'd like
[...]

PR #2849, opened 2020-08-26
scatter: support group level scattering for regions

Yisaer:

Resolve https://github.com/tikv/pd/issues/2732

Add group parameter for the Scatter Region. If the group is defined, the regions with the same group would be scattered in the group level. This would be quite helpful If there are 2 `split table` running at the same time as the `tableID` would become the group for the regions in this case.

### Check List

Tests

- Unit test (simulate the concurrency visiting by regions in different groups.)

Related changes


server/schedule/region_scatterer.go
```diff
@@ -33,44 +33,75 @@ const regionScatterName = "region-scatter"
...
 
-func (s *selectedStores) put(id uint64) bool {
+func (s *selectedStores) put(id uint64, group string) bool {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if _, ok := s.stores[id]; ok {
 		return false
 	}
 	s.stores[id] = struct{}{}
+	if group == "" {
```
nolouch, line 96:
```
put(4, "")
// the follower operator will alway return false
put(4, "t1")
put(4, "t2")
put(4, "t3")
```

server/api/operator.go
```diff
@@ -275,7 +275,11 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
 			h.r.JSON(w, http.StatusBadRequest, "missing region id")
 			return
 		}
-		if err := h.AddScatterRegionOperator(uint64(regionID)); err != nil {
+		group, ok := input["group"].(string)
```
rleungx, line 278: Do we need to support multi-group?

server/schedule/region_scatterer.go
Yisaer: good catch, updated.

server/schedule/region_scatterer.go
```diff
@@ -225,16 +276,33 @@ func (r *RegionScatterer) selectPeerToReplace(stores map[uint64]*core.StoreInfo,
 		return nil
 	}
 
-	target := candidates[rand.Intn(len(candidates))]
+	minPeer := uint64(math.MaxUint64)
+	var selectedCandidateID uint64
+	selectedCandidateID = 0
```
JmPotato, line 281:
```suggestion
```

I think this line is unnecessary.

server/schedule/region_scatterer.go
```diff
@@ -33,59 +33,109 @@ const regionScatterName = "region-scatter"
...
+func (s *selectedLeaderStores) get(id uint64, group string) uint64 {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	return s.stores[id]
+	if group == "" {
+		return s.stores[id]
+	}
+	distribution, ok := s.groupDistribution[group]
+	if !ok {
+		s.groupDistribution[group] = make(map[uint64]uint64)
```
rleungx, line 62: Why do we need it?

server/schedule/region_scatterer.go
```diff
@@ -250,18 +318,30 @@ func (r *RegionScatterer) collectAvailableStores(region *core.RegionInfo, contex
...
+func (r *RegionScatterer) collectAvailableLeaderStores(group string, peers map[uint64]*metapb.Peer, context engineContext) uint64 {
 	m := uint64(math.MaxUint64)
-	id := uint64(0)
+	minTableLeaderStores := map[uint64]struct{}{}
 	for storeID := range peers {
-		count := context.selectedLeader.get(storeID)
+		count := context.selectedLeader.get(storeID, group)
+		if m > count {
+			m = count
+			minTableLeaderStores = map[uint64]struct{}{}
```
rleungx, line 327: why init it twice?

server/schedule/region_scatterer.go
```diff
@@ -250,18 +318,30 @@ func (r *RegionScatterer) collectAvailableStores(region *core.RegionInfo, contex
...
+			minTableLeaderStores[storeID] = struct{}{}
+		} else if m == count {
```
rleungx, line 329: Seems you can directly use if `m >= count`

server/schedule/region_scatterer.go
Yisaer: updated.

server/schedule/region_scatterer.go
Yisaer: The `minTableLeaderStores` here is actually stored the candidates for the target leader store. So if we found a new "min count number" here, we should reset the `minTableLeaderStores`. Also `minTableLeaderStores`'s name is misleading, I rename it to `tableLeaderStoreCandidates`.

server/schedule/region_scatterer.go
Yisaer: ditto.

server/schedule/region_scatterer.go
```diff
@@ -250,18 +316,34 @@ func (r *RegionScatterer) collectAvailableStores(region *core.RegionInfo, contex
 	return targets
 }
 
-func (r *RegionScatterer) searchLeastleaderStore(peers map[uint64]*metapb.Peer, context engineContext) uint64 {
+// selectAvailableLeaderStores select the target leader store from the candidates. The candidates would be collected by
+// the existed peers store depended on the leader counts in the group level.
+func (r *RegionScatterer) selectAvailableLeaderStores(group string, peers map[uint64]*metapb.Peer, context engineContext) uint64 {
 	m := uint64(math.MaxUint64)
-	id := uint64(0)
+	tableLeaderStoreCandidates := map[uint64]struct{}{}
```
nolouch, line 323: Why need `tableLeaderStoreCandidates`? I think we can pick the min leader store directly.

Yisaer: If there exist multipule stores have the same leader count(also the minest) in group level, we will choose the one which have the least leader count in cluster level.

Yisaer: updated.

server/schedule/region_scatterer.go
```diff
@@ -33,59 +33,108 @@ const regionScatterName = "region-scatter"
...
 func newSelectedStores() *selectedStores {
 	return &selectedStores{
-		stores: make(map[uint64]struct{}),
+		stores:            make(map[string]map[uint64]struct{}),
+		groupDistribution: make(map[string]map[uint64]uint64),
 	}
 }
 
-func (s *selectedStores) put(id uint64) bool {
+func (s *selectedStores) put(id uint64, group string) bool {
```
nolouch, line 89: Can merge `selectedStores` and `selectedLeaderStores `?

Yisaer: updated.

server/schedule/region_scatterer.go
```diff
@@ -30,64 +30,84 @@ import (
 
 const regionScatterName = "region-scatter"
 
-type selectedLeaderStores struct {
-	mu     sync.Mutex
-	stores map[uint64]uint64 // storeID -> hintCount
+type selectedStores struct {
+	mu sync.Mutex
+	// If checkExist is true, after each putting operation, an entry with the key constructed by group and storeID would be put
+	// into "stores" map. And the entry with same key(storeId,group) couldn't be put before "stores" being reset
```
JmPotato, line 36:
```suggestion
	// into "stores" map. And the entry with the same key(storeID, group) couldn't be put before "stores" being reset
```

server/schedule/region_scatterer.go
```diff
@@ -225,20 +246,36 @@ func (r *RegionScatterer) selectPeerToReplace(stores map[uint64]*core.StoreInfo,
...
+	minPeer := uint64(math.MaxUint64)
+	var selectedCandidateID uint64
+	for _, candidate := range candidates {
+		count := context.selectedPeer.get(candidate.GetID(), group)
+		if count < minPeer {
+			minPeer = count
+			selectedCandidateID = candidate.GetID()
+		}
+	}
+	if selectedCandidateID < 1 {
```
rleungx, line 258: When will it happen?

Yisaer: Almost never happen

rleungx (approved):
LGTM

JmPotato (approved):
LGTM

ti-srebot (approved):

nolouch (commented):
LGTM. BTW, you need update the `client` interface, maybe you can fire new PR and label it as compatibility changes

Yisaer:
> LGTM. BTW, you need to update the `client` interface, maybe you can fire new PR and label it as compatibility changes

got it.

Issue #2858, opened 2020-08-28
Provide a new way to manage TSO Allocators


JmPotato:
### Describe your feature request related problem

After introducing the `TSO Allocator` through #2803, we need to provide a new way to manage all TSO Allocators a PD server will hold. As you can see from [the design document of TSO Allocator](https://docs.google.com/document/d/1t4D2M5W0dqduNTey1aTute4umuEQrKazU-ipSwFAaLM/edit#heading=h.qolqioffe8ac), a PD server may hold more than one TSO Allocators.

### Describe the feature you'd like

The old Global TSO Allocator couples PD leader election strongly, but after introducing the Local TSO Allocator, we need a more centralized and clear way to manage the TSO allocators.

### Describe alternatives you've considered

I'd like to introduce a new model named `AllocatorManager` to manage the TSO allocators a PD server will hold. As an `AllocatorManager`,  it should be in charge of maintaining TSO allocators' leadership, checking election priority, forwarding TSO allocation requests to correct TSO Allocators, and so on.

JmPotato:
@nolouch @disksing @Yisaer PTAL

PR #2859, opened 2020-08-28
tso: introduce the AllocatorManager

JmPotato:

Solve #2858 

### Check List

Tests

- Unit test
- Integration test

### Release note

* Introduce the `AllocatorManager` to manage all TSO Allocators a PD server will hold.

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -825,7 +825,7 @@ func (h *Handler) GetEmptyRegion() ([]*core.RegionInfo, error) {
 
 // ResetTS resets the ts with specified tso.
 func (h *Handler) ResetTS(ts uint64) error {
-	tsoAllocator := h.s.tsoAllocator
+	tsoAllocator, _ := h.s.tsoAllocatorManager.GetAllocator("global")
 	if tsoAllocator == nil {
 		return ErrServerNotStarted
 	}
--- a/server/handler.go
+++ b/server/handler.go
@@ -825,7 +825,10 @@ func (h *Handler) GetEmptyRegion() ([]*core.RegionInfo, error) {
 
 // ResetTS resets the ts with specified tso.
 func (h *Handler) ResetTS(ts uint64) error {
-	tsoAllocator, _ := h.s.tsoAllocatorManager.GetAllocator("global")
+	tsoAllocator, err := h.s.tsoAllocatorManager.GetAllocator("global")
+	if err != nil {
+		return err
+	}
 	if tsoAllocator == nil {
 		return ErrServerNotStarted
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -470,10 +470,11 @@ func (s *Server) LoopContext() context.Context {
 
 func (s *Server) startServerLoop(ctx context.Context) {
 	s.serverLoopCtx, s.serverLoopCancel = context.WithCancel(ctx)
-	s.serverLoopWg.Add(3)
+	s.serverLoopWg.Add(4)
 	go s.leaderLoop()
 	go s.etcdLeaderLoop()
 	go s.serverMetricsLoop()
+	go s.allocatorLoop()
 }
 
 func (s *Server) stopServerLoop() {
@@ -496,6 +497,15 @@ func (s *Server) serverMetricsLoop() {
 			return
 		}
 	}
+}
+
+func (s *Server) allocatorLoop() {
+	defer logutil.LogPanic()
+	defer s.serverLoopWg.Done()
+
+	ctx, cancel := context.WithCancel(s.serverLoopCtx)
+	defer cancel()
+	s.tsoAllocatorManager.AllocatorDaemon(ctx)
 }
 
 func (s *Server) collectEtcdStateMetrics() {
--- a/server/server.go
+++ b/server/server.go
@@ -119,7 +119,7 @@ type Server struct {
 	// for basicCluster operation.
 	basicCluster *core.BasicCluster
 	// for tso.
-	tsoAllocator tso.Allocator
+	tsoAllocatorManager *tso.AllocatorManager
 	// for raft cluster
 	cluster *cluster.RaftCluster
 	// For async region heartbeat.
@@ -348,10 +348,8 @@ func (s *Server) startServer(ctx context.Context) error {
 	s.member.SetMemberBinaryVersion(s.member.ID(), versioninfo.PDReleaseVersion)
 	s.member.SetMemberGitHash(s.member.ID(), versioninfo.PDGitHash)
 	s.idAllocator = id.NewAllocatorImpl(s.client, s.rootPath, s.member.MemberValue())
-	s.tsoAllocator = tso.NewGlobalTSOAllocator(
-		s.member.GetLeadership(),
-		s.rootPath,
-		s.cfg.TsoSaveInterval.Duration,
+	s.tsoAllocatorManager = tso.NewAllocatorManager(
+		ctx, s.member.Etcd(), s.client, s.rootPath, s.cfg.TsoSaveInterval.Duration,
 		func() time.Duration { return s.persistOptions.GetMaxResetTSGap() },
 	)
 	kvBase := kv.NewEtcdKVBase(s.client, s.rootPath)
@@ -1112,12 +1110,11 @@ func (s *Server) campaignLeader() {
 	go s.member.KeepLeader(ctx)
 	log.Info("campaign pd leader ok", zap.String("campaign-pd-leader-name", s.Name()))
 
-	log.Info("initialize the global TSO allocator")
-	if err := s.tsoAllocator.Initialize(); err != nil {
-		log.Error("failed to initialize the global TSO allocator", zap.Error(err))
+	log.Info("setting up the global TSO allocator")
+	if err := s.tsoAllocatorManager.SetUpAllocator(ctx, cancel, "global", s.member.GetLeadership()); err != nil {
+		log.Error("failed to set up the global TSO allocator", zap.Error(err))
 		return
 	}
-	defer s.tsoAllocator.Reset()
 
 	if err := s.reloadConfigFromKV(); err != nil {
 		log.Error("failed to reload configuration", zap.Error(err))
@@ -1135,8 +1132,6 @@ func (s *Server) campaignLeader() {
 	CheckPDVersion(s.persistOptions)
 	log.Info("PD cluster leader is ready to serve", zap.String("pd-leader-name", s.Name()))
 
-	tsTicker := time.NewTicker(tso.UpdateTimestampStep)
-	defer tsTicker.Stop()
 	leaderTicker := time.NewTicker(leaderTickInterval)
 	defer leaderTicker.Stop()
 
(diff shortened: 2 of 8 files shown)


server/tso/allocator_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	switch dcLocation {
+	case "global":
+		am.allocatorGroups[dcLocation] = &allocatorGroup{
+			ctx:        ctx,
+			cancel:     cancel,
+			leadership: leadership,
+			allocator:  NewGlobalTSOAllocator(leadership, am.getAllocatorPath(dcLocation), am.saveInterval, am.maxResetTSGap),
+		}
+		if err := am.allocatorGroups[dcLocation].allocator.Initialize(); err != nil {
+			return errs.ErrSetAllocator.FastGenByArgs(err)
```
rleungx, line 99: Do we need to define this error in the place where the error is generated?

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,168 @@
...
+
+// AllocatorDaemon is used to update every allocator's TSO.
+func (am *AllocatorManager) AllocatorDaemon(serverCtx context.Context) {
+	tsTicker := time.NewTicker(UpdateTimestampStep)
+	defer tsTicker.Stop()
+
+	for {
+		select {
+		case <-tsTicker.C:
+			am.RLock()
```
rleungx, line 127: Is `RLock` enough?

server/handler.go
```diff
@@ -825,7 +825,7 @@ func (h *Handler) GetEmptyRegion() ([]*core.RegionInfo, error) {
 
 // ResetTS resets the ts with specified tso.
 func (h *Handler) ResetTS(ts uint64) error {
-	tsoAllocator := h.s.tsoAllocator
+	tsoAllocator, _ := h.s.tsoAllocatorManager.GetAllocator("global")
```
rleungx, line 828: Why omit this error?

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	go allocatorManager.AllocatorDaemon(serverCtx)
+	return allocatorManager
+}
+
+func (am *AllocatorManager) getAllocatorPath(dcLocation string) string {
+	return path.Join(am.rootPath, dcLocation)
+}
+
+// SetUpAllocator is used to set up an allocator, which will initialize the allocator and put it into allocator daemon.
+func (am *AllocatorManager) SetUpAllocator(ctx context.Context, cancel context.CancelFunc, dcLocation string, leadership *election.Leadership) error {
```
rleungx, line 87: Do we need a child context?

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	allocatorManager := &AllocatorManager{
+		allocatorGroups: make(map[string]*allocatorGroup),
+		etcd:            etcd,
+		client:          client,
+		rootPath:        path.Join(rootPath, "allocator"),
+		saveInterval:    saveInterval,
+		maxResetTSGap:   maxResetTSGap,
+	}
+	// Start the deamon for allocators
+	go allocatorManager.AllocatorDaemon(serverCtx)
```
disksing, line 78: Looks better to start the background goroutine in `server.startServerLoop`. In that way, we can trace the lifetime of the goroutine better.

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,168 @@
...
+// AllocatorDaemon is used to update every allocator's TSO.
+func (am *AllocatorManager) AllocatorDaemon(serverCtx context.Context) {
+	tsTicker := time.NewTicker(UpdateTimestampStep)
+	defer tsTicker.Stop()
+
+	for {
+		select {
+		case <-tsTicker.C:
+			am.RLock()
+			for dcLocation, allocatorGroup := range am.allocatorGroups {
```
disksing, line 128: maybe better to handle different allocators in separated goroutines. When an allocator gets stuck, it has a smaller impact on others.

server/tso/allocator_manager.go
JmPotato: You mean to define this error inside the `Initialize()` rather than here?

server/tso/allocator_manager.go
JmPotato: Oops, it should be `Lock`.

server/tso/allocator_manager.go
JmPotato: This `context` corresponds to the lifetime of an allocator, so we don't need to create a child context here. But maybe I should change a more clear name.

server/tso/allocator_manager.go
JmPotato: Good idea.

server/tso/allocator_manager.go
JmPotato: Ok, let's do it this way.

server/tso/allocator_manager.go
JmPotato: After checking again, I think we need to update the error handling in `tso` pkg. I will create a new PR to address this.

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,187 @@
...
+	allocatorManager := &AllocatorManager{
+		allocatorGroups: make(map[string]*allocatorGroup),
+		etcd:            etcd,
+		client:          client,
+		rootPath:        path.Join(rootPath, "allocator"),
+		saveInterval:    saveInterval,
+		maxResetTSGap:   maxResetTSGap,
+	}
+	return allocatorManager
+}
```
Yisaer, line 82: It seems we didn't use `serverCtx context.Context` here. Do we need to handle this parameter?

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,187 @@
...
+	for {
+		select {
+		case <-tsTicker.C:
+			// Collect all dc-locations first
+			am.RLock()
+			dcLocations := make([]string, 0, len(am.allocatorGroups))
+			for dcLocation := range am.allocatorGroups {
+				dcLocations = append(dcLocations, dcLocation)
+			}
+			am.RUnlock()
```
Yisaer, line 135: I think we could extract this into a function for `AllocatorManager` like `getAllocatorGroups` and controlled it by rwmutex.

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,187 @@
...
+			// Update each allocator concurrently
+			for _, dcLocation := range dcLocations {
+				am.wg.Add(1)
+				go am.updateAllocator(dcLocation)
+			}
+			am.wg.Wait()
```
Yisaer, line 141: `am.updateAllocator(dcLocation)` is controlled by mutex. Is this necessary to use `waitGroup` here?

Yisaer (commented):
Rest LGTM

tools/pd-backup/pdbackup/backup.go
```diff
@@ -73,7 +73,7 @@ func GetBackupInfo(client *clientv3.Client, pdAddr string) (*BackupInfo, error)
 
 	backInfo.AllocIDMax = allocIDMax
 
-	timestampPath := path.Join(rootPath, "timestamp")
+	timestampPath := path.Join(rootPath, "allocator", "global", "timestamp")
```
disksing, line 76: I think we should not change the global timestamp's store path in case of backward compatibility.

Yisaer (approved):
LGTM

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,193 @@
...
+
+	for {
+		select {
+		case <-tsTicker.C:
+			// Collect all dc-locations first
+			allocatorGroups := am.getAllocatorGroups()
+			// Update each allocator concurrently
+			for _, ag := range allocatorGroups {
+				go am.updateAllocator(ag)
+			}
```
Yisaer, line 149: I mean we don't need sync.waitGroup and `go` anymore as `updateAllocator` functions are controlling by the mutex so they must be executed serially.

JmPotato: Oh I get it!

JmPotato:
One more review needed. PTAL @disksing @rleungx

disksing (approved):

ti-srebot (approved):

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (am *AllocatorManager) getAllocatorGroups() []*allocatorGroup {
+	am.RLock()
+	allocatorGroups := make([]*allocatorGroup, 0, len(am.allocatorGroups))
+	for _, ag := range am.allocatorGroups {
+		allocatorGroups = append(allocatorGroups, ag)
+	}
+	am.RUnlock()
+	return allocatorGroups
+}
```
Yisaer, line 135: I suggest unify the style to use `defer am.RUnlock()` here.

Yisaer (approved):
LGTM

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,195 @@
...
+				go am.updateAllocator(ag)
+			}
+			am.wg.Wait()
+		case <-serverCtx.Done():
+			return
+		}
+	}
+}
+
+// updateAllocator is used to update the allocator in the group
```
lhy1024, line 159:
```suggestion
// updateAllocator is used to update the allocator in the group.
```

server/tso/allocator_manager.go
lhy1024: unify the style is better

server/server.go
```diff
@@ -500,6 +499,16 @@ func (s *Server) serverMetricsLoop() {
 	}
 }
 
+func (s *Server) allocatorLoop() {
```
rleungx, line 502: prefer to rename it to `tsoAllocatorLoop` or add a comment for it

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,195 @@
...
+		// Resetting the allocator will clear TSO in memory
+		ag.allocator.Reset()
+		return
+	default:
+	}
+	if !ag.isInitialized {
+		log.Info("allocator has not been initialized yet", zap.String("dc-location", ag.dcLocation))
+		return
+	}
+	if !ag.leadership.Check() {
```
rleungx, line 175: Do we also need a log here?

server/tso/allocator_manager.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	return path.Join(am.rootPath, dcLocation)
+}
+
+// SetUpAllocator is used to set up an allocator, which will initialize the allocator and put it into allocator daemon.
+func (am *AllocatorManager) SetUpAllocator(parentCtx context.Context, parentCancel context.CancelFunc, dcLocation string, leadership *election.Leadership) error {
+	am.Lock()
+	defer am.Unlock()
+	switch dcLocation {
+	case "global":
```
rleungx, line 98: How about using a constant for "global"?

JmPotato: What about using a random string to be the value of this constant to prevent it from being duplicated from the user's `dc-location`? Like this:

```go
const globalDCLocation = "GLOBAL_b6b3f51996c7da2d8adcbb44baa39750"
```

Yisaer: I think we should prevent "global" from being duplicated by user's input dc-location by adding validation in the api instead of generating a random name which will be misleading for the user in the view.

PR #2861, opened 2020-08-31
autoscaling: fix label and scaled group

rleungx:

We expect to output the complete scaled group even if there is no action of auto scaling. But for now, it will only output nil as a result. 
Another problem is if there is no auto scaling group, it will create a new one. But in the previous implementation, it didn't contain a label in the returned value.

This PR fixes the above problems.

### Check List

Tests

- Manual test

### DIFF
--- a/pkg/autoscaling/calculation.go
+++ b/pkg/autoscaling/calculation.go
@@ -99,16 +99,24 @@ func getPlans(rc *cluster.RaftCluster, querier Querier, strategy *Strategy, comp
 	usage := totalCPUUseTime / totalCPUTime
 	maxThreshold, minThreshold := getCPUThresholdByComponent(strategy, component)
 
+	groups, err := getScaledGroupsByComponent(rc, component, instances)
+	if err != nil {
+		// TODO: error handling
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/autoscaling/calculation.go
```diff
@@ -405,6 +403,16 @@ func findBestGroupToScaleOut(rc *cluster.RaftCluster, strategy *Strategy, scaleO
 		Component:    component.String(),
 		Count:        0,
 		ResourceType: resources[0].ResourceType,
+		Labels: []*metapb.StoreLabel{
+			{
+				Key:   groupLabelKey,
+				Value: autoScalingGroupLabelKeyPrefix + component.String(),
```
howardlau1999, line 409: it is better to leave a TODO here because we should make sure the group name is not duplicated when there are already other groups but PD still decides to create a new group.

howardlau1999:
The rest LGTM

pkg/autoscaling/calculation.go
rleungx: Done.

howardlau1999 (approved):

Yisaer (approved):
LGTM

PR #2864, opened 2020-08-31
scheduler: revert low space threshold hard limit

rleungx:

The previous PR uses a hard limit for low space. If the capacity beyond 500G, it will use 100G as its low space threshold. It doesn't seem reasonable.

This PR changes back to the original way to control low space.

### Check List

Tests

- Unit test

Related changes

### Release note

- Change back to the original way to control the low space threshold

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -31,9 +31,7 @@ import (
 const (
 	// Interval to save store meta (including heartbeat ts) to etcd.
 	storePersistInterval = 5 * time.Minute
-	lowSpaceThreshold    = 100 * (1 << 10) // 100 GB
-	highSpaceThreshold   = 300 * (1 << 10) // 300 GB
-	mb                   = 1 << 20         // megabyte
+	mb                   = 1 << 20 // megabyte
 )
 
 // StoreInfo contains information about a store.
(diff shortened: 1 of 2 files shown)


HunDunDM (approved):
LGTM

zz-jason:
How MySQL and other databases determine the low space event?

lhy1024 (commented):
Don’t we do other judgment on large disks?

zz-jason:
Here is an example of MySQL: https://help.scalegrid.io/docs/mysql-alerts-rules

> These are the default alert rules automated by ScaleGrid for all MySQL clusters:
>
> * CPU - Total (%) greater than 60%
> * Available disk space (%) less than 20%
> * Replication Delay (secs) greater than 300 secs
> * Connections (per second) greater than 1000 per second

rleungx:
> Here is an example of MySQL: https://help.scalegrid.io/docs/mysql-alerts-rules
> 
> > These are the default alert rules automated by ScaleGrid for all MySQL clusters:
> > 
> > * CPU - Total (%) greater than 60%
> > * Available disk space (%) less than 20%
> > * Replication Delay (secs) greater than 300 secs
> > * Connections (per second) greater than 1000 per second

It seems it's the same as our original way.

zz-jason:
CockroachDB [Node is running low on disk space](https://www.cockroachlabs.com/docs/stable/monitoring-and-alerting.html#node-is-running-low-on-disk-space):

> * Rule: Send an alert when a node has less than 15% of free space remaining.
> * How to detect: Divide the capacity metric by the capacity_available metric in the node's _status/vars output.

ZenoTan:
LGTM

rleungx:
> Don’t we do other judgment on large disks?

We need to find a better way.

zz-jason (commented):
LGTM

ti-srebot (approved):

PR #2869, opened 2020-08-31
fix empty http response in scheduler

lhy1024:

we should call `.Error` to convert `error` to `string`, otherwise we will get an empty result.

### Check List

Tests

- Unit test


lhy1024:
@howardlau1999 PTAL

lhy1024:
cc @nolouch 
do we need to cherry-pick it to 4.0?

rleungx (approved):
LGTM

howardlau1999:
LGTM

PR #2870, opened 2020-08-31
*: fix missing calls to .Error()

howardlau1999:
Signed-off-by: Howard Lau <howardlau1999@hotmail.com>

We should call `.Error` to convert `error` to `string`, otherwise we will get an empty result.

Add missing `.Error()` call for API error responses.

### Check List

Tests

- No test

### Release note

- No release note


lhy1024 (approved):

nolouch (approved):
LGTM

ti-srebot (approved):

PR #2877, opened 2020-09-01
[DNM] Test Bench

lhy1024:


zz-jason:
how to read the benchmark result?

lhy1024:
> how to read the benchmark result?

@zz-jason 
it will display some metrics that master branch and this pr branch, in this case, we hope that `balance_region_operator_count` and `balance_leader_operator_count` is less, it means that create less operator when scale out.

lhy1024:
> /bench simulator

Do we need to go to compatible format？

ZenoTan:
> > /bench simulator
> 
> Do we need to go to compatible format？

Yes. I will fix the format later.

nolouch:
What the meaning of `Send Maximum`,`Send Minimum`?

ZenoTan:
`Send Maximum` is is maximum snapshots send in all stores. Refer to `statistics.go` in simulator.

Issue #2884, opened 2020-09-03
refactor task: remove mockoptions and Options interfaces.


disksing:
### Problem
It is too much work to add a new schedule config! Typically, we need to:
- add new config in `server/config.Config` and `server/config.PersistOptions`
- add new methods in interfaces `opt.Options` and `statistics.Options`
- implement a new method in `RaftCluster`
- add new config in `mockoptions` and implement the new method

What's worse, the `mockoptions` used in the tests are not the same as the actual running `Config`, and there is no guarantee that the behaviour of the two is consistent.

### How to fix
- Eliminate the dependency of config package on `schedule/*`
- Replace `mockoptions` with concrete `config.PersistOptions`
- Remove interfaces

PR #2888, opened 2020-09-03
config: resolve schedule dependency

disksing:

First step of https://github.com/tikv/pd/issues/2884

Make sure config does not import `server/schedule` any more.

### Check List

Tests
- Unit test

### Release note

- No release note


server/config/config_test.go
```diff
@@ -39,6 +36,14 @@ var _ = Suite(&testConfigSuite{})
 
 type testConfigSuite struct{}
 
+func (s *testConfigSuite) SetUpSuite(c *C) {
```
lhy1024, line 39: why add it?

lhy1024 (commented):
the rest LGTM

server/config/config_test.go
disksing: Because to avoid dependency, cannot import `schedulers` any more. So we need to register scheduler names that we'll use in tests.

server/schedule/scheduler.go
```diff
@@ -107,12 +108,7 @@ func RegisterSliceDecoderBuilder(typ string, builder ConfigSliceDecoderBuilder)
 		log.Fatal("duplicated scheduler", zap.String("type", typ))
 	}
 	schedulerArgsToDecoder[typ] = builder
-}
-
-// IsSchedulerRegistered check where the named scheduler type is registered.
-func IsSchedulerRegistered(name string) bool {
-	_, ok := schedulerMap[name]
-	return ok
+	config.RegisterScheduler(typ)
```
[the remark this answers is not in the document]
disksing: To avoid `config` depend on `scheduler`. I have to make 2 maps, one is in `scheduler` for creating schedulers, one is in `config` to validate configurations. So the register process should update 2 places at the same time.

lhy1024 (approved):

nolouch (commented):
LGTM

ti-srebot (approved):

PR #2619, opened 2020-07-09
 replication: sync replication mode with sync_timeout

5kbpers:
### What problem does this PR solve?

close https://github.com/pingcap/pd/issues/2490

### What is changed and how it works?

* Add a goroutine for getting the replication status from the pd leader periodically
* Add an HTTP API for getting the replication status and update the last time for sync replication status from followers
* Replicate the status file with timeout

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

Code changes

- Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/pingcap/pd/blob/master/docs/development.md#updating-api-documentation))

Side effects

- Possible performance regression
- Increased code complexity

### Release note

* replication_mode: sync replication status mode with sync_timeout

Issue #2490, opened 2020-06-02
Replication mode: state sync replicate not work when primary and dr disconected


disksing:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
- setup replication mode to dr-auto-sync
- disconnect primary and dr DC

2. What did you expect to see?
switch the async and recover service in primary

3. What did you see instead?
cannot switch state. Because PD is also disconnected, state sync replication fails.

4. What version of PD are you using (`pd-server -V`)?
latest master

5kbpers:
My preliminary thought is to implement it with the [lease API of etcd](https://etcd.io/docs/v3.4.0/learning/api/#lease-api).
Each follower grants a lease which TTL is `wait_pd_timeout`, and renew the lease every `wait_pd_timeout` * 50%. After `wait_pd_timeout` * 95%, follower would think the replication state is not reliable.

5kbpers:
But there is the restriction is that getting the replication status and renewing the lease must be combined as an atomic operation.

disksing:
> But there is the restriction is that getting the replication status and renewing the lease must be combined as an atomic operation.

em, sorry I don't get it. Why do we have this restriction?

5kbpers:
> > But there is the restriction is that getting the replication status and renewing the lease must be combined as an atomic operation.
> 
> em, sorry I don't get it. Why do we have this restriction?

😅 My original thought is that we should save the local replication mode status (stored in the file) before updating the lease.

disksing:
I think it can be a stand alone feature, simply trace connection state between PD nodes, has nothing to do with replication mode. Then replication mode can use it to determine whether to switch state.

nolouch:
fixed by https://github.com/tikv/pd/pull/2898/

PR #2898, opened 2020-09-04
replication: add member wait async timeout

disksing:
Co-authored-by: 5kbpers <tangminghua@pingcap.com>

1/2 PR for #2490
extract from original PR #2619

The main problem is that when PD's leader and follower are disconnected. From the follower side, we cannot know the state at the leader's side.
The fix is to introduce timeout before switching to async and maintain heartbeats between leader and followers. In this way, we can determine the state of the leader in a period of time after the disconnection.

- add new config `waitAsynctimeout`
- add API to allow followers to periodically sync timeout
- leader only switch to async state when it can be sure that all followers are timeout
- add test case

next step:
- periodically sync timeout at the follower side

### Check List

Tests
- Unit test

Code changes
- Has configuration change

### Release note
- No release note


HunDunDM (commented):
Why can't it be changed through the dynamically configured api?

disksing:
@HunDunDM what is dynamically configured api?

server/replication/replication_mode.go
```diff
@@ -209,6 +224,23 @@ func (m *ModeManager) loadDRAutoSync() error {
...
+		return true
+	}
+	// make sure all members are timeout.
+	for _, t := range m.drMemberWaitAsyncTime {
+		if time.Since(t) <= timeout {
+			return false
+		}
+	}
+	// make sure all members that have synced with previous leader are timeout.
+	return time.Since(m.initTime) > timeout
```
Yisaer, line 241: If the pd leader is started at `0:00` and the follower is started at `0:03`, and the `WaitAsyncTimeout` is 2min. If the `drCheckAsyncTimeout` happened in the `2:01` and the follower's request arrive at the `2:02`. The logic here will judge that all the members are timeout. Is this expeceted?

Yisaer (commented):
The rest LGTM

server/replication/replication_mode.go
disksing: Should be ok. In order to determine if data is synced, we need to check if `time.Now() < lastSync.Add(2min)` too.

Yisaer: got it

Yisaer (approved):
LGTM

HunDunDM:
> @HunDunDM what is dynamically configured api?

`[POST] /config/replication-mode`

disksing:
> > @HunDunDM what is dynamically configured api?
> 
> `[POST] /config/replication-mode`

Seems to be a good idea...

disksing:
Looks better to add a sub handler such as `/config/replication-mode/heartbeat`

HunDunDM:
This is not a config update, I think it is not a good idea.

HunDunDM (approved):
LGTM

ti-srebot (approved):

Yisaer:
@disksing   You need to fix the error

```
[2020-09-09T10:38:19.988Z] server/replication/replication_mode_test.go:254:36: undefined: mockoption
```

PR #2905, opened 2020-09-07
client: support scatter group

Yisaer:

Support scatter client in client which is also an compatibility change. ref https://github.com/tikv/pd/pull/2849

### Check List

Tests

- Unit test

Code changes

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -85,7 +85,7 @@ type Client interface {
 	UpdateServiceGCSafePoint(ctx context.Context, serviceID string, ttl int64, safePoint uint64) (uint64, error)
 	// ScatterRegion scatters the specified region. Should use it for a batch of regions,
 	// and the distribution of these regions will be dispersed.
-	ScatterRegion(ctx context.Context, regionID uint64) error
+	ScatterRegion(ctx context.Context, regionID uint64, group string) error
 	// GetOperator gets the status of operator of the specified region.
 	GetOperator(ctx context.Context, regionID uint64) (*pdpb.GetOperatorResponse, error)
 	// Close closes the client.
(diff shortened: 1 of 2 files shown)


nolouch:
This change changes the function signature of the interface, which will cause upgrade problems due to circular dependencies in TiDB. cc @overvenus @HunDunDM 
maybe we can add a new interface "ScatterRegionWithGroup"

Yisaer:
> ScatterRegionWithGroup

got it.

client/client.go
```diff
@@ -86,6 +86,10 @@ type Client interface {
 	// ScatterRegion scatters the specified region. Should use it for a batch of regions,
 	// and the distribution of these regions will be dispersed.
 	ScatterRegion(ctx context.Context, regionID uint64) error
+	// ScatterRegionWithGroup scatters the specified region. Should use it for a batch of regions , and the distribution
+	// of these regions will be dispersed If they have the same group.
+	// If the group is not defined, it will work as ScatterRegion.
+	ScatterRegionWithGroup(ctx context.Context, regionID uint64, group string) error
```
overvenus, line 92: Please consider name it `ScatterRegionWithOption`, so that we can avoid adding new APIs when there is a new feature requirement in the future.
```suggestion
	ScatterRegionWithOption(ctx context.Context, regionID uint64, option Option...) error
```

Yisaer: good idea.

overvenus:
Can it be cherrypicked to release-4.0?

Yisaer:
> Can it be cherrypicked to release-4.0?

sure.

overvenus (approved):
LGTM

nolouch (approved):
LGTM

ti-srebot (approved):

PR #2886, opened 2020-09-03
[DNM] Use Joint Consensus

HunDunDM:
### What problem does this PR solve?

This PR is used to test the whole at any time while #2860 is in progress. 

**NOTE: This PR will not merge.**

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

### Release note

- No release note

Issue #2860, opened 2020-08-28
Use Joint Consensus


HunDunDM:
## Development Task

Etcd has supported joint consensus, we need to port the change and finally support it in TiKV. This issue records the progress in PD. Please refer to [tikv/tikv#7587](https://github.com/tikv/tikv/issues/7587) for the total progress.

Development branch: [hundundm/pd/joint-wip](https://github.com/HunDunDM/pd/tree/joint-wip)

### Development progress

* [x] some preparations
* [x] replace operator.Cluster to opt.Cluster
* [x] steps
  * [x] add DemoteFollower
  * [x] add DemoteVoter
  * [x] add ChangePeerV2Enter
  * [x] add ChangePeerV2Leave
  * [x] add unit test
* [x] builder
  * [x] refine compatible builder
  * [x] add joint consensus builder
  * [x] add leave joint state operator
  * [x] refine unit test
* [x] checker
  * [x] add joint state checker
  * [x] add joint state checker unit test
  * ~~(optional) refine learner checker~~
* [x] fix original test with error

### Split PRs

* [x] some preparations
  * [x] [Merged] Update kvproto in dependencies #2729 
  * [x] [Merged] schedule: add IsFeatureSupported in Cluster Interface #2735 
  * [x] [Merged] testutil: refine operator_check #2887 
  * [x] [Merged] cluster: refine IsFeatureSupported #2889
  * [x] [Merged] client: fix typo file name #2890
  * [x] [Merged] test: refine unit test #2893
  * [x] [Merged] operator: replace operator.Cluster to opt.Cluster #2936
* [x] step
  * [x] [Merged] operator: ConfVerChanged returns uint64 #2919
  * [x] [Merged] schedule: add some operator steps related to joint consensus #2895
* [x] builder
  * [x] [Merged] operator: refactor the builder to prepare to support the joint consensus #2938
  * [x] [Merged] operator: implement joint consensus builder #2944
  * [x] [Merged] operator: optimize builder code for the previous PRs #2947
    * Unite `hasAbilityLeader` and `allowLeader `
    * Use `allowDemote` + `useJointConsensus` instead of `useJointConsensus`
    * Consider `promote learner + demote voter` as a step
    * Rename `Empty` to `IsEmpty`
    * Rename `b2i` to `BoolToInt`
    * Unite `newBuilderWithBasicCheck` and `NewBuilder`
* [x] [Merged] schedule: add joint state checker #2952
* ~~(optional) refine learner checker~~

[...]

PR #2919, opened 2020-09-07
operator: change the return type of ConfVerChanged from bool to uint64

HunDunDM:

This is a child PR of #2886 

For details, please refer to #2860 

* change the return type of `ConfVerChanged` from `bool` to `uint64`
* adjust the `ConfVerChanged` behavior of `RemovePeer`: 
  * when `PeerID` is specified, if `Peer.Id` is changed, it is also regarded as `ConfVerChanged = 1`

### Check List

Tests

- Unit test

### Release note

- No release note


server/schedule/operator/step.go
```diff
@@ -220,12 +231,13 @@ func (pl PromoteLearner) Influence(opInfluence OpInfluence, region *core.RegionI
...
+	FromStore, PeerID uint64
 }
 
-// ConfVerChanged returns true if the conf version has been changed by this step
-func (rp RemovePeer) ConfVerChanged(region *core.RegionInfo) bool {
-	return region.GetStorePeer(rp.FromStore) == nil
+// ConfVerChanged returns the delta value for version increased by this step.
+func (rp RemovePeer) ConfVerChanged(region *core.RegionInfo) uint64 {
+	id := region.GetStorePeer(rp.FromStore).GetId()
+	return typeutil.BoolToUint64(id == 0 || (rp.PeerID != 0 && id != rp.PeerID))
```
rleungx, line 240: Why?

server/schedule/operator/step.go
```diff
@@ -197,19 +208,19 @@ func (pl PromoteLearner) String() string {
...
+		return peer.GetId() == pl.PeerID
 	}
 	return false
 }
 
 // CheckSafety checks if the step meets the safety properties.
 func (pl PromoteLearner) CheckSafety(region *core.RegionInfo) error {
 	peer := region.GetStorePeer(pl.ToStore)
-	if peer == nil {
+	if peer.GetId() != pl.PeerID {
```
Yisaer, line 223: Do we still need to check if peer is `nil` here?

pkg/typeutil/convension.go
```diff
@@ -34,3 +34,11 @@ func Uint64ToBytes(v uint64) []byte {
...
 	return b
 }
+
+// BoolToUint64 converts bool to uint64.
+func BoolToUint64(b bool) uint64 {
+	if b {
+		return 1
+	}
+	return 0
+}
```
Yisaer, line 44: What about adding a unit test here?

Yisaer (commented):
Rest LGTM

server/schedule/operator/step.go
HunDunDM: Add some description.

server/schedule/operator/step.go
HunDunDM: The `GetXX` function of `ProtoBuf` has its own `nil` judgment.

pkg/typeutil/convension.go
HunDunDM: I think this is just a conditional operator.

server/schedule/operator/step.go
disksing: When will `PeerID` be set and when not set?

disksing (approved):

JmPotato (approved):
LGTM

ti-srebot (approved):

Yisaer (approved):
LGTM

server/schedule/operator/step.go
HunDunDM: It will not be set now, and the `builder` code will be updated later to use this new logic.

PR #2935, opened 2020-09-09
config, schedule: use default config in tests

disksing:

Tests do not use default configuration.

When we change the default value of configuration, it does not affect any tests. If the change is bad, we cannot know.

Use default configuration for all tests.

### Check List

Tests
- Unit test

### Release note
- No release note


Yisaer (approved):
LGTM

ti-srebot (approved):

HunDunDM (approved):
LGTM

PR #2925, opened 2020-09-08
config: introduce the dc-location config

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

### What problem does this PR solve?

Part of #2860. Introduce the `dc-location` and tso related configurations.

### What is changed and how it works?

Add a new `LocalTSOConfig` to `config` pkg.

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has configuration change

PR #2937, opened 2020-09-09
test: add a new parameter to pass the server name for config

JmPotato:

Before this PR, the `NewTestCluster` in test can only create servers with the same config. But in some cases, we need to give different servers with different configs such as `dc-location` from #2925. So I update the `ConfigOption` to make the config generator can pass the server name such as `pd1` with this function so that we can do some personalization to the configuration.

Introduce a new parameter to pass the server name for config in `ConfigOption`.

### Check List

Tests

- Unit test


tests/client/client_tls_test.go
```diff
@@ -127,7 +127,7 @@ func (s *clientTLSTestSuite) testTLSReload(
 	useIP bool) {
 	tlsInfo := cloneFunc()
 	// 1. start cluster with valid certs
-	clus, err := tests.NewTestCluster(s.ctx, 1, func(conf *config.Config) {
+	clus, err := tests.NewTestCluster(s.ctx, 1, func(conf *config.Config, _ string) {
```
HunDunDM:
```suggestion
	clus, err := tests.NewTestCluster(s.ctx, 1, func(conf *config.Config, serverName string) {
```

HunDunDM (approved):
LGTM

ti-srebot (approved):

PR #2938, opened 2020-09-09
operator: refactor the builder to prepare to support the joint consensus

HunDunDM:

This is a child PR of #2886 

For details, please refer to #2860 

* operator: refactor the builder to prepare to support the joint consensus

### Check List

Tests

- Unit test

### Release note

- No release note

### DIFF
--- a/server/schedule/operator/builder.go
+++ b/server/schedule/operator/builder.go
@@ -1,4 +1,4 @@
-// Copyright 2019 TiKV Project Authors.
+// Copyright 2020 TiKV Project Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import (
 	"github.com/tikv/pd/server/schedule/filter"
 	"github.com/tikv/pd/server/schedule/opt"
 	"github.com/tikv/pd/server/schedule/placement"
+	"github.com/tikv/pd/server/versioninfo"
 )
 
 // Builder is used to create operators. Usage:
@@ -42,36 +43,48 @@ type Builder struct {
 	rules       []*placement.Rule
 
 	// operation record
-	originPeers  peersMap
-	originLeader uint64
-	targetPeers  peersMap
-	targetLeader uint64
-	err          error
+	originPeers         peersMap
+	originLeaderStoreID uint64
+	targetPeers         peersMap
+	targetLeaderStoreID uint64
+	err                 error
 
 	// flags
-	isLightWeight bool
+	useJointConsensus bool
+	isLightWeight     bool
+	forceTargetLeader bool
 
 	// intermediate states
-	currentPeers               peersMap
-	currentLeader              uint64
-	toAdd, toRemove, toPromote peersMap       // pending tasks.
-	steps                      []OpStep       // generated steps.
-	peerAddStep                map[uint64]int // record at which step a peer is created.
+	currentPeers                         peersMap
+	currentLeaderStoreID                 uint64
+	toAdd, toRemove, toPromote, toDemote peersMap       // pending tasks.
+	steps                                []OpStep       // generated steps.
+	peerAddStep                          map[uint64]int // record at which step a peer is created.
+
+	// comparison function
+	stepPlanPreferFuncs []func(stepPlan) int // for buildStepsWithoutJointConsensus
 }
 
-// NewBuilder creates a Builder.
-func NewBuilder(desc string, cluster opt.Cluster, region *core.RegionInfo) *Builder {
-	var originPeers peersMap
+// newBuilderWithBasicCheck creates a Builder with some basic checks.
+func newBuilderWithBasicCheck(desc string, cluster opt.Cluster, region *core.RegionInfo) *Builder {
+	var err error
+	originPeers := newPeersMap()
+
 	for _, p := range region.GetPeers() {
+		if p == nil || p.GetStoreId() == 0 {
+			err = errors.Errorf("cannot build operator for region with nil peer")
+			break
+		}
 		originPeers.Set(p)
 	}
-	var err error
-	if originPeers.Get(region.GetLeader().GetStoreId()) == nil {
+
+	originLeaderStoreID := region.GetLeader().GetStoreId()
+	if _, ok := originPeers[originLeaderStoreID]; err == nil && !ok {
 		err = errors.Errorf("cannot build operator for region with no leader")
 	}
 
 	var rules []*placement.Rule
-	if cluster.GetOpts().IsPlacementRulesEnabled() {
+	if err == nil && cluster.GetOpts().IsPlacementRulesEnabled() {
 		fit := cluster.FitRegion(region)
 		for _, rf := range fit.RuleFits {
 			rules = append(rules, rf.Rule)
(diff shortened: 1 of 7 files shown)


server/schedule/operator/builder.go
```diff
@@ -163,17 +213,19 @@ func (b *Builder) SetPeers(peers map[uint64]*metapb.Peer) *Builder {
...
 		return b
 	}
-	b.targetPeers = peersMap{}
-	for k, p := range peers {
-		if p.GetStoreId() != k {
-			b.err = errors.Errorf("setPeers with mismatch storeID: %v", peers)
+
+	for key, peer := range peers {
+		if peer == nil || key == 0 || peer.GetStoreId() != key || core.IsInJointState(peer) {
+			b.err = errors.Errorf("setPeers with mismatch peers: %v", peers)
```
Yisaer, line 219: If the error is caused by `IsInJointState`, can we distinguish it from the error message here?

server/schedule/operator/builder.go
```diff
@@ -183,99 +235,190 @@ func (b *Builder) SetLightWeight() *Builder {
...
+				// voter -> learner
+				if b.useJointConsensus {
+					b.toDemote.Set(n)
+				} else {
+					b.toRemove.Set(o)
+					// Need to add `b.toAdd.Set(n)` in the later targetPeers loop
+				}
+			}
 		}
 	}
```
Yisaer, line 325: What about adding a TODO here? We can simply the code here in future.

server/schedule/operator/builder.go
```diff
@@ -183,99 +235,190 @@ func (b *Builder) SetLightWeight() *Builder {
...
+	if peer, ok := b.targetPeers[b.targetLeaderStoreID]; !ok || core.IsLearner(peer) {
+		b.targetLeaderStoreID = 0
+	}
+
+	// If no target leader is specified, try not to change the leader as much as possible.
+	if b.targetLeaderStoreID == 0 {
+		if peer, ok := b.targetPeers[b.originLeaderStoreID]; ok && !core.IsLearner(peer) {
+			b.targetLeaderStoreID = b.originLeaderStoreID
+		}
+	}
```
Yisaer, line 357: Do we need to add a check `if originLeaderStoreID == 0 ` and throw the error if it does?

server/schedule/operator/builder.go
```diff
@@ -183,99 +235,190 @@ func (b *Builder) SetLightWeight() *Builder {
 	return b
 }
 
+// SetForceTargetLeader marks the step of transferring leader to target is forcible. It is used for grant leader.
+func (b *Builder) SetForceTargetLeader() *Builder {
+	b.forceTargetLeader = true
+	return b
+}
```
Yisaer, line 242: What about `EnableForceTargetLeader`, `SetForceTargetLeader` usually receive a parameter. And could you give an example in which case we need to use this function?

server/schedule/operator/builder.go
```diff
@@ -285,74 +428,106 @@ func (b *Builder) buildSteps(kind OpKind) (OpKind, error) {
...
 	}
+
+	// store does not exist
+	if peer.GetStoreId() == b.currentLeaderStoreID {
+		return true
+	}
 	store := b.cluster.GetStore(peer.GetStoreId())
 	if store == nil {
 		return false
 	}
```
Yisaer, line 528: It seems the logic here is as same as the `hasAbilityLeader`, we can directly call this function here.

server/schedule/operator/builder.go
HunDunDM: Yes, the `peer.Role` in the `peers` of the log can be observed.

server/schedule/operator/builder.go
HunDunDM: If `hasAbilityLeader` return `true`, I don't know if it is because `peer.GetStoreId() == b.currentLeaderStoreID` or `store != nil`.

server/schedule/operator/builder.go
HunDunDM: As the comment says `It is used for grant leader`. I'll rename it.

server/schedule/operator/builder.go
HunDunDM: Before that, it would have an error in `NewBuilder`.

disksing:
Bad PR title, generally `refactor` means update naming or structure without changing behaviour. Could be: `operator: make builder support joint consensus`

server/schedule/operator/builder.go
HunDunDM: After the refactoring, `useJointConsensus` and `toDemote` are used repeatedly in the old logic. I think this is very difficult.

HunDunDM:
> Bad PR title, generally `refactor` means update naming or structure without changing behaviour. Could be: `operator: make builder support joint consensus`

But it is not yet supported in this PR.

server/schedule/operator/builder.go
HunDunDM: Look forward to

server/schedule/operator/builder.go
```diff
@@ -373,39 +550,44 @@ func (b *Builder) allowLeader(peer *metapb.Peer) bool {
 // 2. add learner + remove learner.
 // 3. add learner + promote learner + remove voter.
 // 4. promote learner.
-// 5. remove voter/learner.
-// 6. add voter/learner.
+// 5. demote voter.
```
disksing, line 553: should consider promote learner + demote voter as a step?

server/schedule/operator/builder.go
```diff
@@ -163,119 +213,213 @@ func (b *Builder) SetPeers(peers map[uint64]*metapb.Peer) *Builder {
...
+				return "", errors.New("cannot create operator: target leader is impossible")
+			}
+		} else if !b.allowLeader(targetLeader) {
+			return "", errors.New("cannot create operator: target leader is not allowed")
+		}
+	}
+
+	if len(b.toAdd)+len(b.toRemove)+len(b.toPromote)+len(b.toDemote) <= 1 {
+		// If only one peer changed, joint consensus is not used.
+		b.useJointConsensus = false
```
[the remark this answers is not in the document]
HunDunDM: This means I can use it, but I don't use it.

server/schedule/operator/builder.go
HunDunDM: Makes sense.

server/schedule/operator/builder.go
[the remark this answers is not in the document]
disksing: I'd like to propose `allowDemote` + `useJointConsensus`. The first one is determined by cluster version, and the second one is  determined by cluster version & multiple peer changes.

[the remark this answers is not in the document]
HunDunDM: Makes sense.

server/schedule/operator/builder.go
disksing: Or you can change later since it is not used anyway.

server/schedule/operator/builder.go
```diff
@@ -415,49 +597,73 @@ func (b *Builder) planReplace() stepPlan {
...
 			}
 		}
-		if next.promote != nil && b.allowLeader(next.promote) && next.promote.GetStoreId() != next.remove.GetStoreId() {
-			next.leaderRemove = next.promote.GetStoreId()
+		if next.promote != nil &&
+			next.promote.GetStoreId() != next.demote.GetStoreId() &&
+			next.promote.GetStoreId() != next.remove.GetStoreId() &&
+			b.allowLeader(next.promote) {
+			// leaderBeforeRemove does not select nodes to be demote or removed.
+			next.leaderBeforeRemove = next.promote.GetStoreId()
```
Yisaer, line 658: I' wondering for the case `// add learner + promote learner + remove voter` , will the code panic here since ` next.demote` is nil?

server/schedule/operator/builder.go
```diff
@@ -415,49 +597,73 @@ func (b *Builder) planReplace() stepPlan {
...
 			best = b.comparePlan(best, next)
 		}
-		if next.add != nil && b.allowLeader(next.add) && next.add.GetStoreId() != next.remove.GetStoreId() {
-			next.leaderRemove = next.add.GetStoreId()
+		if next.add != nil &&
+			next.add.GetStoreId() != next.demote.GetStoreId() &&
+			next.add.GetStoreId() != next.remove.GetStoreId() &&
```
Yisaer, line 663: ditto.

server/schedule/operator/builder.go
Yisaer: good catch.

server/schedule/operator/builder.go
```diff
@@ -163,119 +213,213 @@ func (b *Builder) SetPeers(peers map[uint64]*metapb.Peer) *Builder {
...
+
+	if b.targetLeaderStoreID != 0 {
+		targetLeader := b.targetPeers[b.targetLeaderStoreID]
+		if b.forceTargetLeader {
+			if !b.hasAbilityLeader(targetLeader) {
+				return "", errors.New("cannot create operator: target leader is impossible")
+			}
+		} else if !b.allowLeader(targetLeader) {
+			return "", errors.New("cannot create operator: target leader is not allowed")
+		}
```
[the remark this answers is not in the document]
Yisaer: agreed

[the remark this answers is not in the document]
HunDunDM: Other places where `allowLeader` is used should not judge `forceTargetLeader`.

server/schedule/operator/builder.go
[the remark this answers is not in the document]
HunDunDM: Will be updated in a future PR.

server/schedule/operator/builder.go
```diff
@@ -373,39 +550,44 @@ func (b *Builder) allowLeader(peer *metapb.Peer) bool {
...
 func (p stepPlan) String() string {
-	return fmt.Sprintf("stepPlan{leaderAdd=%v,add={%s},promote={%s},leaderRemove=%v,remove={%s}}",
-		p.leaderAdd, p.add, p.promote, p.leaderRemove, p.remove)
+	return fmt.Sprintf("stepPlan{leaderBeforeAdd=%v,add={%s},promote={%s},leaderBeforeRemove=%v,demote={%s},remove={%s}}",
+		p.leaderBeforeAdd, p.add, p.promote, p.leaderBeforeRemove, p.demote, p.remove)
 }
 
-func (p stepPlan) empty() bool {
-	return p.promote == nil && p.add == nil && p.remove == nil
+func (p stepPlan) Empty() bool {
```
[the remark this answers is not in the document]
HunDunDM: Will be updated in the future PR.

server/schedule/operator/builder.go
HunDunDM: `kvproto`'s `GetXXX()` allows `nil` calls.

Yisaer (commented):
Basically LGTM. I noticed that there are several improvements mentioned in the comments. What about adding a TODO to track it ?

server/schedule/operator/builder.go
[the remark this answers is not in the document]
HunDunDM: It is a good idea to unify into one function.

HunDunDM:
> Basically LGTM. I noticed that there are several improvements mentioned in the comments which is left in the future. What about adding a TODO to track it ?

Has been recorded in [Development Task Issue](https://github.com/tikv/pd/issues/2860)

Yisaer (approved):
LGTM

server/schedule/operator/builder.go
HunDunDM: and also `add voter + demote voter + remove learner`

HunDunDM: Will be updated in the future PR.

disksing (approved):

ti-srebot (approved):

server/schedule/operator/builder.go
```diff
@@ -163,119 +213,213 @@ func (b *Builder) SetPeers(peers map[uint64]*metapb.Peer) *Builder {
...
+		} else {
+			if core.IsLearner(n) {
+				// voter -> learner
+				if b.useJointConsensus {
+					b.toDemote.Set(n)
+				} else {
+					b.toRemove.Set(o)
+					// Need to add `b.toAdd.Set(n)` in the later targetPeers loop
+				}
+			}
```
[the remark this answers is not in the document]
HunDunDM: Problems left in the modification process, Will be updated in the future PR.

PR #2946, opened 2020-09-14
metrics: optimize the store state filter metrics

### LINKED ISSUE #2933: Improve the metrics of store state filter
We have some metrics to help us locate the reason why the scheduling source or target is filtered. e.g.

Screen Shot 2020-09-09 at 11 01 37 AM

There is an item named *-store-state-filter which contains many situations, but we cannot distinguish them through the monitor directly. 

The logic of the store state filter can be found in https://github.com/tikv/pd/blob/master/server/schedule/filter/filters.go#L220-L358, The metrics related logic can be found in https://github.com/tikv/pd/blob/master/server/schedule/filter/filters.go#L66-L90

rleungx:

Close #2933.

This PR records the detailed reason in the filter type. After this PR:
Screen Shot 2020-09-14 at 2 25 07 PM

### Check List

Tests

- Manual test


server/schedule/filter/filters.go
```diff
@@ -227,67 +227,80 @@ type StoreStateFilter struct {
 	MoveRegion bool
 	// Set true if allows temporary states.
 	AllowTemporaryStates bool
+	// Typ is used to distinguish the reason of store state filter
+	Typ string
```
disksing, line 231: It is too tricky to use a field here, and the name is not accurate.

server/schedule/filter/filters.go
```diff
@@ -227,67 +227,80 @@ type StoreStateFilter struct {
...
+func (f *StoreStateFilter) Type() string {
+	return f.Typ
 }
 
 // conditionFunc defines condition to determine a store should be selected.
 // It should consider if the filter allows temporary states.
 type conditionFunc func(*config.PersistOptions, *core.StoreInfo) bool
 
-func (f StoreStateFilter) isTombstone(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) isTombstone(opt *config.PersistOptions, store *core.StoreInfo) bool {
```
disksing, line 248: you can change it to `isTombstone(opt *config.PersistOptions, store *core.StoreInfo) (reason string, ok bool)`

rleungx: The reason I use a pointer receiver here is to keep consistent with other filters.

server/schedule/filter/filters.go
```diff
@@ -228,67 +228,80 @@ type StoreStateFilter struct {
...
 
-func (f StoreStateFilter) exceedRemoveLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) exceedRemoveLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+	f.Reason = "remove-limit"
 	return !f.AllowTemporaryStates && !store.IsAvailable(storelimit.RemovePeer)
 }
 
-func (f StoreStateFilter) exceedAddLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) exceedAddLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+	f.Reason = "add-limit"
```
Yisaer, line 285: what about `exceed-limit` here. it would be mor clear.

server/schedule/filter/filters.go
```diff
@@ -228,67 +228,80 @@ type StoreStateFilter struct {
 	MoveRegion bool
 	// Set true if allows temporary states.
 	AllowTemporaryStates bool
+	// Reason is used to distinguish the reason of store state filter
```
Yisaer, line 231:
```suggestion
	// Reason is used to point out the reason of store state filter
```

server/schedule/filter/filters.go
```diff
@@ -228,67 +228,80 @@ type StoreStateFilter struct {
...
 
-func (f StoreStateFilter) exceedAddLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) exceedAddLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
+	f.Reason = "add-limit"
 	return !f.AllowTemporaryStates && !store.IsAvailable(storelimit.AddPeer)
 }
 
-func (f StoreStateFilter) tooManySnapshots(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) tooManySnapshots(opt *config.PersistOptions, store *core.StoreInfo) bool {
+	f.Reason = "snapshot"
```
Yisaer, line 290: what about `excess-snapshot`

server/schedule/filter/filters.go
```diff
@@ -228,67 +228,80 @@ type StoreStateFilter struct {
...
 	return !f.AllowTemporaryStates && (uint64(store.GetSendingSnapCount()) > opt.GetMaxSnapshotCount() ||
 		uint64(store.GetReceivingSnapCount()) > opt.GetMaxSnapshotCount() ||
 		uint64(store.GetApplyingSnapCount()) > opt.GetMaxSnapshotCount())
 }
 
-func (f StoreStateFilter) tooManyPendingPeers(opt *config.PersistOptions, store *core.StoreInfo) bool {
+func (f *StoreStateFilter) tooManyPendingPeers(opt *config.PersistOptions, store *core.StoreInfo) bool {
+	f.Reason = "pending-peer"
```
Yisaer, line 297: what about `pending-peer`

Yisaer (commented):
The rest LGTM

server/schedule/filter/filters.go
rleungx: What is the difference?

server/schedule/filter/filters.go
rleungx: Do you mean exceed-add-limit?

disksing (approved):

nolouch:
LGTM

ti-srebot (approved):

Yisaer (approved):
LGTM

PR #2951, opened 2020-09-14
simulator: Visualize new region distribution in import data case

ZenoTan:

Add visualization in import data case.

Use go e-charts to plot 3d html.

### Check List

Tests

- Manual test


ZenoTan:
image
This is what the result is like. For now, new regions are not balanced.

tools/pd-simulator/simulator/cases/import_data.go
```diff
@@ -140,6 +143,37 @@ func newImportData() *Case {
...
+			xAxis := make([]int, 10)
+			for i := 1; i <= 10; i++ {
+				xAxis[i-1] = i
+			}
+			yAxis := make([]int, checkCount)
+			for i := 1; i <= int(checkCount); i++ {
+				yAxis[i-1] = i
+			}
+			bar3d.AddXYAxis(xAxis, yAxis).AddZAxis("bar3d", newRegionCount)
+			f, _ := os.Create("region_3d.html")
```
nolouch, line 171: can we show the total region in another file?

ZenoTan: Now supported.
image

nolouch (approved):
cool, LGTM.

tools/pd-simulator/simulator/cases/import_data.go
rleungx: Can we change the label of each axis?

ZenoTan: Their API now can only add data, no `set title` interface...

lhy1024 (approved):

ti-srebot (approved):

Issue #2759, opened 2020-08-13
Local/Global Transaction Txn Allocator in Cross-DC Deployment Roadmap


JmPotato:
RFC: [Local/Global Transaction in Cross-DC Deployment](https://docs.google.com/document/d/1t4D2M5W0dqduNTey1aTute4umuEQrKazU-ipSwFAaLM/edit#heading=h.efxhx09hxyyo) (In Chinese)
Issue: pingcap/tidb#18253

## Refactoring Phase

- [x] Wrap the election mechanism to improve its reusability to introduce the election of TSO Allocator more easily. (#2767 + #2831 + #2883)
- [x] Modify the `tso` package to be more extendable, maybe using `interface`. (#2803 + #2842 + #2859)

### New Configuration

- [x] Introduce the new DC location configuration. (#2925)
- [x] Write configuration into etcd for persistence. (#2939)
- [ ] Implement the ctl command and API to create and update the configuration.

### Allocator Manager

- [x] Introduce the new Allocator Manager to concurrently manage multi allocations in one PC server. (#2859)
- [x] Implement the basic election of the TSO Allocator. (#2894 + #2992)
- [x] Implement the election priority check. (#2953)
- [x] Implement the differentiation mechanism by using the low bits of TSO's logical part to distinguish the same TSO of different DCs. (#3267 + #3290)

### Local TSO Allocator

- [x] Implement the internal logic of the Local TSO Allocator. (#2986)

### Global TSO Allocator

- [x] Implement the new RPC methods used to synchronize TSO between PD servers. (#2988)
- [x] Be compatible with the old Global TSO Allocator. (#3033)
- [x] Implement the internal logic of the Global TSO Allocator. (#3033)
- [x] Implement the unit tests. (#3033)

### PD Client

- [x] Implement the new Local/Global TSO acquisition mechanism for PD Client. (#3045)

## Test Phase

- [x] Have an instance-level verification with TiDB.

## Notice

The goal of the refactoring phase is doing some preparation work to make the implementation phase can go smoothly and easily. Some of these ideas are not mature yet. So this roadmap's content above may change at any time during the process of development.

lhy1024:
@JmPotato Can we close it?

JmPotato:
> @JmPotato Can we close it?

It has not been released yet, I think we can keep it open.

okJiang:
@JmPotato How about closing it now?👀

JmPotato:
Close due to the stall of this project.

PR #2953, opened 2020-09-14
tso: add election priority for Local TSO Allocator

### LINKED ISSUE #2943: Provide a way to implement the Local TSO Allocator leader election priority

Part of #2759. For more details about TSO Allocator, please look at this [document](https://docs.google.com/document/d/1t4D2M5W0dqduNTey1aTute4umuEQrKazU-ipSwFAaLM/edit#heading=h.qolqioffe8ac) (In Chinese).

For the PD leader election priority, we can do this by changing the etcd leader to make the PD leader been changed because we have a constrain between the PD leader and etcd leader. But for the Local TSO Allocator leader election priority, we don't have such an easy way to move a Local TSO Allocartor leader to another.

We need to provide a new way to guarantee the following rules:

1. For example, the PD server with `dc-location="dc-1"` needs to be elected as the allocator leader with the highest priority.
2. If all PD servers with `dc-location="dc-1"` are down, then the other PD servers of DC could be elected.

We can implement this feature by introducing an extra procedure that will write some keys and values into etcd to make the election priority been guaranteed.

For Local TSO Allocator follower:

1. Each `AllocatorManager` keeps checking if its priority is higher than the current Local TSO Allocator leader by reading the leader key from etcd.
2. If it's true and `AllocatorManager` have an uninitialized Local TSO Allocator, it means this allocator is able to become the DC's Local TSO Allocator with a higher priority. Then it will write into etcd with a key-value pair like `dc-location/dc-1/next-leader: member-id`.
3. While campaigning a Local TSO Allocator leader, the `AllocatorManager` will check the corresponding `dc-location/dc-1/next-leader: member-id` key-value pair. If the member ID is not itself, it will not continue trying to campaign the leader.

For Local TSO Allocator leader:

1. While it keeps maintaining its leadership, it will also check the `dc-location/dc-1/next-leader`, then it will find that the value is not its member ID, which will make it step down and become a follower.

JmPotato:

Part of #2759. Resolve #2943. You can check the detailed design in this issue.

By introducing the `next-leader` key in etcd to implement the Local TSO Allocator leader election priority.

### Check List

Tests

- Unit test


server/tso/allocator_manager.go
```diff
@@ -387,3 +406,116 @@ func (am *AllocatorManager) GetAllocators(filters ...AllocatorGroupFilter) []All
...
+		if allocatorGroup.dcLocation != leaderServerDCLocation && allocatorGroup.dcLocation == myServerDCLocation {
+			log.Info("try to move the local tso allocator",
+				zap.Uint64("old-leader-id", leaderServerID),
+				zap.String("old-dc-location", leaderServerDCLocation),
+				zap.Uint64("next-leader-id", serverID),
+				zap.String("next-dc-location", myServerDCLocation))
+			nextLeaderKey := path.Join(am.rootPath, allocatorGroup.dcLocation, "next-leader")
+			resp, err := kv.NewSlowLogTxn(am.member.Client()).
+				If(clientv3.Compare(clientv3.CreateRevision(nextLeaderKey), "=", 0)).
+				Then(clientv3.OpPut(nextLeaderKey, fmt.Sprint(serverID))).
```
JmPotato, line 454: Need to introduce the lease when we write `next-leader` into etcd, otherwise if the next leader is done, we will never have a new Local TSo Allocator anymore. This change is a low priority, I will update it later.

server/tso/allocator_manager.go
```diff
@@ -231,6 +231,21 @@ func (am *AllocatorManager) allocatorLeaderLoop(parentCtx context.Context, alloc
 			log.Info("local tso allocator leader has changed, try to re-campaign a local tso allocator leader",
 				zap.String("dc-location", allocator.dcLocation))
 		}
+		// Check the next-leader key
+		nextLeader, err := am.getNextLeaderID(allocator.dcLocation)
+		if err != nil {
+			log.Error("get next leader from etcd failed",
+				zap.String("dc-location", allocator.dcLocation),
+				errs.ZapError(err))
```
disksing, line 239: continue?

server/tso/allocator_manager.go
```diff
@@ -387,3 +406,116 @@ func (am *AllocatorManager) GetAllocators(filters ...AllocatorGroupFilter) []All
 	}
 	return allocators
 }
+
+// checkPriority is used to check the election priority of a Local TSO Allocator.
+// In the normal case, if we want to elect a Local TSO Allocator for a certain DC,
+// such as dc-1, we need to make sure the follow priority rules:
+// 1. The PD server with dc-location="dc-1" needs to be elected as the allocator
+// leader with the highest priority.
+// 2. If all PD servers with dc-location="dc-1" are done, then the other PD servers
```
disksing, line 415: done -> down

server/tso/allocator_manager.go
```diff
@@ -304,6 +321,8 @@ func (am *AllocatorManager) AllocatorDaemon(serverCtx context.Context) {
 				go am.updateAllocator(ag)
 			}
 			am.wg.Wait()
+			// The Local TSO Allocator priority checker
+			am.checkPriority()
```
disksing, line 325: It should run with a slower interval to prevent reading etcd too frequently.

disksing: for example, per minute.

disksing (approved):

server/tso/allocator_manager.go
nolouch: Better to create an issue.

nolouch (commented):
LGTM

ti-srebot (approved):

server/tso/allocator_manager.go
JmPotato: Fixed.

PR #2959, opened 2020-09-15
pdctl: fix the issue that flags get lost in interactive mode

disksing:

Fix https://github.com/tikv/pd/issues/2899

The flags used when entering interactive mode need to be retained in subsequent commands.

The previous approach was that each time a command was run, the flags would be restored to the default values, and then the previous flags will be appended to the end of the input. When the command is executed, the restored default values are reset to the previous values again.

The bug is that if the input command is wrong and the execution fails, the flags will be restored to the default value.

The fix of this PR is to directly set flags to the previous value instead of the default value each time the command is run.

### Check List

Tests

Currently, we cannot test pd-ctl in interactive mode, so I only test it manually.

Prepare: start pd at 127.0.0.1:2381

1. wrong command won't clear previous flags
> pd-ctl -i -u "127.0.0.1:2381"
> cluster # ok
> region -x # error
> cluster # ok (without the PR, it reports cannot connect to 127.0.0.1:2379)

2. reset flags
> pd-ctl -i
> cluster # cannot connect 127.0.0.1:2379
> cluster -u "127.0.0.1:2381" # ok
> cluster # ok

### Release note
- No release note


rleungx (approved):
LGTM

nolouch (approved):
LGTM

ti-srebot (approved):

HunDunDM (approved):
LGTM

PR #2966, opened 2020-09-16
scheduler: add leader verify in balance-region scheduler

Yisaer:

If the region have no leader during balance-region, it will pause PD panic and restart over and over.

Check region leader before balance region. If it has no leader, the balance for this region would be skipped.

### Check List

Tests

- Unit test

Related changes

### Release note

- Fix the bug that pd might panic If some regions have no leader when `balance-region` enabled.


server/schedule/filter/filters.go
```diff
@@ -683,10 +683,13 @@ func (f *isolationFilter) Target(opt *config.PersistOptions, store *core.StoreIn
...
-		Id:      leader.Id,
-		StoreId: leader.StoreId,
-		Role:    leader.Role,
+	copyLeader := &metapb.Peer{}
+	if leader != nil {
+		copyLeader = &metapb.Peer{
+			Id:      leader.Id,
+			StoreId: leader.StoreId,
+			Role:    leader.Role,
+		}
```
disksing, line 692: how about ` copyLeader := proto.Clone(leader).(*metapb.Peer)`

Yisaer: good idea. updated.

server/schedulers/balance_region.go
```diff
@@ -173,6 +173,11 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 				schedulerCounter.WithLabelValues(s.GetName(), "region-hot").Inc()
 				continue
 			}
+			// Check region whether have leader
+			if region.GetLeader() == nil {
```
rleungx, line 177: Please also add metrics for it

pkg/pointer/point.go
```diff
@@ -0,0 +1,18 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pointer
```
howardlau1999, line 14: should the filename be `pointer.go`?

server/schedulers/balance_region.go
Yisaer: updated.

pkg/pointer/point.go
Yisaer: good catch,updated.

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -245,6 +245,11 @@ func (mc *Cluster) AddRegionStore(storeID uint64, regionCount int) {
 	mc.PutStore(store)
 }
 
+// AddRegionInfo supports add regionInfo directly
+func (mc *Cluster) AddRegionInfo(regionInfo *core.RegionInfo) {
```
HunDunDM, line 249: Why not use `PutRegion` directly?

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -535,7 +540,7 @@ func (mc *Cluster) UpdateStoreStatus(id uint64) {
 }
 
 func (mc *Cluster) newMockRegionInfo(regionID uint64, leaderStoreID uint64, followerStoreIDs ...uint64) *core.RegionInfo {
-	return mc.MockRegionInfo(regionID, leaderStoreID, followerStoreIDs, []uint64{}, nil)
+	return mc.MockRegionInfo(regionID, &leaderStoreID, followerStoreIDs, []uint64{}, nil)
```
HunDunDM, line 543: When does `nil` need to be passed in here? In addition, `0` can also mean that store does not exist.

pkg/pointer/pointer.go
```diff
@@ -0,0 +1,19 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pointer
```
HunDunDM, line 14: Maybe it can be put in `typeutil` package?

HunDunDM (commented):
rest LGTM

pkg/mock/mockcluster/mockcluster.go
disksing: same question

pkg/mock/mockcluster/mockcluster.go
disksing: I think check 0 is better.

pkg/pointer/pointer.go
rleungx: Maybe we don't need this function

disksing (approved):

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -583,8 +584,12 @@ func (mc *Cluster) MockRegionInfo(regionID uint64, leaderStoreID uint64,
 		EndKey:      []byte(fmt.Sprintf("%20d", regionID+1)),
 		RegionEpoch: epoch,
 	}
-	leader, _ := mc.AllocPeer(leaderStoreID)
-	region.Peers = []*metapb.Peer{leader}
+	region.Peers = []*metapb.Peer{}
```
HunDunDM, line 587:
```suggestion
```

HunDunDM (commented):
rest LGTM

server/schedulers/balance_region.go
```diff
@@ -173,6 +173,12 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
...
+				log.Debug("region have no leader", zap.String("scheduler", s.GetName()), zap.Uint64("region-id", region.GetID()))
```
lhy1024, line 178: it may be `warn`?

lhy1024 (commented):
the rest LGTM

pkg/mock/mockcluster/mockcluster.go
Yisaer: updated.

server/schedulers/balance_region.go
Yisaer: updated.

lhy1024 (approved):

ti-srebot (approved):

PR #2968, opened 2020-09-16
add nil check in GetLeader and GetFollower

lhy1024:

we usually check nil when we directly use `leader[storeID]` , but we did not in some function.

I think we should add it to avoid forgetting it although they are used in test now.

### Check List

Tests

- Unit test


HunDunDM (approved):
LGTM

nolouch (commented):
LGTM

ti-srebot (approved):

nolouch:
Does need cherry-pick to 4.0?

lhy1024:
> Does need cherry-pick to 4.0?

No, it only is used in test.

PR #2969, opened 2020-09-16
misc: unify route path

lhy1024:

schedulerConfigHandler use path from other pkg rather than a string directly, this PR will unify.

### Check List

Tests

- Unit test


disksing (approved):

HunDunDM (commented):
I think `server.SchedulerConfigHandlerPath = "/api/v1/scheduler-config"`, `api.schedulerConfigPrefix = "pd/api/v1/scheduler-config"` need to unify too.

HunDunDM (approved):

ti-srebot (approved):

Issue #2576, opened 2020-06-24
Consider a better way to get the max-alloc-id when recovering PD cluster


nolouch:
### Describe your feature request related problem

If the` max-alloc-id` is set incorrectly when recovering PD cluster, it will cause duplicate region id and cause the cluster to be unavailable. We currently use monitor and log to find it, but it is possible that people forget to set or the ID found in the log is incorrect（may log loss). We need a safer way to get the id.

### Describe alternatives you've considered

consider let recover tool do not need to set the id or a tool can more easy and safe to get the id.
- consider scanning all tikvs to get id
- consider collecting all regions' heartbeat then turn on `allocer`
and so on../
### Teachability, Documentation, Adoption, Migration Strategy
- Go

PR #2978, opened 2020-09-17
Get the max-alloc-id by polling all regions when recovering PD cluster

longfangsong:
Signed-off-by: longfangsong <longfangsong@icloud.com>

#2576

If `max-alloc-id` is not set manually during recovering PD cluster, poll all regions and find the max one, then set `allocID = theMaxOne + 1` .

### Check List

Tests

Side effects

- Increased code complexity

### Release note

PD Recover can decide `max-alloc-id` automatically now.

### DIFF
--- a/tools/pd-recover/main.go
+++ b/tools/pd-recover/main.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"strconv"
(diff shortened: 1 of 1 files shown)


tools/pd-recover/main.go
```diff
@@ -65,17 +68,40 @@ func main() {
...
 
 	rootPath := path.Join(pdRootPath, strconv.FormatUint(clusterID, 10))
 	clusterRootPath := path.Join(rootPath, "raft")
 	raftBootstrapTimeKey := path.Join(clusterRootPath, "status", "raft_bootstrap_time")
 
 	urls := strings.Split(endpoints, ",")
 
+	if allocID == 0 {
+		for _, url := range urls {
+			regionsURL := strings.TrimRight(url, "/") + "/pd/api/v1/regions"
```
rleungx, line 80: Before we use pd-recover, the data of all PD will be cleaned.

PR #2982, opened 2020-09-17
[WIP] schedule: implement multiple dimensional hotspot scheduler

jiyingtk:

It can balance both CPU and IO usages of nodes.

Overview:
1. Load based region split, which can divide regions into arbitrary proportions based on load
    * Physically divides regions, extends the "SplitRegion" operator to tell TiKV to execute the load based split command
2. Heuristic scheduling algorithm, the scheduling decision is made according to the CPU (qps information) and IO (bytes rate information) loads, the whole scheduling process is divided into four stages:
    * Select the regions that may need to be divided
    * Execute the split operator and obtain the flow data of the new divided regions
    * Combining the division information of regions, and then using a heuristic strategy to generate the migration scheme
    * Execute the migration operator and wait for the schedule to complete

TODO:
    1. Logically divides regions, collect load distribution of region from TiKV
    2. Decide when to run scheduling algorithm
    3. The algorithm is static, while the influence of system dynamic characteristics should also be taken into account

### Check List

Tests

running sysbench `select_random_points`: 
image
image

Side effects

### Release note
- Support multiple dimensional hotspot scheduler


rleungx:
@jiyingtk Would you like to draft RFC?

jiyingtk:
> @jiyingtk Would you like to draft RFC?

This PR hasn't been well tested, some steps of the algorithm may need to be revalidated or redesigned, I'll do some tests as soon as possible.

rleungx:
> redesigned

Ok, feel free to ask for help if you have any questions.

jiyingtk:
> > redesigned
> 
> Ok, feel free to ask for help if you have any questions.

OK, thx

JmPotato:

Issue #2984, opened 2020-09-17
Implement the global TSO generation algorithm


JmPotato:
### Describe your feature request related problem

As you can see from the [design document](https://docs.google.com/document/d/1t4D2M5W0dqduNTey1aTute4umuEQrKazU-ipSwFAaLM/edit#heading=h.w99w3m5hht19) of Global TSO Allocator (In Chinese), we need to implement the synchronization algorithm between the Global TSO Allocator and the Local TSO Allocator leaders. Because this change is big, so I open this issue to record the different development parts of it.

### Describe the feature you'd like

Implement the global TSO generation algorithm described in [design document](https://docs.google.com/document/d/1t4D2M5W0dqduNTey1aTute4umuEQrKazU-ipSwFAaLM/edit#heading=h.w99w3m5hht19) of Global TSO Allocator (In Chinese).

PR #2988, opened 2020-09-18
tso: add SyncMaxTS RPC method

JmPotato:

Part of #2984 and #2759. Prepare for the next step of implementing the global TSO generation algorithm.

Add `SyncMaxTS` RPC method.

### Check List

Tests

- Unit test
- Integration test

Related changes

- PR to update pingcap/kvproto#678


server/grpc_service.go
```diff
@@ -908,3 +908,94 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+			return nil, err
+		}
+	}
+	return &pdpb.WriteMaxTSResponse{
+		Header:  s.header(),
+		Written: true,
+	}, nil
+}
+
+// validateInternalRequest checks if server is closed, which is used to valide
```
Yisaer, line 994:
```suggestion
// validateInternalRequest checks if server is closed, which is used to validate
```

server/tso/allocator_manager.go
```diff
@@ -336,6 +336,9 @@ func (am *AllocatorManager) updateAllocator(ag *allocatorGroup) {
 func (am *AllocatorManager) HandleTSORequest(dcLocation string, count uint32) (pdpb.Timestamp, error) {
 	am.RLock()
 	defer am.RUnlock()
+	if len(dcLocation) == 0 {
+		dcLocation = config.GlobalDCLocation
+	}
```
Yisaer, line 341: Why set dcLocation as global instead of throwing error here?

Yisaer (commented):
Basically LGTM. I'm wondering maybe we also need to add the sender information like member ID in request. And each receiver should compare the sender member ID and the leader id in their vision.

server/tso/allocator_manager.go
JmPotato: Considering the forward compatibility, I add this check in order to handle the old TSO request that may not have this field.

JmPotato:
> leader

Good idea. I will update this part later.

Yisaer (dismissed):
LGTM

server/grpc_service.go
```diff
@@ -908,3 +908,98 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+		currentLocalTSO, err := allocator.GetCurrentTSO()
+		if err != nil {
+			return nil, err
+		}
+		// Validate whether the MaxTs is greater than all other local TSOs
+		if currentLocalTSO.Physical > request.MaxTs.Physical {
+			return &pdpb.PrewriteMaxTSResponse{
+				Header:     s.header(),
+				Prewritten: false,
+				MaxLocalTs: &currentLocalTSO,
```
disksing, line 935: It may not be the max. I think we need to iterate all allocators to find the max.

JmPotato: This RPC method just implements the responding behavior of a Local TSO Allocator leader when it receives a prewriting request. After the Global TSO Allocator collects all responses, it will choose the biggest one as the max, which is not included in this PR.

JmPotato: Maybe the problem is about the naming? We should not call it `MaxLocalTs` in this phase technically.

disksing: I know. But it still needs to iterate all allocators, or the global allocator may not able to find the global max of all allocators.

JmPotato: I got it.

server/grpc_service.go
```diff
@@ -908,3 +908,106 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+	if err := s.validateInternalRequest(request.GetHeader()); err != nil {
+		return nil, err
+	}
+	tsoAllocatorManager := s.GetTSOAllocatorManager()
+	// Get all Local TSO Allocator leaders
+	allocatorLeaders, err := tsoAllocatorManager.GetLocalAllocatorLeaders()
+	if err != nil {
+		return nil, err
+	}
+	// Iterate all Local TSO Allocator leaders to validate whether the prewritten TSO is consistent
```
disksing, line 968: Why do we need the prewrite tso? seems we can use the ts in `request` directly?

JmPotato: What if it's a request that arrives very late that the prewritten TSO is no longer valid during an abortive transaction?

JmPotato: The prewritten TSO is a bridge connecting the previous Prewrite request and the later Write request. Otherwise the requests will be separated, which make a Local TSO Allocator leader not know whether it's a correct request to write the corresponding prewritten TSO.

disksing:
update PR title?

server/grpc_service.go
```diff
@@ -908,3 +908,80 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
 		Message: msg,
 	})
 }
+
+// PrewriteMaxTS is the first phase of 2PC for global MaxTS synchronization,
```
disksing, line 912: Seems `prewrite` is not a suitable name now.

server/grpc_service.go
```diff
@@ -908,3 +908,62 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
 		Message: msg,
 	})
 }
+
+// SyncMaxTS is a RPC method used to synchronize the timestamp of TSO between a
+// Global TSO Allocator and multi Local TSO Allocator leaders. It contains two
```
disksing, line 913: __the__ global tso allocator

server/grpc_service.go
```diff
@@ -908,3 +908,62 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+		}, nil
+	}
+	// The second phase of synchronization: do the writing
+	for _, allocator := range allocatorLeaders {
+		if err := allocator.WriteTSO(request.GetMaxTs()); err != nil {
+			return nil, err
+		}
+	}
+	return &pdpb.SyncMaxTSResponse{
+		Header: s.header(),
```
disksing, line 954: need to report handled DCs in case local allocator leader transfer.

server/grpc_service.go
```diff
@@ -908,3 +908,81 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+	if err != nil {
+		return nil, err
+	}
+	processedDCs := make([]string, 0)
+	if request.GetMaxTs() == nil || request.GetMaxTs().Physical == 0 {
+		// The first phase of synchronization: collect the max local ts
+		var maxLocalTS pdpb.Timestamp
+		for _, allocator := range allocatorLeaders {
+			if !allocator.IsStillAllocatorLeader() {
+				return nil, errs.ErrNotLocalAllocatorLeader.FastGenByArgs(
```
disksing, line 936: can just skip it here because global allocator will check if all DCs are handled.

server/grpc_service.go
```diff
@@ -908,3 +908,81 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+	if err := s.validateInternalRequest(request.GetHeader()); err != nil {
+		return nil, err
+	}
+	tsoAllocatorManager := s.GetTSOAllocatorManager()
+	// Get all Local TSO Allocator leaders
+	allocatorLeaders, err := tsoAllocatorManager.GetLocalAllocatorLeaders()
+	if err != nil {
+		return nil, err
+	}
+	processedDCs := make([]string, 0)
```
disksing, line 930: var processedDCs []string

server/grpc_service.go
```diff
@@ -908,3 +908,81 @@ func (s *Server) incompatibleVersion(tag string) *pdpb.ResponseHeader {
...
+		}
+		return &pdpb.SyncMaxTSResponse{
+			Header:     s.header(),
+			MaxLocalTs: &maxLocalTS,
+			Dcs:        processedDCs,
+		}, nil
+	}
+	// The second phase of synchronization: do the writing
+	for _, allocator := range allocatorLeaders {
+		if !allocator.IsStillAllocatorLeader() {
```
disksing, line 959: can skip too.

disksing (approved):

ti-srebot (approved):

PR #2992, opened 2020-09-18
tso: add the allocator checker

JmPotato:

Part of #2984 and #2759. Prepare for the next step of implementing the global TSO generation algorithm.

After merging this pr, the `AllocatorDaemon` will check the `dc-location` config in the cluster periodically. If any new new `dc-location` found, it will set up the corresponding Local TSO Allocator.

### Check List

Tests

- Unit test

### DIFF
--- a/server/tso/allocator_manager.go
+++ b/server/tso/allocator_manager.go
@@ -47,7 +47,10 @@ type AllocatorGroupFilter func(ag *allocatorGroup) bool
 
 type allocatorGroup struct {
 	dcLocation string
-	serverCtx  context.Context
+	// Because an allocator may be set up with different context,
+	// we need to store the parent context for each allocator in
+	// order to receive the Done() signal correctly.
+	parentCtx context.Context
 	// For the Global TSO Allocator, leadership is a PD leader's
 	// leadership, and for the Local TSO Allocator, leadership
 	// is a DC-level certificate to allow an allocator to generate
@@ -158,7 +161,7 @@ func (am *AllocatorManager) getLocalTSOConfigPath() string {
 }
 
 // SetUpAllocator is used to set up an allocator, which will initialize the allocator and put it into allocator daemon.
-func (am *AllocatorManager) SetUpAllocator(serverCtx context.Context, dcLocation string, leadership *election.Leadership) error {
+func (am *AllocatorManager) SetUpAllocator(parentCtx context.Context, dcLocation string, leadership *election.Leadership) error {
 	am.Lock()
 	defer am.Unlock()
 	var allocator Allocator
@@ -170,7 +173,7 @@ func (am *AllocatorManager) SetUpAllocator(serverCtx context.Context, dcLocation
 	// Update or create a new allocatorGroup
 	am.allocatorGroups[dcLocation] = &allocatorGroup{
 		dcLocation: dcLocation,
-		serverCtx:  serverCtx,
+		parentCtx:  parentCtx,
 		leadership: leadership,
 		allocator:  allocator,
 	}
@@ -187,7 +190,7 @@ func (am *AllocatorManager) SetUpAllocator(serverCtx context.Context, dcLocation
 	default:
 		// Join in a Local TSO Allocator election
 		localTSOAllocator, _ := allocator.(*LocalTSOAllocator)
-		go am.allocatorLeaderLoop(serverCtx, localTSOAllocator)
+		go am.allocatorLeaderLoop(parentCtx, localTSOAllocator)
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


server/tso/allocator_manager.go
```diff
@@ -288,28 +289,62 @@ func (am *AllocatorManager) campaignAllocatorLeader(parentCtx context.Context, a
...
+			// Check if we have any new dc-location configured, if yes,
+			// then set up the corresponding local allocator.
+			am.allocatorChecker(serverCtx, serverCtxCancel)
 		case <-serverCtx.Done():
 			return
 		}
 	}
 }
 
+func (am *AllocatorManager) allocatorChecker(ctx context.Context, ctxCancel context.CancelFunc) {
```
Yisaer, line 315: What about `patrolAllocatorGroups`. And What if the `dc-location` is unfounded while its allocator is still running. For example, users want to remove a `dc` zone cluster.

server/tso/allocator_manager.go
```diff
@@ -288,28 +289,62 @@ func (am *AllocatorManager) campaignAllocatorLeader(parentCtx context.Context, a
...
-				am.wg.Add(1)
-				go am.updateAllocator(ag)
-			}
-			am.wg.Wait()
+			// Update the Local TSO Allocator leaders TSO in memory concurrently.
+			am.allocatorUpdater()
+		case <-checkerTicker.C:
+			// Check if we have any new dc-location configured, if yes,
+			// then set up the corresponding local allocator.
+			am.allocatorChecker(serverCtx, serverCtxCancel)
```
Yisaer, line 308: I suggest also adding the comments for these 2 function.

server/tso/allocator_manager.go
```diff
@@ -288,28 +289,62 @@ func (am *AllocatorManager) campaignAllocatorLeader(parentCtx context.Context, a
...
+func (am *AllocatorManager) allocatorUpdater() {
+	// Filter out allocators without leadership and uninitialized
+	allocatorGroups := am.getAllocatorGroups(FilterUninitialized(), FilterUnavailableLeadership())
+	// Update each allocator concurrently
+	for _, ag := range allocatorGroups {
+		am.wg.Add(1)
+		go am.updateAllocator(ag)
+	}
+	am.wg.Wait()
+}
```
Yisaer, line 346: I'm wondering why `allocatorUpdater` doesn't introduce the `ctx` and `cancelFunc`.

Yisaer (commented):
The rest LGTM

server/tso/allocator_manager.go
JmPotato: I found the context usage here is so disorganized...I need to sort it out.

JmPotato: Thanks for pointing out!

JmPotato: Updated. `allocatorUpdater` doesn't need the context because an `allocatorGroup` already has one. `allocatorPatroller` needs context because it needs to set up an allocator which will be added into its `allocatorGroup`.

server/tso/allocator_manager.go
```diff
@@ -144,6 +139,9 @@ func (am *AllocatorManager) GetClusterDCLocations() (map[string][]uint64, error)
 		// The key will contain the member ID and the value is its dcLocation
 		serverPath := strings.Split(string(kv.Key), "/")
 		dcLocation := string(kv.Value)
+		if len(strings.TrimSpace(dcLocation)) == 0 {
```
disksing, line 142: why need trim space?

JmPotato: I'm not sure whether we could possibly get an empty value here from etcd, so I add this to make sure the value is valid as a `dc-location` string. If it's my overthinking, I will delete it though.

server/tso/allocator_manager.go
```diff
@@ -327,11 +331,60 @@ func (am *AllocatorManager) updateAllocator(ag *allocatorGroup) {
...
+		dcLocations = append(dcLocations, dcLocation)
+	}
+	// Get all Local TSO Allocators
+	allocatorGroups := am.getAllocatorGroups(FilterDCLocation(config.GlobalDCLocation))
+	// Set up the new one
+	for _, dcLocation := range dcLocations {
+		if slice.NoneOf(allocatorGroups, func(i int) bool {
+			return allocatorGroups[i].dcLocation == dcLocation
+		}) {
+			if err := am.SetUpAllocator(serverCtx, dcLocation, election.NewLeadership(
```
disksing, line 358: Can we gurantee that ts of the newly created allocator be greater than the last allocated global ts?

JmPotato: Yes. When we synchronize the global ts and write it into each local allocator leader, we will also update the time window in etcd. So if the newly created allocator becomes a leader, it will be greater than the last allocated global ts.

disksing: cool.

disksing (approved):

server/tso/allocator_manager.go
```diff
@@ -36,23 +36,18 @@ import (
...
 type allocatorGroup struct {
 	dcLocation string
-	// allocator's parent ctx and cancel function, which is to
-	// control the allocator's behavior in AllocatorDaemon and
-	// pass the cancel signal to its parent as soon as possible
-	// since it is critical to let parent goroutine know whether
-	// the allocator is still able to work well.
-	parentCtx    context.Context
-	parentCancel context.CancelFunc
+	serverCtx  context.Context
```
Yisaer, line 50: It seems `allocatorGroup` and `allocatorManager` holding the same serverCtx passed from the `Server`, so I think it is not necessary for `allocatorGroup` to hold `serverCtx`. And `allocatorUpdater` should pass the `serverCtx` to the `allocatorGroup` in that function.

Yisaer (commented):
Rest LGTM

Yisaer (approved):
LGTM

ti-srebot (approved):

PR #2993, opened 2020-09-18
operator: add additional info for operator

lhy1024:

When we check the log about create and finish operator, we don't know info about it. Especially, leader score and region score.

### Check List

Tests

- Unit test

### Release note

- add additional info for operator

### DIFF
--- a/server/schedulers/balance_leader.go
+++ b/server/schedulers/balance_leader.go
@@ -274,7 +274,8 @@ func (l *balanceLeaderScheduler) createOperator(cluster opt.Cluster, region *cor
 
 	opInfluence := l.opController.GetOpInfluence(cluster)
 	kind := core.NewScheduleKind(core.LeaderKind, cluster.GetOpts().GetLeaderSchedulePolicy())
-	if !shouldBalance(cluster, source, target, region, kind, opInfluence, l.GetName()) {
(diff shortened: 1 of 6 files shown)


rleungx (commented):
Will it be recorded through the log? Can you show the output example?

lhy1024:
> Will it be recorded through the log? Can you show the output example?

image

yep, but there are some operator without additional info, I am not sure whether its format is suitable.

server/schedulers/balance_leader.go
```diff
@@ -294,5 +295,7 @@ func (l *balanceLeaderScheduler) createOperator(cluster opt.Cluster, region *cor
 		l.counter.WithLabelValues("move-leader", sourceLabel+"-out"),
 		l.counter.WithLabelValues("move-leader", targetLabel+"-in"),
 	)
+	op.AdditionalInfos["sourceScore"] = strconv.FormatFloat(sourceScore, 'f', -1, 64)
```
disksing, line 298: I think precision=2 should be enough.

lhy1024: done

disksing (approved):

nolouch (commented):
LGTM

ti-srebot (approved):

PR #3007, opened 2020-09-22
server: warn useless config change when placement rules enabled

disksing:

After placement rues enabled, updating `max-replicas` or `location-labels` won't work. User may not know it and get confused if they did not read doc carefully.

Give some warning when user want to update config.

### Check List

Tests
- Unit test

### Release note
- No release note

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -808,6 +809,15 @@ func (s *Server) SetReplicationConfig(cfg config.ReplicationConfig) error {
 			}
 		}
 	}
+	if cfg.EnablePlacementRules {
```
rleungx, line 812: I think we can move it to validate and no need to compare it with the old config.

disksing: It looks we cannot know if `MaxReplicas` or `LocationLables` is changed without comparing.

rleungx: Oh, right. But I still think this validate logic should be put into the validate funciton.

disksing: But the config is valid, only the config transition is invalid.

rleungx: make sense

HunDunDM (approved):
LGTM

rleungx (approved):
LGTM

ti-srebot (approved):

Issue #2963, opened 2020-09-15
refactor task: move `server/heartbeat_stream_test.go` to `server/schedule/hbstream`


disksing:
## Development Task

Derived from https://github.com/tikv/pd/issues/2882
It requires some more updates to the unit test to make it possible to move to the new location, so we do it in a separated PR.

disksing:
Please take a look @duduainankai

disksing:
Hi @duduainankai I have proposed a PR to fix this issue. You may want to help review or pick other `help-wanted` issues :)

duduainankai:
Sorry for the slow response since a busy week (I was about to write the code :(

PR #3010, opened 2020-09-22
server: move heartbeat test to heartbeat package

disksing:

Fix https://github.com/tikv/pd/issues/2963

- move code to new place
- Adjust the test method to avoid starting the server

### Check List

Tests
- Unit test

### Release note
- No release note


server/schedule/hbstream/heartbeat_stresms_test.go
```diff
@@ -0,0 +1,66 @@
...
+	cluster := mockcluster.NewCluster(config.NewTestOptions())
+	cluster.AddRegionStore(1, 1)
+	cluster.AddRegionStore(2, 0)
+	cluster.AddLeaderRegion(1, 1)
+	region := cluster.GetRegion(1)
+	msg := &pdpb.RegionHeartbeatResponse{
+		ChangePeer: &pdpb.ChangePeer{Peer: &metapb.Peer{Id: 2, StoreId: 2}, ChangeType: eraftpb.ConfChangeType_AddLearnerNode},
+	}
+
+	hbs := NewHeartbeatStreams(ctx, 1, cluster)
```
duduainankai, line 39: How about using `NewTestHeartbeatStreams` and `cluster.ID` ?

server/schedule/hbstream/heartbeat_stresms_test.go
```diff
@@ -0,0 +1,66 @@
...
+	stream1, stream2 := mockhbstream.NewHeartbeatStream(), mockhbstream.NewHeartbeatStream()
+
+	// Active stream is stream1.
+	hbs.BindStream(1, stream1)
+	time.Sleep(50 * time.Millisecond)
```
duduainankai, line 44: Is there any better way to notify ? `Sleep` maybe not so stable.

duduainankai (commented):
Can we keep the origin `heartbeat_stream_test.go` to do more "upper class" test at `server` level ? But maybe give another name for it.

server/schedule/hbstream/heartbeat_stresms_test.go
disksing: good idea.

server/schedule/hbstream/heartbeat_stresms_test.go
disksing: Seems not easy to do because heartbeat stream provides an async interface. Maybe I should use `testutil.WaitUtil` to make it more predictable.

duduainankai:
lgtm

> Can we keep the origin `heartbeat_stream_test.go` to do more "upper class" test at `server` level ? But maybe give another name for it.

What do you think of this ?

disksing:
> lgtm
> 
> > Can we keep the origin `heartbeat_stream_test.go` to do more "upper class" test at `server` level ? But maybe give another name for it.
> 
> What do you think of this ?

Seems not so necessary. We do have some chaos tests to check if it works fine when network reconnect happens and now we want PD repo contains more unit tests and less integration test (especially that requires running a server).

duduainankai:
> > lgtm
> > > Can we keep the origin `heartbeat_stream_test.go` to do more "upper class" test at `server` level ? But maybe give another name for it.
> > 
> > 
> > What do you think of this ?
> 
> Seems not so necessary. We do have some chaos tests to check if it works fine when network reconnect happens and now we want PD repo contains more unit tests and less integration test (especially that requires running a server).

Got

Yisaer (approved):
LGTM

server/schedule/hbstream/heartbeat_stresms_test.go
```diff
@@ -0,0 +1,65 @@
+package hbstream
```
HunDunDM, line 1: need copyright.

HunDunDM: And there is a typo in the file name.

disksing: Fixed /cc @HunDunDM

ZenoTan:
lgtm

nolouch (commented):
LGTM

ti-srebot (approved):

PR #3011, opened 2020-09-22
logutil, *: implement log desensitization 

### LINKED ISSUE #2852: log: requesting log desensitization
## Development Task

To reinforce the security in the PD, one thing we need to do is to do the log desensitization.  
Here are some examples we need to hide from the logs in my view:

1. region key in logs like following:
```golang
log.Error("wrong range keys",zap.String("start-key", string(HexRegionKey(startKey))),zap.String("end-key", string(HexRegionKey(endKey))))
```

2.  etcd key and value in logs like following:
```golang
log.Warn("kv gets too slow", zap.String("request-key", key), zap.Duration("cost", cost), zap.Error(err))
```

3. store label key and value in logs like following:
```golang
log.Warn("not found the key match with the store label",zap.Stringer("store", s.GetMeta()),zap.String("label-key", key))
```

4. placement rule key and value in logs like following:
```golang
log.Error("rule is in bad format", zap.String("rule-key", k), zap.String("rule-value", v), zap.Error(errs.ErrLoadRule.FastGenByArgs()), zap.NamedError("cause", err))
```

We will add a new configuration like "enable-log-desensitization"(default false). If this configuration is enabled, the sensitive information won't appear on the previous log.

Yisaer:
close #2852 

Implement log desensitization to omit the region key information from the log.

### Check List

Tests

- Unit test

Code changes

- Has configuration change

Related changes

### Release note

- Support configuration `enable-redact-log` to enable log desensitization or not.


cmd/pd-server/main.go
```diff
@@ -78,6 +78,7 @@ func main() {
 	if err != nil {
 		log.Fatal("initialize logger error", errs.ZapError(err))
 	}
+	logutil.SetRedactLog(cfg.EnableRedactLog == int32(1))
```
disksing, line 81: can we do it in `cfg.SetupLogger`?

server/config/config.go
```diff
@@ -144,6 +144,8 @@ type Config struct {
 	Dashboard DashboardConfig `toml:"dashboard" json:"dashboard"`
 
 	ReplicationMode ReplicationModeConfig `toml:"replication-mode" json:"replication-mode"`
+	// EnableRedactLog indicates that whether redact log, 0 is disable. 1 is enable.
+	EnableRedactLog int32 `toml:"enable-redact-log" json:"enable-redact-log"`
```
disksing, line 148: why not use boolean?

tools/pd-simulator/simulator/event.go
```diff
@@ -86,7 +87,7 @@ func (e *WriteFlowOnSpot) Run(raft *RaftEngine, tickCount int64) bool {
 		region := raft.SearchRegion([]byte(key))
 		simutil.Logger.Debug("search the region", zap.Reflect("region", region.GetMeta()))
 		if region == nil {
-			simutil.Logger.Error("region not found for key", zap.String("key", key))
+			simutil.Logger.Error("region not found for key", zap.String("key", logutil.RedactArgIfNeeded(key).(string)))
```
rleungx, line 90: No need to change it?

pkg/logutil/log_test.go
```diff
@@ -109,3 +109,59 @@ func (s *testLogSuite) TestFileLog(c *C) {
...
+
+	for _, testcase := range testcases {
+		c.Log(testcase.name)
+		SetRedactLog(testcase.enableRedactLog)
+		r := RedactArgIfNeeded(testcase.arg)
+		if testcase.expect == nil {
+			c.Assert(r, IsNil)
+			continue
+		}
+		switch testcase.expect.(type) {
```
disksing, line 160: I think we can use `DeepEquals` here.

rleungx:
BTW, I think there are more logs which need to be handled.

pkg/logutil/log.go
```diff
@@ -292,3 +293,39 @@ func LogPanic() {
...
+	return enabledRedactLog.Load().(bool)
+}
+
+// SetRedactLog sets enabledRedactLog
+func SetRedactLog(enabled bool) {
+	enabledRedactLog.Store(enabled)
+}
+
+// RedactArgIfNeeded will omit the argument if RedactLog is enabled
+func RedactArgIfNeeded(arg interface{}) interface{} {
```
disksing, line 316: I think it would be more easy to use if we can provide 2 functions for `string` and `[]byte` separately.

Yisaer:
> BTW, I think there are more logs which need to be handled.

After discussion, only start-key and end-key of the region need to be redacted in the log. @rleungx

server/cluster/cluster.go
```diff
@@ -554,7 +554,7 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		if r.GetVersion() > o.GetVersion() {
 			log.Info("region Version changed",
 				zap.Uint64("region-id", region.GetID()),
-				zap.String("detail", core.DiffRegionKeyInfo(origin, region)),
+				zap.String("detail", logutil.RedactArgIfNeeded(core.DiffRegionKeyInfo(origin, region)).(string)),
```
rleungx, line 557: line 566 also needs to be changed

rleungx:
> > BTW, I think there are more logs which need to be handled.
> 
> After discussion, only start-key and end-key of the region need to be redacted in the log. @rleungx

Does `RegionToHexMeta` need to be changed to `?`?

tools/pd-simulator/simulator/event.go
Yisaer: good catch

server/config/config.go
Yisaer: To keep the same behavior with `tidb` 

https://github.com/pingcap/tidb/blob/fd0299f58ecc174e5ed8f91660299ebac8fa2b9c/config/config.go#L163-L164

disksing: They use int32 because they need to do atomic operations but we don't. I think we don't have to follow them.

pkg/logutil/log.go
```diff
@@ -292,3 +293,49 @@ func LogPanic() {
...
+	return enabledRedactLog.Load().(bool)
+}
+
+// SetRedactLog sets enabledRedactLog
+func SetRedactLog(enabled bool) {
+	enabledRedactLog.Store(enabled)
+}
+
+// RedactBytesArgIfNeeded receives []byte argument and return omitted information if redact log enabled
+func RedactBytesArgIfNeeded(arg []byte) []byte {
```
disksing, line 316: Looks verbose. How about just `RedactBytes`?

Yisaer: updated.

disksing (approved):

HunDunDM (commented):
How about using a function like `func ZapRedact(key string, val interface{}) zap.Field`?

pkg/logutil/log.go
```diff
@@ -292,3 +293,65 @@ func LogPanic() {
...
+	return redactArgIfNeeded(arg).(string)
+}
+
+// RedactStringer receives stringer argument and return omitted information if redact log enabled
+func RedactStringer(arg fmt.Stringer) fmt.Stringer {
+	return redactArgIfNeeded(arg).(fmt.Stringer)
+}
+
+type stringer struct {
+	s string
```
HunDunDM, line 331: In fact, it is always "?". I don't think it is necessary.

pkg/logutil/log.go
```diff
@@ -292,3 +293,65 @@ func LogPanic() {
 		zaplog.Fatal("panic", zap.Reflect("recover", e))
 	}
 }
+
+var (
+	enabledRedactLog atomic.Value
```
HunDunDM, line 298: how about using `atomic.Bool` from `go.uber.org/atomic`?

Yisaer: `go.uber.org/atomic` is good, but I think it's better to unify the `atomic` introducing in PD.

Yisaer:
> How about using a function like `func ZapRedact(key string, val interface{}) zap.Field`?

good idea.

server/schedule/checker/merge_checker.go
```diff
@@ -123,7 +124,9 @@ func (m *MergeChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		return nil
 	}
 
-	log.Debug("try to merge region", zap.Stringer("from", core.RegionToHexMeta(region.GetMeta())), zap.Stringer("to", core.RegionToHexMeta(target.GetMeta())))
+	log.Debug("try to merge region",
+		zap.Stringer("from", logutil.RedactStringer(core.RegionToHexMeta(region.GetMeta()))),
```
HunDunDM, line 128: `logutil.ZapRedactStringer`

HunDunDM (approved):
LGTM

ti-srebot (approved):

PR #3019, opened 2020-09-24
pkg: add conversion tests

rleungx:

Just as the title says.

This PR adds some conversion tests. Moreover, we can utilize this test to decode the etcd value temporarily. e.g.

```
./etcd-dump-db iterate-bucket default.pd-ip-172-16-5-31 key --decode  | grep alloc
rev={main:2381 sub:0}, value=[key "/pd/6873030750099943686/alloc_id" | val "\x00\x00\x00\x00\x00\x00\x03\xe8" | created 2381 | mod 2381 | ver 1]
```

### Check List

Tests

- Unit test


Yisaer (approved):
LGTM

lhy1024 (approved):

ti-srebot (approved):

PR #3020, opened 2020-09-24
simulator: Add cmd arg to specify log file

ZenoTan:

Simulator log output is to stdout by default previously.

Now we can have log output to a file, by arg `simLogFile`.

### Check List

Tests

- No code


lhy1024 (approved):

nolouch:
Is this compatible with the original behavior?

ZenoTan:
> Is this compatible with the original behavior?

If not specified, then it is stdout.

tools/pd-simulator/main.go
```diff
@@ -45,6 +45,7 @@ var (
 	caseName                    = flag.String("case", "", "case name")
 	serverLogLevel              = flag.String("serverLog", "fatal", "pd server log level")
 	simLogLevel                 = flag.String("simLog", "fatal", "simulator log level")
+	simLogFile                  = flag.String("simLogFile", "", "simulator log file")
```
lhy1024, line 48: do we need to add default value?

ZenoTan: It is previously to stdout, so the behavior is as before.

nolouch (commented):
LGTM

ti-srebot (approved):

PR #3027, opened 2020-09-25
tso: Support configuring tso update physical interval

MyonKeminta:

This PR makes the interval to update physical of tso configurable. It will be useful for test purpose.

Makes `UpdateTimestampStep` configurable. But since this config should not be used by users, it's not added to the config example. Nor will it be documented.

I'm currently not very sure if the code is correct.

### Check List

Tests

- Unit test

Code changes

- Has configuration change

Side effects

- Increased code complexity

Related changes

- No

### Release note

* No release note


JmPotato:
Your codes are correct in the term of making this argument become configurable. But I'm wondering whether we should do this because it a kind of internal setting of TSO. If it is changed accidentally, which may cause critical problems to TiDB and TiKV. So maybe you should add some validation for this config, e.g, check whether its value is too big or too small?

MyonKeminta:
The motivation of making this is for testing async commit. Async commit introduces the possibility that transactions' commit ts may not be allocated from PD, but it still need to satisfy the monotonicity in some way. We think if we make PD able to allocate a lot of continuous timestamps, it will be easier for us to find problems in tests.

JmPotato:
> The motivation of making this is for testing async commit. Async commit introduces the possibility that transactions' commit ts may not be allocated from PD, but it still need to satisfy the monotonicity in some way. We think if we make PD able to allocate a lot of continuous timestamps, it will be easier for us to find problems in tests.

Do you have any document about this?

MyonKeminta:
> Do you have any document about this?

For now it's only mentioned in async commit's test plan.

MyonKeminta:
I think I can limit its value in 50ms ~ 10s and print a warning log if it's not the default config. How do you think?

JmPotato:
> I think I can limit its value in 50ms ~ 10s and print a warning log if it's not the default config. How do you think?

Looks good to me.

server/config/config.go
```diff
@@ -86,6 +86,12 @@ type Config struct {
 	// TSOSaveInterval is the interval to save timestamp.
 	TSOSaveInterval typeutil.Duration `toml:"tso-save-interval" json:"tso-save-interval"`
 
+	// The interval to update physical part of timestamp. Usually this config should not be set.
```
JmPotato, line 89:
```suggestion
	// The interval to update the physical part of the timestamp. Usually, this config should not be set.
```

server/config/config.go
```diff
@@ -86,6 +86,12 @@ type Config struct {
...
+	// It's only useful for test purpose.
```
JmPotato, line 90:
```suggestion
	// It's only useful for test purposes.
```

server/config/config.go
```diff
@@ -86,6 +86,12 @@ type Config struct {
...
+	// This config is only valid in 50ms to 10s. If it's configured too long or to short, it will be
```
JmPotato, line 91:
```suggestion
	// This config is only valid in 50ms to 10s. If it's configured too long or too short, it will be
```

server/config/config.go
```diff
@@ -225,6 +231,10 @@ const (
 	defaultDRWaitSyncTimeout  = time.Minute
 	defaultDRWaitAsyncTimeout = 2 * time.Minute
 	defaultEnableRedactLog    = false
+
+	DefaultTSOUpdatePhysicalInterval = 50 * time.Millisecond
```
JmPotato, line 235: Please add some comments for `DefaultTSOUpdatePhysicalInterval `.

JmPotato (commented):
The rest LGTM

server/tso/tso.go
```diff
@@ -55,6 +53,7 @@ type timestampOracle struct {
 	rootPath string
 	// TODO: remove saveInterval
 	saveInterval  time.Duration
+	updatePhysicalInterval time.Duration
```
JmPotato, line 56: Please format this, thx.

JmPotato (approved):
LGTM

JmPotato:
@disksing PTAL

nolouch (commented):
LGTM

ti-srebot (approved):

nolouch:
please resolve the conflicts @MyonKeminta

Issue #3026, opened 2020-09-24
Add panel about distribution of region keys and bytes


lhy1024:
I am not sure whether its format is suitable, I think we can adjust the hot peer threshold by used it

lhy1024:
![image](https://user-images.githubusercontent.com/19542290/94145612-640f3a80-fea5-11ea-929f-e4c03f557f08.png)

lhy1024:
heatmap
![image](https://user-images.githubusercontent.com/19542290/94224743-c90a7500-ff25-11ea-8cd2-de6799cf74ec.png)

nolouch:
Are there only hot spots or all?

nolouch:
Are there only hot spots or all?

lhy1024:
> Are there only hot spots or all?

There is all region.

And I prepare to add an interval statistic with the same method

lhy1024:
![image](https://user-images.githubusercontent.com/19542290/94233632-c7977780-ff3a-11ea-9197-7a30d4c3163e.png)

disksing:
looks great!

PR #3031, opened 2020-09-25
metrics: add heartbeat interval and distribution of region bytes, keys metrics

lhy1024:

to solve #3026 

### Check List

Side effects

Related changes

- Update Grafana

Tests

- Unit test


lhy1024:
image

server/statistics/metrics.go
```diff
@@ -71,6 +71,54 @@ var (
 			Name:      "label_level",
 			Help:      "Number of regions in the different label level.",
 		}, []string{"type"})
+	readByteHist = prometheus.NewHistogram(
```
nolouch, line 74: HotReadBytesHist? It only observes the hot peer

lhy1024: All region which heartbeat will be observed

disksing (approved):

rleungx:
What does the number on the right side mean?

nolouch (commented):
LGTM

ti-srebot (approved):

nolouch:
Does it need a higher version grafana?

PR #3033, opened 2020-09-27
tso: implement the global TSO generation algorithm

### LINKED ISSUE #2984: Implement the global TSO generation algorithm

JmPotato:

Part of #2759. Resolve #2984. This PR includes two parts:

* Implement the new global TSO generation algorithm, which includes about 200 lines of codes.
* Refactor the TSO unit tests, which include about 300 lines of codes, but most of them are just structural adjustments.

The reason I put these two parts together is to have a CI test fully. If this brings you guys serious difficulty in reviewing, I will take these two parts separately though.

### Check List

Tests

- Unit test
- Integration test

Side effects

- Increased code complexity


server/tso/global_allocator.go
```diff
@@ -84,8 +109,145 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+	}
+	// Send maxTS to all Local TSO Allocator leaders to prewrite
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	var maxTS time.Time
+	// Collect the MaxTS with all Local TSO Allocator leaders first
+	if maxTS, err = gta.syncMaxTS(ctx, leaderURLs, dcLocationMap, maxTS); err != nil {
+		return pdpb.Timestamp{}, err
+	}
+	maxTS = maxTS.Add(updateTimestampGuard)
```
disksing, line 143: I think we need to use `max physical+count` here.

server/tso/global_allocator.go
```diff
@@ -84,8 +109,145 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+	close(errCh)
+	// If any error occurs, the synchronization process will fail
+	if err := <-errCh; err != nil {
+		return typeutil.ZeroTime, err
+	}
+	for resp := range respCh {
+		if resp == nil {
+			return typeutil.ZeroTime, errs.ErrSyncMaxTS.FastGenWithCause("got nil response")
+		}
+		if notSyncedDCs := gta.checkSyncedDCs(dcLocationMap, resp.GetDcs()); len(notSyncedDCs) > 0 {
```
disksing, line 201: It looks that if there are multiple responses, it is likely to return error on receiving the first response.

disksing: And I think it is better to retry instead return an error directly.

pkg/tsoutil/tso.go
```diff
@@ -51,3 +51,20 @@ func GenerateTimestamp(physical time.Time, logical uint64) *pdpb.Timestamp {
 		Logical:  int64(logical),
 	}
 }
+
+// CompareTimestamp is used to compare the two timestamps.
+// If tsoOne > tsoTwo, returns 1.
+// If tsoOne = tsoTwo, returns 0.
+// If tsoOne < tsoTwo, returns -1.
+func CompareTimestamp(tsoOne, tsoTwo *pdpb.Timestamp) int {
```
rleungx, line 59: prefer to change it to
```
if a {
    return 1
}

if b {
    return 0
}

return -1
```

server/tso/global_allocator.go
```diff
@@ -84,8 +108,154 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+	dcLocationMap, err := gta.allocatorManager.GetClusterDCLocations()
+	if err != nil {
+		log.Error("get cluster dc-locations info failed", errs.ZapError(err))
+		return pdpb.Timestamp{}, err
+	}
+	// No dc-locations configured in the cluster
+	if len(dcLocationMap) == 0 {
+		return gta.timestampOracle.getTS(gta.leadership, count)
+	}
+	// Colletc all allocator leaders' client URLs
```
rleungx, line 121:
```suggestion
	// Collect all allocator leaders' client URLs
```

server/tso/global_allocator.go
```diff
@@ -84,8 +108,154 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+				leaderURLs = append(leaderURLs, leaderURL)
+			}
+		}
+	}
+	// Send maxTS to all Local TSO Allocator leaders to prewrite
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	maxTSO := &pdpb.Timestamp{}
+	// Collect the MaxTS with all Local TSO Allocator leaders first
+	if err = gta.syncMaxTS(ctx, leaderURLs, dcLocationMap, maxTSO); err != nil {
```
rleungx, line 139: line 139 and line 144 use the same method but the comments are different.

server/tso/global_allocator.go
```diff
@@ -84,8 +108,154 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+				}
+				respCh <- resp
+				wg.Done()
+			}(ctx, leaderConn, respCh, errCh)
+		}
+		wg.Wait()
+		close(respCh)
+		close(errCh)
+		// If any error occurs, the synchronization process will fail
+		if err := <-errCh; err != nil {
```
rleungx, line 194: Do we need to show all errors?

server/tso/global_allocator.go
```diff
@@ -84,8 +108,154 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+}
+
+const (
+	dialTimeout = 3 * time.Second
+	rpcTimeout  = 3 * time.Second
+)
+
+func (gta *GlobalTSOAllocator) syncMaxTS(ctx context.Context, leaderURLs []string, dcLocationMap map[string][]uint64, maxTSO *pdpb.Timestamp) error {
+	maxRetryCount := 1
+	for i := 0; i < maxRetryCount; i++ {
```
rleungx, line 161: Will this affect the latency of getting TSO?

server/tso/global_allocator.go
JmPotato: Though it's the same method, it will have different behaviours according to the argument `maxTSO`.

server/tso/global_allocator.go
JmPotato: By default, this loop will only work once. Even if it needs to retry, it will only retry for once. Also, this circumstance will only appear when the request meets the Local TSO Allocator leader changing, so it won't run twice in most situations.

server/grpc_service.go
```diff
@@ -941,8 +941,8 @@ func (s *Server) SyncMaxTS(ctx context.Context, request *pdpb.SyncMaxTSRequest)
 			if err != nil {
 				return nil, err
 			}
-			if currentLocalTSO.Physical > maxLocalTS.Physical {
-				maxLocalTS.Physical = currentLocalTSO.Physical
+			if tsoutil.CompareTimestamp(&currentLocalTSO, &maxLocalTS) == 1 {
```
disksing, line 944: usually we use `> 0` instead of `== 1` to check return value of compare functions.

server/tso/global_allocator.go
```diff
@@ -58,6 +75,13 @@ func NewGlobalTSOAllocator(leadership *election.Leadership, rootPath string, sav
 			saveInterval:  saveInterval,
 			maxResetTSGap: maxResetTSGap,
 		},
+		allocatorManager: am,
+		localAllocatorConn: struct {
+			sync.RWMutex
+			clientConns map[string]*grpc.ClientConn
+		}{
+			clientConns: make(map[string]*grpc.ClientConn),
+		},
```
disksing, line 84: looks not very tidy :(

server/tso/global_allocator.go
disksing: only retry when leader of local tso changes.

server/tso/global_allocator.go
```diff
@@ -84,8 +108,158 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+		return pdpb.Timestamp{}, err
+	}
+	leaderURLs := make([]string, 0, len(allocatorLeaders))
+	for _, allocator := range allocatorLeaders {
+		for _, leaderURL := range allocator.GetMember().GetClientUrls() {
+			if slice.NoneOf(leaderURLs, func(i int) bool { return leaderURLs[i] == leaderURL }) {
+				leaderURLs = append(leaderURLs, leaderURL)
+			}
+		}
+	}
```
disksing, line 133: I think we cannot calculate `leaderURLs` before entering `gta.syncMaxTS` because leader of local tso may drift to a PD that previously did not contain any tso leader so the PD will not appear in `leaderURLs`.

server/tso/global_allocator.go
```diff
@@ -84,8 +105,173 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+		// If any error occurs, the synchronization process will fail
+		if err := <-errCh; err != nil {
+			errList = append(errList, err)
+		}
+		if len(errList) > 0 {
+			return errs.ErrSyncMaxTS.FastGenWithCause(errList)
+		}
+		var syncedDCs []string
+		for resp := range respCh {
+			if resp == nil {
```
nolouch, line 207: Does it always get the nil responds because you close it in line 196?

JmPotato: No, it doesn't. I close the channel when all requesting goroutines are done, so if the requests are all successful, the results will keep staying in the buffered channel until I use `for range` to consume them.

server/tso/global_allocator.go
```diff
@@ -91,8 +113,173 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+				return errs.ErrSyncMaxTS.FastGenWithCause("got nil response")
```
rleungx, line 216: I prefer to add some log to help debug this case

server/tso/global_allocator.go
```diff
@@ -91,8 +113,181 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+		// Collect all allocator leaders' client URLs
+		allocatorLeaders, err := gta.allocatorManager.GetLocalAllocatorLeaders()
+		if err != nil {
+			return err
+		}
+		leaderURLs := make([]string, 0, len(allocatorLeaders))
+		for _, allocator := range allocatorLeaders {
+			for _, leaderURL := range allocator.GetMember().GetClientUrls() {
+				if slice.NoneOf(leaderURLs, func(i int) bool { return leaderURLs[i] == leaderURL }) {
+					leaderURLs = append(leaderURLs, leaderURL)
```
disksing, line 169: If sever has more than 1 client urls, it will be requested multiple times.

server/tso/global_allocator.go
```diff
@@ -91,8 +113,181 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+				syncCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+				resp, err := pdpb.NewPDClient(conn).SyncMaxTS(syncCtx, request)
+				cancel()
+				if err != nil {
+					errCh <- err
+					log.Warn("sync max ts rpc failed, got an error", zap.String("local-allocator-leader-url", leaderConn.Target()), zap.Error(err))
+				}
+				respCh <- resp
+				if resp == nil {
+					log.Warn("sync max ts rpc failed, got a nil response", zap.String("local-allocator-leader-url", leaderConn.Target()))
```
disksing, line 202: does this log duplicate with L216?

server/tso/global_allocator.go
```diff
@@ -91,8 +113,181 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+				maxRetryCount++
+				continue
+			}
+			return errs.ErrSyncMaxTS.FastGenWithCause(fmt.Sprintf("got unsynced dc-locations: %+v", unSyncedDCs))
+		}
+	}
+	return nil
+}
+
+func (gta *GlobalTSOAllocator) checkSyncedDCs(dcLocationMap map[string][]uint64, syncedDCs []string) []string {
```
disksing, line 247: how about returning a bool instead?

server/tso/tso.go
```diff
@@ -68,11 +68,26 @@ func (t *timestampOracle) setTSOPhysical(next time.Time) {
 	t.tsoMux.Lock()
 	defer t.tsoMux.Unlock()
 	// make sure the ts won't fall back
-	if t.tsoMux.tso == nil || typeutil.SubTimeByWallClock(next, t.tsoMux.tso.physical) >= updateTimestampGuard {
+	if t.tsoMux.tso == nil || typeutil.SubTimeByWallClock(next, t.tsoMux.tso.physical) > 0 {
 		t.tsoMux.tso = &tsoObject{physical: next}
 	}
 }
 
+func (t *timestampOracle) setTSO(nextPhysical time.Time, nextLogical int64) {
```
disksing, line 76: Do we need to update the persisted time window on etcd?

JmPotato: Yes, but this method is only used to set the TSO in memory and the only place we call `setTSO` for now is in `ResetUserTimestamp` which guarantees the time window updating at L221.

disksing: got it

disksing (approved):

server/tso/global_allocator.go
```diff
@@ -91,8 +113,181 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+	defer cancel()
+	maxTSO := &pdpb.Timestamp{}
+	// Collect the MaxTS with all Local TSO Allocator leaders first
+	if err = gta.syncMaxTS(ctx, dcLocationMap, maxTSO); err != nil {
+		return pdpb.Timestamp{}, err
+	}
+	maxTSO.Logical += int64(count)
+	// Sync the MaxTS with all Local TSO Allocator leaders then
+	if err := gta.syncMaxTS(ctx, dcLocationMap, maxTSO); err != nil {
+		return pdpb.Timestamp{}, err
```
nolouch, line 137: If some local allocator's syncMaxTS failed, Does need to rollback the maxTSO of the success local allocator?

JmPotato: I think there is no need to and shouldn't rollback other Local TSO Allocator's TSO. First, we only need to guarantee the TSO won't fallback, and the `MaxTS` we synced before is absolutely greater or equal to the previous local TSO. Second, the Local TSO Allocators already synced successfully may have handled some TSO requests during this period. In this case, the rollback will break the linear constrains.

nolouch: got, Do we need to update the design document？

JmPotato: I will add this to the design document later.

server/tso/global_allocator.go
```diff
@@ -91,8 +113,181 @@ func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
...
+	}
+	var currentGlobalTSO pdpb.Timestamp
+	if currentGlobalTSO, err = gta.getCurrentTSO(); err != nil {
+		return pdpb.Timestamp{}, err
+	}
+	if tsoutil.CompareTimestamp(&currentGlobalTSO, maxTSO) < 0 {
```
nolouch, line 143: Cloud this update guarantee atomicity?

nolouch (commented):
reset LGTM.

server/tso/global_allocator.go
JmPotato: It's not atomic here. But even if the `currentGlobalTSO` changed during the comparison, the later `SetTSO()` will guarantee the atomicity.

nolouch (commented):
LGTM

ti-srebot (approved):

Issue #2997, opened 2020-09-21
Use a separate row for store scaling in the monitor


rleungx:
### Describe your feature request related problem

Sometimes we may encounter the problem that the store can not be offline for a long time, but we don't know the reason. It's usually caused by unhealthy regions that cannot be migrated.

### Describe the feature you'd like

We can provide a separate row for store scaling in the monitor to show the region's health for in the offline store and the estimated time for scaling.

- [ ] add region health status for offline 
- [ ] provide the estimate time for offline
- [ ] change `pd.json` to add a row for scaling

Besides these, we can support more in the future.

PR #3035, opened 2020-09-27
metrics: add offline region metrics

rleungx:

Related to #2997.

This PR adds metrics of regions' health for those offline stores and move the offline info from the stats to offlineStats

### Check List

Tests

- Unit test


server/statistics/region_collection.go
```diff
@@ -100,43 +129,47 @@ func (r *RegionStatistics) Observe(region *core.RegionInfo, stores []*core.Store
...
 	}
 
-	if len(region.GetPeers()) < desiredReplicas {
-		r.stats[MissPeer][regionID] = region
-		peerTypeIndex |= MissPeer
-	} else if len(region.GetPeers()) > desiredReplicas {
-		r.stats[ExtraPeer][regionID] = region
-		peerTypeIndex |= ExtraPeer
-	}
+	var isOffline bool
```
HunDunDM, line 132:
```suggestion
	var hasOffline bool
```

rleungx: IMO, isOffline is more suitable here?

server/statistics/region_collection.go
```diff
@@ -100,43 +129,47 @@ func (r *RegionStatistics) Observe(region *core.RegionInfo, stores []*core.Store
...
 
-	if len(region.GetDownPeers()) > 0 {
-		r.stats[DownPeer][regionID] = region
-		peerTypeIndex |= DownPeer
+	for _, store := range stores {
+		if store.IsOffline() {
+			peer := region.GetStorePeer(store.GetID())
+			if peer != nil {
+				isOffline = true
```
lhy1024, line 138:
```suggestion
				isOffline = true
				break
```

lhy1024 (approved):

server/statistics/region_collection.go
HunDunDM: The region is offline, which looks like the region is not available. But I think maybe we should replace `offline`?

AndreMouche:
LGTM

HunDunDM (approved):

PR #3037, opened 2020-09-27
config: unify the redact log configuration with tikv

Yisaer:

Unify the redact log configuration with tikv

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -1369,7 +1369,7 @@ func (c *LocalTSOConfig) Validate() error {
 
 // SecurityConfig indicates the security configuration for pd server
 type SecurityConfig struct {
-	*grpcutil.SecurityConfig
+	*grpcutil.TLSConfig
 	// RedactInfoLog indicates that whether enabling redact log
 	RedactInfoLog bool `toml:"redact-info-log" json:"redact-info-log"`
 }
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -1369,7 +1369,7 @@ func (c *LocalTSOConfig) Validate() error {
 
 // SecurityConfig indicates the security configuration for pd server
 type SecurityConfig struct {
-	*grpcutil.TLSConfig
+	grpcutil.TLSConfig
 	// RedactInfoLog indicates that whether enabling redact log
 	RedactInfoLog bool `toml:"redact-info-log" json:"redact-info-log"`
 }
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -122,7 +122,7 @@ type Config struct {
 	// an election, thus minimizing disruptions.
 	PreVote bool `toml:"enable-prevote"`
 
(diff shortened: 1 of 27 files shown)


server/config/config.go
```diff
@@ -1373,3 +1366,10 @@ func (c *LocalTSOConfig) Validate() error {
 	}
 	return nil
 }
+
+// SecurityConfig indicates the security configuration for pd server
+type SecurityConfig struct {
+	*grpcutil.SecurityConfig
```
HunDunDM, line 1372: Don't use pointer.

HunDunDM (commented):
rest LGTM

server/config/config_test.go
```diff
@@ -165,7 +165,6 @@ leader-schedule-limit = 0
 	c.Assert(cfg.PreVote, IsTrue)
 	c.Assert(cfg.Schedule.MaxMergeRegionKeys, Equals, uint64(defaultMaxMergeRegionKeys))
 	c.Assert(cfg.PDServerCfg.MetricStorage, Equals, "http://127.0.0.1:9090")
-	c.Assert(cfg.EnableRedactLog, Equals, defaultEnableRedactLog)
```
rleungx, line 168: How about adding another test?

Yisaer: updated.

server/config/config.go
Yisaer: updated.

server/server.go
```diff
@@ -957,7 +957,7 @@ func (s *Server) GetClusterVersion() semver.Version {
 
 // GetSecurityConfig get the security config.
 func (s *Server) GetSecurityConfig() *grpcutil.SecurityConfig {
-	return &s.cfg.Security
+	return &s.cfg.Security.SecurityConfig
```
HunDunDM, line 960: I think the two `SecutityConfig` need to be distinguished.

Yisaer: updated.

server/config/config.go
```diff
@@ -145,8 +145,6 @@ type Config struct {
 	Dashboard DashboardConfig `toml:"dashboard" json:"dashboard"`
 
 	ReplicationMode ReplicationModeConfig `toml:"replication-mode" json:"replication-mode"`
-	// EnableRedactLog indicates that whether redact log, 0 is disable. 1 is enable.
-	EnableRedactLog bool `toml:"enable-redact-log" json:"enable-redact-log"`
```
lhy1024, line 149: Is this configuration compatible with the previous one?

Yisaer: As `EnableRedactLog` haven't been released yet, I think it is not necessary for us to consider being compatible here.

HunDunDM (approved):
LGTM

JmPotato (approved):
LGTM

ti-srebot (approved):

PR #3038, opened 2020-09-28
scheduler: Add auto-gc for Region Scatter

### LINKED ISSUE #3008: Implement auto-gc for the region information stored in scatter
In region scatter, we saved the region information in memory to schedule the target store for the region. see detailed in https://github.com/tikv/pd/pull/2849

As comment in https://github.com/tikv/pd/issues/2732#issuecomment-679079285 , the region information we stored in memory during scattering also needed to be auto gc.

Implement auto-gc for the region information stored in scatter, I think we can directly used `ttl` in the `cache` package.

longfangsong:
Signed-off-by: longfangsong <longfangsong@icloud.com>

fix #3008 

Use `ttl` to do auto-gc in selectedStores in Region Scatter

### Check List

Tests

- Unit test
- Integration test

### Release note

- No release note


server/schedule/region_scatterer.go
```diff
@@ -14,19 +14,21 @@
...
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/server/core"
 	"github.com/tikv/pd/server/schedule/filter"
 	"github.com/tikv/pd/server/schedule/operator"
 	"github.com/tikv/pd/server/schedule/opt"
 	"go.uber.org/zap"
+	"math"
+	"math/rand"
+	"sync"
+	"time"
```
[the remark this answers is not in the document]
longfangsong: > plz format the goimports.

Do you consider add things like 
```makefile
goimports -w .
```
into the Makefile's `lint` target?

server/schedule/region_scatterer.go
```diff
@@ -30,46 +33,69 @@ import (
...
-func newSelectedStores(checkExist bool) *selectedStores {
+func newSelectedStores(ctx context.Context, checkExist bool) *selectedStores {
 	return &selectedStores{
 		checkExist:        checkExist,
-		stores:            make(map[string]map[uint64]struct{}),
-		groupDistribution: make(map[string]map[uint64]uint64),
+		stores:            cache.NewStringTTL(ctx, gcInterval, gcTTL),
+		groupDistribution: cache.NewStringTTL(ctx, gcInterval, gcTTL),
 	}
 }
```
Yisaer, line 55: I think we can directly new a context in the `newSelectedStores` instead of passing it.

Yisaer (commented):
What about add a new unit test to verify the gc?

server/schedule/region_scatterer.go
longfangsong: > I think we can directly new a context in the `newSelectedStores` instead of passing it.

That's my origin idea, but it will cause the ttl's gc goroutine leak, the problem is we do not know when to cancel this context.

I think context is kinda like lifespan for goroutines, and it's reasonable for the ttl's gc goroutine to have the same lifespan with `RegionScatterer`.

server/schedule/region_scatterer_test.go
```diff
@@ -339,5 +348,22 @@ func (s *testScatterRegionSuite) TestScatterGroup(c *C) {
...
+func (s *testScatterRegionSuite) TestSelectedStoreGC(c *C) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	stores := newSelectedStores(ctx, true)
+	stores.put(1, "testgroup")
+	_, ok := stores.getStore("testgroup")
+	c.Assert(ok, Equals, true)
+	_, ok = stores.getGroupDistribution("testgroup")
+	c.Assert(ok, Equals, true)
+	time.Sleep(gcTTL)
```
longfangsong, line 364: I don't know whether it is a good idea to use `Sleep` during the test, it might make the tests run slower.
Any better suggestions?

Yisaer: sleep here is ok.

rleungx: Can we change the gcTTL to shorten the sleep time?

longfangsong: > Can we change the gcTTL to shorten the sleep time?

Maybe we do can make gcTTL a variable. Except the convenience in testing, we may also expost it to the config in the future.

nolouch (commented):
LGTM

rleungx (approved):
LGTM

ti-srebot (approved):

PR #3041, opened 2020-09-28
cluster: check if region is valid before handling it

rleungx:

Currently, we first handle the region heartbeat then check if the region is valid.

This PR moves the check logic to `RegionHeartbeat`

### Check List

Tests

- Unit test


lhy1024 (approved):

lhy1024:
do we need to pick it to 4.0?

JmPotato (approved):
LGTM

ti-srebot (approved):

Issue #2965, opened 2020-09-16
Split Regions handled by PD


Yisaer:
### Describe your feature request related problem
Currently, the split table(regions) in tidb is constructed by following 2 steps:
1. Notify tikv to split regions
2. Notify pd to scatter regions

This solution has several problems listed below which will let the time cost for the whole splitting table become unstable and rather time-consuming.
1. split region without any checking ( see detailed in https://github.com/tikv/pd/issues/2733)
2. any application who wants to using scattering API with pd need to handles `backoff`

Also, as suggested by https://github.com/tikv/pd/issues/2732#issuecomment-682360248 , PD could receive the [startKey,endkey) range directly instead of regionID which could reduce the calling times.

### Describe the feature you'd like

1. PD support split regions (split keys) API with backoff and checking ability.
   - [x] implement region splitter skeleton https://github.com/tikv/pd/pull/3062
      - [x] implement sending splitting operator and watch regions changing. https://github.com/tikv/pd/pull/3072
      - [x] support asserting region valid https://github.com/tikv/pd/pull/3072
      - [x] support splitting regions in concurrency https://github.com/tikv/pd/pull/3113
   - [x] expose the region splitter by Http API. https://github.com/tikv/pd/pull/3078
2. PD support scatter regions[startKey, endKey) API with backoff and checking ability.
   - [x] region scatter supports relocate regions by keyranges or regionID. https://github.com/tikv/pd/pull/3028
   - [x] expose region scatter by Http API. #3051

The API shall be released in v5.0.

Yisaer:
PTAL @nolouch

nolouch:
LGTM

PR #3044, opened 2020-09-29
pkg: support region request sender

Yisaer:

ref https://github.com/tikv/pd/issues/2965

Implement the region request sender by the `RegionRequestSender` in [tikv/client-go](https://github.com/tikv/client-go) without regionCache in order to let PD send kvrpc request directly to TiKV.

### Check List

Tests

- Unit test

Related changes


pkg/client/kvrpc_test.go
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/pingcap/kvproto/pkg/kvrpcpb"
```
rleungx, line 20: move them?

Yisaer: good catch

disksing:
em, I don't get it. Why does PD need to send kvrpc requests?

Yisaer:
> em, I don't get it. Why does PD need to send kvrpc requests?

Because `kvrpc` is a synchronous request, and we can get the newly splitted regionsID in the response while the current region heartbeat rpc is asynchronous request and I don't know how to get the newly splitted regionsID in this way. @disksing

disksing:
> > em, I don't get it. Why does PD need to send kvrpc requests?
> 
> Because `kvrpc` is a synchronous request, and we can get the newly splitted regionsID in the response while the current region heartbeat rpc is asynchronous request and I don't know how to get the newly splitted regionsID in this way. @disksing

I think to introduce tikv client is too heavy. Maybe it could better to use async API and use another way to watch if the region is split.

Issue #2533, opened 2020-06-12
Placement Rules: support add `transfer-region` operator


disksing:
## Development Task
From current API we cannot know each role of the peer, thus cannot support it when placement rules feature is enabled.

## Feature Request

The old way to use `transfer-region` is like following:
```sh
operator add transfer-region 126977 1 4 5
```

The request needs to point out the peer role on the store if the placement rule is enabled like following:
```sh
operator add transfer-region 126977 1 leader 4 follower 5 follower
```

kavirajk:
If I understand correctly,  we need to support both storeIDs(1, 4, 5) and its roles (leader, follower) during transfer region.

I see two changes.
1. accept additional roles in `pd-ctl` and send it to API
2. accept additional roles at the API layer. [here](https://github.com/tikv/pd/blob/master/server/handler.go#L488)

Question: what exactly we do at the API layer, just check if those store ids match the roles that are passed before proceeding? or anything else?

any reference related to transfer-region/placement-rules would be nice!

disksing:
Hi @kavirajk , in fact, this issue is not directly related to the placement rules feature.  Before the introduction of this feature, the long-existing replica role could only be _voter_ (means either _leader_ or _follower_), so it did not need to be specified in the API.

To fix this issue, besides the changes in `pd-ctl` side, at API layer you will need to read the request and determine in which stores the replicas of the region should be distributed, and their respective roles. After composing the target region state, you can call `operator.CreateMoveRegionOperator` to create the operator, inside `CreateMoveRegionOperator`, it uses `operator.Builder` to plan the detailed steps to transform to the target state.

We can support 4 types of roles:
- voter. Same as before
- learner. Need to set `peer.IsLearner=true`.
- leader. Need to make sure the target leader is the peer. (require some change to `CreateMoveRegionOperator`)
- follower. Need to make sure the target leader is not the peer. (require some change to `CreateMoveRegionOperator`)

Yisaer:
clsoed in https://github.com/tikv/pd/pull/3047

PR #3047, opened 2020-09-30
pdctl, server: transfer-region operator support peer role 

mashenjun:
Signed-off-by: Shenjun Ma <mashenjun0902@gmail.com>

`transfer-region` operator is supported with placement rule enabled and user is allowed to point out peer roles. https://github.com/tikv/pd/issues/2533

- pdctl extracts the peer roles information from the user input and put the data in `peer_roles` filed when send post request to '/pd/api/v1/operators' api. 
- add an optional field `peer_roles` to `POST /pd/api/v1/operators` api.
- change parameters of `AddTransferRegionOperator` and `CreateMoveRegionOperator` to support peer roles.
- add codes in `CreateMoveRegionOperator` to process the given peer roles when the build the Operator.
- add field `roles` into builder to store peer role info, which is used to help select target leader. 
- add unit tests for pdctl, api-server and CreateMoveRegionOperator.

### Check List

Tests

- Unit test

Code changes

### Release note

user can point out peer role when perform the transfer-region


Yisaer:
@PhotonQuantum  PTAL

server/api/operator.go
```diff
@@ -337,18 +338,36 @@ func (h *operatorHandler) Delete(w http.ResponseWriter, r *http.Request) {
...
 	}
-	ids := make(map[uint64]struct{})
+	storeIDs := make([]uint64, 0, len(items))
 	for _, item := range items {
 		id, ok := item.(float64)
 		if !ok {
 			return nil, false
 		}
-		ids[uint64(id)] = struct{}{}
+		storeIDs = append(storeIDs, uint64(id))
```
disksing, line 352: I think you can put the id to map directly. `storeIDToPeerRole[id] = ""`, then update it later if role is specified.

server/api/operator.go
```diff
@@ -337,18 +338,36 @@ func (h *operatorHandler) Delete(w http.ResponseWriter, r *http.Request) {
...
 	}
-	return ids, true
+
+	storeIDToPeerRole := make(map[uint64]placement.PeerRoleType)
+
+	peerRoles, ok := roles.([]interface{})
+	if ok && len(peerRoles) >= len(storeIDs) {
```
disksing, line 358: I think it would be better to accept only zero length or same length as storeIDs.

server/handler.go
```diff
@@ -486,22 +487,26 @@ func (h *Handler) AddTransferLeaderOperator(regionID uint64, storeID uint64) err
...
+	if c.GetOpts().IsPlacementRulesEnabled() {
+		// Cannot determine role without peer role when placement rules enabled. Not supported now.
+		for _, role := range storeIDs {
+			if len(role) == 0 {
+				return errors.New("transfer region without peer role is not supported when placement rules enabled")
+			}
+		}
+	}
+
 	if len(storeIDs) > c.GetOpts().GetMaxReplicas() {
```
disksing, line 510: This check is outdated, we can remove it directly.

server/handler.go
```diff
@@ -518,11 +523,13 @@ func (h *Handler) AddTransferRegionOperator(regionID uint64, storeIDs map[uint64
...
-	for id := range storeIDs {
-		peers[id] = &metapb.Peer{StoreId: id}
+	roles := make(map[uint64]placement.PeerRoleType)
+	for id, peerRole := range storeIDs {
+		peers[id] = &metapb.Peer{StoreId: id, Role: peerRole.MetaPeerRole()}
+		roles[id] = peerRole
 	}
 
-	op, err := operator.CreateMoveRegionOperator("admin-move-region", c, region, operator.OpAdmin, peers)
+	op, err := operator.CreateMoveRegionOperator("admin-move-region", c, region, operator.OpAdmin, peers, roles)
```
disksing, line 532: I think we can pass only `roles` and create `metapb.Peer` inside the function.

server/schedule/placement/rule_list.go
```diff
@@ -203,5 +203,8 @@ func (rl ruleList) getRulesForApplyRegion(start, end []byte) []*Rule {
 	if i != len(rl.ranges) && (len(end) == 0 || bytes.Compare(end, rl.ranges[i].startKey) > 0) {
 		return nil
 	}
+	if i == 0 {
```
disksing, line 206: when will it happen?

server/schedule/operator/create_operator.go
```diff
@@ -62,10 +63,23 @@ func CreateForceTransferLeaderOperator(desc string, cluster opt.Cluster, region
...
+	builder := NewBuilder(desc, cluster, region).SetPeers(peers)
+	for storeID, role := range roles {
+		switch role {
+		case placement.Leader:
+			builder.SetLeader(storeID)
+		case placement.Follower:
+			// if all roles are follower, the originLeaderStoreID will be selected in prepare build phase
+			if builder.targetLeaderStoreID == storeID {
+				builder.targetLeaderStoreID = 0
+			}
```
PhotonQuantum, line 76: Maybe this is dead code? The newly created builder has `targetLeaderStoreID = 0` by default, and only `setLeader` can modify it in this function. I wonder whether this can stop the builder from choosing the store as a leader.

For example, suppose you have `1(follower), 3(follower), 4(leader)`, then `transfer-region <region> 1 voter 2 voter 4 follower`, you will get `1(follower), 2(follower), 4(leader)`, which is not expected.

server/api/operator.go
mashenjun: Good point.

server/handler.go
mashenjun: Done.

server/handler.go
mashenjun: Done.

server/schedule/placement/rule_list.go
mashenjun: The `if` statement is unnecessary here since a default rule is inserted during initialization. I added the initialization code in unit tests and removed the statement.

server/api/operator.go
mashenjun: Done.

server/schedule/operator/create_operator.go
mashenjun: Right, setting `targetLeaderStoreID` here is redundant.

server/schedule/operator/builder.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/metapb"
+
```
disksing, line 22: Please remove the empty line

server/schedule/operator/operator.go
```diff
@@ -209,6 +209,9 @@ func (o *Operator) CheckTimeout() bool {
 
 // Len returns the operator's steps count.
 func (o *Operator) Len() int {
+	if o == nil {
```
disksing, line 212: When will this happen? I think better to update test to check nil.

server/schedule/operator/builder.go
mashenjun: Done.

server/schedule/operator/operator.go
mashenjun: Done.

server/schedule/operator/create_operator.go
PhotonQuantum: > I suppose a better plan in the given case is 1(leader), 3(follower), 4(follower) since store 1 already has a peer belonging to the region?

Sounds good. IMO, the point is that as long as you specify store 4 as follower, it should never be assigned as leader.

You may want to take a look at `server/schedule/operator/builder`:`buildStepsWith(out)JointConsensus` to enforce the roles.

disksing (approved):

disksing:
Please help to take a look. @PhotonQuantum @Yisaer

nolouch (commented):
LGTM

ti-srebot (approved):

nolouch:
thanks @mashenjun @PhotonQuantum

PR #3048, opened 2020-10-04
server, pd-ctl: support setting roles in transfer-region operator

### LINKED ISSUE #2533: Placement Rules: support add `transfer-region` operator

PhotonQuantum:
The transfer-region operator needs to accept roles alongside stores to support placement rules feature.

fixes #2533 .

pd-ctl is modified to accept peer roles and send them to backend.

To handle the provided roles, `operator.Builder` now has a new property called `targetFollowerStoreIDs` to specify those peers that mustn't be leaders, and make sure that any of them won't be the leader in the final state. 

`operator.CreateMoveRegionOperator` is modified to specify the target leader and target followers before building the operator.

`handler.AddTransferRegionOperator` is also modified to set learners.

### Check List

Tests

- Unit test
- Integration test

Code changes

### Release note

- `transfer-region` operator is now supported when placement rules feature is enabled, and users must specify roles of stores explicitly in this case.

> P.S.
>
> There's a possibility that the builder generates a redundant leader transfer when joint consensus is disabled. I'm not familiar with Raft, thus I can't eliminate the extra transfer right now without compromising reliability for now.

### DIFF
--- a/server/api/operator.go
+++ b/server/api/operator.go
@@ -21,6 +21,7 @@ import (
 	"github.com/tikv/pd/pkg/apiutil"
 	"github.com/tikv/pd/server"
 	"github.com/tikv/pd/server/schedule/operator"
+	"github.com/tikv/pd/server/schedule/placement"
 	"github.com/unrolled/render"
 )
 
@@ -150,16 +151,16 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
 			h.r.JSON(w, http.StatusBadRequest, "missing region id")
 			return
 		}
-		storeIDs, ok := parseStoreIDs(input["to_store_ids"])
+		storeIDsAndPeerRoles, ok := parseStoreIDsAndPeerRoles(input["to_store_ids"], input["peer_roles"])
 		if !ok {
 			h.r.JSON(w, http.StatusBadRequest, "invalid store ids to transfer region to")
 			return
 		}
-		if len(storeIDs) == 0 {
+		if len(storeIDsAndPeerRoles) == 0 {
 			h.r.JSON(w, http.StatusBadRequest, "missing store ids to transfer region to")
 			return
 		}
-		if err := h.AddTransferRegionOperator(uint64(regionID), storeIDs); err != nil {
+		if err := h.AddTransferRegionOperator(uint64(regionID), storeIDsAndPeerRoles); err != nil {
 			h.r.JSON(w, http.StatusInternalServerError, err.Error())
 			return
 		}
@@ -337,18 +338,33 @@ func (h *operatorHandler) Delete(w http.ResponseWriter, r *http.Request) {
 	h.r.JSON(w, http.StatusOK, "The pending operator is canceled.")
 }
 
-func parseStoreIDs(v interface{}) (map[uint64]struct{}, bool) {
-	items, ok := v.([]interface{})
+func parseStoreIDsAndPeerRoles(rawStoreIDs interface{}, rawPeerRoles interface{}) (map[uint64]placement.PeerRoleType, bool) {
+	storeIDs, ok := rawStoreIDs.([]interface{})
 	if !ok {
 		return nil, false
 	}
-	ids := make(map[uint64]struct{})
-	for _, item := range items {
-		id, ok := item.(float64)
+
+	var peerRoles []interface{}
+	peerRoles, ok = rawPeerRoles.([]interface{})
+	if !ok {
+		peerRoles = nil
+	}
+
+	hasRoles := len(peerRoles) >= len(storeIDs)
+	result := make(map[uint64]placement.PeerRoleType)
+	for i, storeID := range storeIDs {
+		id, ok := storeID.(float64)
 		if !ok {
 			return nil, false
 		}
-		ids[uint64(id)] = struct{}{}
+		if hasRoles {
+			role, ok := peerRoles[i].(string)
+			if ok {
+				result[uint64(id)] = placement.PeerRoleType(role)
+				continue
+			}
+		}
+		result[uint64(id)] = ""
 	}
-	return ids, true
+	return result, true
 }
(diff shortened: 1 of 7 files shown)


nolouch:
Hi @PhotonQuantum, thanks.
This PR is duplicated with https://github.com/tikv/pd/pull/3048. would you like to help do some reviews in that PR?

server/api/operator.go
```diff
@@ -337,18 +338,33 @@ func (h *operatorHandler) Delete(w http.ResponseWriter, r *http.Request) {
...
-	for _, item := range items {
-		id, ok := item.(float64)
+
+	var peerRoles []interface{}
+	peerRoles, ok = rawPeerRoles.([]interface{})
+	if !ok {
+		peerRoles = nil
+	}
+
+	hasRoles := len(peerRoles) >= len(storeIDs)
```
disksing, line 353: I think it would be better to accept only zero length or the same length as storeID.

server/handler.go
```diff
@@ -486,28 +487,32 @@ func (h *Handler) AddTransferLeaderOperator(regionID uint64, storeID uint64) err
...
 
 	region := c.GetRegion(regionID)
 	if region == nil {
 		return ErrRegionNotFound(regionID)
 	}
 
-	if len(storeIDs) > c.GetOpts().GetMaxReplicas() {
-		return errors.Errorf("the number of stores is %v, beyond the max replicas", len(storeIDs))
+	if len(storeIDsAndRoles) > c.GetOpts().GetMaxReplicas() {
+		return errors.Errorf("the number of stores is %v, beyond the max replicas", len(storeIDsAndRoles))
```
disksing, line 511: When placement rules enabled, we cannot use `MaxReplicas` to determine max replicas for a region. I think we can remove this check directly.

server/schedule/operator/create_operator.go
```diff
@@ -62,10 +63,18 @@ func CreateForceTransferLeaderOperator(desc string, cluster opt.Cluster, region
...
-		Build(kind)
+func CreateMoveRegionOperator(desc string, cluster opt.Cluster, region *core.RegionInfo, kind OpKind, peers map[uint64]*metapb.Peer, roles map[uint64]placement.PeerRoleType) (*Operator, error) {
+	builder := NewBuilder(desc, cluster, region).
+		SetPeers(peers)
+	for key, role := range roles {
+		switch role {
+		case placement.Leader:
+			builder = builder.SetLeader(key)
+		case placement.Follower:
+			builder.targetFollowerStoreIDs[key] = struct{}{}
```
disksing, line 74: better to add a new function like `setLeader`.

server/handler.go
```diff
@@ -518,11 +523,13 @@ func (h *Handler) AddTransferRegionOperator(regionID uint64, storeIDs map[uint64
...
-	for id := range storeIDs {
-		peers[id] = &metapb.Peer{StoreId: id}
+	roles := make(map[uint64]placement.PeerRoleType)
+	for id, role := range storeIDsAndRoles {
+		peers[id] = &metapb.Peer{StoreId: id, Role: role.MetaPeerRole()}
+		roles[id] = role
 	}
 
-	op, err := operator.CreateMoveRegionOperator("admin-move-region", c, region, operator.OpAdmin, peers)
+	op, err := operator.CreateMoveRegionOperator("admin-move-region", c, region, operator.OpAdmin, peers, roles)
```
disksing, line 532: I think we can add some check to make sure there is at most 1 "leader" and there are at least 1 "leader" or "voter".

tools/pd-ctl/pdctl/command/operator.go
```diff
@@ -423,3 +430,34 @@ func parseUint64s(args []string) ([]uint64, error) {
...
+		return 0, nil, nil, errors.WithStack(err)
+	}
+
+	if _, err := strconv.ParseUint(args[2], 10, 64); err == nil {
+		// No peer roles provided in arguments
+		storeIDs, err := parseUint64s(args[1:])
+		return regionID, storeIDs, nil, err
+	}
+
+	storeIDs, peerRoles := make([]uint64, 0, len(args)), make([]string, 0, len(args))
```
rleungx, line 446: Why len(args)?

PhotonQuantum: I was trying to preallocate the slice. It seems that I misunderstood how `append` works in go, and I'm removing it. Sorry!

server/handler.go
PhotonQuantum: Sounds good! Check added.

server/api/operator.go
PhotonQuantum: Done.

server/handler.go
PhotonQuantum: I added a condition check here. Only when placement rules disabled will the check be performed now, or should I remove the check completely?

server/schedule/operator/create_operator.go
PhotonQuantum: I added `setFollowers`, `addFollower` and `removeFollower`.

server/api/operator.go
```diff
@@ -150,16 +151,16 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
 			h.r.JSON(w, http.StatusBadRequest, "missing region id")
 			return
 		}
-		storeIDs, ok := parseStoreIDs(input["to_store_ids"])
+		storeIDsAndPeerRoles, ok := parseStoreIDsAndPeerRoles(input["to_store_ids"], input["peer_roles"])
 		if !ok {
 			h.r.JSON(w, http.StatusBadRequest, "invalid store ids to transfer region to")
```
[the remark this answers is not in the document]
PhotonQuantum: Done.

PhotonQuantum:
Duplicate with #3047. Closing this one.

PR #3055, opened 2020-10-09
api: add status query parameter for the scheduler API

### LINKED ISSUE #3052: Add an interface to check the status of paused scheduler.
To solve https://github.com/pingcap/br/issues/549, when br pause the scheduler with a specified delay. we only know the request is succeed or not. then it's not easy for br to add integration test to test the status of scheduler.

It's better to have an interface to know whether the scheduler paused succeed or not.

rleungx:

Closes #3052.

This PR supports the status query parameter for the scheduler API. For now, it only supports the `paused` status. If there is no status query parameter, the default behavior is the same as before.

### Check List

Tests

- Manual test

### DIFF
--- a/server/api/scheduler.go
+++ b/server/api/scheduler.go
@@ -60,7 +60,7 @@ func (h *schedulerHandler) List(w http.ResponseWriter, r *http.Request) {
 	status := r.URL.Query().Get("status")
 	switch status {
 	case "paused":
-		var pausedScehdulers []string
+		var pausedSchedulers []string
 		for _, scheduler := range schedulers {
 			paused, err := h.IsSchedulerPaused(scheduler)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


server/api/scheduler.go
```diff
@@ -56,6 +56,26 @@ func (h *schedulerHandler) List(w http.ResponseWriter, r *http.Request) {
...
+
+			if paused {
+				pausedScehdulers = append(pausedScehdulers, scheduler)
+			}
+		}
+		h.r.JSON(w, http.StatusOK, pausedScehdulers)
+		return
+	default:
+	}
 	h.r.JSON(w, http.StatusOK, schedulers)
```
disksing, line 79: move this line after `default:`?

server/api/scheduler.go
```diff
@@ -56,6 +56,26 @@ func (h *schedulerHandler) List(w http.ResponseWriter, r *http.Request) {
 		h.r.JSON(w, http.StatusInternalServerError, err.Error())
 		return
 	}
+
+	status := r.URL.Query().Get("status")
+	switch status {
+	case "paused":
+		var pausedScehdulers []string
```
howardlau1999, line 63:
```suggestion
		var pausedSchedulers []string
```

howardlau1999:
LGTM

server/api/scheduler.go
```diff
@@ -56,7 +56,27 @@ func (h *schedulerHandler) List(w http.ResponseWriter, r *http.Request) {
...
+				h.r.JSON(w, http.StatusInternalServerError, err.Error())
+				return
+			}
+
+			if paused {
+				pausedSchedulers = append(pausedSchedulers, scheduler)
+			}
+		}
+		h.r.JSON(w, http.StatusOK, pausedSchedulers)
+		return
```
disksing, line 76: no need return now.

disksing (approved):

ti-srebot (approved):

Issue #3039, opened 2020-09-28
testCoordinatorSuite.TestCheckRegion failed after the placement rule enable


nolouch:
### What did you do?
relative https://github.com/tikv/pd/pull/3006, the ci failed, and the report is https://travis-ci.org/github/tikv/pd/builds/730835475

### What did you expect to see?
test pass

### What did you see instead?
the test `testCoordinatorSuite.TestCheckRegion ` will fail because the checker will promote pending peers.
### What version of PD are you using (`pd-server -V`)?
master

ti-srebot:
## Please edit this comment to complete the following information

### Not a bug

1. Remove the 'type/bug' label
2. Add notes to indicate why it is not a bug

### Duplicate bug

1. Add the 'type/duplicate' label
2. Add the link to the original bug

### Bug

Note: Make Sure that 'component', and 'severity' labels are added
Example for how to fill out the template: https://github.com/pingcap/tidb/issues/20100

#### 1. Root Cause Analysis (RCA)


#### 2. Symptom



#### 3. All Trigger Conditions



#### 4. Workaround (optional)

#### 5. Affected versions



#### 6. Fixed versions

ti-srebot:
(RCA AllTriggerConditions FixedVersions AffectedVersions Symptom ) fields are empty.

PR #3059, opened 2020-10-10
operator/builder: check healthy peers before promoting learner

ZenoTan:

Solve #3039 

Builder should check learner pending or down before promote it from learner to voter. Those peers are called `unhealthyPeers`.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

- Fix the bug of promote learners when peer unhealthy.


server/schedule/operator/create_operator.go
```diff
@@ -34,6 +34,9 @@ func CreateAddPeerOperator(desc string, cluster opt.Cluster, region *core.Region
 
 // CreatePromoteLearnerOperator creates an operator that promotes a learner.
 func CreatePromoteLearnerOperator(desc string, cluster opt.Cluster, region *core.RegionInfo, peer *metapb.Peer) (*Operator, error) {
+	if region.GetPendingPeer(peer.GetId()) != nil {
```
nolouch, line 37: This check is more reasonable in the builder.

nolouch (commented):
Could you add a test?

server/schedule/operator/create_operator.go
disksing: +1

ZenoTan: Updated.

nolouch:
PTAL @HunDunDM

HunDunDM (commented):
I think the operator also should not transfer the leader to a pending peer or down peer.

server/schedule/operator/builder.go
```diff
@@ -45,6 +45,7 @@ type Builder struct {
 
 	// operation record
 	originPeers         peersMap
+	pendingPeers        peersMap
```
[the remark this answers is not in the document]
HunDunDM: How about `unhealthyPeers`? @nolouch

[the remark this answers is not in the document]
ZenoTan: Will open another PR to deal with other cases.

HunDunDM:
It would be better to change the title like "operator: do not promote pending or down learners".

lhy1024:
Fix pr description, plz.

HunDunDM (approved):
LGTM

server/schedule/operator/builder_test.go
```diff
@@ -442,3 +443,18 @@ func (s *testBuilderSuite) TestBuild(c *C) {
 		}
 	}
 }
+
+// Test for issue 3039
+func (s *testBuilderSuite) TestPromotePending(c *C) {
```
rleungx, line 448: TestPromoteUnhealthyPeer?

rleungx (approved):
LGTM

ti-srebot (approved):

PR #2976, opened 2020-09-16
security: Encrypt region boundary keys, Part 2 - server changes

yiwu-arbug:
Signed-off-by: Yi Wu <yiwu@pingcap.com>

### What problem does this PR solve?

This is part 2 for adding TDE support to PD. pingcap/tidb#18262 This PR adds changes on the server layer. 

### What is changed and how it works?

This PR adds the following:
* Define config struct for encryption (see config.toml changes)
* Update storage.go and region_storage.go to encrypt region boundary keys
* And empty `KeyManager` implementation. Its implementation will be done in the next PR. With the empty implementation, the encryption is an no-op.
* server.go changes to create `KeyManager` and notify the `KeyManager` when the current node become PD leader. The `KeyManager` is supposed to update encryption metadata (e.g. key rotation, handle dynamic config change) only when the node is the PD leader.

### Check List

Tests

- CI. 
- Some manual end-to-end test with a WIP `KeyManager` implementation.

Side effects

- Shouldn't have any side effects, as the encrypt is an no-op before `KeyManager` is implemented.

Related changes

- Part 1 of the project: #2931 

### Release note

- No release note

PR #2931, opened 2020-09-09
Encrypt region boundary keys, Part 1 - helper functions

yiwu-arbug:
Signed-off-by: Yi Wu <yiwu@pingcap.com>

### What problem does this PR solve?

This is part 1 for adding TDE support to PD. https://github.com/pingcap/tidb/issues/18262 It contains helper methods for encryption.

### What is changed and how it works?

This PR adds utility types and functions:
* encrypt/decrypt data using aes-ctr (for encrypting region boundary keys) and aes-gcm (for encrypting data encryption keys)
* helper methods to encrypt/decrypt region boundary keys in-place
* wrapper of master key, and helper method to read a master key (must be a 256 bit encryption key, stored as hex-string) from a local file

### Check List

Tests

- Unit test

Related changes

- depends on https://github.com/pingcap/kvproto/pull/666

### Release note

* No release note

PR #3060, opened 2020-10-12
security: Encrypt region boundary keys, Part 3 - key manager

yiwu-arbug:

This is part 3 for adding TDE support to PD. pingcap/tidb#18262 This PR implements key manager, which handles encryption key management.

The key manager provides the simple interface of:
* GetKey() to decrypt data with a known key id.
* GetCurrentKey() to provide the newest key to encrypt data

The workflow of the key manager is:
* data encryption keys are persisted in etcd with path "encryption/keys".
* On initialize, key manager add watcher to the "encryption/keys" path, reload encryption keys if the revision is newer.
* For a PD follower, the keys load from watcher is used to provide encryption keys
* For a PD leader, when it become a leader, it checks the current key with its own encryption config, and see if a new key needs to be generate (be it encryption is just being enabled, encryption method changed, or current key expired). Despite whether a new key is generated, it saves the encryption keys with the master key in its config. In this way, if master key config is changed, the data encryption keys are re-encrypted using the new master key, thus finishing master key rotation.
* The update to encryption keys is done through a leader transaction, guarantees only the leader can update encryption keys.
* The leader will also run key rotation check every 10 min, and rotate data key if current key expired.

The current implementation leave room for future support of dynamic config change for encryption. Only the leader is responsible for data key and master key rotation, as a result, only the leader has to aware of config change.

### Check List

Tests

- Unit test
- Manual test with basic workflow (enable/disable encryption, master key/data key rotation)

Code changes

- Has persistent data change
encryption keys are persisted in etcd with path "encryption/keys".

Side effects

- may introduce small overhead when saving and loading regions from storage

Related changes

- part 1 #2931, part 2 #2976 

### Release note

 - No release note


yiwu-arbug:
@HunDunDM @Yisaer PTAL

yiwu-arbug:
Not sure why the new tests are getting the error. `ioutil.TempDir` should work in jenkins?
```
[2020-10-12T02:47:13.382Z] key_manager_test.go:637:
[2020-10-12T02:47:13.382Z]     c.Assert(err, IsNil)
[2020-10-12T02:47:13.382Z] ... value *errors.withStack = [PD:encryption:ErrEncryptionNewMasterKey]fail to get encryption key from file /tmp/test_key_file730139297/key ("[PD:encryption:ErrEncryptionNewMasterKey]fail to get encryption key from file /tmp/test_key_file730139297/key")
```

yiwu-arbug:
the latest test failure is due to race condition in test. it does not block code review. and I will fix soon.

yiwu-arbug:
encryption test should be fixed. we are now getting tso_test timeout. Is it expected?

yiwu-arbug:
CI passes! ping @HunDunDM @Yisaer

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+	return keys, nil
+}
+
+func (m *KeyManager) loadKeys() (keys *encryptionpb.KeyDictionary, err error) {
+	resp, err := etcdutil.EtcdKVGet(m.etcdClient, EncryptionKeysPath)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil || len(resp.Kvs) == 0 {
+		return nil, nil
```
HunDunDM, line 276: If `m.getKeys()` is not `nil` at this time, it means that an accidental deletion has occurred and `err` needs to be returned.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+		if keys.CurrentKeyId == disableEncryptionKeyID {
+			needRotate = true
+		} else {
+			currentKey := keys.Keys[keys.CurrentKeyId]
+			if currentKey == nil {
+				return errs.ErrEncryptionCurrentKeyNotFound.GenWithStack("keyId = %d", keys.CurrentKeyId)
+			}
+			// Rotate key in case of:
+			// * Encryption method is changed.
+			// * Currnet key is exposed.
```
HunDunDM, line 325:
```suggestion
			// * Current key is exposed.
```

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+	go m.startBackgroundLoop(ctx, m.keysRevision)
+	return m, nil
+}
+
+func (m *KeyManager) startBackgroundLoop(ctx context.Context, revision int64) {
+	// Create new context for the loop.
+	loopCtx, loopCancel := context.WithCancel(ctx)
+	defer loopCancel()
+	// Setup key dictionary watcher
+	watcher := clientv3.NewWatcher(m.etcdClient)
```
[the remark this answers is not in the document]
yiwu-arbug: I didn't know I can reuse the same `context`.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+	defer ticker.Stop()
+	// Loop
+	for {
+		select {
+		// Reload encryption keys updated by PD leader (could be ourselves).
+		case resp := <-watchChan:
+			if resp.Canceled {
+				// If the watcher failed, we rely solely on rotateKeyIfNeeded() to reload encryption keys.
+				log.Warn("encryption key watcher canceled")
+				continue
```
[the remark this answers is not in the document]
yiwu-arbug: Good catch. I failed to add the fallback reload mechanism as said in the comment. Fixed.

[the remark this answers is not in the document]
yiwu-arbug: I probably need to add a test.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+
+func (m *KeyManager) rotateKeyIfNeeded(forceUpdate bool) error {
+	if m.leadership == nil || !m.leadership.Check() {
+		// We are not leader.
+		m.leadership = nil
+		return nil
+	}
+	m.helper.eventAfterLeaderCheck()
+	// Reload encryption keys in case we are not up-to-date.
+	keys, err := m.loadKeys()
```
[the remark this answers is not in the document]
yiwu-arbug: renamed to `eventAfterReloadByWatcher`

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+	// Backing storage for key dictionary.
+	etcdClient *clientv3.Client
+	// Encryption method used to encrypt data
+	method encryptionpb.EncryptionMethod
+	// Time interval between data key rotation.
+	dataKeyRotationPeriod time.Duration
+	// Metadata defines the master key to use.
+	masterKeyMeta *encryptionpb.MasterKey
+	// Mutex for updating keys. Used for both of LoadKeys() and rotateKeyIfNeeded().
+	muUpdate sync.Mutex
```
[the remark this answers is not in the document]
yiwu-arbug: nice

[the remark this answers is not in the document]
yiwu-arbug: btw I want to ensure there's no concurrent access to both `leadership` and `keysRevision`, so I use mutex instead of rwmutex.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,452 @@
...
+	}
+	m := &KeyManager{
+		etcdClient:            etcdClient,
+		method:                method,
+		dataKeyRotationPeriod: config.DataKeyRotationPeriod.Duration,
+		masterKeyMeta:         masterKeyMeta,
+		helper:                helper,
+	}
+	// Load encryption keys from storage.
+	_, err = m.loadKeys()
```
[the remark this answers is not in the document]
yiwu-arbug: why? we got error here means we are not able to decrypt existing data (the regions), at least before next reload. I think we should fail PD.

server/encryptionkm/key_manager.go
yiwu-arbug: Not sure if I understand correctly, but for a PD instance without enabling encryption, we could get empty result.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,469 @@
...
+	}
+	keys, err := loadKeysFromKV(kv)
+	if err != nil {
+		return nil, err
+	}
+	m.mu.keysRevision = kv.ModRevision
+	m.keys.Store(keys)
+	log.Info("reloaded encryption keys", zap.Int64("revision", kv.ModRevision))
+	return keys, nil
+}
```
Yisaer, line 279: It seems the `keysRevision` is visited and updated withou holding mutex here.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,469 @@
...
+	_, err = m.loadKeys()
+	if err != nil {
+		return nil, err
+	}
+	// Start periodic check for keys change and rotation key if needed.
+	go m.startBackgroundLoop(ctx, m.mu.keysRevision)
+	return m, nil
+}
+
+func (m *KeyManager) startBackgroundLoop(ctx context.Context, revision int64) {
```
Yisaer, line 196: Can we name it as `parCtx` instead of `ctx`.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,469 @@
...
+				continue
+			}
+			for _, event := range resp.Events {
+				if event.Type != mvccpb.PUT {
+					log.Warn("encryption keys is deleted unexpectedly")
+					continue
+				}
+				m.mu.Lock()
+				_, err := m.loadKeysFromKV(event.Kv)
+				m.mu.Unlock()
```
Yisaer, line 230: I suggest moving mutex holding into `loadKeysFromKV` instead of this way.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,469 @@
...
+				log.Warn("fail to rotate data encryption key", zap.Error(err))
+			}
+			// Fallback mechanism to reload keys if watcher failed.
+			if !watcherEnabled {
+				_, err = m.loadKeys()
+				if err != nil {
+					log.Warn("fail to reload keys after watcher failed", zap.Error(err))
+				}
+			}
+			m.mu.Unlock()
```
Yisaer, line 250: ditto.

Yisaer (commented):
I strongly recommended that holding mutex in following ways: 
```golang
// loadKeysFromKV reload keys from etcd result.
// Require mu lock to be held.
func (m *KeyManager) loadKeysFromKV(
	kv *mvccpb.KeyValue,
) (*encryptionpb.KeyDictionary, error) {
	m.mu.Lock()
	defer m.mu.Unlock()
	// Sanity check if keys revision is in order.
	// etcd docs indicates watcher event can be out of order:
	// https://etcd.io/docs/v3.4.0/learning/api_guarantees/#isolation-level-and-consistency-of-replicas
	if kv.ModRevision <= m.mu.keysRevision {
		return m.getKeys(), nil
	}
	keys, err := loadKeysFromKV(kv)
	if err != nil {
		return nil, err
	}
	m.mu.keysRevision = kv.ModRevision
	m.keys.Store(keys)
	log.Info("reloaded encryption keys", zap.Int64("revision", kv.ModRevision))
	return keys, nil
}
```

instead of:

```golang
m.mu.Lock()
_, err := m.loadKeysFromKV(event.Kv)
m.mu.Unlock()
```

I also found there existed deadlock error during the test. https://travis-ci.org/github/tikv/pd/builds/737556806 Could you take a look?

server/server.go
```diff
@@ -360,12 +360,12 @@ func (s *Server) startServer(ctx context.Context) error {
 	if err = s.tsoAllocatorManager.SetLocalTSOConfig(s.cfg.LocalTSO); err != nil {
 		return err
 	}
-	kvBase := kv.NewEtcdKVBase(s.client, s.rootPath)
-	encryptionKeyManager, err := encryptionkm.NewKeyManager(kvBase, &s.cfg.Security.Encryption)
+	encryptionKeyManager, err := encryptionkm.NewKeyManager(ctx, s.client, &s.cfg.Security.Encryption)
```
HunDunDM, line 363: Need to use `s.serverLoopCtx` and `serverLoopWg` to graceful shutdown.

server/encryptionkm/key_manager.go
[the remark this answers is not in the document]
HunDunDM: You are right.

server/encryptionkm/key_manager.go
HunDunDM: We need to use `serverLoopCtx` and `serverLoopWg` here. Please follow the comment in `server.go`.

server/encryptionkm/key_manager.go
HunDunDM: I am referring to the situation equivalent to `log.Warn("encryption keys is deleted unexpectedly")`

server/encryptionkm/key_manager.go
yiwu-arbug: loadKeysFromKV are supposed to be called with mutex held.

server/encryptionkm/key_manager.go
yiwu-arbug: @Yisaer what's `par` stands for?

server/encryptionkm/key_manager.go
yiwu-arbug: got it. I will add a test.

server/encryptionkm/key_manager.go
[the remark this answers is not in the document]
yiwu-arbug: hmm, I'm not able to simulate watcher cancel, even if I cancel the ctx, cancel event is not sent. I give up.

yiwu-arbug:
@HunDunDM @Yisaer updated. mind take another look? Thanks.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,494 @@
...
+					log.Warn("fail to get encryption keys from watcher result", zap.Error(err))
+				}
+			}
+			m.helper.eventAfterReloadByWatcher()
+		case <-m.helper.tick(ticker):
+			m.checkOnTick(watcherEnabled)
+			m.helper.eventAfterTicker()
+			// Server shutdown.
+		case <-ctx.Done():
+			log.Info("encryption key manager is closed")
```
HunDunDM, line 241: Duplicate the log in `Server.encryptionKeyManagerLoop`.

server/encryptionkm/key_manager.go
```diff
@@ -14,41 +14,494 @@
...
+}
+
+// GetKey gets specific encryption key by key id.
+func (m *KeyManager) GetKey(keyID uint64) (*encryptionpb.DataKey, error) {
+	key := m.getKeyLocal(keyID)
+	if key != nil {
+		return key, nil
+	}
+	// Key not found in memory.
+	// The key could be generated by another PD node, which shouldn't happen normally.
```
HunDunDM, line 455: Does it need a warning log here?

HunDunDM (approved):
LGTM

server/encryptionkm/key_manager.go
yiwu-arbug: I'm hesitate to add logging on a frequently access code path.

Yisaer (commented):
LGTM

ti-srebot (approved):

Issue #3054, opened 2020-10-09
placement rules: imporve bundle API


disksing:
## Development Task

We have introduced the bundle API to update placement rules. There is an issue that the `Group` field of received `Rule` may not match the id of the group. I have sent a PR to fix it previously (https://github.com/tikv/pd/pull/2957) but it seems to be not good enough (see https://github.com/tikv/pd/issues/2680#issuecomment-700400319).

To make greater, we can support setting `Group` field automatically when it is unset. Note that when `Group` is set but different from expected group id, we should still report an error.

nolouch:
Can we close this issue?

PR #3061, opened 2020-10-12
api: auto fill group id in rules when unset

duduainankai:
Signed-off-by: Zheming Li <nkdudu@126.com>

#3054 

- Using `GroupBundle.ID` to auto fill `GroupId` field in Rule when unset.

### Check List

Tests

- Unit test

Code changes

### Release note

- No release note

### DIFF
--- a/server/api/rule.go
+++ b/server/api/rule.go
@@ -454,7 +454,7 @@ func (h *ruleHandler) SetAllGroupBundles(w http.ResponseWriter, r *http.Request)
 				h.rd.JSON(w, http.StatusBadRequest, err.Error())
 				return
 			}
-			if len(rule.GroupID) != 0 {
+			if len(rule.GroupID) == 0 {
 				rule.GroupID = g.ID
 			}
 			if rule.GroupID != g.ID {
@@ -535,6 +535,9 @@ func (h *ruleHandler) SetGroupBundle(w http.ResponseWriter, r *http.Request) {
 	var group placement.GroupBundle
 	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &group); err != nil {
 		return
+	}
+	if len(group.ID) == 0 {
+		group.ID = groupID
 	}
 	if group.ID != groupID {
 		h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))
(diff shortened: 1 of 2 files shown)


server/api/rule.go
```diff
@@ -454,6 +454,9 @@ func (h *ruleHandler) SetAllGroupBundles(w http.ResponseWriter, r *http.Request)
 				h.rd.JSON(w, http.StatusBadRequest, err.Error())
 				return
 			}
+			if len(rule.GroupID) != 0 {
```
disksing, line 457: `!=` -> `==`?

disksing:
I think we can auto set `bundle.ID` when it is empty to avoid this error:
```
	if group.ID != groupID {
		h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))			h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))
		return			return
	}		}
```

server/api/rule.go
duduainankai: Yep, a stupid mistake :(

duduainankai:
> I think we can auto set `bundle.ID` when it is empty to avoid this error:
> 
> ```
> 	if group.ID != groupID {
> 		h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))
> 	}
> ```

I have a second think here when wrote the code, maybe it's safer to keep this check ?

disksing:
> > I think we can auto set `bundle.ID` when it is empty to avoid this error:
> > ```
> > 	if group.ID != groupID {
> > 		h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))
> > 	}
> > ```
> 
> I have a second think here when wrote the code, maybe it's safer to keep this check ?

IMO it is the same with the changes we apply to `rule`. It should be ok if we only set it when it is empty.

duduainankai:
> > > I think we can auto set `bundle.ID` when it is empty to avoid this error:
> > > ```
> > > 	if group.ID != groupID {
> > > 		h.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("group id %s does not match request URI %s", group.ID, groupID))
> > > 	}
> > > ```
> > 
> > 
> > I have a second think here when wrote the code, maybe it's safer to keep this check ?
> 
> IMO it is the same with the changes we apply to `rule`. It should be ok if we only set it when it is empty.

Done. PTAL.

disksing (approved):
Great! Thanks!

Issue #3053, opened 2020-10-09
HotRegion scheduler cause  PD panic


nolouch:
## Bug Report

some stores are disconnect and PD  will panic:
![image](https://user-images.githubusercontent.com/6428910/95557378-516d3780-0a47-11eb-9259-dcf6b08cf45b.png)

### What did you expect to see?
PD runs well.
### What did you see instead?
PD panic periodically.
### What version of PD are you using (`pd-server -V`)?
master

lhy1024:
Due to golang's compilation optimization, the actual number of lines that cause errors is 810 instead of 809.

The define of unhealthy is different in core and schedule. #3063

In 4.0.7, statistics are retained for stores with heartbeats within 10 minutes, while candidates are retained within 30 minutes, and nil may appear, which has been reproduced by adding unit tests.

In master, they are 10 minutes and 20 seconds, the candidates are less, so it is safe.

disksing:
Can we close it now?

PR #3069, opened 2020-10-13
Fix the bug that PD might panic if there is down store with 10 minutes

lhy1024:

Fix the bug that PD might panic if there is down store with 10 minutes
#3053 

replace healthy filter with Connected filter

### Check List

Tests

- Unit test

### Release note

- Fix the bug that PD might panic if there is down store with 10 minutes


HunDunDM (approved):
LGTM

nolouch:
LGTM

ti-srebot (approved):

PR #3072, opened 2020-10-15
scheduler: implement region splitter

Yisaer:
ref https://github.com/tikv/pd/issues/2965

Implement the `SplitRegionsHandler` interface and provide the region split http api.

### Check List

Tests

- Unit test


server/api/region.go
```diff
@@ -726,6 +728,56 @@ func (h *regionsHandler) ScatterRegions(w http.ResponseWriter, r *http.Request)
...
+	s := struct {
+		ProcessedPercentage int      `json:"processed-percentage"`
+		NewRegionsID        []uint64 `json:"regions-id"`
+	}{}
+	percentage, newRegionsID := rc.GetRegionSplitter().SplitRegions(splitKeys, retryLimit)
+	s.ProcessedPercentage = percentage
+	s.NewRegionsID = newRegionsID
+	failpoint.Inject("splitResponses", func() {
+		s.ProcessedPercentage = 100
+		s.NewRegionsID = []uint64{99, 100, 101}
```
disksing, line 776: the values look test case specific. I think it would be better to use [failpoint.Value](https://github.com/pingcap/failpoint#how-to-inject-a-failpoint-to-your-program) to inject from test case.

server/schedule/region_splitter.go
```diff
@@ -119,3 +136,54 @@ func (r *RegionSplitter) groupKeysByRegion(keys [][]byte) (map[uint64][][]byte,
...
+	var regionsID map[uint64]struct{}
+	for {
+		select {
+		case <-ticker.C:
+			regions := h.cluster.ScanRegions(startKey, endKey, expectRegionsCount)
+			regionsID = make(map[uint64]struct{}, expectRegionsCount)
+			for _, region := range regions {
+				regionsID[region.GetID()] = struct{}{}
+			}
+			if len(regionsID) < expectRegionsCount {
```
disksing, line 181: Should we consider that a region may split due to other reason (e.g. oversize)? In this case, even if region count is enough, the split keys may different from expected.

Yisaer: That's a good point. I think we can watch and asserting the regions by comparing split keys and regions' start key.

server/api/region.go
Yisaer: updated.

server/schedule/region_splitter.go
```diff
@@ -119,3 +136,57 @@ func (r *RegionSplitter) groupKeysByRegion(keys [][]byte) (map[uint64][][]byte,
...
+
+type splitRegionsHandler struct {
+	cluster opt.Cluster
+	oc      *OperatorController
+}
+
+func (h *splitRegionsHandler) SplitRegionByKeys(region *core.RegionInfo, splitKeys [][]byte) error {
+	op := operator.CreateSplitRegionOperator("region-splitter", region, 0, pdpb.CheckPolicy_USEKEY, splitKeys)
+	if ok := h.oc.AddOperator(op); !ok {
+		log.Info("add region split operator failed", zap.Uint64("regionID", region.GetID()))
```
rleungx, line 162:
```suggestion
		log.Info("add region split operator failed", zap.Uint64("region-id", region.GetID()))
```
BTW, should the level be `info`?

server/schedule/region_splitter.go
```diff
@@ -72,20 +84,25 @@ func (r *RegionSplitter) splitRegionsByKeys(splitKeys [][]byte, newRegions map[u
...
+		if !r.checkRegionValid(region) {
+			unProcessedKeys = append(unProcessedKeys, keys...)
+			continue
+		}
 		err := r.handler.SplitRegionByKeys(region, keys)
 		if err != nil {
 			unProcessedKeys = append(unProcessedKeys, keys...)
 			continue
 		}
 		// TODO: use goroutine to run watchRegionsByKeyRange asynchronously
```
rleungx, line 100: Does this PR need to include it?

Yisaer: I will implement this enhancement in the next PR.

server/schedule/region_splitter.go
Yisaer: updated.

disksing (approved):

PR #3082, opened 2020-10-19
api: support temporary configuration

Yisaer:

1. Provide a temporary API to support temporary configuration for some configurations.

2. Make `/store/{id}/limit` and `/stores/limit` support ttl param.  

**Note that this API is only supported for lightning and BR as a temporary way. Don't call this API.**

As the temporary configuration only is saved in the memory, a recommended way to use API is calling it with short interval (once a minute) and long ttl (10 minutes).

### DIFF
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -47,7 +47,7 @@ func newConfHandler(svr *server.Server, rd *render.Render) *confHandler {
 }
 
 // @Tags config
-// @Summary Update temporary configuration. Note that this is a temporary api and only support for lightning and BR. Don't call this API.
+// @Summary Update temporary configuration. Note that this is a temporary api and only support for lightning and BR. Don't call this API and it will be removed in release-5.0.
 // @Produce json
 // @Success 200 {object} config.Config
 // @Router /config/ttl [post]
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -47,7 +47,6 @@ func newConfHandler(svr *server.Server, rd *render.Render) *confHandler {
 }
 
 func (h *confHandler) SetTTLConfig(w http.ResponseWriter, r *http.Request) {
-	rc := getCluster(r.Context())
 	var input map[string]interface{}
 	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &input); err != nil {
 		return
@@ -61,7 +60,7 @@ func (h *confHandler) SetTTLConfig(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	rc.SaveTTLConfig(input, time.Duration(ttls)*time.Second)
+	h.svr.SaveTTLConfig(input, time.Duration(ttls)*time.Second)
 	h.rd.JSON(w, http.StatusOK, "success")
 }
 
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -20,7 +20,9 @@ import (
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pingcap/errcode"
 	"github.com/pingcap/errors"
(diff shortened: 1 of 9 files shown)


Yisaer:
@3pointer @glorv @overvenus PTAL

server/config/persist_options.go
```diff
@@ -30,6 +32,8 @@ import (
 // PersistOptions wraps all configurations that need to persist to storage and
 // allows to access them safely.
 type PersistOptions struct {
+	// configuration -> ttl value
+	ttl             map[string]*cache.TTLString
```
rleungx, line 36: What if the PD restart?

Yisaer: As we didn't persist it, so we need lightning and BR to call this API with a short interval (like once a minute)

disksing (approved):

server/api/config_test.go
```diff
@@ -275,3 +275,29 @@ func (s *testConfigSuite) TestConfigDefault(c *C) {
...
+	postData, err := json.Marshal(r)
+	c.Assert(err, IsNil)
+	err = postJSON(testDialClient, addr, postData)
+	c.Assert(err, IsNil)
+	c.Assert(s.svr.GetPersistOptions().GetMaxSnapshotCount(), Equals, uint64(999))
+	c.Assert(s.svr.GetPersistOptions().GetStoreLimit(1).AddPeer, Equals, float64(999))
+	c.Assert(s.svr.GetPersistOptions().GetStoreLimit(1).RemovePeer, Equals, float64(999))
+	c.Assert(s.svr.GetPersistOptions().GetMaxMergeRegionSize(), Equals, uint64(999))
+	c.Assert(s.svr.GetPersistOptions().GetMaxMergeRegionKeys(), Equals, uint64(999))
+	time.Sleep(10 * time.Second)
```
lhy1024, line 297: It will delay the test time.

server/api/config.go
```diff
@@ -44,6 +46,29 @@ func newConfHandler(svr *server.Server, rd *render.Render) *confHandler {
 	}
 }
 
+// @Tags config
+// @Summary Update temporary configuration. Note that this is a temporary api and only support for lightning and BR. Don't call this API and it will be removed in release-5.0.
```
lhy1024, line 50: do we need to add a warning in the return result?

lhy1024 (commented):
the rest LGTM

server/api/config_test.go
```diff
@@ -275,3 +275,29 @@ func (s *testConfigSuite) TestConfigDefault(c *C) {
...
 }
+
+func (s *testConfigSuite) TestConfigTTL(c *C) {
+	addr := fmt.Sprintf("%s/config/ttl?ttlSecond=5", s.urlPrefix)
+	r := map[string]int{
+		"max-snapshot-count":      999,
+		"store-limit-add-peer":    999,
+		"max-merge-region-size":   999,
+		"max-merge-region-keys":   999,
+		"store-limit-remove-peer": 999,
```
overvenus, line 286: BR needs `enable-location-replacement` too.

server/api/config_test.go
Yisaer: updated.

server/api/config_test.go
Yisaer: updated.

server/api/config.go
Yisaer: I think the `Warning` in @Description is enough.

server/api/router.go
```diff
@@ -89,6 +89,7 @@ func createRouter(ctx context.Context, prefix string, svr *server.Server) *mux.R
 	apiRouter.HandleFunc("/config/cluster-version", confHandler.SetClusterVersion).Methods("POST")
 	apiRouter.HandleFunc("/config/replication-mode", confHandler.GetReplicationMode).Methods("GET")
 	apiRouter.HandleFunc("/config/replication-mode", confHandler.SetReplicationMode).Methods("POST")
+	apiRouter.HandleFunc("/config/ttl", confHandler.SetTTLConfig).Methods("POST")
```
nolouch, line 92: why not consist with store api,  use `/config?ttlSecond=xx?`

server/api/store.go
```diff
@@ -385,14 +386,29 @@ func (h *storeHandler) SetLimit(w http.ResponseWriter, r *http.Request) {
 		h.rd.JSON(w, http.StatusBadRequest, err.Error())
 		return
 	}
-
+	var ttl int
+	if ttlSec := r.URL.Query().Get("ttlSecond"); ttlSec != "" {
```
nolouch, line 390: update the swager comment

server/api/router.go
Yisaer: updated.

server/api/store.go
Yisaer: updated.

nolouch:
LGTM

ti-srebot (approved):

PR #3091, opened 2020-10-21
config: fix goroutine leak risk

Yisaer:

The `SetTTLData` may have goroutine leak as `NewStringTTL` would create an extra goroutine to do the gc work. When updating an existed key, we need to call the cancel function to close the extra goroutine.
```golang
// SetTTLData set temporary configuration
func (o *PersistOptions) SetTTLData(ctx context.Context, key string, value interface{}, ttl time.Duration) {
	if data, ok := o.ttl[key]; ok {
		data.Clear()
                // the goroutine would leak here
	}
	o.ttl[key] = cache.NewStringTTL(ctx, 5*time.Second, ttl)
	o.ttl[key].Put(key, value)
}
```

As the origin context is directly use the server context, this request fix it by replace a new context which is success from the server context and maintained its cancel function.

### DIFF
--- a/server/config/persist_options.go
+++ b/server/config/persist_options.go
@@ -35,6 +35,7 @@ import (
 type PersistOptions struct {
 	// configuration -> ttl value
 	ttl             map[string]*cache.TTLString
+	ttlCancel       map[string]context.CancelFunc
 	schedule        atomic.Value
 	replication     atomic.Value
 	pdServerConfig  atomic.Value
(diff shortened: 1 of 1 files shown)


JmPotato (approved):
LGTM

server/config/persist_options.go
```diff
@@ -35,6 +35,7 @@ import (
 type PersistOptions struct {
 	// configuration -> ttl value
 	ttl             map[string]*cache.TTLString
+	ttlCancel       map[string]context.CancelFunc
```
disksing, line 38: Why create `cache.TTLString` for each key, instead of use only 1 `cache.TTLString` for all keys?

Yisaer: For each configuration, they could have different  ttl.

PR #3094, opened 2020-10-22
config: fix goroutine leak risk (#3091)

ti-srebot:
cherry-pick #3091 to release-4.0

---

The `SetTTLData` may have goroutine leak as `NewStringTTL` would create an extra goroutine to do the gc work. When updating an existed key, we need to call the cancel function to close the extra goroutine.
```golang
// SetTTLData set temporary configuration
func (o *PersistOptions) SetTTLData(ctx context.Context, key string, value interface{}, ttl time.Duration) {
	if data, ok := o.ttl[key]; ok {
		data.Clear()
                // the goroutine would leak here
	}
	o.ttl[key] = cache.NewStringTTL(ctx, 5*time.Second, ttl)
	o.ttl[key].Put(key, value)
}
```

As the origin context is directly use the server context, this request fix it by replace a new context which is success from the server context and maintained its cancel function.


JmPotato:
Please address the conflicts.

JmPotato:
LGTM

JmPotato (approved):
LGTM

PR #3100, opened 2020-10-26
metrics: optimize the store state filter metrics (#2946)

### LINKED ISSUE #2933: Improve the metrics of store state filter

ti-srebot:
cherry-pick #2946 to release-4.0

---

Close #2933.

This PR records the detailed reason in the filter type. After this PR:
Screen Shot 2020-09-14 at 2 25 07 PM

### Check List

Tests

- Manual test


disksing:
It looks like this cherry-pick comes with many other commits (some of them may include behaviour changes).

lhy1024 (approved):

nolouch (approved):

nolouch (approved):

HunDunDM (approved):

PR #3105, opened 2020-10-26
tso: fix some bad smelling in allocator

Yisaer:

fix bad smelling.

### DIFF
--- a/server/tso/allocator_manager.go
+++ b/server/tso/allocator_manager.go
@@ -65,14 +65,16 @@ type allocatorGroup struct {
 // It is in charge of maintaining TSO allocators' leadership, checking election
 // priority, and forwarding TSO allocation requests to correct TSO Allocators.
 type AllocatorManager struct {
-	sync.RWMutex
+	mu struct {
+		sync.RWMutex
+		// There are two kinds of TSO Allocators:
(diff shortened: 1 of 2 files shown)


server/tso/allocator_manager.go
```diff
@@ -592,11 +589,21 @@ func (am *AllocatorManager) getAllocatorGroups(filters ...AllocatorGroupFilter)
 	return allocatorGroups
 }
 
+func (am *AllocatorManager) getAllocatorGroup(dcLocation string) (*allocatorGroup, bool) {
+	am.mu.RLock()
+	defer am.mu.RUnlock()
+	if len(dcLocation) == 0 {
```
JmPotato, line 595: Because this part of logic is only available and used in the `HandleTSORequest` method, so if you make it an independent function like `getAllocatorGroup` now, maybe we should add some comments or log to tell the user why we do this?

JmPotato (commented):
The rest LGTM

server/tso/allocator_manager.go
Yisaer: good catch, updated.

JmPotato (approved):
LGTM

PR #3123, opened 2020-10-30
tso: do not fail the Global TSO sync when update memory

JmPotato:

Before returning the `maxTSO` as the result, we need to update the Global TSO in memory even in etcd to make it persistent, but because the setting process is not atomic completely, the TSO in memory may be greater after we call `ResetUserTimestamp` then failed the whole synchronization, which will reduce the success rate of obtaining the Global TSO.

Actually, it's ok to let `SetTSO` fail when we update the Global TSO in memory. To make it not fail the Global TSO sync when update memory, I replace the err returning with a warning log.

### Check List

Tests

- Unit test
- Integration test


nolouch:
LGTM

ti-srebot (approved):

PR #3125, opened 2020-10-30
tso: refine the Global TSO synchronization algorithm

### LINKED ISSUE #3126: The Global TSO synchronization failed because non-atomically resetting
### What did you do?

When a Global TSO Allocator handles the TSO generating requests, we need to do the two-phase synchronization between other Local TSO Allocators. During the second phase, we will forcibly set all the Local TSOs to the `maxTS`. But because the resetting function is not completely atomic, the Local TSO may become greater than the `maxTS` during this process, which will cause the synchronization error.

However, it is acceptable for the synchronization that the Local TSO is already greater than `maxTS`. In this situation, we should make the synchronization ignore the error and finish successfully rather than failing it.

### LINKED ISSUE #3127: Inaccurate judgment on whether the Global TSO synchronization is successful
### What did you do?

image

Once we receive a `SyncMaxTSResponse`, we will think the TSO synchronization between this DC is successful according to these codes below.

We should make the response check more precise by differentiating the different synchronization phases. For example, during the first collecting Local TSO phase, before we think the response is valid, we should make sure the content of `MaxLocalTS` is non-nil or not zero.

### LINKED ISSUE #3161: TestSynchronizedGlobalTSO is not stable
```
[2020-11-05T12:24:21.472Z] ----------------------------------------------------------------------
[2020-11-05T12:24:21.472Z] FAIL: global_tso_test.go:440: testSynchronizedGlobalTSO.TestSynchronizedGlobalTSO
[2020-11-05T12:24:21.472Z] 
[2020-11-05T12:24:21.472Z] wait start
[2020-11-05T12:24:21.472Z] wait start
[2020-11-05T12:24:21.472Z] wait start
[2020-11-05T12:24:21.472Z] global_tso_test.go:481:
[2020-11-05T12:24:21.472Z]     // Get a global TSO then
[2020-11-05T12:24:21.472Z]     globalTSO := s.testGetTimestamp(ctx, c, tsoCount, config.GlobalDCLocation)
[2020-11-05T12:24:21.472Z] global_tso_test.go:515:
[2020-11-05T12:24:21.472Z]     c.Assert(err, IsNil)
[2020-11-05T12:24:21.472Z] ... value *status.statusError = &status.statusError{Code:2, Message:"[PD:tso:ErrSyncMaxTS]sync max ts failed, [rpc error: code = Unknown desc = [PD:tso:ErrResetUserTimestamp]reset user timestamp failed, the specified ts is smaller than now]", Details:[]*any.Any(nil), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0} ("rpc error: code = Unknown desc = [PD:tso:ErrSyncMaxTS]sync max ts failed, [rpc error: code = Unknown desc = [PD:tso:ErrResetUserTimestamp]reset user timestamp failed, the specified ts is smaller than now]")
```

Details can be found in https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/pd_test/runs/12918/nodes/45/steps/115/log/?start=0

JmPotato:

Close #3126, close #3127, close #3161.

This PR basically refines the Global TSO synchronization algorithm from two parts.

* Make the `resetUserTimestamp` method atomically set the TSO in memory.
* Allow the `resetUserTimestamp` method to ignore and skip the smaller TSO error check.
* Make the `syncMaxTS` method handle the gRPC response more precisely.

* Rewrite the `resetUserTimestamp` method.
* Update the `syncMaxTS` method.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/tso/global_allocator.go
+++ b/server/tso/global_allocator.go
@@ -100,7 +100,7 @@ func (gta *GlobalTSOAllocator) UpdateTSO() error {
 
 // SetTSO sets the physical part with given TSO.
 func (gta *GlobalTSOAllocator) SetTSO(tso uint64) error {
-	return gta.timestampOracle.ResetUserTimestamp(gta.leadership, tso)
+	return gta.timestampOracle.resetUserTimestamp(gta.leadership, tso, false)
 }
 
 // GenerateTSO is used to generate a given number of TSOs.
@@ -134,7 +134,7 @@ func (gta *GlobalTSOAllocator) GenerateTSO(count uint32) (pdpb.Timestamp, error)
 	}
 	if tsoutil.CompareTimestamp(&currentGlobalTSO, maxTSO) < 0 {
 		// Update the global TSO in memory
-		if err := gta.SetTSO(tsoutil.GenerateTS(maxTSO)); err != nil {
+		if err := gta.timestampOracle.resetUserTimestamp(gta.leadership, tsoutil.GenerateTS(maxTSO), true); err != nil {
 			log.Warn("update the global tso in memory failed", errs.ZapError(err))
 		}
 	}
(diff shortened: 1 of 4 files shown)


JmPotato:
@nolouch @Yisaer @disksing PTAL

server/tso/tso.go
```diff
@@ -193,40 +176,57 @@ func (t *timestampOracle) isInitialized() bool {
...
+			err = errs.ErrResetUserTimestamp.FastGenByArgs("the specified counter is smaller than now")
+		}
 	}
 	// do not update if next physical time is less/before than prev
-	if typeutil.SubTimeByWallClock(nextPhysical, prevPhysical) < 0 {
+	if typeutil.SubTimeByWallClock(nextPhysical, t.tsoMux.tso.physical) < 0 {
 		tsoCounter.WithLabelValues("err_reset_small_ts").Inc()
-		return errs.ErrResetUserTimestamp.FastGenByArgs("the specified ts is smaller than now")
+		if !ignoreSmaller {
+			err = errs.ErrResetUserTimestamp.FastGenByArgs("the specified ts is smaller than now")
```
rleungx, line 201: Will the error be overwritten?

JmPotato: The conditions that may go wrong are mutually exclusive and do not occur at the same time.

rleungx: Could the maxResetTSGap be changed?

JmPotato: Unless the maxResetTSGap is set to a non-postive value, which I think won't happen and could be found in time during the code review.

rleungx: 0 will make line 205 become true

JmPotato: I mean according to the meaning of the maxRestTSGap, it couldn't be zero, though we can add a check to make sure its value is positive.

rleungx: IMO, it's better to guarantee this situation will not happen by ourselves. We don't check this value for now.

JmPotato: I modified the `adjustDuration` method to make sure the duration value is positive. PTAL.

rleungx (approved):

server/tso/global_allocator.go
```diff
@@ -209,17 +209,41 @@ func (gta *GlobalTSOAllocator) syncMaxTS(ctx context.Context, dcLocationMap map[
...
+			// for each phase. For example, if we think we're in the first phase of the Global TSO
+			// synchronization, the inCollectingPhase will be set to true, and during this phase,
+			// any response with nil or empty MaxLocalTs will be regarded as an invalid response.
+			// Then the whole synchronization will fail.
+			if respCount == 1 && resp.GetMaxLocalTs() != nil && resp.GetMaxLocalTs().GetPhysical() != 0 {
+				inCollectingPhase = true
+			}
+			if inCollectingPhase {
+				// Handle the response of the first phase: collect all the Local TSOs
+				if resp.GetMaxLocalTs() == nil || resp.GetMaxLocalTs().GetPhysical() == 0 {
```
nolouch, line 233: `inCollectingPhase` means this condition always is `false`?

nolouch: and why respCount>=2 means not in  collecting phase?

JmPotato:
```go
if respCount == 1 && resp.GetMaxLocalTs() != nil && resp.GetMaxLocalTs().GetPhysical() != 0 {
	inCollectingPhase = true
}
```

This section means we will judge whether we are in the collecting phase (or the 1st phase) according to the first response we receive.

nolouch (commented):
/lgtm

nolouch (approved):
lgtm

PR #3129, opened 2020-11-02
config: enable cross table merge by default

rleungx:

See https://docs.google.com/document/d/1KKVGJMOY6gmuVrVwqGiA2p0sQL5YqF2zoDbqNSgAhKg/edit#

This PR enables cross table merge by default.

### Check List

Tests

- Unit test

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -751,6 +751,7 @@ const (
 	defaultLeaderSchedulePolicy        = "count"
 	defaultStoreLimitMode              = "manual"
 	defaultEnableJointConsensus        = true
+	defaultEnableCrossTableMerge       = true
 )
 
 func (c *ScheduleConfig) adjust(meta *configMetaData) error {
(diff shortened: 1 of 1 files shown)


HunDunDM (approved):
LGTM

lhy1024 (approved):

ti-srebot (approved):

server/config/config.go
```diff
@@ -802,6 +803,9 @@ func (c *ScheduleConfig) adjust(meta *configMetaData) error {
 	if !meta.IsDefined("enable-joint-consensus") {
 		c.EnableJointConsensus = defaultEnableJointConsensus
 	}
+	if !meta.IsDefined("enable-cross-table-merge") {
+		c.EnableCrossTableMerge = defaultEnableCrossTableMerge
```
nolouch, line 807: Can we let the merge is enabled if an upgrade from 4.0?

rleungx: It depends, do you think it's better?

nolouch: I think some paraments should be adjusted to a better default value, even it is upgraded from old version. such as `tolerance-ratio`

rleungx: make sense, but maybe in another PR?

PR #3139, opened 2020-11-02
metrics: replace histogram with gauge in patrol region

rleungx:

We only need to record the latest time spent on patrol regions. No need to use histogram.

This PR replaces histogram with gauge in patrol region.

### Check List

Tests

- Unit test

### DIFF
--- a/server/cluster/coordinator.go
+++ b/server/cluster/coordinator.go
@@ -157,7 +157,7 @@ func (c *coordinator) patrolRegions() {
 		// Updates the label level isolation statistics.
 		c.cluster.updateRegionsLabelLevelStats(regions)
 		if len(key) == 0 {
-			patrolCheckRegionsHistogram.Observe(time.Since(start).Seconds())
+			patrolCheckRegionsGauge.Set(time.Since(start).Seconds())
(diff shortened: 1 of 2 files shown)


nolouch (commented):
LGTM

lhy1024:
why to replace it with gauge?

rleungx:
> why to replace it with gauge?

We only need to record the time spent on the latest round.

HunDunDM (approved):
/lgtm

HunDunDM (approved):
/lgtm

HunDunDM (approved):

lhy1024 (approved):

disksing:
should we choose another metrics name?

rleungx:
> should we choose another metrics name?

Any suggestion?

disksing:
> > should we choose another metrics name?
> 
> Any suggestion?

I don't know. I wonder if it will cause any problem when prometheus has stored data with same metrics name but different type.

rleungx:
PTAL @disksing

disksing (approved):

PR #3141, opened 2020-11-03
security: Encrypt region boundary keys, Part 4 - KMS

yiwu-arbug:

This is part 4 for adding TDE support to PD. pingcap/tidb#18262 This PR supports AWS KMS as master key.

The previous PR in the series support storing master key as a file. This is only suitable for testing. For production this is insecure. This PR supports using AWS KMS as master key.  KMS is a service holds an encryption key (called a CMK), and provide API to encrypt and decrypt data using the encryption key, without exposing the key itself. We only use two of its interfaces:
```
  GenerateDataKey() (plaintextKey, ciphertextKey),
  Decrypt(ciphertextKey) plaintextKey,
```
`GenerateDataKey()` generates an encryption key from the CMK, returning both the plaintext key, and the key after encrypted by the CMK (the ciphertext key). `Decrypt()` can then used to decrypt the ciphertext key. We could use the KMS data key as our data key, but on PD restart we would need to decrypt the data keys one by one, and AWS KMS doesn't provide a batch API. Instead we use the KMS data key as our master key. So here's another layer of envelope encryption.

KMS CMK (hold by KMS, never exposed to us)
    ↓ encrypts
KMS data key (a.k.a PD master key)
    ↓ encrypts
PD data key

When PD persist data keys (after key rotation):
1. Call `GenerateDataKey` to obtain a new KMS data key (plaintext and ciphertext key pair), use it to encrypt the PD data keys
2. Store the encrypted data keys to etcd, along with the ciphertext key returned by KMS. discard the plaintext key.

When PD restart, or is notified of change to the keys through watcher:
1. load the encrypted data keys from etcd, which contains the ciphertext key.
2. call `Decrypt` to restore plaintext KMS data key from the ciphertext key.
3. decrypt PD data keys using the plaintext key.

This PR only support AWS KMS, but in the future we need to support other KMS services (e.g. GCP KMS). The `vender` field in `encryptionpb.MasterKey` is used for forward compatibility - we will fail if the field is not "AWS".

### Check List

Tests
- New unit test
- Manual test with sysbench on a tidb cluster, with KMS key.

Related changes

- part 1 #2931, part 2 #2976, part 3 #3060

### Release note

- No release note


yiwu-arbug:
@Yisaer @HunDunDM

server/encryptionkm/key_manager.go
```diff
@@ -124,7 +131,10 @@ func saveKeys(
 }
 
 // extractKeysFromKV unpack encrypted keys from etcd KV.
-func extractKeysFromKV(kv *mvccpb.KeyValue) (*encryptionpb.KeyDictionary, error) {
+func extractKeysFromKV(
+	kv *mvccpb.KeyValue,
+	helper ...keyManagerHelper,
```
HunDunDM, line 136: Can the length of `helper` only be 0 or 1?

server/encryptionkm/key_manager.go
```diff
@@ -488,6 +503,7 @@ func (m *KeyManager) SetLeadership(leadership *election.Leadership) error {
 type keyManagerHelper struct {
 	now                          func() time.Time
 	tick                         func(ticker *time.Ticker) <-chan time.Time
+	newMasterKey                 func(*encryptionpb.MasterKey, []byte) (*encryption.MasterKey, error)
```
HunDunDM, line 506: Why do we need to add this to the helper?

HunDunDM (commented):
rest LGTM

server/encryptionkm/key_manager.go
yiwu-arbug: how can I enforce that?

server/encryptionkm/key_manager.go
yiwu-arbug: It is used in the unit test to fake and check ciphertext_key.

HunDunDM (approved):

Yisaer (commented):
LGTM

Yisaer (approved):
LGTM

PR #3142, opened 2020-11-03
simulator: Force stop condition for pd bench

ZenoTan:

In bench bot, if simulator runs indefinitely, it will block other bench. So it is necessary to guarantee simulator case will stop.

image

In import-data case, after import, we wait for another period for it to stop. After that, we force it.

### Check List

Tests

- Manual test


lhy1024 (approved):

ZenoTan:
PTAL @nolouch

nolouch (approved):

PR #3146, opened 2020-11-03
server: Make gc_worker's service safepoint special

### LINKED ISSUE #3136: PD need to set an initial value for "gc_worker"'s service safepoint when the cluster bootstraps

It's not a good idea to regard TiDB's GC safepoint and other service safepoints as symmetrical things.

Consider this case:

1. A cluster bootstraps
2. TiDB bootstraps
3. Before the first GC, service 1 (for example, br) registered a  service safepoint 100, and now the minimal service safepoint is 100
4. Before the first GC, service 2 (for example, cdc) registered a service safepoint 99 (because it allocates the timestamp a little earlier). It's less than the current min safepoint so it's rejected. However GC didn't run so the current theoretical safepoint is 0, so service 2 should be able to start.

After discussing with @leoppro , we think the best solution to this problem is to let PD register the "gc_worker" service safepoint as 0 when the cluster starts.

When the cluster bootstraps, register a service safepoint with id = `"gc_worker"`, value = 0, and TTL = inf.

We also considered handle the problem in TiDB's GCWorker or CDC. However those solutions are all hard to do, tricky, or full of corner cases.

MyonKeminta:
Signed-off-by: MyonKeminta <MyonKeminta@users.noreply.github.com>

Fixes #3136 

This PR makes service safepoint with `gc_worker` a special one:

* It has an initial value `0`
* Its TTL must be math.MaxInt64 (which represents infinity)
* It cannot be deleted

Therefore, gc_worker's service safepoint is no longer symmetrical to other services' safepoints.

By the way, this PR also disallowed registering service safepoint with empty ID.

I'm not sure if returning `errors.New(...)` is a proper way to report errors. If it's not, I need better suggestions.

cc @leoppro

### Check List

Tests

- Unit test

Code changes

\-

Side effects

- Increased code complexity

Related changes

  - release-4.0

### Release note

* No release note.


overvenus (commented):
LGTM

leoppro (commented):
LGTM

MyonKeminta:
@rleungx PTAL. Can this be merged?

server/core/storage.go
```diff
@@ -45,6 +45,8 @@ const (
 	componentPath            = "component"
 	customScheduleConfigPath = "scheduler_config"
 	encryptionKeysPath       = "encryption_keys"
+	// GCWorkerServiceSafePointID is the ID of GCWorker's service GC safe point.
+	GCWorkerServiceSafePointID = "gc_worker"
```
rleungx, line 49: How about changing this name to make it private?

MyonKeminta: It's used in test in tests/client/client_test.go .

rleungx: prefer to use "gc_worker" directly in tests

server/core/storage.go
```diff
@@ -485,10 +495,25 @@ func (s *Storage) SaveServiceGCSafePoint(ssp *ServiceSafePoint) error {
...
 	return s.Remove(key)
 }
 
+func (s *Storage) initServiceGCSafePointForGCWorker() (*ServiceSafePoint, error) {
+	ssp := &ServiceSafePoint{
+		ServiceID: GCWorkerServiceSafePointID,
+		SafePoint: 0,
+		ExpiredAt: math.MaxInt64,
+	}
+	if err := s.SaveServiceGCSafePoint(ssp); err != nil {
```
disksing, line 511: return s.SaveServiceGCSafePoint(ssp)

MyonKeminta: `SaveServiceGCSafePoint` doesn't return the ssp, and I think it's better to return nil for the `ssp` on error.

MyonKeminta:
@rleungx @disksing PTAL again thx

rleungx (approved):

server/core/storage.go
disksing: oh, my mistake.

disksing (approved):

MyonKeminta:
CI failed on lightning's test because lightning sets empty string as serviceID, while this PR added non-empty check.

disksing:
> CI failed on lightning's test because lightning sets empty string as serviceID, while this PR added non-empty check.

sad story. Should we update the lightning test?

MyonKeminta:
> > CI failed on lightning's test because lightning sets empty string as serviceID, while this PR added non-empty check.
> 
> sad story. Should we update the lightning test?

I thin lightning should fix the bug that it didn't set service ID.

By the way it's lucky we used `gc_worker` as service ID  for TiDB, instead of using empty to show its specialness.

MyonKeminta:
… or perhaps another choice is to remove the empty check from this PR first to merge this PR quickly, and add the empty check in another PR...

Then people who tries to use old versions of lightning on new tidb cluster will got the unfriendly error...
I have no idea...

HunDunDM:
Do we need to add support in `pd-recover`?

MyonKeminta:
@HunDunDM Sorry I didn't get it. What to recover?

PR #3148, opened 2020-11-04
schedule: check joint state when creating split/merge operator

HunDunDM:

* PD may try to merge/split the region in the joint state.

* schedule: check joint state when creating split/merge operator

### Check List

Tests

- Unit test

### Release note

- No release note


server/schedule/checker/rule_checker.go
```diff
@@ -78,7 +78,13 @@ func (c *RuleChecker) fixRange(region *core.RegionInfo) *operator.Operator {
 	if len(keys) == 0 {
 		return nil
 	}
-	return operator.CreateSplitRegionOperator("rule-split-region", region, 0, pdpb.CheckPolicy_USEKEY, keys)
```
HunDunDM, line 81: @disksing Please confirm whether the rule checker needs to be updated.

disksing: it's ok

server/schedule/checker/rule_checker.go
```diff
@@ -78,7 +78,13 @@ func (c *RuleChecker) fixRange(region *core.RegionInfo) *operator.Operator {
...
+
+	op, err := operator.CreateSplitRegionOperator("rule-split-region", region, 0, pdpb.CheckPolicy_USEKEY, keys)
+	if err != nil {
+		return nil
```
disksing, line 84: better to log the err

HunDunDM: add a debug log

server/schedule/operator/create_operator.go
```diff
@@ -95,7 +95,11 @@ func CreateMoveLeaderOperator(desc string, cluster opt.Cluster, region *core.Reg
 }
 
 // CreateSplitRegionOperator creates an operator that splits a region.
-func CreateSplitRegionOperator(desc string, region *core.RegionInfo, kind OpKind, policy pdpb.CheckPolicy, keys [][]byte) *Operator {
+func CreateSplitRegionOperator(desc string, region *core.RegionInfo, kind OpKind, policy pdpb.CheckPolicy, keys [][]byte) (*Operator, error) {
+	if core.IsInJointState(region.GetPeers()...) {
+		return nil, errors.Errorf("cannot split region which is in joint state")
```
rleungx, line 100: We need modify the error according to the new error standard.

disksing (approved):

server/schedule/operator/create_operator.go
nolouch: cc @HunDunDM

HunDunDM: I think the `errors` under `operator` should be uniformly modified in another PR.

nolouch: Does the other errors do not meet the error code standard code?

HunDunDM: Yes, so this part of the modification requires cherry-pick.

nolouch (approved):
lgtm

PR #3155, opened 2020-11-04
schedule: cache the unprocessed region due to limit

rleungx:

Use a LRU cache to save the unprocessed regions due to the limit restriction.

This PR adds a `regionWaitingList` in `CheckerController`.
To make it more clear, we mainly have two kinds of situations that need to add the region into the `regionWaitingList`:
1. when the operator is created but the schedule limit reaches the threshold
2. the store limit inside checker reaches the threshold

### Check List

Tests

- Unit test

### DIFF
--- a/server/cluster/cluster_stat.go
+++ b/server/cluster/cluster_stat.go
@@ -18,8 +18,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/tikv/pd/pkg/movingaverage"
-
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/movingaverage"
--- a/server/cluster/cluster_stat.go
+++ b/server/cluster/cluster_stat.go
@@ -18,6 +18,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/tikv/pd/pkg/movingaverage"
+
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/movingaverage"
--- a/server/schedule/checker_controller.go
+++ b/server/schedule/checker_controller.go
@@ -46,16 +46,17 @@ type CheckerController struct {
 // NewCheckerController create a new CheckerController.
 // TODO: isSupportMerge should be removed.
 func NewCheckerController(ctx context.Context, cluster opt.Cluster, ruleManager *placement.RuleManager, opController *OperatorController) *CheckerController {
+	regionWaitingList := cache.NewCache(DefaultCacheSize, cache.DefaultCacheType)
 	return &CheckerController{
 		cluster:           cluster,
 		opts:              cluster.GetOpts(),
 		opController:      opController,
 		learnerChecker:    checker.NewLearnerChecker(cluster),
-		replicaChecker:    checker.NewReplicaChecker(cluster),
+		replicaChecker:    checker.NewReplicaChecker(cluster, regionWaitingList),
 		ruleChecker:       checker.NewRuleChecker(cluster, ruleManager),
 		mergeChecker:      checker.NewMergeChecker(ctx, cluster),
 		jointStateChecker: checker.NewJointStateChecker(cluster),
-		regionWaitingList: cache.NewCache(DefaultCacheSize, cache.DefaultCacheType),
+		regionWaitingList: regionWaitingList,
 	}
 }
 
--- a/server/schedule/checker_controller.go
+++ b/server/schedule/checker_controller.go
@@ -16,6 +16,7 @@ package schedule
 import (
 	"context"
 
+	"github.com/tikv/pd/pkg/cache"
 	"github.com/tikv/pd/server/config"
 	"github.com/tikv/pd/server/core"
 	"github.com/tikv/pd/server/schedule/checker"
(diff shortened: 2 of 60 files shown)


HunDunDM (approved):

lhy1024 (commented):
LGTM. will we merge `checkSuspectRegion` and `checkWaitingRegions`？And will we use different cache according to different checkers?

BusyJay:
Any progress?

disksing:
@BusyJay  test result not good enough now, we are still try to figure it out.

server/cluster/cluster_stat.go
```diff
@@ -14,11 +14,12 @@
 package cluster
 
 import (
-	"github.com/tikv/pd/pkg/movingaverage"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/tikv/pd/pkg/movingaverage"
+
```
lhy1024, line 22: unnecessary empty line?

nolouch:
Does this PR ready to review？

rleungx:
> Does this PR ready to review？

You can review it right now

BusyJay:
PTAL

server/schedule/checker_controller.go
```diff
@@ -34,64 +38,80 @@ type CheckerController struct {
...
 	return &CheckerController{
 		cluster:           cluster,
 		opts:              cluster.GetOpts(),
 		opController:      opController,
 		learnerChecker:    checker.NewLearnerChecker(cluster),
-		replicaChecker:    checker.NewReplicaChecker(cluster),
-		ruleChecker:       checker.NewRuleChecker(cluster, ruleManager),
+		replicaChecker:    checker.NewReplicaChecker(cluster, regionWaitingList),
+		ruleChecker:       checker.NewRuleChecker(cluster, ruleManager, regionWaitingList),
 		mergeChecker:      checker.NewMergeChecker(ctx, cluster),
```
lhy1024, line 55: will we add `regionWaitingList` for `mergeChecker`?

server/schedule/checker_controller.go
```diff
@@ -34,64 +38,80 @@ type CheckerController struct {
...
 		jointStateChecker: checker.NewJointStateChecker(cluster),
+		regionWaitingList: regionWaitingList,
 	}
 }
 
 // CheckRegion will check the region and add a new operator if needed.
-func (c *CheckerController) CheckRegion(region *core.RegionInfo) (bool, []*operator.Operator) { //return checkerIsBusy,ops
+func (c *CheckerController) CheckRegion(region *core.RegionInfo) []*operator.Operator { //return checkerIsBusy,ops
```
lhy1024, line 62: why to remove `CheckerIsBusy`?

server/schedule/checker_controller.go
rleungx: No, for now, we don't save the regions which need to be merged into the cache.

server/schedule/checker_controller.go
rleungx: Since it is not used anymore

server/schedule/checker_controller.go
```diff
@@ -34,64 +38,80 @@
...
 		}
-		if opController.OperatorCount(operator.OpReplica) < c.opts.GetReplicaScheduleLimit() {
-			checkerIsBusy = false
-			if op := c.replicaChecker.Check(region); op != nil {
-				return checkerIsBusy, []*operator.Operator{op}
+		if op := c.replicaChecker.Check(region); op != nil {
+			if opController.OperatorCount(operator.OpReplica) < c.opts.GetReplicaScheduleLimit() {
+				return []*operator.Operator{op}
 			}
+			c.regionWaitingList.Put(region.GetID(), nil)
```
HunDunDM, line 86: It seems that it can be put inside the checker or outside the checker. Can you explain the design?

HunDunDM (commented):
rest LGTM

server/schedule/operator_controller.go
```diff
@@ -123,10 +123,12 @@ func (oc *OperatorController) Dispatch(region *core.RegionInfo, source string) {
 		case operator.SUCCESS:
 			oc.pushHistory(op)
 			if oc.RemoveOperator(op) {
+				operatorWaitCounter.WithLabelValues(op.Desc(), "promote-success").Inc()
```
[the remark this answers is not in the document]
rleungx: They don't have any relationship.

server/schedule/operator_controller.go
```diff
@@ -298,6 +303,7 @@ func (oc *OperatorController) AddWaitingOperator(ops ...*operator.Operator) int
 	}
 
 	oc.Unlock()
+	operatorWaitCounter.WithLabelValues(ops[0].Desc(), "promote-add").Inc()
```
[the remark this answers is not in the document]
rleungx: ditto

server/schedule/checker_controller.go
lhy1024: So will it affect the speed of merge?

server/schedule/checker_controller.go
lhy1024: then we should remove the comment.

server/schedule/checker_controller.go
lhy1024: Maybe will it be better to provide a cache for each checker

Yisaer (approved):
LGTM

Issue #3140, opened 2020-11-02
Compatible with replication setting  after placement rule is enabled


nolouch:
## Feature request
We can set the replication config with `pdctl` if the placement rules is disabled, like:
```
>> pd-ctl config set max-replicas 1
successful!
```
But if we enable the placement rules, the replication of the other config will become invalid:
```
>> pd-ctl config set enable-placement-rules true
>> pd-ctl config set max-replicas 1
Failed to set config: [400] "cannot update MaxReplicas when placement rules feature is enabled, please update rule instead"
```
This issue is want to let `config set` also work if the placement rules only have one default rule.

duduainankai:
Sorry for the slow response.

Does  `placement rules only have one default rule`  means we have one placement rule and the `StarKey` and `EndKey` are both nil ?

And I am supposed to update the `count` of rule when update `max-replicas` which is missing from current mr.

nolouch:
already finished.

PR #3163, opened 2020-11-06
server: enable replication setting when only have one default placeme…

duduainankai:
…nt rule

Signed-off-by: Zheming Li <nkdudu@126.com>

#3140

- When update max replica and placement rule enabled, check if have only one default rule.

### Check List

Tests

- Unit test

### Release note

- No release note


server/server.go
```diff
@@ -857,6 +858,11 @@ func (s *Server) SetReplicationConfig(cfg config.ReplicationConfig) error {
 	return nil
 }
 
+func (s *Server) isOnlyDefaultRule() bool {
+	rules := s.GetRaftCluster().GetRuleManager().GetAllRules()
+	return len(rules) == 1 && rules[0].GroupID == "pd" && rules[0].ID == "default"
```
rleungx, line 863: What if we overwrite this default rule?

tests/pdctl/config/config_test.go
```diff
@@ -567,3 +563,78 @@ func (s *configTestSuite) TestReplicationMode(c *C) {
...
+	c.Assert(err, IsNil)
+	ioutil.WriteFile(fname, b, 0644)
+	_, _, err = pdctl.ExecuteCommandC(cmd, "-u", pdAddr, "config", "placement-rules", "save", "--in="+fname)
+	c.Assert(err, IsNil)
+
+	_, output, err = pdctl.ExecuteCommandC(cmd, "-u", pdAddr, "config", "set", "max-replicas", "4")
+	c.Assert(err, IsNil)
+	c.Assert(strings.Contains(string(output), "please update rule instead"), IsTrue)
+	checkMaxReplicas(3)
+
```
rleungx, line 639: Please remove the empty line.

server/server.go
duduainankai: So what's the definition of 'default rule' ? 

I thought it was only defined by GroupID and ID.

disksing:
em, I think you need not only allow setting the config but also need to sync the new configuration with the default rule.

Yisaer:
I think the origin issue could be described like: 
```
>> pd-ctl config set max-replicas <group-id> <rule-id> 1
successful!
```

so that the pd-ctl could edit max-relicas for each placement rule. WDYT @nolouch

nolouch:
@Yisaer This is an extension, but I hope it can be compatible with the original command.

nolouch:
Hi, @duduainankai , ci failed.
```
FAIL: cluster_test.go:47: testClusterSuite.TestCluster

cluster_test.go:63:
    // Cannot set the replicas when placement rules is enabled.
    c.Assert(s.svr.SetReplicationConfig(r), NotNil)
... value = nil
```

tests/pdctl/config/config_test.go
```diff
@@ -567,3 +563,78 @@ func (s *configTestSuite) TestReplicationMode(c *C) {
...
+			GroupID: "pd",
+			ID:      "test1",
+			Role:    "voter",
+			Count:   1,
+		},
+	}
+	b, err := json.Marshal(rules)
+	c.Assert(err, IsNil)
+	ioutil.WriteFile(fname, b, 0644)
+	_, _, err = pdctl.ExecuteCommandC(cmd, "-u", pdAddr, "config", "placement-rules", "save", "--in="+fname)
```
ZenoTan, line 632: Also check replicas here?

duduainankai:
@nolouch @ZenoTan Got

But I still try to understand the comments:

> What if we overwrite this default rule?

> em, I think you need not only allow setting the config but also need to sync the new configuration with the default rule.

and a little busy these days, maybe need more time, sorry about that :( .

server/api/cluster_test.go
```diff
@@ -56,12 +56,8 @@ func (s *testClusterSuite) TestCluster(c *C) {
 
 	c2 := &metapb.Cluster{}
 	r := config.ReplicationConfig{
-		EnablePlacementRules: true,
```
rleungx, line 59: Why remove it?

duduainankai: This is the "only one default rule when enable placement rules", so `c.Assert(s.svr.SetReplicationConfig(r), NotNil)` is no longer correct here.

And the similar test is already done in `config_test`, so I think just remove the code to simply fix the test is ok.

rleungx (approved):
Overall, LGTM, but there is another case: once we change the default rule, and then we disable the placement rule, the original max replicas will not match.

nolouch (approved):
LGTM. thanks

PR #3178, opened 2020-11-13
client: add max error retry time for client

glorv:

The current implementation will retry upto 100 times if connect meets error. Thus, if use misconfig the pd address, the `pd.NewClient` will return an error after 100s which is too slow and seems like a bug in the view of a user.

See: https://github.com/pingcap/tidb-lightning/issues/462

Add a config `WithMaxErrorRetry` to allow pd client custom connect max retry time when meets error.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note
- Support config client max error retry times


glorv:
@3pointer @overvenus PTAL

overvenus:
LGTM

JmPotato:
@glorv Please update the branch to merge this PR.

PR #3179, opened 2020-11-13
tso: make TSO unique for each dc

### LINKED ISSUE #3056: Implement the differentiation mechanism for local TSO

Two different Local TSO Allocator leaders may generate two same local TSOs now. This may cause some potential problems when we read the data with a specified TSO. We need to make every TSO generated unique.

We need to implement a differentiation mechanism. We can firstly get the DC number by reading the `dc-location` in the cluster and set the low bits of TSO's logical part to distinguish them. For example, if we have dc-1, dc-2, and dc-3. We will make the TSO's low bits of logical part from each DC looks like:

* dc-1: xxxxxxxxxxxxxxxx00
* dc-2: xxxxxxxxxxxxxxxx01
* dc-3: xxxxxxxxxxxxxxxx10

Also, we need to handle both the `maxLogical` and the number of bits well during the shifting to avoid break the linear consistency.

JmPotato:

Resolve #3056. Related PR: pingcap/kvproto#701.

### Check List

Tests

- Unit test
- Integration test

Side effects

- Increased code complexity


server/tso/local_allocator.go
```diff
@@ -99,8 +101,21 @@ func (lta *LocalTSOAllocator) SetTSO(tso uint64) error {
...
 // Make sure you have initialized the TSO allocator before calling.
 func (lta *LocalTSOAllocator) GenerateTSO(count uint32) (pdpb.Timestamp, error) {
-	// Todo: use the low bits of TSO's logical part to distinguish the different local TSO
-	return lta.timestampOracle.getTS(lta.leadership, count)
+	dcLocationMap := lta.allocatorManager.GetClusterDCLocations()
+	dcLocationList := make([]string, 0, len(dcLocationMap))
+	for dcLocation := range dcLocationMap {
+		dcLocationList = append(dcLocationList, dcLocation)
+	}
+	sort.Strings(dcLocationList)
```
JmPotato, line 109: Maybe update this with `clusterDCLocations`.

server/tso/allocator_manager.go
```diff
@@ -451,7 +472,9 @@ func (am *AllocatorManager) ClusterDCLocationChecker() {
 			continue
 		}
 		am.mu.clusterDCLocations[dcLocation] = append(am.mu.clusterDCLocations[dcLocation], serverID)
+		am.mu.sortedDCLocations = append(am.mu.sortedDCLocations, dcLocation)
```
ZenoTan, line 475: Will this cause tso duplication when the configuration changes?

JmPotato: Added a check. PTAL.

zhangjinpeng1987:
How about adding new nodes in a new DC dynamically?

JmPotato:
> How about adding new nodes in a new DC dynamically?

`clusterDCLocations` is dynamically changing according to the cluster status. So we can make sure all PD nodes will know this change in current codes.

nolouch (commented):
If dynamically add a new DC PD member, is it possible to allocate the same TSO? sunch as:
```
1:    DC-B    ->  time1...1[00]
      DC-C    ->  time1...1[01]
# add DC-A member
2:   DC-A    ->  time1...1[00]  allocated.
     DC-B    ->  time1...2[01]
     DC-C    ->  time1...2[10]

```

JmPotato:
> 01

Any good idea to handle this case? The first solution that comes to my mind is to order the `dc-location` by the join time rather than the lexicographic order if etcd provides a strong consistency.

JmPotato:
> If dynamically add a new DC PD member, is it possible to allocate the same TSO? such as:
> 
> ```
> 1:    DC-B    ->  time1...1[00]
>       DC-C    ->  time1...1[01]
> # add DC-A member
> 2:   DC-A    ->  time1...1[00]  allocated.
>      DC-B    ->  time1...2[01]
>      DC-C    ->  time1...2[10]
> ```

Any good idea to handle this case? The first solution that comes to my mind is to order the dc-location by the join time rather than the lexicographic order if etcd provides a strong consistency.

nolouch:
Or sync global tso once when there is a member change? cc @disksing

server/tso/tso.go
```diff
@@ -83,18 +83,39 @@ func (t *timestampOracle) getTSO() (time.Time, int64) {
...
+// Local TSO to do some differentiating work. For example, we have three DCs: dc-1, dc-2 and
+// dc-3. So the shiftNum is 2 bits because we have three DCs + a global scope, and 2 bits are
+// enough to distinguish them. Then, for dc-2, the serialNum is 1 because its index in all
+// sorted dc-locations is 1.
+// Once we get a noramal TSO like this (18 bits): xxxxxxxxxxxxxxxxxx. We will make the TSO's
+// low bits of logical part from each DC looks like:
+//     dc-1: xxxxxxxxxxxxxxxx00
+//     dc-2: xxxxxxxxxxxxxxxx01
+//     dc-3: xxxxxxxxxxxxxxxx10
+//   global: xxxxxxxxxxxxxxxx11
```
Yisaer, line 110: Is it better to let `global tso` have `00` suffix?

server/tso/global_allocator.go
```diff
@@ -129,6 +127,14 @@ func (gta *GlobalTSOAllocator) GenerateTSO(count uint32) (pdpb.Timestamp, error)
 		currentGlobalTSO pdpb.Timestamp
 		err              error
 	)
+	// Differentiate the Global TSO synced from all the other Local TSOs
+	serialNum := gta.allocatorManager.getSortedDCLocationLength()
+	// log2(the number of dc-locations + global)
+	shiftNum := int(math.Ceil(math.Log2(float64(serialNum + 1))))
```
Yisaer, line 133: I tend to use a fixed `shiftNum` instead of a dynamic one. What about set it as `8`? I think `2^8` can hold most of the situations.

server/tso/allocator_manager.go
```diff
@@ -451,6 +476,27 @@ func (am *AllocatorManager) ClusterDCLocationChecker() {
...
 		am.mu.clusterDCLocations[dcLocation] = append(am.mu.clusterDCLocations[dcLocation], serverID)
+		// Only the PD leader can check and update the sortedDCLocations
+		if am.member.IsStillLeader() {
+			// Add the new dc-location found from the etcd into the sortedDCLocations
+			if slice.NoneOf(am.mu.sortedDCLocations, func(i int) bool {
+				return am.mu.sortedDCLocations[i] == dcLocation
+			}) {
+				am.mu.sortedDCLocations = append(am.mu.sortedDCLocations, dcLocation)
+			}
+		}
```
Yisaer, line 487: After `transfer pd leader`, can we maintain the order like before? I think this will be quite useful.

Yisaer (commented):
rest LGTM

Yisaer:
After discussed with @JmPotato , we are going to persist the map order index for each dc when the dc get persisted into etcd instead of on-demanding calculating.

server/tso/allocator_manager.go
```diff
@@ -338,6 +396,18 @@ func (am *AllocatorManager) campaignAllocatorLeader(loopCtx context.Context, all
 	}
 }
 
+func (am *AllocatorManager) getLocalTSOSuffix(dcLocation string) (int, error) {
+	resp, err := etcdutil.EtcdKVGet(am.member.Client(), am.GetLocalTSOSuffixPath(dcLocation))
+	if err != nil {
+		return -1, err
+	}
+	suffix, err := strconv.ParseInt(string(resp.Kvs[0].Value), 10, 32)
```
rleungx, line 404: Is there possible `len(resp.Kvs) == 0`?

server/tso/allocator_manager.go
```diff
@@ -150,10 +146,57 @@ func (am *AllocatorManager) SetLocalTSOConfig(localTSOConfig config.LocalTSOConf
...
+		// The Local TSO suffix is determined by the joining order of this dc-location.
+		localTSOSuffixValue := strconv.FormatInt(int64(len(resp.Kvs)), 10)
+		resp, err := kv.NewSlowLogTxn(client).
+			If(clientv3.Compare(clientv3.CreateRevision(localTSOSuffixKey), "=", 0)).
+			Then(clientv3.OpPut(localTSOSuffixKey, localTSOSuffixValue)).
+			Commit()
+		if err != nil {
+			return errs.ErrEtcdTxn.Wrap(err).GenWithStackByCause()
+		}
+		// If the etcd txn failed, it means there is already a suffix written previously.
```
rleungx: What if txn failed due to some other reasons?

tests/server/tso/manager_test.go
```diff
@@ -92,6 +96,71 @@ func (s *testManagerSuite) TestClusterDCLocations(c *C) {
...
+	for serverName, server := range cluster.GetServers() {
+		tsoAllocatorManager := server.GetTSOAllocatorManager()
+		dcLocation := testCase.dcLocationConfig[serverName]
+		suffixResp, err := etcdutil.EtcdKVGet(
+			cluster.GetEtcdClient(),
+			tsoAllocatorManager.GetLocalTSOSuffixPath(dcLocation))
+		c.Assert(err, IsNil)
+		c.Assert(len(suffixResp.Kvs), Equals, 1)
+		// Test the increment of the suffix
+		allSuffixResp, err := etcdutil.EtcdKVGet(
```
rleungx, line 147: No need to do the test in the loop?

rleungx:
Can you add a join test?

JmPotato:
> Can you add a join test?

Ok, I'm working on it.

server/grpc_service.go
```diff
@@ -1034,23 +1033,12 @@ func (s *Server) GetDCLocations(ctx context.Context, request *pdpb.GetDCLocation
 	if err := s.validateInternalRequest(request.GetHeader(), false); err != nil {
 		return nil, err
 	}
-	if !s.member.IsLeader() {
+	if !s.member.IsStillLeader() {
```
nolouch, line 1036: why rename this?

server/tso/allocator_manager.go
```diff
@@ -145,10 +156,57 @@ func (am *AllocatorManager) SetLocalTSOConfig(localTSOConfig config.LocalTSOConf
 			zap.Uint64("server-id", serverID))
 		return errs.ErrEtcdTxn.FastGenByArgs()
 	}
-	am.ClusterDCLocationChecker()
+	log.Info("write dc-location into etcd",
```
nolouch, line 159:
```suggestion
	log.Info("write dc-location configuration into etcd",
```

server/tso/tso.go
```diff
@@ -39,6 +39,8 @@ const (
 	// When a TSO's logical time reaches this limit,
 	// the physical time will be forced to increase.
 	maxLogical = int64(1 << 18)
+	// SuffixBits indicates the number of suffix bits
+	SuffixBits = 8
```
nolouch, line 43: we already meet the problem about:
image
assume we need 1000,000 QPS, that's mean 1000 ts need to be allocated on average every millisecond, the buffer of logical part here is obviously not enough

server/tso/allocator_manager.go
```diff
@@ -145,10 +154,53 @@ func (am *AllocatorManager) SetLocalTSOConfig(localTSOConfig config.LocalTSOConf
...
+		clientv3.WithPrefix())
+	if err != nil {
+		return clusterDCLocations, err
+	}
+	clusterDCLocations = make(map[string][]uint64)
+	for _, kv := range resp.Kvs {
+		// The key will contain the member ID and the value is its dcLocation
+		serverPath := strings.Split(string(kv.Key), "/")
+		dcLocation := string(kv.Value)
+		// Get serverID from serverPath, e.g, /pd/dc-location/1232143243253 -> 1232143243253
```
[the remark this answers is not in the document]
JmPotato: Theoretically, `IsStillLeader` is more accurate than `IsLeader` to check the leader role.

server/grpc_service.go
JmPotato: Theoretically, `IsStillLeader` is more accurate than `IsLeader` to check the leader role.

server/tso/allocator_manager.go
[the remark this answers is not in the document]
JmPotato: This comment is for the method of getting `serverID` from `serverPath`. The origin one is correct.

server/tso/allocator_manager.go
```diff
@@ -75,6 +77,9 @@ type AllocatorManager struct {
 		allocatorGroups map[string]*allocatorGroup
 		// dc-location/global (string) -> Member ID
 		clusterDCLocations map[string][]uint64
+		// suffixDCLocations is a map of the dc-locations to its TSO suffix.
+		// dc-location (string) -> Suffix number (int)
```
Yisaer, line 81:
```suggestion
		// dc-location (string) -> Suffix sign (int)
```

server/tso/allocator_manager.go
```diff
@@ -430,33 +505,95 @@ func (am *AllocatorManager) allocatorPatroller(serverCtx context.Context) {
...
+		if suffix > maxSuffixNum {
+			maxSuffixNum = suffix
 		}
-		am.mu.clusterDCLocations[dcLocation] = append(am.mu.clusterDCLocations[dcLocation], serverID)
 	}
+	maxSuffixNum++
+	localTSOSuffixKey := am.GetLocalTSOSuffixPath(dcLocation)
+	// The Local TSO suffix is determined by the joining order of this dc-location.
+	localTSOSuffixValue := strconv.FormatInt(maxSuffixNum, 10)
+	// Todo: clean up the useless dc-location suffix
```
Yisaer, line 569: We may don't need to clean the useless dc-location suffix.

server/tso/global_allocator.go
```diff
@@ -121,6 +123,14 @@ func (gta *GlobalTSOAllocator) GenerateTSO(count uint32) (pdpb.Timestamp, error)
...
 	}
 	maxTSO.Logical += int64(count)
+	maxTSO.Logical = gta.timestampOracle.differentiateLogical(maxTSO.Logical)
+	// If the maxTSO's logical part is bigger than maxLogical, just add a updateTimestampGuard
+	// to the physical time and empty the logical part. We just need to make sure it's bigger than
+	// all the other Local TSOs.
+	if maxTSO.GetLogical() > maxLogical {
+		maxTSO.Physical += updateTimestampGuard.Milliseconds()
+		maxTSO.Logical = 0
+	}
```
Yisaer, line 133: we still need to sign the suffix here.

server/tso/allocator_manager.go
```diff
@@ -75,6 +77,9 @@ type AllocatorManager struct {
 		allocatorGroups map[string]*allocatorGroup
 		// dc-location/global (string) -> Member ID
 		clusterDCLocations map[string][]uint64
+		// suffixDCLocations is a map of the dc-locations to its TSO suffix.
+		// dc-location (string) -> Suffix number (int)
+		suffixDCLocations map[string]int32
```
Yisaer, line 82: Can we merge `clusterDCLocations` and `suffixDCLocations` as one to prevent inconsistency here?

Yisaer (commented):
Rest LGTM

server/tso/global_allocator.go
JmPotato: The Global TSO's suffix will always be zero, so it's no need to do the differentiate here.

server/tso/tso.go
JmPotato: Making the bits of suffix change dynamicly may alleviates this problem. but can not to solve it. Do you have any idea?

nolouch: we can make it be dynamic, but we also limit the upper bits.

JmPotato: I have already made it be dynamic and also limit the upper to the 2^8 now. PTAL.

server/tso/global_allocator.go
Yisaer: Got it. what about leave a comment to explain it here?

server/tso/global_allocator.go
```diff
@@ -35,7 +35,7 @@ type Allocator interface {
 	// Initialize is used to initialize a TSO allocator.
 	// It will synchronize TSO with etcd and initialize the
 	// memory for later allocation work.
-	Initialize() error
+	Initialize(suffix int) error
```
Yisaer, line 38:
```suggestion
	Initialize(sign int) error
```

Yisaer (commented):
LGTM

server/tso/tso.go
```diff
@@ -39,6 +39,8 @@ const (
 	// When a TSO's logical time reaches this limit,
 	// the physical time will be forced to increase.
 	maxLogical = int64(1 << 18)
+	// MaxSuffixBits indicates the max number of suffix bits.
+	MaxSuffixBits = 8
```
nolouch, line 43:
```suggestion
	MaxSuffixBits = 4
```

server/tso/allocator_manager.go
```diff
@@ -430,33 +516,108 @@ func (am *AllocatorManager) allocatorPatroller(serverCtx context.Context) {
...
+	txnResp, err := kv.NewSlowLogTxn(am.member.Client()).
+		If(clientv3.Compare(clientv3.CreateRevision(localTSOSuffixKey), "=", 0)).
+		Then(clientv3.OpPut(localTSOSuffixKey, localTSOSuffixValue)).
+		Commit()
+	if err != nil {
+		return -1, errs.ErrEtcdTxn.Wrap(err).GenWithStackByCause()
+	}
+	// If the etcd txn failed, it means there is already a suffix written previously.
+	if !txnResp.Succeeded {
+		log.Warn("write local tso suffix into etcd failed",
```
nolouch, line 603: Will print this log periodically？

server/tso/local_allocator.go
```diff
@@ -99,8 +96,8 @@ func (lta *LocalTSOAllocator) SetTSO(tso uint64) error {
 // GenerateTSO is used to generate a given number of TSOs.
 // Make sure you have initialized the TSO allocator before calling.
 func (lta *LocalTSOAllocator) GenerateTSO(count uint32) (pdpb.Timestamp, error) {
-	// Todo: use the low bits of TSO's logical part to distinguish the different local TSO
-	return lta.timestampOracle.getTS(lta.leadership, count)
+	return lta.timestampOracle.getTS(
+		lta.leadership, count, lta.allocatorManager.GetClusterDCLocationsNumber())
```
nolouch, line 100: Should get the suffix of the dc locator? this is the total number?

server/tso/allocator_manager.go
JmPotato: It will print the log periodically only if the previous writing fails.

server/tso/local_allocator.go
JmPotato: This is the total number.

server/tso/tso.go
JmPotato: I will update it with the `go.mod` file after merging the pingcap/kvproto#701.

server/tso/local_allocator.go
nolouch: So, Is the suffixBits dynamically change?

server/tso/allocator_manager.go
nolouch:
```
// If the etcd txn failed, it means there is already a suffix written previously.
```
I think the comment says the previous write is successful.

JmPotato: Fixed.

server/tso/local_allocator.go
JmPotato: Yes, it will be calculated by the `tso.CalSuffixBits` method according to the given `dc-location` number.

nolouch (commented):
LGTM. Please fire an issue about the corner case with member change.

nolouch (approved):

JmPotato:
> LGTM. Please fire an issue about the corner case in member change.

It's here: #3260.

Yisaer (approved):
LGTM

PR #3188, opened 2020-11-19
client: fix the nil interface conversion bug

JmPotato:

A bug found from pingcap/tidb#20980. There is a certain chance that when a TSO request is received, the corresponding DC's `tsDeadline` is not ready.

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/tidb_ghpr_check_2/detail/tidb_ghpr_check_2/61998/pipeline/49/

Manually check the `tsDeadline` when it's not ready.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -179,6 +179,8 @@ type client struct {
 	tsDeadline sync.Map // Same as map[string]chan deadline
 	// dc-location -> *lastTSO
 	lastTSMap sync.Map // Same as map[string]*lastTSO
+
+	checkTSDeadlineCh chan struct{}
 }
 
 // NewClient creates a PD client.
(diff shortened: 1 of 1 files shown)


lhy1024 (approved):

client/client.go
```diff
@@ -179,6 +179,8 @@ type client struct {
 	tsDeadline sync.Map // Same as map[string]chan deadline
 	// dc-location -> *lastTSO
 	lastTSMap sync.Map // Same as map[string]*lastTSO
+
+	checkTSDeadlineCh chan struct{}
```
HunDunDM, line 183: It seems more reasonable to use `context`.

Yisaer (approved):
LGTM

client/client.go
Yisaer: why?

JmPotato: May I ask why? The channel is more easy to use in this case.

HunDunDM: Sorry, I read something wrong.

HunDunDM (approved):

PR #3192, opened 2020-11-19
Update `average over time`

lhy1024:

Previously, we filled the aot and then emptied it, which not only wasted some data, but also increased latency.

Now we will save the data, while continuing to take the data in the aot size, and for the extra margin in the first data in the queue, we will make the cut.

So we can translate the problem into another one: Taking the flow data from the heartbeat, imagining it as a histogram with area `delta` and width `interval`, and then take the area of `aot_size` and calculate their heights.

image

It also a part of #3201 

### Check List

Tests

- Unit test


HunDunDM (approved):

disksing (approved):

nolouch:
Does need pick to  4.0？

lhy1024:
> Does need pick to 4.0？

I think it would be better to merge it with other statistical modifications.

disksing:
CI broken. Use super power.

PR #3200, opened 2020-11-24
server: check the store version when changing tombstone status

HunDunDM:

* When changing a `Tombstone` store to `Up` or `Offline`, if the `cluster-version` has been upgraded at this time, the operation should fail.

* The store of lower version is no longer allowed to change from `Tombstone` back to `Up`.
* Fix the bug that restarting the store will cause store-limit and statistics to be reset.

### Check List

Tests

- Unit test

Related changes

### Release note

- The store of lower version is no longer allowed to change from `Tombstone` back to `Up`.
- Fix the bug that restarting the store will cause store-limit and statistics to be reset.

### DIFF
--- a/server/statistics/store.go
+++ b/server/statistics/store.go
@@ -45,7 +45,9 @@ func NewStoresStats() *StoresStats {
 func (s *StoresStats) CreateRollingStoreStats(storeID uint64) {
 	s.Lock()
 	defer s.Unlock()
-	s.rollingStoresStats[storeID] = newRollingStoreStats()
+	if _, ok := s.rollingStoresStats[storeID]; !ok {
+		s.rollingStoresStats[storeID] = newRollingStoreStats()
+	}
 }
 
 // RemoveRollingStoreStats removes RollingStoreStats with a given store ID.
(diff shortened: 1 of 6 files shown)


server/statistics/store.go
```diff
@@ -45,7 +45,9 @@ func NewStoresStats() *StoresStats {
 func (s *StoresStats) CreateRollingStoreStats(storeID uint64) {
 	s.Lock()
 	defer s.Unlock()
-	s.rollingStoresStats[storeID] = newRollingStoreStats()
+	if _, ok := s.rollingStoresStats[storeID]; !ok {
```
lhy1024, line 48: we should remove `CreateRollingStoreStats`，only to use `GetOrCreateRollingStoreStats`

lhy1024:
It also is a part of #3201

disksing (approved):

lhy1024 (approved):

server/statistics/store.go
HunDunDM: lgtm

PR #3212, opened 2020-11-25
hot scheduler: remove unused code

disksing:

As `hotPeerFilterType` is always `mixed`, we can remove it.
`hotRegionThreshold` is only used when `hotPeerFilterType` is not `mixed` so we can remove it too.

Removed unused code to make it easier to read. We can add it back in the future if we want it again.

### Check List

Tests
- Unit test

### Release note
- No release note


lhy1024:
Previously, we wanted to use this strategy but then gave up.

PTAL @rleungx  @nolouch

lhy1024 (approved):

HunDunDM (approved):

PR #3224, opened 2020-11-28
server/auth: implement rbac storage, caching, managing and enforcing logic

PhotonQuantum:

server/auth: implement rbac storage, caching, managing and enforcing logic.

A new kvBase interface `kv.TxnBase` and also its etcd implementation is added to handle kv transactions.

server/auth `Manager` is implemented to handle storage and caching logic of users & roles info. It can be used to manage users, roles and permissions, and validate requests.

### Check List

Tests

- Unit test
- Integration test

Code changes

### Release note

No release note

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -1,0 +1,712 @@
+// Copyright 2020 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissionKeys and
+// limitations under the License.
+
+package auth
+
+import (
+	"encoding/json"
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
+	"go.uber.org/zap"
+	"path"
+	"strings"
+	"sync"
+
+	"github.com/pingcap/errors"
+	"github.com/tikv/pd/server/kv"
+	"go.etcd.io/etcd/clientv3"
+)
+
+var (
+	ErrorUserNotFound    = func(name string) error { return errors.Errorf("user not found: %s", name) }
+	ErrorUserExists      = func(name string) error { return errors.Errorf("user already exists: %s", name) }
+	ErrorRoleNotFound    = func(name string) error { return errors.Errorf("role not found: %s", name) }
+	ErrorRoleExists      = func(name string) error { return errors.Errorf("role already exists: %s", name) }
+	ErrorUserHasRole     = func(user string, role string) error { return errors.Errorf("user %s already has role: %s", user, role) }
+	ErrorUserMissingRole = func(user string, role string) error {
+		return errors.Errorf("user %s doesn't have role: %s", user, role)
+	}
+	ErrorRoleHasPermission = func(role string, permission Permission) error {
+		return errors.Errorf("role %s already has permission: %s", role, permission.String())
+	}
+	ErrorRoleMissingPermission = func(role string, permission Permission) error {
+		return errors.Errorf("role %s doesn't have permission: %s", role, permission.String())
+	}
+)
+
+var (
+	userPrefix = "users"
+	rolePrefix = "roles"
+)
+
+type Manager struct {
+	kv    kv.TxnBase
+	users sync.Map // map[string]*User
+	roles sync.Map // map[string]*Role
+}
+
+func NewManager(k kv.TxnBase) *Manager {
+	return &Manager{kv: k}
+}
+
+func (m *Manager) GetUser(name string) (*User, error) {
+	_cachedUser, ok := m.users.Load(name)
+	// Cache hit, so we return cached value.
+	if ok {
+		cachedUser := _cachedUser.(*User)
+		// Nil in cache implies this user doesn't exist
+		if cachedUser == nil {
+			return nil, ErrorUserNotFound(name)
+		}
+		return cachedUser, nil
+	}
+
+	// Cache miss. Load value from etcd.
+	userJson, err := m.kv.Load(path.Join(userPrefix, name))
+	if err != nil {
+		return nil, err
+	}
+
+	var user *User
+	if userJson == "" {
+		user = nil
+	} else {
+		user, err = NewUserFromJson(userJson)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// If another (write) operation has already updated the cache,
+	// we abandon the data we've just retrieved and use the cached version.
+	_latestUser, _ := m.users.LoadOrStore(name, user)
+	latestUser := _latestUser.(*User)
+
+	if latestUser == nil {
+		return nil, ErrorUserNotFound(name)
+	}
+	return latestUser, nil
+}
+
+func (m *Manager) GetUsers() (map[*User]struct{}, error) {
+	// NOTE: GetUsers doesn't use the cache and read data from kv directly to avoid cache inconsistency.
+	usersPath := strings.Join([]string{userPrefix, ""}, "/")
+	_, _users, err := m.kv.LoadRange(usersPath, clientv3.GetPrefixRangeEnd(usersPath), 0)
+	if err != nil {
+		return nil, err
+	}
+
+	users := make(map[*User]struct{})
+	for _, userJson := range _users {
+		user, err := NewUserFromJson(userJson)
+		if err != nil {
+			return nil, err
+		}
+
+		users[user] = struct{}{}
+	}
+
+	return users, nil
+}
+
+func (m *Manager) GetUserKeys() (map[string]struct{}, error) {
+	// NOTE: GetRoleKeys doesn't use the cache and read data from kv directly to avoid cache inconsistency.
+	usersPath := strings.Join([]string{userPrefix, ""}, "/")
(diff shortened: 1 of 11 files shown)


JmPotato:
Could you please split this PR into different parts? Reviewing more than 2000+ lines of code changes in a single PR is kind of tough.

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+	"sync"
+
+	"github.com/pingcap/errors"
+	"github.com/tikv/pd/server/kv"
+	"go.etcd.io/etcd/clientv3"
+)
+
+var (
+	// ErrUserNotFound is error info for user not found.
+	ErrUserNotFound = func(name string) error { return errors.Errorf("user not found: %s", name) }
```
JmPotato, line 32: I think the errors defination should be placed in [here](https://github.com/tikv/pd/blob/master/pkg/errs/errno.go).

JmPotato, line 32: I think the error definition should be placed in [here](https://github.com/tikv/pd/blob/master/pkg/errs/errno.go).

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+	}
+)
+
+var (
+	userPrefix = "users"
+	rolePrefix = "roles"
+)
+
+// Manager is used for the rbac storage, cache, management and enforcing logic.
+type Manager struct {
```
JmPotato, line 61: I think we should use a more specified name here rather than the `Manager`. Maybe the `RBACManager`?

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+
+	// Cache miss. Load value from etcd.
+	userJSON, err := m.kv.Load(path.Join(userPrefix, name))
+	if err != nil {
+		return nil, err
+	}
+
+	var user *User
+	if userJSON == "" {
+		user = nil
```
JmPotato, line 93: I think this part of logic is unnecessary because the default value of use is already `nil`.

JmPotato:
Do we have any design documents or something?

PhotonQuantum:
> Could you please split this PR into different parts? Reviewing more than 2000+ lines of code changes in a single PR is kind of tough.

Maybe I could split the kv.TxnBase part to another PR? However, that part has only 300 lines, so I suspect it won't help much.

The remaining part contains user & role models and their management methods, which work as a whole. Maybe splitting it apart will make the code harder to review because they are closely related, and extracting part of them doesn't make much sense?

> Do we have any design documents or something?

There's an RFC draft written in Chinese. It has already been reviewed by @Yisaer.

https://docs.google.com/document/d/1yZ1pjUqJZAFl4-ynHMaA-IrVbLojXjJKxZPs1hYlPbs/edit?usp=sharing

server/auth/permission.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/pingcap/errors"
+)
+
+var (
+	// ErrInvalidAction is error info for invalid action.
+	ErrInvalidAction = func(action Action) error { return errors.Errorf("invalid action: %s", action) }
```
gregwebs, line 25: Why not just declare this as a function?

server/auth/util.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	patName = regexp.MustCompile("^([A-Za-z])[A-Za-z0-9_]*$")
+)
+
+var (
+	// ErrPasswordMismatch is error info for password mismatch.
+	ErrPasswordMismatch = errors.New("given password doesn't match the hash")
+	// ErrInvalidName is error info for invalid name.
+	ErrInvalidName = errors.New("invalid key name")
```
gregwebs, line 32: This message should include the required pattern

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+	_, err = m.kv.NewTxn().
+		If(kv.Eq(kv.Value(userPath), string(userJSON))).
+		Then(kv.OpRemove(userPath)).
+		Commit()
+	if err != nil {
+		return err
+	}
+
+	// Delete user from memory cache.
+	m.users.Store(name, (*User)(nil))
```
gregwebs, line 210: How does the memory cache interact with the PD leader changing? Will it get refreshed when the leader changes?

Yisaer (commented):
I guess we could split this request into 3 parts: 
1. `user`(including creating and querying)
2. `auth`
3. `permission`

server/auth/auth.go
Yisaer: I think we could name it as ` RBACManager` which contains `userManager` and `authManager`. Each Manger mange its own logic.

gregwebs:
@Yisaer this is an important change. There should be more design review ideally via the [RFC process](github.com/tikv/rfcs)?

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+	userPrefix = "users"
+	rolePrefix = "roles"
+)
+
+// Manager is used for the rbac storage, cache, management and enforcing logic.
+type Manager struct {
+	kv    kv.TxnBase
+	users sync.Map // map[string]*User
+	roles sync.Map // map[string]*Role
+}
```
Yisaer, line 65: I found it all the methods related to `rbac control` is managed by this Manager which makes it so large. As I recommended [here](https://github.com/tikv/pd/pull/3224/files#r533090691). We could name it as `rbacManager` use which contains userManager and roleManager.

The `userManger` manages the methods for the `users` including `querying from the memory` / `loading from the etcd` / `persisting into etcd` / `comparing the password` etc...

The `roleManager ` manages the methods for the `roles`  including `querying from the memory` / `loading from the etcd` / `persisting into etcd` etc...

server/auth/user.go
```diff
@@ -0,0 +1,138 @@
...
+	username string
+	hash     string
+	roleKeys map[string]struct{}
+}
+
+type jsonUser struct {
+	Username string   `json:"username"`
+	Hash     string   `json:"hash"`
+	RoleKeys []string `json:"roles"`
+}
```
Yisaer, line 33: We can maintain the model in a certain file, and directly define the json tag in `User`

server/auth/role.go
```diff
@@ -0,0 +1,123 @@
...
+// Read-Only once created.
+type Role struct {
+	name        string
+	permissions map[Permission]struct{}
+}
+
+type jsonRole struct {
+	Name        string       `json:"name"`
+	Permissions []Permission `json:"permissions"`
+}
```
Yisaer, line 31: ditto.

server/auth/permission.go
```diff
@@ -0,0 +1,111 @@
...
+// Permission represents a permission to a specific pair of resource and action.
+type Permission struct {
+	resource string
+	action   Action
+}
+
+type jsonPermission struct {
+	Resource string `json:"resource"`
+	Action   Action `json:"action"`
+}
```
Yisaer, line 49: ditto.

Yisaer (commented):
Please address the comment and split the request by  `user` / `role` / `permission`.

Yisaer:
> RFC process

Thanks for the reminding. We will soon request a rfc document about this change. @gregwebs

gregwebs:
The goal of this change is to add RBAC for the PD resources?

BTW, I have an auth PR focused around adding [auth to TiKV](https://github.com/tikv/rfcs/pull/46). I think it is orthogonal and works together with this change. In my RFC I tried to avoid the subject of RBAC but did need an admin role and pre-seeding of a root and TiKV user that leverages mTLS. Perhaps all that could be done in this PR.

Yisaer:
> The goal of this change is to add RBAC for the PD resources?

Right.

The original goal is to control the PD Resources so that the different applications could hod the different auth to interact with PD without affecting each other.

gregwebs:
> The original goal is to control the PD Resources so that the different applications could hod the different auth to interact with PD without affecting each other.

I am still a little unclear what this means. In my TiKV auth proposal I was coming up with a solution so that applications could ensure that others did not write to their keys. This will work at the PD level to ensure only certain users can manipulate PD data of certain applications?

Yisaer:
> ys. This will work at the PD level to ensure only certain users can manipulate PD data of certain applications

Right. For exmaple, for some tools (like BR), PD may grant it to configure the `scheduler` and forbid it to remove the `store`.

nolouch:
Could you open the comment permission of the RFC docs?

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissionKeys and
+// limitations under the License.
+
+package auth
+
+import (
+	"encoding/json"
```
rleungx, line 17: remember to format it

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+		return errors.Errorf("role %s already has permission: %s", role, permission.String())
+	}
+	// ErrRoleMissingPermission is error info for role lacks a required role.
+	ErrRoleMissingPermission = func(role string, permission Permission) error {
+		return errors.Errorf("role %s doesn't have permission: %s", role, permission.String())
+	}
+)
+
+var (
+	userPrefix = "users"
```
rleungx, line 56: constants?

server/auth/auth.go
```diff
@@ -0,0 +1,740 @@
...
+}
+
+// NewManager creates a new Manager.
+func NewManager(k kv.TxnBase) *Manager {
+	return &Manager{kv: k}
+}
+
+// GetUser returns a user.
+func (m *Manager) GetUser(name string) (*User, error) {
+	_cachedUser, ok := m.users.Load(name)
```
rleungx, line 74: prefer not to use `_`

PhotonQuantum:
Moved to #3256

Issue #2950, opened 2020-09-14
Supports co-balance range group


nolouch:
### Describe your feature request related problem
the placement rules which can place the regions in the stores which meet the label constraint.  but in some situations, we need the range to share the same balance strategy and the peers always place in the same node. such as https://github.com/pingcap/tidb/issues/19805

### Describe the feature you'd like
The range group can add or remove interleaved ranges, that's mean the range inside is not necessarily continuous. each range group shares the same balance strategy and same placement,  It can be seen as a single region, in the future, TiKV can support to make them into one raft group.

### Teachability, Documentation, Adoption, Migration Strategy
- Go

Yisaer:
The detailed design would be release by @Yisaer recently.

Rustin-Liu:
@nolouch  @Yisaer Could you please modify the description of issue to follow this challenge program template?

nolouch:
@Rustin-Liu  ok

AT-Fieldless:
Hey, I want to participate in this issue

PR #3225, opened 2020-11-29
anti-affinity rules to seperated regions 

PearsonCHN:

Issue Number: https://github.com/tikv/pd/issues/2950

Problem Summary:

Proposal: 
we use a method of anti-affinity(refer to kubernetes), to let regions which are originally gathered at the same store to seperate. Asume that there is a table contains 3 regions, of which key ranges from 0-299, and there are 2 stores(store1, store2) in the tikv cluster. Now these 3 regions' leader are gathered at store1. An anti-affinity rules called antiRule1, of which key ranges cover 0-299, can make one of the 3 regions' region leader transfer to store2. In this case, loads can be seperated among these 2 stores.

What's Changed:

### Related changes

1. anti-rule checker: checks anti-rules that fit given region's key range
2. anti-rule manager: add anti-rule
3. antiRuleScore: a hash map to record the regions' amount in the same store of specified anti-rule.


PearsonCHN:
so sorry for being busy at work. Test results will be added in these two days.

JmPotato:
Did you change the file permission or something? Why all the *.sh files are committed but without any change?

PearsonCHN:
> Did you change the file permission or something? Why all the *.sh files are committed but without any change?

no, I don't modified any of the *.sh files,  I did feel weriod when I commited my codes.

JmPotato:
> > Did you change the file permission or something? Why all the *.sh files are committed but without any change?
> 
> no, I don't modified any of the *.sh files, I did feel weriod when I commited my codes.

According to the CI result, it seems that the permission of all these files is changed. Could you please revert these files to the previous version? Otherwise, the tests won't work.

```shell
[2020-12-01T13:01:32.943Z] /bin/bash: ./scripts/embed-dashboard-ui.sh: Permission denied
```

Maybe this command will be helpful:

```shell
git checkout {HASH} -- my-file.txt
```

server/cluster/cluster.go
```diff
@@ -1504,6 +1509,13 @@ func (c *RaftCluster) FitRegion(region *core.RegionInfo) *placement.RegionFit {
 	return c.GetRuleManager().FitRegion(c, region)
 }
 
+// GetAntiRuleManager returns the anti rule manager reference.
+func (c *RaftCluster) GetAntiRuleManager() *anti.AntiRuleManager {
+	c.RLock()
+	defer c.RUnlock()
```
JmPotato, line 1515: Though the `c.antiRuleManager` could be concurrently read, the only place to set it is in the `InitCluster`. So I'm wondering whether the lock here is necessary.

server/core/region.go
```diff
@@ -37,6 +38,8 @@ var errRegionIsStale = func(region *metapb.Region, origin *metapb.Region) error
 // RegionInfo records detail region info.
 // Read-Only once created.
 type RegionInfo struct {
+	mu                sync.Mutex
+	countInAntiScore  bool
```
JmPotato, line 42: Since the only role this `sync.Mutex` plays here is to lock the `countInAntiScore`, what about using `atomic.Value` here?

lhy1024:
why to chmod those scripts?

JmPotato:

PR #3227, opened 2020-11-30
core: add smoothed store available size 

disksing:

There are spikes of the store's available size that caused by TiKV's compaction. Result in PD creates redundant schedules.

Smooth the available stats and record its deviation. We can use it to calculate region score later.

### Check List

Tests
- Unit test

### Release note
- No release note


server/core/store_stats.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// GetAvailableDeviation returns approximate magnitude of available in the recent period.
+func (ss *storeStats) GetAvailableDeviation() uint64 {
+	ss.mu.RLock()
+	defer ss.mu.RUnlock()
+	return climp0(ss.avgMaxAvailableDeviation.Get())
+}
+
+func climp0(v float64) uint64 {
```
rleungx, line 151: What does climp0 here mean?

server/core/store_stats.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/tikv/pd/pkg/movingaverage"
+)
+
+type storeStats struct {
+	mu       sync.RWMutex
+	rawStats *pdpb.StoreStats
+	// smooth available size and record deviation to calculate region score.
+	avgAvailable             *movingaverage.HMA
+	maxAvailableDeviation    *movingaverage.MaxFilter
+	avgMaxAvailableDeviation *movingaverage.HMA
```
lhy1024, line 30: I think that we should add some comments for them

lhy1024 (approved):

server/core/store_stats.go
disksing: no special meaning 😢  do you have better idea?

server/core/store_stats.go
disksing: ok

server/core/store_stats.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func newStoreStats() *storeStats {
+	return &storeStats{
+		rawStats:                 &pdpb.StoreStats{},
+		avgAvailable:             movingaverage.NewHMA(240),       // 40min when 10s per hb
```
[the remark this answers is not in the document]
disksing: updated to more detailed description.

server/core/store_stats.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"math"
+	"sync"
+
+	"github.com/pingcap/kvproto/pkg/pdpb"
+	"github.com/tikv/pd/pkg/movingaverage"
+)
+
+type storeStats struct {
+	mu       sync.RWMutex
```
rleungx, line 25: Why do we need mutex here?

server/core/store_stats.go
rleungx: Will the float be truncated?

server/core/store_stats.go
disksing: It may be read (hot scheduler) and write (store heartbeat) at the same time.

server/core/store_stats.go
disksing: Yes. It is usually a very large value (unit is byte) so the truncation won't matter.

server/core/store_stats.go
rleungx: It seems we use clone to avoid directly write?

disksing: no. I did not clone this field.

rleungx (approved):

Issue #3230, opened 2020-11-30
Bug: PD follower got error "Got too many pings from the client"


hunterlxt:
## Bug Report

PD follower log got errors "Got too many pings from the client". Unlike tikv and tidb, they are responsible for establishing grpc server by itself, PD uses the grpc server provided by etcd, so the parameters cannot be set.  This error may occur when the client and server parameters `PermitWithoutStream` are not uniform.

HunDunDM:
A similar issue [tidb#17870](https://github.com/pingcap/tidb/issues/17870)

hunterlxt:
> A similar issue [tidb#17870](https://github.com/pingcap/tidb/issues/17870)

similar but not same, this is an issue between pds.

PR #3233, opened 2020-11-30
Fix "Got too many pings" GRPC error in sync region client-side

hunterlxt:
Signed-off-by: Xintao <hunterlxt@live.com>

fix "Got too many pings" GRPC error in sync region client. We should ensure that the grpc parameters are consistent on the client and server sides. Etcd server cannot set `PermitWithoutStream` so it's false by default.

close https://github.com/tikv/pd/issues/3230

make `PermitWithoutStream` false in region_syncer client.

### Check List

Related changes

### Release note

- Fix "Got too many pings" GRPC error in sync region client-side


HunDunDM (approved):

nolouch (approved):
lgtm

PR #3240, opened 2020-12-01
scripts: add script to check missing test

disksing:
Sometimes we may accidentally remove the `TestingT` function causes some tests do not run.

Add script to check it. (Not completely strict, and should be sufficient in most cases.)

### Check List
Tests
- Manual test

### Release note
- No release note


nolouch (approved):
lgtm

lhy1024 (approved):

PR #3247, opened 2020-12-02
server: fix the wrong use of validateInternalRequest

JmPotato:

Fix the wrong use of `validateInternalRequest`.

* `SplitRegions` is not an internal request, it should use `validateRequest` instead.
* `GetDCLocations` is an internal request that may be started by a PD follower, so the sender check is not necessary here.

### Check List

Tests

- Unit test
- Integration test


Yisaer (approved):
LGTM

nolouch (approved):
lgtm

PR #3248, opened 2020-12-02
server/auth: implement role & role manager

PhotonQuantum:

server/auth: implement role & role manager.

This PR is split from #3224.

I'm still working on the RFC document about this change (RBAC support for TiDB PD), and I'll submit it as soon as it's completed. In the meantime, a draft written in Chinese is available at [here](https://docs.google.com/document/d/1yZ1pjUqJZAFl4-ynHMaA-IrVbLojXjJKxZPs1hYlPbs/edit?usp=sharing).

server/auth `roleManager` is implemented to handle role-related CRUD operations, including querying roles in memory, loading them from etcd and persisting them to etcd.

### Check List

Tests

- Unit test

Code changes

### Release note

- No release note

### DIFF
--- a/pkg/errs/errno.go
+++ b/pkg/errs/errno.go
@@ -287,3 +287,12 @@ var (
 	ErrEncryptionSaveDataKeys       = errors.Normalize("failed to save data keys", errors.RFCCodeText("PD:encryption:ErrEncryptionSaveDataKeys"))
 	ErrEncryptionKMS                = errors.Normalize("KMS error", errors.RFCCodeText("PD:ErrEncryptionKMS"))
 )
+
+// auth
+var (
+	ErrRoleNotFound          = errors.Normalize("role not found: %s", errors.RFCCodeText("PD:auth:ErrRoleNotFound"))
+	ErrRoleExists            = errors.Normalize("role already exists: %s", errors.RFCCodeText("PD:auth:ErrRoleExists"))
+	ErrInvalidName           = errors.Normalize("key name may only contain alphanumeric and underscores, and may only start with an alphabetic character.", errors.RFCCodeText("PD:auth:ErrInvalidName"))
+	ErrRoleHasPermission     = errors.Normalize("role %s already has permission: %s", errors.RFCCodeText("PD:auth:ErrRoleHasPermission"))
+	ErrRoleMissingPermission = errors.Normalize("role %s doesn't have permission: %s", errors.RFCCodeText("PD:auth:ErrRoleMissingPermission"))
+)
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -1,0 +1,240 @@
+// Copyright 2020 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissionKeys and
+// limitations under the License.
+//
+
+package auth
+
+import (
+	"encoding/json"
+	"path"
+	"strings"
+	"sync"
+
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/kv"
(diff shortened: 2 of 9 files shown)


server/auth/auth.go
```diff
@@ -0,0 +1,239 @@
...
+
+type roleManager struct {
+	kv    kv.Base
+	lock  *sync.RWMutex
+	roles map[string]*Role
+}
+
+// newRoleManager creates a new roleManager.
+func newRoleManager(kv kv.Base) *roleManager {
+	return &roleManager{kv: kv, lock: new(sync.RWMutex), roles: make(map[string]*Role)}
```
rleungx, line 40: no need to new a mutex?

server/auth/auth.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+// RBACManager is used for the rbac storage, cache, management and enforcing logic.
+type RBACManager struct {
+	roleManager
+}
+
+type roleManager struct {
+	kv    kv.Base
+	lock  *sync.RWMutex
```
rleungx, line 34: prefer to rename it to mu or something else

server/auth/permission.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// Permission represents a permission to a specific pair of resource and action.
+type Permission struct {
+	Resource string `json:"resource"`
+	Action   Action `json:"action"`
+}
+
+// NewPermission safely creates a new permission instance.
+func NewPermission(resource string, action Action) (*Permission, error) {
```
rleungx, line 47: Seem we cannot create a permission with multiple actions?

server/auth/auth.go
PhotonQuantum: This function is only for debugging purposes, so we need to new a mutex to make roleManager work standalone.

server/auth/permission.go
PhotonQuantum: Yes. Currently, I'm defining Permission as a pair of resource and action. Multiple actions to the same resource can be written as several different Permissions.

server/auth/auth.go
rleungx: Hmm...actually I mean the mutex doesn't need to initialize

PhotonQuantum: I was thinking about sharing the same mutex between roleManager & userManager (not implemented in this pr yet). The draft is available at https://github.com/PhotonQuantum/pd/blob/rbac-manager/server/auth/auth.go#L46.

Umm... But it seems that there's no need to have a global mutex between roleManager and userManager. I'm removing it. Thanks!

pkg/errs/errno.go
```diff
@@ -287,3 +287,12 @@ var (
 	ErrEncryptionSaveDataKeys       = errors.Normalize("failed to save data keys", errors.RFCCodeText("PD:encryption:ErrEncryptionSaveDataKeys"))
 	ErrEncryptionKMS                = errors.Normalize("KMS error", errors.RFCCodeText("PD:ErrEncryptionKMS"))
 )
+
+// auth
+var (
+	ErrRoleNotFound          = errors.Normalize("role not found: %s", errors.RFCCodeText("PD:auth:ErrRoleNotFound"))
+	ErrRoleExists            = errors.Normalize("role already exists: %s", errors.RFCCodeText("PD:auth:ErrRoleExists"))
+	ErrInvalidName           = errors.Normalize("key name may only contain alphanumeric and underscores, and may only start with an alphabetic character.", errors.RFCCodeText("PD:auth:ErrInvalidName"))
```
gregwebs, line 295:
```suggestion
	ErrInvalidName           = errors.Normalize("role name may only contain alphanumeric and underscores, and may only start with an alphabetic character.", errors.RFCCodeText("PD:auth:ErrInvalidName"))
```

gregwebs: Should this be `ErrInvalidRoleName` ?

PhotonQuantum:
@Yisaer opened a new branch `rbac-auth` so that we may review rbac related prs first before merging them into `master`.
Moved to #3256.

pkg/errs/errno.go
PhotonQuantum: This error is actually reused when username is invalid. Maybe I should split it into `ErrInvalidRoleName` and `ErrInvalidUserName`?

gregwebs: Users will be confused by "key" since they are dealing with keys and roles.

PR #3249, opened 2020-12-03
scripts: make check_missing_tests compatible with macOS

JmPotato:

In macOS, there are two problems with the `scripts/check-missing-tests.sh`.

* If you don't specify a directory, `grep` will try to search in the standard input, but the recursive search does not make sense there.

图片

* `dirname` on macOS only takes a single pathname.

图片

Update the `check-missing-tests.sh`.

### Check List

Tests

- Unit test
- Integration test
- Manual test

### DIFF
--- a/scripts/check-missing-tests.sh
+++ b/scripts/check-missing-tests.sh
@@ -3,7 +3,7 @@
 # Check if there are any packages foget to add `TestingT` when use "github.com/pingcap/check".
 
 res=$(diff <(grep -rl --include=\*_test.go "github.com/pingcap/check" . | xargs -L 1 dirname | sort -u) \
-     <(grep -rl --include=\*_test.go -E "^\s*(check\.)?TestingT\(" . | xargs  -L 1 dirname | sort -u))
+     <(grep -rl --include=\*_test.go -E "^\s*(check\.)?TestingT\(" . | xargs -L 1 dirname | sort -u))
 
 if [ "$res" ]; then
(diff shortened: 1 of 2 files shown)


rleungx:
How about changing the target to `check-missing-tests`?

rleungx (approved):

scripts/check-missing-tests.sh
```diff
@@ -2,8 +2,8 @@
 
 # Check if there are any packages foget to add `TestingT` when use "github.com/pingcap/check".
 
-res=$(diff <(grep -rl --include=\*_test.go "github.com/pingcap/check" | xargs dirname | sort -u) \
-     <(grep -rl --include=\*_test.go -E "^\s*(check\.)?TestingT\(" | xargs dirname | sort -u))
+res=$(diff <(grep -rl --include=\*_test.go "github.com/pingcap/check" . | xargs -L 1 dirname | sort -u) \
+     <(grep -rl --include=\*_test.go -E "^\s*(check\.)?TestingT\(" . | xargs  -L 1 dirname | sort -u))
```
rleungx, line 6:
```suggestion
     <(grep -rl --include=\*_test.go -E "^\s*(check\.)?TestingT\(" . | xargs -L 1 dirname | sort -u))
```

nolouch (approved):

Yisaer (approved):
LGTM

Issue #3234, opened 2020-11-30
Transfer peer operator should check if the store is down


rleungx:
### What did you do?

Execute `operator add transfer-peer 1 2 3` but 3 is down.

### What did you expect to see?

The operator should be rejected.

### What did you see instead?

The operator is blocked until timeout.

### What version of PD are you using (`pd-server -V`)?
master

ti-srebot:
These labels are not found `severity/mederate`.

PR #3253, opened 2020-12-07
api: check the unhealthy store when adding a new operator through API

rleungx:

Closes https://github.com/tikv/pd/issues/3234

This PR adds a check when adding a new operator through API. If a store is unhealthy, we can prevent doing the operation on it.

### Check List

Tests

- Unit test

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -493,15 +493,9 @@ func (h *Handler) AddTransferRegionOperator(regionID uint64, storeIDs map[uint64
 			}
 		}
 	}
-
-	var store *core.StoreInfo
 	for id := range storeIDs {
-		store = c.GetStore(id)
-		if store == nil {
-			return errs.ErrStoreNotFound.FastGenByArgs(id)
-		}
-		if store.IsTombstone() {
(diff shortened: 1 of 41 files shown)


lhy1024 (commented):
This pr changed so many files because of rename test pkg = =

server/handler.go
```diff
@@ -920,3 +906,17 @@ func (h *Handler) SetStoreLimitTTL(data string, value float64, ttl time.Duration
 		data: value,
 	}, ttl)
 }
+
+func checkStoreState(rc *cluster.RaftCluster, storeID uint64) error {
+	store := rc.GetStore(storeID)
+	if store == nil {
+		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
+	}
+	if store.IsTombstone() {
```
JmPotato, line 915: Maybe we can add one more argument in `ErrStoreUnhealthy` to distinguish these different situations.

HunDunDM (approved):

JmPotato (approved):

PR #3257, opened 2020-12-08
tso: add avg latency metric in tso-bench

ZenoTan:

Add avg latency metric in tso-bench.

Store avg latency in metrics, output it in the report.

### Check List

Tests

- Manual test


rleungx (approved):

nolouch (approved):

JmPotato (approved):
Maybe we can have an easy and elegent way to add this kind of indicator in the future?

PR #3258, opened 2020-12-08
log: add pluggable_logger and implement http_logger for it

HunDunDM:

There are some debug logs, because there are so many, we don't want it to be printed in the log file. But they are often very helpful for troubleshooting certain problems. Currently, we can only obtain them by changing the log level. This PR aims to improve this solution.

* add pluggable_logger: Use the different pluggable_logger to print the debug log. The pluggable_logger will not have any side effects when the actual output is not set. The actual output can be set to GlobalLogger, HTTPLogger or FileLogger, etc.
* implement http_logger for pluggable_logger: When accessing `/admin/log` through API, a set of pluggable_logger output can be obtained through API within a specified time.

[example](https://github.com/HunDunDM/pd/compare/logger...HunDunDM:debug_log): `/pd/api/v1/admin/log?name=hot_region&second=60`

### Check List

Tests

- Unit test

Code changes

### Release note

- No release note

### DIFF
--- a/pkg/logutil/http_logger.go
+++ b/pkg/logutil/http_logger.go
@@ -52,13 +52,17 @@ func (l *HTTPLogger) AddCloseCallback(fs ...func()) {
 func (l *HTTPLogger) Plug(names ...string) {
 	l.AddCloseCallback(func() {
 		for _, name := range names {
-			pl := GetOrCreatePluggableLogger(name)
-			pl.SetLogger(nil)
+			pl := GetPluggableLogger(name, false)
+			if pl != nil {
+				pl.SetLogger(nil)
+			}
 		}
 	})
 	for _, name := range names {
-		pl := GetOrCreatePluggableLogger(name)
-		pl.SetLogger(l.logger)
+		pl := GetPluggableLogger(name, false)
+		if pl != nil {
+			pl.SetLogger(l.logger)
+		}
 	}
 }
 
--- a/pkg/logutil/http_logger.go
+++ b/pkg/logutil/http_logger.go
@@ -60,6 +60,7 @@ func (l *HTTPLogger) Plug(names ...string) {
 			}
 		}
 	})
+	l.writer.WriteHeader(http.StatusOK)
 	for _, name := range names {
 		pl := GetPluggableLogger(name, false)
 		if pl != nil {
--- a/pkg/logutil/http_logger.go
+++ b/pkg/logutil/http_logger.go
@@ -67,7 +67,6 @@ func (l *HTTPLogger) Plug(names ...string) {
 
 // Close will call close callbacks and close all output.
 func (l *HTTPLogger) Close() {
-	defer l.writer.WriteHeader(http.StatusOK)
 	defer l.logger.Sync()
 	for _, f := range l.closeCallbacks {
 		f()
--- a/pkg/logutil/http_logger.go
+++ b/pkg/logutil/http_logger.go
@@ -15,7 +15,7 @@ package logutil
 
 import (
 	"net/http"
-	"time"
+	"sync"
 
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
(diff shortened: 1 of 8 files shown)


pkg/logutil/http_logger_test.go
```diff
@@ -0,0 +1,91 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logutil
+
+import (
+	"bytes"
+	"go.uber.org/zap"
```
lhy1024, line 18: import sort

pkg/logutil/http_logger.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// NewHTTPLogger returns a HTTPLogger.
+func NewHTTPLogger(conf *log.Config, w http.ResponseWriter) (*HTTPLogger, error) {
+	syncer := zapcore.AddSync(w)
+	logger, props, err := log.InitLoggerWithWriteSyncer(conf, syncer, zap.AddStacktrace(zapcore.FatalLevel))
+	if err != nil {
+		return nil, err
+	}
+	props.Level.SetLevel(zapcore.DebugLevel)
```
rleungx, line 38: Why change it here?

HunDunDM: Specific debugging information requires more details.

rleungx: Why not `logger, props, err := log.InitLoggerWithWriteSyncer(conf, syncer, zap.AddStacktrace(zapcore. DebugLevel))`?

HunDunDM: I didn't know it :(

server/api/log.go
```diff
@@ -69,3 +71,52 @@ func (h *logHandler) Handle(w http.ResponseWriter, r *http.Request) {
...
+	logConfig := h.svr.GetConfig().Log
+	logConfig.Format = "json"
+	logConfig.Level = "debug"
+	httpLogger, err := logutil.NewHTTPLogger(&logConfig, w)
+	if err != nil {
+		h.rd.JSON(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	httpLogger.Plug(names...)
```
rleungx, line 115: If there is no logger for a given name, we should return the message to the client?

server/api/log.go
```diff
@@ -69,3 +71,52 @@ func (h *logHandler) Handle(w http.ResponseWriter, r *http.Request) {
...
+	var second int64
+	if secondStr := query.Get("second"); secondStr != "" {
+		var err error
+		second, err = strconv.ParseInt(secondStr, 10, 64)
+		if err != nil {
+			h.rd.JSON(w, http.StatusBadRequest, err.Error())
+			return
+		}
+	}
+	if second <= 0 {
```
rleungx, line 96: Why not 400?

server/api/log.go
```diff
@@ -69,3 +71,52 @@ func (h *logHandler) Handle(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	names := query["name"]
+	if len(names) == 0 {
+		h.rd.JSON(w, http.StatusBadRequest, "empty name.")
+		return
+	}
+
+	logConfig := h.svr.GetConfig().Log
+	logConfig.Format = "json"
```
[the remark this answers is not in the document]
HunDunDM: I think there is no need to use other log levels and formats. If necessary in the future, it is easy to add support to the API.

server/api/log.go
HunDunDM: good idea

server/api/log.go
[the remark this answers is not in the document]
rleungx: So it's better to change the function name or leave a TODO here.

pkg/logutil/log.go
```diff
@@ -254,13 +254,14 @@ func (lg *wrapLogrus) V(l int) bool {
 	return int(lg.Logger.Level) <= logrusLevel
 }
 
-var once sync.Once
+var (
+	initLoggerOnce sync.Once
+	initLoggerErr  error
```
rleungx, line 259: Why change it to a global variable?

pkg/logutil/pluggable_logger.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		l = &PluggableLogger{name: name}
+		m.Loggers[name] = l
+	}
+	return l
+}
+
+var globalLoggerManager loggerManager
+
+func init() {
```
rleungx, line 136: Is there anyway to avoid it?

pkg/logutil/pluggable_logger_test.go
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logutil
+
+import (
+	"bytes"
+	"go.uber.org/zap"
```
rleungx, line 18: need format

pkg/logutil/log.go
HunDunDM: This function will only be executed once. If an error occurs in the first execution, the next execution should continue to return err instead of nil.

pkg/logutil/pluggable_logger.go
HunDunDM: In the example, `GetPluggableLogger` will be used to generate some new global loggers.

server/api/log.go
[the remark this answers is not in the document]
HunDunDM: Already supported.

JmPotato:
Do we still need to finish this?

rleungx:
Can we close it since it has been stale for more than one year?

PR #3263, opened 2020-12-09
tso: add GetLastTS to get last global TSO

Andrewmatilde:
Signed-off-by: Andrewmatilde <davis6813585853062@outlook.com>

ref [#21522](https://github.com/pingcap/tidb/issues/21522)

### Check List

Tests

- Unit test

Side effects

- Increased code complexity

### Release note

 No release note

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -56,6 +56,10 @@ type Client interface {
 	GetLocalTS(ctx context.Context, dcLocation string) (int64, int64, error)
 	// GetLocalTSAsync gets a local timestamp from PD, without block the caller.
 	GetLocalTSAsync(ctx context.Context, dcLocation string) TSFuture
+	// GetLastTS gets the last timestamp given by GetTS.
+	GetLastTS(ctx context.Context) (int64, int64, error)
+	// GetLastLocalTS gets the last timestamp given by GetLastLocalTS.
+	GetLastLocalTS(ctx context.Context, dcLocation string) (int64, int64, error)
 	// GetRegion gets a region and its leader Peer from PD by key.
 	// The region may expire after split. Caller is responsible for caching and
 	// taking care of region change.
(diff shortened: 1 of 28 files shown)


client/client.go
```diff
@@ -638,11 +642,26 @@ func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err
...
+
 func (c *client) GetLocalTS(ctx context.Context, dcLocation string) (physical int64, logical int64, err error) {
 	resp := c.GetLocalTSAsync(ctx, dcLocation)
 	return resp.Wait()
 }
 
+func (c *client) GetLastLocalTS(ctx context.Context, dcLocation string) (int64, int64, error) {
+	lastTSOInterface, ok := c.lastTSMap.Load(dcLocation)
+	if !ok {
+		err := fmt.Errorf("func GetLastLocalTS fail : get last local TS before get first TS or something wrong happen")
```
JmPotato, line 657: Please make the log info look like this:

```shell
[pd] get last local TS before get first TS or something wrong happen
```

client/client.go
```diff
@@ -56,6 +56,10 @@ type Client interface {
 	GetLocalTS(ctx context.Context, dcLocation string) (int64, int64, error)
 	// GetLocalTSAsync gets a local timestamp from PD, without block the caller.
 	GetLocalTSAsync(ctx context.Context, dcLocation string) TSFuture
+	// GetLastTS gets the last timestamp given by GetTS.
+	GetLastTS(ctx context.Context) (int64, int64, error)
+	// GetLastLocalTS gets the last timestamp given by GetLastLocalTS.
+	GetLastLocalTS(ctx context.Context, dcLocation string) (int64, int64, error)
```
JmPotato, line 62: Maybe we can merge these two methods into a single one?

client/client.go
rleungx: How about using errors.New?

Andrewmatilde: thanks for your review, I agree with you.

client/client.go
```diff
@@ -638,11 +642,26 @@ func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err
...
+	return c.GetLastLocalTS(ctx, globalDCLocation)
+}
+
 func (c *client) GetLocalTS(ctx context.Context, dcLocation string) (physical int64, logical int64, err error) {
 	resp := c.GetLocalTSAsync(ctx, dcLocation)
 	return resp.Wait()
 }
 
+func (c *client) GetLastLocalTS(ctx context.Context, dcLocation string) (int64, int64, error) {
+	lastTSOInterface, ok := c.lastTSMap.Load(dcLocation)
```
Yisaer, line 655: According to https://github.com/pingcap/tidb/issues/21522 , we also need to maintain the arrival timestamp in lastTSMap.

Yisaer (commented):
Rest LGTM

Yisaer:
```sh
revive -formatter friendly -config revive.toml $(go list ./...)
  ✘  https://revive.run/r#confusing-results  unnamed results of the same type may be confusing, consider using named results  
  /home/runner/work/pd/pd/client/client.go:654:1
```

plz solve this error

client/client.go
Andrewmatilde: Thanks for your review, I agree with you.

nolouch:
Why need this interface?

client/client.go
Andrewmatilde: Thanks for your review, I agree with you.

Yisaer:
> https://github.com/pingcap/tidb/blob/master/store/tikv/oracle/oracles/pd.go#L134-L152

I miss that, but I think If we implement this function in pdclient, other components could reuse this ability?

nolouch:
@Yisaer  It's not a common request. and there is an issue about how to push the last ts if not have any request, TiDB already solve it.

Yisaer:
> @Yisaer It's not a common request. and there is an issue about how to push forward the `lastTS` if not have any request, TiDB already solves it . BTW, the `GetLastTS` is ambiguous, whether it means the client's `lastTS` or the server's `lastTS`, the interface also confuse me.

got it. That's ok to me.

PR #3264, opened 2020-12-10
server/auth: implement user & user manager.

PhotonQuantum:

server/auth: implement user & user manager.

Related PR: #3256 

server/auth `userManager` is implemented to handle user-related CRUD operations, including querying users in memory, loading them from etcd and persisting them to etcd.

### Check List

Tests

- Unit test

Code changes

### Release note

- No release note

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -272,7 +272,8 @@ func (m *userManager) UpdateCache() error {
 	m.users = make(map[string]*User)
 	for i := range keys {
 		value := values[i]
-		user, err := NewUserFromJSON(value)
+
+		user, err := UnmarshalUser(value)
 		if err != nil {
 			return err
 		}
@@ -493,7 +494,7 @@ func (m *roleManager) UpdateCache() error {
 	m.roles = make(map[string]*Role)
 	for i := range keys {
 		value := values[i]
-		role, err := NewRoleFromJSON(value)
+		role, err := UnmarshalRole(value)
 		if err != nil {
 			return err
 		}
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -24,9 +24,261 @@ import (
 	"go.etcd.io/etcd/clientv3"
 )
 
+const (
+	userPrefix = "users"
+	rolePrefix = "roles"
+)
+
 // RBACManager is used for the rbac storage, cache, management and enforcing logic.
 type RBACManager struct {
+	userManager
 	roleManager
+}
+
+// NewRBACManager creates a new RBACManager.
+func NewRBACManager(kv kv.Base) *RBACManager {
+	return &RBACManager{
+		userManager{
+			kv:    kv,
+			users: make(map[string]*User),
+		},
+		roleManager{
+			kv:    kv,
+			roles: make(map[string]*Role),
+		}}
+}
+
+type userManager struct {
+	kv    kv.Base
+	mu    sync.RWMutex
+	users map[string]*User
+}
+
+// newUserManager creates a new roleManager for debug purposes.
+func newUserManager(kv kv.Base) *userManager {
+	return &userManager{kv: kv, users: make(map[string]*User)}
+}
+
+// GetUser returns a user.
+func (m *userManager) GetUser(name string) (*User, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	user, ok := m.users[name]
+	if !ok {
+		return nil, errs.ErrUserNotFound.FastGenByArgs(name)
+	}
+
+	return user, nil
+}
+
+// GetUsers returns all available roles.
+func (m *userManager) GetUsers() map[string]*User {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return m.users
+}
+
+// CreateUser creates a new user.
+func (m *userManager) CreateUser(name string, password string) error {
+	_, err := m.GetUser(name)
+	if err == nil {
+		return errs.ErrUserExists.GenWithStackByArgs(name)
+	}
+
+	user, err := NewUser(name, GenerateHash(password))
+	if err != nil {
+		return err
+	}
+
+	userJSON, err := json.Marshal(user)
+	if err != nil {
+		return errs.ErrJSONMarshal.Wrap(err).GenWithStackByCause()
+	}
+	userPath := path.Join(userPrefix, name)
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// Add user to kv.
+	err = m.kv.Save(userPath, string(userJSON))
+	if err != nil {
+		return err
+	}
+
+	// Add user to memory cache.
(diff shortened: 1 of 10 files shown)


server/auth/auth.go
```diff
@@ -24,18 +24,270 @@ import (
...
+	user, err := NewUser(name, GenerateHash(password))
+	if err != nil {
+		return err
+	}
+
+	userJSON, err := json.Marshal(user)
+	if err != nil {
+		return errs.ErrJSONMarshal.Wrap(err).GenWithStackByCause()
+	}
+	userPath := path.Join(userPrefix, name)
```
JmPotato, line 99: I saw the use of this function for multi-times, maybe we can make it become a common method?

server/auth/auth.go
```diff
@@ -24,18 +24,270 @@ import (
...
+}
+
 type roleManager struct {
 	kv    kv.Base
 	mu    sync.RWMutex
 	roles map[string]*Role
 }
 
-// newRoleManager creates a new roleManager.
+// newRoleManager creates a new roleManager for debug purposes.
```
Yisaer, line 290:
```suggestion
// newRoleManager creates a new roleManager.
```

server/auth/auth.go
```diff
@@ -24,18 +24,270 @@ import (
...
+
+	if ok := user.appendRole(role); !ok {
+		return errs.ErrUserHasRole.FastGenByArgs(name, role)
+	}
+
+	userJSON, err := json.Marshal(role)
+	if err != nil {
+		return errs.ErrJSONMarshal.Wrap(err).GenWithStackByCause()
+	}
+	userPath := path.Join(userPrefix, name)
```
Yisaer, line 214: I think we can define a function in util like `GetUserPath` and `GetRolePath`.

server/auth/auth.go
```diff
@@ -24,18 +24,270 @@ import (
...
+
+	return nil
+}
+
+// UpdateCache refreshes in-memory cache of users.
+func (m *userManager) UpdateCache() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	userPath := strings.Join([]string{userPrefix, ""}, "/")
```
Yisaer, line 265: I see different methods to build userPath

Yisaer (commented):
plz address the comment

server/auth/auth.go
```diff
@@ -24,18 +24,270 @@ import (
...
+	user.Hash = GenerateHash(password)
+
+	// Update user in kv
+	userJSON, err := json.Marshal(user)
+	if err != nil {
+		return errs.ErrJSONMarshal.Wrap(err).GenWithStackByCause()
+	}
+	userPath := path.Join(userPrefix, name)
+
+	err = m.kv.Save(userPath, string(userJSON))
```
rleungx, line 159: What if we have something wrong when saving kv, will the password be rollback?

gregwebs:
All the auth information is cached in memory. What happens when the PD leader changes or when a different PD is contacted? I am wondering if we can just rely on etcd to cache this information as it sees fit?

I am also wondering if we could be delegating this RBAC to the RBAC that is already [available in etcd](https://etcd.io/docs/v3.3.12/op-guide/authentication/).

server/auth/auth.go
PhotonQuantum: Extracted to util.

server/auth/auth.go
PhotonQuantum: Using this method in `GetUserPath` because it can handle trailing slash correctly.

server/auth/auth.go
PhotonQuantum: Wow, it's a bug! Thank you for pointing it out!

server/auth/user.go
```diff
@@ -0,0 +1,107 @@
...
+	ok := validateName(username)
+	if !ok {
+		return nil, errs.ErrInvalidUserName.FastGenByArgs(username)
+	}
+
+	return &User{Username: username, Hash: hash, RoleKeys: make([]string, 0)}, nil
+}
+
+// NewUserFromJSON safely deserialize a json string to a user instance.
+func NewUserFromJSON(j string) (*User, error) {
```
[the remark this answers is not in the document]
PhotonQuantum: Renamed.

server/auth/user.go
```diff
@@ -0,0 +1,107 @@
...
+	"encoding/json"
+
+	"github.com/tikv/pd/pkg/errs"
+)
+
+// User records user info.
+// Read-Only once created.
+type User struct {
+	Username string   `json:"username"`
+	Hash     string   `json:"hash"`
```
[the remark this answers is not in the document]
PhotonQuantum: Added.

server/auth/auth.go
PhotonQuantum: Extracted into `GetUserPath` and `GetRolePath`.

server/auth/auth.go
PhotonQuantum: Removed.

gregwebs:
Any response to my comments? Does the memory caching work when more than one PD is used?

Since this PR is just to a branch I think we can merge it once tests are passing.

PhotonQuantum:
> Any response to my comments?

Sorry for the late reply! I'm quite busy with my studies these days, so I may not respond in time 😢. But I will try to do so as long as I can.

> All the auth information is cached in memory. What happens when the PD leader changes or when a different PD is contacted?

I believe we may refresh the cache when a new PD leader is elected, but this won't work when a follower is contacted. I'm thinking about redirecting all requests to the leader when auth is enabled, for now.

> I am wondering if we can just rely on etcd to cache this information as it sees fit.

I don't like it because this will reduce the reflexivity of the auth system, and slow down the authentication process (accessing etcd per request is obviously slower than reading from in4-memory cache.).

@Yisaer What's your opinion on this?

Yisaer:
> What happens when the PD leader changes or when a different PD is contacted?

I think we can load the data from etcd while the member is elected as leader. Also we could create a `gouroutine` to refresh the data from etcd to memory to keep consistency.

> delegating this RBAC to the RBAC that is already available in etcd.

delegating RBAC to etcd RBAC strategy means all the resources the PD holding should saved into etcd while etcd is not the only stored backed (pd also use sqlite).

Yisaer (approved):
LGTM

gregwebs:
> Also we could create a gouroutine to refresh the data from etcd to memory to keep consistency.

This will always create inconsistency windows unless we can know if things are out of date. Refresh on leader election and only sending requests to the leader would keep things 100% consistent though.

> I don't like it because this will reduce the reflexivity of the auth system, and slow down the authentication process (accessing etcd per request is obviously slower than reading from in-memory cache.).

We can benchmark perform implementing performance optimizations. Etcd aggressively caches data in memory so I think we may be duplicating this work. Accessing etcd does not necessarily have much overhead since we are using it in an embedded manner.

PR #3267, opened 2020-12-11
tso: make the TSO unique while member changing

### LINKED ISSUE #3260: Make the Local TSO unique while member changing

Currently, we will use the low bits of the logical part of TSO to make every TSO unique among a cluster according to #3179. However, this constraint may be broken during the member changing. For example, we have three DCs, then its suffix sign in a Local TSO's logical part will look like:

`t1: xxxxxxxxxxxxxxxx | 11`, which contains 16 bits normal logical part and 2 bits suffix sign part.

And if we add several new DCs into the cluster, one of the newly added DCs may have a Local TSO's logical part looks like:

`t2: xxxxxxxxxxxxxxx | 111`, which contains 15 bits normal logical part and 3 bits suffix sign part.

In this case, we have may a TSO t1 is the same with t2 like:

* `t1: xxxxxxxxxxxxxxx1 | 11`
* `t2: xxxxxxxxxxxxxxx | 111`

To prevent this kind of situation from happening, we have these solutions:

* Use the fixed bits of the suffix sign.
* Have a Global TSO synchronization after the member changes.
* ...

JmPotato:

Resolve #3260. Also, PTAL at pingcap/kvproto#706.

* Have a Global TSO synchronization when the PD leader found the number of suffix bits changes, and make sure the new Local TSO Allocator won't provide service until the PD synchronization is over by introducing a bool flag named `NeedSyncTSO` (about 179 lines of codes).
* Add a corresponding test for this case (about 334 lines of codes).
* Refine some codes.
* Add more comments.

### Check List

Tests

- Unit test
- Integration test


server/tso/allocator_manager.go
```diff
@@ -291,12 +302,12 @@ func (am *AllocatorManager) getAllocatorPath(dcLocation string) string {
 
 // similar logic with leaderLoop in server/server.go
 func (am *AllocatorManager) allocatorLeaderLoop(ctx context.Context, allocator *LocalTSOAllocator) {
+	defer log.Info("server is closed, return local tso allocator leader loop",
```
rleungx, line 305: There are some other cases that will return

JmPotato: According to the codes, the only case that will make the `allocatorLeaderLoop` return is that `ctx` is canceled, which can only be caused by the server closing. Which other cases do you mean exactly?

rleungx: Oh, you're right, my mistake

server/tso/allocator_manager.go
```diff
@@ -561,9 +581,25 @@ func (am *AllocatorManager) ClusterDCLocationChecker() {
...
+					suffixBitsChanged = true
+				}
+				am.mu.maxSuffix = suffix
+			}
 			am.mu.clusterDCLocations[dcLocation].suffix = suffix
 		}
 	}
+	am.mu.Unlock()
+	if suffixBitsChanged {
+		am.syncGlobalTSO()
```
nolouch, line 601: Is it possible to generate local tso before `syncGlobalTSO`? the suffix already generated before `syncGlobalTSO`.

server/tso/allocator_manager.go
```diff
@@ -309,7 +320,10 @@ func (am *AllocatorManager) allocatorLeaderLoop(ctx context.Context, allocator *
 				zap.String("dc-location", allocator.dcLocation),
 				zap.Int("suffix", suffix),
 				errs.ZapError(err))
-			time.Sleep(200 * time.Millisecond)
+			// PD leader hasn't been elected out, wait for the campaign
+			if !longSleep(ctx, time.Second) {
```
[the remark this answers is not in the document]
JmPotato: Because the PD leader campaign usually lasts for seconds, and I notice this line print a lot of logs during the test, so I increase the sleep time here.

server/tso/allocator_manager.go
JmPotato: Theoretically yes, but it's very unlikely to occur this situation. Because PD leader will sync with the local as soon as the Local TSO Allocator leader is elected out. This should happen in a second. Consdering the PD client won't spot the new Local TSO Allocator so qucikly. I thinks it acceptable (PD client will update the PD leader and Local TSO Allocators every minute).

rleungx: What if there is something wrong when syncGlobalTSO? How long should we wait before the next check?

nolouch: I think we should make `syncGlobalTSO` succeed before providing the TSO service in the newly dc allocator. BTW, Does the member delete not adjust the max suffix?

JmPotato: @nolouch For now, we don't recall the suffix has already allocated, so the member deleting will not affect the max suffix.

JmPotato:
@nolouch @rleungx @Yisaer PTAL

Yisaer:
After disscussed with @JmPotato , the preparing of suffix should be done in `isLeaderAwareOfDCLocation`.

server/grpc_service.go
```diff
@@ -1030,19 +1030,32 @@ func (s *Server) SplitRegions(ctx context.Context, request *pdpb.SplitRegionsReq
...
+		am.ClusterDCLocationChecker()
+		return nil, fmt.Errorf("dc-location %s is not found", request.GetDcLocation())
+	}
+	resp := &pdpb.GetDCLocationInfoResponse{
+		Header: s.header(),
+		Suffix: info.Suffix,
+	}
+	if info.NeedSyncMaxTSO {
+		var err error
+		if resp.MaxTs, err = am.GetMaxLocalTSO(ctx); err != nil {
```
Yisaer, line 1054: I think we don't need to return `maxTS`

server/tso/allocator_manager.go
```diff
@@ -547,62 +598,80 @@ func (am *AllocatorManager) ClusterDCLocationChecker() {
...
+		// Follower should check and update the am.mu.maxSuffix
+		maxSuffix, err := am.getMaxLocalTSOSuffix()
+		if err != nil {
+			log.Error("get the max local tso suffix from etcd failed", errs.ZapError(err))
+			// Rollback the new dc-locations we update before
+			for _, dcLocation := range newDCLocations {
+				delete(am.mu.clusterDCLocations, dcLocation)
+			}
+		} else if maxSuffix > am.mu.maxSuffix {
+			am.mu.maxSuffix = maxSuffix
```
Yisaer, line 654: Why `follower` need to maintain the `maxSuffix`?

Yisaer (commented):
Rest LGTM

server/grpc_service.go
JmPotato: This `MaxTs` will be used [here](https://github.com/tikv/pd/pull/3267/files#diff-0d818fb56aef7677c53762dd2a46dd116a8e91979d743e3bee39cfa1dbc0bb3bR455).

server/tso/allocator_manager.go
JmPotato: If the follower holds a Local TSO Allocator leader, it will be used to pass as an argument [here](https://github.com/tikv/pd/pull/3267/files#diff-864112911592e078c8ec9a72330c17cc7473a525e144f2bde0c2462a8f02d02dR99) while generating a Local TSO.

Yisaer (approved):
LGTM

nolouch (commented):
rest lgtm.

nolouch (commented):
lgtm. please update kvproto

nolouch (approved):

PR #3268, opened 2020-12-13
mockcluster: use realistic store size and region size in unit tests

disksing:

Currently, we use 1GB storage size and 10M region size for tests. It is different from the real configuration.

Use 100GiB storage size and 96MiB region size by default.

### Check List

Tests
- Unit test

### Release note
- No release note


lhy1024 (approved):

Yisaer (approved):
LGTM

PR #3269, opened 2020-12-14
*: add new region score formula

disksing:

Make region score grow smoother and less affected by available changes.

- introduce new formula (v2)
- add config to control formula version
- add test

Default formula for newly created clusters is "v2", all old clusters keeps "v1" unless set manually.

The test result shows that compared to v4.0.7, the new formula can help balance fast to converge and more stable even with a very high store limit configuration.

###  Check List

Tests
- Unit test

### Related Issue
- https://github.com/pingcap/tidb/issues/18012

### Release note
- Introduce new region score formula

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -307,8 +307,11 @@ func (s *StoreInfo) regionScoreV2(delta int64, deviation int) float64 {
 	if A >= C || C < 1 {
 		score = R
 	} else if A > F {
+		// As the amount of data increases (available becomes smaller), the weight of region size on total score
+		// increases. Ideally, all nodes converge at the position where remaining space is F (default 20GiB).
 		score = (K + M*(math.Log(C)-math.Log(A-F+1))/(C-A+F-1)) * R
 	} else {
+		// When remaining space is less then F, the score is mainly determined by available space.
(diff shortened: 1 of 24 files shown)


server/core/store.go
```diff
@@ -279,6 +294,26 @@ func (s *StoreInfo) RegionScore(highSpaceRatio, lowSpaceRatio float64, delta int
...
+	var (
+		K, M float64 = 1, 256 // Experience value to control the weight of the available influence on score
+		F    float64 = 20     // Experience value to prevent some nodes from running out of disk space prematurely.
+	)
+
+	var score float64
+	if A >= C || C < 1 {
+		score = R
+	} else if A > F {
+		score = (K + M*(math.Log(C)-math.Log(A-F+1))/(C-A+F-1)) * R
```
rleungx, line 310: How about adding some comments or some reference to explain why the function look like this?

disksing: good idea.

rleungx (approved):

lhy1024 (commented):
LGTM, do we need to add some test result in this pr desc?

nolouch (approved):
lgtm

disksing:
> LGTM, do we need to add some test result in this pr desc?

ok, I'll add a short description.

PR #3287, opened 2020-12-18
election, member, tso: fix the leader check bug and refine the logic

JmPotato:

If you check the leader immediately after the campaign failed by `leadership.Check()`, it will return true wrongly because the lease is not expired yet. And to be more specific, this bug could cause a PD to make a non-leader Local TSO Allocator participate in a Global TSO synchronization, which will make the synchronization fail, right after the campaign failed. This PR will fix this problem.

Also, I rearrange the logic in `allocatorLeaderLoop` to reduce the RPC requests and etcd IO, which should improve the performance.

* Merge `IsStillLeader` and `IsLeader` methods, they are doing the same thing.
* Improve the leader check to make it more precise.
* Rearrange the logic in `allocatorLeaderLoop` to reduce the RPC requests and etcd IO.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/election/leadership.go
+++ b/server/election/leadership.go
@@ -107,9 +107,11 @@ func (ls *Leadership) Campaign(leaseTimeout int64, leaderData string) error {
 		Commit()
 	log.Info("check campaign resp", zap.Any("resp", resp))
 	if err != nil {
+		ls.getLease().Close()
 		return errs.ErrEtcdTxn.Wrap(err).GenWithStackByCause()
 	}
 	if !resp.Succeeded {
+		ls.getLease().Close()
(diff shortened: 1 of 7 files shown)


nolouch (approved):
LGTM

Yisaer (approved):
LGTM

rleungx:
Does this PR have any tests?

JmPotato:
> Does this PR have any tests?

I add a test in this [PR](https://github.com/tikv/pd/pull/3290/files#diff-adc274c7b56e5c4c0238e59a97673d30cb885c99c4cf588eb194990772d323a1).

Issue #3293, opened 2020-12-23
Unexisted labels in placement rules blocking the scheduel


xhebox:
## Bug Report

PD did not check if posted placement rules has a label constraint that is actually not belong to any tikv instances. And this particular placement rule seems block the scheduel of all other placement rules.

@djshow832 suggested that the step 8 in the below instruction should fail because this label does not exist.

**Update**: Even if label keys existed in the step 8(add `+zone=bj` for tikv instances), the number of peer stiil does not change.

### What did you do?

1. `tiup playground nightly --db 3 --kv 5 --pd 3`
2. connect one of the TiDB
3. `set global tidb_enable_alter_placement=on;`
4. reconnect the TiDB to refresh the global switch
5. `create table test.t1 (a int) partition by hash(a) partitions 2;`
6. `alter table test.t1 alter partition p1 alter placement policy CONSTRAINTS='[]' REPLICAS=2 ROLE=voter;`
7. `show table test.t1 regions;`
8. `alter table test.t1 alter partition p1 alter placement policy CONSTRAINTS='["+zone=sh"]' REPLICAS=1 ROLE=follower;`
9. `show table test.t1 regions;`
10. `alter table test.t1 alter partition p1 alter placement policy CONSTRAINTS='[]' REPLICAS=3 ROLE=voter;`
11. `show table test.t1 regions;`

### What did you expect to see?

step 7 should show p1 partition with 2 peer.
step 9 should show p1 partition with 2 peer since the only existed and default label is tiflash label.
step 11 should show p1 partition with 3 peer.

### What did you see instead?

step 7 showed p1 partition with 2 peer.
step 9 showed p1 partition with 2 peer.
step 11 showed p1 partition with 2 peer.

### What version of PD are you using (`pd-server -V`)?

nightly-20201212

Issue #2255, opened 2020-03-18
TiFlash replicate progress blocked when there is only 1 TiKV node


zhangjinpeng1987:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?

- Bootstrapped a cluster with 1 TiKV, 1 TiFlash, 1 PD and 1 TiDB.
- load some data into TiDB
- and then execute `alter table t set tiflash replica 1` to add the TiFlash replica for table t.

2. What did you expect to see?
- The TiFlash replica progress stays at 0%.
- After changed PD's `max-replicas` from 3 to 1, the progress still 0%.
- After execute `curl -X POST -d '{"group_id":"pd","id":"default","role":"voter","count":1}' http://127.0.0.1:2379/pd/api/v1/config/rule`, the replicate progress start to work.

3. What did you see instead?

- The TiFlash replicate progress works When there is only one TiKV node.
- After change `max-replicas` from 3 to 1, the replicate progress should work

4. What version of PD are you using (`pd-server -V`)?

Release Version: v4.0.0-beta.1-37-g0ba1160b
Git Commit Hash: 0ba1160bf0b5537ce62067643857ce5d191be659
Git Branch: release-4.0
UTC Build Time:  2020-03-11 01:54:01

disksing:
> The TiFlash replicate progress works When there is only one TiKV node.
It is expected behaviour when there is only 1 tikv but the default rule specify 3 voters, because scheduler tries to satisfy rules one by one.

> After change max-replicas from 3 to 1, the replicate progress should work
When placement rules feature is enabled, `max-replicas` does not use any more. I think we can give some error message instead make it work.

PR #3296, opened 2020-12-23
placement, checker: Fix rule_checker blocked by the rule in some cases

Yisaer:
Close https://github.com/tikv/pd/issues/3293

If rule checker found some rule have not matched store in `FitRegion`, it will break the `FitRegion` process directly.

1. Let `FitRegion` fix each `RuleFit`, if it found there is no `fix` for a `RuleFit`, it will try to fix next `RuleFit` instead of exiting.
2. Introducing filter the rules which won't be fixed before `FixRegion`, this would reduce the calculation consuming for each Region.

### Check List

Tests

- Unit test

### Release note

* No release note


Yisaer:
@xhebox  PTAL

server/schedule/placement/rule_filter_test.go
```diff
@@ -0,0 +1,103 @@
...
+							Value: "sh",
+						},
+					},
+				})
+				stores := []*core.StoreInfo{
+					store1,
+				}
+				return WithMatchStoreFilter(stores)
+			}(),
+			rule: &Rule{
```
xhebox, line 80: What about cases with multiple constraints?

server/schedule/operator_controller_test.go
```diff
@@ -663,7 +663,9 @@ func (t *testOperatorControllerSuite) TestAddWaitingOperator(c *C) {
 	cluster := mockcluster.NewCluster(config.NewTestOptions())
 	stream := hbstream.NewTestHeartbeatStreams(t.ctx, cluster.ID, cluster, false /* no need to run */)
 	controller := NewOperatorController(t.ctx, cluster, stream)
-
+	cluster.AddLabelsStore(1, 1, map[string]string{"host": "host1"})
```
xhebox, line 666: Same storeID?

server/schedule/checker/rule_checker_test.go
```diff
@@ -337,3 +337,40 @@ func (s *testRuleCheckerSuite) TestIssue2419(c *C) {
...
+		ID:      "default",
+		Role:    placement.Voter,
+		Count:   3,
+	})
+	c.Assert(err, IsNil)
+	err = s.ruleManager.DeleteRule("pd", "default")
+	c.Assert(err, IsNil)
+	op := s.rc.Check(s.cluster.GetRegion(1))
+	c.Assert(op, NotNil)
+	c.Assert(op.Desc(), Equals, "add-rule-peer")
```
xhebox, line 375: Can we check that it is adding 3 rather than 1 peer?

xhebox (commented):
Rest LGTM, just one more problem: @djshow832 has prepared some placement rules SQL tests, which are expected to fail(return error to MYSQL client) if label keys do not exist or there are some non-matched rules.

Is it reasonable to detect such rules when adding rules? In another word, will it damage the performance? How does that strategy compared to `RuleFilter`, which is applied just before `FitRegion`?

Yisaer:
> Is it reasonable to detect such rules when adding rules? In another word, will it damage the performance? How does that strategy compared to RuleFilter, which is applied just before FitRegion?

I think there is no conflict between checking rules before `adding` or `FitRegion`. Rules could be legal when tidb-server add it, then if the stores down (or the label changed), the rules would be illegal. Thus we always need to filter rules before `FitRegion`. @xhebox

server/schedule/placement/rule_filter_test.go
Yisaer: updated.

server/schedule/operator_controller_test.go
Yisaer: Here is function sign:
```golagn
AddLabelsStore(storeID uint64, regionCount int, labels map[string]string) 
```

server/schedule/placement/rule_filter_test.go
xhebox: No, I mean cases like: `store="+dc=bj" constraints="+dc=bj,host=host1"`. or `store="+dc=bj,host=host1" constraints="+dc=bj"`.

xhebox: Eh.. What about cases like `store="+dc=bj" rule="+dc=bj,host=host1"`.

server/schedule/operator_controller_test.go
xhebox: Yeah, so you invoked with `storeID=1` for three times?

Yisaer: nice catch

server/schedule/placement/rule_filter.go
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package placement
+
+import "github.com/tikv/pd/server/core"
+
+// RuleFilter filter the rules if it doesn't match the conditions.
+type RuleFilter interface {
```
rleungx, line 19: Is there necessary to create a filter?

server/cluster/coordinator_test.go
```diff
@@ -908,6 +908,21 @@ func waitOperator(c *C, co *coordinator, regionID uint64) {
 }
 
 var _ = Suite(&testOperatorControllerSuite{})
+var _ = SerialSuites(&testOperatorControllerSerialSuite{})
```
rleungx, line 911: Why?

server/cluster/coordinator_test.go
```diff
@@ -908,6 +908,21 @@ func waitOperator(c *C, co *coordinator, regionID uint64) {
...
+
+type testOperatorControllerSerialSuite struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+}
+
+func (s *testOperatorControllerSerialSuite) SetUpSuite(c *C) {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	c.Assert(failpoint.Enable("github.com/tikv/pd/server/schedule/unexpectedOperator", "return(true)"), IsNil)
```
rleungx, line 920: Why we need failpoint here?

server/schedule/placement/rule_filter.go
Yisaer: Any advice?

server/schedule/placement/rule_manager.go
```diff
@@ -263,7 +263,13 @@ func (m *RuleManager) GetRulesForApplyRegion(region *core.RegionInfo) []*Rule {
 // FitRegion fits a region to the rules it matches.
 func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *RegionFit {
 	rules := m.GetRulesForApplyRegion(region)
-	return FitRegion(stores, region, rules)
+	filteredRules := make([]*Rule, 0, len(rules))
+	for _, rule := range rules {
+		if filterRule(rule, stores.GetStores()) {
+			filteredRules = append(filteredRules, rule)
+		}
+	}
```
disksing, line 271: I think it is a good idea but may be the wrong place to do it. Because the returned `RegionFit` will not include the rules dropped here. I'd like to sugguest to move the logic to some place inside the `fitWorker`.

Yisaer: I think it's ok that `RegionFit` will not include the dropped rules. What's the difference?

Yisaer: updated.

nolouch (approved):
lgtm

nolouch:
cc @xhebox

xhebox (commented):
LGTM

disksing:
Does it fix https://github.com/tikv/pd/issues/2255 ?

PR #3302, opened 2020-12-24
metrics: keep region health same as before

rleungx:

PR #3035 use an individual gauge to record the offline related region health.

This PR make the metrics look like the original one.

### Check List

Tests

- Manual test


disksing (approved):

lhy1024:
what does this pr change?

rleungx:
> what does this pr change?

Since PR #3035 use another gauge to record the offline related metrics, this PR is going to keep consistent with the original metrics

nolouch (approved):
lgtm

Issue #3299, opened 2020-12-23
placement: need more illegal label constraint check


xhebox:
## Bug Report

Follow #3293, separate the check part from that issue. These are checks that needs to be done when adding rules.

Known illegal cases so far:

1. label key not existed
2. `["+dc=dc1", "+dc=dc2"]`, contradicted label
3. `["+dc=dc1", "-dc=dc1"]`, contradicted label
4. No-match constraint
5. `["+dc=dc1", "+dc=dc1"]`, not necessarily an error, but it is duplicated

All these labels will pass the check successfully.

xhebox:
@Yisaer PTAL, I am still finding if there are other illegal constraint.

PR #3312, opened 2020-12-28
placement: push rules check down to rule manager

xhebox:

It is redundant to check rules two times: one in `api/rule.go`, another in `rule_manager.go`. This PR moves all checks in `api/rule.go` into `adjustRule` from `rule_manager.go`, which will recude one `O(N)` loop for all setting rules methods.

And now we can maintain only one single unique rule validation method `adjustRule` than two copies. My next PR on #3299 depends this PR.

### Check List

Tests

- Unit test
- Integration test

### Release note

- No release note


rleungx (approved):
LGTM

pkg/errs/errno.go
```diff
@@ -228,11 +228,6 @@ var (
 	ErrNetstatTCPSocks = errors.Normalize("TCP socks error", errors.RFCCodeText("PD:netstat:ErrNetstatTCPSocks"))
 )
 
-// hex error
-var (
```
rleungx, line 232: I noticed why remove this error?

xhebox: It is weird to wrap the hex error again with `ErrRuleContent`, looks like `[placement:PD:xx]invalid rule, [placement:PD:xx] invalid hex` or so... And it is only used here, in this function, after merging two check functions.

rleungx: But actually, we need to use an individual error type for those error come from the third-party package.

xhebox: Then, what about not using this error in adjustRule? And I will add it back.

xhebox: What about now?

rleungx: IMO, we can keep it as before and return the BadRequest when the error is ErrHexDecoding

xhebox: One more error check..? OK, later I will send a fix.

xhebox: Done

server/api/rule.go
```diff
@@ -80,14 +78,13 @@ func (h *ruleHandler) SetAll(w http.ResponseWriter, r *http.Request) {
 	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &rules); err != nil {
 		return
 	}
-	for _, rule := range rules {
-		if err := h.checkRule(rule); err != nil {
+	if err := cluster.GetRuleManager().SetKeyType(h.svr.GetConfig().PDServerCfg.KeyType).
+		SetRules(rules); err != nil {
+		if errs.ErrRuleContent.Equal(err) && errs.ErrHexDecodingString.Equal(err) {
```
rleungx, line 83: `||` ?

xhebox: Yeah. sorry. I am doing some other things.

Yisaer (approved):
LGTM

Issue #1905, opened 2019-11-06
api: use `mux.Middleware` to eliminate duplicate checks


disksing:
There are some common logic inside many http handlers. For example, the bootstrapped check for most handlers, and placement is enabled check for all placement rules APIs. We can use `mux.Middleware` to simplify it.
Original suggested from https://github.com/pingcap/pd/pull/1894
/cc @shafreeck

PR #3313, opened 2020-12-28
api: refactor, only check cluster existence in middleware

xhebox:

https://github.com/tikv/pd/issues/1905 is the source of using context to retrive cluster. Its aim is to eliminate common checks. But, all api handlers that invokes `getCluster` actually have a `*server.Server` or `*server.Handler` in their api handler structs, which means they can get the cluster directly without using context.

As our original aim is to remove redundant checks, we can totally avoid using context, using only middleware, for a consistent way to get extra infos in api handlers: passing `*server.Server` down to handler.

### Check List

Tests

- Unit test
- Integration test

### Release note

- No release note


JmPotato (approved):
LGTM. BTW, what do you think about unifying the different API handlers here to merge `h.GetRaftCluster()` and `h.svr.GetRaftCluster()`?

xhebox:
@JmPotato It sounds good to me. I will look into it later.

server/api/admin.go
```diff
@@ -44,7 +44,7 @@ func newAdminHandler(svr *server.Server, rd *render.Render) *adminHandler {
 // @Failure 400 {string} string "The input is invalid."
 // @Router /admin/cache/region/{id} [delete]
 func (h *adminHandler) HandleDropCacheRegion(w http.ResponseWriter, r *http.Request) {
-	rc := getCluster(r.Context())
+	rc := h.svr.GetRaftCluster()
```
rleungx, line 47: Do we need to check if `rc != nil`?

rleungx (approved):

PR #3203, opened 2020-11-24
schedulers: fix check hot region flow

lhy1024:
### What problem does this PR solve?

It will fix #3023 and #3198, it also be a part of #3201

It can be reviewd after #3192 and #3202 merged

### What is changed and how it works?

To deal with the problem of unstable heartbeats, this pr introduces three modifications #3023 
1. use the new aot
2. update the hot degree and anticount records after the sum of the intervals has been accumulated for one heartbeat interval, so as to avoid multiple too short heartbeats leading to rapid cooling or rapid heating. At the same time, we avoid that the whole system becomes too sensitive after aot modification.
3. skip the first heartbeat after transfer leader.

To solve the hot degree instability problem caused by inherited peer statistics #3198
1. We record the distribution of peers at the time of heartbeat and compare whether they are already joined in the next heartbeat, and do not put the old && cold peers into the cache anymore.

### Check List

Tests

- Unit test

PR #3316, opened 2020-12-30
Revert "schedulers: fix check hot region flow (#3203)"

nolouch:
This reverts commit c2688375e6e5271fe1f958a000e73d1a2317464e.

We need more test with this PR, in my test, the hot region is 
Under a random insert workload, a particularly large number of hotspot regions have been identified, resulting in more hotspot scheduling. We need to confirm whether this behavior is reasonable, so firstly revert this PR in the release branch.
image

### Check List

Tests

- Unit test
- Integration test


HunDunDM (approved):

lhy1024 (approved):
So sad

PR #3320, opened 2021-01-04
filter: add label for filterCounter

Yisaer:

Add label for `filterCounter`. If the `Filter` filter the `target` by comparing origin store and target store, we can record the origin store id into metrics.

Then we can query the metrics like this:

image

By this way, we can easily judge whether the filter is reasonable.
 

### Check List

Tests

- Unit test

### Release note

* No release note

### DIFF
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -6232,7 +6232,7 @@
               "format": "time_series",
               "hide": true,
               "intervalFactor": 2,
-              "legendFormat": "{{scope}}-store-{{source}}-{{type}}-target-{{target}}",
+              "legendFormat": "{{scope}}-{{type}}-{{source}}-{{target}}",
               "refId": "B"
             },
             {
@@ -6240,7 +6240,7 @@
               "format": "time_series",
               "hide": true,
               "intervalFactor": 2,
-              "legendFormat": "{{scope}}-store-{{source}}-{{type}}-target-{{target}}",
+              "legendFormat": "{{scope}}-{{type}}-{{source}}-{{target}}",
               "refId": "C"
             },
             {
@@ -6248,7 +6248,7 @@
               "format": "time_series",
               "hide": true,
               "intervalFactor": 2,
-              "legendFormat": "{{scope}}-store-{{source}}-{{type}}-target-{{target}}",
+              "legendFormat": "{{scope}}-{{type}}-{{source}}-{{target}}",
               "refId": "D"
             }
           ],
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -6226,6 +6226,30 @@
               "metric": "pd_scheduler_event_count",
               "refId": "A",
               "step": 4
+            },
+            {
+              "expr": "sum(delta(pd_schedule_filter{action=\"filter-target\",type=\"distinct-filter\"}[1m])) by (source, target, type, scope)",
+              "format": "time_series",
+              "hide": true,
+              "intervalFactor": 2,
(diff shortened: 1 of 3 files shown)


Yisaer:
@rleungx  PTAL

lhy1024:
LGTM

server/schedule/filter/filters.go
```diff
@@ -76,13 +87,23 @@ type Filter interface {
 	Target(opt *config.PersistOptions, store *core.StoreInfo) bool
 }
 
+// ComparingFilter is an interface to filter target store by comparing source and target stores
+type ComparingFilter interface {
+	Filter
+	// GetOldStoreID returns the source store when comparing.
+	GetOldStoreID() uint64
```
rleungx, line 94: Why not source store

Yisaer: updated.

server/schedule/filter/filters.go
```diff
@@ -166,6 +194,7 @@ type distinctScoreFilter struct {
 	stores    []*core.StoreInfo
 	policy    string
 	safeScore float64
+	oldStore  uint64
```
rleungx, line 197:
```suggestion
	srcStore  uint64
```

Yisaer: updated.

rleungx:
BTW, I think we can add the metrics but hide it

lhy1024:
> BTW, I think we can add the metrics but hide it

In some cases, we use tools to export metrics, which may not be convenient for later troubleshooting after the hidden monitoring

Yisaer:
> > BTW, I think we can add the metrics but hide it
> 
> In some cases, we use tools to export metrics, which may not be convenient for later troubleshooting after the hidden monitoring

We can manually display the hidden panel, then expose the metrics.

rleungx (approved):

metrics/grafana/pd.json
```diff
@@ -6226,6 +6226,30 @@
               "metric": "pd_scheduler_event_count",
               "refId": "A",
               "step": 4
+            },
+            {
+              "expr": "sum(delta(pd_schedule_filter{action=\"filter-target\",type=\"distinct-filter\"}[1m])) by (source, target, type, scope)",
+              "format": "time_series",
+              "hide": true,
+              "intervalFactor": 2,
+              "legendFormat": "{{scope}}-store-{{source}}-{{type}}-target-{{target}}",
```
rleungx, line 6235:
```suggestion
              "legendFormat": "{{scope}}-{{type}}-{{source}}-{{target}}",
```

server/schedule/filter/filters.go
```diff
@@ -455,28 +490,33 @@ func (f *ruleFitFilter) Source(opt *config.PersistOptions, store *core.StoreInfo
...
+	scope          string
+	fitter         RegionFitter
+	region         *core.RegionInfo
+	oldFit         *placement.RegionFit
+	srcLeaderStore uint64
 }
 
 // newRuleLeaderFitFilter creates a filter that ensures after transfer leader with new store,
 // the isolation level will not decrease.
 func newRuleLeaderFitFilter(scope string, fitter RegionFitter, region *core.RegionInfo, oldLeaderStoreID uint64) Filter {
```
rleungx, line 513:
```suggestion
func newRuleLeaderFitFilter(scope string, fitter RegionFitter, region *core.RegionInfo, srcLeaderStoreID uint64) Filter {
```

rleungx (approved):

lhy1024 (approved):

PR #3329, opened 2021-01-05
statistics: filter regions by hotDegree

disksing:

Check minHotDegree inside `HotPeerCache` and only return regions that are hot enough so we don't have to consider hotDegree in schedulers.

### Check List

Tests
- Unit test

### Release note
- No release note

### DIFF
--- a/server/statistics/hot_cache.go
+++ b/server/statistics/hot_cache.go
@@ -66,21 +66,21 @@ func (w *HotCache) Update(item *HotPeerStat) {
 }
 
 // RegionStats returns hot items according to kind
-func (w *HotCache) RegionStats(kind FlowKind) map[uint64][]*HotPeerStat {
+func (w *HotCache) RegionStats(kind FlowKind, minHotDegree int) map[uint64][]*HotPeerStat {
 	switch kind {
 	case WriteFlow:
-		return w.writeFlow.RegionStats()
+		return w.writeFlow.RegionStats(minHotDegree)
 	case ReadFlow:
-		return w.readFlow.RegionStats()
+		return w.readFlow.RegionStats(minHotDegree)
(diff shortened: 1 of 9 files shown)


lhy1024 (approved):

disksing:
@lhy1024 Pushed another commit, PTAL!

server/statistics/hot_cache.go
```diff
@@ -66,32 +66,28 @@ func (w *HotCache) Update(item *HotPeerStat) {
...
-func (w *HotCache) RandHotRegionFromStore(storeID uint64, kind FlowKind, hotDegree int) *HotPeerStat {
-	if stats, ok := w.RegionStats(kind)[storeID]; ok {
-		for _, i := range rand.Perm(len(stats)) {
-			if stats[i].HotDegree >= hotDegree {
-				return stats[i]
-			}
-		}
+func (w *HotCache) RandHotRegionFromStore(storeID uint64, kind FlowKind, minHotDegree int) *HotPeerStat {
+	if stats, ok := w.RegionStats(kind, minHotDegree)[storeID]; ok && len(stats) > 0 {
+		return stats[rand.Intn(len(stats))]
```
rleungx, line 82: Do we need a rand seed?

rleungx (commented):
It's better to update the comment of `RegionWriteStats()/RegionReadStats()` in `region_stat_informer.go`. The reset LGTM.

server/statistics/hot_cache.go
disksing: I think no need because rand seed is updated in `CreateServer`. P.S. the original `rand.Perm` is seed releated too.

rleungx (approved):

PR #3335, opened 2021-01-05
server/scheduler: Do not remove the down peer until the normal peers are enough

### LINKED ISSUE #3117: Do not remove the down peer until the normal peers are enough

Currently, when a region has down peer (the region's leader does not receive the heartbeat), we will handle it if the duration is beyond the `max-store-down-time`. See https://github.com/tikv/pd/blob/6dc21259f151f1447bdddff5d9951e8f9a2d551d/server/schedule/checker/replica_checker.go#L85

We will try to replace this peer with a new peer. 
https://github.com/tikv/pd/blob/6dc21259f151f1447bdddff5d9951e8f9a2d551d/server/schedule/checker/replica_checker.go#L229-L239

But if there is something wrong after the learner is added, there will have 4 peers (include 1 learner). Once it enters the checker again, the PD will remove the down peer instead due to the number of peers is beyond the max replicas.
https://github.com/tikv/pd/blob/6dc21259f151f1447bdddff5d9951e8f9a2d551d/server/schedule/checker/replica_checker.go#L218

In this case, the region will have three peers and two normal peers, and one learner. 

I think a better way is to make sure there are enough normal peers then remove the down peer.

AndreMouche:

Issue Number: close #3117 

Do not remove the down peer until the normal peers are enough

### Check List

Tests

- Unit test

### Release note

Do not remove the down peer until the normal peers are enough

### DIFF
--- a/server/schedule/checker/replica_checker.go
+++ b/server/schedule/checker/replica_checker.go
@@ -220,7 +220,7 @@ func (r *ReplicaChecker) checkLocationReplacement(region *core.RegionInfo) *oper
 
 func (r *ReplicaChecker) fixPeer(region *core.RegionInfo, storeID uint64, status string) *operator.Operator {
 	// Check the number of replicas first.
-	if len(region.GetPeers()) > r.opts.GetMaxReplicas() {
+	if len(region.GetVoters()) > r.opts.GetMaxReplicas() {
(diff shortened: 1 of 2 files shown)


disksing (approved):

Yisaer (approved):
LGTM

PR #3336, opened 2021-01-05
placement: check if rule can match any store when adding

### LINKED ISSUE #3299: placement: need more illegal label constraint check

xhebox:

close #3299 

PD did not check if rules can match any TiKV instances on adding. For example, there may be one TiKV with `dc=bj` and another one with `dc=sh`. Then one rule of `+dc=nj` won't match any instance. It is invalid and useless to add it.

This PR adds a check in the function `adjustRule`, which means every rule will be validated before adding.

**NOTE**: `checkRule` function comes from the [commit](https://github.com/xhebox/pd/commit/a1488e9466a8a63ed8b57c6cdc36e8d86b824524).

Placement rules in SQL is not a released feature, so it is OK to break the compatibility. 

`store` information is used exclusively for changing/adding rules.

### Check List

Tests

- Unit test
- Integration test

### Release note

- Disallow adding placement rule that won't match any TiKV store

### DIFF
--- a/server/schedule/checker/rule_checker_test.go
+++ b/server/schedule/checker/rule_checker_test.go
@@ -49,6 +49,7 @@ func (s *testRuleCheckerSuite) TestFixRange(c *C) {
 	s.cluster.AddLeaderStore(1, 1)
 	s.cluster.AddLeaderStore(2, 1)
 	s.cluster.AddLeaderStore(3, 1)
+	s.ruleManager.SetStores(s.cluster.GetStores())
 	s.ruleManager.SetRule(&placement.Rule{
 		GroupID:     "test",
 		ID:          "test",
@@ -83,6 +84,7 @@ func (s *testRuleCheckerSuite) TestAddRulePeerWithIsolationLevel(c *C) {
 	s.cluster.AddLabelsStore(3, 1, map[string]string{"zone": "z1", "rack": "r2", "host": "h1"})
 	s.cluster.AddLabelsStore(4, 1, map[string]string{"zone": "z1", "rack": "r3", "host": "h1"})
 	s.cluster.AddLeaderRegionWithRange(1, "", "", 1, 2)
+	s.ruleManager.SetStores(s.cluster.GetStores())
 	s.ruleManager.SetRule(&placement.Rule{
 		GroupID:        "pd",
 		ID:             "test",
@@ -154,6 +156,7 @@ func (s *testRuleCheckerSuite) TestFixOrphanPeers2(c *C) {
 	s.cluster.AddLabelsStore(2, 1, map[string]string{"foo": "bar"})
 	s.cluster.AddLabelsStore(3, 1, map[string]string{"foo": "baz"})
 	s.cluster.AddLeaderRegionWithRange(1, "", "", 1, 3)
+	s.ruleManager.SetStores(s.cluster.GetStores())
 	s.ruleManager.SetRule(&placement.Rule{
 		GroupID:  "pd",
 		ID:       "r1",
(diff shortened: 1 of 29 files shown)


server/schedule/checker/rule_checker_test.go
```diff
@@ -374,3 +384,116 @@ func (s *testRuleCheckerSuite) TestIssue3293(c *C) {
...
 	c.Assert(op.Desc(), Equals, "add-rule-peer")
 }
+
+func (s *testRuleCheckerSuite) TestIssue3299(c *C) {
+	s.cluster.AddLabelsStore(1, 1, map[string]string{"host": "host1"})
+	s.cluster.AddLabelsStore(2, 1, map[string]string{"dc": "sh"})
+	s.cluster.AddLeaderRegionWithRange(1, "", "", 1, 2)
+	s.ruleManager.SetStores(s.cluster.GetStores())
+
+	err := s.ruleManager.SetRule(&placement.Rule{
```
rleungx, line 394: How about making the test cases shorter since they have some common values in Rule?

xhebox: Done, that is the best I can do.

nolouch (commented):
please update the comment and title. is it only update stores when change the rules?

xhebox:
> please update the comment and title.

I don't what to update...

> is it only update stores when changing the rules?

Yes, I use `store` infomation exclusively for changing/adding rules.

server/schedule/placement/rule_manager.go
```diff
@@ -659,3 +664,11 @@ func (m *RuleManager) SetKeyType(h string) *RuleManager {
 	m.keyType = h
 	return m
 }
+
+// SetStores will update stores info for adjustRule()
+func (m *RuleManager) SetStores(h []*core.StoreInfo) *RuleManager {
```
nolouch, line 669: Its easy to forget `SetStores` before `SetRules`. I suggest passing in a closure to get stores when creating rule management. like: 
```
c.ruleManager = placement.NewRuleManager(c.storage, c.GetStores)
```

xhebox: Nice!

nolouch (approved):
lgtm

HunDunDM (commented):
What is this file `pkg/typeutil/sednSN3dL` used for?

server/schedule/placement/rule_manager.go
```diff
@@ -41,13 +41,15 @@ type RuleManager struct {
...
-	keyType string
+	keyType   string
+	getStores func() []*core.StoreInfo
 }
 
 // NewRuleManager creates a RuleManager instance.
-func NewRuleManager(store *core.Storage) *RuleManager {
+func NewRuleManager(store *core.Storage, getStores func() []*core.StoreInfo) *RuleManager {
 	return &RuleManager{
 		store:      store,
```
HunDunDM, line 51: Maybe we need to rename this to `storage` to avoid ambiguity?

HunDunDM (commented):
rest LGTM

server/schedule/placement/rule_manager.go
xhebox: Done.

server/schedule/placement/rule_manager.go
```diff
@@ -34,20 +34,22 @@ import (
...
 
 	// used for rule validation
-	keyType string
+	keyType   string
+	getStores func() []*core.StoreInfo
 }
 
 // NewRuleManager creates a RuleManager instance.
-func NewRuleManager(store *core.Storage) *RuleManager {
+func NewRuleManager(storage *core.Storage, getStores func() []*core.StoreInfo) *RuleManager {
```
Yisaer, line 49: I think it's better to use `StoreSetInformer` interface here.

Yisaer (changes requested):
Rest LGTM. plz remove `pkg/typeutil/sednSN3dL`

server/schedule/placement/rule_manager.go
xhebox: Addressed.

HunDunDM (approved):

nolouch (approved):

nolouch:
cc @Yisaer

Yisaer (approved):
LGTM

PR #3344, opened 2021-01-07
scheduler: do not balance the empty regions

rleungx:

After `drop table` or such kind of operation, there might be a large number of empty regions which cannot be merged immediately. The balance region scheduler will balance them if the stores' size have the difference. In this case, it may cause a lot of balance operation which affect the performance.

This PR will skip the empty region when selecting the region.

### Check List

Tests

- Unit test

### DIFF
--- a/server/schedule/opt/healthy.go
+++ b/server/schedule/opt/healthy.go
@@ -13,7 +13,12 @@
 
 package opt
 
-import "github.com/tikv/pd/server/core"
+import (
+	"github.com/tikv/pd/server/core"
+)
+
+// BalanceEmptyRegionThreshold is a threshold which allow balance the empty region if the region number is less than this threshold.
+var balanceEmptyRegionThreshold = 50
 
 // IsRegionHealthy checks if a region is healthy for scheduling. It requires the
 // region does not have any down or pending peers. And when placement rules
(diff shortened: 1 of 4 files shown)


server/schedulers/balance_region.go
```diff
@@ -167,6 +171,13 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 			}
 			log.Debug("select region", zap.String("scheduler", s.GetName()), zap.Uint64("region-id", region.GetID()))
 
+			// Skip the empty region
+			if region.GetApproximateSize() <= core.EmptyRegionApproximateSize && regionCount > balanceEmptyRegionThreshold {
```
disksing, line 175: it seems better to define a region filter and filter empty regions in `Rand{Pending, Follower, Leader, Learner}Region`.

server/schedule/opt/healthy.go
```diff
@@ -31,6 +36,12 @@ func IsHealthyAllowPending(cluster Cluster, region *core.RegionInfo) bool {
 	return len(region.GetDownPeers()) == 0
 }
 
+// IsEmptyRegionAllowBalance checks if a region is an empty region and can be balanced.
+func IsEmptyRegionAllowBalance(cluster Cluster, region *core.RegionInfo) bool {
+	return region.GetApproximateSize() > core.EmptyRegionApproximateSize ||
+		(region.GetApproximateSize() <= core.EmptyRegionApproximateSize && cluster.GetRegionCount() < balanceEmptyRegionThreshold)
```
HunDunDM, line 42:
```suggestion
	return region.GetApproximateSize() > core.EmptyRegionApproximateSize || cluster.GetRegionCount() < balanceEmptyRegionThreshold
```

HunDunDM (commented):
rest LGTM

disksing (approved):

HunDunDM (approved):

Issue #3076, opened 2020-10-15
Safe API cause unrecoverable cluster state


breeswish:
### What did you do?

We have a client invoked the following commands, in the scenario that a store (let's call it store_x) 's disk is broken:

```
curl -X POST 'http://{pd_ip}:{pd_port}/pd/api/v1/store/{store_id}/state?state=Tombstone'
pd-ctl remove-tombstone
```

There is no command calling *unsafe* here. However after executing these commands, all TiKV stores are now containing region peers (id=store_x) which are invalid. Error logs are continuously printed out (trying to connect store_x) in all TiKV stores, lead to disk full in short time.

To clean up these invalid region peers, all TiKV stores must be shutdown and user need to perform `unsafe-recover remove-fail-stores` TiKV-ctl command. This lead to service unavailable for sometime.

### What did you expect to see?

Commands / APIs not containing "unsafe" in the name should not lead to an unrecoverable cluster state

### What did you see instead?

User must manually stop all TiKV stores and invoke a highly unsafe command to clean up metadata.

### What version of PD are you using (`pd-server -V`)?

3.0

disksing:
I think the set state command should not be included in any document.

breeswish:
> I think the set state command should not be included in any document.

Now it has already been mentioned in a lot of places, as well as search engines. I was also told by a DBA that this API is used quite frequently although I don't know why.

Maybe we need to disable (like returning errors) this API and instead open a new one containing "unsafe" in the name?

disksing:
> > I think the set state command should not be included in any document.
> 
> Now it has already been mentioned in a lot of places, as well as search engines. I was also told by a DBA that this API is used quite frequently although I don't know why.
> 
> Maybe we need to disable (like returning errors) this API and instead open a new one containing "unsafe" in the name?

Good point.

AndreMouche:
DO you mean merge the following two commands: 

```
curl -X POST 'http://{pd_ip}:{pd_port}/pd/api/v1/store/{store_id}/state?state=Tombstone'
pd-ctl remove-tombstone
```

to the one which may similar to 

``` 
curl -X DELETE 'http://{pd_ip}:{pd_port}/pd/api/v1/unsafe/store/{store_id}'
```

which means we use this command to set the state to `Tombstone` and remove it. and we do not support to set the state to tombstone anymore in the previous `set state` API.


Or just rename the API to

```
[...]

PR #3347, opened 2021-01-08
server/api: add api to up a store from offline

AndreMouche:

It is part of #3076 
A new api to up a store from offline
```
curl -X POST 'http://{pd_ip}:{pd_port}/pd/api/v1/store/{store_id}
```

### Check List

Tests

- Unit test

Code changes

- Has new  HTTP API interfaces (Don't forget to [add the declarative for API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))

### DIFF
--- a/server/api/store_test.go
+++ b/server/api/store_test.go
@@ -257,6 +257,7 @@ func (s *testStoreSuite) TestStoreDelete(c *C) {
 }
 
 func (s *testStoreSuite) TestStoreUp(c *C) {
+	// s.TestStoresList(c)
 	table := []struct {
 		id     int
 		status int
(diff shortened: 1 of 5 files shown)


disksing:
how about `POST /store/restore/{id}` or `POST /store/recall/{id}`?

server/api/store_test.go
```diff
@@ -257,6 +257,7 @@ func (s *testStoreSuite) TestStoreDelete(c *C) {
 }
 
 func (s *testStoreSuite) TestStoreUp(c *C) {
+	// s.TestStoresList(c)
```
JmPotato, line 260: What does this line mean?

HunDunDM:
> how about `POST /store/restore/{id}` or `POST /store/recall/{id}`?

It is not a good idea to have verbs in the path of the URL. How about `[PATCH] /store/{id}?action=[delete|undelete]`. `action=undelete` is the operation of this PR. `action=delete` is equivalent to `[DELETE] /store/{id}`.

PR #3355, opened 2021-01-12
server/api: do not support setting store to Tombstone by any API

AndreMouche:

To close issue https://github.com/tikv/pd/issues/3076

This PR does not support setting a store's state to Tombstone in any API anymore. 

And for the situation in which the  previous API supported, we can use the following API instead:

####  For the situation that the store is physically destroyed and we want to start a new store with the same address: 

When the store is physically destroyed, we can delete it directly:
```
delete /store/{id}?force=true
```
then we can never up the store again, also we can directly start a new store with the same address.

We have changed two API
- the API to update the store's state does not support `Tombstone` anymore:
```
POST /store/{id}/state?state=${state}
```
state query string true "state" Enums(Up, Offline), `Tombstone` is not supported anymore

- The API to offline a store
```
DELETE /store/{id}?force=${force}
```
force is a boolean value:
- true: means the store shouldn't be up again and PD can remove it, also we can up a new store with the same address
- false: means the store is offline, it can be up again.

### Check List

- Unit test
- Integration test

Code changes

### Release note

API: Do not support setting the store to Tombstone by any API

### DIFF
--- a/server/api/store.go
+++ b/server/api/store.go
@@ -171,6 +171,7 @@ func (h *storeHandler) Get(w http.ResponseWriter, r *http.Request) {
 // @Tags store
 // @Summary Take down a store from the cluster.
 // @Param id path integer true "Store Id"
+// @Param force query string true "force" Enums(true, false), when force is true it means the store is physically destroyed and can never up gain
 // @Produce json
 // @Success 200 {string} string "The store is set as Offline."
 // @Failure 400 {string} string "The input is invalid."
@@ -196,13 +197,13 @@ func (h *storeHandler) Delete(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	h.rd.JSON(w, http.StatusOK, "The store is set as Offline or Tombstone.")
+	h.rd.JSON(w, http.StatusOK, "The store is set as Offline.")
 }
 
 // @Tags store
 // @Summary Set the store's state.
 // @Param id path integer true "Store Id"
-// @Param state query string true "state" Enums(Up, Offline, Tombstone)
+// @Param state query string true "state" Enums(Up, Offline)
 // @Produce json
 // @Success 200 {string} string "The store's state is updated."
 // @Failure 400 {string} string "The input is invalid."
--- a/server/api/store.go
+++ b/server/api/store.go
@@ -37,8 +37,7 @@ import (
 // MetaStore contains meta information about a store.
 type MetaStore struct {
 	*metapb.Store
-	StateName           string `json:"state_name"`
-	PhysicallyDestoryed bool   `json:"physically_destroyed"`
+	StateName string `json:"state_name"`
 }
 
 // StoreStatus contains status about a store.
@@ -77,9 +76,8 @@ const (
 func newStoreInfo(opt *config.ScheduleConfig, store *core.StoreInfo) *StoreInfo {
 	s := &StoreInfo{
 		Store: &MetaStore{
-			Store:               store.GetMeta(),
-			StateName:           store.GetState().String(),
-			PhysicallyDestoryed: store.IsPhysicallyDestoryAndOffline(),
+			Store:     store.GetMeta(),
+			StateName: store.GetState().String(),
 		},
 		Status: &StoreStatus{
 			Capacity:           typeutil.ByteSize(store.GetCapacity()),
--- a/server/api/store.go
+++ b/server/api/store.go
@@ -170,7 +170,7 @@ func (h *storeHandler) Get(w http.ResponseWriter, r *http.Request) {
 // @Summary Take down a store from the cluster.
 // @Param id path integer true "Store Id"
 // @Produce json
-// @Success 200 {string} string "The store is set as Offline or Tombstone."
+// @Success 200 {string} string "The store is set as Offline."
 // @Failure 400 {string} string "The input is invalid."
 // @Failure 404 {string} string "The store does not exist."
 // @Failure 410 {string} string "The store has already been removed."
@@ -189,13 +189,8 @@ func (h *storeHandler) Delete(w http.ResponseWriter, r *http.Request) {
 	_, force := r.URL.Query()["force"]
 	err = rc.RemoveStore(storeID, force)
 
-	if errors.ErrorEqual(err, errs.ErrStoreNotFound.FastGenByArgs(storeID)) {
-		h.rd.JSON(w, http.StatusNotFound, err.Error())
-		return
-	}
-
-	if errors.ErrorEqual(err, errs.ErrStoreTombstone.FastGenByArgs(storeID)) {
-		h.rd.JSON(w, http.StatusGone, err.Error())
+	if err != nil {
+		h.responseStoreErr(w, err, storeID)
 		return
 	}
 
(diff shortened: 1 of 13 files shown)


server/cluster/cluster.go
```diff
@@ -1020,18 +1022,25 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
...
+			return errors.Errorf("The store is already physically destroyed and offline")
+		}
+		// Remove an offline store should be OK, nothing to do.
+		if store.IsOffline() {
+			return nil
+		}
+	} else if store.IsPhysicallyDestoryAndOffline() {
+		return nil
+	}
+	newStore := store.Clone(core.OfflineStore(physicallyDestroyed))
```
rleungx, line 1040: prefer to use SetStoreState

server/cluster/cluster.go
```diff
@@ -1094,27 +1106,24 @@ func (c *RaftCluster) AttachAvailableFunc(storeID uint64, limitType storelimit.T
...
+		return errs.ErrStoreTombstone.FastGenByArgs(storeID)
 	}
 
-	newStore := store.Clone(core.SetStoreState(state))
-	log.Warn("store update state",
+	newStore := store.Clone(core.SetStoreState(metapb.StoreState_Up))
+	log.Warn("up store",
 		zap.Uint64("store-id", storeID),
-		zap.Stringer("new-state", state))
+		zap.Stringer("old-state", store.GetState()))
```
rleungx, line 1124: How about printing both the old and the new state?

AndreMouche: The information of the new state is already in the log's prefix  `up store..`?

server/cluster/cluster.go
AndreMouche: `SetStoreState` do not have args `physicallyDestroyed` since only offline a store can use the param

server/cluster/cluster.go
```diff
@@ -83,8 +83,9 @@ type RaftCluster struct {
 
 	running bool
 
-	clusterID   uint64
-	clusterRoot string
+	clusterID                               uint64
+	clusterRoot                             string
+	curPhysicallyDestroyedAndOfflineStoreID uint64
```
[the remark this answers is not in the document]
AndreMouche: The number of stores with the status offline and physically are limited to 1 in order to make sure the number of replications is `relatively` safe according to our last discussion @HunDunDM can you provide more reason?

And about the name, any good suggestion?

server/cluster/cluster.go
rleungx: I prefer to record them not only in the log message

server/cluster/cluster.go
```diff
@@ -1094,26 +1106,22 @@ func (c *RaftCluster) AttachAvailableFunc(storeID uint64, limitType storelimit.T
...
 	store := c.GetStore(storeID)
 	if store == nil {
 		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
 	}
-
-	if store.GetState() == metapb.StoreState_Tombstone && state != metapb.StoreState_Tombstone {
-		if err := c.checkStoreVersion(store.GetMeta()); err != nil {
-			return err
-		}
+	if store.IsTombstone() || store.IsPhysicallyDestoryAndOffline() {
```
rleungx, line 1117: We can return nil if the store is up

server/cluster/cluster.go
```diff
@@ -1020,18 +1022,25 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
...
+		if store.IsOffline() {
+			return nil
+		}
+	} else if store.IsPhysicallyDestoryAndOffline() {
+		return nil
+	}
+	newStore := store.Clone(core.OfflineStore(physicallyDestroyed))
 	log.Warn("store has been offline",
 		zap.Uint64("store-id", newStore.GetID()),
+		zap.Bool("physicalDestroyed", physicallyDestroyed),
```
rleungx, line 1043: Please change the format

server/cluster/cluster.go
```diff
@@ -1053,16 +1061,20 @@ func (c *RaftCluster) BuryStore(storeID uint64, force bool) error {
...
-			return errs.ErrStoreIsUp.FastGenByArgs()
-		}
-		log.Warn("force bury store", zap.Stringer("store", store.GetMeta()))
+		return errs.ErrStoreIsUp.FastGenByArgs()
+	}
+
+	if store.IsPhysicallyDestoryAndOffline() {
+		log.Warn("physically destroyed and offline store has been tombstone",
+			zap.Uint64("store-id", storeID),
+			zap.Uint64("should equal to store-id", c.curPhysicallyDestroyedAndOfflineStoreID))
```
rleungx, line 1076: ditto

server/api/store.go
```diff
@@ -37,7 +37,8 @@ import (
 // MetaStore contains meta information about a store.
 type MetaStore struct {
 	*metapb.Store
-	StateName string `json:"state_name"`
+	StateName           string `json:"state_name"`
+	PhysicallyDestoryed bool   `json:"physically_destroyed"`
```
Yisaer, line 41: I think `destroyed` or `removed` is better

server/api/store.go
```diff
@@ -231,16 +225,39 @@ func (h *storeHandler) SetState(w http.ResponseWriter, r *http.Request) {
...
-	err := rc.SetStoreState(storeID, metapb.StoreState(state))
+	var err error
+	switch metapb.StoreState(state) {
+	case metapb.StoreState_Up:
+		err = rc.UpStore(storeID)
+	case metapb.StoreState_Offline:
+		err = rc.RemoveStore(storeID, false)
+	default:
+		h.rd.JSON(w, http.StatusBadRequest, "invalid state")
+		return
```
Yisaer, line 236: I think we can add a case for `tombstone` and return the response that the tombstone is state is now allowed to be set.

AndreMouche: We already have the case in `TestStoreSetState`, the related test code
```
	// Invalid state.
	info = StoreInfo{}
	err = postJSON(testDialClient, url+"/state?state=Foo", nil)
	c.Assert(err, NotNil)
	err = postJSON(testDialClient, url+"/state?state=Tombstone", nil)
	c.Assert(err, NotNil)
```

Yisaer: I mean the response for `tombstone` should be `not allowed` instead of `invalid state`

server/cluster/cluster.go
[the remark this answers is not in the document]
Yisaer: I wonder what `curPhysicallyDestroyedAndOfflineStoreID` used for?

server/api/store.go
```diff
@@ -231,16 +223,39 @@ func (h *storeHandler) SetState(w http.ResponseWriter, r *http.Request) {
...
 	}
-
-	err := rc.SetStoreState(storeID, metapb.StoreState(state))
+	var err error
+	switch metapb.StoreState(state) {
+	case metapb.StoreState_Up:
+		err = rc.UpStore(storeID)
+	case metapb.StoreState_Offline:
+		err = rc.RemoveStore(storeID, false)
+	default:
```
rleungx, line 232: It seems no necessary to use default here?

server/api/store_test.go
```diff
@@ -254,6 +254,41 @@ func (s *testStoreSuite) TestStoreDelete(c *C) {
...
+	err = readJSON(testDialClient, url, store)
+	c.Assert(err, IsNil)
+	c.Assert(store.Store.State, Equals, metapb.StoreState_Offline)
+	c.Assert(store.Store.PhysicallyDestroyed, Equals, true)
+
+	// try to up store again failed because it is physically destroyed
+	status, _ = requestStatusBody(c, testDialClient, http.MethodPost, fmt.Sprintf("%s/state?state=Up", url))
+	c.Assert(status, Equals, http.StatusGone)
+	// reset store 6
+	s.cleanup()
```
rleungx, line 290: Prefer to use another suite

server/api/store_test.go
```diff
@@ -286,6 +324,11 @@ func (s *testStoreSuite) TestStoreSetState(c *C) {
 	err = readJSON(testDialClient, url, &info)
 	c.Assert(err, IsNil)
 	c.Assert(info.Store.State, Equals, metapb.StoreState_Up)
+
+	// store not found
+	info = StoreInfo{}
+	err = postJSON(testDialClient, s.urlPrefix+"/store/10086"+"/state?state=Offline", nil)
```
rleungx, line 330:
```suggestion
	err = postJSON(testDialClient, s.urlPrefix+"/store/10086/state?state=Offline", nil)
```

server/server.go
```diff
@@ -828,7 +828,7 @@ func (s *Server) SetReplicationConfig(cfg config.ReplicationConfig) error {
 		} else {
 			// NOTE: can be removed after placement rules feature is enabled by default.
 			for _, s := range raftCluster.GetStores() {
-				if !s.IsTombstone() && core.IsTiFlashStore(s.GetMeta()) {
+				if !(s.IsTombstone()) && core.IsTiFlashStore(s.GetMeta()) {
```
rleungx, line 831: Why?

tests/client/client_test.go
```diff
@@ -757,25 +757,28 @@ func (s *testClientSuite) TestGetStore(c *C) {
...
 
-	// Mark the store as tombstone.
-	err = cluster.BuryStore(store.GetId(), true)
+	// Mark the store as physically destroyed and offline.
+	err = cluster.RemoveStore(store.GetId(), true)
 	c.Assert(err, IsNil)
-	tombstoneStore := proto.Clone(store).(*metapb.Store)
-	tombstoneStore.State = metapb.StoreState_Tombstone
+	physicallyDestroyedStoreID := store.GetId()
+	// tombstoneStore := proto.Clone(store).(*metapb.Store)
```
rleungx, line 764: remove them?

PR #3362, opened 2021-01-14
placement: extract range list utility

disksing:

Extract range list utility to separated package. Later we can reuse it.

### Check List

Tests
- Unit test


server/schedule/placement/rule_list.go
```diff
@@ -105,103 +64,69 @@ type ruleContainer interface {
...
+	return rl.ranges[i].rules
 }
 
 func (rl ruleList) getRulesForApplyRegion(start, end []byte) []*Rule {
-	i := sort.Search(len(rl.ranges), func(i int) bool {
-		return bytes.Compare(rl.ranges[i].startKey, start) > 0
-	})
-	if i == 0 || i != len(rl.ranges) && (len(end) == 0 || bytes.Compare(end, rl.ranges[i].startKey) > 0) {
+	i, data := rl.rangeList.GetData(start, end)
+	if len(data) == 0 {
```
rleungx, line 128: Do we need to judge if i < 0?

disksing: ok

server/schedule/rangelist/builder.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// determine items for each range.
+	var l List
+	var si sortedItems
+	for i, p := range b.splitPoints {
+		switch p.ty {
+		case tStart:
+			si.insertItem(p.data, b.compare)
+		case tEnd:
+			si.deleteItem(p.data)
```
AndreMouche, line 102: It makes me confused that the data in the `splitPoints` are inserted by func `AddItem(start, end, data)`, it seems they add two points with the same data. So here it seems the `si` will only include the range whose `end` is empty?

server/schedule/rangelist/builder.go
```diff
@@ -0,0 +1,114 @@
...
+	for i, item := range si.items {
+		if item == del {
+			si.items = append(si.items[:i], si.items[i+1:]...)
+			return
+		}
+	}
+}
+
+// Build creates the key range list.
+func (b *Builder) Build() List {
```
AndreMouche, line 89: Could we implement it with a tree(interval tree or segment tree) that is easier to understand and maintain?

disksing:
I learned about interval tree and segment tree and found that they are not very suitable for our scenario. They are mainly used for range queries on discrete data points, but in our case the original data are ranges.
So reopened this PR. Please take a look @rleungx @Yisaer

server/schedule/rangelist/builder.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type splitPointType int
+
+const (
+	tStart splitPointType = iota
+	tEnd
+)
+
+// splitPoints represents key that exists in items.
```
rleungx, line 28:
```suggestion
// splitPoint represents key that exists in items.
```

rleungx (approved):
The rest LGTM

server/schedule/rangelist/builder.go
Yisaer: The `insert` and `delete` item here will be much `cpu`/`memory` consuming when we have lots of rules (maybe over 10 or 100 thousands) while using segment tree can solve this problem.  As the rules may not be so much, I think using array here is acceptable for now.

Yisaer (approved):
LGTM

PR #3363, opened 2021-01-15
*: remove some unused code

rleungx:

Remove some unused code.

Just as the above says.

### Check List

Tests

- Unit test

### DIFF
--- a/server/statistics/hot_peer_cache.go
+++ b/server/statistics/hot_peer_cache.go
@@ -377,7 +377,7 @@ func (f *hotPeerCache) getDefaultTimeMedian() *movingaverage.TimeMedian {
 }
 
 func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, bytes, keys float64, interval time.Duration) *HotPeerStat {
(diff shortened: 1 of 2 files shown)


server/statistics/hot_peer_cache.go
```diff
@@ -377,7 +377,7 @@ func (f *hotPeerCache) getDefaultTimeMedian() *movingaverage.TimeMedian {
 }
 
 func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, bytes, keys float64, interval time.Duration) *HotPeerStat {
-	if newItem == nil || newItem.needDelete {
+	if newItem.needDelete {
```
disksing, line 380: so it won't be nil?

JmPotato (approved):

server/statistics/hot_peer_cache.go
rleungx: Before we invoke this function, we will create a `newItem` and it cannot be nil in the current implementation.

disksing (approved):

PR #3373, opened 2021-01-20
tso: fix a global TSO generation bug

JmPotato:

Fix a bug that will cause the Global TSO generation to be critically wrong. If `maxTSO.Logical` is bigger than `maxLogical`, it will be set to zero but without adding the count.

Add the count to `maxTSO.Logical` after being set to zero.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/tests/server/tso/global_tso_test.go
+++ b/tests/server/tso/global_tso_test.go
@@ -121,7 +121,7 @@ func (s *testNormalGlobalTSOSuite) testGetNormalGlobalTimestamp(c *C, pdCli pdpb
 	c.Assert(err, IsNil)
 	c.Assert(resp.GetCount(), Equals, req.GetCount())
 	res := resp.GetTimestamp()
-	c.Assert(res.GetLogical(), Greater, int64(req.GetCount()))
+	c.Assert(res.GetLogical(), GreaterEqual, int64(req.GetCount()))
 	return res
 }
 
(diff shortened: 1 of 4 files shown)


Yisaer (commented):
Rest LGTM, plz add a unit test

Yisaer (approved):
LGTM

tests/server/tso/global_tso_test.go
```diff
@@ -498,6 +518,41 @@ func (s *testSynchronizedGlobalTSO) testGetTimestamp(ctx context.Context, c *C,
...
+	defer cluster.Destroy()
+	c.Assert(err, IsNil)
+
+	err = cluster.RunInitialServers()
+	c.Assert(err, IsNil)
+
+	waitAllLeaders(s.ctx, c, cluster, dcLocationConfig)
+
+	for _, dcLocation := range dcLocationConfig {
+		pdName := cluster.WaitAllocatorLeader(dcLocation)
```
rleungx, line 546: Is it duplicated with `waitAllLeaders`?

JmPotato: Updated.

tests/server/tso/global_tso_test.go
```diff
@@ -326,10 +330,9 @@ func (s *testTimeFallBackSuite) testGetTimestamp(c *C, n uint32) *pdpb.Timestamp
 	resp, err := tsoClient.Recv()
 	c.Assert(err, IsNil)
 	c.Assert(resp.GetCount(), Equals, uint32(n))
-
 	res := resp.GetTimestamp()
-	c.Assert(tsoutil.CompareTimestamp(res, tsoutil.GenerateTimestamp(time.Now(), 0)), Equals, 1)
-
+	c.Assert(res.GetPhysical(), Greater, int64(0))
```
rleungx, line 334: Why do we need to change it?

JmPotato: The original code compares `res` with zero. In addition to this, the new codes will also check whether the logical part is bigger than `count`.

rleungx: It seems physical is not zero before?

JmPotato: Oops, my mistake. Fixed.

rleungx (approved):

HunDunDM (approved):

PR #3375, opened 2021-01-20
tso, client: make the TSO computing unique

### LINKED ISSUE #3374: PD client should consider the suffix of Local and Global TSO
## Bug Report

Currently, we involve the suffix scheme to make every TSO unique. However, the PD client doesn't consider this part during the TSO computing, which will cause the TSO we get is not unique as expect.

For example, if we have two Local TSOs `t1` and `t2` which have the same physical and logical parts, but different suffix, one is 1 and another is 2. If the count we need for `t1` is 10 and `t2` is 11. After subtracting each count, they will have exactly the same TSO to use.

We need to make PD client consider the suffix of Local and Global TSO while doing the computing.

JmPotato:

Close #3374 and refine the TSO uniqueness tests.

* Make PD client consider the suffix of Local and Global TSO while doing the computing.
* Make the `CheckTSOUnique` become a method of `TestCluster`.

### Check List

Tests

- Unit test
- Integration test

### Related PR

pingcap/kvproto#713

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -440,7 +440,7 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
 		span := opentracing.StartSpan("pdclient.processTSORequests", opts...)
 		defer span.Finish()
 	}
-	count := len(requests)
+	count := int64(len(requests))
 	start := time.Now()
 	req := &pdpb.TsoRequest{
 		Header:     c.requestHeader(),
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -450,52 +450,55 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
...
 
 	if resp.GetCount() != uint32(len(requests)) {
 		err = errors.WithStack(errTSOLength)
-		c.finishTSORequest(requests, 0, 0, err)
+		c.finishTSORequest(requests, 0, 0, 0, err)
 		return err
 	}
 
-	physical, logical := resp.GetTimestamp().GetPhysical(), resp.GetTimestamp().GetLogical()
+	physical, logical, suffixBits := resp.GetTimestamp().GetPhysical(), resp.GetTimestamp().GetLogical(), resp.GetTimestamp().GetSuffixBits()
```
Yisaer, line 471: I think pass the `suffixBits` is too implementation specific, what about pass the sign itself?

Yisaer (commented):
Rest LGTM

client/client.go
JmPotato: I don't think it's necessary, for the client user, all they get for the TSO result is two `int64` integers, and the `suffixBits` here is rather considered as internal computing usage only. We should make it easy to implement the shift operation.

client/client.go
```diff
@@ -450,52 +450,55 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
...
 
-func (c *client) compareAndSwapTS(dcLocation string, physical, logical, n int64) {
+func (c *client) compareAndSwapTS(dcLocation string, physical, firstLogical, count int64) {
 	lastTSOInterface, ok := c.lastTSMap.Load(dcLocation)
 	if !ok {
 		var loaded bool
 		if lastTSOInterface, loaded = c.lastTSMap.LoadOrStore(dcLocation, &lastTSO{
 			physical: physical,
-			logical:  logical + n - 1,
+			// Save the highest logical part here
```
AndreMouche, line 485: one question, why not use `LoadOrStore` directly on L480, so L480-L490 can reduce to:
```
lastTSOInterface, loaded := c.lastTSMap.LoadOrStore(dcLocation, &lastTSO{
		physical: physical,
		// Save the highest logical part here
		logical: firstLogical + count - 1,
	})
	if !loaded {
		return
	}
```
which I think is easier to understand?

client/client.go
```diff
@@ -505,12 +508,12 @@ func tsLessEqual(physical, logical, thatPhysical, thatLogical int64) bool {
...
 }
 
-func (c *client) finishTSORequest(requests []*tsoRequest, physical, firstLogical int64, err error) {
+func (c *client) finishTSORequest(requests []*tsoRequest, physical, firstLogical int64, suffixBits uint32, err error) {
 	for i := 0; i < len(requests); i++ {
 		if span := opentracing.SpanFromContext(requests[i].requestCtx); span != nil {
 			span.Finish()
 		}
-		requests[i].physical, requests[i].logical = physical, firstLogical+int64(i)
+		requests[i].physical, requests[i].logical = physical, firstLogical+int64(i<<suffixBits)
```
AndreMouche, line 516: It confused me that you say `firstLogical` is the highest ts on L473, while each request's `logical` is equal or bigger than `firstLogical`.

client/client.go
JmPotato: You're right. `LoadOrStore` can already guarantee atomicity. I misunderstand it.

client/client.go
JmPotato: I revised the wording of the comment. PTAL again!

pkg/tsoutil/tso.go
```diff
@@ -44,6 +44,11 @@ func GenerateTS(ts *pdpb.Timestamp) uint64 {
 	return uint64(ts.GetPhysical())<<18 | uint64(ts.GetLogical())&0x3FFFF
 }
 
+// ComposeTS generate an `uint64` TS by passing the physical and logical parts.
+func ComposeTS(physical, logical int64) uint64 {
+	return uint64(physical)<<18 | uint64(logical)&0x3FFFF
```
AndreMouche, line 49: How about call `ComposeTS(ts.GetPhysical(), ts.GetLogical())` in the previous function `GenerateTS` , since the logical are quite similar between L49 and L44

tests/cluster.go
```diff
@@ -380,6 +380,11 @@ func (s *TestServer) GetTSOAllocatorManager() *tso.AllocatorManager {
 type TestCluster struct {
 	config  *clusterConfig
 	servers map[string]*TestServer
+	// tsPool is used to check the TSO uniqueness among the test cluster
+	tsPool struct {
+		sync.Mutex
+		pool map[uint64]struct{}
```
AndreMouche, line 386: Could we use `map[uint64]bool` instead since here we just need a `set`?

tests/client/client_test.go
```diff
@@ -283,31 +292,36 @@ func (s *clientTestSuite) TestLocalTSO(c *C) {
...
+					if err == nil {
+						ts1 = tsoutil.ComposeTS(p1, l1)
+						return true
+					}
+					c.Log(err)
+					return false
+				})
+				c.Assert(cluster.CheckTSOUnique(ts1), IsTrue)
+				time.Sleep(10 * time.Millisecond)
+				testutil.WaitUntil(c, func(c *C) bool {
```
AndreMouche, line 312: Similar logical in the same function is ugly, how about using a temporary to store the previous `TS`, which may similar to 
```

    prevTS := uint64(0)
	for i := 0; i < 5; i++ {
		var curTS uint64
		testutil.WaitUntil(c, func(c *C) bool {
			....
				curTS = tsoutil.ComposeTS(p1, l1)
			...
		})
		c.Assert(cluster.CheckTSOUnique(curTS), IsTrue)
		c.Assert(prevTS, Less, curTS)
		prevTS = curTS
		time.Sleep(10 * time.Millisecond)
	}
```

tests/cluster.go
JmPotato: Because the size of `struct{}` in Golang is zero, so it could be better than `bool` here technically?

AndreMouche (approved):

Yisaer (approved):
LGTM

client/client.go
```diff
@@ -450,52 +450,52 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
...
 	}
 
-	physical, logical := resp.GetTimestamp().GetPhysical(), resp.GetTimestamp().GetLogical()
-	// Server returns the highest ts.
-	logical -= int64(resp.GetCount() - 1)
-	c.compareAndSwapTS(dcLocation, physical, logical, int64(len(requests)))
-	c.finishTSORequest(requests, physical, logical, nil)
+	physical, logical, suffixBits := resp.GetTimestamp().GetPhysical(), resp.GetTimestamp().GetLogical(), resp.GetTimestamp().GetSuffixBits()
+	// logical is the highest ts here, we need to do the subtracting before we finish each TSO request.
+	firstLogical := logical - int64((count-1)<<suffixBits)
```
nolouch, line 473: the `logical` get from the response already `<<suffixBits`.  Is it correct to use `<< suffixBits` in here?

nolouch: we need to add a test about this. check if the logical part same as the server-generated.

JmPotato: The shift operation here acts on `count`, not `logical`. We shift `count` with `suffixBits` to make sure the computing will not be bothered by the suffix part.

JmPotato:
@nolouch @HunDunDM @rleungx PTAL, thx!

nolouch (approved):
lgtm

nolouch:
Plz update kvproto.

PR #3381, opened 2021-01-22
statistics: change read/write cache kinds to leader/peer

disksing:

Replace `ReadCache/WriteCache` with `LeaderCache/PeerCache` so it can reflect the intent better.
Old logic unchanged except for adding some more metrics.

### Check List

Tests
- Unit test

### Release note
- No release note


server/statistics/hot_peer_cache.go
```diff
@@ -209,41 +209,50 @@ func (f *hotPeerCache) CheckRegionFlow(region *core.RegionInfo) (ret []*HotPeerS
...
+func (f *hotPeerCache) CollectMetrics() {
 	for storeID, peers := range f.peersOfStore {
 		store := storeTag(storeID)
 		thresholds := f.calcHotThresholds(storeID)
-		hotCacheStatusGauge.WithLabelValues("total_length", store, typ).Set(float64(peers.Len()))
-		hotCacheStatusGauge.WithLabelValues("byte-rate-threshold", store, typ).Set(thresholds[byteDim])
-		hotCacheStatusGauge.WithLabelValues("key-rate-threshold", store, typ).Set(thresholds[keyDim])
-		// for compatibility
-		hotCacheStatusGauge.WithLabelValues("hotThreshold", store, typ).Set(thresholds[byteDim])
+		hotCacheStatusGauge.WithLabelValues("total_length", store, f.kind.String()).Set(float64(peers.Len()))
```
rleungx, line 224: I find the formats of labels are not consistent

disksing: em, do you mean it should be `total-length`?

rleungx: Yes. We have `byte-rate-threshold`, `hotThreshold`, and `total_length` three different kinds of format.

Issue #2138, opened 2020-02-17
build: add a ci test about Dockerfile


nolouch:
### Describe your feature request related problem: 
Meet a broken change which makes `Dockerfile` do not work, we cannot found it immediately. like: #2137

### Describe the feature you'd like:

add a ci test to verify it.

salman-bhai:
Hi just to understand are we looking at a smoke test something like this?
https://pythonspeed.com/articles/test-your-docker-build/

nolouch:
Hi @salman-bhai , only add a test to make sure the docker image can be built. maybe add a build task on [travis](https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile).

AnubhavUjjawal:
@nolouch could we add a `docker build` task to `Makefile's` https://github.com/tikv/pd/blob/c7aac753461a4ecbd136a05a108119008e4b5c8a/Makefile#L168 Then testing whether a docker image can be built or not will automatically be done in github workflow as one of the steps there in `.github/workflows/check.yaml` is `make check`?

nolouch:
already finished.

PR #3384, opened 2021-01-23
build: add tests for building docker image

AnubhavUjjawal:
Signed-off-by: AnubhavUjjawal <anubhavujjawal@gmail.com>

Issue #2138 #2136 reference difficulties faced when our `docker build`
is broken. 

The commit attempts to mitigate the difficulties by adding a
sanity test in `make check` command. `make check` command is run in the
github workflow as well, so we would be able to locate any issues caused
due to a broken `docker build` quickly.

### Check List

Tests

- Unit test
- Integration test

### Release note

No release note

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,12 +195,10 @@ errdoc: install-go-tools
 
 docker-build-test:
 	$(eval DOCKER_PS_EXIT_CODE=$(shell docker ps > /dev/null 2>&1 ; echo $$?))
-	@if [ $(DOCKER_PS_EXIT_CODE) -eq 0 ]; then \
-	docker build --no-cache -t tikv/pd .; \
-	else \
-	echo "Encountered problem while invoking docker cli. Is docker up and running? Failing docker-build-test. Check exit code for details."; \
-	exit $(DOCKER_PS_EXIT_CODE); \
+	@if [ $(DOCKER_PS_EXIT_CODE) -ne 0 ]; then \
+	echo "Encountered problem while invoking docker cli. Is the docker daemon running?"; \
 	fi
+	docker build --no-cache -t tikv/pd .
 
 check-missing-tests:
 	./scripts/check-missing-tests.sh
--- a/Makefile
+++ b/Makefile
@@ -199,6 +199,7 @@ docker-build-test:
 	docker build --no-cache -t tikv/pd .; \
 	else \
 	echo "Encountered problem while invoking docker cli. Is docker up and running? Skipping docker-build-test."; \
+	exit 1; \
 	fi
 
 check-missing-tests:
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,12 @@ errdoc: install-go-tools
 	./scripts/check-errdoc.sh
 
 docker-build-test:
-	docker build --no-cache -t tikv/pd .
+	$(eval DOCKER_PS_EXIT_CODE=$(shell docker ps > /dev/null 2>&1 ; echo $$?))
+	@if [ $(DOCKER_PS_EXIT_CODE) -eq 0 ]; then \
+	docker build --no-cache -t tikv/pd .; \
+	else \
+	echo "Encountered problem while invoking docker cli. Is docker up and running? Skipping docker-build-test."; \
+	fi
 
 check-missing-tests:
 	./scripts/check-missing-tests.sh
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -193,6 +193,9 @@ errdoc: install-go-tools
 	@echo "generator errors.toml"
 	./scripts/check-errdoc.sh
 
+docker-build-test:
+	docker build --no-cache -t tikv/pd .
```
nolouch, line 197: Thanks, maybe you can check whether docker is installed. Some developers may not have docker locally.

AnubhavUjjawal: what should be the exit code when docker is not installed ?

Yisaer: what about print the message and exit with a non-zero code

AnubhavUjjawal: @Yisaer @nolouch how about now?

Makefile
```diff
@@ -193,6 +193,15 @@ errdoc: install-go-tools
 	@echo "generator errors.toml"
 	./scripts/check-errdoc.sh
 
+docker-build-test:
+	$(eval DOCKER_PS_EXIT_CODE=$(shell docker ps > /dev/null 2>&1 ; echo $$?))
+	@if [ $(DOCKER_PS_EXIT_CODE) -eq 0 ]; then \
+	docker build --no-cache -t tikv/pd .; \
+	else \
+	echo "Encountered problem while invoking docker cli. Is docker up and running? Failing docker-build-test. Check exit code for details."; \
```
Yisaer, line 201:
```suggestion
	echo "Encountered problem while invoking docker cli. Is the docker daemon running?"; \
```

Yisaer (commented):
Rest LGTM

Makefile
AnubhavUjjawal: Hey, @Yisaer, sometimes when the users don't have permissions to run docker, they run docker commands using `sudo docker`. "Check exit code for details." will make them check exit code and hence point them to the correct issue in this case.

Also, "Is docker up and running?" feels a better error msg to me than "Is the docker daemon running?" because we are specifically using docker's cli client, and even if the docker daemon is up and running, there might be issues with the cli client which could fail the test. "Is docker up and running?" covers every scenario where there could be a fault from docker side of things.

For "Failing docker-build-test." I have no opinion. I can remove it if you want.

Please review my comments above. I will make the changes after you give the feedback again.

Yisaer: `Is the docker daemon running` is the default response for docker cli when docker daemon is not started or running, I just tend to keep the same behavior with docker cli.

AnubhavUjjawal: Do we not want to check for conditions where users don't have permission to run docker?
```
user@server:~$ docker ps
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json: dial unix /var/run/docker.sock: connect: permission denied
```

Yisaer: We can throw a simple response and expose the error message to let the users find the reason themselves.

AnubhavUjjawal: So would this work? Basically we are exposing the error msg but running the `docker build` command in the end anyway.
```makefile
docker-build-test:
	$(eval DOCKER_PS_EXIT_CODE=$(shell docker ps > /dev/null 2>&1 ; echo $$?))
	@if [ $(DOCKER_PS_EXIT_CODE) -ne 0 ]; then \
	echo "Encountered problem while invoking docker cli. Is the docker daemon running?"; \
	fi
	docker build --no-cache -t tikv/pd .
```

When docker daemon is not running, it would give output like this
```
$ make docker-build-test                                                                          
Encountered problem while invoking docker cli. Is the docker daemon running?
docker build --no-cache -t tikv/pd .
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
make: *** [docker-build-test] Error 1
```

In case docker is not installed, we would get this error:
```
$ make docker-build-test                                                                          
Encountered problem while invoking docker cli. Is the docker daemon running?
docker build --no-cache -t tikv/pd .
/bin/bash: docker: command not found
make: *** [docker-build-test] Error 127
```

Yisaer: I think This is ok to me.

AnubhavUjjawal: @Yisaer check now

Yisaer (approved):
LGTM

HunDunDM (approved):

PR #3388, opened 2021-01-25
server/api:support offline a store with physically destroyed 

AndreMouche:

It is part of #3076 
This PR support to offline a store and set it physically destroyed which means this store can never up and we can start a new store with the same address 

- The API to offline a store
```
DELETE /store/{id}?force=${force}
```
force is a boolean value:
- true: means the store shouldn't be up again and PD can remove it, also we can up a new store with the same address
- false: means the store is offline, it can be up again.

### Check List

Tests

- Unit test
- Integration test

Code changes

Related changes

https://github.com/pingcap/kvproto/pull/711

### Release note

Offline a store with force true means the store shouldn't be up again and PD can remove it, also we can up a new store with the same address

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1030,7 +1030,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 	}
 
 	if store.IsPhysicallyDestroyed() {
-		return errors.Errorf("The store is already physically destroyed")
+		return errs.ErrStoreDestroyed.FastGenByArgs(storeID)
 	}
 
 	newStore := store.Clone(core.OfflineStore(physicallyDestroyed))
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -932,8 +932,8 @@ func (c *RaftCluster) putStoreImpl(store *metapb.Store, force bool) error {
 
 	// Store address can not be the same as other stores.
 	for _, s := range c.GetStores() {
-		// It's OK to start a new store on the same address if the old store has been removed.
-		if s.IsTombstone() {
+		// It's OK to start a new store on the same address if the old store has been removed or physically destroyed.
+		if s.IsTombstone() || s.IsPhysicallyDestroyed() {
 			continue
 		}
 		if s.GetID() != store.GetId() && s.GetAddress() == store.GetAddress() {
@@ -1011,7 +1011,7 @@ func (c *RaftCluster) checkStoreLabels(s *core.StoreInfo) error {
 
 // RemoveStore marks a store as offline in cluster.
 // State transition: Up -> Offline.
-func (c *RaftCluster) RemoveStore(storeID uint64) error {
+func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) error {
 	c.Lock()
 	defer c.Unlock()
 
@@ -1021,7 +1021,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
 	}
 
 	// Remove an offline store should be OK, nothing to do.
-	if store.IsOffline() {
+	if store.IsOffline() && store.IsPhysicallyDestroyed() == physicallyDestroyed {
 		return nil
 	}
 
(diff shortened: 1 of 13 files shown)


server/cluster/cluster.go
```diff
@@ -1021,30 +1021,34 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
 	}
 
 	// Remove an offline store should be OK, nothing to do.
-	if store.IsOffline() {
+	if store.IsOffline() && store.IsPhysicallyDestroyed() == physicallyDestroyed {
```
Yisaer, line 1024: Do we need to compare `physicallyDestroyed` here ? i think this is unnecessary.

server/cluster/cluster.go
```diff
@@ -1021,30 +1021,34 @@ func (c *RaftCluster) RemoveStore(storeID uint64) error {
...
 	}
 
 	if store.IsTombstone() {
 		return errs.ErrStoreTombstone.FastGenByArgs(storeID)
 	}
 
-	newStore := store.Clone(core.SetStoreState(metapb.StoreState_Offline))
+	if store.IsPhysicallyDestroyed() {
+		return errors.Errorf("The store is already physically destroyed")
+	}
```
Yisaer, line 1034: Maybe we can return `nil` here as the store is already `PhysicallyDestroyed`

Yisaer (commented):
Rest LGTM

server/cluster/cluster.go
AndreMouche: Since physically-destroyed is not allowed from true to false.

server/cluster/cluster.go
AndreMouche: we needn't get the lock and set the status again when the status is already as we expected. By the way, it is possible a store may call the API with the following steps:
1. offline the store normally   DELETE store?force=false
2. Find the store is physically destroyed, do not want it to start again: DELETE store?force=true

Yisaer: > Find the store is physically destroyed, do not want it to start again: DELETE store?force=true

Why we need to forbid it? I think calling the api `delete` with force=false , then `force=true` is common to see.

server/cluster/cluster.go
Yisaer: I think we should define a errorCode like `ErrStoreTombstone` and to explain it as `The store have been force removed`

Yisaer (commented):
Rest LGTM

server/cluster/cluster.go
HunDunDM: If there is no judgment here, first `delete && force=false` and then `delete && force=true`, the latter will not be able to mark the `physicallyDestroyed` correctly. I think it needs to compare `physicallyDestroyed`.

server/core/store_option.go
```diff
@@ -82,6 +82,16 @@ func SetStoreState(state metapb.StoreState) StoreCreateOption {
 	}
 }
 
+// OfflineStore offline a store
+func OfflineStore(physicallyDestroyed bool) StoreCreateOption {
```
rleungx, line 86: Can we just use `func SetStoreState(state metapb.StoreState, Destroyed ...bool) StoreCreateOption` here?

server/core/store_option.go
```diff
@@ -82,6 +82,16 @@ func SetStoreState(state metapb.StoreState) StoreCreateOption {
...
+	return func(store *StoreInfo) {
+		meta := proto.Clone(store.meta).(*metapb.Store)
```
HunDunDM, line 88: Why is `Clone` needed here? It seems that the `StoreCreateOption ` in this file does not need to clone `meta`.

server/core/store_option.go
HunDunDM: I think `... bool` is not a good design.

HunDunDM (approved):

server/core/store_option.go
rleungx: Or maybe an option?

HunDunDM: how about
```

server/cluster/cluster.go
Yisaer: Thanks for clarifying

Yisaer (approved):
LGTM

PR #3407, opened 2021-02-01
api/store:set state do not support set from or to Tombstone anymore

AndreMouche:

To close issue https://github.com/tikv/pd/issues/3076

This PR does not support setting a store's state to or from `Tombstone` by the API `set store state`

- the API to update the store's state does not support `Tombstone` anymore:
```
POST /store/{id}/state?state=${state}
```
state query string true "state" Enums(Up, Offline), `Tombstone` is not supported anymore. Also, this PR does not support to up a store from `tombstone` or `physically destroyed`

### Check List

Tests

- Unit test

Related changes

- https://github.com/pingcap/kvproto/pull/711
- https://github.com/tikv/pd/pull/3388

### Release note 
- No release note


server/core/store_option.go
```diff
@@ -73,21 +73,30 @@ func SetStoreDeployPath(deployPath string) StoreCreateOption {
...
+		meta.State = metapb.StoreState_Up
+		store.meta = meta
+	}
+}
+
+// TombstoneStore set a store to tombstone.
+func TombstoneStore() StoreCreateOption {
+	return func(store *StoreInfo) {
+		meta := proto.Clone(store.meta).(*metapb.Store)
+		meta.State = metapb.StoreState_Tombstone
```
HunDunDM, line 99: Should `PhysicallyDestroyed` be changed to `false`? And maybe we should use `func setStoreState(state metapb.StoreState, physicallyDestroyed bool) StoreCreateOption` to reduce redundant code.

HunDunDM (commented):
rest LGTM

server/api/store.go
```diff
@@ -222,10 +222,18 @@ func (h *storeHandler) SetState(w http.ResponseWriter, r *http.Request) {
 		h.rd.JSON(w, http.StatusBadRequest, "invalid state")
 		return
 	}
+	var err error
+	switch metapb.StoreState(state) {
```
Yisaer, line 226: If the `state` is `tombstone`, I think we can return the response that `tombstone` is not supported to be set anymore.

server/core/store_option.go
Yisaer: Agreed, the api design lgtm while we can use `setStoreState` to reduce the duplicated code.

Yisaer (commented):
Rest LGTM

server/core/store_option.go
AndreMouche: As PR #3388 discussed, SetStoreState are replaced by following `StoreCreateOption`s:
```
- UpStore() 
- OfflineStore(physicallyDestroyed bool)
- TombstoneStore()
```
which have clearer definition and logic for each `StoreCreateOption`

HunDunDM (approved):

Yisaer (approved):
LGTM

nolouch (commented):
rest LGTM.

PR #3411, opened 2021-02-02
*: rename the package name of test

rleungx:

Some test package names are not meet the standard of golang.

This PR renames them and remove an unnecessary test also.

### Check List

Tests

- Unit test


disksing:
why do they not meet standard?

rleungx:
> why do they not meet standard?

See https://blog.golang.org/package-names

disksing:
I think the document does not cover test packages. Actually, use `foo_test` is commonly used. You can find many usages in https://github.com/golang/go

rleungx:
> I think the document does not cover test packages. Actually, `foo_test` style is very commonly used. You can find many usages in https://github.com/golang/go

Make sense. Shall we change other test package to foo_test?

disksing:
Both ways are good, depends on if we want to do black box test or white box test.

lhy1024 (approved):

PR #3412, opened 2021-02-02
test, tso: fix test and the Global TSO Allocator initialization bug

JmPotato:

* Fix a bug that causes `TestLeaderTransfer` test to cost too much time and refine the code logic.
* Fix the Global TSO Allocator initialization bug which will cause the new PD leader after transferring will be unable to provide TSO service.

* Fix `TestLeaderTransfer` test which will take too long time to run before by using `ResignLeader()`.
* Move Global TSO Allocator initialization to `campaignLeader()`.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1195,7 +1195,6 @@ func (s *Server) campaignLeader() {
 	// TSO service is strictly enabled/disabled by PD leader lease for 2 reasons:
 	//   1. lease based approach is not affected by thread pause, slow runtime schedule, etc.
 	//   2. load region could be slow. Based on lease we can recover TSO service faster.
-
 	ctx, cancel := context.WithCancel(s.serverLoopCtx)
 	defer cancel()
 	defer s.member.ResetLeader()
@@ -1206,6 +1205,15 @@ func (s *Server) campaignLeader() {
 	log.Info("setting up the global TSO allocator")
 	if err := s.tsoAllocatorManager.SetUpAllocator(ctx, config.GlobalDCLocation, s.member.GetLeadership()); err != nil {
 		log.Error("failed to set up the global TSO allocator", errs.ZapError(err))
+		return
+	}
+	alllocator, err := s.tsoAllocatorManager.GetAllocator(config.GlobalDCLocation)
+	if err != nil {
+		log.Error("failed to get the global TSO allocator", errs.ZapError(err))
+		return
+	}
+	if err := alllocator.Initialize(0); err != nil {
+		log.Error("failed to initialize the global TSO allocator", errs.ZapError(err))
 		return
 	}
 	// Check the cluster dc-location after the PD leader is elected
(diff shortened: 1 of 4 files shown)


tests/cluster.go
```diff
@@ -135,6 +135,7 @@ func (s *TestServer) Destroy() error {
 func (s *TestServer) ResignLeader() error {
 	s.Lock()
 	defer s.Unlock()
+	s.server.GetMember().ResetLeader()
```
HunDunDM, line 138: Why does it need to be explicitly executed here?

server/tso/allocator_manager.go
```diff
@@ -309,11 +309,7 @@ func (am *AllocatorManager) SetUpAllocator(parentCtx context.Context, dcLocation
 	switch dcLocation {
 	// For Global TSO Allocator
 	case config.GlobalDCLocation:
-		// Because Global TSO Allocator only depends on PD leader's leadership,
-		// so we can directly initialize it here.
-		if err := am.mu.allocatorGroups[dcLocation].allocator.Initialize(0); err != nil {
-			return err
-		}
+		return nil
```
HunDunDM, line 312: Why is it no longer possible to return err?

JmPotato: In the previous code, the initialization of the Global TSO Allocator is carried out with `SetUpAllocator()`. Now it has been moved to `campaignLeader()`, so there is no other work to be done here, just return `nil`.

tests/cluster.go
JmPotato: It is indeed a bit redundant, but I want to clearly express the leader’s resign behavior.

Yisaer (approved):
LGTM

server/tso/allocator_manager.go
```diff
@@ -309,11 +309,10 @@ func (am *AllocatorManager) SetUpAllocator(parentCtx context.Context, dcLocation
 	switch dcLocation {
```
rleungx, line 309: How about using:
```
if dcLocation != config.GlobalDCLocation {
    xxx
}
return nil

Comment (rleungx) [reply server/tso/allocator_manager.go:309]: BTW, it seems no need for error as the return value?

File: server/server.go [hunk 4]
```diff
@@ -1208,6 +1207,16 @@ func (s *Server) campaignLeader() {
...
+	alllocator, err := s.tsoAllocatorManager.GetAllocator(config.GlobalDCLocation)
+	if err != nil {
+		log.Error("failed to get the global TSO allocator", errs.ZapError(err))
+		return
+	}
+	if err := alllocator.Initialize(0); err != nil {
+		log.Error("failed to initialize the global TSO allocator", errs.ZapError(err))
+		return
+	}
+	defer s.tsoAllocatorManager.ResetAllocatorGroup(config.GlobalDCLocation)
```
rleungx, line 1219: The position of `defer` may need to change

rleungx (commented):
The rest LGTM

rleungx (approved):

Issue #3379, opened 2021-01-22
`/region` in the API may return a huge written_keys


HunDunDM:
## Bug Report

`/region` in the API returned a result as follows

```json
{
	"id": 29031958,
	"start_key": "******",
	"end_key": "******",
	"epoch": {
		"conf_ver": 1204,
		"version": 33785
	},
	"peers": [{
		"id": 29031959,
		"store_id": 5678246
	}, {
		"id": 29031960,
		"store_id": 1
	}, {
		"id": 29031961,
		"store_id": 5678148
	}],
	"leader": {
		"id": 29031961,
		"store_id": 5678148
	},
	"pending_peers": [{
		"id": 29031960,
		"store_id": 1
	}],
	"written_bytes": 18446744073709550591,
	"read_bytes": 0,
	"written_keys": 18446744073709551615,
	"read_keys": 0,
	"approximate_size": 93,
	"approximate_keys": 504338
}
```

### What did you do?

* `pd-ctl region`

### What did you expect to see?

* This is a cold region and should have smaller `written_bytes`.

### What did you see instead?

* written_bytes = 18446744073709550591 = uint64(-1024)
* written_keys = 18446744073709551615 = uint64(-1)

### Other information that may be useful

* The information may not be updated for a long time
  * It may be due to a bug fixed by #3116 
* In [TiDB code](https://github.com/pingcap/tidb/blob/v4.0.9/store/helper/helper.go#L433), the type of this field is `int64`, which will cause the following error.
  * `json: cannot unmarshal number 18446744073709550591 into Go struct field .written_bytes of type int64`
* This data comes from TiKV's region heartbeat data.

### What version of PD are you using (`pd-server -V`)?

* v4.0.9

rleungx:
Fixed by https://github.com/tikv/tikv/pull/10119

PR #3415, opened 2021-02-05
cluster: filter region heartbeats that contain overflow flow

HunDunDM:

* Related to #3379 

* Since the flow in the region heartbeat by TiKV may overflow, clean up this part of the stat.

### Check List

Tests

- Unit test

### Release note

- No release note


server/core/region.go
```diff
@@ -83,9 +83,16 @@ func classifyVoterAndLearner(region *RegionInfo) {
...
-const EmptyRegionApproximateSize = 1
+const (
+	// EmptyRegionApproximateSize is the region approximate size of an empty region
+	// (heartbeat size <= 1MB).
+	EmptyRegionApproximateSize = 1
+	// ImpossibleFlowSize is an impossible flow size (such as written_bytes, read_keys, etc.)
+	// It may be caused by overflow, refer to https://github.com/tikv/pd/issues/3379.
+	// They need to be filtered so as not to affect downstream.
+	// (flow size >= 1024TB)
+	ImpossibleFlowSize = 1 << 50
```
rleungx, line 94: IMO, we can directly judge if the value is less than zero and there is no need to create a variable.

nolouch:
Please merge this before 5.0 release.

nolouch (approved):

server/core/region.go
HunDunDM: Values very close to the upper limit of int64 are also very dangerous.

lhy1024 (approved):

PR #3416, opened 2021-02-05
config, tso: add labels configuration to update TSO configuration

### LINKED ISSUE #3414: Change the configuration way of Local TSO

For now, we use `local-tso.dc-location` and `local-tso.enable-local-tso` to configure the Local TSO, which is kind of inconvenient and inconsistent with TiDB and TiKV.

We should introduce the feature of `labels` like TiDB and TiKV, which will be more uniform and reasonable for users to use.

JmPotato:

Close #3414.

* Introduce `Labels` configuration.
* Make TSO use labels to configure.

### Check List

Tests

- Unit test
- Integration test

Related changes

- pingcap/tidb#19212
- pingcap/tidb#22639

### DIFF
--- a/conf/config.toml
+++ b/conf/config.toml
@@ -20,7 +20,15 @@ initial-cluster-state = "new"
 lease = 3
 tso-save-interval = "3s"
 
+## Make sure you set the "zone" label for this PD server before enabling its Local TSO service.
+# enable-local-tso = true
+
 enable-prevote = true
+
+[labels]
+## This means **this** PD server belongs to data center "dc-1". Setting it alone won't have any effect.
+## It should be used with enable-local-tso together if you want to make the Local TSO function work.
+# zone="dc-1"
 
 [security]
 ## Path of file that contains list of trusted SSL CAs. if set, following four settings shouldn't be empty
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -93,8 +93,11 @@ type Config struct {
 	// be automatically clamped to the range.
 	TSOUpdatePhysicalInterval typeutil.Duration `toml:"tso-update-physical-interval" json:"tso-update-physical-interval"`
 
-	// Local TSO service related configuration.
-	LocalTSO LocalTSOConfig `toml:"local-tso" json:"local-tso"`
+	// EnableLocalTSO is used to enable the Local TSO Allocator feature,
+	// which allows the PD server to generate Local TSO for certain DC-level transactions.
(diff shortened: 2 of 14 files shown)


conf/config.toml
```diff
@@ -20,8 +20,17 @@ initial-cluster-state = "new"
...
 
+## Make sure you set the "zone" label for this PD server before enabling its Local TSO service.
+# enable-local-tso = true
+
 enable-prevote = true
 
+[labels]
+foo= "bar"
+group= "abc"
+# This means **this** PD server belongs to data center "dc-1".
```
Yisaer, line 31: I think we can remove these 2 labels.

Yisaer (commented):
Rest LGTM

Yisaer (approved):
LGTM

conf/config.toml
```diff
@@ -20,8 +20,15 @@ initial-cluster-state = "new"
...
 tso-save-interval = "3s"
 
+## Make sure you set the "zone" label for this PD server before enabling its Local TSO service.
+# enable-local-tso = true
+
 enable-prevote = true
 
+[labels]
+# This means **this** PD server belongs to data center "dc-1".
+zone="dc-1"
```
HunDunDM, line 30: Need to comment out this section?

server/config/config.go
```diff
@@ -93,8 +93,11 @@ type Config struct {
 	// be automatically clamped to the range.
 	TSOUpdatePhysicalInterval typeutil.Duration `toml:"tso-update-physical-interval" json:"tso-update-physical-interval"`
 
-	// Local TSO service related configuration.
-	LocalTSO LocalTSOConfig `toml:"local-tso" json:"local-tso"`
+	// EnableLocalTSO is used to enable the Local TSO Allocator feature,
+	// which allows the PD server to generate Local TSO for certain DC-level transactions.
+	// To make this feature meaningful, user has to set the "zone" label for the PD server
+	// to indicate which DC this PD belongs to.
+	EnableLocalTSO bool `toml:"enable-local-tso" json:"enable-local-tso"`
```
HunDunDM, line 100: Incompatible changes need to be confirmed whether it has no effect on downstream.

conf/config.toml
JmPotato: No need to comment out because there are no side effects I think.

server/config/config.go
JmPotato: This feature is not released yet and there is no effect on downstream. It's ok to change.

HunDunDM (approved):

PR #3417, opened 2021-02-05
Fix hot peer cache threshold unstable when interval unstable

lhy1024:

Before PR, we add peer to the hot peer cache when its interval is less than normal, which causes topn will increase a lot when the heartbeat interval is unstable. We calculate the threshold by the 60th item in the hot peer cache, so the threshold will reduce to the default threshold while a lot of peers are added to the hot peer cache.

we add peer to the hot peer cache when its interval is less than normal only when it is hot too.

### Check List

Tests

- Unit test

### DIFF
--- a/server/statistics/hot_peer_cache.go
+++ b/server/statistics/hot_peer_cache.go
@@ -423,14 +423,14 @@ func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, bytes, k
 		newItem.AntiCount = oldItem.AntiCount
 	} else {
 		if f.isOldColdPeer(oldItem, newItem.StoreID) {
-			if newItem.isHot() {
+			if newItem.isFullAndHot() {
 				newItem.HotDegree = 1
 				newItem.AntiCount = hotRegionAntiCount
 			} else {
 				newItem.needDelete = true
 			}
 		} else {
-			if newItem.isHot() {
+			if newItem.isFullAndHot() {
 				newItem.HotDegree = oldItem.HotDegree + 1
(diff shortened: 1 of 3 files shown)


rleungx (commented):
LGTM, but the test is failed

nolouch (commented):
lgtm, plz fix the test.

nolouch (approved):
lgtm

nolouch:
PTAL @rleungx @HunDunDM

server/statistics/hot_peer_cache.go
```diff
@@ -385,11 +385,11 @@ func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, bytes, k
...
 		}
+		isHot := bytes/interval.Seconds() >= newItem.thresholds[byteDim] || keys/interval.Seconds() >= newItem.thresholds[keyDim]
+		if !isHot {
+			return nil
+		}
 		if interval.Seconds() >= RegionHeartBeatReportInterval {
-			isHot := bytes/interval.Seconds() >= newItem.thresholds[byteDim] || keys/interval.Seconds() >= newItem.thresholds[keyDim]
-			if !isHot {
-				return nil
-			}
```
Yisaer, line 392: I'm wondering what's difference here? It seems the result won't be changed with the given `isHot` and `interval.Seconds() >= RegionHeartBeatReportInterval`

lhy1024: it will directly return nil when the interval is less than RegionHeartBeatReportInterval

rleungx (approved):

PR #3422, opened 2021-02-18
schedule: revise region_scatter distribution for multi groups 

Yisaer:

ref https://github.com/pingcap/tidb/issues/22788

Currently region-scatter didn't work well for multi regions in different groups.

For example, there are 50 regions and 6 stores in the cluster. Each region belong to a dedicated group.

region scatter may locate them in the `store1/store2/store3`. Thus from each region in the group, the distribution is well. And for the whole cluster, the distribution is not well. 

During region scatter, we should consider whether the target store already have the most peer or leader, if it is, we should skip it.

### Check List

Tests

- Unit test

I also tested with the testcases in this document: https://docs.google.com/document/d/1cGZU2DCm0UuIueqAQP8TGqTivyavCvB_QiwIzqvP_rA/edit

All the test case results didn't change.

### Release note

* No release note


server/schedule/region_scatterer_test.go
```diff
@@ -197,16 +198,16 @@ func (s *testScatterRegionSuite) scatterSpecial(c *C, numOrdinaryStores, numSpec
...
-		c.Assert(float64(count), LessEqual, 1.1*float64(numRegions*3)/float64(numSpecialStores))
-		c.Assert(float64(count), GreaterEqual, 0.9*float64(numRegions*3)/float64(numSpecialStores))
+		c.Assert(float64(count), LessEqual, 1.3*float64(numRegions*3)/float64(numSpecialStores))
+		c.Assert(float64(count), GreaterEqual, 0.7*float64(numRegions*3)/float64(numSpecialStores))
 	}
 	for _, count := range countOrdinaryLeaders {
-		c.Assert(float64(count), LessEqual, 1.1*float64(numRegions)/float64(numOrdinaryStores))
-		c.Assert(float64(count), GreaterEqual, 0.9*float64(numRegions)/float64(numOrdinaryStores))
+		c.Assert(float64(count), LessEqual, 1.3*float64(numRegions)/float64(numOrdinaryStores))
+		c.Assert(float64(count), GreaterEqual, 0.7*float64(numRegions)/float64(numOrdinaryStores))
```
lhy1024, line 210: why change them？

server/schedule/region_scatterer_test.go
```diff
@@ -432,3 +433,37 @@ func (s *testScatterRegionSuite) TestSelectedStoreGC(c *C) {
 	_, ok = stores.getGroupDistribution("testgroup")
 	c.Assert(ok, Equals, false)
 }
+
+func (s *testScatterRegionSuite) TestRegionFromDifferentGroups(c *C) {
```
lhy1024, line 437: would you like to add more comments for this test?

server/schedule/region_scatterer_test.go
Yisaer: Because we should both guarantee the distribution for the single group and the whole cluster. The effect of the distribution have to become lower a little.

server/schedule/region_scatterer_test.go
Yisaer: updated.

nolouch:
What might become worse if there is concurrency? I remember that there was a situation where there was a concurrent split of different tables.

server/schedule/region_scatterer.go
```diff
@@ -307,15 +324,29 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string) *
...
 			if len(stores) == 0 {
 				context.selectedPeer.reset()
 				stores = r.collectAvailableStores(group, region, context)
 			}
-			if context.selectedPeer.put(peer.GetStoreId(), group) {
-				delete(stores, peer.GetStoreId())
-				targetPeers[peer.GetStoreId()] = peer
-				continue
+			storeCount := context.selectedPeer.storeTotalCount(peer.GetStoreId())
+			if storeCount < maxStoreTotalCount || storeCount == minStoreTotalCount {
```
rleungx, line 344: What is the difference between `storeCount < maxStoreTotalCount` and `storeCount == minStoreTotalCount`?

server/schedule/region_scatterer_test.go
```diff
@@ -197,16 +198,16 @@ func (s *testScatterRegionSuite) scatterSpecial(c *C, numOrdinaryStores, numSpec
 
 	// Each store should have the same number of peers.
 	for _, count := range countOrdinaryPeers {
-		c.Assert(float64(count), LessEqual, 1.1*float64(numRegions*3)/float64(numOrdinaryStores))
-		c.Assert(float64(count), GreaterEqual, 0.9*float64(numRegions*3)/float64(numOrdinaryStores))
+		c.Assert(float64(count), LessEqual, 1.25*float64(numRegions*3)/float64(numOrdinaryStores))
```
rleungx, line 201: I think we need to do more tests in the real environment to evaluate the drawback of this change.

server/schedule/region_scatterer.go
Yisaer: If the storeCounts are all the same( maxStoreTotalCount == minStoreTotalCount), we can pick this store as candidate.

server/schedule/region_scatterer.go
```diff
@@ -312,9 +330,23 @@ func (r *RegionScatterer) selectCandidates(region *core.RegionInfo, sourceStoreI
...
+		}
+		if count < minStoreTotalCount {
+			minStoreTotalCount = count
+		}
+	}
 	for _, store := range stores {
-		if filter.Target(r.cluster.GetOpts(), store, filters) && !store.IsBusy() {
-			candidates = append(candidates, store.GetID())
+		storeCount := context.selectedPeer.storeTotalCount(store.GetID())
+		if storeCount < maxStoreTotalCount || storeCount == minStoreTotalCount {
```
lhy1024, line 346: would you like to add more comment about it?

Yisaer: updated.

server/schedule/region_scatterer_test.go
```diff
@@ -440,3 +441,38 @@ func (s *testScatterRegionSuite) TestSelectedStoreGC(c *C) {
...
 }
+
+// TestRegionFromDifferentGroups test the multi regions. each region have its own group.
+// After scatter, the distribution for the whole cluster should be well.
+func (s *testScatterRegionSuite) TestRegionFromDifferentGroups(c *C) {
+	opt := config.NewTestOptions()
+	tc := mockcluster.NewCluster(opt)
+	// Add 6 stores.
+	storeCount := 6
+	for i := uint64(1); i <= uint64(6); i++ {
```
JmPotato, line 452:
```suggestion
	for i := uint64(1); i <= uint64(storeCount); i++ {
```

server/schedule/region_scatterer_test.go
```diff
@@ -440,3 +441,38 @@ func (s *testScatterRegionSuite) TestSelectedStoreGC(c *C) {
...
+		for i := uint64(1); i <= uint64(storeCount); i++ {
+			count := ss.storeTotalCount(i)
+			if count > max {
+				max = count
+			}
+			if count < min {
+				min = count
+			}
+		}
+		c.Assert(max-min, LessEqual, uint64(regionCount/6))
```
JmPotato, line 475: I think the difference is too big, what about setting it to `(regionCount/6)/2`?

server/schedule/region_scatterer.go
```diff
@@ -93,6 +94,23 @@ func (s *selectedStores) getDistributionByGroupLocked(group string) (map[uint64]
 	return nil, false
 }
 
+func (s *selectedStores) storeTotalCount(storeID uint64) uint64 {
```
rleungx, line 97: The naming is a little bit confusing.

server/schedule/region_scatterer.go
```diff
@@ -351,18 +386,26 @@ func (r *RegionScatterer) selectStore(group string, peer *metapb.Peer, sourceSto
...
+	minStoreGroupLeader := uint64(math.MaxUint64)
+	id := uint64(0)
+	for _, storeID := range leaderCandidateStores {
 		storeGroupLeaderCount := context.selectedLeader.Get(storeID, group)
 		if minStoreGroupLeader > storeGroupLeaderCount {
 			minStoreGroupLeader = storeGroupLeaderCount
 			id = storeID
 		}
 	}
+	if id < 1 {
```
rleungx, line 406: When will it happen?

server/api/region_test.go
```diff
@@ -294,9 +294,10 @@ func (s *testRegionSuite) TestScatterRegions(c *C) {
 	err := postJSON(testDialClient, fmt.Sprintf("%s/regions/scatter", s.urlPrefix), []byte(body))
 	c.Assert(err, IsNil)
 	op1 := s.svr.GetRaftCluster().GetOperatorController().GetOperator(601)
-	c.Assert(op1 != nil, Equals, true)
 	op2 := s.svr.GetRaftCluster().GetOperatorController().GetOperator(602)
-	c.Assert(op2 != nil, Equals, true)
+	op3 := s.svr.GetRaftCluster().GetOperatorController().GetOperator(603)
+	// At least one operator used to scatter region
+	c.Assert(op1 != nil || op2 != nil || op3 != nil, Equals, true)
```
rleungx, line 300: How about `IsTrue`?

server/schedule/region_scatterer.go
Yisaer: updated.

server/schedule/region_scatterer.go
Yisaer: Unreachable code, comment updated.

rleungx: And why do we keep it?

JmPotato (approved):

nolouch (approved):
lgtm

PR #3425, opened 2021-02-19
api: add dclocation in member response

Yisaer:

add dclocation in member response

ref https://github.com/pingcap/kvproto/pull/721

### Check List

Tests

- Unit test

### Release note

* No release note

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/pingcap/errcode v0.3.0
 	github.com/pingcap/errors v0.11.5-0.20201126102027-b0a155152ca3
 	github.com/pingcap/failpoint v0.0.0-20200702092429-9f69995143ce
-	github.com/pingcap/kvproto v0.0.0-20210204054616-1c1ed89bb167
+	github.com/pingcap/kvproto v0.0.0-20210219064844-c1844a4775d6
 	github.com/pingcap/log v0.0.0-20201112100606-8f1e84a3abc8
 	github.com/pingcap/sysutil v0.0.0-20201130064824-f0c8aa6a6966
 	github.com/prometheus/client_golang v1.2.1
@@ -51,7 +51,6 @@ require (
 )
 
 replace (
-	github.com/pingcap/kvproto => github.com/yisaer/kvproto v0.0.0-20210219053205-7ca18c2ae68a
 	github.com/sirupsen/logrus => github.com/sirupsen/logrus v1.2.0
 	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 )
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/BurntSushi/toml v0.3.1
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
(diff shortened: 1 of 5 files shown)


server/api/member.go
```diff
@@ -64,7 +64,12 @@ func getMembers(svr *server.Server) (*pdpb.GetMembersResponse, error) {
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
+	dclocationDistribution, err := svr.GetTSOAllocatorManager().GetClusterDCLocationsFromEtcd()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 	for _, m := range members.GetMembers() {
+		m.DcLocation = "global"
```
JmPotato, line 72: Does it mean this filed is `global` by default? What if we have a PD that doesn't configure `dc-location`? Should it be `global` also?

Yisaer: Yes.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/BurntSushi/toml v0.3.1
-	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
```
JmPotato, line 7: Is there something wrong with this dependency? Why does it change?

server/api/member.go
```diff
@@ -91,6 +96,19 @@ func getMembers(svr *server.Server) (*pdpb.GetMembersResponse, error) {
...
+		found := false
+		for dcLocation, serverIDs := range dclocationDistribution {
+			for _, serverID := range serverIDs {
+				if serverID == m.MemberId {
+					m.DcLocation = dcLocation
+					found = true
+					break
+				}
+			}
+			if found {
```
rleungx, line 108: What is the `found` used for? how about directly `return members, nil` if `serverID == m.MemberId`?

Yisaer: I think we can't. We should fulfill `dcLocation` for every member.

rleungx: oh, my mistake~

server/api/member_test.go
```diff
@@ -38,7 +38,12 @@ type testMemberAPISuite struct {
 }
 
 func (s *testMemberAPISuite) SetUpSuite(c *C) {
-	s.cfgs, s.servers, s.clean = mustNewCluster(c, 3)
+	s.cfgs, s.servers, s.clean = mustNewCluster(c, 3, func(cfg *config.Config) {
+		cfg.EnableLocalTSO = true
+		cfg.Labels = map[string]string{
+			"zone": "dc-1",
```
JmPotato, line 44: Better use `config.ZoneLabel` here.

Yisaer: addressed.

JmPotato (approved):

server/api/member.go
nolouch: How about empty if not config the zone label?

nolouch: "Global" is a concept in the tso allocator, and has no strong correlation with which DC localted.

Yisaer: good point.

rleungx (approved):

rleungx (approved):

PR #3428, opened 2021-02-19
config: provide instance-level config persistence

### LINKED ISSUE #3418: Provide instance-level config persistence

Because we now have a configuration named `Labels` which is instance-level, but the current persisted configurations are all cluster-level. Some TSO APIs will rely on this feature. We may need to introduce some changes to the persistence function to make it support instance-level configuration.

Provide instance-level config persistence feature.

JmPotato:

Close #3418.

Provide instance-level config persistence feature for `labels` and `enable-local-tso` configs.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/config/persist_options.go
+++ b/server/config/persist_options.go
@@ -19,6 +19,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 	"unsafe"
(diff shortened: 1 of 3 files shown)


server/config/persist_options.go
```diff
@@ -35,17 +36,30 @@ import (
...
+	ttl *cache.TTLString
+	// cluster-level configuration
 	schedule        atomic.Value
 	replication     atomic.Value
 	pdServerConfig  atomic.Value
 	replicationMode atomic.Value
 	labelProperty   atomic.Value
-	clusterVersion  unsafe.Pointer
+	// instance-level configuration
+	instance struct {
```
nolouch, line 59: The main reason for persist instance config is the dynamic setting？But I think users rarely change this configration, and restarting the instance can warn users of important changes.

JmPotato: Yes, the main reason to introduce the persisted instance-level config is to provide dynamic changing APIs, which the TiPocket test case needs also. According to this, I think making it persist is more reasonable for both test and useful purposes.

nolouch: The other point is if persist, to change the config should use API or PD control. It's strange for users about the instance level configuration because it's not works in instance configuration. I think should modify the test instead of causing some user behavior problems.

nolouch: cc @rleungx @disksing

rleungx: If this is only used for the tests, it is unnecessary to introduce the instance level config which may confuse the user when using the API.

PR #3431, opened 2021-02-22
client: support forwarding requests for pd client

rleungx:

To solve the network partition problem, we introduce the request forwarding mechanism to improve the availability of the system.

The main changes of this PR are:
- When unary RPC encounters the network error, fall back to send the request to the follower which can be reached
- When streaming RPC encounters the network error, fall back to create a new stream which the other side is a follower to send the request. And when the network back to normal, replace the stream of the follower with the leader/allocator.
- When the server receives the request, it will check whether this request is sent to itself. If not, the request should be sent to the leader/allocator according to the receiver address in the header.

The detailed design can be found in https://docs.google.com/document/d/1H_wkp8KmxlnpGidwIMIQzucrFjzCto-kFaz4ajsxK9o/edit#heading=h.efxhx09hxyyo

### Check List

Tests

- Unit test

Side effects

- Increased code complexity


client/client.go
```diff
@@ -425,6 +529,16 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
 				return
 			}
 			opts = extractSpanReference(requests[:pendingPlus1], opts[:0])
+			select {
+			case s, ok := <-streamCh:
+				if ok {
+					stream = s.stream
+					cancel = s.cancel
+					changedCh <- true
+					log.Info("tso stream has changed")
```
Yisaer, line 538:
```suggestion
					log.Info("tso stream has changed back to leader")
```

client/client.go
```diff
@@ -365,23 +445,47 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
...
+						streamCh = make(chan struct {
+							cancel context.CancelFunc
+							stream pdpb.PD_TsoClient
+						})
+						changedCh = make(chan bool)
+						// the goroutine is used to check the network and change back to the original stream
+						go c.checkAllocator(loopCtx, cancel, dc, url, streamCh, changedCh)
+					}
+				}
+			}
```
Yisaer, line 487: I think we can move these code into an extra function.

client/client.go
```diff
@@ -425,6 +529,16 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
...
+			select {
+			case s, ok := <-streamCh:
+				if ok {
+					stream = s.stream
+					cancel = s.cancel
+					changedCh <- true
+					log.Info("tso stream has changed")
+				}
+			default:
+			}
```
Yisaer, line 541: ditto.

client/client.go
```diff
@@ -365,23 +445,47 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
...
+				err = status.New(codes.Unavailable, "unavailable").Err()
+			})
+			rpcErr, ok := status.FromError(err)
+			// encounter the network error
+			if ok && isNetworkError(rpcErr.Code()) {
+				log.Info("fall back to use follower to redirect tso stream")
+				followerClient := c.followerClient()
+				if followerClient != nil {
+					// create the follower stream
+					cctx, cancel = context.WithCancel(loopCtx)
```
Yisaer, line 475: I think we can't directly cover the `cancel` here which will cause the goroutine leak by the previous cancel in line461.

Yisaer: I think we don't need to create a new context by`cctx, cancel = context.WithCancel(loopCtx)` here. We can directly pass the previous `cctx` which means `checkAllocator` don't need to handnle the redirectCancel. Then the `cancel` only be reference with 1 goroutine and it will be solved in the following error handling.

Yisaer (commented):
The change in `tsloop` basically LGTM. I suggest to add a switch like `ClientOption` to control whether enable the fallback proxy feature for pdclient.

client/base_client.go
```diff
@@ -149,8 +151,8 @@ func (c *baseClient) leaderLoop() {
 		failpoint.Inject("skipUpdateLeader", func() {
```
JmPotato, line 151: `skipUpdateLeader` should be renamed.

server/grpc_service.go
```diff
@@ -101,6 +114,36 @@ func (s *Server) Tso(stream pdpb.PD_TsoServer) error {
...
+			if err := redirectStream.Send(request); err != nil {
+				return errors.WithStack(err)
+			}
+			resp, err := redirectStream.Recv()
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			if err := stream.Send(resp); err != nil {
+				return errors.WithStack(err)
+			}
```
Yisaer, line 143: When we meet error here, do we need to clean the `redirectStream` and cancel the `cancel`?

client/client.go
```diff
@@ -288,18 +292,29 @@ func (c *client) GetAllMembers(ctx context.Context) ([]*pdpb.Member, error) {
...
+				cctx, cancel := context.WithTimeout(ctx, c.timeout)
+				resp, err = followerClient.GetMembers(cctx, req)
+				cancel()
+				if err == nil {
+					return resp.GetMembers(), nil
+				}
+			}
+		}
+
 		cmdFailDurationGetAllMembers.Observe(time.Since(start).Seconds())
```
Yisaer, line 313: I think we should add a label to distinguish whether this is a `redirect` or `direct` request.

client/client.go
```diff
@@ -365,23 +445,47 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
...
+			cctx, cancel = context.WithCancel(loopCtx)
+			stream, err = c.createTsoStream(cctx, cancel, pdpb.NewPDClient(cc))
+			failpoint.Inject("unreachableNetwork", func() {
+				cancel()
+				stream, cancel = nil, nil
+				err = status.New(codes.Unavailable, "unavailable").Err()
+			})
+			rpcErr, ok := status.FromError(err)
+			// encounter the network error
+			if ok && isNetworkError(rpcErr.Code()) {
```
Yisaer, line 470: When ` status.FromError(err)` return `nil, true`, will line 470 panic?

client/client.go
```diff
@@ -464,8 +578,9 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
 	}
 	count := int64(len(requests))
 	start := time.Now()
+	addr, _ := c.getAllocatorLeaderAddrByDCLocation(dcLocation)
```
Yisaer, line 581: Why not assert exists here? The `addr` might have been gc here.

Yisaer (commented):
Rest LGTM

server/grpc_service.go
```diff
@@ -1087,3 +1229,51 @@ func (s *Server) validateInternalRequest(header *pdpb.RequestHeader, onlyAllowLe
...
+	ctx, cancel := context.WithCancel(s.ctx)
+	go func(streamCtx context.Context, cancel context.CancelFunc, done chan struct{}) {
+		select {
+		case <-done:
+			return
+		case <-time.After(3 * time.Second):
+			cancel()
+		case <-streamCtx.Done():
+		}
+		<-done
```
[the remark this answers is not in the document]
rleungx: It seems we can't

server/grpc_service.go
rleungx: When the function returns, defer will cancel the stream.

client/client.go
rleungx: I have changed the way to send the unary request, PTAL

client/client.go
rleungx: `ok` will prevent the nil problem.

client/client.go
```diff
@@ -288,9 +292,9 @@ func (c *client) GetAllMembers(ctx context.Context) ([]*pdpb.Member, error) {
 	start := time.Now()
 	defer func() { cmdDurationGetAllMembers.Observe(time.Since(start).Seconds()) }()
 
-	ctx, cancel := context.WithTimeout(ctx, c.timeout)
-	resp, err := c.leaderClient().GetMembers(ctx, &pdpb.GetMembersRequest{
-		Header: c.requestHeader(),
+	cctx, cancel := context.WithTimeout(ctx, c.timeout)
```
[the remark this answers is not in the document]
rleungx: no special meaning

server/grpc_service.go
```diff
@@ -1087,3 +1229,51 @@ func (s *Server) validateInternalRequest(header *pdpb.RequestHeader, onlyAllowLe
...
+	}(ctx, cancel, done)
+	redirectStream, err := pdpb.NewPDClient(client).Tso(ctx)
+	done <- struct{}{}
+	return redirectStream, cancel, err
+}
```
Yisaer, line 1279: I think we can record a `TODO` for the case if goroutine here timeout when tso stream created successfully.

Yisaer (commented):
Rest LGTM

tests/client/client_test.go
```diff
@@ -367,6 +338,118 @@ func (s *clientTestSuite) TestCustomTimeout(c *C) {
...
+
+	c.Assert(failpoint.Disable("github.com/tikv/pd/client/unreachableNetwork1"), IsNil)
+	time.Sleep(200 * time.Millisecond)
+	r, err = cli.GetRegion(context.Background(), []byte("a"))
+	c.Assert(err, IsNil)
+	c.Assert(r, NotNil)
+}
+
+// case 1: unreachable -> normal
+func (s *clientTestSuite) TestGetTsoFromFollowerClient1(c *C) {
```
JmPotato, line 363: What about adding a case to test the Local TSO getting?

Yisaer (approved):
LGTM

sourya-deepsource:
> replace github.com/pingcap/kvproto => /home/ryan/kvproto

Checks, including tests and DeepSource, are failing due to this line in `go.mod`.

tests/client/client_test.go
rleungx: Yes, but maybe later?

rleungx:
PTAL @Yisaer @JmPotato @nolouch

client/client.go
```diff
@@ -249,6 +259,44 @@ func (c *client) updateTSODispatcher() {
 	})
 }
 
+func (c *client) leaderCheckLoop() {
+	defer c.wg.Done()
+
+	leaderCheckLoopCtx, leaderCheckLoopCancel := context.WithCancel(c.ctx)
+	defer leaderCheckLoopCancel()
+
+	ticker := time.NewTicker(leaderHealthCheckInterval)
```
nolouch, line 268: Is this too frequent?

client/client.go
```diff
@@ -249,6 +259,44 @@ func (c *client) updateTSODispatcher() {
...
+	defer cancel()
+	if cc, ok := c.clientConns.Load(c.GetLeaderAddr()); ok {
+		healthCli := healthpb.NewHealthClient(cc.(*grpc.ClientConn))
+		resp, err := healthCli.Check(ctx, &healthpb.HealthCheckRequest{Service: ""})
+		rpcErr, ok := status.FromError(err)
+		failpoint.Inject("unreachableNetwork1", func() {
+			resp = nil
+			err = status.New(codes.Unavailable, "unavailable").Err()
+		})
+		if (ok && isNetworkError(rpcErr.Code())) || resp.GetStatus() != healthpb.HealthCheckResponse_SERVING {
```
nolouch, line 292: If the request timeout, such as the firewall dropping the packet, Will it be network failure?

client/client.go
```diff
@@ -327,17 +375,16 @@ func (c *client) GetAllMembers(ctx context.Context) ([]*pdpb.Member, error) {
 	defer func() { cmdDurationGetAllMembers.Observe(time.Since(start).Seconds()) }()
 
 	ctx, cancel := context.WithTimeout(ctx, c.timeout)
-	resp, err := c.leaderClient().GetMembers(ctx, &pdpb.GetMembersRequest{
-		Header: c.requestHeader(),
-	})
+	req := &pdpb.GetMembersRequest{Header: c.requestHeader()}
+	ctx = grpcutil.NewReceiverMetadata(ctx, c.GetLeaderAddr())
```
nolouch, line 379: If the leader is in network partition, we may not update the members thought leader.

server/grpc_service.go
```diff
@@ -133,6 +177,15 @@ func (s *Server) Tso(stream pdpb.PD_TsoServer) error {
 
 // Bootstrap implements gRPC PDServer.
 func (s *Server) Bootstrap(ctx context.Context, request *pdpb.BootstrapRequest) (*pdpb.BootstrapResponse, error) {
+	receiverAddr := getReceiverAddr(ctx)
```
nolouch, line 180: Maybe we can create an issue about using gRPC Interceptor like [`grpc-proxy`](https://github.com/mwitkow/grpc-proxy) in the later.

Yisaer (commented):
Rest LGTM, btw i think we need to create a draft pr in tidb to test the new pdclient whether there exists any goroutine leak before we merge this pr.

client/client.go
```diff
@@ -398,23 +510,53 @@ func (c *client) handleDispatcher(dispatcherCtx context.Context, dc string, tsoD
...
+					if followerClient != nil {
+						log.Info("fall back to use follower to redirect tso stream", zap.String("addr", addr))
+						// create the follower stream
+						if addr, ok = c.getAllocatorLeaderAddrByDCLocation(dc); !ok {
+							continue
+						}
+						cctx, cancel = context.WithCancel(dispatcherCtx)
+						cctx = grpcutil.NewReceiverMetadata(cctx, addr)
+						stream, err = c.createTsoStream(cctx, cancel, followerClient)
+						if err == nil {
```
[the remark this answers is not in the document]
rleungx: The cancel will be called in line 568

server/grpc_service.go
```diff
@@ -101,6 +116,48 @@ func (s *Server) Tso(stream pdpb.PD_TsoServer) error {
...
+				if err != nil {
+					return err
+				}
+				redirectStream, cancel, err = s.createTsoRedirectStream(client, receiverAddr)
+				if err != nil {
+					return err
+				}
+				lastReceiverAddr = receiverAddr
+				errCh = make(chan error, 1)
+				go func(redirectStream pdpb.PD_TsoClient, errCh chan error) {
```
nolouch, line 136: Would check the goroutine number in your test？ I think it will growth quickly, it only close when stream meet error.

client/client.go
```diff
@@ -611,6 +765,41 @@ func (c *client) leaderClient() pdpb.PDClient {
...
+			return pdpb.NewPDClient(cc), addr
+		}
+		healthCancel()
+	}
+	return nil, ""
+}
+
+func (c *client) getClient() pdpb.PDClient {
+	if c.enableRedirection && atomic.LoadInt32(&c.leaderNetworkFailure) == 1 {
+		followerClient, addr := c.followerClient()
```
nolouch, line 794: we need to randomly get the client.

nolouch (commented):
rest LGTM

nolouch (commented):
lgtm.

nolouch (approved):

JmPotato (approved):

nolouch:
Can we merge it? cc @rleungx

PR #3432, opened 2021-02-22
client: revise client if mismatch leader id

Yisaer:

After transfer pd leader, the pd client may meet `mismatch leader id` several times during request global tso.

This request make `mismatch` error appearing at most once after transferring pd leader.

### Check List

Tests

- Unit test

### Release note

* No release note

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -26,7 +26,6 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
-	"github.com/tikv/pd/server"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
@@ -1086,5 +1085,5 @@ func addrsToUrls(addrs []string) []string {
 }
 
 func isMismatchLeader(err error) bool {
-	return strings.Contains(err.Error(), server.MismatchLeaderErr)
+	return strings.Contains(err.Error(), errs.MismatchLeaderErr)
 }
--- a/client/client.go
+++ b/client/client.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
@@ -352,8 +353,8 @@ func (c *client) handleDispatcher(loopCtx context.Context, dc string, tsoDispatc
 		cancel     context.CancelFunc
 		stream     pdpb.PD_TsoClient
 		opts       []opentracing.StartSpanOption
-		requests        = make([]*tsoRequest, maxMergeTSORequests+1)
-		needUpdate bool = false
+		requests   = make([]*tsoRequest, maxMergeTSORequests+1)
+		needUpdate = false
 	)
 	defer func() {
 		if cancel != nil {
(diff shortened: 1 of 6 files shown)


client/client.go
```diff
@@ -347,12 +347,13 @@ func (c *client) createTSODispatcher(dcLocation string) {
...
-		stream   pdpb.PD_TsoClient
-		opts     []opentracing.StartSpanOption
-		requests = make([]*tsoRequest, maxMergeTSORequests+1)
+		err        error
+		ctx        context.Context
+		cancel     context.CancelFunc
+		stream     pdpb.PD_TsoClient
+		opts       []opentracing.StartSpanOption
+		requests        = make([]*tsoRequest, maxMergeTSORequests+1)
+		needUpdate bool = false
```
JmPotato, line 356: Type annotation here is unnecessary.

tests/server/tso/global_tso_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	. "github.com/pingcap/check"
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/pdpb"
+	pdClient "github.com/tikv/pd/client"
```
JmPotato, line 26: I think this test better should be placed in `tests/client.go`?

tests/server/tso/global_tso_test.go
```diff
@@ -478,6 +479,20 @@ func (s *testSynchronizedGlobalTSO) TestSynchronizedGlobalTSO(c *C) {
 	c.Assert(err, IsNil)
 
 	cluster.WaitAllLeaders(c, dcLocationConfig)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	pClient, err := pdClient.NewClientWithContext(s.ctx, []string{cluster.GetServer(cluster.GetLeader()).GetAddr()}, pdClient.SecurityOption{})
+	c.Assert(err, IsNil)
+	// assert global tso
+	cluster.WaitLeader()
```
rleungx, line 487: Has it already been contained in `WaitAllLeaders`?

client/client.go
```diff
@@ -1066,3 +1083,7 @@ func addrsToUrls(addrs []string) []string {
 	}
 	return urls
 }
+
+func isMismatchLeader(err error) bool {
+	return strings.Contains(err.Error(), "mismatch leader id")
```
rleungx, line 1088: I prefer not to use the string directly since we may change the error message later by accident.

tests/server/tso/global_tso_test.go
Yisaer: `mismatch leader id` need `pd` enable local tso, so the test should be placed in `global_tso_test`.

JmPotato: You can check `TestGlobalAndLocalTSO` in `tests/client/client_tests.go`. Enabling Local TSO is ok in there.

Yisaer: updated.

JmPotato (approved):

pkg/errs/errs.go
```diff
@@ -33,3 +33,6 @@ func ZapError(err error, causeError ...error) zap.Field {
 	}
 	return zap.Field{Key: "error", Type: zapcore.ErrorType, Interface: err}
 }
+
+// MismatchLeaderErr indicates the the non-leader member received the requests which should be received by leader
+const MismatchLeaderErr = "mismatch leader id"
```
rleungx, line 38: How about moving it to errno.go?

rleungx (approved):

PR #3460, opened 2021-03-10
statistics: make hotPeerCache concurrency safe

Yisaer:

In some cases, pd might update the same region and dimension at the same time which will cause the concurrent problem which may let pd panic.

ref https://github.com/tikv/pd/issues/3461

This request made `dimStat` concurrency safe.

### Check List

Tests

- Unit test

### Release note

*  No release note

### DIFF
--- a/server/statistics/hot_peer.go
+++ b/server/statistics/hot_peer.go
@@ -15,7 +15,6 @@ package statistics
 
 import (
 	"math"
-	"sync"
 	"time"
 
 	"github.com/tikv/pd/pkg/movingaverage"
@@ -29,11 +28,8 @@ const (
 
 type dimStat struct {
 	typ int
-	mu  struct {
-		sync.RWMutex
-		Rolling     *movingaverage.TimeMedian  // it's used to statistic hot degree and average speed.
-		LastAverage *movingaverage.AvgOverTime // it's used to obtain the average speed in last second as instantaneous speed.
-	}
+	Rolling     *movingaverage.TimeMedian  // it's used to statistic hot degree and average speed.
+	LastAverage *movingaverage.AvgOverTime // it's used to obtain the average speed in last second as instantaneous speed.
 }
 
 func newDimStat(typ int) *dimStat {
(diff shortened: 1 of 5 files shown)


lhy1024:
It should be met only when handling some heartbeats from the same region at the same time.
Maybe we also need to learn why there is this case.

server/statistics/region_collection_test.go
```diff
@@ -14,8 +14,6 @@
 package statistics
 
 import (
-	"testing"
```
lhy1024, line 17: why remove it

lhy1024 (commented):
the rest LGTM

server/statistics/hot_peer_cache.go
```diff
@@ -54,24 +55,30 @@ var (
 
 // hotPeerCache saves the hot peer's statistics.
 type hotPeerCache struct {
-	kind           FlowKind
-	peersOfStore   map[uint64]*TopN               // storeID -> hot peers
-	storesOfRegion map[uint64]map[uint64]struct{} // regionID -> storeIDs
+	mu struct {
+		sync.RWMutex
```
rleungx, line 59: Will it have a performance problem?

server/statistics/region_collection_test.go
```diff
@@ -25,10 +23,6 @@ import (
 	"github.com/tikv/pd/server/schedule/placement"
 )
 
-func TestStatistics(t *testing.T) {
```
HunDunDM, line 28: Maybe it is better to delete `func Test(t *testing.T)` in `hot_peer_cache_test.go`.

server/statistics/hot_peer.go
```diff
@@ -34,11 +34,12 @@ type dimStat struct {
 
 func newDimStat(typ int) *dimStat {
 	reportInterval := RegionHeartBeatReportInterval * time.Second
-	return &dimStat{
```
HunDunDM, line 37: Why modify this?

server/statistics/hot_peer_cache.go
```diff
@@ -128,17 +137,19 @@ func (f *hotPeerCache) collectRegionMetrics(byteRate, keyRate float64, interval
 
 // CheckRegionFlow checks the flow information of region.
 func (f *hotPeerCache) CheckRegionFlow(region *core.RegionInfo) (ret []*HotPeerStat) {
+	f.mu.Lock()
```
HunDunDM, line 140: `RLock()`?

server/statistics/hot_peer_cache.go
HunDunDM, line 59: Due to the low probability of conflict,`Mutex` should have better performance than `RWMutex`.

PR #3462, opened 2021-03-10
cluster: save to the region cache when pending-peers or down-peers change

### LINKED ISSUE #3456: Abnormalities in TiKV may cause slow PD heartbeat processing
### What did you do?

* Some TiKVs are damaged

* The cluster is working normally

### What did you see instead?

* Region heartbeat processing is slow
* The region information is not updated in time
* A large number of scheduling operations failed

* v4.0.10

HunDunDM:

close #3456

* Before: save to the region cache when pending-peers or down-peers exists
* After: save to the region cache when pending-peers or down-peers change

### Check List

Tests

- Unit test

Related changes

### Release note

- save to the region cache when pending-peers or down-peers change


nolouch (approved):

server/cluster/cluster.go
```diff
@@ -591,11 +591,13 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 			}
 			saveCache, needSync = true, true
 		}
-		if len(region.GetDownPeers()) > 0 || len(region.GetPendingPeers()) > 0 {
-			saveCache = true
+		if !core.SortedPeersStatsEqual(region.GetDownPeers(), origin.GetDownPeers()) {
+			log.Debug("down-peers changed", zap.Uint64("region-id", region.GetID()))
```
rleungx, line 595: How about printing the difference?

HunDunDM: Personally, I think pending is more common, so I used the `Debug` level.

lhy1024 (approved):

PR #3467, opened 2021-03-11
statistics: fix the bug that the isolation level is wrong when the store lacks label

### LINKED ISSUE #3455: When there are redundant labels in location-labels, the metrics is wrong
## Bug Report

* `location-labels = "region,zone,rack,host"`
* TiKV only used `zone,host`

image

### What did you see instead?

* Should be `zone`

* v3.0.12

HunDunDM:

close #3455

* statistics: fix the bug that the isolation level is wrong when the store lacks label

**example**

* pd location-labels: region,host
* tikv1 label: {host: 1}
* tikv2 label: {host: 2}
* tikv3 label: {host: 3}
* Before
  * image
* After
  *  image

### Check List

Tests

- Unit test

Related changes

### Release note

- fix the bug that the isolation level is wrong when the store lacks label


server/statistics/region_collection.go
```diff
@@ -296,17 +280,35 @@ func getRegionLabelIsolation(stores []*core.StoreInfo, labels []string) string {
...
+			if len(stores) > 1 {
+				res = append(res, stores)
+			}
+		}
+	} else {
+		for _, stores := range valueStoresMap {
+			stores = append(stores, emptyValueStores...)
 			res = append(res, stores)
 		}
 	}
```
JmPotato, line 314: What about adding some comments to these different situations? For example, for `valueStoresMap` is empty, it means no store has this label configured, for `emptyValueStores` is empty, it means every store has this label configured, and for `emptyValueStores` is not empty, it means this is at least one store has this label configured but not all.

HunDunDM: good idea

JmPotato (approved):

PR #3482, opened 2021-03-16
*: add retry mechanism for connecting leader

rleungx:

We need to retry several times when the leader connection is failed to avoid unnecessary forwarding.
BTW, since TiKV uses TSO stream in a unary way, it will print lots of logs in the follower which is responsible to forward the requests when there is a network problem.

This PR adds the retry mechanism for connecting the leader when there is a network problem and changes TSO stream log to debug level. 

### Check List

Tests

- Unit test

Related changes

### Release note

- Add the retry mechanism for connecting leader and change the noisy log to debug level

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -173,6 +173,8 @@ const (
 	tsLoopDCCheckInterval = time.Minute
 	maxMergeTSORequests   = 10000 // should be higher if client is sending requests in burst
 	maxInitClusterRetries = 100
+	retryInterval         = 1 * time.Second
+	maxRetryTimes         = 5
 )
 
 // LeaderHealthCheckInterval might be chagned in the unit to shorten the testing time.
@@ -504,6 +506,8 @@ func (c *client) handleDispatcher(dispatcherCtx context.Context, dc string, tsoD
 			stream pdpb.PD_TsoClient
 		}
 		changedCh chan bool
+		// the number of retrying to connect leader
+		retryTimes uint64
 	)
 	defer func() {
 		log.Info("[pd] exit tso dispatcher", zap.String("dc-location", dc))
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -532,37 +536,51 @@ func (c *client) handleDispatcher(dispatcherCtx context.Context, dc string, tsoD
...
+								changedCh = make(chan bool)
+								// the goroutine is used to check the network and change back to the original stream
+								go c.checkAllocator(dispatcherCtx, cancel, dc, url, streamCh, changedCh)
+							}
 						}
+					} else {
+						// not reach the max retry time, sleep for a while and retry
+						time.Sleep(retryInterval)
+						cancel()
+						continue
```
Yisaer, line 575: Can we extract these as a extra function? I think `retryTimes` can be maintained in this function.

Yisaer (commented):
Rest LGTM

client/client.go
rleungx: `retryTimes` can not be maintained in this function since it is used to reconnect the leader not follower

client/client.go
```diff
@@ -639,6 +606,76 @@ func (c *client) handleDispatcher(dispatcherCtx context.Context, dc string, tsoD
...
+	stream   pdpb.PD_TsoClient
+	cancel   context.CancelFunc
+	streamCh streamCh
+	changeCh chan bool
+}
+
+func (c *client) tryConnect(dispatcherCtx context.Context, dc string) (connectionContext, error) {
+	var (
+		url        string
+		networkErr uint64
```
nolouch, line 619:
```suggestion
		networkErrCouter uint64
```

nolouch (commented):
rest LGTM

Yisaer (approved):
LGTM

nolouch (approved):
lgtm

PR #3483, opened 2021-03-16
schedule: fix panic during hot schedule

Yisaer:

`hot-region` scheduler may cause PD panic in the following order:

1.  load `storesLoads` before hot region scheduling
```golang
	storesLoads := cluster.GetStoresLoads()
```
2. receive new `PutStore` request
```golang
// put Store in grpc_service.go
if err := rc.PutStore(store); err != nil {
	return nil, status.Errorf(codes.Unknown, err.Error())
}
```
3. select candidates from `raft.cluster`
```golang
candidates = bs.cluster.GetStores()
```

This will cause make `pickDstStores` panic in following code because `stLoadDetail` didn't have the storeID which candidates hold as it was newly put before.
```golang
detail := bs.stLoadDetail[store.GetID()]
```

Make the candidates should be searched from `stLoadDetail` during hot region schedule

### Check List

Tests

- Unit test

### Release note

* No release note

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -732,7 +732,6 @@ func (bs *balanceSolver) filterDstStores() map[uint64]*storeLoadDetail {
 	if srcStore == nil {
 		return nil
 	}
-
 	switch bs.opTy {
 	case movePeer:
 		filters = []filter.Filter{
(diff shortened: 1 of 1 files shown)


server/schedulers/hot_region.go
```diff
@@ -753,7 +754,15 @@ func (bs *balanceSolver) filterDstStores() map[uint64]*storeLoadDetail {
...
-		candidates = bs.cluster.GetFollowerStores(bs.cur.region)
+		regionStores := make(map[uint64]*core.StoreInfo, 0)
+		for _, store := range bs.cluster.GetFollowerStores(bs.cur.region) {
+			regionStores[store.GetID()] = store
+		}
+		for storeID := range bs.stLoadDetail {
+			if store, ok := regionStores[storeID]; ok {
+				candidates = append(candidates, store)
+			}
+		}
```
lhy1024, line 765:
```suggestion
		for _, store := range bs.cluster.GetFollowerStores(bs.cur.region) {
			if _, ok := bs.stLoadDetail[store.GetID()]; ok {
				candidates = append(candidates, store)
			}
		}
```

Yisaer: updated.

lhy1024 (approved):

HunDunDM (approved):

PR #3498, opened 2021-03-22
server: fix the problem that get tso cannot work well in network partition environment

rleungx:

TiKV will get TSO in a fixed interval, while it cannot work well when the network partition happens. This is because after sending the request to the forwarding stream, it will continue to handle the next request. But TiKV uses this streaming method in a unary way, the next request will be EOF which will cancel the forwarding mechanism.

Closes https://github.com/tikv/pd/issues/3500

This PR changes the way to handle forwarding TSO requests to avoid handling the next request before the forwarding stream responds.

### Check List

Tests

- Manual test

Related changes

### Release note

- Fix the problem that get TSO cannot work well in a network partition environment

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -420,13 +420,13 @@ func (c *client) createTsoStream(ctx context.Context, cancel context.CancelFunc,
 	return stream, err
 }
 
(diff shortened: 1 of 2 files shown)


JmPotato (commented):
I think we should add a TODO to mark we should update this part of logic after TiKV uses the stream way to get TSO. Rest LGTM.

nolouch (approved):
lgtm

JmPotato (approved):

PR #3528, opened 2021-03-25
scheduler: add cooldown time after transfer hot leader

### LINKED ISSUE #3527: Unstable hot region scheduler after load base split
### What did you do?

```
[root@tidb-bench-0 case2]# cat test.sh 
for i in {1..3}
do
echo $i
mysql -h10.68.208.230 -P4000 -u root < reset.sql
sysbench --rand-type=uniform --table_size=10000 --range_size=100 --point_selects=0 --sum_ranges=0 --order_ranges=0 --distinct_ranges=0 --config-file=config oltp_read_only prepare
sysbench --rand-type=uniform --table_size=10000 --range_size=100 --point_selects=0 --sum_ranges=0 --order_ranges=0 --distinct_ranges=0 --config-file=config oltp_read_only run
sleep 600
done

```

stable hot region schedule

### What did you see instead?

image

left is 5.0 and right is 4.0 and I found byte is drop

image

tikv is 5.0 with 52f658a 
pd is 5.0 with 9a70cb5
pd is 4.0 with 5b5bc83

lhy1024:

To close #3527

After transfer hot read leader, we do the following：
1. skip byte and key statistic because its statistics are calculated by the last leader in this store and are inaccurate
2. maintain anticount and hotdegree to avoid store threshold and hot peers are unstable.

And we filter hot peer those is in cooldown time after transfer leader.

It means the region which just transfer leader will participate in the calculation of the threshold but will not be selected for scheduling

### Check List

Tests

- Unit test
- Integration test

it's less schedule in 5.0 with enable copr cache
image
image

It is also stable with disable copr cache(I reduce threshold of load base split to 500 at 17:40)
image
image

Related changes

### Release note

- No release note


server/statistics/hot_peer_cache.go
```diff
@@ -410,7 +410,7 @@ func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, bytes, k
 	newItem.rollingKeyRate = oldItem.rollingKeyRate
 
 	if newItem.justTransferLeader {
-		newItem.HotDegree = oldItem.HotDegree
+		newItem.HotDegree = 0
```
rleungx, line 413: Can you add more explanation about why we set it to 0? I'm not sure if it is reasonable.

lhy1024: It will prevent the region that has just been transferred leader from being dispatched again, and because of the anti count, this region can still participate in the threshold calculation

server/schedulers/hot_region.go
```diff
@@ -617,7 +617,9 @@ func (bs *balanceSolver) filterHotPeers() []*statistics.HotPeerStat {
 
 	// filter pending region
 	appendItem := func(items []*statistics.HotPeerStat, item *statistics.HotPeerStat) []*statistics.HotPeerStat {
-		if _, ok := bs.sche.regionPendings[item.ID()]; !ok && !item.IsJustTransferLeader() {
+		minHotDegree := bs.cluster.GetOpts().GetHotRegionCacheHitsThreshold()
+		if _, ok := bs.sche.regionPendings[item.ID()]; !ok && !item.IsNeedCoolDownTransferLeader(minHotDegree) {
```
rleungx, line 621: Will it affect the move peer operation?

server/statistics/hot_peer.go
```diff
@@ -132,12 +133,13 @@ func (stat *HotPeerStat) Log(str string, level func(msg string, fields ...zap.Fi
 		zap.Bool("just-transfer-leader", stat.justTransferLeader),
 		zap.Bool("is-leader", stat.isLeader),
 		zap.Bool("need-delete", stat.IsNeedDelete()),
-		zap.String("type", stat.Kind.String()))
+		zap.String("type", stat.Kind.String()),
+		zap.Time("lastTransferLeaderTime", stat.lastTransferLeaderTime))
```
rleungx, line 137: The format need to change.

server/schedulers/hot_region.go
lhy1024: Neither the transfer leader operator nor the move peer operator will be generated within three minutes of the transfer leader

rleungx (approved):

nolouch (approved):
lgtm

Issue #3517, opened 2021-03-24
inconsistent behavior on `member leader resign` and `member leader transfer` commands


mahjonp:
### What did you do?

```bash
[root@3f0e8cd2ec8f ~]# tiup ctl pd -u http://10.0.2.64:2379 member leader show
Starting component `ctl`: /root/.tiup/components/ctl/v5.0.0-rc/ctl pd -u http://10.0.2.64:2379 member leader show
{
  "name": "pd-10.0.2.64-2379",
  "member_id": 6601946604331627876,
  "peer_urls": [
    "http://10.0.2.64:2380"
  ],
  "client_urls": [
    "http://10.0.2.64:2379"
  ]
}
[root@3f0e8cd2ec8f ~]# tiup ctl pd -u http://10.0.2.64:2379 member leader resign
Starting component `ctl`: /root/.tiup/components/ctl/v5.0.0-rc/ctl pd -u http://10.0.2.64:2379 member leader resign
Failed to resign: [500] "no valid pd to transfer etcd leader"
[root@3f0e8cd2ec8f ~]# tiup ctl pd -u http://10.0.2.64:2379 member leader transfer pd-10.0.2.64-2379
Starting component `ctl`: /root/.tiup/components/ctl/v5.0.0-rc/ctl pd -u http://10.0.2.64:2379 member leader transfer pd-10.0.2.64-2379
Success!
```

### What did you expect to see?

Expect the same behavior

### What did you see instead?

Resign fails but leader transfer succeeds.

### What version of PD are you using (`pd-server -V`)?

```
Release Version: v4.0.0-rc.2-483-g48310e03
Edition: Community
Git Commit Hash: 48310e03c91a28a0afddd48ef536b9d3b1ef0f04
Git Branch: pr/3495
UTC Build Time:  2021-03-21 07:48:19
```

PR #3560, opened 2021-04-01
api: fix resign myself error

rickif:

This PR is to solve #3517.

There's no need to resign member when only one member left in etcd cluster, therefore the `ResignEtcdLeader()` should just return in this scenario.

### Check List

Tests

- Unit test
- Integration test
Origin:  
image
Fixed:
image

### Release note

- No release note

### DIFF
--- a/server/api/member.go
+++ b/server/api/member.go
@@ -275,7 +275,7 @@ func newLeaderHandler(svr *server.Server, rd *render.Render) *leaderHandler {
 // @Tags leader
 // @Summary Get the leader PD server of the cluster.
 // @Produce json
-// @Success 200 {string} string "The transfer command is submitted."
+// @Success 200 {object} pdpb.Member
 // @Router /leader [get]
 func (h *leaderHandler) Get(w http.ResponseWriter, r *http.Request) {
 	h.rd.JSON(w, http.StatusOK, h.svr.GetLeader())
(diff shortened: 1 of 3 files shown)


rickif:
@nolouch @rleungx PTAL

disksing (approved):

JmPotato:
@nolouch Should we change the "The transfer command is submitted." to "The resign command is submitted." here?

rickif:
> @nolouch Should we change the "The transfer command is submitted." to "The resign command is submitted." here?

@JmPotato Good idea. The response has been updated.

rickif:
@Yisaer PTAL

JmPotato:
@rickif Actually, I'm wondering that why we should use the "transfer" semantic here. Though resigning the only one PD will cause the unavailability of the whole cluster, it's reasonable to resign a PD leader without transferring it.

rickif:
> @rickif Actually, I'm wondering that why we should use the "transfer" semantic here. Though resigning the only one PD will cause the unavailability of the whole cluster, it's reasonable to resign a PD leader without transferring it.

https://github.com/etcd-io/etcd/blob/57a092b45d0eae6c9e600e62513ffcd2f1f25a92/raft/raft.go#L1349

Based on the etcd impliment, resigning the only one PD would do nothing and never cause the unavailability. Maybe is it better that the "regisn" behaves the same as the "transfer" ?  Namely, let the etcd handle this scenario. @JmPotato

JmPotato (approved):

Issue #3544, opened 2021-03-28
RegionScore not consider the compression ratio for eatch tikv


nolouch:
### What did you do?

`Balance Region` should make the disk usage ratio and disk available ratio peer TiKV on the same level.

### What did you expect to see?
the available ratio almost the same. （the problem also exists in regionscorev1 stage-1）

### What did you see instead?
difference，not in the same level.

![image](https://user-images.githubusercontent.com/6428910/112747939-240f4880-8feb-11eb-9e28-1cbbf7cdcb29.png)

![image](https://user-images.githubusercontent.com/6428910/112747911-f924f480-8fea-11eb-980e-741139fc2e07.png)

![image](https://user-images.githubusercontent.com/6428910/112748073-042c5480-8fec-11eb-88c4-fd52935963aa.png)

### What version of PD are you using (`pd-server -V`)?
master, release-5.0

nolouch:
ref https://github.com/tikv/pd/pull/3570

PR #3570, opened 2021-04-07
core: update region score v2 formula

disksing:

Ref: https://github.com/tikv/pd/issues/3544
The influence weight of available size is not enough in the near out of space stage.

Update formula to increase influence weight of available size.

simulator result (`stores=1200_1200_1200&amps=1.05_1.05_1.37&f=20`)

old formula

image

new formula

image

### Check List

Tests
- no test yet

### Release note
- No release note


disksing:
Have new PR https://github.com/tikv/pd/pull/3592
closing.

PR #3580, opened 2021-04-09
server: Defer region Clone to String() method

frew:
server: Defer region Clone to String() method so that it doesn't trigger if the logging isn't actually happening

Signed-off-by: Fred Wulff <frew@cs.stanford.edu>

In running `perf` on our pd node, we see a lot of the time under the cluster lock in https://github.com/tikv/pd/blob/master/server/core/region_tree.go#L99 - the zap library defers the String() method of a Stringer until it's actually doing the logging (so in this case would never execute it since Debug logging is not enabled), but currently the heavyweight proto Clone() method is called on creation. I verified that HexRegionMeta is only used by logging calls, so the change in the actual contents shouldn't affect anything unexpected.

HexRegionMeta now does the clone and hex-ification in String() rather than constructor

### Check List

Tests

- Has not been tested - let me know if there's something that would be useful.

Side effects

- Theoretically possible it decreases performance, but I can't find anywhere where the String() method is called more than once per object.

Related changes

### Release note

- No release note

### DIFF
--- a/server/core/region.go
+++ b/server/core/region.go
@@ -1122,9 +1122,6 @@ func RegionToHexMeta(meta *metapb.Region) HexRegionMeta {
 	if meta == nil {
 		return HexRegionMeta{}
 	}
-	meta = proto.Clone(meta).(*metapb.Region)
-	meta.StartKey = HexRegionKey(meta.StartKey)
-	meta.EndKey = HexRegionKey(meta.EndKey)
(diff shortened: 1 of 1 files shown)


JmPotato (approved):

JmPotato:
Thx for your contribution!

Yisaer (approved):
LGTM

Issue #3589, opened 2021-04-14
there is huge gap between low space kv and  high space kv where  there has few space.


bufferflies:
## Bug Report

there is huge gap between diferent space. if the small store has no space, but it's score is lower than the high space store,so PD do not balance region from it to high space store, so it will offline because of no space left

![image](https://user-images.githubusercontent.com/23159587/114695315-0e6d8300-9d4e-11eb-8ebc-664375ada59b.png)
 

### What did you do?

 it will can move out some region  after it's score bigger than the bigger space store to avoid store offline 

### What did you expect to see?
I want the low space store's score will increse rapidly when it has few space,
### What did you see instead?
![image](https://user-images.githubusercontent.com/23159587/114695799-96538d00-9d4e-11eb-97f7-a98e13171c83.png)

### What version of PD are you using (`pd-server -V`)?
master

nolouch:
fixed.

PR #3592, opened 2021-04-14
core: change score calculator in few disk space to avoid offline earlier

bufferflies:
issue: #3589

core regionScoreV2 calculator, store's score will increase rapidly when it disk space less than F 
### Check List

Tests

- Integration test

Code changes

Side effects

Related changes

### Release note

- modify score calculator to satisfy isomerous stores

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -308,6 +308,7 @@ func (s *StoreInfo) regionScoreV2(delta int64, deviation int) float64 {
 	var (
 		K, M float64 = 1, 256 // Experience value to control the weight of the available influence on score
 		F    float64 = 20     // Experience value to prevent some nodes from running out of disk space prematurely.
+		B            = 1e5
 	)
 
 	var score float64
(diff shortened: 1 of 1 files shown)


server/core/store.go
```diff
@@ -319,7 +320,7 @@ func (s *StoreInfo) regionScoreV2(delta int64, deviation int) float64 {
 		score = (K + M*(math.Log(C)-math.Log(A-F+1))/(C-A+F-1)) * R
 	} else {
 		// When remaining space is less then F, the score is mainly determined by available space.
-		score = (K+M*math.Log(C)/C)*R + (F-A)*(K+M*math.Log(F)/F)
+		score = (K+M*math.Log(C)/C)*R + B*(F-A)*(K+M*math.Log(F)/F)
```
nolouch, line 323: can we make F configuration and set to a more suitable default value?

rleungx: What is the relationship between `F` and `low-space-ratio`? BTW, I think it is a little bit confusing and misleading.

bufferflies: low-space-ratio: avoid other region move in 
F:  it will increase score after it's free disk less than F, so it will move out region to other 
I agree the two args misleading

bufferflies: I use max(20,capacity*(1-lowSpaceRatio)) to set F, so it will eliminate F and  low-space-ratio meaning

nolouch (approved):
lgtm

HunDunDM (approved):

PR #3616, opened 2021-04-19
tso: fix the bugs of TSO fallback and Local TSO physical time advancing too fast

JmPotato:

* Fix the bug that causes TSO fallback.
* Fix the problem of Local TSO physical time advancing too fast.

`typeutil.SubTimeByWallClock` will use the nanosecond to compare the physical time, though the smallest unit of TSO is millisecond, there is a possibility that on the millisecond level it's the same but on nanosecond level is different when updating the physical time of TSO while restoring the logical time, which will be a fallback for TSO (physical time remains unchanged, logical time goes to 0).

If `enable-local-tso` is false, this bug won't happen because we will call `setTSOPhysical` every `50ms` at least, the interval is big enough to avoid this problem. However, in Local TSO, the TSO may be written frequently by the Global TSO synchronization, which makes this bug more likely to occur.

* Not add `updateTimestampGuard` before resetting anymore.
* Always set `t.tsoMux.tso.physical` and `t.tsoMux.tso.logical` at the same time.
* No longer using the pointer to store a `tsoObject` atomically with an `RWMutex` Lock.
* Only update physical time while the difference is big enough, in this case, it's `updateTimestampGuard`, which is `1ms` by default.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/tso/global_allocator.go
+++ b/server/tso/global_allocator.go
@@ -76,6 +76,7 @@ func NewGlobalTSOAllocator(
 			updatePhysicalInterval: am.updatePhysicalInterval,
 			maxResetTSGap:          am.maxResetTSGap,
 			dcLocation:             GlobalDCLocation,
+			tsoMux:                 &tsoObject{},
 		},
 	}
 	return gta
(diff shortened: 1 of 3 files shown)


Yisaer (approved):

nolouch (approved):
LGTM.
BTW, There are many places that are changing the value of `tso` in memory. If we can maintain tso obeject in on place, may avoid this problem and more clearer.

PR #3591, opened 2021-04-14
statistics: use `[]float64` to represent hot peer loads

lhy1024:
### What problem does this PR solve?
Currently, all hot peer loads are saved in different variables and accessed using different functions.
Too many duplicated code and hard to extend.

### What is changed and how it works?
Save loads using `[]float64` and access them using enums.

### Check List

Tests
- Unit test

### Release note
- No release note

PR #3618, opened 2021-04-20
schedulers: make hot region scheduler's influence classified by rwTy

Yisaer:

Currently, the pending influence holding by hot region scheduler is classified by the `resourceType`, not `read/write type`.
This will cause the write-peer and write-leader can't see each other's influence(either do read-peer and read-leader in the future)

This request make pending influence classified by rwTy which allowed `write-peer` and `write-leader` to see each other's pending influence.

### Check List

Tests

- Unit test

### Release note

* No release not

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/metapb"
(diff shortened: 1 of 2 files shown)


disksing:
can we merge all pending influences together?

Yisaer:
duplicated with https://github.com/tikv/pd/pull/3591

Yisaer:
duplicated with https://github.com/tikv/pd/pull/3654

PR #3620, opened 2021-04-21
schedule: revise region_scatter distribution for multi groups  (#3422)

ti-srebot:
cherry-pick #3422 to release-4.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In pd repo:
git pr https://github.com/tikv/pd/pull/3620
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/pd.git pr/3620:release-4.0-9e60f4d367e3
```

---

ref https://github.com/pingcap/tidb/issues/22788

Currently region-scatter didn't work well for multi regions in different groups.

For example, there are 50 regions and 6 stores in the cluster. Each region belong to a dedicated group.

region scatter may locate them in the `store1/store2/store3`. Thus from each region in the group, the distribution is well. And for the whole cluster, the distribution is not well. 

During region scatter, we should consider whether the target store already have the most peer or leader, if it is, we should skip it.

### Check List

Tests

- Unit test

I also tested with the testcases in this document: https://docs.google.com/document/d/1cGZU2DCm0UuIueqAQP8TGqTivyavCvB_QiwIzqvP_rA/edit

All the test case results didn't change.

### Release note

* No release note


nolouch:
plz fix the conflict

PR #3628, opened 2021-04-23
tso: fix the TSO fallback caused by resetUserTimestamp

JmPotato:

Fix the TSO fallback caused by `resetUserTimestamp`.

In the previous code, we use the nanosecond to compare the next physical with TSO physical in memory, which will cause a TSO fallback in such case:

```
     second    |ms |ns  logical
now  1619094171|741|10     15
next 1619094171|741|00     10
```

While also `ignoreSmaller` flag is true, the next TSO will escape from the comparison check and make a TSO fallback eventually.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/tso/tso.go
+++ b/server/tso/tso.go
@@ -74,7 +74,7 @@ func (t *timestampOracle) setTSOPhysical(next time.Time) {
 	t.tsoMux.Lock()
 	defer t.tsoMux.Unlock()
 	// make sure the ts won't fall back
-	if typeutil.SubTimeByWallClock(next, t.tsoMux.physical).Milliseconds() >= 0 {
+	if typeutil.SubTSOPhysicalByWallClock(next, t.tsoMux.physical) >= 0 {
 		t.tsoMux.physical = next
 		t.tsoMux.logical = 0
 	}
@@ -176,7 +176,7 @@ func (t *timestampOracle) SyncTimestamp(leadership *election.Leadership) error {
 	})
 	// If the current system time minus the saved etcd timestamp is less than `updateTimestampGuard`,
 	// the timestamp allocation will start from the saved etcd timestamp temporarily.
-	if typeutil.SubTimeByWallClock(next, last) < updateTimestampGuard {
+	if typeutil.SubRealTimeByWallClock(next, last) < updateTimestampGuard {
 		log.Error("system time may be incorrect", zap.Time("last", last), zap.Time("next", next), errs.ZapError(errs.ErrIncorrectSystemTime))
 		next = last.Add(updateTimestampGuard)
 	}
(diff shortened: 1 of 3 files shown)


server/tso/tso.go
```diff
@@ -203,46 +206,51 @@ func (t *timestampOracle) isInitialized() bool {
...
-	if typeutil.SubTimeByWallClock(nextPhysical, t.tsoMux.physical) == 0 && int64(nextLogical) <= t.tsoMux.logical {
-		tsoCounter.WithLabelValues("err_reset_small_counter", t.dcLocation).Inc()
-		if !ignoreSmaller {
-			err = errs.ErrResetUserTimestamp.FastGenByArgs("the specified counter is smaller than now")
-		}
-	}
+	var (
+		nextPhysical, nextLogical = tsoutil.ParseTS(tso)
+		logicalDifference         = int64(nextLogical) - t.tsoMux.logical
+		physicalDifference        = typeutil.SubTimeByWallClock(nextPhysical, t.tsoMux.physical)
```
nolouch, line 221: How about make the accuracy of  `t.tsoMux.physical` always at the millisecond level?

JmPotato: If we use `int64` as the `t.tsoMux.physical` here will cause too many code changes because `time.Duration` and `time.Time` are massively used in the TSO code. Using `time.Time` as `t.tsoMux.physical` is more reasonable. However, the previous code is a bit redundant. I change a more clear way. PTAL again!

server/tso/tso.go
```diff
@@ -71,7 +74,7 @@ func (t *timestampOracle) setTSOPhysical(next time.Time) {
 	t.tsoMux.Lock()
 	defer t.tsoMux.Unlock()
 	// make sure the ts won't fall back
-	if typeutil.SubTimeByWallClock(next, t.tsoMux.physical) >= updateTimestampGuard {
+	if typeutil.SubTimeByWallClock(next, t.tsoMux.physical).Milliseconds() >= 0 {
```
rleungx, line 77: How about changing the implementation of `SubTimeByWallClock`?

JmPotato: I think we should keep `SubTimeByWallClock` to use nanosecond precision and leave it up to the caller to decide exactly what precision to use. Although we need to use millisecond precision for TSO comparisons, we still need nanoseconds for physical system time comparison in the `UpdateTimestamp` and `SyncTimestamp` functions.

server/tso/tso.go
nolouch: I have a consider about the new change, If I want to change `updateTimestampGuard` to `2ms`, I cannot just adjust the value.

JmPotato: I think `updateTimestampGuard` is a different term from the TSO's minimal unit. `updateTimestampGuard` should only be used when we compare the real-world time, such as making the etcd persistence determination. The TSO physical time comparison should always be kept in the unit of 1 millisecond, and I haven't thought of any reason to make the unit bigger yet.

JmPotato:
I distinguish between two different time calculation functions and add more comments to make the use of them more clear. PTAL. @nolouch @rleungx

server/tso/tso.go
nolouch: There is a scenario that maybe we can consider reducing the accuracy of the physical time when the logical bits are not enough. Now we update `physicalTS`  with `50 ms` interval.  But use millisecond as minimal unit is ok to me.

nolouch (approved):
lgtm

rleungx (approved):

PR #3639, opened 2021-04-25
[pkg]: fix the bug that the oldest element has the highest weight 

bufferflies:

1.  fix the bug that the oldest element has the highest weight 
2. use O(1) to get wma 

1.  use 2 element to record last score and sum that all element ,so score(n)=score(n-1)-sum+n 
2.  simply test case 

### Check List

Tests

- Unit test
similator:
image

Code changes

Side effects

Related changes


pkg/movingaverage/weight_moving_average.go
```diff
@@ -38,39 +40,36 @@ func NewWMA(sizes ...int) *WMA {
...
+	}
 	w.records[w.count%w.size] = n
 	w.count++
 }
 
 // Get returns the weight average of the data set.
 func (w *WMA) Get() float64 {
 	if w.count == 0 {
-		return 0
+		return w.score / float64(w.size)
```
HunDunDM, line 56: At this time, the return result must be 0?

pkg/movingaverage/moving_average_test.go
```diff
@@ -90,19 +90,19 @@ type testCase struct {
 
 func (t *testMovingAvg) TestMovingAvg(c *C) {
 	var empty float64 = 0
-	data := []float64{120, 130, 140, 150, 145, 136, 121, 132, 145, 156, 148, 157, 175}
```
HunDunDM, line 93: Can the old test case be kept?

bufferflies: it work well,but i think it should be simple for verify by manual

pkg/movingaverage/weight_moving_average.go
bufferflies: yes， it should be 0 when the list has no element

nolouch:
cc @HunDunDM @lhy1024

pkg/movingaverage/moving_average_test.go
lhy1024: add some test with jitter? such as [1,1,1,1,5,1,1,1]?

pkg/movingaverage/weight_moving_average.go
```diff
@@ -38,39 +40,36 @@ func NewWMA(sizes ...int) *WMA {
 
 // Add adds a data point.
 func (w *WMA) Add(n float64) {
+	w.score = w.score - w.sum + n*float64(w.size)
+	if w.count < w.size {
+		w.sum = w.sum + n
+	} else {
+		w.sum = w.sum - w.records[(w.count)%w.size] + n
```
lhy1024, line 47:
```suggestion
		w.sum = w.sum - w.records[w.count%w.size] + n
```

pkg/movingaverage/weight_moving_average.go
lhy1024: w. score is zero when w. count is zero? I think return directly 0 is enough

pkg/movingaverage/weight_moving_average.go
```diff
@@ -38,39 +40,36 @@ func NewWMA(sizes ...int) *WMA {
...
 	}
-
-	sum := 0.0
-
 	if w.count < w.size {
-		for i := 0; i < int(w.count); i++ {
-			sum += w.records[i]
-		}
-		return sum / float64(w.count)
+		return w.score / float64((2.0*w.size-w.count+1)*w.count/2.0)
```
lhy1024, line 59: 2.0*w.size-w.count+1 ?

bufferflies: when w is full , the count=w.size+(w.size-w.count+1),eg 
w.size=5 , the w  has 1,1,2,3 the count=4
the count should be  5+4+3+2=14 
it should be same wiht (w.size+(w.size-w.count+1))*w.count/2=14, equal (the first element+last element)*count/2

lhy1024: I think it will be better to add some comment in code or replace it with `(w.size+(w.size-w.count+1))*w.count/2`

pkg/movingaverage/weight_moving_average.go
HunDunDM: Please fix it

HunDunDM (approved):

nolouch (approved):
lgtm

PR #3634, opened 2021-04-23
tso: only run Local TSO related daemons when EnableLocalTSO

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

### What problem does this PR solve?

Only run Local TSO related daemons when `EnableLocalTSO` is true.

### What is changed and how it works?

Check `EnableLocalTSO` flag before running these daemons.

### Check List

Tests

- Unit test
- Integration test

PR #3641, opened 2021-04-26
tso: test: refine and add a test for EnableLocalTSO

JmPotato:

#3634 makes the TSO Allocator Manager only run the Local TSO daemons when `EnableLocalTSO` is true. This PR fixes some problems the previous change has and adds a test for it.

* `am.ClusterDCLocationChecker()` should run even `EnableLocalTSO` is false to check and maintain the cluster DC info because this kind of PDs could also become PD leaders to allocate the Global TSO.
* Add a test case for this change. 6 PDs while only 3 of them enable Local TSO.

### Check List

Tests

- Unit test
- Integration test

### DIFF
--- a/server/tso/allocator_manager.go
+++ b/server/tso/allocator_manager.go
@@ -1150,3 +1150,8 @@ func (am *AllocatorManager) transferLocalAllocator(dcLocation string, serverID u
 func (am *AllocatorManager) nextLeaderKey(dcLocation string) string {
 	return path.Join(am.rootPath, dcLocation, "next-leader")
 }
+
+// EnableLocalTSO returns the value of AllocatorManager.enableLocalTSO.
+func (am *AllocatorManager) EnableLocalTSO() bool {
+	return am.enableLocalTSO
+}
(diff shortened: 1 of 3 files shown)


HunDunDM (approved):

server/tso/allocator_manager.go
```diff
@@ -533,31 +533,36 @@ func (am *AllocatorManager) campaignAllocatorLeader(
...
+			// Update the initialized TSO Allocator to advance TSO.
+			am.allocatorUpdater()
 		case <-checkerTicker.C:
-			// ClusterDCLocationChecker and PriorityChecker are time consuming and low frequent to run,
-			// we should run them concurrently to speed up the progress.
+			// Check and maintain the cluster's meta info about dc-location distribution.
 			go am.ClusterDCLocationChecker()
-			go am.PriorityChecker()
+			// We won't have any Local TSO Allocator set up in PD without enabling Local TSO.
+			if am.enableLocalTSO {
```
rleungx, line 560: EnableLocalTSO？

rleungx (approved):

PR #3654, opened 2021-04-29
schedulers: unify hot region scheduler's influence

Yisaer:

Currently, the pending influence holding by hot region scheduler is classified by the `resourceType`, not `read/write type`.
This will cause the write-peer and write-leader can't see each other's influence(either do read-peer and read-leader in the future)

1. unify hot region scheduler's influence
2. make read/write pending influence can be summarized by each hot schedule operator

### Check List

Tests

- Unit test

### Release note

* No release not

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -1074,40 +1074,19 @@ func (bs *balanceSolver) buildOperators() ([]*operator.Operator, []InfluenceSumm
 	}
 	if bs.rwTy == read {
 		infls.ReadInfluence = infl
-		infls.WriteInfluence = bs.getInfluence(bs.cur.srcStoreID, bs.cur.region.GetID(), write)
+		infls.WriteInfluence = bs.generateInfluence(bs.cur.srcStoreID, bs.cur.region.GetID(), write)
 	} else {
 		infls.WriteInfluence = infl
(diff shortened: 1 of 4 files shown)


lhy1024 (commented):
It will influence correctly between read and write. But I think unified pending influence will be better.

lhy1024 (approved):

server/schedulers/hot_region.go
```diff
@@ -1099,19 +1093,18 @@ func (h *hotScheduler) GetHotWriteStatus() *statistics.StoreHotPeersInfos {
...
+	return h.copyPendingInfluence()
 }
 
 func (h *hotScheduler) GetReadPendingInfluence() map[uint64]*Influence {
-	return h.copyPendingInfluence(readLeader)
+	return h.copyPendingInfluence()
 }
 
-func (h *hotScheduler) copyPendingInfluence(ty resourceType) map[uint64]*Influence {
+func (h *hotScheduler) copyPendingInfluence() map[uint64]*Influence {
```
rleungx, line 1103: directly use it?

rleungx (commented):
The rest LGTM.

lhy1024 (approved):

rleungx (approved):

PR #3657, opened 2021-05-07
scheduler: replace `move-hot-read-peer` with `move-hot-read-leader`

### LINKED ISSUE #3649: move-hot-read-peer may not take effect successfully
### What did you do?

* The cluster has hot read flow and generates `move-hot-read-peer` scheduling.

* A(L) B(F) C(F) -> B(F) C(F) D(L). Hot reading Leader successfully moved to the new Store-D from Store-A.

### What did you see instead?

* Due to the use of `CreateMovePeerOperator` to generate the schedule. By default, the new Leader will be B or C, not D.

```
Release Version: v5.0.1
Edition: Community
Git Commit Hash: ea1449c7c693d94661b3c60d35b8679ebfb3da34
Git Branch: heads/refs/tags/v5.0.1
UTC Build Time:  2021-04-23 05:58:55
```

It should also exist in 4.0.

lhy1024:

close #3649

1. replace `[opLen]pendingInfluence` with  simple pending influence 
2. replace `move-hot-read-peer` with `move-hot-read-leader`

### Check List

Tests

- Unit test
- Integration test


server/schedulers/hot_region.go
```diff
@@ -244,6 +244,10 @@ func (h *hotScheduler) summaryPendingInfluence() {
 // ended operator
 func (h *hotScheduler) gcRegionPendings() {
 	for regionID, op := range h.regionPendings {
+		if op == nil {
+			delete(h.regionPendings, regionID)
```
rleungx, line 248: When will it happen?

lhy1024: This was the case in the previous test case; btw, this pr #3658 3658 will only involve the refactoring of pending influence

server/schedulers/hot_region.go
```diff
@@ -1000,14 +1000,23 @@ func (bs *balanceSolver) buildOperators() ([]*operator.Operator, []Influence) {
 		srcPeer := bs.cur.region.GetStorePeer(bs.cur.srcStoreID) // checked in getRegionAndSrcPeer
 		dstPeer := &metapb.Peer{StoreId: bs.cur.dstStoreID, Role: srcPeer.Role}
 		desc := "move-hot-" + bs.rwTy.String() + "-peer"
```
rleungx, line 1002: need to change?

rleungx (approved):
Actually, I think adding a new opType is better. But I'm ok with the current implementation.

HunDunDM (approved):

PR #3673, opened 2021-05-13
test: reduce the consumption of TSO-related tests

JmPotato:

Some of the TSO-related tests are covered by [TiPocket cross-region test](https://github.com/pingcap/tipocket/tree/master/testcase/cross-region), and because they are very time-consuming, I reduce the consumption of these tests to make the whole test process could run faster.

Reduce the concurrency of TSO-related tests.

### Check List

Tests

- Unit test
- Integration test


nolouch (approved):

nolouch:
plz reslove the conflict.

PR #3687, opened 2021-05-18
statistics: fix the bug that cold peer stat won't be updated

Yisaer:

Currently, for hot read peer stat reporting, it is depended on the store heartbeat. If the peer become cold after it is marked as hotPeerStat, the store heartbeat won't take it which make the PD can't automatically update cold peer stat.

Check cold peer stat during each store heartbeat to fix this problem.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


server/statistics/hot_cache.go
```diff
@@ -17,11 +17,17 @@ import (
 	"context"
 
 	"github.com/tikv/pd/server/core"
+	"go.uber.org/atomic"
 )
 
 // Denoising is an option to calculate flow base on the real heartbeats. Should
 // only turned off by the simulator and the test.
-var Denoising = true
+var Denoising = atomic.NewBool(true)
```
rleungx, line 25: I think it's unnecessary to use an atomic variable?

lhy1024: there may be race if no atomic.

server/statistics/hot_cache.go
```diff
@@ -17,11 +17,17 @@ import (
 	"context"
 
 	"github.com/tikv/pd/server/core"
+	"go.uber.org/atomic"
```
JmPotato, line 20: Why not just using `sync/atomic` pkg here?

Yisaer: `sync/atomic` also lgtm.

server/statistics/hot_cache.go
Yisaer: we met data race before

lhy1024 (approved):

server/cluster/cluster.go
```diff
@@ -554,10 +556,11 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
 		}
 		peerInfo := core.NewPeerInfo(peer, 0, 0,
 			peerStat.GetReadBytes(), peerStat.GetReadKeys(), interval)
-		item := statistics.NewFlowItem(peerInfo, region, nil)
+		item := statistics.NewPeerInfoItem(peerInfo, region)
 		c.hotStat.CheckReadAsync(item)
 	}
-
+	coldItem := statistics.NewColdItem(storeID, regionIDs, interval)
```
lhy1024, line 562: `coldItem` is confusing, we need a more fit name.

Yisaer: updated.

server/cluster/cluster.go
```diff
@@ -554,10 +556,11 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
 		}
 		peerInfo := core.NewPeerInfo(peer, 0, 0,
 			peerStat.GetReadBytes(), peerStat.GetReadKeys(), interval)
-		item := statistics.NewFlowItem(peerInfo, region, nil)
+		item := statistics.NewPeerInfoItem(peerInfo, region)
 		c.hotStat.CheckReadAsync(item)
 	}
-
+	coldItem := statistics.NewUnReportStatsHandle(storeID, regionIDs, interval)
```
lhy1024, line 562: `coldItem` also need to be updated

server/statistics/hot_cache.go
```diff
@@ -33,19 +48,51 @@ type HotCache struct {
...
+// NewPeerInfoItem creates FlowItem for PeerInfo
+func NewPeerInfoItem(peerInfo *core.PeerInfo, regionInfo *core.RegionInfo) *FlowItem {
 	return &FlowItem{
-		peerInfo:    peerInfo,
-		regionInfo:  regionInfo,
-		expiredStat: expiredStat,
+		peerInfo:            peerInfo,
+		regionInfo:          regionInfo,
+		expiredStat:         nil,
+		unReportStatsHandle: nil,
```
lhy1024, line 85: how about `unreportStatCollect`？

rleungx (approved):

server/statistics/hot_peer_cache.go
```diff
@@ -433,33 +409,24 @@ func (f *hotPeerCache) updateHotPeerStat(newItem, oldItem *HotPeerStat, deltaLoa
 	isFull := newItem.rollingLoads[0].isFull() // The intervals of dims are the same, so it is only necessary to determine whether any of them
 	if !isFull {
 		// not update hot degree and anti count
-		newItem.HotDegree = oldItem.HotDegree
-		newItem.AntiCount = oldItem.AntiCount
+		succeedItemDegree(newItem, oldItem)
```
lhy1024, line 412: how about `inheritItem`

PR #3696, opened 2021-05-21
config: add separte configuration for hot scheudle for read and write

Yisaer:

Currently, the threshold for the hot spot hit cache for read and write is both controlled by `hot-region-cache-hits-threshold`. As the hot read peer stat is reported by store heartbeat while the hot write peer stat is reported by region stat, the threshold for read and write should be controlled by different configuration.

Add `hot-read-region-cache-hits-threshold` and `hot-write-region-cache-hits-threshold`

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


disksing:
It seems that we never need to change the config.

PR #3698, opened 2021-05-21
schedulers: add random-split-scheduler

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>
is:pr is:open 

Sometimes we want to test the stability of TiDB under a frequent region scheduling scenario, for example, test the stability of TiKV client or test the stability of TiKV's configuration changes and splitting or merging regions. PD has `shuffle-leader-scheduler`, `shuffle-region-scheduler` and `random-merge-scheduler` which helps us to create such a scenario but it lacks the `random-split-scheduler` and it's not sufficient to rely on auto-splitting in TiKV.

Add a `random-split-scheduler` that randomly picks a leader region and splits it. Its limit and configuration is the same as `shuffle-leader-scheduler` because they have a similar cost.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/server/schedulers/random_split.go
+++ b/server/schedulers/random_split.go
@@ -1,0 +1,148 @@
+// Copyright 2021 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schedulers
+
+import (
+	"github.com/pingcap/kvproto/pkg/pdpb"
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/core"
+	"github.com/tikv/pd/server/schedule"
+	"github.com/tikv/pd/server/schedule/filter"
+	"github.com/tikv/pd/server/schedule/operator"
+	"github.com/tikv/pd/server/schedule/opt"
+)
+
+const (
+	// RandomSplitName is random split scheduler name.
+	RandomSplitName = "random-split-scheduler"
+	// RandomSplitType is random split scheduler type.
+	RandomSplitType = "random-split"
+)
+
+func init() {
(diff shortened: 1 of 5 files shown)


server/schedulers/random_split.go
```diff
@@ -0,0 +1,148 @@
...
+func (s *randomSplitScheduler) GetType() string {
+	return RandomSplitType
+}
+
+func (s *randomSplitScheduler) EncodeConfig() ([]byte, error) {
+	return schedule.EncodeConfig(s.conf)
+}
+
+func (s *randomSplitScheduler) IsScheduleAllowed(cluster opt.Cluster) bool {
+	allowed := s.OpController.OperatorCount(operator.OpSplit) < cluster.GetOpts().GetLeaderScheduleLimit()
```
rleungx, line 99: Why is it limited by leader schedule limit?

server/schedulers/random_split.go
```diff
@@ -0,0 +1,148 @@
...
+		schedulerCounter.WithLabelValues(s.GetName(), "not-allowed").Inc()
+		return nil
+	}
+
+	op, err := operator.CreateSplitRegionOperator(RandomSplitType, region, operator.OpAdmin, pdpb.CheckPolicy_APPROXIMATE, nil)
+	if err != nil {
+		log.Debug("fail to create split region operator", errs.ZapError(err))
+		return nil
+	}
+	op.SetPriorityLevel(core.HighPriority)
```
rleungx, line 132: No need to set it since OpAdmin will be regarded as a high priority.

server/schedulers/random_split.go
```diff
@@ -0,0 +1,148 @@
...
+		schedulerCounter.WithLabelValues(s.GetName(), "no-region").Inc()
+		return nil
+	}
+
+	if !s.allowSplit(cluster, region) {
+		schedulerCounter.WithLabelValues(s.GetName(), "not-allowed").Inc()
+		return nil
+	}
+
+	op, err := operator.CreateSplitRegionOperator(RandomSplitType, region, operator.OpAdmin, pdpb.CheckPolicy_APPROXIMATE, nil)
```
rleungx, line 127: Perhaps we should support batch split.

server/schedulers/random_split.go
HunDunDM: Can we configure OpSplit limit in scheduler config?

server/schedulers/random_split.go
HunDunDM: Is batch split more different from actual?

JmPotato:
Any updates on this?

PR #3700, opened 2021-05-22
docs: update development doc

tisonkun:

straightforward typo fixes.

### Release note

```release-note
None
```


JmPotato:
The `go.sum` change seems unnecessary. Could you please try to remove it？

tisonkun:
> @tisonkun The `go.sum` change seems unnecessary. Could you please try to remove it？

ok make sense. But why when I execute `make` the `go.sum` file changed? Isn't it stable?

tisonkun:
@JmPotato done. PTAL.

BTW codecov is quite noisy. I don't touch any line of code but coverage decreased.

JmPotato:
> > @tisonkun The `go.sum` change seems unnecessary. Could you please try to remove it？
> 
> 
> 
> ok make sense. But why when I execute `make` the `go.sum` file changed? Isn't it stable?

I think it's caused by some static tools that `make` will use. I will take a look.

JmPotato:
> @JmPotato done. PTAL.
> 
> 
> 
> BTW codecov is quite noisy. I don't touch any line of code but coverage decreased.

Yeah, it's basically not accurate at all… Maybe we can remove it some day.

PR #3706, opened 2021-05-26
schedulers: extract balancePlan for balancers

disksing:

Fix https://github.com/tikv/pd/issues/835

Extract `balancePlan` so we don't have to pass all parameters around.

### Check List

Tests
- Unit test

### Release note
```release-note
None
```

### DIFF
--- a/server/schedulers/utils.go
+++ b/server/schedulers/utils.go
@@ -36,24 +36,65 @@ const (
 	minTolerantSizeRatio    float64 = 1.0
 )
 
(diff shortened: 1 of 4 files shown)


lhy1024 (approved):

server/schedulers/utils.go
```diff
@@ -36,61 +36,102 @@ const (
...
+
+func (p *balancePlan) GetOpInfluence(storeID uint64) int64 {
+	return p.opInfluence.GetStoreInfluence(storeID).ResourceProperty(p.kind)
+}
+
+func (p *balancePlan) SourceStoreID() uint64 {
+	return p.source.GetID()
+}
+
+func (p *balancePlan) SourceStoreLabel() string {
```
rleungx, line 68: The function is a little bit misleading since we have a store label actually but with a different meaning.

disksing: good point.

rleungx (approved):

Issue #3705, opened 2021-05-25
PD cannot add replicas when all Stores in a zone are removed


disksing:
### What did you do?
Consider a 3 replica cluster. It has 6 tikv servers located among 3 zones. We use `pd-ctl` to delete all stores in a zone.

### What did you expect to see?
As now we only have 2 available zones, PD should replace offline replicas with new peer the other zones.

### What did you see instead?
Cannot replace down peer. Because PD refuses to decrease region isolation score.

### What version of PD are you using (`pd-server -V`)?
master

PR #3708, opened 2021-05-26
checker: allow location placement get worse when fix replica

disksing:

Fix https://github.com/tikv/pd/issues/3705

Root cause of the bug:

Before 4.0, PD can first remove the down replica then add a new peer replica.
In 5.0, we restricted that down replicas can only be repaired by replacing in order to maintain a sufficient number of replicas during the process. And replacing comes with the restriction that the isolation level could not be worse than before.

- Rename `SelectStoreToReplace` to `SelectStoreToFix` and remove the use of `LocationSafeguard` filter.
- add tests.

### Check List

Tests
- Unit test

Related changes
- Need to cherry-pick to the release branch (release-5.0)

### Release note
```release-note
Fix the issue that when all tikv in a zone are offline or down, PD does not move replicas to another zone
```

### DIFF
--- a/pkg/mock/mockcluster/config.go
+++ b/pkg/mock/mockcluster/config.go
@@ -128,6 +128,11 @@ func (mc *Cluster) SetLocationLabels(v []string) {
 	mc.updateReplicationConfig(func(r *config.ReplicationConfig) { r.LocationLabels = v })
 }
 
+// SetIsolationLevel updates the IsolationLevel configuration.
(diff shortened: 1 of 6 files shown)


rleungx (approved):

JmPotato (approved):

JmPotato:
@disksing Do we need to cherry-pick this PR?

disksing:
> @disksing Do we need to cherry-pick this PR?

yes. Added the tag.

Issue #3699, opened 2021-05-21
Consider priority to make up replicas for the region with fewer copies


nolouch:
## Development Task

In some disaster recovery scenarios, we may lose major of the replica regions. During the recovery process, we should restore these regions first.

bufferflies:
ok ,I'll try it

bufferflies:
it will separate some steps:

- [ ] implement the queue that supports priority and get all 
- [ ] fix replica checker to support miss regions (it needs merge into release/4.0)
- [ ] fix rule  checker to support miss regions

rleungx:
@bufferflies Can we close it now?

PR #3709, opened 2021-05-26
clsuter: let fewer replicas region has high priority to scheduler

bufferflies:

issue: #3699

check suspect and check waiting regions should be sorted by miss peer count 
### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
let fewer replicas region has a high priority to the scheduler
```


nolouch (commented):
Is it can pick to release 4.0?

server/cluster/cluster.go
```diff
@@ -454,8 +455,21 @@ func (c *RaftCluster) AddSuspectRegions(regionIDs ...uint64) {
 // GetSuspectRegions gets all suspect regions.
 func (c *RaftCluster) GetSuspectRegions() []uint64 {
 	c.RLock()
-	defer c.RUnlock()
-	return c.suspectRegions.GetAllID()
+	regions := c.suspectRegions.GetAllID()
+	c.RUnlock()
+	sort.Slice(regions, func(i, j int) bool {
```
nolouch, line 460: If the store is down, there may many regions.

disksing:
When we put a region to the wait list, we can put some extra data or tag to explain the reason that a region needs to fix. In this way, we don't have to check again.

bufferflies:
> When we put a region to the wait list, we can put some extra data or tag to explain the reason that a region needs to fix. In this way, we don't have to check again.

the pr will be close,release 4.0 do not have waiting queue .so I prepare devide the pr into two pr .  it is a good idea for check miss peer again.

Issue #3697, opened 2021-05-21
cannot fast exit when resign leader


nolouch:
### What did you do?

When the store's number is 180. we resign pd leader  may need one minute or more to finish elect the new leader:
![image](https://user-images.githubusercontent.com/6428910/119088013-8c523780-ba3a-11eb-991f-9d1359f9835f.png)

details replay: https://gist.github.com/disksing/a1bf78007b674da7e15b52a9e25eb3a5 
### What did you expect to see?
a new leader elected in 10 secend.

### What did you see instead?
1 minute
### What version of PD are you using (`pd-server -V`)?
4.0.x, master

PR #3712, opened 2021-05-27
server: make leader can lost lease as soon as possible

nolouch:

Fix https://github.com/tikv/pd/issues/3697
When the store's number is 180. we resign pd leader  may need one minute or more to finish elect the new leader:
image

details replay: https://gist.github.com/disksing/a1bf78007b674da7e15b52a9e25eb3a5 

make leader can be lost lease as soon as possible

### Check List

Tests

- Integration test

### Release note

```release-note
Fix the issue that leader re-election be slow when there are many stores
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1212,8 +1212,11 @@ func (s *Server) campaignLeader() {
 	//   1. lease based approach is not affected by thread pause, slow runtime schedule, etc.
 	//   2. load region could be slow. Based on lease we can recover TSO service faster.
 	ctx, cancel := context.WithCancel(s.serverLoopCtx)
-	defer cancel()
-	defer s.member.ResetLeader()
+	var resetLeaderOnce sync.Once
+	defer resetLeaderOnce.Do(func() {
+		cancel()
+		s.member.ResetLeader()
+	})
 
 	// maintain the PD leadership, after this, TSO can be service.
 	go s.member.KeepLeader(ctx)
@@ -1259,12 +1262,12 @@ func (s *Server) campaignLeader() {
 	s.member.EnableLeader()
 	// Check the cluster dc-location after the PD leader is elected.
 	go s.tsoAllocatorManager.ClusterDCLocationChecker()
-	defer func() {
+	defer resetLeaderOnce.Do(func() {
 		// as soon as cancel the leadership keepalive, then other member have chance
 		// to be new leader.
 		cancel()
 		s.member.ResetLeader()
-	}()
+	})
 
 	CheckPDVersion(s.persistOptions)
 	log.Info("PD cluster leader is ready to serve", zap.String("pd-leader-name", s.Name()))
--- a/server/server.go
+++ b/server/server.go
@@ -646,6 +646,7 @@ func (s *Server) createRaftCluster() error {
 }
 
 func (s *Server) stopRaftCluster() {
+	failpoint.Inject("raftclusterIsBusy", func() {})
 	s.cluster.Stop()
 }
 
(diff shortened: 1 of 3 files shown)


tests/server/member/member_test.go
```diff
@@ -238,12 +239,31 @@ func (s *memberTestSuite) TestLeaderResign(c *C) {
...
+	c.Log("leader2:", leader2)
+	c.Assert(leader2, Not(Equals), leader1)
+}
+
 func (s *memberTestSuite) waitLeaderChange(c *C, cluster *tests.TestCluster, old string) string {
 	var leader string
 	testutil.WaitUntil(c, func(c *C) bool {
 		leader = cluster.GetLeader()
 		if leader == old || leader == "" {
-			time.Sleep(time.Second)
```
lhy1024, line 246: What will be the side effects of removing it?

nolouch: Actually `testutil.WaitUntil` will sleep in the retry with `100 * time. Millisecond` interval, It's already enough.

server/server.go
```diff
@@ -1213,6 +1214,7 @@ func (s *Server) campaignLeader() {
 	ctx, cancel := context.WithCancel(s.serverLoopCtx)
 	defer cancel()
 	defer s.member.ResetLeader()
```
JmPotato, line 1216: These two lines should be removed now.

server/server.go
```diff
@@ -1256,6 +1258,11 @@ func (s *Server) campaignLeader() {
 		return
 	}
 	s.member.EnableLeader()
+	defer func() {
+		s.member.ResetLeader()
+		// as soon as cancel the keepalive.
+		cancel()
+	}()
```
JmPotato, line 1265: I think the TSO resetting better should be moved to here too.

server/server.go
rleungx: I think we still need to cancel the context when there is a problem before we reach the below defer function.

nolouch: yes, some branches will return early.

server/server.go
nolouch: Once leadership is reset and tso will stop service.

nolouch:
PTAL @JmPotato  @rleungx

server/server.go
rleungx: But it will be reset and canceled twice.

tests/server/member/member_test.go
```diff
@@ -238,12 +239,31 @@ func (s *memberTestSuite) TestLeaderResign(c *C) {
...
+
+	err = cluster.RunInitialServers()
+	c.Assert(err, IsNil)
+
+	leader1 := cluster.WaitLeader()
+	addr1 := cluster.GetServer(leader1).GetConfig().ClientUrls
+
+	err = failpoint.Enable("github.com/tikv/pd/server/raftclusterIsBusy", `pause`)
+	defer failpoint.Disable("github.com/tikv/pd/server/raftclusterIsBusy")
+	c.Assert(err, IsNil)
```
JmPotato, line 255: `defer` should come after the `err` check.

server/server.go
```diff
@@ -1255,7 +1255,16 @@ func (s *Server) campaignLeader() {
 		log.Error("failed to sync id from etcd", errs.ZapError(err))
 		return
 	}
+	// EnableLeader to accept the remaining service, such as GetStore, GetRegion.
 	s.member.EnableLeader()
+	// Check the cluster dc-location after the PD leader is elected
```
JmPotato, line 1260:
```suggestion
	// Check the cluster dc-location after the PD leader is elected.
```

JmPotato (approved):
Rest LGTM.

server/server.go
nolouch: yes, it's valid.

rleungx (approved):

PR #3728, opened 2021-06-01
server: make region statistic loss some precision

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve?

In the current situation, a change in traffic will trigger the update of `save cache` in the region heartbeat, which will cause the heartbeat to become more stressful. In fact, this part of the statistical data can work normally as long as it is within a certain accuracy.

### Check List

- Unit test
- Manual test (add detailed scripts or steps below)
- No code

Code changes

- Has configuration change
- Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Possible performance regression
- Increased code complexity
- Breaking backward compatibility

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tidb-ansible`](https://github.com/pingcap/tidb-ansible):
- Need to cherry-pick to the release branch

### Release note

```release-note
add `flow-loss-precision` to reduce the statistics update overhead
```

PR #3740, opened 2021-06-04
config: let the changes of trace-region-flow take effect in time

HunDunDM:

* After the `trace-region-flow` is changed, it needs to be restarted or transfer PD leader to take effect.

* Reload the config regularly.

### Check List

Tests

### Release note

```release-note
None
```


HunDunDM:
Replaced by #3728

Issue #3744, opened 2021-06-04
Performance issues of balance-region-scheduler


disksing:
## Development Task

[tests](https://gist.github.com/disksing/a1bf78007b674da7e15b52a9e25eb3a5) show that when there are too many stores in a cluster, `balance-region-scheduler` could have performance issues.

- [ ] Improve ruleChecker (waiting regions too long)
- [ ] Improve fitRegion
- [x] Improve `GetStoreRegionCount`
- [ ] Improve the big loop of `BalanceRegionScheduler`

PR #3745, opened 2021-06-04
schedulers: avoid recalculate tolerantRatio

disksing:

Improve `GetStoreRegionCount`
Part of https://github.com/tikv/pd/issues/3744

Calculate `tolerantSizeRatio` when the `balancePlan` is created and reuse it later.

### Check List

Tests
- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedulers/utils.go
+++ b/server/schedulers/utils.go
@@ -37,9 +37,10 @@ const (
 )
 
 type balancePlan struct {
-	kind        core.ScheduleKind
-	cluster     opt.Cluster
-	opInfluence operator.OpInfluence
+	kind              core.ScheduleKind
+	cluster           opt.Cluster
+	opInfluence       operator.OpInfluence
+	tolerantSizeRatio float64
 
 	source *core.StoreInfo
 	target *core.StoreInfo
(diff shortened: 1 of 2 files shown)


server/schedulers/utils.go
```diff
@@ -119,24 +121,24 @@ func (p *balancePlan) shouldBalance(scheduleName string) bool {
...
+		return int64(p.tolerantSizeRatio)
 	}
-
 	regionSize := p.region.GetApproximateSize()
 	if regionSize < p.cluster.GetAverageRegionSize() {
 		regionSize = p.cluster.GetAverageRegionSize()
 	}
-	regionSize = int64(float64(regionSize) * adjustTolerantRatio(p.cluster))
-	return regionSize
+	return int64(float64(regionSize) * p.tolerantSizeRatio)
```
lhy1024, line 130: can we move them to `adjustTolerantRatio`？

disksing: The `regionSize` is related to the region which is selected to balance so it cannot be calculated ahead.

rleungx (approved):

nolouch (approved):
lgtm

PR #3748, opened 2021-06-07
schedule: tune waiting region length

disksing:

The regions in the waring list are checked every tick. When there are many regions need to fix, it costs too much CPU.

Part of https://github.com/tikv/pd/issues/3744

Update cache size to a smaller value.

### Check List

Tests
- Unit test

### Release note
```release-note
None
```

### DIFF
--- a/server/schedule/checker_controller.go
+++ b/server/schedule/checker_controller.go
@@ -26,7 +26,7 @@ import (
 )
 
 // DefaultCacheSize is the default length of waiting list.
-const DefaultCacheSize = 1000
+const DefaultCacheSize = 128
 
 // CheckerController is used to manage all checkers.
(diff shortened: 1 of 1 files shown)


rleungx:
IMO, if the cache size is small, it will be less useful when there are a lot of regions.

disksing:
> IMO, if the cache size is small, it will be less useful when there are a lot of regions.

The problem is that for every tick we only scan 128 regions. This means that when the 1000 length cache is filled, we will spend 90% of our CPU time repeatedly checking the region in the cache and only 10% of our CPU time on the scan. This doesn't seem quite suitable anymore.

lhy1024 (approved):

PR #3754, opened 2021-06-09
ci: make github action effective for release-5.1

HunDunDM:

ci: make github action effective for `release-5.1`

### Check List

Tests

- No code

Code changes

Side effects

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/.github/workflows/check-dashboard.yaml
+++ b/.github/workflows/check-dashboard.yaml
@@ -3,16 +3,14 @@ on:
     branches:
       - master
       - release-4.0
-      - release-5.0
-      - release-5.1
+      - release-5.*
     paths:
       - "go.mod"
   pull_request:
     branches:
       - master
       - release-4.0
-      - release-5.0
-      - release-5.1
+      - release-5.*
     paths:
       - "go.mod"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check-dashboard.yaml
```diff
@@ -4,13 +4,15 @@ on:
...
       - release-5.0
+      - release-5.1
     paths:
       - "go.mod"
   pull_request:
     branches:
       - master
       - release-4.0
       - release-5.0
+      - release-5.1
```
disksing, line 15: how about `release-5.*`

disksing: or maybe we can use `release-*` and ignore `release-[123]*`

disksing: examples: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags

HunDunDM: They do not run in `release-3.1` or lower.

disksing (approved):

Issue #3753, opened 2021-06-08
PD panic due to nil pointer dereference


rleungx:
```
[2021-06-08T07:36:37.659Z] [CDC:ErrGetAllStoresFailed]rpc error: code = DeadlineExceeded desc = context deadline exceeded
[2021-06-08T07:36:37.659Z] [negroni] PANIC: runtime error: invalid memory address or nil pointer dereference
[2021-06-08T07:36:37.659Z] goroutine 42383 [running]:
[2021-06-08T07:36:37.659Z] github.com/urfave/negroni.(*Recovery).ServeHTTP.func1(0x7f5bd408ffa0, 0xc00387ddd0, 0xc000534370, 0xc00184da00)
[2021-06-08T07:36:37.659Z] 	/nfs/cache/mod/github.com/urfave/negroni@v0.3.0/recovery.go:159 +0xd2
[2021-06-08T07:36:37.659Z] panic(0x239e8a0, 0x3ce9210)
[2021-06-08T07:36:37.659Z] 	/usr/local/go/src/runtime/panic.go:965 +0x1b9
[2021-06-08T07:36:37.659Z] github.com/tikv/pd/server/cluster.(*RaftCluster).GetMetaStores(...)
[2021-06-08T07:36:37.659Z] 	/home/jenkins/agent/workspace/build_pd_multi_branch_master/go/src/github.com/pingcap/pd/server/cluster/cluster.go:902
[2021-06-08T07:36:37.659Z] github.com/tikv/pd/server/api.(*storesHandler).ServeHTTP(0xc001af6e30, 0x7f5bd408ffa0, 0xc00387ddd8, 0xc0028f8000)
[2021-06-08T07:36:37.659Z] 	/home/jenkins/agent/workspace/build_pd_multi_branch_master/go/src/github.com/pingcap/pd/server/api/store.go:620 +0x50
[2021-06-08T07:36:37.659Z] github.com/tikv/pd/server/api.clusterMiddleware.Middleware.func1(0x7f5bd408ffa0, 0xc00387ddd8, 0xc0028f8000)
[2021-06-08T07:36:37.659Z] 	/home/jenkins/agent/workspace/build_pd_multi_branch_master/go/src/github.com/pingcap/pd/server/api/middleware.go:43 +0x7e
[...]

PR #3756, opened 2021-06-09
api: save cluster in request context

disksing:

ref https://github.com/tikv/pd/issues/3753
We use middleware to verify the cluster is created. But the `GetRaftCluster` can return `nil` after the check. For example, when the server is closing.

It can potentially cause server panic.

In middleware, save the `RaftCluster` in request context. So that later we can always take a `RaftCluster` and use.

### Check List

Tests
- Unit test

### Release note
```release-note
None
```


nolouch (approved):
lgtm

JmPotato (approved):

PR #3770, opened 2021-06-15
*: improve the some metrics

rleungx:

We have some monitors like the balance region scheduler which records the scheduling event. But it's hard to find something when we investigate the problem.

This PR records the total number of events so that we can know the relationship between each event.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedulers/balance_region.go
+++ b/server/schedulers/balance_region.go
@@ -150,6 +150,7 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 	})
 	for _, plan.source = range stores {
 		for i := 0; i < balanceRegionRetryLimit; i++ {
+			schedulerCounter.WithLabelValues(s.GetName(), "total").Inc()
 			// Priority pick the region that has a pending peer.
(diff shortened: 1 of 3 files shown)


server/schedulers/balance_region.go
```diff
@@ -150,6 +150,7 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 	})
 	for _, plan.source = range stores {
 		for i := 0; i < balanceRegionRetryLimit; i++ {
+			schedulerCounter.WithLabelValues(s.GetName(), "total").Inc()
```
lhy1024, line 153: total? balance_total?

rleungx: IMO, total is enough ^_^

lhy1024 (approved):

nolouch (approved):
lgtm

nolouch:
Do we need to cherry-pick?

rleungx:
> Do we need to cherry-pick?

Since it's only a small improvement, I think we don't need to cherry-pick it.

PR #3783, opened 2021-06-16
server: revert #3756, recover #3313, fix panic

xhebox:
Signed-off-by: xhe <xw897002528@gmail.com>

It is known that my #3313 has raised some panics in CI. After some investigations, I think I found the cause. This PR reverted the changes back and fixed the potential race cases.

**What is the problem**:

According to [CI log](https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/4494/pipeline/#step-301-log-213), it is clear that API requests have passed the middleware check. 

But it may be nil if server is about to close or is already stopped, [link](https://github.com/tikv/pd/blob/3a7f0d6dec16fc342b2f21349419e5d5f500d0dc/server/server.go#L1040-L1042). It is not ensured that server is not closed or stopping while serving RPC services.

Maybe it is running while executing middleware checks, but later it is stopped/closed when the real service is serving. But the real service assumes that it is non-nil, because middleware checks have passed.

In short, PD can serve RPC while closing servers. That is the root cause.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix panic if request PD RPC server fast enough
```

### DIFF
--- a/server/api/admin.go
+++ b/server/api/admin.go
@@ -44,7 +44,7 @@ func newAdminHandler(svr *server.Server, rd *render.Render) *adminHandler {
 // @Failure 400 {string} string "The input is invalid."
 // @Router /admin/cache/region/{id} [delete]
 func (h *adminHandler) HandleDropCacheRegion(w http.ResponseWriter, r *http.Request) {
-	rc := getCluster(r)
+	rc := h.svr.GetRaftCluster()
 	vars := mux.Vars(r)
 	regionIDStr := vars["id"]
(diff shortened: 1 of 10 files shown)


disksing:
@xhebox I don't know if it can avoid panic or not. But it is a bad idea for the stability of one module to depend on the execution of statements in a certain order within another module.

xhebox:
@disksing It is also a problem if PD can serve RPC while closing servers. RaftCluster you got is essentially closed/deconstructed.

This PR can't fix the problem, sadly.

PR #3787, opened 2021-06-17
schedulers: rollback the scheduler when failing to persist

### LINKED ISSUE #2572: evict-leader-scheduler may fail to persist configuration
It is a hypothetically speculated bug: when remove a store id from `evict-leader-scheduler`, PD first update the in-memory configuration and then persist it. If it fails to persist to etcd, the operation looks successful, but later when PD leader changes, the removed store id will come up again.

rleungx:

Try to fix #2572.

This PR corrects the status code and rollback the scheduler when failing to persist.

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
Fix the issue that the scheduler may appear again even if we have already executed the delete operation
```


server/cluster/coordinator.go
```diff
@@ -613,7 +613,6 @@ func (c *coordinator) removeScheduler(name string) error {
 
 	s.Stop()
```
nolouch, line 614: also need to move this line?

disksing (approved):

lhy1024 (approved):

PR #3805, opened 2021-06-21
checker: avoid to merge region with high qps

lhy1024:

avoid to merge region with high qps

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
checker: avoid to merge region with high qps
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -634,6 +634,8 @@ type ScheduleConfig struct {
 	MaxMergeRegionKeys uint64 `toml:"max-merge-region-keys" json:"max-merge-region-keys"`
 	// SplitMergeInterval is the minimum interval time to permit merge after split.
 	SplitMergeInterval typeutil.Duration `toml:"split-merge-interval" json:"split-merge-interval"`
+	// SplitQPSThreshold is the qps threshold for split region.
+	SplitQPSThreshold uint64 `toml:"split-qps-threshold" json:"split-qps-threshold"`
 	// EnableOneWayMerge is the option to enable one way merge. This means a Region can only be merged into the next region of it.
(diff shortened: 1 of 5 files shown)


server/config/config.go
```diff
@@ -634,6 +634,8 @@ type ScheduleConfig struct {
 	MaxMergeRegionKeys uint64 `toml:"max-merge-region-keys" json:"max-merge-region-keys"`
 	// SplitMergeInterval is the minimum interval time to permit merge after split.
 	SplitMergeInterval typeutil.Duration `toml:"split-merge-interval" json:"split-merge-interval"`
+	// SplitQPSThreshold is the qps threshold for split region.
+	SplitQPSThreshold uint64 `toml:"split-qps-threshold" json:"split-qps-threshold"`
```
nolouch, line 638: can we know this from TiKV? I think we can use `ConfigManager`.

lhy1024: How to use `ConfigManager`？And I think we can also get `StoreHeartbeatInterval` and `RegionHeartbeatInterval` from it.

disksing (approved):

disksing:
I think we need some document update.

server/config/config.go
rleungx: What if this threshold is not the same between PD and TiKV?

lhy1024: When the pd threshold is higher than the tikv threshold, some regions may not be merged as expected, and conversely, some regions that were just split by qps may be merged again, so we expect them to be the same, or not much different

lhy1024:
Considering that qps will be used as the dimension for reading hotspot scheduling and that the merge checker will skip regions already identified as hotspots, this pr is unnecessary and I will close it.

PR #3813, opened 2021-06-25
pkg: implement preempt priority queue use btree 

bufferflies:
# What problem does this PR solve?
fix #3709
implement preempt priority queue use btree ,and every entry  has retry and last time property , the next execute time= last+(10*patro_interval)*retry. the bigger the retry is , the longer the entry wait

### Check List
Tests
  - Unit test

Code changes

Side effects

Related changes

### Release note
```release-note
None
```


bufferflies:
PTAL @HunDunDM

pkg/cache/priority_queue.go
```diff
@@ -0,0 +1,120 @@
...
+func (pq *PriorityQueue) Len() int {
+	return pq.btree.Len()
+}
+
+// Entry a record in PriorityQueue
+type Entry struct {
+	Priority int
+	Value    interface{}
+	Last     time.Time
+	retry    int
```
HunDunDM, line 112: What are they used for?

pkg/cache/priority_queue.go
```diff
@@ -0,0 +1,120 @@
...
+		return true
+	})
+	return rs
+}
+
+// Remove remove value from queue
+func (pq *PriorityQueue) Remove(value interface{}) {
+	if v, ok := pq.items[value]; ok {
+		if i := pq.btree.Delete(v); i == nil {
+			fmt.Printf("error:%v \n", pq.items)
```
HunDunDM, line 96: use `log.Error`

pkg/cache/priority_queue.go
bufferflies: it will be helper for region retry time to avoid some region retry frequency. pr is https://github.com/tikv/pd/pull/3814

nolouch: I think the retry logic should not maintain in the basic library, that's strange.

bufferflies: ok, it will remove in this pr.

pkg/cache/cache_test.go
```diff
@@ -297,3 +297,61 @@ func (s *testRegionCacheSuite) TestTwoQueueCache(c *C) {
...
+	c.Assert(pq.Put(1, testData[1]), IsFalse)
+
+	// it will have priority-value pair as 1-1 2-2 3-3
+	pq = NewPriorityQueue(3)
+	c.Assert(pq.Put(1, testData[1]), IsTrue)
+	c.Assert(pq.Put(2, testData[2]), IsTrue)
+	c.Assert(pq.Put(3, testData[4]), IsTrue)
+	c.Assert(pq.Put(5, testData[4]), IsTrue)
+	c.Assert(pq.Put(6, testData[5]), IsFalse)
+	c.Assert(pq.Put(3, testData[3]), IsTrue)
```
nolouch, line 321: Does the same priority lead to frequent swaps in and out? is it as expect?

nolouch (commented):
rest LGTM

pkg/cache/cache_test.go
bufferflies: it will swap only if the lowest priority less than the input's priority. As this case, after the 3->4 will become  5->4, it will not be swapped by 6->5 because the 6>5, but it will be swapped by the 3->3 because 3<5.

nolouch: I mean same priority, actually, if you use miss peer count as the priority, there are many regions have the same priority.

pkg/cache/priority_queue.go
```diff
@@ -0,0 +1,128 @@
...
+		return false
+	}
+	id := item.ID()
+	entry, ok := pq.items[id]
+	if !ok {
+		entry = &Entry{Priority: priority, Value: value}
+		if pq.Len() >= pq.capacity {
+			min := pq.btree.Min()
+			// avoid to capacity equal 0
+			if min == nil || entry.Less(min) {
```
HunDunDM, line 44:
```suggestion
			if min == nil || !min.Less(entry) {
```
@nolouch

pkg/cache/cache_test.go
bufferflies: yes, I will change the input condition(!min.Less(entry)) and add unit test case

pkg/cache/priority_queue.go
```diff
@@ -0,0 +1,114 @@
+package cache
```
nolouch, line 1: add license.

pkg/cache/priority_queue.go
```diff
@@ -0,0 +1,127 @@
+// Copyright 2017 TiKV Project Authors.
```
nolouch, line 1:
```suggestion
// Copyright 2021 TiKV Project Authors.
```

HunDunDM (approved):

nolouch (approved):

HunDunDM:
@bufferflies Please fix the sign-off problem

PR #3814, opened 2021-06-25
[server]: region that has more miss peers has high priority to schedule 

bufferflies:
Issue: #3699
it only fix make up replicas not consider the offline or down peer

first It will check region miss  counts if it absense region peer. if it's miss peers more than majority replicas, I will put it to priority queue,so it will run first .
second it will remove from priority queue if it's miss peers less than majority replicas. 
### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
Please add a release note.

```


bufferflies:
PTAL @HunDunDM

server/cluster/coordinator.go
```diff
@@ -155,6 +156,35 @@ func (c *coordinator) patrolRegions() {
 	}
 }
 
+// checkPriorityRegions
```
disksing, line 159: please complete comment or simply remove it.

nolouch (commented):
I recommend you let the logic `check the priority and push to the queue` in a separate place, the logic is too confusing in some checkers. and you also cannot cover the sense that the placement rule is enabled. ptal @HunDunDM

server/cluster/coordinator.go
bufferflies: I will complete it

bufferflies:
> I recommend you let the logic `check the priority and push to the queue` in a separate place, the logic is too confusing in some checkers. and you also cannot cover the sense that the placement rule is enabled. ptal @HunDunDM

the place rule fix will implement in other pr

nolouch:
PTAL @JmPotato @AndreMouche   please help to review this PR.

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// NewPriorityChecker creates a priority checker.
+func NewPriorityChecker(cluster opt.Cluster) *PriorityChecker {
+	opts := cluster.GetOpts()
+	return &PriorityChecker{
+		cluster: cluster,
+		opts:    opts,
+		queue:   cache.NewPriorityQueue(defaultPriorityQueueSize),
+	}
```
JmPotato, line 42:
```suggestion
	return &PriorityChecker{
		cluster: cluster,
		opts:    cluster.GetOpts(),
		queue:   cache.NewPriorityQueue(defaultPriorityQueueSize),
	}
```

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// checkRegionInPlacementRule check region in placement rule mode
+func (p *PriorityChecker) checkRegionInPlacementRule(region *core.RegionInfo) (makeupCount int) {
+	fit := p.cluster.FitRegion(region)
+	if len(fit.RuleFits) == 0 {
+		return
+	}
+	for _, rf := range fit.RuleFits {
+		makeupCount = makeupCount + rf.Rule.Count - len(rf.Peers)
```
JmPotato, line 86: Is it possible that there will be multiple duplicated rules describing the same replica count here?

bufferflies: well, it will also put into priority queue  if there are two rule relate with the same region and one rule not satisfy. then it will make up the deficiency replicas

server/schedule/checker/priority_checker.go
bufferflies: OK

JmPotato (approved):
Rest LGTM.

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,134 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checker
+
+import (
+	"time"
+
+	"github.com/tikv/pd/server/schedule/placement"
+
```
rleungx, line 20: remove the empty line

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,134 @@
...
+		p.queue.Remove(regionID)
+	}
+}
+
+// GetPriorityRegions returns all regions in priority queue that needs rerun
+func (p *PriorityChecker) GetPriorityRegions() (ids []uint64) {
+	entries := p.queue.Elems()
+	for _, e := range entries {
+		re := e.Value.(*RegionPriorityEntry)
+		if t := re.Last.Add(time.Duration(re.Retry*10) * p.opts.GetPatrolRegionInterval()); t.Before(time.Now()) {
```
rleungx, line 124: Can you add a comment to explain it?

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// ID implement PriorityQueueItem interface
+func (r RegionPriorityEntry) ID() uint64 {
+	return r.regionID
+}
+
+// NewRegionEntry construct of region priority entry
+func NewRegionEntry(regionID uint64) *RegionPriorityEntry {
+	return &RegionPriorityEntry{regionID: regionID, Last: time.Now(), Retry: 1}
```
rleungx, line 65: Why does retry equal to 1 for a new entry?

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,134 @@
...
+// Check check region's replicas, it will put into priority queue if the region lack of replicas.
+func (p *PriorityChecker) Check(region *core.RegionInfo) {
+	makeupCount := 0
+	if p.opts.IsPlacementRulesEnabled() {
+		makeupCount = p.checkRegionInPlacementRule(region)
+	} else {
+		makeupCount = p.checkRegionInReplica(region)
+	}
+	priority := 0 - makeupCount
+	p.addPriorityQueue(priority, region.GetID())
```
rleungx, line 77: For those regions which don't need to make it up, do we still need to do it?

bufferflies: this pr solve the recovery situation . it will speed the progress if some store broken and the region should make up replica after recovery by manual

server/schedule/checker/priority_checker.go
bufferflies: the description as below: 
// avoid to some priority region occupy checker, region don't need check on next check interval
// the next run time is : last_time+retry*10*patrol_region_interval

server/schedule/checker/priority_checker.go
rleungx: According to your implementation, each region, no matter if it should make up a replica, will enter addPriorityQueue, right?

server/schedule/checker/priority_checker.go
bufferflies: the region needs run time is : last_time+retry*10*patrol_region_interval, so the retry is 1,the region will rerun after 10*patrol_region_interval

server/schedule/checker/priority_checker.go
bufferflies: yes

nolouch: It's better to add it only make up the replica.

server/schedule/checker/priority_checker_test.go
```diff
@@ -0,0 +1,78 @@
+package checker
```
nolouch, line 1: copyright

server/schedule/checker/priority_checker.go
bufferflies: the addPriorityQueue function will remove the region if it don't need to make up replica, otherwise, it will add the region to priority queue.

bufferflies:
@rleungx PTAL

disksing:
I haven't synchronized this PR progress lately, unassigning myself.

server/schedule/checker/priority_checker.go
```diff
@@ -0,0 +1,136 @@
...
+// it will remove if region's priority equal 0
+// it's Attempt will increase if region's priority equal last
+func (p *PriorityChecker) addOrRemoveRegion(priority int, regionID uint64) {
+	if priority < 0 {
+		if entry := p.queue.Get(regionID); entry != nil && entry.Priority == priority {
+			e := entry.Value.(*RegionPriorityEntry)
+			e.Attempt = e.Attempt + 1
+			e.Last = time.Now()
+		}
+		entry := NewRegionEntry(regionID)
```
nolouch, line 112: It's not intuitive, you took it out, and make a new one.

nolouch (approved):
lgtm

Issue #3807, opened 2021-06-21
merge operator blocks the balance leader operations


rleungx:
### What did you do?

Run a cluster, kill one TiKV, and restart it.

### What did you expect to see?

The leader should be balanced quickly.

### What did you see instead?

The merge operator spends 10 mins to become timeout. In this period, the balance leader cannot run due to the leader limit. Besides, one merge operator will occupy two leader limits if it contains a transfer leader step.
<img width="1733" alt="Screen Shot 2021-06-21 at 8 27 29 PM" src="https://user-images.githubusercontent.com/35896542/122762103-8b2d5800-d2cf-11eb-91c6-84906b2b8083.png">
<img width="1750" alt="Screen Shot 2021-06-21 at 8 28 09 PM" src="https://user-images.githubusercontent.com/35896542/122762125-8f597580-d2cf-11eb-9cf5-87c497e855ab.png">

### What version of PD are you using (`pd-server -V`)?
v5.0.2

rleungx:
not only balance leader, may also block evict leader

Issue #3778, opened 2021-06-15
Reduce the conflict when multiple scheduling is running


nolouch:
## Background

Scenario 1: Due to the drop table operation, a large number of merge operations are continuously occurring, and the balance-leader is restricted by the leader-schedule-limit, resulting in the conflict.

Scenario 2: store limit is shared by multiple schedulers. 

## details
| |  Store Limit - Add Peer| Store Limit -Remove Peer  | leader-schedule-limit |  hot-region-schedule-limit | merge-schedule-limit| region-schedule-limit | replica-schedule-limit |
|  ----  | ----  | --- | ----  | --- |  ----  | --- |  --- |
| BalanceLeader  | × | × |  √ | × | × | × | × |
| BalanceRegion  | √ | √ |  × | × | × | √ | × |
| EvictLeader | × | × |  √ | × | × | × | × |
| HotRegion | √ |  √ |  √ |  √ | × |  √ | × |
| Label | 
| ScatterRange | √ |  √ |  √ | × | × |  √ | × |
| learnerChecker| × | × |  × | × | × | × | × |
| replicaChecker | √ |  √ |  × | × | × |  √ | √ |
| ruleChecker | √ |  √ |  × | × | × |  √ | √ |
| mergeChecker | √ |  √ |  √ | × | √ |  √ | × |

- [ ] Consider do not influence with each other, such as `hot-region`.
- [ ] Improve the shared behavior for the `store limit`, such as allow preemption with priority.

rleungx:
Also some existed problems can be found in https://github.com/tikv/pd/issues/3807

bufferflies:
Also some existed problems can be found in #3749

bufferflies:
there are two way to solved the problem that schedulers influenced each other:
one way: one operator enter to operator control should have one OpKind, the priority is blow:
         OpMerge>OpHotRegion>OpLeader>OpRegion
the other way: modify the scheduler/checker isAllowed condition, such as:
	balance region scheduler: count(OpRegion)-count(OpMerge)-count(OpHotRegion)<region scheduler limit

bufferflies:
the relation between OpKind with scheduler(checker) is blow:
![image](https://user-images.githubusercontent.com/23159587/123762216-29837400-d8f5-11eb-8283-37ace1ad36fe.png)

bufferflies:
Letremove-extras-peer not limit #3865

lhy1024:
@bufferflies Can we close it?

PR #3820, opened 2021-06-29
schedule:  one operator only occupy one limit

bufferflies:

fix: #3807 
	 #3778

operator controller's count will occupy the highest priority OpKind even if the operator has many OpKind. 

the relation between limit and scheduler is blow:
image

red mark need to be changed 
### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
Reduce the conflict due to multiple scheduler running in same time
```


server/schedule/operator/operator.go
```diff
@@ -125,6 +125,18 @@ func (o *Operator) Kind() OpKind {
 	return o.kind
 }
 
+// SchedulerKind return the highest OpKind even if the operator has many OpKind
+// fix #3778
+func (o *Operator) SchedulerKind() OpKind {
+	priority := []OpKind{OpMerge, OpRange, OpReplica, OpHotRegion, OpRegion, OpLeader}
```
HunDunDM, line 131: Is `OpRange` really needed here?

HunDunDM (commented):
`OperatorCount` can `return oc.counts[mask]` directly  now?

server/schedule/operator/operator.go
```diff
@@ -125,6 +125,18 @@ func (o *Operator) Kind() OpKind {
 	return o.kind
 }
 
+// SchedulerKind return the highest OpKind even if the operator has many OpKind
+// fix #3778
+func (o *Operator) SchedulerKind() OpKind {
```
HunDunDM, line 130: We can rearrange OpKind and use the LowBit function to get it directly.
`LowBit(x) = x&(-x)`

server/schedule/operator/operator.go
```diff
@@ -125,6 +125,16 @@ func (o *Operator) Kind() OpKind {
...
+	// LowBit ref: https://zh.wikipedia.org/wiki/%E6%A0%91%E7%8A%B6%E6%95%B0%E7%BB%84
```
HunDunDM: This is a Chinese page. You can use `https://en.wikipedia.org/wiki/Find_first_set`.

server/schedule/operator_controller.go
```diff
@@ -769,7 +769,7 @@ func (oc *OperatorController) updateCounts(operators map[uint64]*operator.Operat
 		delete(oc.counts, k)
 	}
 	for _, op := range operators {
-		oc.counts[op.Kind()]++
+		oc.counts[op.SchedulerKind()]++
```
HunDunDM, line 772: Maybe we can implement a new `OperatorCount` function without changing the existing function?

server/schedule/operator/kind.go
```diff
@@ -24,22 +24,23 @@ type OpKind uint32
...
-	// Initiated by merge checker or merge scheduler. Note that it may not include region merge.
-	OpMerge
-	// Initiated by range scheduler.
-	OpRange
+	// Include peer addition or removal. This means that this operator may take a long time.
+	OpRegion
+	// Include leader transfer.
+	OpLeader
+	// Initiated by admin.
+	OpAdmin
```
nolouch, line 43: should Admin with higher priority?

HunDunDM: LGTM, `OpAdmin` should not occupy the limit.

rleungx: It will make `op show admin` cannot be used.

HunDunDM: > It will make `op show admin` cannot be used.

No, it uses `GetOperatorsOfKind`.

rleungx: Ok

server/schedule/operator/operator.go
rleungx: Perhaps, we can find a better name.

HunDunDM (approved):

lhy1024 (commented):
Do we need additional test?

nolouch (approved):
lgtm

Issue #3569, opened 2021-04-06
Improve the merge operator in Grafana


rleungx:
### Describe your feature request related problem

<img width="925" alt="Screen Shot 2021-04-06 at 7 22 43 PM" src="https://user-images.githubusercontent.com/35896542/113703549-8ebc3480-970d-11eb-823d-a1565e575615.png">

### Describe the feature you'd like

Currently, one merge operation has two merge operators. But one is successful and the other one is canceled which may mislead the user.

lhy1024:
close by #6673

rleungx:
> close by #6673

Why?

lhy1024:
> > close by #6673
> 
> Why?

This pr makes merge operators canceled together, I think it finishes the description in pr.
If you think not, it is free to reopen it.

rleungx:
> 

The problem is that we have two operators, only one can reach the finish status, and the other one will be changed to cancel status. IMO, We only need to record one metrics in this kind of situation.

PR #3832, opened 2021-07-05
schedule: prevent merge operators occupying schedule limit

rleungx:

Another improvement for #3807.

This PR introduces a new operator kind named `OpPlaceholder`, which will not waste the merge schedule limit.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/merge_checker.go
+++ b/server/schedule/checker/merge_checker.go
@@ -153,7 +153,7 @@ func (m *MergeChecker) Check(region *core.RegionInfo) []*operator.Operator {
 	log.Debug("try to merge region",
 		logutil.ZapRedactStringer("from", core.RegionToHexMeta(region.GetMeta())),
 		logutil.ZapRedactStringer("to", core.RegionToHexMeta(target.GetMeta())))
-	ops, err := operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target, operator.OpMerge)
+	ops, err := operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target, 0)
 	if err != nil {
 		log.Warn("create merge region operator failed", errs.ZapError(err))
 		return nil
--- a/server/schedule/operator/create_operator.go
+++ b/server/schedule/operator/create_operator.go
@@ -170,7 +170,7 @@ func CreateMergeRegionOperator(desc string, cluster opt.Cluster, source *core.Re
 
 	brief := fmt.Sprintf("merge: region %v to %v", source.GetID(), target.GetID())
 	op1 := NewOperator(desc, brief, source.GetID(), source.GetRegionEpoch(), kind|OpMerge, steps...)
-	op2 := NewOperator(desc, brief, target.GetID(), target.GetRegionEpoch(), kind&OpAdmin|OpPlaceholder, MergeRegion{
+	op2 := NewOperator(desc, brief, target.GetID(), target.GetRegionEpoch(), kind|OpPlaceholder, MergeRegion{
 		FromRegion: source.GetMeta(),
 		ToRegion:   target.GetMeta(),
 		IsPassive:  true,
--- a/server/schedule/operator/create_operator.go
+++ b/server/schedule/operator/create_operator.go
@@ -170,12 +170,13 @@ func CreateMergeRegionOperator(desc string, cluster opt.Cluster, source *core.Re
 
 	brief := fmt.Sprintf("merge: region %v to %v", source.GetID(), target.GetID())
 	op1 := NewOperator(desc, brief, source.GetID(), source.GetRegionEpoch(), kind|OpMerge, steps...)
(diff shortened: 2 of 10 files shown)


server/schedule/operator/create_operator.go
```diff
@@ -162,12 +162,13 @@ func CreateMergeRegionOperator(desc string, cluster opt.Cluster, source *core.Re
 
 	brief := fmt.Sprintf("merge: region %v to %v", source.GetID(), target.GetID())
 	op1 := NewOperator(desc, brief, source.GetID(), source.GetRegionEpoch(), kind|OpMerge, steps...)
-	op2 := NewOperator(desc, brief, target.GetID(), target.GetRegionEpoch(), kind|OpMerge, MergeRegion{
+	op2 := NewOperator(desc, brief, target.GetID(), target.GetRegionEpoch(), kind&OpAdmin|OpPlaceholder, MergeRegion{
```
HunDunDM, line 165: Why `&OpAdmin`?

HunDunDM (commented):
rest LGTM

server/schedule/operator/create_operator.go
rleungx: This operator can only be `(OpAdmin, OpPlaceholder)` or `(OpPlaceholder)`

server/schedule/checker/merge_checker.go
```diff
@@ -137,7 +137,7 @@ func (m *MergeChecker) Check(region *core.RegionInfo) []*operator.Operator {
 	log.Debug("try to merge region",
 		logutil.ZapRedactStringer("from", core.RegionToHexMeta(region.GetMeta())),
 		logutil.ZapRedactStringer("to", core.RegionToHexMeta(target.GetMeta())))
-	ops, err := operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target, operator.OpMerge)
+	ops, err := operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target, 0)
```
nolouch, line 140: Can we remove the argument `kind`?

server/schedule/operator/create_operator.go
nolouch: Why not directly use `OpPlaceholder`?

server/schedule/checker/merge_checker.go
rleungx: No, we can't because the waiting operator will divide the operators into different buckets according to the priority. If we remove it or just use `OpPlaceholder`, the two operators will be put into two kinds of buckets which is not expected when we manually add a merge operator through API.

nolouch, line 140:
```suggestion
	ops, err := operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target)
```
I don‘t understand, Why kind will influence the operator priority? isn't it set by `SetPriorityLevel`?

rleungx: See https://github.com/tikv/pd/blob/f6fc8fb17af8c6495a282fa20d10e45422b9e02b/server/schedule/operator/operator.go#L61-L63.

nolouch (commented):
If you just want to make the two merge operators only consume 1 limit. how about just change the line:
https://github.com/tikv/pd/blob/671b49f4bedaa0a233c71c6af7236609e588b856/server/schedule/checker_controller.go#L93

to 
```
opController.OperatorCount(operator.OpMerge)/2 < c.opts.GetMergeScheduleLimit()
```

JmPotato:
Any updates on this?

rleungx:
> Any updates on this?

I'm going to solve https://github.com/tikv/pd/issues/3569 in this PR. FYI @nolouch

server/schedule/checker/merge_checker.go
AndreMouche: How about add a new function `NewMergeRegionOperator(cluster,source,target)`?  Just call`CreateMergeRegionOperator("merge-region",cluster,source,target,0)` in the function?

AndreMouche: Why the kind is 0? I think it should be `OpMerge`? https://github.com/tikv/pd/blob/5fdb1fb78ba8e5ca0aff383d2c8652b2554e6bdd/server/schedule/operator/kind.go#L33-L35

AndreMouche (commented):
I've found there are only three different ways to call the function 
```
CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin)
CreateMergeRegionOperator(RandomMergeType, cluster, region, target, operator.OpMerge)
 operator.CreateMergeRegionOperator("merge-region", m.cluster, region, target, 0)
````

And I do not like the last usage whose kind is 0, could we avoid the usage of 0?

rleungx:
This PR is stale, close it for now.

PR #3864, opened 2021-07-09
api: add interval to the region info

HunDunDM:

* Use the `Interval` field in Key Visualizer to filter some noise data.

* api: add `Interval` to the region info

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


nolouch:
Why need this? how about making the field `WrittenBytes` to bytes/s.

lhy1024 (commented):

JmPotato:

PR #3871, opened 2021-07-14
scheduler: support dim priority for hot-region-scheduler

Yisaer:

Hot-region-scheduler will try to balance the rate of each dim for every store while the desired dim may not be balanced after the hot-region scheduling become stable.

This request support priority for dim for hot-region-scheduler

### Check List

Tests

- Unit test

### Release note

```release-note
Support dimension priority for hot-region-scheduler
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -846,14 +846,24 @@ func (bs *balanceSolver) calcProgressiveRank() {
 	dstLd := bs.stLoadDetail[bs.cur.dstStoreID].LoadPred.max()
 	peer := bs.cur.srcPeerStat
 	rank := int64(0)
+	priority := bs.sche.conf.HotDimPriority
 	if bs.rwTy == write && bs.opTy == transferLeader {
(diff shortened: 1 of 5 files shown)


server/schedulers/hot_region.go
```diff
@@ -786,16 +886,32 @@ func (bs *balanceSolver) calcProgressiveRank() {
...
+				// If belong to the case, both byte rate and key rate will be more balanced, the best choice.
+				rank = -3
+			case byteDecRatio <= minorDecRatio && keyHot && keyDecRatio <= greatDecRatio:
+				// If belong to the case, byte rate will be not worsened, key rate will be more balanced.
+				rank = -2
+			case byteHot && byteDecRatio <= greatDecRatio:
+				// If belong to the case, byte rate will be more balanced, ignore the key rate.
+				rank = -1
+			}
+		} else if priority == ReadKeyDimPriority || priority == WriteKeyDimPriority {
```
lhy1024, line 901: Do we need to consider whether other dim will be not worsened?

PR #3875, opened 2021-07-15
statistics: remove justTransferLeader check

Yisaer:

Remove unnecessary justTransferLeader check for hot peer stat as we will support sending truely peer stat in tikv by https://github.com/tikv/tikv/pull/10530

Remove unnecessary justTransferLeader check 

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


lhy1024:
add cooldown because I found it's unstable in sysbench scan small table with copr cache enable #3528.

I think it would be more convincing to test this scenario on this branch

Yisaer:
> add cooldown because I found it's unstable in sysbench scan small table with copr cache enable #3528.
> 
> I think it would be more convincing to test this scenario on this branch

Good point, I will check it sooner

PR #3889, opened 2021-07-19
moving average: MovingAvg supports GetInstantaneous

HunDunDM:

Add `GetInstantaneous` to `MovingAvg` interface.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/movingaverage/weight_moving_average.go
+++ b/pkg/movingaverage/weight_moving_average.go
@@ -75,3 +75,11 @@ func (w *WMA) Set(n float64) {
 	w.Reset()
 	w.Add(n)
 }
+
+// GetInstantaneous returns the value just added.
+func (w *WMA) GetInstantaneous() float64 {
+	if w.count == 0 {
+		return 0
+	}
+	return w.records[(w.count-1)%w.size]
+}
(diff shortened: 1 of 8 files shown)


lhy1024 (approved):

pkg/movingaverage/weight_moving_average.go
```diff
@@ -75,3 +75,11 @@ func (w *WMA) Set(n float64) {
 	w.Reset()
 	w.Add(n)
 }
+
+// GetInstantaneous returns the value just added.
+func (w *WMA) GetInstantaneous() float64 {
+	if w.count == 0 {
+		return 0
+	}
+	return w.records[(w.count-1)%w.size]
```
rleungx, line 84: Why not record the instantaneous value in Add and just return it here?

HunDunDM: Since only `MedianFilter.GetInstantaneous` will be used in the subsequent code, only this is currently modified. Others can be determined according to the frequency of use of `Add` and `GetInstantaneous` when they are used later.

disksing (approved):

Issue #3886, opened 2021-07-19
The replica count of default placement rule stays constant  after replication.max-replicas is updated 


lilinghai:
### What did you do?

deploy tidb cluster with default config
enable placement rules
online update replication.max-replicas to 1
show placement rule
### What did you expect to see?
The replica count of default placement rule  is 3. This is a little confusing whether I complete the update of  replicas count.
### What did you see instead?
The replica count of default placement rule stays same with the config replication.max-replicas.
### What version of PD are you using (`pd-server -V`)?
v5.1.0

disksing:
The cause is that `RuleManager.GetRule` does not return a copy of rule but a pointer to the rule which is stored in rule manager. In `server.go` when we update the rule we are updating the rule in rule manager at the same time. When the rule is about to save to storage, it will be skipped because if we compare the new rule and the rule in rule manager, they are exact the same.

Taken together, the phenomenon is that rule will be temporarily updated. But when the PD is restarted or a leader re-election occurs, the original configuration appears again.

Way to fix: Change methods such as GetRule to return a cloned rule.

adamotonete:
When changing the replication.max-replicas to anything different than 1 the service is dying and the client is returning:
`Failed to set config: [500] PANIC: runtime error: invalid memory address or nil pointer dereference
goroutine 24385 [running]:`

Branch master (v5.2.0-alpha-27-gc1f31284)

disksing:
Hi @adamotonete I have tried by did not reproduce the panic issue. Do you have more detailed steps to reproduce it?

PR #3892, opened 2021-07-20
placement: return rules after clone

disksing:

Fix https://github.com/tikv/pd/issues/3886

Clone the rule before returning to callers. It should be safe to modify and update again.

### Check List

Tests
- Unit test
- Integration test

Related changes

### Release note

```release-note
Fix an issue where data is not stored when using max-replicas or location-labels to indirectly update default placement rule
```


disksing:
Please take a look. @HunDunDM @bufferflies

nolouch (approved):
lgtm

server/server.go
```diff
@@ -1378,3 +1376,15 @@ func (s *Server) SaveTTLConfig(data map[string]interface{}, ttl time.Duration) e
 	}
 	return nil
 }
+
+func stringsEqual(a, b []string) bool {
```
HunDunDM, line 1380: Judge the situation where `len` is all 0, then we can still use `reflect.DeepEqual` directly?

HunDunDM (commented):
Does Copy-on-Write perform better?

server/server.go
bufferflies, line 1380: it will be better  if move the function to typeutil.

disksing:
> Does Copy-on-Write perform better?

copy-on-write should be better. Since go does not have `const`, it is easy to miss use. I think for the methods that are mainly used by HTTP API, it is acceptable to clone every rule.

server/server.go
disksing: ok.

disksing: Then it will be `if len(a) == 0 && len(b) == 0 || reflect.DeepEqual(a, )`, looks not very elegant I think.

bufferflies:
lgtm

pkg/typeutil/comparison.go
```diff
@@ -38,3 +38,16 @@ func MinDuration(a, b time.Duration) time.Duration {
 	}
 	return b
 }
+
+// StringsEqual checks if two string slices are equal. Empyt slice and nil are considered equal.
+func StringsEqual(a, b []string) bool {
```
HunDunDM, line 43: Do we need the unit test?

disksing: Too simple to test. The code was written by AI, which I feel is highly credible.

nolouch:
PTAL @HunDunDM @bufferflies

bufferflies:
lgtm

bufferflies (approved):

PR #3897, opened 2021-07-21
schedule: `isRegionHealthy` only care pending peer and down peer

### LINKED ISSUE #3896: The behavior of opt.HealthRegion is inconsistent for placement rules disable or enable
## Bug Report

image

When placement rules are disable, `isHealthRegion` will return false when the region has a learner generated by the scheduling intermediate state

When placement rules enable, `isHealthRegion` will return true when the region has a learner generated by the scheduling intermediate state

Perhaps we should also determine whether the learner meets the replica count requirement when placement rules enable.

Maybe we should only determine the down peer and pending peer, and not the learner in `isHealthRegion`

lhy1024:

close #3896 

The current implementation, when using only `isRegionHealthy`, unexpect schedule regions with learners when enabling placement rules.

However, in addition to `isRegionHealthy`, there is also `isReplicated`, which are often used together. When used together, it is actually okay because it will be checked in `isReplicated`.

Only when `isRegionHealthy` is used alone, such as in evict-leader, balance-leader, shuffle-leader, grant-leader and random-merge leaders, only `opt. HealthRegion`, but not `isReplicated`. There is not much difference whether it is allowed or not here.

So, to simplify the semantics and ensure consistent behavior, let `isRegionHealthy` only determine pending peer and down peer.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/opt/healthy.go
+++ b/server/schedule/opt/healthy.go
@@ -21,8 +21,7 @@ import (
 var balanceEmptyRegionThreshold = 50
 
 // IsRegionHealthy checks if a region is healthy for scheduling. It requires the
-// region does not have any down or pending peers. And when placement rules
-// feature is disabled, it requires the region does not have any learner peer.
+// region does not have any down or pending peers. 
 func IsRegionHealthy(cluster Cluster, region *core.RegionInfo) bool {
 	return IsHealthyAllowPending(cluster, region) && len(region.GetPendingPeers()) == 0
 }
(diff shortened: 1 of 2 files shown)


disksing (approved):

server/schedule/opt/healthy.go
```diff
@@ -21,18 +21,14 @@ import (
...
 // IsRegionHealthy checks if a region is healthy for scheduling. It requires the
-// region does not have any down or pending peers. And when placement rules
-// feature is disabled, it requires the region does not have any learner peer.
+// region does not have any down or pending peers.
 func IsRegionHealthy(cluster Cluster, region *core.RegionInfo) bool {
 	return IsHealthyAllowPending(cluster, region) && len(region.GetPendingPeers()) == 0
 }
 
 // IsHealthyAllowPending checks if a region is healthy for scheduling.
 // Differs from IsRegionHealthy, it allows the region to have pending peers.
```
rleungx, line 30: Prefer to rewrite the comment of this function.

lhy1024: What would you suggest to add?

rleungx (approved):

PR #3901, opened 2021-07-22
cluster: move stats update into a better place

### LINKED ISSUE #3199: The statistics of miss-peer is not accurate sometimes
### What did you do?

Use pd-ctl to set the max-replicas to 2 with a cluster having 15w regions. And after all extra peers are removed, change back the max-replicas to 3.

Screen Shot 2020-11-23 at 10 44 24 AM

As you can see from the above screenshot, after changing back the parameters, there should be a lot of missing peers. But actually, there are a few missing peers. 

master

### LINKED ISSUE #3286: The label monitor does not match the region numbers
### What did you do?

Create the tables and merge them all.

The number of label counter is equal to the number of regions

### What did you see instead?

Screen Shot 2020-12-18 at 5 48 36 PM

master (b77c8433f48ba5af43171e4978ec088ac45bcd6b)

### LINKED ISSUE #3558: The num of offline Region is unexpected
### What did you do?

offline 3 store
image

The num of offline Region gradually decreases from 2400 accompanied by offline

### What did you see instead?

it increased from 200
image

5.0-GA

rleungx:

Issue Number: Close #3286, close #3558, close #3199, ref #5648.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -138,6 +138,7 @@ type RaftCluster struct {
 	coordinator              *coordinator
 	labelLevelStats          *statistics.LabelStatistics
 	regionStats              *statistics.RegionStatistics
+	lastRegionStats          *statistics.RegionStatistics
 	hotStat                  *statistics.HotStat
 	hotBuckets               *buckets.HotBucketCache
 	ruleManager              *placement.RuleManager
@@ -275,6 +276,7 @@ func (c *RaftCluster) Start(s Server) error {
 	c.storeConfigManager = config.NewStoreConfigManager(c.httpClient)
 	c.coordinator = newCoordinator(c.ctx, cluster, s.GetHBStreams())
 	c.regionStats = statistics.NewRegionStatistics(c.opt, c.ruleManager, c.storeConfigManager)
+	c.lastRegionStats = statistics.NewRegionStatistics(c.opt, c.ruleManager, c.storeConfigManager)
 	c.limiter = NewStoreLimiter(s.GetPersistOptions())
 	c.externalTS, err = c.storage.LoadExternalTS()
 	if err != nil {
@@ -845,19 +847,10 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 	}
 	c.coordinator.CheckTransferWitnessLeader(region)
 
-	hasRegionStats := c.regionStats != nil
 	// Save to storage if meta is updated.
 	// Save to cache if meta or leader is updated, or contains any down/pending peer.
 	// Mark isNew if the region in cache does not have leader.
 	isNew, saveKV, saveCache, needSync := regionGuide(region, origin)
-	if !saveKV && !saveCache && !isNew {
-		// Due to some config changes need to update the region stats as well,
-		// so we do some extra checks here.
-		if hasRegionStats && c.regionStats.RegionStatsNeedUpdate(region) {
-			c.regionStats.Observe(region, c.getRegionStoresLocked(region))
-		}
-		return nil
-	}
 
 	failpoint.Inject("concurrentRegionHeartbeat", func() {
 		time.Sleep(500 * time.Millisecond)
@@ -875,18 +868,7 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		if overlaps, err = c.core.AtomicCheckAndPutRegion(region); err != nil {
 			return err
 		}
-
-		for _, item := range overlaps {
-			if c.regionStats != nil {
-				c.regionStats.ClearDefunctRegion(item.GetID())
-			}
-			c.labelLevelStats.ClearDefunctRegion(item.GetID())
-		}
 		regionUpdateCacheEventCounter.Inc()
-	}
-
-	if hasRegionStats {
-		c.regionStats.Observe(region, c.getRegionStoresLocked(region))
 	}
 
 	if !c.IsPrepared() && isNew {
@@ -1031,6 +1013,11 @@ func (c *RaftCluster) GetStoreRegionCount(storeID uint64) int {
 // GetAverageRegionSize returns the average region approximate size.
 func (c *RaftCluster) GetAverageRegionSize() int64 {
 	return c.core.GetAverageRegionSize()
+}
+
+// IsLastRegionStatsEmpty returns if region statistics is empty. Only for test purpose.
+func (c *RaftCluster) IsLastRegionStatsEmpty() bool {
+	return c.lastRegionStats.IsEmpty()
 }
 
 // DropCacheRegion removes a region from the cache.
@@ -1880,7 +1867,7 @@ func (c *RaftCluster) collectMetrics() {
 
 	c.coordinator.collectSchedulerMetrics()
 	c.coordinator.collectHotSpotMetrics()
-	c.collectClusterMetrics()
+	c.collectHotStatsMetrics()
 	c.collectHealthStatus()
 }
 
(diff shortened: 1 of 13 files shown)


server/cluster/cluster.go
```diff
@@ -714,6 +714,11 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		c.regionStats.Observe(region, c.getRegionStoresLocked(region))
 	}
 
+	// Updates the label level isolation statistics.
+	if c.labelLevelStats != nil {
+		c.labelLevelStats.Observe(region, c.getRegionStoresLocked(region), c.opt.GetLocationLabels())
```
nolouch, line 719: I guess it‘s may cost many CPUs.

disksing: +1

disksing:
I think even if update statistics in heartbeat, it can still lose sync.

server/cluster/cluster.go
```diff
@@ -60,6 +60,7 @@ const (
 	defaultChangedRegionsLimit = 10000
 	// persistLimitRetryTimes is used to reduce the probability of the persistent error
 	// since the once the store is add or remove, we shouldn't return an error even if the store limit is failed to persist.
+	defaultStatsJobLimit   = 10000
```
JmPotato, line 63: The comments above will mismatch.

rleungx:
> I think even if update statistics in heartbeat, it can still lose sync.

Yes. I decide to leave it in region heartbeat process for now.

server/cluster/coordinator_test.go
```diff
@@ -286,10 +286,11 @@ func prepare(setCfg func(*config.ScheduleConfig), setTc func(*testCluster), run
 	}
 	tc := newTestCluster(ctx, opt)
 	hbStreams := hbstream.NewTestHeartbeatStreams(ctx, tc.getClusterID(), tc, true /* need to run */)
+	co := newCoordinator(ctx, tc.RaftCluster, hbStreams)
+	tc.coordinator = co
```
JmPotato, line 290: Maybe we can merge these two lines and use `run(tc.coordinator)` later?

server/cluster/cluster_test.go
```diff
@@ -879,14 +881,20 @@ func (s *testClusterInfoSuite) TestOfflineAndMerge(c *C) {
...
+	// wait for background job goroutine to handle the stats task.
+	time.Sleep(time.Second)
 	c.Assert(cluster.GetOfflineRegionStatsByType(statistics.OfflinePeer), HasLen, len(regions))
 
 	// Merge.
 	for i := 0; i < n; i++ {
 		regions = core.MergeRegions(regions)
 		heartbeatRegions(c, cluster, regions)
+		// wait for background job goroutine to handle the stats task.
+		time.Sleep(100 * time.Millisecond)
```
JmPotato, line 893: I am concerned this kind of waiting is unstable, which will make the test fail in an unstable test environment. 🤔 Though if there is no other way to do this, I'm ok with it.

server/cluster/cluster.go
```diff
@@ -316,10 +316,14 @@ func (c *RaftCluster) LoadClusterInfo() (*RaftCluster, error) {
...
 }
 
-func (c *RaftCluster) runBackgroundJobs(interval time.Duration) {
+func (c *RaftCluster) runBackgroundJobs() {
 	defer logutil.LogPanic()
 	defer c.wg.Done()
 
+	interval := backgroundJobInterval
+	failpoint.Inject("highFrequencyClusterJobs", func() {
+		backgroundJobInterval = 100 * time.Microsecond
```
JmPotato, line 325: Should update `interval` here, not `backgroundJobInterval`.

server/cluster/coordinator_test.go
rleungx: There is not much difference here.

server/cluster/cluster_test.go
rleungx: Yes. But I have no idea about it.

JmPotato (approved):

server/cluster/cluster.go
```diff
@@ -1369,12 +1371,22 @@ func (c *RaftCluster) GetOfflineRegionStatsByType(typ statistics.RegionStatistic
 	return c.regionStats.GetOfflineRegionStatsByType(typ)
 }
 
-func (c *RaftCluster) updateRegionsLabelLevelStats(regions []*core.RegionInfo) {
+func (c *RaftCluster) updateStats(region *core.RegionInfo) {
```
disksing, line 1374: I think we should handle region in batches. Lock and Unlock for every region could be slow.

server/cluster/cluster.go
```diff
@@ -703,14 +705,14 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
-	}
-
 	changedRegions := c.changedRegions
-
 	c.Unlock()
 
+	for _, item := range overlaps {
+		statsJobs = append([]func(){func() { c.clearStats(item) }}, statsJobs...)
+	}
+	c.statsJobCh <- statsJobs
```
disksing, line 714: If the channel is full. It can affect handling heartbeat.

rleungx: good idea

server/cluster/cluster.go
rleungx: good idea

server/cluster/cluster.go
```diff
@@ -338,6 +345,38 @@ func (c *RaftCluster) runBackgroundJobs(interval time.Duration) {
...
+	pendingStatsTasks := make([]*statsTask, 0, maxPendingStatsTasks)
+	for {
+		select {
+		case <-c.ctx.Done():
+			log.Info("process region heartbeat job has been stopped")
+			return
+		case tasks := <-c.statsTaskCh:
+			pendingStatsTasks = append(pendingStatsTasks, tasks...)
+			if len(pendingStatsTasks) < maxPendingStatsTasks {
+				continue
```
JmPotato, line 368: According to the [Go spec](https://go.dev/ref/spec#Continue_statements):

> A "continue" statement begins the next iteration of the innermost "for" loop at its post statement. The "for" loop must be within the same function.

So the `continue` here will continue to the next for loop if the length of `pendingStatsTasks` is smaller than `maxPendingStatsTasks`?

server/cluster/coordinator.go
```diff
@@ -157,6 +157,36 @@ func (c *coordinator) patrolRegions() {
...
+		regions := c.cluster.ScanRegions(key, nil, statsScanRegionLimit)
+		if len(regions) == 0 {
+			// Resets the scan key.
+			key = nil
+			continue
+		}
+
+		c.cluster.updateRegionStats(regions)
+		if len(key) == 0 {
+			c.cluster.collectStatsMetrics()
```
disksing, line 185: We can sleep a while here. If there are not many regions, we don't to collect stats too often.

server/cluster/coordinator.go
```diff
@@ -157,6 +157,36 @@ func (c *coordinator) patrolRegions() {
...
+		}
+
+		regions := c.cluster.ScanRegions(key, nil, statsScanRegionLimit)
+		if len(regions) == 0 {
+			// Resets the scan key.
+			key = nil
+			continue
+		}
+
+		c.cluster.updateRegionStats(regions)
```
disksing, line 183: We can create `regionStats` and `labelLevelStats` locally and set them into cluster after all regions are collected. No need to lock `cluster` every batch.

rleungx: We still need to get the store so that the lock cannot be eliminated.

server/cluster/coordinator.go
```diff
@@ -189,9 +184,10 @@ func (c *coordinator) collectRegionStats() {
 
 		c.cluster.updateRegionStats(regions)
 		if len(key) == 0 {
-			if c.cluster.GetRegionCount() <= maxPauseRegionNum {
-				time.Sleep(pauseInterval)
+			if time.Since(start) < defaultScrapInterval {
+				time.Sleep(defaultScrapInterval - time.Since(start))
```
disksing, line 188: sleep should be after `collectStatsMetrics`.

server/cluster/cluster.go
```diff
@@ -1302,14 +1299,21 @@ func (c *RaftCluster) resetMetrics() {
...
+	c.Lock()
+	defer c.Unlock()
 	if c.regionStats == nil {
-		c.RUnlock()
 		return
 	}
 	c.regionStats.Collect()
 	c.labelLevelStats.Collect()
+	c.lastRegionStats = c.regionStats.Clone()
+	c.labelLevelStats = statistics.NewLabelStatistics()
```
disksing, line 1311: Do we need to handle `labelLevelStats` the same way as `regionStats`?

rleungx: No, we don't, `lastRegionStats` is used for API.

disksing (approved):

PR #3903, opened 2021-07-23
*: Create SECURITY.md

s3nt3:

Announce the way and specifications for users to submit vulnerabilities to the tidb security team through Github-Security-Policy.

Refer to the steps mentioned in [add a security policy to your repository](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) to create SECURITY.md.

### Check List

Tests

- No code

### Release note

```release-note
None
```

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,33 @@
+# Security Vulnerability Disclosure and Response Process
+
+TiDB is a fast-growing open source database. To ensure its security, a security vulnerability disclosure and response process is adopted.
+
+The primary goal of this process is to reduce the total exposure time of users to publicly known vulnerabilities. To quickly fix vulnerabilities of TiDB products, the security team is responsible for the entire vulnerability management process, including internal communication and external disclosure.
+
+If you find a vulnerability or encounter a security incident involving vulnerabilities of TiDB products, please report it as soon as possible to the TiDB security team (security@tidb.io).
+
+Please kindly help provide as much vulnerability information as possible in the following format:
+
+- Issue title*:
+
+- Overview*:
+
+- Affected components and version number*:
+
+- CVE number (if any):
+
+- Vulnerability verification process*:
+
+- Contact information*:
+
+The asterisk (*) indicates the required field.
+
+# Response Time
+
+The TiDB security team will confirm the vulnerabilities and contact you within 2 working days after your submission.
+
+We will publicly thank you after fixing the security vulnerability. To avoid negative impact, please keep the vulnerability confidential until we fix it. We would appreciate it if you could obey the following code of conduct:
+
+The vulnerability will not be disclosed until TiDB releases a patch for it.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,33 @@
+# Security Vulnerability Disclosure and Response Process
+
+TiDB is a fast-growing open source database. To ensure its security, a security vulnerability disclosure and response process is adopted.
+
+The primary goal of this process is to reduce the total exposure time of users to publicly known vulnerabilities. To quickly fix vulnerabilities of TiDB products, the security team is responsible for the entire vulnerability management process, including internal communication and external disclosure.
+
+If you find a vulnerability or encounter a security incident involving vulnerabilities of TiDB products, please report it as soon as possible to the TiDB security team (security@tidb.io).
```
rleungx, line 7: What is the difference between security@tidb.io and tikv-security@lists.cncf.io?

s3nt3: Thanks a lot, this is a good question. It seems that tikv and tidb use different Security Policies and vulnerability reporting emails. I will close this PR and re-propose a new PR to migrate tikv Security Policy to this repo.

s3nt3:
Hi @rleungx, after confirming that the tikv project was donated to cncf, the security vulnerabilities of tikv were submitted through tikv-security@lists.cncf.io. As tikv/pd is a key component of tidb, it is recommended to adopt tidb's security policy. Vulnerabilities should be submitted through security@tidb.io and handled by the tidb security team.

SECURITY.md
disksing: I think we should use the same email as tikv, since it is a project in the tikv org.

s3nt3: It is not just choosing an email, but also determining the security policy. If we use the email address security@tidb.io, the security vulnerability of pd will be responded by the tidb security team. If we use the email address tikv-security@lists.cncf.io, which is a cncf email address, I am not sure whether the vulnerability information is submitted to the foundation or the project maintainer, but the tidb security team may not be able to obtain the vulnerability information in the first time.

disksing: So the tidb security team cares more about `github.com/tikv/pd` but less about the security issues of `github.com/tikv/tikv`?

disksing: To clarify: I didn't suggest using tidb or CNCF email addresses, I just think tikv/pd should be consistent with tikv/tikv because they both belong to the tikv organization.

s3nt3: There are two reasons for submitting this PR. At first, currently PD does not have a security policy to clearly inform users how to submit vulnerabilities; secondly, PD is a key component of tidb, and I personally think that the tidb security team should respond to its vulnerabilities. Anyway, the final choice of security policy is up to the maintainer to decide.

disksing: I still don't understand why tikv and pd have to be handled differently.

PR #3908, opened 2021-07-26
ci:Switch to self-hosted runner.

n0vad3v:

use self-host github action
### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Please add a release note.

```


nolouch:
It works on self-host GitHub action, we can use it with the integration test. for the unit tests, I prefer to use the current official GitHub action.

JmPotato:

PR #3922, opened 2021-07-29
server/schedulers: introduce evict-slow-store-scheduler 

5kbpers:

For addressing https://github.com/tikv/tikv/issues/10539, we will introduce a `slow score` concept into tikv and report it to pd in store heartbeats. We then implement `evict-slow-store-scheduler` to detect and evict all leaders from one store that was considered slow.

Blocked by https://github.com/pingcap/kvproto/pull/789

* implement `evict-slow-store-scheduler`
* show `slow score` in store info

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Related changes

### Release note

```release-note
introduce evict-slow-store-scheduler for detecting and evicting slow stores
```


disksing:
Hi, I have some concerns about it:
- Can it really solve the problems of poor tikv performance, considering that even if the leader is removed, follower writes may still bring other problems when the speed is too slow.
- What should be the behavior if all tikv scores are low?
- The condition whether it is slow or not should be added to the stateFilter, otherwise other schedulers will conflict with evict-slow-scheduler.

5kbpers:
> Hi, I have some concerns about it:
> 
> * Can it really solve the problems of poor tikv performance, considering that even if the leader is removed, follower writes may still bring other problems when the speed is too slow.
> * What should be the behavior if all tikv scores are low?
> * The condition whether it is slow or not should be added to the stateFilter, otherwise other schedulers will conflict with evict-slow-scheduler.

* Currently we can only handle the situation that only one tikv gets slow down. Then the majority of replicas could have the ability to continue to serve requests without serious regression in most cases
* If most of TiKVs get slow down, the only thing we can do is alerting the user and handle it manually

go.mod
```diff
@@ -54,3 +54,5 @@ require (
 	google.golang.org/grpc v1.26.0
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 )
+
+replace github.com/pingcap/kvproto v0.0.0-20210604082642-dda0a102bc6a => github.com/5kbpers/kvproto v0.0.0-20210714075718-bae4198f5778
```
disksing, line 58: can update to latest kvproto now.

server/cluster/cluster.go
```diff
@@ -1155,7 +1166,9 @@ func (c *RaftCluster) checkStores() {
 	var offlineStores []*metapb.Store
 	var upStoreCount int
 	stores := c.GetStores()
+	storesMap := make(map[uint64]*core.StoreInfo)
 	for _, store := range stores {
+		storesMap[store.GetID()] = store
```
disksing, line 1171: what is it used for?

disksing, line 1171: what is it used for?

server/schedulers/evict_slow_store.go
```diff
@@ -0,0 +1,221 @@
...
+				slowStores = append(slowStores, store)
+			}
+		}
+
+		// If there is only one slow store, evict leaders from that store.
+		if len(slowStores) == 1 && slowStores[0].GetSlowScore() >= slowStoreEvictThreshold {
+			store := slowStores[0]
+			log.Info("detected slow store, start to evict leaders",
+				zap.Stringer("store", store.GetMeta()))
+			s.conf.EvictedStores = []uint64{store.GetID()}
```
disksing, line 185: As we can evict leader from only 1 slow store, why use a vector to store the ID>

server/schedulers/evict_slow_store.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (conf *evictSlowStoreSchedulerConfig) getSchedulerName() string {
+	return EvictSlowStoreName
+}
+
+type evictSlowStoreScheduler struct {
+	*BaseScheduler
+	conf                 *evictSlowStoreSchedulerConfig
+	evictLeaderScheduler *evictLeaderScheduler
```
disksing, line 94: I think it is not a very good idea to use another scheduler as a toolkit because each scheduler has its own context. For example, scheduler names will be recorded to the filter metrics, it may happen that we want to record evict-slow-store-scheduler and end up with evict-leader-scheduler.
There are two possible ways to improve this: a) reimplement the similar code again b) extract the common parts and let different schedulers call them

5kbpers: The logic is basically the same so I think it would be better to reuse it directly. About the metrics, I saw that it is based on the name of the scheduler, may setting a customized name is a good solution?

server/schedulers/evict_slow_store.go
5kbpers: I thought there would be a chance to introduce a more complex strategy for this scheduler. So use a vector here for keeping flexibility.

lhy1024:
This pr will release in sprint4, PTAL

tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -267,6 +268,16 @@ func NewBalanceLeaderSchedulerCommand() *cobra.Command {
 	return c
 }
 
+// NewEvictSlowStoreSchedulerCommand returns a command to add a evict-slow-store-scheduler.
+func NewEvictSlowStoreSchedulerCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "evict-slow-store-scheduler",
```
bufferflies, line 274: the schedule should add store_id param???

5kbpers: No, it just needs the id of the previous evicted store which was stored in etcd.

server/schedulers/evict_slow_store.go
disksing: Consider extracting the common part?

server/core/basic_cluster.go
```diff
@@ -429,6 +444,9 @@ type StoreSetController interface {
 	PauseLeaderTransfer(id uint64) error
 	ResumeLeaderTransfer(id uint64)
 
+	SlowStoreEvicted(id uint64) error
```
rleungx, line 447: Can we reuse `PauseLeaderTransfer`/`ResumeLeaderTransfer`?

5kbpers: That will make this scheduler conflicts with evict-leader-scheduler

server/schedulers/evict_slow_store.go
```diff
@@ -0,0 +1,212 @@
...
+		} else if store.GetSlowScore() <= slowStoreRecoverThreshold {
+			log.Info("slow store has been recovered",
+				zap.Stringer("store", store.GetMeta()))
+		} else {
+			return s.schedulerEvictLeader(cluster)
+		}
+		// Stop to evict leaders
+		s.cleanupEvictLeader(cluster)
+		s.conf.EvictedStores = []uint64{}
+		s.conf.Persist()
```
rleungx, line 168: What if the persistent fails? BTW, do we have other ways to control this config?

server/schedulers/evict_slow_store.go
```diff
@@ -0,0 +1,212 @@
...
+	var ops []*operator.Operator
+
+	evictedStores := s.conf.EvictedStores
+	if len(evictedStores) != 0 {
+		store := cluster.GetStore(evictedStores[0])
+		if store == nil || store.IsTombstone() {
+			// Previous slow store had been removed, remove the sheduler and check
+			// slow node next time.
+			log.Info("slow store has been removed",
+				zap.Stringer("store", store.GetMeta()))
```
rleungx, line 158: Maybe id is enough? If you wanna print meta, I think you should use `ZapRedactStringer`.

server/schedule/filter/filters.go
```diff
@@ -378,12 +383,13 @@ func (f *StoreStateFilter) anyConditionMatch(typ int, opt *config.PersistOptions
 	var funcs []conditionFunc
 	switch typ {
 	case leaderSource:
-		funcs = []conditionFunc{f.isTombstone, f.isDown, f.pauseLeaderTransfer, f.isDisconnected}
+		funcs = []conditionFunc{f.isTombstone, f.isDown, f.pauseLeaderTransfer, f.isDisconnected,
+			f.slowStoreEvicted}
```
disksing, line 387: I think slow store can be the source of transfer leader.

server/schedulers/evict_leader.go
```diff
@@ -30,6 +27,10 @@ import (
 	"github.com/tikv/pd/server/schedule/operator"
 	"github.com/tikv/pd/server/schedule/opt"
 	"github.com/unrolled/render"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/log"
```
disksing, line 33: em, why changed the import order?

server/schedulers/evict_leader.go
```diff
@@ -235,35 +236,15 @@ func (s *evictLeaderScheduler) IsScheduleAllowed(cluster opt.Cluster) bool {
...
-		op.SetPriorityLevel(core.HighPriority)
-		op.Counters = append(op.Counters, schedulerCounter.WithLabelValues(s.GetName(), "new-operator"))
-		ops = append(ops, op)
-	}
-	return ops
+	return scheduleEvictLeaderBatch(s.GetName(), cluster, s.conf.StoreIDWithRanges, EvictLeaderBatchSize)
 }
 
-func (s *evictLeaderScheduler) uniqueAppend(dst []*operator.Operator, src ...*operator.Operator) []*operator.Operator {
+func uniqueAppend(dst []*operator.Operator, src ...*operator.Operator) []*operator.Operator {
```
disksing, line 247: `uniqueAppendOperator` may be better.

disksing (approved):

rleungx (approved):

PR #3923, opened 2021-07-29
scheduler: use qps priority with multi priorities in hot region scheduler 

lhy1024:

The current hotspot identification mainly uses byte and key dimensions, which are not very fit for load requests with high CPU overhead but low hard disk overhead, we plan to add QPS dimension to better describe the distribution of CPU resources.

use qps dimension  as the first priority in  hot region scheduler

### Check List

Tests

- Unit test
- 
### Release note

```release-note
use qps dimension  as the first priority in  hot region scheduler
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -219,7 +219,7 @@ func (h *hotScheduler) gcRegionPendings() {
 }
 
 // summaryStoresLoad Load information of all available stores.
-// it will filtered the hot peer and calculate the current and future stat(rate,count) for each store
+// it will filter the hot peer and calculate the current and future stat(rate,count) for each store
 func summaryStoresLoad(
 	stores []*core.StoreInfo,
 	storesLoads map[uint64][]float64,
@@ -1096,7 +1096,7 @@ func (bs *balanceSolver) buildOperator() (op *operator.Operator, infl *Influence
 
 	dim := ""
 	if bs.firstPriorityIsBetter && bs.secondPriorityIsBetter {
-		dim = "both"
+		dim = "all"
 	} else if bs.firstPriorityIsBetter {
 		dim = dimToString(bs.firstPriority)
 	} else if bs.secondPriorityIsBetter {
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -594,7 +594,7 @@ func (bs *balanceSolver) filterSrcStores() map[uint64]*storeLoadDetail {
 
 func (bs *balanceSolver) checkSrcByDimPriorityAndTolerance(minLoad, expectLoad *storeLoad) bool {
 	return slice.AllOf(minLoad.Loads, func(i int) bool {
-		if statistics.IsSelectedDim(i) {
+		if bs.isSelectedDim(i) {
 			return minLoad.Loads[i] > bs.sche.conf.GetSrcToleranceRatio()*expectLoad.Loads[i]
 		}
 		return true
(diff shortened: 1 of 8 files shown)


server/schedulers/utils.go
```diff
@@ -282,12 +284,14 @@ func (load storeLoad) ToLoadPred(rwTy rwType, infl *Influence) *storeLoadPred {
...
 
-func stLdByteRate(ld *storeLoad) float64 {
-	return ld.Loads[statistics.ByteDim]
-}
-
-func stLdKeyRate(ld *storeLoad) float64 {
-	return ld.Loads[statistics.KeyDim]
+func stLdRate(dim int) func(ld *storeLoad) float64 {
+	return func(ld *storeLoad) float64 {
+		switch dim {
```
rleungx, line 289: no need to use `switch`?

server/schedulers/hot_region.go
```diff
@@ -219,7 +219,7 @@ func (h *hotScheduler) gcRegionPendings() {
 }
 
 // summaryStoresLoad Load information of all available stores.
-// it will filtered the hot peer and calculate the current and future stat(byte/key rate,count) for each store
+// it will filtered the hot peer and calculate the current and future stat(rate,count) for each store
```
rleungx, line 222:
```suggestion
// it will filter the hot peer and calculate the current and future stat (rate, count) for each store
```

server/schedulers/hot_region.go
```diff
@@ -414,14 +413,18 @@ type balanceSolver struct {
...
-	rankStep       *storeLoad
-	firstPriority  int
-	secondPriority int
+	maxSrc   *storeLoad
+	minDst   *storeLoad
+	rankStep *storeLoad
 
-	byteIsBetter bool
-	keyIsBetter  bool
+	firstPriority             int
```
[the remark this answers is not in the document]
nolouch: +1, maybe directly use the array?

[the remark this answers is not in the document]
Yisaer: we only consider 2 dimension during calculating rank now, so introducing new dimension won't change the logic here.

server/schedulers/hot_region.go
```diff
@@ -779,43 +801,38 @@ func (bs *balanceSolver) calcProgressiveRank() {
...
+			bs.secondPriorityIsBetter = true
 		case firstPriorityDimHot && firstPriorityDecRatio <= greatDecRatio:
-			// If belong to the case, byte rate will be more balanced, ignore the key rate.
+			// If belong to the case, first priority dim will be more balanced, ignore the second priority dim.
 			rank = -1
-			bs.byteIsBetter = true
+			bs.firstPriorityIsBetter = true
 		}
 	}
-	log.Debug("calcProgressiveRank",
```
Yisaer, line 808: why remove debug log?

server/schedulers/hot_region.go
```diff
@@ -1079,12 +1095,12 @@ func (bs *balanceSolver) buildOperator() (op *operator.Operator, infl *Influence
 	}
 
 	dim := ""
-	if bs.byteIsBetter && bs.keyIsBetter {
+	if bs.firstPriorityIsBetter && bs.secondPriorityIsBetter {
 		dim = "both"
```
Yisaer, line 1099:
```suggestion
		dim = "all"
```

I think `all` is better

Yisaer (commented):
Basically lgtm

server/schedulers/hot_region.go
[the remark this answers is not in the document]
lhy1024: Currently only two dimensions are considered, if a third dimension is introduced, the whole scheduling policy may need to be modified

server/schedulers/hot_region.go
lhy1024: If you use debug mode, a lot of logs will be printed and we can get this information through the firstPriorityIsBetter  metrices

Yisaer (approved):

nolouch:
PTAL @HunDunDM @bufferflies

nolouch (approved):

PR #3926, opened 2021-07-30
scheduler: use pending amp in hot region scheduler

lhy1024:

In the past, for two stores with large differences, it was easy to have redundant scheduling behavior, because as the difference between the two stores approached, the speed might not come down in time, so we limited the speed by adding pending amps, and the closer the two stores were, the fewer regions were allowed to be scheduled at the same time.

Before
image

After
image

isTolerance checks source store and target store by checking the difference value with pendingAmpFactor * pendingPeer. This will make the hot region scheduling slow even serialized running when each 2 store's pending influence is close.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -69,6 +69,8 @@ const (
 
 // schedulePeerPr the probability of schedule the hot peer.
 var schedulePeerPr = 0.66
+
+// pendingAmpFactor will amplify the impact of pending influence, making scheduling slower or even serial when two stores are close together
 var pendingAmpFactor = 8.0
 
 type hotScheduler struct {
@@ -514,7 +516,7 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	switch {
 	case bs.isForWriteLeader():
 		maxZombieDur = bs.sche.conf.GetRegionsStatZombieDuration()
-	case bs.rwTy == write && bs.opTy == movePeer:
+	case bs.isForWritePeer():
 		if bs.best.srcDetail.Info.IsTiFlash {
 			maxZombieDur = bs.sche.conf.GetRegionsStatZombieDuration()
 		} else {
@@ -528,6 +530,10 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 
 func (bs *balanceSolver) isForWriteLeader() bool {
 	return bs.rwTy == write && bs.opTy == transferLeader
+}
+
+func (bs *balanceSolver) isForWritePeer() bool {
+	return bs.rwTy == write && bs.opTy == movePeer
 }
 
 // filterSrcStores compare the min rate and the ratio * expectation rate, if two dim rate is greater than
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -394,13 +394,13 @@ func (bs *balanceSolver) init() {
 	// For write, they are different
 	switch bs.rwTy {
 	case read:
-		bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetReadPriorities(), []string{BytePriority, KeyPriority})
+		bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetReadPriorities(), getReadLeaderPriorities)
 	case write:
 		switch bs.opTy {
 		case transferLeader:
-			bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetWriteLeaderPriorites(), []string{KeyPriority, BytePriority})
+			bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetWriteLeaderPriorities(), getWriteLeaderPriorities)
 		case movePeer:
-			bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetWritePeerPriorites(), []string{BytePriority, KeyPriority})
+			bs.firstPriority, bs.secondPriority = bs.adjustConfig(bs.sche.conf.GetWritePeerPriorities(), getWritePeerPriorities)
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


server/schedulers/hot_region.go
```diff
@@ -795,39 +796,66 @@ func (bs *balanceSolver) checkDstByPriorityAndTolerance(maxLoad, expect *storeLo
...
+			if !bs.isTolerance(src, dst, bs.firstPriority) {
+				return
+			}
+			bs.cur.progressiveRank = -1
 			bs.firstPriorityIsBetter = true
 		}
 	}
-	bs.cur.progressiveRank = rank
+}
+
```
Yisaer, line 847:
```suggestion
`isTolerance` checks source store and target store by checking the difference value with pendingAmpFactor * pendingPeer. This will make the hot region scheduling serializely running when each 2 store's pending influence is close.
```

Yisaer (approved):
basically lgtm

server/schedulers/hot_region.go
```diff
@@ -795,39 +796,68 @@ func (bs *balanceSolver) checkDstByPriorityAndTolerance(maxLoad, expect *storeLo
...
+// isTolerance checks source store and target store by checking the difference value with pendingAmpFactor * pendingPeer.
+// This will make the hot region scheduling slow even serializely running when each 2 store's pending influence is close.
+func (bs *balanceSolver) isTolerance(src, dst *storeLoadPred, dim int) bool {
+	srcRate := src.Current.Loads[dim]
+	dstRate := dst.Current.Loads[dim]
+	if srcRate <= dstRate {
+		return false
+	}
+	pendingAmp := (1 + pendingAmpFactor*srcRate/(srcRate-dstRate))
```
bufferflies, line 856: the queals need src>dst all dim ?

lhy1024: yes, src need to be larger than dst

server/schedulers/hot_region_config.go
```diff
@@ -82,6 +83,7 @@ type hotRegionSchedulerConfig struct {
 	MinorDecRatio          float64  `json:"minor-dec-ratio"`
 	SrcToleranceRatio      float64  `json:"src-tolerance-ratio"`
 	DstToleranceRatio      float64  `json:"dst-tolerance-ratio"`
+	PendingAmpFactor       float64  `json:"pending-amp-factor"`
```
rleungx, line 86: How to choose the right value for this config? Assume that if we encounter a hot spot problem online, what is the right way to change it? If we don't know, then this config is meanless.

server/schedulers/hot_region.go
```diff
@@ -868,39 +868,68 @@ func (bs *balanceSolver) checkDstByPriorityAndTolerance(maxLoad, expect *storeLo
...
+	dst := bs.stLoadDetail[bs.cur.dstStoreID].LoadPred
+	bs.cur.progressiveRank = 0
+	srcLd := src.min()
+	dstLd := dst.max()
 	peer := bs.cur.srcPeerStat
-	rank := int64(0)
+
 	if bs.rwTy == write && bs.opTy == transferLeader {
 		// In this condition, CPU usage is the matter.
 		// Only consider key rate or query rate.
```
rleungx, line 880: Does the comment need to change?

lhy1024: Generally speaking, we would like to use qps or key as the dimension of qps instead of byte. or is it better to prompt the user when he/she sets it?

server/schedulers/hot_region.go
```diff
@@ -868,39 +868,68 @@ func (bs *balanceSolver) checkDstByPriorityAndTolerance(maxLoad, expect *storeLo
...
+
+// isTolerance checks source store and target store by checking the difference value with pendingAmpFactor * pendingPeer.
+// This will make the hot region scheduling slow even serializely running when each 2 store's pending influence is close.
+func (bs *balanceSolver) isTolerance(src, dst *storeLoadPred, dim int) bool {
+	srcRate := src.Current.Loads[dim]
+	dstRate := dst.Current.Loads[dim]
+	if srcRate <= dstRate {
+		return false
+	}
+	pendingAmp := (1 + bs.sche.conf.GetPendingAmpFactor()*srcRate/(srcRate-dstRate))
```
bufferflies, line 928: it will panic if srcRate equal dstRate

server/schedulers/hot_region_config.go
lhy1024: Smaller, meaning that the influence of pending will be smaller, while other solution spaces may be searched for, but maybe unstable.
Larger means that the impact of pending will be greater, while scheduling may be slower and more stable.
As for the exact number, there may not be an exact conclusion, at least not at the moment.

rleungx: What I wanna emphasize is if there is not an exact conclusion for adjusting this config, we'd better not expose it or just use a constant. Once we have a best practice for it, then make it configurable.

server/schedulers/hot_region.go
rleungx: Here you only have a variable named firstPriority. It's hard to get any information from it. If you think we need to make it clear, prefer to move this comment to the place where the value is set.
BTW, how about using a function for `bs.rwTy == write && bs.opTy == transferLeader`? There are lots of places using it.

server/schedulers/hot_region_config.go
lhy1024: use a constant

lhy1024:
@rleungx PTAL

server/schedulers/hot_region_test.go
```diff
@@ -1397,10 +1411,19 @@ func (s *testInfluenceSerialSuite) TestInfluenceByRWType(c *C) {
 	c.Assert(nearlyAbout(pendingInfluence[4].Loads[statistics.RegionReadKeys], 0.5*MB), IsTrue)
 	c.Assert(nearlyAbout(pendingInfluence[4].Loads[statistics.RegionReadBytes], 0.5*MB), IsTrue)
 
+	// consider pending amp, there are nine regions or more.
 	addRegionInfo(tc, write, []testRegionInfo{
```
rleungx, line 1415: It's better to use a for loop.

rleungx (approved):

server/schedulers/utils.go
```diff
@@ -317,6 +317,18 @@ func (lp *storeLoadPred) max() *storeLoad {
...
 
+func (lp *storeLoadPred) pending() *storeLoad {
+	mx, mn := lp.max(), lp.min()
+	loads := make([]float64, len(mx.Loads))
+	for i := range loads {
+		loads[i] = mx.Loads[i] - mn.Loads[i]
+	}
+	return &storeLoad{
+		Loads: loads,
+		Count: mx.Count,
```
HunDunDM, line 328: why max?

server/schedulers/hot_region.go
```diff
@@ -69,6 +69,7 @@ const (
 
 // schedulePeerPr the probability of schedule the hot peer.
 var schedulePeerPr = 0.66
+var pendingAmpFactor = 8.0
```
HunDunDM, line 72: add some comment?

server/schedulers/hot_region.go
```diff
@@ -496,7 +512,7 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	// If the statistics are from the sum of Regions, there will be a longer ZombieDuration.
 	var maxZombieDur time.Duration
 	switch {
-	case bs.rwTy == write && bs.opTy == transferLeader:
+	case bs.isForWriteLeader():
 		maxZombieDur = bs.sche.conf.GetRegionsStatZombieDuration()
 	case bs.rwTy == write && bs.opTy == movePeer:
```
HunDunDM, line 517: isForWritePeer?

server/schedulers/hot_region_test.go
```diff
@@ -401,23 +401,28 @@ func (s *testHotWriteRegionSchedulerSuite) TestByteRateOnlyWithTiFlash(c *C) {
...
 	for _, r := range testRegions {
 		regionBytesSum += r.byteRate
 		regionKeysSum += r.keyRate
+		regionQuerySum += r.queryRate
 	}
 	for _, r := range testRegions[0:3] {
 		hotRegionBytesSum += r.byteRate
 		hotRegionKeysSum += r.keyRate
+		hotRegionQuerySum += r.queryRate
+
```
HunDunDM, line 425:
```suggestion
```

HunDunDM (approved):

PR #3937, opened 2021-08-02
scheduler: set different priorities for write-leader and write-peer

lhy1024:

write-leader and write-peer need to have different priorities because write leader needs to have balanced metrics that reflect the leader, such as sum of key or qps, but write-peer does not, and write-peer does not necessarily have data on qps.

set different priorities for write-leader and write-peer

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/pdctl/scheduler/scheduler_test.go
+++ b/tests/pdctl/scheduler/scheduler_test.go
@@ -310,7 +310,7 @@ func (s *schedulerTestSuite) TestScheduler(c *C) {
 	mustExec([]string{"-u", pdAddr, "scheduler", "config", "balance-hot-region-scheduler"}, &conf1)
 	c.Assert(conf1, DeepEquals, expected1)
 
-	// write-priorities is  divided into write-leader-priorities and write-peer-priorities
+	// write-priorities is divided into write-leader-priorities and write-peer-priorities
 	mustExec([]string{"-u", pdAddr, "scheduler", "config", "balance-hot-region-scheduler", "set", "write-priorities", "key,byte"}, nil)
 	mustExec([]string{"-u", pdAddr, "scheduler", "config", "balance-hot-region-scheduler"}, &conf1)
 	c.Assert(conf1, DeepEquals, expected1)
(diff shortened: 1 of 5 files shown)


nolouch:
It's may conflict? especially we add qps.

lhy1024:
> It's may conflict? especially we add qps.

In fact, the write leader's priority now has key as first priority and byte as second priority, while the write peer's priority has byte as first priority and key as second priority. They are now different.

This pr doesn't change the default behaviour, it just avoids query in the write peer's scheduling priority. because the follower doesn't have this statistic.

tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -539,6 +539,10 @@ func postSchedulerConfigCommandFunc(cmd *cobra.Command, schedulerName string, ar
 					schedulers.KeyPriority))
 				return
 			}
+			if priority == schedulers.QueryPriority && key == "write-peer-priorities" {
+				cmd.Println("For hotspot scheduling of write peers, it is not reasonable to use the qps dimension.")
```
Yisaer, line 543:
```suggestion
				cmd.Println("qps is not allowed to be set in priorities for write-peer-priorities")
```

Yisaer (approved):
Rest LGTM

tests/pdctl/scheduler/scheduler_test.go
```diff
@@ -309,6 +310,15 @@ func (s *schedulerTestSuite) TestScheduler(c *C) {
 	mustExec([]string{"-u", pdAddr, "scheduler", "config", "balance-hot-region-scheduler"}, &conf1)
 	c.Assert(conf1, DeepEquals, expected1)
 
+	// write-priorities is  divided into write-leader-priorities and write-peer-priorities
```
HunDunDM, line 313:
```suggestion
	// write-priorities is divided into write-leader-priorities and write-peer-priorities
```

HunDunDM (approved):
rest LGTM

Issue #3930, opened 2021-08-02
Pd replication mode can't recover from sync_recover after network recover in backup cluster 


mayjiang0203:
## Bug Report
When test dr-autosync feature, found pd stay in sync_recover more than 12 hours, and no process.
>>>12 hours ago.
curl http://172.16.4.67:31664/pd/api/v1/replication_mode/status
{
  "mode": "dr-auto-sync",
  "dr-auto-sync": {
    "label_key": "zone",
    "state": "sync_recover",
    "state_id": 477,
    "total_regions": 1,
    "synced_regions": 3,
    "recover_progress": 0.75
  }
}
>>>After 12 hours.
curl http://172.16.4.67:31664/pd/api/v1/replication_mode/status
{
  "mode": "dr-auto-sync",
  "dr-auto-sync": {
    "label_key": "zone",
    "state": "sync_recover",
    "state_id": 518,
    "total_regions": 1,
    "synced_regions": 2,
    "recover_progress": 0.6666667
  }
}

### What did you do?
Step 1. Create one table, and insert 1000 rows.
Step 2. Interrupt network of the backup cluster more than 1 min.
Step 3. Recover the network of the backup cluster.

### What did you expect to see?
Since there is only a few data in the database, recover should take several seconds or minutes instead of several hours.
### What did you see instead?
Seems recover blocked.
### What version of PD are you using (`pd-server -V`)?
/ # /pd-server -V
Release Version: v5.2.0-alpha-51-gfa0ada7f
Edition: Community
Git Commit Hash: fa0ada7fc4626865fd31be4c9490eb8062e91143
Git Branch: HEAD
UTC Build Time:  2021-07-31 06:35:33

nolouch:
I check the region information:
```
{
  "count": 1,
  "regions": [
    {
      "id": 42,
      "start_key": "",
      "end_key": "",
      "epoch": {
        "conf_ver": 53,
        "version": 72
      },
      "peers": [
        {
          "id": 422,
          "store_id": 1,
          "role_name": "Voter"
        },
        {
          "id": 470,
          "store_id": 6,
          "role_name": "Voter"
        },
        {
          "id": 476,
          "store_id": 5,
          "role_name": "Voter"
        }
      ],
      "leader": {
        "id": 422,
        "store_id": 1,
        "role_name": "Voter"
      },
      "written_bytes": 17676,
      "read_bytes": 570151,
      "written_keys": 30,
      "read_keys": 8461,
[...]

PR #3939, opened 2021-08-02
replication: fix the issue that the state cannot switch to sync when the region is merged

nolouch:
…he region is mergerd

Fix https://github.com/tikv/pd/issues/3930

Still check the region if the region contains the `drRecoverKey `

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix the issue that the state cannot switch to sync when the region is merged
```


nolouch:
PTAL @disksing @rleungx

server/replication/replication_mode.go
```diff
@@ -487,10 +487,11 @@ func (m *ModeManager) estimateProgress() float32 {
 }
 
 func (m *ModeManager) checkRegionRecover(region *core.RegionInfo, startKey []byte) bool {
-	if !bytes.Equal(startKey, region.GetStartKey()) {
+	// if the region not contains the key, log it and return false
+	if bytes.Compare(startKey, region.GetStartKey()) < 0 || (len(region.GetEndKey()) != 0 && bytes.Compare(startKey, region.GetEndKey()) > 0) {
```
disksing, line 491: I think no need to check endkey. Just `if bytes.Compare(startKey, region.GetStartKey()) < 0` should be ok.

disksing (approved):

bufferflies (approved):
lgtm

PR #3962, opened 2021-08-11
scheduler: use store leaders sum instead of hot leaders sum

HunDunDM:

scheduler: use store leaders sum instead of hot leaders sum

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


lhy1024 (commented):

JmPotato:

Issue #3839, opened 2021-07-06
Tracking issue for Region Label Feature


disksing:
Region Label is a system that runs inside PD and is used to identify different types of Regions, based on Region Label we can bind different behaviors to different regions.

For example, by setting labels for different tables of regions, we can control some tables from being merged.

In the first phase, we will support syncing part of the schema from TiDB to PD as region labels, and use region labels to control the behavior of region merge on top of that.

Sub tasks break up:

- [x] Introducing the Region Label framework & Implementing a LabelRule of type KeyRange
    - [x] https://github.com/tikv/pd/pull/3968
- [x] #3985
- [x] #4000
- [x] #4001
- [x] Support synchronization of table attributes CRUD operations from TiDB to PD
   - [x] https://github.com/pingcap/tidb/issues/26692
   - [x] https://github.com/pingcap/tidb/issues/26693
   - [x] https://github.com/pingcap/tidb/issues/26835
   - [x] https://github.com/pingcap/tidb/issues/27019
   - [x] https://github.com/pingcap/tidb/issues/27065
- [x] Support synchronization of partition attributes CRUD operations from TiDB to PD
   - [x] https://github.com/pingcap/tidb/issues/26704
   - [x] https://github.com/pingcap/tidb/issues/26794
   - [x] https://github.com/pingcap/tidb/issues/26836
   - [x] https://github.com/pingcap/tidb/issues/27101
- [x] Support showing the added attributes
   - [x] https://github.com/pingcap/tidb/issues/26694
- [x] Support DDL syntax for adding/resetting attributes
   - [x] https://github.com/pingcap/parser/issues/1288
- [x] Support rule override
   - [x] https://github.com/pingcap/tidb/issues/27784
   - [x] https://github.com/tikv/pd/issues/4031

tisonkun:
Hi @disksing , thanks for creating this tracking issue. Is there a design for this feature? I'd like to know how region label feature is integrated with region merge on top of that.

sunxiaoguang:
Could you please help me to understand why region rule better than placement rule? Is it because we want to configure rules at finer granularity? Thanks.

disksing:
> Hi @disksing , thanks for creating this tracking issue. Is there a design for this feature? I'd like to know how region label feature is integrated with region merge on top of that.

The detailed design is in Chinese so not linked. sorry :( 

Our plan is to support binding some attributes for table (or database or partition) in SQL, and these attributes will be synchronized to PD in the form of label.

[...]

PR #3968, opened 2021-08-12
labeler: add region label

disksing:

Add labeler implementation and support key-range type label rule.

https://github.com/tikv/pd/issues/3839

### Check List

Tests
- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/labeler/labeler.go
+++ b/server/schedule/labeler/labeler.go
@@ -1,0 +1,270 @@
+// Copyright 2021 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package labeler
+
+import (
+	"bytes"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sync"
+
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/core"
+	"github.com/tikv/pd/server/schedule/rangelist"
+	"go.uber.org/zap"
+)
+
+// RegionLabeler is utility to label regions.
+type RegionLabeler struct {
+	storage *core.Storage
+	sync.RWMutex
+	labelRules map[string]*LabelRule
+	rangeList  rangelist.List // sorted LabelRules of the type `KeyRange`
+}
+
+// NewRegionLabeler creates a Labeler instance.
+func NewRegionLabeler(storage *core.Storage) (*RegionLabeler, error) {
+	l := &RegionLabeler{
+		storage:    storage,
+		labelRules: make(map[string]*LabelRule),
+	}
+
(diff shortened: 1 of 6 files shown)


server/schedule/labeler/labeler.go
```diff
@@ -0,0 +1,270 @@
...
+			rules = append(rules, rule)
+		} else {
+			return nil, errs.ErrRegionRuleNotFound.FastGenByArgs(id)
+		}
+	}
+	return rules, nil
+}
+
+// GetLabelRule returns the Rule with the same ID.
+func (l *RegionLabeler) GetLabelRule(id string) *LabelRule {
```
rleungx, line 169: For now, the implementation of TiDB hasn't used `GetLabelRule` yet. Perhaps, it can be replaced by `GetLabelRules` with only one ID in the list.

server/schedule/labeler/labeler.go
```diff
@@ -0,0 +1,270 @@
...
+// GetLabelRules returns the rules that match the given ids.
+func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
+	l.RLock()
+	defer l.RUnlock()
+	rules := make([]*LabelRule, 0, len(ids))
+	for _, id := range ids {
+		if rule, ok := l.labelRules[id]; ok {
+			rules = append(rules, rule)
+		} else {
+			return nil, errs.ErrRegionRuleNotFound.FastGenByArgs(id)
```
rleungx, line 162: I am not sure that under this case, we should omit the invalid in or not.

disksing: which one do you prefer?

rleungx: Both are ok to me.

server/schedule/labeler/labeler.go
```diff
@@ -0,0 +1,270 @@
...
+			return
+		}
+		l.labelRules[r.ID] = &r
+	})
+	if err != nil {
+		return err
+	}
+	for _, d := range toDelete {
+		if err = l.storage.DeleteRegionRule(d); err != nil {
+			return err
```
JmPotato, line 73: Should we continue deleting the rest of `toDelete` rather than returning here?

disksing: If an error is encountered here, it is possible that the underlying etcd storage cannot be written to, and the next operation will most likely fail. I think it would make more sense to delay the deletion until the next load.

server/schedule/labeler/rules.go
```diff
@@ -0,0 +1,67 @@
...
+type KeyRangeRule struct {
+	StartKey    []byte `json:"-"`         // range start key
+	StartKeyHex string `json:"start_key"` // hex format start key, for marshal/unmarshal
+	EndKey      []byte `json:"-"`         // range end key
+	EndKeyHex   string `json:"end_key"`   // hex format end key, for marshal/unmarshal
+}
+
+// IsMatch returns if the region matches the rule.
+func (r *KeyRangeRule) IsMatch(region *core.RegionInfo) bool {
+	return bytes.Compare(region.GetStartKey(), r.StartKey) >= 0 && bytes.Compare(region.GetEndKey(), r.EndKey) <= 0
```
rleungx, line 60: Do we need to consider `""`?

disksing: you are right! In fact this method is not used. I'm going to just remove it.

rleungx (approved):

JmPotato (approved):

nolouch (approved):

PR #3979, opened 2021-08-12
scheduler: filter stores without hotspots

HunDunDM:

* Optimize the calculation of store count when summaryStoresLoad

* Filter stores without hot peers as a general solution
  * Misjudgment of scale-out nodes
  * Misjudgment of low flow nodes

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


lhy1024 (commented):
Maybe we need some tests.

JmPotato:

PR #3984, opened 2021-08-13
api: add region label api

disksing:

Add API for region label rules. Can review now but should be merged after https://github.com/tikv/pd/pull/3968
Review diff from [this commit](https://github.com/tikv/pd/pull/3984/commits/0ac942b3c77abf2f4c5f2f172396eb4138d6db63)

Fix https://github.com/tikv/pd/issues/3985

### Check List

Tests
- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/api/region_label.go
+++ b/server/api/region_label.go
@@ -52,6 +52,8 @@ func (h *regionLabelHandler) GetAllRules(w http.ResponseWriter, r *http.Request)
 // @Summary Update region label rules in batch.
 // @Produce json
 // @Success 200 {array} labeler.LabelRulePatch
+// @Failure 400 {string} string "The input is invalid."
+// @Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /config/region-label/rules [patch]
 func (h *regionLabelHandler) Patch(w http.ResponseWriter, r *http.Request) {
 	cluster := getCluster(r)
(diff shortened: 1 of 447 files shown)


server/api/region_label.go
```diff
@@ -0,0 +1,212 @@
...
+func (h *regionLabelHandler) GetAllRules(w http.ResponseWriter, r *http.Request) {
+	cluster := getCluster(r)
+	rules := cluster.GetRegionLabeler().GetAllLabelRules()
+	h.rd.JSON(w, http.StatusOK, rules)
+}
+
+// @Tags region_label
+// @Summary Update region label rules in batch.
+// @Produce json
+// @Success 200 {array} labeler.LabelRulePatch
```
rleungx, line 54: Also need 400 and 500

server/api/region_label.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	h.rd.JSON(w, http.StatusOK, "Update region label rules successfully.")
+}
+
+// @Tags region_label
+// @Summary Get label rules of cluster by ids.
+// @Param body body []string true "IDs of query rules"
+// @Produce json
+// @Success 200 {array} labeler.LabelRule
+// @Failure 404 {string} string "The rule does not exist."
```
rleungx, line 78: need 500

server/api/region_label.go
rleungx: BTW, `patch` doesn't return an array.

rleungx (approved):

JmPotato (approved):
LGTM. BTW, do we have any integration tests on these APIs with TiDB now?

PR #4002, opened 2021-08-18
schedule: add split checker for region splits

disksing:

Fix https://github.com/tikv/pd/issues/4000

Now that both placement rules and region labels need to split region based on key range, we can extract it to create a `SplitChecker` to handle region splits in the same place.

- add split checker
- remove region split logic from rule checker
- add test

### Check List

Tests
- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/rule_checker.go
+++ b/server/schedule/checker/rule_checker.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/metapb"
-	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/cache"
 	"github.com/tikv/pd/pkg/errs"
@@ -70,10 +69,8 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
 	c.record.refresh(c.cluster)
 
 	if len(fit.RuleFits) == 0 {
-		checkerCounter.WithLabelValues("rule_checker", "fix-range").Inc()
-		// If the region matches no rules, the most possible reason is it spans across
-		// multiple rules.
-		return c.fixRange(region)
+		checkerCounter.WithLabelValues("rule_checker", "need-split").Inc()
+		return nil
 	}
 	op, err := c.fixOrphanPeers(region, fit)
 	if err == nil && op != nil {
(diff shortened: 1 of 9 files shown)


server/schedule/checker/rule_checker.go
```diff
@@ -70,10 +69,10 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
 	c.record.refresh(c.cluster)
 
 	if len(fit.RuleFits) == 0 {
-		checkerCounter.WithLabelValues("rule_checker", "fix-range").Inc()
+		checkerCounter.WithLabelValues("rule_checker", "need-split").Inc()
```
JmPotato, line 72: Do we need to update the Grafana json also?

server/schedule/checker/split_checker.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// We may consider to merge labeler split keys and rule split keys together
+	// before creating operator. It can help to reduce operator count. However,
+	// handle them separately helps to understand the reason for the split.
+	desc := "labeler-split-region"
+	keys := c.labeler.GetSplitKeys(region.GetStartKey(), region.GetEndKey())
+
+	if len(keys) == 0 && c.cluster.GetOpts().IsPlacementRulesEnabled() {
+		desc = "rule-split-region"
+		keys = c.ruleManager.GetSplitKeys(region.GetStartKey(), region.GetEndKey())
```
JmPotato, line 61: What about creating two variables `startKey` and `endKey` here rather than calling the functions every time.

server/schedule/checker/rule_checker.go
disksing: no need to update. it is just an arbitrary label.

server/schedule/checker/split_checker.go
disksing: good idea.

server/schedule/checker_controller.go
```diff
@@ -70,9 +73,13 @@ func (c *CheckerController) CheckRegion(region *core.RegionInfo) []*operator.Ope
...
 		return []*operator.Operator{op}
 	}
-	fit := c.priorityChecker.Check(region)
+
+	if op := c.splitChecker.Check(region); op != nil {
+		return []*operator.Operator{op}
+	}
 
 	if c.opts.IsPlacementRulesEnabled() {
+		fit := c.priorityChecker.Check(region)
```
rleungx, line 82: Why move it here?

JmPotato: Only line#83 uses it.

JmPotato (approved):
LGTM

rleungx (approved):

PR #4007, opened 2021-08-19
scheduler: dynamically adjust the retry limit according to the operator

HunDunDM:

related #3744 

* scheduler: dynamically adjust the retry limit according to the operator

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


nolouch (approved):

server/schedulers/utils.go
```diff
@@ -746,3 +748,53 @@ func filterHotPeers(kind core.ResourceKind, peers []*statistics.HotPeerStat) []*
...
+func (q *retryQuota) Attenuate(store *core.StoreInfo) {
+	newLimit := q.GetLimit(store) / q.attenuation
+	if newLimit < q.minLimit {
+		newLimit = q.minLimit
+	}
+	q.limits[store.GetID()] = newLimit
+}
+
+func (q *retryQuota) GC(keepStores []*core.StoreInfo) {
+	set := make(map[uint64]struct{}, len(keepStores))
```
disksing, line 791: I think it would be better to avoid allocate map here, simply:
```
for _, store := range keepStores {
    delete(q.limits, id)
}
```

server/schedulers/balance_leader.go
```diff
@@ -154,40 +156,46 @@ func (l *balanceLeaderScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
...
+					l.retryQuota.ResetLimit(plan.target)
 					ops[0].Counters = append(ops[0].Counters, l.counter.WithLabelValues("transfer-in", plan.TargetMetricLabel()))
 					return ops
 				}
 			}
+			l.Attenuate(plan.target)
 			log.Debug("no operator created for selected stores", zap.String("scheduler", l.GetName()), zap.Uint64("target", plan.TargetStoreID()))
 		}
 	}
+	l.retryQuota.GC(append(sources, targets...))
```
disksing, line 198: maybe better to avoid reallocate:
```
l.retryQuota.GC(sources)
l.retryQuota.GC(targets)
```

disksing (approved):

server/schedulers/balance_region.go
```diff
@@ -188,11 +191,14 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 			}
 
 			if op := s.transferPeer(plan); op != nil {
+				s.retryQuota.ResetLimit(plan.source)
```
bufferflies, line 194: it may be better that the limit  recover same with the failed?

server/schedulers/utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+	store2 := core.NewStoreInfo(&metapb.Store{Id: 2})
+	keepStores := []*core.StoreInfo{store1}
+
+	// test GetLimit
+	c.Assert(q.GetLimit(store1), Equals, 10)
+
+	// test Attenuate
+	q.Attenuate(store1)
+	c.Assert(q.GetLimit(store1), Equals, 5)
+	for i := 0; i < 10; i++ {
```
rleungx, line 38: no need to loop 10 times

HunDunDM: To test `minLimit`

rleungx: Since the attenuation is 2, you don't need to loop 10 times

PR #4017, opened 2021-08-20
Feature/pprof all

bufferflies:
relate: # 4013

put pprof ,goroutine ,mutex,allocs,config,version to zip file

image
 
### Check List

Tests

Code changes

Side effects

Related changes

### Release note

```release-note
add debug/zip http api to zip file that debug needed
```

### DIFF
--- a/server/api/pprof.go
+++ b/server/api/pprof.go
@@ -47,7 +47,7 @@ func newProfHandler(svr *server.Server, rd *render.Render) *ProfHandler {
 }
 
 func (h *ProfHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="tidb_debug"`+time.Now().Format("20060102150405")+".zip"))
+	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="pd_debug"`+time.Now().Format("20060102_150405")+".zip"))
 
 	// dump goroutine/heap/mutex
 	items := []struct {
@@ -59,6 +59,7 @@ func (h *ProfHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		{name: "goroutine", debug: 2},
 		{name: "heap", gc: 1},
 		{name: "mutex"},
+		{name: "allocs"},
 	}
 	zw := zip.NewWriter(w)
 	for _, item := range items {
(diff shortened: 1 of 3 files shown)


nolouch (approved):
lgtm

server/api/router.go
```diff
@@ -251,6 +251,9 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	apiRouter.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
 	apiRouter.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
 
+	profHandler := newProfHandler(svr, rd)
+	apiRouter.Handle("/debug/zip", profHandler)
```
disksing, line 255: could `/debug/pprof/zip` better?

bufferflies: yes

server/api/router.go
```diff
@@ -241,6 +241,7 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	apiRouter.HandleFunc("/tso/allocator/transfer/{name}", tsoHandler.TransferLocalTSOAllocator).Methods("POST")
 
 	// profile API
+	profHandler := newProfHandler(svr, rd)
```
JmPotato, line 244: What about moving this line to L#254?

bufferflies: ok, it will set as param

server/api/pprof.go
```diff
@@ -0,0 +1,152 @@
...
+	return &ProfHandler{
+		svr: svr,
+		rd:  rd,
+	}
+
+}
+
+// @Summary debug zip of PD servers.
+// @Produce application/octet-stream
+// @Router /pprof/debug/zip [get]
```
disksing, line 52: wrong path

disksing (approved):

server/api/pprof.go
```diff
@@ -0,0 +1,152 @@
...
+func (h *ProfHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="pd_debug"`+time.Now().Format("20060102_150405")+".zip"))
```
rleungx, line 54: The format is a little bit different with TiDB.

server/api/pprof.go
```diff
@@ -0,0 +1,152 @@
...
+		{name: "goroutine", debug: 2},
+		{name: "heap", gc: 1},
+		{name: "mutex"},
+		{name: "allocs"},
+	}
+	zw := zip.NewWriter(w)
+	for _, item := range items {
+		p := pprof.Lookup(item.name)
+		if p == nil {
+			h.rd.JSON(w, http.StatusNotFound, fmt.Sprintf("pprof can not find name:%s", item.name))
```
rleungx, line 72:
```suggestion
			h.rd.JSON(w, http.StatusNotFound, fmt.Sprintf("pprof can not find name: %s", item.name))
```

server/api/pprof.go
bufferflies: yes, it will be better to splite date and time

PR #4024, opened 2021-08-23
placement: add some benchmark cases for FitRegion

Yisaer:

add some benchmark cases for FitRegion

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/placement/fit_region_test.go
+++ b/server/schedule/placement/fit_region_test.go
@@ -1,0 +1,330 @@
+// Copyright 2021 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


server/schedule/placement/fit_region_test.go
```diff
@@ -0,0 +1,329 @@
...
+		FitRegion(storesSet, region, rules)
+	}
+}
+
+func BenchmarkFitRegionMoreVotersSplitRules(b *testing.B) {
+	i := 0
+	region := mockRegion(5, 0)
+	rules := []*Rule{
+		{
+			GroupID:        "pd",
```
rleungx, line 245: Is it on purpose to use the same rule?

Yisaer: yes, this is used to comparing the cpu resource consuming between less rules and more rules.

rleungx: Then how about using a variable?

Yisaer: updated.

server/schedule/placement/fit_region_test.go
```diff
@@ -0,0 +1,288 @@
...
+			ID:             "default",
+			Role:           Voter,
+			Count:          3,
+			LocationLabels: []string{},
+		},
+	}
+	storesSet := newMockStoresSet(100)
+	b.ResetTimer()
+	for i < b.N {
```
rleungx, line 119: use `for i := 0; i < b.N; i++`

Issue #4025, opened 2021-08-24
Add cache layer for placement.FitRegion


Yisaer:
## Development Task

`placement.FitRegion` will consume lots of cpu resources, which would be the scheduling bottleneck in a large cluster with tiflash. This issue introducing cache layer for the `placement.FitRegion` If the calculating variables are not changed, and the region is already in a stable status, we could cache the result in cache layer after calculating by `FitRegion` and use the cache result in next time.

1. [x] introducing `version` in rule, version will be added each time rule updated.
2. [x] introducing `cache layer` for FitRegion
3. [x]  avoding re-calculating if we get cached result from FitRegion

rleungx:
I think we have done this job? @Yisaer

rleungx:
Close it since it has been finished.

PR #4026, opened 2021-08-24
placement: introduing rule version

Yisaer:

refer https://github.com/tikv/pd/issues/4025

Introducing version for rule in order to judge whether the is updated from last checking.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/placement/rule.go
+++ b/server/schedule/placement/rule.go
@@ -81,8 +81,6 @@ func (r *Rule) Clone() *Rule {
 	json.Unmarshal([]byte(r.String()), &clone)
 	clone.StartKey = append(r.StartKey[:0:0], r.StartKey...)
 	clone.EndKey = append(r.EndKey[:0:0], r.EndKey...)
-	clone.Version = r.Version
-	clone.CreateTimestamp = r.CreateTimestamp
 	return &clone
 }
 
--- a/server/schedule/placement/rule.go
+++ b/server/schedule/placement/rule.go
@@ -65,9 +65,9 @@ type Rule struct {
 	LabelConstraints []LabelConstraint `json:"label_constraints,omitempty"` // used to select stores to place peers
 	LocationLabels   []string          `json:"location_labels,omitempty"`   // used to make peers isolated physically
 	IsolationLevel   string            `json:"isolation_level,omitempty"`   // used to isolate replicas explicitly and forcibly
-
-	version uint64     // only set at runtime, add 1 each time rules updated, begin from 0.
-	group   *RuleGroup // only set at runtime, no need to {,un}marshal or persist.
+	Version          uint64            `json:"version,omitempty"`           // only set at runtime, add 1 each time rules updated, begin from 0.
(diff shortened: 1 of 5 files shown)


server/schedule/placement/rule.go
```diff
@@ -66,7 +66,8 @@ type Rule struct {
 	LocationLabels   []string          `json:"location_labels,omitempty"`   // used to make peers isolated physically
 	IsolationLevel   string            `json:"isolation_level,omitempty"`   // used to isolate replicas explicitly and forcibly
 
-	group *RuleGroup // only set at runtime, no need to {,un}marshal or persist.
+	version uint64     // only set at runtime, add 1 each time rules updated, begin from 0.
```
disksing, line 69: I think we can consider making it persistent and public. Clients may benefit from it to avoid some checks.

Yisaer: updated.

server/schedule/placement/rule.go
```diff
@@ -65,8 +65,9 @@ type Rule struct {
 	LabelConstraints []LabelConstraint `json:"label_constraints,omitempty"` // used to select stores to place peers
 	LocationLabels   []string          `json:"location_labels,omitempty"`   // used to make peers isolated physically
 	IsolationLevel   string            `json:"isolation_level,omitempty"`   // used to isolate replicas explicitly and forcibly
-
-	group *RuleGroup // only set at runtime, no need to {,un}marshal or persist.
+	Version          uint64            `json:"version,omitempty"`           // only set at runtime, add 1 each time rules updated, begin from 0.
+	CreateTimestamp  uint64            `json:"create_timestamp,omitempty"`  // only set at runtime, recorded rule create timestamp
```
rleungx, line 69: What does it used for?

JmPotato (approved):

server/schedule/placement/rule.go
```diff
@@ -80,6 +81,8 @@ func (r *Rule) Clone() *Rule {
 	json.Unmarshal([]byte(r.String()), &clone)
 	clone.StartKey = append(r.StartKey[:0:0], r.StartKey...)
 	clone.EndKey = append(r.EndKey[:0:0], r.EndKey...)
+	clone.Version = r.Version
+	clone.CreateTimestamp = r.CreateTimestamp
```
disksing, line 85: can remove. Unmarshal already handles them.

server/schedule/placement/rule.go
Yisaer: we use `createTimestamp` and `version` to compare whether rule is updated. This will avoid thinking rule unchanged after it created,deleted,re-created.

server/schedule/placement/rule.go
Yisaer: updated.

server/schedule/placement/rule.go
rleungx: Seem `createTimestamp` is not used to do the comparison?

Yisaer: we have not introduced comparison yet.

rleungx (approved):

PR #4030, opened 2021-08-25
placement: support RegionFitCache for FitRegion

Yisaer:

refer https://github.com/tikv/pd/issues/4025  FitRegion consume much cpu resources while most regions are not necessary to be recalculated when it gets stable.

Support cache layer for FitRegion

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -15,6 +15,7 @@
 package core
 
 import (
+	"github.com/tikv/pd/pkg/slice"
 	"math"
 	"strings"
 	"time"
@@ -706,21 +707,13 @@ func IsTiFlashStore(store *metapb.Store) bool {
 
 // IsEqualLabels returns whether have same store label
 func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
-	if len(s.GetLabels()) != len(labels) {
+	storeLabels := s.GetLabels()
+	if len(storeLabels) != len(labels) {
 		return false
 	}
-	storeLabels := s.GetLabels()
-	for _, label := range labels {
-		find := false
-		for _, slabel := range storeLabels {
-			if label.Key == slabel.Key && label.Value == slabel.Value {
-				find = true
-				break
-			}
-		}
-		if !find {
-			return false
-		}
-	}
-	return true
+	return slice.AllOf(labels, func(i int) bool {
+		return slice.AnyOf(storeLabels, func(j int) bool {
+			return labels[i].Key == storeLabels[j].Key && labels[i].Value == storeLabels[j].Value
+		})
+	})
 }
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -15,7 +15,6 @@
 package core
 
 import (
-	"github.com/tikv/pd/pkg/slice"
 	"math"
 	"strings"
 	"time"
@@ -24,6 +23,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/slice"
 	"github.com/tikv/pd/server/core/storelimit"
 	"go.uber.org/zap"
 )
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -703,3 +703,21 @@ func IsTiFlashStore(store *metapb.Store) bool {
 	}
 	return false
 }
+
+func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
+	if len(s.GetLabels()) != len(labels) {
+		return false
+	}
+	for _, label := range labels {
+		find := false
+		for _, slabel := range s.GetLabels() {
+			if label.Key == slabel.Key && label.Value == slabel.Value {
+				find = true
+			}
+		}
+		if find {
+			return false
+		}
+	}
+	return true
+}
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -714,9 +714,10 @@ func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
 		for _, slabel := range s.GetLabels() {
 			if label.Key == slabel.Key && label.Value == slabel.Value {
 				find = true
+				break
 			}
 		}
-		if find {
+		if !find {
 			return false
 		}
 	}
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -709,9 +709,10 @@ func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
 	if len(s.GetLabels()) != len(labels) {
 		return false
 	}
+	storeLabels := s.GetLabels()
 	for _, label := range labels {
 		find := false
-		for _, slabel := range s.GetLabels() {
+		for _, slabel := range storeLabels {
 			if label.Key == slabel.Key && label.Value == slabel.Value {
 				find = true
 				break
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -704,6 +704,7 @@ func IsTiFlashStore(store *metapb.Store) bool {
 	return false
 }
 
+// IsEqualLabels returns whether have same store label
 func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
 	if len(s.GetLabels()) != len(labels) {
 		return false
--- a/server/schedule/placement/region_rule_cache.go
+++ b/server/schedule/placement/region_rule_cache.go
@@ -1,0 +1,107 @@
+package placement
+
+import (
+	"sync"
+
+	"github.com/tikv/pd/server/core"
+)
+
+type RegionRuleFitCacheManager struct {
+	mu     sync.RWMutex
+	caches map[uint64]*RegionRuleFitCache
+}
+
+func NewRegionRuleFitCacheManager() *RegionRuleFitCacheManager {
+	return &RegionRuleFitCacheManager{
+		caches: map[uint64]*RegionRuleFitCache{},
+	}
+}
+
+func (manager *RegionRuleFitCacheManager) GetCacheRegionFit(regionID uint64) *RegionFit {
+	manager.mu.RLock()
+	defer manager.mu.RUnlock()
+	return manager.caches[regionID].bestFit
+}
+
+func (manager *RegionRuleFitCacheManager) InValid(regionID uint64) {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	delete(manager.caches, regionID)
+}
+
+func (manager *RegionRuleFitCacheManager) Check(region *core.RegionInfo, rules []*Rule) bool {
+	manager.mu.RLock()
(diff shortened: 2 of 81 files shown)


server/core/store.go
```diff
@@ -703,3 +703,23 @@ func IsTiFlashStore(store *metapb.Store) bool {
...
 }
+
+// IsEqualLabels returns whether have same store label
+func (s *StoreInfo) IsEqualLabels(labels []*metapb.StoreLabel) bool {
+	if len(s.GetLabels()) != len(labels) {
+		return false
+	}
+	for _, label := range labels {
+		find := false
+		for _, slabel := range s.GetLabels() {
```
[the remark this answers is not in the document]
Yisaer: updated.

rleungx (approved):

server/core/store.go
```diff
@@ -703,3 +703,24 @@ func IsTiFlashStore(store *metapb.Store) bool {
...
+	}
+	storeLabels := s.GetLabels()
+	for _, label := range labels {
+		find := false
+		for _, slabel := range storeLabels {
+			if label.Key == slabel.Key && label.Value == slabel.Value {
+				find = true
+				break
+			}
+		}
```
disksing, line 720: hard to read. maybe you can extract a new function or use `slice.AllOf`/`slice.NoneOf` to simplify.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+// 2. RegionFit is satisfied
+// RegionRuleFitCacheManager will invalid the cache for the region only when:
+// 1. region peer topology is changed
+// 2. region have down peers
+// 3. region leader is changed
+// 4. any involved rule is changed
+// 5. stores topology is changed
+// 6. any store label is changed
+// 7. any store state is changed
+type RegionRuleFitCacheManager struct {
```
disksing, line 36: How about `RegionRuleFitCache`?

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+	if len(a) != len(b) {
+		return false
+	}
+	for _, arule := range a {
+		find := false
+		for _, brule := range b {
+			if arule.ID == brule.ID &&
+				arule.GroupID == brule.GroupID &&
+				arule.Version == brule.Version &&
+				arule.CreateTimestamp == brule.CreateTimestamp {
```
disksing, line 147: We can extract it to a new method of `Rule`.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+		if !find {
+			return false
+		}
+	}
+	return true
+}
+
+func isRulesEqual(a, b []*Rule) bool {
```
disksing, line 137: how about `rulesEqual`? `rules` is plural, should not use `is`.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+		return false
+	}
+	if !(isPeersEqual(cache.region, region) &&
+		region.GetLeader().StoreId == cache.region.GetLeader().StoreId) {
+		return false
+	}
+	return true
+}
+
+func isPeersEqual(a, b *core.RegionInfo) bool {
```
disksing, line 118: how about `peersEqual`.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+func (cache *RegionRuleFitCache) isRegionUnchanged(region *core.RegionInfo) bool {
+	// we only cache region when it doesn't have down peers
+	if len(region.GetDownPeers()) > 0 || region.GetLeader() == nil {
+		return false
+	}
+	if !(isPeersEqual(cache.region, region) &&
+		region.GetLeader().StoreId == cache.region.GetLeader().StoreId) {
+		return false
+	}
+	return true
```
disksing, line 115:
```suggestion
	return isPeersEqual(cache.region, region) && region.GetLeader().StoreId == cache.region.GetLeader().StoreId
```

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+	region  *core.RegionInfo
+	rules   []*Rule
+}
+
+// IsUnchanged checks whether the region and rules unchanged for the cache
+func (cache *RegionRuleFitCache) IsUnchanged(region *core.RegionInfo, rules []*Rule) bool {
+	return cache.isRegionUnchanged(region) && cache.isRulesUnchanged(rules)
+}
+
+func (cache *RegionRuleFitCache) isRulesUnchanged(rules []*Rule) bool {
```
disksing, line 102: no need this method. We can use `isRulesEqual` directly.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,157 @@
...
+// 1. region peer topology is changed
+// 2. region have down peers
+// 3. region leader is changed
+// 4. any involved rule is changed
+// 5. stores topology is changed
+// 6. any store label is changed
+// 7. any store state is changed
+type RegionRuleFitCacheManager struct {
+	mu     sync.RWMutex
+	caches map[uint64]*RegionRuleFitCache
```
disksing, line 38: Will it cost too much memory to cache all region's fit result?

server/core/store.go
Yisaer: updated.

server/schedule/placement/region_rule_cache.go
Yisaer: we already have `RegionRuleFitCache`

server/schedule/placement/region_rule_cache.go
Yisaer: updated.

server/schedule/placement/region_rule_cache.go
Yisaer: updated.

server/schedule/placement/region_rule_cache.go
Yisaer: updated.

server/schedule/placement/region_rule_cache.go
Yisaer: I think it won't. each `RegionRuleFitCache` stored pointer for `rule`/`peer` which are already in memory. Thus, it won't cost too much memory here.

server/schedule/placement/rule_manager.go
```diff
@@ -305,9 +308,39 @@ func (m *RuleManager) GetRulesForApplyRegion(region *core.RegionInfo) []*Rule {
 }
 
 // FitRegion fits a region to the rules it matches.
-func (m *RuleManager) FitRegion(stores StoreSet, region *core.RegionInfo) *RegionFit {
+func (m *RuleManager) FitRegion(storeSet StoreSet, region *core.RegionInfo) *RegionFit {
+	stores := storeSet.GetStores()
```
disksing, line 312: Does it get all stores in the cluster? It looks we only need stores that contain the region's peers.

server/core/store.go
```diff
@@ -15,6 +15,7 @@
 package core
 
 import (
+	"github.com/tikv/pd/pkg/slice"
```
disksing, line 18: looks not in good order.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,150 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package placement
+
+import (
+	"github.com/tikv/pd/pkg/slice"
```
disksing, line 18: not in good order.

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (cache *RegionRuleFitCache) isRegionUnchanged(region *core.RegionInfo) bool {
+	// we only cache region when it doesn't have down peers
+	if len(region.GetDownPeers()) > 0 || region.GetLeader() == nil {
+		return false
+	}
+	return region.GetLeader().StoreId == cache.region.GetLeader().StoreId && peersEqual(cache.region, region)
+}
+
+func peersEqual(a, b *core.RegionInfo) bool {
```
disksing, line 113: can we replace it by checking region's confVersion?

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,150 @@
...
+		return false
+	}
+	return slice.AllOf(a, func(i int) bool {
+		return slice.AnyOf(b, func(j int) bool {
+			return equalRules(a[i], b[j])
+		})
+	})
+}
+
+func storeSEqual(a, b []*core.StoreInfo) bool {
```
disksing, line 139: should be `storesEqual`

server/schedule/placement/region_rule_cache.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(a) != len(b) {
+		return false
+	}
+	return slice.AnyOf(a, func(i int) bool {
```
disksing, line 143: should it be `AllOf`?

server/schedule/placement/region_rule_cache.go
Yisaer: updated.

server/schedule/placement/rule_manager.go
Yisaer: good cactch, updated.

disksing: I think you need to make `getStoresByRegion` to accept `storeSet` directly to avoid getting all stores here.

server/schedule/placement/rule_manager.go
```diff
@@ -305,9 +307,24 @@ func (m *RuleManager) GetRulesForApplyRegion(region *core.RegionInfo) []*Rule {
...
+			return fit
+		}
+	}
+	fit := FitRegion(stores, region, rules)
+	if fit.IsSatisfied() && len(region.GetDownPeers()) == 0 && region.GetLeader() != nil {
+		m.cache.SetCache(region, rules, fit, getStoresByRegion(stores, region))
+	} else {
+		m.cache.Invalid(region.GetID())
+	}
+	fit.SetCached(false)
```
disksing, line 326: I'm not quite understand what `Cached` is used for.

Yisaer: In future pr, if the `checker` find that the fit is cached, it will skip the current region instead of generating operator.

disksing: Then I guess we should not set the flag to `true` at L316. Instead, the flag should only be set to `true` by checker when it cannot create any operator to improve the region. Also, I wonder if `checked` is more appropriate than `cached`.

disksing (approved):

PR #4044, opened 2021-08-27
tso, election: remove some unnecessary logs from the TSO hot path

JmPotato:

Remove some unnecessary logs from the TSO hot path.

```go
log.Warn("get timestamp too slow", zap.Duration("cost", elapsed))
```

This indicator can be observed through metrics, so it is not necessary to hit the log here.

```go
log.Info("this PD is a new leader, but sync hasn't been completed yet, wait for a while")
```

In past experience, this log has not been very helpful and may be very noisy.

```go
log.Error("invalid timestamp",
				zap.Any("timestamp-physical", currentPhysical),
				zap.Any("timestamp-logical", currentLogical),
				errs.ZapError(errs.ErrInvalidTimestamp))
```

Ditto.

Remove some unnecessary logs from the TSO hot path.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/errors.toml
+++ b/errors.toml
@@ -611,11 +611,6 @@ error = '''
 get local allocator failed, %s
 '''
 
-["PD:tso:ErrInvalidTimestamp"]
-error = '''
-invalid timestamp
-'''
-
 ["PD:tso:ErrLogicOverflow"]
 error = '''
 logic part overflow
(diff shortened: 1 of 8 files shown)


nolouch (approved):
lgtm

rleungx (approved):

Issue #3149, opened 2020-11-04
Improve the Scalability of TSO Service


ZenoTan:
## Problem Description

When we have multiple PD clients in a cluster, especially in the cases without TiDB, CPU utilization of PD will be high since there will be many connections to the PD leader to request for TSO.

Here is a testing result about the relationship among the number of clients, the concurrency of each client, and the TSO QPS:
<img src="https://user-images.githubusercontent.com/41138939/98064630-b1f86480-1e8d-11eb-89fb-7a485ce1cd93.png" width="80%" height="80%">

## Design

RFC: https://github.com/tikv/rfcs/pull/78

## RoadMap
More details can be found in [PD Scalability (CPU Bound): TSO, Balance](https://docs.google.com/document/d/19wOAS-8mG3Izb073NYz15LK2_EoCkUk9UwKrAf2RIaU)

PR #4054, opened 2021-08-30
tso: add tsoBatchController to control the TSO client batch interval

JmPotato:

Part of https://github.com/tikv/pd/issues/3149.

When we have many TiDB instances (or PD clients) requesting TSO concurrently, the PD leader (without enabling Local TSO) may suffer from very high CPU utilization due to the bad batch size. To solve this, one easy and possible way is to improve the batch size by waiting for a while when we batch the TSO requests at the side of each PD client. This PR implements this optimization.

Introduce `tsoBatchController` to control the TSO client batch interval.

### Check List

Tests

- Unit test
- Integration test

Manual test results will be updated later.

### Release note

```release-note
Improve the batch effect of TSO requests for the PD client.
```


client/client.go
```diff
@@ -156,10 +156,123 @@ type tsoRequest struct {
...
+		return false
+	case firstTSORequest = <-tbc.tsoRequestCh:
+	}
+	// Start to batch when the first TSO request arrives.
+	tbc.batchStartTime = time.Now()
+	tbc.collectedRequestCount = 0
+	tbc.pushRequest(firstTSORequest)
+
+	// This loop is for trying best to collect more requests, so we use `tbc.maxBatchSize` here.
+	for tbc.collectedRequestCount < tbc.maxBatchSize {
```
rleungx, line 198: Will it affect the latency since we don't have a time limit here.

rleungx: BTW, do we need the handle the context in this for loop?

client/client.go
```diff
@@ -156,10 +156,123 @@ type tsoRequest struct {
...
+			// No more pending TSO requests, just return.
+			if tsoReq == nil {
+				return
+			}
+			tbc.pushRequest(tsoReq)
+		case <-after.C:
+			return
+		}
+	}
+	after.Stop()
```
rleungx, line 240: The timer will not stop if it returns early.

disksing: +1

client/client.go
JmPotato: I think it won't affect the latency since the `maxBatchSize` is not that large.

client/client.go
```diff
@@ -156,10 +156,123 @@ type tsoRequest struct {
...
+	tbc.batchStartTime = time.Now()
+	tbc.collectedRequestCount = 0
+	tbc.pushRequest(firstTSORequest)
+
+	// This loop is for trying best to collect more requests, so we use `tbc.maxBatchSize` here.
+	for tbc.collectedRequestCount < tbc.maxBatchSize {
+		select {
+		case tsoReq := <-tbc.tsoRequestCh:
+			// No more pending TSO requests, just return.
+			if tsoReq == nil {
```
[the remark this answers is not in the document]
JmPotato: Actually, only when the `tsoRequestCh` is closed.

client/client.go
```diff
@@ -156,10 +156,136 @@ type tsoRequest struct {
...
+		// Waits too long to collect requests, reduce the target batch size.
+		tbc.bestBatchSize--
+	} else if length > tbc.bestBatchSize+4 /* Hard-coded number, in order to make `tbc.bestBatchSize` stable */ &&
+		tbc.bestBatchSize < tbc.maxBatchSize {
+		tbc.bestBatchSize++
+	}
+}
+
+func (tbc *tsoBatchController) observeBatchSendLatency() {
+	tsoBatchSendLatency.Observe(float64(time.Since(tbc.batchStartTime)))
```
rleungx, line 282: Why not directly use `tsoBatchSendLatency.Observe(float64(time.Since(tbc.batchStartTime)))`?

rleungx (approved):

client/base_client.go
```diff
@@ -55,10 +61,12 @@ type baseClient struct {
...
-	gRPCDialOptions  []grpc.DialOption
-	timeout          time.Duration
-	maxRetryTimes    int
-	enableForwarding bool
+	gRPCDialOptions         []grpc.DialOption
+	timeout                 time.Duration
+	maxRetryTimes           int
+	enableForwarding        bool
+	maxTSOBatchSize         int
+	maxTSOBatchWaitInterval time.Duration
```
HunDunDM, line 69: Do they need to be placed in `baseClient`?

JmPotato: I will file another PR to refactor this part.

client/client.go
```diff
@@ -156,10 +156,120 @@ type tsoRequest struct {
...
+				return true
+			}
+		}
+	}
+
+	// Do an additional non-block try. Here we test the length with `tbc.maxBatchSize` instead
+	// of `tbc.bestBatchSize` because trying best to fetch more requests is necessary so that
+	// we can adjust the `tbc.bestBatchSize` dynamically later.
+	for tbc.collectedRequestCount < tbc.maxBatchSize {
+		if len(tbc.tsoRequestCh) == 0 {
```
HunDunDM, line 238: ditto

client/client.go
HunDunDM: If the `request` can be taken out all the time, it will be taken up to `maxBatchSize`. Otherwise, it will `break`. I don't think it will consume a lot of time. Of course, we can set a maximum value for `maxBatchSize`.

client/client.go
```diff
@@ -232,20 +341,20 @@ func NewClientWithContext(ctx context.Context, pdAddrs []string, security Securi
...
 			dispatcher, _ := c.tsoDispatcher.Load(dcLocation)
 			dispatcherCtx := dispatcher.(*tsoDispatcher).dispatcherCtx
-			tsoRequestCh := dispatcher.(*tsoDispatcher).tsoRequestCh
+			tbc := dispatcher.(*tsoDispatcher).tsoBatchController
 			// Each goroutine is responsible for handling the tso stream request for its dc-location.
 			// The only case that will make the dispatcher goroutine exit
 			// is that the loopCtx is done, otherwise there is no circumstance
 			// this goroutine should exit.
-			go c.handleDispatcher(dispatcherCtx, dcLocation, tsoRequestCh)
+			go c.handleDispatcher(dispatcherCtx, dcLocation, tbc)
```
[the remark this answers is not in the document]
JmPotato: Nice catch.

client/client.go
```diff
@@ -156,10 +156,120 @@ type tsoRequest struct {
...
+	// This loop is for trying best to collect more requests, so we use `tbc.maxBatchSize` here.
+	for tbc.collectedRequestCount < tbc.maxBatchSize {
+		select {
+		case tsoReq := <-tbc.tsoRequestCh:
+			tbc.pushRequest(tsoReq)
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			// break here can only break the select statement.
+			goto fetchMorePendingRequests
```
HunDunDM, line 206: use `break fetch` to break out of multi loop

HunDunDM:
rest LGTM. Maybe we also need to enable `bestBatchSize` only when the PD load is high.

client/client.go
JmPotato: I think this is no different from `goto XXX`.

HunDunDM: When the code changes later, `goto` is more likely to cause problems that are not easy to find.

client/base_client.go
JmPotato: PTAL https://github.com/tikv/pd/pull/4104.

client/client.go
```diff
@@ -547,46 +651,41 @@ func (c *client) handleDispatcher(dispatcherCtx context.Context, dc string, tsoD
 				continue
 			}
 		}
+		if err = tbc.fetchPendingRequests(dispatcherCtx); err != nil {
+			log.Error("[pd] fetch pending tso requests error", zap.String("dc-location", dc), errs.ZapError(errs.ErrClientGetTSO, err))
+			return
+		}
+		tbc.adjustBestBatchSize()
```
nolouch, line 658: should consider the `maxBatchWaitInterval` = 0. when disable batch strategy, it should same as before.

client/base_client.go
```diff
@@ -99,18 +106,26 @@ func WithMaxErrorRetry(count int) ClientOption {
...
-		maxRetryTimes:        maxInitClusterRetries,
+		urls:                    urls,
+		checkLeaderCh:           make(chan struct{}, 1),
+		checkTSODispatcherCh:    make(chan struct{}, 1),
+		ctx:                     ctx1,
+		cancel:                  cancel,
+		security:                security,
+		timeout:                 defaultPDTimeout,
+		maxRetryTimes:           maxInitClusterRetries,
+		maxTSOBatchWaitInterval: defaultMaxTSOBatchInterval,
```
nolouch, line 128: I think it should disable by default, the latency is slightly increased if enabled.

client/client.go
```diff
@@ -710,6 +812,7 @@ func (c *client) processTSORequests(stream pdpb.PD_TsoClient, dcLocation string,
 		c.finishTSORequest(requests, 0, 0, 0, err)
 		return err
 	}
+	tsoBatchSendLatency.Observe(float64(time.Since(tbc.batchStartTime)))
```
nolouch, line 815: Is it differnt with `requestDurationTSO`?

JmPotato: `requestDurationTSO` records the elapsed time between RPC sending and receiving. `tsoBatchSendLatency` records the elapsed time between the first TSO request arrives in a batch and the RPC sending.

nolouch (commented):
I think we need dynamic change the config, also, we need explainable configuration suggestions.

JmPotato:
> I think we need dynamic change the config, also, we need explainable configuration suggestions.

I think we can do this in another independent PR.

PR #4060, opened 2021-09-01
checker: skip generating operator if fit is cached for rule checker

Yisaer:

If the fit is fetched from cache, the rule checker will skip checking instead of generating operator.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/rule_checker.go
+++ b/server/schedule/checker/rule_checker.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/cache"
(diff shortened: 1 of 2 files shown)


server/schedule/checker/rule_checker.go
```diff
@@ -65,6 +66,16 @@ func (c *RuleChecker) Check(region *core.RegionInfo) *operator.Operator {
...
 // CheckWithFit checkWithFit is similar with Checker with placement.RegionFit
 func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.RegionFit) *operator.Operator {
+	// If the fit is fetched from cache, it seems that the region doesn't need cache
+	if fit.IsCached() {
+		checkerCounter.WithLabelValues("rule_checker", "cache").Inc()
+		return nil
+	}
+
+	failpoint.Inject("assertCache", func() {
+		panic("shouldn't be here")
```
disksing, line 76: maybe `RegionFit is not cached`

Yisaer: updated.

disksing (approved):

PR #4063, opened 2021-09-02
server, pkg: introduce CPUCollector to collect the CPU usage

JmPotato:

Part of #3149. See https://github.com/tikv/pd/pull/4054 also.

While adjusting the batch size to improve the batch effect of TSO requests, we also need the CPU usage info from the PD server in real-time to help us measure the actual effectiveness of the current batch size.

Introduce `CPUCollector` to collect the CPU usage of a PD server within a certain interval.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/cpu_collector/cpu_collector.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/pingcap/log"
+	"github.com/shirou/gopsutil/v3/cpu"
+	"github.com/tikv/pd/pkg/errs"
+	"go.uber.org/zap"
+)
+
+// CPUCollector is a daemon to collect the CPU usage of a PD server within a certain interval.
+type CPUCollector struct {
+	// Make sure every collector only has one background goroutine to keep collecting and updating the usage.
+	once     sync.Once
```
rleungx, line 32: Could you give an example about there are more than one goroutine?

JmPotato: Considering that the packages under pkg are somewhat generic, so the main purpose is to prevent the caller from starting the same Collector multiple times. This is not currently a possibility within PD.

rleungx: Perhaps, it should be guaranteed by the caller?

PR #4064, opened 2021-09-02
placement: early exit searching during FitRegion

Yisaer:

FitRegion will search all possible solutions which will consume much cpu resources while for some common cases the best solution is quite similar for other cases.

early exit searching for some common cases in order to save cpu resource.

### Check List

Tests

- Manual Bnechmark Test

#### early exit

```sh
3 default peers
BenchmarkFitRegion-4   	 1054921	      1053 ns/op

5 peers
BenchmarkFitRegion-4   	  657933	      1582 ns/op

3 peers with 1 learner
BenchmarkFitRegion-4   	  763387	      1586 ns/op

3 peers in each zone
BenchmarkFitRegion-4   	  696102	      1553 ns/op
```

#### origin

```sh
3 default peers
BenchmarkFitRegionMoreStores-4   	  829810	      1274 ns/op

5 peers
BenchmarkFitRegionMoreStores-4   	  437342	      2740 ns/op

3 peers with 1 learner
BenchmarkFitRegionMoreStores-4   	  157977	      6533 ns/op

3 peers in each zone
BenchmarkFitRegionMoreStores-4   	   46964	     21685 ns/op
```

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/placement/fit.go
+++ b/server/schedule/placement/fit.go
@@ -195,7 +195,7 @@ func (w *fitWorker) fitRule(index int) bool {
 	if index >= len(w.rules) {
 		// If there is no isolation level and we already find one solution, we can early exit searching instead of
 		// searching the whole cases.
-		if w.bestFit.IsSatisfied() && !w.needIsolation {
+		if !w.needIsolation && w.bestFit.IsSatisfied() {
 			w.exit = true
 		}
 		return false
--- a/server/schedule/placement/fit.go
+++ b/server/schedule/placement/fit.go
@@ -150,11 +150,12 @@ func FitRegion(stores []*core.StoreInfo, region *core.RegionInfo, rules []*Rule)
 }
 
 type fitWorker struct {
-	stores  []*core.StoreInfo
-	bestFit RegionFit  // update during execution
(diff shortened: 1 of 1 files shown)


server/schedule/placement/fit.go
```diff
@@ -186,7 +187,15 @@ func (w *fitWorker) run() {
 // Index specifies the position of the rule.
 // returns true if it replaces `bestFit` with a better alternative.
 func (w *fitWorker) fitRule(index int) bool {
+	if w.exit {
+		return false
+	}
 	if index >= len(w.rules) {
+		// If there is no isolation level and we already find one solution, we can early exit searching instead of
+		// searching the whole cases.
+		if w.bestFit.IsSatisfied() && !needIsolation(w.rules) {
```
disksing, line 196: I'd like to propose to calculate `needIsolation` when creating the worker to save CPU when `needIsolation` is always true.

Yisaer: good catch, updated.

disksing (approved):

JmPotato (approved):

Issue #4045, opened 2021-08-27
PD may remove wrong pending peer if the peer count exceeds the limit


andylokandy:
### What did you do?

The steps to reproduce:

1. Assume we have 3 stores: `Store1` `Store2` `Store3`. `Region1` is a two-replica region and has 1 voter `Peer1` and 1 learner `Peer2`:

| Store1 | Store2 | Store3 |
| ------------- | ------------- | ---- |
| Peer1 (leader)  | Peer2 (learner) | |

2.  PD decides to move the learner to `Store3`, therefore, PD adds a new learner `Peer3` to `Store3`:

| Store1 | Store2 | Store3 |
| ------------- | ------------- | ---- |
| Peer1 (leader)  | Peer2 (learner) | Peer3 (learner, no data, pending) |

3. `Store2` suffers from network jitter, which makes Peer2 fails to catch up with the leader:

| Store1 | Store2 | Store3 |
| ------------- | ------------- | ---- |
| Peer1 (leader)  | Peer2 (learner, pending) | Peer3 (learner, no data, pending) |

4. `Store3`'s network becomes unstable and thus unable to get a snapshot from the leader. PD cancels the `add peer` operation on timeout. And then 3 peers exist for `Region1`, and PD is going to delete one of the peers:

| Store1 | Store2 | Store3 |
| ------------- | ------------- | ---- |
| Peer1 (leader)  |  | Peer3 (learner, no data, pending) |

5. `Store1` goes down, then no data is left for `Region1` in the TiKV cluster:

| Store1 | Store2 | Store3 |
| ------------- | ------------- | ---- |
|  |  | Peer3 (learner, no data, pending) |

### What did you expect to see?

PD should not delete extra peers if the number of normal peers does not exceed the limit.

### What version of PD are you using (`pd-server -V`)?

master

zz-jason:
will this bug exist if there are at least 3 voters?

zz-jason:
> PD should not delete extra peers if the number of normal peers does not exceed the limit.

Sorry I didn't get the point, could you elaborate more on:

1. what are the "extra peers"?
2. what are "normal peers"?
3. what is the "limit"?

andylokandy:
@zz-jason 

> will this bug exist if there are at least 3 voters?

There will be.

>  what are the "extra peers"? what is the "limit"?

If a region is desired to have n replicas, and we now have m replicas in fact (m > n), then there are (m - n) extra peers.

> what are "normal peers"?

[...]

PR #4067, opened 2021-09-02
placement: do not delete orphan peers if some peers selected by RuleFit is down or pending

andylokandy:

fixes https://github.com/tikv/pd/issues/4045

Related changes

- Need to cherry-pick to the release-5.0

### Release note

```release-note
None
```


disksing:
If pending peer is considered *not satisfied*, the checker will create an operator to replace the pending peer immediately. But I think in most cases, we should wait for the status to become normal rather than replace it.

andylokandy:
@disksing I've reimplemented the patch, PTAL.

disksing (approved):

disksing:
btw, can we add a test case for it?

nolouch (commented):
could you add a test?

nolouch (approved):
lgtm

PR #4069, opened 2021-09-03
placement: fix nil error during FitRegion cache

Yisaer:

Fix nil error during FitRegion Cache. Previously, `SetCache` is called by `rule_checker` and `InvalidCache` is called by `Fitregion`. This made `Set` and `Invalid` would run in different goroutine and which may caused nil error in following situation.

```sh
     goroutine1  goroutine2 
1.   CheckCache
2.               Invalid Cache
3.   GetCache
```

Let `rule_checker` invalid cache so that `set` and `invalid` will run in one goroutine. We also merge `check` and `get` in order to guarantee by the same mutex.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/rule_checker.go
+++ b/server/schedule/checker/rule_checker.go
@@ -65,16 +65,22 @@ func (c *RuleChecker) Check(region *core.RegionInfo) *operator.Operator {
 }
 
 // CheckWithFit checkWithFit is similar with Checker with placement.RegionFit
-func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.RegionFit) *operator.Operator {
+func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.RegionFit) (op *operator.Operator) {
 	// If the fit is fetched from cache, it seems that the region doesn't need cache
 	if fit.IsCached() {
+		failpoint.Inject("assertShouldNotCache", func() {
+			panic("cached shouldn't be used")
+		})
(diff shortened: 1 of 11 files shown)


disksing:
why delete `TestFitRegionCache`?

Yisaer:
> why delete `TestFitRegionCache`?

As the `invalidCache` in moved into `checker`, the related testcases are moved into `TestRuleCache`.

server/schedule/checker/rule_checker.go
```diff
@@ -65,17 +65,23 @@ func (c *RuleChecker) Check(region *core.RegionInfo) *operator.Operator {
 }
 
 // CheckWithFit checkWithFit is similar with Checker with placement.RegionFit
```
JmPotato, line 67:
```suggestion
// CheckWithFit is similar with Checker with placement.RegionFit
```

server/schedule/placement/rule_manager.go
```diff
@@ -328,6 +324,11 @@ func (m *RuleManager) SetRegionFitCache(region *core.RegionInfo, fit *RegionFit)
 	m.cache.SetCache(region, fit)
 }
 
+// InvalidCache invalid cache
```
JmPotato, line 327:
```suggestion
// InvalidCache invalids the cache.
```

JmPotato (approved):
The rest LGTM.

disksing (approved):

PR #4078, opened 2021-09-06
checker: judging that the peer is down is no longer based on DownSeconds

### LINKED ISSUE #4077: PD will not repair down-peer in time
### What did you do?

In a cluster with sufficient resources, one TiKV is down.

After half an hour, the peer on this TiKV will be moved to other TiKVs.

### What did you see instead?

The remaining Peer has not been processed.

v5.2.0

HunDunDM:

Fix #4077 

After #3462, `DownSeconds` no longer update with the heartbeat. It will cause the down-peer to not be repaired in time.

Judging that the peer is down is no longer based on `DownSeconds`.

### Check List

Tests

- Unit test

Code changes

Side effects

  - PD judge that the peer is down is no longer based on `DownSeconds`. In extreme cases, down-peers that would not have been processed will be fixed.

Related changes

### Release note

```release-note
Fix the bug that PD would not fix down-peer in time.
```

### DIFF
--- a/server/schedule/checker/replica_checker.go
+++ b/server/schedule/checker/replica_checker.go
@@ -110,10 +110,6 @@ func (r *ReplicaChecker) checkDownPeer(region *core.RegionInfo) *operator.Operat
 		if store.DownTime() < r.opts.GetMaxStoreDownTime() {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


lhy1024 (approved):

disksing:
does it break compatibility?

nolouch (approved):

PR #4089, opened 2021-09-07
placement: use struct in order to avoid ignoring changing cache

Yisaer:

Previously, we use pointer to store the related variable of `FitRegionCache` which may cause the following bug.

```sh
1. FitRegionCache saves pointer of rules, pointer of region,pointer of stores.
2. rules changed
3. FitRegionCache didn't find the variable is changed as the stored variables are also changed.

```

use interval struct to store the required variable in order to avoid the aboved problem. Also, use struct will avoid the nil panic during FitRegionCache.

### Check List

Tests

- Unit test

Side effects

As each `FitRegionCache` will use struct to store the variables, each cache is nearly about `96` byte. If we save 1000k FitRegionCache, it will consume about 91MB memory, thus I think the memory consuming is acceptable.

Related changes

### Release note

```release-note
None
```


server/schedule/placement/region_rule_cache.go
```diff
@@ -77,63 +71,162 @@ func (manager *RegionRuleFitCacheManager) CheckAndGetCache(region *core.RegionIn
...
-			return equalRules(a[i], b[j])
+			return a[i].storeEqual(b[j])
 		})
 	})
 }
 
-func storesEqual(a, b []*core.StoreInfo) bool {
-	if len(a) != len(b) {
+func validateRegion(region *core.RegionInfo) bool {
+	if region == nil || region.GetLeader() == nil || len(region.GetDownPeers()) > 0 || region.GetRegionEpoch() == nil {
```
rleungx, line 128: directly use return

server/schedule/placement/region_rule_cache.go
```diff
@@ -77,63 +71,162 @@ func (manager *RegionRuleFitCacheManager) CheckAndGetCache(region *core.RegionIn
...
+			id:       rule.ID,
+			group:    rule.GroupID,
+			version:  rule.Version,
+			createTS: rule.CreateTimestamp,
 		})
+	}
+	return c
+}
+
+type storeCache struct {
```
rleungx, line 169: prefer to use pointer receiver

disksing:
I don't understand. How can a `Rule` be changed? I think we always clone the rule then update it.

server/schedule/placement/region_rule_cache.go
Yisaer: Why use pointer here? I tend to use struct in order to avoid panic error

disksing (approved):

server/schedule/placement/region_rule_cache.go
```diff
@@ -77,63 +71,163 @@ func (manager *RegionRuleFitCacheManager) CheckAndGetCache(region *core.RegionIn
...
-			return equalRules(a[i], b[j])
+			return a[i].storeEqual(b[j])
 		})
 	})
 }
 
-func storesEqual(a, b []*core.StoreInfo) bool {
-	if len(a) != len(b) {
+func validateRegion(region *core.RegionInfo) bool {
+	return !(region == nil || region.GetLeader() == nil || len(region.GetDownPeers()) > 0 || region.GetRegionEpoch() == nil)
```
rleungx, line 128:
```suggestion
	return region != nil && region.GetLeader() != nil && len(region.GetDownPeers()) <= 0 && region.GetRegionEpoch() != nil
```

server/schedule/placement/region_rule_cache_test.go
```diff
@@ -141,18 +144,53 @@ func (s *testRuleSuite) TestRegionRuleFitCache(c *C) {
...
+			name:      "invalid input3",
+			region:    mockRegion(3, 0),
+			rules:     nil,
+			unchanged: false,
+		},
 	}
 	for _, testcase := range testcases {
 		c.Assert(cache.IsUnchanged(testcase.region, testcase.rules, mockStores(3)), Equals, testcase.unchanged)
 	}
+	// Invalid Input3
```
rleungx, line 177: Invalid Input3?

rleungx (approved):
The rest LGTM

PR #4100, opened 2021-09-09
tools: add cancel-delete store command for pd-ctl

### LINKED ISSUE #4098: support abort deleting a store

It's possible that user accidentally deletes a undesired store. Currently, there is an undocumented way to abort the operation by:
```
curl -X POST http://<ip>:<port>/pd/api/v1/store/<store_id>/state?state=Up
```
. The operation is dangerous as it allows changing state manually. A more safe way is providing a public and well documented command/API to abort the operation.

rleungx:

Issue Number: Close #4098

```commit-message
```

```
» store --help
manipulate or query stores

Usage:
  pd-ctl store [command] [flags]
  pd-ctl store [command]

Available Commands:
  cancel-delete    cancel delete the store
  check            Check all the stores with specified status
  delete           delete the store
  label            set a store's label value
  limit            show or set a store's rate limit
  limit-scene      show or set the limit value for a scene
  remove-tombstone remove tombstone record if only safe
  weight           set a store's leader and region balance weight

Flags:
  -h, --help            help for store
      --jq string       jq query
      --state strings   state filter

Use "pd-ctl store [command] --help" for more information about a command.
```

First, delete store 2:
```
» store delete 2
Success!
```
```
"store": {
   "id": 2,
   "state_name": "Offline"
}
```
Then, cancel delete store 2:
```
» store cancel-delete 2
Success!
```

```
"store": {
   "id": 2,
   "state_name": "Up"
}
```

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -117,7 +117,7 @@ type RaftCluster struct {
 	storage            storage.Storage
 	minResolvedTS      uint64
 	// Keep the previous store limit settings when removing a store.
-	prevStoreLimit map[uint64]map[storelimit.Type]*storelimit.StoreLimit
+	prevStoreLimit map[uint64]map[storelimit.Type]float64
 
 	// This below fields are all read-only, we cannot update itself after the raft cluster starts.
 	clusterID                uint64
@@ -222,7 +222,7 @@ func (c *RaftCluster) InitCluster(
 	c.hotStat = statistics.NewHotStat(c.ctx)
 	c.progressManager = progress.NewManager()
 	c.changedRegions = make(chan *core.RegionInfo, defaultChangedRegionsLimit)
-	c.prevStoreLimit = make(map[uint64]map[storelimit.Type]*storelimit.StoreLimit)
+	c.prevStoreLimit = make(map[uint64]map[storelimit.Type]float64)
 }
 
 // Start starts a cluster.
@@ -1115,12 +1115,13 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 	if err == nil {
 		c.progressManager.AddProgress(encodeRemovingProgressKey(storeID), float64(c.core.GetStoreRegionSize(storeID)))
 		c.resetProgress(storeID, store.GetAddress(), preparingAction)
+		// record the current store limit in memory
+		c.prevStoreLimit[storeID] = map[storelimit.Type]float64{
+			storelimit.AddPeer:    c.GetStoreLimitByType(storeID, storelimit.AddPeer),
+			storelimit.RemovePeer: c.GetStoreLimitByType(storeID, storelimit.RemovePeer),
+		}
 		// TODO: if the persist operation encounters error, the "Unlimited" will be rollback.
 		// And considering the store state has changed, RemoveStore is actually successful.
-		c.prevStoreLimit[storeID] = map[storelimit.Type]*storelimit.StoreLimit{
-			storelimit.AddPeer:    store.GetStoreLimit(storelimit.AddPeer),
-			storelimit.RemovePeer: store.GetStoreLimit(storelimit.RemovePeer),
-		}
 		_ = c.SetStoreLimit(storeID, storelimit.RemovePeer, storelimit.Unlimited)
 	}
 	return err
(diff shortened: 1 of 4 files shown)


tests/pdctl/store/store_test.go
```diff
@@ -276,6 +277,17 @@ func (s *storeTestSuite) TestStore(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(strings.Contains(string(output), "Unknown state: Invalid_state"), IsTrue)
 
+	// store undelete <store_id> command
+	args = []string{"-u", pdAddr, "store", "undelete", "1"}
```
lhy1024, line 281: revote-delete? revote-offline?

rleungx: Do you mean revoke or revert?

lhy1024: I think `revoke` is better than `undelete`

HunDunDM:
I think it is necessary to confirm whether the Store state is `Offline` to prevent users from using pd-ctl to connect to the old version PD and change the Store state from `Tombstone` to `Up`.

rleungx:
> I think it is necessary to confirm whether the Store state is `Offline` to prevent users from using pd-ctl to connect to the old version PD and change the Store state from `Tombstone` to `Up`.

make sense

BusyJay:
Perhaps "cancel" is a more precise word for this situation.

rleungx:
> Perhaps "cancel" is a more precise word for this situation.

I think both of them are ok to me.

BusyJay:
Revoke usually means take away, like revoke privileges or licenses. Cancel means abort, stop doing sth.

tools/pd-ctl/pdctl/command/store_command.go
```diff
@@ -304,54 +327,124 @@ func deleteStoreCommandFunc(cmd *cobra.Command, args []string) {
...
+	}
+
+	prefix := fmt.Sprintf(storePrefix, args[0])
+	r, err := doRequest(cmd, prefix, http.MethodGet)
+	if err != nil {
+		cmd.Printf("Failed to get store: %s\n", err)
+		return
+	}
+
+	storeInfo := struct {
```
HunDunDM, line 361: Can the defined struct be used here?

tools/pd-ctl/pdctl/command/store_command.go
```diff
@@ -304,54 +327,124 @@ func deleteStoreCommandFunc(cmd *cobra.Command, args []string) {
...
+			Address string `json:"address"`
+			State   int32  `json:"state"`
+		} `json:"store"`
+	}{}
+	if err = json.Unmarshal([]byte(r), &storeInfo); err != nil {
+		cmd.Printf("Failed to parse store info: %s\n", err)
+		return
+	}
+
+	if storeInfo.Store.State != 1 {
```
HunDunDM, line 373: `1` It's so magical.

tools/pd-ctl/pdctl/command/store_command.go
rleungx: We can but I don't think it's necessary.

HunDunDM (approved):

BusyJay:
This is not enough, store limit should be reset to normal after canceling.

HunDunDM:
Since it was not a reversible operation before, the value before offline will not be recorded. Will it be a problem if it always changes back to 15?

rleungx:
> Since it was not a reversible operation before, the value before offline will not be recorded. Will it be a problem if it always changes back to 15?

I think it needs further discussion, hold for now.

nolouch (approved):

nolouch (changes requested):
Let us change back to the default store limit first?

rleungx:
> Let us change back to the default store limit first?

I don't think it's a good option if the users change the all stores' limit before.

rleungx:
After discussing with our PM, he suggests adding a limit option when delete or cancel-delete store. 
e.g.
```
store delete 1 1000(represent the limit)
store cancel-delete 1 15(represent the limit)
```
WDYT @BusyJay @nolouch

nolouch:
can we save the previous store limit?

lhy1024 (commented):

BusyJay:
I don't have strong opinion on how to restore previous settings.

nolouch (approved):
lgtm

server/cluster/cluster.go
```diff
@@ -1277,23 +1278,26 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
...
-	if limiter[storelimit.AddPeer] != nil {
-		options = append(options, core.ResetStoreLimit(storelimit.AddPeer, limiter[storelimit.AddPeer].Rate()))
-	}
-	if limiter[storelimit.RemovePeer] != nil {
-		options = append(options, core.ResetStoreLimit(storelimit.RemovePeer, limiter[storelimit.RemovePeer].Rate()))
+	// get the previous store limit recorded in memory
+	limiter, exist := c.prevStoreLimit[storeID]
+	if exist {
+		options = append(options,
+			core.ResetStoreLimit(storelimit.AddPeer, limiter[storelimit.AddPeer]),
```
JmPotato, line 1286: Don't we need to check whether the `limiter[storelimit.AddPeer]` is nil anymore?

rleungx: If the store limit has existed, add peer won't be nil.

JmPotato (approved):

PR #4102, opened 2021-09-09
election: fix the data race of the leadership resetting

JmPotato:

Fix the data race of the leadership resetting.

image

ref: https://ci2.pingcap.net/blue/rest/organizations/jenkins/pipelines/pd_test/runs/2203/nodes/50/steps/94/log/?start=0

Make sure the leadership is not expired before resetting it.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/election/lease.go
+++ b/server/election/lease.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/etcdutil"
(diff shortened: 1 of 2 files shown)


rleungx:
Do we need to cherry-pick this PR?

JmPotato:
> Do we need to cherry-pick this PR?

I think cherry-pick is needed since this data race may affect the normal leadership campaign, i.e, cancel the context immediately after it's set.

nolouch (approved):

PR #4104, opened 2021-09-13
client: refactor the client options to make the scope clear

JmPotato:

We have many options inside both the `client` and `baseClient`, however, each of them may work in different scopes. To make the scope clear, we need to do some refactoring.

Distinguish between different client scopes of options.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/base_client.go
+++ b/client/base_client.go
@@ -55,10 +55,10 @@ type baseClient struct {
 
 	security SecurityOption
 
-	gRPCDialOptions  []grpc.DialOption
-	timeout          time.Duration
-	maxRetryTimes    int
-	enableForwarding bool
+	// BaseClient options.
+	gRPCDialOptions []grpc.DialOption
+	timeout         time.Duration
+	maxRetryTimes   int
 }
 
 // SecurityOption records options about tls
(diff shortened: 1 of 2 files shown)


HunDunDM (approved):

client/base_client.go
```diff
@@ -55,10 +55,10 @@ type baseClient struct {
...
 	security SecurityOption
 
-	gRPCDialOptions  []grpc.DialOption
-	timeout          time.Duration
-	maxRetryTimes    int
-	enableForwarding bool
+	// BaseClient options.
+	gRPCDialOptions []grpc.DialOption
+	timeout         time.Duration
+	maxRetryTimes   int
```
rleungx, line 61: only move out enableForwarding?

JmPotato: Because the usage scope of `enableForwarding` is in `client`, and the others in `baseClient`. We also need to consider the scope of the new configuration in the future before adding it to the corresponding scope. For example, `maxBatchWaitInterval` is an option that should be placed in the `client` rather than `baseClient`.

HunDunDM: https://github.com/tikv/pd/pull/4054#discussion_r706986632

rleungx: The base client is abstracted as a common part for dynamic configuration. Since that feature is changed a lot, there is no need to distinguish between client and base client.

JmPotato: I think this change is more about code style. Putting different options into its scope can help understand the role of an option better.

rleungx (approved):

PR #4106, opened 2021-09-13
scheduler: allow empty region to be scheduled and use a sperate tolerance config in scatter range scheduler

lhy1024:

1. we expect the tolerance ratio in range cluster should be 2, but it doesn't take effect.
https://github.com/tikv/pd/blob/3f53f1ea41f35b1cfd4576c890b122e44f60a746/server/schedulers/scatter_range.go#L220
https://github.com/tikv/pd/blob/3f53f1ea41f35b1cfd4576c890b122e44f60a746/server/schedulers/utils.go#L154

2. we expect it can schedule empty region, but it can not and  test can not detect
https://github.com/tikv/pd/blob/3f53f1ea41f35b1cfd4576c890b122e44f60a746/server/core/region.go#L96
https://github.com/tikv/pd/blob/3f53f1ea41f35b1cfd4576c890b122e44f60a746/server/schedulers/balance_test.go#L1103-L1111

allow empty region to be scheduled and use a sperate tolerance config in scatter range scheduler

### Check List

Tests

- Unit test

Add scatter range scheduler

Before this pr
image
After this pr
image

Related changes

### Release note

```release-note
allow empty region to be scheduled and use a sperate tolerance config in scatter range scheduler
```

### DIFF
--- a/server/schedule/opt/healthy.go
+++ b/server/schedule/opt/healthy.go
@@ -54,13 +54,7 @@ func HealthAllowPending(cluster Cluster) func(*core.RegionInfo) bool {
 
 // AllowBalanceEmptyRegion returns a function that checks if a region is an empty region and can be balanced.
 func AllowBalanceEmptyRegion(cluster Cluster) func(*core.RegionInfo) bool {
-	switch cluster.GetClusterType() {
-	case core.RangeCluster:
-		// allow empty region to be scheduled in range cluster
-		return func(region *core.RegionInfo) bool { return true }
-	default:
-		return func(region *core.RegionInfo) bool { return IsEmptyRegionAllowBalance(cluster, region) }
-	}
+	return func(region *core.RegionInfo) bool { return IsEmptyRegionAllowBalance(cluster, region) }
 }
 
 // IsRegionReplicated checks if a region is fully replicated. When placement
--- a/server/schedule/opt/healthy.go
+++ b/server/schedule/opt/healthy.go
@@ -54,7 +54,13 @@ func HealthAllowPending(cluster Cluster) func(*core.RegionInfo) bool {
 
 // AllowBalanceEmptyRegion returns a function that checks if a region is an empty region and can be balanced.
 func AllowBalanceEmptyRegion(cluster Cluster) func(*core.RegionInfo) bool {
-	return func(region *core.RegionInfo) bool { return IsEmptyRegionAllowBalance(cluster, region) }
+	switch cluster.GetClusterType() {
+	case core.RangeCluster:
+		// allow empty region to be scheduled in range cluster
+		return func(region *core.RegionInfo) bool { return true }
+	default:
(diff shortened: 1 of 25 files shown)


server/schedulers/utils.go
```diff
@@ -151,7 +152,14 @@ func (p *balancePlan) getTolerantResource() int64 {
...
 
 func adjustTolerantRatio(cluster opt.Cluster, kind core.ScheduleKind) float64 {
-	tolerantSizeRatio := cluster.GetOpts().GetTolerantSizeRatio()
+	var tolerantSizeRatio float64
+	switch cluster.GetClusterType() {
+	case core.RangeCluster:
+		// range cluster use a separate configuration
+		tolerantSizeRatio = cluster.(*schedule.RangeCluster).GetTolerantSizeRatio()
+	default:
+		tolerantSizeRatio = cluster.GetOpts().GetTolerantSizeRatio()
```
rleungx, line 161: How about adding GetTolerantSizeRatio in raft cluster?

server/schedule/opt/healthy.go
```diff
@@ -54,7 +54,13 @@ func HealthAllowPending(cluster Cluster) func(*core.RegionInfo) bool {
 
 // AllowBalanceEmptyRegion returns a function that checks if a region is an empty region and can be balanced.
 func AllowBalanceEmptyRegion(cluster Cluster) func(*core.RegionInfo) bool {
-	return func(region *core.RegionInfo) bool { return IsEmptyRegionAllowBalance(cluster, region) }
+	switch cluster.GetClusterType() {
```
rleungx, line 57: Can we use type assertion?

lhy1024: it will introduce cycle import between `schedule`  and `schedule/opt`

server/schedulers/utils.go
lhy1024: If so, we will need implementation for all cluster

disksing (approved):

server/schedulers/balance_region.go
```diff
@@ -152,24 +152,35 @@ func (s *balanceRegionScheduler) Schedule(cluster opt.Cluster) []*operator.Opera
 		return stores[i].RegionScore(opts.GetRegionScoreFormulaVersion(), opts.GetHighSpaceRatio(), opts.GetLowSpaceRatio(), iOp) >
 			stores[j].RegionScore(opts.GetRegionScoreFormulaVersion(), opts.GetHighSpaceRatio(), opts.GetLowSpaceRatio(), jOp)
 	})
+
+	var allowBalanceEmptyRegion func(*core.RegionInfo) bool
+
+	switch cluster.(type) {
+	case *schedule.RangeCluster:
+		// allow empty region to be scheduled in range cluster
+		allowBalanceEmptyRegion = func(region *core.RegionInfo) bool { return true }
```
rleungx, line 161: directly use true?

rleungx (approved):

PR #4120, opened 2021-09-14
labeler: change key-range config to a list of ranges

disksing:

Then TiDB uses table partition, a table's range can have multiple key-ranges. If we want to set label rule for a table, we need to update many rules.

Support new type `key-ranges` to allow set multiple ranges for a rule.

### Check List

Tests
- Unit test

### Release note

```release-note
None
```


server/schedule/labeler/labeler.go
```diff
@@ -96,46 +96,75 @@ func (l *RegionLabeler) adjustRule(rule *LabelRule) error {
...
+		rs := make([]*KeyRangeRule, 0, len(rules))
+		for _, r := range rules {
+			rr, err := l.adjustKeyRangeRule(r)
+			if err != nil {
+				return err
+			}
+			rs = append(rs, rr)
 		}
-		rule.Rule = &r
+		rule.Rule = rs
```
rleungx, line 118: How about using another name for `rule.Rule`? See https://github.com/pingcap/tidb/pull/27660#pullrequestreview-753499559

server/schedule/labeler/rules.go
```diff
@@ -32,6 +32,8 @@ type LabelRule struct {
 const (
 	// KeyRange is the rule type that labels regions by key range.
 	KeyRange = "key-range"
+	// KeyRanges is the rule type that specifies a list of key ranges.
+	KeyRanges = "key-ranges"
```
JmPotato, line 36: The name `key-ranges` is too similar to the `key-range`, what about using `key-range-list` instead?

server/schedule/labeler/labeler.go
disksing: I'm not quite undertand. Would you like to provide a better name?

server/schedule/labeler/rules.go
disksing: @rleungx suggests to keep only one type. So I'm going to merge the 2 types to one (name `key-range`, but is a list of ranges)

server/schedule/labeler/labeler.go
rleungx: I have no idea about it =。=

server/schedule/labeler/labeler.go
```diff
@@ -96,16 +96,12 @@ func (l *RegionLabeler) adjustRule(rule *LabelRule) error {
...
 	case KeyRange:
-		r, err := l.adjustKeyRangeRule(rule.Rule)
-		if err != nil {
-			return err
-		}
-		rule.Rule = r
-		return nil
-	case KeyRanges:
-		rules, ok := rule.Rule.([]map[string]interface{})
+		rules, ok := rule.Rule.([]interface{})
```
rleungx, line 99: How about change it to `rule.Rule.([]interface{})`

disksing: but it is not required to be plural. This field contains the rule logic of the rule. Its content type depends on the rule-type. Maybe we can use some abstract word, like `Data` or `Detail`.

rleungx: make sence

rleungx: Will it be changed in this PR?

disksing: use `Data`?

rleungx: ok

server/schedule/labeler/rules.go
```diff
@@ -26,7 +26,7 @@ type LabelRule struct {
 	Index    int           `json:"index"`
 	Labels   []RegionLabel `json:"labels"`
 	RuleType string        `json:"rule_type"`
-	Rule     interface{}   `json:"rule"`
+	Data     interface{}   `json:"Data"`
```
rleungx, line 29:
```suggestion
	Data     interface{}   `json:"data"`
```

rleungx (approved):
The rest LGTM.

JmPotato (approved):
LGTM. BTW, so the constraint that we need to make sure the number of `key-range` is even on `ruel.Data` is guaranteed by TiDB?

disksing:
> LGTM. BTW, so the constraint that we need to make sure the number of `key-range` is even on `ruel.Data` is guaranteed by TiDB?

There is no such constraint. I think you have misunderstood it. The `makeKeyRanges` function is only used in tests.

PR #4143, opened 2021-09-22
*: simplify assert

rleungx:

Make assert simpler. 

- replace `assert(len(xxx), Equals, xxx)` with `assert(xxx, HasLen, xxx)`
- replace `assert(xxx, Equals, true/false)` with `assert(xxx, IsTrue/IsFalse)`

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


tests/server/cluster/cluster_test.go
```diff
@@ -414,7 +414,7 @@ func (s *clusterTestSuite) TestGetPDMembers(c *C) {
 	resp, err := grpcPDClient.GetMembers(context.Background(), req)
 	c.Assert(err, IsNil)
 	// A more strict test can be found at api/member_test.go
-	c.Assert(len(resp.GetMembers()), Not(Equals), 0)
+	c.Assert((resp.GetMembers(), Not(HasLen), 0)
```
JmPotato, line 417: Oops.

JmPotato (approved):

PR #4151, opened 2021-09-24
checker: rename priority checker to inspector

rleungx:

Here is a bug that when we pause the priority checker, it will skip the rule checker also. BTW, the name for the priority checker is a little bit confusing.

This PR renames the priority checker to the priority inspector. And if we call a checker, we think it will return operators.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


bufferflies (approved):
lgtm

Issue #4140, opened 2021-09-18
PD cannot satisfy 3 Voter and 1 learner rules


nolouch:
### What did you do?
 the region
```
peer 2 in zone2, peer 3  in zone3, peer 4 in zone3
```
add the rule 
```
[]string{"1/voter/zone=zone1//", "1/voter/zone=zone2/", "1/voter/zone=zone3/", "1/learner/zone=zone4/"
```

### What did you expect to see?

the region became 3voter + 1 learner

### What did you see instead?

the operator adds peer 5 in `zone1`, and blocking on 4 voters.

### What version of PD are you using (`pd-server -V`)?
master

nolouch:
the reason is only 3 rules can find the fit peer, the `learner rule` is not satisfied. but `fixOrphanPeers` only works on all rules are satisfied.

PR #4166, opened 2021-09-27
checker: support demote voter in rule checker

Yisaer:

close https://github.com/tikv/pd/issues/4140

Rule checker didn't support to demote voters in order to satisfy the learner rule. This request fixes it by making voter and learner rules as loose match during `FitRegion`.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/core/peer.go
+++ b/server/core/peer.go
@@ -23,6 +23,11 @@ func IsLearner(peer *metapb.Peer) bool {
 	return peer.GetRole() == metapb.PeerRole_Learner
 }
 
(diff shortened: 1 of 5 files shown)


disksing (commented):
do we need to consider if joint-consensus is enabled?

Yisaer:
> do we need to consider if joint-consensus is enabled?

I think we don't need to consider it unless we both demote voter and promote learner in one operator.

disksing (approved):

rleungx (approved):

Issue #3936, opened 2021-08-02
Cancel `LoadRegionsOnce` when exits `WatchLeader`


disksing:
## Bug Report

When the PD starts, if there is a leader (or if all PDs restart at the same time and etcd still has the leader key left over from the restart), the PD enters the follower state and loads the region from the region storage and starts the region syncer client to synchronize the region.
If all PDs are restarted at the same time, or if the old leader steps down, the newly started PD cannot start the election until the region is loaded.

Translated with www.DeepL.com/Translator (free version)

### What version of PD are you using (`pd-server -V`)?
master

PR #4175, opened 2021-09-28
core: allow cancel load region

disksing:

Fix https://github.com/tikv/pd/issues/3936

- add cancel support for core.LoadRegion functions
- replace `chan bool` with `contextes` in region syncer client
- add test case

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue that PD may not elect leader as soon as leader step down
```

### DIFF
--- a/server/core/storage_test.go
+++ b/server/core/storage_test.go
@@ -15,6 +15,7 @@
 package core
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"math"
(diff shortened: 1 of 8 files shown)


server/core/storage_test.go
```diff
@@ -167,7 +168,7 @@ func (s *testKVSuite) TestLoadRegionsToCache(c *C) {
 
 	n = 20
 	mustSaveRegions(c, storage, n)
-	c.Assert(storage.LoadRegionsOnce(cache.SetRegion), IsNil)
+	c.Assert(storage.LoadRegionsOnce(context.Background(), cache.SetRegion), IsNil)
```
JmPotato, line 171: Maybe we can create only one context in a single test case.

server/region_syncer/server.go
```diff
@@ -95,7 +96,7 @@ func NewRegionSyncer(s Server) *RegionSyncer {
 		tlsConfig: s.GetTLSConfig(),
 	}
 	syncer.mu.streams = make(map[string]ServerStream)
-	syncer.mu.closed = make(chan struct{})
+	syncer.mu.clientCtx, syncer.mu.clientCancel = context.WithCancel(context.Background())
```
JmPotato, line 99: Just a question, why not create a sub-context from the server rather than using a `context.Background()`?

PR #4179, opened 2021-09-29
schedule: remove the store limit from OperatorController

rleungx:

Simplify the code of the store limit.

- Remove `AttachAvailableFunc` from interface
- Add `GetBasicCluster` interface
- Remove the store limit from `OperatorController`
- Use Get/ResetStoreLimit instread

### Check List

Tests

- Unit test
- 
### Release note

```release-note
None
```


disksing:
The rest LGTM.

PR #4196, opened 2021-10-09
*: clean up the opt package

rleungx:

Some code enhancement

Clean up the unnecessary code in opt package.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


HunDunDM:
`Cluster` is an `Interface` defined in `opt`, which can be `RaftCluster` or `MockCluster`. Using `RuleManager` to `FitRegion` is the behavior defined on the `Cluster` interface, not the logic in `RaftCluster`.

rleungx:
> `Cluster` is an interface defined in `opt`, which can be `RaftCluster` or `MockCluster`. `CacheCluster` should be based on `Cluster` interface instead of `RaftCluster`.
> 
> Using `RuleManager` to `FitRegion` is the behavior defined on the `Cluster` interface like `IsRegionHealthy`.

IMO, CacheCluster is not very useful in mock cluster. As for FitRegion, we have already had RuleManager in cluster interface. Maybe remove it make the package more clear?

server/config/config.go
```diff
@@ -1184,6 +1184,10 @@ type StoreLabel struct {
 	Value string `toml:"value" json:"value"`
 }
 
+// RejectLeader is the label property type that suggests a store should not
+// have any region leaders.
+const RejectLeader = "reject-leader"
```
HunDunDM, line 1189: how about `server/core/store.go`?

HunDunDM (commented):
rest LGTM

server/config/config.go
rleungx: I think it can be regarded as a kind of config to some degree.

HunDunDM (approved):

Issue #4095, opened 2021-09-08
Enhance Split Table Stability and Usability


nolouch:
## Description

In the [high concurrency practice of TiDB](https://docs.pingcap.com/tidb/stable/high-concurrency-best-practices), split table is a common operation for users. It can help users solve hotspot issues in advance and prevent sudden traffic from being unable to handle it.

## Document Collection

- Proposal doc: ${proposal doc}
- Weekly report: ${weekly report link}

## Talent Challenge Program information

- Mentor of this issue: @Yisaer 
- Recommended skills:  Go
- Estimated Workloads: 2XL

## Milestones and action items

Milestone 1:  Enhance the `split table` stability

- [ ] dive to the problem https://github.com/pingcap/tidb/issues/24295
- [ ] Let PD check regions before splitting regions https://github.com/tikv/pd/issues/2733
- [ ] All downstream use new API.

Milestone 2: Enhance the `split table` usability
refer: https://github.com/wentaojin/tidba, we can extend the `split table` syntax.

- [ ] TBD
- [ ] TBD

PR #4212, opened 2021-10-15
grpc-service: Add splitAndScatterRegions

hzh0425:
ref: https://github.com/pingcap/tidb/issues/29034

Add a method splitAndScatterRegions in grpc-service
let pd handle split and scatter regions


rleungx:
Does this PR have any context?

hzh0425:
> Does this PR have any context?

yep, this is a talent challenge program, see here : https://github.com/tikv/pd/issues/4095
my mentor is @Yisaer

nolouch:
Also need to update the context in this issue, and do you have detail design RFC?

hzh0425:
close #4224

Yisaer:
> I know it's a subtask, But the design is different from the issue. Such as [pingcap/tidb#25281](https://github.com/pingcap/tidb/issues/25281), the first thing is to write the design RFC. we recommend writing proposal RFC before coding, for this topic you can discuss it at https://internals.tidb.io/t/topic/399.

Currently this task is target to merge `split` and `scatter` in PD and allow tidb to directly call this merged function. I think it only needs to use `Issue` to track this task in TiDB repo instead of rfc. This task isn't complex enough to propose a rfc.

hzh0425:
ref: https://github.com/pingcap/tidb/issues/29034

PR #4214, opened 2021-10-15
schedulers: evict leader support multiple targets

MrCroxx:

Part of https://github.com/tikv/tikv/issues/10602.

close [#4229](https://github.com/tikv/pd/issues/4229)

Evict leader scheduler picks all valid stores, let TiKV decide the best store to transfer leader to.

### Check List

Tests

- [x] Unit test
- [x] Integration test
- [x] Manual test, see [tikv#10602](tikv/tikv#10602).

Side effects

- Evict leader may lead to leader unbalanced when some store is slow.

Related changes

- [kvproto#822](https://github.com/pingcap/kvproto/pull/822)
- [tikv#11063](https://github.com/tikv/tikv/pull/11063)

### Release note

```release-note
Evict leader automatically pick the fastest store to transfer leader to.
```

Signed-off-by: MrCroxx <mrcroxx@outlook.com>


HunDunDM:
* Maybe using another name would be better than `TransferLeaderV2`, also in `kvproto`, it is a special step, not an enhancement of `TransferLeader`. (Refer to `ChangePeer` and `ChangePeerV2`)
* Refer to `versioninfo.JointConsensus` to implement backward compatibility and ensure that the `TransferLeader` is sent to TiKVs that do not support this feature.

MrCroxx:
> * Maybe using another name would be better than `TransferLeaderV2`, also in `kvproto`, it is a special step, not an enhancement of `TransferLeader`. (Refer to `ChangePeer` and `ChangePeerV2`)

From PD's view, the new evict leader implementation is not that similar with transfer leader, but from TiKV's view they are. IMO, reuse the origin `TransferLeader` proto with tiny modifications (`peer` => `peers`) is just okay? What do yo think? @BusyJay @NingLin-P

nolouch:
I wonder even the evict leader picks the best one, another scheduler like `balance-leader` will schedule the leader to the other store in order to maintain the leader number is balanced.

MrCroxx:
> I wonder even the evict leader picks the best one, another scheduler like `balance-leader` will schedule the leader to the other store in order to maintain the leader number is balanced.

@nolouch IMO, this modification is only for speeding up evict leader ops and making the store can be brought offline faster. It's fine that the transfered leaders need to be balanced smoothly later.

MrCroxx:
Let me summarize the conversations above.

1. Reuse the origin `TransferLeader` proto message by replacing singular field `peer` with repeatable field `peers`.
2. Add an config in PD to decide whether to enable the multi-target leader evicting feature.
3. Check versioninfo to decide whether to enable the feature. (or simply use random order in `peers` field to simplify the compatible problem?)

@HunDunDM Any updates?

BusyJay:
>     1. Reuse the origin `TransferLeader` proto message by replacing singular field `peer` with repeatable field `peers`.

Unfortunately, according to https://github.com/tikv/pd/pull/4214#issuecomment-946553963 it's not backward compatible, hence should use a new field instead.

nolouch:
Could you test the branch to show how many speeds improve？

server/schedule/operator/create_operator.go
```diff
@@ -73,6 +74,19 @@ func CreateForceTransferLeaderOperator(desc string, cluster opt.Cluster, region
 		Build(kind)
 }
 
+// CreateEvictLeaderOperator creates an operator that transfers the leader from a source store to one of the given target stores
+// or the given target store when incompatible.
+func CreateEvictLeaderOperator(desc string, cluster opt.Cluster, region *core.RegionInfo, sourceStoreID uint64, targetStoreID uint64, targetStoreIDs []uint64, kind OpKind) (*Operator, error) {
```
rleungx, line 79: Can we use a more general name since it could be used in other scenes?

MrCroxx: Emm, in my first edition implementation, I directly modified `TransferLeader` to support multi targets. But then we found only evict leader needs it, and other ops only needs one target. So I created a seperate `EvictLeader`

MrCroxx: @rleungx Any updates about the naming? I've not come up with a good idea yet.

rleungx: No, but it needs to be general. ^ ^

server/schedule/operator/step.go
```diff
@@ -82,6 +82,70 @@ func (tl TransferLeader) Influence(opInfluence OpInfluence, region *core.RegionI
 	to.LeaderCount++
 }
 
+// EvictLeader is an OpStep that transfers a region's leader to one of the targets.
+type EvictLeader struct {
+	FromStore uint64
+	ToStore   uint64
```
rleungx, line 88: Why do we still need it?

MrCroxx: For compatible, [ref](https://github.com/tikv/pd/pull/4214#issuecomment-947587897).

rleungx: I think we don't need this field here, it can be appended when we send it to TiKV?

MrCroxx:
Duplicated #4401 .

PR #4223, opened 2021-10-19
operator: check store status for running operators

### LINKED ISSUE #3353: PD keeps transfering leader to a down store
### What did you do?

I used nightly build of PD to test joint consensus.

I enabled shuffle region scheduling and set max store down time to 30s. After killing two stores in the same label, a region will stuck and keep 4 replicas until 10min. One example looked like following:
```
"region_id": 129,
"peers": [
    {
      "id": 180171,
      "store_id": 1
    },
    {
      "id": 181496,
      "store_id": 4
    },
    {
      "id": 181653,
      "store_id": 13
    },
    {
      "id": 181850,
      "store_id": 5,
      "role": 1
    }
  ],
  "leader": {
    "id": 180171,
    "store_id": 1
  },
  "down_peers": [
    {
      "peer": {
        "id": 181496,
        "store_id": 4
      },
      "down_seconds": 360
    }
  ],
```
, no pending peer was reported. Store 4 was killed and PD's log kept report 
```
[2021/01/12 12:29:23.613 +08:00] [INFO] [operator_controller.go:626] ["send schedule command"] [region-id=129] [step="transfer leader from store 1 to store 4"] [source="active push"]
```

disksing:

Fix #3353 

- For `TransferLeader`, `AddLearner`, `AddPeer`, check if the target store is down and cancel the operator if need.
- Other types of `OperatorStep` may or may not get blocked by down store, we don't restrict them in the PR.
- Add tests

### Check List

Tests
- Unit test

Related changes

### Release note

```release-note
Fix the issue that operator can get blocked due to down store
```


server/schedule/operator_controller_test.go
```diff
@@ -165,6 +165,25 @@ func (t *testOperatorControllerSuite) TestFastFailOperator(c *C) {
 	c.Assert(oc.GetOperator(region.GetID()), IsNil)
 }
 
+// Issue 3353
+func (t *testOperatorControllerSuite) TestFastFailOperator2(c *C) {
```
HunDunDM, line 169: How about `TestFastFailWithUnhealthyStore`?

HunDunDM (approved):

server/schedule/operator_controller_test.go
disksing: ok

PR #4230, opened 2021-10-21
cluster: handle region after report split

### LINKED ISSUE #4157: Update Region Meta in advance during ReportBatchSplit

After the split occurs in the Region, Heartbeat is relied on to update the Region Meta. The Range hole will be generated during the update, and the Region corresponding to the Range cannot be queried in a short time.

When the PD heartbeat processing is not fast enough, the problem will be magnified.

Update Region Meta in advance during `ReportBatchSplit`

bufferflies:

closed #4157

put region after batch report split 
### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
handle region after report split
```


server/cluster/cluster_worker.go
```diff
@@ -196,6 +196,16 @@ func (c *RaftCluster) HandleReportSplit(request *pdpb.ReportSplitRequest) (*pdpb
 	originRegion := proto.Clone(right).(*metapb.Region)
 	originRegion.RegionEpoch = nil
 	originRegion.StartKey = left.GetStartKey()
+
+	for _, region := range []*metapb.Region{left, right} {
```
JmPotato, line 200: We will do some checks before calling `HandleRegionHeartbeat` in a normal Region heartbeat process, should we do the same thing here?

server/cluster/cluster_worker.go
```diff
@@ -196,6 +196,16 @@ func (c *RaftCluster) HandleReportSplit(request *pdpb.ReportSplitRequest) (*pdpb
...
+		regionInfo := core.NewRegionInfo(region, nil)
+		if err = c.HandleRegionHeartbeat(regionInfo); err != nil {
+			log.Warn("region handler error",
+				zap.Stringer("region-meta", region),
+				errs.ZapError(err))
+		}
+	}
+
```
JmPotato, line 208: I think some metrics are needed to see how many regions and the splitting event is handled here.

server/cluster/cluster_worker.go
bufferflies: I think it does not need to checker, it will call HandleRegionHeartbeat too.

server/cluster/cluster_worker.go
bufferflies: I think the metrics can see grpc metric in pd.

server/cluster/cluster_worker.go
```diff
@@ -196,6 +196,16 @@ func (c *RaftCluster) HandleReportSplit(request *pdpb.ReportSplitRequest) (*pdpb
 	originRegion := proto.Clone(right).(*metapb.Region)
 	originRegion.RegionEpoch = nil
 	originRegion.StartKey = left.GetStartKey()
+
+	for _, region := range []*metapb.Region{left, right} {
+		regionInfo := core.NewRegionInfo(region, nil)
+		if err = c.HandleRegionHeartbeat(regionInfo); err != nil {
```
HunDunDM, line 202: It is not a real heartbeat and lacks a lot of information. I think it would be better to write an updated logic alone.

HunDunDM: like this: https://github.com/tikv/pd/blob/fe6fab9268d2d6fd34cd22edd1cf31a302e8dc5c/server/region_syncer/client.go#L199

server/cluster/cluster_worker.go
```diff
@@ -216,10 +226,17 @@ func (c *RaftCluster) HandleBatchReportSplit(request *pdpb.ReportBatchSplitReque
 	}
 	last := len(regions) - 1
 	originRegion := proto.Clone(regions[last]).(*metapb.Region)
-	hrm = core.RegionsToHexMeta(regions[:last])
 	log.Info("region batch split, generate new regions",
 		zap.Uint64("region-id", originRegion.GetId()),
 		zap.Stringer("origin", hrm),
 		zap.Int("total", last))
+	for _, region := range hrm {
```
rleungx, line 233: Why hex？

server/cluster/cluster_worker.go
```diff
@@ -216,10 +226,17 @@ func (c *RaftCluster) HandleBatchReportSplit(request *pdpb.ReportBatchSplitReque
...
+		regionInfo := core.NewRegionInfo(region, nil)
+		if err = c.HandleRegionHeartbeat(regionInfo); err != nil {
```
rleungx, line 235: I think we need a lightweight way to handle them.

server/cluster/cluster_worker.go
```diff
@@ -216,10 +226,17 @@ func (c *RaftCluster) HandleBatchReportSplit(request *pdpb.ReportBatchSplitReque
...
+			log.Warn("region handler error",
```
rleungx, line 236: The test case is not enough since it won't return an error, only print a log.

server/cluster/cluster.go
```diff
@@ -664,6 +663,58 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+		} else if i == len(regions)-1 {
+			isNew = false
+			r, o := region.GetRegionEpoch(), origin.GetRegionEpoch()
+			if r.GetVersion() > o.GetVersion() || r.GetConfVer() > o.GetConfVer() {
+				update = true
+				needSync = false
+			}
+		}
+		if update {
+			if e := c.saveRegionLock(isNew, true, true, needSync, region, origin); e != nil {
```
rleungx, line 706: looks like it's not a locked function?

server/cluster/cluster.go
```diff
@@ -665,6 +664,58 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+func (c *RaftCluster) processRegionSpiltReport(regions []*metapb.Region) (err error) {
+	if err = checkSplitRegions(regions); err != nil {
+		return err
+	}
+	c.RLock()
+	coreCluster := c.core
+	c.RUnlock()
+	total := len(regions) - 1
+	for i := range regions {
+		// It should update rightmost because it will override by left .
```
JmPotato, line 681:
```suggestion
		// It should update rightmost because it will override by left.
```

server/cluster/cluster.go
```diff
@@ -665,6 +664,58 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+		if perr != nil {
+			err = errors.Wrap(err, perr.Error())
+			continue
+		}
+		update, isNew, needSync := false, false, true
+		// The region that is not the rightmost element should update if origin is nil.
+		// The rightmost region should update if epoch greater than origin.
+		if origin == nil {
+			update = true
+		} else if i == len(regions)-1 {
```
JmPotato, line 698:
```suggestion
		} else if i == total {
```

server/cluster/cluster.go
```diff
@@ -665,6 +664,58 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+		if update {
+			if e := c.saveRegionLock(isNew, true, true, needSync, region, origin); e != nil {
+				err = errors.Wrap(err, e.Error())
+			}
+		}
+	}
+	return err
+}
+
+func (c *RaftCluster) saveRegionLock(isNew, saveKV, saveCache, needSync bool, region *core.RegionInfo, origin *core.RegionInfo) error {
```
JmPotato, line 715: The name `xxxLock` or `xxxLocked` should only be used when it's in the situation that the caller handles the lock.

server/cluster/cluster_worker.go
bufferflies: yes

server/cluster/cluster_worker.go
bufferflies: it will remove

server/cluster/cluster.go
bufferflies: the lock in the fuction.

server/cluster/cluster.go
bufferflies: yes, it will rename as saveRegion

server/cluster/cluster_worker.go
bufferflies: yes, it will simplify the process and remove some update:
1. hotstat should not update.
2. regionGuide will replace with conf or new.

server/cluster/cluster_worker.go
```diff
@@ -182,44 +165,32 @@ func (c *RaftCluster) checkSplitRegions(regions []*metapb.Region) error {
...
+		zap.Uint64("region-id", right.GetId()),
 		logutil.ZapRedactStringer("region-meta", core.RegionToHexMeta(left)))
 	return &pdpb.ReportSplitResponse{}, nil
 }
 
 // HandleBatchReportSplit handles the batch report split request.
 func (c *RaftCluster) HandleBatchReportSplit(request *pdpb.ReportBatchSplitRequest) (*pdpb.ReportBatchSplitResponse, error) {
 	regions := request.GetRegions()
-
-	hrm := core.RegionsToHexMeta(regions)
```
rleungx, line 209: Why remove it?

server/cluster/cluster_worker_test.go
```diff
@@ -17,6 +17,8 @@ package cluster
 import (
 	"context"
 
+	"github.com/gogo/protobuf/proto"
+
```
rleungx, line 21: remove the empty line

server/cluster/cluster_worker_test.go
```diff
@@ -41,28 +43,95 @@ func (s *testClusterWorkerSuite) SetUpTest(c *C) {
...
+	origin.Peers = mockRegionPeer(idAllc, []uint64{1, 2, 3})
+
+	c.Assert(cluster.putRegion(core.NewRegionInfo(origin, nil)), IsNil)
+	old := cluster.GetRegion(2)
+	c.Assert(core.HexRegionKeyStr(old.GetStartKey()), Equals, core.HexRegionKeyStr(origin.GetStartKey()))
+
+	left := &metapb.Region{Id: 1, StartKey: []byte("a"), EndKey: []byte("c"), RegionEpoch: &metapb.RegionEpoch{ConfVer: 1, Version: 1}}
+	right := &metapb.Region{Id: 2, StartKey: []byte("c"), EndKey: []byte("d"), RegionEpoch: &metapb.RegionEpoch{ConfVer: 1, Version: 1}}
+	c.Assert(cluster.GetRegion(1), IsNil)
+	//case1: left endKey should equal to right startKey
```
rleungx, line 76:
```suggestion
	// case1: left endKey should equal to right startKey
```

server/cluster/cluster_worker_test.go
```diff
@@ -41,28 +43,95 @@ func (s *testClusterWorkerSuite) SetUpTest(c *C) {
...
+	_, err = cluster.HandleReportSplit(&pdpb.ReportSplitRequest{Left: proto.Clone(right).(*metapb.Region), Right: proto.Clone(left).(*metapb.Region)})
 	c.Assert(err, NotNil)
+	c.Assert(cluster.GetRegion(1), IsNil)
+	old = cluster.GetRegion(2)
+	c.Assert(old, NotNil)
+	c.Assert(core.HexRegionKeyStr(old.GetStartKey()), Equals, core.HexRegionKeyStr(origin.GetStartKey()))
+	// case1: peer can not be nil
```
rleungx, line 83:
```suggestion
	// case2: peer can not be nil
```

server/cluster/cluster.go
```diff
@@ -699,19 +747,18 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
 	if c.regionStats != nil {
 		c.regionStats.Observe(region, c.getRegionStoresLocked(region))
 	}
-
 	changedRegions := c.changedRegions
-
 	c.Unlock()
-
+	c.RLock()
+	storage := c.storage
```
rleungx, line 760: How about putting it into the lock section above?

server/cluster/cluster.go
```diff
@@ -665,6 +664,55 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+		return err
+	}
+	c.RLock()
+	coreCluster := c.core
+	c.RUnlock()
+	total := len(regions) - 1
+	for i := range regions {
+		// It should update rightmost because it will override by left.
+		v := regions[total-i]
+		if v.Peers == nil {
```
rleungx, line 683: GetPeers

server/cluster/cluster.go
```diff
@@ -665,6 +664,55 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+			err = errors.Wrap(err, "region peer is nil")
```
rleungx, line 684: Will it wrap multiple times?

server/cluster/cluster.go
```diff
@@ -665,6 +664,55 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+			continue
+		}
+		region := core.NewRegionInfo(v, v.Peers[0])
+		origin, perr := coreCluster.PreCheckPutRegion(region)
```
rleungx, line 688: no need lock?

bufferflies: no, it will lock after it needs to  update in saveRegion function.

server/cluster/cluster_worker.go
bufferflies: yes, it will revert this change.

server/cluster/cluster.go
```diff
@@ -665,6 +664,56 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+	total := len(regions) - 1
+	for i := range regions {
+		// It should update rightmost because it will override by left.
+		index := total - i
+		v := regions[index]
+		if v.GetPeers() == nil {
+			err = errors.Wrap(err, fmt.Sprintf("region:%d peer is nil", v.GetId()))
+			continue
+		}
+		region := core.NewRegionInfo(v, v.Peers[0])
```
rleungx, line 688: Why `v.Peers[0]`?

server/cluster/cluster.go
```diff
@@ -665,6 +664,56 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+			err = errors.Wrap(err, perr.Error())
+			continue
+		}
+		update, isNew, needSync := false, false, true
+		// The region that is not the rightmost element should update if origin is nil.
+		// The rightmost region should update if epoch greater than origin.
+		if origin == nil {
+			update = true
+		} else if index == total {
+			isNew = false
```
rleungx, line 700: When `isNew` will be true?

rleungx:
Also, https://github.com/tikv/pd/pull/4230#discussion_r773651498 is not solved.

server/cluster/cluster.go
bufferflies: left's origin is exist or rightmost's conf greater than  origin conf .

server/cluster/cluster.go
bufferflies: It needs to resign one leader to get router but the leader may not elect.

server/cluster/cluster.go
bufferflies: yes, the error message will add region-id.

server/cluster/cluster.go
rleungx: I don't find anywhere to set it to true.

server/cluster/cluster.go
rleungx: But in PD's view, it is the leader.

server/cluster/cluster.go
bufferflies: yes , it's bug.

server/cluster/cluster.go
rleungx: I think it may have something unexpected with the current implementation.

rleungx:
> Also, [#4230 (comment)](https://github.com/tikv/pd/pull/4230#discussion_r773651498) is not solved.

PTAL

bufferflies:
PTAL @rleungx

server/cluster/cluster.go
```diff
@@ -699,19 +762,16 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
 		c.prepareChecker.collect(region)
 	}
 
 	if c.regionStats != nil {
 		c.regionStats.Observe(region, c.getRegionStoresLocked(region))
 	}
-
 	changedRegions := c.changedRegions
-
+	storage := c.storage
```
HunDunDM, line 773: race risk

server/cluster/cluster.go
```diff
@@ -665,6 +664,70 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
...
+		if update {
+			if err := c.saveRegion(isNew, true, true, needSync, region, origin); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	}
+	return errs
+}
+
+func (c *RaftCluster) saveRegion(isNew, saveKV, saveCache, needSync bool, region *core.RegionInfo, origin *core.RegionInfo) error {
```
HunDunDM, line 730: Maybe it is necessary to put `isNew`, `saveKV`, `saveCache`, `needSync` in a `Struct`.

bufferflies:
ref #6867

Issue #4249, opened 2021-10-29
Handle the EOF/Context canceled error caused by TSO Follower Proxy disabling


JmPotato:
### Describe your feature request related problem

Now we can enable/disable TSO Follower Proxy online, however, when we change it from on to off, there are two corner cases that will occur.

#### EOF/Context canceled Error

`c.tryConnectWithProxy` will rebuild the `connectionCtxs` and cancel the old stream, if we have an ongoing TSO gRPC request, this will cause an `EOF/Context canceled` error.

#### Old closed stream chosen

Because we will randomly choose a stream and wait for the incoming TSO request from a channel, so if any `connectionCtxs` rebuilding happens after the stream choosing, it will make the TSO requests in the future fail due to the chosen stream being closed.

These two cases won't affect the TiDB now because it has the TSO backoff strategy. However, I think it's still worth being fixed.

### Describe the feature you'd like

Enabling/Disabling the TSO Follower Proxy online won't affect the TSO request at all. Everything will run smoothly with no error occurring.

PR #4251, opened 2021-10-31
client: fix the case where the chosen TSO stream is canceled later

JmPotato:

Fix a part of https://github.com/tikv/pd/issues/4249.

Moving the stream choosing part behind the TSO batch to make sure that the stream being canceled after being chosen will never happen again.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


nolouch (commented):
ptal @HunDunDM

disksing (approved):

client/client.go
```diff
@@ -753,6 +742,46 @@ func (c *client) handleDispatcher(
 		if maxTSOBatchWaitInterval >= 0 {
 			tbc.adjustBestBatchSize()
 		}
+		// Choose a stream to send the TSO gRPC request.
+	streamChoosingLoop:
```
nolouch, line 746: I think you still cannot guarantee the order, because they are all called concurrently.

JmPotato: This PR is only to solve the case in which the stream is canceled while we are trying to batch the TSO requests. Yeah, it can not guarantee the order between concurrently writing from `UpdateConfig`.

nolouch (approved):

Issue #3453, opened 2021-03-04
Unify the KV interface call


JmPotato:
### Describe your feature request related problem

We have a KV interface definition and some KV storage wraps in [server/kv](https://github.com/tikv/pd/tree/master/server/kv). However, some of the PD codes are using etcd client directly rather than using the KV interface.

### Describe the feature you'd like

We should unify the KV interface call in PD. And furthermore, base on the unified KV storage, we could richer KV storage options for different cases. For example, for the test purpose, we can simply provide a KV storage based on a map struct to speed up the tests and reduce resource consumption.

JmPotato:
After diving into the codes, I think this is a change that we should discuss first. The amount of work is not small.

rleungx:
Perhaps we can expose some APIs for external use so that the other component can use it instead of directly using etcd client.

PR #4278, opened 2021-11-03
*: sort out the usage of core.Storage and kv.Base

JmPotato:

Part of https://github.com/tikv/pd/issues/3453. Trying to sort out the usage of `core.Storage` and `kv.Base`.

#### Background

As https://github.com/tikv/pd/issues/3453 said, we have the unified `kv.Base` and `core.Storage` inside the PD. However, both of them have some design and usage problems.

#### Details

Let's make some basic points before any further discussion.

1. `kv.Base` is the abstraction of the lowest level KV storage, which should provide some fundamental KV operations such as `Load(key)` and `Save(key, value)` etc.
2. `core.Storage` is the abstraction of the concrete `kv.Base` usage. So `kv.Base` should be the backend of a `core.Storage`.
3. The upper service use `core.Storage` to store data and defines some special operations that rely on it.

Base on the three rules, these changes are made.

> Change `core.Storage` into an interface.

Since different storages may have different implementations on the same method, it's better to make `core.Storage` an interface rather than an all-in-one hodgepodge. For example, we have two storages for the region info now, one is etcd and another one is LevelDB. `SaveRegion` is a method that both storages need to implement in different ways.

> Introduce three kinds of storage.

For now, we have three kinds of storage.

* `EtcdStorage`, the storage that stores data in etcd, which is used by the PD server. Most PD data is stored in this one.
* `LevelDBStorage`, the storage that stores data in LevelDB, which is used in the region-related storage.
* `MemoryStorage`, the storage that stores data in a memory B-Tree without any locks, which should only be used in tests.

Specifying a certain storage type will help us use these three storages more properly, which will also make the developer know which storage this service is using while reading the source code.

> Make `RegionStorage` an independent service based on any given storage and the `LevelDBStorage`.

`RegionStorage` should act more like a storage user rather than the storage itself, which is misleading in nowadays code. The related region Load/Save methods are moved into the `RegionStorage`. Also, the switch between the `core.Storage` (usually `EtcdStorage`) is moved into it.

#### Compromises

 * `HotRegionStorage` is still constructed based on the LevelDB KV Base rather than the `LevelDBStorage` due to some special LevelDB operations use, such as `Batch` and `Write`.

#### Review Guide

You can focus on the changes of `server/core/*`, especially the `server/core/storage.go` and `server/core/region_storage.go`. Many other differences are caused by the interface change.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/component/manager.go
+++ b/pkg/component/manager.go
@@ -27,13 +27,13 @@ import (
 // Manager is used to manage components.
 type Manager struct {
 	sync.RWMutex
-	storage *core.Storage
+	storage core.Storage
 	// component -> addresses
 	Addresses map[string][]string `json:"address"`
 }
 
 // NewManager creates a new component manager.
-func NewManager(storage *core.Storage) *Manager {
+func NewManager(storage core.Storage) *Manager {
 	return &Manager{
 		storage:   storage,
 		Addresses: make(map[string][]string),
--- a/pkg/component/manager_test.go
+++ b/pkg/component/manager_test.go
@@ -20,7 +20,6 @@ import (
 
 	. "github.com/pingcap/check"
 	"github.com/tikv/pd/server/core"
-	"github.com/tikv/pd/server/kv"
 )
 
 func Test(t *testing.T) {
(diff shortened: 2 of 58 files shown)


disksing:
The diff is too big. Need split.

JmPotato:
Close due to moving to https://github.com/tikv/pd/pull/4335.

PR #4294, opened 2021-11-05
*: use context to manage RaftCluster

### LINKED ISSUE #4293: Only use context to manage RaftCluster
## Development Task

Currently, there are both `ctx context.Context` and `quit chan struct{}` in `RaftCluster` to control the start and stop, where `ctx` is actually `Server.ctx`. Currently, it has the following problems:

* The control sub-module needs to pass parameters `ctx` and `quit` at the same time
* The `suspectRegions` and `suspectKeyRanges` only use `ctx` without `quit`, which leads to leaks after changing from Leader to Follower.
* `RegionSyncer` did not make full use of quit to terminate the service as soon as possible.

We can solve them by combining `ctx` and `quit`.

HunDunDM:

close #4293

* `RaftCluster` uses its own `ctx` to replace the original `quit`.
* `RegionSyncer` can stop immediately when `ctx` is `Done` during service.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/server/region_syncer/server.go
+++ b/server/region_syncer/server.go
@@ -43,6 +43,7 @@ const (
 )
 
 var (
+	// ErrNotStarted is returned when current server is starting.
 	ErrNotStarted = status.Errorf(codes.Unavailable, "server not started")
 )
 
--- a/server/region_syncer/server.go
+++ b/server/region_syncer/server.go
@@ -42,11 +42,6 @@ const (
 	defaultHistoryBufferSize = 10000
 )
 
-var (
-	// ErrNotStarted is returned when current server is starting.
-	ErrNotStarted = status.Errorf(codes.Unavailable, "server not started")
-)
-
 // ClientStream is the client side of the region syncer.
 type ClientStream interface {
 	Recv() (*pdpb.SyncRegionResponse, error)
@@ -78,7 +73,6 @@ type RegionSyncer struct {
 		streams      map[string]ServerStream
 		clientCtx    context.Context
 		clientCancel context.CancelFunc
-		serverCtx    context.Context
 	}
 	server    Server
 	wg        sync.WaitGroup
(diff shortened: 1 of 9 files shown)


server/region_syncer/server.go
```diff
@@ -42,6 +42,11 @@ const (
 	defaultHistoryBufferSize = 10000
 )
 
+var (
+	// ErrNotStarted is returned when current server is starting.
+	ErrNotStarted = status.Errorf(codes.Unavailable, "server not started")
```
rleungx, line 47: How about reducing the duplicated definision?

server/region_syncer/server.go
```diff
@@ -148,7 +166,21 @@ func (s *RegionSyncer) RunServer(regionNotifier <-chan *core.RegionInfo, quit ch
 // Sync firstly tries to sync the history records to client.
 // then to sync the latest records.
 func (s *RegionSyncer) Sync(stream pdpb.PD_SyncRegionsServer) error {
+	s.mu.RLock()
+	ctx := s.mu.serverCtx
+	s.mu.RUnlock()
+
+	if ctx == nil {
```
rleungx, line 173: When will it become nil?

HunDunDM: `RaftCluster` is not started.

server/region_syncer/server.go
HunDunDM: I didn't think of a good way. Maybe only a normal error is returned in region_syncer?

server/region_syncer/server.go
rleungx: Why not move this check to server so that we can avoid the duplicated definition.

rleungx (approved):

nolouch (approved):
lgtm

PR #4298, opened 2021-11-07
Add show command under pd-ctl scheduler config

### LINKED ISSUE #3951: pd-ctl: support `show` and `list` in `pd-ctl config` and `pd-ctl scheduler config`
image
image

It bothers me.

Supported `show` and `list` at the same time.

Go

matchge-ca:
Signed-off-by: Hua Lu <hua.lu@pingcap.com>

Resolves #3951 

- Add new command `show` under `pd-ctl scheduler config balance-hot-region-scheduler` with same behavior as `list`
- Add deprecated warning message when execute `list` command under `pd-ctl scheduler config balance-hot-region-scheduler`
- Modify short description of `list` with deprecate message.
- Add test case for `show` command, usage the same logic as `list` command

### Check List
Tests
- Unit test
   - Compare outputs from `pd-ctl scheduler config balance-hot-region-scheduler show` and `pd-ctl scheduler config balance-hot-region-scheduler list`
   - Check `pd-ctl scheduler config balance-hot-region-scheduler list` has deprecated message printed.

Code changes
- Has configuration change

Related changes

### Release note
```release-note
None
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/scheduler.go
+++ b/tools/pd-ctl/pdctl/command/scheduler.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
 	"github.com/spf13/cobra"
 	"github.com/tikv/pd/server/schedulers"
 )
@@ -423,14 +424,22 @@ func newConfigHotRegionCommand() *cobra.Command {
 		Short: "balance-hot-region-scheduler config",
 		Run:   listSchedulerConfigCommandFunc,
 	}
+
+	// Deprecated: 'list' command will be deprecated in future version, use 'show' command instead.
 	c.AddCommand(&cobra.Command{
 		Use:   "list",
+		Short: "list the config item (will be deprecated in feature version, use show command instead)",
+		Run:   listSchedulerConfigCommandFunc,
+	}, &cobra.Command{
+		Use:   "show",
 		Short: "list the config item",
-		Run:   listSchedulerConfigCommandFunc})
-	c.AddCommand(&cobra.Command{
+		Run:   listSchedulerConfigCommandFunc,
+	}, &cobra.Command{
 		Use:   "set <key> <value>",
 		Short: "set the config item",
-		Run:   func(cmd *cobra.Command, args []string) { postSchedulerConfigCommandFunc(cmd, c.Name(), args) }})
+		Run:   func(cmd *cobra.Command, args []string) { postSchedulerConfigCommandFunc(cmd, c.Name(), args) },
+	})
+
 	return c
 }
 
@@ -512,6 +521,9 @@ func listSchedulerConfigCommandFunc(cmd *cobra.Command, args []string) {
 	}
 	p := cmd.Name()
 	if p == "list" {
+		p = cmd.Parent().Name()
+		log.Info("The list command will be deprecated in future version, use show command instead")
+	} else if p == "show" {
 		p = cmd.Parent().Name()
 	}
 	path := path.Join(schedulerConfigPrefix, p, "list")
--- a/tools/pd-ctl/pdctl/command/scheduler.go
+++ b/tools/pd-ctl/pdctl/command/scheduler.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 
 	"github.com/pingcap/errors"
-	"github.com/pingcap/log"
 	"github.com/spf13/cobra"
 	"github.com/tikv/pd/server/schedulers"
 )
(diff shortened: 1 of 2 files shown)


tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -423,14 +424,22 @@ func newConfigHotRegionCommand() *cobra.Command {
 		Short: "balance-hot-region-scheduler config",
 		Run:   listSchedulerConfigCommandFunc,
 	}
+
+	// Deprecated: 'list' command will be deprecated in future version, use 'show' command instead.
 	c.AddCommand(&cobra.Command{
 		Use:   "list",
+		Short: "list the config item (will be deprecated in feature version, use show command instead)",
+		Run:   listSchedulerConfigCommandFunc,
```
rleungx, line 432: How about adding `Deprecated`?

lhy1024 (approved):

nolouch:
ptal @HunDunDM

tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -423,14 +424,22 @@ func newConfigHotRegionCommand() *cobra.Command {
...
+	}, &cobra.Command{
```
bufferflies, line 433: how about use ‘Aliases’ to replase it.

tools/pd-ctl/pdctl/command/scheduler.go
matchge-ca: Something like this?
`"list the config item (Deprecated: list command will be deprecated in future version, use show command instead.)"`

tools/pd-ctl/pdctl/command/scheduler.go
matchge-ca: Below is the comparison of `-h` output between command and alias. As you can see, alias will not be printed in help section which could cause confusion.

Define `show` as alias of `list`
```
Available Commands:
  list        list the config item (Deprecated: will be deprecated in feature version, use show command instead)
  set         set the config item
```

Define `show` as command:
```
Available Commands:
  list        list the config item (Deprecated: will be deprecated in feature version, use show command instead)
  set         set the config item
  show        list the config item
```

tools/pd-ctl/pdctl/command/scheduler.go
rleungx: `cobra.Command` has a `Deprecated` field and you can refer to `NewStoresCommand()`.

matchge-ca: I see, but I have few concerns using `Deprecated` field at the moment:
- The `list` command has not been deprecated yet, most likely will be in next major release which means the `Deprecated` field should be added then
- The command with `Deprecated` field will be hidden which indicates an incorrect impression to users.
  ``` 
  Available Commands:
    set         set the config item
    show        list the config item
  ```
- The command with `Deprecated` field will print extra message to screen which could break script if it is only expect the json output.
  ```
  Command "list" is deprecated, use show instead
  ```

rleungx: Ok, SGTM. First, we should change the document to let users know when we going to make it deprecated. Then we can mark it deprecated.

tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -513,6 +522,9 @@ func listSchedulerConfigCommandFunc(cmd *cobra.Command, args []string) {
 	p := cmd.Name()
 	if p == "list" {
 		p = cmd.Parent().Name()
+		log.Info("The list command will be deprecated in future version, use show command instead")
```
rleungx, line 525: Can we directly print instead of using log?

tools/pd-ctl/pdctl/command/scheduler.go
matchge-ca: Yes, that is why the short description contains the message `list command will be deprecated in future version`. I will create an issue to update the document.

tools/pd-ctl/pdctl/command/scheduler.go
HunDunDM: like `cmd.Println`?

matchge-ca: As discussed with @rleungx, print message directly is similar to adding `Deprecated` field to the command which could break a script if it only expects the json output. 
```
The list command will be deprecated in future version, use show command instead
```
It is better to not change the command behavior before deprecate it, thus `log.Info` should be removed as well.

rleungx (approved):

PR #4309, opened 2021-11-09
server: Application-layer ACL evaluation and support for SAN-based TLS client authentication

LINKIWI:

This change adds the ability to verify TLS clients based on their SAN values (DNS names and URIs), rather than just CN. Additionally, it implements ACL evaluation in the application/HTTP layer rather than deferring to etcd, which performs this verification at handshake time. This implementation makes it more consistent with the behavior of TiKV.

* Add a `cert-allowed-san` field to configuration
* Add a new TLS authentication middleware that evaluates client certificate identities before passing it to the next handler
* Apply the above middleware at the root handler for `/pd/api/`
* Remove the code that passes `cert-allowed-cn` to embedded etcd, in favor of evaluating it in the new authentication middleware.

### Check List

Tests

[ ] Unit test
[ ] Integration test
[x] Manual test (add detailed scripts or steps below)
[ ] No code

I will add a unit test after getting some general feedback on the direction of the approach and whether this patch is something that is desired for upstream.

Manual test:

Case 1: Start PD with config field `cert-allowed-san = ["invalid"]` and curl `/pd/api/v1/members`. Observe HTTP 401 with body `no matching allowed CNs or SANs in client certificate`.
Case 2: Start PD with config field `cert-allowed-san = ["<hostname>"]` and `/pd/api/v1/members`. Observe HTTP 200.

Code changes

[x] Has configuration change
[ ] Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
[ ] Has persistent data change

Side effects

[ ] Possible performance regression
[ ] Increased code complexity
[ ] Breaking backward compatibility

Potential concern: today, client certificate validation happens at transport/handshake layer via embedded etcd. If the CN does not match, PD will fail the TLS handshake and clients will see a connection error. This change moves the evaluation to application layer (within PD itself, rather than etcd), causing the handshake to succeed but for clients to receive an HTTP 401 response.

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn): I will file a PR to update the docs if this implementation approach is accepted.

Related TiKV change: tikv/tikv#11164

### Release note

```release-note
server: Application-layer ACL evaluation and support for SAN-based TLS client authentication
```


nolouch (commented):
Hi, @LINKIWI Thanks for your contribution.
the modification is only on the HTTP layer, can the grpc layer still work?

LINKIWI:
Thanks for taking a look. Correct, this change should only apply to the HTTP layer, but this is something I need to double check.

Is the expected behavior for `cert-allowed-cn` that it applies to both the HTTP and gRPC layers, or just one? I can update the implementation as needed to conform to the expected behavior.

nolouch:
Hi, @LINKIWI, I think it should be both.

JmPotato:
@LINKIWI Hi, sorry to bother you, do you have any updates on this PR?

LINKIWI:
@JmPotato Sorry for the delay, I've been caught up with some other work and just came back from vacation. I'm planning to continue work on this PR this week and will post an update here with test results once I have them.

PR #4335, opened 2021-11-16
*: introduce the new Storage interface

JmPotato:

Part of #3453. Trying to sort out the usage of `core.Storage` and `kv.Base`.

~~This is a sub-PR of https://github.com/tikv/pd/pull/4278~~, which mainly focuses on the new `Storage` interface introducing.

For more details, please read the PR description of https://github.com/tikv/pd/pull/4278.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


disksing (approved):

server/core/storage_v2.go
```diff
@@ -0,0 +1,612 @@
...
+	"github.com/tikv/pd/pkg/encryption"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/encryptionkm"
+	"github.com/tikv/pd/server/kv"
+	"go.etcd.io/etcd/clientv3"
+)
+
+// StorageV2 is the interface for the backend storage of the PD.
+// TODO: replace the core.Storage with this interface later.
+type StorageV2 interface {
```
nolouch, line 41: I don't think it is good that a storage interface has so many methods.

JmPotato: Another approach is to define these methods in the components that actually need them, rather than in one centralized place. What do you think?

JmPotato: For example, we can define `LoadReplicationStatus` inside the `replication` package with an independent interface rather than all in the `Storage`.

JmPotato: Or we can just split this interface into several different sub-interfaces.

JmPotato: After some experimentation, I think splitting Interface is more feasible. PTAL.

rleungx: From my perspective, Storage should not take care of the detailed implementation of PD.

JmPotato: @rleungx You are right and I agree with that, but as I said in #4278, `kv.Base` should be the role that should not take care of the detailed implementation of PD. `Storage` is the abstraction of the concrete `kv.Base` usage. So `kv.Base` should be the backend of a `core.Storage`. It's ok to make `core.Storage` full of detailed implementations because `kv.Base` is closer to the role you are talking about.

server/core/storage_v2.go
```diff
@@ -0,0 +1,623 @@
...
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/encryption"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/encryptionkm"
+	"github.com/tikv/pd/server/kv"
+	"go.etcd.io/etcd/clientv3"
+)
+
+// StorageV2 is the interface for the backend storage of the PD.
+// V2 is just a temp name, we will rename it to Storage eventually.
```
rleungx, line 39: How about directly using Storage

JmPotato: This PR is a sub-PR of #4278, it will be changed to `Storage` once we deleted the old storage struct.

server/core/storage_v2.go
```diff
@@ -0,0 +1,623 @@
...
+// TODO: replace the core.Storage with this interface later.
+type StorageV2 interface {
+	ConfigStorage
+	MetaStorage
+	RuleStorage
+	ComponentStorage
+	ReplicationStatusStorage
+	GCSafePointStorage
```
rleungx, line 47: It's a little bit specific.

server/core/storage_v2.go
```diff
@@ -0,0 +1,623 @@
...
+	return s.Remove(configPath)
+}
+
+// MetaStorage defines the storage operations on the PD cluster meta info.
+type MetaStorage interface {
+	LoadMeta(meta *metapb.Cluster) (bool, error)
+	SaveMeta(meta *metapb.Cluster) error
+	LoadStore(storeID uint64, store *metapb.Store) (bool, error)
+	SaveStore(store *metapb.Store) error
+	SaveStoreWeight(storeID uint64, leader, region float64) error
```
rleungx, line 117: Can we use SaveStore to replace it?

server/core/storage_v2.go
rleungx: If we do need them, I prefer to use an individual package and separate them into different files.

server/core/storage_v2.go
```diff
@@ -0,0 +1,623 @@
...
+)
+
+// StorageV2 is the interface for the backend storage of the PD.
+// V2 is just a temp name, we will rename it to Storage eventually.
+// TODO: replace the core.Storage with this interface later.
+type StorageV2 interface {
+	ConfigStorage
+	MetaStorage
+	RuleStorage
+	ComponentStorage
```
rleungx, line 45: It's not used anymore.

server/core/storage_v2.go
JmPotato: It seems that they are saving different things. So having two methods is reasonable to me.

server/core/storage_v2.go
JmPotato: It contains some specific `GCSafePoint` handling logic, it's hard to make it a generic definition I think.

server/core/storage_v2.go
JmPotato: I think the deprecating work can be done later.

server/core/storage_v2.go
JmPotato: I split them into the `server/storage/base/*` and only expose the final `Storage` interface, etcd, and memory storage to the outside. PTAL.

server/core/storage_v2.go
rleungx: It just clones a store and uses SaveStore which I think we don't need in the interface.

server/storage/base/storage.go
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+package base
+
+import (
+	"github.com/tikv/pd/server/encryptionkm"
+	"github.com/tikv/pd/server/kv"
+)
+
+const (
```
rleungx, line 22: move them into meta?

pkg/component/manager_test.go
```diff
@@ -32,7 +31,7 @@ var _ = Suite(&testManagerSuite{})
 type testManagerSuite struct{}
 
 func (s *testManagerSuite) TestManager(c *C) {
-	m := NewManager(core.NewStorage(kv.NewMemoryKV()))
+	m := NewManager(storage.NewMemoryStorage())
```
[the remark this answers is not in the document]
nolouch: or use builder look likes `NewStorageBuilder.WithEtcdBackend.Build()`.

nolouch (commented):
Overall LGTM, Great works.

pkg/component/manager_test.go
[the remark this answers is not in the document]
JmPotato: Good idea. I think changing it to `XXXStorageBackend` is better.

server/storage/etcd_storage.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/tikv/pd/server/kv"
+	"github.com/tikv/pd/server/storage/base"
+	"go.etcd.io/etcd/clientv3"
+)
+
+var _ Storage = (*EtcdStorage)(nil)
+
+// EtcdStorage is a storage that stores data in etcd,
+// which is used by the PD server.
+type EtcdStorage struct {
```
nolouch, line 27: Can`EtcdStoage` merge with `MemoryStorage` as `Storage`? I think etcd or memdb is the backend kv engine, and the storage is level higher than them.

JmPotato: After this PR, I will try to use `EtcdStorage` and `LevelDBStorage` (maybe `BadgerDBStorage` later) to be the backend of the region storage, which will make it an independent storage component. In order to switch the backend storage smoothly, there is a need to implement different `LoadRegion`/`SaveRegion`/`DeleteRegion` according to different storage backends. For now, it's implemented in different places, the etcd implementation is in `core.Storage`, the LevelDB one is in `core.RegionStorage`. Keeping `EtcdStorage` and `LevelDBStorage` separate is a good way to make different implementations for the same method.

server/storage/builder.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// WithEtcdBackend sets the backend of the storage to etcd with the given client.
+func (sb *Builder) WithEtcdBackend(client *clientv3.Client) *Builder {
+	sb.client = client
+	return sb
+}
+
+// WithMemoryBackend sets the backend of the storage to memory.
+func (sb *Builder) WithMemoryBackend() *Builder {
```
rleungx, line 53: Can we create a `kv` in this function?

server/storage/etcd_storage.go
nolouch: This is indeed a problem, the main problem is that region storage is special... But I think a good abstraction is that I don't need to be concerned with what the underlying backend engine is.  I just know there is one storage, the storage can save/load config, rules, regions and so on. we may make region storage behavior unify for the all backend firstly, such as we can add a `WriteBatch` interface for kv.Base etc.

nolouch: The other path is what you do now, to separate them. But several concepts confuse me. As I mentioned above, ·MemoryStorage`, `EtcdStorage`,`LevelDBStroage` are different with `ConfigStorage`, `RuleStorage`... but they are all called Storage.

server/cluster/cluster.go
```diff
@@ -69,7 +71,7 @@ type Server interface {
 	GetAllocator() id.Allocator
 	GetConfig() *config.Config
 	GetPersistOptions() *config.PersistOptions
-	GetStorage() *core.Storage
+	GetStorage() storage.Storage /* NOTICE: this method will be renamed to GetStorage later. */
```
rleungx, line 74: It has already been GetStorage?

server/storage/memory_backend.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2021 TiKV Project Authors.
```
rleungx, line 1:
```suggestion
// Copyright 2022 TiKV Project Authors.
```

server/storage/base_backend/component.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package backend
+
+import (
+	"encoding/json"
+
+	"github.com/pingcap/errors"
+	"github.com/tikv/pd/pkg/errs"
+	storage "github.com/tikv/pd/server/storage/base_storage"
```
rleungx, line 22: `base_storage` is confused.

server/storage/endpoint/config.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package endpoint
+
+// ConfigStorage defines the storage operations on the config.
```
nolouch, line 17: I prefer move the implements to the `endpoint` package with the interface together.  and remove `bask_bankend` packages, such as:

```
-- endpoint.go
type StorageEndpoint struct {
	kv.Base
	encryptionKeyManager *encryptionkm.KeyManager
}

-- config.go
// ConfigStorage defines the storage operations on the config.
type ConfigStorage interface {
	LoadConfig(cfg interface{}) (bool, error)
	SaveConfig(cfg interface{}) error
	LoadAllScheduleConfig() ([]string, []string, error)
	SaveScheduleConfig(scheduleName string, data []byte) error
	RemoveScheduleConfig(scheduleName string) error
}

func (se * StorageEndpoint) LoadConfig(cfg interface{}) (bool, error) {...}
func (se * StorageEndpoint) SaveConfig(cfg interface{}) error

-- etcd_backend.go
type etcdBackendStorage struct {
	*endpoint.StorageEndpoints
}

// newEtcdBackendStorage is used to create a new etcd backend.
func newEtcdBackendStroage(client *clientv3.Client, rootPath string) *etcdBackend {
	return &etcdBackendStorage{endpoint.NewStorageEndpoint(kv.NewEtcdKVBase(client, rootPath), nil)}
}

....

```
cc @rleungx

rleungx: Maybe put it into an individual file instead of `config.go`

rleungx: BTW, we can use the package name to simplify the naming, like `config.Storage`

nolouch (approved):
lgtm!

nolouch:
ptal @rleungx

server/storage/storage.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx context.Context,
+	rootPath string,
+	ekm *encryptionkm.KeyManager,
+) (*levelDBBackend, error) {
+	return newLevelDBBackend(ctx, rootPath, ekm)
+}
+
+// TODO: support other KV storage backends like BadgerDB in the future.
+
+type pdStorage struct {
```
rleungx, line 64: prefer to use another name

JmPotato: This storage contains both etcd and LevelDB to save meta and region info in most cases. I think it's OK to make its name start with "pd" to indicate the special use by the `pd-server`.

rleungx (approved):

PR #4338, opened 2021-11-17
Mark failed stores as tombstones before executing unsafe recovery

v01dstar:

To avoid users restarting failed stores after [unsafe recovery](https://github.com/tikv/tikv/issues/10483) which may put the cluster into a conflicting state, forcibly mark failed stores as tombstones at the beginning of [unsafe recovery](https://github.com/tikv/tikv/issues/10483).

Unsafe Recovery Issue: https://github.com/tikv/tikv/issues/10483

Once a `remove-failed-stores` request is received, directly put failed stores (provided in the request) into tombstone state.

### Check List

Tests

- [x] Unit test

Code changes

N/A

Side effects

N/A

Related changes

N/A

### Release note

```release-note
Mark failed stores as tombstones before executing unsafe recovery
```


server/cluster/cluster.go
```diff
@@ -1114,6 +1114,37 @@ func (c *RaftCluster) buryStore(storeID uint64) error {
 	return err
 }
 
+// ForceRemoveStore set a store as tombstone directly without waiting for the data to be moved somewhere else.
+func (c *RaftCluster) ForceRemoveStore(storeID uint64) error {
```
rleungx, line 1118: Can we use `BuryStore`?

server/cluster/cluster.go
```diff
@@ -1114,6 +1114,37 @@ func (c *RaftCluster) buryStore(storeID uint64) error {
...
+	c.Lock()
+	defer c.Unlock()
+
+	store := c.GetStore(storeID)
+	if store == nil {
+		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
+	}
+
+	// Bury a tombstone store should be OK, nothing to do.
+	if store.IsTombstone() {
```
Connor1996, line 1128: and check if the store is up now, if that, reject to remove the store

v01dstar: I intentionally created a new function (in addition to `buryStore`) which does not check if the store is up or not. Because, it seems that, whether or not a store is [up or offline](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L150) is totally set by the user/admin, while the background liveness check sets [a different state here](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L475). If we really want to check the liveness of a store, we probably should check `IsDisconnected()`. Updated, PTAL

v01dstar: I intentionally created a new function (in addition to `buryStore`) which does not check if the store is up or not. Because, it seems that, whether or not a store is [up or offline](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L150) is totally set by the user/admin, while the background liveness check sets [a different state here](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L475). If we really want to check the liveness of a store, we probably should check `IsDisconnected()`. Updated, PTAL

server/cluster/cluster.go
v01dstar: I intentionally created a new function, because, given the fact that [offline](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L150) does not mean [disconnected](https://github.com/tikv/pd/blob/3448e55fc5e7974e261a282f66204379e70a7855/server/core/store.go#L475), bury a store that is "up" is not allowed by `buryStore`, but is allowed in unsafe recovery.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -80,7 +80,7 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
 		return errors.Errorf("Another request is working in progress")
 	}
 	u.reset()
-	for failedStore, _ := range failedStores {
+	for failedStore := range failedStores {
```
Connor1996, line 83: make sure all stores are legal and then remove them

server/cluster/cluster.go
rleungx: You can take a look at the implementation of `BuryStore` in release 4.0. We can add a flag to control the behavior instead of creating a new function.

rleungx: BTW, what if we only have a small number of regions that need to be recovered? e.g. we can use `-r` in tikv-ctl.

rleungx: For this kind of operation, we must give a prompt to the users to let the users confirm their operations.

Connor1996: > BTW, what if we only have a small number of regions that need to be recovered? e.g. we can use `-r` in tikv-ctl.

@v01dstar Seems, we should support it later

v01dstar: > BTW, what if we only have a small number of regions that need to be recovered? e.g. we can use `-r` in tikv-ctl.

Not sure what concerns you in this case, could you be more specific?

I have updated the implementation which reuses the `BuryStore` function, PTAL

server/cluster/unsafe_recovery_controller.go
v01dstar: Done. PTAL.

server/cluster/cluster.go
rleungx: > Not sure what concerns you in this case, could you be more specific?

In this case, we don't need delete store.

v01dstar: It is still better to remove/bury/delete store even if a small number of regions are affected, right? e.g. in "ONCALL-3811", several regions were not able to elect leaders since replicas in the same group had conflicting states, which we believe was due to the restarting of failed stores. This PR prevent such things from happening by marking failed stores as tombstones. Not deleting the store means the store can still be booted? In that case, unsafe recovery should not be used at the first place?

rleungx: What if there are some bugs which lead to some regions unavailable? There is no need to delete the store.

v01dstar: Was there an existed issue that required using unsafe recovery to mitigate while having a store to be kept alive? I thought we only use this to `remove failed stores` right?

rleungx: For example: ONCALL-1659

v01dstar: I see. Currently, Online unsafe recovery does not support removing failed stores from a specific region (or `-r` option). If we are going to implement that in the future we will make sure we don't mark the store as tombstone, but so far I don't think that's a problem, admin or oncall should use online unsafe recovery if the entire store is broken, and still use tikv ctl unsafe recovery if they want to remove a peer from a specific group.

Connor1996 (approved):
LGTM

server/cluster/unsafe_recovery_controller.go
```diff
@@ -79,6 +80,18 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
 		return errors.Errorf("Another request is working in progress")
 	}
 	u.reset()
+	for failedStore := range failedStores {
+		store := u.cluster.GetStore(failedStore)
+		if store != nil && store.IsUp() && !store.IsDisconnected() {
+			return errors.Errorf("Store %v is up and connected", failedStore)
+		}
+	}
+	for failedStore := range failedStores {
```
rleungx, line 89: can we do it in the above loop?

v01dstar: They are intentionally separated since we want to bury stores if and only if all of them are able to be buried.

server/cluster/cluster.go
```diff
@@ -1095,7 +1094,11 @@ func (c *RaftCluster) buryStore(storeID uint64) error {
 	}
 
 	if store.IsUp() {
-		return errs.ErrStoreIsUp.FastGenByArgs()
+		if !forceBury {
+			return errs.ErrStoreIsUp.FastGenByArgs()
+		} else if !store.IsDisconnected() {
```
rleungx, line 1099: I think we don't need this condition.

v01dstar: This is also part of the design, we are encouraging our customers only use this feature only when the store is broken and unrecoverable, which means, if a store is up and connected, we disallow the operation.

rleungx (approved):

Issue #4348, opened 2021-11-18
unstable test `regionSyncerTestSuite.TestRegionSyncer`


nolouch:
## Development Task
`regionSyncerTestSuite.TestRegionSyncer` easy to failed.
details: https://ci2.pingcap.net/blue/organizations/jenkins/pd_test/detail/pd_test/3446/pipeline

HunDunDM:
ref #4294

rleungx:
https://ci2.pingcap.net/blue/organizations/jenkins/pd_test/detail/pd_test/3506/ happened again

nolouch:
ref https://github.com/tikv/pd/pull/4360

rleungx:
Closed via #4360.

PR #4360, opened 2021-11-22
server: make TestRegionSynce stable

nolouch:

ref #4348

add more failpoint

### Check List

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/region_syncer/server.go
+++ b/server/region_syncer/server.go
@@ -156,8 +156,8 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 // GetAllDownstreamNames tries to get the all bind stream's name.
 // Only for test
 func (s *RegionSyncer) GetAllDownstreamNames() []string {
-	var names []string
 	s.mu.RLock()
+	names := make([]string, 0, len(s.mu.streams))
 	for name := range s.mu.streams {
 		names = append(names, name)
 	}
--- a/server/region_syncer/server.go
+++ b/server/region_syncer/server.go
@@ -153,6 +153,18 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 	}
 }
 
+// GetAllDownstreamNames tries to get the all bind stream's name.
+// Only for test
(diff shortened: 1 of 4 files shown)


HunDunDM (approved):

server/region_syncer/server.go
```diff
@@ -153,6 +153,18 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 	}
 }
 
+// GetAllDownstreamNames tries to get the all bind stream's name.
+// Only for test
+func (s *RegionSyncer) GetAllDownstreamNames() []string {
+	var names []string
```
JmPotato, line 159: Better initialize it with a capability.

server/region_syncer/server.go
```diff
@@ -153,6 +153,18 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
...
+	s.mu.RLock()
+	for name := range s.mu.streams {
+		names = append(names, name)
+	}
```
JmPotato, line 163: What about using `copy()`?

tests/cluster.go
```diff
@@ -534,6 +534,31 @@ func (c *TestCluster) WaitLeader(ops ...WaitOption) string {
...
+	for i := 0; i < option.retryTimes; i++ {
+		name := c.GetLeader()
+		if len(name) == 0 {
+			time.Sleep(option.waitInterval)
+			continue
+		}
+		leaderServer := c.GetServer(name)
+		clus := leaderServer.GetServer().GetRaftCluster()
+		if clus != nil {
+			if len(clus.GetRegionSyncer().GetAllDownstreamNames()) == n {
```
JmPotato, line 555: Since we only use its length, why bother copying the whole slice?

nolouch: I think it is more clear to return the name of the current downstream than to return only the number, although it is only used in test now.

server/region_syncer/server.go
nolouch: this is a map, do you mean copy map?

PR #4364, opened 2021-11-23
server: extract GrpcServer from Server

### LINKED ISSUE #4363: Extract `GrpcServer` from `Server`
## Development Task

Now some of the methods of `Server` are used as the gRPC server, some of them are used internally. Sometimes we want to declare different versions of the method to allow it to be used in both ways. We can extract GrpcServer from Server to make it easy to distinguish methods use internally or for public.

disksing:

close #4363

Define `GrpcServer` to wrap `Server` into a grpc service.

### Check List

Tests
- Unit test

### Release note
```release-note
None
```

### DIFF
--- a/server/api/member.go
+++ b/server/api/member.go
@@ -15,7 +15,6 @@
 package api
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -59,9 +58,8 @@ func (h *memberHandler) ListMembers(w http.ResponseWriter, r *http.Request) {
 	h.rd.JSON(w, http.StatusOK, members)
 }
 
-func getMembers(svr *server.Server) (*pdpb.GetMembersResponse, error) {
-	req := &pdpb.GetMembersRequest{Header: &pdpb.RequestHeader{ClusterId: svr.ClusterID()}}
-	members, err := svr.GetMembers(context.Background(), req)
+func getMembers(svr *server.Server) ([]*pdpb.Member, error) {
+	members, err := svr.GetMembers()
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
@@ -69,7 +67,7 @@ func getMembers(svr *server.Server) (*pdpb.GetMembersResponse, error) {
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	for _, m := range members.GetMembers() {
+	for _, m := range members {
 		m.DcLocation = ""
 		binaryVersion, e := svr.GetMember().GetMemberBinaryVersion(m.GetMemberId())
 		if e != nil {
@@ -228,7 +226,7 @@ func (h *memberHandler) SetMemberPropertyByName(w http.ResponseWriter, r *http.R
 
 	var memberID uint64
 	name := mux.Vars(r)["name"]
-	for _, m := range members.GetMembers() {
+	for _, m := range members {
 		if m.GetName() == name {
 			memberID = m.GetMemberId()
 			break
--- a/server/api/member.go
+++ b/server/api/member.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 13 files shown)


server/grpc_service.go
```diff
@@ -1451,7 +1453,7 @@ var mockLocalAllocatorLeaderChangeFlag = false
 
 // SyncMaxTS will check whether MaxTS is the biggest one among all Local TSOs this PD is holding when skipCheck is set,
 // and write it into all Local TSO Allocators then if it's indeed the biggest one.
-func (s *Server) SyncMaxTS(ctx context.Context, request *pdpb.SyncMaxTSRequest) (*pdpb.SyncMaxTSResponse, error) {
+func (s *GrpcServer) SyncMaxTS(ctx context.Context, request *pdpb.SyncMaxTSRequest) (*pdpb.SyncMaxTSResponse, error) {
```
JmPotato, line 1456: This method is used internally also, could you please help move this into `Server`?

server/grpc_service.go
```diff
@@ -1556,12 +1558,12 @@ func (s *Server) SplitRegions(ctx context.Context, request *pdpb.SplitRegionsReq
...
 
 // SplitAndScatterRegions split regions by the given split keys, and scatter regions
-func (s *Server) SplitAndScatterRegions(ctx context.Context, request *pdpb.SplitAndScatterRegionsRequest) (*pdpb.SplitAndScatterRegionsResponse, error) {
+func (s *GrpcServer) SplitAndScatterRegions(ctx context.Context, request *pdpb.SplitAndScatterRegionsRequest) (*pdpb.SplitAndScatterRegionsResponse, error) {
 	panic("unimplemented")
 }
 
 // GetDCLocationInfo gets the dc-location info of the given dc-location from PD leader's TSO allocator manager.
-func (s *Server) GetDCLocationInfo(ctx context.Context, request *pdpb.GetDCLocationInfoRequest) (*pdpb.GetDCLocationInfoResponse, error) {
+func (s *GrpcServer) GetDCLocationInfo(ctx context.Context, request *pdpb.GetDCLocationInfoRequest) (*pdpb.GetDCLocationInfoResponse, error) {
```
JmPotato, line 1566: Ditto.

JmPotato (commented):
The rest LGTM.

server/grpc_service.go
disksing: Sure. I noticed it after pushing the PR.

disksing: @JmPotato Found that they are not used internally. The `allocator manager` does have a method with the same name.  Maybe you're misremembering.

server/api/member.go
```diff
@@ -61,7 +61,8 @@ func (h *memberHandler) ListMembers(w http.ResponseWriter, r *http.Request) {
 
 func getMembers(svr *server.Server) (*pdpb.GetMembersResponse, error) {
 	req := &pdpb.GetMembersRequest{Header: &pdpb.RequestHeader{ClusterId: svr.ClusterID()}}
-	members, err := svr.GetMembers(context.Background(), req)
+	grpcServer := &server.GrpcServer{Server: svr}
+	members, err := grpcServer.GetMembers(context.Background(), req)
```
JmPotato, line 65: Can we provide a method for `*server.Server` like `s.GetRPCServer()` so that we can write these two lines into one.

```suggestion
	members, err := svr.GetRPCServer().GetMembers(context.Background(), req)
```

JmPotato (commented):
The rest LGTM.

server/grpc_service.go
```diff
@@ -52,11 +57,8 @@ var (
 )
 
 // GetMembers implements gRPC PDServer.
-func (s *Server) GetMembers(context.Context, *pdpb.GetMembersRequest) (*pdpb.GetMembersResponse, error) {
-	if s.IsClosed() {
-		return nil, status.Errorf(codes.Unknown, "server not started")
-	}
-	members, err := cluster.GetMembers(s.GetClient())
+func (s *GrpcServer) GetMembers(context.Context, *pdpb.GetMembersRequest) (*pdpb.GetMembersResponse, error) {
```
rleungx, line 60: BTW, do we need to check the cluster ID?

disksing: It is by design. The first request client sends to pd-server is `GetMembers` and the client does not know the cluster ID.

disksing: I think I can add some comments here.

server/api/member.go
disksing: em, I think one more line here is trivial, compared to introducing the gRPC concept to `server.Server`.

JmPotato: Ok. Fair enough.

JmPotato (approved):

rleungx (approved):

PR #4396, opened 2021-11-26
statistics: fix the problem that the hot cache cannot be emptied when the interval is less than 60

### LINKED ISSUE #4390: The hot cache cannot be cleared, when the interval is less than 60
### What did you do?

stop bench

hot cache is cleared

### What did you see instead?

there are still many hot peers

v5.0.4

lhy1024:

close #4390 

Our current heartbeat process looks like this.

If the interval is less than the default interval for heartbeats (60 seconds), then we will put it into the cache temporarily and wait until it has made up 60 seconds before considering whether it is hot enough.

For a peer that has just been reported, if the region is also in the hot cache, then there are three cases.
1. the store is the same, then the old item is used directly
2. the store is different, but there is a move peer or transfer leader in the previous round, then the inheritable item is directly used as the old item
3. store is different, and there is no inheritable peer, then choose any peer as old item

The problem occurs in the third branch, if the old peer is used directly without clone, then the old item and the new item will be written at the same time. And when the new peer interval is less than 60 seconds, it means that it will be temporarily put into the cache.

If the old peer will be cooled down at this time, it will keep the peer in the hot cache for a long time and cannot be exited.

This pr does a clone for the third branch to avoid this problem, and changes the store and region heartbeats to configurable in another pr.

### Check List

Tests

- Unit test

### Release note

```release-note
fix the problem that the hot cache cannot be emptied when the interval is less than 60 
```


nolouch (commented):
The issue and PR comments cannot let me understand the problem.

disksing (approved):

rleungx:
Could you provide more information on how it works?

lhy1024:
@rleungx @nolouch pr desc update

server/statistics/kind.go
```diff
@@ -97,6 +97,27 @@ func (k StoreStatKind) String() string {
...
+	Origin  SourceKind = iota // there is a corresponding peer in this store
+	Inherit                   // there is no a corresponding peer in this store and there is a peer just deleted
+	Adopt                     // there is no a corresponding peer in this store and there is no a peer just deleted, we need to copy from other store
+)
+
+func (k SourceKind) String() string {
+	switch k {
+	case Origin:
+		return "origin"
+	case Inherit:
```
rleungx, line 113: What is `Inherit` used for? It looks we only need `Adopt` right now.

server/statistics/kind.go
```diff
@@ -97,6 +97,27 @@ func (k StoreStatKind) String() string {
 	return "unknown StoreStatKind"
 }
 
+// SourceKind represents the statistics item source.
+type SourceKind int
+
+const (
+	Origin  SourceKind = iota // there is a corresponding peer in this store
+	Inherit                   // there is no a corresponding peer in this store and there is a peer just deleted
+	Adopt                     // there is no a corresponding peer in this store and there is no a peer just deleted, we need to copy from other store
```
rleungx, line 106:
```suggestion
	Adopt                     // there is no corresponding peer in this store and there is no peer just deleted, we need to copy from other stores
```

server/statistics/kind.go
```diff
@@ -97,6 +97,27 @@ func (k StoreStatKind) String() string {
 	return "unknown StoreStatKind"
 }
 
+// SourceKind represents the statistics item source.
+type SourceKind int
+
+const (
+	Origin  SourceKind = iota // there is a corresponding peer in this store
```
rleungx, line 104: seem we don't need to export it.

pkg/movingaverage/queue.go
```diff
@@ -53,3 +53,17 @@ func (sq *SafeQueue) PopFront() interface{} {
 	defer sq.mu.Unlock()
 	return sq.que.PopFront()
 }
+
+func (sq *SafeQueue) Clone() *SafeQueue {
+	sq.mu.Lock()
+	defer sq.mu.Unlock()
+	q := queue.New().Init()
+	for i := 0; i < sq.que.Len(); i++ {
+		node := sq.que.PopFront()
```
rleungx, line 62:
```suggestion
		v := sq.que.PopFront()
```

server/statistics/kind.go
lhy1024: yep, only distinct different source

rleungx (approved):

disksing (approved):

PR #4401, opened 2021-11-29
schedulers: evict leader support multiple targets

### LINKED ISSUE #4229: Evict leader supports multi targets
Subtask of [tikv#10602](https://github.com/tikv/tikv/issues/10602).

Briefly, we can make PD give TiKV all valid peers when evicting leader, and let TikV to decide which one to transfer leader to.  To compat with old TiKV version, we can simply introduce a new repeated field in the origin TransferLeader message, and fill in both fields.

MrCroxx:

Part of [tikv#10602](https://github.com/tikv/tikv/issues/10602).

Close #4229 .

Evict leader scheduler will pick all valid stores, and let TiKV decide the best store to transfer leader to.

This pr is refactored from #4214, see [more conversations](https://github.com/tikv/pd/pull/4214).

### Check List

Tests

- [x] Unit test
- [x] Integration test
- [x] Manual test (add detailed scripts or steps below)

Code changes

- Has configuration change
- Has persistent data change

Side effects

- Evict leader may lead to leader unbalanced when some store is slow (which is acceptable).

Related changes

- [kvproto#822](https://github.com/pingcap/kvproto/pull/822)
- [tikv#11063](https://github.com/tikv/tikv/pull/11063)

### Release note

```release-note
Evict leader automatically pick the fastest store to transfer leader to.
```


MrCroxx:
This pr is refactored from #4214 , which initially didn't take compatibility into consideration, and made the code verbose later. As @rleungx suggested, the new field (`peers` in `TransferLeader`) doesn't intersect in function with the older version, so we can reuse the origin struct.

server/schedule/operator/step.go
```diff
@@ -42,7 +42,10 @@ type OpStep interface {
 
 // TransferLeader is an OpStep that transfers a region's leader.
 type TransferLeader struct {
+	// Compatible with old TiKV's TransferLeader.
 	FromStore, ToStore uint64
+	// Multi-target transfer leader.
+	ToStores []uint64
```
rleungx, line 48: Perhaps, we can remove ToStore?

MrCroxx: I don't think that's a good idea. `ToStores` is only used during evicting leader, whose targets don't need to be filtered during `BuldJointXXX` or `BuildNonJointXXX`. It's better to separate `ToStore` and `ToStores` to avoid misleading.

MrCroxx: cc @HunDunDM

TennyZhuang (approved):

MrCroxx:
Any updates?

rleungx (commented):
Mostly LGTM, except for https://github.com/tikv/pd/pull/4401#discussion_r763567741

server/schedule/operator/builder.go
```diff
@@ -243,6 +244,28 @@ func (b *Builder) SetLeader(storeID uint64) *Builder {
...
 
+// SetLeaders records all valid target leaders in Builder.
+func (b *Builder) SetLeaders(storeIDs []uint64) *Builder {
+	if b.err != nil {
+		return b
+	}
+	for _, storeID := range storeIDs {
+		if peer, ok := b.targetPeers[storeID]; !ok {
+			continue
+		} else if core.IsLearner(peer) {
```
disksing, line 255:
```suggestion
		if core.IsLearner(peer) {
```
can avoid use `else`.

disksing:
The rest LGTM.

server/schedule/operator/step.go
disksing: separate 2 fields +1

server/schedule/operator/builder.go
MrCroxx: Refactored, but not completely in the suggested way.

disksing (approved):

MrCroxx:
Ready to merge? XD

rleungx (approved):

PR #4410, opened 2021-12-01
api: use heartbeat status to make store state more clear

### LINKED ISSUE #1566: Misleading store state
Currently, the status of down/disconnect is not reflected in the meta information of the store. As long as you do not manually go offline, the meta is always Up. This leads to the inability to determine the down state when using the grpc or HTTP interface to query the meta information of the store, and it is easy to mistake it as normal (but not).

Proposal: Define new state to use in the API, distinguish from that persistent to storage.

rleungx:

close #1566

This PR separates store state and heartbeat status and add two more heartbeat status.

### Check List

Tests

- Unit test

Code changes

Related changes

### Release note

```release-note
Use an individual heartbeat status for the store.
```

### DIFF
--- a/server/api/store.go
+++ b/server/api/store.go
@@ -50,6 +50,7 @@ type MetaStore struct {
 	LastHeartbeat       int64                `json:"last_heartbeat,omitempty"`
 	PhysicallyDestroyed bool                 `json:"physically_destroyed,omitempty"`
 	StateName           string               `json:"state_name"`
+	HeartbeatStatus     string               `json:"heartbeat_status"`
 }
 
 // NewMetaStore convert metapb.Store to MetaStore without State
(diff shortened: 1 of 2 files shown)


server/api/store.go
```diff
@@ -117,8 +118,10 @@ type StoreInfo struct {
 }
 
 const (
-	disconnectedName = "Disconnected"
-	downStateName    = "Down"
+	alive        = "Alive"
+	disconnected = "Disconnected"
+	down         = "Down"
+	dead         = "Dead"
```
disksing, line 124: I'm not sure if it is a good name. It may lead to some misunderstandings, as the tikv process may still be running.

rleungx: When it enters the dead status, the TiKV has already been Tombstone.

disksing (approved):

server/api/store.go
disksing: May need some opinions from other reviewers.

bufferflies: I think the down and disconnect is fuzzy for user if they do not see the code, and they may not known down will step to dead after the store has no regions.

JmPotato: `Dead` is kind of arbitrary. Maybe a word like `Removed`, `Outcasted`, or `Detached` is a better choice to indicate that the TiKV node does not belong to the cluster anymore.

PR #4412, opened 2021-12-01
server/schedule: prevent removing voters directly in 2 replica raft group

### LINKED ISSUE #4411: Removing voter from 2 voter replicas raft group would make region unavailable
### What did you do?

Remove a voter from 2 voter replicas.  Details in https://github.com/tikv/tikv/issues/10595.

Region shouldn't be unavailable.

### What did you see instead?

A region would be unavailable.

w41ter-l:

close #4411 

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
Prevent removing a voter from 2 replicas raft group directly, instead, demote the voter firstly.
```


PatrickNicholas:
@nolouch @rleungx PTAL, feel free to let me known if you have any question.
cc @BusyJay @NingLin-P

HunDunDM:
So currently,  we can’t change from 3 replicas to 1 replica?

PatrickNicholas:
> So currently, we can’t change from 3 replicas to 1 replica?

In https://github.com/tikv/tikv/issues/10595 has described a situation:
> Assume the leader wants to remove another voter. After it sends the message with the commitment of the confchange and before it persists the commit index (leader can send messages before persisting data) the leader node crashed, but the follower can still apply the confchange and destroy itself. And when the leader node recovers and the peer start as a follower, it will keep trying to request a vote from the destroyed follower since the commit index is not persisted, but it will never get it because the follower is destroyed.

Just change replicas from 3 to 1 is ok if cluster is normaly.

HunDunDM:
TiKV seems to be unable to execute the DemoteFollower command now(v5.3.0).
ref https://github.com/tikv/tikv/issues/11463

HunDunDM:
Can we add a test that requires the three steps of TransferLeader, DomoteFollower, and RemoveLearner?

PatrickNicholas:
> TiKV seems to be unable to execute the DemoteFollower command now(v5.3.0). ref [tikv/tikv#11463](https://github.com/tikv/tikv/issues/11463)

Ok, I will try to solve this problem first.

> Can we add a test that requires the three steps of TransferLeader, DomoteFollower, and RemoveLearner?

Ok

HunDunDM:
After merging #4534 , the issue will be easily implemented. Only need to change the judgment of `useJointConsensus`, use `buildStepsWithJointConsensus`, the step will be `ChangePeerV2(DemoteVoter) + RemovePeer`.

HunDunDM:
@w41ter-l will you continue this PR? Or do I submit a new PR?

w41ter-l:
@HunDunDM I haven't free time in recent days, please feel free to submit new PR.

HunDunDM:
Replaced by #4564 .

PR #4149, opened 2021-09-24
statistics: refacter store load in hot region scheduler

rleungx:
### What problem does this PR solve?

Move store-related statistics for scheduler package to statistics package.

### What is changed and how it works?

- move the store load statistics from `util.go` to `store_load.go`

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

PR #4424, opened 2021-12-05
statistics: fix the hot region API cannot work without hot scheduler

### LINKED ISSUE #3879: hot region should work if hot region schedule is disable
## Bug Report

I find hot region api will not work if hot region schedule is disable 
### What did you do?
schedule remove balance-hot-region-scheduler, See:
https://github.com/tikv/pd/blob/48963369ef04efaa9144eff27135c9ae14429a80/server/cluster/coordinator.go#L444-L456

hot region api should not be contorlled by  hot region schedule config

rleungx:

close #3879. But this PR should be reviewed after the following PRs are merged:

- [x] #4426 
- [x] #4149

This PR mainly makes the following changes:

- remove `hasHotStatus` interface
- no need to run a hot scheduler to get hot statistics
- move all calculation logic of hot region into `statistics` package

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


server/statistics/metrics.go
```diff
@@ -170,6 +170,14 @@ var (
 			Name:      "flow_queue_status",
 			Help:      "Status of the hotspot flow queue.",
 		}, []string{"type"})
+
+	hotPeerSummary = prometheus.NewGaugeVec(
```
bufferflies, line 174: Will it  move all hot perr summary from hot region scheduler to statustucs?

rleungx: > will the metrics works well if hot regions scheduler is disable.

The metrics don't have any relationship with the hot scheduler. 
> Will it move all hot perr summary from hot region scheduler to statistics?

I don't get it.

rleungx:
The lock problem needs to pay attention to.

server/cluster/coordinator.go
```diff
@@ -535,35 +517,33 @@ func (c *coordinator) resetSchedulerMetrics() {
 
 func (c *coordinator) collectHotSpotMetrics() {
```
bufferflies, line 518: There is some difference between collection and  GetHotXXXXRegions.

bufferflies (approved):

nolouch:
ptal @lhy1024

server/cluster/cluster.go
```diff
@@ -1590,23 +1581,17 @@ func (checker *prepareChecker) collect(region *core.RegionInfo) {
 
 // GetHotWriteRegions gets hot write regions' info.
 func (c *RaftCluster) GetHotWriteRegions(storeIDs ...uint64) *statistics.StoreHotPeersInfos {
-	c.RLock()
```
lhy1024, line 1593: Do we no longer need locks?

server/statistics/store_hot_peers_infos.go
```diff
@@ -14,6 +14,19 @@
...
 
+import (
+	"fmt"
+
+	"github.com/tikv/pd/server/core"
+)
+
+const (
+	// HotReadRegionType is hot read region scheduler type.
+	HotReadRegionType = "hot-read-region"
```
lhy1024, line 25: Where do we use it?

server/cluster/cluster.go
rleungx: I think so.

lhy1024 (approved):

Issue #4322, opened 2021-11-11
Enable more linters


rleungx:
We can utilize the linters integrated with golangci-lint to enhance code quality.

PR #4437, opened 2021-12-07
*: enable bodyclose linter

rleungx:

ref #4322

Enable bodyclose linter.

### Check List

Tests

- Unit test
### Release note

```release-note
None
```

### DIFF
--- a/server/api/util.go
+++ b/server/api/util.go
@@ -64,7 +64,7 @@ func readJSON(client *http.Client, url string, data interface{}) error {
 }
 
 func readJSONWithBody(client *http.Client, url string, body []byte, data interface{}) error {
-	req, err := http.NewRequest("GET", url, bytes.NewBuffer(body))
+	req, err := http.NewRequest(http.MethodGet, url, bytes.NewBuffer(body))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


lhy1024 (approved):

server/api/util.go
```diff
@@ -127,17 +127,18 @@ func patchJSON(client *http.Client, url string, body []byte) error {
...
+	req, err := http.NewRequest(http.MethodDelete, url, nil)
 	if err != nil {
-		return nil, err
+		return http.StatusBadRequest, err
 	}
+	// TODO: find a better way to return status code.
 	res, err := client.Do(req)
 	if err != nil {
-		return nil, err
+		return http.StatusInternalServerError, err
```
disksing, line 138: can we `return 0, err` here?

rleungx: The caller will get 0 as the status code which is not good. I think an alternative is to return the status code according to the detailed error. But I'm going to leave it for now.

tests/server/api/api_test.go
```diff
@@ -70,6 +70,7 @@ func (s *serverTestSuite) TestReconnect(c *C) {
 		if name != leader {
 			res, e := http.Get(s.GetConfig().AdvertiseClientUrls + "/pd/api/v1/version")
 			c.Assert(e, IsNil)
+			res.Body.Close()
```
JmPotato, line 73: Why not `defer`?

JmPotato (commented):
The rest LGTM.

tests/server/api/api_test.go
rleungx: It's a for loop, closing it directly is better.

server/api/util.go
disksing: 0 is ok to use as an _invalid status code_ when err is not nil. If you construct status code in client side, we cannot distiguish the cases that we receive the status code from server or construct it from client.

rleungx: Make sense, change to 0.

rleungx:
@JmPotato @disksing PTAL

JmPotato (approved):

Issue #4444, opened 2021-12-09
Demoting single voter directly isn't working


w41ter-l:
## Bug Report

Details see issue https://github.com/tikv/tikv/issues/11463.

This is introduced in https://github.com/tikv/pd/pull/4166, to fix https://github.com/tikv/pd/issues/4140. I decide to change the field `ChangePeer` to `ChangePeerV2` if demoting a voter directlry. This can avoid breaking compatibility.

PR #4445, opened 2021-12-09
server/schedule: use ChangePeerV2 to demote single voter directly

w41ter-l:

ref #4444.
ref https://github.com/tikv/tikv/issues/11463.

 Change the field `ChangePeer` to `ChangePeerV2` if demoting a voter directlry. This can avoid breaking compatibility.

### Check List

Tests

- No code

Related changes

### Release note

```release-note
Fix the problem that demoting a follower to learner isn't working.
```

### DIFF
--- a/server/schedule/operator/step.go
+++ b/server/schedule/operator/step.go
@@ -409,6 +409,20 @@ func (df DemoteFollower) CheckInProgress(cluster opt.Cluster, region *core.Regio
 // Influence calculates the store difference that current step makes.
 func (df DemoteFollower) Influence(opInfluence OpInfluence, region *core.RegionInfo) {}
 
+// GetRequest get the ChangePeerV2 request
+func (df DemoteFollower) GetRequest() *pdpb.ChangePeerV2 {
+	return &pdpb.ChangePeerV2{
+		Changes: []*pdpb.ChangePeer{{
+			ChangeType: eraftpb.ConfChangeType_AddLearnerNode,
+			Peer: &metapb.Peer{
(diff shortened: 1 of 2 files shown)


PatrickNicholas:
In addition to this PR, please also help me to review : https://github.com/pingcap/kvproto/pull/839.

HunDunDM:
Since `ChangePeerV2` is used, `useJointConsensus == true` is required. This shows that the judgment condition of `useJointConsensus` and the logic of `buildStepsWithJointConsensus` need to be changed. At the same time, `DemoteFollower` will not be used in `buildStepsWithoutJointConsensus`, it must be split into `RemovePeer` and `AddLearner`.

HunDunDM:
We need a new way of implementation.

nolouch:
ptal @HunDunDM

server/schedule/operator/step.go
```diff
@@ -409,6 +409,20 @@ func (df DemoteFollower) CheckInProgress(cluster opt.Cluster, region *core.Regio
 // Influence calculates the store difference that current step makes.
 func (df DemoteFollower) Influence(opInfluence OpInfluence, region *core.RegionInfo) {}
 
+// GetRequest get the ChangePeerV2 request
+func (df DemoteFollower) GetRequest() *pdpb.ChangePeerV2 {
```
bufferflies, line 413: how about add unit test?

HunDunDM:
#4444 is implemented in #4534, maybe we can close this PR? If you have any questions about #4534, please ask.

PR #4446, opened 2021-12-09
statistics: fix hot peer cache

### LINKED ISSUE #4390: The hot cache cannot be cleared, when the interval is less than 60

lhy1024:

close #4390 

> Our current heartbeat process looks like this.
> 
> If the interval is less than the default interval for heartbeats (60 seconds), then we will put it into the cache temporarily and wait until it collects 60 seconds before considering whether it is hot enough.
> 
> For a peer that has just been reported, if the region is also in the hot cache, then there are three cases.
> 
> 1. the store is the same, then the old item is used directly
> 2. the store is different, but there is a move peer or transfer leader in the previous round, then the inheritable item is directly used as the old item
> 3. store is different, and there is no inheritable peer, then choose any peer as old item
> 
> The problem occurs in the third branch, if the old peer is used directly without clone, then the old item and the new item will be written at the same time. And when the new peer interval is less than 60 seconds, it means that it will be temporarily put into the cache.
> 
> If the old peer will be cooled down at this time, it will keep the peer in the hot cache for a long time and cannot be exited.

For `store is different, and there is no inheritable peer, then choose any peer as old item`, we call it as `adopt`, taking the child of a relative as own son or daughter.

Suppose there is a cluster with region heartbeat 5s.

At t1, when a peer in store1 is changed to cold, and other peers still are hot, it will adopt from store2 as own, it is expected. 

|        | source                  | hot/cold | interval of sum |
| ------ | ----------------------- | -------- | --------------- |
| store1 | adopt from store2 at t1 | cold     | 60s             |
| store2 | inherit                 | hot      | 55s             |
| store3 | inherit                 | hot      | 55s             |

But other peers will be changed to cold, the problem occurs. A peer in store2 or store3 adopts from store1 at t2, but this child is from store2 at t1.

|        | source                                             | hot/cold | interval of sum |
| ------ | -------------------------------------------------- | -------- | --------------- |
| store1 | adopt from store2 at t1                            | cold     | 5s              |
| store2 | adopt from store1 at t2 (it is from store2 at t1!) | cold     | 60s             |
| store3 | adopt from store1 at t2 (it is from store2 at t1!) | cold     | 60s             |

So we should avoid to adopt from adopted item before it is confirmed whether it is hot.

add `justFromAdopt` to avoid to adopt from adopted item before it is confirmed whether is hot.

Before this pr, there are still some hot peers in cluster after workload stopping.

After this pr, it is expected

image

### Check List

Tests

- Unit test
- Integration test

Related changes

### Release note

```release-note
 None.
```


rleungx:
It's hard to understand what is the figure above going to talk about.

lhy1024:
> It's hard to understand what is the figure above going to talk about.

I am very sorry that it was too late last night to add just a brief description; I will add the complete one later.

server/statistics/hot_peer.go
```diff
@@ -109,6 +109,10 @@ type HotPeerStat struct {
 	// If the peer didn't been send by store heartbeat when it is already stored as hot peer stat,
 	// we will handle it as cold peer and mark the inCold flag
 	inCold bool
+	source sourceKind
+	// it is uncertain whether it is hot and it is adopt from other store
+	// it will be changed to false until it is confirmed to be hot
+	justFromAdopt bool
```
disksing, line 115: How about `allowAdopt`? (and swap true/false)

lhy1024:
@disksing @rleungx PTAL

nolouch (commented):
Could you show the origin data that you meet in the test cluster?  I cannot understand the example that you shown.

lhy1024:
> Could you show the origin data that you meet in the test cluster? I cannot understand the example that you shown. and also the test cannot make me understand, I think it does not match you issue mentioned.

log:
```
tick1
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=10] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=2] [hot-anti-count=1] [sum-interval=55s] [need-delete=false] [source=direct]
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=1] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
tick2
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=2] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=1] [hot-anti-count=1] [sum-interval=5s] [need-delete=false] [source=adopt]
[2021/12/17 00:12:40.940 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=10] [hot-anti-count=1] [sum-interval=5s] [need-delete=false] [source=adopt]
tick*n

tick3
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=10] [hot-anti-count=1] [sum-interval=55s] [need-delete=false] [source=direct]
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=2] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=1] [hot-anti-count=1] [sum-interval=55s] [need-delete=false] [source=direct]
tick4
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=10] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat delete from cache"] [store=1] [hot-anti-count=0] [sum-interval=0s] [need-delete=true] [source=direct]
[2021/12/17 00:12:40.942 +08:00] [INFO] [hot_peer.go:132] ["region heartbeat update"] [store=2] [hot-anti-count=1] [sum-interval=5s] [need-delete=false] [source=adopt]

```
Firstly store 1 or store 10 adopt from store2 at tick2
Secondly store 2 adopt item from store10 or store1 at tick4

server/statistics/hot_peer_cache.go
```diff
@@ -101,7 +101,9 @@ func (f *hotPeerCache) RegionStats(minHotDegree int) map[uint64][]*HotPeerStat {
 // Update updates the items in statistics.
 func (f *hotPeerCache) Update(item *HotPeerStat) {
 	if item.IsNeedDelete() {
-		f.putInheritItem(item)
+		if item.AntiCount > 0 {
```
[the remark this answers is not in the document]
lhy1024: ok

disksing (approved):

server/statistics/hot_peer_cache.go
```diff
@@ -525,6 +529,7 @@ func coldItem(newItem, oldItem *HotPeerStat) {
 func hotItem(newItem, oldItem *HotPeerStat) {
 	newItem.HotDegree = oldItem.HotDegree + 1
 	newItem.AntiCount = hotRegionAntiCount
+	newItem.allowAdopt = true
```
rleungx, line 532: What about `coldItem`?

server/statistics/hot_peer_cache.go
```diff
@@ -517,6 +521,7 @@ func (f *hotPeerCache) removeItem(item *HotPeerStat) {
 func coldItem(newItem, oldItem *HotPeerStat) {
 	newItem.HotDegree = oldItem.HotDegree - 1
 	newItem.AntiCount = oldItem.AntiCount - 1
+	newItem.allowAdopt = true
```
rleungx, line 524: If the item needs to be deleted, shall we allow `adopt`?

lhy1024: update

nolouch (commented):
It is best to add a better test case that matches the description of the issue

server/statistics/hot_peer.go
```diff
@@ -127,15 +132,18 @@ func (stat *HotPeerStat) Log(str string, level func(msg string, fields ...zap.Fi
 		zap.Uint64("interval", stat.interval),
 		zap.Uint64("region-id", stat.RegionID),
 		zap.Uint64("store", stat.StoreID),
+		zap.Bool("is-leader", stat.isLeader),
```
rleungx, line 135: Do we really need this log with so many fields?

rleungx:
BTW, what about `inheritItemDegree`?

lhy1024:
> BTW, what about `inheritItemDegree`?

It will not change whether allow adopt, because we are not sure whether is hot.
And new item can get value from old item in https://github.com/tikv/pd/blob/39bcb80b7ce387f2df574f068f1fb31336383905/server/statistics/hot_peer_cache.go#L395

server/statistics/hot_peer.go
lhy1024: I think we can remove some after #4461

server/statistics/hot_peer_cache.go
```diff
@@ -519,12 +523,15 @@ func coldItem(newItem, oldItem *HotPeerStat) {
 	newItem.AntiCount = oldItem.AntiCount - 1
 	if newItem.AntiCount <= 0 {
 		newItem.needDelete = true
+	} else {
+		newItem.allowAdopt = true
```
nolouch, line 527: Is there a problem? such as `allowAdopt = false` for a new item, and then a heartbeat turns to `allowAdopt = true`, and then other stores can adopt from it, and the same problem occurs repeatedly.

lhy1024: t1: peer in store1 adopt from store2, with anticount is 2 and is not allowed to adopt
t2. store1 heartbeat and cold, peer in store1 is with anticount is 1 and allowed to adopt 

So I think it doesn't occur repeatedly, because anticount decrease

nolouch (commented):
lgtm

nolouch:
ptal @rleungx

nolouch (approved):

Issue #4449, opened 2021-12-10
core: hot history region storage cant get is_leader info and format error


qidi1:
### What did you do?

when we use  `TIDB_HOT_REGIONS_HISTORY` table to get hot region history info,The `isLeader` field is always false,and The `startKey` and `endKey` fields formatted like `dIAAAAAAAAD/aF9ygAAAAAD/CrGkAAAAAAD6` is also not what we want to see

### What did you expect to see?

we expect to get  `isLeader`  info and  The `startKey` and `endKey` fields formatted like `dIAAAAAAAAD/aF9ygAAAAAD/CrGkAAAAAAD6` can be see like this `7480000000000001FF8800000000000000F8`.

rleungx:
Closed via #4450.

PR #4450, opened 2021-12-10
core:fix style of hot history region info of startkey and endkey,fix bug isleader always false

qidi1:
related issues:
* https://github.com/tikv/pd/issues/4449

when we use TIDB_HOT_REGIONS_HISTORY table to get hot region history info,The isLeader field is always false,and The startKey and endKey fields formatted like dIAAAAAAAAD/aF9ygAAAAAD/CrGkAAAAAAD6 is also not what we want to see

we expect to get isLeader info and The startKey and endKey fields formatted like dIAAAAAAAAD/aF9ygAAAAAD/CrGkAAAAAAD6 can be see like this 7480000000000001FF8800000000000000F8.

remove `omit empty` behind struct json format,change is_leader judge way and change the style of start_key and end_key store in storage

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

previous commit:
* https://github.com/tikv/pd/pull/4020
### Release note

```release-note
None
```


server/core/hot_region_storage_test.go
```diff
@@ -63,8 +63,8 @@ func (m *MockPackHotRegionInfo) GenHistoryHotRegions(num int, updateTime time.Ti
 			FlowBytes:     rand.Float64() * 100,
 			KeyRate:       rand.Float64() * 100,
 			QueryRate:     rand.Float64() * 100,
-			StartKey:      []byte(fmt.Sprintf("%20d", i)),
```
rleungx, line 66: Do we need more tests to test key format?

qidi1: I add some test about key format

qidi1: this function is used by benchmark,not test

server/core/hot_region_storage.go
```diff
@@ -254,16 +254,16 @@ func (h *HotRegionStorage) packHistoryHotRegions(historyHotRegions []HistoryHotR
 	for i := range historyHotRegions {
 		region := &metapb.Region{
 			Id:             historyHotRegions[i].RegionID,
-			StartKey:       historyHotRegions[i].StartKey,
-			EndKey:         historyHotRegions[i].EndKey,
+			StartKey:       HexRegionKey([]byte(historyHotRegions[i].StartKey)),
```
nolouch, line 257: I think we can store origin format, just covert it in API.

nolouch (commented):
rest LGTM

server/core/hot_region_storage.go
qidi1: If we use Json.Marshal to Encode the StartKey and EndKey and then use Json.UnMarshal to decode them, the values we get will change. The []byte will be escaped. For example, the original key is "t\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\x15_i\x80\x00\x00\x00\x00\x00\x00\x00\xff\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xfa", and after json.Marshal and then UnMarshal becomes "t�\x00\x00\x00\x00\x00\x00\x00\x00\x00��\x15_i�\x00\x00\x00\x00\x00\x00\x00\x00��\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"

nolouch (approved):
lgtm, ptal @rleungx

rleungx:
https://github.com/tikv/pd/pull/4450#discussion_r767485149 is still not solved.

server/core/hot_region_storage.go
```diff
@@ -338,21 +338,20 @@ func (it *HotRegionStorageIterator) Next() (*HistoryHotRegion, error) {
...
+		StartKey:       []byte(message.StartKey),
+		EndKey:         []byte(message.EndKey),
 		EncryptionMeta: message.EncryptionMeta,
 	}
 	if err := encryption.DecryptRegion(region, it.encryptionKeyManager); err != nil {
 		return nil, err
 	}
-	message.StartKey = region.StartKey
-	message.EndKey = region.EndKey
+	message.StartKey = String(region.StartKey)
```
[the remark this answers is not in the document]
nolouch: ptal @qidi1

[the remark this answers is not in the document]
nolouch: Here is a `DecryptRegion` operator, I think here want to use decrypt key.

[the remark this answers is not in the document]
rleungx: got

rleungx (approved):

Issue #4422, opened 2021-12-03
Fail to execute remove tombstone 


rleungx:
### What did you do?

Try to remove tombstone record from Grafana.

### What did you expect to see?

The store can be removed from PD successfully.

### What did you see instead?

```
[2021/12/03 07:46:41.333 +00:00] [WARN] [cluster.go:1190] ["skip removing tombstone"] [store="id:5 address:\"172.17.31.167:20160\" state:Tombstone version:\"4.0.13\" status_address:\"172.17.31.167:20180\" git_hash:\"a448d617f79ddf545be73931525bb41af0f790f3\" start_timestamp:1624611301 deploy_path:\"/data1/deploy/install/deploy/tikv-20160/bin\" last_heartbeat:1634311455424568570 "]
[2021/12/03 07:46:41.333 +00:00] [WARN] [cluster.go:1190] ["skip removing tombstone"] [store="id:7 address:\"172.17.31.168:20160\" state:Tombstone version:\"4.0.13\" status_address:\"172.17.31.168:20180\" git_hash:\"a448d617f79ddf545be73931525bb41af0f790f3\" start_timestamp:1624611301 deploy_path:\"/data1/deploy/install/deploy/tikv-20160/bin\" last_heartbeat:1634311454289591213 "]
```

https://github.com/tikv/pd/blob/cefca8069820fcffebcaeec4e3951e5045147fd4/server/cluster/cluster.go#L1251

We should use `GetStoreRegionCount` here.

### What version of PD are you using (`pd-server -V`)?
v4.0.13

rleungx:
Closed via #4451.

PR #4451, opened 2021-12-10
fixbug-#4422

Dagon0577:

This is to solve the error skip code found in https://github.com/tikv/pd/issues/4422
ref #4422
we should gets the total count of a store's leader and follower RegionInfo by storeID rather than get the Region count of the store.
### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


rleungx (approved):
Thanks. LGTM.

nolouch (approved):

PR #4453, opened 2021-12-10
Add parameter prompt for scheduler show --status ref #4405

### LINKED ISSUE #4405: Add parameter prompt for scheduler show --status
There is no prompt for `scheduler show --status` command to indicate the value of status. See:

```
» scheduler show --status
Usage:
  pd-ctl scheduler show [flags]

Flags:
  -h, --help            help for show
      --status string   the scheduler status

flag needs an argument: --status
```

Dagon0577:
close #4405

Add parameter prompt for scheduler show --status ref #4405

Just add parameter prompt for scheduler show --status
### Check List

Tests

- Unit test
- Integration test
### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/scheduler.go
+++ b/tools/pd-ctl/pdctl/command/scheduler.go
@@ -87,7 +87,7 @@ func NewShowSchedulerCommand() *cobra.Command {
 		Short: "show schedulers",
 		Run:   showSchedulerCommandFunc,
 	}
-	c.Flags().String("status", "", "the scheduler status")
+	c.Flags().String("status", "", "the scheduler status value can be paused | disabled")
 	return c
 }
 
--- a/tools/pd-ctl/pdctl/command/scheduler.go
+++ b/tools/pd-ctl/pdctl/command/scheduler.go
@@ -87,7 +87,7 @@ func NewShowSchedulerCommand() *cobra.Command {
 		Short: "show schedulers",
 		Run:   showSchedulerCommandFunc,
 	}
-	c.Flags().String("status", "", "the scheduler status value can be paused | disabled")
+	c.Flags().String("status", "", "the scheduler status value can be [paused | disabled]")
 	return c
 }
 


rleungx:
Perhaps we need another status which could be named "running" or "enabled" and make the default output of `scheduler show` command to list all running schedulers. /cc @nolouch

JayiceZ:
> Perhaps we need another status which could be named "running" or "enabled" and make the default output of `scheduler show` command to list all running schedulers. /cc @nolouch

But it seems like if the status is meaningless, then aall running schedulers will be output by default

see https://github.com/tikv/pd/blob/eba75c3080a0db2d5edda7a2a8628e3a08c1af87/server/api/scheduler.go#L94

rleungx:
> > Perhaps we need another status which could be named "running" or "enabled" and make the default output of `scheduler show` command to list all running schedulers. /cc @nolouch
> 
> But it seems like if the status is meaningless, then all running schedulers will be output by default
> 
> see
> 
> https://github.com/tikv/pd/blob/eba75c3080a0db2d5edda7a2a8628e3a08c1af87/server/api/scheduler.go#L94

It will actually print all existing schedulers, but we cannot tell if it is running.

disksing (approved):

Dagon0577:
> 

maybe you can do it, i will close this

tools/pd-ctl/pdctl/command/scheduler.go
```diff
@@ -87,7 +87,7 @@ func NewShowSchedulerCommand() *cobra.Command {
 		Short: "show schedulers",
 		Run:   showSchedulerCommandFunc,
 	}
-	c.Flags().String("status", "", "the scheduler status")
+	c.Flags().String("status", "", "the scheduler status value can be [paused | disabled]")
```
bufferflies, line 90: the paused and disbaled will support all or one should be deprecated.

nolouch:
> Perhaps we need another status which could be named "running" or "enabled" and make the default output of `scheduler show` command to list all running schedulers. /cc @nolouch

It's good point, we can do it in another PR or issues.

nolouch (approved):
lgtm

PR #4458, opened 2021-12-14
scheduler: add approximateSize property to operator struct

### LINKED ISSUE #4457: Operator add region size properties
## Enhancement Task
In past, it will bury operator if it's running duration  exceeds  than 10 minutes, it can not work well if region size is big like 10GB. The timeout should depend on operator step and region's size. It is the first step to solve operator life cycle.
More detail is described at #4456

bufferflies:
Operator timeout will depend on region size not fixed value and close #4457.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


nolouch:
plz describe you issue and PR clearly.

disksing:
better to change title to `scheduler: add approximateSize property to operator struct`

disksing (approved):

lhy1024 (commented):
Would you like to add more descriptions?

lhy1024 (commented):
LGTM

lhy1024 (approved):

rleungx:
We need more information before this PR is merged.

server/schedule/operator/operator.go
```diff
@@ -363,3 +367,11 @@ func (o *Operator) GetAdditionalInfo() string {
 	}
 	return ""
 }
+
+// mock region default region size is 96Mb.
+const mockRegionSize = 96 * (1 << 20)
+
+// NewTestOperator creates a test operator.
+func NewTestOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, steps ...OpStep) *Operator {
```
rleungx, line 375: Why do we need it?

server/schedule/operator/operator.go
```diff
@@ -81,7 +83,9 @@ func (o *Operator) String() string {
 	for i := range o.steps {
 		stepStrs[i] = o.steps[i].String()
 	}
-	s := fmt.Sprintf("%s {%s} (kind:%s, region:%v(%v,%v), createAt:%s, startAt:%s, currentStep:%v, steps:[%s])", o.desc, o.brief, o.kind, o.regionID, o.regionEpoch.GetVersion(), o.regionEpoch.GetConfVer(), o.GetCreateTime(), o.GetStartTime(), atomic.LoadInt32(&o.currentStep), strings.Join(stepStrs, ", "))
+	s := fmt.Sprintf("%s {%s} (kind:%s, region:%v(%v,%v), createAt:%s, startAt:%s, currentStep:%v, size:%v，steps:[%s])",
```
rleungx, line 86:
```suggestion
	s := fmt.Sprintf("%s {%s} (kind:%s, region:%v(%v,%v), createAt:%s, startAt:%s, currentStep:%v, size:%v, steps:[%s])",
```

server/schedule/operator/operator.go
bufferflies: This constructor of the operator is used for unit test, it will decrease works for unit test.

server/schedule/operator/operator.go
```diff
@@ -363,3 +367,15 @@ func (o *Operator) GetAdditionalInfo() string {
 	}
 	return ""
 }
+
+// mock region default region size is 96Mb.
```
rleungx, line 371: MB?

server/schedule/operator/operator.go
```diff
@@ -363,3 +367,15 @@ func (o *Operator) GetAdditionalInfo() string {
 	}
 	return ""
 }
+
+const (
```
rleungx, line 371: Please add a comment to emphasize these constants are only used for unit test.

rleungx (approved):
The rest LGTM.

Issue #4109, opened 2021-09-13
dcLocation may have the same name as an existing module and cause an accident


HunDunDM:
### What did you do?

* Create a cluster with local tso. Their `dc-location` is named `leader`, `member`, `alloc_id`, `raft` etc.

### What did you expect to see?

* everything is normal

### What did you see instead?

* [allocator_manager.go#L330](https://github.com/tikv/pd/blob/v5.2.1/server/tso/allocator_manager.go#L330)
* `dcLocation` may have the same name as an existing module and cause an accident

### What version of PD are you using (`pd-server -V`)?

v5.2.1

PR #4464, opened 2021-12-15
tso: add a local TSO prefix to prevent being conflicted with other system key paths

JmPotato:

Close https://github.com/tikv/pd/issues/4109.

Add a prefix to the root path to prevent being conflicted with other system key paths such as leader, member, alloc_id, raft, etc.

### Check List

Tests

- Unit test
- Integration test

Side effects

**Since the Local TSO has not been released yet, I think it's ok to break the backward compatibility.**

### Release note

```release-note
None.
```

### DIFF
--- a/server/tso/allocator_manager.go
+++ b/server/tso/allocator_manager.go
@@ -47,8 +47,8 @@ const (
 	patrolStep                  = 1 * time.Second
 	defaultAllocatorLeaderLease = 3
 	leaderTickInterval          = 50 * time.Millisecond
-	localTSOAllocatorEtcdPrefix = "local-tso-allocator"
-	localTSOSuffixEtcdPrefix    = "local-tso-suffix"
+	localTSOAllocatorEtcdPrefix = "lta"
+	localTSOSuffixEtcdPrefix    = "lts"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


server/tso/allocator_manager.go
```diff
@@ -47,6 +47,7 @@ const (
 	patrolStep                  = 1 * time.Second
 	defaultAllocatorLeaderLease = 3
 	leaderTickInterval          = 50 * time.Millisecond
+	localTSOAllocatorEtcdPrefix = "local-tso-allocator"
```
rleungx, line 50: Do we have a compatibility issue?

JmPotato: Since the Local TSO has not been released yet, I think it's ok to break the backward compatibility.

nolouch: It's better to be shorter, like "lta"?

nolouch (commented):
BTW, is there exist tests?

server/tso/allocator_manager.go
rleungx: BTW, should we use `-` or `_`?

JmPotato: I think `_` is better since we already have keys like `alloc_id`.

nolouch (approved):
lgtm

rleungx (approved):

Issue #4242, opened 2021-10-27
Support online recovery for Raft majority failure


v01dstar:
This is mirror of https://github.com/tikv/tikv/issues/10483

PR #4475, opened 2021-12-17
unsafe recovery: Pause all schedulers and checkers when doing unsafe recovery 

v01dstar:

Issue Number: ref #4242 which is part of tikv/tikv#10483

Split/merge and conf change can cause PD make wrong decisions during unsafe recovery. Previously, we expect the user to stop all the schedulers themselves, but that is not reliable, Unsafe Recovery should be responsible for pausing all the schedulers.

Pause all the schedulers before collecting store reports. Note that, to avoid reporting stale info, tikv also need to do its own parts, such as making sure apply index is equal to commit index.

### Check List

Tests

- Unit test

Code changes

- N/A

Side effects

- All schedulers are paused until the recovery is finished or timed out.

Related changes

- N/A

### Release note

```release-note
None
```

### DIFF
--- a/server/api/unsafe_operation.go
+++ b/server/api/unsafe_operation.go
@@ -59,13 +59,7 @@ func (h *unsafeOperationHandler) RemoveFailedStores(w http.ResponseWriter, r *ht
 	for _, store := range storeSlice {
 		stores[store] = ""
 	}
-	timeout := int64(600)
-	rawTimeout, exists := input["timeout"]
-	if exists {
-		timeout = int64(rawTimeout.(float64))
-	}
-
-	if err := rc.GetUnsafeRecoveryController().RemoveFailedStores(stores, timeout); err != nil {
+	if err := rc.GetUnsafeRecoveryController().RemoveFailedStores(stores); err != nil {
 		h.rd.JSON(w, http.StatusInternalServerError, err.Error())
 		return
 	}
--- a/server/api/unsafe_operation.go
+++ b/server/api/unsafe_operation.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 
 	"github.com/tikv/pd/pkg/apiutil"
+	"github.com/tikv/pd/pkg/typeutil"
 	"github.com/tikv/pd/server"
 	"github.com/unrolled/render"
 )
@@ -36,6 +37,8 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
 
 // @Tags unsafe
 // @Summary Remove failed stores unsafely.
+// @Accept json
+// @Param body body object true "json params"
 // @Produce json
 // Success 200 {string} string "Request has been accepted."
 // Failure 400 {string} string "The input is invalid."
@@ -43,15 +46,26 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
 // @Router /admin/unsafe/remove-failed-stores [POST]
 func (h *unsafeOperationHandler) RemoveFailedStores(w http.ResponseWriter, r *http.Request) {
 	rc := getCluster(r)
-	var stores map[uint64]string
-	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &stores); err != nil {
+	var input map[string]interface{}
+	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &input); err != nil {
 		return
 	}
-	if len(stores) == 0 {
-		h.rd.JSON(w, http.StatusBadRequest, "No store specified")
+	storeSlice, ok := typeutil.JSONToUint64Slice(input["stores"])
+	if !ok {
+		h.rd.JSON(w, http.StatusBadRequest, "Store ids are invalid")
 		return
 	}
-	if err := rc.GetUnsafeRecoveryController().RemoveFailedStores(stores); err != nil {
+	stores := make(map[uint64]string)
+	for _, store := range storeSlice {
+		stores[store] = ""
+	}
+	timeout := int64(600)
+	rawTimeout, exists := input["timeout"]
+	if exists {
+		timeout = int64(rawTimeout.(float64))
+	}
+
+	if err := rc.GetUnsafeRecoveryController().RemoveFailedStores(stores, timeout); err != nil {
 		h.rd.JSON(w, http.StatusInternalServerError, err.Error())
 		return
 	}
--- a/server/api/unsafe_operation_test.go
+++ b/server/api/unsafe_operation_test.go
@@ -45,9 +45,16 @@ func (s *testUnsafeAPISuite) TearDownSuite(c *C) {
 }
 
 func (s *testUnsafeAPISuite) TestRemoveFailedStores(c *C) {
-	input := map[uint64]string{1: ""}
-	data, err := json.Marshal(input)
-	c.Assert(err, IsNil)
+	input := map[string]interface{}{"stores": []uint64{}}
+	data, _ := json.Marshal(input)
+	err := postJSON(testDialClient, s.urlPrefix+"/remove-failed-stores", data)
+	c.Assert(err.Error(), Equals, "\"[PD:unsaferecovery:ErrUnsafeRecoveryInvalidInput]invalid input no store specified\"\n")
+	input = map[string]interface{}{"stores": []string{"abc", "def"}}
+	data, _ = json.Marshal(input)
+	err = postJSON(testDialClient, s.urlPrefix+"/remove-failed-stores", data)
+	c.Assert(err.Error(), Equals, "\"Store ids are invalid\"\n")
+	input = map[string]interface{}{"stores": []uint64{1, 2}, "timeout": 3600}
+	data, _ = json.Marshal(input)
 	err = postJSON(testDialClient, s.urlPrefix+"/remove-failed-stores", data)
 	c.Assert(err, IsNil)
 	// Test show
--- a/server/api/unsafe_operation_test.go
+++ b/server/api/unsafe_operation_test.go
@@ -53,7 +53,7 @@ func (s *testUnsafeAPISuite) TestRemoveFailedStores(c *C) {
 	data, _ = json.Marshal(input)
 	err = postJSON(testDialClient, s.urlPrefix+"/remove-failed-stores", data)
 	c.Assert(err.Error(), Equals, "\"Store ids are invalid\"\n")
-	input = map[string]interface{}{"stores": []uint64{1, 2}, "timeout": 3600}
+	input = map[string]interface{}{"stores": []uint64{1, 2}}
 	data, _ = json.Marshal(input)
 	err = postJSON(testDialClient, s.urlPrefix+"/remove-failed-stores", data)
 	c.Assert(err, IsNil)
--- a/server/cluster/cluster_worker.go
+++ b/server/cluster/cluster_worker.go
@@ -42,6 +42,9 @@ func (c *RaftCluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 
 // HandleAskSplit handles the split request.
 func (c *RaftCluster) HandleAskSplit(request *pdpb.AskSplitRequest) (*pdpb.AskSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errs.ErrUnsafeRecoveryIsRunning.FastGenByArgs()
+	}
 	reqRegion := request.GetRegion()
 	err := c.ValidRequestRegion(reqRegion)
 	if err != nil {
@@ -94,6 +97,9 @@ func (c *RaftCluster) ValidRequestRegion(reqRegion *metapb.Region) error {
 
 // HandleAskBatchSplit handles the batch split request.
 func (c *RaftCluster) HandleAskBatchSplit(request *pdpb.AskBatchSplitRequest) (*pdpb.AskBatchSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errs.ErrUnsafeRecoveryIsRunning.FastGenByArgs()
+	}
 	reqRegion := request.GetRegion()
 	splitCount := request.GetSplitCount()
 	err := c.ValidRequestRegion(reqRegion)
--- a/server/cluster/coordinator.go
+++ b/server/cluster/coordinator.go
@@ -116,6 +116,10 @@ func (c *coordinator) patrolRegions() {
 			log.Info("patrol regions has been stopped")
 			return
 		}
+		if c.cluster.GetUnsafeRecoveryController() != nil && c.cluster.GetUnsafeRecoveryController().IsRunning() {
+			// Skip patrolling regions during unsafe recovery.
+			continue
+		}
 
 		// Check priority regions first.
 		c.checkPriorityRegions()
@@ -722,6 +726,20 @@ func (c *coordinator) pauseOrResumeScheduler(name string, t int64) error {
 	return err
 }
 
+// isSchedulerAllowed returns whether a scheduler is allowed to schedule, a scheduler is not allowed to schedule if it is paused or blocked by unsafe recovery.
+func (c *coordinator) isSchedulerAllowed(name string) (bool, error) {
+	c.RLock()
+	defer c.RUnlock()
+	if c.cluster == nil {
+		return false, errs.ErrNotBootstrapped.FastGenByArgs()
+	}
+	s, ok := c.schedulers[name]
+	if !ok {
+		return false, errs.ErrSchedulerNotFound.FastGenByArgs()
+	}
+	return s.AllowSchedule(), nil
+}
+
 func (c *coordinator) isSchedulerPaused(name string) (bool, error) {
 	c.RLock()
 	defer c.RUnlock()
@@ -882,7 +900,7 @@ func (s *scheduleController) GetInterval() time.Duration {
 
 // AllowSchedule returns if a scheduler is allowed to schedule.
 func (s *scheduleController) AllowSchedule() bool {
-	return s.Scheduler.IsScheduleAllowed(s.cluster) && !s.IsPaused()
+	return s.Scheduler.IsScheduleAllowed(s.cluster) && !s.IsPaused() && !(s.cluster.GetUnsafeRecoveryController() != nil && s.cluster.GetUnsafeRecoveryController().IsRunning())
 }
 
 // isPaused returns if a scheduler is paused.
--- a/server/cluster/unsafe_recovery_controller.go
+++ b/server/cluster/unsafe_recovery_controller.go
@@ -79,26 +79,34 @@ func newUnsafeRecoveryController(cluster *RaftCluster) *unsafeRecoveryController
 	}
 }
 
+// IsRunning returns whether there is ongoing unsafe recovery process. If yes, further unsafe
+// recovery requests, schedulers, checkers, AskSplit and AskBatchSplit requests are blocked.
+func (u *unsafeRecoveryController) IsRunning() bool {
+	u.RLock()
+	defer u.RUnlock()
+	return u.stage != ready && u.stage != finished
+}
+
 // RemoveFailedStores removes failed stores from the cluster.
-func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string) error {
+func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string, timeout int64) error {
+	if u.IsRunning() {
+		return errs.ErrUnsafeRecoveryIsRunning.FastGenByArgs()
+	}
 	u.Lock()
 	defer u.Unlock()
 	if len(failedStores) == 0 {
-		return errors.Errorf("No store specified")
-	}
-	if u.stage != ready && u.stage != finished {
-		return errors.Errorf("Another request is working in progress")
+		return errs.ErrUnsafeRecoveryInvalidInput.FastGenByArgs("no store specified")
 	}
 	u.reset()
 	for failedStore := range failedStores {
 		store := u.cluster.GetStore(failedStore)
 		if store != nil && (store.IsPreparing() || store.IsServing()) && !store.IsDisconnected() {
-			return errors.Errorf("Store %v is up and connected", failedStore)
+			return errs.ErrUnsafeRecoveryInvalidInput.FastGenByArgs(fmt.Sprintf("store %v is up and connected", failedStore))
 		}
 	}
 	for failedStore := range failedStores {
 		err := u.cluster.BuryStore(failedStore, true)
-		if !errors.ErrorEqual(err, errs.ErrStoreNotFound.FastGenByArgs(failedStore)) {
+		if err != nil && !errors.ErrorEqual(err, errs.ErrStoreNotFound.FastGenByArgs(failedStore)) {
 			return err
 		}
 	}
@@ -165,6 +173,7 @@ func (u *unsafeRecoveryController) HandleStoreHeartbeat(heartbeat *pdpb.StoreHea
 				u.executionReports[heartbeat.Stats.StoreId] = heartbeat.StoreReport
 				u.numStoresPlanExecuted++
 				if u.numStoresPlanExecuted == len(u.storeRecoveryPlans) {
+					u.cluster.PauseOrResumeScheduler("all", 0)
 					log.Info("Recover finished.")
 					go func() {
 						for _, history := range u.History() {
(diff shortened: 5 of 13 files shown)


Connor1996 (commented):
Need to pause split too which TiKV performs an `AskSplit` call

server/cluster/unsafe_recovery_controller.go
```diff
@@ -102,6 +102,7 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
 		}
 		u.storeReports[s.GetID()] = nil
 	}
+	u.cluster.PauseOrResumeScheduler("all", 3600)
```
Connor1996, line 105: 1h maybe too long, how about 10min?

bufferflies: The time can load from request, if not, it  can use default value.

server/cluster/unsafe_recovery_controller_test.go
```diff
@@ -516,6 +517,8 @@ func (s *testUnsafeRecoverSuite) TestReportCollection(c *C) {
 func (s *testUnsafeRecoverSuite) TestPlanExecution(c *C) {
 	_, opt, _ := newTestScheduleConfig()
 	cluster := newTestRaftCluster(s.ctx, mockid.NewIDAllocator(), opt, core.NewStorage(kv.NewMemoryKV()), core.NewBasicCluster())
+	// Manually fill the coordinator up to allow calling on cluster.PauseOrResumeSchedulers().
+	cluster.coordinator = newCoordinator(s.ctx, cluster, hbstream.NewTestHeartbeatStreams(s.ctx, cluster.getClusterID(), cluster, true))
```
bufferflies, line 521: It will be better if you checker the scheduler is pause  status.

server/cluster/unsafe_recovery_controller.go
v01dstar: How long should the schedulers be paused does not really matter, right? In my opinion, they should be paused "indefinitely" until released by unsafe recovery controller again.

If you guys agree with it, we can achieve this goal by checking `c.UnsafeReocoveryController().IsRunning()` whenever PD wants to schedule somethings, instead of using `PauseOrResumeScheduler()`, like we did to `AskSplit`.

bufferflies: In your opinion, the recovery duration should be less than 1h, Is it right.

v01dstar: In most cases, it should, but based on the "no assumption" rule, I think it may be better to make it dynamic (suspend all the schedulers until release them explicitly), so it still works in some "extreme cases". I am proposing that, instead of pausing for a fixed amount of time, make it infinite. If everything works fine, this approach may be better. While, the up side of using "pause" is that, if "unsafe recovery" gets stuck, the schedulers won't be blocked forever. However, if something really goes wrong with the unsafe recovery, we need to restart the PD, and, after that, the schedulers will be resumed anyway.

tonyxuqqi: please make it configurable, in case there's need to change in a specific scenario.

v01dstar: It is now configurable through the optional command line argument.

server/cluster/unsafe_recovery_controller_test.go
v01dstar: Done

server/api/unsafe_operation_test.go
```diff
@@ -45,7 +45,13 @@ func (s *testUnsafeAPISuite) TearDownSuite(c *C) {
 }
 
 func (s *testUnsafeAPISuite) TestRemoveFailedStores(c *C) {
-	input := map[uint64]string{1: ""}
+	input := map[string]interface{}{
+		"stores": map[uint64]interface{}{
+			1: 1,
```
Connor1996, line 50: can we use `[]uint64`?

Connor1996 (commented):
rest LGTM

server/api/unsafe_operation_test.go
v01dstar: Done.

tests/pdctl/unsafe/unsafe_operation_test.go
```diff
@@ -49,6 +49,9 @@ func (s *unsafeOperationTestSuite) TestRemoveFailedStores(c *C) {
 	args := []string{"-u", pdAddr, "unsafe", "remove-failed-stores", "1,2,3"}
 	_, err = pdctl.ExecuteCommand(cmd, args...)
 	c.Assert(err, IsNil)
+	args = []string{"-u", pdAddr, "unsafe", "remove-failed-stores", "1,2,3", "3600"}
```
Connor1996, line 52: we may need to add more options like `--do-recreate --allow-learner --timeout 3600`, can we support specifying it with a option name?

server/api/unsafe_operation.go
```diff
@@ -43,15 +44,43 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
...
+	for _, storeStr := range storeList {
+		storeID, err := strconv.ParseUint(storeStr, 10, 64)
+		if err != nil {
+			h.rd.JSON(w, http.StatusBadRequest, "Cannot parse the store ID")
+			return
+		}
+		stores[storeID] = ""
+	}
+	var pauseDuration int64
+	rawPauseDuration, exists := input["pause"]
```
Connor1996, line 72: let's call it `timeout`, also indicate the max execution duration of recovery

tests/pdctl/unsafe/unsafe_operation_test.go
v01dstar: Done.

server/api/unsafe_operation.go
v01dstar: Done.

server/api/unsafe_operation.go
```diff
@@ -43,15 +44,44 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
...
+			return
+		}
+		stores[storeID] = ""
+	}
+	var timeout int64
+	rawTimeout, exists := input["timeout"]
+	if !exists {
+		timeout = 600
+	} else {
+		// String can only be parsed into float
```
Connor1996, line 76: why `strconv.ParseUint` can't?

server/cluster/cluster_worker.go
```diff
@@ -97,6 +100,9 @@ func (c *RaftCluster) ValidRequestRegion(reqRegion *metapb.Region) error {
 
 // HandleAskBatchSplit handles the batch split request.
 func (c *RaftCluster) HandleAskBatchSplit(request *pdpb.AskBatchSplitRequest) (*pdpb.AskBatchSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errors.New("Unsafe recovery is running, splitting needs to be paused")
```
Connor1996, line 104:
```suggestion
		return nil, errors.New("Unsafe recovery is running, split is paused")
```

server/cluster/cluster_worker.go
```diff
@@ -45,6 +45,9 @@ func (c *RaftCluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 
 // HandleAskSplit handles the split request.
 func (c *RaftCluster) HandleAskSplit(request *pdpb.AskSplitRequest) (*pdpb.AskSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errors.New("Unsafe recovery is running, splitting needs to be paused")
```
Connor1996, line 49:
```suggestion
		return nil, errors.New("Unsafe recovery is running, split is paused")
```

tests/pdctl/unsafe/unsafe_operation_test.go
v01dstar: Changed it to a "if" condition, since we currently only have 1 parameter and golang linter does not allow me use "switch".

tools/pd-ctl/pdctl/command/unsafe_command.go
```diff
@@ -67,21 +67,41 @@ func NewRemoveFailedStoresHistoryCommand() *cobra.Command {
 
 func removeFailedStoresCommandFunc(cmd *cobra.Command, args []string) {
 	prefix := fmt.Sprintf("%s/remove-failed-stores", unsafePrefix)
-	if len(args) != 1 {
+	if len(args) != 1 || len(args) != 2 {
```
Connor1996: With `remove-failed-stores 1,2,3 --timeout 60`, seems the len(args) should be 3?

server/api/unsafe_operation.go
v01dstar: https://pkg.go.dev/encoding/json@go1.17.5#Unmarshal

server/cluster/cluster_worker.go
v01dstar: Changed

server/cluster/cluster_worker.go
v01dstar: Changed

tools/pd-ctl/pdctl/command/unsafe_command.go
```diff
@@ -67,21 +68,28 @@ func NewRemoveFailedStoresHistoryCommand() *cobra.Command {
...
 		}
-		stores[strStore] = ""
+		stores = append(stores, store)
 	}
-	postJSON(cmd, prefix, stores)
+	postInput := map[string]interface{}{
+		"stores": stores,
+	}
+	timeout, err := cmd.Flags().GetFloat64("timeout")
+	if err == nil && timeout != 0 {
```
Connor1996, line 89: what if `err != nil`

Connor1996 (commented):
rest LGTM

server/api/unsafe_operation.go
```diff
@@ -36,22 +36,43 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
...
+	storeList := rawStores.([]interface{})
+	if len(storeList) == 0 {
+		h.rd.JSON(w, http.StatusBadRequest, "Invalid input for the stores")
+		return
+	}
+	stores := make(map[uint64]string)
+	for _, rawStore := range storeList {
+		stores[uint64(rawStore.(float64))] = ""
+	}
+	var timeout int64
```
rleungx, line 67:
```suggestion
	timeout := int64(600)
	rawTimeout, exists := input["timeout"].(float64)
	if exist {
	    timeout = int64(rawTimeout)
	}
```

tools/pd-ctl/pdctl/command/unsafe_command.go
```diff
@@ -42,6 +42,7 @@ func NewRemoveFailedStoresCommand() *cobra.Command {
 		Short: "Remove failed stores unsafely",
 		Run:   removeFailedStoresCommandFunc,
 	}
+	cmd.PersistentFlags().Float64("timeout", 0, "timeout in seconds")
```
rleungx, line 45: the default value is 0?

tools/pd-ctl/pdctl/command/unsafe_command.go
v01dstar: Done

tools/pd-ctl/pdctl/command/unsafe_command.go
v01dstar: Thanks for pointing out, it is better to show 600 here which is the "actual" default value.

server/api/unsafe_operation.go
```diff
@@ -36,22 +36,43 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
...
 // Failure 400 {string} string "The input is invalid."
 // Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /admin/unsafe/remove-failed-stores [POST]
 func (h *unsafeOperationHandler) RemoveFailedStores(w http.ResponseWriter, r *http.Request) {
 	rc := getCluster(r)
-	var stores map[uint64]string
-	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &stores); err != nil {
+	var input map[string]interface{}
+	if err := apiutil.ReadJSONRespondError(h.rd, w, r.Body, &input); err != nil {
+		h.rd.JSON(w, http.StatusBadRequest, "Cannot parse the request")
```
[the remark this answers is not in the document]
v01dstar: What do you mean by "directly return"?

server/api/unsafe_operation.go
v01dstar: Changed, slightly different, since we can't put `.(float64)` right after extracting the value from the map.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -112,6 +116,7 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
 		}
 		u.storeReports[s.GetID()] = nil
 	}
+	u.cluster.PauseOrResumeScheduler("all", timeout)
```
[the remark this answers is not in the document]
v01dstar: You mean those who have been triggered already? No

server/api/unsafe_operation.go
```diff
@@ -36,22 +36,43 @@ func newUnsafeOperationHandler(svr *server.Server, rd *render.Render) *unsafeOpe
...
-	if err := rc.GetUnsafeRecoveryController().RemoveFailedStores(stores); err != nil {
+	storeList := rawStores.([]interface{})
+	if len(storeList) == 0 {
+		h.rd.JSON(w, http.StatusBadRequest, "Invalid input for the stores")
+		return
+	}
+	stores := make(map[uint64]string)
+	for _, rawStore := range storeList {
+		stores[uint64(rawStore.(float64))] = ""
+	}
```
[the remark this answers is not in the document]
v01dstar: Done

server/cluster/unsafe_recovery_controller.go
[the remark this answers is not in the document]
v01dstar: @Connor1996 Do you think we need to enforce the timeout to be at least 10 minutes? Or we leave the option to the caller?

server/api/unsafe_operation.go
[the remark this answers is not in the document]
rleungx: We have `ErrorResp` in `ReadJSONRespondError` which does the same thing.

server/cluster/unsafe_recovery_controller.go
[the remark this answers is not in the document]
rleungx: `PauseOrResumeScheduler` only prevents the schedulers from generating the scheduling operator. The other things like checkers also will generate the operator.

server/cluster/cluster_worker.go
```diff
@@ -45,6 +45,9 @@ func (c *RaftCluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 
 // HandleAskSplit handles the split request.
 func (c *RaftCluster) HandleAskSplit(request *pdpb.AskSplitRequest) (*pdpb.AskSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
```
rleungx, line 48: Do we need a lock?

server/cluster/cluster_worker.go
```diff
@@ -97,6 +100,9 @@ func (c *RaftCluster) ValidRequestRegion(reqRegion *metapb.Region) error {
 
 // HandleAskBatchSplit handles the batch split request.
 func (c *RaftCluster) HandleAskBatchSplit(request *pdpb.AskBatchSplitRequest) (*pdpb.AskBatchSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
```
rleungx, line 103: ditto

server/api/unsafe_operation.go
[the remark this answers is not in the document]
v01dstar: Done.

server/cluster/cluster_worker.go
v01dstar: Added a lock inside the `IsRunning` method. As for the call site, I don't think we need a lock. `unsafeRecoveryController` is created once and never get destroyed, If you were worrying about the race condition here.

server/cluster/cluster_worker.go
v01dstar: Same above

server/cluster/unsafe_recovery_controller.go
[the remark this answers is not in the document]
v01dstar: Please take a look at the new code added in `server/schedule/checker/checker_controller.go`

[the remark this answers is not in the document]
v01dstar: BTW, do you feel comfortable we do the same to the schedulers? i.e. checking if `unsafeRecoveryController.IsRunning()` instead of pause schedulers for a fixed amount of time? We may add another cmd in a separate PR to interrupt unsafe recovery process, if you also worry about unsafe recovery hangs forever.

server/cluster/cluster_worker.go
```diff
@@ -45,6 +45,9 @@ func (c *RaftCluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 
 // HandleAskSplit handles the split request.
 func (c *RaftCluster) HandleAskSplit(request *pdpb.AskSplitRequest) (*pdpb.AskSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errors.New("Unsafe recovery is running, split is paused")
```
Connor1996, line 49: return error with unknown type in the response header instead of grpc error

server/cluster/cluster_worker.go
```diff
@@ -97,6 +100,9 @@ func (c *RaftCluster) ValidRequestRegion(reqRegion *metapb.Region) error {
 
 // HandleAskBatchSplit handles the batch split request.
 func (c *RaftCluster) HandleAskBatchSplit(request *pdpb.AskBatchSplitRequest) (*pdpb.AskBatchSplitResponse, error) {
+	if c.GetUnsafeRecoveryController() != nil && c.GetUnsafeRecoveryController().IsRunning() {
+		return nil, errors.New("Unsafe recovery is running, split is paused")
```
Connor1996, line 104: ditto

server/cluster/cluster_worker.go
v01dstar: What's the consideration behind this? I did a little bit code search, seems that header errors are not widely using?

Connor1996: TiKV may panic on the grpc error, but will pass through the error in the header. https://github.com/pingcap/tikv/blob/5c48a7a7a540d45fe301da193ce06aeb2cee00b9/components/pd_client/src/client.rs#L701

v01dstar: @Connor1996 @rleungx Please take a look at the proposed code changes in `server/grpc_service.go`.

server/cluster/unsafe_recovery_controller.go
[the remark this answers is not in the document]
rleungx: For now, it's ok to me. But I think it's a common feature to pause all scheduling. It's better not to rely on the unsafe recovery.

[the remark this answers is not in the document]
v01dstar: Not applicable anymore, we now disallow scheduling during the unsafe recovery process instead of pausing it for a fixed amount of time.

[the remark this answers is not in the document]
v01dstar: PTAL

server/cluster/cluster_worker.go
v01dstar: Same above

Connor1996 (approved):
LGTM

server/cluster/unsafe_recovery_controller.go
```diff
@@ -79,16 +79,22 @@ func newUnsafeRecoveryController(cluster *RaftCluster) *unsafeRecoveryController
 	}
 }
 
+func (u *unsafeRecoveryController) IsRunning() bool {
```
JmPotato, line 82: A comment is needed.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -79,16 +79,22 @@ func newUnsafeRecoveryController(cluster *RaftCluster) *unsafeRecoveryController
...
+	u.RLock()
+	defer u.RUnlock()
+	return u.stage != ready && u.stage != finished
+}
+
 // RemoveFailedStores removes failed stores from the cluster.
-func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string) error {
+func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string, timeout int64) error {
+	if u.IsRunning() {
+		return errors.Errorf("Another request is working in progress")
```
JmPotato, line 91: Same problem as above.

server/cluster/coordinator.go
```diff
@@ -711,6 +715,19 @@ func (c *coordinator) pauseOrResumeScheduler(name string, t int64) error {
 	return err
 }
 
+func (c *coordinator) isSchedulerAllowed(name string) (bool, error) {
```
rleungx, line 718: Is it only used for test purposes?

Connor1996:
ping @v01dstar

server/cluster/cluster_worker.go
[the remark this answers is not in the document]
v01dstar: Done

server/cluster/coordinator.go
v01dstar: Yes

rleungx: Please add a comment.

v01dstar: Added.

tests/pdctl/unsafe/unsafe_operation_test.go
```diff
@@ -49,6 +49,12 @@ func (s *unsafeOperationTestSuite) TestRemoveFailedStores(c *C) {
 	args := []string{"-u", pdAddr, "unsafe", "remove-failed-stores", "1,2,3"}
 	_, err = pdctl.ExecuteCommand(cmd, args...)
 	c.Assert(err, IsNil)
+	args = []string{"-u", pdAddr, "unsafe", "remove-failed-stores", "1,2,3", "--timeout", "3600"}
+	_, err = pdctl.ExecuteCommand(cmd, args...)
+	c.Assert(err, IsNil)
+	args = []string{"-u", pdAddr, "unsafe", "remove-failed-stores", "1,2,3", "--timeout", "abc"}
+	_, err = pdctl.ExecuteCommand(cmd, args...)
+	c.Assert(err, Not(IsNil))
```
rleungx, line 57:
```suggestion
	c.Assert(err, NotNil))
```

server/cluster/unsafe_recovery_controller.go
```diff
@@ -79,26 +79,34 @@ func newUnsafeRecoveryController(cluster *RaftCluster) *unsafeRecoveryController
...
+// recovery requests, schedulers, checkers, AskSplit and AskBatchSplit requests are blocked.
+func (u *unsafeRecoveryController) IsRunning() bool {
+	u.RLock()
+	defer u.RUnlock()
+	return u.stage != ready && u.stage != finished
+}
+
 // RemoveFailedStores removes failed stores from the cluster.
-func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string) error {
+func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string, timeout int64) error {
```
rleungx, line 91: Where does the timeout being used?

tests/pdctl/unsafe/unsafe_operation_test.go
v01dstar: N/A

server/cluster/unsafe_recovery_controller.go
v01dstar: It wasn't used, removed.

Connor1996:
@v01dstar Please add back the timeout parameter, it'll will be used

v01dstar:
> @v01dstar Please add back the timeout parameter, it'll will be used

Maybe we should add it iff we need it. It is indeed a little bit weird to introduce a non-using parameter.

Connor1996:
> > @v01dstar Please add back the timeout parameter, it'll will be used
> 
> Maybe we should add it iff we need it. It is indeed a little bit weird to introduce a non-using parameter.

It should be used now to make sure the scheduler is paused with a timeout

JmPotato (approved):

PR #4479, opened 2021-12-20
 ctl: pd-ctl cmd will work well in char device mode

### LINKED ISSUE #4478: tools: pd-ctl cmd  will work well in crontab
## Bug Report
#!/bin/sh ./.tiup/components/ctl/v5.0.0-rc/pd-ctl -u http://xxx.xx.xx.xx:2379 scheduler show is ok.

But it will work wrong if I run the cmd in crontab:

image

### What did you do?
```
 * * * * * /data2/robert/soft/golang/pd/bin/pd-ctl -u xxx:xxx:xxx:xxx:2379 scheduler show >> /data2/robert/cron.log 2>&1
```

It will run well in crontab 
5.0.0~master: fe6fab92

bufferflies:
close #4478

String split mention If s does not contain sep and sep is not empty, Split returns a
slice of length 1 whose only element is s. It will add args if the arg is nil in char device mode.
So it needs to check the args is nil. 
### Check List

Tests

image

Code changes

Side effects

Related changes

### Release note

```release-note
pdctl args lens is wrong in char device mode.
```

### DIFF
--- a/tools/pd-ctl/main.go
+++ b/tools/pd-ctl/main.go
@@ -57,7 +57,10 @@ func main() {
 			fmt.Println(err)
 			return
 		}
-		input = strings.Split(strings.TrimSpace(string(b)), " ")
+		s := strings.TrimSpace(string(b))
+		if s != "" {
+			input = strings.Split(s, " ")
+		}
 	}
 
 	pdctl.MainStart(append(os.Args[1:], input...))
--- a/tools/pd-ctl/main.go
+++ b/tools/pd-ctl/main.go
@@ -53,6 +53,7 @@ func main() {
 		in, err := pdctl.ReadStdin(os.Stdin)
 		if err != nil {
 			fmt.Println(err)
+			return
 		}
 		inputs = in
 	}
--- a/tools/pd-ctl/main.go
+++ b/tools/pd-ctl/main.go
@@ -16,10 +16,8 @@ package main
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"os/signal"
-	"strings"
 	"syscall"
 
 	"github.com/tikv/pd/tools/pd-ctl/pdctl"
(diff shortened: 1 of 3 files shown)


lhy1024 (commented):
Do we need to add a test for it？

bufferflies:
> Do we need to add a test for it？

I think it should need.

lhy1024 (approved):

tools/pd-ctl/pdctl/ctl.go
```diff
@@ -198,3 +198,15 @@ func genCompleter(cmd *cobra.Command) []readline.PrefixCompleterInterface {
...
 	return pc
 }
+
+// ReadStdin convert stdin to string array
+func ReadStdin(r io.Reader) (input []string, err error) {
+	b, err := io.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+	if s := strings.TrimSpace(string(b)); s != "" {
```
JmPotato, line 208:
```suggestion
	if s := strings.TrimSpace(string(b)); len(s) > 0 {
```

tools/pd-ctl/main.go
```diff
@@ -49,16 +47,14 @@ func main() {
...
 
-	var input []string
+	var inputs []string
 	stat, _ := os.Stdin.Stat()
 	if (stat.Mode() & os.ModeCharDevice) == 0 {
-		b, err := io.ReadAll(os.Stdin)
+		in, err := pdctl.ReadStdin(os.Stdin)
 		if err != nil {
 			fmt.Println(err)
-			return
```
JmPotato, line 58: Why not `return err` here?

tools/pd-ctl/pdctl/ctl.go
```diff
@@ -198,3 +198,15 @@ func genCompleter(cmd *cobra.Command) []readline.PrefixCompleterInterface {
 	}
 	return pc
 }
+
+// ReadStdin convert stdin to string array
+func ReadStdin(r io.Reader) (input []string, err error) {
```
JmPotato, line 203: Since the name of it is `ReadStdin`, I think there is no need to pass the `os.Stdin` through a parameter?

JmPotato (commented):
The rest LGTM.

tools/pd-ctl/pdctl/ctl.go
bufferflies: the parameter design to test simple.

JmPotato (approved):

PR #4498, opened 2021-12-24
api, pd-ctl: fix the region range hole end key check and add some help info

### LINKED ISSUE #4216: Add some help info for the `region range-holes` command

Part of #4187. After merging https://github.com/tikv/pd/pull/4202, we have a `pd-ctl` command `region range-holes`, but it lacks some help info.

Add some help info for the `region range-holes` command.

### LINKED ISSUE #4496: `range-holes` should check if the end key is `""`
## Bug Report

Consider such a case that we have only two regions with the given range:
```
region A: ["", "a")
region B: ["a", "b")
```

Actually, we have a range hole from `b` to `""`, but we cannot get the right result when using `range-holes` API.

master

JmPotato:

Close #4496, close #4216.

* Check if the last end key is not empty.
* Add some help info for the `region range-holes` command.

### Check List

Tests

- Unit test
- Integration test

```shell
❯ ./bin/pd-ctl region range-holes -h
There are some cases that the region range is not continuous, for example, the region doest't send the heartbeat to PD after a spliting.
This command will output all empty ranges without any region info.

Usage:
  pd-ctl region range-holes [flags]

Examples:

  If PD now holds the region ranges info like ["", "a"], ["b", "x"], ["x", "z"]. The the output will be like:

  [
    [
      "a",
      "b"
    ],
    [
      "z",
      ""
    ],
  ]


Flags:
  -h, --help   help for range-holes

Global Flags:
      --cacert string   path of file that contains list of trusted SSL CAs
      --cert string     path of file that contains X509 certificate in PEM format
      --key string      path of file that contains X509 key in PEM format
  -u, --pd string       address of pd (default "http://127.0.0.1:2379")
```

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/region_command.go
+++ b/tools/pd-ctl/pdctl/command/region_command.go
@@ -519,12 +519,33 @@ func showRegionWithStoreCommandFunc(cmd *cobra.Command, args []string) {
 	cmd.Println(r)
 }
 
+const (
+	rangeHolesLongDesc = `There are some cases that the region range is not continuous, for example, the region doest't send the heartbeat to PD after a spliting.
+This command will output all empty ranges without any region info.`
+	rangeHolesExample = `
+  If PD now holds the region ranges info like ["", "a"], ["b", "x"], ["x", "z"]. The the output will be like:
+
+  [
+    [
+      "a",
+      "b"
+    ],
+    [
+      "z",
+      ""
+    ],
+  ]
+`
+)
+
 // NewRangesWithRangeHolesCommand returns ranges with range-holes subcommand of regionCmd
 func NewRangesWithRangeHolesCommand() *cobra.Command {
 	r := &cobra.Command{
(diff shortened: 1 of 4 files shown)


tools/pd-ctl/pdctl/command/region_command.go
```diff
@@ -519,12 +519,33 @@ func showRegionWithStoreCommandFunc(cmd *cobra.Command, args []string) {
 	cmd.Println(r)
 }
 
+const (
+	rangeHolesLongDesc = `There are some cases that the region range is not continuous, for example, the region doest't send the heartbeat to PD after a spliting.
```
rleungx, line 523:
```suggestion
	rangeHolesLongDesc = `There are some cases that the region range is not continuous, for example, the region doesn't send the heartbeat to PD after a splitting.
```

disksing (approved):

rleungx (approved):

PR #4508, opened 2021-12-28
api: return more replication info

### LINKED ISSUE #4509: pd: mirror issue of pingcap/tidb#30960
## Development Task

Check pingcap/tidb#30960

xhebox:
Signed-off-by: xhe <xw897002528@gmail.com>

Previously, only a boolean value is returned to indicate whether it is scheduled. But there are a lot of cases, when pd can not really make the schedule and waiting for something. A typical example is you ask for more replications than there is. And this requirement origins from pingcap/tidb#30960.

Close #4509

Now it returns a three-valued string.

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

### Release note

```release-note
return three-valued state string for regions replication check api
```


server/api/region.go
```diff
@@ -262,14 +262,20 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
 	}
 
 	regions := rc.ScanRegions(startKey, endKey, -1)
-	replicated := true
+	state := "REPLICATED"
 	for _, region := range regions {
 		if !opt.IsRegionReplicated(rc, region) {
-			replicated = false
+			state = "INPROGRESS"
+			for _, item := range rc.GetCoordinator().GetWaitingRegions() {
```
rleungx, line 269: What is the difference between `INPROGRESS` and `WAITING`. Sound like they have a similar meaning. BTW, waiting regions is a cache that may be empty sometimes.

server/api/region.go
```diff
@@ -262,14 +262,20 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
...
+				if item.Key == region.GetID() {
+					state = "WAITING"
+					break
+				}
+			}
 			break
 		}
 	}
-	h.rd.JSON(w, http.StatusOK, replicated)
+	h.rd.JSON(w, http.StatusOK, state)
```
lcwangchao, line 278: I think we can return a json object instead of the pure string for easy to extend in the future.

xhebox: It is actually JSON.

server/api/region.go
xhebox: `INPROGRESS` means there is an operator executing, while `WAITING` basically means the scheduling stalls somehow. If you have a better name, I'd like to replace them.

xhebox: It does not really matter if it is empty. This is not so strict anyway as long as it is expected under good network and load condition.

xhebox: It is changed to "PENDING" according to morgo, https://github.com/pingcap/tidb/pull/31073#issuecomment-1002610790

server/api/region.go
```diff
@@ -262,14 +262,20 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
 	}
 
 	regions := rc.ScanRegions(startKey, endKey, -1)
-	replicated := true
+	state := "REPLICATED"
```
rleungx, line 265: If the length of regions is 0, will it return "REPLICATED"?

xhebox: Yes, it is expected.

disksing: we already have a pending status means that a region peer has log lag. Maybe we can choose another word?

xhebox: I can't come up with one, any suggestion?

xhebox: @disksing What about 'PAUSE'?

disksing: how about `WAITING`? because the region is in the waiting list.

xhebox: > What is the difference between INPROGRESS and WAITING. Sound like they have a similar meaning. BTW, waiting regions is a cache that may be empty sometimes.


Have used that before, but from @rleungx , can we just decide one? :joy:

disksing (approved):

server/api/region.go
disksing: Then I guess any one is fine.

rleungx (approved):

PR #4512, opened 2021-12-28
statistic: replace old inherit with adopt in hot statistic

### LINKED ISSUE #4510: statistic: refactor inherit in hot statistic
## Enhancement Task

For the current hotspot statistics, we want to be able to use the previous data after move-peer and transfer-leader to reduce fluctuations and increase stability.

So we save the statistics of the source peer of the scheduling for the next time, but with many refactorings, this mechanism has failed.

For move peer, since we now handle hot data at peer granularity, and considering that we actually add learner, promote learner, and then remove peer for each move peer, the temporary save for inherit will no longer be used as expected. Here is an example：

> At T0, there are three peer in store1, store2 and store3.
> At T1, there is a operator, which move peer from store1 to store4. 
> At T2, adding learner in store4, and its data is adopted from store1, store2 and store3 randomly at T3
> At T3, remove peer in store1 and put its data to inherit items
> At T4-T10, there is read and write workload.
> At T11, there is a operator, which move peer from store2 to store5.
> At T12, add learner in store5, and its data is from store1 at T3, because inherit items is not empty 

|      | store1 | store2 | store3 | store4                   | store5                                         | Inherit  items   |
| ---- | ------ | ------ | ------ | ------------------------ | ---------------------------------------------- | ----------- |
| T0   | peer1  | peer2  | peer3  |                          |                                                | empty       |
| T2  | peer1  | peer2  | peer3  | peer4(copy random at t1)(blue) |                                                | empty       |
| T3   | remove | peer2  | peer3  | peer4                    |                                                | peer1 at t1 |
| T12   |        | peer2  | peer3  | peer4                    | peer5 is from inherit, as known as peer1 at t1(red) |             |

middle_img_v2_223ceedc-0fa5-4bf8-aecc-c2259327062g

For the transfer leader, after the introduction of stale read, the data of the transfer leader will no longer be inherited from the leader but will be directly used from the target store.

So:
1. for transfer leader, we expect to inherit old data in strong read and write, but not.
2. for move peer, we inherit error data.

| inherit     | transfer leader | move peer        |
| ----------- | --------------- | ---------------- |
| stale read  | no inherit             | inherit error data          |
| strong read | no inherit (expect to inherit)  | inherit error data (expect no to inherit) |
| write       | no inherit (expect to inherit)   | inherit error data            |

For the above reasons, I have the following suggestions

1. remove the inherit-related mechanism and keep only direct and adopt
2. for move peer, get the data from the operator's queue, so as to determine which operator the step of add learner actually belongs to, and then specify the source peer to be copied, instead of adopting one at random.
3. for transfer leader, under strong read, mark the difference between stale read and strong read from the tikv side, but I think his development cost is a bit high, and CoolDownTransferLeader will avoid frequent scheduling.

I prefer to use option 1 or 2

lhy1024:

To close #4510 

- remove inherit in hot statistic
- rename `directly` to `inherit`

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/statistics/kind.go
+++ b/server/statistics/kind.go
@@ -102,8 +102,7 @@ type sourceKind int
 
 const (
 	direct  sourceKind = iota // there is a corresponding peer in this store.
-	inherit                   // there is no a corresponding peer in this store and there is a peer just deleted.
-	adopt                     // there is no corresponding peer in this store and there is no peer just deleted, we need to copy from other stores.
+	inherit                   // there is no corresponding peer in this store and we need to copy from other stores.
 )
 
 func (k sourceKind) String() string {
@@ -112,8 +111,6 @@ func (k sourceKind) String() string {
 		return "direct"
 	case inherit:
 		return "inherit"
-	case adopt:
-		return "adopt"
 	}
 	return "unknown"
 }
--- a/server/statistics/hot_peer.go
+++ b/server/statistics/hot_peer.go
@@ -107,11 +107,11 @@ type HotPeerStat struct {
 	// If the peer didn't been send by store heartbeat when it is already stored as hot peer stat,
 	// we will handle it as cold peer and mark the inCold flag
 	inCold bool
-	// source represents the statistics item source, such as directly, inherit, adopt.
+	// source represents the statistics item source, such as direct, inherit.
 	source sourceKind
-	// If the item in storeA is just adopted from storeB,
-	// then other store, such as storeC, will be forbidden to adopt from storeA until the item in storeA is hot.
-	allowAdopt bool
+	// If the item in storeA is just inherited from storeB,
+	// then other store, such as storeC, will be forbidden to inherit from storeA until the item in storeA is hot.
+	allowInherited bool
 }
 
 // ID returns region ID. Implementing TopNItem.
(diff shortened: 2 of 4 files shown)


rleungx:
We may need some tests for this PR.

lhy1024:
> We may need some tests for this PR.

yep, I will test it with regression tests

lhy1024:
> > We may need some tests for this PR.
> 
> yep, I will test it with regression tests

Its query is higher, and its stddev is less

image

rleungx (approved):

server/statistics/kind.go
```diff
@@ -101,15 +101,12 @@ func (k StoreStatKind) String() string {
 type sourceKind int
 
 const (
-	direct  sourceKind = iota // there is a corresponding peer in this store.
-	inherit                   // there is no a corresponding peer in this store and there is a peer just deleted.
-	adopt                     // there is no corresponding peer in this store and there is no peer just deleted, we need to copy from other stores.
+	inherit sourceKind = iota // there is a corresponding peer in this store.
```
nolouch, line 104: Does use `direct, inherit` more suitable? change adopt to inherit and remove old inherit.

lhy1024: Good idea

lhy1024:
@nolouch PTAL

HunDunDM (commented):
LGTM, but can we add some tests that will fail before modification?

lhy1024:
According to #4510, I add a test and it failed in master
image

HunDunDM (approved):

PR #4523, opened 2021-12-29
dr-ausosync: add new intermediate state

### LINKED ISSUE #4522: DR-AutoSync: Add intermediate state to avoid break ACID integraty
## Development Task

1. Add new state `ASYNC_WAIT`. When DR is down or disconnected with Primary, the state will first switch to `ASYNC_WAIT`. Then after all available TiKV servers have switched to `ASYNC_WAIT`, PD can continue switch state to `ASYNC`.
2. Add `available_stores` field to replication status. When the state is `ASYNC_WAIT` or `ASYNC`, TiKV should make sure that they don't exchange messages with TiKV servers that are not in the list. In this way, we can guarentee the ACID integraty of DR.
3. Add `drStoreStatus` (a map) field to track replication state of TiKV. TiKV reports its state to PD throught StoreHeartbeat.
4. Make sure PD only switch state from `ASYNC_WAIT` to `ASYNC` when all available stores have switched to the latest state. (they won't exchange messages with DR nodes)
5. Make sure when the state is `ASYNC` and the `available_stores` list needs to change, it can only change when all stores in the new `availabel_stores` list are up to date. (prevent them from exchanging messages with DR nodes)

disksing:

close #4522

### Check List

Tests
- Unit test

Code changes
- Has persistent data change: introduced new state

Side effects
- Breaking backward compatibility: the replication mode feature cannot work with old tikv

### Release note

```release-note
None
```


disksing:
Please take a look. @lhy1024 @nolouch

server/replication/replication_mode.go
```diff
@@ -396,78 +411,158 @@ func (m *ModeManager) tickDR() {
...
+
+func (m *ModeManager) drCheckStoreStateUpdated(stores []uint64) bool {
+	state := m.GetReplicationStatus().GetDrAutoSync()
+	for _, s := range stores {
+		status, ok := m.drStoreStatus.Load(s)
+		if !ok {
+			return false
+		}
+		drStatus := status.(*pb.StoreDRAutoSyncStatus)
+		if drStatus.GetState() != state.GetState() || drStatus.GetStateId() != state.GetStateId() {
```
nolouch, line 561: Which state of the store should be in `AsyncWait`?

nolouch (commented):
Does `AsynWait` cannot be directly recovered? there should has ACID issue?

server/replication/replication_mode.go
disksing: `AsyncWait` is the intermediate state from `sync` to `async`. (the [chart](https://github.com/tikv/pd/blob/656ce79a6c4bbacc6b7300453359ead6be7e2595/server/replication/replication_mode.go#L440-L456) shows how states change) We can recover dr replica from `AsyncWait` state, it will not have ACID issue.

nolouch: once the TiKV switch to `AsyncWait`, what tikv's behavior looks like？

disksing: Yes. Though it is not implemented in tikv yet. Now TiKV only switches states.

server/replication/replication_mode.go
```diff
@@ -396,78 +411,158 @@ func (m *ModeManager) tickDR() {
...
+	*/
+
+	switch m.drGetState() {
+	case drStateSync:
+		// If hasMajority is false, the cluster is always unavailable. Switch to async won't help.
+		if !canSync && hasMajority && m.drCheckAsyncTimeout() {
+			m.drSwitchToAsyncWait(stores[primaryUp])
+		}
+	case drStateAsyncWait:
+		if canSync {
```
nolouch, line 465: Can this directly switch to `sync`? may the follower replica have a large logging lag during `AsyncWait`?

nolouch (commented):
rest LGTM

server/replication/replication_mode.go
disksing: It's ok. In `AsyncWait` state, TiKV cannot write data so there will be no log lag.

nolouch: `Sync` -> `AsyncWait` already wait for a timeout, why cannot write data? In my understanding, once switch to `AsyncWait`, it means there are majority replicas in primary DC, but cannot replicate to DR DC's replica.

disksing: | State  | Can commit raftlog locally | Transport between DC |
| ------------- | ------------- | --------------- |
| Sync  | No | Connected |
| AsyncWait  | No  | Disconnected |
| Async | Yes | Disconnected |

Without the `AsyncWait` state, it will appear that some tikv are in Sync state and some TiKV are in Async state. Because TiKV in Async state can commit successfully with only 1DC participating in the replication log, the ACID of the dr dc can be broken.
From the table we can see in `AsyncWait` state, the network is cut off, and it cannot commit raftlog locally, so it in fact cannot commit raft log successfully.

nolouch: got

nolouch (approved):
lgtm

server/replication/replication_mode.go
```diff
@@ -396,78 +411,158 @@ func (m *ModeManager) tickDR() {
...
-
-	if m.drGetState() == drStateSyncRecover {
-		m.updateProgress()
-		progress := m.estimateProgress()
-		drRecoverProgressGauge.Set(float64(progress))
-
-		if progress == 1.0 {
+	/*
+
+	           +----+      all region sync     +------------+
```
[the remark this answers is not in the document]
disksing: If one of primary hosts is down, the state should be unchange. I think it's ok to now show it.

server/grpc_service.go
```diff
@@ -1787,3 +1791,8 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
 	err = server.Send(&pdpb.WatchGlobalConfigResponse{Changes: ls})
 	return err
 }
+
+// ReportBuckets receives region buckets from tikv.
```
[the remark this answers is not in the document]
disksing: emm, but the RPC is named buckets.

lhy1024 (approved):

PR #4564, opened 2022-01-12
schedule: prevent removing voters directly in 2 replica raft group

### LINKED ISSUE #4411: Removing voter from 2 voter replicas raft group would make region unavailable

HunDunDM:

close #4411 

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
Prevent removing a voter from 2 replicas raft group directly, instead, demote the voter firstly.
```

### DIFF
--- a/server/schedule/operator/builder_test.go
+++ b/server/schedule/operator/builder_test.go
@@ -167,7 +167,6 @@ func (s *testBuilderSuite) TestPrepareBuild(c *C) {
 
 func (s *testBuilderSuite) TestBuild(c *C) {
 	type testCase struct {
-		name              string
 		useJointConsensus bool
 		originPeers       []*metapb.Peer // first is leader
 		targetPeers       []*metapb.Peer // first is leader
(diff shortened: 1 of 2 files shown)


server/schedule/operator/builder_test.go
```diff
@@ -496,10 +495,40 @@ func (s *testBuilderSuite) TestBuild(c *C) {
...
+					PromoteLearners: []PromoteLearner{},
+					DemoteVoters:    []DemoteVoter{{ToStore: 1}},
+				},
+				RemovePeer{FromStore: 1},
+			},
+		},
 	}
 
 	for _, tc := range cases {
-		c.Log(tc.name)
```
lhy1024, line 502: Why remove it?

HunDunDM: It prints unnecessary and malformed logs.

nolouch (approved):
lgtm

server/schedule/operator/builder_test.go
HunDunDM: I made a mistake, I will roll back this part of the edit

PR #4569, opened 2022-01-12
client: use separate go.mod

### LINKED ISSUE #4568: client: use separate go.mod
## Development Task

Make it possible to import pd client with more lightweight dependency.

disksing:
Fix #4568

- add go.mod for pd/client
- copy utility packages to `pd/client` (`errs` / `grpcutil` / `testutil`)
- avoid `pd/client` depending on `etcd` (add `tlsutil`)
- add go.mod for `pd-tso-bench` (because it uses client)
- add go.mod for `tests/client`

### Check List

Tests
- Unit test

### Release note
```release-note
None
```


client/errs/errno.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020 PingCAP, Inc.
```
JmPotato, line 1: Should be like this?

```suggestion
// Copyright 2022 TiKV Project Authors.
```

client/grpcutil/grpcutil.go
```diff
@@ -0,0 +1,139 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
+)
+
+// ForwardMetadataKey is used to record the forwarded host of PD.
+const ForwardMetadataKey = "pd-forwarded-host"
+
+// TLSConfig is the configuration for supporting tls.
+type TLSConfig struct {
```
rleungx, line 34: How about moving them to `tlsconfig.go`?

rleungx:
I found there are lots of logs in https://github.com/tikv/pd/runs/5124147745?check_suite_focus=true even if there is no any error.

disksing:
> I found there are lots of logs in https://github.com/tikv/pd/runs/5124147745?check_suite_focus=true even if there is no error.

Thanks it's now fixed.

Makefile
```diff
@@ -147,17 +147,20 @@ install-tools:
...
-	@revive -formatter friendly -config revive.toml $(PACKAGES)
+	@ revive -formatter friendly -config revive.toml $(PACKAGES)
+
+	@ for mod in $(SUBMODULES); do cd $$mod && $(MAKE) static && cd - > /dev/null; done
 
 tidy:
-	go mod tidy
-	git diff --quiet go.mod go.sum
+	@ go mod tidy
+	@ git diff --quiet go.mod go.sum
```
JmPotato, line 161: I suggest adding some output info for this command, otherwise, it will be kind of confusing for a new-bee developer to know where the thing goes wrong.

client/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	cd .. && $(MAKE) install-tools
+
+static: install-tools
+	@ gofmt -s -l -d . 2>&1 | awk '{ print } END { if (NR > 0) { exit 1 } }'
+	@ golangci-lint run -c ../.golangci.yml ./...
+	@ revive -formatter friendly -config ../revive.toml .
+
+tidy:
+	@ go mod tidy
+	@ git diff --quiet go.mod go.sum
```
JmPotato, line 38: Ditto.

client/errs/errno.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	// NotLeaderErr indicates the the non-leader member received the requests which should be received by leader.
+	NotLeaderErr = "is not leader"
+	// MismatchLeaderErr indicates the the non-leader member received the requests which should be received by leader.
+	MismatchLeaderErr = "mismatch leader id"
+)
+
+// client errors
+var (
+	ErrClientCreateTSOStream = errors.Normalize("create TSO stream failed", errors.RFCCodeText("PD:client:ErrClientCreateTSOStream"))
```
JmPotato, line 28: This line changed recently, please update it with the latest master.

tests/client/Makefile
```diff
@@ -0,0 +1,50 @@
...
+SHELL := env PATH='$(PATH)' GOBIN='$(GO_TOOLS_BIN_PATH)' $(shell which bash)
+
+static: install-tools
+	@ gofmt -s -l -d . 2>&1 | awk '{ print } END { if (NR > 0) { exit 1 } }'
+	@ golangci-lint run ./...
+	@ revive -formatter friendly -config ../../revive.toml .
+
+tidy:
+	@ go mod tidy
+	@ git diff --quiet go.mod go.sum
```
JmPotato, line 26: Same here.

rleungx (approved):

Makefile
```diff
@@ -82,7 +82,7 @@ pd-server-basic:
 pd-ctl:
 	CGO_ENABLED=0 go build -gcflags '$(GCFLAGS)' -ldflags '$(LDFLAGS)' -o $(BUILD_BIN_PATH)/pd-ctl tools/pd-ctl/main.go
 pd-tso-bench:
-	CGO_ENABLED=0 go build -o $(BUILD_BIN_PATH)/pd-tso-bench tools/pd-tso-bench/main.go
+	cd tools/pd-tso-bench && CGO_ENABLED=0 go build -o $(BUILD_BIN_PATH)/pd-tso-bench main.go
```
rleungx, line 85: BTW, why do we need to change it?

Makefile
disksing: ok

client/errs/errno.go
disksing: ok. I'll recheck copied code.

Makefile
disksing: we need to cd to the directory to use the `go.mod` of submodule.

JmPotato (approved):

Makefile
rleungx: Oh, I get it.

PR #4570, opened 2022-01-13
region_scatterer: fix the bug that could generate schedule with too few peers

### LINKED ISSUE #4565: Peer is accidently removed when scatter region
## Bug Report

Screen Shot 2022-01-12 at 4 28 21 PM

### What did you do?

Use API to scatter regions.

There should be only `scatter-region` operators.

### What did you see instead?
Screen Shot 2022-01-12 at 4 31 55 PM

master and v5.4.0

HunDunDM:

close #4565

If an originPeer is already in `selectStore`, `continue` directly. Otherwise, it will end up picking a `targetPeer` which is either itself or some other store that is not in the initial peer list. Since entering the select phase, it is guaranteed that it is not in the `selectStore`, so it is always guaranteed that the worst-case selection itself is feasible at this time. In the end, the number of peers before and after can be guaranteed to remain unchanged.

### Check List

Tests

- Unit test

When not modified, the following operator may be generated:

scatter-region {rm peer: store [3]} (kind:region,leader, region:201(0,0), createAt:2022-01-13 18:24:32.375399 +0800 CST m=+0.014571751, startAt:0001-01-01 00:00:00 +0000 UTC, currentStep:0, steps:[transfer leader from store 3 to store 4, remove peer on store 3])

Code changes

Side effects

Related changes

### Release note

```release-note
Fix the bug that the region scatterer may generate the schedule with too few peers.
```

### DIFF
--- a/server/schedule/region_scatterer.go
+++ b/server/schedule/region_scatterer.go
@@ -273,7 +273,7 @@ func (r *RegionScatterer) Scatter(region *core.RegionInfo, group string) (*opera
 
 func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string) *operator.Operator {
 	ordinaryFilter := filter.NewOrdinaryEngineFilter(r.name)
-	ordinaryPeers := make(map[uint64]*metapb.Peer)
+	ordinaryPeers := make(map[uint64]*metapb.Peer, len(region.GetPeers()))
 	specialPeers := make(map[string]map[uint64]*metapb.Peer)
 	// Group peers by the engine of their stores
 	for _, peer := range region.GetPeers() {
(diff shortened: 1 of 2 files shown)


lhy1024 (approved):

server/schedule/region_scatterer.go
```diff
@@ -323,6 +323,14 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string) *
 		scatterWithSameEngine(peers, ctx)
 	}
 
+	// If the number of target peers is too small, change it back to the original.
```
nolouch, line 326: I think it's better add a filter that can filter other peer stores in `selectCandidates `, like:
```
func (r *RegionScatterer) selectCandidates(region *core.RegionInfo, sourceStoreID uint64, selectedStores map[uint64]struct{}, context engineContext) []uint64 {
...
excludedPeersStores := make(map[uint64]struct{})
		if peer.GetId() != sourceStoreID {
			excludedPeersStores[peer.GetStoreId()] = struct{}{}
		}
	}
	filters := []filter.Filter{
		filter.NewExcludedFilter(r.name, nil, selectedStores),
		filter.NewExcludedFilter(r.name, nil, excludedPeersStores),
	}
...
}

```
This will make the region distribution more dispersed than the current strategy.

HunDunDM: It is necessary to have some regions unchanged.

nolouch: the goal is scatter them.

HunDunDM: itself is excluded, it will choose the smallest of the remaining ones resulting in redundant scheduling.

Yisaer (approved):

nolouch (approved):

PR #4575, opened 2022-01-13
test: update github action

disksing:
CI (especially jenkin CI) runs too slow.

- All-in GitHub Actions
- Split unit tests by weight
- Run unit test and coverage in parallel 

After merging this PR, we can discard unit test in Jenkins

### Check List

Tests
- Unit test

### Release note
```release-note
None
```


scripts/ci-subtask.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# ./ci-subtask.sh <TOTAL_TASK_N> <TASK_INDEX>
+
+packages=(`go list ./...`)
+dirs=(`find . -iname "*_test.go" -exec dirname {} \; | sort -u | sed -e "s/^\./github.com\/tikv\/pd/"`)
+tasks=($(comm -12 <(printf "%s\n" "${packages[@]}") <(printf "%s\n" "${dirs[@]}")))
+
+weight () {
+    [[ $1 == "github.com/tikv/pd/server/api" ]] && return 30
+    [[ $1 == "github.com/tikv/pd/server/tso" ]] && return 30
```
JmPotato, line 11: All TSO tests are in the `tests` folder, and they are running independently through `tso-function-test`. I think we should exclude them here.

```suggestion
    [[ $1 == "github.com/tikv/pd/tests/server/tso" ]] && return 30
```

.github/workflows/pd-tests.yaml
```diff
@@ -16,13 +16,11 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        worker_id: [1, 2, 3, 4]
-    outputs:
-      job-total: ${{ strategy.job-total }}
+        worker_id: [1, 2, 3, 4, 5, 6, 7, 8]
```
bufferflies, line 19: why the work_id is  different between pd-cover and pd-tests ?

disksing: `pd-cover` does not enable `race` or `deadlock` so it can run faster. From the test time cost, maybe, change to `pd-cover: 4, pd-tests: 6` is more suitable

server/schedule/region_scatterer_test.go
```diff
@@ -63,28 +63,13 @@ var _ = Suite(&testScatterRegionSuite{})
...
-
-func (s *testScatterRegionSuite) TestSixSpecialStores(c *C) {
-	s.scatterSpecial(c, 3, 6, 100)
-	s.scatterSpecial(c, 3, 6, 1000)
-}
-
-func (s *testScatterRegionSuite) TestFiveSpecialStores(c *C) {
-	s.scatterSpecial(c, 5, 5, 100)
-	s.scatterSpecial(c, 5, 5, 1000)
+func (s *testScatterRegionSuite) TestScatterRegions(c *C) {
```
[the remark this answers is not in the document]
disksing: it runs too slow when `-race` is enabled.

JmPotato (approved):

rleungx (approved):

Issue #4284, opened 2021-11-04
Unify the key path management


JmPotato:
### Describe your feature request related problem

PD server will store many different keys into the backend etcd storage in terms of different usage. However, the key path management now is very chaotic. For example, we have two different places to make a bootstrap time key.

https://github.com/tikv/pd/blob/e5d7be15786a1a369106131962b835c4e5736891/server/util.go#L138-L140

https://github.com/tikv/pd/blob/e5d7be15786a1a369106131962b835c4e5736891/server/core/storage.go#L129-L132

Besides, the etcd kv is initialized with the `rootPath`, the developer may easily miss out on this point to access a key with the wrong key path.

https://github.com/tikv/pd/blob/e5d7be15786a1a369106131962b835c4e5736891/server/kv/etcd_kv.go#L42-L48

### Describe the feature you'd like

We better introduce a data structure or interface to provide a unified key path management feature.

rleungx:
Some compatibility problems also need to be considered.

JmPotato:
Another problem with the same root case: https://github.com/tikv/pd/issues/4109

JmPotato:
Another problem with the same root case: https://github.com/tikv/pd/issues/4109

PR #4595, opened 2022-01-17
*: unify the root key path joining

JmPotato:

Close https://github.com/tikv/pd/issues/4284.

Use `server/storage/endpoint/key_path.go` to manage the key path joining.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,6 @@ static: install-tools
 	@revive -formatter friendly -config revive.toml $(PACKAGES)
 
 tidy:
-	@echo "go mod tidy"
 	go mod tidy
 	git diff --quiet go.mod go.sum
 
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -647,6 +650,10 @@ func (s *Server) bootstrapCluster(req *pdpb.BootstrapRequest) (*pdpb.BootstrapRe
 	}, nil
 }
 
+func (s *Server) appendToRootPath(key string) string {
```
rleungx, line 653: How about moving it to `key_path.go`?

Makefile
```diff
@@ -156,7 +156,6 @@ static: install-tools
 	@revive -formatter friendly -config revive.toml $(PACKAGES)
 
 tidy:
-	@echo "go mod tidy"
```
rleungx, line 159: Why remove it?

JmPotato: The cmd `go mod tidy` itself will be outputted if there is no `@` at the beginning. So the `echo` is unncessary.

rleungx (approved):

lhy1024 (approved):

HunDunDM (approved):

Issue #4515, opened 2021-12-29
Reduce the dependency of `RaftCluster` for `coordinator`


rleungx:
## Enhancement Task
The mutex between `RaftCluster` and `coordinator` is complicated which is easy to cause a deadlock when writing codes. We need to make it more clear and reduce the dependency of `RaftCluster` for `coordinator`.

### Subtask
- [x] https://github.com/tikv/pd/pull/4516
- [x] https://github.com/tikv/pd/pull/4544
- [x] https://github.com/tikv/pd/pull/4594
- [x] https://github.com/tikv/pd/pull/4597

PR #4597, opened 2022-01-18
*: optimize lock usage

rleungx:

Issue Number: Ref #4515.

This PR removes some necessary lock.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -501,8 +501,6 @@ func (c *RaftCluster) GetAllocator() id.Allocator {
 
 // GetRegionSyncer returns the region syncer.
 func (c *RaftCluster) GetRegionSyncer() *syncer.RegionSyncer {
-	c.RLock()
-	defer c.RUnlock()
 	return c.regionSyncer
 }
 
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -31,7 +31,6 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
-	"github.com/tikv/pd/pkg/component"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/etcdutil"
 	"github.com/tikv/pd/pkg/logutil"
@@ -487,11 +486,6 @@ func (c *RaftCluster) PauseOrResumeChecker(name string, t int64) error {
 // IsCheckerPaused returns if checker is paused
 func (c *RaftCluster) IsCheckerPaused(name string) (bool, error) {
 	return c.coordinator.isCheckerPaused(name)
-}
-
-// GetComponentManager returns component manager.
-func (c *RaftCluster) GetComponentManager() *component.Manager {
-	return c.componentManager
 }
 
 // GetAllocator returns cluster's id allocator.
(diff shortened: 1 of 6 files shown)


HunDunDM (commented):
I'm worried about some nil pointers after breaking consistency.

server/cluster/cluster.go
```diff
@@ -217,10 +212,7 @@ func (c *RaftCluster) InitCluster(
 	c.ctx, c.cancel = context.WithCancel(c.serverCtx)
```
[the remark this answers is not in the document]
rleungx: I checked the RPC calls and only `SplitRegions` that needs to be modified.

server/cluster/cluster.go
```diff
@@ -1136,9 +1155,6 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
 
 // SetStoreWeight sets up a store's leader/region balance weight.
 func (c *RaftCluster) SetStoreWeight(storeID uint64, leaderWeight, regionWeight float64) error {
-	c.Lock()
```
bufferflies, line 1139: Is it safe if update store weight by multi thread?

disksing (approved):

server/cluster/cluster.go
```diff
@@ -418,62 +413,106 @@ func (c *RaftCluster) Context() context.Context {
...
 }
 
-// GetHeartbeatStreams returns the heartbeat streams.
-func (c *RaftCluster) GetHeartbeatStreams() *hbstream.HeartbeatStreams {
-	c.RLock()
-	defer c.RUnlock()
-	return c.coordinator.hbStreams
+// GetMergeChecker returns merge checker.
+func (c *RaftCluster) GetMergeChecker() *checker.MergeChecker {
+	return c.coordinator.checkers.GetMergeChecker()
```
AndreMouche, line 438: Removing the `Lock` LGTM, however, should we consider the situation that `c.coordinator` is nil(IsRunning==false)?  
Here contains two issues:
1. panic with nil: if the c.coordinator is nil
2. data race: if there exists a goroutine that is starting the server(updating the c.coordinator) at the same time

If we are the only programmer who knows how to use the lib, we should run `c.Start` successfully at first. However the later programmer may not know this condition, so to be a better `lib`, Could you please check `c.IsRunning` first(which can both avoid the nil issue and keep thread safe)?

AndreMouche (commented):
Rest LGTM

server/cluster/cluster.go
```diff
@@ -1430,31 +1441,23 @@ func (c *RaftCluster) resetMetrics() {
 }
 
 func (c *RaftCluster) collectClusterMetrics() {
-	c.RLock()
 	if c.regionStats == nil {
```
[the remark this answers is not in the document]
rleungx: This is guaranteed by our internal logic.

server/statistics/region_collection.go
```diff
@@ -274,6 +287,8 @@ func NewLabelStatistics() *LabelStatistics {
 
 // Observe records the current label status.
 func (l *LabelStatistics) Observe(region *core.RegionInfo, stores []*core.StoreInfo, labels []string) {
+	l.Lock()
+	defer l.Unlock()
```
AndreMouche, line 291: The lock could start after L293

server/statistics/region_collection.go
```diff
@@ -308,6 +327,8 @@ func (l *LabelStatistics) ClearDefunctRegion(regionID uint64) {
 
 // GetLabelCounter is only used for tests.
 func (l *LabelStatistics) GetLabelCounter() map[string]int {
+	l.RLock()
```
AndreMouche, line 330: Although it is only used in tests,  there are two issues:
- First, the lock is unnecessary
- secondary, It is not threadsafe to return the pointer(l.labelCounter)

AndreMouche (approved):
LGTM

PR #4615, opened 2022-01-26
api: introduce API v2

### LINKED ISSUE #4640: Introduce HTTP API v2
## Development Task
The current HTTP API doesn't follow the RESTful API conventions. We have summarised some problems of the current status:
- mixed-use of `_` and `-`
- some methods are not accurate, e.g., we do not distinguish `PUT` and `POST`
- the resource should use a noun
- query parameter should not occur in the path
- mixed-use of singular and plural nouns

In addition, we can use `gin` as our HTTP web framework which is more popular and has a better ecosystem.

rleungx:

Close #4640.

This PR tries to support the v2 HTTP API which is RESTful and more standard. The reason why it is needed can be found in the above issue. Fow now, this PR only supports two kinds of APIs with the common prefix `/pd/api/v2`:
- `/meta/stores`
- `/members`

**Notice: They are very likely to be changed in the next few days.** 

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Introduce the v2 HTTP API
```

### DIFF
--- a/server/apiv2/handlers/members.go
+++ b/server/apiv2/handlers/members.go
@@ -1,0 +1,107 @@
+package handlers
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/pingcap/kvproto/pkg/pdpb"
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server"
+	"go.uber.org/zap"
+)
+
+// GetMembers returns the PD members.
+func GetMembers() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		svr := c.MustGet("server").(*server.Server)
+		members, err := svr.GetMembers()
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{
+				"error": err.Error(),
+			})
+			return
+		}
+		dclocationDistribution, err := svr.GetTSOAllocatorManager().GetClusterDCLocationsFromEtcd()
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{
+				"error": err.Error(),
+			})
+			return
+		}
+		for _, m := range members {
+			m.DcLocation = ""
+			binaryVersion, e := svr.GetMember().GetMemberBinaryVersion(m.GetMemberId())
+			if e != nil {
+				log.Error("failed to load binary version", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+			}
+			m.BinaryVersion = binaryVersion
+			deployPath, e := svr.GetMember().GetMemberDeployPath(m.GetMemberId())
+			if e != nil {
+				log.Error("failed to load deploy path", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+			}
+			m.DeployPath = deployPath
+			if svr.GetMember().GetEtcdLeader() == 0 {
+				log.Warn("no etcd leader, skip get leader priority", zap.Uint64("member", m.GetMemberId()))
+				continue
+			}
+			leaderPriority, e := svr.GetMember().GetMemberLeaderPriority(m.GetMemberId())
+			if e != nil {
+				log.Error("failed to load leader priority", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+				continue
+			}
+			m.LeaderPriority = int32(leaderPriority)
+			gitHash, e := svr.GetMember().GetMemberGitHash(m.GetMemberId())
+			if e != nil {
+				log.Error("failed to load git hash", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+				continue
+			}
+			m.GitHash = gitHash
+			found := false
+			for dcLocation, serverIDs := range dclocationDistribution {
+				for _, serverID := range serverIDs {
+					if serverID == m.MemberId {
+						m.DcLocation = dcLocation
+						found = true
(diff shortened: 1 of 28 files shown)


nolouch:
PTAL @CabinfeverB

server/apiv2/router.go
```diff
@@ -0,0 +1,75 @@
...
+	router.Use(func(c *gin.Context) {
+		c.Set("server", svr)
+		c.Next()
+	})
+	router.Use(middlewares.Redirector())
+	root := router.Group(apiV2Prefix)
+	meta := root.Group("meta")
+	meta.Use(middlewares.BootstrapChecker())
+	meta.GET("/stores", handlers.GetStores())
+	meta.GET("/stores/:id", handlers.GetStoreByID())
```
CabinfeverB, line 60: ref #4640, it is a problem that query parameters should not occur in the path.  Do we need change it?

server/apiv2/router.go
```diff
@@ -0,0 +1,75 @@
...
+	meta.DELETE("/stores/:id", handlers.DeleteStoreByID())
```
CabinfeverB, line 61: how about “/store/:id"? This delete method has only one param

server/apiv2/handlers/types.go
```diff
@@ -0,0 +1,237 @@
...
+)
+
+const (
+	aliveStateName        = "Alive"
+	disconnectedStateName = "Disconnected"
+	downStateName         = "Down"
+)
+
+// StoresInfo records stores' info.
+type StoresInfo struct {
```
CabinfeverB, line 35: It seems that there is no difference between apiv1 and apiv2. How about only one declaration?

rleungx: I'm not quite sure if it will be changed later, so just copy it.

server/apiv2/router.go
rleungx: It's actually not a query parameter I think.

server/apiv2/handlers/members.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return nil, err
+	}
+	dclocationDistribution, err := svr.GetTSOAllocatorManager().GetClusterDCLocationsFromEtcd()
+	if err != nil {
+		return nil, err
+	}
+	for _, m := range members.GetMembers() {
+		m.DcLocation = ""
+		binaryVersion, e := svr.GetMember().GetMemberBinaryVersion(m.GetMemberId())
```
AndreMouche, line 174: how about use `m.BinaryVersion` directly? So we needn't worry about missing L178?
```
          
  var e error
  m.BinaryVersion, e = svr.GetMember().GetMemberBinaryVersion(m.GetMemberId())
```

server/apiv2/handlers/members.go
```diff
@@ -0,0 +1,215 @@
...
+		if e != nil {
+			log.Error("failed to load binary version", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+		}
+		m.BinaryVersion = binaryVersion
+		deployPath, e := svr.GetMember().GetMemberDeployPath(m.GetMemberId())
```
AndreMouche, line 179: m.DeployPath, e = svr.GetMember().GetMemberDeployPath(m.GetMemberId())

server/apiv2/handlers/members.go
```diff
@@ -0,0 +1,215 @@
...
+			log.Error("failed to load leader priority", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+			continue
+		}
+		m.LeaderPriority = int32(leaderPriority)
+		gitHash, e := svr.GetMember().GetMemberGitHash(m.GetMemberId())
+		if e != nil {
+			log.Error("failed to load git hash", zap.Uint64("member", m.GetMemberId()), errs.ZapError(e))
+			continue
+		}
+		m.GitHash = gitHash
```
AndreMouche, line 199: dit too

server/apiv2/handlers/members.go
```diff
@@ -0,0 +1,215 @@
...
+		found := false
+		for dcLocation, serverIDs := range dclocationDistribution {
+			for _, serverID := range serverIDs {
+				if serverID == m.MemberId {
+					m.DcLocation = dcLocation
+					found = true
```
AndreMouche, line 205: Too many `{}` in `{}`. how about define an independent function and make it friendly for our reviewer? For example replace L200-L213 as following:
```
          for dcLocation, serverIDs := range dclocationDistribution {
			found := contains(serverIDs, m.MemberId)
			if found {
				m.DcLocation = dcLocation
				break
			}
		}
```

server/apiv2/handlers/members.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/etcdutil"
+	"github.com/tikv/pd/server"
+	"go.uber.org/zap"
+)
+
+// GetMembers returns the PD members.
+func GetMembers() gin.HandlerFunc {
+	return func(c *gin.Context) {
```
AndreMouche, line 32: Could we define `GetMembers` as a gin.HandlerFunc directly such as the following?
```
func GetMembers(c *gin.Context) {
	svr := c.MustGet("server").(*server.Server)
	members, err := getAllMembers(svr)
	if err != nil {
		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
		return
	}
	c.IndentedJSON(http.StatusOK, members)
}
```
This looks better for me aha 😄

PR #4619, opened 2022-01-28
api: fix invalid interface conversion

### LINKED ISSUE #4617: Some rest api handle query body incorrectly
## Bug Report
Some rest api use map[string]interface{} to parse request body and cast them to int or []uint. Howerer go json does parse these field as float or []interface{}, which lead to invalid interface conversion. For example.
https://github.com/tikv/pd/blob/c1d1ecdd9e19325023a9b91d176c08f3b44c534a/server/api/region.go#L798
https://github.com/tikv/pd/blob/c1d1ecdd9e19325023a9b91d176c08f3b44c534a/server/api/region.go#L832
### What did you do?
curl --request POST \
  --url http://127.0.0.1:2379/pd/api/v1/regions/scatter \
  --data '{"regions_id": [10]}'

The server response OK.

### What did you see instead?
The server response Internal Server Error
``
PANIC: interface conversion: interface {} is []interface {}, not []uint64
``

### What version of PD are you using (pd-server -V)?
Release Version: v5.5.0-alpha-43-gc1d1ecdd
Edition: Community
Git Commit Hash: c1d1ecdd9e19325023a9b91d176c08f3b44c534a
Git Branch: master
UTC Build Time:  2022-01-28 03:19:49

rleungx:

Fix #4617.

Correctly convert `[]interface{}` to `[]uint64` after JSON `Unmarshal`.

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the Region ID list cannot be converted to `uint64`
```

### DIFF
--- a/pkg/typeutil/conversion_test.go
+++ b/pkg/typeutil/conversion_test.go
@@ -15,6 +15,9 @@
 package typeutil
 
 import (
+	"encoding/json"
+	"reflect"
+
 	. "github.com/pingcap/check"
 )
 
@@ -34,4 +37,24 @@ func (s *testUint64BytesSuite) TestUint64ToBytes(c *C) {
 	b := Uint64ToBytes(a)
 	str := "\x00\x00\x00\x00\x00\x00\x03\xe8"
 	c.Assert(b, DeepEquals, []byte(str))
+}
+
+var _ = Suite(&testJSONSuite{})
+
+type testJSONSuite struct{}
+
+func (s *testJSONSuite) TestJSONToUint64Slice(c *C) {
+	type testArray struct {
+		Array []uint64 `json:"array"`
+	}
+	a := testArray{
+		Array: []uint64{1, 2, 3},
+	}
+	bytes, _ := json.Marshal(a)
+	var t map[string]interface{}
+	err := json.Unmarshal(bytes, &t)
+	c.Assert(err, IsNil)
+	res, ok := JSONToUint64Slice(t["array"])
+	c.Assert(ok, IsTrue)
+	c.Assert(reflect.TypeOf(res[0]).Kind(), Equals, reflect.Uint64)
 }
--- a/pkg/typeutil/conversion_test.go
+++ b/pkg/typeutil/conversion_test.go
@@ -54,7 +54,26 @@ func (s *testJSONSuite) TestJSONToUint64Slice(c *C) {
 	var t map[string]interface{}
 	err := json.Unmarshal(bytes, &t)
 	c.Assert(err, IsNil)
+	// valid case
 	res, ok := JSONToUint64Slice(t["array"])
 	c.Assert(ok, IsTrue)
 	c.Assert(reflect.TypeOf(res[0]).Kind(), Equals, reflect.Uint64)
+	// invalid case
+	_, ok = t["array"].([]uint64)
+	c.Assert(ok, IsFalse)
+
+	// invalid type
+	type testArray1 struct {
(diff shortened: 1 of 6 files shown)


lightmelodies:
Maybe other case need fix too.
https://github.com/tikv/pd/blob/c1d1ecdd9e19325023a9b91d176c08f3b44c534a/server/api/region.go#L808
https://github.com/tikv/pd/blob/c1d1ecdd9e19325023a9b91d176c08f3b44c534a/server/api/region.go#L887
https://github.com/tikv/pd/blob/c1d1ecdd9e19325023a9b91d176c08f3b44c534a/server/api/operator.go#L291

server/api/operator.go
```diff
@@ -286,14 +286,19 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
 		// support both receiving key ranges or regionIDs
 		startKey, _ := input["start_key"].(string)
 		endKey, _ := input["end_key"].(string)
-		regionIDs, _ := input["region_ids"].([]uint64)
-		group, _ := input["group"].(string)
-		retryLimit, ok := input["retry_limit"].(int)
+		regionIDs, _ := input["region_ids"].([]interface{})
```
disksing, line 289: should it better be `interface{}` ? because `toUint64Slice` accepts `interface{}`.

server/api/operator.go
```diff
@@ -286,14 +286,19 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
...
+		ids, ok := toUint64Slice(regionIDs)
```
disksing, line 290: better add some test to cover it.

pkg/typeutil/conversion_test.go
```diff
@@ -35,3 +38,23 @@ func (s *testUint64BytesSuite) TestUint64ToBytes(c *C) {
...
+var _ = Suite(&testJSONSuite{})
+
+type testJSONSuite struct{}
+
+func (s *testJSONSuite) TestJSONToUint64Slice(c *C) {
+	type testArray struct {
+		Array []uint64 `json:"array"`
+	}
+	a := testArray{
+		Array: []uint64{1, 2, 3},
```
JmPotato, line 51: What about adding an invalid test case?

server/api/region_test.go
```diff
@@ -344,7 +348,7 @@ func (s *testRegionSuite) TestSplitRegions(c *C) {
 	mustRegionHeartbeat(c, s.svr, r1)
 	mustPutStore(c, s.svr, 13, metapb.StoreState_Up, []*metapb.StoreLabel{})
 	newRegionID := uint64(11)
-	body := fmt.Sprintf(`{"retry_limit":%v, "split_keys": ["%s","%s","%s"]}`, 0,
+	body := fmt.Sprintf(`{"retry_limit":%v, "split_keys": ["%s","%s","%s"]}`, 3,
```
JmPotato, line 351: Do we have a test case to test the default value of the `retry_limit`?

rleungx: No, I test it manually.

JmPotato (approved):

pkg/typeutil/conversion_test.go
JmPotato: @rleungx I mean a case that the `JSONToUint64Slice` function which will return nil and false.

AndreMouche (approved):
LGTM

disksing (approved):

PR #4620, opened 2022-01-28
Support to pause scheduling on specific regions

lightmelodies:

Currently Pd can only pause scheduler for whole cluster, which does not work well for [online bulk load](https://github.com/tikv/rfcs/blob/master/text/0072-online-bulk-load-for-rawkv.md). This PR add support to pause scheduling on specific regions.

1. Add a TTLCache in RaftCluster struct to store pinned regions.
2. Modify most schedulers and checkers to skip pinned regions when calc operators.
3. Add rest api to pin regions by key range or regions id with given duration, as well as list currently pinned regions.
4. The pin api should be used immediately after split and scatter in most case.

### Check List

Tests
- Unit test

### Release note
```release-note
Support to pause scheduling on specific regions
```

### DIFF
--- a/pkg/mock/mockcluster/mockcluster.go
+++ b/pkg/mock/mockcluster/mockcluster.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/cache"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/mock/mockid"
 	"github.com/tikv/pd/server/config"
@@ -53,6 +54,7 @@ type Cluster struct {
 	*config.PersistOptions
 	ID             uint64
 	suspectRegions map[uint64]struct{}
+	pinnedRegions  *cache.TTLUint64
 }
 
 // NewCluster creates a new Cluster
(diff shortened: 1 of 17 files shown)


nolouch:
I prefer to do this with region labels, cc @disksing @rleungx

lightmelodies:
> I prefer to do this with region labels, cc @disksing @rleungx

There are some problems with current region labels implemention.

1. No TTL support, so the pinning status cannot be automatically recovered if client crash.
2. Only Key range are support now, which may be slower than lookup of regions id.

Anyway, This PR can be changed to use region labels strategy by modify [IsRegionPinned](https://github.com/lightmelodies/pd/blob/e63348dc7eb2d9100d0faa39f73ab8a40ec6d555/server/cluster/cluster.go#L513) 's implemention like [this](https://github.com/lightmelodies/pd/blob/e63348dc7eb2d9100d0faa39f73ab8a40ec6d555/server/schedule/checker/merge_checker.go#L247), which is not a huge work.

disksing:
region label +1

lightmelodies:
For region label way, check this [commit](https://github.com/lightmelodies/pd/commit/09035939774ab99b759687f28ebeab869a023d13) @nolouch @disksing

nolouch:
Hi, @lightmelodies.  Thanks for your contribution.  Sorry for that I ignored the changed commit with the region label. I saw you reviewed the pr https://github.com/tikv/pd/pull/4649, which already meets a part of your requirement. and why you need a region ID rather than the `key range`?

lightmelodies:
> Hi, @lightmelodies. Thanks for your contribution. Sorry for that I ignored the changed commit with the region label. I saw you reviewed the pr #4649, which already meets a part of your requirement. and why you need a region ID rather than the `key range`?

Not really need `region ID`. Though it's easy to implement with TTL.  #4649 (with #4695) should be enough to meet my requirement after some prototype test. So I'll close this PR.

nolouch:
Thank your contribution.

Issue #4623, opened 2022-02-07
id: reduce WithLabelValues of prometheus


nolouch:
## Development Task
improve the metrics in the id allocator.
![image](https://user-images.githubusercontent.com/6428910/152715989-ed1d84d0-25d4-4d95-9a28-55fef7fb535c.png)

PR #4624, opened 2022-02-07
server/id: pre-alloc id gauge metrics

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/4623

```commit-message
server/id: pre-alloc id gauge metrics
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -120,7 +120,7 @@ func (alloc *allocatorImpl) rebaseLocked() error {
 	}
 
 	log.Info("idAllocator allocates a new id", zap.Uint64("alloc-id", end))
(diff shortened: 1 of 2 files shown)


disksing:
I think alloc id should be a low-frequency enough operation that it doesn't seem worth doing such an optimization.

lhy1024 (approved):

JmPotato (approved):
LGTM. But should this close the issue https://github.com/tikv/pd/issues/4623? The TSO part still needs to be addressed.

nolouch:
@disksing Yes, Here is. I mainly want to remove the usage of `withLabelValues` in the key service packages. otherwise, it's may guide developers to use the same way.

PR #4308, opened 2021-11-08
implement `GlobalConfig` for grpc server and client

LemonHX:
Signed-off-by: lemonhx <lemonhx@lemonhx.tech>

### What problem does this PR solve?
previously we have discussions about adding global config into PD and store in etcd for easier to maintain the config and for the ease of usage of  all components such as tikv tidb and tiflash

### What is changed and how it works?
adding three functions into kvproto for loading ,storing and watching the global config, and implement it for etcd
### Check List
- [x] does load information from etcd
- [x] does write information into etcd
- [x] does listen the prefix `/global/config` correctly

Tests

-  [x] Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Side effects

-  [x] Possible performance regression 
-  [x] Increased code complexity 
-  [x] Breaking backward compatibility

Related changes

- kvproto https://github.com/pingcap/kvproto/pull/828

### Release note

```release-note
None
```

PR #4635, opened 2022-02-11
config: add `region-max-size ` property

### LINKED ISSUE #4459: sync tikv config
## Enhancement Task
Config supports `region-max-size` property. This value should only be set by file not by api.

bufferflies:

Issue Number: Close #4459

`config.PersistOptions` should save the initial config and not update it .

```commit-message
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
config adds`region-max-size ` property 
```


server/config/persist_options.go
```diff
@@ -50,11 +50,12 @@ type PersistOptions struct {
 	replicationMode atomic.Value
 	labelProperty   atomic.Value
 	clusterVersion  unsafe.Pointer
+	cfg             *Config
```
rleungx, line 53: How about adding an interface function into Cluster?

rleungx:
Please add a unit test to check if changing the parameter in the config file works.

server/config/persist_options.go
bufferflies: The most values of the  persistOptions is transferred by config and I think the cluster should export one method to manage all config.

rleungx: Those configs in persist option can be changed dynamically, I prefer not to put `cfg` inside it.

nolouch:
How this config influence the TiKV?

conf/config.toml
```diff
@@ -23,7 +23,7 @@
 
 ## Join to an existing cluster. The value should be cluster's ${advertise-client-urls}
 # join = ""
-
+# max-region-size = 96
```
disksing, line 26: insert an empty line?

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -51,18 +51,20 @@ type Cluster struct {
...
-		BasicCluster:   core.NewBasicCluster(),
-		IDAllocator:    mockid.NewIDAllocator(),
-		HotStat:        statistics.NewHotStat(ctx),
-		PersistOptions: opts,
-		suspectRegions: map[uint64]struct{}{},
+		BasicCluster:    core.NewBasicCluster(),
+		IDAllocator:     mockid.NewIDAllocator(),
+		HotStat:         statistics.NewHotStat(ctx),
+		PersistOptions:  opts,
+		ImmutableConfig: config.NewTestImmutableOptions(),
```
disksing, line 66: Does `immutable` means that it cannot be updated by pd-ctl?

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -78,6 +80,11 @@ func (mc *Cluster) GetOpts() *config.PersistOptions {
 	return mc.PersistOptions
 }
 
+// GetConfig returns the cluster immutable configuration.
+func (mc *Cluster) GetConfig() *config.ImmutableConfig {
```
rleungx, line 84: Prefer to change the name.

server/config/immutable_options.go
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+// ImmutableConfig is a readonly config.
+type ImmutableConfig struct {
+	maxRegionSize uint64
+	leaderLease   int64
```
rleungx, line 20: Why only leaderLease?

server/config/immutable_options.go
```diff
@@ -0,0 +1,48 @@
...
+// GetMaxRegionSize returns the max region size.
+func (c *ImmutableConfig) GetMaxRegionSize() uint64 {
+	return c.maxRegionSize
+}
+
+// ImmutableConfigCreateOption used to create ImmutableConfig.
+type ImmutableConfigCreateOption func(config *ImmutableConfig)
+
+// WithMaxRegionSize set max region size.
+func WithMaxRegionSize(maxRegionSize uint64) ImmutableConfigCreateOption {
```
rleungx, line 44: As it is an immutable config, perhaps we don't need this function.

server/schedule/operator/builder.go
```diff
@@ -33,6 +33,7 @@ import (
 type ClusterInformer interface {
 	GetBasicCluster() *core.BasicCluster
 	GetOpts() *config.PersistOptions
+	GetConfig() *config.ImmutableConfig
```
rleungx, line 36: How about moving it to `Cluster`?

bufferflies:
> How this config influence the TiKV?

the configuration will not influence the TiKVs, maybe can consider this pr:#4308

pkg/mock/mockcluster/mockcluster.go
bufferflies: yes, pd will reload config after restarting.

conf/config.toml
bufferflies: fix

server/schedule/operator/builder.go
bufferflies: this immutable config can be used by some checks or schedulers.

bufferflies: fix

pkg/mock/mockcluster/mockcluster.go
bufferflies: fix to GetImmutableCfg

server/config/immutable_options.go
bufferflies: `leaderLease` will be  remove and add `max-split-size`.

bufferflies:
It is not graceful to add some tikv config into the config file, so I will use call back tikv to get this config.

PR #4642, opened 2022-02-14
server/api/operator.go: tiny refactor

AndreMouche:
This PR tiny refactor api/operator.go

 This PR tiny refactored the file operator.go, it used `const` for  the keys for the arguments, so we can share the key between different scope.  Also it define serval small functions to simplify the main logic.

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/api/operator.go
```diff
@@ -79,7 +92,7 @@ func (h *operatorHandler) List(w http.ResponseWriter, r *http.Request) {
 		err     error
 	)
 
-	kinds, ok := r.URL.Query()["kind"]
+	kinds, ok := r.URL.Query()[pKind]
```
nolouch, line 95: Will this have compatibility issues?

pkg/typeutil/conversion.go
```diff
@@ -68,3 +68,20 @@ func JSONToUint64Slice(from interface{}) ([]uint64, bool) {
 	}
 	return to, true
 }
+
+// JSONToStringSlice converts JSON slice to string slice.
+func JSONToStringSlice(from interface{}) ([]string, bool) {
```
JmPotato, line 73: Can we use `reflect` to merge this function with `JSONToUint64Slice`?

server/api/operator.go
AndreMouche: First of all, this PR we only adjust(change) the response body. 
And I'm not sure if it is ok haha. 
However, the error response for invalid or missing parameters before seems defined casually,  examples :
- (region_id)`region_ids is invalid`
- (region_id)`missing region id`
- (to_store_id) `"missing store id to transfer leader to"`
- (from_store_id) `"invalid store id to transfer peer from"`

in this PR there were changed to 
- (region_id) `missing region_id`
- (to_store_id) `missing to_store_id`
- (from_store_id) `invalid from_store_id`

Thus the response is totally different, I'm not sure if it is ok, any suggestion?

pkg/typeutil/conversion.go
AndreMouche: Maybe not? It seems we only need  to support string and uint64, while using `reflect` costs more and also harder to maintain?

server/api/operator.go
```diff
@@ -125,177 +169,168 @@ func (h *operatorHandler) Post(w http.ResponseWriter, r *http.Request) {
...
 		}
-		if err := h.AddTransferLeaderOperator(uint64(regionID), uint64(storeID)); err != nil {
+		if err := h.AddTransferLeaderOperator(regionID, storeID); err != nil {
 			h.r.JSON(w, http.StatusInternalServerError, err.Error())
 			return
 		}
-	case "transfer-region":
-		regionID, ok := input["region_id"].(float64)
+	case opTransferRegion:
+		regionID, ok := parseUInt64Param(pRegionID)
```
rleungx, line 203: IMO, the original way is simple enough.

AndreMouche: here we use constant `pRegionID` instead of "region_id" so we needn't check whether the spelling of `region_id` is correct everywhere on those it used as the param for region's ID and we wish they should keep the same.

rleungx: I mean `input[pRegionID].(float64)`

AndreMouche: yes, the original way is very simple , the function `parseUIntParam` is used to parse and check the `UInt` param from `input` which only contains 3~4 lines and seems simple enough. However, the same logic of "parse and check the `Uint` param from `input`" appears serval times, so why we need the function `parseUInt64Param`:

- They allow us to conceive of our program as a bunch of sub-steps. (Each sub-step can be its own function. When any program seems too hard, just break the overall program into sub-steps!)

- They allow us to reuse code instead of rewriting it.

PR #4649, opened 2022-02-17
server/schedule/labeler: support schedule-disabled for region

AndreMouche:

This PR is going to support disabling schedule for the region.

### What is changed and how it work?

From TiDB（SQL）
 image

with the above labels, the regions for the table would stop schedule except operators generated by `join state checker`. However, we can still create operators manually by `pd-ctl`(admin)
It works with region-label, for example:
```
{
 ID: "rule1", 
 Labels: []RegionLabel{
        {Key: "schedule", Value: "disabled"}
    },
 RuleType: "key-range",
 Data: makeKeyRanges("1234", "5678")
}
```

### Check List

Tests

- [x] - Unit test
- [x] - Manual test (add detailed scripts or steps below)

### Release note

```release-note
region-labels: support disable schedule by region.
```


server/schedule/checker/checker_controller.go
```diff
@@ -76,6 +76,13 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
 	// Don't check isRaftLearnerEnabled cause it maybe disable learner feature but there are still some learners to promote.
 	opController := c.opController
 
+	if cl, ok := c.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
```
lightmelodies, line 79: Should we ignore **schedule deny label** for join state and replica checker? Dealing with unhealthy region does not make sense in most case.

server/schedule/operator_controller.go
```diff
@@ -409,6 +410,14 @@ func (oc *OperatorController) checkAddOperator(ops ...*operator.Operator) bool {
 			operatorWaitCounter.WithLabelValues(op.Desc(), "exceed-max").Inc()
 			return false
 		}
+		if cl, ok := oc.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
```
lightmelodies, line 413: Maybe better to allow operator generated by manual split and scatter request to bypass this check?

server/schedule/labeler/labeler.go
```diff
@@ -287,6 +288,12 @@ func (l *RegionLabeler) GetRegionLabel(region *core.RegionInfo, key string) stri
 	return value
 }
 
+// ScheduleDisabled returns true if the region is lablelld with schedule-disabled.
+func (l *RegionLabeler) ScheduleDisabled(region *core.RegionInfo) bool {
```
rleungx, line 292: We should skip the admin operators.

nolouch (commented):
will schedulers still work on the range? such as `balance-region-scheduler`

server/schedule/checker/checker_controller.go
AndreMouche: Yes,  `ignore from join state` is done.  But we do not ignore the label for `replica check` because `the replica operators generated by manual` can still work when `schedule=deny`.

server/schedule/operator_controller.go
AndreMouche: done

server/schedule/labeler/labeler.go
AndreMouche: done

AndreMouche:
> will schedulers still work on the range? such as `balance-region-scheduler`

yes

server/schedule/operator_controller.go
```diff
@@ -369,7 +369,7 @@ func (oc *OperatorController) PromoteWaitingOperator() {
 // - The region already has a higher priority or same priority operator.
 // - Exceed the max number of waiting operators
 // - At least one operator is expired.
-func (oc *OperatorController) checkAddOperator(ops ...*operator.Operator) bool {
+func (oc *OperatorController) checkAddOperator(fromAddmin bool, ops ...*operator.Operator) bool {
```
rleungx, line 376:
```suggestion
func (oc *OperatorController) checkAddOperator(fromAdmin bool, ops ...*operator.Operator) bool {
```

server/schedule/operator_controller.go
```diff
@@ -410,6 +410,9 @@ func (oc *OperatorController) checkAddOperator(ops ...*operator.Operator) bool {
 			operatorWaitCounter.WithLabelValues(op.Desc(), "exceed-max").Inc()
 			return false
 		}
+		if fromAddmin || op.IsLeaveJointStateOperator() {
```
rleungx, line 417: prefer to use op.Kind()

AndreMouche: It seems the `LeaveJointStateOperator` share the same `kind ` with other operators.

rleungx: Oh, I mean `SchedulerKind`

AndreMouche: can you provide a further heavy hint? which `ScheduleKind` should it be when it is create by `leave joint state`? 😂

rleungx: There is no need to use `fromAddmin`, maybe you can use `op.SchedulerKind() == OpAdmin`

disksing (approved):

server/schedule/operator_controller.go
```diff
@@ -413,6 +414,18 @@ func (oc *OperatorController) checkAddOperator(ops ...*operator.Operator) bool {
...
 		}
+
+		if op.Kind().ContainsOpAdmin() || op.IsLeaveJointStateOperator() {
+			continue
+		}
+		if cl, ok := oc.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
+			l := cl.GetRegionLabeler()
+			if l.ScheduleDisabled(region) {
+				log.Debug("schedule disabled", zap.Uint64("region-id", op.RegionID()))
+				operatorWaitCounter.WithLabelValues(op.Desc(), "schedule-disabled").Inc()
```
nolouch, line 425: I think we missing these metrics in Grafana. need to update the template in another PR.

nolouch (commented):
LGTM

nolouch (approved):

server/schedule/operator/create_operator.go
```diff
@@ -217,6 +217,13 @@ func CreateScatterRegionOperator(desc string, ci ClusterInformer, origin *core.R
 		Build(0)
 }
 
+const opDescLeaveJointState = "leave-joint-state"
+
+// NewLeaveJointStateOperator  creates an operator that let region leave joint state with default desc.
+func NewLeaveJointStateOperator(ci ClusterInformer, origin *core.RegionInfo) (*Operator, error) {
```
rleungx, line 223: What is the difference between `NewLeaveJointStateOperator` and `CreateLeaveJointStateOperator`?

AndreMouche: Similar to the `CreateXXXOperator`, the user can define the desc in `CreateLeaveJointStateOperator`, such as in the test. While `NewLeaveJointStateOperator ` use a constant as the desc so we can easily find the operator that was created by `leave joint state`

server/schedule/operator_controller.go
AndreMouche: OK

rleungx (approved):

PR #4652, opened 2022-02-17
scheduler:  allow `balance-leader-scheduler` generate multiple operators

CabinfeverB:
Signed-off-by: Cabinfever_B <cabinfeveroier@gmail.com>

### What problem does this PR solve?

Issue Number: ref #4610
speed up balance leader

### What is changed and how it works?

Add a Batch param like #4008 that the scheduler can generate multi operators.

```commit-message
speed up balance leader by batch
```

### Check List

Tests

- Unit test

### Release note

```release-note
Speed up balance leader
```

Issue #4610, opened 2022-01-25
Speed up balance leader


nolouch:
## Development Task
Currently, balance leader only have max ops 100 op/s. we want to increase the speed when a big cluster restart (2M regions).

one way like:  https://github.com/tikv/pd/pull/4008

nolouch:
cc @CabinfeverB, Would you like to take a look? ptal @rleungx

CabinfeverB:
I will take a look

CabinfeverB:
### Describe alternatives you've considered
Add a `Batch` param like #4008 that the scheduler can generate multi operators.

Subtasks
- [ ] #4652
- [ ] add `balance-leader` config handler and improve config
- [ ] support config `balance-leader` in pd-ctl

CabinfeverB:
cc @mayjiang0203

PR #4655, opened 2022-02-18
api, scheduler: add `balance-leader` config handler

CabinfeverB:

Issue Number: ref https://github.com/tikv/pd/issues/4610
This PR should be merged after #4652

```commit-message
Add `balance-leader-leader` config API.
```

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

### Release note

```release-note
Add `balance-leader-leader` config API.
```


nolouch:
there are some conflicts

server/schedulers/balance_leader.go
```diff
@@ -192,6 +296,8 @@ func (cs *candidateStores) reSort(stores ...*core.StoreInfo) {
 }
 
 func (l *balanceLeaderScheduler) Schedule(cluster schedule.Cluster) []*operator.Operator {
+	l.conf.mu.RLock()
```
rleungx, line 299: How about locking it only when it's going to be read?

server/schedulers/balance_leader.go
```diff
@@ -71,15 +80,106 @@ func init() {
...
+	return &balanceLeaderSchedulerConfig{
+		Ranges: conf.Ranges,
+		Batch:  conf.Batch,
+	}
+}
+
+func (conf *balanceLeaderSchedulerConfig) Persist() error {
+	conf.mu.RLock()
+	defer conf.mu.RUnlock()
+	data, err := schedule.EncodeConfig(conf)
```
bufferflies, line 135:
```suggestion
return conf.persistLocked()
```

server/schedulers/balance_leader.go
```diff
@@ -71,15 +80,106 @@ func init() {
 }
 
 type balanceLeaderSchedulerConfig struct {
-	Name   string          `json:"name"`
-	Ranges []core.KeyRange `json:"ranges"`
-	Batch  int             `json:"batch"`
+	mu      sync.RWMutex
+	storage endpoint.ConfigStorage
+	Name    string          `json:"name"`
```
bufferflies, line 85: how about remove Name?

CabinfeverB: I want to abstract `BaseSchedulerConfig` like `BaseScheduler` in the future, it will use `Name` when persisting config

rleungx: Does that mean we can change the name through API?

CabinfeverB: we can't change the name. I will remove it.

server/schedulers/balance_leader.go
CabinfeverB: I add a var `batch` to read it. After reading,  unlock it immediately

rleungx (approved):

nolouch (approved):

PR #4670, opened 2022-02-23
server: pd can support bucket steam and save them.

### LINKED ISSUE #4668: TiKV/PD support processing flow report in a standalone stream.

### LINKED ISSUE #4669: pd can handler bucket steam and  support searching function.
## Enhancement Task
PD support bucket  heartbeat report and support search.

bufferflies:

Issue Number: Close #4669

```commit-message
```

### Check List

Tests

- Unit test

Code changes

- Has configuration change

Side effects

Related changes

### Release note

```release-note
None.
```


bufferflies:
PTAL @lhy1024 @rleungx

server/cluster/cluster.go
```diff
@@ -591,6 +592,27 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
...
+// processBucketHeartbeat update the bucket information.
+func (c *RaftCluster) processBucketHeartbeat(buckets *metapb.Buckets) error {
+	c.RLock()
+	region := c.core.GetRegion(buckets.GetRegionId())
+	c.RUnlock()
+	if region == nil {
+		return errors.Errorf("region %v not found", buckets.GetRegionId())
+	}
+	for i := 0; i < retry; i++ {
+		old := region.GetBuckets()
```
nolouch, line 604: It breaks the comment:
```
// Read-Only once created.
type RegionInfo struct {
	term              uint64
```
Is there any other place to modify it directly?

nolouch, line 604: It breaks the comment:
```
// Read-Only once created.
type RegionInfo struct {
	term              uint64
```
Is there any other place to modify it directly?

nolouch: And after the members of the region change with the epoch increase, new `regionInfo` will be created and replace the old one, is the bucket missing?

nolouch: And after the members of the region change with the epoch increase, new `regionInfo` will be created and replace the old one, is the bucket missing?

bufferflies: if the confVer of the region has changed,  the bucket of the new leader should be same, so  the origin region should inherit the buckets in the processor of region heartbeat deatils in region.Inherit(origin). so the bucket of RegionInfo should not miss.

server/cluster/cluster_worker.go
```diff
@@ -223,3 +223,8 @@ func (c *RaftCluster) HandleBatchReportSplit(request *pdpb.ReportBatchSplitReque
 		zap.Int("total", last))
 	return &pdpb.ReportBatchSplitResponse{}, nil
 }
+
+// HandleBucketHeartbeat processes RegionInfo reports from client
+func (c *RaftCluster) HandleBucketHeartbeat(buckets *metapb.Buckets) error {
+	return c.processBucketHeartbeat(buckets)
```
rleungx, line 229: Why do we need this abstruction?

server/core/region.go
```diff
@@ -165,18 +167,22 @@ func RegionFromHeartbeat(heartbeat *pdpb.RegionHeartbeatRequest, opts ...RegionC
...
+		}
 		return
 	}
-
-	if origin != nil {
+	if r.approximateSize == 0 {
 		r.approximateSize = origin.approximateSize
-	} else {
-		r.approximateSize = EmptyRegionApproximateSize
+	}
```
rleungx, line 183: I think the original code is easier for reading.

server/core/region.go
```diff
@@ -406,6 +412,30 @@ func (r *RegionInfo) GetStat() *pdpb.RegionStat {
 	}
 }
 
+// SetBuckets sets the buckets of the region.
+func (r *RegionInfo) SetBuckets(buckets *metapb.Buckets) bool {
```
rleungx, line 416: I prefer not to use `setxxx` for region directly.

rleungx:
The implementation is almost the same as region heartbeat, can we abstract the common part to reduce code?

server/cluster/cluster_worker.go
bufferflies: In the first, HandleBucketHeartbeat should handle the change of the bucket meta(keys and version), but in the second , it will handle the statistics.

server/core/region.go
bufferflies: fix

server/core/region.go
bufferflies: How about using UpdateBuckets?

bufferflies:
PTAL @lhy1024 @rleungx again.

bufferflies:
PTAL @rleungx  @nolouch  again

bufferflies:
PTAL @rleungx @AndreMouche see again

server/grpc_service.go
```diff
@@ -648,9 +648,48 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
 	return resp, nil
 }
 
-const regionHeartbeatSendTimeout = 5 * time.Second
+const heartbeatSendTimeout = 5 * time.Second
 
-var errSendRegionHeartbeatTimeout = errors.New("send region heartbeat timeout")
+var errSendHeartbeatTimeout = errors.New("send heartbeat timeout")
+
+// bucketHeartServer wraps PD_ReportBucketsServer to ensure when any error
```
rleungx, line 655:
```suggestion
// bucketHeartbeatServer wraps PD_ReportBucketsServer to ensure when any error
```

server/cluster/cluster.go
```diff
@@ -606,6 +607,33 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
 	return nil
 }
 
+// processBucketHeartbeat update the bucket information.
+func (c *RaftCluster) processBucketHeartbeat(buckets *metapb.Buckets) error {
+	c.RLock()
+	region := c.core.GetRegion(buckets.GetRegionId())
+	c.RUnlock()
```
AndreMouche, line 614: Why we need lock? It seems `region := c.core.GetRegion(buckets.GetRegionId())` is thread safe

server/cluster/cluster.go
```diff
@@ -606,6 +607,33 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
 	return nil
 }
 
+// processBucketHeartbeat update the bucket information.
+func (c *RaftCluster) processBucketHeartbeat(buckets *metapb.Buckets) error {
+	c.RLock()
```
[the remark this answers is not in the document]
bufferflies: yes, buckets will lose if the duration of the region heartbeat is too slow.

server/cluster/cluster.go
```diff
@@ -606,6 +607,33 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
...
+	region := c.core.GetRegion(buckets.GetRegionId())
+	c.RUnlock()
+	if region == nil {
+		bucketEventCounter.WithLabelValues("region_cache_miss").Inc()
+		return errors.Errorf("region %v not found", buckets.GetRegionId())
+	}
+
+	for i := 0; i < retry; i++ {
```
AndreMouche, line 620: It seems the only thing we want to retry is `UpdateBuckets`? So could we try it directly, the L621-L626 seems shouldn't be included in the retry logic.

server/metrics.go
```diff
@@ -85,6 +91,15 @@ var (
 			Buckets:   prometheus.ExponentialBuckets(0.0005, 2, 13),
 		})
 
+	bucketReportHandleDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "pd",
+			Subsystem: "scheduler",
```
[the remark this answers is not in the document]
bufferflies: how about server?

server/grpc_service.go
bufferflies: fix

server/cluster/cluster.go
```diff
@@ -606,6 +607,33 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
...
+		old := region.GetBuckets()
+		// region should not update if the version of the buckets is less than the old one.
+		if old != nil && old.Version >= buckets.Version {
+			bucketEventCounter.WithLabelValues("version_not_match").Inc()
+			return nil
+		}
+		if ok := region.UpdateBuckets(buckets); ok {
+			log.Info("update buckets successful", zap.Uint64("region-id", buckets.GetRegionId()),
```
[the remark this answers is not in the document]
bufferflies: fix

server/core/region.go
```diff
@@ -39,7 +40,8 @@ func errRegionIsStale(region *metapb.Region, origin *metapb.Region) error {
 }
 
 // RegionInfo records detail region info.
-// Read-Only once created.
+// most properties is Read-Only once created exclude buckets.
+// buckets should be modified through report buckets and the version is greater than the current.
```
AndreMouche, line 44:
```suggestion
// the `buckets` could be modified by the request `report buckets` with greater version.
```
Does the above make sense?

server/core/region.go
```diff
@@ -39,7 +40,8 @@ func errRegionIsStale(region *metapb.Region, origin *metapb.Region) error {
 }
 
 // RegionInfo records detail region info.
-// Read-Only once created.
+// most properties is Read-Only once created exclude buckets.
```
AndreMouche, line 43:
```suggestion
// the properties are Read-Only once created except buckets.
```

server/core/region.go
```diff
@@ -165,18 +168,21 @@ func RegionFromHeartbeat(heartbeat *pdpb.RegionHeartbeatRequest, opts ...RegionC
 	return region
 }
 
+// Inherit inherits the buckets and region size from the parent region.
 // CorrectApproximateSize correct approximate size by the previous size if here exists an reported RegionInfo.
```
AndreMouche, line 172: So what is `CorrectApproximateSize`, should we remove it?

server/core/region.go
```diff
@@ -406,6 +413,30 @@ func (r *RegionInfo) GetStat() *pdpb.RegionStat {
 	}
 }
 
+// UpdateBuckets sets the buckets of the region.
+func (r *RegionInfo) UpdateBuckets(buckets *metapb.Buckets) bool {
+	// bucket is only nil in test cases.
```
AndreMouche, line 418: The bucket shouldn't be nil except in the test cases?

server/core/region.go
```diff
@@ -58,6 +60,7 @@ type RegionInfo struct {
 	replicationStatus *replication_modepb.RegionReplicationStatus
 	QueryStats        *pdpb.QueryStats
 	flowRoundDivisor  uint64
+	buckets           unsafe.Pointer
```
AndreMouche, line 63: Why not define as `*metapb.Buckets` directly?

server/cluster/cluster.go
bufferflies: fix

server/core/region.go
bufferflies: fix

server/core/region.go
bufferflies: It may be updated by concurrency。

server/core/region.go
bufferflies: the comment will be " correct approximate size and buckets by the previous size if here exists a reported RegionInfo."

server/cluster/cluster.go
bufferflies: fix

server/metrics.go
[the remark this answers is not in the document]
rleungx: I think it should be `server`.

server/metrics.go
```diff
@@ -24,7 +24,13 @@ var (
...
 			Help:      "Counter of system time jumps backward.",
 		})
-
+	bucketReportCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "pd",
+			Subsystem: "server",
+			Name:      "bucket_report",
+			Help:      "Counter of bucket report.",
+		}, []string{"type", "status"})
```
rleungx, line 33: need store?

server/metrics.go
```diff
@@ -85,6 +91,15 @@ var (
...
 		})
 
+	bucketReportHandleDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "pd",
+			Subsystem: "server",
+			Name:      "handle_bucket_report_duration_seconds",
+			Help:      "Bucketed histogram of processing time (s) of handled bucket report requests.",
+			Buckets:   prometheus.ExponentialBuckets(0.0001, 2, 29), // 0.1ms ~ 7hours
+		}, []string{})
```
rleungx, line 101: ditto

server/grpc_service.go
```diff
@@ -689,6 +728,83 @@ func (s *heartbeatServer) Recv() (*pdpb.RegionHeartbeatRequest, error) {
...
+		}
+		start := time.Now()
+		bucketReportCounter.WithLabelValues("report", "recv").Inc()
+		buckets := request.GetBuckets()
+		if buckets == nil || len(buckets.Keys) == 0 {
+			continue
+		}
+		err = rc.HandleBucketHeartbeat(request.Buckets)
+		if err != nil {
+			log.Warn("update bucket failed", zap.Error(err))
```
rleungx, line 799: Prefer to remove this log or change the level.

server/metrics.go
bufferflies: the request of the stream does not container the information of the region leader or store id,  it does not need to distinguish the report count by store.

server/grpc_service.go
bufferflies: fix

server/core/region.go
bufferflies: fix

server/metrics.go
rleungx: Then do we know if the buckets are from one store or different stores?

server/cluster/cluster.go
```diff
@@ -606,6 +606,23 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
...
+// processBucketHeartbeat update the bucket information.
+func (c *RaftCluster) processBucketHeartbeat(buckets *metapb.Buckets) error {
+	region := c.core.GetRegion(buckets.GetRegionId())
+	if region == nil {
+		bucketEventCounter.WithLabelValues("region_cache_miss").Inc()
+		return errors.Errorf("region %v not found", buckets.GetRegionId())
+	}
+
+	// region should not update if the version of the buckets is less than the old one.
+	if old := region.GetBuckets(); old != nil && old.Version >= buckets.Version {
```
JmPotato, line 618: Shouldn't it be `old.Version > buckets.Version`?

server/cluster/cluster_test.go
```diff
@@ -427,10 +427,51 @@ func (s *testClusterInfoSuite) TestRegionHeartbeatHotStat(c *C) {
...
+	newRegion := regions[0].Clone(core.WithIncConfVer())
+	c.Assert(cluster.processRegionHeartbeat(newRegion), IsNil)
+	c.Assert(cluster.GetRegion(uint64(0)).GetBuckets(), NotNil)
+}
+
 func (s *testClusterInfoSuite) TestRegionHeartbeat(c *C) {
 	_, opt, err := newTestScheduleConfig()
 	c.Assert(err, IsNil)
 	cluster := newTestRaftCluster(s.ctx, mockid.NewIDAllocator(), opt, storage.NewStorageWithMemoryBackend(), core.NewBasicCluster())
+
```
JmPotato, line 474:
```suggestion
```

server/core/region.go
JmPotato: We better add a comment to explain the concurrency case here.

server/core/region.go
```diff
@@ -165,18 +168,20 @@ func RegionFromHeartbeat(heartbeat *pdpb.RegionHeartbeatRequest, opts ...RegionC
...
-// CorrectApproximateSize correct approximate size by the previous size if here exists an reported RegionInfo.
-//
+// Inherit inherits the buckets and region size from the parent region.
+// correct approximate size and buckets by the previous size if here exists a reported RegionInfo.
 // See https://github.com/tikv/tikv/issues/11114
-func (r *RegionInfo) CorrectApproximateSize(origin *RegionInfo) {
-	if r.approximateSize != 0 {
-		return
+func (r *RegionInfo) Inherit(origin *RegionInfo) {
+	// regionSize should not be zero if region is not empty.
```
JmPotato, line 175: I think we can check whether the `origin` is nil at the beginning of this function and return directly if it is.

server/grpc_service.go
```diff
@@ -689,6 +728,82 @@ func (s *heartbeatServer) Recv() (*pdpb.RegionHeartbeatRequest, error) {
 	return req, nil
 }
 
+// ReportBuckets implements gRPC PDServer
+func (s *GrpcServer) ReportBuckets(stream pdpb.PD_ReportBucketsServer) error {
```
JmPotato, line 732: Why not keep the convention to name it `BucketsHeartbeat`?

server/grpc_service.go
```diff
@@ -689,6 +728,82 @@ func (s *heartbeatServer) Recv() (*pdpb.RegionHeartbeatRequest, error) {
...
+			return err
+		}
+		start := time.Now()
+		bucketReportCounter.WithLabelValues("report", "recv").Inc()
+		buckets := request.GetBuckets()
+		if buckets == nil || len(buckets.Keys) == 0 {
+			continue
+		}
+		err = rc.HandleBucketHeartbeat(request.Buckets)
+		if err != nil {
```
JmPotato, line 798: I think the `err` message needs to be output somewhere so we can check what's wrong.

server/grpc_service.go
```diff
@@ -788,19 +788,26 @@ func (s *GrpcServer) ReportBuckets(stream pdpb.PD_ReportBucketsServer) error {
...
-		start := time.Now()
-		bucketReportCounter.WithLabelValues("report", "recv").Inc()
 		buckets := request.GetBuckets()
 		if buckets == nil || len(buckets.Keys) == 0 {
 			continue
 		}
-		err = rc.HandleBucketHeartbeat(request.Buckets)
+		store := rc.GetLeaderStoreByRegionID(buckets.GetRegionId())
+		if store == nil {
+			return errors.Errorf("region's store %d not found", buckets.GetRegionId())
```
rleungx, line 797: The error message is not clear

server/metrics.go
bufferflies: fix

server/metrics.go
bufferflies: fix

server/cluster/cluster.go
bufferflies: region will not update if version is not change.

server/core/region.go
bufferflies: region size will be 1 if origin is nil and the size of the  new region  is zero.

server/core/region.go
bufferflies: fix

server/grpc_service.go
bufferflies: In past, the one error is that the region is not found.

server/grpc_service.go
bufferflies: how about `the store of the bucket %v is not found`

rleungx: `%v` actually is region ID

server/grpc_service.go
bufferflies: yes, the name called the `BucketsHeartbeat` is better than `ReportBuckets`,  but I think the past internal grpc function is also ok .

server/grpc_service.go
bufferflies: the bucket has no constant ID and it will report all buckets once that belongs one same region.

rleungx: How about `the store of the bucket in region %v is not found`

bufferflies: fix

server/cluster/cluster.go
JmPotato: It seems the region heartbeat only checks `<` here. As for a region, the version only changes when splitting or merging happens, so it's still necessary to update other statistical information while the version remains the same. I assume a bucket should also work like that, right?

https://github.com/tikv/pd/blob/f605a2cb030cb03376e0f927b1fdc0bc8b60c488/server/core/basic_cluster.go#L362

rleungx: Is there a possibility that we have an ABA problem here? Consider the following cases: 
There are two requests coming at the same time. Both request A and request B have passed the version check, A is a little bit earlier than B. But B which has a larger version is finished earlier than A. After A is finished, B is lost.

bufferflies: > It seems the region heartbeat only checks `<` here. As for a region, the version only changes when splitting or merging happens, so it's still necessary to update other statistical information while the version remains the same. I assume a bucket should also work like that, right?
> 
> https://github.com/tikv/pd/blob/f605a2cb030cb03376e0f927b1fdc0bc8b60c488/server/core/basic_cluster.go#L362

the version of the bucket maybe happened at cron job or wirte skew, so it should drop the buckets info if the version is same.

server/cluster/cluster.go
```diff
@@ -606,6 +606,32 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
...
+		// region should not update if the version of the buckets is less than the old one.
+		if old != nil && buckets.GetVersion() <= old.GetVersion() {
+			bucketEventCounter.WithLabelValues("version_not_match").Inc()
+			return nil
+		}
+		if ok := region.UpdateBuckets(buckets, old); ok {
+			return nil
+		}
+	}
+	bucketEventCounter.WithLabelValues("bucket_update_failed").Inc()
```
rleungx, line 631:
```suggestion
	bucketEventCounter.WithLabelValues("update_failed").Inc()
```

rleungx (approved):
Mostly LGTM. And we need a concurrency test for this PR.

bufferflies:
> Mostly LGTM. And we need a concurrency test for this PR.

I have added one concurrency test for ABA  and one benchmark test for region update.

server/cluster/cluster.go
bufferflies: > Is there a possibility that we have an ABA problem here? Consider the following case: There are two requests coming at the same time. Both request A and request B have passed the version check, A is a little bit earlier than B. But B which has a larger version is finished earlier than A. After A is finished, B is lost.

yes, the conor will be happened , I will use CAS to avoid the latest buckets to be replaced.

JmPotato (approved):

PR #4672, opened 2022-02-23
api: validate config changes before it updates.

### LINKED ISSUE #4659: region may be splited after the region is merged due to some wrong configuration.
## Enhancement Task
One big region that's regionSize greater than `region-split-size` will be split some small regions. And the smallest splited region may less than `merge-region-size` will be merge. So the merge check will always run .

example 1: 
`merge-region-size`: 70Mb
`split-max-size`: 96Mb
`region-max-size`: 144Mb

the target region's size is 144Mb, it will split two region:
`region-A`:96Mb
`region-B`: 48Mb

the region-B can be merged because the region size less than `merge-region-size`, and will split again.

### Todo 
the necessary and sufficient conditions of the problem is below:

```
region-max-size > region-split_size (1)
region-split-size > merge-region-size (2)
split-max-size - region-split-size> merge-region-size (3)
mod(region-max-size+merge-region-size, region-split-size)> merge-region-size (4)
```

bufferflies:

Issue Number: Close #4659
wait: #4635

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
validate config changes before it updates.
```


conf/config.toml
```diff
@@ -23,6 +23,8 @@
 
 ## Join to an existing cluster. The value should be cluster's ${advertise-client-urls}
 # join = ""
+# max-region-size = 96
```
JmPotato, line 26: What about adding a comment to tell the unit here?

server/cluster/cluster.go
```diff
@@ -488,6 +491,11 @@ func (c *RaftCluster) GetOpts() *config.PersistOptions {
 	return c.opt
 }
 
+// GetConfig gets the cluster configuration.
```
JmPotato, line 494: Please fix this line with the real function name.

disksing (approved):

disksing:
LGTM. But I'm thinking do we need to update `tiup` to make sure the config is synced with tikv's config.

server/config/config.go
```diff
@@ -164,6 +164,10 @@ type Config struct {
 	ReplicationMode ReplicationModeConfig `toml:"replication-mode" json:"replication-mode"`
 
 	EnableAuditMiddleware bool
+
+	MaxRegionSize uint64 `toml:"max-region-size" json:"max-region-size"`
```
nolouch, line 168: Does this PR need to be closed? We don't need the configuration anymore?

bufferflies: yes

bufferflies:
close, it will reopen after get config from tikv

Issue #4400, opened 2021-11-29
api,pd-ctl: support endKey in ScanRegions http interface


IcePigZDB:
## Development Task
Tracking Issue: pingcap/tidb/pull/28330

- [ ] #4142
- [ ] #4371

PR #4674, opened 2022-02-24
api: set scanregions with endkey default limit as 16

IcePigZDB:
ref: #4400
Signed-off-by: IcePigZDB <icepigzdb@gmail.com>

In the current [implementation](https://github.com/tikv/pd/blob/318f9ceddcb0a35a5a61a2121741a251c85064ec/server/api/region.go#L326) of `scanRegions` , the default limit is `defaultRegionLimit(16)` without `endkey` and `noRegionLimit(-1)` with `endkey`.
> Use this command to query all Regions in a given range [startkey, endkey). It supports range without endKey, the default value of limit is 16 without endKey and -1 (no limit)  with endKey. 

  * When one try to use `startKey = a`, `endKey = ""` try to get all regions in `[a, )`, the default limit is `defaultRegionLimit(16)`, need to set `limit = -1` to get all regions, this is inconsistent with the case where `endKey` is not empty.
  * No default limit when using pd-ctl to query `endKey` may cause a lot of network IO cost.
  
For these two reasons, tend to set the default limit with `endKey` to `defaultRegionLimit(16)` the same as without `endKey`. One can set `limit = -1` to get all regions.

### Release note

```release-note
None
```


IcePigZDB:
@nolouch PTAL

disksing:
I think this patch makes sense. Do we have any special reason before? @nolouch @rleungx

IcePigZDB:
> I think this patch makes sense. Do we have any special reason before? @nolouch @rleungx

The support for endkey was added by me before, which was not fully considered. It seems that this part of code has been introduced by 5.4.0, and I wonder whether this path will affect compatibility :worried:.

disksing (approved):

nolouch (approved):

PR #4676, opened 2022-02-25
server: small refactor for grpc-server

### LINKED ISSUE #4675: grpc-service has many similar code to handler the forward request.
## Enhancement Task
some similar code can be exracted, like this :
https://github.com/tikv/pd/blob/4b6f00c3ed0ee52d7cc60b15a446357a7d27b53b/server/grpc_service.go#L388-L400

bufferflies:

Issue Number: Close #4675

```commit-message
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


server/grpc_service.go
```diff
@@ -59,6 +59,28 @@ var (
 	ErrNotStarted = status.Errorf(codes.Unavailable, "server not started")
 )
 
+type Header interface {
```
JmPotato, line 62: Comments are needed.

JmPotato (commented):
The reset LGTM.

server/grpc_service.go
```diff
@@ -59,6 +59,28 @@ var (
...
+	forwardedHost := getForwardedHost(ctx)
+	if !s.isLocalRequest(forwardedHost) {
+		client, err := s.getDelegateClient(ctx, forwardedHost)
+		if err != nil {
+			return nil, err
+		}
+		ctx = grpcutil.ResetForwardContext(ctx)
+		return fn(ctx, client)
+	}
+	if err := s.validateRequest(header.GetHeader()); err != nil {
```
rleungx, line 78: I prefer not to contain it in `forwardRequest`.

disksing (approved):

server/grpc_service.go
bufferflies: fix

server/grpc_service.go
bufferflies: I think the forward request should validate the request also .

server/grpc_service.go
```diff
@@ -59,6 +59,29 @@ var (
 	ErrNotStarted = status.Errorf(codes.Unavailable, "server not started")
 )
 
+// Header returns header metadata.
+type Header interface {
+	GetHeader() *pdpb.RequestHeader
```
rleungx, line 64: We don't need this interface. How about using `ResponseHeader` directly?

server/grpc_service.go
rleungx: Yes, but the function name should be changed. For now, it sounds like only forward requests need to be validated.

server/grpc_service.go
bufferflies: fix

tests/server/global_config/global_config_test.go
```diff
@@ -235,11 +235,3 @@ func (s *GlobalConfigTestSuite) TestClientWatch(c *C) {
 		}
 	}
 }
-
-func (s *GlobalConfigTestSuite) TestClientWatchTimeout(c *C) {
```
JmPotato, line 239: Could we delete this gRPC method directly?

bufferflies: the test function is not steady, and  the global config is not using.

rleungx (approved):

JmPotato:
CI failed. PTAL @bufferflies 

/merge cancel

server/grpc_service.go
```diff
@@ -350,15 +371,12 @@ func watchTSDeadline(ctx context.Context, tsDeadlineCh <-chan deadline) {
...
-	if !s.isLocalRequest(forwardedHost) {
-		client, err := s.getDelegateClient(ctx, forwardedHost)
-		if err != nil {
-			return nil, err
-		}
-		ctx = grpcutil.ResetForwardContext(ctx)
+	fn := func(ctx context.Context, client *grpc.ClientConn) (interface{}, error) {
 		return pdpb.NewPDClient(client).Bootstrap(ctx, request)
 	}
+	if rsp, err := s.unaryMiddleware(ctx, request.GetHeader(), fn); err == nil && rsp != nil {
```
rleungx, line 377: We should handle the cases when the error is not nil.

bufferflies: yes , it will return err if err is not nil

bufferflies:
PTAL @rleungx  again

rleungx (approved):

server/grpc_service.go
```diff
@@ -350,15 +371,14 @@ func watchTSDeadline(ctx context.Context, tsDeadlineCh <-chan deadline) {
...
+	fn := func(ctx context.Context, client *grpc.ClientConn) (interface{}, error) {
 		return pdpb.NewPDClient(client).Bootstrap(ctx, request)
 	}
+	if rsp, err := s.unaryMiddleware(ctx, request.GetHeader(), fn); err != nil {
+		return nil, err
+	} else if rsp != nil {
+		return rsp.(*pdpb.BootstrapResponse), nil
+	}
 
 	if err := s.validateRequest(request.GetHeader()); err != nil {
```
rleungx, line 383: remove it?

bufferflies: fix

PR #4682, opened 2022-02-28
pkg: add progress indicator

rleungx:

Issue Number: Ref #4640.

This PR adds a progress indicator and supports showing offline progress through API or metrics.

```commit-message
```

### Check List

Tests

- Unit test
- Manual test (TODO)

Code changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -20,7 +20,7 @@ import (
 	"time"
 )
 
-// Manager is used to maintain the progresses we are care about.
+// Manager is used to maintain the progresses we care about.
 type Manager struct {
 	sync.RWMutex
 	progesses map[string]*progressIndicator
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -1,0 +1,96 @@
+// Copyright 2022 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package progress
+
+import (
+	"sync"
+	"time"
+)
+
+// Manager is used to maintain the progresses we are care about.
+type Manager struct {
+	sync.RWMutex
+	progesses map[string]*progressIndicator
+}
+
+// NewManager creates a new Manager.
+func NewManager() *Manager {
+	return &Manager{
+		progesses: make(map[string]*progressIndicator),
+	}
+}
+
+// progressIndicator reflects a specified progress.
+type progressIndicator struct {
+	total          float64
+	current        float64
+	startTime      time.Time
+	lastUpdateTime time.Time
+	speedPerSec    float64
+}
+
+// AddProgressIndicator adds a progress into manager.
+func (m *Manager) AddProgressIndicator(progress string, total float64) {
+	m.Lock()
+	defer m.Unlock()
+
+	m.progesses[progress] = &progressIndicator{
+		total:     total,
+		startTime: time.Now(),
+	}
+}
+
+// RemoveProgressIndicator removes a progress from manager.
+func (m *Manager) RemoveProgressIndicator(progress string) {
+	m.Lock()
+	defer m.Unlock()
+
+	delete(m.progesses, progress)
+}
+
+// UpdateProgressIndicator updates the progress of a given name.
+func (m *Manager) UpdateProgressIndicator(progress string, count float64) {
+	m.Lock()
+	defer m.Unlock()
+
+	m.progesses[progress].current += count
+	m.progesses[progress].speedPerSec = count / time.Since(m.progesses[progress].lastUpdateTime).Seconds()
+	m.progesses[progress].lastUpdateTime = time.Now()
+}
+
+// LeftSeconds returns the left seconds until finishing.
+func (m *Manager) LeftSeconds(progress string) float64 {
+	m.RLock()
+	defer m.RUnlock()
+
+	return time.Since(m.progesses[progress].startTime).Seconds() /
+		m.progesses[progress].current * (m.progesses[progress].total - m.progesses[progress].current)
+}
+
+// Process returns the current progress of a give name.
+func (m *Manager) Process(progress string) float64 {
+	m.RLock()
+	defer m.RUnlock()
+
+	return m.progesses[progress].current / m.progesses[progress].total
+}
+
+// CurrentSpeed returns the current speed of a given name.
+func (m *Manager) CurrentSpeed(progress string) float64 {
+	m.RLock()
+	defer m.RUnlock()
+
+	return m.progesses[progress].speedPerSec
+}
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -48,45 +48,36 @@ func (m *Manager) Reset() {
 	m.progesses = make(map[string]*progressIndicator)
 }
 
-// AddProgressIndicator adds a progress into manager if it doesn't exist.
-func (m *Manager) AddProgressIndicator(progress string, total float64) bool {
+// AddOrUpdateProgress adds a progress into manager if it doesn't exist.
+func (m *Manager) AddOrUpdateProgress(progress string, current float64) (exist bool) {
 	m.Lock()
 	defer m.Unlock()
 
-	if _, exist := m.progesses[progress]; exist {
-		return true
-	}
-	m.progesses[progress] = &progressIndicator{
-		total:     total,
-		startTime: time.Now(),
-	}
-	return false
-}
-
-// RemoveProgressIndicator removes a progress from manager.
-func (m *Manager) RemoveProgressIndicator(progress string) bool {
-	m.Lock()
-	defer m.Unlock()
-
-	if _, exist := m.progesses[progress]; exist {
-		delete(m.progesses, progress)
-		return true
-	}
-	return false
-}
-
-// UpdateProgressIndicator updates the progress of a given name.
-func (m *Manager) UpdateProgressIndicator(progress string, current float64) {
-	m.Lock()
-	defer m.Unlock()
-
-	if _, exist := m.progesses[progress]; exist {
+	if _, exist = m.progesses[progress]; exist {
 		m.progesses[progress].current = current
 		if m.progesses[progress].total < current {
 			m.progesses[progress].total = current
 		}
 		m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].startTime).Seconds()
+		return
 	}
+	m.progesses[progress] = &progressIndicator{
+		total:     current,
+		startTime: time.Now(),
+	}
+	return
+}
+
+// RemoveProgress removes a progress from manager.
+func (m *Manager) RemoveProgress(progress string) (exist bool) {
+	m.Lock()
+	defer m.Unlock()
+
+	if _, exist = m.progesses[progress]; exist {
+		delete(m.progesses, progress)
+		return
+	}
+	return
 }
 
 // Process returns the current progress of a give name.
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -40,7 +40,7 @@ type progressIndicator struct {
 	speedPerSec float64
 }
 
-// RemoveProgressIndicator removes a progress from manager.
+// Reset resets the progress manager.
 func (m *Manager) Reset() {
 	m.Lock()
 	defer m.Unlock()
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -53,12 +53,13 @@ func (m *Manager) AddOrUpdateProgress(progress string, current float64) (exist b
 	m.Lock()
 	defer m.Unlock()
 
-	if _, exist = m.progesses[progress]; exist {
-		m.progesses[progress].current = current
-		if m.progesses[progress].total < current {
-			m.progesses[progress].total = current
+	var p *progressIndicator
+	if p, exist = m.progesses[progress]; exist {
+		p.current = current
+		if p.total < current {
+			p.total = current
 		}
-		m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].startTime).Seconds()
+		p.speedPerSec = (p.total - p.current) / time.Since(p.startTime).Seconds()
 		return
 	}
 	m.progesses[progress] = &progressIndicator{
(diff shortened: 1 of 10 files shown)


pkg/progress/progress.go
```diff
@@ -0,0 +1,95 @@
...
+	delete(m.progesses, progress)
+}
+
+// UpdateProgressIndicator updates the progress of a given name.
+func (m *Manager) UpdateProgressIndicator(progress string, current float64) {
+	m.Lock()
+	defer m.Unlock()
+
+	m.progesses[progress].current = current
+	m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].lastUpdateTime).Seconds()
```
JmPotato, line 69: Maybe it's better to provide a speed in recent x duration rather than the whole time?

server/cluster/cluster.go
```diff
@@ -1068,6 +1076,7 @@ func (c *RaftCluster) BuryStore(storeID uint64, forceBury bool) error {
 	if err == nil {
 		// clean up the residual information.
 		c.RemoveStoreLimit(storeID)
+		c.resetProgress(storeID, store.GetAddress(), removingAction)
```
JmPotato, line 1079: What about we forcibly set the store state back to the `Up` state? Do we also need to reset the progress?

JmPotato, line 1079: What about we forcibly set the store state back to the `Up` state? Do we also need to reset the progress?

rleungx: Yes, we need to remove it from progress manager.

pkg/progress/progress.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// AddProgressIndicator adds a progress into manager.
+func (m *Manager) AddProgressIndicator(progress string, total float64) {
+	m.Lock()
+	defer m.Unlock()
+	m.progesses[progress] = &progressIndicator{
+		total:          total,
+		startTime:      time.Now(),
+		lastUpdateTime: time.Now(),
```
AndreMouche, line 51: how about defining a variable `now=time.Now()`  to decrease the number of calling `time.Now()`? It seems [`time.Now`](https://pkg.go.dev/time) costs lot(according the implementation).

pkg/progress/progress.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// UpdateProgressIndicator updates the progress of a given name.
+func (m *Manager) UpdateProgressIndicator(progress string, current float64) {
+	m.Lock()
+	defer m.Unlock()
+
+	m.progesses[progress].current = current
+	m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].lastUpdateTime).Seconds()
+	m.progesses[progress].lastUpdateTime = time.Now()
```
AndreMouche, line 70: Ditto, could we try to decrease the number of calling `time.Now()`? It seems L69 has already called a `time.Now()` in `time.Since(..)`

pkg/progress/progress.go
AndreMouche, line 69: I think the speed should be `(total-current)/(now-start)`?
```suggestion
	m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].startTime).Seconds()
```

pkg/progress/progress.go
rleungx: remove `lastUpdateTime`

server/cluster/cluster.go
```diff
@@ -296,7 +303,11 @@ func (c *RaftCluster) LoadClusterInfo() (*RaftCluster, error) {
 		zap.Duration("cost", time.Since(start)),
 	)
 	for _, store := range c.GetStores() {
-		c.hotStat.GetOrCreateRollingStoreStats(store.GetID())
+		storeID := store.GetID()
+		c.hotStat.GetOrCreateRollingStoreStats(storeID)
+		if store.IsRemoving() {
+			c.progressManager.AddProgressIndicator(fmt.Sprintf("%s-%d", removingAction, storeID), float64(c.core.GetStoreRegionSize(storeID)))
```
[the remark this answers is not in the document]
rleungx: The answer is no, we don't need a lock also since it is in the start process.

pkg/progress/progress.go
```diff
@@ -0,0 +1,95 @@
...
+	startTime      time.Time
+	lastUpdateTime time.Time
+	speedPerSec    float64
+}
+
+// AddProgressIndicator adds a progress into manager.
+func (m *Manager) AddProgressIndicator(progress string, total float64) {
+	m.Lock()
+	defer m.Unlock()
+	m.progesses[progress] = &progressIndicator{
```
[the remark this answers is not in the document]
rleungx: Only two kinds of situations the function will be called: 1. The API triggers. In this case, it will prevent calling twice in the same store. 2. In the start process, we only call it once.

pkg/progress/progress.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return 1 - m.progesses[progress].current/m.progesses[progress].total
+}
+
+// LeftSeconds returns the left seconds until finishing.
+func (m *Manager) LeftSeconds(progress string) float64 {
+	m.RLock()
+	defer m.RUnlock()
+
+	return m.progesses[progress].current / ((m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].startTime).Seconds())
```
JmPotato, line 83: I don't get this calculation. Shouldn't the remaining progress be divided by the speed?

server/cluster/cluster.go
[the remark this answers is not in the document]
AndreMouche: Oh, I mean the lock in `progressManager` ^^

pkg/progress/progress.go
rleungx: Yes, it is. When deleting a store, the current is the remaining progress.

server/api/store.go
```diff
@@ -720,3 +761,12 @@ func parseStoreLimitType(typeName string) (storelimit.Type, error) {
 	}
 	return typeValue, err
 }
+
+func checkProgressAction(action string) error {
+	switch action {
+	case metapb.NodeState_Preparing.String(), metapb.NodeState_Removing.String():
+		return nil
+	default:
+		return errors.New("unknown type")
```
JmPotato, line 770:
```suggestion
		return errors.New("unknown action type")
```

pkg/progress/progress.go
```diff
@@ -0,0 +1,113 @@
...
+// RemoveProgressIndicator removes a progress from manager.
+func (m *Manager) RemoveProgressIndicator(progress string) {
+	m.Lock()
+	defer m.Unlock()
+
+	delete(m.progesses, progress)
+}
+
+// UpdateProgressIndicator updates the progress of a given name.
+func (m *Manager) UpdateProgressIndicator(progress string, current float64) {
```
JmPotato, line 72: What about making it accept `int64` and doing the float64 conversion inside the function?

JmPotato (approved):
LGTM

server/api/store.go
```diff
@@ -610,6 +610,47 @@ func (h *storesHandler) GetStoreLimitScene(w http.ResponseWriter, r *http.Reques
...
+	LeftSeconds  float64 `json:"left_seconds"`
+}
+
+// @Tags stores
+// @Summary Get store progress in the cluster.
+// @Produce json
+// @Success 200 {object} StoreProgress
+// @Failure 400 {string} string "The input is invalid."
+// @Failure 500 {string} string "PD server failed to proceed the request."
+// @Router /stores/{id}/progress [get]
```
nolouch, line 628: Can it be?
```
/stores/progress/{id}
/stores/progress/{action}
```

rleungx: Of cause, we can, but I'm not sure which one is better in API v1.

pkg/progress/progress.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Reset resets the progress manager.
+func (m *Manager) Reset() {
+	m.Lock()
+	defer m.Unlock()
+
+	for progress := range m.progesses {
+		delete(m.progesses, progress)
+	}
```
AndreMouche, line 50: how about `m.progress = make(map[string]*progressIndicator)`?

server/cluster/cluster.go
```diff
@@ -1207,6 +1239,28 @@ func (c *RaftCluster) checkStores() {
 	}
 }
 
+func (c *RaftCluster) updateProgress(storeID uint64, storeAddress string, action string, current int64) {
+	storeLabel := fmt.Sprintf("%d", storeID)
+	progress := fmt.Sprintf("%s-%s", action, storeLabel)
+	if !c.progressManager.IsProgressExist(progress) {
+		c.progressManager.AddProgressIndicator(progress, float64(current))
```
AndreMouche, line 1246: Should we consider the following situation(t1<t2<t3<t4)?
t1: thread 1 has finished L1245 and gets a `false` from `IsProgressExist`
t2: thread 2  share the same `progress` with thread 1 and gets a `false` from `IsProgressExist`
t3: thread 1 finish L1246 `AddProgressIndicator`, and now the value in `progressManager` was initted by current1
t4: thread 2 finish L1246 `AddProgressIndicator`, and now the value in `progressManager` was replaced by current2

pkg/progress/progress.go
```diff
@@ -0,0 +1,113 @@
...
+
+	delete(m.progesses, progress)
+}
+
+// UpdateProgressIndicator updates the progress of a given name.
+func (m *Manager) UpdateProgressIndicator(progress string, current float64) {
+	m.Lock()
+	defer m.Unlock()
+
+	m.progesses[progress].current = current
```
AndreMouche, line 76: Was it possible that `m.progresses[progress]` did not exist?

rleungx: It will be checked by the caller to make sure this situation does not happen.

rleungx: move the check into the function

server/cluster/cluster.go
rleungx: ditto

pkg/progress/progress.go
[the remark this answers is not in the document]
AndreMouche: How about merge the function `AddProgressIndicator`, `Update` into 1
```
// AddProgressIndicator inits or updates the progress of a given name. If the progress with the given name not exist, update it,
// returns true when the progress already exist.
func (m *Manager) AddProgressIndicator(progress string, current float64) (exist bool) {
	m.Lock()
	defer m.Unlock()
	curPro, ok := m.progesses[progress]
	if !ok {
		m.progesses[progress] = &progressIndicator{
			total:     current,
			startTime: time.Now(),
		}
		return ok
	}
	curPro.current = current
	if curPro.total < current {
		curPro.total = current
	}
	m.progesses[progress].speedPerSec = (m.progesses[progress].total - m.progesses[progress].current) / time.Since(m.progesses[progress].startTime).Seconds()
	return ok
}
```

pkg/progress/progress.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// AddOrUpdateProgress adds a progress into manager if it doesn't exist.
+func (m *Manager) AddOrUpdateProgress(progress string, current float64) (exist bool) {
+	m.Lock()
+	defer m.Unlock()
+
+	if _, exist = m.progesses[progress]; exist {
+		m.progesses[progress].current = current
+		if m.progesses[progress].total < current {
```
AndreMouche, line 58: Could you please create a variable pointer to `m.progress[progress]` on L56, so we can use the pointer directly when we want to visit or change the value. Using `m.progress[progress]` to find the pointer in the map is expensive: ref https://github.com/golang/go/blob/master/src/runtime/map.go#L395

server/api/store.go
```diff
@@ -720,3 +761,12 @@ func parseStoreLimitType(typeName string) (storelimit.Type, error) {
 	}
 	return typeValue, err
 }
+
+func checkProgressAction(action string) error {
+	switch action {
+	case metapb.NodeState_Preparing.String(), metapb.NodeState_Removing.String():
+		return nil
+	default:
```
nolouch, line 769: I think the user does not need to specify an `action`. one store must has one action?

server/api/store.go
```diff
@@ -610,6 +610,47 @@ func (h *storesHandler) GetStoreLimitScene(w http.ResponseWriter, r *http.Reques
...
+		apiutil.ErrorResp(h.rd, w, errcode.NewInvalidInputErr(errParse))
+		return
+	}
+
+	action := r.URL.Query().Get("action")
+	if err := checkProgressAction(action); err != nil {
+		h.rd.JSON(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	progress, currentSpeed, leftSeconds := h.Handler.GetStoreProgressByID(action, storeID)
```
nolouch, line 642: Same as below, `GetStoreProgressByID` need action, how get progress by `GetStoreProgressByAction` if I want to remove or add multiple nodes.

server/api/store.go
rleungx: For those stores which don't in the preparing or removing state, they don't have a progress indicator.

server/api/store.go
rleungx: The problem is that do we want to know the progress of each store? We can add another interface to return specified action progress.

nolouch: I think it's needed, because there may exist removing stores and preparing stores during the same time period, the user may separately check it.

server/api/store.go
nolouch: we can return empty if there don't have a progress indicator.

AndreMouche (approved):
LGTM

server/cluster/cluster.go
```diff
@@ -1073,13 +1082,18 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 
 	newStore := store.Clone(core.OfflineStore(physicallyDestroyed))
 	log.Warn("store has been offline",
-		zap.Uint64("store-id", newStore.GetID()),
+		zap.Uint64("store-id", storeID),
 		zap.String("store-address", newStore.GetAddress()),
 		zap.Bool("physically-destroyed", newStore.IsPhysicallyDestroyed()))
 	err := c.putStoreLocked(newStore)
 	if err == nil {
+		c.progressManager.AddProgress(fmt.Sprintf("%s-%d", removingAction, store.GetID()), float64(c.core.GetStoreRegionSize(storeID)))
```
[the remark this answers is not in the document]
rleungx: IMO, `AddProgress` is a common method, we may have other actions which are different from the current format.

pkg/progress/progress.go
```diff
@@ -0,0 +1,122 @@
...
+
+package progress
+
+import (
+	"math"
+	"sync"
+	"time"
+)
+
+// Manager is used to maintain the progresses we are care about.
```
JmPotato, line 23:
```suggestion
// Manager is used to maintain the progresses we care about.
```

pkg/progress/progress.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+// RemoveProgress removes a progress from manager.
+func (m *Manager) RemoveProgress(progress string) (exist bool) {
+	m.Lock()
+	defer m.Unlock()
+
+	if _, exist = m.progesses[progress]; exist {
+		delete(m.progesses, progress)
```
JmPotato, line 87: If `m.progesses` is nil or there is no such element, `delete` is a no-op. So I think there is no need to check the existence first.

rleungx: We need to return if the progress has existed to the caller.

JmPotato (approved):

server/cluster/cluster.go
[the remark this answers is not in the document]
nolouch: why string is a type of process? It's better to enum, so I suggest changing it to explicit action type.

nolouch (approved):

AndreMouche (approved):
LGTM

PR #4687, opened 2022-03-01
*: add node state field

rleungx:

Issue Number: Ref #1566

This PR adds a field in store named `NodeState` and tries to replace `State` in the future. The output of API keeps the same as before for compatibility but the internal logic will use this new field.
kvproto: https://github.com/pingcap/kvproto/pull/868

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -147,19 +147,24 @@ func (s *StoreInfo) IsAvailable(limitType storelimit.Type) bool {
 	return true
 }
 
-// IsUp checks if the store's state is Up.
-func (s *StoreInfo) IsUp() bool {
-	return s.GetState() == metapb.StoreState_Up
+// IsPreparing checks if the store's state is preparing.
+func (s *StoreInfo) IsPreparing() bool {
+	return s.GetNodeState() == metapb.NodeState_Preparing
 }
 
-// IsOffline checks if the store's state is Offline.
-func (s *StoreInfo) IsOffline() bool {
-	return s.GetState() == metapb.StoreState_Offline
+// IsServing checks if the store's state is serving.
+func (s *StoreInfo) IsServing() bool {
+	return s.GetNodeState() == metapb.NodeState_Serving
 }
 
-// IsTombstone checks if the store's state is Tombstone.
-func (s *StoreInfo) IsTombstone() bool {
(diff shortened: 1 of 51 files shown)


server/core/store.go
```diff
@@ -147,19 +147,24 @@ func (s *StoreInfo) IsAvailable(limitType storelimit.Type) bool {
 	return true
 }
 
-// IsUp checks if the store's state is Up.
-func (s *StoreInfo) IsUp() bool {
-	return s.GetState() == metapb.StoreState_Up
+// IsPreparing checks if the store's state is preparing.
```
AndreMouche, line 150: Maybe we can define a new function IsOnline or rewrite the function IsUp, which returns true if the NodeState is Serving or Preparing? Since The function `IsPreparing` and `IsServing` are used together(more than once) which is used to check if the store are online..

AndreMouche (approved):
Rest LGTM

server/core/store.go
lhy1024: LGTM

CabinfeverB:
Some asserts may need to add. Such as 
https://github.com/rleungx/pd/blob/0534d3cf61b046ede320786794161f8822a39fbd/tests/pdctl/store/store_test.go#L278

```
c.Assert(storeInfo.Store.NodeState, Equals, metapb.NodeState_Removed)
```

tests/pdctl/store/store_test.go
```diff
@@ -80,6 +80,7 @@ func (s *storeTestSuite) TestStore(c *C) {
 				Store: &metapb.Store{
 					Id:            2,
 					State:         metapb.StoreState_Tombstone,
+					NodeState:     metapb.NodeState_Removed,
```
CabinfeverB, line 83: Do we need add some assert in this file to test `NodeState `

CabinfeverB (approved):
Rest LGTM

server/core/store.go
rleungx: In the future, maybe we will divide them apart. For now, I just make it work well as before. That's why I doesn't create a new function.

JmPotato (approved):
LGTM

lhy1024 (approved):

PR #4693, opened 2022-03-02
*: Replace juju ratelimit

### LINKED ISSUE #4206: Replace juju/ratelimit
## Development Task
we need to replace juju/ratelimit, consider one:
 - [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)
 - [uber-go/ratelimit](https://github.com/uber-go/ratelimit)

CabinfeverB:

Issue Number: Close #4206 

This PR implements `RateLimiter` which is a rate limiter based on `golang.org/x/time/rate`.
But we need an available function in PD to pre-check store limit. `RateLimiter` implements `Available` function which is not included in `golang.org/x/time/rate`. It uses `ReserveN`, and check  `ReserveN.Delay()`, finally cancel this reserve.

```commit-message
Replace juju ratelimit
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Available is shorthand for AvailableN(1).
+func (l *RateLimiter) Available() bool {
+	return l.AvailableN(1)
+}
+
+// AvailableN returns whether limiter has enough tokens.
+// Note: AvailableN will increase the wait time of WaitN.
+func (l *RateLimiter) AvailableN(n int) bool {
```
rleungx, line 45: we only need `Available` or `AvailableN`

pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,68 @@
...
+	defer l.mu.Unlock()
+	now := time.Now()
+	r := l.Limiter.ReserveN(now, n)
+	delay := r.Delay()
+	r.CancelAt(now)
+	return delay <= 0
+}
+
+// Allow is same as `rate.Limiter.Allow`.
+func (l *RateLimiter) Allow() bool {
```
rleungx, line 56: ditto

pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,68 @@
...
+// AvailableN returns whether limiter has enough tokens.
+// Note: AvailableN will increase the wait time of WaitN.
+func (l *RateLimiter) AvailableN(n int) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	now := time.Now()
+	r := l.Limiter.ReserveN(now, n)
+	delay := r.Delay()
+	r.CancelAt(now)
+	return delay <= 0
```
rleungx, line 52: delay won't be less than 0

pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,68 @@
...
+// RateLimiter is a rate limiter based on `golang.org/x/time/rate`.
+// It implements `Available` function which is not included in `golang.org/x/time/rate`.
+// Note: AvailableN will increase the wait time of WaitN.
+type RateLimiter struct {
+	mu sync.Mutex
+	*rate.Limiter
+}
+
+// NewRateLimiter returns a new Limiter that allows events up to rate r and permits
+// bursts of at most b tokens.
```
JmPotato, line 33: Better to tell the unit of `r` is second?

JmPotato (commented):
Should we involve some tests to make sure its behavior keeps the same with `juju/ratelimit`?

pkg/ratelimit/ratelimiter.go
CabinfeverB: has fixed it

pkg/ratelimit/ratelimiter.go
CabinfeverB: updated

pkg/ratelimit/ratelimiter.go
CabinfeverB: to overwirte /time/rate.Limiter.Allow()

rleungx (approved):

pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,63 @@
...
+// NewRateLimiter returns a new Limiter that allows events up to rate r (it means limiter refill r token per second)
+// and permits bursts of at most b tokens.
+func NewRateLimiter(r float64, b int) *RateLimiter {
+	return &RateLimiter{Limiter: rate.NewLimiter(rate.Limit(r), b)}
+}
+
+// Available returns whether limiter has enough tokens.
+// Note: Available will increase the wait time of WaitN.
+func (l *RateLimiter) Available(n int) bool {
+	l.mu.Lock()
```
nolouch, line 41: Why need the locks?

CabinfeverB: > Why need the locks?

Considering concurrency. If we don't add this lock when an operator named 1 is promoting, exactly right an operator named 2 is doing a store limit check, the tokens will be occupied, and operator named 1 will not be executed

nolouch (approved):
lgtm

PR #4695, opened 2022-03-02
server/schedule/labeler: support ttl for region label

### LINKED ISSUE #4694: Support TTL for region labels
## Feature Request

support `ttl` for region labels.

For example, for the label `schudule=deny` would stop the scheduling work for the special region, with the support of `ttl`, the scheduling work can recover once the region label  expired.

For region_labels,  a `ttl` can be setted.

AndreMouche:

Issue Number: Close #4694 

```commit-message
support ttl for region label
```

usage , post a rule with ttl
```
#!/bin/bash

curl -X POST http://yourip:2379/pd/api/v1/config/region-label/rule -H "Content-Type: application/json" --data-binary @- <<DATA
{"id":"ttltest5",
 "labels": [{
   			"key": "schedule", 
                          "value": "deny",
			  "ttl":"2m" }], 
 "rule_type":"key-range",
 "data": [{
            "start_key": "7480000000000000ff455f720000000000fa",
            "end_key": "7480000000000000ff465f720000000000fa"
        }]
}
DATA

```

### Check List

Tests

- [x] - Unit test
- [x] - Integration test

### Release note

```release-note
support `ttl` for region label.
```


server/schedule/labeler/labeler.go
```diff
@@ -124,7 +183,17 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
 func (l *RegionLabeler) GetLabelRule(id string) *LabelRule {
 	l.RLock()
 	defer l.RUnlock()
-	return l.labelRules[id]
+	return l.getLabelRuleInLock(id, time.Now())
+}
+
+func (l *RegionLabeler) getLabelRuleInLock(id string, now time.Time) *LabelRule {
```
rleungx, line 189:
```suggestion
func (l *RegionLabeler) getLabelRuleLocked(id string, now time.Time) *LabelRule {
```

server/schedule/labeler/labeler.go
```diff
@@ -124,7 +183,17 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
...
+	if rule, ok := l.labelRules[id]; ok {
+		if rule.expireBefore(now) {
+			return nil
+		}
+		return rule
+	}
+	return nil
```
rleungx, line 196:
```suggestion
	if rule, ok := l.labelRules[id]; ok && !rule.expireBefore(now) {
		return rule
	}
	return nil
```

server/cluster/cluster.go
```diff
@@ -236,7 +236,7 @@ func (c *RaftCluster) Start(s Server) error {
 		}
 	}
 
-	c.regionLabeler, err = labeler.NewRegionLabeler(c.storage)
+	c.regionLabeler, err = labeler.NewRegionLabeler(c.ctx, c.storage, time.Hour)
```
rleungx, line 239: How about using a constant?

server/schedule/labeler/labeler.go
```diff
@@ -31,23 +33,71 @@ import (
...
+		case <-ticker.C:
+			toDelete := l.checkAndClearExpiredRulesInMemory()
+			for _, d := range toDelete {
+				if err := l.storage.DeleteRegionRule(d); err != nil {
+					log.Error("failed to remove expired label rule", zap.String("rule-key", d), zap.Error(err))
+				}
+			}
+			log.Debug("RegionLabeler GC items", zap.Int("count", len(toDelete)))
+		case <-l.ctx.Done():
+			log.Debug("GC stopped")
```
rleungx, line 71: Maybe use info level?

server/schedule/labeler/labeler.go
```diff
@@ -122,9 +178,26 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
...
 func (l *RegionLabeler) GetLabelRule(id string) *LabelRule {
-	l.RLock()
-	defer l.RUnlock()
-	return l.labelRules[id]
+	return l.getAndCheckRule(id, time.Now())
+}
+
+func (l *RegionLabeler) getAndCheckRule(id string, now time.Time) *LabelRule {
+	l.Lock()
+	defer l.Unlock()
```
nolouch, line 186: can release the lock before save to storage?

server/schedule/labeler/labeler.go
```diff
@@ -122,9 +178,26 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
 
 // GetLabelRule returns the Rule with the same ID.
 func (l *RegionLabeler) GetLabelRule(id string) *LabelRule {
-	l.RLock()
-	defer l.RUnlock()
-	return l.labelRules[id]
+	return l.getAndCheckRule(id, time.Now())
```
nolouch, line 181: Can we reduce the passing of the `now` parameter, which currently seems unnecessary?

AndreMouche: sorry, we can't. since we should check if there exist expired labels before return the current rule.

server/schedule/labeler/labeler.go
AndreMouche: sorry we can't , since the rule is a pointer get from `labelRules`,  Without lock it could be updated in the other goroutine(s) which is not thread safe.

server/cluster/cluster.go
```diff
@@ -54,6 +54,8 @@ import (
 
 var backgroundJobInterval = 10 * time.Second
 
+const regionLabelGCInterval = time.Hour
```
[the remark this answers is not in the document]
AndreMouche: I think we needn't check the `earlistExpireTime` too frequently because lots of labels expired in 1 hour shouldn't be a normal phenomenon？

server/schedule/labeler/rules.go
```diff
@@ -82,6 +150,15 @@ func (rule *LabelRule) checkAndAdjust() error {
...
 			return errs.ErrRegionRuleContent.FastGenByArgs("empty region label value")
 		}
+		if err := l.checkAndAdjustExpire(); err != nil {
+			err := fmt.Sprintf("region label with invalid ttl info %v", err)
+			return errs.ErrRegionRuleContent.FastGenByArgs(err)
+		}
+	}
+	rule.checkAndRemoveExpireLabels(time.Now())
+	if len(rule.Labels) == 0 {
+		err := fmt.Errorf("region label with expired ttl")
```
rleungx, line 160: directly use the string is ok

server/schedule/labeler/labeler.go
```diff
@@ -31,23 +33,73 @@ import (
...
+	}
+	var err error
+	deleted := false
+
+	for key, rule := range l.labelRules {
+		if !rule.checkAndRemoveExpireLabels(now) {
+			continue
+		}
+		if len(rule.Labels) == 0 {
+			err = l.storage.DeleteRegionRule(key)
```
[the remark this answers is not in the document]
AndreMouche: It will just log the error info.  However, if there is something wrong with the storage last for long, the restart of the server would be long with too much expired items.

server/schedule/labeler/labeler.go
```diff
@@ -122,9 +178,26 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
...
+	defer l.Unlock()
+	rule, ok := l.labelRules[id]
+	if !ok {
+		return nil
+	}
+	if !rule.checkAndRemoveExpireLabels(now) {
+		return rule
+	}
+	if len(rule.Labels) == 0 {
+		l.storage.DeleteRule(id)
```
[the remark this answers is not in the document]
AndreMouche: Here It will do nothing here and the item wouldn't be deleted in the storage until `restart`.  However, if there is something wrong with the storage last for long, the restart of the server would be long with too many expired items.

server/schedule/labeler/labeler_test.go
```diff
@@ -234,6 +261,104 @@ func (s *testLabelerSuite) TestKeyRange(c *C) {
...
+			break
+		}
+	}
+
+	// no rule was cleared because the gc interval is big.
+	c.Assert(labeler.labelRules, HasLen, len(ttls))
+
+	labeler.checkAndClearExpiredLabels()
+
+	labeler.RLock()
```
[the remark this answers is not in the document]
AndreMouche: Because there is a goroutine(The GC function) running which may change the `labeler.labelRules ` concurrently

server/schedule/labeler/rules.go
```diff
@@ -28,18 +29,23 @@ import (
 // RegionLabel is the label of a region.
 // NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
 type RegionLabel struct {
-	Key   string `json:"key"`
-	Value string `json:"value"`
+	Key     string `json:"key"`
+	Value   string `json:"value"`
+	TTL     string `json:"ttl,omitempty"`
+	StartAt string `json:"start_at,omitempty"`
+	start   time.Time
```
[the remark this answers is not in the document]
AndreMouche: `StartAt` is used to store and return to the client, and omit with `no-ttl` labels. with `start` is the time we used to compute.  And Good catch, I would like to try to remote the `start`

server/schedule/labeler/rules.go
```diff
@@ -68,6 +74,68 @@ type LabelRulePatch struct {
...
 
+func (l *RegionLabel) expireBefore(t time.Time) bool {
+	return l.getExpire().Before(t)
+}
+
+var unlimittedExpire time.Time
+
+func init() {
+	unlimittedExpire, _ = time.Parse(time.UnixDate, "Mon Jan 12 22:22:22 MST 2222")
+}
```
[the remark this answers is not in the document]
AndreMouche: Unfortunately, Go doesn't support defining struct constants.  That's why here `time.Time` can't be defined to a const.

server/schedule/labeler/rules.go
```diff
@@ -68,6 +73,69 @@ type LabelRulePatch struct {
...
+		if l.expireBefore(now) {
+			continue
+		}
+		labels = append(labels, l)
+		if l.expireBefore(minExpire) {
+			minExpire = l.expire
+		}
+	}
+
+	if len(labels) == len(rule.Labels) && rule.minExpire.Equal(minExpire) {
```
rleungx, line 131: Why not directly compare with `rule.minExpire` for each label?

server/schedule/labeler/labeler.go
```diff
@@ -78,7 +130,11 @@ func (l *RegionLabeler) loadRules() error {
 
 func (l *RegionLabeler) buildRangeList() {
 	builder := rangelist.NewBuilder()
+	l.earlistExpireTime = unlimittedExpire
```
rleungx, line 133: Perhaps, we can use the first rule's expiration time as the earlistExpireTime then we can avoid defining `unlimittedExpire` as a global variable.

server/schedule/labeler/rules.go
AndreMouche: Good catch

server/schedule/labeler/labeler.go
AndreMouche: Yes, we can use the first rule's expire-time instead of `unlimittedExpire` here, however, to totally void to use `unlimittedExpire`, there may be too many conditions we need to consider, for example, With the `Labels` is empty we need to check both the `Labels` empty thing and `earlistExpireTime` in GC.

rleungx: The purpose is trying to avoid using the init function.

AndreMouche:
Manual test:

tidb+pd+tikv

set region-label with `ttl` by pd for table sbtest2

```
#!/bin/bash

curl -X POST http://yourip:2379/pd/api/v1/config/region-label/rule -H "Content-Type: application/json" --data-binary @- <<DATA

{"id":"schema/sbtest/sbtest2",
 "labels": [{
   			  "key": "schedule", 
               "value": "deny",
			   "ttl":"2h" 
			  },
            {
               "key":"maintainer",
               "value":"shirly"
            }
           ], 
 "rule_type":"key-range",
 "data": [{
            "start_key": "7480000000000000ff455f720000000000fa",
            "end_key": "7480000000000000ff465f720000000000fa"
        }]
}
DATA
```

List region-label in tidb-client

```
mysql> select * from information_schema.attributes\G;
*************************** 1. row ***************************
        ID: schema/sbtest/sbtest2
      TYPE: key-range
ATTRIBUTES: "schedule=deny","maintainer=shirly"
    RANGES: [7480000000000000ff455f720000000000fa, 7480000000000000ff465f720000000000fa]

```

server/schedule/labeler/labeler.go
```diff
@@ -78,14 +130,22 @@ func (l *RegionLabeler) loadRules() error {
 
 func (l *RegionLabeler) buildRangeList() {
 	builder := rangelist.NewBuilder()
+	minExpireInitted := false
 	for _, rule := range l.labelRules {
+		if !minExpireInitted || rule.minExpire.Before(l.earlistExpireTime) {
```
nolouch, line 135: What does `minExpireInitted` do? I think it always becomes true.

server/schedule/labeler/rules.go
```diff
@@ -28,18 +29,22 @@ import (
 // RegionLabel is the label of a region.
 // NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
 type RegionLabel struct {
-	Key   string `json:"key"`
-	Value string `json:"value"`
+	Key     string `json:"key"`
+	Value   string `json:"value"`
+	TTL     string `json:"ttl,omitempty"`
+	StartAt string `json:"start_at,omitempty"`
+	expire  time.Time
```
disksing, line 36: same as above, maybe we can use `*time.Time`

server/schedule/labeler/labeler.go
AndreMouche: The `minExpireInitted` is used to mark that whether  the `l.earlistExpireTime` is initted by the first rule's `minExpire`. So we can avoid init `l.earlistExpireTime` with `unlimitExpire` according to the previous comments cc @rleungx

nolouch (approved):
LGTM

server/schedule/labeler/labeler.go
```diff
@@ -122,9 +179,26 @@ func (l *RegionLabeler) GetLabelRules(ids []string) ([]*LabelRule, error) {
...
+	}
+	if !rule.checkAndRemoveExpireLabels(now) {
+		return rule
+	}
+	if len(rule.Labels) == 0 {
+		l.storage.DeleteRule(id)
+		delete(l.labelRules, id)
+		return nil
+	}
+	l.storage.SaveRegionRule(id, rule)
```
disksing, line 200: why need save here?

server/schedule/labeler/rules.go
```diff
@@ -68,20 +73,79 @@ type LabelRulePatch struct {
...
 	}
-	for _, l := range rule.Labels {
+	for id, l := range rule.Labels {
 		if l.Key == "" {
 			return errs.ErrRegionRuleContent.FastGenByArgs("empty region label key")
 		}
 		if l.Value == "" {
 			return errs.ErrRegionRuleContent.FastGenByArgs("empty region label value")
 		}
+		if err := rule.Labels[id].checkAndAdjustExpire(); err != nil {
```
disksing, line 141:
```suggestion
		if err := l.checkAndAdjustExpire(); err != nil {
```

server/schedule/labeler/labeler.go
AndreMouche: Because there exist expired label which need to be cleared.

server/schedule/labeler/rules.go
AndreMouche: Here we should use `rule.Labels[id]` instead of `l` because `l` is a *copy*(not the address) of the element `rule.Labels[id]`,  and we need to adjust the value of `rule.Labels[i]` here.

disksing: got.

server/schedule/labeler/labeler.go
```diff
@@ -33,21 +35,72 @@ type RegionLabeler struct {
...
 }
 
 // NewRegionLabeler creates a Labeler instance.
-func NewRegionLabeler(storage endpoint.RuleStorage) (*RegionLabeler, error) {
+func NewRegionLabeler(ctx context.Context, storage endpoint.RuleStorage, gcInterval time.Duration) (*RegionLabeler, error) {
 	l := &RegionLabeler{
 		storage:    storage,
 		labelRules: make(map[string]*LabelRule),
+		ctx:        ctx,
+		minExpire:  nil,
```
disksing, line 48:
```suggestion
```

disksing (commented):
LGTM

disksing (approved):

Issue #4698, opened 2022-03-03
Client supports getting region with buckets


youjiali1995:
## Development Task

Adjust the PD client interface to support buckets.

PR #4699, opened 2022-03-03
client: support getting region with buckets

youjiali1995:

Issue Number: Close https://github.com/tikv/pd/issues/4698

```commit-message
Update [kvproto](https://github.com/pingcap/kvproto/pull/865) and PD client to support getting region with buckets.
```

### Check List

Tests

- No code

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


youjiali1995:
@bufferflies PTAL

client/client.go
```diff
@@ -76,13 +77,13 @@ type Client interface {
 	// taking care of region change.
 	// Also it may return nil if PD finds no Region for the key temporarily,
 	// client should retry later.
-	GetRegion(ctx context.Context, key []byte) (*Region, error)
+	GetRegion(ctx context.Context, key []byte, needBuckets bool) (*Region, error)
```
rleungx, line 80: Is it a breaking change?

youjiali1995: Yes. Another way is to add `GetRegionAndBucket()`/`GetPrevRegionAndBucket()`/`GetRegionAndBucketByID()`, but there are many places the `Client` interface is implemented for a mock client, it's also a breaking change and doesn't reduce the workload. Since the main(maybe the only one) user of PD client is client-go and TiDB, I think we can choose either way to utilize the new interface.

bufferflies:
how about add some unit test for needBuckets is true?

youjiali1995:
> how about add some unit test for needBuckets is true?

I don't know how to do it when the server doesn't support buckets now. I depend on you to test it...

client/client.go
disksing: can we use the option pattern. Or can we put the option into context?

client/client.go
```diff
@@ -353,6 +354,8 @@ func WithMaxErrorRetry(count int) ClientOption {
 	}
 }
 
+var _ Client = &client{}
```
disksing, line 357: we don't need this. If method does not match, `NewClient` won't compile.

client/client.go
youjiali1995: OK. I will use the option pattern.

youjiali1995:
@disksing @bufferflies PTAL again, thanks!

tools/pd-tso-bench/go.sum
```diff
@@ -163,9 +161,8 @@ go.uber.org/atomic v1.7.0/go.mod h1:fEN4uk6kAWBTFdckzkM89CLk9XfWZrxpCo0nPH17wJc=
 go.uber.org/atomic v1.9.0 h1:ECmE8Bn/WFTYwEW/bpKD3M8VtR/zQVbavAoalC1PYyE=
 go.uber.org/atomic v1.9.0/go.mod h1:fEN4uk6kAWBTFdckzkM89CLk9XfWZrxpCo0nPH17wJc=
 go.uber.org/goleak v1.1.10/go.mod h1:8a7PlsEVH3e/a/GLqe5IIrQx6GzcnRmZEufDUTk4A7A=
+go.uber.org/goleak v1.1.11 h1:wy28qYRKZgnJTxGxvye5/wgWr1EKjmUDGYox5mGlRlI=
 go.uber.org/goleak v1.1.11/go.mod h1:cwTWslyiVhfpKIDGSZEM2HlOvcqm+tG4zioyIeLoqMQ=
-go.uber.org/goleak v1.1.12 h1:gZAh5/EyT/HQwlpkCy6wTpqfH9H8Lz8zbm3dZh+OyzA=
```
bufferflies, line 167: why degrade it.

bufferflies (approved):
the rest LGTM.

client/client.go
```diff
@@ -76,13 +77,13 @@ type Client interface {
...
-	GetRegion(ctx context.Context, key []byte) (*Region, error)
+	GetRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
 	// GetRegionFromMember gets a region from certain members.
 	GetRegionFromMember(ctx context.Context, key []byte, memberURLs []string) (*Region, error)
 	// GetPrevRegion gets the previous region and its leader Peer of the region where the key is located.
-	GetPrevRegion(ctx context.Context, key []byte) (*Region, error)
+	GetPrevRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
 	// GetRegionByID gets a region and its leader Peer from PD by id.
-	GetRegionByID(ctx context.Context, regionID uint64) (*Region, error)
+	GetRegionByID(ctx context.Context, regionID uint64, opts ...GetRegionOption) (*Region, error)
```
AndreMouche, line 86: Oh, I think it is not a good choice to support `GetRegionByID` with other opts(which may make this API too complex)?

AndreMouche (approved):
LGTM

tools/pd-tso-bench/go.sum
youjiali1995: `go mod tidy` did it.

nolouch (approved):
lgtm

PR #4700, opened 2022-03-03
placement: change ValidateStores condition

### LINKED ISSUE #4690: Flawed judgment conditions when validating stores in rule checker
## Bug Report
There is a problem that has a flawed judgment condition when validating stores in rule checker.
If a store is offline or disconnected, it is not in Up state.
image

master

CabinfeverB:

Issue Number: Close #4690

change `and` to `or`

```commit-message
change ValidateStores condition
```

### Check List

Tests

- Unit test
- 
### Release note

```release-note
 None.
```


rleungx (approved):

lhy1024 (approved):
Do we need to add a test to avoid similar code?

CabinfeverB:
> Do we need to add a test to avoid similar code?

I don't think it's easy to avoid similar code with additional tests. It's good behavior that when creating stores, try to add heartbeat information

Yisaer (approved):

PR #4704, opened 2022-03-03
unsafe recover: multiple stage recovery plan generation

Connor1996:
Signed-off-by: Connor1996 <zbk602423539@gmail.com>

Issue Number: ref https://github.com/tikv/pd/issues/4887

Simplify the original recovery plan, and introduce two stages for recovery:
1. force leader stage: force one peer of the unhealthy region to be leader to make all the peers catch up with the latest log
2. recover stage: dispatch recover operations including remove-peer, create-region and tombstone-peer

Output would be like this:
image

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
Introduce multiple stage online unsafe recovery
```


Connor1996:
tests are to be added

server/cluster/unsafe_recovery_controller.go
```diff
@@ -116,48 +133,52 @@ func (u *unsafeRecoveryController) HandleStoreHeartbeat(heartbeat *pdpb.StoreHea
...
-				resp.RequireDetailedReport = true
-			}
-		} else if report, exist := u.storeReports[heartbeat.Stats.StoreId]; exist && report == nil {
-			u.storeReports[heartbeat.Stats.StoreId] = heartbeat.StoreReport
-			u.numStoresReported++
-			if u.numStoresReported == len(u.storeReports) {
-				log.Info("Reports have been fully collected, generating plan...")
-				go u.generateRecoveryPlan()
-			}
+		if u.collectAndCheckStoreReport(heartbeat, resp) {
```
rleungx, line 136: How about moving it before switch?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -166,210 +187,305 @@ func (u *unsafeRecoveryController) reset() {
...
+	return true
+}
+
+func (u *unsafeRecoveryController) getRecoveryPlan(storeID uint64) *pdpb.RecoveryPlan {
+	storeRecoveryPlan, exists := u.storeRecoveryPlans[storeID]
+	if !exists {
+		u.storeRecoveryPlans[storeID] = &pdpb.RecoveryPlan{}
+		storeRecoveryPlan = u.storeRecoveryPlans[storeID]
+	}
+	return storeRecoveryPlan
```
rleungx, line 399: How about
```suggestion
	_, exists := u.storeRecoveryPlans[storeID]
	if !exists {
		u.storeRecoveryPlans[storeID] = &pdpb.RecoveryPlan{}
	}
	return u.storeRecoveryPlans[storeID]
```

server/cluster/unsafe_recovery_controller.go
```diff
@@ -116,48 +133,52 @@ func (u *unsafeRecoveryController) HandleStoreHeartbeat(heartbeat *pdpb.StoreHea
 	}
 	switch u.stage {
```
rleungx, line 134: Do we need to add a default implementation?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -166,210 +187,305 @@ func (u *unsafeRecoveryController) reset() {
...
 	}
-	region.Peers = newPeerList
+	return false
 }
 
-type peerStorePair struct {
-	peer    *pdpb.PeerReport
-	storeID uint64
+// Less returns true if the region start key is less than the other.
+func (r *regionItem) Less(other btree.Item) bool {
```
rleungx, line 281: Is there possible to avoid the duplicated implementation?

Connor1996: it's hard as the original `regionTree` accepts `RegionInfo`, but this needs peer report

Connor1996:
Tests to be added, It's ready for review. PTAL @rleungx @v01dstar

server/grpc_service.go
```diff
@@ -595,20 +595,19 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+		return &pdpb.StoreHeartbeatResponse{
+			Header: s.errorHeader(pberr),
+		}, nil
+	}
+	storeID := request.GetStats().GetStoreId()
+	store := rc.GetStore(storeID)
+	if store == nil {
+		return nil, errors.Errorf("store %v not found", storeID)
+	}
+
```
rleungx, line 608: `checkStore` also will check it.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -17,102 +17,123 @@ package cluster
...
+	log.Info(msg)
+}
+
 // RemoveFailedStores removes failed stores from the cluster.
-func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]string) error {
+func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]interface{}) error {
 	u.Lock()
 	defer u.Unlock()
+
 	if len(failedStores) == 0 {
```
rleungx, line 97: We have already checked it outside of this function.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		if u.numStoresReported == len(u.storeReports) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (u *unsafeRecoveryController) GetStage() unsafeRecoveryStage {
+	u.Lock()
```
rleungx, line 242: RLock is enough

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
-		region := peerReport.RegionState.Region
-		if _, ok := toBeRemovedRegions[region.Id]; ok {
-			return false
-		} else if target, ok := targetRegions[region.Id]; ok {
-			if bytes.Equal(target.StartKey, region.StartKey) && bytes.Equal(target.EndKey, region.EndKey) && !u.containsFailedPeers(region) {
-				numFinished += 1
-			} else {
-				return false
+
+	if report, exists := u.storeReports[storeID]; exists && report == nil { // if receive duplicated report from same TiKV, just ignore
```
rleungx, line 230: Perhaps we can update the existing one.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		// TODO: abort the process directly
+		return false
 	}
-	for _, del := range u.storeRecoveryPlans[storeID].Deletes {
-		toBeRemovedRegions[del] = true
+
+	if _, find := u.storeReports[storeID]; !find {
+		// No need to collect the report of the store
+		return false
 	}
```
rleungx, line 228: line 230 contains this case

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		}
+		regions := ""
+		for i, regionID := range forceLeaders.GetEnterForceLeaders() {
+			regions += fmt.Sprintf("%d", regionID)
+			if i != len(forceLeaders.GetEnterForceLeaders())-1 {
+				regions += ", "
 			}
 		}
+		output = append(output, fmt.Sprintf(" - store %d", storeID))
+		output = append(output, fmt.Sprintf("   - force leader on regions: %s", regions))
```
rleungx, line 296: `-` can only work well if there are no other logs.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
 			}
-			storeRecoveryPlan.Deletes = append(storeRecoveryPlan.Deletes, region.Id)
 		}
 	}
-	// There may be ranges that are covered by no one. Find these empty ranges, create new regions that cover them and evenly distribute newly created regions among all stores.
+	return hasPlan
+}
+
+func (u *unsafeRecoveryController) generateCreateEmptyRegionPlan() bool {
+	newestRegionTree, _ := u.buildUpFromReports()
```
rleungx, line 678: How about using a flag to control if we need to build the peer map?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -17,102 +17,123 @@ package cluster
...
+		stage:              idle,
+		failedStores:       make(map[uint64]interface{}),
+		storeReports:       make(map[uint64]*pdpb.StoreReport),
+		numStoresReported:  0,
+		storePlanExpires:   make(map[uint64]time.Time),
+		storeRecoveryPlans: make(map[uint64]*pdpb.RecoveryPlan),
 	}
 }
 
+func (u *unsafeRecoveryController) reset() {
```
rleungx, line 77: Does reset need to clean the output?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +142,629 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
-		peers += "(" + getPeerDigest(peer) + "), "
+func (u *unsafeRecoveryController) getReportStatus() []string {
+	var status []string
+	if u.numStoresReported != len(u.storeReports) {
+		status = append(status, fmt.Sprintf("Collecting reports from alive stores(%d/%d):", u.numStoresReported, len(u.storeReports)))
+		var reported, unreported, undispatched string
+		for storeID, report := range u.storeReports {
+			if report == nil {
+				if _, requested := u.storePlanExpires[storeID]; !requested {
+					undispatched += strconv.FormatUint(storeID, 10) + ", "
```
rleungx, line 733: use a variable?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -17,102 +17,123 @@ package cluster
...
 
 type unsafeRecoveryStage int
 
 const (
-	storeReportRequestInterval = time.Second * 60
-	storePlanRequestInterval   = time.Second * 60
+	storeRequestInterval = time.Second * 60
 )
 
 const (
```
rleungx, line 39: Please add a stage transition graph for it.

server/grpc_service.go
Connor1996: `checkStore` doesn't return an error when the store id is not found

server/cluster/unsafe_recovery_controller.go
Connor1996: do you have any better idea?

server/cluster/unsafe_recovery_controller.go
Connor1996: I believe it would be optimized by the compiler somehow?

server/cluster/unsafe_recovery_controller.go
Connor1996: yes

server/cluster/unsafe_recovery_controller.go
Connor1996: what do you mean?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -121,452 +172,664 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+
+	foundSelf := false
+	for _, old := range overlaps {
+		if item.Region().GetId() == old.Region().GetId() {
+			foundSelf = true
+			// not only check epoch for peers of the same region
+			if item.IsStale(old) {
+				return false
+			}
+		} else if item.IsEpochStale(old) {
```
v01dstar, line 579: I am not sure this is what we really want. For example, imagine that there are multiple overlaps, the first overlap is newer (than the one is currently examining), and the second one is older, this code returns false after the first overlap is checked. I know that we not longer pursue "best effort", but ignoring the "newer" one but keeping the "older" one seems also problematic? Shouldn't we also remove the second overlap from the tree map as well?

server/api/unsafe_operation.go
```diff
@@ -55,11 +55,17 @@ func (h *unsafeOperationHandler) RemoveFailedStores(w http.ResponseWriter, r *ht
 		h.rd.JSON(w, http.StatusBadRequest, "Store ids are invalid")
 		return
 	}
-	stores := make(map[uint64]string)
+	stores := make(map[uint64]interface{})
```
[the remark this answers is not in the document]
Connor1996: it complains `server/api/unsafe_operation.go:63:24: invalid operation: input["timeout"] (map index expression of type struct{}) is not an interface`

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +169,744 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+	if u.numStoresReported != len(u.storeReports) {
+		status = append(status, fmt.Sprintf("Collecting reports from alive stores(%d/%d):", u.numStoresReported, len(u.storeReports)))
+		var reported, unreported, undispatched string
+		for storeID, report := range u.storeReports {
+			if report == nil {
+				if _, requested := u.storePlanExpires[storeID]; !requested {
+					undispatched += strconv.FormatUint(storeID, 10) + ", "
+				} else {
+					unreported += strconv.FormatUint(storeID, 10) + ", "
+				}
```
rleungx, line 900:
```suggestion
			if report == nil {
			        str := strconv.FormatUint(storeID, 10) + ", "
				if _, requested := u.storePlanExpires[storeID]; !requested {
					undispatched += str
				} else {
					unreported += str
				}
```

server/cluster/unsafe_recovery_controller.go
Connor1996: I didn't find any way to stringify the zap fields

server/api/unsafe_operation.go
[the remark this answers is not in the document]
rleungx: Why is it related to input?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -36,71 +35,120 @@ import (
 type unsafeRecoveryStage int
 
 const (
-	storeReportRequestInterval = time.Second * 60
-	storePlanRequestInterval   = time.Second * 60
+	storeRequestInterval = time.Second * 60
 )
 
+// Stage transition graph: for more details, please check `unsafeRecoveryController.HandleStoreHeartbeat()`
```
rleungx, line 41: How about adding `forceLeaderForCommitMerge` also?

server/api/unsafe_operation.go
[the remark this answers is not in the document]
Connor1996: Oh, no need to change the `input map[string]interface{}` to `input map[string]struct{}`. change it.

server/cluster/unsafe_recovery_controller.go
Connor1996: Now it's printed in json structure. And all actions in one stage will be output in one log.

server/cluster/unsafe_recovery_controller.go
Connor1996: done

server/api/unsafe_operation_test.go
```diff
@@ -39,6 +41,7 @@ func (s *testUnsafeAPISuite) SetUpSuite(c *C) {
 	s.urlPrefix = fmt.Sprintf("%s%s/api/v1/admin/unsafe", addr, apiPrefix)
 
 	mustBootstrapCluster(c, s.svr)
+	mustPutStore(c, s.svr, 1, metapb.StoreState_Offline, metapb.NodeState_Serving, nil)
```
rleungx, line 44:
```suggestion
	mustPutStore(c, s.svr, 1, metapb.StoreState_Offline, metapb.NodeState_Removing, nil)
```

tools/pd-ctl/pdctl/command/unsafe_command.go
```diff
@@ -84,25 +75,20 @@ func removeFailedStoresCommandFunc(cmd *cobra.Command, args []string) {
...
-
-func removeFailedStoresShowCommandFunc(cmd *cobra.Command, args []string) {
-	var resp string
-	var err error
-	prefix := fmt.Sprintf("%s/remove-failed-stores/show", unsafePrefix)
-	resp, err = doRequest(cmd, prefix, http.MethodGet, http.Header{})
+	timeout, err := cmd.Flags().GetFloat64("timeout")
 	if err != nil {
-		cmd.Println(err)
+		cmd.Usage()
```
rleungx, line 80: Should we print err?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +182,754 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		output.Info = "Unsafe recovery enters force leader stage"
+		output.Actions = u.getForceLeaderPlanDigest()
+	case demoteFailedVoter:
+		output.Info = "Unsafe recovery enters demote failed voter stage"
+		output.Actions = u.getDemoteFailedVoterPlanDigest()
+	case createEmptyRegion:
+		output.Info = "Unsafe recovery enters create empty region stage"
+		output.Actions = u.getCreateEmptyRegionPlanDigest()
+	case finished:
+		u.cluster.PauseOrResumeScheduler("all", 0)
```
rleungx, line 416: Do we need to resume here? The scheduler can run depending on if unsafe recovery is running.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +182,754 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		if u.step > 1 {
+			// == 1 means no operation has done, no need to invalid cache
+			u.cluster.DropCacheAllRegion()
+		}
+		output.Info = "Unsafe recovery finished"
+	case failed:
+		u.cluster.PauseOrResumeScheduler("all", 0)
```
rleungx, line 423: ditto

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +182,754 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+		output.Info = fmt.Sprintf("Unsafe recovery failed: %v", u.err)
+	}
+
+	u.output = append(u.output, output)
+	data, err := json.Marshal(output)
+	if err != nil {
+		panic(err)
```
rleungx, line 430: Should we panic? Maybe an error log is enough.

tools/pd-ctl/pdctl/command/unsafe_command.go
Connor1996: yes, changed

server/cluster/unsafe_recovery_controller.go
Connor1996: no need, changed

server/cluster/unsafe_recovery_controller.go
Connor1996: it can't happen, so I just panic straightforward here

rleungx: Yes, but this is only a log problem, we don't need to panic here.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +182,736 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+
+func (u *unsafeRecoveryController) getCreateEmptyRegionPlanDigest() map[string][]string {
+	outputs := make(map[string][]string)
+	for storeID, plan := range u.storeRecoveryPlans {
+		if plan.GetCreates() == nil {
+			continue
+		}
+		output := []string{}
+		for _, region := range plan.GetCreates() {
+			output = append(output, fmt.Sprintf("create region %v", core.RegionToHexMeta(region)))
```
rleungx, line 494: Should be redacted?

tools/pd-ctl/pdctl/command/unsafe_command.go
rleungx: line 62 also need to be changed

Connor1996: I think line 70 should be changed, but not line 62

server/cluster/unsafe_recovery_controller.go
Connor1996: necessary to know the region range

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +182,737 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
-			storeRecoveryPlan.Creates = append(storeRecoveryPlan.Creates, creates...)
-			if update != nil {
-				storeRecoveryPlan.Updates = append(storeRecoveryPlan.Updates, update)
+			if forCommitMerge && !report.HasCommitMerge {
+				// check force leader only for ones has commit merge to avoid the case that
+				// target region can't catch up log for the source region due to force leader
+				// propose an empty raft log on being leader
+				return true
+			} else if !forCommitMerge && report.HasCommitMerge {
+				panic("unreachable")
```
rleungx, line 771: How about not using panic

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +184,747 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+	if _, isFailedStore := u.failedStores[storeID]; isFailedStore {
+		return false, errors.Errorf("Receive heartbeat from failed store %d", storeID)
+	}
+
+	if heartbeat.StoreReport == nil {
+		return false, nil
+	}
+
+	if heartbeat.StoreReport.GetStep() != u.step {
+		log.Info("Unsafe recovery receives invalid store report",
```
rleungx, line 360: Does the log level need to be changed?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +184,747 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+
+// Insert the peer report of one region int the tree.
+// It finds and deletes all the overlapped regions first, and then
+// insert the new region.
+func (t *regionTree) insert(item *regionItem) bool {
+	overlaps := t.getOverlaps(item)
+
+	if t.contains(item.Region().GetId()) {
+		// it's ensured by the `buildUpFromReports` that only insert the latest peer of one region.
+		panic("region shouldn't be updated twice")
```
rleungx, line 676: It's better to remove all panic since if it failed to recover, we shouldn't let the whole service down.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -120,462 +184,747 @@ func (u *unsafeRecoveryController) RemoveFailedStores(failedStores map[uint64]st
...
+	}
+
+	for _, old := range overlaps {
+		// it's ensured by the `buildUpFromReports` that peers are inserted in epoch descending order.
+		if old.IsEpochStale(item) {
+			panic("region's epoch shouldn't be staler than old ones")
```
rleungx, line 682: ditto

server/cluster/unsafe_recovery_controller.go
Connor1996: it's not an error, it's possible when suffering network issues

rleungx (approved):

v01dstar (approved):
LGTM

JmPotato (approved):

PR #4705, opened 2022-03-04
Periodically instead of limitlessly request for store report.

v01dstar:

Issue: tikv/tikv#10483

Previously, once the unsafe recovery process is triggered, PD requests for store report or sends a recovery plan in every store heartbeat. If a store is just slow to respond, neither dead nor failed, this type of limitless action will only cause backlog. 

Instead of requesting for report or sending plan in every heartbeat, we do these periodically. PD issues the next request only after the previous one has expired and no response has been received.

### Check List

Tests

- Unit test

Code changes

- N/A

Side effects

- All schedulers are paused until the recovery is finished or timed out.

Related changes

- N/A

### Release note

```release-note
None
```


JmPotato (approved):
LGTM. Could you please change the PR title format to "server: periodically instead of limitlessly request for store report"? thx!

server/cluster/unsafe_recovery_controller.go
```diff
@@ -34,6 +35,11 @@ import (
 
 type unsafeRecoveryStage int
 
+const (
+	storeReportRequestInterval = time.Second * 60
+	storePlanRequestInterval   = time.Second * 60
```
rleungx, line 40: Can we use merge them into one?

Connor1996: +1

Connor1996: I'll do it in the next PR, let's merge it first

Connor1996 (approved):
LGTM

PR #4706, opened 2022-03-04
server: get tikv config by status api

### LINKED ISSUE #4459: sync tikv config

bufferflies:

Issue Number: Close #4459

PD should know the tikv configuration to schedule like merge_check.
```commit-message
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


server/config/store_config.go
```diff
@@ -0,0 +1,138 @@
...
+	if c == nil {
+		return value
+	}
+	body := fn()
+	var bs typeutil.ByteSize
+	if err := json.Unmarshal(body, &bs); err != nil {
+		return value
+	}
+	return bs.ToMiB()
+}
```
AndreMouche, line 79: It is too complex, it seems you just need a function to parse MBsize from a string ?

server/config/store_config.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.getMbSize(fn, defaultRegionMaxSize)
+}
+
+// GetRegionSplitSize returns the region split size in MB
+func (c *StoreConfig) GetRegionSplitSize() uint64 {
+	fn := func() []byte {
+		return []byte(`"` + c.Coprocessor.RegionSplitSize + `"`)
+	}
+	return c.getMbSize(fn, defaultRegionSplitSize)
+}
```
AndreMouche, line 95: How about 
```
// GetRegionMaxSize returns the max region size in MB
func (c *StoreConfig) GetRegionMaxSize() uint64 {
	return parseMBFromTxt(c.Coprocessor.RegionMaxSize, defaultRegionMaxSize)

}

// GetRegionSplitSize returns the region split size in MB
func (c *StoreConfig) GetRegionSplitSize() uint64 {
	return parseMBFromTxt(c.Coprocessor.RegionSplitSize, defaultRegionSplitSize)
}

func parseMBFromTxt(text string, defaultSizeInMB uint64) uint64 {
	b := typeutil.ByteSize(0)
	err := b.UnmarshalText([]byte(text))
	if err != nil {
		return defaultSizeInMB
	} else {
		return b.ToMiB()
	}
}


```

bufferflies: yes, I will fix it.

server/grpc_service.go
```diff
@@ -522,6 +522,14 @@ func (s *GrpcServer) PutStore(ctx context.Context, request *pdpb.PutStoreRequest
 	log.Info("put store ok", zap.Stringer("store", store))
 	CheckPDVersion(s.persistOptions)
 
+	go func(url string) {
+		// tikv may not ready to server.
```
lhy1024, line 526: What happens to down tikv?

bufferflies: Keeping  the config of all the tikv  to be same in the cluster is the precondition，so the pd should be sync the tikv config even through some tikv is down.  if the precondition is not exist, the cluster will work well in past.
 In upgrade, the configuration of the TiKV may be different,syncing with tikv is ok in the end.

bufferflies:
PTAL @lhy1024 @AndreMouche again.

bufferflies:
PTAL @AndreMouche @lhy1024 again.

rleungx:
I'm worrying that we had lots of ways to support saving TiKV configs in PD before and all of them are deprecated. Those ways are more general than this PR. So I think we need a further discussion of it before moving forward.

AndreMouche (approved):
LGTM

bufferflies:
> I'm worrying that we had lots of ways to support saving TiKV configs in PD before and all of them are deprecated. Those ways are more general than this PR. So I think we need a further discussion of it before moving forward.

I I very much agree with your opinion about the consistency with other component,  it does need a further discussion. But in past, this is the smallest cost to get the config of the tikv and tikv doesn't need to add other codes. it will need to rewrite the update logic.  I will propose one issue about the centralized configuration,.

server/grpc_service.go
disksing: Does TiFlash have the status API too?

bufferflies: it is a good question, it should ignore tiflash.

disksing (approved):

pkg/typeutil/size_test.go
```diff
@@ -16,7 +16,6 @@ package typeutil
 
 import (
 	"encoding/json"
-
```
lhy1024, line 19: unnecessary modify

server/config/store_config.go
```diff
@@ -0,0 +1,154 @@
...
+
+// NewStoreConfigManager creates a new StoreConfigManager.
+func NewStoreConfigManager(config *SecurityConfig) *StoreConfigManager {
+	manager := &StoreConfigManager{
+		schema: "http",
+	}
+	if config == nil {
+		return manager
+	}
+	if cfg, err := config.ToTLSConfig(); err == nil && cfg != nil {
```
lhy1024, line 56: Do we need to add a tls test?

server/config/store_config.go
```diff
@@ -0,0 +1,154 @@
...
+// StoreConfig is the config of store like TiKV.
+// generated by https://mholt.github.io/json-to-go/.
+// nolint
+type StoreConfig struct {
+	Coprocessor `json:"coprocessor"`
+}
+
+// Coprocessor is the config of coprocessor.
+type Coprocessor struct {
+	RegionMaxSize   string `json:"region-max-size"`
```
lhy1024, line 74: how about using atomic value to avoid introducing lock?

server/schedule/checker/merge_checker.go
```diff
@@ -32,7 +32,10 @@ import (
 	"github.com/tikv/pd/server/schedule/placement"
 )
 
-const maxTargetRegionSize = 500
+const (
+	maxTargetRegionSize   = 500
+	maxTargetRegionFactor = 4
```
lhy1024, line 37: why use it?

tests/server/cluster/cluster_test.go
```diff
@@ -660,7 +660,7 @@ func (s *clusterTestSuite) TestLoadClusterInfo(c *C) {
 	tc.WaitLeader()
 	leaderServer := tc.GetServer(tc.GetLeader())
 	svr := leaderServer.GetServer()
-	rc := cluster.NewRaftCluster(s.ctx, svr.ClusterID(), syncer.NewRegionSyncer(svr), svr.GetClient(), svr.GetHTTPClient())
+	rc := cluster.NewRaftCluster(s.ctx, svr.ClusterID(), syncer.NewRegionSyncer(svr), svr.GetClient(), svr.GetHTTPClient(), svr.GetStoreConfigManager())
```
lhy1024, line 663: I think we need to add a test for modifying config

server/config/store_config.go
bufferflies: good idea.

tests/server/cluster/cluster_test.go
bufferflies: store manage expose the  `UpdateConfig` api.

pkg/typeutil/size_test.go
bufferflies: fix

server/config/store_config.go
bufferflies: yes , fixed

lhy1024 (approved):

PR #4015, opened 2021-08-20
metrics: replace delta with rate

bufferflies:
Signed-off-by: bufferflies <1045931706@qq.com>

### What problem does this PR solve?
scheduler some metrics using delta is not correct. 
[delta](https://prometheus.fuckcloudnative.io/di-san-zhang-prometheus/di-4-jie-cha-xun/functions#delta): return the last element - the first element in time zone ,so it could not express the ops 
[rate](https://prometheus.fuckcloudnative.io/di-san-zhang-prometheus/di-4-jie-cha-xun/functions#rate): return the average increase rate in the time zone

### What is changed and how it works?
replace delta with rate 
### Check List

Tests

Code changes

Side effects

Related changes

- Need to cherry-pick to the release branch

### Release note

```release-note
replace delta with rate in scheduler panel
```

PR #4708, opened 2022-03-04
metrics: convert all delta() usage to rate()

### LINKED ISSUE #4489: `delta()` should not be used in counter
## Bug Report

According to https://prometheus.io/docs/prometheus/latest/querying/functions/#delta: _"delta should only be used with gauges"_.
But we have some counters that use the delta function which may cause the wrong results. These metrics can be found in `pd.json`.

### What did you see instead?

Like previous issue: https://github.com/tikv/pd/issues/3561

Another one is the region heartbeat in the test cluster:
Screen Shot 2021-12-22 at 5 10 43 PM

master

BenMusch:
Signed-off-by: Ben Muschol <ben.muschol@airbnb.com>

`delta()` is used on counter metrics in grafana, though the [documentation](https://prometheus.io/docs/prometheus/latest/querying/functions/#delta) says not to do this:

> delta should only be used with gauges.

Issue Number: Close #4489

All usage of `delta()` has been replaced with [rate()](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate). A previous PR https://github.com/tikv/pd/pull/4015 did this for a subset of the metrics, this PR applies that change to the rest of the mtrics

### Check List

Tests

- Manual test ✅ (add detailed scripts or steps below)
- No code

Manually checked all referenced metrics and confirmed none were `gauge()`. Modified the grafana dashboards to confirm that the output was still as expected.

Code changes

Side effects

Related changes

### Release note

```release-note
Replaced all usage of delta() in the grafana dashboards with rate()
```


rleungx (approved):
LGTM

lhy1024 (approved):

metrics/grafana/pd.json
```diff
@@ -2240,7 +2240,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(delta(pd_schedule_operators_count{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=\"$instance\", event=\"create\"}[1m])) by (type)",
+              "expr": "sum(rate(pd_schedule_operators_count{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=\"$instance\", event=\"create\"}[1m])) by (type)",
```
nolouch, line 2243: The `yaxes`  is `opm`, but this formula changes it to `ops`.

BenMusch: great catch, changing the units to `ops` on the dashboards and multiplying the expressions for alerts by `60`

nolouch: I think some units should keep `opm`, because some parameters are converted into minutes, for example, we have a parameter of [ `store limit`](https://docs.pingcap.com/tidb/stable/configure-store-limit#view-setting-of-the-current-store). 
I will check it one by one, thanks.

rleungx:
PTAL @nolouch

metrics/grafana/pd.json
BenMusch: Oops, just saw this comment. Let me update the queries to be `opm` then

lhy1024 (commented):
Would you like to resolve conflict?

BenMusch:
Apologies, abandoning this PR on my end. Was working on this to onboard to TiDB at a previous job but no longer am working on TiDB so dont have the bandwidth for this

PR #4709, opened 2022-03-07
server: add evict leader logic for damaged regions when handling store heartbeat

hunterlxt:
Signed-off-by: Xintao <hunterlxt@live.com>

Issue Number: ref https://github.com/tikv/tikv/issues/10578

```commit-message
Add evict leader logic for damaged regions when handling store heartbeat
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- only logic changed

Side effects

- when this kind of errors happen, user need intervention manually with pd-ctl

Related changes

### Release note

```release-note
Add evict leader logic for damaged regions when handling store heartbeat
```


server/cluster/cluster.go
```diff
@@ -588,6 +589,8 @@ func (c *RaftCluster) HandleStoreHeartbeat(stats *pdpb.StoreStats) error {
 	}
 	// Here we will compare the reported regions with the previous hot peers to decide if it is still hot.
 	c.hotStat.CheckReadAsync(statistics.NewCollectUnReportedPeerTask(storeID, regions, interval))
+
+	c.handleDamagedStore(stats)
```
JmPotato, line 593: It seems that this function doesn't need to acquire the lock, right? I don't think putting it here is a good idea.

server/cluster/cluster.go
```diff
@@ -1781,3 +1784,42 @@ func newCacheCluster(c *RaftCluster) *cacheCluster {
...
+	// and need to be removed in the future
+	damaged_regions := stats.GetDamagedRegionsId()
+	if len(damaged_regions) == 0 {
+		return
+	}
+
+	log.Warn("store has damaged regions",
+		zap.Uint64("store-id", stats.GetStoreId()),
+		zap.Uint64s("region-ids", damaged_regions))
+	if exist, err := c.IsSchedulerExisted(schedulers.EvictLeaderName); !exist {
```
JmPotato, line 1801: Also, `c.IsSchedulerExisted` will acquire a read lock, if you put `handleDamagedStore` inside a lock itself, there will be a deadlock here.

JmPotato, line 1801: Another question is that if the `evict-leader-scheduler` exists, we may still need to update it because it can work on different stores. This is done by updating the scheduler config. You can check the implementation of the API to see a similar logic.

https://github.com/tikv/pd/blob/9bbd6df243b483ac998fb41448d6d71c7dc42eb8/server/api/scheduler.go#L257-L263

hunterlxt: yep, and i found `c.GetOperatorController` and `c.AddScheduler` also acquire a lock.Looks like just unlock before `c.handleDamagedStore(stats)`?

server/cluster/cluster.go
hunterlxt: commented below

server/cluster/cluster.go
```diff
@@ -1781,3 +1784,42 @@ func newCacheCluster(c *RaftCluster) *cacheCluster {
 		stores:      c.GetStores(),
 	}
 }
+
+// Evict the leaders when the store is damaged. Damaged regions are emergency errors
+// and requires user to manually remove the `evict-leader-scheduler` with pd-ctl
+func (c *RaftCluster) handleDamagedStore(stats *pdpb.StoreStats) {
+	// ToDO: regions have no special process for the time being
+	// and need to be removed in the future
+	damaged_regions := stats.GetDamagedRegionsId()
```
rleungx, line 1793: What is the definition of the damaged region? Is there any issue related to this PR?

server/grpc_service.go
```diff
@@ -616,6 +617,10 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
 		if err != nil {
 			return nil, status.Errorf(codes.Unknown, err.Error())
 		}
+		s.handleDamagedStore(request.Stats)
+		if err != nil {
```
rleungx, line 621: Where does the error come from?

server/util.go
```diff
@@ -157,3 +160,31 @@ func checkBootstrapRequest(clusterID uint64, req *pdpb.BootstrapRequest) error {
 
 	return nil
 }
+
+func postJSON(client *http.Client, url string, data []byte, checkOpts ...func([]byte, int)) error {
```
rleungx, line 164: `checkOpts` is used for testing purposes. Besides, I think we can avoid making an internal HTTP call.

hunterlxt: I don't want making an internal HTTP call either, but api/handler do this before. My original intention is to use the existing handler logic and avoid reimplementing the logic of add scheduler. But `addEvictOrGrant` post a request I don't know why such an internal post was done before

server/grpc_service.go
hunterlxt: fixed

server/util.go
hunterlxt: what's your suggestion?

server/cluster/cluster.go
hunterlxt: Finished by move `RedirectSchedulerUpdate` from`api/handler` to `handler`, which allows us to use `server.handler.RedirectSchedulerUpdate`

server/util.go
hunterlxt: @rleungx and I had a brief voice exchange: 1. Move `postJSON` to `apiutil` to avoid redundant code caused by circular dependencies but continue to use the existing interface 2. Reimplement add scheduler logic independent of handler logic.
Because the latter is too complicated and needs to maintain two sets of similar logic, for now use the existing handler interface. cc @JmPotato

hunterlxt:
I found PD Test / chunks (4) always failed and I exec `go test -timeout 30s -run ^TestAPIServer$ github.com/tikv/pd/server/api` locally based master branch, it failed again.

server/grpc_service.go
```diff
@@ -616,6 +617,10 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
 		if err != nil {
 			return nil, status.Errorf(codes.Unknown, err.Error())
 		}
+		err = s.handleDamagedStore(request.Stats)
```
JmPotato, line 620: I suggest change all `request.Status` to `request.GetStats()`.

server/grpc_service.go
```diff
@@ -1796,3 +1801,38 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
...
+	damagedRegions := stats.GetDamagedRegionsId()
+	if len(damagedRegions) == 0 {
+		return nil
+	}
+
+	log.Error("store damaged and leaders will be evicted, you might fix the store and remove evict-leader-scheduler manually",
+		zap.Uint64("store-id", stats.GetStoreId()),
+		zap.Uint64s("region-ids", damagedRegions))
+
+	h := s.GetHandler()
```
JmPotato, line 1819: Could you please help add a TODO comment here to tell that better not rely on an API calling internally? We could rewrite this part in an independent PR after mering this one.

server/grpc_service.go
```diff
@@ -1796,3 +1801,38 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
...
+// and requires user to manually remove the `evict-leader-scheduler` with pd-ctl
+func (s *GrpcServer) handleDamagedStore(stats *pdpb.StoreStats) error {
+	// ToDO: regions have no special process for the time being
+	// and need to be removed in the future
+	damagedRegions := stats.GetDamagedRegionsId()
+	if len(damagedRegions) == 0 {
+		return nil
+	}
+
+	log.Error("store damaged and leaders will be evicted, you might fix the store and remove evict-leader-scheduler manually",
```
JmPotato, line 1815: I think a warning is better?

server/grpc_service.go
```diff
@@ -1796,3 +1801,38 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
...
+		}
+		err = h.AddEvictLeaderScheduler(stats.GetStoreId())
+		if err != nil {
+			return err
+		}
+	} else {
+		if err := h.RedirectSchedulerUpdate(schedulers.EvictLeaderName, float64(stats.GetStoreId())); err != nil {
+			return err
+		}
+		log.Info("update scheduler",
```
JmPotato, line 1832: What about adding some info to help distinguish this part from the same API logs?

JmPotato (commented):
The rest LGTM.

server/grpc_service.go
hunterlxt: yeah in tikv-server it's absolutely an critical error, but i'm afraid when inexperienced users look at pd logs first, they couldn't notice the warnings (because too many warnings while running daily)

server/grpc_service.go
```diff
@@ -1796,3 +1801,40 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
...
+// and requires user to manually remove the `evict-leader-scheduler` with pd-ctl
+func (s *GrpcServer) handleDamagedStore(stats *pdpb.StoreStats) error {
+	// TODO: regions have no special process for the time being
+	// and need to be removed in the future
+	damagedRegions := stats.GetDamagedRegionsId()
+	if len(damagedRegions) == 0 {
+		return nil
+	}
+
+	log.Error("store damaged and leaders will be evicted, you might fix the store and remove evict-leader-scheduler manually",
```
rleungx, line 1815: Will it be printed for every store heartbeat?

server/grpc_service.go
```diff
@@ -1796,3 +1801,40 @@ func (s *GrpcServer) sendAllGlobalConfig(ctx context.Context, server pdpb.PD_Wat
...
+		}
+		err = h.AddEvictLeaderScheduler(stats.GetStoreId())
+		if err != nil {
+			return err
+		}
+	} else {
+		if err := h.RedirectSchedulerUpdate(schedulers.EvictLeaderName, float64(stats.GetStoreId())); err != nil {
+			return err
+		}
+		log.Info("update scheduler because of damaged regions",
```
rleungx, line 1833: ditto

server/grpc_service.go
hunterlxt: Preferably, my original intention is to warn users by all means possible to deal with this problem immediately. In the tikv part, there are critical error metric and log errors. If the user does not deal with that, damaged tikv will panic 24hours after the problem is first discovered.

server/grpc_service.go
hunterlxt: same as above

rleungx (approved):

AndreMouche (approved):
LGTM

JmPotato (approved):

PR #4712, opened 2022-03-09
client: observe the TSO cmd duration in all cases

JmPotato:

`cmdDurationWait` and `cmdDurationTSO` only record the cmd duration when the TSO request successfully, which is kind of misleading sometimes. For example, if a TSO error occurs, we can't see any differences from the TSO Wait Duration metrics unless you check the PD-side TSO handling duration or `cmdFailDurationTSO` specifically.

```commit-message
client: observe the TSO cmd duration in all cases
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1288,28 +1288,34 @@ type TSFuture interface {
 }
 
 func (req *tsoRequest) Wait() (physical int64, logical int64, err error) {
-	// If tso command duration is observed very high, the reason could be it
-	// takes too long for Wait() be called.
 	start := time.Now()
+	// `cmdDurationTSOAsyncWait` is useful to see how long a TSFuture takes
+	// until the Wait() function is called.
+	// It represents the duration between the TSO request and actually being used.
 	cmdDurationTSOAsyncWait.Observe(start.Sub(req.start).Seconds())
 	select {
 	case err = <-req.done:
 		err = errors.WithStack(err)
 		defer tsoReqPool.Put(req)
 		if err != nil {
+			// `cmdFailDurationTSO` is useful to see how long a TSO request takes until an error is met.
 			cmdFailDurationTSO.Observe(time.Since(req.start).Seconds())
-			return 0, 0, err
+		} else {
+			physical, logical = req.physical, req.logical
 		}
-		physical, logical = req.physical, req.logical
-		now := time.Now()
-		cmdDurationWait.Observe(now.Sub(start).Seconds())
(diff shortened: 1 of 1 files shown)


CabinfeverB (approved):
LGTM

client/client.go
```diff
@@ -1288,28 +1288,34 @@ type TSFuture interface {
...
+		err = errors.WithStack(req.requestCtx.Err())
 	case <-req.clientCtx.Done():
-		return 0, 0, errors.WithStack(req.clientCtx.Err())
-	}
+		err = errors.WithStack(req.clientCtx.Err())
+	}
+	now := time.Now()
+	// `cmdDurationWait` is useful to see how long it takes to Wait() a TSO result.
+	// It partially or fully represents the client-side pressure.
+	cmdDurationWait.Observe(now.Sub(start).Seconds())
```
rleungx, line 1314: How do we distinguish between success and failed requests in the metrics?

client/client.go
```diff
@@ -1288,28 +1288,34 @@ type TSFuture interface {
...
-		return
 	case <-req.requestCtx.Done():
-		return 0, 0, errors.WithStack(req.requestCtx.Err())
+		err = errors.WithStack(req.requestCtx.Err())
 	case <-req.clientCtx.Done():
-		return 0, 0, errors.WithStack(req.clientCtx.Err())
-	}
+		err = errors.WithStack(req.clientCtx.Err())
+	}
+	now := time.Now()
```
AndreMouche, line 1311: How about putting L1311-1317 in `defer`?

client/client.go
JmPotato: I think the problem here is that the old cmdDurationWait only shows the part of the durations, i.e, the successful part. This PR extends these metrics to all TSO durations. IMO, distinguishing success or not is not that important even unnecessary for this one. It just shows the whole picture as a TSO Wait indicator.

client/client.go
JmPotato: It can be, but I don't think this change makes any difference.

AndreMouche: But that's why `defer` exist in `golang` ? For this case we need to run these logic at the end of the function,  I think it is more friendly for the reviewer and maintainers if we use `defer`.

JmPotato: IMO, `defer` is useful when the function is complex, e.g has a long body or multiple branches. It can help us to release some resources eventually without forgetting. However, this function is simple enough, so I think the `defer` is optional.

PR #4716, opened 2022-03-09
dr-ausosync: report min resolved ts

lhy1024:

Issue Number: ref https://github.com/tikv/pd/issues/4686

### Check List

Tests

- Unit test
- Integration test

image

### Release note

```release-note
dr-ausosync: report min resolved ts 
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -124,6 +124,8 @@ type Client interface {
 	StoreGlobalConfig(ctx context.Context, items []GlobalConfigItem) error
 	// WatchGlobalConfig returns an stream with all global config and updates
 	WatchGlobalConfig(ctx context.Context) (chan []GlobalConfigItem, error)
+	// ReportMinResolvedTS reports the min resolved ts to pd.
+	ReportMinResolvedTS(ctx context.Context, storeID, minResolvedTS uint64) error
 	// UpdateOption updates the client option.
 	UpdateOption(option DynamicOption, value interface{}) error
 	// Close closes the client.
@@ -1864,4 +1866,25 @@ func (c *client) WatchGlobalConfig(ctx context.Context) (chan []GlobalConfigItem
 		}
 	}()
 	return globalConfigWatcherCh, err
+}
+
+func (c *client) ReportMinResolvedTS(ctx context.Context, storeID, minResolvedTS uint64) error {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span = opentracing.StartSpan("pdclient.ReportMinResolvedTS", opentracing.ChildOf(span.Context()))
+		defer span.Finish()
+	}
+	ctx, cancel := context.WithTimeout(ctx, c.option.timeout)
+	req := &pdpb.ReportMinResolvedTsRequest{
+		Header:        c.requestHeader(),
+		StoreId:       storeID,
+		MinResolvedTs: minResolvedTS,
+	}
+	ctx = grpcutil.BuildForwardContext(ctx, c.GetLeaderAddr())
+	_, err := c.getClient().ReportMinResolvedTS(ctx, req)
+	cancel()
+
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	return nil
 }
--- a/client/client.go
+++ b/client/client.go
@@ -124,8 +124,6 @@ type Client interface {
 	StoreGlobalConfig(ctx context.Context, items []GlobalConfigItem) error
 	// WatchGlobalConfig returns an stream with all global config and updates
 	WatchGlobalConfig(ctx context.Context) (chan []GlobalConfigItem, error)
-	// ReportMinResolvedTS reports the min resolved ts to pd.
-	ReportMinResolvedTS(ctx context.Context, storeID, minResolvedTS uint64) error
 	// UpdateOption updates the client option.
 	UpdateOption(option DynamicOption, value interface{}) error
 	// Close closes the client.
@@ -1866,25 +1864,4 @@ func (c *client) WatchGlobalConfig(ctx context.Context) (chan []GlobalConfigItem
 		}
 	}()
 	return globalConfigWatcherCh, err
-}
-
-func (c *client) ReportMinResolvedTS(ctx context.Context, storeID, minResolvedTS uint64) error {
-	if span := opentracing.SpanFromContext(ctx); span != nil {
-		span = opentracing.StartSpan("pdclient.ReportMinResolvedTS", opentracing.ChildOf(span.Context()))
-		defer span.Finish()
-	}
-	ctx, cancel := context.WithTimeout(ctx, c.option.timeout)
-	req := &pdpb.ReportMinResolvedTsRequest{
-		Header:        c.requestHeader(),
-		StoreId:       storeID,
-		MinResolvedTs: minResolvedTS,
-	}
-	ctx = grpcutil.BuildForwardContext(ctx, c.GetLeaderAddr())
-	_, err := c.getClient().ReportMinResolvedTS(ctx, req)
-	cancel()
-
-	if err != nil {
-		return errors.WithStack(err)
-	}
-	return nil
 }
--- a/server/api/min_resolved_ts.go
+++ b/server/api/min_resolved_ts.go
@@ -1,0 +1,66 @@
+// Copyright 2022 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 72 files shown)


client/client.go
```diff
@@ -1865,3 +1867,24 @@ func (c *client) WatchGlobalConfig(ctx context.Context) (chan []GlobalConfigItem
 	}()
 	return globalConfigWatcherCh, err
 }
+
+func (c *client) ReportMinResolvedTS(ctx context.Context, storeID, minResolvedTS uint64) error {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
```
disksing, line 1872: maybe we can panic directly? I think only TiKV can use it.

server/grpc_service.go
```diff
@@ -1819,3 +1819,40 @@ func (s *GrpcServer) handleDamagedStore(stats *pdpb.StoreStats) error {
...
+	rc := s.GetRaftCluster()
+	if rc == nil {
+		return &pdpb.ReportMinResolvedTsResponse{Header: s.notBootstrappedHeader()}, nil
+	}
+
+	var storage endpoint.MinResolvedTSStorage = s.storage
+	storeID := request.StoreId
+	minResolvedTS := request.MinResolvedTs
+
+	if err := storage.SaveMinResolvedTS(storeID, minResolvedTS); err != nil {
```
disksing, line 1848: If tikv sync resolved_ts every 1s, will it cause too many etcd writings and lead to performance issues?

client/client.go
lhy1024: how about removing it?

server/cluster/cluster.go
```diff
@@ -53,6 +54,7 @@ import (
 )
 
 var backgroundJobInterval = 10 * time.Second
+var saveMinResolvedTSInterval = 1 * time.Second
```
lhy1024, line 57: Do we need to make it configurable?

lhy1024: cc @disksing @nolouch

disksing: I think we need a configuration. maybe we should make it disabled by default.

server/api/min_resolved_ts.go
```diff
@@ -0,0 +1,58 @@
...
+
+func newMinResolvedTSHandler(svr *server.Server, rd *render.Render) *minResolvedTSHandler {
+	return &minResolvedTSHandler{
+		svr: svr,
+		rd:  rd,
+	}
+}
+
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
+type listMinResolvedTS struct {
```
disksing, line 37: Should not name it `list` any more because it only returns a single ts.

server/api/min_resolved_ts.go
```diff
@@ -0,0 +1,58 @@
...
+	MinResolvedTS uint64 `json:"min_resolved_ts"`
+}
+
+// @Tags minresolvedts
+// @Summary Get min resolved ts.
+// @Produce json
+// @Success 200 {array} listMinResolvedTS
+// @Failure 500 {string} string "PD server failed to proceed the request."
+// @Router /min-resolved-ts [get]
+func (h *minResolvedTSHandler) List(w http.ResponseWriter, r *http.Request) {
```
nolouch, line 47: Do you want list all store`resolved ts` or get cluster level min `resolved ts`

server/cluster/cluster.go
```diff
@@ -1650,6 +1654,64 @@ func (c *RaftCluster) RemoveStoreLimit(storeID uint64) {
...
+	store := c.GetStore(storeID)
+	if store == nil {
+		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
+	}
+
+	newStore := store.Clone(
+		core.SetMinResolvedTS(minResolvedTS),
+	)
+
+	return c.putStoreLocked(newStore)
```
nolouch, line 1690: I think this has unnecessary writes to etcd.

server/storage/endpoint/min_resolved_ts.go
```diff
@@ -0,0 +1,56 @@
...
+package endpoint
+
+import (
+	"strconv"
+)
+
+// MinResolvedTSPoint is the min resolved ts for a store
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
+type MinResolvedTSPoint struct {
+	StoreID       uint64 `json:"store_id"`
```
nolouch, line 24: Why need store ID？

nolouch (commented):
this resloved ts is cluster-level, do we need the data center-lever?  such as, Primary DC scope resolved ts and DR DC scop resolved ts. cc @disksing

server/cluster/cluster.go
lhy1024: ok

server/storage/endpoint/min_resolved_ts.go
lhy1024: It's a retain code, I'll fix it

disksing:
> this resloved ts is cluster-level, do we need the data center-level? such as, Primary DC scope resolved ts and DR DC scop resolved ts. cc @disksing

It should be cluster scope.

server/cluster/cluster.go
lhy1024: cc @disksing

AndreMouche (commented):
1. Could or should we store the `min_resolve_ts` once we receive the request instead store it in goroutine?
2. the path `min_resolved_ts` is too big, maybe save it to a sub path?

lhy1024:
> 1. Could or should we store the `min_resolve_ts` once we receive the request instead store it in goroutine?
> 2. the path `min_resolved_ts` is too big, maybe save it to a sub path?

1. if there are many stores, it could cause a performance problem. https://github.com/tikv/pd/pull/4716#discussion_r823418531

3. do you have a good idea about the path?

server/storage/endpoint/min_resolved_ts.go
```diff
@@ -0,0 +1,55 @@
...
+// LoadMinResolvedTS loads the min resolved ts from storage.
+func (se *StorageEndpoint) LoadMinResolvedTS() (uint64, error) {
+	value, err := se.Load(MinResolvedTSPath())
+	if err != nil {
+		return 0, err
+	}
+	if value == "" {
+		return 0, nil
+	}
+	minResolvedTS, err := strconv.ParseUint(value, 16, 64)
```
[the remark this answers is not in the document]
lhy1024: we need to wrap err

server/cluster/cluster.go
disksing: It should be ok to write to etcd. If not write to etcd, it will be lost when PD is shut down or PD leader is changed.

disksing (approved):

server/config/config.go
```diff
@@ -1102,6 +1103,8 @@ type PDServerConfig struct {
 	TraceRegionFlow bool `toml:"trace-region-flow" json:"trace-region-flow,string,omitempty"`
 	// FlowRoundByDigit used to discretization processing flow information.
 	FlowRoundByDigit int `toml:"flow-round-by-digit" json:"flow-round-by-digit"`
+	// SaveMinResolvedTSInterval is the interval to save the min resolved ts.
+	SaveMinResolvedTSInterval typeutil.Duration `toml:"save-min-resolved-ts-interval" json:"save-min-resolved-ts-interval"`
```
nolouch, line 1107: consider a better name

server/cluster/cluster.go
```diff
@@ -1650,6 +1654,66 @@ func (c *RaftCluster) RemoveStoreLimit(storeID uint64) {
...
+// GetMinResolvedTS returns the min resolved ts of all stores.
+func (c *RaftCluster) GetMinResolvedTS() uint64 {
+	c.RLock()
+	defer c.RUnlock()
+	if !c.isInitialized() {
+		return math.MaxUint64
+	}
+	min := uint64(math.MaxUint64)
+	for _, s := range c.GetStores() {
+		if !core.IsAvailableForMinResolvedTS(s) {
```
lhy1024, line 1666: Do we need to set a ttl for min resolved-ts reported? It will solve a store with leader but not report min resolved-ts for a long time @disksing

disksing: If tikv has bug or wrong config and it does not report `min-resolved-ts`, I guess the global min-resolved-ts should not advance, or the result could be wrong.

nolouch (approved):

PR #4718, opened 2022-03-10
*: use topology weight to estimate preparing transition

rleungx:

Issue Number: Ref #4640

This PR supports calculating a threshold for each store to change the node state from `Preparing` to `Serving`. The default threshold is 90% of the final store size. 

If there are some placement rules, it will calculate the store size range by range and accumulate the size. Once it reaches the `max-store-preparing-time` or the threshold, the state will be changed.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Support changing the node state from `Preparing` to `Serving`
```


server/cluster/cluster.go
```diff
@@ -1176,6 +1181,34 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
 	return c.putStoreLocked(newStore)
 }
 
+// ReadyToServe change store's node state to Serving.
+func (c *RaftCluster) ReadyToServe(store *core.StoreInfo) error {
+	c.Lock()
+	defer c.Unlock()
+	storeID := store.GetID()
+	if store == nil {
```
AndreMouche, line 1189: I think it should be panic on L1188 if store is nil.

server/cluster/cluster.go
```diff
@@ -1176,6 +1181,34 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
 	return c.putStoreLocked(newStore)
 }
 
+// ReadyToServe change store's node state to Serving.
+func (c *RaftCluster) ReadyToServe(store *core.StoreInfo) error {
+	c.Lock()
```
AndreMouche, line 1186: What are we try to protect before L1208?

CabinfeverB: > What are we try to protect before L1208?

Maybe to protect the store from a state change

server/cluster/cluster.go
```diff
@@ -1252,6 +1300,115 @@ func (c *RaftCluster) checkStores() {
...
+			}
+		}
+	}
+
+	return weight / count
+}
+
+func buildTopology(s *core.StoreInfo, stores []*core.StoreInfo, locationLabels []string) (map[string]interface{}, float64) {
+	topology := make(map[string]interface{})
+	storeCount := 1.0
```
CabinfeverB, line 1369: It seems that it returns as least 2 because stores includes `s`

server/cluster/cluster.go
```diff
@@ -1252,6 +1300,115 @@ func (c *RaftCluster) checkStores() {
...
+			zap.String("rule", rule.String()),
+			zap.Int64("region-size", regionSize),
+			zap.Float64("weight", weight),
+			zap.Float64("store-size", storeSize),
+		)
+	}
+	return storeSize
+}
+
+func getStoreTopoWeight(store *core.StoreInfo, stores []*core.StoreInfo, locationLabels []string) float64 {
```
CabinfeverB, line 1349: If the isolation level does not meet the number of replica, there seems to be some offset.

server/cluster/cluster.go
```diff
@@ -1213,12 +1248,25 @@ func (c *RaftCluster) checkStores() {
...
+	keys := c.ruleManager.GetSplitKeys([]byte(""), []byte(""))
 	for _, store := range stores {
 		// the store has already been tombstone
 		if store.IsRemoved() {
 			continue
 		}
 
+		storeID := store.GetID()
+		if store.IsPreparing() {
+			if store.GetUptime() > c.opt.GetMaxStorePreparingTime() || float64(store.GetRegionSize()) >= c.getThreshold(stores, store, keys) {
```
CabinfeverB, line 1261: What happen if there are no rule？

CabinfeverB: It will return 0 because of `calculateRange` returns 0

server/cluster/cluster.go
AndreMouche: However, I think it does not work.

rleungx: I modified the function implementation to keep consistent with others, PTAL.

server/cluster/cluster.go
```diff
@@ -1244,12 +1274,24 @@ func (c *RaftCluster) checkStores() {
...
 	stores := c.GetStores()
+
 	for _, store := range stores {
 		// the store has already been tombstone
 		if store.IsRemoved() {
 			continue
 		}
 
+		storeID := store.GetID()
+		if store.IsPreparing() {
```
bufferflies, line 1285:
```suggestion
                // it can  decrease the nested judgment.
		isReady:=store.IsPreparing()&&(store.GetUptime() > c.opt.GetMaxStorePreparingTime() || float64(store.GetRegionSize()) >= c.getThreshold(stores, store))
		if err := c.ReadyToServe(storeID);isReady&&err!=nil{
		}
```

server/cluster/cluster.go
```diff
@@ -270,12 +274,10 @@ func (c *RaftCluster) Start(s Server) error {
...
 
-	c.wg.Add(6)
+	c.wg.Add(7)
 	go c.runCoordinator()
-	failpoint.Inject("highFrequencyClusterJobs", func() {
-		backgroundJobInterval = 1 * time.Microsecond
-	})
-	go c.runBackgroundJobs(backgroundJobInterval)
+	go c.runMetricsCollectionJob()
+	go c.runNodeStateCheckJob()
```
JmPotato, line 280: No action is strongly needed in this PR. But I'm thinking that since the function structures here are so similar that we could reuse these codes to provide a common interface to run these daemons and no need to modify this number manually every time.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,159 @@ func (c *RaftCluster) checkStores() {
...
-	c.progressManager.UpdateProgress(progress, left)
+	c.progressManager.UpdateProgressLeft(progress, left)
 	process, ls, _ := c.progressManager.Status(progress)
 	storesProgressGauge.WithLabelValues(storeAddress, storeLabel, action).Set(process)
 	storesETAGauge.WithLabelValues(storeAddress, storeLabel, action).Set(ls)
 }
 
-func (c *RaftCluster) resetRemovingProgress(storeID uint64, storeAddress string, action string) {
+func (c *RaftCluster) resetProgress(storeID uint64, storeAddress string, action string) {
 	storeLabel := fmt.Sprintf("%d", storeID)
```
CabinfeverB, line 1572: Maybe `strconv.FormatUint(storeID, 10)` is better

server/cluster/cluster.go
```diff
@@ -1192,7 +1218,8 @@ func (c *RaftCluster) BuryStore(storeID uint64, forceBury bool) error {
 		// clean up the residual information.
 		delete(c.prevStoreLimit, storeID)
 		c.RemoveStoreLimit(storeID)
-		c.resetRemovingProgress(storeID, store.GetAddress(), removingAction)
+		c.resetProgress(storeID, store.GetAddress(), removingAction)
+		c.resetProgress(storeID, store.GetAddress(), preparingAction)
```
CabinfeverB, line 1222: It seems that currently we don't call `AddProgress` for preparingAction?

rleungx: It will be called in `checkStores`

pkg/progress/progress.go
```diff
@@ -110,7 +126,7 @@ func (m *Manager) Status(progress string) (process, leftSeconds, currentSpeed fl
 	defer m.RUnlock()
 	if p, exist := m.progesses[progress]; exist {
 		process = 1 - p.left/p.total
-		speedPerSec := (p.total - p.left) / time.Since(p.startTime).Seconds()
+		speedPerSec := (p.last - p.left) / time.Since(p.lastTime).Seconds()
```
CabinfeverB, line 129: I think this approach is unreliable and has a high interpretation cost. Because `Status()` will return zero speed and infinite remaining time after an `UpdateProgressLeft()` operation whose `last` is equal to `left` or an initialization. It would be unwise to expose this inaccurate result to the user.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,159 @@ func (c *RaftCluster) checkStores() {
...
+
+	return topology, storeCount
+}
+
+func getSortedLabels(storeLabels []*metapb.StoreLabel, locationLabels []string) []*metapb.StoreLabel {
+	var sortedLabels []*metapb.StoreLabel
+	for _, ll := range locationLabels {
+		for _, sl := range storeLabels {
+			if ll == sl.Key {
+				sortedLabels = append(sortedLabels, sl)
```
AndreMouche, line 1531: Should we break if we find the current `ll`? Or Is it possible that the same labels would exist in the `storeLabels`?

server/statistics/store_collection.go
```diff
@@ -43,6 +43,10 @@ type storeStatistics struct {
 	RegionCount     int
 	LeaderCount     int
 	LabelCounter    map[string]int
+	Preparing       int
+	Serving         int
+	Removing        int
+	Removed         int
```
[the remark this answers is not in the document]
rleungx: The `nodeStore` is going to replace the `storeState`.

tests/server/cluster/cluster_test.go
```diff
@@ -798,10 +798,11 @@ func (s *clusterTestSuite) TestTiFlashWithPlacementRules(c *C) {
 	rep.EnablePlacementRules = false
 	err = svr.SetReplicationConfig(rep)
 	c.Assert(err, NotNil)
-	err = svr.GetRaftCluster().RemoveStore(11, true)
+	err = svr.GetRaftCluster().BuryStore(11, true)
```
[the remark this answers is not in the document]
rleungx: This is a bug introduced by another PR that makes the test unstable.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,160 @@ func (c *RaftCluster) checkStores() {
...
+func (c *RaftCluster) calculateRange(stores []*core.StoreInfo, store *core.StoreInfo, startKey, endKey []byte) float64 {
+	var storeSize float64
+	rules := c.ruleManager.GetRulesForApplyRange(startKey, endKey)
+	for _, rule := range rules {
+		if !placement.MatchLabelConstraints(store, rule.LabelConstraints) {
+			continue
+		}
+
+		var matchStores []*core.StoreInfo
+		for _, store := range stores {
```
AndreMouche, line 1464: Please rename the name `store` which shares the same name with the arguments of the current function.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,160 @@ func (c *RaftCluster) checkStores() {
...
+	}
+
+	return weight / count
+}
+
+func buildTopology(s *core.StoreInfo, stores []*core.StoreInfo, locationLabels []string) (map[string]interface{}, float64) {
+	topology := make(map[string]interface{})
+	storeCount := 1.0
+	for _, store := range stores {
+		if store.IsServing() || store.IsPreparing() {
```
AndreMouche, line 1510: According the previous logic, this condition will always be true.

AndreMouche (commented):
It is too complex to compute the `threshold` of a store, How about wrap an struct to compute it?

server/cluster/cluster.go
rleungx: There is more than one place to call this function and not all of them can ensure it.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,160 @@ func (c *RaftCluster) checkStores() {
...
+	topology, count := buildTopology(store, stores, locationLabels)
+	weight := 1.0
+	topo := topology
+	storeLabels := getSortedLabels(store.GetLabels(), locationLabels)
+	for _, label := range storeLabels {
+		if _, ok := topo[label.Value]; ok {
+			weight /= float64(len(topo))
+			topo, ok = topo[label.Value].(map[string]interface{})
+			if !ok {
+				return weight / count
```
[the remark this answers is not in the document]
rleungx: the last level could be false

server/cluster/cluster_test.go
```diff
@@ -1131,6 +1153,186 @@ func (s *testClusterInfoSuite) TestUpdateStorePendingPeerCount(c *C) {
...
+	cluster.ruleManager.DeleteRule("pd", "default")
+
+	regions := newTestRegions(100, 10, 5)
+	for _, region := range regions {
+		c.Assert(cluster.putRegion(region), IsNil)
+	}
+
+	stores := cluster.GetStores()
+	store := cluster.GetStore(1)
+	// 100 * 100 * 5 (total region size) / 5 * 2 (zone) / 4 (host) * 0.9 = 4500
```
CabinfeverB, line 1254:
```suggestion
	// 100 * 100 * 2 (placement rule) / 4 (host) * 0.9 = 4500
```

CabinfeverB (approved):

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,160 @@ func (c *RaftCluster) checkStores() {
...
+	for _, store := range stores {
+		if store.IsServing() || store.IsPreparing() {
+			updateTopology(topology, getSortedLabels(store.GetLabels(), locationLabels))
+		}
+
+		if store.GetID() == s.GetID() {
+			continue
+		}
+
+		if s.CompareLocation(store, locationLabels) == -1 {
```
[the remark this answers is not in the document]
rleungx: It might be complex to decide if we need to update the `topology`.

server/cluster/cluster.go
```diff
@@ -1342,22 +1426,161 @@ func (c *RaftCluster) checkStores() {
...
+	var sortedLabels []*metapb.StoreLabel
+	for _, ll := range locationLabels {
+		for _, sl := range storeLabels {
+			if ll == sl.Key {
+				sortedLabels = append(sortedLabels, sl)
+				break
+			}
+		}
+		// TODO: we need to improve this logic to make the label calculation more accurate if the user has the wrong label settings.
+		sortedLabels = append(sortedLabels, &metapb.StoreLabel{Key: ll, Value: ""})
```
AndreMouche, line 1535: e , I think the one who has the `ll` label do not need to append the empty label again.

server/cluster/cluster.go
```diff
@@ -1259,7 +1287,40 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
...
 	err := c.putStoreLocked(newStore)
 	if err == nil {
-		c.resetRemovingProgress(storeID, store.GetAddress(), removingAction)
+		c.resetProgress(storeID, store.GetAddress(), removingAction)
+	}
+	return err
+}
+
+// ReadyToServe change store's node state to Serving.
+func (c *RaftCluster) ReadyToServe(storeID uint64) error {
```
AndreMouche, line 1296: Could you please merge it with the function `UpStore`, it seems the logic is similar except for the `options`?

server/cluster/cluster.go
```diff
@@ -1301,12 +1362,35 @@ func (c *RaftCluster) checkStores() {
...
+			regionSize := float64(store.GetRegionSize())
+			if store.GetUptime() > c.opt.GetMaxStorePreparingTime() || regionSize >= threshold {
+				if err := c.ReadyToServe(storeID); err != nil {
+					log.Error("change store to serving failed",
+						zap.Stringer("store", store.GetMeta()),
+						errs.ZapError(err))
+				}
+			} else {
+				remaining := threshold - regionSize
+				if remaining < 0 {
```
AndreMouche, line 1385: It will never happen since it will meet the condition on L1377 first.

AndreMouche (commented):
Could we split this PR ? it is too big for me，1 of them may be: checks  the preparing store and up the store if needed 😄

server/cluster/cluster.go
rleungx: These two functions are a little bit different, the common part is only checking the store state.

AndreMouche (commented):
LGTM

rleungx:
PTAL @JmPotato @bufferflies

JmPotato (approved):

PR #4731, opened 2022-03-16
*: update dashboard 

Yisaer:

Issue Number: ref https://github.com/pingcap/tidb/issues/29661

Update dashboard in order to use optimizer trace feature

iqFjVKo91Z

```commit-message
```

### Release note

```release-note
None
```


breeswish:
@Yisaer I think we can drive this PR instead https://github.com/tikv/pd/pull/4740

Issue #4399, opened 2021-11-29
Meta issue


lhy1024:
## Development Task

This issue aims to provide an anchor for links to those misc changes, and it is important to notice that if your PR has sufficient descriptions, it is a better option to create a separate issue.

PR #4736, opened 2022-03-16
server: add persist file log

disksing:

Issue Number: ref #4399

add a log for persisting file.

### Release note

```release-note
None
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1464,6 +1464,7 @@ func (s *Server) ReplicateFileToMember(ctx context.Context, member *pdpb.Member,
 
 // PersistFile saves a file in DataDir.
 func (s *Server) PersistFile(name string, data []byte) error {
+	log.Info("persist file", zap.String("name", name), zap.Binary("data", data))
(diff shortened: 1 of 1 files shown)


lhy1024 (approved):

AndreMouche (approved):
LGTM

server/server.go
```diff
@@ -1464,6 +1464,7 @@ func (s *Server) ReplicateFileToMember(ctx context.Context, member *pdpb.Member,
 
 // PersistFile saves a file in DataDir.
 func (s *Server) PersistFile(name string, data []byte) error {
+	log.Info("persist file", zap.String("name", name), zap.Binary("data", data))
```
rleungx, line 1467: Is it readable?

disksing: Not readable. It will use base64. Because the API supports write arbitrary data, binary should be a good choice.

nolouch: why need to log the data？may be an md5 or checksum?

disksing: It should not be very long. Sometime if we really want to debug, we can decode and check the content.

nolouch (approved):

PR #4743, opened 2022-03-17
core: add buckets property to regionInfo struct

bufferflies:

This is the first step for the final storage of buckets in memery. 
Some api and storage depences on it.
Issue Number: Relate #4668

```commit-message
```

### Check List

Tests

Code changes

Side effects

Related changes

### Release note

```release-note
 None.
```

### DIFF
--- a/server/core/region.go
+++ b/server/core/region.go
@@ -58,6 +58,7 @@ type RegionInfo struct {
 	replicationStatus *replication_modepb.RegionReplicationStatus
 	QueryStats        *pdpb.QueryStats
 	flowRoundDivisor  uint64
+	buckets           *metapb.Buckets
 }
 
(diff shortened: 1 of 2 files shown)


server/core/region.go
```diff
@@ -214,6 +215,14 @@ func (r *RegionInfo) Clone(opts ...RegionCreateOption) *RegionInfo {
 	return region
 }
 
+// GetBuckets returns the buckets of the region.
+func (r *RegionInfo) GetBuckets() *metapb.Buckets {
+	if r == nil {
```
rleungx, line 220: When will it happen?

bufferflies: this is the first step,  sync and some api will use it, I will split them into different pr.

rleungx: I mean `r == nil`

PR #4744, opened 2022-03-17
dr-ausosync: ensure cluster min resolved ts increase

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/4686

This PR will ensure cluster min resolved ts increase, it will compare the real-time one with the old one,  and only save when the old one is smaller.

This PR will make  min resolved ts persistent interval dynamic configurable 

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/api/min_resolved_ts.go
+++ b/server/api/min_resolved_ts.go
@@ -17,6 +17,7 @@ package api
 import (
 	"net/http"
 
+	"github.com/tikv/pd/pkg/typeutil"
 	"github.com/tikv/pd/server"
 	"github.com/unrolled/render"
 )
@@ -35,7 +36,9 @@ func newMinResolvedTSHandler(svr *server.Server, rd *render.Render) *minResolved
 
 // NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
 type minResolvedTS struct {
-	MinResolvedTS uint64 `json:"min_resolved_ts"`
+	IsRealTime      bool              `json:"is_real_time,omitempty"`
+	MinResolvedTS   uint64            `json:"min_resolved_ts"`
+	PersistInterval typeutil.Duration `json:"persist_interval,omitempty"`
 }
 
 // @Tags minresolvedts
@@ -45,13 +48,12 @@ type minResolvedTS struct {
 // @Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /min-resolved-ts [get]
 func (h *minResolvedTSHandler) GetMinResolvedTS(w http.ResponseWriter, r *http.Request) {
-	storage := h.svr.GetStorage()
-	value, err := storage.LoadMinResolvedTS()
-	if err != nil {
-		h.rd.JSON(w, http.StatusInternalServerError, err.Error())
-		return
-	}
+	c := h.svr.GetRaftCluster()
+	value := c.GetMinResolvedTS()
+	persistInterval := c.GetOpts().GetPDServerConfig().MinResolvedTSPersistenceInterval
 	h.rd.JSON(w, http.StatusOK, minResolvedTS{
-		MinResolvedTS: value,
+		MinResolvedTS:   value,
+		PersistInterval: persistInterval,
+		IsRealTime:      persistInterval.Duration != 0,
 	})
 }
--- a/server/api/min_resolved_ts.go
+++ b/server/api/min_resolved_ts.go
@@ -44,7 +44,7 @@ type minResolvedTS struct {
 // @Success 200 {array} minResolvedTS
 // @Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /min-resolved-ts [get]
-func (h *minResolvedTSHandler) Get(w http.ResponseWriter, r *http.Request) {
+func (h *minResolvedTSHandler) GetMinResolvedTS(w http.ResponseWriter, r *http.Request) {
 	storage := h.svr.GetStorage()
 	value, err := storage.LoadMinResolvedTS()
 	if err != nil {
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -97,12 +97,13 @@ type RaftCluster struct {
 	clusterID uint64
 
 	// cached cluster info
-	core    *core.BasicCluster
-	meta    *metapb.Cluster
-	opt     *config.PersistOptions
-	storage storage.Storage
-	id      id.Allocator
-	limiter *StoreLimiter
+	core          *core.BasicCluster
+	meta          *metapb.Cluster
+	opt           *config.PersistOptions
+	storage       storage.Storage
+	id            id.Allocator
+	limiter       *StoreLimiter
+	minResolvedTS uint64
 
 	changedRegions chan *core.RegionInfo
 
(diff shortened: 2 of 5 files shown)


tests/server/cluster/cluster_test.go
```diff
@@ -1136,11 +1137,26 @@ func (s *clusterTestSuite) putRegionWithLeader(c *C, rc *cluster.RaftCluster, id
...
 
+func (s *clusterTestSuite) checkMinResolvedTSFromStorage(c *C, rc *cluster.RaftCluster, expect uint64) {
+	time.Sleep(time.Millisecond * 10)
+	ts2, err := rc.GetStorage().LoadMinResolvedTS()
+	c.Assert(err, IsNil)
+	c.Assert(ts2, Equals, expect)
+}
+
+func (s *clusterTestSuite) setMinResolvedTSPersistenceInterval(c *C, rc *cluster.RaftCluster, svr *server.Server, interval time.Duration) {
+	cfg := rc.GetOpts().GetPDServerConfig()
```
rleungx, line 1148: need clone

lhy1024: thanks

server/cluster/cluster.go
```diff
@@ -1690,30 +1693,59 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
...
 		case <-ticker.C:
-			minResolvedTS := c.GetMinResolvedTS()
-			if minResolvedTS != math.MaxUint64 {
-				c.Lock()
-				c.storage.SaveMinResolvedTS(minResolvedTS)
-				c.Unlock()
+			interval = c.opt.GetMinResolvedTSPersistenceInterval()
+			if interval != 0 {
+				minResolvedTSRealtime := c.GetMinResolvedTS()
+				if c.isLarger(minResolvedTSRealtime) {
```
AndreMouche, line 1723: It seems the function `GetMinResolvedTS` and `isLarger` are used together and only here, could we merge these two functions into one, so we just need to c.RLock&RUnlock 1 time? 
Also， I would like to move the L1722 - L1727 into one function for the same reason, there are too many `lock` & `unlock`, maybe we could make the logic easier(reduce a bit performance but friendly to our maintainers), how about calling the following function instead(as an example)?

```
func (c *RaftCluster) checkAndUpdateMinResovedTS() {
	c.Lock()
	defer c.Unlock()

	if !c.isInitialized() {
		return
	}
	curMinResolvedTS := uint64(math.MaxUint64)
	for _, s := range c.GetStores() {
		if !core.IsAvailableForMinResolvedTS(s) {
			continue
		}
		if curMinResolvedTS > s.GetMinResolvedTS() {
			curMinResolvedTS = s.GetMinResolvedTS()
		}
	}
	if curMinResolvedTS == math.MaxUint64 || curMinResolvedTS <= c.minResolvedTS {
		return
	}
	c.minResolvedTS = curMinResolvedTS
	c.storage.SaveMinResolvedTS(curMinResolvedTS)
}
```

server/cluster/cluster.go
```diff
@@ -1690,30 +1693,59 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
...
+					c.Lock()
+					c.minResolvedTS = minResolvedTSRealtime
+					c.storage.SaveMinResolvedTS(minResolvedTSRealtime)
+					c.Unlock()
+				}
+			} else {
+				interval = DefaultMinResolvedTSPersistenceInterval
```
AndreMouche, line 1730: So when the interval is set to zero we will skip the check?

AndreMouche (commented):
Rest LGTM

server/api/min_resolved_ts.go
```diff
@@ -44,7 +44,7 @@ type minResolvedTS struct {
 // @Success 200 {array} minResolvedTS
 // @Failure 500 {string} string "PD server failed to proceed the request."
 // @Router /min-resolved-ts [get]
-func (h *minResolvedTSHandler) Get(w http.ResponseWriter, r *http.Request) {
+func (h *minResolvedTSHandler) GetMinResolvedTS(w http.ResponseWriter, r *http.Request) {
```
disksing, line 47: can it load minTS from `cluster`?

lhy1024: if a store with leader peers doesn't report, the cluster will return zero, and the storage will return the min resolved ts in the last round.

server/cluster/cluster.go
lhy1024: yes, we will not persist the value.

lhy1024:
PTAL @AndreMouche @disksing @rleungx

server/cluster/cluster.go
```diff
@@ -1690,30 +1674,79 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
...
+		}
+		if curMinResolvedTS > s.GetMinResolvedTS() {
+			curMinResolvedTS = s.GetMinResolvedTS()
+		}
+	}
+	if curMinResolvedTS == math.MaxUint64 || curMinResolvedTS <= c.minResolvedTS {
 		return
 	}
+	c.minResolvedTS = curMinResolvedTS
+	c.storage.SaveMinResolvedTS(curMinResolvedTS)
```
disksing, line 1697: Can we save storage without holding the lock?

server/cluster/cluster.go
```diff
@@ -1690,30 +1674,79 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
...
 			log.Info("min resolved ts background jobs has been stopped")
 			return
 		case <-ticker.C:
-			minResolvedTS := c.GetMinResolvedTS()
-			if minResolvedTS != math.MaxUint64 {
-				c.Lock()
-				c.storage.SaveMinResolvedTS(minResolvedTS)
-				c.Unlock()
+			interval = c.opt.GetMinResolvedTSPersistenceInterval()
+			if interval != 0 {
```
nolouch, line 1719: It's may strange that user can set the interval be `1s`,`2s` ...`10s`, but when the user set be `0`, it will be `10s` in fact?

lhy1024: if the user set it to 10s, it will not persist value and only update the config.

server/cluster/cluster.go
AndreMouche: Good catch, the `SaveMinResolveTS` is a big operator and here we do needn't the lock, I think to save without holding the lock should be better.

server/cluster/cluster.go
```diff
@@ -1690,30 +1674,79 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
...
+			} else {
+				interval = DefaultMinResolvedTSPersistenceInterval
 			}
+			ticker.Reset(interval)
 		}
 	}
 }
 
+func (c *RaftCluster) loadMinResolvedTS() {
+	c.Lock()
```
AndreMouche, line 1730: Ditto, I think we needn't holding the lock during `c.storage.LoadMinResolvedTS()` too cc @disksing @lhy1024

AndreMouche (approved):
LGTM

disksing (approved):

Issue #4763, opened 2022-03-24
scatter region failed due to region 0 not found during lightning import


fubinzh:
### What did you do?
1. Do a lightning import to tidb cluster with 15 TiKV nodes using lightning local backend

### What did you expect to see?
1. Data imported in TiKV should be scattered among all TiKV nodes.

### What did you see instead?
1. scatter region failed due to below PD error, as a result, data only imported to 3 TiKV nodes, and make these TiKV disk full and not able to finish the import.

```
[2022/03/23 11:44:34.987 +08:00] [WARN] [localhelper.go:385] ["scatter region failed"] [regionCount=3] [failedCount=3] [error="rpc error: code = Unknown desc = region 0 not found"] [retry=0]
```

<img width="604" alt="pd_scatter_failed" src="https://user-images.githubusercontent.com/7403864/159883532-38cf60b2-a452-4f06-b122-7bad10d36f1a.png">

### What version of PD are you using (`pd-server -V`)?
Release Version: v6.0.0
Edition: Community
Git Commit Hash: f1b8f8066f74149c4fce924f67103ffd2d4c7451
Git Branch: heads/refs/tags/v6.0.0
UTC Build Time:  2022-03-18 07:33:08

PR #4582, opened 2022-01-14
*: clean up some lint comments

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

### What problem does this PR solve?

Ref #4322

### What is changed and how it works?

Remove some lint comments and make the code better.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

PR #4766, opened 2022-03-24
server: use old interface for scatter region

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/4763

```commit-message
server: use old interface for scatter region
```

### Check List

Tests

### Release note

```release-note
None
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1203,7 +1203,8 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 			FinishedPercentage: uint64(percentage),
 		}, nil
 	}
-
+	// TODO: Deprecate it use `request.GetRegionsID`.
+	//nolint
 	region := rc.GetRegion(request.GetRegionId())
 	if region == nil {
 		if request.GetRegion() == nil {
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1208,6 +1208,7 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 	region := rc.GetRegion(request.GetRegionId())
 	if region == nil {
 		if request.GetRegion() == nil {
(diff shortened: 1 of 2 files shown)


server/grpc_service.go
```diff
@@ -1204,7 +1204,7 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 		}, nil
 	}
 
-	region := rc.GetRegion(request.GetRegion().GetId())
+	region := rc.GetRegion(request.GetRegionId())
```
JmPotato, line 1207: I think we should support both fields and completely remove the deprecated one in a future release.

nolouch: the new field is `GetRegionsId`, then `regionID` deprecated in https://github.com/pingcap/kvproto/pull/686. Both field `regionID` and `regionsID` are supported. here is incorrectly use `Region` field.

server/grpc_service.go
```diff
@@ -1203,8 +1203,9 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
...
 		}, nil
 	}
-
-	region := rc.GetRegion(request.GetRegion().GetId())
+	// TODO: Deprecate it use `request.GetRegionsID`.
+	//nolint
+	region := rc.GetRegion(request.GetRegionId())
 	if region == nil {
 		if request.GetRegion() == nil {
 			return nil, errors.Errorf("region %d not found", request.GetRegion().GetId())
```
JmPotato, line 1211: This should also be changed back? ref: https://github.com/tikv/pd/pull/4582/files#diff-eb404b45d0a569b05f58f4e4bf2b42b32150be75ec2ed72f86330c4b5d739eabL1194

JmPotato (approved):

lhy1024 (approved):

disksing (approved):

PR #4774, opened 2022-03-30
placement: make meta region isolate by default

### LINKED ISSUE #4772: Metadata should be isolated from other region by default
Ref: https://github.com/pingcap/tiflow/issues/4756

CharlesCheung96:
Signed-off-by: CharlesCheung96 <xiaoshuaizx@gmail.com>

Issue Number: Close #4772

Separate default pacement rule by key range.

### Check List

Tests

- Unit test

Code changes

- Has configuration change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Please add a release note.

```


disksing:
Hi @CharlesCheung96 We should not do it like this. Technicall, PD can run independently of TiDB, so no TiDB specific logic should be introduced into it. A better way might be to set it up in TiDB.

CharlesCheung96:
@disksing Thank you for your suggestion, we will implement this requirement through Region Label.

Issue #3998, opened 2021-08-17
evict-leader-scheduler may cause that PD cannot offline Store


HunDunDM:
### What did you do?

* Create a cluster
* Create `evict-leader-scheduler` for a lot of Stores.
* Offline a Store

### What did you expect to see?

* The Store state becomes Tombstone.

### What did you see instead?

* The Store status remains Offline.

### What version of PD are you using (`pd-server -V`)?

* v5.0.1

lhy1024:
maybe we need to add `move leader` for `evict-leader-scheduler`

HunDunDM:
It cannot completely solve this issue.

nolouch:
@AndreMouche  Would you help to solve this issue?

AndreMouche:
It seems we have already fixed this issue in master, here is my test steps:
1. evict leader on store 1
```
» scheduler add evict-leader-scheduler 1
Success!
» store 1
{
  "store": {
    "id": 1,
    "address": "127.0.0.1:20161",
    "version": "5.4.0-alpha",
    "status_address": "127.0.0.1:20181",
    "git_hash": "cb5cb230e77665e017853ce9292162c5fd250cbb",
    "start_timestamp": 1647934885,
    "deploy_path": "/Users/wuxuelian/workspace/tikv/target/release",
    "last_heartbeat": 1648018539943302000,
    "state_name": "Up"
  },
  "status": {
    "capacity": "460.4GiB",
    "available": "315.5GiB",
    "used_size": "6.407GiB",
    "leader_count": 38,
    "leader_weight": 1,
    "leader_score": 38,
    "leader_size": 2993,
    "region_count": 38,
    "region_weight": 1,
    "region_score": 5153.193036757978,
    "region_size": 2993,
    "slow_score": 1,
    "start_ts": "2022-03-22T15:41:25+08:00",
    "last_heartbeat_ts": "2022-03-23T14:55:39.943302+08:00",
    "uptime": "23h14m14.943302s"
  }
}

```

delete store 1
```
» store delete 1
Success!
» store 1
{
  "store": {
    "id": 1,
    "address": "127.0.0.1:20161",
    "version": "5.4.0-alpha",
    "status_address": "127.0.0.1:20181",
    "git_hash": "cb5cb230e77665e017853ce9292162c5fd250cbb",
    "start_timestamp": 1647934885,
    "deploy_path": "/Users/wuxuelian/workspace/tikv/target/release",
[...]

PR #4775, opened 2022-03-30
server/cluster: consider the number of left evict-leader &  online store before offline a store.

AndreMouche:

Issue Number: part of  #3998 

```commit-message
Consider the number of left online stores and evict-leader stores before removing(offline) a store.
```

### Check List

Tests

- [x] - Unit test
- [ ] - Manual test (add detailed scripts or steps below)

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1061,8 +1061,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
 	upStores := c.getUpStores()
 	expectUpStoresNum := len(upStores) - 1
-	// When placement rules feature is enabled. It is hard to determine required replica count precisely.
-	if !c.opt.IsPlacementRulesEnabled() && expectUpStoresNum < c.opt.GetMaxReplicas() {
+	if expectUpStoresNum < c.opt.GetMaxReplicas() {
 		return fmt.Errorf("can not remove store %d since there are no extra up store has enough space to accommodate the extra replica（%d）", storeID, c.opt.GetMaxReplicas())
 	}
 
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1153,13 +1153,13 @@ func (c *RaftCluster) getEvictLeaderStores() (evictStores []uint64) {
 }
 
 func (c *RaftCluster) getUpStores() []uint64 {
-	upstores := make([]uint64, 0)
+	upStores := make([]uint64, 0)
 	for _, store := range c.GetStores() {
 		if store.IsPreparing() || store.IsServing() {
-			upstores = append(upstores, store.GetID())
+			upStores = append(upStores, store.GetID())
 		}
 	}
-	return upstores
+	return upStores
 }
 
 // BuryStore marks a store as tombstone in cluster.
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -45,6 +45,7 @@ import (
 	"github.com/tikv/pd/server/schedule/hbstream"
 	"github.com/tikv/pd/server/schedule/labeler"
 	"github.com/tikv/pd/server/schedule/placement"
+	"github.com/tikv/pd/server/schedulers"
 	"github.com/tikv/pd/server/statistics"
 	"github.com/tikv/pd/server/storage"
 	"github.com/tikv/pd/server/storage/endpoint"
@@ -1037,6 +1038,12 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 		return errs.ErrStoreDestroyed.FastGenByArgs(storeID)
 	}
 
+	if store.IsPreparing() || store.IsServing() {
+		if err := c.checkReplicaBeforeOfflineStore(storeID); err != nil {
+			return err
+		}
+	}
+
 	newStore := store.Clone(core.OfflineStore(physicallyDestroyed))
 	log.Warn("store has been offline",
 		zap.Uint64("store-id", newStore.GetID()),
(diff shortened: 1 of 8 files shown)


server/cluster/cluster.go
```diff
@@ -1051,6 +1058,65 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 	return err
 }
 
+func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
+	upStores := c.getUpStores()
+	expectUpStoresNum := len(upStores) - 1
+	// When placement rules feature is enabled. It is hard to determine required replica count precisely.
+	if !c.opt.IsPlacementRulesEnabled() && expectUpStoresNum < c.opt.GetMaxReplicas() {
+		return fmt.Errorf("can not remove store %d since there are no extra up store has enough space to accommodate the extra replica（%d）", storeID, c.opt.GetMaxReplicas())
```
rleungx, line 1066: We have another place to give the warning which is duplicated with this, I think we can remove it since it's better to prevent the operation.

server/schedulers/evict_leader.go
```diff
@@ -189,6 +200,11 @@ func newEvictLeaderScheduler(opController *schedule.OperatorController, conf *ev
 	}
 }
 
+// EvictStores returns the IDs of the evict-stores.
+func (s *evictLeaderScheduler) EvictStoreIDs() []uint64 {
```
rleungx, line 204: need lock?

server/cluster/cluster.go
```diff
@@ -1051,6 +1058,65 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
+		}
+	}
+
+	// returns error if there is no store for leader.
+	if evictLeaderStoresNum == len(expectUpstores) {
+		return fmt.Errorf("can not remove store %d since there are no extra up store to store the leader", storeID)
+	}
+
+	return nil
+}
```
rleungx, line 1095: need an empty line

server/schedulers/evict_leader.go
AndreMouche: We do not need a lock here since the method of `s.conf.getStores()` is thread safe

server/cluster/cluster.go
AndreMouche: As we discussed, It is ok we check the `placementRules` twice ^^

server/schedulers/evict_leader.go
```diff
@@ -193,6 +204,11 @@ func newEvictLeaderScheduler(opController *schedule.OperatorController, conf *ev
...
```
disksing, line 208: why not merge this function and `getStores` into one?

server/cluster/cluster.go
```diff
@@ -1071,6 +1072,12 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 		return errs.ErrStoreDestroyed.FastGenByArgs(storeID)
 	}
 
+	if (store.IsPreparing() || store.IsServing()) && !physicallyDestroyed {
+		if err := c.checkReplicaBeforeOfflineStore(storeID); err != nil {
```
disksing, line 1076: When the placement rule is enabled and we cannot easily tell if it should be allowed or not, I think we should always return no error and allow the operation.

server/schedulers/evict_leader.go
AndreMouche: The function `EvictStoreIDs`  should be exist in `evictLeaderScheduler` since it should be public for `evictLeaderScheduler` (`Scheduler`), and the `evictLeaderSchedulerConfig` are not accessible for the package `cluster`.  
We do not implement the function in the `EvictStoreIDs` directly because all we need in this function is to visit the private attributes in the `conf`   with threadsafe.

server/cluster/cluster.go
AndreMouche: However, we think removing the store directly when the placement rule is enabled is not safe, and if the client does want to offline the store even when the number of up stores is not enough and the placement rule is enabled, he can still remove the store with `physicallyDestroyed` true

disksing (approved):

server/cluster/cluster.go
```diff
@@ -1085,6 +1092,65 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
+	evictLeaderStoresNum := 0
+	for _, evitStoreID := range evictStores {
+		if _, ok := expectUpstores[evitStoreID]; ok {
+			evictLeaderStoresNum++
+		}
+	}
+
+	// returns error if there is no store for leader.
+	if evictLeaderStoresNum == len(expectUpstores) {
+		return fmt.Errorf("can not remove store %d since there are no extra up store to store the leader", storeID)
```
JmPotato, line 1124: Ditto.

server/cluster/cluster.go
```diff
@@ -1092,6 +1099,69 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
 }
 
+func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
+	upStores := c.getUpStores()
+	expectUpStoresNum := len(upStores) - 1
+	if expectUpStoresNum < c.opt.GetMaxReplicas() {
+		return errs.ErrStoresNotEnough.FastGenByArgs(storeID, expectUpStoresNum, c.opt.GetMaxReplicas())
+	}
+
+	// Check if thre are extra up store to store the leaders of the regions.
```
JmPotato, line 1109:
```suggestion
	// Check if there are extra up stores to store the leaders of the regions.
```

server/cluster/cluster.go
```diff
@@ -1092,6 +1099,69 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
+	}
+	h, ok := handler.(evictLeaderHandler)
+	if !ok {
+		return
+	}
+	return h.EvictStoreIDs()
+}
+
+func (c *RaftCluster) getUpStores() []uint64 {
+	upstores := make([]uint64, 0)
```
JmPotato, line 1156:
```suggestion
	upStores := make([]uint64, 0)
```

server/cluster/cluster_test.go
```diff
@@ -259,6 +264,62 @@ func (s *testClusterInfoSuite) TestSetOfflineStore(c *C) {
...
+	cluster.coordinator = newCoordinator(s.ctx, cluster, nil)
+
+	// Put 4 stores.
+	for _, store := range newTestStores(4, "2.0.0") {
+		c.Assert(cluster.PutStore(store.GetMeta()), IsNil)
+	}
+
+	c.Assert(cluster.RemoveStore(2, false), IsNil)
+	// should be failed since no enough store to accommodate the extra replica.
+	c.Assert(cluster.RemoveStore(3, false), NotNil)
```
JmPotato, line 280: Could we check the error content to make sure it's the expected one?

server/cluster/cluster_test.go
```diff
@@ -259,6 +264,62 @@ func (s *testClusterInfoSuite) TestSetOfflineStore(c *C) {
...
+		c.Assert(cluster.PutStore(store.GetMeta()), IsNil)
+	}
+	_, err = addEvictLeaderScheduler(cluster, 1)
+
+	c.Assert(err, IsNil)
+	c.Assert(cluster.RemoveStore(2, false), IsNil)
+
+	// should be failed since there is only 1 store left and it is the evict-leader store.
+	err = cluster.RemoveStore(3, false)
+	c.Assert(err, NotNil)
```
JmPotato, line 317: Ditto.

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

PR #4779, opened 2022-03-31
schedule: avoid that the smallest region of the spliting regions  can be merged again.

### LINKED ISSUE #4659: region may be splited after the region is merged due to some wrong configuration.

bufferflies:

Issue Number: Close #4659

notice: 

1. pdctl `config set scheduler ` only check the related key not all, but post api will check all.

2. it can't be checked before start because the configuration of the store may be not initialized.

Weakest think：
should we change the `max-merge-region-size` or `max-merge-split-size` if the configuration of the store has changed.
for example:
the max-split-size has been setted from 98MB to 140Mb, so the `max-merge-region-size` can not work well.  

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
validate config changes before it updates.
```


server/config/store_config.go
```diff
@@ -118,6 +119,48 @@ func (c *StoreConfig) GetRegionMaxKeys() uint64 {
...
+	splitCount := maxRegionSize / splitRegionSize
+
+	// the merge size should be less the min region size
+	if size := maxRegionSize - splitRegionSize*splitCount; size < threshold {
+		threshold = size
+	}
+	// the sum of merge-region-size+max-region-size should be less than splitCount * max-split-size
+	if size := splitRegionSize*(splitCount+1) - maxRegionSize; size < threshold {
+		threshold = size
+	}
```
disksing, line 157: could you explain more about these section?

server/config/store_config.go
```diff
@@ -118,6 +119,32 @@ func (c *StoreConfig) GetRegionMaxKeys() uint64 {
...
 
+// CheckRegionSize return error if the smallest region's size is less than mergeSize
+func (c *StoreConfig) CheckRegionSize(size, mergeSize uint64) error {
+	if size < c.GetRegionMaxSize() {
+		return nil
+	}
+
+	if smallSize := size % c.GetRegionSplitSize(); smallSize <= mergeSize && smallSize != 0 {
+		log.Info("region size is too small", zap.Uint64("size", size), zap.Uint64("mergeSize", mergeSize), zap.Uint64("smallSize", smallSize))
+		return errs.ErrCheckerMergeAgain.FastGenByArgs("the smallest region of the split regions is less than max-merge-region-size, " +
```
rleungx, line 130: Maybe preventing it from splitting is not always a good way. BTW, we have `split-merge-interval` to make sure the split region won't be merged again within a fixed time interval. If we introduce this mechanism, the `split-merge-interval` will be less useful?

bufferflies: In this case ,  the `split-merge-interval`  mechanism  only prevent from merge new region,  but the region will be merged again after `split-merge-interval` seconds ago.

server/config/store_config.go
```diff
@@ -118,6 +119,32 @@ func (c *StoreConfig) GetRegionMaxKeys() uint64 {
...
+			"it will be merged again")
+	}
+	return nil
+}
+
+// CheckRegionKeys return error if the smallest region's keys is less than mergeKeys
+func (c *StoreConfig) CheckRegionKeys(keys, mergeKeys uint64) error {
+	if keys < c.GetRegionMaxKeys() {
```
nolouch, line 138: better to add some comments on `RegionMaxKeys`,  `Region SplitKeys`。

nolouch (commented):
rest LGTM

server/config/store_config.go
rleungx: Oh, I misunderstand the meaning. Better to add a simple example to explain why we do it.

rleungx (approved):
The rest LGTM

server/config/store_config.go
bufferflies: fix

server/config/store_config.go
```diff
@@ -24,6 +24,9 @@ import (
 
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/netutil"
+	"github.com/tikv/pd/pkg/slice"
+
```
rleungx, line 29: remove the line?

bufferflies: fix

nolouch (approved):

PR #4781, opened 2022-04-01
config: reject the config setting request if TTL config exists

### LINKED ISSUE #4780: Set config conflicts with TTL setting
## Enhancement Task
If we wanna change the config in the schedule section but here has already had a TTL setting, It won't take effect until the TTL ends. This operation should be forbidden or raise a warning for users to let them decide if they want to clean up the TTL setting firstly.

rleungx:

Issue Number: Close #4780.

If we change this config which has a TTL setting, the error will be returned.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -145,6 +145,9 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value interfa
 	kp := strings.Split(key, ".")
 	switch kp[0] {
 	case "schedule":
+		if h.svr.IsTTLConfigExist(key) {
+			return errors.Errorf("need to clean up TTL first for %s", key)
+		}
 		return h.updateSchedule(cfg, kp[len(kp)-1], value)
 	case "replication":
 		return h.updateReplication(cfg, kp[len(kp)-1], value)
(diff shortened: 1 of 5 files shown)


server/api/config.go
```diff
@@ -145,6 +145,9 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value interfa
 	kp := strings.Split(key, ".")
 	switch kp[0] {
 	case "schedule":
+		if h.svr.IsTTLConfigExist(key) {
+			return errors.Errorf("need to clean up TTL first for %s", key)
```
AndreMouche, line 149: How about 
```suggestion
			return errors.Errorf("Firstly,  you need to clean up the TTL for %s", key)
```

AndreMouche (commented):
Rest LGTM

server/api/config_test.go
```diff
@@ -349,3 +349,22 @@ func (s *testConfigSuite) TestConfigTTL(c *C) {
...
+	err = postJSON(testDialClient, addr, postData)
+	c.Assert(err, IsNil)
+	assertTTLConfig(c, s.svr.GetPersistOptions(), Equals)
+
+	cfg := map[string]interface{}{"max-snapshot-count": 30}
+	postData, err = json.Marshal(cfg)
+	c.Assert(err, IsNil)
+	addr = fmt.Sprintf("%s/config", s.urlPrefix)
+	err = postJSON(testDialClient, addr, postData)
+	c.Assert(err.Error(), Equals, "\"need to clean up TTL first for schedule.max-snapshot-count\"\n")
```
lhy1024, line 366: ditto

lhy1024 (approved):

server/api/config.go
rleungx: IMO, there is not much difference since this function may also be called by other components, the `you` here is not important.

JmPotato (approved):

PR #4794, opened 2022-04-07
server/schedulers: support update the config of evict_leader concurrently while schedule

### LINKED ISSUE #4793: Should PD support updating the config of EvictLeaderSchedule concurrently while schedule
## Enhancement Task
It seems that `pd` do doesn't support updating the config of `EvictLeaderSchedler` while schedule since there is a lock of conf in the `Scheule` func https://github.com/tikv/pd/blob/master/server/schedulers/evict_leader.go#L244

AndreMouche:

This PR  supports updating the config of EvictLeaderSchedule concurrently while schedule
Issue Number: Close #4793 

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```


server/schedulers/evict_slow_store.go
```diff
@@ -89,6 +89,20 @@ func (conf *evictSlowStoreSchedulerConfig) getSchedulerName() string {
 	return EvictSlowStoreName
 }
 
+func (conf *evictSlowStoreSchedulerConfig) getStores() []uint64 {
+	// TODO: shoud we add a lock?
```
lhy1024, line 93: maybe?

server/schedulers/evict_leader.go
```diff
@@ -279,14 +291,19 @@ func scheduleEvictLeaderBatch(name, typ string, cluster schedule.Cluster, storeR
 	return ops
 }
 
-func scheduleEvictLeaderOnce(name, typ string, cluster schedule.Cluster, storeRanges map[uint64][]core.KeyRange) []*operator.Operator {
-	ops := make([]*operator.Operator, 0, len(storeRanges))
-	for id, ranges := range storeRanges {
+func scheduleEvictLeaderOnce(name, typ string, cluster schedule.Cluster, conf evictLeaderStoresConf) []*operator.Operator {
+	stores := conf.getStores()
```
bufferflies, line 295: Is it safe after store has changed?

server/schedulers/evict_slow_store.go
```diff
@@ -89,6 +89,20 @@ func (conf *evictSlowStoreSchedulerConfig) getSchedulerName() string {
...
+	return conf.EvictedStores
+}
+
+func (conf *evictSlowStoreSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRange {
+	for _, storeID := range conf.EvictedStores {
```
rleungx, line 98: no need for loop for now I think

server/schedulers/evict_slow_store.go
rleungx: seem no need

server/schedulers/evict_leader.go
```diff
@@ -262,10 +269,15 @@ func uniqueAppendOperator(dst []*operator.Operator, src ...*operator.Operator) [
 	return dst
 }
 
-func scheduleEvictLeaderBatch(name, typ string, cluster schedule.Cluster, storeRanges map[uint64][]core.KeyRange, batchSize int) []*operator.Operator {
+type evictLeaderStoresConf interface {
```
rleungx, line 272: may we don't need this interface

rleungx, line 272: Can we move it to a common place?

AndreMouche: We need this interface since we need to use the function `scheduleEvictLeaderBatch` both for `evictLeaderScheduler` and `evictSlowStoreScheduler`.

server/schedulers/evict_slow_store.go
AndreMouche: yes, we do not need a loop,  furthermore, we do not need a slice for `EvictedStores` too. However, with a slice of `EvictStores` , I think a loop may cause less confusion. Here the usage of the loop is cheap so we needn't consider the performance here.

server/schedulers/evict_slow_store.go
AndreMouche: yes, we do not need it now, but with the design, we need it in the future.

lhy1024 (approved):

server/schedulers/evict_leader.go
AndreMouche: yes

CabinfeverB (approved):
LGTM

PR #4803, opened 2022-04-12
schedulers/evict_slow_store: return directly  when there is more than 1 slow store.

### LINKED ISSUE #4802: schedulers/evict_slow_store: speed up the check when there are more than 1 slow store.
## Enhancement Task
It seems we can return when we find there is more than 1 slow store in the function `Schedule` while we try to find the next evict-slow store: https://github.com/tikv/pd/blob/v6.0.0/server/schedulers/evict_slow_store.go#L175-L187

AndreMouche:

Issue Number: Close #4802 

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedulers/evict_slow_store.go
+++ b/server/schedulers/evict_slow_store.go
@@ -80,9 +80,6 @@ func (conf *evictSlowStoreSchedulerConfig) evictStore() uint64 {
 }
 
 func (conf *evictSlowStoreSchedulerConfig) setStoreAndPersist(id uint64) error {
-	if id == conf.evictStore() {
-		return nil
-	}
 	conf.EvictedStores = []uint64{id}
 	return conf.Persist()
 }
--- a/server/schedulers/evict_slow_store.go
+++ b/server/schedulers/evict_slow_store.go
@@ -80,6 +80,9 @@ func (conf *evictSlowStoreSchedulerConfig) evictStore() uint64 {
 }
 
 func (conf *evictSlowStoreSchedulerConfig) setStoreAndPersist(id uint64) error {
+	if id == conf.evictStore() {
+		return nil
+	}
 	conf.EvictedStores = []uint64{id}
 	return conf.Persist()
 }
@@ -111,8 +114,9 @@ func (s *evictSlowStoreScheduler) EncodeConfig() ([]byte, error) {
 }
 
 func (s *evictSlowStoreScheduler) Prepare(cluster schedule.Cluster) error {
-	if len(s.conf.EvictedStores) != 0 {
-		return s.prepareEvictLeader(cluster)
+	evictStore := s.conf.evictStore()
+	if evictStore != 0 {
+		return cluster.SlowStoreEvicted(evictStore)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


server/schedulers/evict_slow_store.go
```diff
@@ -175,36 +182,35 @@ func (s *evictSlowStoreScheduler) Schedule(cluster schedule.Cluster) []*operator
...
+	var slowStore *core.StoreInfo
+
 	for _, store := range cluster.GetStores() {
 		if store.IsRemoved() {
 			continue
 		}
 
 		if (store.IsPreparing() || store.IsServing()) && store.IsSlow() {
-			slowStores = append(slowStores, store)
+			// If there is more than onw slow store do nothing.
```
bufferflies, line 193:
```suggestion
			// It will do nothing If there is more than one slow store.
```

server/schedulers/evict_slow_store.go
```diff
@@ -175,36 +182,35 @@ func (s *evictSlowStoreScheduler) Schedule(cluster schedule.Cluster) []*operator
...
+			if slowStore != nil {
```
bufferflies, line 194: should it report some metrics to analyze the reason why not generates operators?

server/schedulers/evict_slow_store.go
```diff
@@ -121,8 +122,14 @@ func (s *evictSlowStoreScheduler) Cleanup(cluster schedule.Cluster) {
 	s.cleanupEvictLeader(cluster)
 }
 
-func (s *evictSlowStoreScheduler) prepareEvictLeader(cluster schedule.Cluster) error {
-	return cluster.SlowStoreEvicted(s.conf.EvictedStores[0])
+func (s *evictSlowStoreScheduler) prepareEvictLeader(cluster schedule.Cluster, storeID uint64) error {
+	err := s.conf.setStoreAndPersist(storeID)
+	if err != nil {
+		log.Info("evict-slow-store-scheduler persist config failed")
```
bufferflies, line 128: how about add one log field about store_id?

server/schedulers/evict_slow_store.go
AndreMouche: Maybe we could report the number of slow stores in another PR.

bufferflies (approved):

server/schedulers/evict_slow_store.go
```diff
@@ -111,8 +111,9 @@ func (s *evictSlowStoreScheduler) EncodeConfig() ([]byte, error) {
 }
 
 func (s *evictSlowStoreScheduler) Prepare(cluster schedule.Cluster) error {
-	if len(s.conf.EvictedStores) != 0 {
-		return s.prepareEvictLeader(cluster)
+	evictStore := s.conf.evictStore()
```
rleungx, line 114: The original way can reduce the number of function calls.

rleungx (commented):
The rest LGTM.

server/schedulers/evict_slow_store.go
AndreMouche: The origin way visits the arguments of `s.conf` directly which is not a good way, using the function of `s.conf` instead would be clear, that is way we need to define the struct of `conf` II think

rleungx (approved):

PR #4816, opened 2022-04-18
server/cluster/coordinator.go: new function to avoid visiting private attributes from another struct

### LINKED ISSUE #4815: server/cluster/coordinator: avoid visiting private attributes(ctx,wg) of coordinator directly from cluster
## Enhancement Task

The `RaftCluster` visited the private attributes `ctx` and `wg`  of `coordinator` directly in the function `runCoordinator` which is not a good habit, Maybe we could define a function in `coordinator` to avoid the usage.  https://github.com/tikv/pd/blob/v6.1.0-alpha/server/cluster/cluster.go#L359-L363

AndreMouche:

Issue Number: Close #4815 

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -368,13 +368,7 @@ func (c *RaftCluster) runStatsBackgroundJobs() {
 func (c *RaftCluster) runCoordinator() {
 	defer logutil.LogPanic()
 	defer c.wg.Done()
-	defer func() {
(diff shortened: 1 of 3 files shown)


lhy1024 (approved):

nolouch (approved):
lgtm

Issue #4666, opened 2022-02-22
Add HTTP rate limit middleware


CabinfeverB:
## Development Task
ref #4480

PR #4839, opened 2022-04-24
*: Add Limiter Config

CabinfeverB:

Issue Number: Ref #4666
This PR is used to add limiter config and save it into the config of server

It add `DimensionConfig` for `Limiter`. `map[string]DimensionConfig` .
You can think of `map[string]DimensionConfig` as a configuration snapshot of the QPS limiter and the concurrent limiter.
Checking `map[string]DimensionConfig` firstly to reduce intrusion into the limiter itself when it comes to configuration changes.

```commit-message
Add Rate Limiter Config for server
```

### Check List

Tests

- Unit test

Code changes

- Has configuration change
- Has persistent data change

### Release note

```release-note
None.
```


pkg/ratelimit/limiter.go
```diff
@@ -81,7 +105,7 @@ func (l *Limiter) GetQPSLimiterStatus(label string) (limit rate.Limit, burst int
 }
 
 // DeleteQPSLimiter deletes QPS limiter of given label
```
rleungx, line 107:
```suggestion
// deleteQPSLimiter deletes QPS limiter of a given label
```

pkg/ratelimit/limiter.go
```diff
@@ -95,7 +119,7 @@ func (l *Limiter) GetConcurrencyLimiterStatus(label string) (limit uint64, curre
 }
 
 // DeleteConcurrencyLimiter deletes concurrency limiter of given label
```
rleungx, line 121: ditto

pkg/ratelimit/option.go
```diff
@@ -16,39 +16,112 @@ package ratelimit
...
+	ConcurrencyNoChange
+	// ConcurrencyChanged shows that limiter's config is changed and not deleted.
+	ConcurrencyChanged
+	// ConcurrencyDeleted shows that limiter's config is deleted.
+	ConcurrencyDeleted
+	// InAllowList shows that limiter's config isn't changed because it is in in allow list.
+	InAllowList
+	// Ignore shows that the status can be ignored when initiated
+	Ignore
+)
```
rleungx, line 40: Why do we need these flags?

CabinfeverB: It is used to return update results for users in update config API in the future.

pkg/ratelimit/limiter.go
```diff
@@ -20,8 +20,27 @@ import (
...
+// NewLimiterConfig returns a new LimiterConfig
+func NewLimiterConfig() LimiterConfig {
+	return make(map[string]DimensionConfig)
+}
+
+// DimensionConfig is the limit dimension config of one label
+type DimensionConfig struct {
+	// qps conifg
+	QPS      float64
+	QPSBrust int
```
JmPotato, line 35:
```suggestion
	QPSBurst int
```

server/config/config.go
```diff
@@ -1111,7 +1113,11 @@ type PDServerConfig struct {
 	// MinResolvedTSPersistenceInterval is the interval to save the min resolved ts.
 	MinResolvedTSPersistenceInterval typeutil.Duration `toml:"min-resolved-ts-persistence-interval" json:"min-resolved-ts-persistence-interval"`
 	// EnableAudit controls the switch of the audit middleware
-	EnableAudit bool `toml:"enable-audit" json:"enable-audit"`
+	EnableAudit bool `toml:"enable-audit" json:"enable-audit,string"`
+	// EnableRateLimit controls the switch of the rate limit middleware
+	EnableRateLimit bool `toml:"enable-rate-limit" json:"enable-rate-limit,string"`
+	// RateLimitConfig is the config of rate limit middleware
+	RateLimitConfig ratelimit.LimiterConfig `toml:"rate-limit-config" json:"rate-limit-config"`
```
nolouch, line 1120: I prefer an independent config that is not in the server config; other than the server config may look big.

server/config/persist_options.go
```diff
@@ -449,6 +449,11 @@ func (o *PersistOptions) IsAuditEnabled() bool {
 	return o.GetPDServerConfig().EnableAudit
 }
 
+// IsRateLimitEnabled returns whether rate limit middleware is enabled
+func (o *PersistOptions) IsRateLimitEnabled() bool {
```
JmPotato, line 453: Will this method be called in the future? It seems that there is no caller in this PR.

pkg/ratelimit/option.go
```diff
@@ -16,39 +16,112 @@ package ratelimit
...
+		return Ignore
+	}
+}
+
+func updateConcurrencyConfig(l *Limiter, label string, limit uint64) UpdateStatus {
+	l.configMux.Lock()
+	defer l.configMux.Unlock()
+
+	cfg := l.labelConfig[label]
+	if cfg.ConcurrencyLimit == limit {
```
JmPotato, line 60: Is it possible that the `label` does not exist, so the `cfg.ConcurrencyLimit` is always equal to a zero limit which will make the deleting can not be performed forever.

CabinfeverB: > Is it possible that the `label` does not exist, so the `cfg.ConcurrencyLimit` is always equal to a zero `limit` which will make the deleting can not be performed forever.

I think if the `label` does not exist and the `limit` is equal to 0, then the `label` has no limiter, so it does not need to be deleted and there is no change.

pkg/ratelimit/limiter.go
```diff
@@ -30,7 +49,10 @@ type Limiter struct {
 
 // NewLimiter returns a global limiter which can be updated in the later.
 func NewLimiter() *Limiter {
-	return &Limiter{labelAllowList: make(map[string]struct{})}
+	return &Limiter{
+		labelAllowList: make(map[string]struct{}),
+		labelConfig:    NewLimiterConfig(),
```
rleungx, line 54: `make(map[string]DimensionConfig)` directly?

pkg/ratelimit/option.go
```diff
@@ -16,39 +16,112 @@ package ratelimit
...
+	QPSDeleted
+	// ConcurrencyNoChange shows that limiter's config isn't changed.
+	ConcurrencyNoChange
+	// ConcurrencyChanged shows that limiter's config is changed and not deleted.
+	ConcurrencyChanged
+	// ConcurrencyDeleted shows that limiter's config is deleted.
+	ConcurrencyDeleted
+	// InAllowList shows that limiter's config isn't changed because it is in in allow list.
+	InAllowList
+	// Ignore shows that the status can be ignored when initiated
```
rleungx, line 38: Should it be added?

pkg/ratelimit/limiter.go
```diff
@@ -65,10 +87,12 @@ func (l *Limiter) Release(label string) {
 }
 
 // Update is used to update Ratelimiter with Options
-func (l *Limiter) Update(label string, opts ...Option) {
+func (l *Limiter) Update(label string, opts ...Option) UpdateStatus {
```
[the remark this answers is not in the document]
CabinfeverB: It will be used in `server.go`

pkg/ratelimit/limiter.go
```diff
@@ -20,8 +20,22 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// LimiterConfig is the config of Limiter
+type LimiterConfig map[string]DimensionConfig
```
rleungx, line 24: Do we still need it?

pkg/ratelimit/limiter.go
```diff
@@ -20,8 +20,19 @@ import (
...
+	QPS      float64
+	QPSBurst int
+	// concurrency config
+	ConcurrencyLimit uint64
+}
+
 // Limiter is a controller for the request rate.
 type Limiter struct {
+	configMux          sync.Mutex
+	labelConfig        map[string]DimensionConfig
```
AndreMouche, line 35: Could we use pointer instead ?
```suggestion
	labelConfig        map[string]*DimensionConfig
```

pkg/ratelimit/option.go
AndreMouche: Init the labelConfig directly when it not exist will make me(the reviewer) happier  I think😄

pkg/ratelimit/limiter.go
```diff
@@ -80,8 +96,8 @@ func (l *Limiter) GetQPSLimiterStatus(label string) (limit rate.Limit, burst int
 	return 0, 0
 }
 
-// DeleteQPSLimiter deletes QPS limiter of given label
-func (l *Limiter) DeleteQPSLimiter(label string) {
+// deleteQPSLimiter deletes QPS limiter of a given label
```
AndreMouche, line 99:
```suggestion
// deleteQPSLimiter deletes QPS limiter with a given label
```

pkg/ratelimit/limiter.go
```diff
@@ -94,8 +110,8 @@ func (l *Limiter) GetConcurrencyLimiterStatus(label string) (limit uint64, curre
 	return 0, 0
 }
 
-// DeleteConcurrencyLimiter deletes concurrency limiter of given label
-func (l *Limiter) DeleteConcurrencyLimiter(label string) {
+// deleteConcurrencyLimiter deletes concurrency limiter of a given label
```
AndreMouche, line 113:
```suggestion
// deleteConcurrencyLimiter deletes concurrency limiter of the given label
```

rleungx (approved):

pkg/ratelimit/limiter.go
```diff
@@ -20,8 +20,19 @@ import (
 	"golang.org/x/time/rate"
 )
 
+// DimensionConfig is the limit dimension config of one label
+type DimensionConfig struct {
```
AndreMouche, line 24: Could you please try to remove `DimensionConfig` and use `LabelLimitters` instead of `Limitter`?
Since we  already store the config in `RateLimiter` and `concurrencyLimiter` and they are all theadsafe
```
type limiter struct {
	qps         *RateLimiter
	concurrency *concurrencyLimiter
}

type LabelLimitters struct {
	limiters sync.Map // label->limiter
	labelAllowList map[string]struct{}
}
```

AndreMouche (commented):
any update?

CabinfeverB:
> any update?

Just updated

pkg/ratelimit/option.go
```diff
@@ -16,39 +16,101 @@ package ratelimit
...
+
+	if (float64(oldQPSLimit)-limit < eps && float64(oldQPSLimit)-limit > -eps) && oldBurst == burst {
+		return QPSNoChange
+	}
+	if limit <= 0 || burst < 1 {
+		l.QPSUnlimit(label)
+		return QPSDeleted
 	}
+	if limiter, exist := l.qpsLimiter.LoadOrStore(label, NewRateLimiter(limit, burst)); exist {
+		limiter.(*RateLimiter).SetLimit(rate.Limit(limit))
```
AndreMouche, line 80: Should we make it thread-safe? For example, is it possible that `SetLimit` happens concurrently?

AndreMouche (commented):
Rest LGTM

AndreMouche (approved):
LGTM

PR #4840, opened 2022-04-24
api: Refactor Service Middleware

### LINKED ISSUE #4846: Service Middleware doesn't put `next` handler into middleware actually
### What did you do?
I want to get the HTTP duration of `GetRegions`from Prometheus audit.

If regions is up to 10K, it will take hundreds of ms.

### What did you see instead?
It will take a few ms.

https://github.com/tikv/pd/blob/1fc68b2abd664e6bf73bdda874399c8db4435cf6/server/api/router.go#L103-L108
Audit backends whose `ProcessBeforeHandler()` is false will process before handler processing not after.

It will also affect Rate-Limit Middleware.

master

CabinfeverB:

Issue Number: Close #4846 Ref 

Create one middleware for each HTTP API interface.
And I remove failpoint in audit middleware because it has two backends to test. 

```commit-message
Refactor Service Middleware
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/api/middleware.go
+++ b/server/api/middleware.go
@@ -17,11 +17,9 @@ package api
 import (
 	"context"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/pingcap/failpoint"
-	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/audit"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/requestutil"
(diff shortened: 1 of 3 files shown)


server/api/middleware.go
```diff
@@ -106,23 +121,15 @@ func (s *auditMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request, next
...
-	requestInfo, ok := requestutil.RequestInfoFrom(r.Context())
-	if !ok {
-		log.Error("failed to get request info when auditing")
-		next(w, r)
-		return
-	}
+	// There is no need to check whether requestInfo is available like getting RaftCluster
+	requestInfo, _ := requestutil.RequestInfoFrom(r.Context())
 
 	labels := s.svr.GetServiceAuditBackendLabels(requestInfo.ServiceLabel)
```
rleungx, line 127: If there is no requestInfo, we don't need to get it?

CabinfeverB: > If there is no requestInfo, we don't need to get it?

If we open audit or rate-limit switch, there must be a requestInfo.

rleungx: Please update the comment here.

rleungx (approved):
The rest LGTM.

nolouch (approved):

PR #4842, opened 2022-04-25
server: add limiter config and reload mechanism

CabinfeverB:

Issue Number: Ref #4666
**This PR should be merged after #4839**
**This PR should be merged after #4869 **
This PR is used to reload the limit config for limiter after leader resign or restart and do some integration tests

update limiter config when reload presist config

```commit-message
update limiter config when reload presist config
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/server.go
```diff
@@ -1475,6 +1506,14 @@ func (s *Server) reloadConfigFromKV() error {
 	return nil
 }
 
+func (s *Server) loadRateLimitConfig() {
```
rleungx, line 1509: Why do we need it?

server/config/service_middleware_config.go
```diff
@@ -14,22 +14,30 @@
...
+	RateLimitConfig `json:"rate-limit"`
 }
 
 // NewServiceMiddlewareConfig returns a new service middleware config
 func NewServiceMiddlewareConfig() *ServiceMiddlewareConfig {
 	audit := AuditConfig{
 		EnableAudit: defaultEnableAuditMiddleware,
 	}
+	ratelimit := RateLimitConfig{
+		EnableRateLimit: defaultEnableRateLimitMiddleware,
```
rleungx, line 36: need make?

server/server.go
CabinfeverB: The configuration of the Limiter differs from that of EnableRateLimit. The value of EnableRateLimit is fetched from PersistOption on every request, but Limiter does not. So after resigning leader, we should reload the associated configuration.

server/api/service_middleware_test.go
```diff
@@ -51,23 +51,27 @@ func (s *testServiceMiddlewareSuite) TearDownSuite(c *C) {
...
 		"audit.enable-audit": "false",
+		"enable-rate-limit":  "false",
 	}
 	postData, err = json.Marshal(ms)
 	c.Assert(err, IsNil)
 	c.Assert(tu.CheckPostJSON(testDialClient, addr, postData, tu.StatusOK(c)), IsNil)
 	sc = &config.ServiceMiddlewareConfig{}
 	c.Assert(tu.ReadGetJSON(c, testDialClient, addr, sc), IsNil)
 	c.Assert(sc.EnableAudit, Equals, false)
+	c.Assert(sc.EnableRateLimit, Equals, false)
```
AndreMouche, line 74: Could you please separate the test for rate-limit from `audit`?

server/server.go
```diff
@@ -986,6 +990,27 @@ func (s *Server) SetAuditConfig(cfg config.AuditConfig) error {
...
+// GetRateLimitConfig gets the rate limit config information.
+func (s *Server) GetRateLimitConfig() *config.RateLimitConfig {
+	return s.serviceMiddlewarePersistOptions.GetRateLimitConfig().Clone()
+}
+
+// SetRateLimitConfig sets the rate limit config.
+func (s *Server) SetRateLimitConfig(cfg config.RateLimitConfig) error {
+	old := s.serviceMiddlewarePersistOptions.GetRateLimitConfig()
+	s.serviceMiddlewarePersistOptions.SetRateLimitConfig(&cfg)
+	if err := s.serviceMiddlewarePersistOptions.Persist(s.storage); err != nil {
```
AndreMouche, line 1002: Oh, Should we make this function threadsafe? Was it possible that two goroutines run this function, for example, should we avoid the following case:
1. go1 finished L1001 and update the cfg to false
2. go2 finished 1001 and update the cfg to true in cache
3. go2 persist the value with true (from cache) and return success
4. go1 persist the value with true(from cache) while it expect the value is false.
and more different cases may meet in the `Persist` with it not protect while read and write..

CabinfeverB: This happens to all config that needs to `Persist`, and if optimization is needed, I think we can fix it together in another PR

server/config/service_middleware_config.go
CabinfeverB: update

server/api/service_middleware_test.go
CabinfeverB: > Meanwhile, could you please add some tests of updating the item of `limiter-config`?

Some tests will be added in #4843 after providering update API

tests/server/config/config_test.go
```diff
@@ -0,0 +1,106 @@
...
+var _ = Suite(&testConfigPresistSuite{})
+
+type testConfigPresistSuite struct {
+	cleanup func()
+	cluster *tests.TestCluster
+}
+
+func (s *testConfigPresistSuite) SetUpSuite(c *C) {
+	ctx, cancel := context.WithCancel(context.Background())
+	server.EnableZap = true
```
rleungx, line 51: This config is no longer used.

server/api/service_middleware_test.go
CabinfeverB: > Meanwhile, could you please add some tests of updating the item of `limiter-config`?

I add a simple test, more test will in #4843 by specific API

server/api/service_middleware.go
```diff
@@ -129,3 +132,24 @@ func (h *serviceMiddlewareHandler) updateAudit(config *config.ServiceMiddlewareC
...
 	return err
 }
+
+func (h *serviceMiddlewareHandler) updateRateLimit(config *config.ServiceMiddlewareConfig, key string, value interface{}) error {
+	data, err := json.Marshal(map[string]interface{}{key: value})
+	if err != nil {
+		return err
+	}
+
+	updated, found, err := mergeConfig(&config.RateLimitConfig, data)
```
rleungx, line 142: Do we need it?

server/api/service_middleware_test.go
```diff
@@ -44,11 +45,11 @@ func (s *testServiceMiddlewareSuite) SetUpSuite(c *C) {
 	s.urlPrefix = fmt.Sprintf("%s%s/api/v1", addr, apiPrefix)
 }
 
-func (s *testServiceMiddlewareSuite) TearDownSuite(c *C) {
+func (s *testAuditMiddlewareSuite) TearDownSuite(c *C) {
 	s.cleanup()
 }
 
-func (s *testServiceMiddlewareSuite) TestConfigAudit(c *C) {
+func (s *testAuditMiddlewareSuite) TestConfigAuditSwitch(c *C) {
```
rleungx, line 52: Please add a test for the default value.

CabinfeverB: update

rleungx (approved):

AndreMouche (approved):
LGTM

PR #4845, opened 2022-04-25
server/api: create  scatter_region operator with opAdmin

### LINKED ISSUE #4833: Allow manual scatter region when region scheduler is disabled
## Enhancement Task

When importing large scale data, Lightning will first disable scheduler for regions in the target key range, and then split and scatter regions. But scatter region doesn't work due to region scheduler is disabled. 

I think we should allow manual scheduling, even if the automatic region scheduler is disabled. This would provide many benefits. For Lightning, scatter regions should work. Also, it extends the capabilities of pd and other components can implement its custom scheduler, kind of like K8s.

AndreMouche:

Issue Number: Close #4833

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1272,7 +1272,7 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 		region = core.NewRegionInfo(request.GetRegion(), request.GetLeader())
 	}
 
-	op, err := rc.GetRegionScatter().Scatter(region, request.GetGroup())
+	op, err := rc.GetRegionScatter().Scatter(region, request.GetGroup(), 0)
 	if err != nil {
 		return nil, err
 	}
@@ -1644,7 +1644,7 @@ func (s *GrpcServer) SplitAndScatterRegions(ctx context.Context, request *pdpb.S
 
 // scatterRegions add operators to scatter regions and return the processed percentage and error
 func scatterRegions(cluster *cluster.RaftCluster, regionsID []uint64, group string, retryLimit int) (int, error) {
-	ops, failures, err := cluster.GetRegionScatter().ScatterRegionsByID(regionsID, group, retryLimit)
+	ops, failures, err := cluster.GetRegionScatter().ScatterRegionsByID(regionsID, group, retryLimit, 0)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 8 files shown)


server/handler.go
```diff
@@ -809,7 +809,7 @@ func (h *Handler) AddSplitRegionOperator(regionID uint64, policyStr string, keys
 }
 
 // AddScatterRegionOperator adds an operator to scatter a region.
-func (h *Handler) AddScatterRegionOperator(regionID uint64, group string) error {
+func (h *Handler) AddScatterRegionOperator(regionID uint64, group string, kind operator.OpKind) error {
```
rleungx, line 812: How about only changing the kind inside these two functions?

server/grpc_service.go
```diff
@@ -1644,7 +1644,7 @@ func (s *GrpcServer) SplitAndScatterRegions(ctx context.Context, request *pdpb.S
 
 // scatterRegions add operators to scatter regions and return the processed percentage and error
 func scatterRegions(cluster *cluster.RaftCluster, regionsID []uint64, group string, retryLimit int) (int, error) {
-	ops, failures, err := cluster.GetRegionScatter().ScatterRegionsByID(regionsID, group, retryLimit)
+	ops, failures, err := cluster.GetRegionScatter().ScatterRegionsByID(regionsID, group, retryLimit, 0)
```
JmPotato, line 1647: Using `0` here is kind of confusing, could we have it to become something like a const or alias?

server/schedule/operator/kind.go
```diff
@@ -25,6 +25,8 @@ type OpKind uint32
 
 // Flags for operators.
 const (
+	// Default value.
+	OpDefault OpKind = 0
```
bufferflies, line 29: where is it used? should the function `flagToName`  add the default kind?

server/grpc_service.go
```diff
@@ -1272,7 +1273,7 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 		region = core.NewRegionInfo(request.GetRegion(), request.GetLeader())
 	}
 
-	op, err := rc.GetRegionScatter().Scatter(region, request.GetGroup())
+	op, err := rc.GetRegionScatter().Scatter(region, request.GetGroup(), operator.OpDefault)
```
bufferflies, line 1276: why is it different between http and grpc?

PR #4852, opened 2022-04-27
server/api: create scatter_region operator with opAdmin

### LINKED ISSUE #4833: Allow manual scatter region when region scheduler is disabled

AndreMouche:

Issue Number: Close #4833

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```

### DIFF
--- a/server/api/region.go
+++ b/server/api/region.go
@@ -885,6 +885,7 @@ func (h *regionsHandler) ScatterRegions(w http.ResponseWriter, r *http.Request) 
 	}
 	// If there existed any operator failed to be added into Operator Controller, add its regions into unProcessedRegions
 	for _, op := range ops {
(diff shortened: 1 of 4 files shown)


lhy1024 (commented):
some tools, such as br, could create scatter operator too.

rleungx (approved):

nolouch (approved):

PR #4873, opened 2022-04-28
api: fix the potential race problem

### LINKED ISSUE #4872: There might be a data race problem between API v2 and dashboard
## Bug Report
Both dashboard and API v2 will set the gin mode and read it when configuring the router. But the mode here is a global variable, so we need to use `once` to prevent it from being set twice.

rleungx:

Issue Number: Close #4872.

This PR uses `once` to prevent `mode` from being set twice.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/apiv2/router.go
+++ b/server/apiv2/router.go
@@ -39,6 +39,7 @@ const apiV2Prefix = "/pd/api/v2/"
 // NewV2Handler creates a HTTP handler for API.
 func NewV2Handler(_ context.Context, svr *server.Server) (http.Handler, server.ServiceGroup, error) {
 	once.Do(func() {
+		// See https://github.com/pingcap/tidb-dashboard/blob/f8ecb64e3d63f4ed91c3dca7a04362418ade01d8/pkg/apiserver/apiserver.go#L84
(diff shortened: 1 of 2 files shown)


lhy1024 (approved):

server/apiv2/router.go
```diff
@@ -34,7 +38,12 @@ const apiV2Prefix = "/pd/api/v2/"
 
 // NewV2Handler creates a HTTP handler for API.
 func NewV2Handler(_ context.Context, svr *server.Server) (http.Handler, server.ServiceGroup, error) {
-	gin.SetMode(gin.ReleaseMode)
+	once.Do(func() {
+		// These global modification will be effective only for the first invoke.
+		_ = godotenv.Load()
```
JmPotato, line 43: What's this for?

rleungx: Copied from the dashboard, I guess because it supports configuring the mode through environment variables.

AndreMouche (commented):
LGTM

JmPotato (approved):

PR #4896, opened 2022-05-06
server/scheduler/labeler: delete region rule instead of delete placement rule when region rule expired

### LINKED ISSUE #4895: region/labeler: shouldn't delete placement_rule when the label_rule expired
## Bug Report
It seems there is a bug that we remove the placement rule(`DeleteRule(id)`) instead of region_label(`DeleteRegionRule`).
[https://github.com/tikv/pd/blob/16bc5cd65175eb75f69eaa88d1d330c21adbb76c/server/schedule/labeler/labeler.go#L195-L197](https://github.com/tikv/pd/blob/16bc5cd65175eb75f69eaa88d1d330c21adbb76c/server/schedule/labeler/labeler.go#L195-L197)

The influence depend on the possibility that the  placement_rule shares the same key with the region_label.

AndreMouche:

Issue Number: Close #4895 

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/labeler/labeler.go
+++ b/server/schedule/labeler/labeler.go
@@ -193,7 +193,7 @@ func (l *RegionLabeler) getAndCheckRule(id string, now time.Time) *LabelRule {
 		return rule
 	}
 	if len(rule.Labels) == 0 {
-		l.storage.DeleteRule(id)
+		l.storage.DeleteRegionRule(id)
 		delete(l.labelRules, id)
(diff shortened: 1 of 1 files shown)


HunDunDM (approved):

rleungx (approved):
LGTM, except for the testing part.

disksing (approved):

PR #4907, opened 2022-05-07
test: fix data race in store limit test

### LINKED ISSUE #4906: Store Limit test happen Data race
### Which jobs are failing
[chunks (2)](https://github.com/tikv/pd/runs/6333649140?check_suite_focus=true#logs)

### CI link
https://github.com/tikv/pd/runs/6333649140?check_suite_focus=true

data race

CabinfeverB:

Issue Number: Close #4906

clone config

```commit-message
fix data race in store limit test
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024 (approved):

rleungx:
Do other tests have this problem?

bufferflies (approved):

AndreMouche (approved):
LGTM, good catch!

CabinfeverB:
> Do other tests have this problem?

I found the trigger for the data race needs two conditions, the first is there is a point which is direct from `persistOptions` , the second is the need to use the Set method in the ` Server ` (the Set method in the ` Server ` will not Clone, `PersistOptions` will)

rleungx (approved):

PR #4927, opened 2022-05-12
server/cluster/cluster.go: use atomic for running check instead of sharing lock with all attributes

### LINKED ISSUE #4823: server/cluster/cluster.go: share the lock with all attribbutes for running is expensive
## Enhancement Task
The attribute `running ` shares the lock with other attributes in the `RaftCluster`, which is expensive, could we use `atomic.StoreInt64` instead similar with `Server.isServing`
[https://github.com/tikv/pd/blob/56a015c32925022cda9c803830c058916ca02d38/server/cluster/cluster.go#L405-L408](https://github.com/tikv/pd/blob/56a015c32925022cda9c803830c058916ca02d38/server/cluster/cluster.go#L405-L408)

ref `Server.IsClosed` [https://github.com/tikv/pd/blob/56a015c32925022cda9c803830c058916ca02d38/server/server.go#L494-L496](https://github.com/tikv/pd/blob/56a015c32925022cda9c803830c058916ca02d38/server/server.go#L494-L496)

e1ijah1:

Issue Number: Close #4823 

```commit-message
server/cluster/cluster.go: use atomic for running check instead of sharing lock with all attribbutes
```

### Check List

Tests

- Unit test

Code changes

- N/A

Side effects

- N/A

Related changes

- N/A

### Release note

```release-note
None
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -119,7 +119,7 @@ type RaftCluster struct {
 	etcdClient *clientv3.Client
 	httpClient *http.Client
 
-	running            int64
+	running            atomic.Bool
 	meta               *metapb.Cluster
 	storeConfigManager *config.StoreConfigManager
 	storage            storage.Storage
@@ -291,8 +291,8 @@ func (c *RaftCluster) Start(s Server) error {
 	go c.runMinResolvedTSJob()
 	go c.runSyncConfig()
 	go c.runUpdateStoreStats()
-	atomic.StoreInt64(&c.running, 1)
 
+	c.running.Store(true)
 	return nil
 }
 
@@ -493,7 +493,7 @@ func (c *RaftCluster) runReplicationMode() {
 
 // Stop stops the cluster.
 func (c *RaftCluster) Stop() {
-	if !atomic.CompareAndSwapInt64(&c.running, 1, 0) {
+	if !c.running.CompareAndSwap(true, false) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


server/cluster/cluster.go
```diff
@@ -464,16 +461,12 @@ func (c *RaftCluster) Stop() {
...
-	return c.running
+	return atomic.LoadInt64(&c.running) == 1
 }
 
 // Context returns the context of RaftCluster.
 func (c *RaftCluster) Context() context.Context {
-	c.RLock()
-	defer c.RUnlock()
-	if c.running {
+	if c.IsRunning() {
```
JmPotato, line 469: I'm worried that this may cause some unexpected panics. Since the PR makes the `c.running` become zero earlier than before. Would you like to take a deep dive into this to see if this will happen?

e11jah: I think it's ok. coz the context of the cluster is used for the `syncHistoryRegion` function, and if the `c.running` has been set to zero before canceling the context, it will not return the context. And if the context has been cancelled, the `RegionSyncer` will stop syncing historic regions.

JmPotato:
@AndreMouche PTAL, thx!

nolouch (commented):
I think `IsRunning` not in hotpath. cc @AndreMouche

server/cluster/cluster.go
```diff
@@ -112,7 +113,7 @@ type RaftCluster struct {
 	etcdClient *clientv3.Client
 	httpClient *http.Client
 
-	running            bool
+	running            int64
```
rleungx, line 116: How about using atomic bool?

rleungx:
Any update?

rleungx:
CI blocked by #5740

HunDunDM (approved):

PR #4928, opened 2022-05-12
*: add support for tiflash_mpp label/tiflash ReadNode

### LINKED ISSUE #4836: Support for Decouping of Compute and Storage of TiFlash

 The feature requests are from the **Decoupling of Compute and Storage for TiFlash**.
- [ ] Add new label for TiFlash ReadNode: Now the label of **TiFlash** is <engine, tiflash>, we need to to add a new lable <engine, mpp> for tiflash nodes which only used for MPP computation(aka ReadNode). And the region data will be stored in WriteNode.
- [ ] Add node cache information(like region cache), then TiDB use node information when dispatch MPPTask.

guo-shaoge:
Signed-off-by: guo-shaoge <shaoge1994@163.com>

Issue Number: Close #4836

```commit-message
*: add support for tiflash_mpp label/tiflash ReadNode
```
Make PD be compatible with new tiflash_mpp label(<engine, tiflash_mpp>).
### Check List

Tests

N/A

Code changes

N/A

Side effects

N/A

Related changes

N/A

### Release note

```release-note
None
```


server/schedulers/hot_region.go
```diff
@@ -472,6 +472,16 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	if bs.best == nil || len(bs.ops) == 0 {
 		return false
 	}
+
+    // Because there is no region data on the TiFlash ReadNodes,
+    // so neither srcStore nor dstStore should be tiflash_mpp node.
+    if bs.best.srcStore.IsTiFlashMPP() || bs.best.dstStore.IsTiFlashMPP() {
+        log.Error("Got hot region(srcStore) in tiflash_mpp node or scheduling to tiflash_mpp node",
+            zap.Bool("srcStore is tiflash_mpp", bs.best.srcStore.IsTiFlashMPP()),
+            zap.Bool("dstStore is tiflash_mpp", bs.best.dstStore.IsTiFlashMPP()))
```
JmPotato, line 481: I think adding metrics here is better, just like line 486. A log is not that helpful and necessary.

guo-shaoge: I thinks it's ok to add metrics. But why logging is not helpful, we can check log to know this error situation?

server/schedulers/hot_region.go
```diff
@@ -472,6 +472,20 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	if bs.best == nil || len(bs.ops) == 0 {
 		return false
 	}
+
+	// Because there is no region data on the TiFlash ReadNodes,
+	// so neither srcStore nor dstStore should be tiflash_mpp node.
+	if bs.best.srcStore.IsTiFlashMPP() {
+		schedulerCounter.WithLabelValues(bs.sche.GetName(), "tiflash-mpp-as-src-hot-store").Inc()
+		log.Error("tiflash_mpp node is source hot store, which is unexpected")
```
rleungx, line 480: prefer to remove thes logs

guo-shaoge: done

server/schedulers/hot_region.go
guo-shaoge: done

JmPotato (approved):

nolouch (commented):
Could you add some tests?

guo-shaoge:
> Could you add some tests?

Sorry for the late reply. The following cases are added:
1. testHotWriteRegionSchedulerSuite.TestWithTiFlashMPP: test hot_region_scheduler with tiflash_mpp nodes.
2. testClusterInfoSuite.TestEngineLabel: test if engine label is correct when `PutStore`
3. testClusterInfoSuite.TestRegionLabelIsolationLevel: isolation level of tiflash/tiflash_mpp will be ignored(see updateRegionsLabelLevelStats())

I'm writing more complicated cases using test-infra.

PR #4931, opened 2022-05-12
scheduler: support the hot-region-scheduler to generate multiple operators at the same time

HunDunDM:

Issue Number: Ref #4949 

Allows a solution to contain multiple Operators to expand the search set, guaranteeing that a better solution can be found in some cases.

### Check List

Tests

- Unit test

### Release note

```release-note
Optimized the search strategy of the balance-hot-region-scheduler.
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -634,6 +634,7 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 		maxZombieDur = bs.sche.conf.GetStoreStatZombieDuration()
 	}
 
+	// TODO: Process operators atomically.
 	// main peer
 	srcStoreID := bs.best.srcStore.GetID()
 	dstStoreID := bs.best.dstStore.GetID()
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -335,9 +335,9 @@ type solution struct {
 	region       *core.RegionInfo // The region of the main balance effect. Relate mainPeerStat. srcStore -> dstStore
 	mainPeerStat *statistics.HotPeerStat
 
-	dstStore        *statistics.StoreLoadDetail
-	revertRegions   []*core.RegionInfo // The regions to hedge back effects. Relate revertPeersStat. dstStore -> srcStore
-	revertPeersStat []*statistics.HotPeerStat
+	dstStore       *statistics.StoreLoadDetail
+	revertRegion   *core.RegionInfo // The regions to hedge back effects. Relate revertPeerStat. dstStore -> srcStore
+	revertPeerStat *statistics.HotPeerStat
 
 	cachedPeersRate []float64
 
@@ -375,8 +375,8 @@ func (s *solution) calcPeersRate(rw statistics.RWType, dims ...int) {
 	for _, dim := range dims {
 		kind := statistics.GetRegionStatKind(rw, dim)
 		peersRate := s.mainPeerStat.GetLoad(kind)
-		for _, revertPeer := range s.revertPeersStat {
-			peersRate -= revertPeer.GetLoad(kind)
+		if s.revertPeerStat != nil {
+			peersRate -= s.revertPeerStat.GetLoad(kind)
 		}
 		s.cachedPeersRate[dim] = peersRate
 	}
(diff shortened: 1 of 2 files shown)


nolouch:
ptal @lhy1024

server/schedulers/hot_region.go
```diff
@@ -497,8 +537,43 @@ func (bs *balanceSolver) solve() []*operator.Operator {
 				bs.cur.dstStore = dstStore
 				bs.calcProgressiveRank()
 				tryUpdateBestSolution()
+
+				if searchRevertRegions && (bs.cur.progressiveRank >= -1 && bs.cur.progressiveRank <= 0) &&
```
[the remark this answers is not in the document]
lhy1024: I think we should support rank1 later.

server/schedulers/hot_region.go
```diff
@@ -521,12 +561,52 @@ func (bs *balanceSolver) solve() []*operator.Operator {
...
 }
 
+func (bs *balanceSolver) allowSearchRevertRegions() bool {
+	// The next solve is allowed to search-revert-regions only when the following conditions are met.
+	// * No best solution was found this time.
+	// * The progressiveRank of the best solution is -1.
+	// * The best solution contain revert regions.
+	return bs.best == nil || bs.best.progressiveRank >= -1 || len(bs.best.revertRegions) > 0
+}
+
```
lhy1024, line 609: For the multi operator, pending should be the result of taking the absolute value.

server/schedulers/hot_region.go
```diff
@@ -1094,12 +1194,24 @@ func (bs *balanceSolver) buildOperators() (ops []*operator.Operator, infl *stati
...
+			currentOp, typ, err = createOperator(revertRegion, dstStoreID, srcStoreID)
+			if err != nil {
+				break
+			}
+			bs.decorateOperator(currentOp, true, targetLabel, sourceLabel, typ, dim)
+			ops = append(ops, currentOp)
+			for j, load := range bs.cur.dstPeersStat[i].Loads {
+				infl.Loads[j] += load
+			}
+			infl.Count++
```
bufferflies, line 1213: should we need keep the influence disappear by atomicity?

lhy1024 (approved):
When the resources are sufficient, the two branches of master and multiop behave similarly, and the solution can be found (split.qps-threshold=500) and there is almost no scheduling after stabilization, and there is still scheduling after the solution cannot be found (split.qps-threshold=3000) ; When resources are insufficient, the number of peak scheduling of multiop is more

server/schedulers/hot_region.go
```diff
@@ -1095,18 +1207,22 @@ func (bs *balanceSolver) buildOperators() (ops []*operator.Operator, infl *stati
...
-		infl = &statistics.Influence{
-			Loads: append(bs.cur.srcPeerStat.Loads[:0:0], bs.cur.srcPeerStat.Loads...),
-			Count: 1,
+		for _, revertRegion := range bs.cur.revertRegions {
+			currentOp, typ, err = createOperator(revertRegion, dstStoreID, srcStoreID)
+			if err != nil {
+				break
+			}
+			bs.decorateOperator(currentOp, true, targetLabel, sourceLabel, typ, dim)
+			ops = append(ops, currentOp)
```
nolouch, line 1218: I wonder if the `RandBuckets` in`waiting_operator` will be influenced. The `GetOperator` may only get first one operator and miss the second one.

server/schedulers/hot_region.go
```diff
@@ -329,10 +331,15 @@ func (h *hotScheduler) balanceHotWriteRegions(cluster schedule.Cluster) []*opera
...
-	srcPeerStat *statistics.HotPeerStat
-	region      *core.RegionInfo
-	dstStore    *statistics.StoreLoadDetail
+	srcStore     *statistics.StoreLoadDetail
+	region       *core.RegionInfo // The region of the main balance effect. Relate mainPeerStat. srcStore -> dstStore
+	mainPeerStat *statistics.HotPeerStat
+
+	dstStore        *statistics.StoreLoadDetail
+	revertRegions   []*core.RegionInfo // The regions to hedge back effects. Relate revertPeersStat. dstStore -> srcStore
+	revertPeersStat []*statistics.HotPeerStat
```
nolouch, line 340:
```suggestion
	revertPeersStat *statistics.HotPeerStat
```
I prefer it like this. if you want to support more operators, may also need to make `mainPeerStat` to `[]*statistics.HotPeerStat` in the future.

nolouch (commented):
rest LGTM

lhy1024 (approved):

nolouch (approved):

Issue #4949, opened 2022-05-13
Improvement of hot scheduler when meets dimension conflict


lhy1024:

nolouch:
Please add more descriptions.

lhy1024:
![image](https://user-images.githubusercontent.com/19542290/200800692-f133d846-a4d5-473b-93db-c3fe9b2b88a8.png)

Performance improvement greater than 1% marked green.
performance regression greater than 1% marked red, stability evaluated by coefficient of variation, v2 greater than 0.05 and positive variation, marked red.
results are related to load and environment, marked yellow.
v2 is more balanced compared with v1, but the qps is not improved, marked blue; the scheduling convergence result is related to the split result, marked light blue.

PR #4962, opened 2022-05-16
scheduler: redesigned `progressiveRank` of the hot-region-scheduler

HunDunDM:

Issue Number: Ref #4949

The score is expanded from -3 ~ 0 to -4 ~ 1, and the previous design problem that the score of the first dimension is better and the second dimension is not worsened is -1, and the score of the first dimension is not worsened and the second dimension is better is -2. The solution still needs to take a score of -1 and below. The original 0 points become 0 and 1.

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -332,7 +332,10 @@ type solution struct {
 
 	// progressiveRank measures the contribution for balance.
 	// The smaller the rank, the better this solution is.
-	// If rank < 0, this solution makes thing better.
+	// If progressiveRank <= 0, this solution makes thing better.
+	// 0 indicates that this is a solution that cannot be used directly, but can be optimized.
+	// 1 indicates that this is a non-optimizable solution.
+	// See `calcProgressiveRank` for more about progressive rank.
 	progressiveRank int64
 }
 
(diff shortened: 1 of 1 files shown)


server/schedulers/hot_region.go
```diff
@@ -1029,12 +1037,12 @@ func (bs *balanceSolver) buildOperators() (ops []*operator.Operator, infl *stati
 	targetLabel := strconv.FormatUint(dstStoreID, 10)
 	dim := ""
 	switch bs.cur.progressiveRank {
-	case -3:
+	case -4:
 		dim = "all"
+	case -3, -1:
```
lhy1024, line 1042: Do we need to distinguish this two cases in metrics?

server/schedulers/hot_region.go
```diff
@@ -882,37 +895,32 @@ func (bs *balanceSolver) betterThan(old *solution) bool {
...
-					return secondCmp > 0
-				}
-				fallthrough
-			case -1: // firstPriorityDecRatio <= greatDecRatio
-				if firstCmp != 0 {
-					// prefer region with larger first priority rate, to converge faster
-					return firstCmp > 0
-				}
+			return secondCmp > 0
+		case -3, -1:
```
lhy1024, line 910: I think we need to split these two cases. -3 should be more similar with -2

HunDunDM: I wrote them separately at first, but the two seem to have the same processing logic at the moment.

server/schedulers/hot_region.go
```diff
@@ -778,41 +778,46 @@ func (bs *balanceSolver) checkDstByPriorityAndTolerance(maxLoad, expect *statist
...
 
 	isFirstBetter, isSecondBetter := bs.isBetter(bs.firstPriority), bs.isBetter(bs.secondPriority)
-	isFirstNotWorsened := !isFirstBetter && bs.isNotWorsened(bs.firstPriority)
-	isSecondNotWorsened := !isSecondBetter && bs.isNotWorsened(bs.secondPriority)
+	isFirstNotWorsened := isFirstBetter || bs.isNotWorsened(bs.firstPriority)
+	isSecondNotWorsened := isSecondBetter || bs.isNotWorsened(bs.secondPriority)
 	switch {
 	case isFirstBetter && isSecondBetter:
-		// If belong to the case, two dim will be more balanced, the best choice.
+		// If belonging to the case, all two dim will be more balanced, the best choice.
```
lhy1024, line 801:
```suggestion
		// If belonging to the case, both two dim will be more balanced, the best choice.
```

lhy1024, line 801:
```suggestion
		// If belonging to the case, both two dim will be more balanced, the best choice.
```

server/schedulers/hot_region.go
```diff
@@ -876,37 +900,32 @@ func (bs *balanceSolver) betterThan(old *solution) bool {
...
 			kind := statistics.GetRegionStatKind(statistics.Write, bs.firstPriority)
-			switch {
-			case bs.cur.srcPeerStat.GetLoad(kind) > old.srcPeerStat.GetLoad(kind):
-				return true
-			case bs.cur.srcPeerStat.GetLoad(kind) < old.srcPeerStat.GetLoad(kind):
-				return false
+			return bs.cur.srcPeerStat.GetLoad(kind) > old.srcPeerStat.GetLoad(kind)
+		}
+
+		firstCmp, secondCmp := bs.getRkCmpPriorities(old)
```
lhy1024, line 908: need to add comment `We will firstly consider ensuring converge faster, secondly reduce oscillation`

server/schedulers/hot_region.go
lhy1024: I think it will be friendly to read code and modify if split these cases.

nolouch (approved):
lgtm

lhy1024 (approved):

PR #4963, opened 2022-05-17
pkg: make speed estimation more accurate

rleungx:

Issue Number: Ref #4640.

This PR changes the way of calculating progress speed. After this PR, we use a sliding window to get the speed through the delta in a fixed time interval.

Now, the monitor becomes:
Screen Shot 2022-05-17 at 11 37 45 AM

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -23,8 +23,8 @@ import (
 	"github.com/tikv/pd/pkg/syncutil"
 )
 
-// speedStatisticalInterval is the speed calculation interval
-var speedStatisticalInterval = 5 * time.Minute
+// speedStatisticalWindow is the speed calculation window
+var speedStatisticalWindow = 10 * time.Minute
 
 // Manager is used to maintain the progresses we care about.
 type Manager struct {
@@ -43,10 +43,10 @@ func NewManager() *Manager {
 type progressIndicator struct {
 	total     float64
 	remaining float64
-	// we use a fixed interval to calculate the latest average speed.
-	lastTimeRemaining float64
-	lastSpeed         float64
-	lastTime          time.Time
+	// We use a fixed interval's history to calculate the latest average speed.
+	// It is a FIFO struct.
+	history   []float64
+	lastSpeed float64
 }
 
 // Reset resets the progress manager.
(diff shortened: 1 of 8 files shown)


pkg/progress/progress.go
```diff
@@ -83,18 +82,28 @@ func (m *Manager) UpdateProgressRemaining(progress string, remaining float64) {
 		if p.total < remaining {
 			p.total = remaining
 		}
-		if p.lastTimeRemaining < remaining {
-			p.lastTimeRemaining = remaining
+
+		// We use window size / update interval to get the history length.
+		historyLen := int(speedStatisticalWindow / updateInterval)
```
AndreMouche, line 87: Could we set `historyLen` as a constant(an attribute for `Manager`) once `Manager` created?

pkg/progress/progress.go
```diff
@@ -83,18 +82,28 @@ func (m *Manager) UpdateProgressRemaining(progress string, remaining float64) {
...
+		if len(p.history) > historyLen {
+			p.history = p.history[1:]
```
AndreMouche, line 89: Would this cause a GC problem? It seems once meet the historyLen, this operator will be done as frequently as the `UpdateProgress` called?  How about use a slice with constant length to implement the FIFOQueue?  For example, we assume the `historyLength` keep 5, and there is a slice::
1,2,3,4,5 (first=0,last:4)
when 6 comes, the slice changed to
6,2,3,4,5 (first=1,last=0)
and 7 comes: 
6,7,3,4,5 (first=2,last=1)

pkg/progress/progress.go
```diff
@@ -83,18 +89,26 @@ func (m *Manager) UpdateProgressRemaining(progress string, remaining float64) {
 		if p.total < remaining {
 			p.total = remaining
 		}
-		if p.lastTimeRemaining < remaining {
-			p.lastTimeRemaining = remaining
+
+		if p.history.Len() > p.historyLen {
```
AndreMouche, line 93:
```suggestion
		if p.history.Len() >= p.historyLen {
```

rleungx: This is on purpose because we will put one record when initialization.

AndreMouche: So the actually length limit of p.history is `historyLen+1`?

CabinfeverB: > So the actually length limit of p.history is `historyLen+1`?

IMO, `historyLen` is used for time window. For example, if updateInterval is equal to one minute, then historyLen is equal to 10, so it takes 11 points

rleungx: change it to windowLengthLimit

AndreMouche (approved):
LGTM

CabinfeverB (approved):

nolouch (approved):

Issue #4865, opened 2022-04-27
Add PD APIs for RawKV GC


AmoebaProtozoa:
## Development Task

Added new PD APIs to support RawKV GC as described by https://github.com/tikv/rfcs/pull/90

- [ ] Merge RFC https://github.com/tikv/rfcs/pull/90
- [x] Merge kvproto https://github.com/pingcap/kvproto/pull/908 https://github.com/pingcap/kvproto/pull/915

Merge PD RawKV GC API ( Split from combined PR: https://github.com/tikv/pd/pull/4866 )
- [x] Merge PD Storage https://github.com/tikv/pd/pull/4937
- [ ] Merge PD Client https://github.com/tikv/pd/pull/4943
- [ ] Merge PD Server

PR #4984, opened 2022-05-18
storage: Use etcd lease to manage safe point expiration

AmoebaProtozoa:
Signed-off-by: AmoebaProtozoa <8039876+AmoebaProtozoa@users.noreply.github.com>

Use etcd Lease feature to manage service safe point life time
Add revision related storage functions
Issue Number: ref https://github.com/tikv/pd/issues/4865
RFC: https://github.com/tikv/rfcs/pull/90

```commit-message
APIv2: Using ETCD Lease to manage service safe point expiration
```

### Check List

Tests

- Unit test

Code changes

- Has configuration change
No
No
- Has persistent data change
No

Side effects
No

### Release note

```release-note
None
```


server/storage/kv/etcd_kv.go
```diff
@@ -103,6 +109,55 @@ func (kv *etcdKVBase) Save(key, value string) error {
...
+			errs.ZapError(err))
+	}
+	log.Info("lease granted",
+		zap.Int64("lease-id", int64(grantResp.ID)),
+		zap.Int64("lease-timeout", ttlSeconds))
+	return grantResp.ID, nil
+}
+
+func (kv *etcdKVBase) SaveWithTTL(key, value string, ttlSeconds int64) error {
+	leaseID, err := kv.GrantLease(ttlSeconds)
```
AndreMouche, line 137: Could we use `etcdutil.EtcdKVPutWithTTL(context.TODO(),kv.client,key,value,ttlSeconds)` directly?

AmoebaProtozoa: Hmm, I was trying to put `SaveWithTTL` into a `SlowLogTxn` like other update methods for timeout and logging. But I guess that created more complexity.

Should we skip timeout / logging for this operation like other read methods?

tests/client/go.sum
```diff
@@ -24,6 +24,7 @@ github.com/VividCortex/mysqlerr v1.0.0/go.mod h1:xERx8E4tBhLvpjzdUyQiSfUxeMcATEQ
 github.com/Xeoncross/go-aesctr-with-hmac v0.0.0-20200623134604-12b17a7ff502 h1:L8IbaI/W6h5Cwgh0n4zGeZpVK78r/jBf9ASurHo9+/o=
 github.com/Xeoncross/go-aesctr-with-hmac v0.0.0-20200623134604-12b17a7ff502/go.mod h1:pmnBM9bxWSiHvC/gSWunUIyDvGn33EkP2CUjxFKtTTM=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
+github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 h1:JYp7IbQjafoB+tBA3gMyHYHrpOtNuDiK/uB5uXxq5wM=
```
AndreMouche, line 27: why?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/AlekSi/gocov-xml v1.0.0
 	github.com/BurntSushi/toml v0.3.1
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
```
AndreMouche, line 8: why?

go.sum
```diff
@@ -24,6 +24,7 @@ github.com/VividCortex/mysqlerr v1.0.0/go.mod h1:xERx8E4tBhLvpjzdUyQiSfUxeMcATEQ
 github.com/Xeoncross/go-aesctr-with-hmac v0.0.0-20200623134604-12b17a7ff502 h1:L8IbaI/W6h5Cwgh0n4zGeZpVK78r/jBf9ASurHo9+/o=
 github.com/Xeoncross/go-aesctr-with-hmac v0.0.0-20200623134604-12b17a7ff502/go.mod h1:pmnBM9bxWSiHvC/gSWunUIyDvGn33EkP2CUjxFKtTTM=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
+github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 h1:JYp7IbQjafoB+tBA3gMyHYHrpOtNuDiK/uB5uXxq5wM=
```
AndreMouche, line 27: ditto>

server/storage/endpoint/gc_key_space.go
```diff
@@ -51,7 +47,7 @@ type KeySpaceGCSafePointStorage interface {
 var _ KeySpaceGCSafePointStorage = (*StorageEndpoint)(nil)
 
 // SaveServiceSafePoint saves service safe point under given key-space.
-func (se *StorageEndpoint) SaveServiceSafePoint(spaceID string, ssp *ServiceSafePoint) error {
+func (se *StorageEndpoint) SaveServiceSafePoint(spaceID string, ssp *ServiceSafePoint, ttl int64) error {
```
AndreMouche, line 50: Could we store the ttl in the `ServiceSafePoint`, for example replace `ExpireAt`?

go.mod
AmoebaProtozoa: Hmm, it was required by generated `docs/swagger/docs.go`, but that file was not removed in `Make clean` 
So if I ran `go mod tidy` after a `make build`, this will be added automatically.

server/storage/kv/levedb_kv.go
```diff
@@ -72,6 +78,11 @@ func (kv *LevelDBKV) Save(key, value string) error {
 	return errors.WithStack(kv.Put([]byte(key), []byte(value), nil))
 }
 
+// SaveWithTTL not supported on LevelDBKV
+func (kv *LevelDBKV) SaveWithTTL(key, value string, ttlSeconds int64) error {
```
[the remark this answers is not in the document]
AmoebaProtozoa: Hmm, it was required by generated `docs/swagger/docs.go`, but that file was not removed in `Make clean`
So if I ran `go mod tidy` after a `make build`, this will be added automatically.

tests/client/go.sum
AmoebaProtozoa: Hmm, it was required by generated `docs/swagger/docs.go`, but that file was not removed in `Make clean`
So if I ran `go mod tidy` after a `make build`, this will be added automatically.

AmoebaProtozoa: I'll clean it up

server/storage/endpoint/gc_key_space.go
```diff
@@ -38,9 +33,10 @@ type KeySpaceGCSafePoint struct {
 // KeySpaceGCSafePointStorage defines the storage operations on KeySpaces' safe points
 type KeySpaceGCSafePointStorage interface {
 	// Service safe point interfaces.
-	SaveServiceSafePoint(spaceID string, ssp *ServiceSafePoint) error
+	// NOTE: field ServiceSafePoint.ExpiredAt will be ignored, use etcd's lease to manage lifetime instead.
```
JmPotato, line 36: Do we need a deprecated flag for this field to ensure no one will use it anymore?

server/storage/kv/etcd_kv.go
```diff
@@ -85,6 +91,34 @@ func (kv *etcdKVBase) LoadRange(key, endKey string, limit int) ([]string, []stri
 	return keys, values, nil
 }
 
+func (kv *etcdKVBase) SaveWithTTL(key, value string, ttlSeconds int64) error {
+	key = path.Join(kv.rootPath, key)
+	ctx, cancel := context.WithTimeout(kv.client.Ctx(), requestTimeout)
+	start := time.Now()
```
JmPotato, line 97:
```suggestion
	start := time.Now()
	ctx, cancel := context.WithTimeout(kv.client.Ctx(), requestTimeout)
```

server/storage/kv/kv.go
```diff
@@ -14,10 +14,16 @@
...
 
+// RevisionUnavailable is the value of unavailable revision,
+// when the kv does not exist (etcd_kv), or is not supported (mem_kv & leveldb_kv).
+const RevisionUnavailable = -1
+
 // Base is an abstract interface for load/save pd cluster data.
 type Base interface {
 	Load(key string) (string, error)
 	LoadRange(key, endKey string, limit int) (keys []string, values []string, err error)
+	LoadRevision(key string) (string, int64, error)
```
JmPotato, line 25: `revision` here is a very specific term that is related to etcd, is there a better way to avoid introducing it in this generic KV interface?

server/storage/endpoint/gc_key_space.go
AmoebaProtozoa: Probably not, we should probably still store when safepoint will be expired, so if and when client need this information, we can retrieve directly from ServiceSafePoint, instead of looking it up by lease.  

`ExpiredAt` field will be calculated at Service level, using `TTL` client provided with server's `time.now`, here we just simply store it. 

I'll modify the comment to make it more clear

server/storage/kv/kv.go
AmoebaProtozoa: I tried a little bit but I don't think there's a way to avoid modifying `Base` interface unless we change `etcdKVBase` to exported type.
I'm just not sure if thats the way to go, feels kinda hacky. What do you think? @JmPotato

PR #5010, opened 2022-05-23
checker: only temp state should be added in waiting list

### LINKED ISSUE #4920: PD CPU 100% on an idle cluster
### What did you do?

Start a playground cluster, load some data:

```
tiup playground --tiflash 0
tiup br restore full -s 's3://bank/shema-full?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http://minio.pingcap.net:9000&force-path-style=true/' -u 'http://127.0.0.1:2379/'
```

Then the cluster is idle, but I find PD CPU is nearly 100%

PD CPU usage should be low.

### What did you see instead?

Note the CPU usage below:

image
image

Profile shows the CPU is used in the `patolRegions` function

image

I believe it's caused by this loop:

https://github.com/tikv/pd/blob/d20eacc1212a558cde18ebc0d873236dee46d1fc/server/cluster/coordinator.go#L112-L130

Maybe **the actual work takes more time then the ticker interval**, so it repeat the execution endlessly... and cost 100% CPU

6.0.0

rleungx:

Issue Number: Close #4920.

This PR stops adding regions, which cannot be fixed due to no proper store instead of a temporary store state, to the waiting list.

```commit-message
```

### Check List

Tests

- Manual test

Screen Shot 2022-05-23 at 2 23 33 PM

It is not affected by the waiting list anymore. Here is the CPU cost at the beginning:
Screen Shot 2022-05-23 at 2 26 20 PM

After BR is finished:
Screen Shot 2022-05-23 at 2 29 22 PM

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/replica_checker.go
+++ b/server/schedule/checker/replica_checker.go
@@ -157,11 +157,13 @@ func (r *ReplicaChecker) checkMakeUpReplica(region *core.RegionInfo) *operator.O
 	}
 	log.Debug("region has fewer than max replicas", zap.Uint64("region-id", region.GetID()), zap.Int("peers", len(region.GetPeers())))
 	regionStores := r.cluster.GetRegionStores(region)
-	target := r.strategy(region).SelectStoreToAdd(regionStores)
+	target, filterByTempState := r.strategy(region).SelectStoreToAdd(regionStores)
 	if target == 0 {
 		log.Debug("no store to add replica", zap.Uint64("region-id", region.GetID()))
 		checkerCounter.WithLabelValues("replica_checker", "no-target-store").Inc()
-		r.regionWaitingList.Put(region.GetID(), nil)
+		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
+			r.regionWaitingList.Put(region.GetID(), nil)
+		}
 		return nil
 	}
 	newPeer := &metapb.Peer{StoreId: target}
@@ -210,7 +212,7 @@ func (r *ReplicaChecker) checkLocationReplacement(region *core.RegionInfo) *oper
 		checkerCounter.WithLabelValues("replica_checker", "all-right").Inc()
 		return nil
 	}
-	newStore := strategy.SelectStoreToImprove(regionStores, oldStore)
+	newStore, _ := strategy.SelectStoreToImprove(regionStores, oldStore)
 	if newStore == 0 {
 		log.Debug("no better peer", zap.Uint64("region-id", region.GetID()))
 		checkerCounter.WithLabelValues("replica_checker", "not-better").Inc()
@@ -240,12 +242,14 @@ func (r *ReplicaChecker) fixPeer(region *core.RegionInfo, storeID uint64, status
 	}
 
 	regionStores := r.cluster.GetRegionStores(region)
-	target := r.strategy(region).SelectStoreToFix(regionStores, storeID)
+	target, filterByTempState := r.strategy(region).SelectStoreToFix(regionStores, storeID)
 	if target == 0 {
 		reason := fmt.Sprintf("no-store-%s", status)
 		checkerCounter.WithLabelValues("replica_checker", reason).Inc()
-		r.regionWaitingList.Put(region.GetID(), nil)
 		log.Debug("no best store to add replica", zap.Uint64("region-id", region.GetID()))
+		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
+			r.regionWaitingList.Put(region.GetID(), nil)
+		}
 		return nil
 	}
 	newPeer := &metapb.Peer{StoreId: target}
--- a/server/schedule/checker/replica_checker.go
+++ b/server/schedule/checker/replica_checker.go
@@ -161,7 +161,7 @@ func (r *ReplicaChecker) checkMakeUpReplica(region *core.RegionInfo) *operator.O
 	if target == 0 {
 		log.Debug("no store to add replica", zap.Uint64("region-id", region.GetID()))
 		checkerCounter.WithLabelValues("replica_checker", "no-target-store").Inc()
-		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
+		if filterByTempState {
 			r.regionWaitingList.Put(region.GetID(), nil)
 		}
 		return nil
@@ -247,7 +247,7 @@ func (r *ReplicaChecker) fixPeer(region *core.RegionInfo, storeID uint64, status
 		reason := fmt.Sprintf("no-store-%s", status)
 		checkerCounter.WithLabelValues("replica_checker", reason).Inc()
 		log.Debug("no best store to add replica", zap.Uint64("region-id", region.GetID()))
-		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
+		if filterByTempState {
 			r.regionWaitingList.Put(region.GetID(), nil)
 		}
 		return nil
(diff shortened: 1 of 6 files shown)


server/schedule/checker/replica_checker.go
```diff
@@ -157,11 +157,13 @@ func (r *ReplicaChecker) checkMakeUpReplica(region *core.RegionInfo) *operator.O
 	}
 	log.Debug("region has fewer than max replicas", zap.Uint64("region-id", region.GetID()), zap.Int("peers", len(region.GetPeers())))
 	regionStores := r.cluster.GetRegionStores(region)
-	target := r.strategy(region).SelectStoreToAdd(regionStores)
+	target, filterByTempState := r.strategy(region).SelectStoreToAdd(regionStores)
 	if target == 0 {
 		log.Debug("no store to add replica", zap.Uint64("region-id", region.GetID()))
 		checkerCounter.WithLabelValues("replica_checker", "no-target-store").Inc()
-		r.regionWaitingList.Put(region.GetID(), nil)
+		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
```
AndreMouche, line 164: I think check the `list.Len` before `Put` to ensure the list not overflow without a lock is not threadsafe, And fortunally, we have already check the length in the function `regionWaitingList.Put`. 
```suggestion
		if filterByTempState  {
```

AndreMouche: However, I've found that here `regionWaitingList` is a LRU and it will remove the earlist item when overflow, But here we skip the latest item. So Does this matters?

server/schedule/checker/replica_checker.go
```diff
@@ -240,12 +242,14 @@ func (r *ReplicaChecker) fixPeer(region *core.RegionInfo, storeID uint64, status
...
 
 	regionStores := r.cluster.GetRegionStores(region)
-	target := r.strategy(region).SelectStoreToFix(regionStores, storeID)
+	target, filterByTempState := r.strategy(region).SelectStoreToFix(regionStores, storeID)
 	if target == 0 {
 		reason := fmt.Sprintf("no-store-%s", status)
 		checkerCounter.WithLabelValues("replica_checker", reason).Inc()
-		r.regionWaitingList.Put(region.GetID(), nil)
 		log.Debug("no best store to add replica", zap.Uint64("region-id", region.GetID()))
+		if filterByTempState && r.regionWaitingList.Len() <= DefaultCacheSize {
```
AndreMouche, line 250: ditto

server/schedule/checker/replica_strategy.go
```diff
@@ -70,33 +70,37 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 
 	isolationComparer := filter.IsolationComparer(s.locationLabels, coLocationStores)
 	strictStateFilter := &filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true}
-	target := filter.NewCandidates(s.cluster.GetStores()).
+	targetCandidate := filter.NewCandidates(s.cluster.GetStores()).
```
AndreMouche, line 73: How about 
```suggestion
	targetCandidateWithTemp := filter.NewCandidates(s.cluster.GetStores()).
```

rleungx: I think it doesn't matter, `targetCandidateWithTemp` is too long.

server/schedule/checker/replica_strategy.go
```diff
@@ -70,33 +70,37 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
...
+		Sort(filter.RegionScoreComparer(s.cluster.GetOpts()))     // less region score is better
+	candidateLen := targetCandidate.Len()
+	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).PickFirst() // the filter does not ignore temp states
 	if target == nil {
-		return 0
+		if candidateLen == 0 {
+			return 0, false
+		}
+		return 0, true // filter by temporary states
 	}
```
JmPotato, line 84: Maybe this part can be rewritten like this:

```go
if candidateLen == 0 {
	return 0, false
}
if target == nil {
	return 0, true // filter by temporary states
}
```

server/schedule/checker/replica_strategy.go
```diff
@@ -70,33 +70,37 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
...
-		Sort(isolationComparer).Reverse().Top(isolationComparer).        // greater isolation score is better
-		Sort(filter.RegionScoreComparer(s.cluster.GetOpts())).           // less region score is better
-		FilterTarget(s.cluster.GetOpts(), strictStateFilter).PickFirst() // the filter does not ignore temp states
+		Sort(isolationComparer).Reverse().Top(isolationComparer). // greater isolation score is better
+		Sort(filter.RegionScoreComparer(s.cluster.GetOpts()))     // less region score is better
+	candidateLen := targetCandidate.Len()
+	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).PickFirst() // the filter does not ignore temp states
+	if candidateLen == 0 {
+		return 0, false
+	}
```
JmPotato, line 81:
```suggestion
	if targetCandidate.Len() == 0 {
		return 0, false
	}
	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).PickFirst() // the filter does not ignore temp states
```

JmPotato (approved):

server/schedule/checker/replica_strategy.go
AndreMouche: OK

server/schedule/checker/rule_checker.go
```diff
@@ -161,10 +161,12 @@ func (c *RuleChecker) fixRulePeer(region *core.RegionInfo, fit *placement.Region
 func (c *RuleChecker) addRulePeer(region *core.RegionInfo, rf *placement.RuleFit) (*operator.Operator, error) {
 	checkerCounter.WithLabelValues("rule_checker", "add-rule-peer").Inc()
 	ruleStores := c.getRuleFitStores(rf)
-	store := c.strategy(region, rf.Rule).SelectStoreToAdd(ruleStores)
+	store, filterByTempState := c.strategy(region, rf.Rule).SelectStoreToAdd(ruleStores)
 	if store == 0 {
 		checkerCounter.WithLabelValues("rule_checker", "no-store-add").Inc()
-		c.regionWaitingList.Put(region.GetID(), nil)
+		if filterByTempState && c.regionWaitingList.Len() <= DefaultCacheSize {
```
AndreMouche, line 167:
```suggestion
		if filterByTempState  {
```

server/schedule/checker/rule_checker.go
```diff
@@ -179,10 +181,12 @@ func (c *RuleChecker) addRulePeer(region *core.RegionInfo, rf *placement.RuleFit
 // The peer's store may in Offline or Down, need to be replace.
 func (c *RuleChecker) replaceUnexpectRulePeer(region *core.RegionInfo, rf *placement.RuleFit, fit *placement.RegionFit, peer *metapb.Peer, status string) (*operator.Operator, error) {
 	ruleStores := c.getRuleFitStores(rf)
-	store := c.strategy(region, rf.Rule).SelectStoreToFix(ruleStores, peer.GetStoreId())
+	store, filterByTempState := c.strategy(region, rf.Rule).SelectStoreToFix(ruleStores, peer.GetStoreId())
 	if store == 0 {
 		checkerCounter.WithLabelValues("rule_checker", "no-store-replace").Inc()
-		c.regionWaitingList.Put(region.GetID(), nil)
+		if filterByTempState && c.regionWaitingList.Len() <= DefaultCacheSize {
```
AndreMouche, line 187:
```suggestion
		if filterByTempState  {
```

AndreMouche (approved):
LGTM

PR #5029, opened 2022-05-25
cluster: calculate speed until cluster is prepared

### LINKED ISSUE #5025: Offline left time and scale speed will jump in 10m after change pd leader
### What did you do?

Down pd leader during scale out.

Left time and speed should be  stable and reaction actual situation.

### What did you see instead?
image

image

./pd-server -V
 Release Version: v6.1.0-nightly
Edition: Community
Git Commit Hash: b3eeaae3ea3d0e058d7363e134a8c84737f4798f
Git Branch: heads/refs/tags/v6.1.0-nightly
UTC Build Time:  2022-05-24 06:13:20

rleungx:

Issue Number: Close #5025

The reason why the speed isn't accurate is that when the leader changes, the check stores won't wait for all regions' heartbeats to finish. So it will push a small store region size into the statistics window. Then in the later 10 mins, PD will use the latest store region size, which could be big after the heartbeat finish, minus that one to get the delta for calculating the speed. After the 10min passed away, the first value of the statistics window will be removed, then the speed becomes normal.

```commit-message
```

### Check List

Tests

- Unit test
- Manual test 

After：
origin_img_v2_41cbe360-ffc6-474d-816d-ee6d38e2781g
middle_img_v2_4abd4eba-a75d-42d7-9e18-9ea394ca349g

Related changes

### Release note

```release-note
Fix the issue that the store online speed is not accurate when the PD leader transfers.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1449,7 +1449,7 @@ func (c *RaftCluster) checkStores() {
 						zap.Stringer("store", store.GetMeta()),
 						errs.ZapError(err))
 				}
-			} else {
+			} else if c.IsPrepared() {
 				threshold := c.getThreshold(stores, store)
 				log.Debug("store serving threshold", zap.Uint64("store-id", storeID), zap.Float64("threshold", threshold))
 				regionSize := float64(store.GetRegionSize())
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -496,6 +496,13 @@ func (c *RaftCluster) GetOperatorController() *schedule.OperatorController {
 	return c.coordinator.opController
 }
 
+// SetPrepared set the prepare check to prepared. Only for test purpose.
+func (c *RaftCluster) SetPrepared() {
+	c.coordinator.prepareChecker.Lock()
+	defer c.coordinator.prepareChecker.Unlock()
+	c.coordinator.prepareChecker.prepared = true
+}
+
 // GetRegionScatter returns the region scatter.
 func (c *RaftCluster) GetRegionScatter() *schedule.RegionScatterer {
 	return c.coordinator.regionScatterer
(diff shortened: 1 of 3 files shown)


tests/server/api/api_test.go
```diff
@@ -685,8 +683,16 @@ func (s *testProgressSuite) TestPreparingProgress(c *C) {
 	output = sendRequest(c, leader.GetAddr()+"/pd/api/v1/stores/progress?id=4", http.MethodGet, http.StatusNotFound)
 	c.Assert(strings.Contains((string(output)), "no progress found for the given store ID"), IsTrue)
 
+	// is not prepared
 	time.Sleep(2 * time.Second)
+	output = sendRequest(c, leader.GetAddr()+"/pd/api/v1/stores/progress?action=preparing", http.MethodGet, http.StatusNotFound)
+	c.Assert(strings.Contains((string(output)), "no progress found for the action"), IsTrue)
+	output = sendRequest(c, leader.GetAddr()+"/pd/api/v1/stores/progress?id=4", http.MethodGet, http.StatusNotFound)
+	c.Assert(strings.Contains((string(output)), "no progress found for the given store ID"), IsTrue)
```
JmPotato, line 691: Do we need to output something like "the given store's progress is calculating"?

lhy1024 (approved):

JmPotato (approved):

CabinfeverB (approved):

server/cluster/cluster.go
```diff
@@ -496,6 +496,13 @@ func (c *RaftCluster) GetOperatorController() *schedule.OperatorController {
 	return c.coordinator.opController
 }
 
+// SetPrepared set the prepare check to prepared. Only for test purpose.
+func (c *RaftCluster) SetPrepared() {
+	c.coordinator.prepareChecker.Lock()
+	defer c.coordinator.prepareChecker.Unlock()
+	c.coordinator.prepareChecker.prepared = true
+}
```
AndreMouche, line 504: This function uses prepareCheck's lock to lock the update operator for prepareChecker's item, could you please make a new function for SetPrepared for `prepareChecker` directly?  Using its lock in `prepareChecker` can help us to protect the usage of lock and avoid deadlock.

rleungx: It's only used in tests.

PR #5038, opened 2022-05-26
cluster: calculate offline speed until cluster is prepared

rleungx:

Issue Number: Ref #5025

Offline progress has the same problem which will change the ETA to N/A. This PR will check if the cluster is prepared before the offline progress calculation.

```commit-message
```

### Check List

Tests

- Unit test
- Manual test

Screen Shot 2022-05-27 at 10 11 15 AM

Related changes

### Release note

```release-note
Fix the issue that the store offline speed is N/A when the PD leader transfers.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1485,7 +1485,9 @@ func (c *RaftCluster) checkStores() {
 		offlineStore := store.GetMeta()
 		id := offlineStore.GetId()
 		regionSize := c.core.GetStoreRegionSize(id)
-		c.updateProgress(id, store.GetAddress(), removingAction, float64(regionSize), float64(regionSize), false /* dec */)
+		if c.IsPrepared() {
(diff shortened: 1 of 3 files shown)


server/cluster/cluster.go
```diff
@@ -1485,7 +1485,9 @@ func (c *RaftCluster) checkStores() {
 		offlineStore := store.GetMeta()
 		id := offlineStore.GetId()
 		regionSize := c.core.GetStoreRegionSize(id)
-		c.updateProgress(id, store.GetAddress(), removingAction, float64(regionSize), float64(regionSize), false /* dec */)
+		if c.IsPrepared() {
```
JmPotato, line 1488: Is it better to move this check inside the `updateProgress` function?

rleungx: We can avoid calculating the threshold in the preparing stage.

JmPotato (approved):

PR #5045, opened 2022-05-26
unsafe recovery: Tombstone tiflash learners which have the latest log

### LINKED ISSUE #5044: Consider the case that Tiflash is the only left store for unsafe recovery
## Bug Report

For the case that Tiflash is the only left store for unsafe recovery, current online unsafe recovery would ignore the Tiflash, and finish with doing nothing directly. 
As Tiflash can't be Raft leader, so for this case we need tombstone the peer of Tiflash, and create empty region for the range.

master

Connor1996:
Signed-off-by: Connor1996 <zbk602423539@gmail.com>

Issue Number: Close #5044 

```commit-message
tombstone tiflash learners which have the latest log
```

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
None
```


server/cluster/unsafe_recovery_controller.go
```diff
@@ -823,21 +871,49 @@ func (u *unsafeRecoveryController) buildUpFromReports() (*regionTree, map[uint64
 	return newestRegionTree, peersMap, nil
 }
 
-func (u *unsafeRecoveryController) generateForceLeaderPlan(newestRegionTree *regionTree, peersMap map[uint64][]*regionItem, forCommitMerge bool) bool {
+func (u *unsafeRecoveryController) selectLeader(peersMap map[uint64][]*regionItem, region *metapb.Region) *regionItem {
+	var leader *regionItem
+	for _, peer := range peersMap[region.GetId()] {
+		if leader == nil || leader.IsRaftStale(peer, u) {
```
rleungx, line 877: What if only have the TiFlash learner?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -657,27 +693,39 @@ func (r *regionItem) IsEpochStale(other *regionItem) bool {
...
-				return true
-			} else if rs.GetHardState().GetCommit() == os.GetHardState().GetCommit() {
-				// better use voter rather than learner
-				for _, peer := range r.Region().GetPeers() {
-					if peer.StoreId == r.storeID {
-						if peer.Role == metapb.PeerRole_DemotingVoter || peer.Role == metapb.PeerRole_Learner {
-							return true
-						}
+func (r *regionItem) IsRaftStale(origin *regionItem, u *unsafeRecoveryController) bool {
+	cmps := []func(a, b *regionItem) int{
```
rleungx, line 697: Why not directly return bool?

server/cluster/unsafe_recovery_controller.go
```diff
@@ -823,21 +871,49 @@ func (u *unsafeRecoveryController) buildUpFromReports() (*regionTree, map[uint64
...
+	newestRegionTree.tree.Ascend(func(item btree.Item) bool {
+		region := item.(*regionItem).Region()
+		if !u.canElectLeader(region, false) {
+			leader := u.selectLeader(peersMap, region)
+			if leader == nil {
+				u.err = errors.Errorf("can't select leader for region %v", region)
+				return false
+			}
+			storeID := leader.storeID
+			if u.cluster.GetStore(storeID).IsTiFlash() {
```
rleungx, line 899: Only a peer is in TiFlash and selected as a leader will be tombstone?

server/cluster/unsafe_recovery_controller.go
Connor1996: it continue the cmp range only when the return value is zero, if it's not zero it return v < 0

server/cluster/unsafe_recovery_controller.go
Connor1996: yes, as we can't make Tiflash to be the leader

Connor1996: more essentially, Tiflash can't generate snapshot

server/cluster/unsafe_recovery_controller.go
```diff
@@ -823,21 +871,49 @@ func (u *unsafeRecoveryController) buildUpFromReports() (*regionTree, map[uint64
...
-		var leader *regionItem
-		for _, peer := range peersMap[region.GetId()] {
-			if leader == nil || leader.IsRaftStale(peer) {
-				leader = peer
+	newestRegionTree.tree.Ascend(func(item btree.Item) bool {
+		region := item.(*regionItem).Region()
+		if !u.canElectLeader(region, false) {
+			leader := u.selectLeader(peersMap, region)
+			if leader == nil {
+				u.err = errors.Errorf("can't select leader for region %v", region)
```
rleungx, line 895: Will it print the key?

v01dstar (approved):

rleungx (approved):

nolouch (approved):

PR #5050, opened 2022-05-27
server/cluster: store the min version of stores as the cluster_version

### LINKED ISSUE #5049: server/cluster：store the min version of stores as cluster version
## Bug Report
On L1860 : the clusterVersion = max(curVersion,minVersion of all store), this makes the clusterVersion depends on the sequence of `PutStore` happens. 
[https://github.com/tikv/pd/blob/aafc77dd7c583a37b9b27821beb59594e1f7ccce/server/cluster/cluster.go#L1840-L1871](https://github.com/tikv/pd/blob/aafc77dd7c583a37b9b27821beb59594e1f7ccce/server/cluster/cluster.go#L1840-L1871)
### What did you do?
When we update the version of store in PutStore, for example,
3 stores with version [5.0.5,5.0.3,5.0.1].

- PutStoreWithVersion(5.0.3)

> success

- GetClusterVersion    

> get 5.0.3

- PutStoreWithVersion 5.0.1 

> success

- GetClusterVersion 
> get 5.0.3 
_**while here we expect to get 5.0.1 instead.**_

- PutStoreWithVersion 5.0.5 
> success
- GetClusterVersion 

> get 5.0.3 
_**while here we expect to get 5.0.1 instead.**_

master

AndreMouche:

Issue Number: Close #5049 

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024 (approved):

server/cluster/cluster.go
```diff
@@ -1879,19 +1879,21 @@ func (c *RaftCluster) onStoreVersionChangeLocked() {
...
-			log.Error("cluster version changed by API at the same time")
-		}
-		err := c.opt.Persist(c.storage)
-		if err != nil {
-			log.Error("persist cluster version meet error", errs.ZapError(err))
-		}
-		log.Info("cluster version changed",
-			zap.Stringer("old-cluster-version", clusterVersion),
-			zap.Stringer("new-cluster-version", minVersion))
+	if !c.opt.CASClusterVersion(clusterVersion, minVersion) {
```
bufferflies, line 1886: How about add some retry ?

AndreMouche: I'd like to keep the origin logic, which means if the cluster version changed by API at the same time, we should keep the version of API

server/cluster/cluster.go
```diff
@@ -1879,19 +1879,20 @@ func (c *RaftCluster) onStoreVersionChangeLocked() {
 	failpoint.Inject("versionChangeConcurrency", func() {
 		time.Sleep(500 * time.Millisecond)
 	})
+	if minVersion == nil || clusterVersion.Equal(*minVersion) {
```
rleungx, line 1882: What if the cluster version is greater than minVersion?

AndreMouche: The case is :
1. the cluster version 5.0.5
2. put a `tikv` with v4.0.5 will be failed
3. put a `tikv` with v5.0.1 will be success and the cluster version should be v5.0.1
4. put a `tikv` with v5.0.2 should be success and the cluster version should be still v5.0.1

rleungx: So when we use API to change the cluster version, it might be changed to another value?

rleungx: The previous logic is when we set the cluster to a higher version, it won't be changed. But right now, it might be changed to another value.

rleungx (approved):

PR #5070, opened 2022-05-30
server/grpc_service: make update gc_safepoint concurrently safe

### LINKED ISSUE #5018: server/grpc_service: updateGCSafepoint may lost the max safepoint
## Bug Report
In the function `UpdateGCSafePoint`, if another thread receive and update the safepoint after `LoadGCSafePoint` and before `SaveGCSafePoint`, we may miss the max safepoint . 
[https://github.com/tikv/pd/blob/efb4c9f687620afd0654296bfcd932515fce5a46/server/grpc_service.go#L1339-L1349](https://github.com/tikv/pd/blob/efb4c9f687620afd0654296bfcd932515fce5a46/server/grpc_service.go#L1339-L1349)
### What did you do?
For example:
- client1 LoadGCSafePoint (value=5)
- client2 LoadGCSafePoint (value=5)
- client 2 SaveGCSafePoint with value = 10 since it is bigger than oldValue 5.
- client 1 SaveGCSafePoint with value=7 since it is bigger than the oldValue 5 it get.

Here the max safepoint 10 was lost.

AndreMouche:

Issue Number: Close #5018 

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/grpc_service.go
```diff
@@ -1335,22 +1334,17 @@ func (s *GrpcServer) UpdateGCSafePoint(ctx context.Context, request *pdpb.Update
...
-	// Only save the safe point if it's greater than the previous one
+	// safepoint updated
 	if newSafePoint > oldSafePoint {
-		if err := storage.SaveGCSafePoint(newSafePoint); err != nil {
-			return nil, err
-		}
 		log.Info("updated gc safe point",
 			zap.Uint64("safe-point", newSafePoint))
-	} else if newSafePoint < oldSafePoint {
+	} else {
```
bufferflies, line 1347: what does it do if newSafePoint is eqaul to oldSafePoint?

server/server.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/pingcap/sysutil"
+	gc_server "github.com/tikv/pd/cmd/gc/server"
```
nolouch, line 42: I think server do not relay on `cmd` package, maybe move to `pkg`; see https://github.com/golang-standards/project-layout/blob/master/README_zh.md

server/grpc_service.go
AndreMouche: nothing

server/server.go
AndreMouche: ok~

server/grpc_service.go
```diff
@@ -1335,22 +1334,17 @@ func (s *GrpcServer) UpdateGCSafePoint(ctx context.Context, request *pdpb.Update
...
 
-	newSafePoint := request.SafePoint
-
-	// Only save the safe point if it's greater than the previous one
+	// safepoint updated
 	if newSafePoint > oldSafePoint {
-		if err := storage.SaveGCSafePoint(newSafePoint); err != nil {
-			return nil, err
-		}
 		log.Info("updated gc safe point",
```
rleungx, line 1345: How about putting the log into where the safe point is being updated.

AndreMouche: I try to keep the origin logic Because  I was considering  that if someone or someplace grep the log with the source file's name

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,87 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gc
+
+import (
+	"sync"
+	"testing"
+
+	. "github.com/pingcap/check"
```
JmPotato, line 21: What about using testify here? Just like https://github.com/tikv/pd/pull/5061.

bufferflies (commented):
rest LGTM

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/tikv/pd/server/storage/endpoint"
+	"github.com/tikv/pd/server/storage/kv"
+)
+
+func newGCStorage() endpoint.GCSafePointStorage {
+	return endpoint.NewStorageEndpoint(kv.NewMemoryKV(), nil)
+}
+
+func TestGCSafePointUpdateSequentially(t *testing.T) {
+	gcSafePointManager := newGCSafePointManager(newGCStorage())
```
JmPotato, line 31: We could create a `re := require.New(t)` here and use `re.NoError(err)` later. Just like:

https://github.com/tikv/pd/blob/3362ce2de17945b5ee5466d0b7a3ce00f2420682/pkg/apiutil/apiutil_test.go#L27-L43

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,78 @@
...
+	require.NoError(t, err)
+	require.Equal(t, oldSafePoint, safePoint)
+	curSafePoint, err = gcSafePointManager.LoadGCSafePoint()
+	require.NoError(t, err)
+	// current safePoint should not change since the update value was smaller
+	require.Equal(t, curSafePoint, safePoint)
+}
+
+func TestGCSafePointUpdateCurrently(t *testing.T) {
+	gcSafePointManager := newGCSafePointManager(newGCStorage())
```
JmPotato, line 59: Ditto.

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func TestGCSafePointUpdateSequentially(t *testing.T) {
+	gcSafePointManager := newGCSafePointManager(newGCStorage())
+	curSafePoint := uint64(0)
+	// update gc safePoint with asc value.
+	for id := 10; id < 20; id++ {
+		safePoint, err := gcSafePointManager.LoadGCSafePoint()
+		require.NoError(t, err)
+		require.Equal(t, safePoint, curSafePoint)
```
JmPotato, line 37: This function's signature is `func Equal(expected interface{}, actual interface{})`, so the expected value should be placed on the first argument.

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/tikv/pd/server/storage/kv"
+)
+
+func newGCStorage() endpoint.GCSafePointStorage {
+	return endpoint.NewStorageEndpoint(kv.NewMemoryKV(), nil)
+}
+
+func TestGCSafePointUpdateSequentially(t *testing.T) {
+	gcSafePointManager := newGCSafePointManager(newGCStorage())
+	req := require.New(t)
```
JmPotato, line 32: Better be consistent with other packages' styles.

```suggestion
	re := require.New(t)
```

server/gc/safepoint_test.go
```diff
@@ -0,0 +1,80 @@
...
+	req.NoError(err)
+	// current safePoint should not change since the update value was smaller
+	req.Equal(safePoint, curSafePoint)
+}
+
+func TestGCSafePointUpdateCurrently(t *testing.T) {
+	gcSafePointManager := newGCSafePointManager(newGCStorage())
+	maxSafePoint := uint64(1000)
+	wg := sync.WaitGroup{}
+	req := require.New(t)
```
JmPotato, line 63: Ditto.

nolouch (approved):

server/grpc_service.go
```diff
@@ -1335,19 +1334,13 @@ func (s *GrpcServer) UpdateGCSafePoint(ctx context.Context, request *pdpb.Update
 		return &pdpb.UpdateGCSafePointResponse{Header: s.notBootstrappedHeader()}, nil
 	}
 
-	var storage endpoint.GCSafePointStorage = s.storage
-	oldSafePoint, err := storage.LoadGCSafePoint()
+	newSafePoint := request.SafePoint
```
JmPotato, line 1337:
```suggestion
	newSafePoint := request.GetSafePoint()
```

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

PR #5080, opened 2022-05-31
*: fix the wrong pending status

### LINKED ISSUE #5095: `/regions/replicated` may return the wrong results
## Bug Report

Use `/regions/replicated` to query if the regions in the specified range are all replicated. But this API only checks the regions inside the waiting list which is not accurate and may return the wrong results. See https://github.com/tikv/pd/blob/fc55b446f9ad0b32119e86b68e1353ba78c928aa/server/api/region.go#L283 

It is because the waiting list is shared by some checkers and saves 1000 regions at most. And in the recent modification, we only put the regions, which are filtered by the temporary store state, into it. Then, the result could be more misleading.

rleungx:

Issue Number: Close #5095.

This PR adds a `pendingList` in the `ruleChecker` to record the region that cannot be scheduled due to some long-term states. e.g., no enough stores, no enough space and etc. Once these regions can be scheduled or filtered by the temp state, they will be removed from the `pendingList`.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Related changes

### Release note

```release-note
Fix the issue that `/regions/replicated` may return the wrong status
```


server/schedule/checker/rule_checker.go
```diff
@@ -166,6 +172,9 @@ func (c *RuleChecker) addRulePeer(region *core.RegionInfo, rf *placement.RuleFit
 		checkerCounter.WithLabelValues("rule_checker", "no-store-add").Inc()
 		if filterByTempState {
 			c.regionWaitingList.Put(region.GetID(), nil)
+			c.pendingList.Remove(region.GetID())
+		} else {
+			c.pendingList.Put(region.GetID(), nil)
 		}
```
JmPotato, line 178: These operations can be extracted as a function since they always work together.

JmPotato (commented):
The rest LGTM.

PR #5084, opened 2022-06-01
config: the defualt value of `max-merge-region-keys` is related with `max-merge-region-size`

### LINKED ISSUE #5083: make merge config is easy.
## Enhancement Task
In past, PD have two configuration about merge behavior such as `merge-max-size` and `merge-max-key`. It should be simply to set them in dynamic region scene. 

Like TIkv configuration:
the default value of `region-split-keys` is the 10000 of the `region-max-size`. 

https://github.com/tikv/tikv/blob/f7edbcf610cd6bb5c9040317edd3260188a6e87d/components/raftstore/src/coprocessor/config.rs#L102-L106

bufferflies:

Issue Number: Close #5083

`max-merge-region-keys` should be zero rather than 20*10000, and `GetMaxmergeRegionKeys` will return 10000 times of `max-merge-region-size` if  `max-merge-region-keys` is not setted.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has configuration change
-

Side effects

Related changes

### Release note

```release-note
the default value of  `merge-max-keys` is 10000 times of`merge-max-size`.
```


server/config/config.go
```diff
@@ -911,6 +908,14 @@ func (c *ScheduleConfig) migrateConfigurationMap() map[string][2]*bool {
 	}
 }
 
+// GetMaxMergeRegionKeys returns the max merge keys.
+func (c *ScheduleConfig) GetMaxMergeRegionKeys() uint64 {
+	if keys := c.MaxMergeRegionKeys; keys != 0 {
+		return keys
+	}
+	return c.MaxMergeRegionSize * 10000
```
lhy1024, line 916: 10000？

tests/pdctl/config/config_test.go
```diff
@@ -129,6 +129,23 @@ func (s *configTestSuite) TestConfig(c *C) {
 	c.Assert(json.Unmarshal(output, &scheduleCfg), IsNil)
 	c.Assert(&scheduleCfg, DeepEquals, svr.GetScheduleConfig())
 
+	c.Assert(int(svr.GetScheduleConfig().MaxMergeRegionSize), Equals, 20)
+	c.Assert(int(svr.GetScheduleConfig().MaxMergeRegionKeys), Equals, 0)
+	c.Assert(int(svr.GetScheduleConfig().GetMaxMergeRegionKeys()), Equals, 20*10000)
+
+	// set max-merge-region-size to 40MB
+	args = []string{"-u", pdAddr, "config", "set", "max-merge-region-size", "40"}
+	output, err = pdctl.ExecuteCommand(cmd, args...)
```
lhy1024, line 138: plz fix ci

server/config/config.go
bufferflies: yes， the value is same with tikv: https://github.com/tikv/tikv/blob/f7edbcf610cd6bb5c9040317edd3260188a6e87d/components/raftstore/src/coprocessor/config.rs#L102-L106

lhy1024 (commented):
plz fix ci

zhangjinpeng1987 (approved):
LGTM

lhy1024 (approved):

server/config/config.go
JmPotato: Could you please help add a comment to tell this is to keep consistent with TiKV?

server/config/persist_options.go
```diff
@@ -245,7 +245,7 @@ func (o *PersistOptions) GetMaxMergeRegionSize() uint64 {
 
 // GetMaxMergeRegionKeys returns the max number of keys.
 func (o *PersistOptions) GetMaxMergeRegionKeys() uint64 {
-	return o.getTTLUintOr(maxMergeRegionKeysKey, o.GetScheduleConfig().MaxMergeRegionKeys)
+	return o.getTTLUintOr(maxMergeRegionKeysKey, o.GetScheduleConfig().GetMaxMergeRegionKeys())
```
JmPotato, line 248: Do we need to make sure the `MaxMergeRegionKeys` will be `MaxMergeRegionSize * 10000` when `MaxMergeRegionKeys` is set to 0 with TTL?

server/config/config.go
bufferflies: fix

server/config/persist_options.go
bufferflies: yes, it will return size * 10000 if `MaxMergeRegionKeys ` is not exist .

server/config/persist_options.go
```diff
@@ -244,8 +244,17 @@ func (o *PersistOptions) GetMaxMergeRegionSize() uint64 {
 }
 
 // GetMaxMergeRegionKeys returns the max number of keys.
+// It returns size * 10000 if the key of max-merge-region-Keys doesn't exist.
 func (o *PersistOptions) GetMaxMergeRegionKeys() uint64 {
-	return o.getTTLUintOr(maxMergeRegionKeysKey, o.GetScheduleConfig().GetMaxMergeRegionKeys())
+	keys, exist, err := o.getTTLUint(maxMergeRegionKeysKey)
+	if exist && err == nil {
```
JmPotato, line 250: I think this should be `exist && err == nil && keys == 0`?

JmPotato (approved):

PR #5125, opened 2022-06-08
filter: Split `StoreStateFilter` into 2 filters

### LINKED ISSUE #5130: Split the Store state filter into temporary state and long-term state
## Enhancement Task
Currently, when we do store state filtering, we mix temporary state and long-term state together, which is not convenient for us to do scheduling diagnosis more clearly.

### LINKED ISSUE #5304: StoreStateFilterReasonTest has a problem
### Which jobs are failing
#5168 
use `Source` twice
https://github.com/tikv/pd/blob/dc41d5235d4a7772a3f4e33896bfe72743b5d848/server/schedule/filter/filters_test.go#L208-L211

CabinfeverB:

Issue Number: Close #5130 Close #5304

remove `StoreStateFilter` and add `LongtermStateFilter` and. `TemporaryStateFilter`

```commit-message
remove `StoreStateFilter` and add `LongtermStateFilter` and. `TemporaryStateFilter`
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
 None.
```

### DIFF
--- a/server/schedule/checker/replica_strategy.go
+++ b/server/schedule/checker/replica_strategy.go
@@ -56,7 +56,7 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 		filter.NewExcludedFilter(s.checkerName, nil, s.region.GetStoreIDs()),
 		filter.NewStorageThresholdFilter(s.checkerName),
 		filter.NewSpecialUseFilter(s.checkerName),
-		&filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true, AllowTemporaryStates: true},
+		filter.NewLongTermStateFilter(s.checkerName, filter.MoveRegion),
 	}
 	if len(s.locationLabels) > 0 && s.isolationLevel != "" {
 		filters = append(filters, filter.NewIsolationFilter(s.checkerName, s.isolationLevel, s.locationLabels, coLocationStores))
@@ -69,7 +69,7 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 	}
 
 	isolationComparer := filter.IsolationComparer(s.locationLabels, coLocationStores)
-	strictStateFilter := &filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true}
+	strictStateFilter := filter.NewTemporaryStateFilter(s.checkerName, filter.MoveRegion)
 	targetCandidate := filter.NewCandidates(s.cluster.GetStores()).
 		FilterTarget(s.cluster.GetOpts(), filters...).
 		KeepTheTopStores(isolationComparer, false) // greater isolation score is better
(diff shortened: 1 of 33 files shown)


server/schedule/checker/replica_strategy.go
```diff
@@ -56,7 +56,7 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 		filter.NewExcludedFilter(s.checkerName, nil, s.region.GetStoreIds()),
 		filter.NewStorageThresholdFilter(s.checkerName),
 		filter.NewSpecialUseFilter(s.checkerName),
-		&filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true, AllowTemporaryStates: true},
+		&filter.LongTermStateFilter{ActionScope: s.checkerName, MoveRegion: true},
```
rleungx, line 59: Should it be TemporaryStateFilter?

CabinfeverB: AllowTemporaryStates, if true, means only long-term status checks.

server/schedule/filter/filters.go
```diff
@@ -264,135 +264,206 @@ func (f *distinctScoreFilter) GetSourceStoreID() uint64 {
...
+// TemporaryStateFilter is used to determine whether a store can be selected as the
+// source or target of the schedule based on the store's temporary state.
+type TemporaryStateFilter struct {
 	ActionScope string
 	// Set true if the schedule involves any transfer leader operation.
 	TransferLeader bool
 	// Set true if the schedule involves any move region operation.
 	MoveRegion bool
 	// Set true if the scatter move the region
 	ScatterRegion bool
```
AndreMouche, line 301: Could we merge the usage info `TransferLeader`,`MoveRegion`,`ScatterRegion` into one int value, such as `OpKind` did, so we could support more in the future.

server/schedule/filter/filters.go
```diff
@@ -264,135 +264,206 @@ func (f *distinctScoreFilter) GetSourceStoreID() uint64 {
...
+	}
+	return true
 }
 
-func (f *StoreStateFilter) exceedAddLimit(opt *config.PersistOptions, store *core.StoreInfo) bool {
-	f.Reason = "exceed-add-limit"
-	return !f.AllowTemporaryStates && !store.IsAvailable(storelimit.AddPeer)
+// LongTermStateFilter is used to determine whether a store can be selected as the
+// source or target of the schedule based on the store's long-term state.
+type LongTermStateFilter struct {
```
AndreMouche, line 401: How about define the StateFile like
```
type StateFilter struct {
	ActionScope string
	// TransferLeader,MoveRegion,ScatterRegion infomations.
	kind int
	// Reason is used to distinguish the reason of store state filter
	Reason string
	// StateType is used to distinguish the type of store state
	StateType int
}

type TemporaryStateFilter struct {
	*StateFilter
}

type LongTermStateFilter struct {
	*StateFilter
}
```

AndreMouche (commented):
How about

AndreMouche (approved):
LGTM

CabinfeverB:
No need to split `StoreStateFilter`

Issue #4813, opened 2022-04-15
Tracking Issue for Test Restructure


JmPotato:
This is the tracking issue for restructuring PD tests, focusing on unit tests and integration tests, especially on unifying test infrastructure in [testify](https://github.com/stretchr/testify). See https://github.com/pingcap/tidb/issues/26022 for more details on the TiDB side. Before you start, please read the code of conduct below carefully.

## Code of Conduct

* Make the PR as small as possible.
	* Modify only one or two packages at a time if possible.
	* Only focus on the test framework migration work, do not change the test logic. If you feel there is a strong need to refactor some tests, please file an issue to work on it later.
* Carefully use the suite of testify because it runs its inside tests serially.
* If a test needs to run a mock test cluster or something that is not lightweight, it's a good idea to put several tests inside a single suite to ensure maximum resource sharing.
* If you need to call `require.Xxx(t, a, b)` multiple times, please create a `re := require.New(t)` first to avoid passing the `t` all around.
* Use only `require.Xxx` or `suite.Xxx` whenever possible, do not mix the two.
* For a function that will return `error` type, please use `require.NoError/Error` rather than `require.Nil/NotNil`.
* For an equality comparison, please place the expected value as the first argument to fit the function signature, which will make the error message clear if the test fails.
	* `re.Equal(expected, actual)` ✅
	* `re.Equal(actual, expected)` ❌
* Comment on this issue first to tell others which package you will work on.

## Migration

- [x] migrate test-infra to testify for `client` (#5081)
- [x] migrate test-infra to testify for `pkg/apiutil` (#5061)
- [x] migrate test-infra to testify for `pkg/assertutil` (#5067)
- [x] migrate test-infra to testify for `pkg/audit` (#5061)
- [x] migrate test-infra to testify for `pkg/autoscaling` (#5061)
- [x] migrate test-infra to testify for `pkg/cache` (#5061)
- [x] migrate test-infra to testify for `pkg/codec` (#5066)
- [x] migrate test-infra to testify for `pkg/encryption` (#5066)
- [x] migrate test-infra to testify for `pkg/errs` (#5066)
- [x] migrate test-infra to testify for `pkg/etcdutil` (#5067)
- [x] migrate test-infra to testify for `pkg/grpcutil` (#5067)
- [x] migrate test-infra to testify for `pkg/keyutil` (#5067)
- [x] migrate test-infra to testify for `pkg/logutil` (#5067)
[...]

Issue #5104, opened 2022-06-06
Use `require.Equal` to replace `reflect.DeepEqual`


JmPotato:
## Enhancement Task

The implementation of `ObjectsAreEqual` inside `require.Equal` is effectively just a call to [reflect.DeepEqual](https://golang.org/pkg/reflect/#DeepEqual), which means we don't need `require.True(reflect.DeepEqual(expected, actual))`. So it's a good idea to replace all `reflect.DeepEqual` in tests to `require.Equal`.

PR #5139, opened 2022-06-09
storage: migrate test framework to testify

LLThomas:
Signed-off-by: LLThomas <zs033@qq.com>

Issue Number: Ref #4813

```commit-message
As the title says.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/storage/hot_region_storage_test.go
+++ b/server/storage/hot_region_storage_test.go
@@ -25,7 +25,7 @@ import (
 	"testing"
 	"time"
 
-	. "github.com/pingcap/check"
+	"github.com/stretchr/testify/require"
 	"github.com/tikv/pd/server/core"
 )
 
@@ -104,21 +104,11 @@ func (m *MockPackHotRegionInfo) ClearHotRegion() {
 	m.historyHotWrites = make([]HistoryHotRegion, 0)
 }
 
-var _ = SerialSuites(&testHotRegionStorage{})
-
-type testHotRegionStorage struct {
-	ctx    context.Context
-	cancel context.CancelFunc
-}
-
-func (t *testHotRegionStorage) SetUpSuite(c *C) {
-	t.ctx, t.cancel = context.WithCancel(context.Background())
-}
-
-func (t *testHotRegionStorage) TestHotRegionWrite(c *C) {
+func TestHotRegionWrite(t *testing.T) {
+	re := require.New(t)
 	packHotRegionInfo := &MockPackHotRegionInfo{}
 	store, clean, err := newTestHotRegionStorage(10*time.Minute, 1, packHotRegionInfo)
-	c.Assert(err, IsNil)
+	re.NoError(err)
 	defer clean()
 	now := time.Now()
 	hotRegionStorages := []HistoryHotRegion{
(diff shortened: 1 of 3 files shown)


LLThomas:
When I run test, TestLoadMinServiceSafePoint can't pass. 
The error occurs in "re.Nil(ssp)".
```
  // verify that service safe point with ServiceID 0 has been removed
  ssp, err := storage.LoadServiceSafePoint(testKeySpace, "0")
  re.NoError(err)
  re.Nil(ssp)
```
Then,  I test the original code.
```
  // verify that service safe point with ServiceID 0 has been removed
  ssp, err := storage.LoadServiceSafePoint(testKeySpace, "0")
  c.Assert(err, IsNil)
  c.Assert(ssp, IsNil)
```
"c.Assert(ssp, IsNil)" also has the same error.
Maybe it is a bug.

server/storage/hot_region_storage_test.go
```diff
@@ -207,7 +198,7 @@ func (t *testHotRegionStorage) TestHotRegionDelete(c *C) {
 	num := 0
 	for next, err := iter.Next(); next != nil && err == nil; next, err = iter.Next() {
 		num++
-		c.Assert(reflect.DeepEqual(next, &historyHotRegions[defaultRemainDay-num]), IsTrue)
+		re.True(reflect.DeepEqual(next, &historyHotRegions[defaultRemainDay-num]))
```
rleungx, line 201: Equal is enough.

LLThomas: I think "re.True" is more suitable for functions that return bool value.
And I see an example that has been merged.

```
- c.Assert(rep.GetReplicationStatus().GetDrAutoSync().GetPauseRegionSplit(), IsTrue)
+ re.True(rep.GetReplicationStatus().GetDrAutoSync().GetPauseRegionSplit())
```
https://github.com/tikv/pd/pull/5100/files#diff-f36769a34db4e8e3412376a3a7c5be18b2f7f8ab90287fc48c7f4445bc716218L221

JmPotato: See https://github.com/tikv/pd/issues/5104, `require.Equal` is used to replace `reflect.DeepEqual`.

```suggestion
		re.Equal(&historyHotRegions[defaultRemainDay-num], next)
```

LLThomas: Oh, I got it, I'll fix it.

LLThomas:
> > When I run test, TestLoadMinServiceSafePoint can't pass. The error occurs in "re.Nil(ssp)".
> > ```
> >   // verify that service safe point with ServiceID 0 has been removed
> >   ssp, err := storage.LoadServiceSafePoint(testKeySpace, "0")
> >   re.NoError(err)
> >   re.Nil(ssp)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, I test the original code.
> > ```
> >   // verify that service safe point with ServiceID 0 has been removed
> >   ssp, err := storage.LoadServiceSafePoint(testKeySpace, "0")
> >   c.Assert(err, IsNil)
> >   c.Assert(ssp, IsNil)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > "c.Assert(ssp, IsNil)" also has the same error. Maybe it is a bug.
> 
> This test relies on the failpoint injection, you can see it is enabled on line 107.
> 
> https://github.com/tikv/pd/blob/86a68215c89936ff9bd242f6ee0dc9bc95a74969/server/storage/storage_gc_test.go#L107
> 
> To make the failpoint work, you should call `make failpoint-enable` before running the test.

Thanks! I got it.

rleungx (approved):

JmPotato (approved):

PR #5144, opened 2022-06-13
replica_strategy:speed up and reduce the complexity of selectStore to O(n)

### LINKED ISSUE #5143: replica_strategy: could we reduce the complexity of selectStore to O(n)
## Enhancement Task
It seems when we are try to select a store to remove or add, we are just try to find a store with a max or min score with some algorithm. However, it seems we always need a sort function with O(NLogN), could we reduce the complexity to O(n) ?
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/replica_strategy.go#L123-L129](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/replica_strategy.go#L123-L129)
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/replica_strategy.go#L73-L80](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/replica_strategy.go#L73-L80)

AndreMouche:

Issue Number: Close #5143 

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/checker/replica_strategy.go
+++ b/server/schedule/checker/replica_strategy.go
@@ -72,12 +72,12 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 	strictStateFilter := &filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true}
 	targetCandidate := filter.NewCandidates(s.cluster.GetStores()).
 		FilterTarget(s.cluster.GetOpts(), filters...).
-		KeepTheTopItems(isolationComparer, false) // greater isolation score is better
+		KeepTheTopStores(isolationComparer, false) // greater isolation score is better
 	if targetCandidate.Len() == 0 {
 		return 0, false
 	}
 	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).
-		PickTheTopItem(filter.RegionScoreComparer(s.cluster.GetOpts()), true) // less region score is better
+		PickTheTopStore(filter.RegionScoreComparer(s.cluster.GetOpts()), true) // less region score is better
 	if target == nil {
 		return 0, true // filter by temporary states
 	}
@@ -124,8 +124,8 @@ func (s *ReplicaStrategy) SelectStoreToRemove(coLocationStores []*core.StoreInfo
 	isolationComparer := filter.IsolationComparer(s.locationLabels, coLocationStores)
 	source := filter.NewCandidates(coLocationStores).
 		FilterSource(s.cluster.GetOpts(), &filter.StoreStateFilter{ActionScope: replicaCheckerName, MoveRegion: true}).
-		KeepTheTopItems(isolationComparer, true).
-		PickTheTopItem(filter.RegionScoreComparer(s.cluster.GetOpts()), true)
+		KeepTheTopStores(isolationComparer, true).
+		PickTheTopStore(filter.RegionScoreComparer(s.cluster.GetOpts()), true)
 	if source == nil {
 		log.Debug("no removable store", zap.Uint64("region-id", s.region.GetID()))
 		return 0
--- a/server/schedule/checker/replica_strategy.go
+++ b/server/schedule/checker/replica_strategy.go
@@ -72,12 +72,12 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 	strictStateFilter := &filter.StoreStateFilter{ActionScope: s.checkerName, MoveRegion: true}
 	targetCandidate := filter.NewCandidates(s.cluster.GetStores()).
 		FilterTarget(s.cluster.GetOpts(), filters...).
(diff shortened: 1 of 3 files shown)


server/schedule/checker/replica_strategy.go
```diff
@@ -72,12 +72,12 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
...
 		FilterTarget(s.cluster.GetOpts(), filters...).
-		Sort(isolationComparer).Reverse().Top(isolationComparer). // greater isolation score is better
-		Sort(filter.RegionScoreComparer(s.cluster.GetOpts()))     // less region score is better
+		KeepTheTopStores(isolationComparer, false) // greater isolation score is better
 	if targetCandidate.Len() == 0 {
 		return 0, false
 	}
-	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).PickFirst() // the filter does not ignore temp states
+	target := targetCandidate.FilterTarget(s.cluster.GetOpts(), strictStateFilter).
+		PickTheTopStore(filter.RegionScoreComparer(s.cluster.GetOpts()), true) // less region score is better
```
rleungx, line 80: The logic here is different from the previous implementation.

AndreMouche: If you mean the check of `targetCandidate`, it should keep the same with the previous since the second sort won't change the length. The only logic which is changed here is we do `filter` before sort, which may make it faster, but shouldn't change the final result.

rleungx: The RegionScoreComparer should be a long-term state which should be included in the first stage.

server/schedule/filter/candidates_test.go
```diff
@@ -15,9 +15,10 @@
 package filter
 
 import (
+	"testing"
+
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/stretchr/testify/require"
-	"testing"
```
rleungx, line 22: remove the empty line

server/schedule/filter/candidates.go
```diff
@@ -52,6 +52,7 @@ func (c *StoreCandidates) Sort(less StoreComparer) *StoreCandidates {
 }
 
 // Reverse reverses the candidate store list.
+// Deprecated function, only used in the tests.
 func (c *StoreCandidates) Reverse() *StoreCandidates {
```
rleungx, line 56: Maybe we can remove it?

CabinfeverB (commented):
If we wanted to implement a retry mechanism in ReplicaStrategy, this method may be not useful

CabinfeverB (approved):

server/schedule/filter/candidates.go
```diff
@@ -51,32 +51,49 @@ func (c *StoreCandidates) Sort(less StoreComparer) *StoreCandidates {
...
+		}
+		c.Stores[idx], c.Stores[topIdx] = c.Stores[topIdx], c.Stores[idx]
 	}
-	c.Stores = c.Stores[:i]
+	c.Stores = c.Stores[:topIdx+1]
 	return c
 }
 
+// PickTheTopStore returns the first store order by asc.
+// it returns the min item when asc is true, return the max item when asc is false.
```
bufferflies, line 82:
```suggestion
// It returns the min item when asc is true, return the max item when asc is false.
```

server/schedule/filter/candidates_test.go
bufferflies, line 22: remove this line.

server/schedule/filter/candidates_test.go
```diff
@@ -87,7 +91,7 @@ func TestCandidates(t *testing.T) {
 	re.Less(store.GetID(), uint64(8))
 
 	cs = newTestCandidates(10, 15, 23, 20, 33, 32, 31)
-	cs.Sort(idComparer).Reverse().Top(idComparer2)
+	cs.KeepTheTopStores(idComparer2, false)
```
bufferflies, line 94: How about add one unit test case for `KeepTheTopStores(cmp, true)`?

server/schedule/filter/candidates.go
```diff
@@ -51,32 +51,49 @@ func (c *StoreCandidates) Sort(less StoreComparer) *StoreCandidates {
...
+		}
+		c.Stores[idx], c.Stores[topIdx] = c.Stores[topIdx], c.Stores[idx]
 	}
-	c.Stores = c.Stores[:i]
+	c.Stores = c.Stores[:topIdx+1]
 	return c
 }
 
+// PickTheTopStore returns the first store order by asc.
+// It returns the min item when asc is true, return the max item when asc is false.
```
bufferflies, line 82:
```suggestion
// It returns the min item when asc is true, returns the max item when asc is false.
```

bufferflies (approved):
rest LGTM

PR #5153, opened 2022-06-13
tests: use `T.TempDir` to create temporary test directory

### LINKED ISSUE #5152: Simplify tests using `T.TempDir`
## Enhancement Task

Currently we are using `os.MkdirTemp` and `os.RemoveAll` to create and cleanup temporary directory in tests. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Juneezee:

Issue Number: Close #5152 

This pull request replaces `os.MkdirTemp` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/encryption/master_key_test.go
+++ b/pkg/encryption/master_key_test.go
@@ -99,8 +99,7 @@ func TestNewFileMasterKeyMissingPath(t *testing.T) {
 func TestNewFileMasterKeyMissingFile(t *testing.T) {
 	t.Parallel()
 	re := require.New(t)
-	dir, err := os.MkdirTemp("", "test_key_files")
-	re.NoError(err)
+	dir := t.TempDir()
 	path := dir + "/key"
 	config := &encryptionpb.MasterKey{
 		Backend: &encryptionpb.MasterKey_File{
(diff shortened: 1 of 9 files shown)


JmPotato:
Thanks for your contribution! After global searching `os.MkdirTemp`, it seems that this part of code could be replaced too, would you like to rewrite it also?

https://github.com/tikv/pd/blob/7baf94c8644ccd55930d213e298fc0104bc7a453/pkg/etcdutil/etcdutil.go#L188

Juneezee:
@JmPotato Thanks! I missed that one because I was only searching in `*_test.go` files. Please take a look again. Thank you.

JmPotato (approved):
LGTM. Thx.

nolouch (approved):

PR #5157, opened 2022-06-14
scheduler/balance_leader: fix data race in the function of clone for config

### LINKED ISSUE #5156: scheduler/balance_leader: potential data race in the function of clone for config
## Bug Report

It may cause data race since it do not deeply clone the slice of `conf.Ranges` in the function of `Clone` for config:
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedulers/balance_leader.go#L128-L133](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedulers/balance_leader.go#L128-L133)

AndreMouche:

Issue Number: Close #5156 

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```

### DIFF
--- a/server/schedulers/balance_leader.go
+++ b/server/schedulers/balance_leader.go
@@ -128,9 +128,10 @@ func (conf *balanceLeaderSchedulerConfig) validate() bool {
 func (conf *balanceLeaderSchedulerConfig) Clone() *balanceLeaderSchedulerConfig {
 	conf.mu.RLock()
 	defer conf.mu.RUnlock()
-	var ranges []core.KeyRange
+	ranges := make([]core.KeyRange, len(conf.Ranges))
+	copy(ranges, conf.Ranges)
 	return &balanceLeaderSchedulerConfig{
-		Ranges: append(ranges, conf.Ranges...),
+		Ranges: ranges,
 		Batch:  conf.Batch,
 	}
 }
--- a/server/schedulers/balance_leader.go
+++ b/server/schedulers/balance_leader.go
@@ -128,8 +128,9 @@ func (conf *balanceLeaderSchedulerConfig) validate() bool {
 func (conf *balanceLeaderSchedulerConfig) Clone() *balanceLeaderSchedulerConfig {
 	conf.mu.RLock()
 	defer conf.mu.RUnlock()
+	var ranges []core.KeyRange
 	return &balanceLeaderSchedulerConfig{
-		Ranges: conf.Ranges,
+		Ranges: append(ranges, conf.Ranges...),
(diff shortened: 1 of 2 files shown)


CabinfeverB (approved):

server/schedulers/balance_leader.go
```diff
@@ -128,8 +128,9 @@ func (conf *balanceLeaderSchedulerConfig) validate() bool {
 func (conf *balanceLeaderSchedulerConfig) Clone() *balanceLeaderSchedulerConfig {
 	conf.mu.RLock()
 	defer conf.mu.RUnlock()
+	var ranges []core.KeyRange
```
bufferflies, line 131: How about define the capacity for the ranges first?  some difference between copy and append : https://stackoverflow.com/questions/71316774/is-it-equivalent-use-copy-or-append-slice-type-in-golang

server/schedulers/balance_leader_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+package schedulers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBalanceLeaderSchedulerConfigClone(t *testing.T) {
```
bufferflies, line 23: Maybe not need this unit test?

AndreMouche: The test is used to prove this PR which should be failed without this PR.

rleungx:
Does other scheduler have the same problems?

AndreMouche:
> Does other scheduler have the same problems?

Yes, I've review schedulers, and find that almost all Clone function for those config that contains slice or map, exist the similar problem.
How about create one or more issues and fix it one by one?

rleungx:
Does this PR need to cherry-pick? If so, I suggest merging them into one.

AndreMouche:
Maybe not? since the read function happens rarely

> Does this PR need to cherry-pick? If so, I suggest merging them into one.

server/schedulers/balance_leader_test.go
```diff
@@ -0,0 +1,38 @@
...
+	re := require.New(t)
+	keyRanges1, _ := getKeyRanges([]string{"a", "b", "c", "d"})
+	conf := &balanceLeaderSchedulerConfig{
+		Ranges: keyRanges1,
+		Batch:  10,
+	}
+	con2 := conf.Clone()
```
rleungx, line 30:
```suggestion
	conf2 := conf.Clone()
```

rleungx (approved):

PR #5168, opened 2022-06-16
schedule: fix the wrong reason problem in StoreStateFilter

### LINKED ISSUE #5150: scheduler/filter: could we define the reason when needed to improve GC
## Enhancement Task
In the `StoreStateFilter`, we need to check the condition and record the failed reason when failed, However, we always record the reason before we check the condition, since we will stop once we meet false so the final reason should be right, while it may case GC problem if we update the `reason` in every condition-check function.
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/filter/filters.go#L297-L310](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/filter/filters.go#L297-L310)

xueshanfeishu:
filter: fix the bug that StoreStateFilter get the wrong reason (#5150) 

Issue Number: Close #5150

```commit-message

    filter: fix the bug that StoreStateFilter get the wrong reason (#5150)
    close tikv/pd#5150
    fix the bug that StoreStateFilter get the wrong reason
    
    Signed-off-by: lijujie <lijunjie.zb@ccbft.com>

```

### Check List

Tests

- Integration test
func TestStoreStateFilterReason(t *testing.T) {
	re := require.New(t)
	filters := []Filter{
		&StoreStateFilter{TransferLeader: true},
		&StoreStateFilter{MoveRegion: true},
		&StoreStateFilter{TransferLeader: true, MoveRegion: true},
		&StoreStateFilter{MoveRegion: true, AllowTemporaryStates: true},
	}
	opt := config.NewTestOptions()
	store := core.NewStoreInfoWithLabel(1, 0, map[string]string{})

	type testCase struct {
		filterIdx    int
		sourceReason string
		targetReason string
	}

	check := func(store *core.StoreInfo, testCases []testCase) {
		for _, tc := range testCases {
			filters[tc.filterIdx].Source(opt, store)
			re.Equal(tc.sourceReason, filters[tc.filterIdx].(*StoreStateFilter).Reason)
			filters[tc.filterIdx].Source(opt, store)
			re.Equal(tc.targetReason, filters[tc.filterIdx].(*StoreStateFilter).Reason)
		}
	}

	// No reason catched
	store = store.Clone(core.SetLastHeartbeatTS(time.Now()))
	testCases := []testCase{
		{2, "", ""},
	}
	check(store, testCases)

	// Disconn
	store = store.Clone(core.SetLastHeartbeatTS(time.Now().Add(-5 * time.Minute)))
	testCases = []testCase{
		{0, "disconnected", "disconnected"},
		{1, "", ""},
		{2, "disconnected", "disconnected"},
		{3, "", ""},
	}
	check(store, testCases)

	// Busy
	store = store.Clone(core.SetLastHeartbeatTS(time.Now())).
		Clone(core.SetStoreStats(&pdpb.StoreStats{IsBusy: true}))
	testCases = []testCase{
		{0, "", ""},
		{1, "busy", "busy"},
		{2, "busy", "busy"},
		{3, "", ""},
	}
	check(store, testCases)
}

Code changes

- No configuration change
- No HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- No persistent data change

Side effects

Related changes

### Release note

```release-note
None.
```


server/schedule/filter/filters.go
```diff
@@ -295,66 +295,126 @@ func (f *StoreStateFilter) Type() string {
 type conditionFunc func(*config.PersistOptions, *core.StoreInfo) bool
 
 func (f *StoreStateFilter) isRemoved(opt *config.PersistOptions, store *core.StoreInfo) bool {
-	f.Reason = "tombstone"
-	return store.IsRemoved()
+	isRemoved := store.IsRemoved()
+	if isRemoved {
+		f.Reason = "tombstone"
+	} else {
```
rleungx, line 301: We can remove else?

xueshanfeishu: I don't think else can be removed. In function anyConditionMatch, any condition matches , it will return and keep the reason in Reason, but if none matches, anyConditionMatch return false, but Reason will keep the last reason before calling anyConditionMatch.

server/schedule/filter/filters.go
```diff
@@ -295,66 +295,126 @@ func (f *StoreStateFilter) Type() string {
 type conditionFunc func(*config.PersistOptions, *core.StoreInfo) bool
 
 func (f *StoreStateFilter) isRemoved(opt *config.PersistOptions, store *core.StoreInfo) bool {
-	f.Reason = "tombstone"
-	return store.IsRemoved()
+	isRemoved := store.IsRemoved()
```
nolouch, line 298: hou about:
```suggestion
        f.Reason = ""
	isRemoved := store.IsRemoved()
	if isRemoved {
	      f.Reason = "tombstone"
	}
	return isRemoved
```
others also can change like it.

nolouch (commented):
rest LGTM

server/schedule/filter/filters.go
xueshanfeishu: Thanks for reviewing!   I think both can get the same result, use if{}else{}  Reason will be set  only once, but set reason and use if{}, Reason may be set twice,  which may have a little influence on the performance.

rleungx (approved):

nolouch (approved):

PR #5181, opened 2022-06-18
client: add AllocID interface for pd client

3pointer:

Issue Number: Close https://github.com/pingcap/tidb/issues/35489
According to ebs backup trace issue https://github.com/pingcap/tidb/issues/33860 , we need to get pd alloc id for backup, and recover it in another cluster.

``` client: add AllocID interface for pd client

```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has configuration change
- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Please add a release note.

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```


client/client.go
```diff
@@ -57,6 +57,8 @@ type GlobalConfigItem struct {
 // Client is a PD (Placement Driver) client.
 // It should not be used after calling Close().
 type Client interface {
+	// GetAllocID gets the alloc ID from PD.
+	GetAllocID(ctx context.Context) (uint64, error)
```
rleungx, line 61: How about just `AllocID`?

client/client.go
```diff
@@ -1247,6 +1249,27 @@ var tsoReqPool = sync.Pool{
 	},
 }
 
+// GetAllocID gets the alloc ID from PD.
+func (c *client) GetAllocID(ctx context.Context) (uint64, error) {
```
rleungx, line 1253: Do we need to record the execution time?

client/client.go
3pointer: It looks like all function in `Client` interface has the `Get` prefix. so I think maybe keep Get here is better?🤔

```
        // GetClusterID gets the cluster ID from PD.
	GetClusterID(ctx context.Context) uint64
	// GetAllMembers gets the members Info from PD
	GetAllMembers(ctx context.Context) ([]*pdpb.Member, error)
	// GetLeaderAddr returns current leader's address. It returns "" before
	// syncing leader from server.
	GetLeaderAddr() string
	// GetTS gets a timestamp from PD.
	GetTS(ctx context.Context) (int64, int64, error)
	// GetTSAsync gets a timestamp from PD, without block the caller.
	GetTSAsync(ctx context.Context) TSFuture
	// GetLocalTS gets a local timestamp from PD.
	GetLocalTS(ctx context.Context, dcLocation string) (int64, int64, error)
	// GetLocalTSAsync gets a local timestamp from PD, without block the caller.
	GetLocalTSAsync(ctx context.Context, dcLocation string) TSFuture
	// GetRegion gets a region and its leader Peer from PD by key.
	// The region may expire after split. Caller is responsible for caching and
	// taking care of region change.
	// client should retry later.
	GetRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
	GetRegionFromMember(ctx context.Context, key []byte, memberURLs []string) (*Region, error)
	// GetPrevRegion gets the previous region and its leader Peer of the region where the key is located.
	GetPrevRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
	// GetRegionByID gets a region and its leader Peer from PD by id.
	GetRegionByID(ctx context.Context, regionID uint64, opts ...GetRegionOption) (*Region, error)

	// GetStore gets a store from PD by store id.
	// The store may expire later. Caller is responsible for caching and taking care
	// of store change.
	GetStore(ctx context.Context, storeID uint64) (*metapb.Store, error)
	
        // GetAllStores gets all stores from pd.
	// The store may expire later. Caller is responsible for caching and taking care
	// of store change.
	GetAllStores(ctx context.Context, opts ...GetStoreOption) ([]*metapb.Store, error)	

	// GetOperator gets the status of operator of the specified region.
	GetOperator(ctx context.Context, regionID uint64) (*pdpb.GetOperatorResponse, error)

```

client/client.go
3pointer: ok，I'll remove it

client/client.go
D3Hunter: > It looks like all function in `Client` interface has the `Get` prefix. so I think maybe keep Get here is better?🤔

impl calls `AllocID` which alloc a new id,  but seems other interfaces with `Get` prefix are just read only operation.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/AlekSi/gocov-xml v1.0.0
 	github.com/BurntSushi/toml v0.3.1
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
```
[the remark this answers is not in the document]
rleungx: You can remove `docs/swagger/docs.go` and re-run `go mod tidy`.

disksing:
maybe consider using the pd-backup tool?

D3Hunter:
all backup/restore task should better be done using one tool, pd should expose those api to `br`

> maybe consider using the pd-backup tool?

server/api/admin.go
```diff
@@ -97,7 +106,21 @@ func (h *adminHandler) ResetTS(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err = handler.ResetTS(ts); err != nil {
+	forceUseLarger := false
+	forceUseLargerVal, contains := input["force-use-larger"]
+	if contains {
+		if forceUseLarger, ok = forceUseLargerVal.(bool); !ok {
+			h.rd.JSON(w, http.StatusBadRequest, "invalid force-use-larger value")
```
rleungx, line 113: need return?

server/api/admin.go
```diff
@@ -97,7 +106,21 @@ func (h *adminHandler) ResetTS(w http.ResponseWriter, r *http.Request) {
...
+	var ignoreSmaller, skipUpperBoundCheck bool
+	if forceUseLarger {
+		ignoreSmaller, skipUpperBoundCheck = true, true
+	} else {
+		ignoreSmaller, skipUpperBoundCheck = false, false
+	}
+
+	if err = handler.ResetTS(ts, ignoreSmaller, skipUpperBoundCheck); err != nil {
 		if err == server.ErrServerNotStarted {
 			h.rd.JSON(w, http.StatusInternalServerError, err.Error())
```
rleungx, line 125: ditto

server/api/admin.go
```diff
@@ -127,3 +150,78 @@ func (h *adminHandler) SavePersistFile(w http.ResponseWriter, r *http.Request) {
...
+		_ = h.rd.Text(w, http.StatusForbidden, "can only recover alloc id when recovering mark marked")
+		return
+	}
+
+	leader := h.svr.GetLeader()
+	if leader == nil {
+		_ = h.rd.Text(w, http.StatusServiceUnavailable, errs.ErrLeaderNil.FastGenByArgs().Error())
+		return
+	}
+	ctx := grpcutil.BuildForwardContext(r.Context(), leader.ClientUrls[0])
```
rleungx, line 214: Why does it create a gRPC request in HTTP API?

server/handler.go
```diff
@@ -920,15 +920,19 @@ func (h *Handler) GetOfflinePeer(typ statistics.RegionStatisticType) ([]*core.Re
 }
 
 // ResetTS resets the ts with specified tso.
-func (h *Handler) ResetTS(ts uint64) error {
+func (h *Handler) ResetTS(ts uint64, ignoreSmaller, skipUpperBoundCheck bool) error {
+	log.Info("reset ts",
+		zap.Uint64("new ts", ts),
+		zap.Bool("ignore smaller", ignoreSmaller),
```
rleungx, line 926: use the dash to connect the word

server/server.go
```diff
@@ -1602,3 +1604,39 @@ func (s *Server) IsTTLConfigExist(key string) bool {
 	}
 	return false
 }
+
+// MarkRecovering mark pd that we're recovering
+// tikv will get this state during BR EBS restore.
+func (s *Server) MarkRecovering() error {
+	log.Info("mark recovering")
```
rleungx, line 1611: I suggest printing the log after the operation is successful or printing the log in both the start and end place.

server/server.go
```diff
@@ -84,6 +84,8 @@ const (
 	// idAllocPath for idAllocator to save persistent window's end.
 	idAllocPath  = "alloc_id"
 	idAllocLabel = "idalloc"
+
+	pdRecoveringMarkPath = "/pd/recovering-mark"
```
rleungx, line 88: Does it need a cluster_id prefix?

server/grpc_service.go
```diff
@@ -462,6 +462,41 @@ func (s *GrpcServer) AllocID(ctx context.Context, request *pdpb.AllocIDRequest)
 	}, nil
 }
 
+// RecoverAllocID implements gRPC PDServer.
+func (s *GrpcServer) RecoverAllocID(ctx context.Context, request *pdpb.RecoverAllocIDRequest) (*pdpb.RecoverAllocIDResponse, error) {
```
rleungx, line 466: Do we need both HTTP and gRPC interfaces?

server/grpc_service.go
```diff
@@ -462,6 +462,41 @@ func (s *GrpcServer) AllocID(ctx context.Context, request *pdpb.AllocIDRequest)
...
+	if rsp, err := s.unaryMiddleware(ctx, request.GetHeader(), fn); err != nil {
+		return nil, err
+	} else if rsp != nil {
+		return rsp.(*pdpb.RecoverAllocIDResponse), err
+	}
+
+	// We can use an allocator for all types ID allocation.
+	err := s.idAllocator.SetBase(request.Id)
+	if err != nil {
+		return nil, status.Errorf(codes.Unknown, err.Error())
```
rleungx, line 479: Should the error be put into the response header?

server/api/admin.go
```diff
@@ -97,7 +106,21 @@ func (h *adminHandler) ResetTS(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err = handler.ResetTS(ts); err != nil {
+	forceUseLarger := false
+	forceUseLargerVal, contains := input["force-use-larger"]
```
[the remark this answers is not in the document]
D3Hunter: no, if the key not exist, cannot do cast

server/api/admin.go
D3Hunter: yes, previous bug

server/api/admin.go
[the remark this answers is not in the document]
rleungx: The `ok` will get false.

server/api/admin.go
D3Hunter: only leader can set it, so make it an rpc. http wraps it.

server/server.go
D3Hunter: those better be done using a middleware, won't do it here

server/server.go
D3Hunter: no need now

server/api/admin.go
rleungx: The HTTP request will redirect automatically, see [server/api/server.go](https://github.com/tikv/pd/blob/6d883d0e8386c64ac12ade25d00893a608b33a14/server/api/server.go#L39)

server/grpc_service.go
D3Hunter: https://github.com/tikv/pd/pull/5181#discussion_r948666633

server/grpc_service.go
D3Hunter: no. return it is ok for now

server/api/admin.go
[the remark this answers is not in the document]
D3Hunter: try this in your local env
```
m := make(map[string]interface{})
	m["a"] = true
	right := m["b'"].(bool)
	fmt.Println(right)
```

server/server.go
rleungx: Will it be better to add a config instead of directly write to etcd?

server/api/admin.go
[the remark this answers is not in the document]
rleungx: What you need is:
```
	m := make(map[string]interface{})
	m["a"] = true
	_, right := m["b"].(bool)
	fmt.Println(right)
```

server/grpc_service.go
rleungx: It affects how caller handles the error, it's better to make it clear.

server/api/admin.go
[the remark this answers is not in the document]
D3Hunter: i forgot. that way we cannot difference between invalid value or not exist
it's optional param

server/server.go
D3Hunter: ok for now

server/api/router.go
```diff
@@ -288,8 +288,13 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	adminHandler := newAdminHandler(svr, rd)
 	registerFunc(clusterRouter, "/admin/cache/region/{id}", adminHandler.DeleteRegionCache, setMethods(http.MethodDelete), setAuditBackend(localLog))
 	registerFunc(clusterRouter, "/admin/cache/regions", adminHandler.DeleteAllRegionCache, setMethods(http.MethodDelete), setAuditBackend(localLog))
-	registerFunc(clusterRouter, "/admin/reset-ts", adminHandler.ResetTS, setMethods(http.MethodPost), setAuditBackend(localLog))
+	registerFunc(apiRouter, "/admin/reset-ts", adminHandler.ResetTS, setMethods(http.MethodPost), setAuditBackend(localLog))
```
rleungx, line 291: Why change it?

server/id/id.go
```diff
@@ -103,58 +108,78 @@ func (alloc *allocatorImpl) Alloc() (uint64, error) {
...
-	resp, err := t.Then(clientv3.OpPut(key, string(value))).Commit()
+	resp, err := txn.If(cmps...).Then(clientv3.OpPut(key, string(value))).Commit()
 	if err != nil {
 		return errs.ErrEtcdTxnInternal.Wrap(err).GenWithStackByArgs()
 	}
 	if !resp.Succeeded {
 		return errs.ErrEtcdTxnConflict.FastGenByArgs()
 	}
 
-	log.Info("idAllocator allocates a new id", zap.String("label", alloc.label), zap.Uint64("alloc-id", end))
```
rleungx, line 154: Current PD recovery documentation relies on it.

server/tso/tso.go
```diff
@@ -237,7 +237,12 @@ func (t *timestampOracle) isInitialized() bool {
 // resetUserTimestamp update the TSO in memory with specified TSO by an atomically way.
 // When ignoreSmaller is true, resetUserTimestamp will ignore the smaller tso resetting error and do nothing.
 // It's used to write MaxTS during the Global TSO synchronization whitout failing the writing as much as possible.
+// cannot set timestamp to one which >= current + maxResetTSGap
 func (t *timestampOracle) resetUserTimestamp(leadership *election.Leadership, tso uint64, ignoreSmaller bool) error {
+	return t.resetUserTimestampInner(leadership, tso, ignoreSmaller, false)
+}
+
+func (t *timestampOracle) resetUserTimestampInner(leadership *election.Leadership, tso uint64, ignoreSmaller, skipUpperBoundCheck bool) error {
```
rleungx, line 245: What is the difference between `resetUserTimestamp` and `resetUserTimestampInner`?

server/api/router.go
D3Hunter: ebs restore phase 1 will reset ts, but at that time the cluser hasn't bootstrapped

server/tso/tso.go
D3Hunter: extra skipUpperBoundCheck param

D3Hunter:
@rleungx  ptal

server/tso/tso.go
rleungx: get

server/server.go
```diff
@@ -1602,3 +1604,46 @@ func (s *Server) IsTTLConfigExist(key string) bool {
...
+	// the value doesn't matter, set to a static string
+	_, err := kv.NewSlowLogTxn(s.client).
+		If(clientv3.Compare(clientv3.CreateRevision(pdRecoveringMarkPath), "=", 0)).
+		Then(clientv3.OpPut(pdRecoveringMarkPath, "on")).
+		Commit()
+	if err != nil {
+		return err
+	}
+	// if other client already marked, return success too
+	return nil
```
rleungx, line 1624:
```suggestion
	return err
```

server/server.go
```diff
@@ -1602,3 +1604,46 @@ func (s *Server) IsTTLConfigExist(key string) bool {
...
+		return false, err
+	}
+	return len(resp.Kvs) > 0, nil
+}
+
+// UnmarkRecovering unmark recovering mark
+func (s *Server) UnmarkRecovering(ctx context.Context) error {
+	log.Info("unmark recovering")
+	_, err := s.client.Delete(ctx, pdRecoveringMarkPath)
+	if err != nil {
```
rleungx, line 1640: ditto

server/id/id.go
```diff
@@ -155,6 +179,8 @@ func (alloc *allocatorImpl) rebaseLocked() error {
 	alloc.metrics.idGauge.Set(float64(end))
 	alloc.end = end
 	alloc.base = end - alloc.step
+	log.Info("rebase id allocator", zap.String("label", alloc.label),
```
rleungx, line 182: Do we need to check `checkCurrEnd` before printing the log?

D3Hunter: print it too. https://github.com/tikv/pd/pull/5181/commits/0f0a83347c4b6ad6eb8c7552c0e9197cbeca840a

D3Hunter: see https://github.com/tikv/pd/pull/5181/files#diff-f7546c173fbab9027155ba08485b80bc7d570970a5e38d86a016d82cf7a6effeR181-R184

server/server.go
rleungx: /cc @nolouch

rleungx (approved):

server/api/router.go
```diff
@@ -288,8 +288,14 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
...
-	registerFunc(clusterRouter, "/admin/reset-ts", adminHandler.ResetTS, setMethods(http.MethodPost), setAuditBackend(localLog))
+	// br ebs restore phase 1 will reset ts, but at that time the cluster hasn't bootstrapped, so cannot use clusterRouter
+	registerFunc(apiRouter, "/admin/reset-ts", adminHandler.ResetTS, setMethods(http.MethodPost), setAuditBackend(localLog))
 	registerFunc(apiRouter, "/admin/persist-file/{file_name}", adminHandler.SavePersistFile, setMethods(http.MethodPost), setAuditBackend(localLog))
+	registerFunc(apiRouter, "/admin/persist-file/{file_name}", adminHandler.SavePersistFile, setMethods(http.MethodPost), setAuditBackend(localLog))
+	registerFunc(apiRouter, "/admin/snapshot-recovering", adminHandler.IsSnapshotRecovering, setMethods(http.MethodGet), setAuditBackend(localLog))
+	registerFunc(apiRouter, "/admin/snapshot-recovering", adminHandler.MarkSnapshotRecovering, setMethods(http.MethodPost), setAuditBackend(localLog))
+	registerFunc(apiRouter, "/admin/snapshot-recovering", adminHandler.UnmarkSnapshotRecovering, setMethods(http.MethodDelete), setAuditBackend(localLog))
```
nolouch, line 297: What actions need to be done before removing the marks? And can you ensure that this mark will be removed? What would be the impact if it wasn't removed?

D3Hunter: ebs-based restore phase 2 will delete data after `backuped resolved-ts`, before that the mark is keeped.
after ebs-based restore phase 2 finish, this mark will be removed

> What would be the impact if it wasn't removed?

this mark is only marked during ebs-based-restore, if it's unmarked after ebs-based restore phase 2 which means it has failed, and the cluster should be destroyed too.

server/api/admin.go
```diff
@@ -127,3 +148,70 @@ func (h *adminHandler) SavePersistFile(w http.ResponseWriter, r *http.Request) {
 	}
 	h.rd.Text(w, http.StatusOK, "")
 }
+
+func (h *adminHandler) MarkSnapshotRecovering(w http.ResponseWriter, r *http.Request) {
+	if err := h.svr.MarkSnapshotRecovering(); err != nil {
+		_ = h.rd.Text(w, http.StatusInternalServerError, err.Error())
```
nolouch, line 154: no need `JSON` format?

server/server.go
```diff
@@ -86,6 +86,8 @@ const (
 	// idAllocPath for idAllocator to save persistent window's end.
 	idAllocPath  = "alloc_id"
 	idAllocLabel = "idalloc"
+
+	pdRecoveringMarkPath = "/pd/snapshot-recovering"
```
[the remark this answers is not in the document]
nolouch: We may replace ${cluster_id} with ${keyspace_id} in APIv2.  In v1, better consist with before.

server/api/admin.go
D3Hunter: yes

nolouch (approved):
lgtm

nolouch:
@D3Hunter  CI failed.

PR #5182, opened 2022-06-19
4504 enhace region scatter observability

### LINKED ISSUE #4504: Can't know why scatter operator failed
## Enhancement Task
Screen Shot 2021-12-27 at 3 32 32 PM

From the monitor, we only can know it fails. And It's hard to know the detailed reason why the scatter operator failed since we only have debug level logs.

matchge-ca:

Issue Number: Close #4504 

- Define new error types in builder
- Increase the metric counter based on the error types

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

None

### DIFF
--- a/server/schedule/operator/builder.go
+++ b/server/schedule/operator/builder.go
@@ -87,6 +87,30 @@ type Builder struct {
 // BuilderOption is used to create operator builder.
 type BuilderOption func(*Builder)
 
+type BuilderNilPeerError struct{}
+
+func (b *BuilderNilPeerError) Error() string {
+	return "cannot build operator for region with nil peer"
+}
+
+type BuilderNoLeaderError struct{}
+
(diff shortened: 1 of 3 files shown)


server/schedule/operator/builder.go
```diff
@@ -87,6 +87,62 @@ type Builder struct {
 // BuilderOption is used to create operator builder.
 type BuilderOption func(*Builder)
 
+type NilPeerError struct{}
```
rleungx, line 90: How about just using errors.New?

HunDunDM (commented):
Something like `b *NilPeerError` can be changed to `b NilPeerError`.
rest LGTM.

matchge-ca:
I revisited this PR and the branch has behind the master a lot, will raise another one to replace this.

PR #5185, opened 2022-06-20
checker: refactor `fixRulePeer`

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/rule_checker.go
+++ b/server/schedule/checker/rule_checker.go
@@ -107,15 +107,10 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
 		// multiple rules.
 		return nil
 	}
-	op, err := c.fixOrphanPeers(region, fit)
-	if err != nil {
-		log.Debug("fail to fix orphan peer", errs.ZapError(err))
-	} else if op != nil {
-		c.pendingList.Remove(region.GetID())
-		return op
-	}
+
+	skip := len(fit.OrphanPeers) == 0
 	for _, rf := range fit.RuleFits {
-		op, err := c.fixRulePeer(region, fit, rf)
+		op, err := c.fixRulePeer(region, fit, rf, skip)
 		if err != nil {
 			log.Debug("fail to fix rule peer", zap.String("rule-group", rf.Rule.GroupID), zap.String("rule-id", rf.Rule.ID), errs.ZapError(err))
 			continue
@@ -125,6 +120,7 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
 			return op
 		}
 	}
+
 	if c.cluster.GetOpts().IsPlacementRulesCacheEnabled() {
 		if placement.ValidateFit(fit) && placement.ValidateRegion(region) && placement.ValidateStores(fit.GetRegionStores()) {
 			// If there is no need to fix, we will cache the fit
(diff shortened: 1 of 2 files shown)


server/schedule/checker/rule_checker.go
```diff
@@ -135,22 +131,64 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
-		if c.isOfflinePeer(peer) {
-			checkerCounter.WithLabelValues("rule_checker", "replace-offline").Inc()
-			return c.replaceUnexpectRulePeer(region, rf, fit, peer, offlineStatus)
+
+		// check orphan peer.
+		if !skipOrphan {
+			for _, pendingPeer := range region.GetPendingPeers() {
+				if pendingPeer.Id == peer.Id {
+					checkerCounter.WithLabelValues("rule_checker", "skip-remove-pending-peer").Inc()
+					skipOrphan = true
```
AndreMouche, line 163: could we break here?

server/schedule/checker/rule_checker.go
```diff
@@ -135,22 +131,64 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
+				}
+			}
+			for _, downPeer := range region.GetDownPeers() {
```
AndreMouche, line 166: should we check `skipOrphan` before loop?

server/schedule/checker/rule_checker.go
```diff
@@ -135,22 +131,64 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
+				if downPeer.Peer.Id == peer.Id {
+					checkerCounter.WithLabelValues("rule_checker", "skip-remove-down-peer").Inc()
+					skipOrphan = true
```
AndreMouche, line 169: ditto, could we break here?

server/schedule/checker/rule_checker.go
```diff
@@ -135,22 +131,64 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
+		if selectedPeer == nil && c.isOfflinePeer(peer) {
+			selectedPeer, kind = peer, offlineStatus
 		}
-		if c.isOfflinePeer(peer) {
-			checkerCounter.WithLabelValues("rule_checker", "replace-offline").Inc()
-			return c.replaceUnexpectRulePeer(region, rf, fit, peer, offlineStatus)
+
+		// check orphan peer.
+		if !skipOrphan {
+			for _, pendingPeer := range region.GetPendingPeers() {
```
AndreMouche, line 160: how about define some function like  containsPeer(peers,peerID) bool?
if it is only used in this func, define it as a temp variable should be ok, for example,
```
	constainsPeer := func(peers []*pdpb.PeerStats, peerID uint64) bool {
		for _, peer := range peers {
			if peer.Peer.Id == peerID {
				return true
			}
		}
		return false
	}
```

server/schedule/checker/rule_checker.go
```diff
@@ -135,22 +131,64 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
+		}
+		if selectedPeer == nil && c.isOfflinePeer(peer) {
+			selectedPeer, kind = peer, offlineStatus
 		}
-		if c.isOfflinePeer(peer) {
-			checkerCounter.WithLabelValues("rule_checker", "replace-offline").Inc()
-			return c.replaceUnexpectRulePeer(region, rf, fit, peer, offlineStatus)
+
+		// check orphan peer.
+		if !skipOrphan {
```
AndreMouche, line 159: Too many {} in {}, could we continue directly when skipOrphan is true?

server/schedule/checker/rule_checker.go
```diff
@@ -107,15 +107,10 @@ func (c *RuleChecker) CheckWithFit(region *core.RegionInfo, fit *placement.Regio
...
-		log.Debug("fail to fix orphan peer", errs.ZapError(err))
-	} else if op != nil {
-		c.pendingList.Remove(region.GetID())
-		return op
-	}
+
+	skip := len(fit.OrphanPeers) == 0
 	for _, rf := range fit.RuleFits {
-		op, err := c.fixRulePeer(region, fit, rf)
+		op, err := c.fixRulePeer(region, fit, rf, skip)
```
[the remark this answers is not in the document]
rleungx: It's unnecessary to check it every loop.

server/schedule/checker/rule_checker.go
rleungx: What do you mean `continue`?

server/schedule/checker/rule_checker.go
rleungx: No, we can't.

server/schedule/checker/rule_checker.go
rleungx: The down peer and pending peer are stored in the different struct, which cannot share the same function.

rleungx:
Close it for now.

Issue #3952, opened 2021-08-04
config: check whether is valid for config of hot scheduler in http


lhy1024:
### Describe your feature request related problem

After #3920, we only check config for config of hot scheduler in pd-ctl, we also need check it in http

### Teachability, Documentation, Adoption, Migration Strategy

Go

aidendou:
hi @lhy1024, I am onboarding to TiDB at work, trying to take some of the PD issues. For this particular one, are we just looking to be able to get the hot region scheduler config via restful api?

lhy1024:
> hi @lhy1024, I am onboarding to TiDB at work, trying to take some of the PD issues. For this particular one, are we just looking to be able to get the hot region scheduler config via restful api?

Now we can configure the hot scheduler via http, but we don't check if the setting is valid

njuwelkin:
Let me repeat the requirement, just to avoid mis-understandings.
On the client side(pd-ctl), we do http request validations like  below:
`
if priority != schedulers.BytePriority && priority != schedulers.KeyPriority && priority != schedulers.QueryPriority {
    ....
}
if priority == schedulers.QueryPriority && key == "write-peer-priorities" {
    ....
}
`
But actually we need do this on server side. For example, maybe can do it in handler AddScheduler().

Am I right? Or It's actually a more complex issue than what I though?

lhy1024:
> 



> Let me repeat the requirement, just to avoid mis-understandings. On the client side(pd-ctl), we do http request validations like below: `if priority != schedulers.BytePriority && priority != schedulers.KeyPriority && priority != schedulers.QueryPriority { .... } if priority == schedulers.QueryPriority && key == "write-peer-priorities" { .... }` But actually we need do this on server side. For example, in handler AddScheduler().
> 
> Am I right? Or It's actually a more complex issue than what I though?

yes, we hope to check if the setting is valid when set config via http.

njuwelkin:
i'd like to fix this, pls assign to me.

njuwelkin:
i'd like to fix this, pls assign to me.

PR #3920, opened 2021-07-28
scheduler: support priority for each dimension in hot scheduling

Yisaer:
Signed-off-by: yisaer <disxiaofei@163.com>

### What is changed and how it works?

support priority for each dimension in hot scheduling

### Check List

Tests

- Unit test

I used `oltp_read_only` to test the dimension priority in hot scheduling, here is the result.

image

### Release note

```release-note
Support priority for each dimension in hot scheduleing
```

PR #5208, opened 2022-06-21
server/scheduler: check whether is valid for config of hot scheduler …

njuwelkin:
…in http

After https://github.com/tikv/pd/pull/3920, we only check config for config of hot scheduler in pd-ctl, we also need check it in http.

Issue Number: Close tikv/pd#3952

Added a method 'validPriority' for structure hotRegionSchedulerConfig.
when receiving a http request to config hot region scheduler, we make below validations:
- whether scheduling dimension is valid.
- whether have at least two priorities
- whether the priorities are repeated
- whether the priority is set to QueryPriority for write-peer-priorities

```commit-message
```

### Check List
Tests
- Unit test

### Release note
None


server/schedulers/hot_region_config.go
```diff
@@ -299,6 +301,38 @@ func (conf *hotRegionSchedulerConfig) handleGetConfig(w http.ResponseWriter, r *
 	rd.JSON(w, http.StatusOK, conf.getValidConf())
 }
 
+func (conf *hotRegionSchedulerConfig) validPriority() error {
+	isValid := func(priorities []string) (map[string]bool, error) {
+		priorityMap := map[string]bool{}
+		for _, p := range priorities {
+			if p != BytePriority && p != KeyPriority && p != QueryPriority {
+				return nil, errs.ErrSchedulerConfig.FastGenByArgs("Invalid scheduling dimensions.")
```
lhy1024, line 309:
```suggestion
				return nil, errs.ErrSchedulerConfig.FastGenByArgs("invalid scheduling dimensions.")
```

lhy1024 (commented):
the rest LGTM

lhy1024:
plz sign commit, and let me know if you encounter any problems

njuwelkin:
there were two commits, and the 1st was not signed. so I merged two commits to one, in order to sign the 1st one.

lhy1024 (approved):

HunDunDM (approved):

PR #5193, opened 2022-06-21
*: migrate test framework to testify

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

### What problem does this PR solve?

Issue Number: Ref #4813.

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

PR #5211, opened 2022-06-22
*: some misc modify for testify

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/4813.

wait #5193

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/errs/errs_test.go
+++ b/pkg/errs/errs_test.go
@@ -141,5 +141,5 @@ func TestErrorWithStack(t *testing.T) {
 	re.GreaterOrEqual(idx1, -1)
 	idx2 := strings.Index(m2, "[stack=")
 	re.GreaterOrEqual(idx2, -1)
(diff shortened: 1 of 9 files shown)


pkg/cache/cache_test.go
```diff
@@ -32,10 +32,10 @@ func TestExpireRegionCache(t *testing.T) {
 	// Test Pop
 	cache.PutWithTTL(9, "9", 5*time.Second)
 	cache.PutWithTTL(10, "10", 5*time.Second)
-	re.Equal(2, cache.Len())
+	re.Len(cache, 2)
```
JmPotato, line 35: `.Len()` is a method of `cache`, it's not the `re.Len` can only be used for Array, Chan, Map, Slice, or String.

JmPotato (approved):

nolouch (approved):

PR #5235, opened 2022-06-27
filter: use `labelConstraintFilter` instead

rleungx:

Issue Number: Ref #5257.

This PR removes the `engineFilter`, `ordinaryEngineFilter` and `specialUseFilter` since they are all special cases of `labelConstraintFilter`. For `specialUseFilter`, this PR also removes the conditional judgment of low space.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/checker/replica_strategy.go
+++ b/server/schedule/checker/replica_strategy.go
@@ -19,7 +19,6 @@ import (
 	"github.com/tikv/pd/server/core"
 	"github.com/tikv/pd/server/schedule"
 	"github.com/tikv/pd/server/schedule/filter"
-	"github.com/tikv/pd/server/schedule/placement"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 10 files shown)


server/schedule/checker/replica_strategy.go
```diff
@@ -55,7 +56,8 @@ func (s *ReplicaStrategy) SelectStoreToAdd(coLocationStores []*core.StoreInfo, e
 	filters := []filter.Filter{
 		filter.NewExcludedFilter(s.checkerName, nil, s.region.GetStoreIDs()),
 		filter.NewStorageThresholdFilter(s.checkerName),
-		filter.NewSpecialUseFilter(s.checkerName),
```
JmPotato, line 58: Could we keep this function as an alias of `filter.NewLabelConstaintFilter` rather than removing it totally? Otherwise, we need to write long code every time we need the same usage filter.

rleungx: use some variables to simplify the code, PTAL

HunDunDM (approved):

server/schedule/checker/rule_checker.go
```diff
@@ -371,7 +371,7 @@ func (c *RuleChecker) strategy(region *core.RegionInfo, rule *placement.Rule) *R
 		isolationLevel: rule.IsolationLevel,
 		locationLabels: rule.LocationLabels,
 		region:         region,
-		extraFilters:   []filter.Filter{filter.NewLabelConstaintFilter(c.name, rule.LabelConstraints)},
+		extraFilters:   []filter.Filter{filter.NewLabelConstaintFilter(c.name, rule.LabelConstraints, true)},
```
JmPotato, line 374: Should be false?

server/schedule/region_scatterer.go
```diff
@@ -129,7 +130,7 @@ func NewRegionScatterer(ctx context.Context, cluster Cluster) *RegionScatterer {
 		ctx:            ctx,
 		name:           regionScatterName,
 		cluster:        cluster,
-		ordinaryEngine: newEngineContext(ctx, filter.NewOrdinaryEngineFilter(regionScatterName)),
+		ordinaryEngine: newEngineContext(ctx, filter.NewLabelConstaintFilter(regionScatterName, filter.NotTiFlashEngine, false)),
```
CabinfeverB, line 133: How about using core.EngineTiKV

CabinfeverB (approved):
rest LGTM

server/schedule/region_scatterer.go
rleungx: It's different because we use `NotIn`, the store without any label will be counted. But we use `In`, we must set that label for the store.

PR #5237, opened 2022-06-28
storage: optimize the storage interface

HunDunDM:

Issue Number: Ref #3453

Currently, in the interface definition of `RegionStorage`, `LoadRegions` and `LoadRegionsOnce` are repeated, and the implementation of the `LoadRegionsOnce` is not always consistent with the definition. In addition, some interface conversions based on `coreStorage` are tricky.

```commit-message
storage: optimize the storage interface
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/storage/storage.go
+++ b/server/storage/storage.go
@@ -83,23 +83,60 @@ func NewCoreStorage(defaultStorage Storage, regionStorage endpoint.RegionStorage
 	}
 }
 
-// GetRegionStorage gets the internal region storage.
-func (ps *coreStorage) GetRegionStorage() endpoint.RegionStorage {
-	return ps.regionStorage
+// TryGetLocalRegionStorage gets the local region storage. Returns nil if not present.
+func TryGetLocalRegionStorage(s Storage) endpoint.RegionStorage {
+	switch ps := s.(type) {
+	case *coreStorage:
+		return ps.regionStorage
+	case *levelDBBackend, *memoryStorage:
+		return ps
+	default:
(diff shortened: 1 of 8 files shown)


JmPotato (commented):
Basically LGTM. Would you like to add some test cases for the newly added function `TrySwitchRegionStorage`?

nolouch (approved):
lgtm

nolouch:
> Basically LGTM. Would you like to add some test cases for the newly added function `TrySwitchRegionStorage`?

cc @HunDunDM

AndreMouche (approved):
LGTM

server/storage/storage.go
```diff
@@ -135,6 +139,22 @@ func (ps *coreStorage) LoadRegionsOnce(ctx context.Context, f func(region *core.
 	return nil
 }
 
+// LoadRegion loads one region from storage.
+func (ps *coreStorage) LoadRegion(regionID uint64, region *metapb.Region) (ok bool, err error) {
```
rleungx, line 143: Seem this function only be used in the tests

nolouch:
ptal @HunDunDM

lhy1024 (approved):

server/storage/storage.go
HunDunDM: It is an implementation of interface.

rleungx: Yes, but it's only used for tests.

Issue #4684, opened 2022-02-28
Add metrics, API and processing logic for Disconnect Region


HunDunDM:
### Describe your feature request related problem

When the Region cannot elect a Leader due to some problems, the Region will not be able to send a heartbeat to the PD. The relevant information in the PD will always maintain the information of the last heartbeat. There are currently no statistics for this type of Region.

### Describe the feature you'd like

Add `Disconnect Region` statistics in `Region Healthy` and add related APIs.

PR #5272, opened 2022-07-06
checker: add a checker to check abnormal regions

LLThomas:

Issue Number: Ref #4684

In order to track abnormal regions, this pr has built a generic checker that currently only detects regions that have expired due to machine downtime.
```commit-message
checker: add a checker to record abnormal regions
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


JmPotato (commented):
Why not implement this feature through [RegionStatistics](https://github.com/tikv/pd/blob/3abc703fa00cef164cd75426c7590d59bd0649c1/server/statistics/region_collection.go#L152)? `Checker` should be more like a struct that will generate certain operators for scheduling purpose.

LLThomas:
> Why not implement this feature through [RegionStatistics](https://github.com/tikv/pd/blob/3abc703fa00cef164cd75426c7590d59bd0649c1/server/statistics/region_collection.go#L152)? `Checker` should be more like a struct that will generate certain operators for scheduling purpose.

You are right, I misunderstood the issue. I'll make a change.

rleungx:
FYI, https://github.com/tikv/pd/pull/3901 is another way to implement it. But there is an extra cost since we need to scan more times.

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/tikv/pd/server/config"
+	"github.com/tikv/pd/server/core"
+)
+
+// regionStateType represents the type of region's state.
+type regionStateType uint32
+
+// region state type
+const (
+	RegionStateDown regionStateType = 1 << iota
```
HunDunDM, line 30: Variables of type private should not be public.
Why use `1 << iota`?

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// regionState is used to record abnormal regions.
+type regionState struct {
+	sync.RWMutex
+	opt    *config.PersistOptions
+	states map[regionStateType]map[uint64]*core.RegionInfo
```
HunDunDM, line 37: Replace `map` with a fixed-length array.

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+type regionStateType uint32
+
+// region state type
+const (
+	RegionStateDown regionStateType = 1 << iota
+)
+
+// regionState is used to record abnormal regions.
+type regionState struct {
+	sync.RWMutex
```
HunDunDM, line 35: The current design does not require a lock?

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+	r := &regionState{
+		opt:    opt,
+		states: make(map[regionStateType]map[uint64]*core.RegionInfo),
+	}
+	r.states[RegionStateDown] = make(map[uint64]*core.RegionInfo)
+	return r
+}
+
+// GetRegionStateByType gets the states of the region by types. The regions here need to be cloned, otherwise, it may cause data race problems.
+func (r *regionState) getRegionStateByType(typ regionStateType) []*core.RegionInfo {
```
HunDunDM, line 51: Is this function only used in tests?

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+	defer r.RUnlock()
+	res := make([]*core.RegionInfo, 0, len(r.states[typ]))
+	for _, r := range r.states[typ] {
+		res = append(res, r.Clone())
+	}
+	return res
+}
+
+// Check verifies a region's state, recording it if need.
+func (r *regionState) observe(region *core.RegionInfo) {
```
HunDunDM, line 62: Maybe we could pass in a slice of regions and compute `time.Now` only once.

server/cluster/region_stat.go
```diff
@@ -0,0 +1,81 @@
...
+	if time.Now().UnixNano()-int64(region.GetInterval().GetEndTimestamp()) >= r.opt.GetMaxStoreDownTime().Nanoseconds() {
+		_, exist := r.states[RegionStateDown][regionID]
+		if !exist {
+			r.states[RegionStateDown][regionID] = region
+		}
+	}
+}
+
+// Collect collects the metrics of the regions' states.
+func (r *regionState) collect() {
```
HunDunDM, line 77: How to clean up a region if it does not exist in the region tree?

nolouch:
PTAL @HunDunDM

JmPotato (approved):

LLThomas:
Picture of down-region-count :

image

server/cluster/region_stat.go
```diff
@@ -0,0 +1,94 @@
...
+// When the number of times the collect() function is executed is exceeds this threshold,
+// regions that are not in region tree will be deleted.
+const clearThreshold = 10
+
+// regionStateType represents the type of region's state.
+type regionStateType int
+
+// region state type
+const (
+	regionStateDown regionStateType = iota
```
HunDunDM, line 34: Consider replacing `Down` with `NoLeader` to prevent confusion with `down-peer`?

server/cluster/metrics.go
```diff
@@ -119,6 +119,14 @@ var (
 			Name:      "store_sync",
 			Help:      "The state of store sync config",
 		}, []string{"address", "state"})
+
+	regionStateGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
```
HunDunDM, line 124: A bit of a duplication with `regionStatusGauge` of `server/statistics/metrics.go`.

server/cluster/region_stat.go
```diff
@@ -0,0 +1,94 @@
...
+	regionStateTypeLen
+)
+
+// regionState is used to record abnormal regions which could not be detected by the region heartbeat.
+// For example, the down regions won’t send any heartbeat anymore.
+type regionState struct {
+	cluster schedule.Cluster
+	states  [regionStateTypeLen]map[uint64]*core.RegionInfo
+	count   int
```
HunDunDM, line 43: Use a more specific name.

nolouch:
ptal @LLThomas

PR #5277, opened 2022-07-06
schedule/placement/fit: cache the match info for rules to peers

### LINKED ISSUE #5276: server/schedule/placement/fit: pick out the candidates for a rule should be faster when cache the match result for peers to rules
## Enhancement Task
The function `fitRule(idx)` are try to find the best fit recursively, and the following part is to pick out the candidates for the current rule, which may repeatedly call the func `MatchLabelConstraints ` when there are lots of rules. so, how about cache the result that if the peer(store, since 1 store can only have 1 peer, so we could use the peer directly) can match the rule
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/placement/fit.go#L214-L225](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/placement/fit.go#L214-L225)

AndreMouche:

Issue Number: Close #5276 

### Check List

Tests

- Unit test

- benchmark
#### current branch
```
➜  placement git:(fit_cache) ✗ go test -benchmem -run=^$ -bench ^BenchmarkFitRegionWithMoreRulesAndStoreLabels -benchtime=100x
goos: darwin
goarch: amd64
pkg: github.com/tikv/pd/server/schedule/placement
cpu: VirtualApple @ 2.50GHz
BenchmarkFitRegionWithMoreRulesAndStoreLabels-8              100           5371767 ns/op          356880 B/op      12879 allocs/op
PASS
ok      github.com/tikv/pd/server/schedule/placement    1.601s
```

#### master
```
➜  placement git:(master) ✗ go test -benchmem -run=^$ -bench ^BenchmarkFitRegionWithMoreRulesAndStoreLabels -benchtime=100x
goos: darwin
goarch: amd64
pkg: github.com/tikv/pd/server/schedule/placement
cpu: VirtualApple @ 2.50GHz
BenchmarkFitRegionWithMoreRulesAndStoreLabels-8              100          47968061 ns/op         2627010 B/op     107788 allocs/op
PASS
ok      github.com/tikv/pd/server/schedule/placement    5.801s
```

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/placement/fit.go
+++ b/server/schedule/placement/fit.go
@@ -155,12 +155,15 @@ func fitRegion(stores []*core.StoreInfo, region *core.RegionInfo, rules []*Rule)
 }
 
 type fitWorker struct {
-	stores        []*core.StoreInfo
-	bestFit       RegionFit  // update during execution
-	peers         []*fitPeer // p.selected is updated during execution.
-	rules         []*Rule
-	needIsolation bool
-	exit          bool
+	stores  []*core.StoreInfo
+	bestFit RegionFit  // update during execution
+	peers   []*fitPeer // p.selected is updated during execution.
+	rules   []*Rule
+	// caches for match info for rules to peers. match,exist:= rulesMatchCache[i][j]
+	// exist is false means haven't been cached. match is true means peers[j] can match rules[i]
+	rulesMatchCache map[int]map[int]bool
+	needIsolation   bool
+	exit            bool
 }
 
 func newFitWorker(stores []*core.StoreInfo, region *core.RegionInfo, rules []*Rule) *fitWorker {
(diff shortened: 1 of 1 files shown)


server/schedule/placement/fit.go
```diff
@@ -180,12 +182,21 @@ func newFitWorker(stores []*core.StoreInfo, region *core.RegionInfo, rules []*Ru
 		return si > sj || (si == sj && peers[i].GetId() < peers[j].GetId())
 	})
 
+	matchRules := make(map[int]map[int]bool, len(rules))
+	for id, rule := range rules {
+		matchPeers := make(map[int]bool, 0)
+		if checkRule(rule, stores) {
+			matchRules[id] = make(map[int]bool)
+		}
+		matchRules[id] = matchPeers
```
rleungx, line 191: It seems `matchRules[id]` is always equal to `matchPeers`, so what is the purpose of the if condition here?

PR #5281, opened 2022-07-06
client, tests: make client check if the cluster ID matches

### LINKED ISSUE #5278: lacking legality checks for pdAddrs in pd.NewClientWithContext()
## Enhancement Task

Hi, here is a problem I encountered that requires an enhancement. 

When trying to use `pd.NewClientWithContext()` to get a new PDClient, it seems that not adequate checks were done on the legality of the given PD addresses. 
If the user passed in unconnectable addresses mixed with correct ones, or, if the user mixed the addresses of 2 PD clusters, the method will still yield a correct PD client. Could we add those checks and returned an error in it? 

In TiCDC, it's often the case when we need such checks since users may handle multiple TiDB clusters at the same time. And, with the progress of decoupling CDC and TiDB, that scenario will be more. Not to mention the code will be redundant and inefficient when trying to do those checks outside the PD.   

The example scenarios I mention are as follows. 
Given 2 Pd clusters
- PD cluster 1: `[http://127.0.0.1:xxx1, http://127.0.0.1:xxx2, http://127.0.0.1:xxx3]`
- PD cluster 2: `[http://127.0.0.1:xxx4, http://127.0.0.1:xxx5, http://127.0.0.1:xxx6]`

The problem is that if the user uses `[http://127.0.0.1:xxx5, http://127.0.0.1:xxx6, http://127.0.0.1:xxx7]` (one address unconnectable) or `[http://127.0.0.1:xxx3, http://127.0.0.1:xxx4, http://127.0.0.1:xxx5]` (mixed), PDClient can be created and no error thrown

JmPotato:

Issue Number: Close #5278.

```commit-message
Let the client check if the cluster ID matches during the initialization and updating.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/base_client.go
+++ b/client/base_client.go
@@ -245,22 +245,46 @@ func (c *baseClient) gcAllocatorLeaderAddr(curAllocatorMap map[string]*pdpb.Memb
 func (c *baseClient) initClusterID() error {
 	ctx, cancel := context.WithCancel(c.ctx)
 	defer cancel()
+	var clusterID uint64
 	for _, u := range c.GetURLs() {
 		members, err := c.getMembers(ctx, u, c.option.timeout)
 		if err != nil || members.GetHeader() == nil {
 			log.Warn("[pd] failed to get cluster id", zap.String("url", u), errs.ZapError(err))
 			continue
 		}
-		c.clusterID = members.GetHeader().GetClusterId()
-		return nil
+		if clusterID == 0 {
+			clusterID = members.GetHeader().GetClusterId()
+			continue
+		}
+		failpoint.Inject("skipClusterIDCheck", func() {
+			failpoint.Continue()
+		})
+		// All URLs passed in should have the same cluster ID.
+		if members.GetHeader().GetClusterId() != clusterID {
+			return errors.WithStack(errUnmatchedClusterID)
+		}
 	}
-	return errors.WithStack(errFailInitClusterID)
+	// Failed to init the cluster ID.
+	if clusterID == 0 {
+		return errors.WithStack(errFailInitClusterID)
+	}
+	c.clusterID = clusterID
+	return nil
 }
 
 func (c *baseClient) updateMember() error {
-	for _, u := range c.GetURLs() {
+	for i, u := range c.GetURLs() {
+		failpoint.Inject("skipFirstUpdateMember", func() {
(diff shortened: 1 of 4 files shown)


client/base_client.go
```diff
@@ -245,22 +245,46 @@ func (c *baseClient) gcAllocatorLeaderAddr(curAllocatorMap map[string]*pdpb.Memb
...
+		failpoint.Inject("skipFirstUpdateMember", func() {
+			if i == 0 {
+				failpoint.Continue()
+			}
+		})
 		members, err := c.getMembers(c.ctx, u, updateMemberTimeout)
-
+		// Check the cluster ID.
+		if err == nil && members.GetHeader().GetClusterId() != c.clusterID {
+			err = errs.ErrClientUpdateMember.FastGenByArgs("cluster id does not match")
```
bufferflies, line 285: why not use "errors.WithStack(errUnmatchedClusterID)"?

JmPotato: I think it's because `errUnmatchedClusterID` is an error that will be returned to the upper caller, e.g, the TiDB. So it's not wrapped by the errs definition. However, in this function, it's an error that will be logged out, so it's necessary to wrap it first.

crelax:
Hi, may I ask what's the expected behavior if some PD addresses are not connectable?

JmPotato:
> Hi, may I ask what's the expected behavior if some PD addresses are not connectable?

I didn't add special handling logic for this because it's hard to tell if "unreachable" is a temporary state or a permanent error from the client's point of view. Given the availability, as long as there is a working PD address in it, then I think the client is ready to start. It is unacceptable to deny the entire service just because of an unreachable URL.

lhy1024 (approved):

rleungx (approved):

crelax:
> 



> > Hi, may I ask what's the expected behavior if some PD addresses are not connectable?
> 
> I didn't add special handling logic for this because it's hard to tell if "unreachable" is a temporary state or a permanent error from the client's point of view. Given the availability, as long as there is a working PD address in it, then I think the client is ready to start. It is unacceptable to deny the entire service just because of an unreachable URL.

Got it. Thx for the explanation!

PR #5284, opened 2022-07-07
id: make id allocator general purpose

### LINKED ISSUE #5294: make id allocator more general purpose
## Development Task

A new `keyspace id` is needed every-time a new `keyspace` is created. PD already has an `idAllocator` that utilize persistent window boundary to efficiently assign new ID. It would be great if its implementation can be reused for `keyspace` management.

However, the current implementation of id allocator permits only one id allocator instance due to its fixed `alloc path` $rootPath/alloc_id and metrics label.

Currently, the single id allocator instance(`server.idAllocator`) was used for assigning regionID, storeID, peerID. This makes it non-ideal for allocating keyspace id, which has different bounds (1 - uint24MAX) and would preferable to be less sparse. 

Therefore, parametrize id allocator's alloc path and metrics label becomes necessary.

AmoebaProtozoa:
Signed-off-by: David <8039876+AmoebaProtozoa@users.noreply.github.com>

The current implementation of id allocator permits only one id allocator due to fixed alloc path `$rootPath/alloc_id` and metrics label.
This PR attempts to make it more general purpose by parametrize its alloc path and metrics label.
Issue Number: Close #5294 

```commit-message
make id allocator general purpose
```

### Check List

Tests

- Unit tests

### Release note

```release-note
None.
```

### DIFF
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -55,7 +55,7 @@ type allocatorImpl struct {
 }
 
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
+func NewAllocator(client *clientv3.Client, rootPath string, allocPath string, label string, member string) Allocator {
 	return &allocatorImpl{
 		client:    client,
 		rootPath:  rootPath,
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -50,8 +50,8 @@ type allocatorImpl struct {
 	rootPath  string
 	allocPath string
 	label     string
-	metrics   *metrics
 	member    string
+	metrics   *metrics
 }
 
 // metrics is a collection of idAllocator's metrics
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -18,6 +18,7 @@ import (
 	"path"
 
 	"github.com/pingcap/log"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/etcdutil"
 	"github.com/tikv/pd/pkg/syncutil"
@@ -45,14 +46,24 @@ type allocatorImpl struct {
 	base uint64
 	end  uint64
 
-	client   *clientv3.Client
-	rootPath string
-	member   string
+	client    *clientv3.Client
+	rootPath  string
+	allocPath string
+	label     string
+	metric    prometheus.Gauge
+	member    string
 }
 
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath string, member string) Allocator {
-	return &allocatorImpl{client: client, rootPath: rootPath, member: member}
+func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
+	return &allocatorImpl{
+		client:    client,
+		rootPath:  rootPath,
+		allocPath: allocPath,
+		label:     label,
+		metric:    idGauge.WithLabelValues(label),
+		member:    member,
+	}
 }
 
 // Alloc returns a new id.
@@ -119,13 +130,13 @@ func (alloc *allocatorImpl) rebaseLocked() error {
 		return errs.ErrEtcdTxnConflict.FastGenByArgs()
 	}
 
-	log.Info("idAllocator allocates a new id", zap.Uint64("alloc-id", end))
-	idallocGauge.Set(float64(end))
+	log.Info("idAllocator allocates a new id", zap.String("label", alloc.label), zap.Uint64("alloc-id", end))
+	alloc.metric.Set(float64(end))
 	alloc.end = end
 	alloc.base = end - allocStep
 	return nil
 }
 
 func (alloc *allocatorImpl) getAllocIDPath() string {
-	return path.Join(alloc.rootPath, "alloc_id")
+	return path.Join(alloc.rootPath, alloc.allocPath)
 }
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -66,8 +66,8 @@ func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member st
 		rootPath:  rootPath,
 		allocPath: allocPath,
 		label:     label,
-		metrics:   &metrics{idGauge: idGauge.WithLabelValues(label)},
 		member:    member,
+		metrics:   &metrics{idGauge: idGauge.WithLabelValues(label)},
 	}
 }
 
--- a/server/id/id.go
+++ b/server/id/id.go
@@ -50,18 +50,23 @@ type allocatorImpl struct {
 	rootPath  string
 	allocPath string
 	label     string
-	metric    prometheus.Gauge
+	metrics   *metrics
 	member    string
 }
 
+// metrics is a collection of idAllocator's metrics
+type metrics struct {
+	idGauge prometheus.Gauge
+}
+
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath string, allocPath string, label string, member string) Allocator {
+func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
 	return &allocatorImpl{
 		client:    client,
 		rootPath:  rootPath,
 		allocPath: allocPath,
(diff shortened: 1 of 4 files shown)


rleungx:
Can you describe more details of your use case?

AmoebaProtozoa:
> Can you describe more details of your use case?

AmoebaProtozoa:
> Can you describe more details of your use case?

This is mostly for the Keyspace management in PD. 
Keyspace needs a new id everytime a keyspace is created, something very similar to what id allocator does. It would be great if we can reuse id allocator's logic.
However, since it need to be stored on a different storage path and impose some extra restrictions, using the same id allocator instance doesn't seems possible.

server/id/id.go
```diff
@@ -45,14 +46,24 @@ type allocatorImpl struct {
...
+	allocPath string
+	label     string
+	metric    prometheus.Gauge
+	member    string
 }
 
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath string, member string) Allocator {
-	return &allocatorImpl{client: client, rootPath: rootPath, member: member}
+func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
```
ystaticy, line 58: The data type of the parameter should be defined here?
For example，```rootPath``` should be ```rootPath string``` ?

AmoebaProtozoa: It can be omitted since they are of the same type. (there is a string at the end)
or is this more of a concern of style?

server/server.go
```diff
@@ -381,7 +381,7 @@ func (s *Server) startServer(ctx context.Context) error {
 	s.member.SetMemberDeployPath(s.member.ID())
 	s.member.SetMemberBinaryVersion(s.member.ID(), versioninfo.PDReleaseVersion)
 	s.member.SetMemberGitHash(s.member.ID(), versioninfo.PDGitHash)
-	s.idAllocator = id.NewAllocator(s.client, s.rootPath, s.member.MemberValue())
+	s.idAllocator = id.NewAllocator(s.client, s.rootPath, "alloc_id", "idalloc", s.member.MemberValue())
```
ystaticy, line 384: Suggest to define constants for ```alloc_id``` and ```idalloc```

server/id/id.go
ystaticy, line 58: Suggest to add test case for different allocPath

server/id/id_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// put memberValue to leaderPath to simulate an election success.
+	_, err = client.Put(context.Background(), leaderPath, memberVal)
+	re.NoError(err)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < 3; i++ {
+		iStr := strconv.Itoa(i)
+		wg.Add(1)
+		// allocators share rootPath and memberVal, but have different allocPaths and labels
```
ystaticy, line 64: ```// All allocators share rootPath and memberVal, but they have different allocPaths and labels.```

server/id/id_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	ep := cfg.LCUrls[0].String()
+	client, err := clientv3.New(clientv3.Config{
+		Endpoints: []string{ep},
+	})
+	re.NoError(err)
+
+	<-etcd.Server.ReadyNotify()
+
+	// put memberValue to leaderPath to simulate an election success.
```
ystaticy, line 56: '''Put'''

server/id/id.go
```diff
@@ -55,7 +55,7 @@ type allocatorImpl struct {
 }
 
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
+func NewAllocator(client *clientv3.Client, rootPath string, allocPath string, label string, member string) Allocator {
```
sunxiaoguang, line 58: why you want to change arguments to specify identical type for each individual argument?

AmoebaProtozoa: It was suggested to stay consistent with the style https://github.com/tikv/pd/pull/5284#discussion_r915597292

server/id/id.go
```diff
@@ -119,13 +130,13 @@ func (alloc *allocatorImpl) rebaseLocked() error {
 		return errs.ErrEtcdTxnConflict.FastGenByArgs()
 	}
 
-	log.Info("idAllocator allocates a new id", zap.Uint64("alloc-id", end))
-	idallocGauge.Set(float64(end))
+	log.Info("idAllocator allocates a new id", zap.String("label", alloc.label), zap.Uint64("alloc-id", end))
+	alloc.metric.Set(float64(end))
```
nolouch, line 134: Could you enhance the readable? maybe like` alloc.metrics.idallocgauge.Set(float64(end))`, and also helpful for adding more metrics for each allocator.

ystaticy:
LGTM

ystaticy (approved):

server/id/id_test.go
```diff
@@ -0,0 +1,83 @@
...
+		}(re, allocator)
+	}
+	wg.Wait()
+}
+
+// testAllocator keep updated given allocator and check if values are expected
+func testAllocator(re *require.Assertions, allocator Allocator) {
+	startID, err := allocator.Alloc()
+	re.NoError(err)
+	for i := startID + 1; i < startID+allocStep*2; i++ {
```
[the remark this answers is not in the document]
AmoebaProtozoa: I have increased it to startID + allocStep*20

server/id/id.go
```diff
@@ -119,13 +135,13 @@ func (alloc *allocatorImpl) rebaseLocked() error {
 		return errs.ErrEtcdTxnConflict.FastGenByArgs()
 	}
 
-	log.Info("idAllocator allocates a new id", zap.Uint64("alloc-id", end))
-	idallocGauge.Set(float64(end))
+	log.Info("idAllocator allocates a new id", zap.String("label", alloc.label), zap.Uint64("alloc-id", end))
```
rleungx, line 138: The docs of the PD recovery process may need to be updated.

rleungx (approved):
The rest LGTM.

server/id/id.go
sunxiaoguang: Curiously, is there anything broken so we have to update PD recovery process?

server/id/id.go
```diff
@@ -45,14 +46,29 @@ type allocatorImpl struct {
...
+	client    *clientv3.Client
+	rootPath  string
+	allocPath string
+	label     string
+	member    string
+	metrics   *metrics
+}
+
+// metrics is a collection of idAllocator's metrics
+type metrics struct {
```
sunxiaoguang, line 58: Why we need to have this type instead of using prometheus.Guage directly?

AmoebaProtozoa: It was to improve the readability and make it easier to add more metrics, as suggested https://github.com/tikv/pd/pull/5284#discussion_r919728599

server/id/id.go
AmoebaProtozoa: > Curiously, is there anything broken so we have to update PD recovery process?

It's this block related to finding the maximum allocated ID
https://github.com/pingcap/docs/blob/master/pd-recover.md#get-allocated-id-from-pd-log

server/id/id.go
```diff
@@ -45,14 +46,29 @@ type allocatorImpl struct {
...
+	idGauge prometheus.Gauge
 }
 
 // NewAllocator creates a new ID Allocator.
-func NewAllocator(client *clientv3.Client, rootPath string, member string) Allocator {
-	return &allocatorImpl{client: client, rootPath: rootPath, member: member}
+func NewAllocator(client *clientv3.Client, rootPath, allocPath, label, member string) Allocator {
```
sunxiaoguang, line 63: We need to make allocation step configurable as well. Even though we have 3 bytes reserved for Keyspace ID, 1000 is still too much.

nolouch (approved):
lgtm

nolouch:
ptal @sunxiaoguang

server/id/id.go
rleungx: Once we recover PD, do we also need to recover all id allocators?

server/id/id.go
rleungx: Once we recover PD, do we also need to recover all id allocators?

PR #5291, opened 2022-07-08
server: keyspace management

AmoebaProtozoa:
Added keyspace manager
Issue Number: Ref #5293 

```commit-message
storage: keyspace management
```

### Check List

Tests

- Unit test

Code changes

### Release note

```release-note
None
```


server/keyspace/keyspace.go
```diff
@@ -0,0 +1,240 @@
...
+// UpdateKeyspaceRequest represents necessary arguments to update keyspace.
+type UpdateKeyspaceRequest struct {
+	Name string
+	// Whether to update the state of keyspace,
+	// if set to false, NewState and Now will be ignored.
+	UpdateState bool
+	NewState    keyspacepb.KeyspaceState
+	Now         time.Time
+	// New KVs to put in keyspace config.
+	ToPut map[string]string
```
sunxiaoguang, line 69: Why there is extra 'To' prefix?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,240 @@
...
+
+// CreateKeyspace create a keyspace meta with initial config and save it to storage.
+func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspacepb.KeyspaceMeta, error) {
+	// allocate new id
+	newID, err := manager.allocID()
+	if err != nil {
+		return nil, err
+	}
+	// bind name to that id
+	if err = manager.createNameToID(newID, request.Name); err != nil {
```
sunxiaoguang, line 125: Name to id and meta should be updated in a transaction. Other wise user could see name to id mapping without metadata associated. 

For the case if we really can't use transaction, we need to reorder updates and make sure user can't see it before it's fully initialized and collect garbages.

server/keyspace/keyspace.go
AmoebaProtozoa: Mostly to disambiguate with go's `delete` built-in, which was used in this package.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+	manager.metaLock.Lock()
+	defer manager.metaLock.Unlock()
+	// load keyspace by id
+	keyspace, err := manager.LoadKeyspace(request.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	// disallow modifying archived keyspace
+	if keyspace.State == keyspacepb.KeyspaceState_ARCHIVED {
```
ystaticy, line 197: The variable name should indicate that this is the old keyspace information

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, errKeyspaceArchived
+	}
+
+	// update keyspace state and config
+	if request.UpdateState {
+		if err = changeKeyspaceState(keyspace, request.NewState, request.Now); err != nil {
+			return nil, err
+		}
+	}
+	changeKeyspaceConfig(keyspace, request.ToPut, request.ToDelete)
```
ystaticy, line 207: Can we change the config when keyspace.state is DISABLE?

server/storage/keyspace_test.go
```diff
@@ -0,0 +1,137 @@
...
+		re.Equal(ids[i], id)
+	}
+	// loading non-existing id should return false, 0, nil
+	success, id, err := storage.LoadKeyspaceID("non-existing")
+	re.NoError(err)
+	re.False(success)
+	re.Equal(uint32(0), id)
+}
+
+func testKeyspaces() []*keyspacepb.KeyspaceMeta {
```
ystaticy, line 100: The function name needs to be able to simply describe the function logic.

server/storage/keyspace_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/pingcap/kvproto/pkg/keyspacepb"
+	"github.com/stretchr/testify/require"
+)
+
+func TestKeyspace(t *testing.T) {
```
ystaticy, line 25: The function name needs to be able to simply describe the function logic.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// changeKeyspaceState changes the state of given keyspace meta.
+// It also records the time state change happened.
+func changeKeyspaceState(keyspace *keyspacepb.KeyspaceMeta, newState keyspacepb.KeyspaceState, now time.Time) error {
+	if keyspace.State == newState {
+		return nil
+	}
+	// ARCHIVED is the terminal state that cannot be changed from.
+	if keyspace.State == keyspacepb.KeyspaceState_ARCHIVED {
```
ystaticy, line 227: Suggest add test case for this situation

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+		return errKeyspaceArchived
+	}
+
+	keyspace.StateChangedAt = now.Unix()
+	keyspace.State = newState
+	return nil
+}
+
+// changeKeyspaceConfig update a keyspace's config according to toPut and toDelete.
+func changeKeyspaceConfig(keyspace *keyspacepb.KeyspaceMeta, toPut map[string]string, toDelete []string) {
```
ystaticy, line 237: If we implement delete here, it is possible that the same key is updated first and then deleted during the update process.So it will increase the complexity of API.

Divide update and delete into different interfaces is better?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// update keyspace state and config
+	if request.UpdateState {
+		if err = changeKeyspaceState(keyspace, request.NewState, request.Now); err != nil {
+			return nil, err
+		}
+	}
+	changeKeyspaceConfig(keyspace, request.ToPut, request.ToDelete)
+	// save keyspace
+	if err = manager.store.SaveKeyspace(keyspace); err != nil {
```
ystaticy, line 209: If the change of state and config is already saved in front. 
Is there any change for KeySpace ?

ystaticy: Is it need complete keyspace information in request?

server/keyspace/keyspace_test.go
```diff
@@ -0,0 +1,229 @@
...
+	// check that eventually all test keyspaces' test config reaches end
+	for _, request := range requests {
+		keyspace, err := manager.LoadKeyspace(request.Name)
+		re.NoError(err)
+		re.Equal(keyspace.Config[testConfig], strconv.Itoa(end))
+	}
+}
+
+// matchCreateRequest verifies a keyspace meta matches a create request.
+func matchCreateRequest(re *require.Assertions, request *CreateKeyspaceRequest, meta *keyspacepb.KeyspaceMeta) {
```
ystaticy, line 183: It can be named ```checkKeyspace```. Like https://github.com/tikv/pd/blob/34a4cce2cb07f8aa4283e28ba40a667e1321a1fb/server/cluster/cluster_test.go#L1763

server/keyspace/keyspace_test.go
```diff
@@ -0,0 +1,229 @@
...
+	re.Equal(request.Name, meta.Name)
+	re.Equal(request.Now.Unix(), meta.CreatedAt)
+	re.Equal(request.Now.Unix(), meta.StateChangedAt)
+	re.Equal(keyspacepb.KeyspaceState_ENABLED, meta.State)
+	re.Equal(request.InitialConfig, meta.Config)
+}
+
+// matchUpdateRequest verifies a keyspace meta could be the immediate result of an update request.
+func matchUpdateRequest(re *require.Assertions, request *UpdateKeyspaceRequest, meta *keyspacepb.KeyspaceMeta) {
```
ystaticy, line 192: Ditto.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+	// Create and save keyspace metadata.
+	keyspace := &keyspacepb.KeyspaceMeta{
+		Id:             newID,
+		Name:           request.Name,
+		State:          keyspacepb.KeyspaceState_ENABLED,
+		CreatedAt:      request.Now.Unix(),
+		StateChangedAt: request.Now.Unix(),
+		Config:         request.InitialConfig,
```
ystaticy, line 135: Should ```gc_life_time```, ```gc_run_interval``` need to be initialized here?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,244 @@
...
+	return keyspace, nil
+}
+
+// UpdateKeyspace update keyspace's state and config according to request.
+// It returns error if saving failed, operation not allowed, or if keyspace not exists.
+func (manager *Manager) UpdateKeyspace(request *UpdateKeyspaceRequest) (*keyspacepb.KeyspaceMeta, error) {
+	manager.metaLock.Lock()
+	defer manager.metaLock.Unlock()
+	// load keyspace by id
+	keyspace, err := manager.LoadKeyspace(request.Name)
```
ystaticy, line 191: The variable name should indicate that here is the old keyspace meta.

server/keyspace/keyspace_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestLoadRangeKeyspace(t *testing.T) {
+	re := require.New(t)
+	manager := newKeyspaceManager()
+	requests := createKeyspaceRequests(10)
+
+	startID := 100
+	// force keyspace id start with startID
+	for i := 0; i < startID-1; i++ {
+		_, _ = manager.idAllocator.Alloc()
```
ystaticy, line 115: How about add a function to set startid when init a new allocator? Or it can be add as a TODO task.

server/storage/keyspace_test.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, keyspace := range keyspaces {
+		re.NoError(storage.SaveKeyspace(keyspace))
+	}
+
+	// load all keyspaces.
+	loadedKeyspaces, err := storage.LoadRangeKeyspace(keyspaces[0].GetId(), 0)
+	re.NoError(err)
+	re.ElementsMatch(keyspaces, loadedKeyspaces)
+
+	// load keyspaces that with id no less than second test keyspace.
```
ystaticy, line 66: // Load keyspaces that with id >= second test keyspace id.

server/storage/keyspace_test.go
ystaticy: For example: https://github.com/tikv/pd/blob/34a4cce2cb07f8aa4283e28ba40a667e1321a1fb/server/api/region_label_test.go#L109

server/keyspace/keyspace.go
AmoebaProtozoa: Yes, it should be allowed.
Consider the scenario where we are re-enabling a disable keyspace with updated config.
If we only permits changing config after it's ENABLED, then keyspace watchers may see outdated config.

server/keyspace/keyspace_test.go
```diff
@@ -0,0 +1,275 @@
...
+		re.NoError(err)
+		re.Equal(updated.State, keyspacepb.KeyspaceState_ARCHIVED)
+		re.Equal(updated.StateChangedAt, newTime.Unix())
+		// Changing state of an ARCHIVED keyspace is not allowed.
+		_, err = manager.UpdateKeyspaceState(createRequest.Name, keyspacepb.KeyspaceState_ENABLED, newTime)
+		re.Error(err)
+	}
+}
+
+func TestLoadRangeKeyspace(t *testing.T) {
```
ystaticy, line 159: Suggest check that if the parameter range out of keyspaces actual range in storage, it will return keyspaces in actual range. And need consider the case that startID is greater than max(keyspaceID) in storage

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,299 @@
...
+	if keyspaceExists {
+		return nil, ErrKeyspaceExists
+	}
+	// Save keyspace meta before saving id.
+	if err = manager.store.SaveKeyspace(keyspace); err != nil {
+		return nil, err
+	}
+	// Create name to ID entry,
+	// if this failed, previously stored keyspace meta should be removed.
+	if err = manager.createNameToID(newID, request.Name); err != nil {
```
AndreMouche, line 134: I think we could check whether the name is duplicated before saving keyspace.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,299 @@
...
+	// Note: OpDel is idempotent, deleting a non-existing key
+	// will not result in error.
+	OpDel
+)
+
+// UpdateKeyspaceConfig changes target keyspace's config in the order specified in mutations.
+// It returns error if saving failed, operation not allowed, or if keyspace not exists.
+func (manager *Manager) UpdateKeyspaceConfig(name string, mutations []*Mutation) (*keyspacepb.KeyspaceMeta, error) {
+	manager.metaLock.Lock()
+	defer manager.metaLock.Unlock()
```
AndreMouche, line 192: I think it is not a good idea to share a lock for all key-spaces.

AmoebaProtozoa: I changed the global keyspace lock to a map of mutexes (`lockGroup`) that uses the lower byte of spaceID as key.

This should levitate the lock contention, especially during consecutive keyspace creation (since spaceIDs are allocated sequentially), and avoid using too much memory to create a mutex for every keyspace. 

It's implementation and corresponding tests are located in `keyspace/util.go` & `keyspace/util_test.go`, PTAL and let me know what you think.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,318 @@
...
+)
+
+const (
+	// AllocStep set idAllocator's step when write persistent window boundary.
+	// Use a lower value for denser idAllocation in the event of frequent pd leader change.
+	AllocStep = uint64(100)
+	// AllocLabel is used to label keyspace idAllocator's metrics.
+	AllocLabel = "keyspace-idAlloc"
+	// DefaultKeyspaceName is the name reserved for default keyspace.
+	DefaultKeyspaceName = "DEFAULT"
```
rleungx, line 34: Is there any reason to use big letters?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,318 @@
...
+	// Check if default keyspace already exists.
+	defaultExists, err := manager.store.LoadKeyspace(DefaultKeyspaceID, &keyspacepb.KeyspaceMeta{})
+	if err != nil {
+		return nil, err
+	}
+	if defaultExists {
+		return manager, nil
+	}
+	// Initialize default keyspace.
+	defaultKeyspace := &keyspacepb.KeyspaceMeta{
```
rleungx, line 78: Why not use CreateKeyspace to init the default keyspace? Seems it has more information like creating time and so on.

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,318 @@
...
+		return nil, err
+	}
+	return keyspace, nil
+}
+
+// LoadRangeKeyspace load up to limit keyspaces starting from keyspace with startID.
+func (manager *Manager) LoadRangeKeyspace(startID uint32, limit int) ([]*keyspacepb.KeyspaceMeta, error) {
+	// Load Start should fall within acceptable ID range.
+	if startID > spaceIDMax {
+		return nil, errIllegalID
```
rleungx, line 292: How about adding the startID into the error?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,318 @@
...
+
+// allocID allocate a new keyspace id.
+func (manager *Manager) allocID() (uint32, error) {
+	id64, err := manager.idAllocator.Alloc()
+	if err != nil {
+		return 0, err
+	}
+	id32 := uint32(id64)
+	// Skip reserved space ID.
+	if id32 == DefaultKeyspaceID {
```
rleungx, line 305: When will it happen? Alloc will return 0 only when there is an error.

server/keyspace/util.go
```diff
@@ -0,0 +1,124 @@
...
+	if !isValid {
+		return errIllegalName
+	}
+	if name == DefaultKeyspaceName {
+		return errIllegalName
+	}
+	return nil
+}
+
+// lockGroup is a map of mutex that locks each keyspace separately.
```
rleungx, line 76: How about moving the below section to pkg/syncutil?

server/keyspace/keyspace.go
AmoebaProtozoa: nothing in particular, it's just feels kinda natural to use for name with special meaning.
Is there any style reference for this?

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,306 @@
...
+	// Save the updated keyspace.
+	if err = manager.store.SaveKeyspace(keyspace); err != nil {
+		return nil, err
+	}
+	return keyspace, nil
+}
+
+// UpdateKeyspaceState updates target keyspace to the given state if it's not already in that state.
+// It returns error if saving failed, operation not allowed, or if keyspace not exists.
+func (manager *Manager) UpdateKeyspaceState(name string, newState keyspacepb.KeyspaceState, now time.Time) (*keyspacepb.KeyspaceMeta, error) {
```
rleungx, line 238: Why do we need to pass now as a parameter?

server/keyspace/keyspace.go
rleungx: No, AFAICT.

server/keyspace/keyspace.go
AmoebaProtozoa: Mostly for two reasons:
1. It makes it easier to generate test data for future time related functionalities (For example, archiving keyspace that has been disabled for more than x days)
2. It's possible that now is parsed from tso rather than from `time.Now` directly. (See [`UpdateServiceGCSafePoint`](https://github.com/tikv/pd/blob/e089172a12c221655044b92210acc0477cf9ccb2/server/grpc_service.go#L1434))

But I'm not sure how necessary it is to used TSO to record keyspace info. So I'm open to remove it from parameter and generate it at the manager.

pkg/syncutil/lock_group.go
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package syncutil
+
+import "fmt"
+
+// LockGroup is a map of mutex that locks each keyspace separately.
```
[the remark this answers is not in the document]
AmoebaProtozoa: I have made lockGroup more general and takes custom hash function.

server/keyspace_service.go
```diff
@@ -0,0 +1,141 @@
...
+	case keyspace.ErrKeyspaceNotFound:
+		return s.wrapErrorToHeader(pdpb.ErrorType_ENTRY_NOT_FOUND, err.Error())
+	default:
+		return s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN, err.Error())
+	}
+}
+
+// validateRequest checks that server is serving and cluster id matches the requested.
+func (s *KeyspaceServer) validateRequest(header *pdpb.RequestHeader) error {
+	if s.IsClosed() {
```
rleungx, line 50: Do we need to check the leader?

server/keyspace/util.go
```diff
@@ -0,0 +1,79 @@
...
+	// Name must be non-empty and contains only alphanumerical, `_` and `-`.
+	namePattern = "^[-A-Za-z0-9_]+$"
+)
+
+var (
+	// ErrKeyspaceNotFound is used to indicate target keyspace does not exist.
+	ErrKeyspaceNotFound = errors.New("keyspace does not exist")
+	// ErrKeyspaceExists indicates target keyspace already exists.
+	// Used when creating a new keyspace.
+	ErrKeyspaceExists   = errors.New("keyspace already exists")
```
rleungx, line 35: Why are these two errors exposed?

server/keyspace_service.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	rc := s.GetRaftCluster()
+	if rc == nil {
+		return &keyspacepb.LoadKeyspaceResponse{Header: s.notBootstrappedHeader()}, nil
+	}
+
+	manager := s.GetKeyspaceManager()
+	meta, err := manager.LoadKeyspace(request.Name)
+	if err != nil {
+		return &keyspacepb.LoadKeyspaceResponse{Header: s.getErrorHeader(err)}, err
```
rleungx, line 75: Should it be
```suggestion
		return &keyspacepb.LoadKeyspaceResponse{Header: s.getErrorHeader(err)}, nil
```

server/keyspace_service.go
```diff
@@ -0,0 +1,141 @@
...
+
+// WatchKeyspaces captures and sends keyspace metadata changes to the client via gRPC stream.
+// Note: It sends all existing keyspaces as it's first package to the client.
+func (s *KeyspaceServer) WatchKeyspaces(request *keyspacepb.WatchKeyspacesRequest, stream keyspacepb.Keyspace_WatchKeyspacesServer) error {
+	if err := s.validateRequest(request.GetHeader()); err != nil {
+		return err
+	}
+	rc := s.GetRaftCluster()
+	if rc == nil {
+		return stream.Send(&keyspacepb.WatchKeyspacesResponse{Header: s.notBootstrappedHeader()})
```
rleungx, line 91: When will it return a response or just an error?

server/keyspace_service.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/tikv/pd/server/keyspace"
+	"github.com/tikv/pd/server/storage/endpoint"
+	"go.etcd.io/etcd/clientv3"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// KeyspaceServer wraps GrpcServer to provide keyspace service.
+type KeyspaceServer struct {
+	*GrpcServer
```
rleungx, line 33: Why not
```suggestion
	*Server
```

AmoebaProtozoa: Wraping the `GrpcServer` instead of `Server` enables `KeyspaceServer` to reuse many `GrpcServer`'s utility functions, such as `wrapErrorToHeader`, otherwise the same logic will have to be repeated.
This also make it possible to use `unaryMiddleware` to redirect request it's deemed necessary.

Also, logically speaking, `KeyspaceServer` is a extension of `GrpcServer`, so I think it's quite natural to embed it.

server/keyspace/util.go
AmoebaProtozoa: It's exported to set the error code in response header.
Server can check if the manager returned these two error to set correct ErrorType.

server/keyspace_service.go
AmoebaProtozoa: I don't think it's necessary, since both requests are read-only

rleungx: But the raftcluster will be nil?

AmoebaProtozoa: Ok, I think you are right.

server/keyspace_service.go
AmoebaProtozoa: I think a rule of thumb is that if the error has a pdpb.ErrorType (e.g., cluster not bootstrapped), it should be send back in the stream.
Otherwise it should return error.
A example of this can be seen in [`ReportBuckets`](https://github.com/tikv/pd/blob/0791fe818b13398c723b7d6214afe230ed429f6a/server/grpc_service.go#L791)

nolouch:
ptal @rleungx @CabinfeverB

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,314 @@
...
+	// If default keyspace already exists, skip initialization.
+	defaultExist, _, err := manager.store.LoadKeyspaceIDByName(DefaultKeyspaceName)
+	if err != nil {
+		return nil, err
+	}
+	if defaultExist {
+		return manager, nil
+	}
+	// Initialize default keyspace.
+	now := time.Now()
```
rleungx, line 77:
```suggestion
	now := time.Now().Unix()
```

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil && err != ErrKeyspaceExists {
+		return nil, err
+	}
+	return manager, nil
+}
+
+// CreateKeyspace create a keyspace meta with given config and save it to storage.
+func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspacepb.KeyspaceMeta, error) {
+	// Validate purposed name's legality.
+	if err := validateName(request.Name); err != nil {
```
rleungx, line 95:
```suggestion
	if err := validateName(request.GetName()); err != nil {
```

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, err
+	}
+	// Create and save keyspace metadata.
+	keyspace := &keyspacepb.KeyspaceMeta{
+		Id:             newID,
+		Name:           request.Name,
+		State:          keyspacepb.KeyspaceState_ENABLED,
+		CreatedAt:      request.Now.Unix(),
+		StateChangedAt: request.Now.Unix(),
+		Config:         request.Config,
```
rleungx, line 110: How about using request.GetXXX

server/keyspace/keyspace.go
```diff
@@ -0,0 +1,314 @@
...
+	defer manager.idLock.Unlock()
+	// Check if keyspace id with that name already exists.
+	nameExists, _, err := manager.store.LoadKeyspaceIDByName(keyspace.Name)
+	if err != nil {
+		return nil, err
+	}
+	if nameExists {
+		return nil, ErrKeyspaceExists
+	}
+	manager.metaLock.Lock(keyspace.Id)
```
rleungx, line 126: ditto

rleungx (commented):
The rest LGTM

server/keyspace/keyspace.go
```diff
@@ -55,7 +55,29 @@ type CreateKeyspaceRequest struct {
...
+	Now int64
+}
+
+// NewCreateKeyspaceRequest returns a create keyspace request with necessary arguments.
+// name is the name of the keyspace to be created; config is its initial config; now is the current timestamp.
+func NewCreateKeyspaceRequest(name string, config map[string]string, now int64) *CreateKeyspaceRequest {
+	return &CreateKeyspaceRequest{Name: name, Config: config, Now: now}
+}
+
+// GetName returns the name of the create keyspace request.
```
rleungx, line 68: I think the function has already been generated by protobuf

AmoebaProtozoa: No, the createKeyspace is only used by HTTP API, thus is not generated by ProtoBuf

rleungx: My bad, then it doesn't matter if wrap it with a function.

rleungx (approved):

nolouch (approved):

PR #5300, opened 2022-07-13
*: improve the topology weight calculation

### LINKED ISSUE #5234: Online progress hung in 83% when scale out with imbalance labels
### What did you do?
Cluster Config:
ZO745v59e0

usl19PhI0e

Scale out topo:
U8RSQz6Fkq

When balance stop, preparing should stop also. 

### What did you see instead?
No balance operator created more than 1 hours, but online store progress still stay in 83%.

image

master

rleungx:

Issue Number: Close #5234.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue that the online process is not accurate when having invalid label settings.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -36,6 +36,7 @@ import (
 	"github.com/tikv/pd/pkg/logutil"
 	"github.com/tikv/pd/pkg/netutil"
 	"github.com/tikv/pd/pkg/progress"
+	"github.com/tikv/pd/pkg/slice"
 	"github.com/tikv/pd/pkg/syncutil"
 	"github.com/tikv/pd/pkg/typeutil"
 	"github.com/tikv/pd/server/config"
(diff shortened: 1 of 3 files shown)


server/cluster/cluster.go
```diff
@@ -1594,24 +1601,42 @@ func getStoreTopoWeight(store *core.StoreInfo, stores []*core.StoreInfo, locatio
...
 
+	validLabels := locationLabels
+	var isMatch bool
+	for i, c := range totalLabelCount {
+		if count/c == 0 {
+			validLabels = validLabels[:i]
+			break
+		}
+		if count/c == 1 {
+			validLabels = []string{validLabels[i]}
```
CabinfeverB, line 1625: a store with label(dc1,zone1) and a store with label(dc2 zone1) are different. So I think it should be 
```
validLabels = validLabels[:i+1]
```
Otherwise, it will affect the results of CompareLocation.

AndreMouche (approved):
LGTM

CabinfeverB (approved):

nolouch (approved):

Issue #5257, opened 2022-07-04
Tracking issue for the scheduling diagnosis


rleungx:
## Background
Previously, when we encountered scheduling problems, it was hard to know the root cause for people unfamiliar with PD according to the existing log and monitoring. Sometimes, it's even hard for us because the scheduling has many random factors which cause the undetermined result.
Under this circumstance, we want to have an easier way to figure out why there is no expected schedule happening and why the unexpected schedule happening.

### Stage 1: support triggering diagnosis through API
- [x] https://github.com/tikv/pd/pull/5228
- [x] https://github.com/tikv/pd/pull/5266
- [x] https://github.com/tikv/pd/pull/5283
- [x] https://github.com/tikv/pd/pull/5135
- [ ] https://github.com/tikv/pd/pull/5125

### Stage 2: support continuously diagnosis
TBD

PR #5303, opened 2022-07-13
*: collect the checker dry run information

rleungx:

Issue Number: Ref #5257.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


server/cluster/diagnosis.go
```diff
@@ -0,0 +1,85 @@
...
+	var result *diagnosisResult
+	// if it's a scheduler.
+	_, ok := d.schedulers[name]
+	if ok {
+		result = newDiagnosisResult(d.schedulers[name].DiagnoseDryRun())
+	}
+	// if it's a checker.
+	switch name {
+	case "merge-checker":
+		regions := d.cluster.GetRegionStatsByType(statistics.UndersizedRegion, 100)
```
AndreMouche, line 43: If it do not have a `region` info, check all regions with `UndersizedRegion` should be ok, otherwise we should consider how to show the rest regions result==

AndreMouche: Or, maybe we could provide an API to get `GetRegionStatsByType`, and the region id was needed for dry-run for checker.

server/cluster/diagnosis.go
```diff
@@ -0,0 +1,81 @@
+package cluster
```
disksing, line 1: no license

rleungx:
Close it temporarily, will support it in the future.

PR #5315, opened 2022-07-16
server/schedule: fix invalid store id

### LINKED ISSUE #5279: Cannot failover well after recover pd from `pd-recover` and one tikv has gone
### What did you do?

one TiKV and all PDs has gone, I do recovery. 

I use `pd-recovery` to recreate the PD cluster, and all existing TiKVs reconnect to PD.  and then I scale a new TiKV. 

I want to see PD can fix all unhealthy peers, which means scheduling to move peers from missing TiKV to the new TiKV.

### What did you see instead?
**Note**: Due to PD is recreated, so PD no any information about the missing  TiKV. this is the main reason about the follow behavior.

I can see there actually do some scheduling operator:
image

Here exist main two kinds of operators:
- `add-rule-peers`:  caused by the rule fit, the peer in the missing store is not fit. so PD think the region's peer count not meet the requirement.
- `remove-orphan-peer`: caused by the rule fit. because the peer in the missing store is not fit, so the peer is an `orphan peer`. PD will remove after the peer count of the region meets the requirement. 

If the operator succeeds, it will satisfy our expectations. but all the `remove-orphan-peer` operator is timeout:
image

 After I check the log. I can see the operator command cannot send to TiKV:

image

master

LLThomas:

Issue Number: Close #5279

Fix the problem that "remove-orphan-operator" cannot be sended to tikv due to invalid store id.
```commit-message
 fix invalid store id
```

### Check List

Tests

Steps:
1. deploy a cluster with 1 pd, 3 tikv
2. delete pd data directory to simulate downtime operation and kill one tikv
3. use pd-recover to recover pd
4. scale out a new tikv

Result: 
There is no "add-orphan-peer" operator with timeout status any more.

image

image

### Release note

```release-note
None.
```


nolouch:
cc @HunDunDM

nolouch (commented):
lgtm

nolouch (approved):

nolouch:
PTAL @HunDunDM

HunDunDM (approved):

PR #5326, opened 2022-07-18
dashboard: fix a bug that could not handle proxy requests correctly

### LINKED ISSUE #5321: Internal proxy cannot work as expected
### What did you do?

Add an [external load balancer](https://cloud.google.com/load-balancing/docs/https#cloud-console) above PD and enable the internal proxy.

The internal proxy can work.

### What did you see instead?

Return 508 error.

Because we check if `Via` header is existed to prevent the loop, but the external proxy maybe already have this header. Once the request is sent to the PD follower, it will be rejected due to the check.

master

HunDunDM:

Issue Number: Close #5321 

Only process headers added by PD.

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
Fix a bug that could not handle dashboard proxy requests correctly.
```


lhy1024 (approved):

pkg/dashboard/adapter/redirector_test.go
```diff
@@ -0,0 +1,112 @@
...
+	suite.noRedirectHTTPClient.CloseIdleConnections()
+}
+
+func (suite *redirectorTestSuite) TestReverseProxy() {
+	redirectorServer := httptest.NewServer(http.HandlerFunc(suite.redirector.ReverseProxy))
+	defer redirectorServer.Close()
+
+	suite.redirector.SetAddress(suite.tempServer.URL)
+	// Test normal forwarding
+	req, err := http.NewRequest("GET", redirectorServer.URL, nil)
```
rleungx, line 73:
```suggestion
	req, err := http.NewRequest(http.MethodGet, redirectorServer.URL, nil)
```
Please replace all with `http.MethodGet`.

rleungx (approved):

PR #5341, opened 2022-07-21
scheduler: speed up exit

### LINKED ISSUE #5274: Coordinator can't stop in time, because of background jobs is still running
## Bug Report
I have posted a post on the forum,   see  https://asktug.com/t/topic/694191/3 .
In short, the problem is that when coordinator is stopping the scheduler will still keep running until  all jobs are finished.

The following figure shows that it takes a long time for the coordinator to close from the beginning to the end,  in this case more than 10 minutes.
image

### What did you do?
pd leader lease timeout.
image

Follower campaign for leader in time.

### What did you see instead?
Follower can't campaign for leader in 10 minutes, and keep print the following log.
image

5.3.0

in-han:

Issue Number: Close #5274

```in scheduler loop,  try to read exit signal.
     if coordinator's ctx has been closed, don't do more schedule. 
```

### Check List

Tests

For our cluster with more than 2000 tikv-servers,  pd transfer leader can be done in 1 second.  

Code changes

Side effects

Related changes

### Release note

```release-note
speed up scheduler exit  
```


server/schedulers/balance_leader.go
```diff
@@ -364,6 +364,11 @@ func (l *balanceLeaderScheduler) Schedule(cluster schedule.Cluster, dryRun bool)
 
 	result := make([]*operator.Operator, 0, batch)
 	for sourceCandidate.hasStore() || targetCandidate.hasStore() {
+		// if coordinator is stopping, speed up exit.
+		if l.BaseScheduler.OpController.Ctx().Err() != nil {
```
nolouch, line 368: Maybe we can consider adding the `context` field for all schedulers?

nolouch (commented):
Thanks for you contribution.

server/schedulers/balance_leader.go
in-han: Every scheduler have a BaseScheduler, and BaseScheduler have opController with a context filed, I think it's no difference.

PR #5363, opened 2022-07-27
dashboard: fix a bug that could not handle proxy requests correctly (#5326)

### LINKED ISSUE #5321: Internal proxy cannot work as expected

HunDunDM:
This is a cherry-pick of #5326

Issue Number: Close #5321 

Only process headers added by PD.

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
Fix a bug that could not handle dashboard proxy requests correctly.
```

### DIFF
--- a/pkg/dashboard/adapter/redirector.go
+++ b/pkg/dashboard/adapter/redirector.go
@@ -76,7 +76,7 @@ func (h *Redirector) SetAddress(addr string) {
 	defaultDirector := h.proxy.Director
 	h.proxy.Director = func(r *http.Request) {
 		defaultDirector(r)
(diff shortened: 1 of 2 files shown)


HunDunDM:
What is needed is version v5.4.2, this PR will be closed.

PR #5375, opened 2022-07-28
statistics: add region score 

### LINKED ISSUE #5372: provide replica read traffic schedule
## Enhancement Task

In https://github.com/pingcap/tidb/pull/28033 and https://github.com/pingcap/tidb/pull/35927 we have introduced 2 type of replica read the prefers read from replica within the same AZ.

As our current balance strategy(leader/region) aims at the balance between different stores. If the traffic across different AZ are not even(in extrame case, all read triffic are from one AZ), current schedule can't balance the traffic because the schedule do not change region's AZ.

In order to solve this problem, I'd like to add a new traffic scheduler. When closest follower read is enable, we maintain a score for each region peer that based on the resource usage(e.g. CPU usage). The score represent the percent that client should dispatch a read traffic to region leader instead of the closest replica. If one store's resource usage exceeds the config threshold, we increase the score; otherwise, we decrease the score. This adjust is triggered at each store heartbeat.

In the client size, the client should maintain the score for active region, it should periodically refresh the score of each region in a short interval(e.g. 5s).

glorv:

Issue Number: Close #5372

```

```
This PR add a new statistics the maintains a score(represent the leader traffic percent) for each region replica. When closest replica read is enabled, tikv client can use this score to dispatch read request to region leader instead of the closest replica. By dynamically adjust the score at StoreHearbeat, we can ensure that all store resource can be fully used even when the read traffic is not even.

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has configuration change
- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Please add a release note.

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -85,6 +85,8 @@ type Client interface {
 	GetPrevRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
 	// GetRegionByID gets a region and its leader Peer from PD by id.
 	GetRegionByID(ctx context.Context, regionID uint64, opts ...GetRegionOption) (*Region, error)
+	//
+	BatchGetRegionScore(ctx context.Context, regionIDs uint64) ([]*pdpb.RegionScore, error)
 	// ScanRegion gets a list of regions, starts from the region that contains key.
 	// Limit limits the maximum number of regions returned.
 	// If a region has no leader, corresponding leader will be placed by a peer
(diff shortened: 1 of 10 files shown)


glorv:
@JmPotato @rleungx  PTAL

nolouch:
Any design related materials?

glorv:
> Any design related materials?

Here is the original desgin [doc](https://pingcap.feishu.cn/docx/doxcnUkCUzPrvDCk9X8HdTQW1Og). The original algorithm is implemented at the client side. @BusyJay suggested that we should maintain this kind of schedule in pd because the traffic dispatch maybe affect region schedule. So I reimplemented the logic in pd and the algorithm is almost the same.

> and how to count the not evenly distributed problem with closest replica read.

Read traffic unevenly distribution can happen with closest replica read is enable in following scenarios:
- the number of tidb instances in each AZ is not same. If the load balancer balance traffic across all instances.
- The user traffic is not even. In case the application traffic distribution are different across AZs. This may exist in multi-region deployment as the latency across different region is high.

PR #5387, opened 2022-08-02
filter: don't record tombstone store in filter metrics

### LINKED ISSUE #5386: filter metrics should not record the tombstone store
## Enhancement Task
If we have many tombstone stores, there is a performance issue in both Grafana and PD itself. Since these tombstone stores have been removed completely, it doesn't matter if we don't record them. Furthermore, we can exclude the store with the tombstone state when getting them.

rleungx:

Issue Number: Close #5386.

```commit-message
```

### Check List

Tests

Before:
Screen Shot 2022-08-02 at 1 37 48 PM
After:
Screen Shot 2022-08-02 at 1 41 24 PM

### Release note

```release-note
None
```


JmPotato (commented):
Unnecessary change?

image

rleungx:
> Unnecessary change?
> 
> image

The previous commit modify the permission, I just change it back.

server/schedule/filter/filters.go
```diff
@@ -34,7 +34,8 @@ import (
 func SelectSourceStores(stores []*core.StoreInfo, filters []Filter, opt *config.PersistOptions) []*core.StoreInfo {
 	return filterStoresBy(stores, func(s *core.StoreInfo) bool {
 		return slice.AllOf(filters, func(i int) bool {
-			if !filters[i].Source(opt, s).IsOK() {
+			status := filters[i].Source(opt, s)
+			if !status.IsOK() && status != statusStoreTombstone {
```
lhy1024, line 38: Maybe we can add 'not tombstone' to `status.IsOK()`?

rleungx: I think it belongs to the different types. The previous change has a problem, I have updated it.

nolouch (approved):

nolouch:
PTAL @CabinfeverB

CabinfeverB (approved):

PR #5433, opened 2022-08-16
api: add region fit  placement rule API

### LINKED ISSUE #5432: Add an API to get rules and matched peers related to the given region
## Development Task
Sometimes, after configuring a placement rule, users want to make sure that it applies to the corresponding region or check for related configuration issues. So we need an API to get matched placement rule and peers.

CabinfeverB:

Issue Number: Close #5432 

```commit-message
add an API to return rules matched by a given region

```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
add an API to return rules matched by a given region.
```


bufferflies:
Can you offer the detail result of this api?

server/api/router.go
```diff
@@ -230,6 +230,7 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	regionHandler := newRegionHandler(svr, rd)
 	registerFunc(clusterRouter, "/region/id/{id}", regionHandler.GetRegionByID, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(clusterRouter.UseEncodedPath(), "/region/key/{key}", regionHandler.GetRegion, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(clusterRouter, "/region/rule/{id}", regionHandler.CheckRegionPlacementRule, setMethods(http.MethodGet), setAuditBackend(prometheus))
```
rleungx, line 233: How about
```suggestion
	/region/id/{id}/rules
```

CabinfeverB:
> Can you offer the detail result of this API?

```
{
  "rule-fits": [
    {
      "rule": {
        "group_id": "pd",
        "id": "default",
        "start_key": "",
        "end_key": "",
        "role": "voter",
        "count": 1
      },
      "peers": [
        {
          "id": 102,
          "store_id": 1
        }
      ],
      "peers-different-role": null,
      "isolation-score": 0
    }
  ],
  "orphan-peers": [
    {
      "id": 103,
      "store_id": 2
    }
  ]
}
```

server/api/router.go
CabinfeverB: I move it to ruleHandler

nolouch (approved):
lgtm

nolouch:
ptal @rleungx

server/api/rule.go
```diff
@@ -117,32 +119,58 @@ func (h *ruleHandler) GetRuleByGroup(w http.ResponseWriter, r *http.Request) {
...
+	regionFit := cluster.GetRuleManager().FitRegion(cluster, region)
+	h.rd.JSON(w, http.StatusOK, regionFit)
+}
+
+func (h *ruleHandler) preCheckForRegionAndRule(w http.ResponseWriter, r *http.Request) (*cluster.RaftCluster, *core.RegionInfo, bool) {
 	cluster := getCluster(r)
 	if !cluster.GetOpts().IsPlacementRulesEnabled() {
 		h.rd.JSON(w, http.StatusPreconditionFailed, errPlacementDisabled.Error())
-		return
+		return cluster, nil, false
```
rleungx, line 160: We don't need to return a bool?

tests/pdctl/config/config_test.go
```diff
@@ -346,6 +346,16 @@ func TestPlacementRules(t *testing.T) {
 	re.Equal([2]string{"pd", "default"}, rules2[0].Key())
 	re.Equal([2]string{"pd", "test1"}, rules2[1].Key())
 
+	// test rule region detail
+	pdctl.MustPutRegion(re, cluster, 1, 1, []byte("a"), []byte("b"))
+	fit := &placement.RegionFit{}
+	// need clear up args, so create new comman. Otherwise gourp still exists.
```
rleungx, line 352: comman?

tests/pdctl/helper.go
```diff
@@ -41,6 +42,16 @@ func ExecuteCommand(root *cobra.Command, args ...string) (output []byte, err err
 	return buf.Bytes(), err
 }
 
+// CreateCommandAndExecuteCommand is used for test purpose when should clean up args.
```
rleungx, line 45: Why do we need it?

server/api/rule.go
CabinfeverB: we should return if false

tests/pdctl/helper.go
CabinfeverB: Because the args of command won't initialize at every executing if we use only one `cobra.Command` in a test case.
For example, in `TestPlacementRules` we firstly test with "--group=pd". And if we test with  "--region=1" next, the arg `group` still exists.

rleungx (approved):

PR #5437, opened 2022-08-17
config: Relax limitation of tso-update-physical-interval

### LINKED ISSUE #5440: Relax minimal limitation of `tso-update-physical-interval` to generate more TSOs

Currently PD can generate at most 1<<18 (262144) TSOs in a period of `tso-update-physical-interval`. As the minimal of `tso-update-physical-interval` is 50ms, the maximum number of PD can provide is 5242880 (262144 x 1s / 50ms) per second.

The value is big enough for most scenario.

But when [RawKV API v2](https://docs.pingcap.com/tidb/stable/tikv-configuration-file#api-version-span-classversion-marknew-in-v610span) use TSO as timestamp of entries, 5242880 would be not big enough due to:

1. A cluster of RawKV can reach more than millions of QPS.

2. Some customers request fault tolerant of PD leader, so that system need to cache TSOs for about 8 seconds before TSO service recovery (see [tikv#12794](https://github.com/tikv/tikv/issues/12794)).

Relax the minimal limitation of `tso-update-physical-interval` to `1ms`. Users can choose the value according to their scenario.

Shorten PD cluster recovery duration after leader is killed (see [#5122](https://github.com/tikv/pd/issues/5122)). But it is not easy to do that, as leader election is the main part of the duration, which depends on ETCD and would be more difficult to speed up.

pingyu:
Issue Number: #5440

Signed-off-by: pingyu <yuping@pingcap.com>

Issue Number: Close #5440

```commit-message
Minimal valid value of config item `tso-update-physical-interval` is changed to 1ms.
```

### Check List

Tests

- Unit test
- Integration test
- Benchmark

| tso-update-physical-interval | CPU without workload | CPU with pd-tso-bench | pd-tso-bench QPS | pd-tso-bench P99 latency |
|------------------------------|----------------------|-----------------------|------------------|--------------------------|
| 1 ms                         | 15%                  | 28%                   | 458306           | 3.59 ms                  |
| 3 ms                         | 9.3%                 | 25%                   | 470455           | 3.57 ms                  |
| 5 ms                         | 7.6%                 | 25%                   | 354989           | 8.64 ms                  |
| 50 ms                        | 6%                   | 7%                    | 44466            | 54.1 ms                  |

Notes:
1. The benchmark is performed on a 40C 125GB NVMe machine, with 1 pd, 4 kv, 1 db, deployed by TiUP playground.
2. `count` in TSO request is set to 100x (`uint32(count) * 100` [here](https://github.com/tikv/pd/blob/6d883d0e8386c64ac12ade25d00893a608b33a14/client/client.go#L1072)) to make it overflow the logical interval.
3. There is no significant difference on memory usage.
4. The benchmark showed that `3ms` is good enough, and `1ms` has no improvement but just causes more CPU. But `1ms` can still provide 3 times TSOs than `3ms`, so `1ms` is still a valid value.

Code changes

- Has configuration change
`tso-update-physical-interval`

Side effects

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn): TBD
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup): TBD

### Release note

```release-note
Minimal valid value of config item `tso-update-physical-interval` is changed to 1ms.
```


haojinming (approved):
LGTM~

pingyu:
> /cc @nolouch

PTAL, thanks~

nolouch (approved):
lgtm

nolouch:
PTAL @bufferflies

server/tso/tso.go
```diff
@@ -44,6 +44,10 @@ const (
 	maxLogical = int64(1 << 18)
 	// MaxSuffixBits indicates the max number of suffix bits.
 	MaxSuffixBits = 4
+	// jetLagWarningThreshold is the warning threshold of jetLag in `timestampOracle.UpdateTimestamp`.
+	// In case of small `updatePhysicalInterval`, the `3 * updatePhysicalInterval` would also is small,
+	// and trigger many un-necessary warnings about clock offset.
+	jetLagWarningThreshold = 30 * time.Millisecond
```
bufferflies, line 50: Why the value is 30ms?

pingyu: It's not strictly calculated, but an empirical value.

On benchmark with `updatePhhsicalInterval = 1ms`, I observed a frequent log about jet lag ("clock offset ..."), saying a `jet-lag` varying between `5ms ~ 30ms`. I think it is not caused by a clock offset but other reasons (such as go runtime scheduling duration).

And I also think `30ms` would be a reasonable go routine wait duration on a busy system. So I choose such a value.

bufferflies: yes， I think the original value is 150ms (3 *minTSOUpdatePhysicalInterval(50ms) ). I think the 30ms maybe too small .

pingyu: Fixed, PTAL~

bufferflies (approved):

PR #5441, opened 2022-08-18
dr-auto-sync: enable min-resolved-ts persist by default

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/4399

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
dr-auto-sync: enable min-resolved-ts persist by default
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -62,7 +62,8 @@ import (
 
 var (
 	// DefaultMinResolvedTSPersistenceInterval is the default value of min resolved ts persistence interval.
-	DefaultMinResolvedTSPersistenceInterval = 10 * time.Second
+	// If interval in config is zero, it means not to persist resolved ts and check config with this DefaultMinResolvedTSPersistenceInterval
+	DefaultMinResolvedTSPersistenceInterval = config.DefaultMinResolvedTSPersistenceInterval
 )
 
 // regionLabelGCInterval is the interval to run region-label's GC work.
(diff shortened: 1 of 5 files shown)


server/cluster/cluster.go
```diff
@@ -62,7 +62,8 @@ import (
 
 var (
 	// DefaultMinResolvedTSPersistenceInterval is the default value of min resolved ts persistence interval.
-	DefaultMinResolvedTSPersistenceInterval = 10 * time.Second
+	// If interval in config is zero, it means not to persist resolved ts and check config with this DefaultMinResolvedTSPersistenceInterval
+	DefaultMinResolvedTSPersistenceInterval = time.Second
```
nolouch, line 66: Why need to reduce the interval?

rleungx, line 66: Why not keep it as before?

lhy1024: This default interval is also the interval for detecting whether the config has changed. During the manual test, it was found that the previous configuration was too long, and it took ten seconds for the configuration to take effect.

lhy1024: This default interval is also the interval for detecting whether the config has changed. During the manual test, it was found that the previous configuration was too long, and it took ten seconds for the configuration to take effect.

server/config/config.go
```diff
@@ -236,7 +236,7 @@ const (
 	defaultFlowRoundByDigit                 = 3 // KB
 	maxTraceFlowRoundByDigit                = 5 // 0.1 MB
 	defaultMaxResetTSGap                    = 24 * time.Hour
-	defaultMinResolvedTSPersistenceInterval = 0
+	defaultMinResolvedTSPersistenceInterval = time.Second
```
nolouch, line 239: Can it shared with `DefaultMinResolvedTSPersistenceInterval`?

rleungx:
CI failed

disksing (approved):

nolouch:
ci failed.

lhy1024:
> ci failed.

I fixing it.

nolouch (approved):
lgtm

Issue #5443, opened 2022-08-18
Support mode setting for PD


rleungx:
### Describe the feature you'd like

There is a common need to stop/pause PD to do the scheduling. For example, BR and Lightning will pause the PD scheduler and change the PD config temporarily, and unsafe recovery also needs to stop PD scheduling. And right now, we have a feature request to backup/restore on EBS. 
We can provide a unified way to control the scheduling. Instead of changing lots of configs, we can only set one config to do it. 
Moreover, we can add some scheduling configuration templates to let users choose them according to their different workloads and cluster performance. Also, they can easily change the config back to default if they want.

PR #5457, opened 2022-08-29
config: support mode setting

rleungx:

Issue Number: Ref #5443.

This PR support manually switching mode in different situations. Three modes are added:
- Normal: the default mode if we don't do anything.
- Suspend: in this mode, all scheduling will be stopped.
- Scaling: in this mode, the store limit of all stores will be changed to 200 to accelerate the scaling process.

And we use a new config item `schedule-mode` to manage the mode. It includes two configurations: `mode` and `auto-tune`. The latter will be used in the future to support auto switch modes.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -586,7 +586,7 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
 	if err := c.PDServerCfg.adjust(configMetaData.Child("pd-server")); err != nil {
 		return err
 	}
-	if err := c.ScheduleMode.adjust(configMetaData.Child("mode")); err != nil {
+	if err := c.ScheduleMode.adjust(configMetaData.Child("schedule-mode")); err != nil {
 		return err
 	}
 
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -756,6 +756,7 @@ type ScheduleConfig struct {
 	// overwrite the auto-tuned value by pd-ctl, when the value
 	// is overwritten, the value is fixed until it is deleted.
 	// Default: manual
+	// WARN: StoreLimitMode is deprecated.
 	StoreLimitMode string `toml:"store-limit-mode" json:"store-limit-mode"`
 
 	// Controls the time interval between write hot regions info into leveldb.
@@ -770,6 +771,9 @@ type ScheduleConfig struct {
 
 	// EnableDiagnostic is the the option to enable using diagnostic
 	EnableDiagnostic bool `toml:"enable-diagnostic" json:"enable-diagnostic,string"`
+
+	// Mode is a state of scheduling. Different mode have different strategies in configuration.
+	Mode string `toml:"mode" json:"mode"`
 }
 
 // Clone returns a cloned scheduling configuration.
@@ -820,6 +824,7 @@ const (
 	defaultHotRegionsReservedDays      = 7
 	// It means we skip the preparing stage after the 48 hours no matter if the store has finished preparing stage.
 	defaultMaxStorePreparingTime = 48 * time.Hour
+	defaultMode                  = Normal
 )
 
 func (c *ScheduleConfig) adjust(meta *configMetaData, reloading bool) error {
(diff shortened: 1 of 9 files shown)


server/config/config.go
```diff
@@ -751,6 +756,7 @@ type ScheduleConfig struct {
 	// overwrite the auto-tuned value by pd-ctl, when the value
 	// is overwritten, the value is fixed until it is deleted.
 	// Default: manual
+	// WARN: StoreLimitMode is deprecated.
 	StoreLimitMode string `toml:"store-limit-mode" json:"store-limit-mode"`
```
bufferflies, line 760: why not delete it?

server/config/mode.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server/storage/endpoint"
+	"go.uber.org/zap"
+)
+
+// Scheduling mode
+const (
+	Normal  = "normal"
+	Suspend = "suspend"
+	Scaling = "scaling"
```
bufferflies, line 15: Does it to distinguish the scale-in or scale-out?Can all the states switch to any state?

server/config/config.go
rleungx: IMO, we need to mark it as deprecated for about 2 major versions, then remove it totally.

server/config/mode.go
rleungx: > Does it to distinguish the scale-in or scale-out?

No need for now perhaps. 

> Can all the states switch to any state?

I think yes. WDYT?

server/server.go
```diff
@@ -896,25 +897,58 @@ func (s *Server) GetScheduleConfig() *config.ScheduleConfig {
...
+	s.persistOptions.SetScheduleModeConfig(&newCfg)
+	if err := s.persistOptions.Persist(s.storage); err != nil {
+		s.persistOptions.SetScheduleModeConfig(oldCfg)
+		log.Error("failed to update schedule mode config",
+			zap.Reflect("new-mode", newCfg),
+			zap.Reflect("old-mode", oldCfg),
 			errs.ZapError(err))
 		return err
 	}
-	log.Info("schedule config is updated", zap.Reflect("new", cfg), zap.Reflect("old", old))
```
CabinfeverB, line 917: we can add a log

server/config/mode.go
```diff
@@ -0,0 +1,89 @@
...
+	switch mode {
+	case Suspend:
+		newCfg.MergeScheduleLimit = 0
+		newCfg.LeaderScheduleLimit = 0
+		newCfg.RegionScheduleLimit = 0
+		newCfg.HotRegionScheduleLimit = 0
+		newCfg.ReplicaScheduleLimit = 0
+	case Scaling:
+		for storeID := range newCfg.StoreLimit {
+			newCfg.StoreLimit[storeID] = StoreLimitConfig{AddPeer: 200, RemovePeer: 200}
```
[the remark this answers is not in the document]
CabinfeverB: Maybe When we release, we can add a note for Scaling and Suspend. Changing some param in Suspend is useless and Changing StoreLimit in Scaling is useless.

CabinfeverB (commented):
rest LGTM

server/config/mode.go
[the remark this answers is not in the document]
rleungx: > Does it means that saving the config of StoreLimit in Scaling mode is useless

No, it will only be called when there is no scaling mode config.

server/config/config.go
rleungx: I'm going to delete it in the next PR.

server/config/mode.go
```diff
@@ -0,0 +1,89 @@
...
+	if !existed {
+		// We must have the Normal config because when we first switch the mode, the normal mode settings will be persisted.
+		_, err := storage.LoadScheduleMode(Normal, newCfg)
+		if err != nil {
+			return err
+		}
+		updateScheduleConfig(newCfg, newMode.Mode)
+	}
+	if newMode.Mode == Scaling { // reset the store limit for all store to avoid new added store using default limit in scaling mode.
+		for storeID := range oldCfg.StoreLimit {
```
CabinfeverB, line 53: we add a check whether store is in `newCfg.StoreLimit`

rleungx:
Close it since it's stale.

PR #5458, opened 2022-08-30
operator: allows to skip placement rules checks

### LINKED ISSUE #5401: A small number of TiFlash replica cannot be created
### What did you do?
Enable `shuffle-region-scheduler`
create some new tables and create TiFlash replica for these tables.

All table's TiFlash replicas will be AVAILABLE normally.

### What did you see instead?
A TiFlash replica is UNAVAILABLE.
image
And found region info
image

v6.2.0-alpha-146-g99e38eb7

HunDunDM:

Issue Number: Close #5401 

Allows skipping placement rules checks when creating operators.

### Check List

Tests

- Unit test

Code changes

Related changes

### Release note

```release-note
Fixed the bug where the Learner Peer of TiFlash Replica might not be created.
```

### DIFF
--- a/server/schedule/operator/builder.go
+++ b/server/schedule/operator/builder.go
@@ -67,7 +67,7 @@ type Builder struct {
 	targetLeaderStoreIDs []uint64 // This field is only used during multi-target evict leader, and will not be filtered during `Build`.
 	err                  error
 
-	// skip origin check flags
+	// skip check flags
 	skipOriginJointStateCheck bool
 	skipPlacementRulesCheck   bool
 
@@ -781,7 +781,7 @@ func (b *Builder) allowLeader(peer *metapb.Peer, ignoreClusterLimit bool) bool {
 	}
 
 	// placement rules
-	if len(b.rules) == 0 {
+	if b.skipPlacementRulesCheck || len(b.rules) == 0 {
 		return true
 	}
 	for _, r := range b.rules {
--- a/server/schedule/operator/builder.go
+++ b/server/schedule/operator/builder.go
@@ -69,6 +69,7 @@ type Builder struct {
 
 	// skip origin check flags
 	skipOriginJointStateCheck bool
+	skipPlacementRulesCheck   bool
 
 	// build flags
 	useJointConsensus bool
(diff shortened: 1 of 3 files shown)


server/schedule/operator/builder.go
```diff
@@ -138,7 +138,7 @@ func NewBuilder(desc string, ci ClusterInformer, region *core.RegionInfo, opts .
 
 	// placement rules
 	var rules []*placement.Rule
-	if err == nil && b.GetOpts().IsPlacementRulesEnabled() {
+	if err == nil && !b.skipCorrectnessCheck && b.GetOpts().IsPlacementRulesEnabled() {
```
CabinfeverB, line 141: If a reigon needs to be split, should we allow it to transfer to the leader

CabinfeverB (commented):
rest LGTM

server/schedule/operator/builder.go
HunDunDM: It should only be valid for `CreateLeaveJointStateOperator` and `CreateForceTransferLeaderOperator`. Now I use two flags instead.

CabinfeverB (approved):

server/schedule/operator/create_operator_test.go
```diff
@@ -1078,3 +1080,61 @@ func (suite *createOperatorTestSuite) TestMoveRegionWithoutJointConsensus() {
...
+	cluster.AddRegionStore(1, 1)
+	cluster.AddRegionStore(2, 1)
+	cluster.AddRegionStore(3, 1)
+	cluster.AddRegionStore(4, 1)
+	originPeers := []*metapb.Peer{
+		{Id: 3, StoreId: 3, Role: metapb.PeerRole_DemotingVoter},
+		{Id: 4, StoreId: 4, Role: metapb.PeerRole_IncomingVoter},
+	}
+
+	region := core.NewRegionInfo(&metapb.Region{Id: 1, Peers: originPeers, StartKey: []byte("a"), EndKey: []byte("c")}, originPeers[0])
```
nolouch, line 1127: why cannot fit any rules in this case?

nolouch (approved):
LGTM

server/schedule/operator/create_operator_test.go
CabinfeverB: It crosses two rule(range)

nolouch: I expect the default, is it not normal?

CabinfeverB: It is not normal if crossing multiple ranges, because of RegionFit will not work. The region have to be split first.

Issue #5468, opened 2022-09-01
simulator: make simulator great again


lhy1024:
## Enhancement Task

- [ ] support load hot region info from file
- [ ] support report hot read region in store heartbeat
- [ ] support joint consensus
- [ ] add more log

bufferflies:
how about new sub task to using the describe file to generate workload.

lhy1024:
It is related with https://github.com/tikv/pd/issues/2500

PR #5472, opened 2022-09-02
test: fix query statistic

lhy1024:

Issue Number: Ref #5468

1. rename `QueryStats` to `queryStats`
2. add `queryStats` to `region.Clone`, which is used in test and simulator
3. add `addQueryStats`, which is used in test and simulator

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```

### DIFF
--- a/server/core/region_option.go
+++ b/server/core/region_option.go
@@ -220,7 +220,7 @@ func SetWrittenQuery(v uint64) RegionCreateOption {
 func SetQueryNum(read, write uint64) RegionCreateOption {
 	r := RandomKindReadQuery(read)
 	w := RandomKindWriteQuery(write)
-	q := MergeQueryStat(r, w)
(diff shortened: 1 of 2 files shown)


bufferflies (approved):

server/core/region_option.go
```diff
@@ -326,3 +345,28 @@ func SetFromHeartbeat(fromHeartbeat bool) RegionCreateOption {
 		region.fromHeartbeat = fromHeartbeat
 	}
 }
+
+// MergeQueryStat merges two query stats.
+func MergeQueryStat(q1, q2 *pdpb.QueryStats) *pdpb.QueryStats {
```
bufferflies, line 350: Does it need to expose ? How about private?

nolouch (approved):

Issue #5491, opened 2022-09-13
multi-region-dr: pd leader panic repeatly 


mayjiang0203:
### What did you see instead?

![image](https://user-images.githubusercontent.com/9443637/189796009-a42fb889-f462-47c5-978e-ef98051447d0.png)

```
[2022/09/11 13:18:38.927 +00:00] [INFO] [operator_controller.go:450] ["add operator"] [region-id=86007] [operator="\"rule-split-region {split: region 86007 use policy USEKEY and keys [7480000000000000FF5800000000000000F8]} (kind:split, region:86007(61, 39), createAt:2022-09-11 13:18:38.92786507 +0000 UTC m=+116.187335222, startAt:0001-01-01 00:00:00 +0000 UTC, currentStep:0, size:1, steps:[split region with policy USEKEY])\""] [additional-info="{\"region-end-key\":\"7480000000000000FF5900000000000000F8\",\"region-start-key\":\"7480000000000000FF5700000000000000F8\"}"]
[2022/09/11 13:18:38.928 +00:00] [INFO] [operator_controller.go:652] ["send schedule command"] [region-id=86007] [step="split region with policy USEKEY"] [source=create]
[...]

PR #5494, opened 2022-09-13
checker: fix the issue that unhealthy region cause panic

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/5491

```commit-message
checker: fix the issue that unhealthy region cause panic
```

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue that unhealthy region cause panic
```

### DIFF
--- a/server/schedule/checker/rule_checker.go
+++ b/server/schedule/checker/rule_checker.go
@@ -37,6 +37,7 @@ var (
 	errNoStoreToReplace   = errors.New("no store to replace peer")
 	errPeerCannotBeLeader = errors.New("peer cannot be leader")
 	errNoNewLeader        = errors.New("no new leader")
+	errRegionNoLeader     = errors.New("region no leader")
 )
 
 const maxPendingListLen = 100000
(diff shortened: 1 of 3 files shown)


server/schedule/checker/rule_checker.go
```diff
@@ -236,10 +236,14 @@ func (c *RuleChecker) fixLooseMatchPeer(region *core.RegionInfo, fit *placement.
 		checkerCounter.WithLabelValues("rule_checker", "fix-peer-role").Inc()
 		return operator.CreatePromoteLearnerOperator("fix-peer-role", c.cluster, region, peer)
 	}
+	if region.GetLeader() == nil {
+		checkerCounter.WithLabelValues("rule_checker", "region-no-leader").Inc()
+		return nil, nil
```
CabinfeverB, line 241: Maybe we can return an error, so L158 can return

HunDunDM: agree

rleungx: If the region doesn't have a leader, should we check it?

nolouch:
ptal @HunDunDM @CabinfeverB

server/schedule/checker/rule_checker.go
nolouch: @rleungx  I think no need, address it.

disksing (approved):

PR #5496, opened 2022-09-13
scheduler: fix target filter in hot scheduler when source peer is leader

### LINKED ISSUE #5492: scheduler: continuous `move-hot-read-peer` when exists `evict-leader-scheduler`
### What did you do?

bench tpcc in 6 node cluster, when there is `evict-leader-scheduler`

stable schedule

### What did you see instead?

continuous `move-hot-read-peer` and a lot of `create-operator-fail` in hot scheduler event
image

master

lhy1024:

Issue Number: Close #5492 

For the read load, there are actually two kinds, one is to read data from the leader, and the other is to read data from the follower, which includes stale read and follower read.

For strong read, there are two kinds of operators, including transfer-leader and move-leader, move-leader is a combination of transfer-leader and move-peer; for stale read and follower read, we only have one kind of operator, that is, move-peer.

In the previous strategy, for the move-peer operator, if the role of the source peer is leader, it will be changed to move-leader. That is to say, the judgment and filtering of the move-leader operator are completely based on the move-peer. So, we don't consider whether the target store can receive the leader peer. For example, if there is an evict leader scheduler on the target node, then it will appear lots of `create-operators-fail` in the scheduler event. 

image

This is not the worst. The move-leader was originally used to schedule strong read. Obviously, the stddev and expectation should be calculated according to the distribution of the load on the leader. However, since the move-leader operator is generated based on the move-peer in the previous strategy, the calculation of stddev and expectation are also based on peers rather than leaders, so when a node has an evict leader scheduler, its variance will never be balanced, so it will always try to schedule.
```
[2022/09/11 14:10:27.654 +00:00] [Info] [hot_region.go:570] ["filter uniform store"] [type=transfer-leader] [src=8] [dst=1] [equal=true] ["src first stddev"=0.15777655162390164] ["dst first stddev"=0.15777655162390164] ["src second stddev"=0.0431061068388591] ["dst second stddev"=0.0431061068388591]

[2022/09/11 14:10:27.655 +00:00] [Info] [hot_region.go:570] ["filter uniform store"] [type=move-peer] [src=6] [dst=4] [equal=true] ["src first stddev"=0.4794498183238738] ["dst first stddev"=0.4794498183238738] ["src second stddev"=0.44969963724264195] ["dst second stddev"=0.44969963724264195]
```

In the new strategy, we decide the type of schedule generated according to the role of the source peer, in another word, the generation of move-leader is no longer based on move-peer but transfer-leader. 

For source peers whose roles are leader, we will try to select all stores, for those stores that contain followers, the transfer-leader will be generated, if it does not contain followers, the move-leader will be generated; for source peers whose roles are follower, we will only generate move-peer.

### Check List

Tests

- Unit test
- Integration test
- Manual test

a cluster with 6 nodes but a node is evicted leader, strict-pick-store is false.

there are both uniform workloads before this pr and after this pr, but there is a lot of operators before this pr.
image
image

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -605,11 +605,12 @@ type ruleLeaderFitFilter struct {
 	region           *core.RegionInfo
 	oldFit           *placement.RegionFit
 	srcLeaderStoreID uint64
+	allowMoveLeader  bool
 }
 
 // newRuleLeaderFitFilter creates a filter that ensures after transfer leader with new store,
 // the isolation level will not decrease.
-func newRuleLeaderFitFilter(scope string, cluster *core.BasicCluster, ruleManager *placement.RuleManager, region *core.RegionInfo, srcLeaderStoreID uint64) Filter {
+func newRuleLeaderFitFilter(scope string, cluster *core.BasicCluster, ruleManager *placement.RuleManager, region *core.RegionInfo, srcLeaderStoreID uint64, allowMoveLeader bool) Filter {
 	return &ruleLeaderFitFilter{
 		scope:            scope,
 		cluster:          cluster,
@@ -617,6 +618,7 @@ func newRuleLeaderFitFilter(scope string, cluster *core.BasicCluster, ruleManage
 		region:           region,
 		oldFit:           ruleManager.FitRegion(cluster, region),
 		srcLeaderStoreID: srcLeaderStoreID,
+		allowMoveLeader:  allowMoveLeader,
 	}
 }
 
@@ -634,13 +636,17 @@ func (f *ruleLeaderFitFilter) Source(options *config.PersistOptions, store *core
 
 func (f *ruleLeaderFitFilter) Target(options *config.PersistOptions, store *core.StoreInfo) *plan.Status {
 	targetPeer := f.region.GetStorePeer(store.GetID())
+	newRegionOptions := []core.RegionCreateOption{core.WithLeader(targetPeer)}
 	if targetPeer == nil {
-		log.Warn("ruleLeaderFitFilter couldn't find peer on target Store", zap.Uint64("target-store", store.GetID()))
-		return statusStoreNotMatchRule
+		if !f.allowMoveLeader {
+			log.Warn("ruleLeaderFitFilter couldn't find peer on target Store", zap.Uint64("target-store", store.GetID()))
+			return statusStoreNotMatchRule
+		}
+		newRegionOptions = append(newRegionOptions, core.WithReplacePeerStore(f.srcLeaderStoreID, store.GetID()))
 	}
 	copyRegion := createRegionForRuleFit(f.region.GetStartKey(), f.region.GetEndKey(),
-		f.region.GetPeers(), f.region.GetLeader(),
-		core.WithLeader(targetPeer))
+		f.region.GetPeers(), f.region.GetLeader(), newRegionOptions...,
+	)
 	newFit := f.ruleManager.FitRegion(f.cluster, copyRegion)
 	if placement.CompareRegionFit(f.oldFit, newFit) <= 0 {
 		return statusOK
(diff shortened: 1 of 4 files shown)


server/schedulers/hot_region.go
```diff
@@ -826,29 +826,57 @@ func (bs *balanceSolver) filterDstStores() map[uint64]*statistics.StoreLoadDetai
...
 			filter.NewSpecialUseFilter(bs.sche.GetName(), filter.SpecialUseHotRegion),
 			filter.NewPlacementSafeguard(bs.sche.GetName(), bs.GetOpts(), bs.GetBasicCluster(), bs.GetRuleManager(), bs.cur.region, srcStore),
 		}
-
 		for _, detail := range bs.stLoadDetail {
 			candidates = append(candidates, detail)
 		}
 
 	case transferLeader:
+		if !bs.cur.mainPeerStat.IsLeader() { // move leader and transfer leader
```
nolouch, line 843: use better comments..

server/schedule/filter/filters.go
```diff
@@ -664,9 +670,13 @@ func NewPlacementSafeguard(scope string, opt *config.PersistOptions, cluster *co
 // NewPlacementLeaderSafeguard creates a filter that ensures after transfer a leader with
 // existed peer, the placement restriction will not become worse.
 // Note that it only worked when PlacementRules enabled otherwise it will always permit the sourceStore.
-func NewPlacementLeaderSafeguard(scope string, opt *config.PersistOptions, cluster *core.BasicCluster, ruleManager *placement.RuleManager, region *core.RegionInfo, sourceStore *core.StoreInfo) Filter {
+func NewPlacementLeaderSafeguard(scope string, opt *config.PersistOptions, cluster *core.BasicCluster, ruleManager *placement.RuleManager, region *core.RegionInfo, sourceStore *core.StoreInfo, allowMoveLeaders ...bool) Filter {
```
nolouch, line 673: Do not use `...bool`, `option` or add a new parameter.

nolouch (commented):
overall lgtm. ptal @HunDunDM

server/schedule/filter/filters.go
```diff
@@ -634,13 +636,17 @@ func (f *ruleLeaderFitFilter) Source(options *config.PersistOptions, store *core
...
 	targetPeer := f.region.GetStorePeer(store.GetID())
+	newRegionOptions := []core.RegionCreateOption{core.WithLeader(targetPeer)}
 	if targetPeer == nil {
-		log.Warn("ruleLeaderFitFilter couldn't find peer on target Store", zap.Uint64("target-store", store.GetID()))
-		return statusStoreNotMatchRule
+		if !f.allowMoveLeader {
+			log.Warn("ruleLeaderFitFilter couldn't find peer on target Store", zap.Uint64("target-store", store.GetID()))
+			return statusStoreNotMatchRule
+		}
+		newRegionOptions = append(newRegionOptions, core.WithReplacePeerStore(f.srcLeaderStoreID, store.GetID()))
```
rleungx, line 645: Will the `core.WithLeader(targetPeer)` actually be `core.WithLeader(nil)`?

HunDunDM (approved):

server/schedulers/hot_region.go
```diff
@@ -826,29 +826,57 @@ func (bs *balanceSolver) filterDstStores() map[uint64]*statistics.StoreLoadDetai
...
+				}
+				// transfer leader
+				if slice.AnyOf(peers, func(i int) bool {
+					return peers[i].GetStoreId() == storeID
+				}) {
+					candidates = append(candidates, detail)
+					continue
+				}
+				// move leader
+				if filter.Target(bs.GetOpts(), detail.StoreInfo, moveLeaderFilters) {
```
rleungx, line 868: Will it be executed twice?

lhy1024: No

rleungx (approved):

nolouch (approved):
lgtm

PR #5499, opened 2022-09-14
cache: add a function for FIFO

CabinfeverB:

Issue Number: Ref #5257 

add a function to get continuous items that have the same comparable attribute with the the lastest one.

```commit-message
get continuous items that have the same comparable attribute with the the lastest one.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -221,6 +221,41 @@ func TestFifoCache(t *testing.T) {
 	re.Equal(0, cache.Len())
 }
 
+type testStruct struct {
+	value string
+}
+
+func TestFifoFromLastSameElems(t *testing.T) {
+	t.Parallel()
+	re := require.New(t)
+	cache := NewFIFO(4)
+	cache.Put(1, &testStruct{value: "1"})
+	cache.Put(1, &testStruct{value: "2"})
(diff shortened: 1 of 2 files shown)


pkg/cache/cache_test.go
```diff
@@ -221,6 +221,41 @@ func TestFifoCache(t *testing.T) {
 	re.Equal(0, cache.Len())
 }
 
+type testStruct struct {
```
rleungx, line 224: How about putting it inside of the test function?

pkg/cache/fifo.go
```diff
@@ -89,6 +89,25 @@ func (c *FIFO) FromElems(key uint64) []*Item {
 	return elems
 }
 
+// FromLastSameElems returns continuous items that have the same comparable attribute with the the lastest one.
+func (c *FIFO) FromLastSameElems(checkFunc func(interface{}) bool, comparableFunc func(interface{}) string) []*Item {
```
bufferflies, line 93: It is not friendly to use two function to check item. howa bout  checkFunc func(interface{})bool,string

rleungx (approved):
LGTM, and maybe we can make it more general in the future.

bufferflies (approved):

PR #5511, opened 2022-09-15
scheduler: fix write leader max zombie duration

lhy1024:

Issue Number: Ref #4949

`maxZombieDur` is not correct when the first priority is query in hot write scheduler, query is reported by store heartbeat rather than region heartbeat, so the interval should be base store heartbeat rather than region heartbeat

It's the same reason with [`tikvCollector.GetLoads`](https://github.com/tikv/pd/blob/bca2afae417921c9e31fd2c208eea485d4d3597d/server/statistics/collector.go#L52)

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -656,6 +656,9 @@ func (bs *balanceSolver) calcMaxZombieDur() time.Duration {
 	switch bs.resourceTy {
 	case writeLeader:
 		if bs.firstPriority == statistics.QueryDim {
+			// We use store query info rather than total of hot write leader to guide hot write leader scheduler
+			// when its first priority is QueryDim`, because `Write-peer` does not have `QueryDim`.
+			// The reason is similar with `tikvCollector.GetLoads`
 			return bs.sche.conf.GetStoreStatZombieDuration()
 		}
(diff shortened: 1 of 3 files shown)


HunDunDM (approved):

nolouch:
plz describe why this changes.

lhy1024:
> query is reported by store heartbeat rather than region heartbeat.

query is reported by store heartbeat rather than region heartbeat, so the interval should be base store heartbeat rather than region heartbeat

server/schedulers/hot_region.go
```diff
@@ -664,6 +650,25 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	return true
 }
 
+// Depending on the source of the statistics used, a different ZombieDuration will be used.
+// If the statistics are from the sum of Regions, there will be a longer ZombieDuration.
+func (bs *balanceSolver) calcMaxZombieDur() time.Duration {
+	switch bs.resourceTy {
+	case writeLeader:
+		if bs.firstPriority == statistics.QueryDim {
```
nolouch, line 658: add some comments.

nolouch (approved):

PR #5521, opened 2022-09-16
statistic: collect both read and write pending influence at the same time

lhy1024:

Issue Number: Ref #4949

Although there is a `regionPendings` in the hot scheduler, there is only read or write load during `tryAddPendingInfluence`.

Hot write scheduler and hot read scheduler may affect each other but cannot be perceived by each other.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedulers/hot_region.go
+++ b/server/schedulers/hot_region.go
@@ -420,6 +420,8 @@ type balanceSolver struct {
 	minorDecRatio float64
 	maxPeerNum    int
 	minHotDegree  int
+
+	checkByPriorityAndTolerance func(loads []float64, f func(int) bool) bool
 }
 
 func (bs *balanceSolver) init() {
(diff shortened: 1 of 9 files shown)


server/statistics/kind.go
```diff
@@ -154,6 +154,17 @@ func (rw RWType) RegionStats() []RegionStatKind {
...
 
+// Inverse returns the opposite of kind.
+func (rw RWType) Inverse() RWType {
+	switch rw {
+	case Write:
+		return Read
+	case Read:
+		return Write
+	}
+	return Read
```
nolouch, line 165: why default read?

server/schedulers/hot_region.go
```diff
@@ -651,6 +649,17 @@ func (bs *balanceSolver) tryAddPendingInfluence() bool {
 	return true
 }
 
+func (bs *balanceSolver) collectPendingInfluence(peer *statistics.HotPeerStat) statistics.Influence {
+	infl := statistics.Influence{Loads: make([]float64, statistics.RegionStatCount), Count: 1}
+	bs.rwTy.SetFullLoadRates(infl.Loads, peer.GetLoads())
+	inverse := bs.rwTy.Inverse()
+	another := bs.GetHotPeerStat(inverse, peer.RegionID, peer.StoreID)
```
nolouch, line 656: I prefer summary all loads at a same time (including read and write) in `prepareForBalance`, then get it from the `stLoadInfos`. it is at a same "snapshot" rather than get it one by one.

HunDunDM: Now we don't need to use read and write statistics together. I think we can do this when we unify read and write schedulers later.

server/statistics/kind.go
HunDunDM: == `Case Write:`

lhy1024: there are only two kind in RWType

server/schedulers/hot_region.go
lhy1024: yes, it will be placed into together when we unify read and write schedulers.

rleungx (approved):

server/statistics/kind.go
nolouch: I think we can panic if not match

server/statistics/hot_peer_cache.go
```diff
@@ -366,13 +366,19 @@ func (f *hotPeerCache) isRegionHotWithPeer(region *core.RegionInfo, peer *metapb
...
+	}
+	return false
+}
+
+func (f *hotPeerCache) getHotPeerStat(regionID, storeID uint64) *HotPeerStat {
 	if peers, ok := f.peersOfStore[storeID]; ok {
-		if stat := peers.Get(region.GetID()); stat != nil {
-			return stat.(*HotPeerStat).HotDegree >= hotDegree
+		if stat := peers.Get(regionID); stat != nil {
+			return stat.(*HotPeerStat)
```
HunDunDM, line 378: Are there times when there may be insufficient statistical confidence?

HunDunDM (commented):
rest LGTM

HunDunDM (approved):

PR #5526, opened 2022-09-17
scheduler: fix policy with `write-leader` in hot scheduler

HunDunDM:

Issue Number: Ref #4949

The write leader dimension only compares the first dimension when it is good, and should only compare the first dimension when filtering.

### Check List

Tests

### Release note

```release-note
None
```


lhy1024 (approved):

nolouch (approved):
lgtm

PR #5534, opened 2022-09-20
schedulers: Add nil check for Collector

CabinfeverB:

reduce the cost of `runtime.newObject` for `Option`

Issue Number: ref #5257 

add nil check for collector
```commit-message
add nil check for collector
```

### Check List

Tests

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/filter/filters.go
+++ b/server/schedule/filter/filters.go
@@ -40,7 +40,9 @@ func SelectSourceStores(stores []*core.StoreInfo, filters []Filter, opt *config.
 				targetID := ""
 				filterCounter.WithLabelValues("filter-source", s.GetAddress(),
 					sourceID, filters[i].Scope(), filters[i].Type(), sourceID, targetID).Inc()
-				collector.Collect(plan.SetResource(s), plan.SetStatus(status))
+				if collector != nil {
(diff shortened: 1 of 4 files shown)


bufferflies (approved):

server/schedule/filter/filters.go
```diff
@@ -40,7 +40,9 @@ func SelectSourceStores(stores []*core.StoreInfo, filters []Filter, opt *config.
 				targetID := ""
 				filterCounter.WithLabelValues("filter-source", s.GetAddress(),
 					sourceID, filters[i].Scope(), filters[i].Type(), sourceID, targetID).Inc()
-				collector.Collect(plan.SetResource(s), plan.SetStatus(status))
+				if collector != nil {
+					collector.Collect(plan.SetResource(s), plan.SetStatus(status))
```
nolouch, line 44: Do you need to check it? I think already do it in the `Collect` function.

CabinfeverB: Because `plan.SetResource(s)` and `SetResource` will be executed before executing `Collect ` and they will create `newObject` for `Option`

rleungx (approved):

PR #5556, opened 2022-09-26
Redefine the total label query for balance leader and region schedulers

### LINKED ISSUE #5393: The `total` in monitor doesn't mean the real total number of scheduling
### What did you do?

Try to find why balance does not work.

The total should equal the sum of other labels.

### What did you see instead?

Screen Shot 2022-08-03 at 5 12 33 PM

It's even smaller than `skip`. This happens in both the balance leader scheduler and the balance region scheduler.

master

matchge-ca:
Signed-off-by: Hua Lu <hualu@gmail.com>

Issue Number: Close #5393 

```commit-message
* Remove the total label metric counters for leader and region
* Add new query in PD's grafana balance leader scheduler and balance region scheduler panels for the sum of all labels and name it as total
```

### Check List

Tests

  - Build source code, copy over pd-server and pd.json in tiup directory
  - Start tiup playground and monitor the two scheduler panels
  - Verify the total label is the sum of all other labels
Screen Shot 2022-09-26 at 11 12 45 AM

Code changes

Side effects

Related changes

### Release note

```release-note
None
```


lhy1024 (approved):

rleungx:
The previous "total" means the number of scheduling including retry. But right now, it's slightly different.

matchge-ca:
> The previous "total" means the number of scheduling including retry. But right now, it's slightly different.

Yes, we can still collect the metric of total number of scheduling, but with different name, such as "schedule-sum", or "total-scheduling", and keep the "total" label as the sum of all other labels.

nolouch (approved):
I think before `total` is meaningless, remove it LGTM. 
PTAL @rleungx

rleungx:
> > The previous "total" means the number of scheduling including retry. But right now, it's slightly different.
> 
> Yes, we can still collect the metric of total number of scheduling, but with different name, such as "schedule-sum", or "total-scheduling", and keep the "total" label as the sum of all other labels.

Maybe we can just remove it, the sum of all other labels is also meaningless.

PR #5561, opened 2022-09-27
schedule: batch to report the metrics of target filter

### LINKED ISSUE #5538: the metrics of target filter shoule batch to report .
In big cluster, there are lots of filter metrics to record , but the counter need to hash, so it is too slow. We can batch them or by sampling. 
If I closed this metrics, it imporve 1.2x performance.

bufferflies:

Issue Number: Close #5538

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
master:
image
image

This PR：

image

image

Code changes

Side effects

Related changes

### Release note

```release-note
 None.
```


server/schedule/filter/metrics.go
```diff
@@ -23,7 +23,7 @@ var (
 			Subsystem: "schedule",
 			Name:      "filter",
 			Help:      "Counter of the filter",
-		}, []string{"action", "address", "store", "scope", "type", "source", "target"})
+		}, []string{"action", "scope", "type", "store"})
```
rleungx, line 26: The prom query also needs to be changed.

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,221 @@
...
+func (s scope) String() string {
+	if s >= ScopeLen {
+		return "unknown"
+	}
+	return scopes[s]
+}
+
+type filterType int
+
+const (
```
rleungx, line 91: IMO, this could be a little hard to maintain in the future.

server/schedule/filter/metrics.go
bufferflies: the sub pr is: https://github.com/tikv/pd/pull/5577

server/schedule/filter/counter.go
bufferflies: yes, I will fix it by add the sub filter type to the filter type.

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,205 @@
...
+	excludedFilterType filterType = iota
```
rleungx, line 91: I prefer to remove `FilterType` suffix which makes the variable shorter.

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,205 @@
...
+	"storage-threshold-filter",
+	"distinct-filter",
+	"label-constraint-filter",
+	"rule-fit-filter",
+	"rule-fit-leader-filter",
+	"engine-filter",
+	"special-use-filter",
+	"isolation-filter",
+	"region-score-filter",
+	"idFilter",
```
rleungx, line 131: Can we remove it since it is only used in the test? If not, I prefer to add a comment.

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,205 @@
...
+	storeStateExceedRemoveLimitFilterType
+	storeStateExceedAddLimitFilterType
+	storeStateTooManySnapshotFilterType
+	storeStateTooManyPendingPeerFilterType
+	storeStateRejectLeaderFilterType
+
+	filtersLen
+)
+
+var filters = [filtersLen]string{
```
rleungx, line 120: Please add a test or use some way to make sure the sequence are same as the definition of FilterType.

server/schedule/filter/counter.go
bufferflies: fixed

server/schedule/filter/counter.go
bufferflies: fixed by adding unit test.

server/schedule/filter/counter.go
bufferflies: fixed by add `used by test`

server/schedulers/balance_leader.go
```diff
@@ -392,6 +394,7 @@ func (l *balanceLeaderScheduler) Schedule(cluster schedule.Cluster, dryRun bool)
 			}
 		}
 	}
+	l.filterCounter.Flush()
```
rleungx, line 397: Should we flush it at a fixed interval?

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,218 @@
...
+		counter[i] = make([]map[uint64]map[uint64]int, filtersLen)
+		for k := range counter[i] {
+			counter[i][k] = make(map[uint64]map[uint64]int)
+		}
+	}
+	return &Counter{counter: counter, scope: scope}
+}
+
+// Add adds the filter counter.
+func (c *Counter) inc(action action, filterType filterType, sourceID uint64, targetID uint64) {
```
rleungx, line 192: It seems that this function is only used in the test?

server/schedule/filter/counter.go
```diff
@@ -0,0 +1,218 @@
...
+	distinctScore
+	labelConstraint
+	ruleFit
+	ruleLeader
+	engine
+	specialUse
+	isolation
+	regionScore
+	// used in test.
+	idFilterType
```
rleungx, line 114: I still think we should remove it.

server/schedule/filter/filters.go
```diff
@@ -88,7 +98,7 @@ func filterStoresBy(stores []*core.StoreInfo, keepPred func(*core.StoreInfo) boo
 type Filter interface {
 	// Scope is used to indicate where the filter will act on.
 	Scope() string
-	Type() string
+	Type() filterType
```
rleungx, line 101: Why not just keep use string?

bufferflies: It is used in array to check the index . if keeps the string type, the counter should be defined as counter []map[string]map[uint64]map[uint64] int , it add one hash calculate .

server/schedulers/balance_leader.go
bufferflies: I think it will do it  in next pr .

server/schedule/filter/counter.go
bufferflies: fixed

rleungx (approved):

CabinfeverB (approved):

PR #5575, opened 2022-10-09
schedule: refine the operator priority

### LINKED ISSUE #5388: Support more accurate operator priority
## Enhancement Task
At present, we indeed have the priority for operators. But many kinds of operators share the high priority, which means they cannot replace each other. We need to tell these kinds of operators apart from the priority. e.g. The admin operator should have the highest priority.

rleungx:

Issue Number: Close #5388

This PR does the following things:
1. For admin operators, we use Urgent priority. These operators will also pass the store limit check.
2. For shuffle/random operators, which are used in the testing environment, we use Low priority.
3. For hot/replica/evict/grant operators, we use High priority.
4. The others use Medium priority.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/core/kind.go
+++ b/server/core/kind.go
@@ -22,7 +22,7 @@ const (
 	Low PriorityLevel = iota
 	Medium
 	High
-	VeryHigh
+	Urgent
 )
 
 // ScheduleKind distinguishes resources and schedule policy.
--- a/server/core/kind.go
+++ b/server/core/kind.go
@@ -19,9 +19,10 @@ type PriorityLevel int
 
 // Built-in priority level
 const (
-	LowPriority PriorityLevel = iota
-	NormalPriority
-	HighPriority
+	Low PriorityLevel = iota
+	Medium
+	High
+	VeryHigh
 )
 
 // ScheduleKind distinguishes resources and schedule policy.
--- a/server/schedule/operator_controller.go
+++ b/server/schedule/operator_controller.go
@@ -831,6 +831,10 @@ func (oc *OperatorController) ExceedStoreLimit(ops ...*operator.Operator) bool {
 
 // exceedStoreLimitLocked returns true if the store exceeds the cost limit after adding the operator. Otherwise, returns false.
 func (oc *OperatorController) exceedStoreLimitLocked(ops ...*operator.Operator) bool {
+	// The operator with Urgent priority, like admin operators, should ignore the store limit check.
+	if len(ops) != 0 && ops[0].GetPriorityLevel() == core.Urgent {
+		return false
+	}
 	opInfluence := NewTotalOpInfluence(ops, oc.cluster)
 	for storeID := range opInfluence.StoresInfluence {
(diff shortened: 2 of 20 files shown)


server/core/kind.go
```diff
@@ -19,9 +19,10 @@ type PriorityLevel int
 
 // Built-in priority level
 const (
-	LowPriority PriorityLevel = iota
-	NormalPriority
-	HighPriority
+	Low PriorityLevel = iota
+	Medium
+	High
+	VeryHigh
```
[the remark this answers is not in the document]
rleungx: Ok, I prefer Urgent

server/schedulers/grant_leader.go
```diff
@@ -252,7 +252,7 @@ func (s *grantLeaderScheduler) Schedule(cluster schedule.Cluster, dryRun bool) (
 			continue
 		}
 		op.Counters = append(op.Counters, schedulerCounter.WithLabelValues(s.GetName(), "new-operator"))
-		op.SetPriorityLevel(core.HighPriority)
+		op.SetPriorityLevel(core.High)
```
bufferflies, line 255: `grant leader` and `grant hot leader` are used in test, so the priority should be same with the shuffle.

server/schedulers/evict_leader.go
```diff
@@ -348,7 +348,7 @@ func scheduleEvictLeaderOnce(name, typ string, cluster schedule.Cluster, conf ev
 			log.Debug("fail to create evict leader operator", errs.ZapError(err))
 			continue
 		}
-		op.SetPriorityLevel(core.HighPriority)
+		op.SetPriorityLevel(core.High)
```
bufferflies, line 351: I think the priority of `evict leader` should be urgent, It should be higher than the hot region.

server/schedulers/grant_leader.go
rleungx: IMO, grant leader can also be used in the production environment.

server/schedulers/evict_leader.go
rleungx: make sense

server/schedule/operator_controller.go
```diff
@@ -831,6 +831,10 @@ func (oc *OperatorController) ExceedStoreLimit(ops ...*operator.Operator) bool {
 
 // exceedStoreLimitLocked returns true if the store exceeds the cost limit after adding the operator. Otherwise, returns false.
 func (oc *OperatorController) exceedStoreLimitLocked(ops ...*operator.Operator) bool {
+	// The operator with Urgent priority, like admin operators, should ignore the store limit check.
+	if len(ops) != 0 && ops[0].GetPriorityLevel() == core.Urgent {
```
bufferflies, line 835: the condition `len(ops)>0` can be ignored because ops can not be nil in this step,.

bufferflies (approved):
LGTM

nolouch (approved):

Issue #5539, opened 2022-09-21
some filter should run once


bufferflies:
The storeStateFilter, SpecialUseFilter, RegionScoreFilter can be run once , not every target. 
some related code:
https://github.com/tikv/pd/blob/0433eea9bf6de3526fa40b8a96a88ac7891fc9b0/server/schedulers/balance_region.go#L230-L236

PR #5591, opened 2022-10-12
schdeuler: Set resource with step for collecting plan

CabinfeverB:

Issue Number: ref #5539
If we determine  which stores cannot be targets at the beginning of the scheduling,  we should set resource with step. 

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/schedule/plan/plan.go
+++ b/server/schedule/plan/plan.go
@@ -23,6 +23,10 @@ type Plan interface {
 
 	Clone(ops ...Option) Plan // generate plan for clone option
 	SetResource(interface{})
+	// SetResourceWithStep is used to set resource for specific step.
+	// The meaning of step is different for different plans.
+	// Such as balancePlan, pickSource = 0, pickRegion = 1, pickTarget = 2
+	SetResourceWithStep(resource interface{}, step int)
 	SetStatus(*Status)
 }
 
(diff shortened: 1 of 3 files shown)


server/schedule/plan/plan.go
```diff
@@ -23,6 +23,7 @@ type Plan interface {
 
 	Clone(ops ...Option) Plan // generate plan for clone option
 	SetResource(interface{})
+	SetResourceWithStep(interface{}, int)
```
bufferflies, line 26: how about use enum?

server/schedule/plan/plan.go
```diff
@@ -83,3 +84,10 @@ func SetResource(resource interface{}) Option {
 		plan.SetResource(resource)
 	}
 }
+
+// SetResource is used to generate Resource for plan
+func SetResourceWithStep(resource interface{}, step int) Option {
```
rleungx, line 89: Where is it used?

CabinfeverB: https://github.com/tikv/pd/pull/5544 will need it

bufferflies (approved):
lgtm

rleungx (approved):

PR #5609, opened 2022-10-19
simulator: support change peer v2

### LINKED ISSUE #5469: simulator: support joint consensus
## Development Task
If we don't support joint consensus, then tikv after v4.0 could not run in the simulator, which led to some feature could not enable, such as hot schedule based query.

HunDunDM:

Issue Number: Close #5469 

Refactor `task.go`

* Split `AddVoter` and `PromoteLearner`
* Merge `AddVoter` and `AddLearner`
  * `PendingPeer` will now be added first, and will be removed after the simulated snap transfer is complete.
* Support original `ChangePeer` in v2 (`ChangePeerV2` and `len(Changes) == 1`)
  * Support `DemoteVoter`, and check if the Peer is not the Leader
* Support enter joint state (`ChangePeerV2` and `len(Changes) > 1`)
* Support leave joint state (`ChangePeerV2` and `len(Changes) == 0`)
* `TransferLeader` adds `Voter` checks
* `TransferLeader` supports the parsing field `Peers` (used to let TiKV find the optimal Peer by itself)
* `RemovePeer` now always also removes the corresponding `DownPeer` and `PendingPeer`

### Check List

Tests

### Release note

```release-note
None
```

### DIFF
--- a/server/core/region_option.go
+++ b/server/core/region_option.go
@@ -325,45 +325,12 @@ func WithAddPeer(peer *metapb.Peer) RegionCreateOption {
 	}
 }
 
-// WithPromoteLearner promotes the learner.
-func WithPromoteLearner(peerID uint64) RegionCreateOption {
+// WithRole changes the role.
+func WithRole(peerID uint64, role metapb.PeerRole) RegionCreateOption {
 	return func(region *RegionInfo) {
 		for _, p := range region.GetPeers() {
 			if p.GetId() == peerID {
-				p.Role = metapb.PeerRole_Voter
-			}
-		}
-	}
-}
-
-// WithPromoteLearnerEnter promotes the learner to the joint enter stage.
-func WithPromoteLearnerEnter(peerID uint64) RegionCreateOption {
-	return func(region *RegionInfo) {
-		for _, p := range region.GetPeers() {
-			if p.GetId() == peerID {
-				p.Role = metapb.PeerRole_IncomingVoter
-			}
-		}
-	}
-}
-
-// WithDemoteVoter demotes the voter.
-func WithDemoteVoter(peerID uint64) RegionCreateOption {
-	return func(region *RegionInfo) {
-		for _, p := range region.GetPeers() {
-			if p.GetId() == peerID {
-				p.Role = metapb.PeerRole_Learner
-			}
-		}
-	}
-}
-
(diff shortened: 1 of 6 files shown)


server/core/region_option.go
```diff
@@ -336,6 +336,39 @@ func WithPromoteLearner(peerID uint64) RegionCreateOption {
 	}
 }
 
+// WithPromoteLearnerEnter promotes the learner to the joint enter stage.
+func WithPromoteLearnerEnter(peerID uint64) RegionCreateOption {
```
bufferflies, line 340: how about exract one same function `WithRole(peerID unit64,role metapb.PeerRole)`

tools/pd-simulator/simulator/task.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 TiKV Project Authors.
+// Copyright 2022 TiKV Project Authors.
```
bufferflies, line 1: why change it?

tools/pd-simulator/simulator/task.go
```diff
@@ -49,380 +53,462 @@ const (
...
+	case resp.GetMerge() != nil:
 		targetRegion := resp.GetMerge().GetTarget()
-		return &mergeRegion{
-			regionID:     regionID,
-			epoch:        epoch,
-			targetRegion: targetRegion,
+		desc = fmt.Sprintf("merge region %d into %d", regionID, targetRegion.GetId())
+		op = &mergeRegion{targetRegion: targetRegion}
+	case resp.GetSplitRegion() != nil:
+		// TODO: support split region
```
lhy1024, line 116: when will we support it?

lhy1024 (commented):
Do we need some manual test?

tools/pd-simulator/simulator/task.go
HunDunDM: Caused by the newly created file, it has been rolled back.

tools/pd-simulator/simulator/task.go
HunDunDM: We need to discuss how to perform for the three different splits.

bufferflies (approved):

nolouch (commented):
/merge

nolouch (approved):

PR #5610, opened 2022-10-19
api: using index to replace tree scan if only returns count 

### LINKED ISSUE #5607: Add filter in /pd/api/v1/stats/region API

TiDB need to get region count between `startKey` and `endKey` from `/pd/api/v1/stats/region` to calculate TiFlash replica progress. `GetRegionStatus` will scan regions in `BTree` so it will have heavy cost if there are many regions.
In fact, [TiDB only need region count](https://github.com/pingcap/tidb/blob/dc223253f0cb1e5101ddf92b677e190da292811d/ddl/ddl_tiflash_api.go#L387), if `/pd/api/v1/stats/region` can add parameter to only get region count, the cost of `/pd/api/v1/stats/region` will reduce from O(N) to O(logN).

1 million regions, call `/pd/api/v1/stats/region` per 2 seconds, flame graph on PD:
7bae2d26-3aa0-43a9-b86f-dc7f79e39a7d

Add a parameter in `/pd/api/v1/stats/region` to get only region count.

bufferflies:

Issue Number: Close #5607

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
 None.
```


server/core/region.go
```diff
@@ -1160,6 +1161,27 @@ func (r *RegionInfo) GetWriteLoads() []float64 {
...
+	var endIndex int
+	var item rangetree.RangeItem
+	// it should return the length of the tree if endKey is nil.
+	if len(endKey) == 0 {
+		endIndex = r.tree.tree.Len() - 1
+	} else {
+		item, endIndex = r.tree.tree.GetWithIndex(end)
+		// it should return the endIndex - 1 if the endKey is the startKey of a region.
+		if item != nil && bytes.Equal(item.GetStartKey(), endKey) {
+			endIndex--
```
JmPotato, line 1179: Why not just return 1 here to make it more clear?

bufferflies:
the fixed pr flame:
image

server/core/region.go
bufferflies: It needs to consider the start key index if the endKey is equal the item startKey.

JmPotato: Do we have a test case to test this kind of situation?

JmPotato (commented):
The rest LGTM.

server/core/region.go
bufferflies: yes， the first case(startkey="",endKey="") can be found in the stats_test.go.

bufferflies (commented):
Yes, the ("","") will be found in the api unit test.

server/core/region.go
```diff
@@ -1174,6 +1175,27 @@ func (r *RegionInfo) GetWriteLoads() []float64 {
...
+func (r *RegionsInfo) GetRangeCount(startKey, endKey []byte) int {
+	start := &regionItem{&RegionInfo{meta: &metapb.Region{StartKey: startKey}}}
+	end := &regionItem{&RegionInfo{meta: &metapb.Region{StartKey: endKey}}}
+	// it returns 0 if startKey is nil.
+	_, startIndex := r.tree.tree.GetWithIndex(start)
+	var endIndex int
+	var item rangetree.RangeItem
+	// it should return the length of the tree if endKey is nil.
+	if len(endKey) == 0 {
+		endIndex = r.tree.tree.Len() - 1
```
CabinfeverB, line 1188: If start equals "", it seems return wrong ans?

bufferflies: No, the first region key range is start with "".

CabinfeverB (approved):

nolouch (approved):
lgtm

PR #5614, opened 2022-10-21
scheduler: modify baseSchedulePlan to fit new balance region impl

CabinfeverB:

Issue Number: ref #5257 

After #5544 , the top-score source store will not has status, so I add a status firstly.
Moreover, some top-score store will be `StatusStoreRemoveLimitThrottled` in past impl. And After #5544, these store will also has no status. So I pick status when they are filterd as source store.

```commit-message
modify baseSchedulePlan to fit new balance region impl 
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024 (commented):
would you like to add more details in pr desc?

nolouch:
PTAL @bufferflies

server/schedulers/balance_plan.go
```diff
@@ -74,11 +74,20 @@ func (p *balanceSchedulerPlan) GetResource(step int) uint64 {
 	}
 	switch step {
 	case pickSource:
-		return p.source.GetID()
+		if p.source != nil {
```
bufferflies, line 77: why  change it ? And when the p.source is nil?

CabinfeverB: When the top-score store is as target for `ScoreDisallowed`, I was worried about the potential misusage that would cause panic, but not so far. I will add a comment

server/schedulers/balance_test.go
```diff
@@ -787,15 +787,15 @@ func TestBalanceRegionSchedule1(t *testing.T) {
 	// test region replicate not match
 	opt.SetMaxReplicas(3)
 	ops, plans := sb.Schedule(tc, true)
-	re.Len(plans, 100)
+	re.Len(plans, 101)
 	re.Empty(ops)
-	re.Equal(int(plans[0].GetStatus().StatusCode), plan.StatusRegionNotReplicated)
+	re.Equal(int(plans[1].GetStatus().StatusCode), plan.StatusRegionNotReplicated)
```
bufferflies, line 792: how about one check for the plan[0]

bufferflies (approved):

rleungx (approved):

PR #5639, opened 2022-10-26
schedulers: add witness transfer leader scheduler

### LINKED ISSUE #5638: Add transfer leader scheduler
## Development Task

Add transfer leader scheduler when receive region heart beat.

Refer to https://github.com/tikv/tikv/issues/12876

ethercflow:
close tikv/pd#5638

Issue Number: Close #5638 

```commit-message
add transfer leader scheduler
```

### Check List

Tests

- Integration test

Code changes

- add a new scheduler

Side effects

- Increased code complexity

Related changes

- none

### Release note

```release-note
add transfer leader scheduler
```


server/config/config.go
```diff
@@ -1042,6 +1042,7 @@ var DefaultSchedulers = SchedulerConfigs{
 	{Type: "balance-leader"},
 	{Type: "hot-region"},
 	{Type: "split-bucket"},
+	{Type: "transfer-leader"},
```
rleungx, line 1045: How about promote-witness or something else, the current name sounds like the balance leader.

server/cluster/cluster.go
```diff
@@ -793,6 +793,7 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo) error {
 		peerInfo := core.NewPeerInfo(peer, region.GetWriteLoads(), interval)
 		c.hotStat.CheckWriteAsync(statistics.NewCheckPeerTask(peerInfo, region))
 	}
+	c.coordinator.CheckTransferLeader(region)
```
rleungx, line 796: How about dropping it once the channel is full?

server/schedulers/transfer_leader.go
```diff
@@ -0,0 +1,142 @@
...
+func (s *transferLeaderScheduler) GetName() string {
+	return TransferLeaderName
+}
+
+func (s *transferLeaderScheduler) GetType() string {
+	return TransferLeaderType
+}
+
+func (s *transferLeaderScheduler) IsScheduleAllowed(cluster schedule.Cluster) bool {
+	allowed := s.OpController.OperatorCount(operator.OpLeader) < cluster.GetOpts().GetLeaderScheduleLimit()
```
rleungx, line 76: Will it need to be restricted by the leader schedule limit?

server/config/config.go
ethercflow: Sure, thanks!

server/cluster/cluster.go
ethercflow: Sure, thanks!

server/schedulers/transfer_leader.go
ethercflow: I'm not sure about this place. Normally, when this scheduler is needed, it's dealing with a failure case scenario, so it's important and urgent but maybe not common. What do you think this limit should be set to? (Or just construct the scenario test later to get a reasonable limit)

server/cluster/coordinator.go
```diff
@@ -966,3 +967,17 @@ func (c *coordinator) getPausedSchedulerDelayUntil(name string) (int64, error) {
 	}
 	return s.GetDelayUntil(), nil
 }
+
+func (c *coordinator) CheckTransferWitnessLeader(region *core.RegionInfo) {
+	c.RLock()
+	defer c.RUnlock()
+	if s, ok := c.schedulers[schedulers.TransferWitnessLeaderName]; ok {
+		if schedulers.NeedTransferWitnessLeader(region) {
```
rleungx, line 975: How about checking it first before locking the coordinator?

server/config/config.go
```diff
@@ -1042,6 +1042,7 @@ var DefaultSchedulers = SchedulerConfigs{
 	{Type: "balance-leader"},
 	{Type: "hot-region"},
 	{Type: "split-bucket"},
+	{Type: "transfer-witness-leader"},
```
rleungx, line 1045: How about enabling it after all tests are finished?

server/cluster/coordinator.go
```diff
@@ -966,3 +967,17 @@ func (c *coordinator) getPausedSchedulerDelayUntil(name string) (int64, error) {
 	}
 	return s.GetDelayUntil(), nil
 }
+
+func (c *coordinator) CheckTransferWitnessLeader(region *core.RegionInfo) {
+	c.RLock()
+	defer c.RUnlock()
+	if s, ok := c.schedulers[schedulers.TransferWitnessLeaderName]; ok {
```
JmPotato, line 974: I suggest releasing the read lock as soon as possible here since the `select` below may have the possibility to block.

```suggestion
	c.RLock()
	scheduler, ok := c.schedulers[schedulers.TransferWitnessLeaderName];
	c.RUnlock()
```

server/cluster/coordinator.go
```diff
@@ -966,3 +967,17 @@ func (c *coordinator) getPausedSchedulerDelayUntil(name string) (int64, error) {
...
+		if schedulers.NeedTransferWitnessLeader(region) {
+			select {
```
JmPotato, line 976: Will this block for a long time? I'm afraid it may slow down the heartbeat process speed. Also, what will happen if the region number is enormous? Will this become a bottleneck?

server/schedulers/transfer_witness_leader.go
```diff
@@ -0,0 +1,148 @@
...
+	return TransferWitnessLeaderName
+}
+
+func (s *transferLeaderScheduler) GetType() string {
+	return TransferWitnessLeaderType
+}
+
+func (s *transferLeaderScheduler) IsScheduleAllowed(cluster schedule.Cluster) bool {
+	// TODO: make sure the restriction is reasonable
+	allowed := s.OpController.OperatorCount(operator.OpLeader) < cluster.GetOpts().GetLeaderScheduleLimit()
```
rleungx, line 77: It could be blocked if there are balance leader operators.

server/schedulers/transfer_witness_leader.go
```diff
@@ -0,0 +1,148 @@
...
+			return nil
+		}
+	})
+
+	schedule.RegisterScheduler(TransferWitnessLeaderType, func(opController *schedule.OperatorController, _ endpoint.ConfigStorage, _ schedule.ConfigDecoder) (schedule.Scheduler, error) {
+		return newTransferWitnessLeaderScheduler(opController), nil
+	})
+}
+
+type transferLeaderScheduler struct {
```
JmPotato, line 54:
```suggestion
type transferWitnessLeaderScheduler struct {
```

server/cluster/coordinator.go
ethercflow: Sure, fixed, thanks!

server/schedulers/transfer_witness_leader.go
ethercflow: Yes, wait for the scene test to observe, and then determine how to adjust, add a TODO

server/cluster/coordinator.go
ethercflow: Fixed, thanks!

server/schedulers/transfer_witness_leader.go
ethercflow: Fixed, thanks!

server/cluster/coordinator.go
ethercflow: For failover scenarios only, and the scene is filtered before sending to the channel. Therefore, there will not be a large number of regions. Even if the size of the channel is exceeded, it will be directly discarded without blocking.

server/cluster/coordinator.go
```diff
@@ -966,3 +967,18 @@ func (c *coordinator) getPausedSchedulerDelayUntil(name string) (int64, error) {
 	}
 	return s.GetDelayUntil(), nil
 }
+
+func (c *coordinator) CheckTransferWitnessLeader(region *core.RegionInfo) {
+	if schedulers.NeedTransferWitnessLeader(region) {
```
rleungx, line 972: I think we can move this function to core package.

server/cluster/coordinator.go
```diff
@@ -966,3 +967,18 @@ func (c *coordinator) getPausedSchedulerDelayUntil(name string) (int64, error) {
...
+		c.RLock()
+		s, ok := c.schedulers[schedulers.TransferWitnessLeaderName]
+		c.RUnlock()
+		if ok {
+			select {
+			case schedulers.RecvRegionInfo(s.Scheduler) <- region:
+			default:
+				log.Warn("transfer witness leader recv region channel full", zap.Uint64("drop region-id", region.GetID()))
```
rleungx, line 980:
```suggestion
				log.Warn("drop transfer witness leader due to recv region channel full", zap.Uint64("region-id", region.GetID()))
```

server/schedulers/transfer_witness_leader.go
```diff
@@ -0,0 +1,148 @@
...
+	// TransferWitnessLeaderName is transfer witness leader scheduler name.
+	TransferWitnessLeaderName = "transfer-witness-leader-scheduler"
+	// TransferWitnessLeaderType is transfer witness leader scheduler type.
+	TransferWitnessLeaderType = "transfer-witness-leader"
+	// TransferWitnessLeaderBatchSize is the number of operators to to transfer
+	// leaders by one scheduling
+	TransferWitnessLeaderBatchSize = 3
+	// TransferWitnessLeaderRecvMaxRegionSize is the max number of region can receive
+	// TODO: make it a reasonable value
+	TransferWitnessLeaderRecvMaxRegionSize = 1000
```
rleungx, line 39: These two constants don't need to be exposed.

rleungx (approved):
The rest LGTM.

server/schedulers/transfer_witness_leader.go
```diff
@@ -0,0 +1,148 @@
...
+	return operator.CreateTransferLeaderOperator(typ, cluster, region, region.GetLeader().GetStoreId(), target.GetID(), targetIDs, operator.OpLeader)
+}
+
+// NeedTransferWitnessLeader is used to judge if the region's leader is a witness
+func NeedTransferWitnessLeader(region *core.RegionInfo) bool {
+	if region == nil || region.GetLeader() == nil {
+		return false
+	}
+	return region.GetLeader().IsWitness
+}
```
JmPotato, line 143: Can we reuse the same function in `core/region.go`?

JmPotato (commented):
The rest LGTM.

server/schedulers/transfer_witness_leader.go
JmPotato: It seems that this function is not used.

JmPotato (approved):

PR #5600, opened 2022-10-13
operator: the operator timeout duation depends on all the step not separated

bufferflies:
Signed-off-by: bufferflies <1045931706@qq.com>

### What problem does this PR solve?
Relax the timeout duration limit for slow operator. 
If the duration of region heartbeat request is higher than 60s , the cluster can not be healthy. 

Issue Number: ref #5596

### What is changed and how does it work?
change the interface `timeout(start time.time,regionSize uint64)bool` to `timeout(regionSize)time.duration`.
1. calculate the max duation depends on the steps 
2. checkout timeout will only check the total duation not all the step

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

- Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

PR #5661, opened 2022-11-01
config: add new configuration to control operator timeout dynamic. 

### LINKED ISSUE #5596: The timeout should consider the transfer latency especially in big cluster
## Bug Report
In big cluster, the duration of region heartbeat is bigger than 10s. For example:
the region heartbeat qps for every store : 1w ops
there are 10w region in this store, so the latency is 10w/1w=10s
but some short operator step timeout limit is 10s, such as promote/demote/transfer leader, so the operator failed because the duration is bigger than 10s. 

some log:
```
[2022/10/13 00:52:23.560 +00:00] [INFO] [operator_controller.go:590] ["operator timeout"] [region-id=1820635949] [takes=1m20.047056718s] [operator="\"balance-region {mv peer: store [14057] to [14053]} (kind:region, region:1820635949(540, 1763), createAt:2022-10-13 00:51:03.513592708 +0000 UTC m=+461704.590481929, startAt:2022-10-13 00:51:03.513683079 +0000 UTC m=+461704.590572320, currentStep:1, size:16, steps:[add learner peer 1832839605 on store 14053, use joint consensus, promote learner peer 1832839605 on store 14053 to voter, demote voter peer 1820635950 on store 14057 to learner, leave joint state, promote learner peer 1832839605 on store 14053 to voter, demote voter peer 1820635950 on store 14057 to learner, remove peer on store 14057]) timeout\""] [additional-info="{\"sourceScore\":\"5538250.44\",\"targetScore\":\"5519343.11\"}"]
```
there are 5w regions in each store.

the operator step duration
image
the duration of region heartbeat 
image

operators are successful.
### What did you see instead?
operators are failed
master, v6.1.0

bufferflies:

Issue Number: Close #5596
based on: #5600

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has configuration change

Side effects

Related changes

### Release note

```release-note
 None.
```


server/api/diagnostic_test.go
```diff
@@ -82,17 +82,20 @@ func (suite *diagnosticTestSuite) TestSchedulerDiagnosticAPI() {
 
 	suite.NoError(tu.ReadGetJSON(re, testDialClient, addr, cfg))
 	suite.False(cfg.Schedule.EnableDiagnostic)
+	suite.Equal(float64(0), cfg.Schedule.OperatorTimeoutOffset.Seconds())
 
 	ms := map[string]interface{}{
-		"enable-diagnostic": "true",
-		"max-replicas":      1,
+		"enable-diagnostic":       "true",
+		"operator-timeout-offset": "10s",
```
lhy1024, line 89: Is it necessary to add new config? Maybe we can dynamic control offset by heartbeat interval

bufferflies: The duration is between the tikv send and the pd begin to handle the heartbeat request. The interval will be reset before tikv sends the request. so the interval has no relation of the handler duration.

server/schedule/operator/operator.go
```diff
@@ -54,11 +54,13 @@ type Operator struct {
 }
 
 // NewOperator creates a new operator.
-func NewOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, approximateSize int64, steps ...OpStep) *Operator {
+func NewOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, approximateSize int64,
+	offset time.Duration, steps ...OpStep) *Operator {
```
CabinfeverB, line 58: It seems `offset` not used?

bufferflies: good catch

CabinfeverB (approved):

server/config/config.go
```diff
@@ -803,6 +805,7 @@ const (
 	defaultSplitMergeInterval        = time.Hour
 	defaultSwitchWitnessInterval     = time.Hour
 	defaultEnableDiagnostic          = false
+	defaultOperatorTimeoutOffset     = 0 * time.Second
```
lhy1024, line 808: default close?

server/schedule/operator/operator.go
```diff
@@ -54,12 +54,14 @@ type Operator struct {
 }
 
 // NewOperator creates a new operator.
-func NewOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, approximateSize int64, steps ...OpStep) *Operator {
+func NewOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, approximateSize int64,
```
lhy1024, line 57: this function has so many params.

server/config/config.go
bufferflies: yes, It may be used in big cluster such region heartbeat bottleneck .

server/schedule/operator/operator.go
bufferflies: yes, it should use builder model to refactor it.

server/config/config.go
```diff
@@ -774,6 +774,8 @@ type ScheduleConfig struct {
 	// EnableDiagnostic is the the option to enable using diagnostic
 	EnableDiagnostic bool `toml:"enable-diagnostic" json:"enable-diagnostic,string"`
 
+	// OperatorTimeoutOffset is the offset of operator timeout, default is 0.
+	OperatorTimeoutOffset typeutil.Duration `toml:"operator-timeout-offset" json:"operator-timeout-offset"`
```
rleungx, line 778: It's a little bit hard to understand the purpose through the name.

bufferflies: yes, I think this configuration should not used in normal cluster.

bufferflies:
closed by feedback control

PR #5673, opened 2022-11-03
tools: use metrics to record statistics

rleungx:

Issue Number: Ref #5468.

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None
```


tools/pd-simulator/simulator/metrics.go
```diff
@@ -11,8 +11,35 @@ var (
...
+		prometheus.CounterOpts{
+			Namespace: "pd",
+			Subsystem: "schedule",
+			Name:      "snap_send_count",
+			Help:      "Counter of sending region snapshot",
+		}, []string{"store"})
 )
 
 func init() {
 	prometheus.MustRegister(snapDuration)
```
bufferflies, line 41: Do we need to add one grafana panel for the simulator later?

rleungx: Yes, I think we can

tools/pd-simulator/simulator/metrics.go
```diff
@@ -11,8 +11,35 @@ var (
...
+
+	snapRecvCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "pd",
+			Subsystem: "schedule",
+			Name:      "snap_recv_count",
+			Help:      "Counter of receiving region snapshot",
+		}, []string{"store"})
+
+	snapSendCounter = prometheus.NewCounterVec(
```
bufferflies, line 31: how about use one counter by add another attribute such as type.

bufferflies (approved):

PR #5670, opened 2022-11-02
tools: fix the inconsistency of available space

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

### What problem does this PR solve?

Issue Number: Ref #5468

### What is changed and how does it work?
Prevent the following situation from happening.

Screen Shot 2022-11-03 at 3 06 20 PM

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None
```

PR #5677, opened 2022-11-03
tools: support triggering an event through HTTP API

### LINKED ISSUE #5451: Make simulator more flexible
## Enhancement Task
We have supported starting the simulator by using the tiup. But when a case is finished, the process will exit. We can add an HTTP interface for checking the results instead of exiting directly. BTW, we can set up the TiKV or trigger the event through that interface.

rleungx:

Issue Number: Ref #5468, close #5451.

This PR supports the following functionality. And it can be reviewed until #5670 is merged.
1. random down a store through `curl <ip>:<status_port>/event\?event=down-node` 
2. scale out a new store through `curl <ip>:<status_port>/event\?event=add-node`

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None
```


tools/pd-simulator/simulator/cases/cases.go
```diff
@@ -61,37 +61,10 @@ type Case struct {
...
-	return a.id
-}
-
-// IDAllocator is used to alloc unique ID.
-var IDAllocator idAllocator
-
 // CaseMap is a mapping of the cases to the their corresponding initialize functions.
 var CaseMap = map[string]func() *Case{
 	"balance-leader":            newBalanceLeader,
 	"redundant-balance-region":  newRedundantBalanceRegion,
```
lilinghai, line 67: by the way ,why not balance region case registered?

tools/pd-simulator/simulator/cases/balance_region.go
```diff
@@ -36,7 +36,7 @@ func newRedundantBalanceRegion() *Case {
 
 	for i := 0; i < storeNum; i++ {
```
lilinghai, line 37: if we add/down-node the case checker should check all the stores balance , but it still use storeNum.
it may exists somewhere else.

lilinghai (approved):
LGTM

tools/pd-simulator/simulator/simutil/id.go
```diff
@@ -0,0 +1,25 @@
+package simutil
```
HuSharp, line 1: please add Copyright

HuSharp (approved):

PR #5681, opened 2022-11-04
Lower down the frequency of sending AwakenRegions.

### LINKED ISSUE #5680: Lower down the frequency of AwakenRegions.
## Enhancement Task
As the title shown, this issue is used to lower down the frequency from 30s (2 times of one full round of SlowScore ticks) to 10 min (2 times of the least time for reseting SlowScore from 100 to 1.)
The reason why we need it is that we should avoid redundant and too frequent AwakenRegions.

LykxSassinator:

Issue Number: Close #5680 

```commit-message
Lower down the frequency of AwakenRegions from 30s to 10min.
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Just default values on AwakenTimeInterval has been modified.

Side effects

- None

Related changes

- Need to cherry-pick to the release branch v6.4

### Release note

```release-note
None
```

### DIFF
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -33,7 +33,7 @@ const (
 	storePersistInterval = 5 * time.Minute
 	initialMinSpace      = 8 * units.GiB // 2^33=8GB
 	slowStoreThreshold   = 80
-	awakenStoreInterval  = 10 * time.Minute
+	awakenStoreInterval  = 10 * time.Minute // 2 * slowScoreRecoveryTime
 
 	// EngineKey is the label key used to indicate engine.
 	EngineKey = "engine"
--- a/server/core/store.go
+++ b/server/core/store.go
@@ -33,7 +33,7 @@ const (
 	storePersistInterval = 5 * time.Minute
 	initialMinSpace      = 8 * units.GiB // 2^33=8GB
 	slowStoreThreshold   = 80
-	awakenStoreInterval  = 30 * time.Second
+	awakenStoreInterval  = 10 * time.Minute
 
 	// EngineKey is the label key used to indicate engine.
 	EngineKey = "engine"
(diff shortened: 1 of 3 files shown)


LykxSassinator:
This PR needs to be cherry-picked to release branch `release-6.4`

server/core/store.go
```diff
@@ -33,7 +33,7 @@ const (
 	storePersistInterval = 5 * time.Minute
 	initialMinSpace      = 8 * units.GiB // 2^33=8GB
 	slowStoreThreshold   = 80
-	awakenStoreInterval  = 30 * time.Second
+	awakenStoreInterval  = 10 * time.Minute // 2 * slowScoreRecoveryTime
```
lhy1024, line 36: Do we need to test this config change?

LykxSassinator: > Do we need to test this config change?

Yes, ut `TestAwakenStore` has contained the related testing on it. And the e-2-e testing has proved that it's valid.

server/core/store.go
```diff
@@ -67,13 +67,12 @@ type StoreInfo struct {
 // NewStoreInfo creates StoreInfo with meta data.
 func NewStoreInfo(store *metapb.Store, opts ...StoreCreateOption) *StoreInfo {
 	storeInfo := &StoreInfo{
-		meta:           store,
-		storeStats:     newStoreStats(),
-		leaderWeight:   1.0,
-		regionWeight:   1.0,
-		limiter:        make(map[storelimit.Type]*storelimit.StoreLimit),
-		minResolvedTS:  0,
-		lastAwakenTime: time.Now(),
```
bufferflies, line 76: some used code need to fix.

rleungx, line 76: Have u tested when the cluster restarts?

LykxSassinator: Yep, when restarting PD, PD will do the checking to determine whether it should send AwakenRegions or not. And if there still existed slow stores, parsing from etcd, it will return slowStoreIDs in AwakenRegions to TiKV side.

bufferflies (approved):

PR #5685, opened 2022-11-08
api: Add the readiness endpoint for PD

### LINKED ISSUE #5658: Need readiness check endpoint for PD

When doing a rolling update for a PD cluster, we need to turn down a PD instance, then bring up a PD instance with the updated config, and repeat this process for all PD instances one at a time. Typically we use tidb-operator for automating this, and a PD instance is considered "Running" as long as the pod is up for some duration. However, it could be the PD instance is still picking up with the leader and receiving etcd raft logs, and cannot serve the request at the moment. If let's say we have a deployment of PDs that has 3 replicas, the first two pods are updated but not ready for serving new requests, then at this point the PD is unavailable since the remaining PD cannot write quorum (as the other two are still syncing etcd).

It would be helpful if PD instance can expose info (maybe via a restful endpoint) whether its internal data is synced with leader and ready for serving new requests.

An alternative approach in tidb-operator scenario is to add a init container in PD's pod spec that sleeps for certain time. Before sleep is done, the Pod will stuck in `Init` state so the rolling update process will wait til the sleep completes. Hopefully, during this time, the etcd data for that PD instance is synced.

The endpoint should be something new, and since old system is depending on it, we don't need to worry about the migration.

hanlins:
Added a readiness endpoint that exposes the information whether the corresponding etcd member of the PD is synced with the leader.

Issue Number: Close #5658

```commit-message
Add the readiness endpoint for PD
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has configuration change
- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Add a readiness check endpoint for checking etcd readiness status
```


nolouch:
I think once member down, the peer role still a follower, do not change to the learner.

hanlins:
> I think once member down, the peer role still a follower, do not change to the learner.

Hi @nolouch, thanks for the feedback. Added a check for the lag between the leader's committed index and the current server's applied index, if the lag is larger than 5000, then the readiness check fails. What about that?

cc @rleungx

hanlins:
Some updates on the issue. I've updated the response payload schema to include richer information. Set up a TiDB cluster using the patch and it seems the response we got from PD is expected:
```
/ # ./curl-amd64 -v 127.0.0.1:2379/pd/api/v1/ready
*   Trying 127.0.0.1:2379...
* Connected to 127.0.0.1 (127.0.0.1) port 2379 (#0)
> GET /pd/api/v1/ready HTTP/1.1
> Host: 127.0.0.1:2379
> User-Agent: curl/7.86.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: accept, content-type, authorization
< Access-Control-Allow-Methods: POST, GET, OPTIONS, PUT, DELETE
< Access-Control-Allow-Origin: *
< Content-Type: application/json; charset=UTF-8
< Date: Thu, 10 Nov 2022 21:23:57 GMT
< Content-Length: 70
<
{
  "is_learner": false,
  "curr_index": 610,
  "leader_index": 610
}
* Connection #0 to host 127.0.0.1 left intact
```
For my next steps, I'm going to work together with some of our users and see if it can resolve issues in their scenario, and will also update the test cases in parallel.

hanlins:
Added some basic tests and seems working:
```
➜  pd git:(topic/5658/readiness-endpoint) go test -v github.com/tikv/pd/server/api -run Etcd
=== RUN   TestEtcdTestSuite
...
=== RUN   TestEtcdTestSuite/TestEtcdReady
    etcd_test.go:58: response: '{IsLearner:false CurrIndex:28 LeaderIndex:28}'
    etcd_test.go:58: response: '{IsLearner:false CurrIndex:28 LeaderIndex:28}'
    etcd_test.go:58: response: '{IsLearner:false CurrIndex:29 LeaderIndex:29}'
...
[2022/11/18 20:42:13.501 +00:00] [INFO] [etcd.go:571] ["stopped serving peer traffic"] [address=127.0.0.1:33701]
[2022/11/18 20:42:13.501 +00:00] [INFO] [etcd.go:373] ["closed etcd server"] [name=pd1] [data-dir=/tmp/test_pd3505621251] [advertise-peer-urls="[http://127.0.0.1:33701]"] [advertise-client-urls="[http://127.0.0.1:44639]"]
[2022/11/18 20:42:13.502 +00:00] [INFO] [server.go:522] ["close server"]
--- PASS: TestEtcdTestSuite (13.55s)
    --- PASS: TestEtcdTestSuite/TestEtcdReady (0.16s)
PASS
ok  	github.com/tikv/pd/server/api	13.584s
```
Wondering do you have any concerns about the endpoint naming? The meaning for `/ready` could be a little vague, maybe we should rename it as `/etcd/ready`? Or just leave it as `/ready` for simplicity?

Issue #5692, opened 2022-11-09
Hot region consumes too much memory


rleungx:
## Enhancement Task
When there is a large number of regions, it will cause PD OOM sometimes.

<img width="1920" alt="Screen Shot 2022-11-09 at 3 56 20 PM" src="https://user-images.githubusercontent.com/35896542/200771357-498ed900-3f3b-4cd9-afc9-2590b3f6c2eb.png">

rleungx:
BTW, it also consumes lots of CPU.

lhy1024:
- [x] #5709
- [x] #5716
- [x] #5721
- [x] #5728
- [x] #5729
- [x] #5734
- [x] #5700
- [ ] #5749
- [ ] #5752

Before: v6.4 ea9b1e9
BenchmarkCheckRegionFlow-40        64240             19952 ns/op            5849 B/op         87 allocs/op
After: master 500b0d6
BenchmarkCheckRegionFlow-40        78693             15800 ns/op            4023 B/op         60 allocs/op

Before: v6.4 ea9b1e9
462 + 3340
![image](https://user-images.githubusercontent.com/19542290/205590586-706ac782-434a-4281-8513-760674e92ac9.png)

After: master 500b0d6
307 + 2891
![image](https://user-images.githubusercontent.com/19542290/205590640-a48feeb1-a9b8-4340-866e-db8c20bbc372.png)

PR #5700, opened 2022-11-10
scheduler: reduce memory used in median filter

lhy1024:

Issue Number: Ref #5692

When there are many stores and frequent flow region update, hot statistic and hot scheduler cost a lot of ram.

1. Reduce unnecessary clone in `summaryStoresLoadByEngine`, we don't need to clone all info in hot peer stat.
2. Reduce unnecessary calculate and slice in `MedianFilter`.

### Check List

Tests

- Unit test
- Integration test
```
tiup playground nightly --kv=1 --kv.binpath=bin/pd-heartbeat-bench --kv.config=tools/pd-heartbeat-bench/config-template.toml --pd=1 --pd.binpath=bin/pd-server --db=0 --host xx.xx.xx.xx
```
notes: tests should be the same duration

Before this pr
image

After this pr

image

### Release note

```release-note
None.
```


rleungx:
CI failed

pkg/movingaverage/median_filter.go
```diff
@@ -14,57 +14,77 @@
...
+	"sync"
+
+	"github.com/montanaflynn/stats"
+)
 
 // MedianFilter works as a median filter with specified window size.
 // There are at most `size` data points for calculating.
 // References: https://en.wikipedia.org/wiki/Median_filter.
 type MedianFilter struct {
+	sync.RWMutex
```
rleungx, line 27: Will it consume more resources?

lhy1024: avoid potential data race

rleungx:
I think we still need to add a benchmark.

server/statistics/hot_peer.go
```diff
@@ -182,13 +182,30 @@ func (stat *HotPeerStat) GetThresholds() []float64 {
 
 // Clone clones the HotPeerStat.
 func (stat *HotPeerStat) Clone() *HotPeerStat {
-	ret := *stat
```
HunDunDM, line 185: Why not use `ret := *stat`?

HunDunDM (commented):
rest LGTM

rleungx (approved):

CabinfeverB (approved):

PR #5716, opened 2022-11-17
statistics: reduce unnecessary clone and init in hot cache

lhy1024:
Signed-off-by: lhy1024 <admin@liudos.us>

### What problem does this PR solve?

Issue Number: Ref #5692

### What is changed and how does it work?

1. reduce unnecessary `newItem`
2. reduce unnecessary queue init

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

PR #5721, opened 2022-11-21
statistics: remove unnecessary field and cached thresholds

lhy1024:

Issue Number: Ref #5692

1. remove unnecessary field: Kind、Peer、source、lastUpdatedTime、isLearner
2. cached thresholds and updated in cache rather than every times calculate.
3. reduce unnecessary sleep in test.

### Check List

Tests

- Unit test
- Integration test
- Manual test

goos: linux
goarch: amd64
pkg: github.com/tikv/pd/server/statistics
cpu: Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz

this PR:
BenchmarkCheckRegionFlow-40        73910             13547 ns/op            4576 B/op         78 allocs/op
#5716:
BenchmarkCheckRegionFlow-40 75266 14751 ns/op 5705 B/op 78 allocs/op
6.4:
BenchmarkCheckRegionFlow-40 78902 15529 ns/op 5850 B/op 87 allocs/op

### Release note

```release-note
None.
```


pkg/cache/cache_test.go
```diff
@@ -28,10 +28,10 @@ func TestExpireRegionCache(t *testing.T) {
 	re := require.New(t)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	cache := NewIDTTL(ctx, time.Second, 2*time.Second)
+	cache := NewIDTTL(ctx, 10*time.Millisecond, 20*time.Millisecond)
```
bufferflies, line 31: how about use constant for it.

server/core/region_option.go
```diff
@@ -284,6 +284,13 @@ func SetReportInterval(v uint64) RegionCreateOption {
 	}
 }
 
+// SetInterval sets the report interval for the region.
```
bufferflies, line 287: Is this function can use test only?

server/statistics/util.go
```diff
@@ -34,3 +34,10 @@ const (
 func storeTag(id uint64) string {
 	return fmt.Sprintf("store-%d", id)
 }
+
+func hotStatReportInterval(kind RWType) int {
```
bufferflies, line 38: how about move this function to the RWtype sub function like this: `func(kind RWType) interval ()int{}`

pkg/cache/cache_test.go
lhy1024: I will split them

lhy1024: It is splited to #5727 #5728 #5729

server/statistics/hot_peer_cache.go
```diff
@@ -456,14 +454,15 @@ func (f *hotPeerCache) updateHotPeerStat(region *core.RegionInfo, newItem, oldIt
 		if newItem.inCold {
 			f.coldItem(newItem, oldItem)
 		} else {
+			thresholds := f.calcHotThresholds(newItem.StoreID)
 			if f.isOldColdPeer(oldItem, newItem.StoreID) {
```
bufferflies, line 458: The judgement is too difficult，maybe need to be refactored  it .

server/cluster/coordinator.go
```diff
@@ -484,16 +484,27 @@ func (c *coordinator) getHotRegionsByType(typ statistics.RWType) *statistics.Sto
...
-	return nil
+	updateRegionInfo := func(stores statistics.StoreHotPeersStat) {
+		for _, store := range stores {
+			for _, hotPeer := range store.Stats {
+				region := c.cluster.GetRegion(hotPeer.RegionID)
+				hotPeer.UpdateRegionInfo(region)
+			}
+		}
+	}
+	updateRegionInfo(infos.AsLeader)
```
bufferflies, line 505: It is strange that one function is call twice .

server/statistics/hot_peer_cache.go
lhy1024: It is not the content of this pr

bufferflies (approved):
the rest lgtm

server/statistics/hot_peer.go
```diff
@@ -75,32 +74,29 @@ func (d *dimStat) Clone() *dimStat {
...
+	// stores contains the all peer's storeID in this region.
+	stores []uint64
+	// actionType is the action type of the region, add, update or remove.
+	actionType ActionType
+	// isLeader is true means that the region has a leader on this store.
+	isLeader bool
+	// interval is the time interval that the region report interval
+	// For read peer, it's the store heartbeat interval
+	// For write peer, it's the region heartbeat interval
+	interval uint64
```
rleungx, line 94: Can we remove it?

server/statistics/hot_regions_stat.go
```diff
@@ -44,3 +46,12 @@ type HotPeerStatShow struct {
 	AntiCount      int       `json:"anti_count"`
 	LastUpdateTime time.Time `json:"last_update_time"`
 }
+
+// UpdateRegionInfo updates the region information, such as `IsLearner` and `LastUpdateTime`.
+func (h *HotPeerStatShow) UpdateRegionInfo(region *core.RegionInfo) {
```
rleungx, line 51: Actually, it updates the hot peer info according to region information?

lhy1024: yes, do you have a good idea about its name?

server/statistics/hot_peer.go
lhy1024: you are right

server/statistics/hot_regions_stat.go
rleungx: Maybe `UpdateHotPeerStat`?

rleungx (approved):

PR #5722, opened 2022-11-22
*: update subtree asynchronously

### LINKED ISSUE #5712: Process the subtrees asynchronously
## Enhancement Task
Currently, the region tree and the subtrees share the same lock. See
https://github.com/tikv/pd/blob/0e0f544a48bf79e75cd32c325f9e66aa11f8a625/server/core/region.go#L727-L736

We can use different locks and moreover, we can make the process of updating subtrees asynchronously which can help us improve the lock contention in processing the region heartbeat.

rleungx:

Issue Number: Close #5712.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


rleungx:
Close since it's stale.

PR #5728, opened 2022-11-23
statistics: cached thresholds in hot peer cache

lhy1024:

Issue Number: Ref #5692

It is a part of #5721

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/statistics/hot_peer_cache.go
```diff
@@ -60,6 +60,14 @@ var MinHotThresholds = [RegionStatCount]float64{
 	RegionWriteQueryNum: 32,
 }
 
+// DefaultThresholdsUpdateInterval is the default interval to update thresholds.
+var DefaultThresholdsUpdateInterval = StoreHeartBeatReportInterval * time.Second
```
bufferflies, line 64: 1. the `StoreHeartBeatReportInterval` can be adjusted by tikv config 
2. the refresh interval should be less than store hearbeat interval to keep the next calculate must use the latest threshold?

server/statistics/hot_peer_cache.go
```diff
@@ -294,19 +311,23 @@ func (f *hotPeerCache) getOldHotPeerStat(regionID, storeID uint64) *HotPeerStat
...
-	}
 	tn, ok := f.peersOfStore[storeID]
 	if !ok || tn.Len() < TopNN {
-		return ret
+		return f.defaultThresholds
 	}
-	for i := range ret {
-		ret[i] = math.Max(tn.GetTopNMin(i).(*HotPeerStat).GetLoad(i)*HotThresholdRatio, ret[i])
+	thresholds, ok := f.thresholdsOfStore[storeID]
+	if ok && time.Since(thresholds.updatedTime) <= DefaultThresholdsUpdateInterval {
```
rleungx, line 319: Will it affect write threshold?

lhy1024: Yes, but less interval means update more frequently, so it will only make write cache more accurate

server/statistics/hot_peer_cache.go
lhy1024: 1. I will make store heartbeat interval sync with tikv in another pr.
2. Good idea

server/statistics/hot_cache.go
```diff
@@ -202,3 +202,15 @@ func (w *HotCache) GetFilledPeriod(kind RWType) int {
 	}
 	return movingaverage.NewTimeMedian(DefaultAotSize, rollingWindowsSize, time.Duration(reportIntervalSecs)*time.Second).GetFilledPeriod()
 }
+
+// GetThresholds returns thresholds.
```
bufferflies, line 206:
```suggestion
// GetThresholds returns thresholds only for test
```

server/statistics/hot_peer_cache.go
```diff
@@ -69,16 +78,25 @@ type hotPeerCache struct {
...
-		peersOfStore:   make(map[uint64]*TopN),
-		storesOfRegion: make(map[uint64]map[uint64]struct{}),
-		regionsOfStore: make(map[uint64]map[uint64]struct{}),
-		taskQueue:      make(chan FlowItemTask, queueCap),
+		kind:              kind,
+		peersOfStore:      make(map[uint64]*TopN),
+		storesOfRegion:    make(map[uint64]map[uint64]struct{}),
+		regionsOfStore:    make(map[uint64]map[uint64]struct{}),
+		taskQueue:         make(chan FlowItemTask, queueCap),
+		defaultThresholds: defaultThresholds,
```
bufferflies, line 98: I think the default thresholds can remove, how about put this values into the thresholds

server/statistics/hot_peer_cache.go
```diff
@@ -69,16 +78,25 @@ type hotPeerCache struct {
...
+		thresholdsOfStore: make(map[uint64]*thresholdWithTime),
```
bufferflies, line 99: Can it be delete after deleting the store ?

server/statistics/hot_cache.go
lhy1024: L207

server/statistics/hot_peer_cache.go
```diff
@@ -60,6 +60,15 @@ var MinHotThresholds = [RegionStatCount]float64{
 	RegionWriteQueryNum: 32,
 }
 
+// DefaultThresholdsUpdateInterval is the default interval to update thresholds.
+// the refresh interval should be less than store heartbeat interval to keep the next calculate must use the latest threshold.
+var DefaultThresholdsUpdateInterval = StoreHeartBeatReportInterval * 0.8 * time.Second
```
rleungx, line 65: How about putting them into the same place with `StoreHeartBeatReportInterval` and 8s directly?

server/statistics/hot_peer_cache.go
```diff
@@ -60,6 +60,15 @@ var MinHotThresholds = [RegionStatCount]float64{
...
+
+type thresholdWithTime struct {
```
rleungx, line 67:
```suggestion
type thresholds struct {
```

server/statistics/hot_peer_cache.go
lhy1024: It cannot be const, because tests need.

rleungx (approved):

server/statistics/hot_peer_cache.go
```diff
@@ -69,24 +78,22 @@ type hotPeerCache struct {
...
-	} else {
-		c.reportIntervalSecs = ReadReportInterval
+	return &hotPeerCache{
+		kind:               kind,
+		peersOfStore:       make(map[uint64]*TopN),
+		storesOfRegion:     make(map[uint64]map[uint64]struct{}),
+		regionsOfStore:     make(map[uint64]map[uint64]struct{}),
+		taskQueue:          make(chan FlowItemTask, queueCap),
+		thresholdsOfStore:  make(map[uint64]*thresholds),
+		reportIntervalSecs: kind.ReportInterval(),
```
bufferflies, line 94: remove this.

bufferflies (commented):
rest LGTM

bufferflies (approved):

PR #5734, opened 2022-11-28
pkg: use another `Median`

rleungx:

Issue Number: Ref #5692.

```
name             old time/op    new time/op    delta
MedianFilter-40     618ns ± 8%     156ns ± 9%  -74.74%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
MedianFilter-40      104B ± 0%       80B ± 0%  -23.08%  (p=0.008 n=5+5)

name             old allocs/op  new allocs/op  delta
MedianFilter-40      2.00 ± 0%      1.00 ± 0%  -50.00%  (p=0.008 n=5+5)
```

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/movingaverage/median_filter_test.go
+++ b/pkg/movingaverage/median_filter_test.go
@@ -1,0 +1,32 @@
+// Copyright 2022 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


pkg/movingaverage/median_filter_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"testing"
+)
+
+func BenchmarkMedianFilter(b *testing.B) {
+	data := []float64{2, 1, 3, 4, 1, 1, 3, 3, 2, 0, 5}
+
+	mf := NewMedianFilter(10)
+	for _, n := range data {
```
lhy1024, line 25: Maybe we need to bench both add and get

lhy1024: In hot statistics and hot scheduler,get is more frequently than add

lhy1024: So I plan to calculate median in every add.cc @CabinfeverB

lhy1024 (approved):

nolouch (approved):

PR #5744, opened 2022-12-01
statistics: make `HotThresholdRatio` configurable

lhy1024:

Issue Number: Ref #5692

The current threshold for hotspot statistics, when the number of hot regions is too large, we will use the value of top60 as the threshold. This top60 will count the hot regions within three minutes.

For clusters with drastic load changes, hotspot statistics will consume a lot of memory. As a temporary means, we need to be able to configure the threshold of statistics
image

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mock/mockcluster/config.go
+++ b/pkg/mock/mockcluster/config.go
@@ -111,6 +111,11 @@ func (mc *Cluster) SetHotRegionCacheHitsThreshold(v int) {
 	mc.updateScheduleConfig(func(s *config.ScheduleConfig) { s.HotRegionCacheHitsThreshold = uint64(v) })
 }
 
+// SetHotThresholdRatio updates the HotThresholdRatio configuration.
+func (mc *Cluster) SetHotThresholdRatio(v int) {
+	mc.updateScheduleConfig(func(s *config.ScheduleConfig) { s.HotThresholdRatio = float64(v) })
+}
+
 // SetEnablePlacementRules updates the EnablePlacementRules configuration.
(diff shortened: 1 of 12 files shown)


pkg/mock/mockcluster/config.go
```diff
@@ -111,6 +111,11 @@ func (mc *Cluster) SetHotRegionCacheHitsThreshold(v int) {
 	mc.updateScheduleConfig(func(s *config.ScheduleConfig) { s.HotRegionCacheHitsThreshold = uint64(v) })
 }
 
+// SetHotThresholdRatio updates the HotThresholdRatio configuration.
+func (mc *Cluster) SetHotThresholdRatio(v int) {
```
rleungx, line 115: used?

server/config/config.go
```diff
@@ -776,6 +776,9 @@ type ScheduleConfig struct {
 
 	// EnableWitness is the option to enable using witness
 	EnableWitness bool `toml:"enable-witness" json:"enable-witness,string"`
+
+	// HotThresholdRatio is used to calculate hot thresholds
+	HotThresholdRatio float64 `toml:"hot-threshold-ratio" json:"hot-threshold-ratio,omitempty"`
```
rleungx, line 781: Do we need to validate the value of this config?

lhy1024: add non-negative check

lhy1024:
This PR won't be released in v6.5.0.

PR #5764, opened 2022-12-07
*: move all util relative package to utils

nolouch:

Issue Number: Ref #4399 

For the adjustment of code organization, for microservice-orient,  `pkg` package should be as clear as possible.

```commit-message
*: move all util relative package to utils
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


rleungx:
If we already put these functions under `utils/`, some package names might delete the suffix, e.g., `utils/logutil` -> `utils/log`.

nolouch:
@rleungx  I think we can keep the suffix for a short time, some reasons:
- the changes is too much
- the `package name` maybe more general than util, like `utils/api`, the package name is `api`, we may need to alias it in some scenario.

binshi-bing (commented):
LGTM

binshi-bing (commented):
read

binshi-bing (commented):
reviewed

binshi-bing (commented):
reviewed

PR #5765, opened 2022-12-08
scheduler: cache statistics to avoid frequent copy

lhy1024:

Issue Number: Ref #5692

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


rleungx:
Does this affect the hot region scheduling?

lhy1024:
> Does it affect the hot region scheduling?

Before this PR, hot scheduler gets hot statistics info by every schedule, usually 10ms. But in fact, we don’t need such high precision. 

I think 1 second is enough. In fact, the minimum time granularity of statistical information is also one second. Currently, the interval of store heartbeat and region heartbeat are also 10 seconds and 60 seconds.

HunDunDM (approved):

rleungx (approved):

Issue #5648, opened 2022-10-28
Improve the performance of handling region heartbeats


rleungx:
## Enhancement Task
A large cluster, whose number of regions is more than 1M, could have the performance bottleneck of region heartbeat. The issue is used to track the progress of improving the heartbeat.

Here are some improvements we might do in the future:
- [ ] #5586 
- [ ] #1488
- [ ] #5712
- [ ] #5707
- [ ] #3901 

Something we have done:
- [x] #5587 @rleungx (**-52%**)
- [x] #5628 @rleungx (~)
- [x] #5636 @rleungx (~)
- [x] #5649 @bufferflies (~)
- [x] #5693 @JmPotato (~)
- [x] #5652 @rleungx (**-16%**)
- [x] #5608 @rleungx (~)
- [x] #5720 @rleungx (**-37%**)
- [x] #5653 @rleungx 
- [x] #5615 @rleungx 
- [x] #5692 @lhy1024 (**-31%**)

rleungx:
Close it for now since most of jobs are finished.

PR #5615, opened 2022-10-21
*: reduce get overlap operations

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

### What problem does this PR solve?

Issue Number: Ref #5648, #5606.

### What is changed and how does it work?
```
name                                       old time/op    new time/op    delta
LoadRegionsByVolume/input_size_10000-40      0.05ns ±32%    0.04ns ±18%     ~     (p=0.095 n=5+5)
LoadRegionsByVolume/input_size_100000-40     0.58ns ± 2%    0.46ns ± 4%  -21.66%  (p=0.008 n=5+5)
LoadRegionsByVolume/input_size_1000000-40     6.29s ±11%     4.88s ± 4%  -22.34%  (p=0.008 n=5+5)

name                                       old alloc/op   new alloc/op   delta
LoadRegionsByVolume/input_size_10000-40       0.00B          0.00B          ~     (all equal)
LoadRegionsByVolume/input_size_100000-40      0.00B          0.00B          ~     (all equal)
LoadRegionsByVolume/input_size_1000000-40     934MB ± 0%     933MB ± 0%   -0.01%  (p=0.032 n=5+5)

name                                       old allocs/op  new allocs/op  delta
LoadRegionsByVolume/input_size_10000-40        0.00           0.00          ~     (all equal)
LoadRegionsByVolume/input_size_100000-40       0.00           0.00          ~     (all equal)
LoadRegionsByVolume/input_size_1000000-40     15.2M ± 0%     15.2M ± 0%     ~     (p=0.421 n=5+5)
```

```
name                                        old time/op    new time/op    delta
LoadRegionsByRandomMerge/merge_ratio_0-40      6.51s ± 4%     4.92s ±10%  -24.48%  (p=0.008 n=5+5)
LoadRegionsByRandomMerge/merge_ratio_20-40     8.06s ± 4%     4.74s ± 8%  -41.23%  (p=0.008 n=5+5)
LoadRegionsByRandomMerge/merge_ratio_40-40     9.74s ± 3%     4.93s ± 4%  -49.36%  (p=0.008 n=5+5)
LoadRegionsByRandomMerge/merge_ratio_60-40     11.1s ± 3%      4.9s ± 4%  -55.85%  (p=0.008 n=5+5)
LoadRegionsByRandomMerge/merge_ratio_80-40     11.9s ± 3%      5.1s ± 7%  -57.32%  (p=0.008 n=5+5)

name                                        old alloc/op   new alloc/op   delta
[...]

PR #5768, opened 2022-12-08
*: fix lock in `CheckAndPutRegion`

rleungx:

Issue Number: Ref #5648. 

The problem is the same as the above issue and was introduced by #5615. This problem only exists in the master, no need to cherry-pick.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/server/core/region.go
+++ b/server/core/region.go
@@ -759,6 +759,7 @@ func (r *RegionsInfo) CheckAndPutRegion(region *RegionInfo) []*RegionInfo {
 	if err != nil {
 		log.Debug("region is stale", zap.Stringer("origin", origin.GetMeta()), errs.ZapError(err))
 		// return the state region to delete.
+		r.t.Unlock()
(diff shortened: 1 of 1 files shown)


server/core/region.go
```diff
@@ -759,6 +759,7 @@ func (r *RegionsInfo) CheckAndPutRegion(region *RegionInfo) []*RegionInfo {
 	if err != nil {
 		log.Debug("region is stale", zap.Stringer("origin", origin.GetMeta()), errs.ZapError(err))
 		// return the state region to delete.
+		r.t.Unlock()
```
lhy1024, line 762: Do we need to cherry pick?

lhy1024 (approved):

server/core/region.go
rleungx: No, you can check the PR description.

Issue #5766, opened 2022-12-08
Split into multi-service architecture


rleungx:
### Describe your feature request related problem

Currently, PD has more and more management services. As TiDB clusters are becoming larger and larger, it is increasingly important to ensure the quality of service on the PD. In most large cluster cases, the main bottleneck of PD is CPU resources, and it is difficult to effectively control and isolate the CPU resources of multiple services in the same golang program. Therefore, we need to split these services into individual ones and deploy them independently for management, which is better to control the use and isolation of resources. Also, the microservice architecture is also very suitable for the cloud environment.

### Describe the feature you'd like

One of the most important services which PD provided is the TSO service. Deploy it through an independent binary will improve its QoS.

#### PreWorks
- [x] https://github.com/tikv/pd/issues/5838 @rleungx 
 - [x] Multi-Service Mode
    - [x] [**Introduce the service registry**](https://github.com/tikv/pd/pull/5779) @nolouch 
    - [x] https://github.com/tikv/pd/issues/5837 @lhy1024 
 - [x] Separate Services
    - [x] [**Provide meta storage interface**](https://github.com/tikv/pd/pull/6025) @rleungx
    - [x] https://github.com/tikv/pd/issues/5836 @binshi-bing
    - [x] https://github.com/tikv/pd/issues/5895 @binshi-bing
    - [x] https://github.com/tikv/pd/issues/5839 @rleungx
    - [x] https://github.com/tikv/pd/issues/5849 @rleungx 
 - [x] High availability 
    - [x] Each service has a high availability mechanism 
    - [x] Client-go discover the right service @lhy1024 
- [x] Compatibility
    - [x] Deploy Tools
        - [x] [TiUP Supports](https://github.com/pingcap/tiup/issues/2429) @rleungx 
        - [x] [Playground Supports](https://github.com/pingcap/tiup/issues/2429) @rleungx 
        - [x] [Operator Supports](https://github.com/tikv/pd/issues/7519) @HuSharp

nolouch:
PTAL @rleungx, The region Informer and storage interface maybe high priority, then the service can decoupling rely on the `pd.Server`.

PR #5773, opened 2022-12-12
*: move kv to pkg

rleungx:

Issue Number: Ref #5766.

This PR moves `kv` package from `server` to `pkg` so that tso logic won't rely on the server.

```commit-message
```

### Check List

Tests

- Unit test
### Release note

```release-note
None
```


lhy1024 (approved):

binshi-bing (commented):
reviewed

PR #5792, opened 2022-12-16
*: decouple the dependency between config and schedule

rleungx:

Issue Number: Ref #5838.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


server/schedule/config/config.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// IsSchedulerRegistered checks if the named scheduler type is registered.
+func IsSchedulerRegistered(name string) bool {
+	_, ok := schedulerMap.Load(name)
+	return ok
+}
+
+// Config is the interface that wraps the Config related method.
+type Config interface {
```
nolouch, line 31: Each scheduler also has it's own config, How to distinguish them?

rleungx: A config instance that implements these interfaces can be used to start the scheduling services. And this interface is used for compatibility. As for the scheduler config, it doesn't rely on the server package so it can be defined without implementing this interface.

nolouch: maybe we can reduce the functions.

nolouch (approved):
lgtm

server/config/config.go
```diff
@@ -995,11 +995,6 @@ func (c *ScheduleConfig) Validate() error {
 	if c.LeaderSchedulePolicy != "count" && c.LeaderSchedulePolicy != "size" {
 		return errors.Errorf("leader-schedule-policy %v is invalid", c.LeaderSchedulePolicy)
 	}
-	for _, scheduleConfig := range c.Schedulers {
```
lhy1024, line 998: Is it removed?

server/schedule/config/config.go
```diff
@@ -0,0 +1,95 @@
...
+	schedulerMap.Store(typ, struct{}{})
+}
+
+// IsSchedulerRegistered checks if the named scheduler type is registered.
+func IsSchedulerRegistered(name string) bool {
+	_, ok := schedulerMap.Load(name)
+	return ok
+}
+
+// Config is the interface that wraps the Config related method.
```
lhy1024, line 30:
```suggestion
// Config is the interface that wraps the Config related methods.
```

server/config/config.go
rleungx: Yes.

rleungx:
@binshi-bing

server/schedule/config/config.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// IsSchedulerRegistered checks if the named scheduler type is registered.
+func IsSchedulerRegistered(name string) bool {
+	_, ok := schedulerMap.Load(name)
+	return ok
+}
+
+// Config is the interface that wraps the Config related methods.
+type Config interface {
```
binshi-bing, line 31: I feel this part is hard to maintain and is error-prone in the future. Just wondering if we can reuse ScheduleConfig in server/config/config.go and move it to server/scheduler to remove dependency. It seems that we want to remove the dependency (on server/config) from server/scheduler, but I'm sure if we can let server/config depend on  server/scheduler/config. I mightn't fully understand our work here. I'll sync with you two after holiday. @rleungx @nolouch ^

binshi-bing (approved):
LGTM

HunDunDM (approved):

PR #5830, opened 2023-01-05
Support resource group watch

### LINKED ISSUE #5794: Watch the resource group changes
## Feature Request
ref  https://github.com/tikv/pd/issues/5794
Provide a way let the client can watch the configuration changes, then TiKV/Storage can updates the priority settings.

### Option

I prefer provide a common mechanism, that's client can watch different `key` or `keyprfix` in etcd, when it changes, the client can watch it. A basic implement in befor: https://github.com/tikv/pd/pull/4308

Note: 
- the config truth is the configuration stored in configuration server (Now is pd)
- the client(TiKV/TiFlash Nodes) watch the configuration and updates the changes in memory, affect the local config. (I suggest not rely on config file, the client only use it in memory)

HuSharp:

Issue Number: Close #5794 Ref https://github.com/tikv/tikv/pull/14022

```commit-message
Support resource group watch
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None
```


client/resourcemanager_client.go
```diff
@@ -124,12 +129,58 @@ func (c *client) DeleteResourceGroup(ctx context.Context, resourceGroupName stri
 	return resp.GetBody(), nil
 }
 
+// WatchResourceGroup watches resource groups changes.
+// It returns a stream of slices of resource groups.
+// The first message in stream contains all current resource groups,
+// all subsequent messages contains new events[PUT/DELETE] for all resource groups.
+func (c *client) WatchResourceGroup(ctx context.Context, revision int64) (chan []*rmpb.ResourceGroup, error) {
+	configChan, err := c.WatchGlobalConfig(ctx, groupSettingsPathPrefix, revision)
```
CabinfeverB, line 137: we need add PD rootPath, because we save resourceGroup by `storage()` which using `rootPath` prefix

server/grpc_service.go
```diff
@@ -1900,59 +1901,66 @@ func (s *GrpcServer) StoreGlobalConfig(_ context.Context, request *pdpb.StoreGlo
...
+func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, server pdpb.PD_WatchGlobalConfigServer) error {
 	ctx, cancel := context.WithCancel(s.Context())
 	defer cancel()
-	err := s.sendAllGlobalConfig(ctx, server)
-	if err != nil {
+	if err := s.sendAllGlobalConfig(ctx, server, req.GetConfigPath(), req.GetRevision()); err != nil {
 		return err
 	}
-	watchChan := s.client.Watch(ctx, globalConfigPath, clientv3.WithPrefix())
+	watchChan := s.client.Watch(ctx, req.GetConfigPath(), clientv3.WithPrefix(), clientv3.WithRev(req.GetRevision()))
```
nolouch, line 1935: Does the config path is rigth?

server/server.go
```diff
@@ -730,6 +730,11 @@ func (s *Server) GetHTTPClient() *http.Client {
 	return s.httpClient
 }
 
+// GetEtcdPath returns the etcd path.
+func (s *Server) GetEtcdPath(configPath string) string {
```
nolouch, line 734: it's pd config path. maybe `GetFinalPathWithinPD` or better one.

server/grpc_service.go
```diff
@@ -1900,59 +1901,66 @@ func (s *GrpcServer) StoreGlobalConfig(_ context.Context, request *pdpb.StoreGlo
...
-	// TODO: complete this function with new implementation
-	return &pdpb.LoadGlobalConfigResponse{}, nil
+	configPath := request.GetConfigPath()
+	r, err := s.client.Get(ctx, s.GetEtcdPath(configPath), clientv3.WithPrefix())
+	if err != nil {
+		item := &pdpb.GlobalConfigItem{Name: configPath, Error: &pdpb.Error{Type: pdpb.ErrorType_UNKNOWN, Message: err.Error()}}
+		return &pdpb.LoadGlobalConfigResponse{Items: []*pdpb.GlobalConfigItem{item}}, err
+	} else if len(r.Kvs) == 0 {
+		msg := "key " + configPath + " not found"
+		item := &pdpb.GlobalConfigItem{Name: configPath, Error: &pdpb.Error{Type: pdpb.ErrorType_GLOBAL_CONFIG_NOT_FOUND, Message: msg}}
```
glorv, line 1916: Why return this error if the config does not exist? It can happen in the normal logic.

HuSharp: You are right, we don't need to worry about loading null and just put all the error in func's error to return.

BornChanger:
image
Is lower code coverage a blocker issue ?

HuSharp:
> image
> 
> Is lower code coverage a blocker issue ?

This does not block merge

client/client.go
```diff
@@ -1821,18 +1821,69 @@ func trimHTTPPrefix(str string) string {
...
+		return err
+	}
 	return nil
 }
 
 func (c *client) WatchGlobalConfig(ctx context.Context, configPath string, revision int64) (chan []GlobalConfigItem, error) {
-	// TODO: complete this function with new implementation.
-	return nil, nil
+	// register watch components
+	globalConfigWatcherCh := make(chan []GlobalConfigItem, 16)
```
glorv, line 1852: Do we need to return the max revision of each []GlobalConfigItem so the caller can do incremental retry?

server/grpc_service.go
```diff
@@ -1900,59 +1901,74 @@ func (s *GrpcServer) StoreGlobalConfig(_ context.Context, request *pdpb.StoreGlo
...
+	// If the revision is compacted, will meet required revision has been compacted error.
+	// - If required revision < CompactRevision, we need to reload all configs to avoid losing data.
+	// - If required revision >= CompactRevision, just keep watching.
+	watchChan := s.client.Watch(ctx, s.GetFinalPathWithinPD(req.GetConfigPath()), clientv3.WithPrefix(), clientv3.WithRev(revision))
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
 		case res := <-watchChan:
+			if res.CompactRevision != 0 && req.GetRevision() < res.CompactRevision {
```
glorv, line 1939: This check is not correct after receiving the first batch response. We should maintain a minRevision and update it after receiving each response

server/grpc_service.go
```diff
@@ -1900,59 +1901,74 @@ func (s *GrpcServer) StoreGlobalConfig(_ context.Context, request *pdpb.StoreGlo
...
-// just reconnect to it.
-func (s *GrpcServer) WatchGlobalConfig(_ *pdpb.WatchGlobalConfigRequest, server pdpb.PD_WatchGlobalConfigServer) error {
+// or stopped by whatever reason, just reconnect to it.
+func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, server pdpb.PD_WatchGlobalConfigServer) error {
 	ctx, cancel := context.WithCancel(s.Context())
 	defer cancel()
-	err := s.sendAllGlobalConfig(ctx, server)
-	if err != nil {
+	revision := req.GetRevision()
+	if err := s.sendAllGlobalConfig(ctx, server, req.GetConfigPath(), revision); err != nil {
```
glorv, line 1927: Why we need this logic? WatchGlobalConfig with a revision v should mean to return all message that revision > v right?

server/grpc_service.go
```diff
@@ -1900,61 +1901,65 @@ func (s *GrpcServer) StoreGlobalConfig(_ context.Context, request *pdpb.StoreGlo
...
+			}
+			if res.CompactRevision != 0 && revision < res.CompactRevision {
+				if err := server.Send(&pdpb.WatchGlobalConfigResponse{
+					Revision: res.CompactRevision,
+					Header: s.wrapErrorToHeader(pdpb.ErrorType_DATA_COMPACTED,
+						fmt.Sprintf("required watch revision: %d is smaller than current compact/min revision. %d", revision, res.CompactRevision)),
+				}); err != nil {
+					return err
+				}
+			}
```
glorv, line 1948:
```suggestion
			if revision > res.CompactRevision {
				if err := server.Send(&pdpb.WatchGlobalConfigResponse{
					Revision: res.CompactRevision,
					Header: s.wrapErrorToHeader(pdpb.ErrorType_DATA_COMPACTED,
						fmt.Sprintf("required watch revision: %d is smaller than current compact/min revision. %d", revision, res.CompactRevision)),
				}); err != nil {
					return err
				}
			}
			revision = res.Header.GetRevision()
```

client/resourcemanager_client.go
```diff
@@ -124,12 +128,56 @@ func (c *client) DeleteResourceGroup(ctx context.Context, resourceGroupName stri
...
+					close(resourceGroupWatcherChan)
+					return
+				}
+				groups := make([]*rmpb.ResourceGroup, 0, len(res))
+				for _, item := range res {
+					switch item.EventType {
+					case pdpb.EventType_PUT:
+						group := &rmpb.ResourceGroup{}
+						if err := proto.Unmarshal([]byte(item.Value), group); err != nil {
+							return
```
glorv, line 161: missing close(resourceGroupWatcherChan)

client/resourcemanager_client.go
```diff
@@ -124,12 +128,56 @@ func (c *client) DeleteResourceGroup(ctx context.Context, resourceGroupName stri
 	return resp.GetBody(), nil
 }
 
+// WatchResourceGroup [just for TEST] watches resource groups changes.
+// It returns a stream of slices of resource groups.
+// The first message in stream contains all current resource groups,
+// all subsequent messages contains new events[PUT/DELETE] for all resource groups.
+func (c *client) WatchResourceGroup(ctx context.Context, revision int64) (chan []*rmpb.ResourceGroup, error) {
+	configChan, err := c.WatchGlobalConfig(ctx, groupSettingsPathPrefix, revision)
```
glorv, line 136: I suggest add checking err != nil here to error return early

server/grpc_service.go
HuSharp: Why does watch revision > CompactRevision will return error? I think we can keep watching?

glorv: oo, my mistake. So I think the `res.CompactRevision != 0` is redundant and should update revision after this check.

client/client.go
HuSharp: since resource client is just for testing, I will add a retry mechanism after I solve pd client/tidb/CDC compatibility...

HuSharp:
@glorv @nolouch @BornChanger @CabinfeverB pls take a look, thx!

client/client.go
```diff
@@ -1821,18 +1821,70 @@ func trimHTTPPrefix(str string) string {
...
+				m, err := res.Recv()
+				if err != nil {
+					return
+				}
+				arr := make([]GlobalConfigItem, len(m.Changes))
+				for j, i := range m.Changes {
+					arr[j] = GlobalConfigItem{i.GetKind(), i.GetName(), i.GetValue()}
+				}
+				globalConfigWatcherCh <- arr
+			}
```
glorv, line 1884:
```suggestion
				m, err := res.Recv()
				if err != nil {
					return
				}
				arr := make([]GlobalConfigItem, len(m.Changes))
				for j, i := range m.Changes {
					arr[j] = GlobalConfigItem{i.GetKind(), i.GetName(), i.GetValue()}
				}
			select {
			case <-ctx.Done():
				return
			case globalConfigWatcherCh <- arr:
			}
```

glorv (approved):
LGTM

nolouch (approved):

CabinfeverB (approved):

Issue #5825, opened 2023-01-04
Introduce the informer 


rleungx:
## Development Task
We are going to split PD from a single binary into multiple services, e.g., API service, TSO service, and scheduling service. Some services need the meta info such as regions and stores. So we need to provide a uniform way to synchronize the different info among the different services.
The informer is used to do the job. Here are the sub-tasks to implement the informer:
- [ ] https://github.com/pingcap/kvproto/pull/1034 @rleungx 
- [ ] #5831 @rleungx 
- [ ] #5846 @rleungx 
- [ ] #5829 @rleungx 
- [ ] #5820 @rleungx 
- [ ] #5832 @rleungx

rleungx:
Please use a new issue to track if needed

PR #5846, opened 2023-01-10
*: support list and watch

rleungx:

Issue Number: Ref #5825.

This PR adds `List` and `Watch` methods for the server and adds a new way to synchronize regions among the PD servers. The main changes of this PR are listed as follows:
- implement the `List` and `Watch` in `grpc_service.go`
- add `StartWatchLeader` on the client side, which plays the same role as the original `StartSyncWithLeader`.
- add `RunWatchServer` on the server side, which plays the same role as the original `RunServer`
- split the original sync process into two phases: list and watch. If the local history index is 0, it will first call `listRegion` to get all regions, otherwise, it will only call `watchRegion` with a given history index

```commit-message
```

### Check List

Tests

- Unit test (TODO)

### Release note

```release-note
None
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -2073,3 +2073,50 @@ func (s *GrpcServer) GetExternalTimestamp(ctx context.Context, request *pdpb.Get
 		Timestamp: timestamp,
 	}, nil
 }
+
+// List implements gRPC PDServer.
+func (s *GrpcServer) List(request *pdpb.ListRequest, stream pdpb.PD_ListServer) error {
+	if s.IsClosed() || s.cluster == nil {
+		return ErrNotStarted
+	}
+	ctx := s.cluster.Context()
+	if ctx == nil {
+		return ErrNotStarted
+	}
+	clusterID := request.GetHeader().GetClusterId()
+	if clusterID != s.ClusterID() {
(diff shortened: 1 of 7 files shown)


server/grpc_service.go
```diff
@@ -2073,3 +2073,50 @@ func (s *GrpcServer) GetExternalTimestamp(ctx context.Context, request *pdpb.Get
...
+	ctx := s.cluster.Context()
+	if ctx == nil {
+		return ErrNotStarted
+	}
+	clusterID := request.GetHeader().GetClusterId()
+	if clusterID != s.ClusterID() {
+		return status.Errorf(codes.FailedPrecondition, "mismatch cluster id, need %d but got %d", s.ClusterID(), clusterID)
+	}
+
+	if request.GetResource() == "regionstat" {
```
lhy1024, line 2091: how about using `regionStatResource`

server/grpc_service.go
```diff
@@ -2073,3 +2073,50 @@ func (s *GrpcServer) GetExternalTimestamp(ctx context.Context, request *pdpb.Get
...
+		syncer := s.cluster.GetRegionSyncer()
+		start := time.Now()
+		if err := syncer.FullSync(ctx, stream); err != nil {
```
lhy1024, line 2094: what will we do when timeout?

Issue #5849, opened 2023-01-10
Provide API service interfaces


rleungx:
## Development Task
Before we separate the other services, we need to change the original internal calls to RPC calls.
- [x] #5874 @rleungx 
- [x] #5911 @rleungx 
- [x] #6025 @rleungx

PR #5852, opened 2023-01-11
*: support tso gRPC interfaces

rleungx:

Issue Number: Ref #5849.

This PR implements `Load/SaveTimestamp` which will be used by the TSO service to persist the timestamp. Most of the change in this PR is `go.sum`, which is due to the upgrade of `google.golang.org/grpc`.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


rleungx:
@binshi-bing PTAL

client/client.go
```diff
@@ -139,6 +139,8 @@ type Client interface {
 	KeyspaceClient
 	// ResourceManagerClient manages resource group metadata and token assignment.
 	ResourceManagerClient
+	// TsoClient manages TSO data.
+	TsoClient
```
JmPotato, line 143: Maybe it's arguable to say whether we should use `TSO` rather than `Tso` 👀.

server/tso_service.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	client := s.GetClient()
+	if client == nil {
+		return &tsopb.SaveTimestampResponse{Header: s.notBootstrappedHeader()}, nil
+	}
+	key := request.GetKey()
+	data := typeutil.Uint64ToBytes(request.GetTimestamp())
+	txn := kv.NewSlowLogTxn(client)
+	resp, err := txn.
+		Then(clientv3.OpPut(key, string(data))).
```
JmPotato, line 88: Do we need to compare the old one before putting in a new TSO to prevent it from fallback?

rleungx: IMO, this part of logic will be done by the TSO service itself. The API service only needs to provide the storage interface.

JmPotato: Got it. Make sense.

JmPotato: I'm wondering whether there will be a case in that different TSO services update the same key at the same time without knowing each other, maybe we should provide an option in the request to determine whether the check should be done. Otherwise, the services need to maintain some kind of mechanism to make sure consistency, which has a higher cost. On this hand, providing a simple check could save a lot of work. Since it's used to provide the TSO service, it makes sense to have such a customized change.

rleungx: make sense

lhy1024:
Do we need to move `externalTS`?

rleungx:
> Do we need to move `externalTS`?

Maybe. But this PR won't do it.

rleungx:
Closed since we are not going to wrap etcd API on the API service side.

Issue #5851, opened 2023-01-11
Tracking issue for resource manager


CabinfeverB:
## Development Task

Ref https://github.com/pingcap/tidb/issues/38825

- [x] #5785
- [x] #5809
- [x] #5810
- [x] #5811
- [x] #5862
- [x] #5830
- [x] https://github.com/tikv/pd/issues/5947

PR #5867, opened 2023-01-15
resouce_manager: fix the bug of loading resource group and using copy

CabinfeverB:

Issue Number: ref #5851 

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/server/manager.go
```diff
@@ -109,8 +112,8 @@ func (m *Manager) DeleteResourceGroup(name string) error {
 	return nil
 }
 
-// GetResourceGroup returns a copy of a resource group.
-func (m *Manager) GetResourceGroup(name string) *ResourceGroup {
+// GetResourceDuplicateGroup returns a copy of a resource group.
+func (m *Manager) GetResourceDuplicateGroup(name string) *ResourceGroup {
```
JmPotato, line 116: `GetResourceDuplicateGroup` sounds like it will get all duplicated groups, which is confusing.

```suggestion
func (m *Manager) GetResourceGroupCopy(name string) *ResourceGroup {
```

tests/msc/resource_manager/resource_manager_test.go
```diff
@@ -137,6 +137,18 @@ func (suite *resourceManagerClientTestSuite) TestAcquireTokenBucket() {
...
+	re.Less(gresp.RUSettings.RRU.Tokens, groups[0].RUSettings.RRU.Tokens)
+
+	// to test persistent
+	leaderName := suite.cluster.WaitLeader()
+	leader := suite.cluster.GetServer(leaderName)
+	leader.Stop()
+	suite.cluster.RunServers([]*tests.TestServer{leader})
+	_, err = cli.GetResourceGroup(suite.ctx, groups[0].GetName())
+	re.NoError(err)
+
```
JmPotato, line 151: Do we need to compare this one with the previous `gresp` to make sure the content is correct?

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -45,11 +47,12 @@ func NewManager(srv *server.Server) *Manager {
...
-		if err := json.Unmarshal([]byte(v), &group); err != nil {
+		group := &rmpb.ResourceGroup{}
+		if err := proto.Unmarshal([]byte(v), group); err != nil {
+			log.Error("err", zap.Error(err), zap.String("k", k), zap.String("v", v))
 			panic(err)
 		}
-		m.groups[group.Name] = &group
+		m.groups[group.Name] = FromProtoResourceGroup(group)
 	}
 	m.storage().LoadResourceGroups(handler)
```
HuSharp, line 57: for loading groups, do we need to change `loadRangeByPrefix(resourceGroupPath, f)` to `loadRangeByPrefix(resourceGroupPath+"/settings", f)`🤔

nolouch (approved):
lgtm

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -119,6 +122,16 @@ func (m *Manager) GetResourceGroup(name string) *ResourceGroup {
 	return nil
 }
 
+// GetResourceGroup returns a resource group.
+func (m *Manager) GetResourceGroup(name string) *ResourceGroup {
```
nolouch, line 126: maybe default use copy, and new function like `GetMutableResourceGroup`.

pkg/storage/endpoint/resource_group.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // ResourceGroupStorage defines the storage operations on the rule.
 type ResourceGroupStorage interface {
-	LoadResourceGroups(f func(k, v string)) error
+	LoadResourceGroups(prefix string, f func(k, v string)) error
```
JmPotato, line 25: It seems that this prefix remains the same all the time, can we make it internal rather than passing it all around?

CabinfeverB: In addition to setting, we may load other state information in the future.  But maybe we can change the name. Do you have any suggestions?

JmPotato: We could name it `LoadResourceGroupsSetting` to prevent exposing the prefix. If we want to load other info in the future, we could also create a new function like `LoadResourceGroupsXXX`.

HuSharp (approved):
LGTM!

nolouch:
the test failed.

pkg/storage/endpoint/resource_group.go
```diff
@@ -20,27 +20,31 @@ import (
 	"github.com/gogo/protobuf/proto"
 )
 
+const (
+	// groupSettingsPathPrefix is the prefix of the resource group path to store group settings.
+	groupSettingsPathPrefix = "/settings"
```
JmPotato, line 25: Put it to `pkg/storage/endpoint/key_path.go` please.

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

PR #5873, opened 2023-01-18
schedulers: split transfer-witness-leader's limit from balance-leader

ethercflow:
ref tikv/pd#5638

Issue Number: ref #5638

Separate the limit configuration of `transfer-witness-leader` and `balance-leader`. Otherwise, when the witness becomes the leader, the op could not be created in time because the quota was full. The client will return a backoff error: `is_witness`.

```commit-message
schedulers: split transfer-witness-leader's limit from balance-leader
```

### Check List

Tests

Code changes

- Has configuration change

Side effects

- None

Related changes

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/operator/kind.go
+++ b/server/schedule/operator/kind.go
@@ -42,32 +42,36 @@ const (
 	OpRegion
 	// Include leader transfer.
 	OpLeader
+	// Include witness leader transfer.
+	OpWitnessLeader
 	// Include witness transfer.
 	OpWitness
 	opMax
 )
 
 var flagToName = map[OpKind]string{
-	OpLeader:    "leader",
-	OpRegion:    "region",
-	OpSplit:     "split",
-	OpAdmin:     "admin",
-	OpHotRegion: "hot-region",
-	OpReplica:   "replica",
-	OpMerge:     "merge",
-	OpRange:     "range",
-	OpWitness:   "witness",
+	OpLeader:        "leader",
+	OpRegion:        "region",
+	OpSplit:         "split",
+	OpAdmin:         "admin",
+	OpHotRegion:     "hot-region",
+	OpReplica:       "replica",
+	OpMerge:         "merge",
+	OpRange:         "range",
+	OpWitness:       "witness",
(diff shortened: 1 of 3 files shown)


server/schedule/operator/kind.go
```diff
@@ -42,32 +42,36 @@ const (
 	OpRegion
 	// Include leader transfer.
 	OpLeader
+	// Include witness leader transfer.
+	OpWitnessLeader
```
bufferflies, line 46: where is it used?

disksing, line 46: not used?

ethercflow: Fixed, sorry, the change was in `git stashed`...

ethercflow: Fixed, sorry, the change was in `git stashed`...

server/schedulers/transfer_witness_leader.go
```diff
@@ -80,12 +80,7 @@ func (s *trasferWitnessLeaderScheduler) GetType() string {
 }
 
 func (s *trasferWitnessLeaderScheduler) IsScheduleAllowed(cluster schedule.Cluster) bool {
-	// TODO: make sure the restriction is reasonable
-	allowed := s.OpController.OperatorCount(operator.OpLeader) < cluster.GetOpts().GetLeaderScheduleLimit()
-	if !allowed {
-		operator.OperatorLimitCounter.WithLabelValues(s.GetType(), operator.OpLeader.String()).Inc()
-	}
-	return allowed
+	return true
```
bufferflies, line 83: no any limit?

ethercflow: Because when the witness becomes the leader, it cannot provide read and write requests, which will trigger the client to enter backoff to retry. If the transfer cannot be made in time, the client will time out. In addition, the strategy of witness raft log gc has been changed to ensure that there are enough logs and snapshots will not be generated during the transfer leader process.

bufferflies (approved):

rleungx (approved):

Issue #5864, opened 2023-01-13
Reduce memory usage of Placement Rules Cache


HunDunDM:
## Development Task

When the Placement Rules Cache is enabled, if the number of Regions is large, a very large amount of memory will be used. 

Reduce memory usage to reduce the risk of PD OOM when this feature is enabled.

HunDunDM:
ref #5473

PR #5879, opened 2023-01-20
schedule: placement rules cache reduce cache low usage fit

HunDunDM:

Issue Number: Ref #5864

`bestFit` uses more memory, but it is rarely used.

part of #5860

Only cache bestFits that have been used more than 10 times.

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


lhy1024 (approved):
Maybe we need more testing on the real cluster or simulater, but given that cache is not on by default, I think we can merge first

nolouch (approved):
lgtm

PR #5881, opened 2023-01-29
resource_manager: supports burstable mode

nolouch:

Issue Number: ref #4399 ref https://github.com/pingcap/tidb/issues/40380

```commit-message
resource_manager: supports burstable mode
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/mcs/resource_manager/client/client.go
+++ b/pkg/mcs/resource_manager/client/client.go
@@ -523,11 +523,11 @@ func (gc *groupCostController) handleTokenBucketTrickEvent(ctx context.Context) 
 func (gc *groupCostController) updateAvgRaWResourcePerSec() {
 	isBurstable := true
 	for typ, counter := range gc.run.resourceTokens {
-		if !gc.calcAvg(counter, getRawResourceValueFromConsumption(gc.run.consumption, typ)) {
-			continue
-		}
 		if counter.limiter.GetBurst() >= 0 {
 			isBurstable = false
+		}
+		if !gc.calcAvg(counter, getRawResourceValueFromConsumption(gc.run.consumption, typ)) {
+			continue
 		}
 		log.Debug("[resource group controller] update avg raw resource per sec", zap.String("name", gc.Name), zap.String("type", rmpb.RawResourceType_name[int32(typ)]), zap.Float64("avgRUPerSec", counter.avgRUPerSec))
 	}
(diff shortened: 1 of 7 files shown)


pkg/mcs/resource_manager/client/client.go
```diff
@@ -351,6 +352,9 @@ type groupCostController struct {
 		consumption *rmpb.Consumption
 	}
 
+	// fast path to make once token limit with un-limit burst.
+	burstable *atomic.Bool
```
JmPotato, line 356: It seems that the standard library has an `atomic.Bool` type, can we use that one?

https://pkg.go.dev/sync/atomic#Bool

nolouch: addressed

pkg/mcs/resource_manager/client/client.go
```diff
@@ -764,7 +784,9 @@ func (gc *groupCostController) onResponse(req RequestInfo, resp ResponseInfo) {
 	for _, calc := range gc.calculators {
 		calc.AfterKVRequest(delta, req, resp)
 	}
-
+	if gc.burstable.Load() {
+		return
```
CabinfeverB, line 788: we can goto `ret` like handle in `onRequestWait`

CabinfeverB (commented):
rest LGTM

pkg/mcs/resource_manager/client/client.go
```diff
@@ -516,21 +521,31 @@ func (gc *groupCostController) handleTokenBucketTrickEvent(ctx context.Context)
...
+		}
 		log.Debug("[resource group controller] update avg raw resource per sec", zap.String("name", gc.Name), zap.String("type", rmpb.RawResourceType_name[int32(typ)]), zap.Float64("avgRUPerSec", counter.avgRUPerSec))
 	}
+	gc.burstable.Store(isBurstable)
 }
 
 func (gc *groupCostController) updateAvgRUPerSec() {
+	isBurstable := true
 	for typ, counter := range gc.run.requestUnitTokens {
+		if counter.limiter.GetBurst() >= 0 {
```
JmPotato, line 540: Should we place this check before or after the below check? I noticed that its place is different with `updateAvgRaWResourcePerSec`.

JmPotato (commented):
The rest LGTM.

pkg/mcs/resource_manager/client/client.go
nolouch: Good Catch, it's need before the below check.

CabinfeverB (approved):

nolouch:
ptal @JmPotato

JmPotato (approved):

PR #5888, opened 2023-01-31
resource_manager: unify RRU and WRU into RU for token limit

CabinfeverB:

Issue Number: ref #5851 

If resource group has enough tokens in RU mode, it can consume these tokens for write or read with freedom. So we should provider a unified token limiter for WRU and RRU.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


BornChanger (approved):
lgtm

client/go.mod
```diff
@@ -15,3 +15,5 @@ require (
 	go.uber.org/zap v1.20.0
 	google.golang.org/grpc v1.51.0
 )
+
+replace github.com/pingcap/kvproto => github.com/CabinfeverB/kvproto v0.0.0-20230130100540-c8385fbaf594
```
BornChanger, line 19: please update with new kvproto master

pkg/mcs/resource_manager/client/config.go
```diff
@@ -21,11 +21,10 @@ import (
 )
 
 var (
-	requestUnitList map[rmpb.RequestUnitType]struct{} = map[rmpb.RequestUnitType]struct{}{
-		rmpb.RequestUnitType_RRU: {},
-		rmpb.RequestUnitType_WRU: {},
+	requestUnitLimitTypeList map[rmpb.RequestUnitType]struct{} = map[rmpb.RequestUnitType]struct{}{
```
HuSharp, line 24: Do we still need this `map` now? If not we can remove `for range` where it is called.

JmPotato: Since the proto keeps `RequestUnitType` (though it only has 1 type), so I think it's reasonable to keep a list here also.

pkg/mcs/resource_manager/client/model.go
```diff
@@ -74,7 +74,8 @@ func (kc *KVCalculator) BeforeKVRequest(consumption *rmpb.Consumption, req Reque
 		// Write bytes are knowable in advance, so we can calculate the WRU cost here.
 		writeBytes := float64(req.WriteBytes())
 		consumption.WriteBytes += writeBytes
-		consumption.WRU += float64(kc.WriteBaseCost) + float64(kc.WriteBytesCost)*writeBytes
+		wru := float64(kc.WriteBaseCost) + float64(kc.WriteBytesCost)*writeBytes
```
JmPotato, line 77: Why create a new variable?

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

nolouch (approved):

Issue #5627, opened 2022-10-24
Handle voter down with witness 


ethercflow:
## Development Task
- Rule checker 
  - When witness placement rule is enabled
    - Promotes the witness to voter when region has one down peer
    - Promotes the witness to voter when region has lagging behind peer
    - Replace the witness with witness when offline a store
  - No matter whether witness placement rule is enabled or disabled (to improve availability)
    - When peer's downtime exceeds the threshold(30min), add a witness and remove the down peer. Then witness is promoted to non-witness gradually.

Refer to https://github.com/tikv/tikv/issues/12876

PR #5896, opened 2023-02-02
rule_checker: promote witness when has pending voter

ethercflow:
ref tikv/pd#5627

Issue Number: ref tikv/pd#5627

```commit-message
When there's pending voter, we should promote the witness to the voter to avoid accumulating too many logs leading to disk full.
```

### Check List

Tests

- None

Code changes

- None

Side effects

- None

Related changes

- None

### Release note

```release-note
None
```


Connor1996 (approved):
LGTM

server/schedule/checker/rule_checker.go
```diff
@@ -216,6 +216,13 @@ func (c *RuleChecker) fixRulePeer(region *core.RegionInfo, fit *placement.Region
 			ruleCheckerReplaceOfflineCounter.Inc()
 			return c.replaceUnexpectRulePeer(region, rf, fit, peer, offlineStatus)
 		}
+
+		if c.isWitnessEnabled() && c.isPendingVoter(region, peer) {
+			if witness, ok := c.hasAvailableWitness(region, peer); ok {
+				ruleCheckerPromoteWitnessCounter.Inc()
+				return operator.CreateNonWitnessPeerOperator("promote-witness", c.cluster, region, witness)
```
rleungx, line 223: How about telling apart fixing pending voter and fixing down peer since the operator are both "promote-witness"?

rleungx (approved):

PR #5891, opened 2023-01-31
resource_manager/client: fix the infinite loop issue

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve?

Issue Number: ref #4399 ref https://github.com/pingcap/tidb/issues/40876 

### What is changed and how does it work?

```commit-message
resource_manager/client: fix the infinite loop issue
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
image

Code changes

- Has configuration change
- Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Possible performance regression
- Increased code complexity
- Breaking backward compatibility

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- Need to cherry-pick to the release branch

### Release note

```release-note
None
```

PR #5900, opened 2023-02-02
resource_manager, server: implement the resource manager failover

JmPotato:

Issue Number: ref #5851.

```commit-message
- Refine the resource manager initialization to support failover via PD election.
- Implement the client failover.
- Fix a race bug.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/client/client.go
```diff
@@ -535,7 +535,6 @@ func (gc *groupCostController) handleTokenBucketTrickEvent(ctx context.Context,
 				threshold := counter.notify.setupNotificationThreshold
 				counter.notify.mu.Unlock()
 				counter.limiter.SetupNotificationThreshold(now, threshold)
-				gc.updateRunState(ctx)
```
BornChanger, line 538: Why removed this line?

JmPotato: https://github.com/tikv/pd/pull/5891 introduced this race bug which will have the WAW problem with other `updateRunState` of `mainLoop`.

pkg/mcs/registry/registry.go
```diff
@@ -44,26 +44,18 @@ type RegistrableService interface {
 // It implements the `ServiceRegistry` interface.
 type ServiceRegistry struct {
 	builders map[string]ServiceBuilder
-	services map[string]RegistrableService
```
JmPotato, line 47: Remove this field otherwise the local integration test will "share" the same builder for the whole cluster, which will cause undefined behavior.

client/resourcemanager_client.go
```diff
@@ -248,35 +268,51 @@ func (c *client) createTokenDispatcher() {
...
 	var connection resourceManagerConnectionContext
 	if err := c.tryResourceManagerConnect(dispatcherCtx, &connection); err != nil {
-		log.Warn("get stream error", zap.Error(err))
+		log.Warn("[resource_manager] get token stream error", zap.Error(err))
 	}
 
+	toReconnect := false
 	for {
+		// Fetch the request from the channel.
 		var firstRequest *tokenRequest
```
HuSharp, line 277: How to change `var firstRequest *tokenRequest` to `case firstRequest := <-tbc.tokenRequestCh:`,
and move all the code inside the case

JmPotato: I refine this part to make sure we can avoid creating variables in the loop. PTAL again.

pkg/mcs/registry/registry.go
```diff
@@ -72,13 +64,7 @@ func (r *ServiceRegistry) InstallAllGRPCServices(srv *server.Server, g *grpc.Ser
 // InstallAllRESTHandler installs all registered REST services.
 func (r *ServiceRegistry) InstallAllRESTHandler(srv *server.Server, h map[string]http.Handler) {
```
HuSharp, line 65: How about turn `RegisterService` func to pointer receivers
`func (r *ServiceRegistry) RegisterService`

JmPotato: Maybe we could file another PR to address this. Let's keep this PR focus on the failover.

CabinfeverB (commented):
Except for the registration part, rest LGTM

CabinfeverB (approved):

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -48,34 +49,42 @@ type Manager struct {
...
 		if err := proto.Unmarshal([]byte(v), group); err != nil {
 			log.Error("err", zap.Error(err), zap.String("k", k), zap.String("v", v))
 			panic(err)
 		}
 		m.groups[group.Name] = FromProtoResourceGroup(group)
-	}
-	m.storage.LoadResourceGroupSettings(handler)
+	})
+	// Start the background metrics flusher.
+	go m.backgroundMetricsFlush(ctx)
```
nolouch, line 86: will this goroutine leak after the leader switched?

JmPotato: This is called in the leader election callback. So it will get the leader context which will be canceled after the switch.

nolouch (approved):
lgtm

PR #5901, opened 2023-02-02
*: move `tso` to `pkg`

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

### What problem does this PR solve?

Issue Number: Ref #5836.

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

Issue #5839, opened 2023-01-09
Separate scheduling into an independent service


rleungx:
### Phase 1: separate the scheduler into the scheduling service 

- [x] Decouple the dependency between server and schedule (mv to `/pkg/schedule`) @rleungx 
  - [x] https://github.com/tikv/pd/pull/5792
  - [x] https://github.com/tikv/pd/pull/6133
  - [x] https://github.com/tikv/pd/pull/6481
  - [x] https://github.com/tikv/pd/pull/6491
  - [x] https://github.com/tikv/pd/pull/6490
  - [x] https://github.com/tikv/pd/pull/6489
  - [x] https://github.com/tikv/pd/pull/6503
  - [x] https://github.com/tikv/pd/pull/6418
  - [x] https://github.com/tikv/pd/pull/6857
- [x] Refactor cluster and coordinator to reduce the dependencies @rleungx
  - [x] https://github.com/tikv/pd/pull/6551
  - [x] https://github.com/tikv/pd/pull/6552
  - [x] https://github.com/tikv/pd/pull/6553
  - [x] https://github.com/tikv/pd/pull/6557
  - [x] https://github.com/tikv/pd/pull/6769
  - [x] https://github.com/tikv/pd/pull/6771
  - [x] https://github.com/tikv/pd/pull/6773
  - [x] https://github.com/tikv/pd/pull/6774
  - [x] https://github.com/tikv/pd/pull/6776
- [x] Run scheduling service independently @rleungx 
  - [x] https://github.com/tikv/pd/pull/6542
  - [x] https://github.com/tikv/pd/pull/6826
  - [x] https://github.com/tikv/pd/pull/6891
  - [x] https://github.com/tikv/pd/pull/6893
  - [x] https://github.com/tikv/pd/pull/6896
  - [x] https://github.com/tikv/pd/pull/7236
  - [x] https://github.com/tikv/pd/pull/7039
  - [x] https://github.com/tikv/pd/pull/7061
- [x] Dynamic configuration @JmPotato
  - [x] https://github.com/tikv/pd/pull/6857
  - [x] https://github.com/tikv/pd/pull/6862
  - [x] https://github.com/tikv/pd/pull/6902
  - [x] https://github.com/tikv/pd/pull/6921
  - [x] https://github.com/tikv/pd/pull/6955
  - [x] https://github.com/tikv/pd/pull/7213
  - [x] https://github.com/tikv/pd/pull/7205
[...]

PR #5902, opened 2023-02-02
config: refactor TSO config

rleungx:

Issue Number: Ref #5839.

This PR can be reviewed after #5901 is merged.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```


rleungx:
@binshi-bing PTAL

binshi-bing (approved):
LGTM. Thanks for making this change!

pkg/tso/allocator_manager.go
```diff
@@ -134,17 +134,18 @@ type AllocatorManager struct {
 func NewAllocatorManager(
 	m *member.Member,
 	rootPath string,
-	cfg config,
+	cfg Config,
```
binshi-bing, line 137: cfg *Config

binshi-bing (changes requested):

server/config/config.go
```diff
@@ -409,6 +412,27 @@ func (c *Config) Parse(arguments []string) error {
 				c.Log.Level = c.LogLevelDeprecated
 			}
 		}
+		if c.EnableLocalTSO {
```
binshi-bing, line 415: why do we still keep EnableLocalTSO, TSOSaveInterval, and TSOUpdatePhysicalInterval in the server config after we added TSOConfig tso.Config?

server/config/config.go
```diff
@@ -409,6 +412,27 @@ func (c *Config) Parse(arguments []string) error {
...
+			msg := fmt.Sprintf("enable-local-tso in %s is deprecated, use [tso.enable-local-tso] instead", c.configFile)
```
binshi-bing, line 416: I think there is no need to change it for now, because we might want to share the same behavior between all-in-one PD and TSO microservice, and the current behavior is perfect for TSO microservice and is evetual behavior that we want after we move all the clients to TSO microservice.

server/config/config.go
```diff
@@ -589,6 +613,7 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
 	}
 
 	c.adjustLog(configMetaData.Child("log"))
+	c.adjustTSOConfig(configMetaData.Child("tso"))
```
binshi-bing, line 616: suggest not to change it to cause the compatibility issue for now

server/config/config.go
binshi-bing: If it's possible, we can defer this change to the future to avoid compatibility issue

server/config/config.go
```diff
@@ -1306,19 +1346,14 @@ func (c *Config) GetConfigFile() string {
 	return c.configFile
 }
 
-// IsLocalTSOEnabled returns if the local TSO is enabled.
-func (c *Config) IsLocalTSOEnabled() bool {
-	return c.EnableLocalTSO
-}
-
 // GetTSOUpdatePhysicalInterval returns TSO update physical interval.
 func (c *Config) GetTSOUpdatePhysicalInterval() time.Duration {
```
binshi-bing, line 1350: no one uses these APIs, shall we can just delete them?

server/config/config.go
binshi-bing: OK, I know the reason. It's to make toml.Decode(Config) happy.

server/config/config.go
binshi-bing: OK, I know the reason. It's to make toml.Decode(Config) happy.

pkg/tso/allocator_manager.go
```diff
@@ -134,17 +132,20 @@ type AllocatorManager struct {
 func NewAllocatorManager(
 	m *member.Member,
 	rootPath string,
-	cfg config,
+	enableLocalTSO bool,
```
binshi-bing, line 135: passing all tso configuration fields here seems to be a litter weird, but we can refine later.

binshi-bing (commented):
LGTM

pkg/tso/allocator_manager.go
rleungx: Yes, we can define an interface later to avoid it.

binshi-bing (approved):
LGTM

server/config/config.go
```diff
@@ -613,6 +570,49 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
 	return nil
 }
 
+// Utility to test if a configuration is defined.
+type configMetaData struct {
```
rleungx, line 574: This part will be moved to a common package later.

lhy1024 (approved):

nolouch (approved):
lgtm

server/config/config.go
```diff
@@ -613,6 +570,49 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
...
+	meta *toml.MetaData
+	path []string
```
JmPotato, line 576: This `path` is kind of confusing to me. Does it mean the config file path or the internal config key path? What about using `keyPath`?

JmPotato (commented):
The rest LGTM,

server/config/config.go
binshi-bing: configMetaData is an existing structure and the issue is a legacy issue. This PR just moves configMetaData from other place without modifying anything except the place, so I think we can address @JmPotato's concern in separate PR.

Issue #5568, opened 2022-09-30
Modify placement rule to support witness


ethercflow:
## Development Task
Refer to https://github.com/tikv/tikv/issues/12876

PR #5904, opened 2023-02-03
fit: optimize the selection of witness rule candidates

ethercflow:
ref tikv/pd#5568

Issue Number: ref tikv/pd#5568

1. If the `IsWitness` of the `Rule` is `true`, the leader will be removed from the candidate list, so that when the best rule is selected, it is impossible for the leader to exist in the witness rule;
2. When sorting the candidate peers, if the two peers participating in the comparison belong to the same `storestate`, then 50% of them will be randomly sorted to prevent the sorted peer list from being fixed, so that when the best rule is selected, the fixed peer will always be returned, causing some store balance, a store is 0;

before:

img_v2_b7ccd415-efc8-431d-a11f-7c77ee78430g

After:

a325ce27-abb7-43b2-949f-1764ccad3219

Update: 
If the rule remains unchanged, each time the rule_checker checks the region, it will directly fetch the previously calculated best rule from the cache without being randomly affected. But If the rule is manually modified, the cache will become invalid, and the rule_checker will recalculate each region, which may cause unnecessary scheduling for existing peers, such as adding TiFlash. So  remove the random strategy.

```commit-message

```

### Check List

Tests

- None

Code changes

- None

Side effects

- None

Related changes

- None

### Release note

```release-note
None
```


server/schedule/placement/fit.go
```diff
@@ -205,7 +207,7 @@ func newFitWorker(stores []*core.StoreInfo, region *core.RegionInfo, rules []*Ru
 	sort.Slice(peers, func(i, j int) bool {
 		// Put healthy peers in front of priority to fit healthy peers.
 		si, sj := stateScore(region, peers[i].GetId()), stateScore(region, peers[j].GetId())
-		return si > sj || (si == sj && peers[i].GetId() < peers[j].GetId())
```
rleungx, line 208: The previous comment says "Sort peers to keep the match result deterministic."

ethercflow:
Random might cause some test cases to fail, I'll think again (:

rleungx (approved):

bufferflies (approved):

PR #5906, opened 2023-02-05
resource_manager: save tokens in persistent storage

CabinfeverB:

Issue Number: ref #5851 

save GroupTokenBucketState every 1 minite

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes
- Has persistent data change

Side effects

- Increased code complexity

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/server/resource_group.go
```diff
@@ -145,17 +145,23 @@ func FromProtoResourceGroup(group *rmpb.ResourceGroup) *ResourceGroup {
 	switch group.GetMode() {
 	case rmpb.GroupMode_RUMode:
 		if settings := group.GetRUSettings(); settings != nil {
-			ruSettings = &RequestUnitSettings{
-				RU: NewGroupTokenBucket(settings.GetRU()),
+			ruSettings = &RequestUnitSettings{}
+			if settings.GetRU() != nil {
```
BornChanger, line 149: what will happen if settings.GetRU() is nil?

pkg/mcs/resource_manager/server/resource_group.go
```diff
@@ -164,10 +170,17 @@ func FromProtoResourceGroup(group *rmpb.ResourceGroup) *ResourceGroup {
 }
 
 // RequestRU requests the RU of the resource group.
-func (rg *ResourceGroup) RequestRU(now time.Time, neededTokens float64, targetPeriodMs uint64) *rmpb.GrantedRUTokenBucket {
+func (rg *ResourceGroup) RequestRU(now time.Time, neededTokens float64, targetPeriodMs uint64, storage endpoint.ResourceGroupStorage) *rmpb.GrantedRUTokenBucket {
 	rg.Lock()
 	defer rg.Unlock()
-	if rg.RUSettings == nil {
+	grant := rg.requestRU(now, neededTokens, targetPeriodMs)
```
BornChanger, line 176: How much overhead will it pay here? And how often is it called and do we need to call it so often?

pkg/mcs/resource_manager/server/resource_group.go
```diff
@@ -164,10 +170,17 @@ func FromProtoResourceGroup(group *rmpb.ResourceGroup) *ResourceGroup {
...
+	rg.persistStates(storage)
```
BornChanger, line 177: How much overhead will it pay here? And how often is it called and do we need to call it so often?

pkg/mcs/resource_manager/server/resource_group.go
CabinfeverB: `RUSettings.RU` will be an empty GroupTokenBucket which has nil pointer of `Settings`.  If `Settings` is a nil pointer, GroupTokenBucket can not handle RequestRU

nolouch (approved):
lgtm. 
maybe we can improve it if the group no updates in the future.

BornChanger (approved):

nolouch:
ptal @JmPotato

JmPotato (approved):

PR #5912, opened 2023-02-06
resource_manager: adjust the RU model config

JmPotato:

Issue Number: ref #5851.

```commit-message
- Adjust the RU model config.
- Unify the SQL and KV CPU time RU cost.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


nolouch (approved):
will replicas consider for write?

JmPotato:
> will replicas consider for write?

To calculate the replica, we need to merge https://github.com/tikv/client-go/pull/684, which may need to be further reviewed and discussed.

PR #5930, opened 2023-02-07
step: fix merge region failed caused by mismatch confver

### LINKED ISSUE #5656: merge check match witness
## Development Task

Consider witness merge in merge scheduler

ref https://github.com/tikv/tikv/issues/12876

ethercflow:
ref tikv/pd#5656

Issue Number: Close #5656

When region A's witness and region B's witness not on the same store. It will gen the op with

step1:  Demote witness voter -> witness learner
step2: switch witness learner -> non-witness leaner
step3: Promote non-witness learner -> voter

When calculating the conf ver change, after the third step is executed, the conf ver change condition of the first step is not satisfied, resulting in the difference between the actual region conf ver change and the calculated one.

So add witness change as a condition to solve this.

```commit-message
step: fix merge region failed caused by mismatch confver
```

### Check List

Tests

- Manual test 
   - test with sysbench prepare

Code changes

- None

Side effects

- None

Related changes

- None

### Release note

```release-note
None
```

### DIFF
--- a/server/schedule/operator/step.go
+++ b/server/schedule/operator/step.go
@@ -414,7 +414,7 @@ func (bsw BatchSwitchWitness) Influence(opInfluence OpInfluence, region *core.Re
 
 // Timeout returns duration that current step may take.
 func (bsw BatchSwitchWitness) Timeout(regionSize int64) time.Duration {
-	count := uint64(len(bsw.ToWitnesses)+len(bsw.ToNonWitnesses)) + 1
+	count := len(bsw.ToNonWitnesses)) + 1
 	return slowStepWaitDuration(regionSize) * time.Duration(count)
 }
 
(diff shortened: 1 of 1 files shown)


rleungx (approved):

server/schedule/operator/step.go
```diff
@@ -415,7 +415,7 @@ func (bsw BatchSwitchWitness) Influence(opInfluence OpInfluence, region *core.Re
 // Timeout returns duration that current step may take.
 func (bsw BatchSwitchWitness) Timeout(regionSize int64) time.Duration {
 	count := uint64(len(bsw.ToWitnesses)+len(bsw.ToNonWitnesses)) + 1
-	return fastStepWaitDuration(regionSize) * time.Duration(count)
+	return slowStepWaitDuration(regionSize) * time.Duration(count)
```
bufferflies, line 418: how about only use `ToNonWitness`?

bufferflies (commented):
rest lgtm

bufferflies (approved):

PR #5932, opened 2023-02-07
grpc: Use WithPrevKV() to get the previous key-value pair when get Delete Event

HuSharp:
Signed-off-by: husharp <jinhao.hu@pingcap.com>

Issue Number: Ref #5851

```commit-message
- Use WithPrevKV() to get the previous key-value pair when get Delete Event
- for client will get key with config path
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None
```


server/grpc_service.go
```diff
@@ -1950,7 +1950,7 @@ func (s *GrpcServer) LoadGlobalConfig(ctx context.Context, request *pdpb.LoadGlo
 	}
 	res := make([]*pdpb.GlobalConfigItem, len(r.Kvs))
 	for i, value := range r.Kvs {
-		res[i] = &pdpb.GlobalConfigItem{Kind: pdpb.EventType_PUT, Name: string(value.Key), Payload: value.Value}
+		res[i] = &pdpb.GlobalConfigItem{Kind: pdpb.EventType_PUT, Name: string(value.Key[len(configPath):]), Payload: value.Value}
```
nolouch, line 1953: why not handle in client. I think you use `clientv3.WithPrefix()` that expect to get the key with the prefix.

HuSharp: The purpose of the `clientv3.WithPrefix()`  is to get all the keys under this configuration path
I think it makes more sense to return keys that don't care about prefixes.
For loading configurations, the server's current implementation is to provide two parameters, Names and ConfigPath.
- Filling `Names` will literately get the values from `ConfigPath/Name`, the keys now have no prefix 
- If `Names` is empty, all values of the current`ConfigPath` will be obtained, but the keys are with prefix.
Consistency cannot be achieved

HuSharp: Okay, keep the client get key with config path and just return prekv when is a delete event.

glorv (commented):
LGTM

nolouch (approved):

server/grpc_service.go
```diff
@@ -1990,7 +1991,12 @@ func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, serve
 			for _, e := range res.Events {
 				// Since item value needs to support marshal of different struct types,
 				// it should be set to `Payload bytes` instead of `Value string`.
-				cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.Kv.Value, Kind: pdpb.EventType(e.Type)})
+				switch e.Type {
+				case clientv3.EventTypePut:
+					cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.Kv.Value, Kind: pdpb.EventType(e.Type)})
+				case clientv3.EventTypeDelete:
+					cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.PrevKv.Value, Kind: pdpb.EventType(e.Type)})
```
JmPotato, line 1998: According to the document of `WithPrevKV`:

> WithPrevKV gets the previous key-value pair before the event happens. If the previous KV is already compacted, nothing will be returned.

When the previous KV is already compacted, will `e.PrevKv.Value` panic? If not, what should we do to handle the empty value?

glorv: If the previous KV is compacted, it should not be a delete event.

glorv: ‘the previous KV is already compacted’ can only happen when delete is called multiple times, but this should not be our case.

HuSharp: If the previous KV is already compacted, e.PrevKv will be nil, so accessing its fields will not panic, just judge for nil
do u have any idea?

JmPotato (approved):

nolouch (approved):

server/grpc_service.go
```diff
@@ -1990,7 +1991,17 @@ func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, serve
 			for _, e := range res.Events {
```
YuJuncen, line 1991: Perhaps we'd better check the `res.Err()` before handling it? That method can detect both revision compacted and other things made the channel no longer be able to be used.
```go
// Err is the error value if this WatchResponse holds an error.
func (wr *WatchResponse) Err() error {
	switch {
	case wr.closeErr != nil:
		return v3rpc.Error(wr.closeErr)
	case wr.CompactRevision != 0:
		return v3rpc.ErrCompacted
	case wr.Canceled:
		if len(wr.cancelReason) != 0 {
			return v3rpc.Error(status.Error(codes.FailedPrecondition, wr.cancelReason))
		}
		return v3rpc.ErrFutureRev
	}
	return nil
}
```

nolouch: cc @HuSharp

HuSharp: done it

YuJuncen: LGTM

server/grpc_service.go
```diff
@@ -1969,28 +1969,44 @@ func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, serve
...
 				// Since item value needs to support marshal of different struct types,
 				// it should be set to `Payload bytes` instead of `Value string`.
-				cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.Kv.Value, Kind: pdpb.EventType(e.Type)})
+				switch e.Type {
+				case clientv3.EventTypePut:
+					cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.Kv.Value, Kind: pdpb.EventType(e.Type)})
+				case clientv3.EventTypeDelete:
+					if e.PrevKv != nil {
+						cfgs = append(cfgs, &pdpb.GlobalConfigItem{Name: string(e.Kv.Key), Payload: e.PrevKv.Value, Kind: pdpb.EventType(e.Type)})
+					} else {
```
glorv, line 2005: I think in this case we should just skip this message because prev-kv is compacted means their must be a delete event before this one, that means this is just a duplicated event, so we can just ignore it.

server/grpc_service.go
```diff
@@ -1969,18 +1969,23 @@ func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, serve
...
-			if revision < res.CompactRevision {
-				if err := server.Send(&pdpb.WatchGlobalConfigResponse{
-					Revision: res.CompactRevision,
-					Header: s.wrapErrorToHeader(pdpb.ErrorType_DATA_COMPACTED,
-						fmt.Sprintf("required watch revision: %d is smaller than current compact/min revision. %d", revision, res.CompactRevision)),
-				}); err != nil {
+			if res.Err() != nil {
+				var resp pdpb.WatchGlobalConfigResponse
+				if revision < res.CompactRevision {
+					resp.Header = s.wrapErrorToHeader(pdpb.ErrorType_DATA_COMPACTED,
```
nolouch, line 1982: directly return `error`.

HuSharp: The errors in Err() are that need to be retried so return error at the end of the if

server/grpc_service.go
```diff
@@ -1969,28 +1969,46 @@ func (s *GrpcServer) WatchGlobalConfig(req *pdpb.WatchGlobalConfigRequest, serve
...
+						fmt.Sprintf("required watch revision: %d is smaller than current compact/min revision %d.", revision, res.CompactRevision))
+				} else {
+					resp.Header = s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN,
+						fmt.Sprintf("watch channel meet other error %s.", res.Err().Error()))
+				}
+				if err := server.Send(&resp); err != nil {
 					return err
 				}
+				// Err() indicates that this WatchResponse holds a channel-closing error.
+				return res.Err()
```
glorv, line 1992: We should remove this return. return `res.Err` will let grpc deliver this error to the client and close the client, but we already notify the client about this error in the last sent message, so we'd better let the client to choose what to do next, in some situation, the client can just go on without retry creating a new client.

HuSharp: for these context-cancel errors there is no need to receive the channel, do you mean to return nil here?

glorv: ok. That make sense, I think return res.Err is ok here.

server/grpc_service.go
glorv: The client is likely to need special logic for DataCompacted error, if we just return err here, can the client distinguish between different errors

glorv (commented):
LGTM

nolouch (approved):

Issue #5837, opened 2023-01-09
Make the service start with independent mode


rleungx:
## Development Task

- [ ] parse mode #5858
- [x] #5857
- [x] #5921
- [ ] #5933
- [ ] #5836
- [ ] #5956
- [ ] run independent mode with resource manager

lhy1024:
ok

PR #5858, opened 2023-01-12
cmd: add service mode param

lhy1024:
Signed-off-by: lhy1024 <admin@liudos.us>

### What problem does this PR solve?

Issue Number: Close #5837

### What is changed and how does it work?

1. cmd: split main.go #5915
2. decouple the dependency between server and mcs
3. run independent mode with resource manager

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Code changes

- Has configuration change
- Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Possible performance regression
- Increased code complexity
- Breaking backward compatibility

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- Need to cherry-pick to the release branch

### Release note

```release-note
Please add a release note.

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

If you don't think this PR needs a release note then fill it with None.
```

Issue #5833, opened 2023-01-06
Balance leader may be triggered when transfer PD leader


rleungx:
## Enhancement Task
We use prepare checker to ensure most regions have sent the heartbeat before starting to schedule. But when the cluster has many regions, assume 200M, if the 180M regions finish the heartbeat, the new PD leader can start to schedule. See https://github.com/tikv/pd/blob/f8c061232841aade419c5ef1b1f4e73efa557348/server/cluster/coordinator.go#L49

However, the new PD leader may not contains all regions in memory due to restart and region synchronizer failing to sync. In this situation, the leader number of each store may not balance since some regions haven't finished the heartbeat. So balance leader scheduler may create operators to make them balanced.

PR #5933, opened 2023-02-07
*: decouple the dependency between server and mcs

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/5837

1. move `APIServiceGroup` to `apiutil`
2. replace `server` with `basic_server` in `mcs`
3. remove `init` of `mcs` in main.go

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
 None.
```


lhy1024:
cc @rleungx @binshi-bing

server/util.go
```diff
@@ -164,63 +163,6 @@ func checkBootstrapRequest(clusterID uint64, req *pdpb.BootstrapRequest) error {
...
-	InstallAllGRPCServices(srv *Server, g *grpc.Server)
-	InstallAllRESTHandler(srv *Server, userDefineHandler map[string]http.Handler)
-}
-
-// NewServiceRegistry is a hook for mcs code which implements the micro service.
-var NewServiceRegistry = func() ServiceRegistry {
-	return dummyServiceRegistry{}
-}
-
-type dummyServiceRegistry struct{}
```
lhy1024, line 180: No longer need dummy

pkg/mcs/registry/registry.go
```diff
@@ -22,7 +21,7 @@ import (
 	"net/http"
 
 	"github.com/pingcap/log"
-	"github.com/tikv/pd/server"
+	bs "github.com/tikv/pd/pkg/basic_server"
```
rleungx, line 24:
```suggestion
	bs "github.com/tikv/pd/pkg/basicserver"
```

binshi-bing: how about basicsvr?

binshi-bing (changes requested):
This pr seems to split from #5858. Regarding this pr, I have the same ask in #5858 and pasted below:

@lhy1024 , could you please split the common part (e.g., move APIServiceGroup from server to pkg/utils/apiutil, registry refactor) required by other services, into a separate pr and check in first. The common part in this pr, including the way to parse subcommand as discussed, and tso mcs have higher priority than resource manager mcs. In this way, we can both work on tso mcs and resource manager mcs more efficiently.

cc @rleungx

binshi-bing (approved):
This pr seems to split from #5858. Regarding this pr, I have the same ask in #5858 and pasted below:

@lhy1024 , thanks for spliting the common part (e.g., move APIServiceGroup from server to pkg/utils/apiutil, registry refactor) required by other services, from #5833 into this separate pr. They have higher priority than the remaining part in the original pr (mainly resource manager mcs related). Could you merge this one earlier so that we can both work on tso mcs and resource manager mcs more efficiently.

pkg/basic_server/basic_server.go
```diff
@@ -27,14 +28,18 @@ type Server interface {
...
 	Close()
-
 	// GetClient returns builtin etcd client.
 	GetClient() *clientv3.Client
 	// GetHTTPClient returns builtin etcd client.
 	GetHTTPClient() *http.Client
+	// AddStartCallback adds a callback in the startServer phase.
+	AddStartCallback(callbacks ...func())
+	// GetMember returns the member information.
+	GetMember() *member.Member
```
binshi-bing, line 42: Do we need to put this API into basicserver.Server? The member.Member defines leaders, uses embed.tcd, and member is pdpb.Member. Different server should have different definition for Member.

`
// Member is used for the election related logic.
type Member struct {

	leadership *election.Leadership
	leader     atomic.Value // stored as *pdpb.Member
	// etcd and cluster information.
	etcd     *embed.Etcd
	client   *clientv3.Client
	id       uint64       // etcd server id.
	member   *pdpb.Member // current PD's info.
	rootPath string
	// memberValue is the serialized string of `member`. It will be save in
	// etcd leader key when the PD node is successfully elected as the PD leader
	// of the cluster. Every write will use it to check PD leadership.
	memberValue string
}
`

pkg/basic_server/basic_server.go
```diff
@@ -27,14 +28,18 @@ type Server interface {
...
+	// AddLeaderCallback adds a callback in the leader campaign phase.
+	AddLeaderCallback(callbacks ...func(context.Context))
```
binshi-bing, line 44: Do we need to put this API into basicserver.Server? This API means that this server will elect for leader, correct? It doesn't sound a must-havior behavior of a **basic** server.

lhy1024:
> This pr seems to split from #5858. Regarding this pr, I have the same ask in #5858 and pasted below:
> 
> @lhy1024 , thanks for spliting the common part (e.g., move APIServiceGroup from server to pkg/utils/apiutil, registry refactor) required by other services, from #5833 into this separate pr. They have higher priority than the remaining part in the original pr (mainly resource manager mcs related). Could you merge this one earlier so that we can both work on tso mcs and resource manager mcs more efficiently.

In master branch, resource management is imported by install, which means it will always start however service mode. So we need to change it for tso mode.

pkg/basic_server/basic_server.go
lhy1024: I also consider to replace it with primary

lhy1024: add todo

pkg/mcs/registry/registry.go
lhy1024: I also think `basicserver` is better than `basicsvr`

rleungx:
Please change the file back to 100644

rleungx (approved):

pkg/mcs/registry/registry.go
```diff
@@ -97,9 +97,3 @@ func (r *ServiceRegistry) InstallAllRESTHandler(srv *server.Server, h map[string
 func (r ServiceRegistry) RegisterService(name string, service ServiceBuilder) {
```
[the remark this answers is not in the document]
lhy1024: done

server/testutil.go
```diff
@@ -137,3 +140,21 @@ func MustWaitLeader(re *require.Assertions, svrs []*Server) *Server {
 	})
 	return leader
 }
+
+// CreateMokHandler creates a mock handler for test.
+func CreateMokHandler(re *require.Assertions, ip string) HandlerBuilder {
```
HuSharp, line 145: I'm not sure why it was called mok before, is mock written wrong?

lhy1024: done

HuSharp: `server_test.go` and `func CreateMokHandler` also has some legacy `mok`

nolouch (approved):

server/testutil.go
nolouch: ptal @lhy1024

lhy1024: replace name in url

Issue #5836, opened 2023-01-09
Separate TSO into an independent service


rleungx:

binshi-bing:
assign to me?

rleungx:
> assign to me?

Yes, but I cannot find you on the assignee list. lol

binshi-bing:
## pre-work tracked below

- [x] #5779 
- [x] #5771 
- [x] (Cancelled) [Support tso service #1038 in kvproto repo](https://github.com/pingcap/kvproto/pull/1038/files)
- [x] (Cancelled) #5852
- [ ] #5837 
- [x] #5857
- [ ] #5858
- [ ] #5874

binshi-bing:
@rleungx could you me grant me the access to modify this issue?

binshi-bing:
Define an independent TSO gRPC server #1053 https://github.com/pingcap/kvproto/pull/1053

binshi-bing:
Closed by mistake. Reopen it.

binshi-bing:
Closed by [#5986](https://github.com/tikv/pd/pull/5986) by mistake. @rleungx , could you help to reopen this issue? It seems that I don't have permission to do it.

PR #5942, opened 2023-02-08
tso: migrate tso key to new one

rleungx:

Issue Number: Ref #5836.

Before this PR, we have the following keys in etcd after enabling local TSO feature:
```
/pd/7198087827415338272/dc-location/9320049945780417218
/pd/7198087827415338272/lta/dc-1
/pd/7198087827415338272/lta/dc-1/timestamp
/pd/7198087827415338272/lts/dc-1
/pd/7198087827415338272/timestamp
```
After this PR 
```
/pd/7198087827415338272/dc-location/9320049945780417218 -> /pd/7198091555267330021/ms/tso/default/dc-location/9320049945780417218
/pd/7198087827415338272/lta/dc-1 -> /pd/7198091555267330021/ms/tso/default/lts/dc-1
/pd/7198087827415338272/lta/dc-1/timestamp -> /pd/7198091555267330021/ms/tso/default/lts/dc-1/timestamp
/pd/7198087827415338272/lts/dc-1 -> /pd/7198091555267330021/ms/tso/default/lta/dc-1
/pd/7198087827415338272/timestamp -> /pd/7198091555267330021/ms/tso/default/gts/timestamp
```

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/member/member.go
+++ b/pkg/member/member.go
@@ -31,6 +31,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/election"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/storage/endpoint"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
 	"go.etcd.io/etcd/clientv3"
@@ -302,7 +303,7 @@ func (m *Member) getMemberLeaderPriorityPath(id uint64) string {
 
 // GetDCLocationPathPrefix returns the dc-location path prefix of the cluster.
 func (m *Member) GetDCLocationPathPrefix() string {
-	return path.Join(m.rootPath, dcLocationConfigEtcdPrefix)
+	return path.Join(m.rootPath, endpoint.MicroserviceKey, endpoint.TSOServiceKey, "default", dcLocationConfigEtcdPrefix)
 }
 
 // GetDCLocationPath returns the dc-location path of a member with the given member ID.
--- a/pkg/member/member.go
+++ b/pkg/member/member.go
@@ -303,7 +303,7 @@ func (m *Member) getMemberLeaderPriorityPath(id uint64) string {
 
 // GetDCLocationPathPrefix returns the dc-location path prefix of the cluster.
 func (m *Member) GetDCLocationPathPrefix() string {
-	return path.Join(m.rootPath, endpoint.MicroserviceKey, endpoint.TSOServiceKey, "default", dcLocationConfigEtcdPrefix)
+	return path.Join(m.rootPath, endpoint.TSOMicroserviceKey, "default", dcLocationConfigEtcdPrefix)
 }
 
 // GetDCLocationPath returns the dc-location path of a member with the given member ID.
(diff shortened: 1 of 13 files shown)


pkg/tso/tso.go
```diff
@@ -169,36 +169,33 @@ func (t *timestampOracle) loadTimestamp() (time.Time, error) {
...
 	})
-
-	last, err := t.loadTimestamp()
+	// It is for compatibility with the old version and will try to load timestamp from the old path at the first time.
+	// Then, switch to the new path.
+	last, kvs, err := t.loadTimestamp()
+	if err != nil {
+		return err
+	}
+	// If old path exists, we should switch to the new path.
```
binshi-bing, line 188: We still need a configuration to turn on/off this path format change. Without this configuration, once this change is deployed to the prod, it will switch to new format even before the entire mcs feature is ready, which results in breaking change.

binshi-bing: "If old path exists, we should switch to the new path" -- this will become the mysterious logic in the code. Would it be better if we can control the behavior with explicit hot configuration? Probably we need introduce bi-directional smooth switching.

pkg/tso/tso.go
```diff
@@ -230,6 +228,30 @@ func (t *timestampOracle) SyncTimestamp(leadership *election.Leadership) error {
 	return nil
 }
 
+func (t *timestampOracle) switchToNewPath(keyValue *mvccpb.KeyValue) error {
```
binshi-bing, line 231: As my previous comment mentioned, need configuration and switchToOldPath too to support bi-directional smooth switching.

binshi-bing (commented):
left some comments.

pkg/tso/tso.go
rleungx: I think we need a discussion about it. I'm not sure whether we should support downgrade operations.

pkg/member/member.go
```diff
@@ -302,7 +303,7 @@ func (m *Member) getMemberLeaderPriorityPath(id uint64) string {
 
 // GetDCLocationPathPrefix returns the dc-location path prefix of the cluster.
 func (m *Member) GetDCLocationPathPrefix() string {
-	return path.Join(m.rootPath, dcLocationConfigEtcdPrefix)
+	return path.Join(m.rootPath, endpoint.MicroserviceKey, endpoint.TSOServiceKey, "default", dcLocationConfigEtcdPrefix)
```
JmPotato, line 306: I think `endpoint.MicroserviceKey` + `service_key` could be extracted by a common function to be used later maybe.

nolouch (commented):
maybe we can replace the cluster-id with keyspace-id or namespace(teneant group) in the new way? cluster id is useless i think.

pkg/member/member.go
rleungx: sounds good

JmPotato (approved):

rleungx:
> maybe we can replace the cluster-id with keyspace-id or namespace(teneant group) in the new way? cluster id is useless i think.

After thinking about it, I decided to keep it, no sure if the different clusters will have the same tenant group/ID.

PR #5966, opened 2023-02-10
cluster: no more redundant etcd writes when `ReportMinResolvedTS` is called

### LINKED ISSUE #5965: Frequent `ReportMinResolvedTS` can cause PD OOM
### What did you do?

ETCD write latency burst up.

image

There will be no adverse effects.

### What did you see instead?

TiKV keeps reconnecting to PD.
image

The number of ReportMinResolvedTS is increased.
image

PD OOM
image

v6.5.0

HunDunDM:

Issue Number: Close #5965

no more redundant etcd writes when `SetMinResolvedTS` is called

### Check List

Tests

- Unit test
  - This PR is based on c64be1f858c4c79356fecdc0f232ba786b1c9d70
  - Make `putStoreLocked` called by `SetMinResolvedTS` slower [7a47041](https://github.com/HunDunDM/pd/commit/7a47041234ec37d92774fee2e8f78080e7d865da)
  - `tiup playground v6.5.0 --kv=15 --db=1 --pd=1 --tiflash=0 --pd.binpath=bin/pd-server`
  - The problem can be reproduced in grafana
    - image
    - image
  - `putStoreLocked` was removed in this PR and replaced with `c.core.PutStore(newStore)`, so it can be considered fixed.

Code changes

Side effects

Related changes

### Release note

```release-note
Fixed an issue where ETCD write delay could cause PD OOM.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -2326,11 +2326,9 @@ func (c *RaftCluster) SetMinResolvedTS(storeID, minResolvedTS uint64) error {
 		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
 	}
 
-	newStore := store.Clone(
-		core.SetMinResolvedTS(minResolvedTS),
-	)
-
-	return c.putStoreLocked(newStore)
+	newStore := store.Clone(core.SetMinResolvedTS(minResolvedTS))
+	c.core.PutStore(newStore)
+	return nil
 }
 
 func (c *RaftCluster) checkAndUpdateMinResolvedTS() (uint64, bool) {


lhy1024 (approved):

lhy1024:
do we need to pick to 6.6？

nolouch (commented):
better to add a test. could you add a test or post the manual test result.

HunDunDM:
Adding unit tests to the removed code was difficult, I added a manual test to prove that the original code was buggy.

nolouch (approved):

CabinfeverB:
It seems that we don't need to persist store in `SetMinResolvedTS`, and persist it only in heartbeat? 
What happens if the PD goes down after SetMinResolvedTS?

lhy1024:
> It seems that we don't need to persist store in `SetMinResolvedTS`, and persist it only in heartbeat? What happens if the PD goes down after SetMinResolvedTS?

we persist min-resolved-ts of the cluster in another goroutine

PR #5964, opened 2023-02-10
server: move client function to util

lhy1024:
Signed-off-by: lhy1024 <admin@liudos.us>

### What problem does this PR solve?

Issue Number: Close #5837

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
 None.
```

PR #5971, opened 2023-02-13
*: add service subcommand

rleungx:

Issue Number: Ref #5836.

This PR has a similar functionality to #5964 but uses the independent config for starting the TSO server. Besides, we can use the same function to init log for all binaries now. 
Update: 
Replace the original implementation with the cobra.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/cmd/pd-server/main.go
+++ b/cmd/pd-server/main.go
@@ -24,8 +24,8 @@ import (
 	grpcprometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
+	"github.com/spf13/cobra"
 	"github.com/tikv/pd/pkg/autoscaling"
-	bs "github.com/tikv/pd/pkg/basicserver"
 	"github.com/tikv/pd/pkg/dashboard"
 	"github.com/tikv/pd/pkg/errs"
 	tso "github.com/tikv/pd/pkg/mcs/tso/server"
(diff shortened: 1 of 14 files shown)


JmPotato (approved):

lhy1024 (approved):

cmd/pd-server/main.go
```diff
@@ -40,8 +41,27 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	serviceMode    = "service"
+	tsoServiceMode = "tso-service"
```
binshi-bing, line 46: nit: can we combine these two parameters into 'service-mode-tso'? This is the subcommand format that I know. If "service tso-service" is TiDB convention, please ignore my comment.

cmd/pd-server/main.go
```diff
@@ -40,8 +41,27 @@ import (
...
+)
+
 func main() {
-	ctx, cancel, svr := createServerWrapper(os.Args[1:])
+	var (
+		ctx    context.Context
+		cancel context.CancelFunc
+		svr    bs.Server
+	)
+	if len(os.Args) >= 3 && os.Args[1] == serviceMode {
```
binshi-bing, line 55: shouldn't it be case insensitive?

cmd/pd-server/main.go
```diff
@@ -40,8 +41,27 @@ import (
...
+		switch os.Args[2] {
+		case tsoServiceMode:
+			ctx, cancel, svr = tso.CreateServerWrapper(os.Args[3:])
+		default:
+			log.Fatal("service are not implemented")
```
binshi-bing, line 60: After log.Fatal, will the command output be friendly to user like displaying the help info and the supported service modes? If not, probably we want to log error and print help info here.

binshi-bing (approved):
Left few comments, could you please take a look? The other parts LGTM

rleungx:
Since the PR has been rewritten, need two more LGTM before merging.

lhy1024 (approved):

cmd/pd-server/main.go
```diff
@@ -41,43 +42,88 @@ import (
...
-	case syscall.SIGTERM:
-		exit(0)
-	default:
-		exit(1)
+// NewTSOServiceCommand returns the unsafe remove failed stores command.
+func NewTSOServiceCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tso [args]",
+		Short: "run a tso service",
+		Run:   tso.CreateServerWrapper,
```
binshi-bing, line 98: so this is why we removed the return data types and have to duplicate server run() and exit code. LGTM

binshi-bing (approved):
Looks like with Cobra, we have to duplicate server.run and exit code. LGTM

lhy1024:
> Looks like with Cobra, we have to duplicate server.run and exit code. LGTM

I think we can will move some functions to `util.go` after ensuring `todo` in tso server

rleungx:
@JmPotato @nolouch Please take a look when you are available.

JmPotato (approved):

Issue #5953, opened 2023-02-09
*: `make test` is failed about `cgroup`


lhy1024:
### What did you do?

make test

### What did you expect to see?

success

### What did you see instead?

failed after https://github.com/tikv/pd/pull/5848 merged

```
--- FAIL: TestGetCgroupCPU (0.00s)
    cgroup_cpu_test.go:45: 
                Error Trace:    /home/lhy1024/pd/pkg/cgroup/cgroup_cpu_test.go:45
                Error:          Received unexpected error:
                                no cpu controller detected
                                github.com/tikv/pd/pkg/cgroup.getCgroupCPUHelper
                                        /home/lhy1024/pd/pkg/cgroup/cgroup_cpu.go:33
                                github.com/tikv/pd/pkg/cgroup.GetCgroupCPU
                                        /home/lhy1024/pd/pkg/cgroup/cgroup_cpu_linux.go:28
                                github.com/tikv/pd/pkg/cgroup.TestGetCgroupCPU
                                        /home/lhy1024/pd/pkg/cgroup/cgroup_cpu_test.go:44
                                testing.tRunner
                                        /usr/local/go/src/testing/testing.go:1446
                                runtime.goexit
                                        /usr/local/go/src/runtime/asm_amd64.s:1594
                Test:           TestGetCgroupCPU
[2023/02/09 14:11:28.732 +08:00] [WARN] [cgroup_memory.go:84] ["no cgroup memory controller detected"]
[2023/02/09 14:11:28.780 +08:00] [WARN] [cgroup_memory.go:55] ["no cgroup memory controller detected"]
[2023/02/09 14:11:28.796 +08:00] [WARN] [cgroup_memory.go:115] ["no cgroup memory controller detected"]
FAIL
coverage: 76.5% of statements
FAIL    github.com/tikv/pd/pkg/cgroup   0.184s
```
```
--- FAIL: TestFinalizer (0.01s)
    finalizer_test.go:40: 
                Error Trace:    /home/lhy1024/pd/pkg/gctuner/finalizer_test.go:40
                Error:          Not equal: 
                                expected: 1
                                actual  : 0
                Test:           TestFinalizer
[2023/02/09 14:18:20.957 +08:00] [INFO] [memory_limit_tuner.go:49] [debug.SetMemoryLimit] [limit=858993459] [ret=9223372036854775807]
[...]

PR #5975, opened 2023-02-13
add enable_flaky_tests and skip some flaky tests in gctuner

hnes:
Signed-off-by: Sen Han <00hnes@gmail.com>

Issue Number: ref #5953

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/gctuner/finalizer_test.go
+++ b/pkg/gctuner/finalizer_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build enable_flaky_tests
+
 package gctuner
 
 import (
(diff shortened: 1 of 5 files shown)


lhy1024 (commented):
`cgroup_cpu_test.go` also need

pkg/gctuner/finalizer_test.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build enable_flaky_tests
```
lhy1024, line 15: `cgroup_cpu_test.go` also need

hnes: The test failed in `cgroup_cpu_test.go` we discussed before may only occur in some OS env. I don't know whether it is appropriate to classify it into flaky tests.

lhy1024:
I think we also need to pick it to 6.6

pkg/gctuner/finalizer_test.go
hnes: Done

PR #6003, opened 2023-02-16
resource_group_client: statistic sql cpu cost

HuSharp:

Issue Number: Ref #5851

```commit-message
statistic tidb cpu cost
```

### Release note

```release-note
None
```


client/resource_group/controller/controller.go
```diff
@@ -35,7 +35,7 @@ const (
 
 // ResourceGroupKVInterceptor is used as quota limit controller for resource group using kv store.
 type ResourceGroupKVInterceptor interface {
-	// OnRequestWait is used to check whether resource group has enough tokens. It maybe needs wait some time.
+	// OnRequestWait is used to check whether resource group has enough tokens. It maybe needs to wait some time.
```
JmPotato, line 38:
```suggestion
	// OnRequestWait is used to check whether the resource group has enough tokens. It maybe needs to wait sometime.
```

client/resource_group/controller/model.go
```diff
@@ -205,3 +212,14 @@ func sub(custom1 *rmpb.Consumption, custom2 *rmpb.Consumption) {
 	custom1.KvReadRpcCount -= custom2.KvReadRpcCount
 	custom1.KvWriteRpcCount -= custom2.KvWriteRpcCount
 }
+
+// getCPUTime returns the cumulative user+system time (in ms) since the process start.
+func getCPUTime() float64 {
```
JmPotato, line 217:
```suggestion
func getSQLProcessCPUTime() float64 {
```

nolouch (commented):
I think we need add a metrics for it.

HuSharp:
> I think we need add a metrics for it.

`sqlCPUMetrics` have been in `backgroundMetricsFlush`

nolouch (commented):
Could you show the metrics panel as an example?

HuSharp:
> Could you show the metrics panel as an example?

update TiDB and client-go as this patch, run `sysbench read_and_write` will get:
@nolouch @JmPotato  PTAL, thx!

img_v2_b2c1328e-0767-4279-b529-1a43a97f17bg

client/resource_group/controller/model.go
```diff
@@ -133,9 +138,11 @@ func newSQLCalculator(cfg *Config) *SQLCalculator {
 	return &SQLCalculator{Config: cfg}
 }
 
-// Trickle ...
-// TODO: calculate the SQL CPU cost and related resource consumption.
+// Trickle Update Sql Layer CPU consumption.
 func (dsc *SQLCalculator) Trickle(ctx context.Context, consumption *rmpb.Consumption) {
+	delta := getSQLProcessCPUTime() - consumption.SqlLayerCpuTimeMs
+	consumption.SqlLayerCpuTimeMs = delta
```
nolouch, line 144: I think it's an accumulative value, not a delta value. we can calculate the delta within `collectRequestAndConsumption`? ptal @JmPotato

HuSharp: Yes, this will eventually be calculated in `collectRequestAndConsumption`. 
The reason why delta is calculated here is because it will be added here. 
https://github.com/tikv/pd/pull/6003/files#diff-ac1c6525251add035bc46a2bce40ac9d39ee05a8f1b9c062657e869d10506497R499

HuSharp:
> Could you show the metrics panel as an example?

after modify pd client in tidb and client-go
when run sysbench:
image

CabinfeverB (approved):

client/resource_group/controller/model.go
```diff
@@ -133,9 +138,11 @@ func newSQLCalculator(cfg *Config) *SQLCalculator {
 	return &SQLCalculator{Config: cfg}
 }
 
-// Trickle ...
-// TODO: calculate the SQL CPU cost and related resource consumption.
+// Trickle Update Sql Layer CPU consumption.
 func (dsc *SQLCalculator) Trickle(ctx context.Context, consumption *rmpb.Consumption) {
+	delta := getSQLProcessCPUTime() - consumption.SqlLayerCpuTimeMs
```
JmPotato, line 143: The `consumption.SqlLayerCpuTimeMs` seems should be zero here, why use `getSQLProcessCPUTime()` to minus it?

HuSharp: Since the `get pid` function of `gosigar` can only get the cpu time from the start of the thread to this point.
And to satisfy the purpose of getRUValue/getRawValue to remove tokens from delta
so the last cpu time is passed before the trickle.

> And to satisfy the purpose of getRUValue/getRawValue to remove tokens from delta

Because before this pr the tidb cpu is also to count RU's, but Actually, I'm not quite sure what the purpose of this now.
Do you think i need to refine the `updateRunState` and `trickle` code?

client/resource_group/controller/controller.go
```diff
@@ -35,7 +35,7 @@ const (
 
 // ResourceGroupKVInterceptor is used as quota limit controller for resource group using kv store.
 type ResourceGroupKVInterceptor interface {
-	// OnRequestWait is used to check whether resource group has enough tokens. It maybe needs wait some time.
+	// OnRequestWait is used to check whether the resource group has enough tokens. It maybe needs to wait sometime.
```
nolouch, line 38:
```suggestion
	// OnRequestWait is used to check whether the resource group has enough tokens. It maybe needs to wait for a while.
```

nolouch (commented):
need add a switch about used in tenant/keyspace to group  mode.

client/resource_group/controller/config.go
```diff
@@ -89,9 +91,17 @@ func defaultRequestUnitConfig() *RequestUnitConfig {
 		WriteBaseCost:    defaultWriteBaseCost,
 		WriteCostPerByte: defaultWriteCostPerByte,
 		CPUMsCost:        defaultCPUMsCost,
+		IsServerless:     false,
 	}
 }
 
+// DefaultServerlessRUConfig returns the default configuration of Serverless.
+func DefaultServerlessRUConfig() *RequestUnitConfig {
```
nolouch, line 99:
```suggestion
func SetSingleGroupByKeyspaceMode() *RequestUnitConfig {
```
It's better not to distinguish between cloud and op

client/resource_group/controller/config.go
```diff
@@ -79,19 +79,29 @@ type RequestUnitConfig struct {
...
-func defaultRequestUnitConfig() *RequestUnitConfig {
+// DefaultRequestUnitConfig returns the default request unit configuration.
+func DefaultRequestUnitConfig() *RequestUnitConfig {
 	return &RequestUnitConfig{
 		ReadBaseCost:     defaultReadBaseCost,
 		ReadCostPerByte:  defaultReadCostPerByte,
 		WriteBaseCost:    defaultWriteBaseCost,
 		WriteCostPerByte: defaultWriteCostPerByte,
 		CPUMsCost:        defaultCPUMsCost,
+		IsServerless:     false,
```
nolouch, line 94: need to change this field too

client/resource_group/controller/config.go
```diff
@@ -79,19 +79,29 @@ type RequestUnitConfig struct {
...
 	}
 }
 
+// SetSingleGroupByKeyspaceMode returns the default configuration of Serverless.
+func SetSingleGroupByKeyspaceMode() *RequestUnitConfig {
```
nolouch, line 99:
```suggestion
func SetSingleGroupByKeyspace() *RequestUnitConfig {
```

client/resource_group/controller/config.go
```diff
@@ -79,19 +79,29 @@ type RequestUnitConfig struct {
...
+	cfg := DefaultRequestUnitConfig()
+	cfg.IsServerless = true
```
nolouch, line 101:
```suggestion
	cfg.IsSingleGroupByKeyspace = true
```

nolouch:
BTW, maybe we need add a metrics that already convert SQLLayer CPU Time to RU.

tests/mcs/resource_manager/resource_manager_test.go
```diff
@@ -293,15 +293,7 @@ func (suite *resourceManagerClientTestSuite) TestResourceGroupController() {
 		re.Contains(resp, "Success!")
 	}
 
-	cfg := &controller.RequestUnitConfig{
```
CabinfeverB, line 296: This configuration is for the convenience of construction consumption speed case. Don't delete it. Otherwise modify `tokenConsumptionPerSecond` in this test

client/resource_group/controller/config.go
CabinfeverB, line 94: IMO, this field is not related to RequestUnit. Can we put it into `Config` or `ResourceGroupsController`?

CabinfeverB (commented):
How about setting `IsSingleGroupByKeyspace` by adding Option in `NewResourceGroupController`

tests/mcs/resource_manager/resource_manager_test.go
HuSharp: thx for ur advice.

client/resource_group/controller/config.go
HuSharp: > How about setting `IsSingleGroupByKeyspace` by adding Option in `NewResourceGroupController`

done. PTAL! thx!

HuSharp:
> BTW, maybe we need add a metrics that already convert SQLLayer CPU Time to RU.

done, PTAL, thx!

client/resource_group/controller/model.go
```diff
@@ -205,3 +212,26 @@ func sub(custom1 *rmpb.Consumption, custom2 *rmpb.Consumption) {
 	custom1.KvReadRpcCount -= custom2.KvReadRpcCount
 	custom1.KvWriteRpcCount -= custom2.KvWriteRpcCount
 }
+
+// getSQLProcessCPUTime returns the cumulative user+system time (in ms) since the process start.
+func getSQLProcessCPUTime(isServerless bool) float64 {
```
CabinfeverB, line 217:
```suggestion
func getSQLProcessCPUTime(IsSingleGroupByKeyspace bool) float64 {
```

client/resource_group/controller/controller.go
```diff
@@ -491,31 +506,14 @@ func (gc *groupCostController) initRunState() {
...
+	gc.mu.Lock()
 	for _, calc := range gc.calculators {
-		calc.Trickle(ctx, deltaConsumption)
+		calc.Trickle(gc.mu.consumption, gc.mainCfg.IsSingleGroupByKeyspace)
 	}
-	gc.mu.Lock()
-	add(gc.mu.consumption, deltaConsumption)
 	*gc.run.consumption = *gc.mu.consumption
 	gc.mu.Unlock()
-	// remove tokens
```
CabinfeverB, line 504: If IsSingleGroupByKeyspace is false, do we need to add SQL CPU to RU consumption as well?

CabinfeverB: I mean we can keep this code.

client/resource_group/controller/model.go
```diff
@@ -205,3 +212,26 @@ func sub(custom1 *rmpb.Consumption, custom2 *rmpb.Consumption) {
...
+	pid := os.Getpid()
+	cpuTime := gosigar.ProcTime{}
+	if err := cpuTime.Get(pid); err != nil {
+		log.Error("getCPUTime get pid failed", zap.Error(err))
+	}
+
+	return float64(cpuTime.User + cpuTime.Sys)
+}
+
+// TODO: will be implemented by Goroutine after https://github.com/golang/go/issues/41554 merged
```
nolouch, line 234: no need to rely on it, we can do it in another way first.

HuSharp: Can you be more detailed?

client/resource_group/controller/controller.go
HuSharp: I think this time pr is mainly for the IsSingleGroupByKeyspace case.
maybe better add this code when implementing multiple groups.

client/resource_group/controller/config.go
```diff
@@ -102,12 +102,14 @@ type Config struct {
 	WriteBaseCost  RequestUnit
 	WriteBytesCost RequestUnit
 	CPUMsCost      RequestUnit
+	// The CPU statistics need to distinguish between different environments.
+	IsSingleGroupByKeyspace bool
```
JmPotato, line 106: This name is too specified. Could we name it something more general? For example, `EnableTiDBProcessCPUMetering`. This should be more clear to tell its use and purpose.

client/resource_group/controller/controller.go
```diff
@@ -491,31 +506,14 @@ func (gc *groupCostController) initRunState() {
...
 }
 
-func (gc *groupCostController) updateRunState(ctx context.Context) {
+func (gc *groupCostController) updateRunState() {
 	newTime := time.Now()
-	deltaConsumption := &rmpb.Consumption{}
+	gc.mu.Lock()
 	for _, calc := range gc.calculators {
-		calc.Trickle(ctx, deltaConsumption)
+		calc.Trickle(gc.mu.consumption, gc.mainCfg.IsSingleGroupByKeyspace)
```
JmPotato, line 513: Considering it will hold the lock now, will `Trickle` cost a long time to process here to make the locking time unexpectedly long?

nolouch (approved):
lgtm

client/resource_group/controller/controller.go
HuSharp: done.

client/resource_group/controller/config.go
HuSharp: The purpose of this switch is to distinguish whether enable single group by keyspace feature.
I changed it to a private field and only changed the option in the controller

client/resource_group/controller/controller.go
HuSharp: - We need to get the previous sql cpu time to calculate the delta consumption, so we need to hold lock to pass the race test.
-  For now impl, Trickle doesn't take long time to get the SQL time.

client/resource_group/controller/controller.go
```diff
@@ -491,31 +506,14 @@ func (gc *groupCostController) initRunState() {
...
 }
 
-func (gc *groupCostController) updateRunState(ctx context.Context) {
+func (gc *groupCostController) updateRunState() {
 	newTime := time.Now()
-	deltaConsumption := &rmpb.Consumption{}
+	gc.mu.Lock()
 	for _, calc := range gc.calculators {
-		calc.Trickle(ctx, deltaConsumption)
+		calc.Trickle(gc.mu.consumption, gc.mainCfg.isSingleGroupByKeyspace)
```
JmPotato, line 513: Currently, all `ResourceCalculator`s own the config inside, so there is no need to pass it again here.

client/resource_group/controller/model.go
```diff
@@ -133,9 +144,15 @@ func newSQLCalculator(cfg *Config) *SQLCalculator {
...
-func (dsc *SQLCalculator) Trickle(ctx context.Context, consumption *rmpb.Consumption) {
+func (dsc *SQLCalculator) SetConfig(config *Config) {
+	dsc.Config = config
+}
+
+// Trickle update sql layer CPU consumption.
+func (dsc *SQLCalculator) Trickle(consumption *rmpb.Consumption) {
+	delta := getSQLProcessCPUTime(dsc.isSingleGroupByKeyspace) - consumption.SqlLayerCpuTimeMs
+	consumption.TotalCpuTimeMs += delta
+	consumption.SqlLayerCpuTimeMs += delta
```
nolouch, line 155: we need make it start from 0.

nolouch (changes requested):

client/resource_group/controller/model.go
HuSharp: We can assign the sql time at the time of initialization?

client/resource_group/controller/controller.go
```diff
@@ -468,7 +483,7 @@ func newGroupCostController(
 		gc.handleRespFunc = gc.handleRawResourceTokenResponse
 	}
 
-	gc.mu.consumption = &rmpb.Consumption{}
+	gc.mu.consumption = &rmpb.Consumption{SqlLayerCpuTimeMs: getSQLProcessCPUTime(mainCfg.isSingleGroupByKeyspace)}
```
nolouch, line 486: I think should make the `lastRequestConsumption` init with it. then `collectRequestAndConsumption` will only calculate the delta part.

HuSharp: You are right, the run.consumption represents the total, not the delta

HuSharp:
metrics 
image

pkg/mcs/resource_manager/server/metrics.go
```diff
@@ -46,6 +46,14 @@ var (
 			Help:      "Bucketed histogram of the write request unit cost for all resource groups.",
 			Buckets:   prometheus.ExponentialBuckets(3, 10, 5), // 3 ~ 300000
 		}, []string{resourceGroupNameLabel})
+	sqlLayerRequestUnitCost = prometheus.NewHistogramVec(
```
nolouch, line 49: maybe we do no need histogram.

HuSharp: have change to vec

client/resource_group/controller/controller.go
```diff
@@ -109,15 +120,19 @@ func NewResourceGroupController(
...
 		lowTokenNotifyChan:    make(chan struct{}, 1),
 		tokenResponseChan:     make(chan []*rmpb.TokenBucketResponse, 1),
 		tokenBucketUpdateChan: make(chan *groupCostController, maxNotificationChanLen),
-		calculators:           []ResourceCalculator{newKVCalculator(config), newSQLCalculator(config)},
-	}, nil
+	}
+	for _, opt := range opts {
+		opt(controller)
+	}
+	controller.calculators = []ResourceCalculator{newKVCalculator(config), newSQLCalculator(config)}
```
JmPotato, line 134: You should use `controller.config`.

HuSharp: done.

nolouch (approved):

JmPotato (approved):

nolouch:
ptal @JmPotato @CabinfeverB

PR #6004, opened 2023-02-16
*: support starting server in API service mode

rleungx:

Issue Number: Ref #5836.

This PR is trying to start the API server when we are going to use microservice architecture. Take TSO as an example: after starting, all TSO-related services will be prevented once we support redirection and the old client's request will be redirected to the TSO service.

```commit-message
```

### Check List

Tests

- Manual test 
Successfully running `pd-server service api --name="api-service" --data-dir="api-service" --client-urls="http://127.0.0.1:23379" --peer-urls="http://127.0.0.1:22380"`

### Release note

```release-note
None
```


rleungx:
@binshi-bing PTAL

server/api/router.go
```diff
@@ -358,6 +352,14 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	registerFunc(clusterRouter, "/admin/unsafe/remove-failed-stores/show",
 		unsafeOperationHandler.GetFailedStoresRemovalStatus, setMethods(http.MethodGet), setAuditBackend(prometheus))
 
+	if !svr.IsAPIMode() {
```
binshi-bing, line 355: What if "pd-server services api,tso XXXX"

rleungx: nice catch. I think the current implementation cannot well support this kind of input, we need to use a flag to specify the mode instead of using subcommand.

server/server.go
```diff
@@ -661,6 +673,19 @@ func (s *Server) stopRaftCluster() {
 	s.cluster.Stop()
 }
 
+// IsServiceEnabled returns whether the service is enabled.
+func (s *Server) IsServiceEnabled(serviceName string) bool {
+	if len(s.services) == 0 {
```
lhy1024, line 678: I think it's a bit of a hack to start all services by default if the length is equal to 0. Perhaps it would be better to pass in all the services names that have been split for the legacy pd server?

rleungx: IMO, they are in different modes, in PD only mode, there is no need to introduce service conception.

server/api/router.go
rleungx: use alias to solve this problem

server/server.go
rleungx: But I will reconsider your suggestion.

nolouch: Maybe we need to know the mode?

server/server.go
```diff
@@ -1465,7 +1494,7 @@ func (s *Server) campaignLeader() {
 	})
 
 	CheckPDVersion(s.persistOptions)
-	log.Info("PD cluster leader is ready to serve", zap.String("pd-leader-name", s.Name()))
+	log.Info("leader is ready to serve", zap.String("leader-name", s.Name()))
```
nolouch, line 1497: If we know the mode, we can log it by the log. sometimes, etcd also log some log similar `leader is`, it's may confuse.

server/server.go
```diff
@@ -506,12 +532,15 @@ func (s *Server) LoopContext() context.Context {
...
 func (s *Server) startServerLoop(ctx context.Context) {
 	s.serverLoopCtx, s.serverLoopCancel = context.WithCancel(ctx)
-	s.serverLoopWg.Add(5)
+	s.serverLoopWg.Add(4)
 	go s.leaderLoop()
 	go s.etcdLeaderLoop()
 	go s.serverMetricsLoop()
-	go s.tsoAllocatorLoop()
 	go s.encryptionKeyManagerLoop()
+	if !s.IsAPIServiceMode() || (s.IsAPIServiceMode() && s.IsServiceEnabled(TSOServiceName)) {
```
lhy1024, line 540: How about adding a function to judge whether enable tso feature?

lhy1024 (commented):
The rest lgtm

cmd/pd-server/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test(t *testing.T) {
+	re := require.New(t)
+
+	output := generateAlias(nil)
+	re.Empty(output)
+	output = generateAlias([]string{"api", "tso", "resource-manager"})
+	re.Len(output, 11)
```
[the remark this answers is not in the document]
rleungx: Yes.

server/server.go
rleungx: IsServiceEnabled is used for that purpose

server/server.go
```diff
@@ -91,8 +91,21 @@ const (
 	idAllocLabel = "idalloc"
 
 	recoveringMarkPath = "cluster/markers/snapshot-recovering"
+	// TSOServiceName is the name of TSO service.
```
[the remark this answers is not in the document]
rleungx: Do you mean to define a type?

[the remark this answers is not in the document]
lhy1024: yes

server/server.go
lhy1024: func (s *server) enableTSO() bool{
    return !s.IsAPIServiceMode() || (s.IsAPIServiceMode() && s.IsServiceEnabled(TSOServiceName))
}

lhy1024: it occurs four times.

server/server.go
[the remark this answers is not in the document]
rleungx: Maybe do it later.

lhy1024 (approved):

rleungx:
@binshi-bing @nolouch Would you please take another look?

server/server.go
```diff
@@ -92,6 +92,11 @@ const (
 	idAllocLabel = "idalloc"
 
 	recoveringMarkPath = "cluster/markers/snapshot-recovering"
+
+	// PDMode represents that server is in PD mode.
+	PDMode = "PD"
+	// APIServiceMode represents that server is in API service mode.
+	APIServiceMode = "API service"
```
binshi-bing, line 99: It needs to support API Service Mode and TSO Service Mode at the same time if TSO service falls back to API service -- how do we represent it?

server/server.go
```diff
@@ -187,14 +192,21 @@ type Server struct {
...
 
 // CreateServer creates the UNINITIALIZED pd server with given configuration.
-func CreateServer(ctx context.Context, cfg *config.Config, legacyServiceBuilders ...HandlerBuilder) (*Server, error) {
-	log.Info("PD Config", zap.Reflect("config", cfg))
+func CreateServer(ctx context.Context, cfg *config.Config, services []string, legacyServiceBuilders ...HandlerBuilder) (*Server, error) {
+	var mode string
+	if len(services) == 0 {
+		mode = PDMode
+	} else {
+		mode = APIServiceMode
```
binshi-bing, line 207: As above -- how do we represent tso service is enabled or not on API service?

server/server.go
```diff
@@ -508,12 +521,14 @@ func (s *Server) LoopContext() context.Context {
...
 func (s *Server) startServerLoop(ctx context.Context) {
 	s.serverLoopCtx, s.serverLoopCancel = context.WithCancel(ctx)
-	s.serverLoopWg.Add(5)
+	s.serverLoopWg.Add(4)
 	go s.leaderLoop()
 	go s.etcdLeaderLoop()
 	go s.serverMetricsLoop()
-	go s.tsoAllocatorLoop()
 	go s.encryptionKeyManagerLoop()
+	// TODO: after support redirect TSO request we need to skip it.
```
binshi-bing, line 529: then no need change this code for now, or you can add a TSO request redirected flag, set it to the default value false, and put the two lines below to the if (! tsoRequestRedirected) block. This code now looks not good.

server/server.go
```diff
@@ -187,14 +192,21 @@ type Server struct {
...
 type HandlerBuilder func(context.Context, *Server) (http.Handler, apiutil.APIServiceGroup, error)
 
 // CreateServer creates the UNINITIALIZED pd server with given configuration.
-func CreateServer(ctx context.Context, cfg *config.Config, legacyServiceBuilders ...HandlerBuilder) (*Server, error) {
-	log.Info("PD Config", zap.Reflect("config", cfg))
+func CreateServer(ctx context.Context, cfg *config.Config, services []string, legacyServiceBuilders ...HandlerBuilder) (*Server, error) {
+	var mode string
+	if len(services) == 0 {
+		mode = PDMode
+	} else {
```
binshi-bing, line 206: where do we do the sanity check to see if 'api' is actually passed in services string array?

binshi-bing (approved):
left few trivial comments

server/server.go
rleungx: For the original PD related starting process, we only support either PD-only mode or API service mode.

server/server.go
rleungx: change it back

server/server.go
rleungx: I use cmd.CalledAs() to do the check. In the current situation, only `api` will be added to the services slice.

server/server.go
binshi-bing: I see

server/server.go
binshi-bing: ok

binshi-bing (approved):
LGTM

nolouch (approved):

PR #6009, opened 2023-02-17
*: fix default config when using config file

rleungx:

Issue Number: Ref #5836.

This PR removes all configs in the command line and lets adjust to set the default value if there is no one.

```commit-message
```

### Check List

Tests

- Manual test 

Use the config file to set `client-urls` and `peer-urls`, then start the server to validate if it is right.

### Release note

```release-note
None.
```


rleungx:
@glorv PTAL

bufferflies (approved):

lhy1024 (approved):

glorv (commented):
LGTM

JmPotato (approved):

PR #6037, opened 2023-02-22
TSO Client implement and provide general client side service discovery framework and general gPRC stream handling framework.

binshi-bing:
Refactor pd client with template design pattern to provide general client side service discovery framework and general gPRC stream batching, forwarding, async and pooling framework.

Issue Number: Ref #5836

```commit-message
Changes:
1. Define the interface BaseClient which is generally for service discovery on a quorum-based cluster or a primary/secondy configured cluster so that the grpc client logic layer can decouple from the server discovery layer.
2. Rename baseClient to pdBaseClient then refactor it to implements BaseClient interface. It provides a basic implementation of service discovery on a quorum-based cluster.
3. Refactor pd client with template design patter to provide general gPRC stream batching, forwarding, async and pooling framework.
4. Add skeleton of tsoBaseClient which is a basic implementation of server discover on a primary/secondary configured cluster.
```

### Check List

Tests

- Unit test
- Manual test

**pd-tso-bench testing result:**
~/code/pingcap/my-pd/bin   tso-integration ●  ./pd-tso-bench -v -duration 5s -pd "[127.0.0.1:3379](http://127.0.0.1:3379/)"                                                                                                                                                                                                1 ↵  4577  17:46:06
Start benchmark #0, duration: 5s
Create 1 client(s) for benchmark
[2023/02/23 17:46:10.110 -08:00] [INFO] [client.go:431] ["[pd] create pd(tso) client with endpoints"] [pd-address="[[127.0.0.1:3379](http://127.0.0.1:3379/)]"]
[2023/02/23 17:46:10.110 -08:00] [INFO] [tso_client.go:336] ["[tso] switch primary"] [new-primary=http://127.0.0.1:3379/] [old-primary=]
[2023/02/23 17:46:10.110 -08:00] [INFO] [client.go:762] ["[pd] tso dispatcher created"] [dc-location=global]
count: 495783, max: 4.3819ms, min: 0.1444ms, avg: 2.0098ms
<1ms: 10410, >1ms: 228832, >2ms: 256541, >5ms: 0, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 481110, max: 4.6088ms, min: 0.0558ms, avg: 2.0063ms
<1ms: 24825, >1ms: 165334, >2ms: 290951, >5ms: 0, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 479289, max: 4.3606ms, min: 0.3493ms, avg: 2.0835ms
<1ms: 515, >1ms: 190880, >2ms: 287894, >5ms: 0, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 478138, max: 4.4927ms, min: 0.3457ms, avg: 2.0902ms
<1ms: 342, >1ms: 185102, >2ms: 292694, >5ms: 0, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
Total:
count: 1934320, max: 4.6088ms, min: 0.0558ms, avg: 2.0471ms
<1ms: 36092, >1ms: 770148, >2ms: 1128080, >5ms: 0, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 1934320, <1ms: 1.87%, >1ms: 39.81%, >2ms: 58.32%, >5ms: 0.00%, >10ms: 0.00%, >30ms: 0.00%, >50ms: 0.00%, >100ms: 0.00%, >200ms: 0.00%, >400ms: 0.00%, >800ms: 0.00%, >1s: 0.00%
P0.5: 2.0442ms, P0.8: 2.2536ms, P0.9: 2.4216ms, P0.99: 3.1541ms

**Remaining issues:**
Killing etcd (remote)backend and the primary in sequence, the secondary doesn't switch to primary. This is because leadership.Watch() in the secondary is stuck. This seems to be a bug in monolithic architecture but uncovered by microservice (detached etcd) scenario.
pd(tso)-client exit isn't clean.

### Release note

```release-note
None.
```


client/base_client.go
```diff
@@ -32,32 +32,110 @@ import (
...
+	CloseClientConns()
+	// GetClusterID returns the ID of the cluster
+	GetClusterID(context.Context) uint64
+	// GetTSOAllocators returns {dc-location -> TSO allocator leader URL} connection map
+	GetTSOAllocators() *sync.Map
+	// GetTSOAllocatorLeaderAddrByDCLocation returns the tso allocator of the given dcLocation
+	GetTSOAllocatorLeaderAddrByDCLocation(dcLocation string) (string, bool)
+	// GetTSOAllocatorClientConnByDCLocation returns the tso allocator grpc client connection
+	// of the given dcLocation
+	GetTSOAllocatorClientConnByDCLocation(dcLocation string) (*grpc.ClientConn, string)
```
JmPotato, line 69: As a base client interface, could we only provide some general methods to build up the specified client upon it? Introducing TSO-related methods here does not sound good to me.

binshi-bing: We should move TSO-related methods out, but we can do it incrementally. The reason why we have TSO-related methods here is that previous baseClient(pdBaseClient) mixed up service discovery and tso related service discovery logic. There is historical reason, now we need to refactor it step by step. @JmPotato

JmPotato: Ok, if the final goal is like that, then this makes sense to me.

binshi-bing: BTW, I'll move CreateTsoStream, ProcessTSORequests and the related private functions to another interface. @JmPotato

client/base_client.go
```diff
@@ -364,32 +477,38 @@ func (c *baseClient) updateURLs(members []*pdpb.Member) {
...
 
-	if _, err := c.getOrCreateGRPCConn(addr); err != nil {
+	if _, err := c.GetOrCreateGRPCConn(addr); err != nil {
 		log.Warn("[pd] failed to connect leader", zap.String("leader", addr), errs.ZapError(err))
 		return err
 	}
 	// Set PD leader and Global TSO Allocator (which is also the PD leader)
 	c.leader.Store(addr)
-	c.allocators.Store(globalDCLocation, addr)
-	c.scheduleUpdateTokenConnection()
```
rleungx, line 387: Previously, we only run `scheduleUpdateTokenConnection`. But now, we do both `scheduleCheckTSODispatcher` and `scheduleUpdateTokenConnection`.

client/base_client.go
```diff
@@ -321,7 +435,6 @@ func (c *baseClient) updateMember() error {
 		if err := c.switchLeader(members.GetLeader().GetClientUrls()); err != nil {
 			return err
 		}
-		c.scheduleCheckTSODispatcher()
```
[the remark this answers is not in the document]
binshi-bing: I didn't remove. Moved to switchLeader actually. We registered a leaderSwitchedCallback which calls c.scheduleCheckTSODispatcher(). Inside of c.switchLeader(), when the leader switched succeeds, all the leaderSwitchedCallbacks will be invoked, including c.scheduleCheckTSODispatcher() and c.scheduleUpdateTokenConnection(). The refactor doesn't change any logic.

client/base_client.go
binshi-bing: As the same answer above. Also pasted below:
I didn't remove. Moved to switchLeader actually. We registered a leaderSwitchedCallback which calls c.scheduleCheckTSODispatcher(). Inside of c.switchLeader(), when the leader switched succeeds, all the leaderSwitchedCallbacks will be invoked, including c.scheduleCheckTSODispatcher() and c.scheduleUpdateTokenConnection(). The refactor doesn't change any logic.

binshi-bing: The refactor didn't intend to change any logic, but I indeed find there is difference introduced by my change. scheduleUpdateTokenConnection is called when switchLeader() returns nil which have two cases: 
1. the addr switched to is equal to the existing leader, so no leader switching happens in this case; L376
2. successfully switched leader to another host

With my change, if addr == oldLeader L376,   scheduleUpdateTokenConnection isn't called. ** if there is no change, do we need to tryResourceManagerConnect? My understanding is no. @JmPotato to confirm**
After leader successfully being switched, both scheduleUpdateTokenConnection and scheduleCheckTSODispatcher are invoked.

client/tso_client.go
```diff
@@ -55,5 +64,216 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
 // without block the caller.
 // TODO: implement the following API
 func (c *client) GetLocalTSWithinKeyspaceAsync(ctx context.Context, dcLocation string, keyspaceID uint32) TSFuture {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span = opentracing.StartSpan("GetLocalTSAsync", opentracing.ChildOf(span.Context()))
```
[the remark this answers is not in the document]
binshi-bing: good catch. Changed locally.

client/tso_client.go
```diff
@@ -55,5 +64,216 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
...
+func (c *tsoBaseClient) createTsoStreamInternal(ctx context.Context, cancel context.CancelFunc, client tsopb.TSOClient) (interface{}, error) {
+	done := make(chan struct{})
+	// TODO: we need to handle a conner case that this goroutine is timeout while the stream is successfully created.
+	go c.checkStreamTimeout(ctx, cancel, done)
+	stream, err := client.Tso(ctx)
+	done <- struct{}{}
+	return stream, err
+}
+
+func (c *tsoBaseClient) checkStreamTimeout(ctx context.Context, cancel context.CancelFunc, done chan struct{}) {
```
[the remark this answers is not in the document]
binshi-bing: changed. Is this necessary? it uses c.option.timeout

client/base_client.go
binshi-bing: @JmPotato @rleungx Could you please help to double check I**f there is no leader change, do we need to tryResourceManagerConnect?**

JmPotato: There is no need to reconnect the resource manager if the leader remains the same.

binshi-bing: Thanks for the confirmation! Then my refactor should be ok, though it changed the previous logic that tryResourceManagerConnect was triggered even the leader kept the same.

client/base_client.go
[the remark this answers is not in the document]
binshi-bing: I shouldn't move. After it's moved to inside of switchLeader(), it won't be called if any local tso allocator leader changed, which caused some tests to fail. I have added the equivalent logic back in this commit https://github.com/tikv/pd/pull/6037/commits/6ce5806648f3c416f33f7c01ffae0205f8f902fb.

client/base_client.go
binshi-bing: As mentioned above in another thread (also pasted below), I have moved scheduleCheckTSODispatcher back.

_I shouldn't move. After it's moved to inside of switchLeader(), it won't be called if any local tso allocator leader changed, which caused some tests to fail. I have added the equivalent logic back in this commit https://github.com/tikv/pd/commit/6ce5806648f3c416f33f7c01ffae0205f8f902fb._

client/grpcutil/grpcutil.go
```diff
@@ -64,3 +72,29 @@ func BuildForwardContext(ctx context.Context, addr string) context.Context {
...
+	if ok {
+		return conn.(*grpc.ClientConn), nil
+	}
+	tc, err := tlsCfg.ToTLSConfig()
+	if err != nil {
+		return nil, err
+	}
+	dCtx, cancel := context.WithTimeout(ctx, dialTimeout)
+	defer cancel()
+	cc, err := GetClientConn(dCtx, addr, tc, do...)
```
lhy1024, line 89: how about renaming to reduce ambiguity？
```suggestion
	cc, err := GetClientConn(dCtx, addr, tls, opt...)
```

client/base_client.go
```diff
@@ -39,63 +39,119 @@ const (
...
+	GetBackupEndpointsAddrs() []string
+	// GetOrCreateGRPCConn returns the corresponding grpc client connection of the given addr
+	GetOrCreateGRPCConn(addr string) (*grpc.ClientConn, error)
+	// ScheduleCheckIfMembershipChanged is used to trigger a check to see if there is any
+	// membership change among the leader/followers in a quorum-based cluster or among
+	// the primary/secondaries in a primary/secondy configured cluster.
+	ScheduleCheckIfMembershipChanged()
+	// Immediately checkif there is any membership change among the leader/followers in a
+	// quorum-based cluster or among the primary/secondaries in a primary/secondy configured cluster.
+	CheckIfMembershipChanged() error
```
lhy1024, line 90: their name is so long, how about `ScheduleCheckMemberChanged` and `CheckMemberChanged`

client/base_client.go
```diff
@@ -39,63 +39,119 @@ const (
...
+	// which is the leader in a quorum-based cluster or the primary in a primary/secondy
+	// configured cluster.
+	GetServingEndpointClientConn() *grpc.ClientConn
+	// GetServingEndpointAddr returns the serving endpoint which is the leader
+	// in a quorum-based cluster or the primary in a primary/secondy configured cluster.
+	GetServingEndpointAddr() string
+	// GetBackupEndpointsAddrs gets the addresses of the current reachable and healthy
+	// backup service endpoints randomly. Backup service endpoints are followers in a
+	// quorum-based cluster or secondaries in a primary/secondary configured cluster.
+	GetBackupEndpointsAddrs() []string
```
lhy1024, line 81: I think `endpoint` of name is unnecessary, current name is so long. How about `GetServingAddr` and `GetBackupAddrs`?

client/base_client.go
```diff
@@ -39,63 +39,119 @@ const (
...
+	// AddServiceEndpointSwitchedCallback adds callbacks which will be called when the leader
+	// in a quorum-based cluster or the primary in a primary/secondary configured cluster
+	// is switched.
+	AddServiceEndpointSwitchedCallback(callbacks ...func())
+	// AddServiceEndpointsChangedCallback adds callbacks which will be called when any leader/follower
+	// in a quorum-based cluster or any primary/secondary in a primary/secondary configured cluster is changed.
+	AddServiceEndpointsChangedCallback(callbacks ...func())
+	// AddTSOAllocatorServiceEndpointSwitchedCallback adds callbacks which will be called
+	// when any global/local tso allocator service endpoint is switched.
+	AddTSOAllocatorServiceEndpointSwitchedCallback(callbacks ...func())
```
lhy1024, line 100: ditto

client/base_client.go
binshi-bing: No. It's to schedule a check instead of checking immediatelly.

binshi-bing: I'll Change from ScheduleCheckIfMembershipChanged to ScheduleCheckMemberChanged

client/grpcutil/grpcutil.go
binshi-bing: changed.

client/base_client.go
binshi-bing: changed.

client/base_client.go
binshi-bing: changed.

client/tso_client.go
```diff
@@ -55,5 +69,233 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
...
+	if err != nil {
+		return nil
+	}
+
+	return bc
+}
+
+// Init initialize the concrete client underlying
+func (c *tsoBaseClient) Init() error {
+	return nil
```
lhy1024, line 153: do nothing?

client/tso_client.go
```diff
@@ -55,5 +69,233 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
...
+}
+
+// Close all grpc client connnections
+func (c *tsoBaseClient) CloseClientConns() {
+	c.clientConns.Range(func(_, cc interface{}) bool {
+		if err := cc.(*grpc.ClientConn).Close(); err != nil {
+			log.Error("[pd] failed to close gRPC clientConn", errs.ZapError(errs.ErrCloseGRPCConn, err))
```
lhy1024, line 160: [pd]?

client/tso_client.go
```diff
@@ -55,5 +69,233 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
...
+	}
+	return url.(string), true
+}
+
+// GetTSOAllocatorClientConnByDCLocation returns the tso allocator grpc client connection
+// of the given dcLocation
+func (c *tsoBaseClient) GetTSOAllocatorClientConnByDCLocation(dcLocation string) (*grpc.ClientConn, string) {
+	url, ok := c.tsoAllocators.Load(dcLocation)
+	if !ok {
+		panic(fmt.Sprintf("the allocator leader in %s should exist", dcLocation))
```
lhy1024, line 196: how about log.Fatal?

client/tso_client.go
```diff
@@ -55,5 +69,233 @@ func (c *client) GetTSWithinKeyspaceAsync(ctx context.Context, keyspaceID uint32
...
+	}
+	cc, ok := c.clientConns.Load(url)
+	if !ok {
+		panic(fmt.Sprintf("the client connection of %s in %s should exist", url, dcLocation))
```
lhy1024, line 200: ditto?

client/tso_request_dispatcher.go
```diff
@@ -0,0 +1,751 @@
...
+		// cancel the forward stream
+		forwardCancel()
+		requestForwarded.WithLabelValues(forwardedHostTrim, addrTrim).Set(0)
+	}()
+	cc, u := c.bc.GetTSOAllocatorClientConnByDCLocation(dc)
+	healthCli := healthpb.NewHealthClient(cc)
+	for {
+		// the pd/allocator leader change, we need to re-establish the stream
+		if u != url {
+			log.Info("[pd] the leader of the allocator leader is changed", zap.String("dc", dc), zap.String("origin", url), zap.String("new", u))
```
lhy1024, line 235: also [pd]?

client/tso_request_dispatcher.go
```diff
@@ -0,0 +1,751 @@
...
+		if err == nil && resp.GetStatus() == healthpb.HealthCheckResponse_SERVING {
+			streamBuilders[addr] = c.tsoStreamBuilderFactory.makeBuilder(cc)
+		}
+	}
+	return streamBuilders
+}
+
+// TryConnectToTSOWithProxy will create multiple streams to all the service endpoints to work as
+// a TSO proxy to reduce the pressure of the main serving service endpoint.
+func (c *client) TryConnectToTSOWithProxy(dispatcherCtx context.Context, dc string, connectionCtxs *sync.Map) error {
```
lhy1024, line 635: need to export?

client/tso_stream.go
```diff
@@ -0,0 +1,176 @@
...
+
+	physical, logical, suffixBits = resp.GetTimestamp().GetPhysical(), resp.GetTimestamp().GetLogical(), resp.GetTimestamp().GetSuffixBits()
+	return
+}
+
+type tsoTSOStream struct {
+	stream tsopb.TSO_TsoClient
+}
+
+func (s *tsoTSOStream) processRequests(clusterID uint64, dcLocation string, requests []*tsoRequest,
```
lhy1024, line 144: this two function is so similiar

client/tso_request_dispatcher.go
```diff
@@ -0,0 +1,751 @@
...
+	// Same as above, we save the largest logical part here.
+	lastTSOPointer.logical = largestLogical
+}
+
+func tsLessEqual(physical, logical, thatPhysical, thatLogical int64) bool {
+	if physical == thatPhysical {
+		return logical <= thatLogical
+	}
+	return physical < thatPhysical
+}
```
lhy1024, line 741: if move to tso util, it will be better

client/client.go
```diff
@@ -396,39 +253,132 @@ type client struct {
...
+	// Register callbacks and start the daemons.
+	if enableTSO {
+		c.bc.AddTSOAllocatorServingAddrSwitchedCallback(c.scheduleCheckTSODispatcher)
+		c.bc.AddServiceAddrsSwitchedCallback(c.scheduleUpdateTSOConnectionCtxs)
+		c.updateTSODispatcher()
+		c.wg.Add(2)
+		go c.tsLoop()
+		go c.tsCancelLoop()
+	}
+	if enableAdmissionCtl {
```
lhy1024, line 353: what does it mean?

lhy1024 (commented):
diff so much, we need to review it carefully

client/tso_request_dispatcher.go
binshi-bing: there is only one place to use these functions after refactor, so it isn't necessary to move it to tso util for now, is it?

client/client.go
binshi-bing: tso mcs mode doesn't need to enable admisstion control module, so this flag is needed.

binshi-bing: Do you mean we can have a better name, such as enableAdmissionControl?

client/client.go
```diff
@@ -396,39 +253,132 @@ type client struct {
...
+	return c, nil
+}
+
+// NewTSOClientWithContext creates a TSO client with context.
+func NewTSOClientWithContext(ctx context.Context, svrAddrs []string, security SecurityOption, opts ...ClientOption) (Client, error) {
+	log.Info("[pd(tso)] create tso client with endpoints", zap.Strings("tso-address", svrAddrs))
+	c, clientCtx, clientCancel, tlsCfg := createClient(ctx, &security)
+	c.tsoStreamBuilderFactory = &tsoTSOStreamBuilderFactory{}
+	c.bc = newTSOBaseClient(clientCtx, clientCancel, &c.wg, addrsToUrls(svrAddrs), tlsCfg, c.option)
+	if err := c.setup(true, false, opts...); err != nil {
```
rleungx, line 303: I prefer not to use options. The client shouldn't know anything about if the tso or resource manager is enabled.

rleungx: I didn't notice here previously.

client/tso_stream.go
binshi-bing: It's not easy to remove duplicate code now, because the request in stream.Send(), the stream, the response from stream.Recv() are in different types from pdpb and tsopb.  We have used several design patterns, including template, Abstract Factory, Builder as you can see in base_client.go and tso_stream.go, so that we can share the service discovery code and tso batching/forwarding/async framework. These two functions are closest to the protobuf code and only different in types so it is even not possible to dedup further.

client/base_client.go
binshi-bing: Moved to tso_stream. Template, Abstract Factor, Builder design patterns are used to make it possible.

client/grpcutil/grpcutil.go
```diff
@@ -64,3 +72,29 @@ func BuildForwardContext(ctx context.Context, addr string) context.Context {
...
 }
+
+// GetOrCreateGRPCConn returns the corresponding grpc client connection of the given addr.
+// Returns the old one if's already  existed in the clientConns; otherwise creates a new one and returns it.
+func GetOrCreateGRPCConn(ctx context.Context, clientConns *sync.Map, addr string, tlsCfg *tlsutil.TLSConfig, opt ...grpc.DialOption) (*grpc.ClientConn, error) {
+	conn, ok := clientConns.Load(addr)
+	if ok {
+		return conn.(*grpc.ClientConn), nil
+	}
+	tc, err := tlsCfg.ToTLSConfig()
```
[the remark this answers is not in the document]
binshi-bing: tls is the package name and mightn't be appropriate for the config name. tlsCfg is already used for tlsCfg *tlsutil.TLSConfig passed to this function as func param. I changed the name to tlsConfig.

client/base_client.go
```diff
@@ -39,63 +39,120 @@ const (
...
+	GetTSOAllocators() *sync.Map
+	// GetTSOAllocatorServingAddrByDCLocation returns the tso allocator of the given dcLocation
+	GetTSOAllocatorServingAddrByDCLocation(dcLocation string) (string, bool)
+	// GetTSOAllocatorClientConnByDCLocation returns the tso allocator grpc client connection
+	// of the given dcLocation
+	GetTSOAllocatorClientConnByDCLocation(dcLocation string) (*grpc.ClientConn, string)
+	// GetServingEndpointClientConn returns the grpc client connection of the serving endpoint
+	// which is the leader in a quorum-based cluster or the primary in a primary/secondy
+	// configured cluster.
+	GetServingEndpointClientConn() *grpc.ClientConn
```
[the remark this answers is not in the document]
binshi-bing: ServingClientConn doesn't sound the serving endpont's client connection.

client/tso_client.go
binshi-bing: This is the old code introduced by previous commit made 2 years ago.

client/tso_client.go
binshi-bing: These functions are used by both tso APIs in pd client and tso APIs in the independent tso client. I changed name to [pd/tso]

client/tso_request_dispatcher.go
binshi-bing: These functions are used by both tso APIs in pd client and tso APIs in the independent tso client. I changed name to [pd/tso]

binshi-bing: Also changed every [pd] to [pd/tso] in this file.

client/client.go
binshi-bing: I remove the flags for now. The original purpose of adding these flags is to disable some clients when the server side doesn't provide the correspoinding services. For example, disable tso client when the server side is resource management microservice, or disable resource management client when the server side the tso microservice.

client/client.go
binshi-bing: I removed the flags for now. The original purpose of adding these flags is to disable some clients when the server side doesn't provide the correspoinding services. For example, disable tso client when the server side is resource management microservice, or disable resource management client when the server side the tso microservice.

lhy1024 (approved):
mostly LGTM

client/tso_client.go
binshi-bing: Haven't integrated with service discovery logic, so do nothing for now.

nolouch (commented):
looks awesome!

PR #6040, opened 2023-02-23
mcs: register resource manager to etcd

rleungx:
This PR is going to register the resource manager into etcd after starting.

Issue Number: Ref #5766.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resource_manager/server/server.go
+++ b/pkg/mcs/resource_manager/server/server.go
@@ -50,6 +50,7 @@ const (
 	defaultGRPCGracefulStopTimeout = 5 * time.Second
 	// defaultHTTPGracefulShutdownTimeout is the default timeout to wait for http server to gracefully shutdown
 	defaultHTTPGracefulShutdownTimeout = 5 * time.Second
+	defaultLeaseInSeconds              = 3
 )
 
 // Server is the resource manager server, and it implements bs.Server.
@@ -283,8 +284,7 @@ func (s *Server) startServer() error {
 
 	// Server has started.
 	atomic.StoreInt64(&s.isServing, 1)
-	// register to etcd
-	s.serviceRegister = discovery.NewServiceRegister(s.ctx, s.etcdClient, "resource_manager", s.cfg.ListenAddr, s.cfg.ListenAddr, 3)
+	s.serviceRegister = discovery.NewServiceRegister(s.ctx, s.etcdClient, "resource_manager", s.cfg.ListenAddr, s.cfg.ListenAddr, defaultLeaseInSeconds)
 	s.serviceRegister.Register()
 	return nil
 }
--- a/pkg/mcs/resource_manager/server/server.go
+++ b/pkg/mcs/resource_manager/server/server.go
@@ -33,6 +33,7 @@ import (
 	"github.com/soheilhy/cmux"
 	"github.com/spf13/cobra"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/mcs/discovery"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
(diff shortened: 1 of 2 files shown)


pkg/mcs/resource_manager/server/server.go
```diff
@@ -275,6 +279,9 @@ func (s *Server) startServer() error {
 
 	// Server has started.
 	atomic.StoreInt64(&s.isServing, 1)
+	// register to etcd
+	s.serviceRegister = discovery.NewServiceRegister(s.ctx, s.etcdClient, "resource_manager", s.cfg.ListenAddr, s.cfg.ListenAddr, 3)
```
lhy1024, line 283: 3？

pkg/mcs/resource_manager/server/server.go
```diff
@@ -275,6 +279,9 @@ func (s *Server) startServer() error {
 
 	// Server has started.
 	atomic.StoreInt64(&s.isServing, 1)
+	// register to etcd
```
lhy1024, line 282: do we also need to register when running legacy server?

lhy1024 (commented):
the rest LGTM

pkg/mcs/resource_manager/server/server.go
rleungx: 3 seconds

pkg/mcs/resource_manager/server/server.go
rleungx: No, it will only register when resource manager is started as a independent binary.

lhy1024 (approved):

pkg/mcs/resource_manager/server/server.go
bufferflies: how about refactor the function's signature for the ttl unit, and make it more clear.

lhy1024: If more clear, it will be better.

pkg/mcs/resource_manager/server/server.go
bufferflies: I think this comment is useless.

tests/mcs/resource_manager/server_test.go
```diff
@@ -100,3 +101,39 @@ func TestResourceManagerServer(t *testing.T) {
...
+
+	err = cluster.RunInitialServers()
+	re.NoError(err)
+
+	leaderName := cluster.WaitLeader()
+	leader := cluster.GetServer(leaderName)
+
+	cfg := rm.NewConfig()
+	cfg.BackendEndpoints = leader.GetAddr()
+	cfg.ListenAddr = "127.0.0.1:8087"
```
bufferflies, line 122: What does it happen if the address has been bound?

bufferflies (commented):
reset LGTM

pkg/mcs/resource_manager/server/server.go
rleungx: How about using a constant? like leaseInSeconds? I just keep consistent with etcd.

pkg/mcs/resource_manager/server/server.go
rleungx: I will remove it.

tests/mcs/resource_manager/server_test.go
rleungx: I copied from `TestResourceManagerServer`, seems it also has the problem. @lhy1024

lhy1024: I will fix it

lhy1024: replace it with `tempurl.Alloc()` in #6047

bufferflies (approved):

PR #6047, opened 2023-02-24
resource_manager: make test stable

lhy1024:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/server/server.go
```diff
@@ -250,6 +250,8 @@ func (s *Server) startServer() error {
 	if err != nil {
 		return err
 	}
+	s.cfg.ListenAddr = strings.TrimPrefix(s.cfg.ListenAddr, "http://")
+	s.cfg.ListenAddr = strings.TrimPrefix(s.cfg.ListenAddr, "https://")
```
rleungx, line 254: no need to do it

lhy1024: use more general function

bufferflies (approved):

rleungx (approved):

PR #6041, opened 2023-02-23
*: write RU model config into the storage

JmPotato:
### What problem does this PR solve?

Issue Number: ref #6038.

### What is changed and how does it work?

```commit-message
- Introduce the RU model config.
- Write the RU model config into the storage while initializing the resource manager.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

Issue #6038, opened 2023-02-23
Allow to configure the RU cost model globally


JmPotato:
It would be more convenient if we could configure the RU cost model directly via the config file or API. This issue is used to track the related implementation tasks.

- [ ] Introduce the config and save it into the storage when the resource manager starts. 
- [ ] Read from the remote storage when initializing the controller client.

JmPotato:
/cc @nolouch @CabinfeverB @HuSharp 

Please help take a look also, thx.

PR #6048, opened 2023-02-24
resource_controller: load RU config from the remote server when initializing controller

JmPotato:

Issue Number: ref #6038. Should merge after #6041.

```commit-message
- Load RU config from the remote server when initializing controller. 
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/resource_group/controller/controller.go
+++ b/client/resource_group/controller/controller.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"math"
 	"sync"
 	"sync/atomic"
@@ -24,11 +25,13 @@ import (
 	"github.com/pingcap/errors"
 	rmpb "github.com/pingcap/kvproto/pkg/resource_manager"
 	"github.com/pingcap/log"
+	pd "github.com/tikv/pd/client"
 	"github.com/tikv/pd/client/errs"
 	"go.uber.org/zap"
 )
 
 const (
+	requestUnitConfigPath  = "resource_group/ru_config"
 	defaultMaxWaitDuration = time.Second
 	maxRetry               = 3
 	maxNotificationChanLen = 200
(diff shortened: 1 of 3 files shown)


JmPotato:
/cc @HuSharp

Please help take a look also.

client/resource_group/controller/controller.go
```diff
@@ -91,14 +95,20 @@ type ResourceGroupsController struct {
...
+	provider ResourceGroupProvider,
+	requestUnitConfig *RequestUnitConfig,
+) (*ResourceGroupsController, error) {
+	var err error
+	if requestUnitConfig == nil {
+		if requestUnitConfig, err = loadRequestUnitConfig(ctx, provider); err != nil {
+			return nil, err
+		}
+	}
+	// TODO: do we need to reload the config from remote server periodically?
```
HuSharp, line 110: Do u mean we need watch it?

JmPotato: Yes, but since the RU model config won't be modified frequently, I think it's ok to leave this as a nice-to-have option.

CabinfeverB, line 110: IMO, reload is not required for now, and the cluster control surface can restart tidb

CabinfeverB (approved):

client/resource_group/controller/controller.go
```diff
@@ -91,14 +95,20 @@ type ResourceGroupsController struct {
...
+func NewResourceGroupController(
+	ctx context.Context,
+	clientUniqueID uint64,
+	provider ResourceGroupProvider,
+	requestUnitConfig *RequestUnitConfig,
+) (*ResourceGroupsController, error) {
+	var err error
+	if requestUnitConfig == nil {
+		if requestUnitConfig, err = loadRequestUnitConfig(ctx, provider); err != nil {
+			return nil, err
```
HuSharp, line 107: If we don't need to reload, maybe we can move the error in if block.

client/resource_group/controller/controller.go
nolouch: Maybe we can consider adding an API to the triage reload. it's more "manual" and no need to restart the node.

client/resource_group/controller/controller.go
JmPotato: This is because `requestUnitConfig` has been defined as the function parameter, we can not re-assign it with `:=`.

client/resource_group/controller/controller.go
JmPotato: I just removed this TODO.

client/resource_group/controller/controller.go
HuSharp: I mean what about 
```
	if requestUnitConfig == nil {
	var err error
```

JmPotato: Done.

HuSharp (approved):
LGTM!

nolouch (approved):

PR #6063, opened 2023-02-28
resource_manager: add degraded mode

CabinfeverB:

Issue Number: ref #5851 

Add a timer after each token request is sent. If the timer does not return successfully after one second, the controller will enter degraded mode. In degraded mode, a resource group in low-token process will receive the same fill rate as the configured.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

### Release note

```release-note
None.
```


kevin-xianliu:
We can have a switch to enable/disable degraded mode.

client/resource_group/controller/controller.go
```diff
@@ -297,6 +320,10 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
 		Requests:              requests,
 		TargetRequestPeriodMs: uint64(defaultTargetPeriod / time.Millisecond),
 	}
+	if c.run.responseDeadline == nil {
+		c.run.responseDeadline = time.NewTimer(time.Second)
```
JmPotato, line 324: I suggest using `Stop()` and `Rest()` here to re-use the timer as much as possible rather than creating a new one each time.

client/resource_group/controller/config.go
```diff
@@ -104,23 +118,26 @@ type Config struct {
...
 func DefaultConfig() *Config {
 	return GenerateConfig(
 		DefaultRequestUnitConfig(),
+		DefaultRMServerConfig(),
 	)
 }
 
 // GenerateConfig generates the configuration by the given request unit configuration.
-func GenerateConfig(ruConfig *RequestUnitConfig) *Config {
+func GenerateConfig(ruConfig *RequestUnitConfig, rmServerConfig *RMServerConfig) *Config {
```
JmPotato, line 133: Can we merge the `ruConfig` and `rmServerConfig` into one?

client/resource_group/controller/controller.go
```diff
@@ -210,6 +247,7 @@ func (c *ResourceGroupsController) Stop() error {
 		return errors.Errorf("resource groups controller does not start")
 	}
 	c.loopCancel()
+	c.run.responseDeadline.Stop()
```
JmPotato, line 250: Is this necessary since there is a `defer` already before?

client/resource_group/controller/controller.go
```diff
@@ -112,14 +117,17 @@ func NewResourceGroupController(
 	requestUnitConfig *RequestUnitConfig,
```
JmPotato, line 117: Do we still need this parameter?

client/resource_group/controller/controller.go
```diff
@@ -31,7 +32,7 @@ import (
 )
 
 const (
-	requestUnitConfigPath  = "resource_group/ru_config"
+	controllerConfigPath   = "resource_group/control"
```
JmPotato, line 35: Maybe a little bit more specific would be better.

```suggestion
	controllerConfigPath   = "resource_group/controller_config"
```

client/resource_group/controller/config.go
```diff
@@ -61,8 +61,28 @@ const (
...
 )
 
+// ControllerConfig is the configuration of the resource manager controller which includes some option for client needed.
+type ControllerConfig struct {
+	// EnableDegradedMode is to control whether resource control client enable degraded mode when server is disconnect.
+	EnableDegradedMode bool `toml:"enable-degraded-mode" json:"enable-degraded-mode"`
+
+	// RequestUnit is the configuration determines the coefficients of the RRU and WRU cost.
+	// This configuration should be modified carefully.
+	RequestUnit RequestUnitConfig
```
JmPotato, line 75: Where are the toml and json tags?

pkg/mcs/resource_manager/server/config.go
```diff
@@ -76,11 +78,31 @@ type Config struct {
...
+}
+
+// ControllerConfig is the configuration of the resource manager controller which includes some option for client needed.
+type ControllerConfig struct {
+	// EnableDegradedMode is to control whether resource control client enable degraded mode when server is disconnect.
+	EnableDegradedMode bool `toml:"enable-degraded-mode" json:"enable-degraded-mode"`
+
 	// RequestUnit is the configuration determines the coefficients of the RRU and WRU cost.
 	// This configuration should be modified carefully.
 	RequestUnit RequestUnitConfig
```
JmPotato, line 91: Ditto.

client/resource_group/controller/controller.go
CabinfeverB: I can't decide if we need to use the specified config on the client side, so I keep it.

client/resource_group/controller/controller.go
CabinfeverB: Do we need the suffix?  Other config key path don't have "_config".

client/resource_group/controller/controller.go
```diff
@@ -161,6 +169,9 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 	c.initRunState()
 	c.loopCtx, c.loopCancel = context.WithCancel(ctx)
 	go func() {
+		c.run.responseDeadline = time.NewTimer(time.Second)
+		c.run.responseDeadline.Stop()
```
nolouch, line 173: remove this line?

client/resource_group/controller/controller.go
```diff
@@ -344,6 +381,10 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
 		Requests:              requests,
 		TargetRequestPeriodMs: uint64(defaultTargetPeriod / time.Millisecond),
 	}
+	if c.responseDeadlineCh == nil {
+		c.run.responseDeadline.Reset(time.Second)
```
nolouch, line 385: How about considering making the deadline timeout be configurable?

client/resource_group/controller/controller.go
```diff
@@ -309,6 +336,16 @@ func (c *ResourceGroupsController) updateAvgRequestResourcePerSec() {
 }
 
 func (c *ResourceGroupsController) handleTokenBucketResponse(resp []*rmpb.TokenBucketResponse) {
+	if c.responseDeadlineCh != nil {
+		if c.run.responseDeadline.Stop() {
+			select {
+			case <-c.run.responseDeadline.C:
```
nolouch, line 342: why need this line?

client/resource_group/controller/controller.go
CabinfeverB: ref https://github.com/tikv/pd/pull/6063#discussion_r1122573946, we new a Timer but don't need it send Time channel at beginning.

client/resource_group/controller/controller.go
CabinfeverB: Good point

client/resource_group/controller/controller.go
CabinfeverB: As the comment of `Timer.Stop()`, 
> Stop does not close the channel, to prevent a read from the channel succeedingincorrectly. To ensure the channel is empty after a call to Stop, check the return value and drain the channel.

client/resource_group/controller/controller.go
```diff
@@ -195,6 +212,8 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 				c.updateAvgRequestResourcePerSec()
 				if !c.run.requestInProgress {
 					c.collectTokenBucketRequests(c.loopCtx, "low_ru", true /* only select low tokens resource group */)
+				} else if c.run.inDegradedMode {
```
nolouch, line 215: I wonder not in else may avoid some unexpect in here.
```
if c.run.inDegradedMode
```

client/resource_group/controller/controller.go
```diff
@@ -701,6 +757,37 @@ func (gc *groupCostController) handleRUTokenResponse(resp *rmpb.TokenBucketRespo
...
+		failpoint.Inject("degradedModeRU", func() {
+			cfg.NewRate = 99999999
+		})
+		counter.limiter.Reconfigure(gc.run.now, cfg, resetLowProcess())
+	}
+}
+
+func (gc *groupCostController) applyBasicConfigForRawResourceTokenCounter() {
+	for typ, counter := range gc.run.resourceTokens {
+		if !counter.limiter.IsLowTokens() {
```
nolouch, line 779: maybe we can not skip it to make sure be in downgrade, or add log here.

nolouch (commented):
rest lgtm

client/resource_group/controller/controller.go
CabinfeverB: I want low-token buckets to enter degraded mode. It will log if successfully enter degraded mode

client/resource_group/controller/controller.go
```diff
@@ -112,14 +117,17 @@ func NewResourceGroupController(
...
+	if err != nil {
+		return nil, err
+	}
+	if requestUnitConfig != nil {
+		controllerConfig.RequestUnit = *requestUnitConfig
 	}
-	config := GenerateConfig(requestUnitConfig)
+	if err != nil {
+		return nil, err
+	}
```
HuSharp, line 129: It looks duplicated

nolouch (approved):
lgtm

pkg/storage/endpoint/key_path.go
```diff
@@ -45,7 +45,7 @@ const (
 	// resource group storage endpoint has prefix `resource_group`
 	resourceGroupSettingsPath = "settings"
 	resourceGroupStatesPath   = "states"
-	requestUnitConfigPath     = "ru_config"
+	requestUnitConfigPath     = "controller"
```
HuSharp, line 48: Do we also need to change this `requestUnitConfigPath` name?

pkg/storage/endpoint/resource_group.go
```diff
@@ -61,7 +61,7 @@ func (se *StorageEndpoint) LoadResourceGroupStates(f func(k, v string)) error {
 	return se.loadRangeByPrefix(resourceGroupStatesPath+"/", f)
 }
 
-// SaveRequestUnitConfig stores the request unit config to storage.
-func (se *StorageEndpoint) SaveRequestUnitConfig(config interface{}) error {
+// SaveControllerConfig stores the request unit config to storage.
```
HuSharp, line 64: ditto for comment.

HuSharp (approved):
rest LGTM

nolouch (approved):
lgtm

Issue #6069, opened 2023-03-01
PD data races report


hnes:
## Bug Report

There are some data races in the code.

### What did you do?

```
$ WITH_RACE=1 make build
# and run benchmark
```

### What did you expect to see?

No data race warning.

### What did you see instead?

Git Commit Hash: f999ad5287d4310d4689989902d96d16e4f977ad https://github.com/tikv/pd/tree/master

```
WARNING: DATA RACE
Write at 0x00c0007f44a0 by main goroutine:
  github.com/tikv/pd/server.(*Server).startServer()
      /x/pd/server/server.go:337 +0xcd
  github.com/tikv/pd/server.(*Server).Run()
      /x/pd/server/server.go:480 +0x145
  main.createServerWrapper()
      /x/pd/cmd/pd-server/main.go:201 +0xd92
  github.com/spf13/cobra.(*Command).execute()
      /x/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0xb53
  github.com/spf13/cobra.(*Command).ExecuteC()
      /x/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x5da
  github.com/spf13/cobra.(*Command).Execute()
      /x/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887 +0x633
  main.main()
      /x/pd/cmd/pd-server/main.go:73 +0x634

Previous read at 0x00c0007f44a0 by goroutine 417:
  github.com/tikv/pd/server.(*GrpcServer).errorHeader()
      /x/pd/server/grpc_service.go:1520 +0x1c8
  github.com/tikv/pd/server.(*GrpcServer).wrapErrorToHeader()
      /x/pd/server/grpc_service.go:92 +0x7f
  github.com/tikv/pd/server.(*GrpcServer).GetMembers()
      /x/pd/server/grpc_service.go:105 +0x69
  github.com/pingcap/kvproto/pkg/pdpb._PD_GetMembers_Handler.func1()
      /x/go/pkg/mod/github.com/pingcap/kvproto@v0.0.0-20230228041042-1e9aca94bab6/pkg/pdpb/pdpb.pb.go:8178 +0x88
  github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1()
[...]

PR #5310, opened 2022-07-14
*: put gRPC unknown error into the header.

HuSharp:
Signed-off-by: husharp <jinhao.hu@pingcap.com>

### What problem does this PR solve?
https://github.com/tikv/pd/issues/5161

Issue Number: Close #5161 

### What is changed and how does it work?

```commit-message
put the unknown error into the header instead of directly returning a gRPC error.
```

### Check List

Tests
- Unit test

### Release note

```release-note
Put the unknown error into the header instead of directly returning a gRPC error.
And modify related uint tests and code.
```

PR #6070, opened 2023-03-02
*: fix member race problem

rleungx:

Issue Number: Ref #6069.

#5310 did some changes to the header but for GetMembers, we may encounter a kind of situation: The server hasn't served but the request is coming. Because the cluster id could not be initialized, there might be a data race when reading/writing the cluster id. 

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


rleungx:
@HuSharp PTAL

server/grpc_service.go
```diff
@@ -101,9 +101,7 @@ func (s *GrpcServer) GetMembers(context.Context, *pdpb.GetMembersRequest) (*pdpb
 	// at startup and needs to get the cluster ID with the first request (i.e. GetMembers).
 	members, err := s.Server.GetMembers()
 	if err != nil {
-		return &pdpb.GetMembersResponse{
-			Header: s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN, err.Error()),
-		}, nil
+		return nil, status.Errorf(codes.Unknown, err.Error())
```
JmPotato, line 104: I think TiKV may relays on this error to retry, changing it to a gRPC error seems will change the behavior.

rleungx: Yes, updated.

server/grpc_service.go
```diff
@@ -102,7 +102,12 @@ func (s *GrpcServer) GetMembers(context.Context, *pdpb.GetMembersRequest) (*pdpb
 	members, err := s.Server.GetMembers()
 	if err != nil {
 		return &pdpb.GetMembersResponse{
-			Header: s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN, err.Error()),
+			Header: &pdpb.ResponseHeader{
+				Error: &pdpb.Error{
```
JmPotato, line 106: Will the missing cluster ID field be a problem?

JmPotato (approved):
Basically LGTM.

server/grpc_service.go
rleungx: I distinguish the two situations in the original `s.Server.GetMembers()`.

server/grpc_service.go
```diff
@@ -99,7 +99,17 @@ func (s *GrpcServer) wrapErrorToHeader(errorType pdpb.ErrorType, message string)
...
+		return &pdpb.GetMembersResponse{
+			Header: &pdpb.ResponseHeader{
+				Error: &pdpb.Error{
+					Type:    pdpb.ErrorType_UNKNOWN,
+					Message: errs.ErrServerNotStarted.FastGenByArgs().Error(),
+				},
+			},
+		}, nil
+	}
+	members, err := cluster.GetMembers(s.GetClient())
```
HuSharp, line 112: BTW, can u modify `func (s *Server) GetMembers()` return `cluster.GetMembers(s.GetClient())` directly?

server/server.go
```diff
@@ -788,11 +788,7 @@ func (s *Server) StartTimestamp() int64 {
 
 // GetMembers returns PD server list.
 func (s *Server) GetMembers() ([]*pdpb.Member, error) {
-	if s.IsClosed() {
```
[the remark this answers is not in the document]
rleungx: IMO, Maybe it's better to handle it outside. But I change it back to avoid the unnecessary problem.

HuSharp (approved):

binshi-bing (approved):

PR #6094, opened 2023-03-06
mcs: reduce duplicated definitions 

rleungx:

Issue Number: Ref #5836.

This PR inits cluster ID for tso and resource manager server and move some common definition to another package.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/tso/server/server.go
```diff
@@ -540,12 +536,9 @@ func (s *Server) startGRPCAndHTTPServers(l net.Listener) {
 }
 
 func (s *Server) startServer() (err error) {
-	// TODO: uncomment the following code to generate a unique cluster id from the given tsoClusterIDPath
-	// after we add rpc for the client to retrieve the clusgter id from the server then use it in every
-	// request for verification.
-	// if s.clusterID, err = etcdutil.InitClusterID(s.etcdClient, tsoClusterIDPath); err != nil {
-	//	return err
-	// }
+	if s.clusterID, err = etcdutil.GetClusterID(s.etcdClient, utils.ClusterIDPath); err != nil {
```
binshi-bing, line 539: See my TODO: comment on the left side. We can't enable cluster id initialization for tso service for now -- every rpc will check if the cluster id in the request header is equal to the one in the server. For pd, when the pd client starts, it calls GetMemebers to get the cluster id, but tso service/client hasn't defined the equivalent rpc to retrieve the cluster id from the server and use it in the requests thereafter. We need a separate pr which combines this change in tso server, the change in tso protobuf and the tso client initialization. @rleungx

binshi-bing (changes requested):

pkg/mcs/tso/server/server.go
binshi-bing: @rleungx , how about leaving the code as what it is here? I'll make the change mentioned above after you merge this pr.

rleungx: done

lhy1024 (approved):

JmPotato (approved):

PR #6098, opened 2023-03-07
resource_manager_client: remove stale resource group

HuSharp:

Issue Number: Ref #5851 

```commit-message
remove stale resource group which have not been requested
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```


CabinfeverB (commented):
can remove this line https://github.com/tikv/pd/blob/15d361ac7aad796ea658730d893b28bae8743877/client/resource_group/controller/controller.go#L257

client/resource_group/controller/controller.go
```diff
@@ -325,6 +324,14 @@ func (c *ResourceGroupsController) collectTokenBucketRequests(ctx context.Contex
 	requests := make([]*rmpb.TokenBucketRequest, 0)
 	c.groupsController.Range(func(name, value any) bool {
 		gc := value.(*groupCostController)
+		// Check stale resource group
+		if equalRU(gc.run.lastRequestConsumption, gc.run.consumption) {
+			if gc.checkStaleTimes > defaultSlotStalePeriod {
+				c.groupsController.Delete(name)
```
CabinfeverB, line 330: Trigger `collectTokenBucketRequest` interval may be short, resource group may be cleared quickly.

client/resource_group/controller/controller.go
```diff
@@ -254,11 +260,21 @@ func (c *ResourceGroupsController) cleanUpResourceGroup(ctx context.Context) err
...
 	c.groupsController.Range(func(key, value any) bool {
 		resourceGroupName := key.(string)
 		if _, ok := latestGroups[resourceGroupName]; !ok {
 			c.groupsController.Delete(key)
+			return true
+		}
+
+		gc := value.(*groupCostController)
+		// Check stale resource group
+		if equalRU(gc.run.lastRequestConsumption, gc.run.consumption) {
```
CabinfeverB, line 272: I think we should use gc.mu.consumption. Consider this scenario: the resource group has some requests, but the runState has not been updated, resulting in the cleanup. Then the resource group will be recreated, but the previously applied RU token will be lost, which may cause the QPS to decrease.

client/resource_group/controller/config.go
```diff
@@ -45,7 +45,9 @@ const (
 	consumptionsReportingThreshold = 100
 	extendedReportingPeriodFactor  = 4
 	// defaultGroupCleanupInterval is the interval to clean up the deleted resource groups in memory.
-	defaultGroupCleanupInterval = 10 * time.Minute
+	defaultGroupCleanupInterval = 5 * time.Minute
+	// defaultSlotStalePeriod is the interval to clean up the stale resource groups in memory.
+	defaultSlotStalePeriod = 5
```
JmPotato, line 50: What is the unit of this const? It's best to reflect the unit in the name or explain it through comments.

HuSharp: Previously it was intended to remove resource group which was not changed more than the const defaultSlotStalePeriod( now is 6 * 5min) , otherwise it was just +1 to reach this const
How much times do you think would be better to clear

HuSharp: @CabinfeverB What do you think a better time to remove?

CabinfeverB: I agree this implement, because we don't have a var to save consumption change time. But I prefer `defalutStaleCheckThreshold`, and add some comments.

client/resource_group/controller/controller.go
```diff
@@ -254,11 +260,21 @@ func (c *ResourceGroupsController) cleanUpResourceGroup(ctx context.Context) err
...
+			if gc.checkStaleTimes > defaultSlotStalePeriod {
+				c.groupsController.Delete(resourceGroupName)
+				return true
+			}
+			gc.checkStaleTimes += 1
 		}
```
CabinfeverB, line 278: `checkStaleTimes` should be reset?

client/resource_group/controller/config.go
CabinfeverB: maybe 10min

CabinfeverB (approved):

client/resource_group/controller/controller.go
```diff
@@ -261,11 +267,26 @@ func (c *ResourceGroupsController) cleanUpResourceGroup(ctx context.Context) err
...
+		if equalRU(gc.mu.consumption, gc.run.consumption) {
+			gc.mu.Unlock()
+			if gc.tombstone {
+				c.groupsController.Delete(resourceGroupName)
+				return true
+			}
+			gc.tombstone = true
+		} else {
+			gc.mu.Unlock()
+			gc.tombstone = false
```
JmPotato, line 289:
```suggestion
		gc.mu.Lock()
		latestConsumption := gc.mu.consumption
		gc.mu.Unlock()
		if equalRU(latestConsumption, gc.run.consumption) {
			if gc.tombstone {
				c.groupsController.Delete(resourceGroupName)
				return true
			}
			gc.tombstone = true
		} else {
			gc.tombstone = false
```

HuSharp: Done, but it's a pointer so needs to take a value to prevent data race

nolouch:
ptal @JmPotato

client/resource_group/controller/controller.go
```diff
@@ -921,3 +945,8 @@ ret:
 	add(gc.mu.consumption, delta)
 	gc.mu.Unlock()
 }
+
+func (c *ResourceGroupsController) CheckResourceGroupExist(name string) bool {
```
JmPotato, line 949: Need comment.

HuSharp: done.

JmPotato (approved):

PR #6101, opened 2023-03-07
pkg: use tso storage

rleungx:

Issue Number: Ref #5836.

The PR uses the TSO storage to replace `etcdutil` when loading and saving TSO.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/storage/kv/etcd_kv.go
+++ b/pkg/storage/kv/etcd_kv.go
@@ -77,11 +77,17 @@ func (kv *etcdKVBase) LoadRange(key, endKey string, limit int) ([]string, []stri
 	// As a result, when we try to scan from "foo/", it ends up scanning from "/pd/foo"
 	// internally, and returns unexpected keys such as "foo_bar/baz".
 	key = strings.Join([]string{kv.rootPath, key}, "/")
-	endKey = strings.Join([]string{kv.rootPath, endKey}, "/")
+	var OpOption []clientv3.OpOption
+	// If endKey is "\x00", it means to scan with prefix.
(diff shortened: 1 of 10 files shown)


pkg/storage/kv/etcd_kv.go
```diff
@@ -77,11 +77,17 @@ func (kv *etcdKVBase) LoadRange(key, endKey string, limit int) ([]string, []stri
 	// As a result, when we try to scan from "foo/", it ends up scanning from "/pd/foo"
 	// internally, and returns unexpected keys such as "foo_bar/baz".
 	key = strings.Join([]string{kv.rootPath, key}, "/")
-	endKey = strings.Join([]string{kv.rootPath, endKey}, "/")
+	var OpOption []clientv3.OpOption
+	// If endKey is "\x00", it means to scan with prefix.
+	if endKey == "\x00" {
```
JmPotato, line 82: What's this change for?

JmPotato (commented):
The rest LGTM.

pkg/storage/kv/etcd_kv.go
rleungx: We don't rely on the root path to generate the key before calling `LoadTimestamp`, so for global TSO, the key will be `""` and  `GetPrefixRangeEnd` is no longer effective and return a zero byte. We need to handle this situation at the kv level to add a prefix option.

JmPotato (approved):

lhy1024 (approved):

PR #6102, opened 2023-03-07
resource_manager: change cpu time unit from ms to ns

glorv:

Issue Number: Ref #5851

```
Change the cpu time uint from ms to ns because ms is too coarse-grained that it will ignore duration small than 1ms.
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/resource_group/controller/model.go
+++ b/client/resource_group/controller/model.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"os"
+	"time"
 
 	"github.com/elastic/gosigar"
 	"github.com/pingcap/log"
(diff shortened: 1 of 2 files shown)


client/resource_group/controller/model.go
```diff
@@ -38,9 +39,9 @@ type RequestInfo interface {
 
 // ResponseInfo is the interface of the response information provider. A response should be
 // able to tell how many bytes it read and KV CPU cost in milliseconds.
-type ResponseInfo interface {
+type RespontokenConsumptionPerSecondseInfo interface {
```
JmPotato: Why change the naming?

glorv: fix, this was a accidental typo

JmPotato (approved):

nolouch (approved):

PR #6103, opened 2023-03-07
pkg: add new store limit type `SendSnapshot`

### LINKED ISSUE #6104: add new store limit type for limit the  snapshot sender.

bufferflies:

Issue Number: Close #6104

```commit-message
1. using `SendSnapshot` to replace the send cost 
2 rate limit affect the `AddPeer`,`RemovePeer` and the sliding windows affects the `SendSnapshot`
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/storelimit/limit.go
+++ b/pkg/core/storelimit/limit.go
@@ -40,6 +40,4 @@ type StoreLimit interface {
 	Take(count int64, typ Type, level constant.PriorityLevel) bool
 	// Reset resets the store limit
 	Reset(rate float64, typ Type)
-	// Feedback adjust the capacity dynamic.
-	Feedback(err float64, typ Type)
 }
--- a/pkg/core/storelimit/limit.go
+++ b/pkg/core/storelimit/limit.go
@@ -26,6 +26,8 @@ const (
 	AddPeer Type = iota
 	// RemovePeer indicates the type of store limit that limits the removing peer rate
 	RemovePeer
+	// SendSnapshot indicates the type of sending snapshot.
+	SendSnapshot
 
 	storeLimitTypeLen
 )
@@ -38,4 +40,6 @@ type StoreLimit interface {
 	Take(count int64, typ Type, level constant.PriorityLevel) bool
 	// Reset resets the store limit
 	Reset(rate float64, typ Type)
+	// Feedback adjust the capacity dynamic.
+	Feedback(err float64, typ Type)
 }
--- a/pkg/core/storelimit/sliding_window.go
+++ b/pkg/core/storelimit/sliding_window.go
@@ -22,12 +22,19 @@ import (
 const (
 	// minSnapSize is the min value to check the windows has enough size.
 	minSnapSize = 10
+
+	// DefaultCapocity is the default size of the sliding windows.
+	DefaultCapacity = 1000
+
+	defaultProportion = 20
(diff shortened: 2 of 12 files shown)


pkg/core/storelimit/limit.go
```diff
@@ -26,6 +26,8 @@ const (
 	AddPeer Type = iota
 	// RemovePeer indicates the type of store limit that limits the removing peer rate
 	RemovePeer
+	// SendSnapshot indicates the type of sending snapshot.
+	SendSnapshot
```
rleungx, line 30: Will it affect the TiKV threshold?

bufferflies: No, It only affect the sending snapshot of every tikv. operator should be rejected if the origin's leader store has no available limit.

pkg/core/storelimit/sliding_window.go
```diff
@@ -46,7 +46,10 @@ func NewSlidingWindows(cap float64) *SlidingWindows {
 
 // Reset resets the capacity of the sliding windows.
 // It doesn't clear all the used, only set the capacity.
-func (s *SlidingWindows) Reset(cap float64, _ Type) {
+func (s *SlidingWindows) Reset(cap float64, typ Type) {
+	if typ == SendSnapshot {
```
rleungx, line 50: equal?

bufferflies: good catch , fixed it later.

rleungx (approved):

server/cluster/cluster.go
```diff
@@ -88,7 +88,7 @@ const (
 	defaultChangedRegionsLimit   = 10000
 	gcTombstoreInterval          = 30 * 24 * time.Hour
 	// persistLimitRetryTimes is used to reduce the probability of the persistent error
-	// since the once the store is add or remove, we shouldn't return an error even if the store limit is failed to persist.
+	// since the once the store is added or remove, we shouldn't return an error even if the store limit is failed to persist.
```
CabinfeverB, line 91:
```suggestion
	// since the once the store is added or removed, we shouldn't return an error even if the store limit is failed to persist.
```

pkg/core/storelimit/sliding_window.go
```diff
@@ -46,7 +46,10 @@ func NewSlidingWindows(cap float64) *SlidingWindows {
 
 // Reset resets the capacity of the sliding windows.
 // It doesn't clear all the used, only set the capacity.
-func (s *SlidingWindows) Reset(cap float64, _ Type) {
+func (s *SlidingWindows) Reset(cap float64, typ Type) {
+	if typ != SendSnapshot {
```
CabinfeverB, line 50: I don't know exactly what the next implementation looks like. Can we unify `SlidingWindows` and `limit` into a interface?  Introduce `Ack` for `StoreRateLimit` and `limit`.

CabinfeverB (approved):
LGTM

pkg/core/storelimit/sliding_window.go
bufferflies: yes, I will expose the ack in the later pr.

server/cluster/cluster.go

PR #6108, opened 2023-03-07
online recovery: fix online recovery timeout mechanism

### LINKED ISSUE #6107: Online recovery timeout doesn't work
### What did you do?

Run online recovery with the TiKV without the support of online recovery

keeps in collecting reports stage 

### What did you see instead?

timeout

master

Connor1996:

Issue Number: Close #6107 

```commit-message
fix online recovery timeout mechanism
```

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue that online recovery timeout may not work as expected
```


Connor1996:
PTAL @v01dstar

v01dstar:
Can you please explain a little bit more about what the bug is? I can tell that with this change, the whole process will exit faster when timeout happens (the old / existing also exit after timeout, i believe? ). Besides, I think, with this change, we may leave some regions in exit force leader state when timeout?

Connor1996:
> Can you please explain a little bit more about what the bug is? I can tell that with this change, the whole process will exit faster when timeout happens (the old / existing also exit after timeout, i believe? ). Besides, I think, with this change, we may leave some regions in exit force leader state when timeout?

Suppose that, one TiKV always returns store heartbeat but without store report. Then in the existing impl, it would never trigger timeout and keep in the collecting stage forever.

v01dstar:
> > Can you please explain a little bit more about what the bug is? I can tell that with this change, the whole process will exit faster when timeout happens (the old / existing also exit after timeout, i believe? ). Besides, I think, with this change, we may leave some regions in exit force leader state when timeout?
> 
> Suppose that, one TiKV always returns store heartbeat but without store report for somewhat reason. Then in the existing impl, it would never trigger timeout and keep in the collecting stage forever.

I think the existing code still exit? Just with a longer wait time, `timeout + 2* store_heartbeat_interval`. Am I miss something here?

Connor1996:
> > > Can you please explain a little bit more about what the bug is? I can tell that with this change, the whole process will exit faster when timeout happens (the old / existing also exit after timeout, i believe? ). Besides, I think, with this change, we may leave some regions in exit force leader state when timeout?
> > 
> > 
> > Suppose that, one TiKV always returns store heartbeat but without store report for somewhat reason. Then in the existing impl, it would never trigger timeout and keep in the collecting stage forever.
> 
> I think the existing code still exit? Just with a longer wait time, `timeout + 2* store_heartbeat_interval`. Am I miss something here?

Please check `checkTimeout` in existing code, it never change the state.

server/cluster/unsafe_recovery_controller.go
```diff
@@ -274,128 +279,139 @@ func (u *unsafeRecoveryController) HandleErr(err error) bool {
...
+		// automatically after being for a long time.
 		u.changeStage(failed)
 		return true
 	}
+	// When encountering an error for the first time, we will try to exit force
+	// leader before turning into failed stage to avoid the leaking force leaders
+	// blocks reads and writes.
+	u.storePlanExpires = make(map[uint64]time.Time)
+	u.storeRecoveryPlans = make(map[uint64]*pdpb.RecoveryPlan)
+	u.timeout = time.Now().Add(storeRequestInterval)
```
v01dstar, line 293: Isn't one heartbeat interval too aggressive? Maybe *2 to make it more stable?

v01dstar (approved):

rleungx (approved):

nolouch (approved):

PR #6119, opened 2023-03-08
client/resource-manager: fix the issue that some groups cannot be reported

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/6118

```
1. 
create resource group if not exists rg1 RU_PER_SEC=1000;
create resource group if not exists rg2 RU_PER_SEC=2000;

2. 
alter user 'user1' resource group rg1;
alter user 'user2' resource group rg2;
3.
# shell user1
sysbench --mysql-host=127.0.0.1 --mysql-user=user1 --mysql-port=4000 --mysql-db=sbtest --percentile=99 --report-interval=1 --tables=4 --table_size=1000000 oltp_read_only --range_selects --threads=4 --time=1800 run

# shell user2
sysbench --mysql-host=127.0.0.1 --mysql-user=user2 --mysql-port=4000 --mysql-db=sbtest --percentile=99 --report-interval=1 --tables=4 --table_size=1000000 oltp_read_only --range_selects --threads=4 --time=1800 run
4. 
alter resource group rg2 RU_PER_SEC=1000 BURSTABLE;
alter resource group rg2 RU_PER_SEC=500;
```
in step 4, rg1 may continuous trigger `low_ru` report, but it's only report rg1, and update the `lastReuqestTime`, then rg2 no chance to report, then rg2 still think itself in burstable mode.

```commit-message
client/resource-manager: fix the issue that some groups cannot be reported
```

### Check List

Tests

- Unit test
- Integration test
can be normal
image

### Release note

```release-note
None.
```


nolouch:
ptal @CabinfeverB @JmPotato

client/resource_group/controller/controller.go
```diff
@@ -141,6 +142,12 @@ func (c *ResourceGroupsController) GetConfig() *Config {
 	return c.config
 }
 
+// Source List
+var (
```
JmPotato, line 146: Why not use const?

client/resource_group/controller/controller.go
```diff
@@ -141,6 +142,12 @@ func (c *ResourceGroupsController) GetConfig() *Config {
 	return c.config
 }
 
+// Source List
+const (
+	FromPeriodReport = "period_report"
+	FromLowRU        = "low_ru"
```
JmPotato, line 148: These two are not needed to be exported.

JmPotato (approved):

client/resource_group/controller/controller.go
```diff
@@ -269,7 +277,13 @@ func (c *ResourceGroupsController) shouldReportConsumption() bool {
...
 		return false
 	}
-	timeSinceLastRequest := c.run.now.Sub(c.run.lastRequestTime)
+	maxDuration := func(a, b time.Duration) time.Duration {
+		if a > b {
+			return a
+		}
+		return b
+	}
+	timeSinceLastRequest := maxDuration(c.run.now.Sub(c.run.lastRequestTime), c.run.now.Sub(c.run.lastRequestTimeLowRU))
```
CabinfeverB, line 286: This will cause resource groups to still request tokens during trickle time. How about lastRequestTime for each resource group?

nolouch: It's more changes, I think we can optimize this part in later. including:
- hibernate group
- each group maintain self report logic
- more metrics

CabinfeverB: OK, hibernate group may be fixed by https://github.com/tikv/pd/pull/6098

CabinfeverB (approved):

Issue #6099, opened 2023-03-07
v6.5.1 lightning import data to master fail for pd reset the connection


seiya-annie:
### What did you do?

tidb cluster upgrade from 6.0.0 to 6.5.1-pre
import data to 6.0.0 cluster using 6.5.1 lightning

### What did you expect to see?
import success
### What did you see instead?
import fail
print many warning:
{"level":"warn","ts":"2023-03-07T03:08:03.956Z","logger":"etcd-client","caller":"v3@v3.5.2/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc000c22380/src-tidb-pd.fb-testbed-comp-2nrpj:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = "transport: Error while dialing dial tcp 10.244.7.55:2379: connect: connection refused""}

### What version of PD are you using (`pd-server -V`)?
v6.5.1-pre

seiya-annie:
[lightning.log.2023-03-06T16.54.59Z.gz](https://github.com/tikv/pd/files/10905139/lightning.log.2023-03-06T16.54.59Z.gz)

seiya-annie:
[pd-downstream.log.gz](https://github.com/tikv/pd/files/10905150/pd-downstream.log.gz)
[pd-upstream.log.gz](https://github.com/tikv/pd/files/10905151/pd-upstream.log.gz)

seiya-annie:
https://github.com/pingcap/tidb/issues/41973

bufferflies:
<img width="795" alt="image" src="https://user-images.githubusercontent.com/23159587/223670426-56fc35cd-04d7-4522-87e0-fd6ce5904499.png">
the restart related log :
```
[2023/03/06 16:56:27.610 +00:00] [INFO] [region.go:679] ["region Version changed"] [region-id=8021] [detail="StartKey Changed:{7480000000000007FFA300000000000000F8} -> {7480000000000007FFA35F728000000000FF27AC410000000000FA}, EndKey:{748000FFFFFFFFFFFFF900000000000000F8}"] [old-version=1056] [new-version=1059]
[2023/03/06 16:56:44.671 +00:00] [INFO] [versioninfo.go:89] ["Welcome to Placement Driver (PD)"]
[...]

PR #6123, opened 2023-03-09
*: add defer to logs the panic reason and stack

HuSharp:

Issue Number: Ref #6099 

```commit-message
add defer logs the panic reason and stack
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```


HuSharp:
> /need_ok_test

maybe /needs-ok-to-test

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -121,7 +122,10 @@ func (m *Manager) Init(ctx context.Context) {
 	m.storage.LoadResourceGroupStates(tokenHandler)
 	// Start the background metrics flusher.
 	go m.backgroundMetricsFlush(ctx)
-	go m.persistLoop(ctx)
+	go func() {
```
rleungx, line 125: What is the difference between backgroundMetricsFlush and persistLoop?

server/server.go
```diff
@@ -486,6 +486,7 @@ func (s *Server) IsClosed() bool {
 // Run runs the pd server.
 func (s *Server) Run() error {
 	go systimemon.StartMonitor(s.ctx, time.Now, func() {
+		defer logutil.LogPanic()
```
rleungx, line 489: We have already had a defer in StartMonitor

pkg/mcs/resource_manager/server/manager.go
HuSharp: `persistLoop` function requires a look at the code to know that it is a goroutine and is not used anywhere else.
To avoid adding logpanic to confuse people, it was changed to this

server/server.go
HuSharp: done

bufferflies (approved):

rleungx (approved):

HuSharp:
@bufferflies @rleungx can u leave a merge label? thx!

PR #6130, opened 2023-03-09
mcs: support forwarding requests for tso server

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/5836

```commit-message
```

1. support forwarding requests for tso server when api server is disable for tso
2. add switch in api server for tso
3. add test cluster for api server

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024:
cc @binshi-bing  @rleungx

server/grpc_service.go
```diff
@@ -176,6 +183,49 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		}
 
 		streamCtx := stream.Context()
+		if !s.IsTSOEnabled() {
```
rleungx, line 186: How about just using api mode?

lhy1024: I am not sure whether we support downgrade in api service mode, when tso service is not available.

tests/mcs/tso/server_test.go
```diff
@@ -137,3 +139,46 @@ func (suite *tsoServerTestSuite) TestTSOServerRegister() {
...
+
+	leaderName := cluster.WaitLeader()
+	pdLeader := cluster.GetServer(leaderName)
+	backendEndpoints := pdLeader.GetAddr()
+
+	{ // cannot get ts from api server addr
+		pdClient, err := pd.NewClientWithContext(ctx, []string{backendEndpoints}, pd.SecurityOption{})
+		re.NoError(err)
+		_, _, err = pdClient.GetTS(ctx)
+		re.Error(err)
```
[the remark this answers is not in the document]
lhy1024: ok

server/grpc_service.go
```diff
@@ -1844,6 +1933,37 @@ func forwardRegionHeartbeatClientToServer(forwardStream pdpb.PD_RegionHeartbeatC
...
+	go checkStream(ctx, cancel, done)
+	forwardStream, err := tsopb.NewTSOClient(client).Tso(ctx)
+	done <- struct{}{}
+	return forwardStream, cancel, err
+}
+
+func forwardMCSTSOClientToServer(forwardStream tsopb.TSO_TsoClient, server *tsoServer, errCh chan error) {
+	defer close(errCh)
+	for {
+		respTSO, err := forwardStream.Recv()
```
binshi-bing, line 1948: We better add tsoHandleDuration metrics to record the duration between forwardStream.send() and forwardStream.recv().

server/grpc_service.go
```diff
@@ -176,6 +183,49 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
...
+				}
+				lastForwardedHost = forwardedHost
+				errCh = make(chan error, 1)
+				go forwardMCSTSOClientToServer(forwardStream, server, errCh)
+			}
+			header := request.GetHeader()
+			forwardRequest := &tsopb.TsoRequest{
+				Header: &tsopb.RequestHeader{
+					ClusterId: s.clusterID,
+					SenderId:  header.GetSenderId(),
```
binshi-bing, line 212: shouldn't be zero. I'll ask the serviceless team to provide this value

server/grpc_service.go
```diff
@@ -176,6 +183,49 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
...
+					// TODO: whether set keyspace info to 0
+				},
+				Count:      request.GetCount(),
+				DcLocation: request.GetDcLocation(),
+			}
+			if err := forwardStream.Send(forwardRequest); err != nil {
```
binshi-bing, line 218: I worry about that if TSO service has problem and here API service still asynchronously sends requests to tso service as fast as possible, API service could be in trouble and impact all our services. We need some protection mechanism. One way I can think of is to call forwardStream.Send() and forwardStream.recv() sequentially in a goroutine.

binshi-bing (commented):
left some comments

server/grpc_service.go
binshi-bing, line 218: what will happen if the first forwardStream.Recv() in the goroutine forwardMCSTSOClientToServer()  is executed before the first forwardStream() is called here?

binshi-bing: I asked this question "regarding grpc.ClientStream, can we call recv() before calling send()?" on ChatGPT. Here is the answer -- No, it is not recommended to call recv() on a grpc.ClientStream before calling send(). In fact, the grpc.ClientStream API is designed such that you must call send() before calling recv().

server/grpc_service.go
```diff
@@ -176,6 +183,49 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		}
 
 		streamCtx := stream.Context()
+		if !s.IsTSOEnabled() {
+			_, forwardedHost, err := s.GetServicePrimaryAddr(ctx, "tso")
```
rleungx, line 187: The first value cannot be omitted.

server/grpc_service.go
```diff
@@ -176,6 +183,49 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
...
+				forwardStream, cancel, err = s.createMCSTSOForwardStream(client)
+				if err != nil {
+					return err
+				}
+				lastForwardedHost = forwardedHost
+				errCh = make(chan error, 1)
+				go forwardMCSTSOClientToServer(forwardStream, server, errCh)
+			}
+			header := request.GetHeader()
+			forwardRequest := &tsopb.TsoRequest{
```
rleungx, line 209: Can we use dispatchTSORequest too?

server/grpc_service.go
lhy1024: thanks

server/grpc_service.go
lhy1024: ok

pkg/mcs/tso/server/server.go
```diff
@@ -560,10 +560,10 @@ func (s *Server) startServer() (err error) {
 	// TODO: uncomment the following code to generate a unique cluster id from the given ClusterIDPath
 	// after we add rpc for the client to retrieve the cluster id from the server then use it in every
 	// request for verification.
-	// if s.clusterID, err = etcdutil.GetClusterID(s.etcdClient, utils.ClusterIDPath); err != nil {
-	// 	return err
-	// }
-	log.Info("init cluster id", zap.Uint64("cluster-id", s.clusterID))
+	if s.clusterID, err = etcdutil.GetClusterID(s.etcdClient, utils.ClusterIDPath); err != nil {
```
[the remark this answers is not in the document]
lhy1024: If so, cluster id of tso server will be always zero in test.

server/grpc_service.go
lhy1024: done

server/grpc_service.go
```diff
@@ -176,6 +181,26 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		}
 
 		streamCtx := stream.Context()
+		if !s.IsTSOEnabled() {
+			ok, forwardedHost, err := s.GetServicePrimaryAddr(ctx, "tso")
+			if !ok {
+				return errors.WithStack(ErrNotFoundTSOADDR)
```
rleungx, line 187: WithStack?

server/grpc_service.go
rleungx: Please change it

server/grpc_service.go
```diff
@@ -222,33 +247,38 @@ type tsoRequest struct {
 	stream        pdpb.PD_TsoServer
 }
 
-func (s *GrpcServer) dispatchTSORequest(ctx context.Context, request *tsoRequest, forwardedHost string, doneCh <-chan struct{}, errCh chan<- error) {
+func (s *GrpcServer) dispatchTSORequest(ctx context.Context, request *tsoRequest, forwardedHost string, doneCh <-chan struct{}, errCh chan<- error, workingMCS bool) {
```
rleungx, line 250: How about using a clearer name instead of `workingMCS`?

lhy1024: Do you have a good idea?

lhy1024:
cc@binshi-bing @rleungx

server/grpc_service.go
binshi-bing: The difference inside of dispatchTSORequest is actually about which protocolbuf to use, pdpb or tsopb? How about 'tsoProto bool'. The tso service itself needs to use this framework too.

server/grpc_service.go
```diff
@@ -180,6 +185,26 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		}
 
 		streamCtx := stream.Context()
+		if s.IsAPIServiceMode() {
+			ok, forwardedHost, err := s.GetServicePrimaryAddr(ctx, "tso")
```
rleungx, line 189: We need a cache for it maybe.

server/grpc_service.go
```diff
@@ -60,6 +61,7 @@ var (
 	ErrNotLeader            = status.Errorf(codes.Unavailable, "not leader")
 	ErrNotStarted           = status.Errorf(codes.Unavailable, "server not started")
 	ErrSendHeartbeatTimeout = status.Errorf(codes.DeadlineExceeded, "send heartbeat timeout")
+	ErrNotFoundTSOADDR      = status.Errorf(codes.NotFound, "not found tso address")
```
rleungx, line 64:
```suggestion
	ErrNotFoundTSOAddr      = status.Errorf(codes.NotFound, "not found tso address")
```

server/grpc_service.go
```diff
@@ -310,26 +340,55 @@ errHandling:
...
-		DcLocation: requests[0].request.GetDcLocation(),
-	}
-	// Send to the leader stream.
-	if err := forwardStream.Send(req); err != nil {
-		return err
+	var (
+		resp tsoResp
+		err  error
+	)
+	if forwardStream != nil {
```
rleungx, line 358: Will both forwardStream and forwardMCSStream not be nil?

server/grpc_service.go
```diff
@@ -1884,6 +1961,40 @@ func checkStream(streamCtx context.Context, cancel context.CancelFunc, done chan
...
+		return pdpb.Timestamp{}, errors.WithStack(err)
+	}
+	client, err := s.getDelegateClient(ctx, forwardedHost)
+	if err != nil {
+		return pdpb.Timestamp{}, err
+	}
+	done := make(chan struct{})
+	ctx, cancel := context.WithTimeout(s.ctx, defaultTSOProxyTimeout)
+	go checkStream(ctx, cancel, done)
+	forwardStream, err := tsopb.NewTSOClient(client).Tso(ctx)
```
rleungx, line 1979: What if the primary changes?

server/grpc_service.go
lhy1024: Line 281 ensure one of them will be not nil

server/grpc_service.go
lhy1024: ok, I will do it in another pr

server/grpc_service.go
lhy1024: ok

server/grpc_service.go
lhy1024: it will change by https://github.com/tikv/pd/blob/5744ddd321d3deb728d47dbec92c6efbe8bd0972/server/grpc_service.go#L1965

rleungx: How about the original stream, will it be closed?

server/grpc_service.go
lhy1024: done

server/grpc_service.go
```diff
@@ -374,8 +375,9 @@ func (s *GrpcServer) processTSORequests(forwardStream pdpb.PD_TsoClient, forward
 	if forwardMCSStream != nil {
 		req := &tsopb.TsoRequest{
 			Header: &tsopb.RequestHeader{
-				ClusterId: requests[0].request.GetHeader().GetClusterId(),
-				// TODO: set keyspace info
+				ClusterId:       requests[0].request.GetHeader().GetClusterId(),
+				KeyspaceId:      utils.DefaultKeySpaceGroupID,
+				KeyspaceGroupId: utils.DefaultKeySpaceGroupID,
```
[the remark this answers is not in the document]
lhy1024: just filled it with default value

server/grpc_service.go
lhy1024: original stream? with pdpb proto?

rleungx: If we create a stream to A, after that, B becomes primary. Then, we create a stream to B. What about A now?

lhy1024: Its behaviour will be similar with other stream in clientConns

rleungx: Yes, but do we need to consider that?

lhy1024: The current number of streams is directly related to the tso server, the current number is not much, I will add a todo first. and do we need to think about the eof issue, do we keep a todo first too?

server/grpc_service.go
```diff
@@ -1884,6 +1963,41 @@ func checkStream(streamCtx context.Context, cancel context.CancelFunc, done chan
...
+		return pdpb.Timestamp{}, errors.WithStack(err)
+	}
+	client, err := s.getDelegateClient(ctx, forwardedHost)
+	if err != nil {
+		return pdpb.Timestamp{}, err
+	}
+	done := make(chan struct{})
+	ctx, cancel := context.WithTimeout(s.ctx, defaultTSOProxyTimeout)
+	go checkStream(ctx, cancel, done)
+	forwardStream, err := tsopb.NewTSOClient(client).Tso(ctx)
```
rleungx, line 1981: Why do we need to create a stream here?

server/grpc_service.go
```diff
@@ -310,26 +341,56 @@ errHandling:
...
 	}
-	resp, err := forwardStream.Recv()
-	if err != nil {
-		return err
+	if forwardMCSStream != nil {
+		req := &tsopb.TsoRequest{
+			Header: &tsopb.RequestHeader{
+				ClusterId:       requests[0].request.GetHeader().GetClusterId(),
+				KeyspaceId:      utils.DefaultKeySpaceID,
+				KeyspaceGroupId: utils.DefaultKeySpaceGroupID,
```
rleungx, line 380: Do we really need these two for now?

lhy1024: only filled it

server/grpc_service.go
lhy1024: I will cache it until forward-host changed

rleungx (approved):

binshi-bing (approved):
LGTM

JmPotato (approved):

tests/mcs/tso/server_test.go
```diff
@@ -118,17 +118,43 @@ func (suite *tsoServerTestSuite) TestTSOServerStartAndStopNormally() {
 	}
 }
 
-func (suite *tsoServerTestSuite) TestTSOPath() {
-	re := suite.Require()
+func TestTSOPath(t *testing.T) {
+	re := require.New(t)
+	checkTSOPath(re, true /*isAPIServiceMode*/)
+	checkTSOPath(re, false /*isAPIServiceMode*/)
```
rleungx, line 124: The comment is the same?

PR #6135, opened 2023-03-10
config: adjust config when tikv engine is `raft-kv2`

### LINKED ISSUE #6167: The default config should be changed if tikv's engine is `raft kv2`
## Enhancement Task
In using the `raft-kv2` , the influence of `split` and `merge` operator is too high, PD should disable it and can modify it by manual .

bufferflies:

Issue Number: Close #6167

```commit-message
1. raft store v2 has not support merge 
2. the split's cost is high
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


server/config/config.go
```diff
@@ -792,6 +797,10 @@ func (c *ScheduleConfig) adjust(meta *configutil.ConfigMetaData, reloading bool)
 		c.EnableDiagnostic = defaultEnableDiagnostic
 	}
 
+	if !meta.IsDefined("store-config-synced") {
+		c.EnableDiagnostic = defaultStoreConfigSynced
```
rleungx, line 801: EnableDiagnostic?

server/config/persist_options.go
```diff
@@ -196,6 +197,21 @@ func (o *PersistOptions) SetMaxReplicas(replicas int) {
...
+// disable merge check
+// disable split buckets
+func (o *PersistOptions) UseRaftV2() {
+	v := o.GetScheduleConfig().Clone()
+	if v.StoreConfigSynced {
+		return
+	}
+	v.MaxMergeRegionSize = 0
+	v.MaxMovableHotPeerSize = math.MaxInt64
+	v.StoreConfigSynced = true
```
rleungx, line 211: Will it become false after it becomes true?

bufferflies: never become true, because the default config should be update only once. so the default config can be modified by manually.

server/config/config.go
bufferflies: fixed

server/config/persist_options.go
rleungx: What if we disable v2?

bufferflies: region merge and bucket split?

rleungx: I mean we enable it and then disable it.

bufferflies: It will edit the merge size again after it restart.

server/cluster/cluster.go
```diff
@@ -380,30 +380,43 @@ func (c *RaftCluster) startGCTuner() {
 func (c *RaftCluster) runSyncConfig() {
 	defer logutil.LogPanic()
 	defer c.wg.Done()
-
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
+
 	stores := c.GetStores()
+	syncFunc := func() {
+		success, switchRaftV2Config := syncConfig(c.storeConfigManager, stores)
```
lhy1024, line 388:
```suggestion
		synced, switchRaftV2Config := syncConfig(c.storeConfigManager, stores)
```

server/cluster/cluster.go
```diff
@@ -380,30 +380,43 @@ func (c *RaftCluster) startGCTuner() {
...
-			}
+			syncFunc()
 		}
 	}
 }
 
-func syncConfig(manager *config.StoreConfigManager, stores []*core.StoreInfo) bool {
+// syncConfig syncs the config of the stores.
+// synced is true if sync config from one tikv.
+// switchRaftV2 is true if the config of tikv engine is changed and engine is raft-kv2.
```
lhy1024, line 414: do we supported downgrade from kv2?

lhy1024 (approved):

lhy1024 (changes requested):
cancel LGT1 label

rleungx (approved):

server/cluster/cluster.go
bufferflies: no, raftkv2 is incompatible with raftkv1.

server/cluster/cluster.go

lhy1024 (approved):

PR #6138, opened 2023-03-10
Add and initialize cluster id in the TSO service

binshi-bing:

Issue Number: Ref #5836 

```commit-message
Add and initialize cluster id in the TSO service.
PD/API service discovery gets cluster id from the API service.
It then passes the cluster id to the TSO service because both share the same cluster id.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -275,7 +275,7 @@ func NewClient(svrAddrs []string, security SecurityOption, opts ...ClientOption)
 // NewClientWithContext creates a PD client with context.
 func NewClientWithContext(ctx context.Context, svrAddrs []string, security SecurityOption, opts ...ClientOption) (Client, error) {
 	log.Info("[pd] create pd client with endpoints", zap.Strings("pd-address", svrAddrs))
-	c, clientCtx, clientCancel, tlsCfg := createClient(ctx, &security)
+	c, clientCtx, clientCancel, tlsCfg := createClient(ctx, 0, &security)
 	// Inject the client options.
 	for _, opt := range opts {
 		opt(c)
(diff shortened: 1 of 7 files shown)


client/client.go
```diff
@@ -376,8 +384,8 @@ func (c *client) scheduleUpdateTokenConnection() {
 }
 
 // GetClusterID returns the ClusterID.
-func (c *client) GetClusterID(ctx context.Context) uint64 {
-	return c.svcDiscovery.GetClusterID(ctx)
+func (c *client) GetClusterID(context.Context) uint64 {
```
rleungx, line 387: Do we need the context?

binshi-bing: No need, but it's defined in the Client interfact. I can't change it.

type Client interface {
	// GetClusterID gets the cluster ID from PD.
	GetClusterID(ctx context.Context) uint64

client/client.go
```diff
@@ -275,7 +275,7 @@ func NewClient(svrAddrs []string, security SecurityOption, opts ...ClientOption)
 // NewClientWithContext creates a PD client with context.
 func NewClientWithContext(ctx context.Context, svrAddrs []string, security SecurityOption, opts ...ClientOption) (Client, error) {
 	log.Info("[pd] create pd client with endpoints", zap.Strings("pd-address", svrAddrs))
-	c, clientCtx, clientCancel, tlsCfg := createClient(ctx, &security)
+	c, clientCtx, clientCancel, tlsCfg := createClient(ctx, 0, &security)
```
rleungx, line 278: Please add a comment to explain that `0` represents the default group.

rleungx (approved):
The rest LGTM

lhy1024 (approved):

client/client.go
binshi-bing: added

Issue #5762, opened 2022-12-07
add balance witness scheduler


ethercflow:
## Development Task

ref https://github.com/tikv/tikv/issues/12876

PR #6149, opened 2023-03-12
schedulers: opt balance witness when store nr larger than peer nr

nathan-edgar:
ref tikv/pd#5762

Issue Number: ref #5762 

1. When stores nr larger than peers nr, move witness instead of transfer witenss
2. For the same region, when witness switching occurs, the order of step generation is guaranteed:  demote witness(peer1) to witness learner, non-witness learner, voter, voter (peer2) -> witness, to avoid there is a non-witness learner applying snapshot and a witness at the same time
3. When verifying the legitimacy of the targetLeader in the builder, the judgment is made based on the result of the transfer witness instead of the intermediate state
4. Add metrics in pd.json

```commit-message
schedulers: opt balance witness when store nr larger than peer nr
```

### Check List

Tests

- Unit test
- Integration test
- Manual test 
   - Test with sysbench prepare

Code changes

- None

Side effects

- None

Related changes

- None

### Release note

```release-note
None.
```


pkg/core/basic_cluster.go
```diff
@@ -91,6 +91,27 @@ func (bc *BasicCluster) GetNonWitnessVoterStores(region *RegionInfo) []*StoreInf
 	return Stores
 }
 
+// GetOptionalStoresForRegion returns all Stores that can be used to store the region's peer.
+func (bc *BasicCluster) GetOptionalStoresForRegion(region *RegionInfo) []*StoreInfo {
```
bufferflies, line 95: how about add new filter for this store? like `excludedFilter`.

PR #6152, opened 2023-03-13
mcs: add retry mechanism for initializing the cluster ID

rleungx:

Issue Number: Ref #5836.

This PR adds a retry mechanism when initializing the cluster ID to prevent using the wrong TSO path.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/server/server.go
```diff
@@ -350,8 +351,22 @@ func (s *Server) GetPrimary() bs.MemberProvider {
 	return s.participant.GetLeader()
 }
 
+func (s *Server) initClusterID() (err error) {
```
lhy1024, line 354: how about adding it to `utils` to avoid duplicate functions in `server.go`

lhy1024 (approved):

pkg/mcs/tso/server/server.go
```diff
@@ -555,8 +556,22 @@ func (s *Server) startGRPCAndHTTPServers(l net.Listener) {
...
 
+func (s *Server) initClusterID() (err error) {
+	for i := 0; i < utils.MaxRetryTimes; i++ {
+		if s.clusterID, err = etcdutil.GetClusterID(s.etcdClient, utils.ClusterIDPath); err == nil && s.clusterID != 0 {
+			return nil
+		}
+		select {
+		case <-s.ctx.Done():
+			return err
+		case <-time.After(time.Second):
```
lhy1024, line 567: how about adding it to constant.go?

rleungx:
@binshi-bing PTAL

lhy1024 (approved):

pkg/mcs/utils/util.go
```diff
@@ -0,0 +1,48 @@
...
+	// clusterIDPath is the path to store cluster id
+	clusterIDPath = "/pd/cluster_id"
+	// retryInterval is the interval to retry.
+	retryInterval = time.Second
+)
+
+// InitClusterID initializes the cluster ID.
+func InitClusterID(ctx context.Context, client *clientv3.Client) (id uint64, err error) {
+	for i := 0; i < maxRetryTimes; i++ {
+		if clusterID, err := etcdutil.GetClusterID(client, clusterIDPath); err == nil && clusterID != 0 {
```
JmPotato, line 38: According to your PR description, this PR tries to prevent using the wrong TSO path while initializing the cluster ID, but the retry here always seems to use the same path.

rleungx: The PR aims to use the same TSO path for the default group no matter in pd mode or microservice mode. But consider that TSO server may start first and we need the cluster ID to get the original root path. So we add the retry here to ensure it can get the original cluster ID and access the TSO path.

JmPotato (approved):

PR #6160, opened 2023-03-14
Support service mode switching for tso and change the value type of tso primary to tsopb.Participant

binshi-bing:
Support service mode switching for and change the value type of tso primary

Issue Number: Ref #5836 

```commit-message
Support tso service switched between PD mode and API mode;
Add server/client implemenation for the GetClusterInfo rpc
Add tsopb.Participant as the value type of tso primary
```

### Check List

Tests

- Unit test
- Integration test
- Performance test

**Conclusion: no performance regression.**

**w/ change**
Total:
count: 28949005, max: 6.1310ms, min: 0.0410ms, avg: 2.0541ms
<1ms: 479588, >1ms: 11500091, >2ms: 16969184, >5ms: 142, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 28949005, <1ms: 1.66%, >1ms: 39.73%, >2ms: 58.62%, >5ms: 0.00%, >10ms: 0.00%, >30ms: 0.00%, >50ms: 0.00%, >100ms: 0.00%, >200ms: 0.00%, >400ms: 0.00%, >800ms: 0.00%, >1s: 0.00%
P0.5: 2.0470ms, P0.8: 2.2553ms, P0.9: 2.4288ms, P0.99: 3.1310ms

**w/o change**
Total:
count: 28181983, max: 9.9771ms, min: 0.0616ms, avg: 2.0791ms
<1ms: 367685, >1ms: 10210926, >2ms: 17602085, >5ms: 1287, >10ms: 0, >30ms: 0, >50ms: 0, >100ms: 0, >200ms: 0, >400ms: 0, >800ms: 0, >1s: 0
count: 28181983, <1ms: 1.30%, >1ms: 36.23%, >2ms: 62.46%, >5ms: 0.00%, >10ms: 0.00%, >30ms: 0.00%, >50ms: 0.00%, >100ms: 0.00%, >200ms: 0.00%, >400ms: 0.00%, >800ms: 0.00%, >1s: 0.00%
P0.5: 2.0682ms, P0.8: 2.2742ms, P0.9: 2.4449ms, P0.99: 3.1436ms

### Release note

```release-note
None.
```


client/pd_service_discovery.go
```diff
@@ -345,14 +370,29 @@ func (c *pdServiceDiscovery) initClusterID() error {
 	return nil
 }
 
+func (c *pdServiceDiscovery) updateServiceMode() {
+	leaderAdrr := c.getLeaderAddr()
+	if len(leaderAdrr) > 0 {
```
rleungx, line 375:
```suggestion
	if len(leaderAddr) > 0 {
```

client/pd_service_discovery.go
```diff
@@ -393,6 +433,25 @@ func (c *pdServiceDiscovery) updateMember() error {
...
+		return nil, err
+	}
+	clusterInfo, err := pdpb.NewPDClient(cc).GetClusterInfo(ctx, &pdpb.GetClusterInfoRequest{})
+	if err != nil {
+		attachErr := errors.Errorf("error:%s target:%s status:%s", err, cc.Target(), cc.GetState().String())
+		return nil, errs.ErrClientGetMember.Wrap(attachErr).GenWithStackByCause()
+	}
+	if clusterInfo.GetHeader().GetError() != nil {
+		attachErr := errors.Errorf("error:%s target:%s status:%s", clusterInfo.GetHeader().GetError().String(), cc.Target(), cc.GetState().String())
+		return nil, errs.ErrClientGetMember.Wrap(attachErr).GenWithStackByCause()
```
rleungx, line 450: The error should be changed

pkg/mcs/meta_storage/server/grpc_service.go
```diff
@@ -75,7 +75,7 @@ func (s *Service) RegisterRESTHandler(userDefineHandlers map[string]http.Handler
 }
 
 func (s *Service) checkServing() error {
-	if !s.manager.srv.IsServing() {
+	if s == nil || s.manager == nil || s.manager.srv == nil || !s.manager.srv.IsServing() {
```
rleungx, line 78: Maybe only checking the server is enough?

pkg/mcs/tso/server/server.go
```diff
@@ -68,7 +68,7 @@ const (
 	pdRootPath = "/pd"
 	// tsoPrimaryPrefix defines the key prefix for keyspace group primary election.
 	// The entire key is in the format of "/ms/<cluster-id>/tso/<group-id>/primary" in which
-	// <group-id> is 5 digits integer with leading zeros. For now we use 0 as the default cluster id.
+	// <group-id> is 5 digits integer with leading zeros. 0 is the default cluster id.
```
rleungx, line 71: Maybe we don't need the default cluster ID?

client/pd_service_discovery.go
binshi-bing: changed.

client/pd_service_discovery.go
binshi-bing: changed to ErrClientGetClusterInfo

pkg/mcs/meta_storage/server/grpc_service.go
binshi-bing: nope, it isn't enough. The reason is s.manager == nil or s.manager.srv == nil. If you remove one of them, you can almost constantly repro the issue by killing pd/api server while keeping sending requests by tools like pd-tso-bench

pkg/mcs/tso/server/server.go
binshi-bing: you're right. use the cluster id generated instead.

client/client.go
```diff
@@ -1158,5 +1201,5 @@ func (c *client) respForErr(observer prometheus.Observer, start time.Time, err e
 // GetTSOAllocators returns {dc-location -> TSO allocator leader URL} connection map
 // For test only.
 func (c *client) GetTSOAllocators() *sync.Map {
-	return c.tsoClient.GetTSOAllocators()
+	return c.getTSOClient().GetTSOAllocators()
```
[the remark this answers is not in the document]
binshi-bing: Make sense. changed to "if the client fails to switch to new service mode, keep the current mode instead of swtiching to unknown mode"

client/client.go
```diff
@@ -376,6 +358,61 @@ func (c *client) Close() {
...
+			c.pdSvcDiscovery, c.pdSvcDiscovery.(tsoAllocatorEventSource), &pdTSOStreamBuilderFactory{})
+		newTSOCli.Setup()
+	} else {
+		tsoSvcDiscovery := newTSOServiceDiscovery(ctx, cancel, wg, MetaStorageClient(c),
+			c.GetClusterID(c.ctx), c.keyspaceID, c.svrUrls, c.tlsCfg, c.option)
+		newTSOCli = newTSOClient(ctx, cancel, wg, c.option, c.keyspaceID,
+			tsoSvcDiscovery, tsoSvcDiscovery.(tsoAllocatorEventSource), &tsoTSOStreamBuilderFactory{})
+		if err := tsoSvcDiscovery.Init(); err != nil {
+			cancel()
+			c.serviceMode = pdpb.ServiceMode_UNKNOWN_SVC_MODE
```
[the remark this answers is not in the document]
binshi-bing: Make sense. changed to "if the client fails to switch to new service mode, keep the current mode instead of swtiching to unknown mode"

client/client.go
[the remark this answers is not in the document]
binshi-bing: changed to:
```
        tsoClient := c.getTSOClient()
	if tsoClient == nil {
		return nil
	}
	return tsoClient.GetTSOAllocators()
```

client/pd_service_discovery.go
```diff
@@ -345,14 +374,29 @@ func (c *pdServiceDiscovery) initClusterID() error {
 	return nil
 }
 
+func (c *pdServiceDiscovery) updateServiceMode() {
+	leaderAddr := c.getLeaderAddr()
+	if len(leaderAddr) > 0 {
+		clusterInfo, err := c.getClusterInfo(c.ctx, leaderAddr, c.option.timeout)
+		if err != nil {
+			log.Warn("[pd] failed to get cluster info fror the leader", zap.String("leader-addr", leaderAddr), errs.ZapError(err))
```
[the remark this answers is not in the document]
binshi-bing: good catch. changed.

rleungx (approved):

rleungx:
CI failed.

client/go.mod
```diff
@@ -39,3 +39,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/pingcap/kvproto => github.com/binshi-bing/kvproto v0.0.0-20230314013108-d80ecb25d075
```
lhy1024, line 43: update it plz

binshi-bing: done

pkg/basicserver/basic_server.go
```diff
@@ -39,8 +33,8 @@ type Server interface {
 	Run() error
 	// Close closes the server.
 	Close()
-	// GetPrimary returns the primary of the server.
-	GetPrimary() MemberProvider
+	// GetLeaderListenUrls gets service endpoints from the leader in election group.
+	GetLeaderListenUrls() []string
```
lhy1024, line 37: how about return `string` directly? There is only one leader or none.

client/client.go
```diff
@@ -240,19 +240,25 @@ func WithMaxErrorRetry(count int) ClientOption {
...
-	// svcDiscovery is for pd service discovery
-	svcDiscovery    ServiceDiscovery
-	tsoClient       *tsoClient
+	keyspaceID      uint32
+	svrUrls         []string
+	pdSvcDiscovery  ServiceDiscovery
 	tokenDispatcher *tokenDispatcher
 
+	// For service mode switching
+	mutex           sync.RWMutex
```
JmPotato, line 249: How about declaring an anonymous struct to tell which fields this lock is protecting?

```suggestion
	struct {
		mutex           sync.RWMutex
		xxx
	}
```

client/client.go
```diff
@@ -272,56 +278,16 @@ func NewClient(svrAddrs []string, security SecurityOption, opts ...ClientOption)
...
-	c.tsoClient = newTSOClient(clientCtx, clientCancel, &c.wg, c.option, c.keyspaceID, tsoSvcDiscovery, tsoSvcDiscovery.(tsoAllocatorEventSource), &tsoTSOStreamBuilderFactory{})
-	if err := c.tsoClient.setup(); err != nil {
-		c.cancel()
-		return nil, err
-	}
-	return c, nil
-}
+// NewClientWithKeyspace creates a client with context and the specified keyspace id.
+func NewClientWithKeyspace(ctx context.Context, keyspaceID uint32, svrAddrs []string, security SecurityOption, opts ...ClientOption) (Client, error) {
+	log.Info("[pd] create pd client with endpoints", zap.Strings("pd-address", svrAddrs))
```
JmPotato, line 289: What about logging the `keyspaceID` here?

client/client.go
```diff
@@ -376,6 +358,68 @@ func (c *client) Close() {
...
+
+	if newMode == pdpb.ServiceMode_UNKNOWN_SVC_MODE {
+		log.Warn("intend to switch to unknown service mode. do nothing")
+		return
+	}
+
+	var newTSOCli *tsoClient
+	tsoSvcDiscovery := c.tsoSvcDiscovery
+	ctx, cancel := context.WithCancel(c.ctx)
+	wg := &sync.WaitGroup{}
```
JmPotato, line 380: Why pass the `wg` as a parameter rather than creating the wait group inside each constructor function?

client/client.go
```diff
@@ -542,10 +586,16 @@ func (c *client) GetLocalTSAsync(ctx context.Context, dcLocation string) TSFutur
 	req.keyspaceID = c.keyspaceID
 	req.dcLocation = dcLocation
 
-	if err := c.tsoClient.dispatchRequest(dcLocation, req); err != nil {
+	tsoClient := c.getTSOClient()
+	if tsoClient == nil {
+		req.done <- errs.ErrClientGetTSO
+		return req
+	}
```
JmPotato, line 593: Maybe we can do this check ASAP right after `req` is constructed to prevent some unnecessary variable copying and assignment.

client/client.go
binshi-bing:
```
// serviceModeKeeper is for service mode switching
type serviceModeKeeper struct {
	svcModeMutex    sync.RWMutex
	serviceMode     pdpb.ServiceMode
	tsoClient       atomic.Value
	tsoSvcDiscovery ServiceDiscovery
}

type client struct {
        ...
	serviceModeKeeper
        ...
```

pkg/basicserver/basic_server.go
binshi-bing: It's the listenUrls passed to the pd which becomes leader. It can be multiple urls and there is one place in admin actually accessing all urls.

client/client.go
binshi-bing: added logging for keyspaceID

client/client.go
binshi-bing: It is shared acroos tsoClient and tsoServiceDiscovery. Yes, I can separate them.

**Separted**.

client/client.go
binshi-bing: You're right. Moved.

```
	req := tsoReqPool.Get().(*tsoRequest)
	tsoClient := c.getTSOClient()
	if tsoClient == nil {
		req.done <- errs.ErrClientGetTSO
		return req
	}

	req.requestCtx = ctx
	req.clientCtx = c.ctx
	req.start = time.Now()
	req.keyspaceID = c.keyspaceID
	req.dcLocation = dcLocation
```

binshi-bing: I reverted, because the test failures show that after "req.done <- errs.ErrClientGetTSO", the TSFuture.Wait() could use some fields whose initialization are skipped here.

pkg/basicserver/basic_server.go
lhy1024: I found we use listenUrls[0] as primary by default.

tests/mcs/tso/server_test.go
```diff
@@ -222,51 +227,55 @@ func (suite *APIServerForwardTestSuite) TearDownTest() {
...
 	time.Sleep(time.Duration(utils.DefaultLeaderLease) * time.Second) // wait for leader lease timeout
+	mcs.WaitForPrimaryServing(suite.Require(), serverMap)
 	primary, exist := suite.pdLeader.GetServer().GetServicePrimaryAddr(suite.ctx, utils.TSOServiceName)
 	suite.True(exist)
 	suite.NotEqual(oldPrimary, primary)
 	suite.checkAvailableTSO()
 
 	// can use the tso-related interface with old primary again
-	_, cleanup := mcs.StartSingleTSOTestServer(suite.ctx, suite.Require(), suite.backendEndpoints, oldPrimary)
+	newPrimary := tempurl.Alloc()
```
lhy1024, line 263: According to the comment, we should test to start the previous server with the same address to see if it works, instead of using a new address

lhy1024: use `tempurl.Alloc()` change test behavior

server/server.go
```diff
@@ -1711,15 +1710,20 @@ func (s *Server) watchServicePrimaryAddrLoop(serviceName string) {
...
+	primary := &tsopb.Participant{}
+	ok, _, err := etcdutil.GetProtoMsgWithModRev(s.client, serviceKey, primary)
 	if err != nil {
 		log.Error("get service primary addr failed", zap.String("service-key", serviceKey), zap.Error(err))
 	}
-	if err == nil && ok {
-		s.servicePrimaryMap.Store(serviceName, leader.GetName())
+	listenUrls := primary.GetListenUrls()
+	if ok && len(listenUrls) > 0 {
+		s.servicePrimaryMap.Store(serviceName, listenUrls[0])
```
lhy1024, line 1724: where it is ensured that use [0] as primary?

pkg/mcs/resource_manager/server/server.go
```diff
@@ -242,12 +241,12 @@ func (s *Server) AddStartCallback(callbacks ...func()) {
...
 // IsServing returns whether the server is the leader, if there is embedded etcd, or the primary otherwise.
 func (s *Server) IsServing() bool {
-	return s.participant.IsLeader()
+	return !s.IsClosed() && s.participant.IsLeader()
 }
 
 // IsClosed checks if the server loop is closed
 func (s *Server) IsClosed() bool {
-	return atomic.LoadInt64(&s.isServing) == 0
+	return s != nil && atomic.LoadInt64(&s.isServing) == 0
```
lhy1024, line 249: maybe we also  should rename the atomic `isServing`, because it is different but similar with `IsServing()`

pkg/mcs/resource_manager/server/server.go
```diff
@@ -403,14 +399,14 @@ func (s *Server) startServer() (err error) {
...
 	// Server has started.
-	atomic.StoreInt64(&s.isServing, 1)
 	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.ListenAddr}
 	serializedEntry, err := entry.Serialize()
 	if err != nil {
 		return err
 	}
 	s.serviceRegister = discovery.NewServiceRegister(s.ctx, s.etcdClient, utils.ResourceManagerServiceName, s.cfg.ListenAddr, serializedEntry, discovery.DefaultLeaseInSeconds)
 	s.serviceRegister.Register()
+	atomic.StoreInt64(&s.isServing, 1)
```
lhy1024, line 409: why move it

client/client.go
```diff
@@ -272,56 +281,16 @@ func NewClient(svrAddrs []string, security SecurityOption, opts ...ClientOption)
...
-		c.cancel()
-		return nil, err
-	}
-	if err := c.tsoClient.setup(); err != nil {
-		c.cancel()
-		return nil, err
-	}
-	return c, nil
+	// pass 0 as the default keyspace id
+	return NewClientWithKeyspace(ctx, 0, svrAddrs, security, opts...)
```
lhy1024, line 287: use constant.go?

client/pd_service_discovery.go
```diff
@@ -345,14 +374,29 @@ func (c *pdServiceDiscovery) initClusterID() error {
...
 
+func (c *pdServiceDiscovery) updateServiceMode() {
+	leaderAddr := c.getLeaderAddr()
+	if len(leaderAddr) > 0 {
+		clusterInfo, err := c.getClusterInfo(c.ctx, leaderAddr, c.option.timeout)
+		if err != nil {
+			log.Warn("[pd] failed to get cluster info for the leader", zap.String("leader-addr", leaderAddr), errs.ZapError(err))
+			return
+		}
+		c.serviceModeUpdateCb(clusterInfo.ServiceModes[0])
```
lhy1024, line 385: we only use [0]?

tests/mcs/tso/server_test.go
binshi-bing: changed back

server/server.go
binshi-bing: When calling new Participant. For now, we only save one url. using array is for extensibility in future.

pkg/mcs/resource_manager/server/server.go
binshi-bing: We can do it in a separate pr to refactor all of them. Some func and variable names look confusing.

pkg/mcs/resource_manager/server/server.go
binshi-bing: We should move it, shouldn't we? After we register the service, then change the state to serving, will be more reasonable.

client/pd_service_discovery.go
binshi-bing: Yes, for now.

binshi-bing: Either API or PD. The array is for extensibility in future.

client/client.go
binshi-bing: This client package, and it seems that the client won't use  an server side package -- I have confirmed this with Yutong.

binshi-bing: but I just defined a constant value in client.go and  use it here.

Changed to NewClientWithKeyspace(ctx, defaultKeyspaceID, svrAddrs, security, opts...)

pkg/basicserver/basic_server.go
binshi-bing: Yes, so any problem with that? Using []string is also for keeping compatibility with both pdpb.Member and tsopb.Participant. For former, it's clientUrls which is an array. For the latter, it's the listenUrl, a.k.a, listenUrls[0], and we may extend it to multiple addresses of participants which join the election group and use listenUrls[0] as the primary.

pkg/mcs/resource_manager/server/server.go
binshi-bing: By the way, the order doesn't impact any running result. It seems to me that, starting the service, registering the service then changing the state to 'isServing' seems to be more reasonable.

client/client.go
```diff
@@ -376,6 +367,67 @@ func (c *client) Close() {
...
+	log.Info("service mode changed", zap.String("old-mode", pdpb.ServiceMode_name[int32(c.serviceMode)]),
+		zap.String("new-mode", pdpb.ServiceMode_name[int32(newMode)]))
+	c.serviceMode = newMode
+}
+
+func (c *client) getTSOClient() *tsoClient {
+	if tsoCli := c.tsoClient.Load(); tsoCli != nil {
+		return tsoCli.(*tsoClient)
+	}
+	return nil
```
JmPotato, line 428: Just returning `nil` is not a good idea as we have already met some similar panics before.

binshi-bing: This is to retrieve a pointer from an atomic value, so there is no need to differentiate the default value and non-existed cases. The caller anyway needs to check if the returned pointer is nil or not.

There are three places calling this function getTSOClient -- two places have checked if the returned pointer is nil and the third place doesn't check (I just added the nil check at the third place anyway). The third place doesn't check because once PD client is initialized, c.tsoClient will always have a non-nil pointer.

pkg/mcs/resource_manager/server/server.go
```diff
@@ -403,14 +399,17 @@ func (s *Server) startServer() (err error) {
...
-	atomic.StoreInt64(&s.isServing, 1)
 	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.ListenAddr}
 	serializedEntry, err := entry.Serialize()
 	if err != nil {
 		return err
 	}
 	s.serviceRegister = discovery.NewServiceRegister(s.ctx, s.etcdClient, utils.ResourceManagerServiceName, s.cfg.ListenAddr, serializedEntry, discovery.DefaultLeaseInSeconds)
-	s.serviceRegister.Register()
+	if err := s.serviceRegister.Register(); err != nil {
+		log.Error("failed to regiser the service", zap.String("service-name", utils.TSOServiceName), errs.ZapError(err))
```
lhy1024, line 409: This should be resource management

pkg/mcs/tso/server/server.go
```diff
@@ -324,12 +328,12 @@ func (s *Server) AddStartCallback(callbacks ...func()) {
...
 // if there is embedded etcd, or the primary otherwise.
 func (s *Server) IsServing() bool {
-	return s.participant.IsLeader() && atomic.LoadInt64(&s.isServing) == 1
+	return atomic.LoadInt64(&s.isServing) == 1 && s.participant.IsLeader()
 }
 
-// GetPrimary returns the primary provider of this tso server.
-func (s *Server) GetPrimary() bs.MemberProvider {
-	return s.participant.GetLeader()
+// GetLeaderListenUrls gets service endpoints from the leader in election group.
```
lhy1024, line 334: Do we need also to add comment for index 0?

server/server.go
```diff
@@ -1732,10 +1737,16 @@ func (s *Server) watchServicePrimaryAddrLoop(serviceName string) {
...
 				case clientv3.EventTypePut:
-					if err := proto.Unmarshal(event.Kv.Value, leader); err != nil {
+					primary.ListenUrls = nil // reset the field
+					if err := proto.Unmarshal(event.Kv.Value, primary); err != nil {
 						log.Error("watch service primary addr failed", zap.String("service-key", serviceKey), zap.Error(err))
 					} else {
-						s.servicePrimaryMap.Store(serviceName, leader.GetName())
+						listenUrls = primary.GetListenUrls()
+						if len(listenUrls) > 0 {
+							s.servicePrimaryMap.Store(serviceName, listenUrls[0])
```
lhy1024, line 1746: Please add comment for index 0 too

client/client.go
```diff
@@ -1096,6 +1156,9 @@ func addrsToUrls(addrs []string) []string {
 
 // IsLeaderChange will determine whether there is a leader change.
 func IsLeaderChange(err error) bool {
+	if err == errs.ErrClientTSOStreamClosed {
```
lhy1024, line 1159: Why judge it?

lhy1024 (commented):
The rest lgtm

pkg/mcs/resource_manager/server/server.go
binshi-bing: changed

pkg/mcs/tso/server/server.go
binshi-bing: added "// The entry at the index 0 is the primary's service endpoint."

pkg/mcs/resource_manager/server/server.go
binshi-bing: good catch!

server/server.go
binshi-bing: added

client/client.go
binshi-bing: There is specific change for this. When TSO service switched, the stream(TSO_TSOClient or PD_TSOClient)'s send/recv could return io.EOF. In this case, we need to treat it as LeaderChanged and update leader address immediate. Previously, we didn't handle this case, and now this change handles it.

lhy1024 (approved):

JmPotato (approved):

PR #6162, opened 2023-03-14
pkg: move operator_check out from test_util

lhy1024:

Issue Number: Ref #4399

remove unnecessary dependency for test_util.

before this pr, test_util need import operator

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/utils/operatorutil/operator_check.go
bufferflies, line 1: I think the path should be `/utils/operator/check.go `, the util is too much.

bufferflies (approved):
rest LGTM

nolouch (approved):

pkg/utils/operatorutil/operator_check.go
lhy1024: I think current path is acceptable

PR #6166, opened 2023-03-15
mcs: use cluster ID for election

rleungx:

Issue Number: Ref #5836.

This PR uses cluster id for the election instead of using 0.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1667,8 +1667,7 @@ func (s *Server) UnmarkSnapshotRecovering(ctx context.Context) error {
 
 // GetServicePrimaryAddr returns the primary address for a given service.
 func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string) (bool, string, error) {
-	// TODO: replace default group name after we make a decision.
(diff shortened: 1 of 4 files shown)


lhy1024 (approved):

server/server.go
```diff
@@ -1667,8 +1667,7 @@ func (s *Server) UnmarkSnapshotRecovering(ctx context.Context) error {
 
 // GetServicePrimaryAddr returns the primary address for a given service.
 func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string) (bool, string, error) {
-	// TODO: replace default group name after we make a decision.
-	key := path.Join("/ms/0", serviceName, fmt.Sprintf("%05d", 0), "primary")
+	key := fmt.Sprintf("/ms/%d/%s/%s/%s", s.clusterID, serviceName, fmt.Sprintf("%05d", 0), "primary")
```
JmPotato, line 1670: Will this break compatibility?

rleungx: It's not used for now, so I think there is no compatibility issue.

JmPotato (approved):

PR #6183, opened 2023-03-20
tests: make the TSO testing as independent as possible

JmPotato:
### What problem does this PR solve?

Issue Number: ref #6181.

### What is changed and how does it work?

```commit-message
- Support testing both legacy and microservice TSO.
- Make the TSO testing as independent as possible.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

Issue #6181, opened 2023-03-20
Integrate the multi-tenant TSO to the TSO consistency test


JmPotato:
Part of #5895.

Currently, the TSO consistency test only runs on the legacy TSO mode. We should integrate the multi-tenant TSO to it also.

PR #6198, opened 2023-03-21
tests: divide all tests into the CI chunks including submodule tests

JmPotato:

Issue Number: ref #6181.

After #6183, the time cost for `chunks (10)` occasionally exceeds 10 minutes. This PR ensures that submodule tests can also be segmented into different chunks to enhance overall test speed.

```commit-message
Divide all tests into the CI chunks including submodule tests.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


Makefile
```diff
@@ -216,12 +216,7 @@ basic-test: install-tools
 ci-test-job: install-tools dashboard-ui
 	@$(FAILPOINT_ENABLE)
 	CGO_ENABLED=1 go test -timeout=15m -tags deadlock -race -covermode=atomic -coverprofile=covprofile -coverpkg=./... $(shell ./scripts/ci-subtask.sh $(JOB_COUNT) $(JOB_INDEX))
-	@$(FAILPOINT_DISABLE)
-
-ci-test-job-submod: install-tools dashboard-ui
-	@$(FAILPOINT_ENABLE)
-	@ for mod in $(SUBMODULES); do cd $$mod && $(MAKE) ci-test-job && cd $(ROOT_PATH) > /dev/null && cat $$mod/covprofile >> covprofile; done
-	@$(FAILPOINT_DISABLE)
+	@ for mod in $(shell ./scripts/ci-subtask.sh $(JOB_COUNT) $(JOB_INDEX) 1); do cd $$mod && $(MAKE) ci-test-job && cd $(ROOT_PATH) > /dev/null && cat $$mod/covprofile >> covprofile; done
```
HuSharp, line 219: Do we also need `@$(FAILPOINT_DISABLE)` or just by `clean`?

JmPotato: This task is only used for the CI enviroment, so there is no need to do the clean-up work.

tests/integrations/client/Makefile
```diff
@@ -12,37 +12,35 @@
...
+test: failpoint-enable
+	CGO_ENABLED=1 go test -v -tags deadlock -race -cover || { $(MAKE) failpoint-disable && exit 1; }
+	$(MAKE) failpoint-disable
 
-basic-test:
-	# skip
-
-ci-test-job: enable-codegen
-	CGO_ENABLED=1 go test -tags deadlock -race -covermode=atomic -coverprofile=covprofile -coverpkg=../../... github.com/tikv/pd/tests/client
+ci-test-job:
```
rleungx, line 34: Do we need to enable the failpoint?

JmPotato: It will be enabled by the root Makefile, so we don't need to enable it here again.

rleungx:
I find that chunk 1 is much longer than before.

JmPotato:
> I find that chunk 1 is much longer than before.

I manually hardcode the chunk index here, it seems that the speed is ok now.

nolouch (approved):

rleungx (approved):

Issue #5895, opened 2023-02-02
Support Multi-tenant and Sharded TSO


binshi-bing:
### Describe your feature request related problem

The current TSO solution is known for its poor scalability and single point of failure. As we are moving to 'Cloud' providing DBaaS and Serverless Computing, supporting multi-tenant is one of primary goals which brings more requirements to the TSO service, including:
1. Horizontal scalability across tenants.
2. High resilience and reliability across tenants, i.e., reduce Blast Radius

### Describe the feature you'd like
By sharding TSO service across tenants, we aim to achieve the following goals:
1. Provide unlimited horizontal scalability across all tenants.
2. In all failure scenarios, there is no single point failure for all the tenants.
3. Provide no-degraded service during rolling update deployment and normal operations.  
4. Keep compatibility with the existing features.

#### Big Picture
<img width="2330" alt="image" src="https://user-images.githubusercontent.com/39923490/216216206-79aa7cf8-e555-4591-9777-ad58204339e9.png">

1. A TSO node can manage multiple tenant groups.
2. Each tenant group is an independent TSO service instance.
3. Each tenant group has multiple replicas which are placed in different failure domains to provide HA.
4. The following utility tools will be provided from command console, control plane and even web admin UI.
     - The balancer --  manages the sharding state of the TSO service via etcd.
     - The cluster membership manager -- manages membership of the TSO nodes.
     - The tenant group membership manager -- manages the mapping of tenant -> tenant group.

For more details, please refer to the RFC (TODO: add link to the RFC).

#### Milestone 1 - (goal) deliver single-group tso microservice (**Completed**)
All work are tracked in [#5836](https://github.com/tikv/pd/issues/5836#issuecomment-1412980551) 

#### Milestone 2 - (goal) code complete for multi-group tso microservice (4/21/2023) (**Completed**)
- [x] (02/10) drive design review; finalize design; write RFC @binshi-bing
- [x] **Develop multi-tenant and sharded TSO Service**
    - [x] (04/17) #6231 @rleungx @binshi-bing
    - [x] (04/21) #6232 @binshi-bing @JmPotato @rleungx  @lhy1024 
    - [x] (04/17) #6233 @lhy1024

#### Milestone 3 - (goal) deploy multi-group tso microservice in dev env (5/5/2023) (Note: May 1st - 3rd, holiday)
- [x] tests
    - [x] Add multi-group tests utilities @JmPotato @rleungx 
[...]

PR #6199, opened 2023-03-21
*: use txn for saving timestamp

rleungx:

Issue Number: Ref #5895.

This PR supports using Txn to save the timestamp to prevent TSO from falling back when two places advance at the same time.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -29,7 +29,7 @@ import (
 // TSOStorage is the interface for timestamp storage.
 type TSOStorage interface {
 	LoadTimestamp(prefix string) (time.Time, error)
-	SaveTimestamp(prefix string, key string, ts time.Time) error
+	SaveTimestamp(key string, ts time.Time) error
 }
 
 var _ TSOStorage = (*StorageEndpoint)(nil)
@@ -66,26 +66,18 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
 }
 
 // SaveTimestamp saves the timestamp to the storage.
-func (se *StorageEndpoint) SaveTimestamp(prefix string, key string, ts time.Time) error {
+func (se *StorageEndpoint) SaveTimestamp(key string, ts time.Time) error {
 	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
-		prefixEnd := clientv3.GetPrefixRangeEnd(prefix)
-		keys, values, err := txn.LoadRange(prefix, prefixEnd, 0)
+		value, err := txn.Load(key)
 		if err != nil {
 			return err
 		}
 
 		previousTS := typeutil.ZeroTime
-		for i, key := range keys {
-			key := strings.TrimSpace(key)
-			if !strings.HasSuffix(key, timestampKey) {
-				continue
-			}
-			tsWindow, err := typeutil.ParseTimestamp([]byte(values[i]))
+		if value != "" {
+			previousTS, err = typeutil.ParseTimestamp([]byte(value))
 			if err != nil {
-				continue
-			}
-			if typeutil.SubRealTimeByWallClock(tsWindow, previousTS) > 0 {
-				previousTS = tsWindow
+				log.Warn("parse timestamp failed", zap.String("key", key), zap.String("value", value), zap.Error(err))
 			}
 		}
 		if previousTS != typeutil.ZeroTime && typeutil.SubRealTimeByWallClock(ts, previousTS) <= 0 {
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -77,11 +77,11 @@ func (se *StorageEndpoint) SaveTimestamp(key string, ts time.Time) error {
 		if value != "" {
 			previousTS, err = typeutil.ParseTimestamp([]byte(value))
 			if err != nil {
-				log.Warn("parse timestamp failed", zap.String("key", key), zap.String("value", value), zap.Error(err))
+				log.Error("parse timestamp failed", zap.String("key", key), zap.String("value", value), zap.Error(err))
+				return err
 			}
 		}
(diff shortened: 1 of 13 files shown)


lhy1024 (approved):

JmPotato (approved):

lhy1024:
ci failed

pkg/storage/endpoint/tso.go
```diff
@@ -64,7 +66,33 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
 }
 
 // SaveTimestamp saves the timestamp to the storage.
-func (se *StorageEndpoint) SaveTimestamp(key string, ts time.Time) error {
-	data := typeutil.Uint64ToBytes(uint64(ts.UnixNano()))
-	return se.Save(key, string(data))
+func (se *StorageEndpoint) SaveTimestamp(prefix string, key string, ts time.Time) error {
+	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
+		prefixEnd := clientv3.GetPrefixRangeEnd(prefix)
+		keys, values, err := txn.LoadRange(prefix, prefixEnd, 0)
```
binshi-bing, line 72: If my understanding is correct, for global location, we use '/pd/[cluster-id]' to do prefix scan, but there are a lot of things, under this path, irrelevant to tso. In order to "“LoadTimestamp will get all time windows of Local/Global TSOs from etcd and return the biggest one.”, shall we use the /pd/<cluster-id>/timestamp/<dc-X> for local TSOs?

pkg/storage/endpoint/tso.go
```diff
@@ -64,7 +66,33 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+		if err != nil {
+			return err
+		}
+
+		previousTS := typeutil.ZeroTime
+		for i, key := range keys {
+			key := strings.TrimSpace(key)
+			if !strings.HasSuffix(key, timestampKey) {
```
binshi-bing, line 80: Isn't this condition "!strings.HasSuffix(key, timestampKey)" too loose, considering we're using "/pd/[cluster-id]" to do prefix range scan? Could there be chance that 'timestamp' suffix are added to the path for other use cases?

pkg/storage/endpoint/tso.go
```diff
@@ -64,7 +66,33 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+			if typeutil.SubRealTimeByWallClock(tsWindow, previousTS) > 0 {
+				previousTS = tsWindow
+			}
+		}
+		if previousTS != typeutil.ZeroTime && typeutil.SubRealTimeByWallClock(ts, previousTS) <= 0 {
+			log.Warn("save timestamp failed, the timestamp is not bigger than the previous one", zap.Time("previous", previousTS), zap.Time("current", ts))
+			return nil
+		}
+		data := typeutil.Uint64ToBytes(uint64(ts.UnixNano()))
+		return txn.Save(key, string(data))
```
binshi-bing, line 96: Because LoadTimestamp() use the logic to do prefix scan then return the largest timestamp among all timestamps, it seems that we don't need the whole change in this pr -- tso service doesn't need to use the same path as pd's but the same prefix for timestamp range scan, then anyway  LoadTimestamp()  will return the largest one among the last writes of pd and tso service.

binshi-bing (commented):
left a few comments.

pkg/storage/endpoint/tso.go
```diff
@@ -65,6 +67,24 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
-	data := typeutil.Uint64ToBytes(uint64(ts.UnixNano()))
-	return se.Save(key, string(data))
+	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
+		value, err := txn.Load(key)
+		if err != nil {
+			return err
+		}
+
+		previousTS := typeutil.ZeroTime
+		if value != "" {
```
binshi-bing, line 77:
```suggestion
		if len(value) > 0 {
```

pkg/storage/endpoint/tso.go
```diff
@@ -65,6 +67,24 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+			previousTS, err = typeutil.ParseTimestamp([]byte(value))
+			if err != nil {
+				log.Warn("parse timestamp failed", zap.String("key", key), zap.String("value", value), zap.Error(err))
+			}
+		}
+		if previousTS != typeutil.ZeroTime && typeutil.SubRealTimeByWallClock(ts, previousTS) <= 0 {
+			log.Warn("save timestamp failed, the timestamp is not bigger than the previous one", zap.Time("previous", previousTS), zap.Time("current", ts))
```
binshi-bing: Do we need this log line? It seems to be a normal case.

binshi-bing (approved):
LGTM

pkg/storage/endpoint/tso.go
```diff
@@ -64,7 +66,33 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+				previousTS = tsWindow
+			}
+		}
+		if previousTS != typeutil.ZeroTime && typeutil.SubRealTimeByWallClock(ts, previousTS) <= 0 {
+			log.Warn("save timestamp failed, the timestamp is not bigger than the previous one", zap.Time("previous", previousTS), zap.Time("current", ts))
+			return nil
+		}
+		data := typeutil.Uint64ToBytes(uint64(ts.UnixNano()))
+		return txn.Save(key, string(data))
+	})
```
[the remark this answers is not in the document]
rleungx: already change to `Load` which will only get one key.

pkg/storage/endpoint/tso.go
rleungx: I think there is not much difference.

lhy1024:
ci failed

PR #6201, opened 2023-03-21
client: refine serviceModeKeeper code

JmPotato:

Issue Number: ref #5895.

```commit-message
Some code refinements for `serviceModeKeeper`.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -240,15 +240,39 @@ func WithMaxErrorRetry(count int) ClientOption {
 
 var _ Client = (*client)(nil)
 
+// serviceModeKeeper is for service mode switching.
+type serviceModeKeeper struct {
+	// RMutex here is for the future usage that there might be multiple goroutines
+	// triggering service mode switching concurrently.
+	sync.RWMutex
+	serviceMode     pdpb.ServiceMode
+	tsoClient       atomic.Value // *tsoClient
+	tsoSvcDiscovery ServiceDiscovery
+}
+
+func (smk *serviceModeKeeper) close() {
+	smk.Lock()
+	defer smk.Unlock()
+	switch smk.serviceMode {
+	case pdpb.ServiceMode_API_SVC_MODE:
+		smk.tsoSvcDiscovery.Close()
+		fallthrough
+	case pdpb.ServiceMode_PD_SVC_MODE:
+		if tsoCli := smk.tsoClient.Load(); tsoCli != nil {
+			tsoCli.(*tsoClient).Close()
+		}
+	case pdpb.ServiceMode_UNKNOWN_SVC_MODE:
+	}
+}
+
 type client struct {
 	keyspaceID      uint32
 	svrUrls         []string
 	pdSvcDiscovery  ServiceDiscovery
 	tokenDispatcher *tokenDispatcher
 
-	serviceMode     pdpb.ServiceMode
-	tsoClient       atomic.Value // *tsoClient
-	tsoSvcDiscovery ServiceDiscovery
+	// For service mode switching.
+	serviceModeKeeper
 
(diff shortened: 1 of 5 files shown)


rleungx (approved):

client/client.go
```diff
@@ -364,57 +359,56 @@ func (c *client) Close() {
 }
 
 func (c *client) setServiceMode(newMode pdpb.ServiceMode) {
-	c.svcModeMutex.Lock()
```
binshi-bing, line 367: I prefer to keep this lock. Originally, I used the corresponding read lock in the getTSOClient, but eventually I used atomic.Value() to store tsoClient and removed the read lock. After that, I still decided to keep this lock and critical section in the setServiceMode() for the following reasons:
1. Have plan to improve error handling. After service mode switched at the server side, the requests will fail and the client needs to call serviceModeUpdate() directly and immediately instead of waiting serviceModeUpdateInterval 3s (I even changed the interval from the original 10s to the current value 3s) for this purpose. When the leader/primary being switched on the server side, we had the similar issue -- for this case, today we fail the requests immediately without retries though we schedule membership check immediately. All these need improvement.
2. This lock's cost can be overlooked (invoked every 3s) and it's inconvenient to add it back.
3. I prefer to move client.setServiceMode to serviceModeKeeper.setServiceMode so that it can be more explicit for changing serviceMode, tsoClient and tsoSvcDiscovery atomically, thus it's more modular.

binshi-bing (commented):
left few comments. We can discuss in the group channel.

client/client.go
JmPotato: I added the lock back. PTAL.

client/tso_client.go
```diff
@@ -91,8 +91,11 @@ type tsoClient struct {
 }
 
 // newTSOClient returns a new TSO client.
-func newTSOClient(ctx context.Context, cancel context.CancelFunc, option *option, keyspaceID uint32,
-	svcDiscovery ServiceDiscovery, eventSrc tsoAllocatorEventSource, factory tsoStreamBuilderFactory) *tsoClient {
+func newTSOClient(
```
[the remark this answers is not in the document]
JmPotato: There are no strong principles here, I just want to make the parameters group as lines to reduce the one-line length.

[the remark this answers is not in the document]
binshi-bing: ok. I don't have strong preference. Your change here looks good to me.

client/client.go
binshi-bing: Sounds good. The change looks good to me.

client/client.go
```diff
@@ -364,57 +378,56 @@ func (c *client) Close() {
...
-	log.Info("service mode changed", zap.String("old-mode", pdpb.ServiceMode_name[int32(c.serviceMode)]),
-		zap.String("new-mode", pdpb.ServiceMode_name[int32(newMode)]))
+	oldTSOClient.Close()
+	// Replace the old TSO service discovery if needed.
+	oldTSOSvcDiscovery := c.tsoSvcDiscovery
+	if newTSOSvcDiscovery != nil {
+		c.tsoSvcDiscovery = newTSOSvcDiscovery
+		// Close the old TSO service discovery safely after both the old client
+		// and service discovery are replaced.
+		if oldTSOSvcDiscovery != nil {
```
binshi-bing, line 423: Placeing "if oldTSOSvcDiscovery != nil {...}" here is very accurate, but reduced readiness.

binshi-bing: either way is ok.

binshi-bing (approved):

PR #6206, opened 2023-03-22
resource_control: create the default resource group

nolouch:

Issue Number: ref #4399 ref https://github.com/pingcap/tidb/issues/42403 

```commit-message
resource_control: create the default resource group
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/mcs/resource_manager/server/manager.go
```diff
@@ -37,6 +37,9 @@ const (
 	defaultConsumptionChanSize = 1024
 	metricsCleanupInterval     = time.Minute
 	metricsCleanupTimeout      = 20 * time.Minute
+
+	defaultGroupName = "default"
```
JmPotato, line 41: I suggest adding a `reserved` to indicate it's an internal use resource group name, otherwise, it may be regarded as some default name setting.

```suggestion
	reservedDefaultGroupName = "default"
```

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -121,6 +124,23 @@ func (m *Manager) Init(ctx context.Context) {
 		}
 	}
 	m.storage.LoadResourceGroupStates(tokenHandler)
+
+	// add default group
+	if _, ok := m.groups[defaultGroupName]; !ok {
```
JmPotato, line 129: Is it possible that we overwrite some user-modified config here?

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -121,6 +124,23 @@ func (m *Manager) Init(ctx context.Context) {
...
+		m.groups[defaultGroupName] = &ResourceGroup{
```
BornChanger, line 130: So default resource group is in memory only?

glorv (approved):

pkg/mcs/resource_manager/server/manager.go
nolouch: It will persists periodically.

pkg/mcs/resource_manager/server/manager.go
nolouch: Here load groups firstly, if user modified, the default group should already exists.

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -121,6 +124,23 @@ func (m *Manager) Init(ctx context.Context) {
 		}
 	}
 	m.storage.LoadResourceGroupStates(tokenHandler)
+
+	// add default group
+	if _, ok := m.groups[reservedDefaultGroupName]; !ok {
```
JmPotato, line 129: Why not use the `AddResourceGroup` to update? So it could persist also.

tests/mcs/resource_manager/resource_manager_test.go
```diff
@@ -128,6 +128,9 @@ func (suite *resourceManagerClientTestSuite) cleanupResourceGroups() {
 	groups, err := cli.ListResourceGroups(suite.ctx)
 	suite.NoError(err)
 	for _, group := range groups {
+		if group.Name == "default" {
```
CabinfeverB, line 131: Can we refuse to delete or modify the default resource group in `manager.go`?

BornChanger: It's already blocked at TiDB side.

JmPotato (approved):

tests/mcs/resource_manager/resource_manager_test.go
nolouch: I also added in pd side.  ptal @CabinfeverB

nolouch:
ptal @CabinfeverB

CabinfeverB (approved):

PR #6212, opened 2023-03-23
mcs: support advertise address

rleungx:

Issue Number: Ref #5836.

This PR supports the advertise address for tso server and resource manager server.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/tso/server/server.go
```diff
@@ -574,7 +574,7 @@ func (s *Server) startServer() (err error) {
 
 	s.participant = member.NewParticipant(s.etcdClient)
 	s.participant.InitInfo(uniqueName, uniqueID, fmt.Sprintf(tsoSvcDiscoveryPrefixFormat, s.clusterID, mcsutils.DefaultKeyspaceID),
```
binshi-bing, line 576: Do we need to add AdvertiseListenAddr to tsopb.Participant?

binshi-bing: It seems that tsopb.Participant just need to store AdvertiseListenAddr as ListenAddr for this case?

binshi-bing (approved):
LGTM

pkg/mcs/tso/server/server.go
rleungx: Yes. There is no need to change the definition of `Participant`.

JmPotato (approved):

Issue #6231, opened 2023-03-27
Keyspace group membership management


binshi-bing:
## Development Task
- [x] Complete the design @binshi-bing @rleungx 
- [x] Implement keyspace group membership management RESTful APIs @rleungx 
- [ ] Implement keyspace group membership management utility @rleungx
- [x] Add storage endpoint for keyspace group @rleungx
- [ ] Support the keyspace group membership management in pd-ctl

Issues: Ref #5895

zhangjinpeng1987:
Do not use internal link. @binshi-bing

PR #6241, opened 2023-03-29
*: define user kind for keyspace group

rleungx:

Issue Number: Ref #6231.

This PR defines the user kind for the keyspace group and adds the interface for assigning the group ID to the keyspace.

```commit-message
```

### Check List

Tests

- Unit test

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -46,6 +46,10 @@ const (
 	regionLabelIDPrefix = "keyspaces/"
 	// regionLabelKey is the key for keyspace id in keyspace region label.
 	regionLabelKey = "id"
+	// UserKindKey is the key for user kind in keyspace config.
+	UserKindKey = "user_kind"
+	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
+	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
 )
 
 // Config is the interface for keyspace config.
@@ -68,6 +72,7 @@ type Manager struct {
 	ctx context.Context
 	// config is the configurations of the manager.
 	config Config
+	kgm    *GroupManager
 }
 
 // CreateKeyspaceRequest represents necessary arguments to create a keyspace.
@@ -85,6 +90,7 @@ func NewKeyspaceManager(store endpoint.KeyspaceStorage,
 	cluster schedule.Cluster,
 	idAllocator id.Allocator,
 	config Config,
+	kgm *GroupManager,
 ) *Manager {
 	return &Manager{
 		metaLock:    syncutil.NewLockGroup(syncutil.WithHash(keyspaceIDHash)),
@@ -93,6 +99,7 @@ func NewKeyspaceManager(store endpoint.KeyspaceStorage,
 		cluster:     cluster,
 		ctx:         context.TODO(),
 		config:      config,
+		kgm:         kgm,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/storage/endpoint/tso_keyspace_group.go
```diff
@@ -22,6 +22,44 @@ import (
 	"go.etcd.io/etcd/clientv3"
 )
 
+// UserKind represents the user kind.
+type UserKind int
+
+// Different user kinds.
+const (
```
binshi-bing, line 29: shall we define it as enum in keyspace.proto then we get all the following functions for free?

binshi-bing (commented):
left one comment

server/keyspace/keyspace.go
```diff
@@ -98,14 +106,22 @@ func (manager *Manager) Bootstrap() error {
...
+	if err != nil {
+		return err
+	}
 	defaultKeyspace := &keyspacepb.KeyspaceMeta{
 		Id:             DefaultKeyspaceID,
 		Name:           DefaultKeyspaceName,
 		State:          keyspacepb.KeyspaceState_ENABLED,
 		CreatedAt:      now,
 		StateChangedAt: now,
+		Config: map[string]string{
```
binshi-bing, line 119: How do we add UserKind to the existing keyspaces? I didn't see they have UpdateKeyspace API. UpdateKeyspaceState API doesn't apply.

binshi-bing: Regarding group id, my understanding is that AddKeyspacesToGroups() API will do the assignment and ad d group id to the existing keyspaces, which is ok.

pkg/storage/endpoint/tso_keyspace_group.go
rleungx: Good question, we need a discussion about it.

server/keyspace/keyspace.go
rleungx: Will AddKeyspacesToGroups be exposed?

pkg/storage/endpoint/tso_keyspace_group.go
rleungx: From my perspective, it's ok to keep it for now since we don't know if the current definition meets the future requirement. It's easier to modify compared with using proto.

lhy1024 (approved):

pkg/keyspace/keyspace.go
```diff
@@ -46,6 +46,10 @@ const (
 	regionLabelIDPrefix = "keyspaces/"
 	// regionLabelKey is the key for keyspace id in keyspace region label.
 	regionLabelKey = "id"
+	// UserKindKey is the key for user kind in keyspace config.
+	UserKindKey = "user_kind"
+	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
+	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
```
JmPotato, line 52: Suggest reflecting the hierarchy from the naming, so maybe `KeyspaceGroupIDTSOKey` is a better name.

pkg/keyspace/keyspace.go
```diff
@@ -68,6 +72,7 @@ type Manager struct {
 	ctx context.Context
 	// config is the configurations of the manager.
 	config Config
+	kgm    *GroupManager
```
JmPotato, line 75: This manager has the same name as:

https://github.com/tikv/pd/blob/9fcde123fd7c7baddc6d1898dfc07cf2a461f6f7/pkg/tso/keyspace_group_manager.go#L57

I strongly feel like we should unify them or give this keyspace group storage manager another name.

pkg/keyspace/keyspace.go
```diff
@@ -148,6 +171,16 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	if err != nil {
 		return nil, err
 	}
+	userKind := endpoint.StringUserKind(request.Config[UserKindKey])
+	id, err := manager.kgm.GetAvailableKeyspaceGroupIDByKind(userKind)
+	if err != nil {
+		return nil, err
+	}
+	if request.Config == nil {
```
JmPotato, line 179: Why do we do the nil check right after we try to get a key from it at line 174?

rleungx: The nil map can be accessed but cannot be assigned.

pkg/keyspace/keyspace.go
rleungx: But it sounds strange. lol

pkg/keyspace/keyspace.go
rleungx: They are two different things. And any suggestion for the name?

JmPotato (approved):

PR #6248, opened 2023-03-30
*: unify the key prefix

rleungx:

Issue Number: Ref #5836.

Unify the micro service key prefix, we use /ms and cluster ID as the prefix.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/tso/server/server.go
+++ b/pkg/mcs/tso/server/server.go
@@ -448,14 +448,15 @@ func (s *Server) startServer() (err error) {
 	}
 
 	// Server has started.
-	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.ListenAddr}
+	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.AdvertiseListenAddr}
 	serializedEntry, err := entry.Serialize()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


pkg/mcs/tso/server/server.go
```diff
@@ -448,14 +448,15 @@ func (s *Server) startServer() (err error) {
 	}
 
 	// Server has started.
-	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.ListenAddr}
+	entry := &discovery.ServiceRegistryEntry{ServiceAddr: s.cfg.AdvertiseListenAddr}
```
binshi-bing, line 451: shall we remove http:// or https://? Probably doesn't matter

binshi-bing (approved):

pkg/mcs/tso/server/server.go
rleungx: I think it doesn't matter unless we have a strong inclination about it.

binshi-bing: bette remove http://, because register use path.Join, and path.Join("aaa", "http://bbb/")返回"aaa/**http:/**bbb" in which it contains http:/ instead of http://

lhy1024: do we need to add related test?

lhy1024 (approved):
the rest LGTM

PR #6249, opened 2023-03-30
*: remove cluster dependency from keyspace

rleungx:

Issue Number: Ref #6231.

Previously, the keyspace is only initialized when bootstrapping the cluster. If we upgrade the cluster from an old version that doesn't support keyspace. The default keyspace won't be initialized.
This PR moves the keyspace to pkg and changes the bootstrap process.

**Update: only move the keyspace to pkg**

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/cluster/cluster.go
```diff
@@ -276,6 +279,14 @@ func (c *RaftCluster) Start(s Server) error {
 		return err
 	}
 
+	if err := s.GetKeyspaceGroupManager().Bootstrap(); err != nil {
+		log.Warn("bootstrap keyspace group manager failed", errs.ZapError(err))
```
binshi-bing, line 283:
```suggestion
		log.Warn("failed to bootstrap keyspace group manager", errs.ZapError(err))
```

binshi-bing: Or
```suggestion
		log.Warn("bootstrapping keyspace group manager failed", errs.ZapError(err))
```

server/cluster/cluster.go
```diff
@@ -276,6 +279,14 @@ func (c *RaftCluster) Start(s Server) error {
...
+	}
+
+	if err := s.GetKeyspaceManager().Bootstrap(); err != nil {
+		log.Warn("bootstrap keyspace manager failed", errs.ZapError(err))
```
binshi-bing, line 287: ditto

binshi-bing (approved):
Left few comments. LGTM

rleungx:
I'm going to remove the changes that put the bootstrapping logic in the start process since we need to control the keyspace initialization in the cloud and op environment. /cc @AmoebaProtozoa @binshi-bing

rleungx:
This PR has two approvals but one of them doesn't have permission. So, changing the label to require-LGT1 to let it go on.

pkg/keyspace/keyspace.go
```diff
@@ -207,18 +212,21 @@ func (manager *Manager) splitKeyspaceRegion(id uint32) error {
 	})
 
 	keyspaceRule := makeLabelRule(id)
-	err := manager.rc.GetRegionLabeler().SetLabelRule(keyspaceRule)
-	if err != nil {
-		log.Warn("[keyspace] failed to add region label for keyspace",
+	if cl, ok := manager.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
```
lhy1024, line 215: why add this judgment?

lhy1024 (approved):
the rest LGTM

pkg/keyspace/keyspace.go
rleungx: remove cluster dependency

PR #6251, opened 2023-03-31
server: fix watch keyspace revision

rleungx:

Issue Number: Ref #5895.

Previously, we don't specify the revision of the watch, it may lose some events if we update the KV between `sendAllKeyspaceMeta` and `Watch`.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


nolouch (approved):

binshi-bing (approved):
LGTM

server/keyspace_service.go
```diff
@@ -82,16 +84,24 @@ func (s *KeyspaceServer) WatchKeyspaces(request *keyspacepb.WatchKeyspacesReques
...
 
-	err := s.sendAllKeyspaceMeta(ctx, stream)
+	revision, err := s.sendAllKeyspaceMeta(ctx, stream)
 	if err != nil {
 		return err
 	}
-	watchChan := s.client.Watch(ctx, path.Join(s.rootPath, endpoint.KeyspaceMetaPrefix()), clientv3.WithPrefix())
+
 	for {
+		watchChan := s.client.Watch(ctx, path.Join(s.rootPath, endpoint.KeyspaceMetaPrefix()), clientv3.WithPrefix(), clientv3.WithRev(revision))
```
binshi-bing, line 93: I suspect this implementation of watch (w/ revision) isn't efficient:
1. if watchChan isn't closed/cancelled or doesn't return error, we should continue reusing watchChan instead of calling s.client.Watch after watchChan returning PUT/DETELE event every time.
2. When calling s.client.Watch, it shouldn't use the revision get from line 87. The revision should be updated with res.Revision around line 105 - 107.

binshi-bing (commented):
left a comment

server/keyspace_service.go
binshi-bing, line 93: I suspect this implementation of watch (w/ revision) isn't efficient:
1. if watchChan isn't closed/cancelled or doesn't return error, we should continue reusing watchChan instead of calling s.client.Watch after watchChan returning PUT/DETELE event every time.
2. When calling s.client.Watch, it shouldn't use the revision get from line 87. The revision should be updated with res.Revision around line 105 - 107.

binshi-bing (changes requested):
please check my comment

server/keyspace_service.go
binshi-bing, line 93: If I'm right, /Users/binshi/code/pingcap/pd/pkg/encryption/key_manager.go has a good implementation of watch in StartBackgroundLoop()

rleungx: Oh, right. I intend to use for range.

server/keyspace_service.go
```diff
@@ -82,18 +85,32 @@ func (s *KeyspaceServer) WatchKeyspaces(request *keyspacepb.WatchKeyspacesReques
...
+					zap.Int64("required-revision", revision),
+					zap.Int64("compact-revision", wresp.CompactRevision))
+				revision = wresp.CompactRevision
+				break
+			}
+			if wresp.Canceled {
+				log.Error("watcher is canceled with",
+					zap.Int64("revision", revision),
+					errs.ZapError(errs.ErrEtcdWatcherCancel, wresp.Err()))
+				return nil
```
binshi-bing, line 110: if wresp.Canceled is true, does it mean watcher.Watch() was cancelled or the watcher (got from clientv3.NewWatcher) was cancelled? If it's the former, can we call watcher.Watch() again?

rleungx: I think when the ctx is canceled, both of them are canceled.

binshi-bing: OK, if wresp.Canceled is true, we don't need to call watcher.Watch() because this is only case that watcher won't call  watcher.Watch() by itself. I checked https://github.com/Homebrew/brew/blob/4.0.10/bin/go/pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20220915004622-85b640cee793/clientv3/watch.go.

For your case, you can just return error to the client and let the client to retry. For my case, keyspace group manager has to retry until etcd is recovered. Probablly I should call clientv3.NewWatcher(kgm.etcdClient) to create a new watcher.

binshi-bing (approved):
LGTM

rleungx:
@AmoebaProtozoa PTAL

AmoebaProtozoa (approved):
LGTM

nolouch (approved):

PR #6255, opened 2023-03-31
mcs: fix ctx in tso forward stream

### LINKED ISSUE #6254: mcs: update gc safe point doesn't work normally
### What did you do?

`~/pingcap/tiup/bin/tiup-playground v6.6.0  --pd.micro '[{"api":true, "amount":1},{"tso":true,"amount":1}]' --pd.binpath ~/pd-cse/bin/pd-server --db.binpath ~/tidb-cse/bin/tidb-server --kv 3 --db 1 --tiflash 0 --db.port 14000 --host 172.16.4.3`
### What did you see instead?
`[2023/03/31 18:18:00.646 +08:00] [ERROR] [gc_worker.go:744] ["[gc worker] failed to update service safe point"] [uuid=61cdc8da8100014] [error="rpc error: code = DeadlineExceeded desc = context deadline exceeded"]`
master

lhy1024:

Issue Number: Close #6254

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
`~/pingcap/tiup/bin/tiup-playground v6.6.0  --pd.micro '[{"api":true, "amount":1},{"tso":true,"amount":1}]' --pd.binpath ~/pd/bin/pd-server --db.binpath ~/tidb-cse/bin/tidb-server --kv 3 --db 1 --tiflash 0 --db.port 14000 --host 172.16.4.3`

not related error log.

### Release note

```release-note
None.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1766,7 +1766,7 @@ func (s *GrpcServer) getGlobalTSOFromTSOServer(ctx context.Context) (pdpb.Timest
 	if !ok || forwardedHost == "" {
 		return pdpb.Timestamp{}, ErrNotFoundTSOAddr
 	}
-	forwardStream, err := s.getTSOForwardStream(ctx, forwardedHost)
+	forwardStream, err := s.getTSOForwardStream(forwardedHost)
 	if err != nil {
 		return pdpb.Timestamp{}, err
 	}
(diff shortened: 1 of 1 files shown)


server/grpc_service.go
```diff
@@ -1804,18 +1805,18 @@ func (s *GrpcServer) getTSOForwardStream(ctx context.Context, forwardedHost stri
...
 
 	// Now let's create the client connection and the forward stream
-	client, err := s.getDelegateClient(ctx, forwardedHost)
+	client, err := s.getDelegateClient(s.ctx, forwardedHost)
 	if err != nil {
 		return nil, err
 	}
 	done := make(chan struct{})
-	ctx, cancel := context.WithTimeout(s.ctx, tsoutil.DefaultTSOProxyTimeout)
+	ctx, cancel := context.WithCancel(s.ctx)
```
binshi-bing, line 1813: From the code and your fix, it seems that #6254 has two issues: one is "context deadline exceeded", another is goroutine leak if TSOClient/tso-stream creation continuously returns error. 

It fixed the first issue by removing the tsoutil.DefaultTSOProxyTimeout, and I suspect this is the right fix. What's the reason to cause TSOClient/tso-stream creation to timeout? Shall we set larger timeout value instead of removing it?

binshi-bing (changes requested):

server/grpc_service.go
binshi-bing: Just realized that checkStream has checked "case <-time.After(3 * time.Second):", that's why you removed tsoutil.DefaultTSOProxyTimeout here. checkStream() hard code the timeout value to 3 seconds, do we need to parameterize the timeout value and pass tsoutil.DefaultTSOProxyTimeout  to checkStream()?

binshi-bing (approved):
LGTM

server/grpc_service.go
lhy1024: yes, you are right, `tsoutil.DefaultTSOProxyTimeout` is 3 seconds, too.

rleungx (approved):

PR #6267, opened 2023-04-04
cgroup: fix the path problem due to special container name

### LINKED ISSUE #6266: cgroup path may be wrong
## Bug Report

See https://github.com/pingcap/tidb/issues/42659 for more details.

rleungx:

Issue Number: Close #6266.

The root cause of this problem is that some container names contain a special symbol. it will lead to getting the wrong cgroup path.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


JmPotato (approved):
LGTM. BTW, is it possible to generate some random container names in the test case to make sure the universality?

rleungx:
> 

Good idea, maybe later.

lhy1024 (approved):

PR #6269, opened 2023-04-04
tso, server: refine the TSO allocator manager parameters

JmPotato:

Issue Number: ref #5895.

```commit-message
- Refine the TSO allocator manager parameters.
- Always run `tsoAllocatorLoop` to advance the Global TSO.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/allocator_manager.go
+++ b/pkg/tso/allocator_manager.go
@@ -231,6 +231,8 @@ func NewAllocatorManager(
 
 	// Set up the Global TSO Allocator here, it will be initialized once the member campaigns leader successfully.
 	am.SetUpGlobalAllocator(am.ctx, am.member.GetLeadership(), startGlobalLeaderLoop)
+	am.svcLoopWG.Add(1)
+	go am.tsoAllocatorLoop()
 
 	return am
 }
@@ -252,9 +254,6 @@ func (am *AllocatorManager) SetUpGlobalAllocator(ctx context.Context, leadership
 		leadership: leadership,
 		allocator:  allocator,
 	}
-
-	am.svcLoopWG.Add(1)
-	go am.tsoAllocatorLoop()
 }
 
 // setUpLocalAllocator is used to set up an allocator, which will initialize the allocator and put it into allocator daemon.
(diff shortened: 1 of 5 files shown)


pkg/tso/allocator_manager.go
```diff
@@ -181,7 +191,7 @@ type AllocatorManager struct {
 	// in etcd, otherwise etcd will expire the leader key and other servers can campaign
 	// the primary/leader again. Etcd only supports seconds TTL, so here is second too.
 	leaderLease    int64
-	maxResetTSGap  func() time.Duration
+	maxResetTSGap  time.Duration
```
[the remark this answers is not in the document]
binshi-bing: If we change it, PD will have regression. In the future, TSO might need to support this hot configuration too.

binshi-bing (changes requested):
Left critical comment -- if my understanding is correct.

pkg/tso/allocator_manager.go
```diff
@@ -143,6 +143,16 @@ type ElectionMember interface {
 	PrecheckLeader() error
 }
 
+// ConfigProvider is used to provide TSO configuration.
+type ConfigProvider interface {
```
binshi-bing, line 147: It seems that there is no need to define a new interface. The newly added interface tso.ServiceConfig in pkg/tso/config.go was designed for this purpose. It's already used by tso.KeyspaceGroupManager and planned to use it for simplifying the interface of NewAllocatorManager too.

pd server config.Config just needs to implement  tso.ServiceConfig, i.e., only needs to add the following methods (the other methods have already been implemented, and GeBackendEndpoints() can be removed).

// GetName returns the Name
func (c *Config) GetName() string {
	return c.Name
}

// GetListenAddr returns the ListenAddr
func (c *Config) GetListenAddr() string {
	return c.ClientUrls
}

// GetAdvertiseListenAddr returns the AdvertiseListenAddr
func (c *Config) GetAdvertiseListenAddr() string {
	return c.AdvertiseClientUrls
}


// GetMaxResetTSGap returns the MaxResetTSGap.
func (c *Config) GetMaxResetTSGap() time.Duration {
	return ...
}

binshi-bing: I just noticed that, for PD/API GetMaxResetTSGap() needs to return s.persistOptions.GetMaxResetTSGap(). Yes, you're right to let Server implement the interface. Shall we combine ConfigProvider and tso.ServiceConfig into one interface? I'm ok to either way.

rleungx (approved):

pkg/tso/allocator_manager.go
JmPotato: Sorry for merging this PR without noticing your comment. I filed a new PR to address this problem. PTAL.

https://github.com/tikv/pd/pull/6272

PR #6274, opened 2023-04-06
mcs: add balancer for keyspace group

### LINKED ISSUE #6233: Develop utility tools for multi-tenant and sharded TSO service
## Development Task
- [ ] Implement the balancer @lhy1024 
- [ ] Support the keyspace group membership management @rleungx @lhy1024

Issues: Ref #5895

lhy1024:

Issue Number: Close #6233

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/balancer/round_robin.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// RoundRobin is a balancer that selects nodes in a round-robin fashion.
+type RoundRobin[T uint32 | string] struct {
+	sync.RWMutex
+	nodes  []T
```
bufferflies, line 25: how about use `LinkedList`

pkg/balancer/round_robin.go
```diff
@@ -0,0 +1,87 @@
...
+	next := atomic.AddUint32(&r.next, 1)
+	node := r.nodes[(int(next)-1)%len(r.nodes)]
+	return node
+}
+
+// Reset updates nodes
+func (r *RoundRobin[T]) Reset(nodes []T) {
+	r.Lock()
+	defer r.Unlock()
+	r.nodes = nodes
```
bufferflies, line 54: the exist needs to update also?

pkg/keyspace/keyspace.go
```diff
@@ -68,14 +68,17 @@ type Manager struct {
...
 type CreateKeyspaceRequest struct {
 	// Name of the keyspace to be created.
 	// Using an existing name will result in error.
-	Name   string
-	Config map[string]string
+	Name     string
+	Config   map[string]string
+	UserKind string
 	// Now is the timestamp used to record creation time.
 	Now int64
```
bufferflies, line 83: how about rename `CreateTime`

pkg/balancer/round_robin.go
lhy1024: good catch

pkg/balancer/round_robin.go
lhy1024: slice is enough, delete op is little.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -106,8 +237,77 @@ func (m *GroupManager) saveKeyspaceGroups(keyspaceGroups []*endpoint.KeyspaceGro
...
+	}
+	return 0, errKeyspaceGroupKindNoExist
+}
+
+// Note: the replica should be less than the number of nodes.
+func (m *GroupManager) allocNodes(replica int) []endpoint.KeyspaceGroupMember {
+	exists := make(map[string]struct{})
+	nodes := make([]endpoint.KeyspaceGroupMember, replica)
+	for i := 0; i < replica; i++ {
+		addr := m.nodesBalancer.Next()
```
rleungx, line 300: Next advances the index, but it returns nil here if the node is not enough.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -50,13 +82,112 @@ func (m *GroupManager) Bootstrap() error {
...
+					m.nodesBalancer.Put(addr)
+				case clientv3.EventTypeDelete:
+					m.nodesBalancer.Delete(addr)
+				}
+			}
+		}
+	}
+}
+
+func (m *GroupManager) initKeyspaceGroupBalancer(groups []*endpoint.KeyspaceGroup) {
```
rleungx, line 134: What if we transfer the leader?

tests/server/keyspace/keyspace_test.go
```diff
@@ -52,16 +58,20 @@ func TestKeyspaceTestSuite(t *testing.T) {
 func (suite *keyspaceTestSuite) SetupTest() {
 	ctx, cancel := context.WithCancel(context.Background())
 	suite.cancel = cancel
-	cluster, err := tests.NewTestCluster(ctx, 3, func(conf *config.Config, serverName string) {
+	cluster, err := tests.NewTestAPICluster(ctx, 3, func(conf *config.Config, serverName string) {
```
rleungx, line 61: Please move the mcs related test under mcs package.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -16,24 +16,56 @@ package keyspace
...
+func NewKeyspaceGroupManager(ctx context.Context, store endpoint.KeyspaceGroupStorage, client *clientv3.Client, clusterID uint64) *GroupManager {
 	return &GroupManager{
-		ctx:   ctx,
-		store: store,
+		ctx:       ctx,
+		store:     store,
+		client:    client,
+		clusterID: clusterID,
+		policy:    defaultBalancerPolicy,
+		groups: struct {
```
rleungx, line 63: The name is confusing since it is not group level but user kind level.

pkg/keyspace/keyspace.go
```diff
@@ -160,16 +165,32 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
...
 			zap.Error(err),
 		)
 		return nil, err
 	}
-	log.Info("[keyspace] keyspace created",
-		zap.Uint32("ID", keyspace.GetId()),
-		zap.String("name", keyspace.GetName()),
-	)
+	enableKeyspaceGroupManager := true
+	failpoint.Inject("disableKeyspaceGroupManager", func() {
```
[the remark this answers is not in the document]
lhy1024: To be compatible with some tests. But the part that allocates keyspace groups to keyspaces has been removed.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -50,13 +82,112 @@ func (m *GroupManager) Bootstrap() error {
 	if err != nil && err != ErrKeyspaceGroupExists {
 		return err
 	}
-
+	if err == nil {
```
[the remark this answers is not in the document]
lhy1024: load keyspace group data. But the part that allocates keyspace groups to keyspaces has been removed.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -106,8 +237,77 @@ func (m *GroupManager) saveKeyspaceGroups(keyspaceGroups []*endpoint.KeyspaceGro
...
+			return err
+		}
+		if oldKG == nil {
+			return errKeyspaceGroupNotFound
+		}
+		newKG := &endpoint.KeyspaceGroup{
+			ID:        oldKG.ID,
+			UserKind:  oldKG.UserKind,
+			Replica:   oldKG.Replica,
+			Keyspaces: append(oldKG.Keyspaces, keyspaceID),
```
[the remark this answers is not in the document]
lhy1024: yes, there is an error. But the part that allocates keyspace groups to keyspaces has been removed.

pkg/keyspace/tso_keyspace_group.go
lhy1024: `Next` will return "" when there is no node. So this function will return nil.

pkg/keyspace/keyspace.go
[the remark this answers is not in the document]
rleungx: It's better not to use the failpoint to pass the original test unless it really needs.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -50,13 +82,112 @@ func (m *GroupManager) Bootstrap() error {
...
+	withRange := clientv3.WithRange(endKey)
+	resp, err := etcdutil.EtcdKVGet(m.client, key, withRange)
+	if err != nil {
+		return err
+	}
+	nodes := make([]string, 0)
+	for _, item := range resp.Kvs {
+		nodes = append(nodes, string(item.Value))
+	}
+	m.nodesBalancer.Reset(nodes)
```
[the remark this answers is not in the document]
lhy1024: yes

pkg/keyspace/tso_keyspace_group.go
lhy1024: maybe. It has been removed.

pkg/keyspace/tso_keyspace_group.go
rleungx: Consider the case that we use 1,3,1,3,1,3... as the request and there are only two nodes, all requests with 1 replica will be allocated in the same node?

lhy1024: the node of the balancer is deduplicated. And it will return "" when the balancer is empty and the internal index will not increase.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -50,13 +82,112 @@ func (m *GroupManager) Bootstrap() error {
...
+	watchChan := m.client.Watch(ctx, key, withRange)
```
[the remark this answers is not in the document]
lhy1024: done

tests/server/keyspace/keyspace_test.go
lhy1024: I think it is more related to keyspace.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -49,12 +84,94 @@ func (m *GroupManager) Bootstrap() error {
...
+				return nil
+			case wresp := <-watchChan:
+				if wresp.CompactRevision != 0 {
+					log.Warn("required revision has been compacted, the watcher will watch again with the compact revision",
+						zap.Int64("required-revision", revision),
+						zap.Int64("compact-revision", wresp.CompactRevision))
+					revision = wresp.CompactRevision
+					break
+				}
+				if wresp.Err() != nil {
```
binshi-bing, line 137: my understanding is that wresp.Err() != nil means two things here:
1. it isn't equal to ctx being cancelled. So if we return here and exit watchLoop but the API service is still serving and try to recover from some bad thing happened, once it's recovered we have no loop to watch the primary address change.
2. It doesn't help to call m.client.Watch to retry because watch.go will retry Watch() as much as it can unless it thinks it doesn't help.

It seems that we can't exit the watch loop but wait a while then retry "watcher := clientv3.NewWatcher(m.client) ; then watch" endlessly until ctx is cancelled. + @rleungx to  provide possible comment on this.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -16,24 +16,59 @@ package keyspace
...
+	client *clientv3.Client
+
+	// tsoServiceKey is the path of TSO service in etcd.
+	tsoServiceKey string
+	// tsoServiceEndKey is the end key of TSO service in etcd.
+	tsoServiceEndKey string
+
+	policy balancer.Policy
+
+	nodesBalancer balancer.Balancer[string]
```
binshi-bing, line 57: It seems that we just select nodes in the round-robin. We need two-levels scheduling here. The first level is to use the following policies to place keyspace groups according to their UserKind, within each UserKind then user round-robin. We probably can't skip these two-levels scheduling for now. + @rleungx, thoughts? 

1. The TSO system tries to evenly distribute the keyspace with the User Kind "business" or "enterprise", across different keyspace groups as much as possible.
2. The TSO system tries to evenly distribute the primary/secondaries of the keyspace groups with the User Kind "business" or "enterprise", across the TSO nodes as much as possible.
3. The TSO system tries to assign the 'basic' keyspace groups' primaries to as few TSO nodes as possible.

binshi-bing (commented):
Left few comments. @rleungx for more thoughts about my comment

pkg/keyspace/tso_keyspace_group.go
lhy1024: So where will the correspondence between tso node and user kind be found?

pkg/balancer/balancer.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Policy is the policy of balancer.
+type Policy int
+
+const (
+	// PolicyRoundRobin is the round robin policy.
+	PolicyRoundRobin Policy = iota
+	// PolicyLeast is the policy to return the least used node.
+	// TODO: move indexed heap to pkg and use it.
+	PolicyLeast
```
bufferflies, line 25: where is it used?

pkg/balancer/balancer.go
```diff
@@ -0,0 +1,59 @@
...
+	// GetAll returns all nodes.
+	GetAll() []T
+}
+
+// GenByPolicy generates a balancer by policy.
+func GenByPolicy[T uint32 | string](policy Policy) Balancer[T] {
+	switch policy {
+	case PolicyRoundRobin:
+		return NewRoundRobin[T]()
+	default: // only round-robin is supported now.
```
bufferflies, line 56: why not return round robin directly?

tests/server/keyspace/keyspace_test.go
```diff
@@ -81,8 +81,8 @@ func (suite *keyspaceTestSuite) TestRegionLabeler() {
 	var err error
 	for i := 0; i < count; i++ {
 		keyspaces[i], err = manager.CreateKeyspace(&keyspace.CreateKeyspaceRequest{
-			Name: fmt.Sprintf("test_keyspace%d", i),
-			Now:  now,
+			Name:       fmt.Sprintf("test_keyspace%d", i),
```
bufferflies, line 84:
```suggestion
			Name:       fmt.Sprintf("test_keyspace_%d", i),
```

pkg/mcs/discovery/key_path.go
```diff
@@ -28,3 +31,8 @@ func registryPath(clusterID, serviceName, serviceAddr string) string {
 func discoveryPath(clusterID, serviceName string) string {
 	return strings.Join([]string{registryPrefix, clusterID, serviceName, registryKey}, "/")
 }
+
+// TSOAddrsPath returns the path to store TSO addresses.
+func TSOAddrsPath(clusterID uint64) string {
```
bufferflies, line 36:
```suggestion
func TSOPath(clusterID uint64) string {
```

pkg/balancer/balancer.go
lhy1024: it will do at anotherr pr

bufferflies: why not add it in another pr?

pkg/keyspace/tso_keyspace_group.go
binshi-bing: The balancer assigns and specifies.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -49,12 +84,94 @@ func (m *GroupManager) Bootstrap() error {
...
 	return nil
 }
 
+// Close closes the manager.
+func (m *GroupManager) Close() {
+	m.cancel()
+	m.wg.Wait()
+}
+
+func (m *GroupManager) startWatchLoop(revision int64) {
```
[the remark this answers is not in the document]
lhy1024: just so

pkg/keyspace/tso_keyspace_group.go
lhy1024: I will do a fix similiar with #6279

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -116,3 +233,36 @@ func (m *GroupManager) GetAvailableKeyspaceGroupIDByKind(userKind endpoint.UserK
...
+	exists := make(map[string]struct{})
+	nodes := make([]endpoint.KeyspaceGroupMember, 0, replica)
+	for len(exists) < replica {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+		}
+		addr := m.nodesBalancer.Next()
+		if addr == "" { // no node
```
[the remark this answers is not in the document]
lhy1024: They is the same

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -44,6 +45,10 @@ type CreateKeyspaceGroupParams struct {
 
 // CreateKeyspaceGroups creates keyspace groups.
 func CreateKeyspaceGroups(c *gin.Context) {
```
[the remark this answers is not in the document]
lhy1024: ok, I will split it.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -430,3 +571,58 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitToID uint32) error {
...
+				continue
+			}
+			exists[addr] = struct{}{}
+			nodes = append(nodes, endpoint.KeyspaceGroupMember{Address: addr})
+		}
+		kg.Members = nodes
+		m.store.SaveKeyspaceGroup(txn, kg)
+		return nil
+	})
+	if err != nil {
```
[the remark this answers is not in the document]
lhy1024: if meet error, we will not alloc node.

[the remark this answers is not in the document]
rleungx:
```suggestion
	return nodes, err
```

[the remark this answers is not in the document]
lhy1024: If there is an error during the loop, we have not actually alloc any nodes to the keyspace group, but the `node` slice is already non-empty, so I don't think we need to return a non-empty slice in this case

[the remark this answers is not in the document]
rleungx: ok

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -86,9 +127,111 @@ func (m *GroupManager) Bootstrap() error {
...
 	return nil
 }
 
+// Close closes the manager.
+func (m *GroupManager) Close() {
+	m.cancel()
+	m.wg.Wait()
+}
+
+func (m *GroupManager) startWatchLoop() {
```
rleungx, line 145: log panic?

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -183,6 +190,48 @@ func FinishSplitKeyspaceByID(c *gin.Context) {
...
+	}
+	svr := c.MustGet(middlewares.ServerContextKey).(*server.Server)
+	manager := svr.GetKeyspaceGroupManager()
+	allocParams := &AllocNodeForKeyspaceGroupParams{}
+	err = c.BindJSON(allocParams)
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, errs.ErrBindJSON.Wrap(err).GenWithStackByCause())
+		return
+	}
+	if manager.GetNodesNum() < allocParams.Replica || allocParams.Replica < 1 {
```
rleungx, line 213: Assume that the group has a specific user kind, here we just GetNodesNum to do the check. Does it make sense?

lhy1024: https://github.com/tikv/pd/pull/6274#discussion_r1160380892

rleungx (approved):
Mostly, LGTM

PR #6279, opened 2023-04-06
mcs: fix watch primary address revision and update cache when meets not leader 

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/5895.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/server.go
```diff
@@ -1737,15 +1737,26 @@ func (s *Server) watchServicePrimaryAddrLoop(serviceName string) {
...
+		watchChan := watcher.Watch(ctx, serviceKey, clientv3.WithPrefix(), clientv3.WithRev(revision))
+		for wresp := range watchChan {
+			if wresp.CompactRevision != 0 {
+				log.Warn("required revision has been compacted, use the compact revision",
+					zap.Int64("required-revision", revision),
+					zap.Int64("compact-revision", wresp.CompactRevision))
+				revision = wresp.CompactRevision
+				break
+			}
+			if wresp.Canceled {
```
binshi-bing, line 1753: It seems that, you haven't forgotten your comment in my pr and didn't check my reply :)

At the same place, you asked "Do we need to solve other errors?", I replied:

I checked the code in etcd watch.go. There is no other errors to solve, but I do think here using wrep.Err() != nil is better than using wresp.Canceled, because the former can accommodate future etcd changes.

Etcd Watch() returns three types of errors -- closed error, compacted error and cancelled error. closed error is captured by range watchChan, compacted error is captured by wresp.CompactRevision != 0, and cancelled error is captured here. Etcd Watch() will retry Watch() for all other errors. Using wrep.Err() != nil is better than using wresp.Canceled.

Changed to using wrep.Err() != nil here.

Member
Author
@[binshi-bing](https://github.com/binshi-bing) binshi-bing Pending
I checked the code in etcd watch.go. There is no other errors to solve, but I do think here using wrep.Err() != nil is better than using wresp.Canceled, because the former can accommodate future etcd changes.

Etcd Watch() returns three types of errors -- closed error, compacted error and cancelled error. closed error is captured by range watchChan, compacted error is captured by wresp.CompactRevision != 0, and cancelled error is captured here. Etcd Watch() will retry Watch() for all other errors. Using wrep.Err() != nil is better than using wresp.Canceled.

Changed to using wrep.Err() != nil here.

binshi-bing, line 1753: This change made some improvement, but mightn't be enough. When wresp.Canceled(), it doesn't mean context has been cancelled, so if the primary address loop returns here but the API service is still serving, we'll have improper function. Do we need to try to recreate Watcher and re-watch in a endless loop until context is cancelled, as what I do in KeyspaceGroupManager? 

```
	select {
	case reqc <- wr:
		ok = true
	case <-wr.ctx.Done():
	case <-donec:
		if wgs.closeErr != nil {
			closeCh <- WatchResponse{Canceled: true, closeErr: wgs.closeErr}
			break
		}
		// retry; may have dropped stream from no ctxs
		return w.Watch(ctx, key, opts...)
	}
```

server/server.go
```diff
@@ -1726,7 +1726,7 @@ func (s *Server) watchServicePrimaryAddrLoop(serviceName string) {
 	log.Info("start to watch", zap.String("service-key", serviceKey))
 
 	primary := &tsopb.Participant{}
-	ok, rev, err := etcdutil.GetProtoMsgWithModRev(s.client, serviceKey, primary)
+	ok, revision, err := etcdutil.GetProtoMsgWithModRev(s.client, serviceKey, primary)
```
binshi-bing, line 1729: do we only need to watch keyspace group 0?

binshi-bing (approved):
LGTM for the change having improved things, though we might still need more improvements.

server/server.go
lhy1024: yes, `serviceKey := fmt.Sprintf("/ms/%d/%s/%s/%s", s.clusterID, serviceName, fmt.Sprintf("%05d", 0), "primary")`

server/server.go
lhy1024: sorry, I forgot about it when the copilot finished

lhy1024: ok

server/server.go
```diff
@@ -1737,15 +1737,26 @@ func (s *Server) watchServicePrimaryAddrLoop(serviceName string) {
 	} else {
 		log.Warn("service primary addr doesn't exist", zap.String("service-key", serviceKey))
 	}
+	watcher := clientv3.NewWatcher(s.client)
```
rleungx, line 1740: There are many codes with the same logic, the only difference is the key. How about abstracting a function for them?

lhy1024: ok, I will try it later.

server/server.go
```diff
@@ -1716,43 +1720,110 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+	defer watcher.Close()
+
+	for {
+	WatchChan:
+		watchChan := watcher.Watch(s.serverLoopCtx, serviceKey, clientv3.WithPrefix(), clientv3.WithRev(revision))
+		select {
+		case <-ctx.Done():
+			return revision, nil
+		case <-s.updateServicePrimaryAddrCh:
+			revision, err = s.updateServicePrimaryAddr(serviceName)
```
binshi-bing, line 1803: Most likely even after we update the service primary address, we still have one more problem -- s.clientConns stores the forwarded hosts' grpc.ClientConn. We never update the broken connections. if a forwarded host's connection broke，e.g., the forwarded host restarted and broken the existing connection, we'll retrieve the broken connection for this forwarded host continuously.

binshi-bing (changes requested):
there is one more comment which might need you to check and confirm.

binshi-bing (approved):
LGTM

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -57,20 +58,20 @@ func NewTSODispatcher(tsoProxyHandleDuration, tsoProxyBatchSize prometheus.Histo
 
 // DispatchRequest is the entry point for dispatching/forwarding a tso request to the detination host
 func (s *TSODispatcher) DispatchRequest(
-	ctx context.Context, req Request, tsoProtoFactory ProtoFactory, doneCh <-chan struct{}, errCh chan<- error) {
+	ctx context.Context, req Request, tsoProtoFactory ProtoFactory, doneCh <-chan struct{}, errCh chan<- error, updateServicePrimaryAddrChs ...chan<- struct{}) {
```
JmPotato, line 61: Could you help divide these parameters into multiple lines? It looks a little bit long now.

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -57,20 +58,20 @@ func NewTSODispatcher(tsoProxyHandleDuration, tsoProxyBatchSize prometheus.Histo
...
+		go s.dispatch(ctx, tsoProtoFactory, req.getForwardedHost(), req.getClientConn(), reqCh, tsDeadlineCh, doneCh, errCh, updateServicePrimaryAddrChs...)
 		go watchTSDeadline(ctx, tsDeadlineCh)
 	}
 	reqCh <- req
 }
 
 func (s *TSODispatcher) dispatch(
 	ctx context.Context, tsoProtoFactory ProtoFactory, forwardedHost string, clientConn *grpc.ClientConn,
-	tsoRequestCh <-chan Request, tsDeadlineCh chan<- deadline, doneCh <-chan struct{}, errCh chan<- error) {
+	tsoRequestCh <-chan Request, tsDeadlineCh chan<- deadline, doneCh <-chan struct{}, errCh chan<- error, updateServicePrimaryAddrChs ...chan<- struct{}) {
```
JmPotato, line 74: Ditto.

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -121,6 +122,11 @@ func (s *TSODispatcher) dispatch(
 				log.Error("proxy forward tso error",
 					zap.String("forwarded-host", forwardedHost),
 					errs.ZapError(errs.ErrGRPCSend, err))
+				if len(updateServicePrimaryAddrChs) > 0 {
+					if strings.Contains(err.Error(), errs.NotLeaderErr) || strings.Contains(err.Error(), errs.MismatchLeaderErr) {
+						updateServicePrimaryAddrChs[0] <- struct{}{}
+					}
+				}
```
JmPotato, line 129: I suggest putting this part into a function outside the loop, which may help the compiler to do some inline optimization.

server/grpc_service.go
```diff
@@ -224,7 +224,7 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 			}
 
 			tsoRequest := tsoutil.NewPDProtoRequest(forwardedHost, clientConn, request, stream)
-			s.tsoDispatcher.DispatchRequest(ctx, tsoRequest, tsoProtoFactory, doneCh, errCh)
+			s.tsoDispatcher.DispatchRequest(ctx, tsoRequest, tsoProtoFactory, doneCh, errCh, s.updateServicePrimaryAddrCh)
```
JmPotato, line 227: It seems that the channel is always passed to the function, then why use an optional parameter?

server/server.go
```diff
@@ -1716,43 +1720,110 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+		default:
+		}
+		nextRevision, err := s.watchServicePrimaryAddr(ctx, serviceName, revision)
+		if err != nil {
+			log.Error("watcher canceled unexpectedly and a new watcher will start after a while",
+				zap.Int64("next-revision", nextRevision),
+				zap.Time("retry-at", time.Now().Add(watchKEtcdChangeRetryInterval)),
+				zap.Error(err))
+			revision = nextRevision
+			time.Sleep(watchKEtcdChangeRetryInterval)
```
JmPotato, line 1764: Suggest using a ticker to select rather than sleeping.

server/server.go
```diff
@@ -104,6 +104,8 @@ const (
 	maxRetryTimesGetServicePrimary = 25
 	// retryIntervalGetServicePrimary is the retry interval for getting primary addr.
 	retryIntervalGetServicePrimary = 100 * time.Millisecond
+	// TODO: move it to etcdutil
+	watchKEtcdChangeRetryInterval = 1 * time.Second
```
rleungx, line 108:
```suggestion
	watchEtcdChangeRetryInterval = 1 * time.Second
```

server/server.go
```diff
@@ -1716,43 +1720,110 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+	s.servicePrimaryMap.Store(serviceName, addr)
+}
+
+// updateServicePrimaryAddr updates the primary address from etcd with get operation.
+func (s *Server) updateServicePrimaryAddr(serviceName string) (nextRevision int64, err error) {
+	serviceKey := s.servicePrimaryKey(serviceName)
+	primary := &tsopb.Participant{}
+	ok, revision, err := etcdutil.GetProtoMsgWithModRev(s.client, serviceKey, primary)
+	listenUrls := primary.GetListenUrls()
+	if !ok || err != nil || len(listenUrls) == 0 {
```
rleungx, line 1781: There are two cases we may return `0, nil` and it breaks the retry loop, is it expected?

server/server.go
```diff
@@ -1716,43 +1720,110 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+
+// watchServicePrimaryAddr watches the primary address on etcd.
+func (s *Server) watchServicePrimaryAddr(ctx context.Context, serviceName string, revision int64) (nextRevision int64, err error) {
+	serviceKey := s.servicePrimaryKey(serviceName)
+	watcher := clientv3.NewWatcher(s.client)
+	defer watcher.Close()
+
+	for {
+	WatchChan:
+		watchChan := watcher.Watch(s.serverLoopCtx, serviceKey, clientv3.WithPrefix(), clientv3.WithRev(revision))
```
rleungx, line 1798: Do we still need the prefix?

server/server.go
```diff
@@ -1716,43 +1720,110 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+		select {
+		case <-ctx.Done():
+			return revision, nil
+		case <-s.updateServicePrimaryAddrCh:
+			revision, err = s.updateServicePrimaryAddr(serviceName)
+			goto WatchChan
```
rleungx, line 1804: Is it possible that we call watch many times?

server/server.go
lhy1024: it will try to create a new connection automatically when the existed connection is closed.

server/grpc_service.go
lhy1024: `DispatchRequest` is also used by tso server, tso server is no needed to watch api key

server/server.go
lhy1024: In fact, we need to wait for a retry here, instead of periodically going to use tick

server/server.go
lhy1024: I will fix

server/server.go
lhy1024: the size of the channel only is one

rleungx: So there might be two watch at the same time?

lhy1024: we only watch in the one goroutine.

server/server.go
```diff
@@ -1716,43 +1721,98 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+	var (
+		revision int64
+		err      error
+	)
+	for i := 0; i < maxRetryTimesGetServicePrimary; i++ {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(retryIntervalGetServicePrimary):
+		}
```
JmPotato, line 1741: Should we call `s.updateServicePrimaryAddr(serviceName)` at the beginning of the loop? Otherwise, we have to wait for a `retryIntervalGetServicePrimary` before the first time update.

server/server.go
```diff
@@ -1716,43 +1721,98 @@ func (s *Server) GetServicePrimaryAddr(ctx context.Context, serviceName string)
...
+				zap.Error(err))
+			revision = nextRevision
+			time.Sleep(watchEtcdChangeRetryInterval)
+		}
+	}
+}
+
+// SetServicePrimaryAddr sets the primary address directly.
+// Note: This function is only used for test.
+func (s *Server) SetServicePrimaryAddr(serviceName, addr string) {
```
rleungx, line 1772: Where do we use it?

rleungx (commented):
Here is a comment still left: https://github.com/tikv/pd/pull/6279/files#r1161400175

rleungx (approved):

Issue #6232, opened 2023-03-27
Support multi-keyspace-group and sharding in tso grpc service


binshi-bing:
## Development Task
- [x] Overall refactor
    - [x] #6175 @binshi-bing
- [x] #6214 @binshi-bing
- [x] #6247  @binshi-bing
- [x] #6282 @binshi-bing 
    - [x] follow-up pr #6305 @binshi-bing
    - [x] follow-up pr #6306 @binshi-bing  
- [x] Support keyspace group 0 split into multiple keyspace groups @JmPotato
    - [x] #6293  @JmPotato
        - [x] (follow-up pr) #6309 
    - [x] #6313 @JmPotato
- [x] #6321  @binshi-bing 
    - [x] https://github.com/pingcap/kvproto/pull/1101

Issues: Ref #5895

binshi-bing:
done

PR #6282, opened 2023-04-07
Apply multi-keyspace-group membership to tso service and handle inconsistency issue

binshi-bing:

Issue Number: Ref #6232 

```commit-message
Apply multi-keyspace-group membership to tso service and handle inconsistency issue
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/tso/keyspace_group_manager.go
```diff
@@ -446,6 +447,10 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
 		if kgm.ams[group.ID].Load() != nil {
 			log.Info("keyspace group already initialized, so update meta only",
 				zap.Uint32("keyspace-group-id", group.ID))
+			group.KeyspaceLookupTable = make(map[uint32]struct{})
+			for _, kid := range group.Keyspaces {
+				group.KeyspaceLookupTable[kid] = struct{}{}
```
JmPotato, line 452: Do we need to clean up the keyspaces that may not be in this group anymore in this situation?

pkg/tso/keyspace_group_manager.go
```diff
@@ -495,39 +505,86 @@ func (kgm *KeyspaceGroupManager) deleteKeyspaceGroup(id uint32) {
...
+	return ts, keyspaceGroupID, err
 }
 
 func (kgm *KeyspaceGroupManager) checkKeySpaceGroupID(id uint32) error {
 	if id < mcsutils.MaxKeyspaceGroupCountInUse {
 		return nil
 	}
 	return errs.ErrKeyspaceGroupIDInvalid.FastGenByArgs(
-		fmt.Sprintf("invalid keyspace group id %d which shouldn't >= %d", id, mcsutils.MaxKeyspaceGroupCountInUse))
+		fmt.Sprintf("invalid keyspace group id %d which shouldn't >= %d",
```
rleungx, line 581: I think the error message "invalid keyspace group id" has already been included in `ErrKeyspaceGroupIDInvalid`.

pkg/tso/keyspace_group_manager.go
```diff
@@ -495,39 +505,86 @@ func (kgm *KeyspaceGroupManager) deleteKeyspaceGroup(id uint32) {
...
+	if err := kgm.checkKeySpaceGroupID(keyspaceGroupID); err != nil {
+		return pdpb.Timestamp{}, keyspaceGroupID, err
+	}
+	am, err := kgm.GetAMWithMembershipCheck(keyspaceID, keyspaceGroupID)
 	if err != nil {
-		return pdpb.Timestamp{}, err
+		// The keyspace doesn't belong to this keyspace group, we should check if it belongs to any other
+		// keyspace groups, and return the correct keyspace group ID to the client.
+		for i := 0; i < int(mcsutils.MaxKeyspaceGroupCountInUse); i++ {
+			if ksg := kgm.ksgs[i].Load(); ksg == nil {
```
rleungx, line 564: Will it cost too much?

pkg/errs/errno.go
```diff
@@ -52,6 +52,7 @@ var (
 	ErrLoadKeyspaceGroupsTimeout       = errors.Normalize("load keyspace groups timeout", errors.RFCCodeText("ErrLoadKeyspaceGroupsTimeout"))
 	ErrLoadKeyspaceGroupsTerminated    = errors.Normalize("load keyspace groups terminated", errors.RFCCodeText("ErrLoadKeyspaceGroupsTerminated"))
 	ErrLoadKeyspaceGroupsRetryExhaustd = errors.Normalize("load keyspace groups retry exhausted, %s", errors.RFCCodeText("ErrLoadKeyspaceGroupsRetryExhaustd"))
+	ErrKeyspaceNotAssigned             = errors.Normalize("the keyspace isn't assigned to any keyspace group, %s", errors.RFCCodeText("PD:tso:ErrKeyspaceNotAssigned"))
```
rleungx, line 55: Where do we use it?

pkg/tso/keyspace_group_manager.go
binshi-bing: Indeed. Changed to "fmt.Sprintf("%d which shouldn't >= %d", id, mcsutils.MaxKeyspaceGroupCountInUse))"

pkg/errs/errno.go
binshi-bing: Changed to

binshi-bing: Made change (check **return pdpb.Timestamp{}, keyspaceGroupID, errs.ErrKeyspaceNotAssigned.FastGenByArgs(keyspaceID)**):
```
	am, err := kgm.GetAMWithMembershipCheck(keyspaceID, keyspaceGroupID)
	if err != nil {
		// The keyspace doesn't belong to this keyspace group, we should check if it belongs to any other
		// keyspace groups, and return the correct keyspace group ID to the client.
		for i := 0; i < int(mcsutils.MaxKeyspaceGroupCountInUse); i++ {
                         // return if the keyspace is found in the keyspace lookup tables.
		}
		return pdpb.Timestamp{}, keyspaceGroupID, errs.ErrKeyspaceNotAssigned.FastGenByArgs(keyspaceID)
	}
```

pkg/tso/keyspace_group_manager.go
binshi-bing: This is the rare case, but I reimplement with new algorithm to reduce time complexity from O(mlogn) to O(logn), where m is the number of keyspace groups and n the number of keyspaces.

pkg/tso/keyspace_group_manager.go
```diff
@@ -475,59 +484,169 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
...
+		for _, kid := range kg.Keyspaces {
+			kgm.keyspaceLookupTable.CompareAndDelete(kid, kg.ID)
+		}
 	}
-	am.close()
-	log.Info("deleted keyspace group", zap.Uint32("keyspace-group-id", id))
+	am := kgm.ams[groupID].Swap(nil)
+	if am != nil {
+		am.close()
+	}
```
rleungx, line 569: Should we make it atomic?

binshi-bing: discussed offline. leave what it is for now, and we may improve it in the future.

rleungx: And maybe we can improve it later.

rleungx (approved):

pkg/tso/keyspace_group_manager.go
binshi-bing: Very tiny unreliable/inconsistent window and we can just let client retry to get over it. After all, there are other unreliable/inconsistency windows during kg-split/keyspace-movement/kg-movement which can't be avoided. It won't cause incorrect result for tso requests, because the timeline split problem will be specifically handled by kg-split/keyspace-movement and there is no time split problem because of primary election among replicas within a kg.

PR #6295, opened 2023-04-10
dr-autosync: add recover timeout

disksing:
Issue Number: Ref #4399

add a configuration item to delay dr state switch from async to sync-recover

```commit-message
```

### Check List

Tests
- Unit test

### Release note

```release-note
None.
```


pkg/replication/replication_mode.go
```diff
@@ -477,7 +488,7 @@ func (m *ModeManager) tickUpdateState() {
 			m.drSwitchToAsync(storeIDs[primaryUp])
 		}
 	case drStateAsync:
-		if canSync {
+		if canSync && m.drDurationSinceAsyncStart() > m.config.DRAutoSync.WaitRecoverTimeout.Duration {
```
lhy1024, line 491: Is this config only used to switch from `async` to `syncRecover`?

pkg/replication/replication_mode.go
```diff
@@ -262,7 +263,8 @@ func (m *ModeManager) drSwitchToAsyncWithLock(availableStores []uint64) error {
 		log.Warn("failed to switch to async state", zap.String("replicate-mode", modeDRAutoSync), errs.ZapError(err))
 		return err
 	}
-	dr := drAutoSyncStatus{State: drStateAsync, StateID: id, AvailableStores: availableStores}
+	now := time.Now()
```
lhy1024, line 266: Do we need to pick it to 6.5 and other branches?

lhy1024 (approved):

pkg/replication/replication_mode.go
disksing: yes!

pkg/replication/replication_mode.go
disksing: need cherry-pick 6.5. i have added the label

nolouch (approved):

PR #6306, opened 2023-04-12
bootstrap default keyspace group in the tso service

binshi-bing:

Issue Number: Ref #6232

```commit-message
Changes:

1. Introduce the initialization logic of the default keyspace group.
- If the default keyspace group isn't configured in the etcd, every tso node/pod should initialize it and join the election for the primary of this group.
- If the default keyspace group is configured in the etcd, the tso nodes/pods which are assigned with this group will initialize it and join the election for the primary of this group.

2. Introduce the keyspace group membership restriction -- default keyspace always belongs to default keyspace group.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -58,6 +58,10 @@ func NewKeyspaceGroupManager(ctx context.Context, store endpoint.KeyspaceGroupSt
 
 // Bootstrap saves default keyspace group info and init group mapping in the memory.
 func (m *GroupManager) Bootstrap() error {
+	// Force the membership restriction that the default keyspace must belong to default keyspace group.
+	// Have no information to specify the distribution of the default keyspace group replicas, so just
+	// leave the replica/member list empty. The TSO service will assign the default keyspace group replica
+	// to every tso node/pod by default.
 	defaultKeyspaceGroup := &endpoint.KeyspaceGroup{
 		ID:        utils.DefaultKeyspaceGroupID,
 		UserKind:  endpoint.Basic.String(),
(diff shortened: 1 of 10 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -71,15 +76,8 @@ func (m *GroupManager) Bootstrap() error {
 		return err
 	}
 
-	userKind := endpoint.StringUserKind(defaultKeyspaceGroup.UserKind)
```
binshi-bing, line 74: delete line 74 ~ 79 for two reasons:
1. This is covered by the for loop below in which it loads all keyspace groups, including the default keyspace group, and initialize m.groups.
2. Theoretically,  at line 79, we put defaultKeyspaceGroup into m.groups, but defaultKeyspaceGroup might be different from the latest version persisted in the store, though later on the current logic of indexedHeap will overwrite the existing one if put a new one.

binshi-bing: @rleungx ^

binshi-bing: Let me know if I missed anything.

rleungx: You're right.

pkg/tso/keyspace_group_manager.go
```diff
@@ -313,10 +320,15 @@ func (kgm *KeyspaceGroupManager) initAssignment(ctx context.Context) (int64, err
 		for _, group := range groups {
 			select {
 			case <-ctx.Done():
-				return watchStartRevision, errs.ErrLoadKeyspaceGroupsTerminated
+				err = errs.ErrLoadKeyspaceGroupsTerminated
```
rleungx, line 323: I think we can return nil here instead of an error.

JmPotato (approved):

rleungx (approved):

PR #6309, opened 2023-04-12
keyspace: add split-from field for endpoint.KeyspaceGroup

JmPotato:

Issue Number: ref #6232. This field is necessary since we have to know which keyspace group this new one comes from, so we can get the corresponding member info to implement the election constraint during the TSO split.

```commit-message
Add `split-from` field for `endpoint.KeyspaceGroup`.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/storage/endpoint/tso_keyspace_group.go
+++ b/pkg/storage/endpoint/tso_keyspace_group.go
@@ -71,6 +71,8 @@ type KeyspaceGroup struct {
 	UserKind string `json:"user-kind"`
 	// InSplit indicates whether the keyspace group is in split.
 	InSplit bool `json:"in-split"`
+	// SplitFrom is the keyspace group ID which the keyspace group is split from.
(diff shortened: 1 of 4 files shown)


rleungx (approved):

pkg/storage/endpoint/tso_keyspace_group.go
```diff
@@ -71,6 +71,8 @@ type KeyspaceGroup struct {
 	UserKind string `json:"user-kind"`
 	// InSplit indicates whether the keyspace group is in split.
 	InSplit bool `json:"in-split"`
+	// SplitFrom is the keyspace group ID which the keyspace group is split from.
```
binshi-bing, line 74:
```suggestion
	// SplitFrom is the keyspace group ID from which the keyspace group is split.
```

binshi-bing (approved):
LGTM. left one trivial comment.

PR #6313, opened 2023-04-12
tso, member: support TSO split based on keyspace group split

JmPotato:

Issue Number: ref #6232.

```commit-message
Support TSO split based on keyspace group split.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -147,7 +147,7 @@ func (m *GroupManager) DeleteKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGro
 		if kg == nil {
 			return nil
 		}
-		if kg.InSplit {
+		if kg.InSplit() {
```
rleungx, line 150: How about `IsSplitting`?

pkg/storage/endpoint/tso_keyspace_group.go
```diff
@@ -84,6 +88,30 @@ type KeyspaceGroup struct {
 	KeyspaceLookupTable map[uint32]struct{} `json:"-"`
 }
 
+// InSplit checks if the keyspace group is in split state.
+func (kg *KeyspaceGroup) InSplit() bool {
+	return kg != nil && kg.SplitState != nil
+}
+
+// IsSplitTo checks if the keyspace group is in split state and is the split-to keyspace group.
+func (kg *KeyspaceGroup) IsSplitTo() bool {
```
rleungx, line 97: How about `IsSplitTarget`?

pkg/tso/keyspace_group_manager.go
```diff
@@ -521,61 +528,75 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
...
+	log.Info("joining primary election",
+		zap.Uint32("keyspace-group-id", group.ID),
+		zap.String("participant-name", uniqueName),
+		zap.Uint64("participant-id", uniqueID))
+	// Initialize the participant info to join the primary election.
+	participant := member.NewParticipant(kgm.etcdClient)
+	participant.InitInfo(
+		uniqueName, uniqueID, path.Join(kgm.tsoSvcRootPath, fmt.Sprintf("%05d", group.ID)),
+		primaryElectionSuffix, "keyspace group primary election", kgm.cfg.GetAdvertiseListenAddr())
+	// If the keyspace group is in split, we should ensure that the primary elected by the new keyspace group
```
binshi-bing, line 559: How do we force the constraint that the primaries of the split-from kg and split-to kg are on the same tso node? Do we reuse the logic of "nextLeader" in local tso allocators? Will this election algorithm change be in this pr or in separate pr?

pkg/tso/keyspace_group_manager.go
```diff
@@ -731,3 +760,69 @@ func genNotServedErr(perr *perrors.Error, keyspaceGroupID uint32) error {
...
 			keyspaceGroupID, errs.NotServedErr))
 }
+
+// checkTSOSplit checks if the given keyspace group is in split state, and if so, it will make sure the
+// newly split TSO keep consistent with the original one.
+func (kgm *KeyspaceGroupManager) checkTSOSplit(
+	keyspaceGroupID uint32,
+	dcLocation string,
+) error {
+	splitAM, splitGroup := kgm.getKeyspaceGroupMeta(keyspaceGroupID)
```
binshi-bing, line 770: What if there are multiple TSO requests coming in and they all pass line 804 and cause splitAllocator.SetTSO() and kgm.finishSplitKeyspaceGroup to be invoked multiple times? Do we need to put the function into a critical section?

pkg/tso/keyspace_group_manager.go
```diff
@@ -731,3 +760,69 @@ func genNotServedErr(perr *perrors.Error, keyspaceGroupID uint32) error {
...
+	return kgm.finishSplitKeyspaceGroup(keyspaceGroupID)
+}
+
+const keyspaceGroupsAPIPrefix = "/pd/api/v2/tso/keyspace-groups"
+
+func (kgm *KeyspaceGroupManager) finishSplitKeyspaceGroup(id uint32) error {
+	if kgm.httpClient == nil {
+		return nil
+	}
+	statusCode, err := apiutil.DoDelete(kgm.httpClient, kgm.cfg.GeBackendEndpoints()+keyspaceGroupsAPIPrefix+fmt.Sprintf("/%d/split", id))
```
binshi-bing, line 819: This line is too long

pkg/tso/keyspace_group_manager.go
```diff
@@ -731,3 +760,69 @@ func genNotServedErr(perr *perrors.Error, keyspaceGroupID uint32) error {
...
+	if err != nil {
+		return err
+	}
+	if statusCode != http.StatusOK {
+		log.Warn("failed to finish split keyspace group", zap.Uint32("keyspace-group-id", id), zap.Int("status-code", statusCode))
```
binshi-bing, line 824: multiple lines?

binshi-bing (commented):
left some comments

pkg/storage/endpoint/tso_keyspace_group.go
```diff
@@ -84,6 +88,30 @@ type KeyspaceGroup struct {
...
+	return kg.InSplit() && kg.SplitState.SplitFrom == kg.ID
+}
+
+// SplitFrom returns the split-from keyspace group ID. When the keyspace group is the split-from group
+// itself, it will return its own ID.
+func (kg *KeyspaceGroup) SplitFrom() uint32 {
+	if kg.InSplit() {
+		return kg.SplitState.SplitFrom
+	}
+	return 0
```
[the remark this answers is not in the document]
JmPotato: Currently, we check the split status before every `SplitSource()` call in all logic, so there won't be any problems. However, you are right that to avoid potential bugs in the future, I should modify this part.

pkg/tso/keyspace_group_manager.go
JmPotato: This is guaranteed by two constraints:

- Once a keyspace group is in the split state, its meta info could not be modified after #6316.
- `participant.SetPreCampaignChecker` below makes sure the primary could only be elected by the participant which owns the split source keyspace group primary also.

pkg/tso/keyspace_group_manager.go
JmPotato: It's not a problem since `SetTSO` can only set the TSO to a bigger value.

rleungx (approved):

pkg/tso/keyspace_group_manager.go
binshi-bing: That part is fine. My real worry is that if tso request concurrency is high, it will cause finishSplitKeyspaceGroup to be invoked too many times, which isn't cheap and result in transaction conflicts and high load on etcd, though one transaction will succeed eventually.

binshi-bing: I'll sign off for now. Suggest to make some protection here.

binshi-bing (approved):
LGTM

pkg/tso/keyspace_group_manager.go
JmPotato: OK, I will make some refinements here today.

PR #6314, opened 2023-04-12
scheduler: cache history loads in hot region scheduler

### LINKED ISSUE #6297: `Hot-region-scheduler` should be more stable if the loads are not stable
## Enhancement Task
In past, `Balance-hot-region-scheduler` depences on the current lodas to move hot region. But if the loads is not stable, there are many useless operators .
5481fa3a-96a8-4b77-9a54-76fa3bf75092

bufferflies:

Issue Number: Close #6297, Ref #6328, close #https://github.com/tikv/tikv/issues/14458

In past, the store pick strategy only consider the current loads, it can't work well if the loads is unstable, it brings many repeat operator to cost the net bandwidth. 
In this pr, hot scheduler will save the history loads and the pick strategy will consider it, it decrease operator count if some store loads are unstable.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
image

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```


pkg/statistics/store_load.go
```diff
@@ -240,3 +243,98 @@ func MaxLoad(a, b *StoreLoad) *StoreLoad {
...
+	s.count++
+	s.update = time.Now()
+}
+
+func (s *storeHistoryLoad) get() [][]float64 {
+	if s.count < s.size {
+		loads := make([][]float64, len(s.loads))
+		for i := range s.loads {
+			loads[i] = make([]float64, s.count)
+			copy(loads[i], s.loads[i][:s.count])
```
nolouch, line 335: can we reduce this copy? I think it's used in the hot scheduler, and there is only one goroutine to read/write.

pkg/schedule/schedulers/hot_region.go
```diff
@@ -789,12 +794,17 @@ func (bs *balanceSolver) filterSrcStores() map[uint64]*statistics.StoreLoadDetai
...
-			ret[id] = detail
-			hotSchedulerResultCounter.WithLabelValues("src-store-succ", strconv.FormatUint(id, 10)).Inc()
-		} else {
-			hotSchedulerResultCounter.WithLabelValues("src-store-failed", strconv.FormatUint(id, 10)).Inc()
+		if !bs.checkSrcByPriorityAndTolerance(detail.LoadPred.Min(), &detail.LoadPred.Expect, srcToleranceRatio) {
+			hotSchedulerResultCounter.WithLabelValues("src-store-failed-"+bs.resourceTy.String(), strconv.FormatUint(id, 10)).Inc()
+			continue
 		}
+
+		if !bs.checkSrcHistoryLoadsByPriorityAndTolerance(&detail.LoadPred.Current, &detail.LoadPred.Expect, srcToleranceRatio) {
```
nolouch, line 802: This will slow down the convergence speed of hotspot scheduling. Have you compared the previous cases with small region? it's expect stable at lease 5mins now. 
BTW, for conservative, we can only do this check in large regions.

bufferflies: good guggestion, return loads direct.

pkg/statistics/store_load.go
bufferflies: good guggestion, return loads direct.

pkg/schedule/schedulers/hot_region.go
bufferflies: Q1: Have you compared the previous cases with small region?
yes, the issue has shown that `hot-region-schduler` brings some useless scheduler in small region. 
Q2: it's expect stable at lease 5mins now.
not yet, PD can scheduler the source or target store if the store load is higher or lower than the average continue. the scheduled store's loads can be unstable in this 5min.
Q3:  we can only do this check in large regions?
maybe, I should one configuration to switch this check.

bufferflies: I have add condition only for the `raftkv2`, so the pr has no influence for the `raftkv1`.

nolouch (approved):
lgtm

pkg/statistics/store_hot_peers_infos.go
```diff
@@ -208,6 +230,13 @@ func summaryStoresLoadByEngine(
 		expectLoads[i] = allStoreLoadSum[i] / float64(allStoreCount)
 	}
 
+	expectHistoryLoads := make([][]float64, DimLen)
+	for i := range allStoreHistoryLoadSum {
+		expectHistoryLoads[i] = make([]float64, len(allStoreHistoryLoadSum[i]))
+		for j := range allStoreHistoryLoadSum[i] {
+			expectHistoryLoads[i][j] = allStoreHistoryLoadSum[i][j] / float64(allStoreCount)
```
lhy1024, line 237: If I understand correctly, the policy now adds an additional set of judgments that require both to be greater than or less than the historical mean value sampled in order to be allowed to be scheduled.Perhaps we can subsequently take other moving average and more lenient probabilities.

lhy1024 (commented):
mostly LGTM if only use in multi rocksdb

pkg/schedule/schedulers/hot_region.go
```diff
@@ -805,6 +820,17 @@ func (bs *balanceSolver) checkSrcByPriorityAndTolerance(minLoad, expectLoad *sta
 	})
 }
 
+func (bs *balanceSolver) checkSrcHistoryLoadsByPriorityAndTolerance(current, expectLoad *statistics.StoreLoad, toleranceRatio float64) bool {
+	if len(current.HistoryLoads) == 0 {
+		return true
+	}
+	return bs.checkHistoryLoadsByPriority(current.HistoryLoads, func(i int) bool {
+		return slice.AllOf(current.HistoryLoads[i], func(j int) bool {
```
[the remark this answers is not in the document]
bufferflies: yes，it should ensune all the sample value is bigger than the history mean loads, not one point.

pkg/schedule/schedulers/hot_region.go
```diff
@@ -461,6 +463,7 @@ type balanceSolver struct {
 	minorDecRatio float64
 	maxPeerNum    int
 	minHotDegree  int
+	isRaftKV2     bool
```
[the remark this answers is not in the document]
bufferflies: yes， In this pr，the new strateegy only work  on the  raftkv2. I will test it on more the test env for single rocksdb.

rleungx (approved):
Mostly, LGTM

pkg/schedule/schedulers/hot_region_test.go
```diff
@@ -563,16 +564,19 @@ func TestHotWriteRegionScheduleByteRateOnlyWithTiFlash(t *testing.T) {
...
 			tc.UpdateStorageWrittenBytes(i, storesBytes[i])
 		}
 	}
+
 	{ // Check the load expect
 		aliveTiKVCount := float64(aliveTiKVLastID - aliveTiKVStartID + 1)
 		allowLeaderTiKVCount := aliveTiKVCount - 1 // store 5 with evict leader
 		aliveTiFlashCount := float64(aliveTiFlashLastID - aliveTiFlashStartID + 1)
 		tc.ObserveRegionsStats()
+		clearPendingInfluence(hb)
```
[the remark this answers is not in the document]
bufferflies: fixed

pkg/statistics/store_hot_peers_infos.go
lhy1024: The average has a disadvantage that it is easily affected by extreme or minimal values, which can make the final result very large or small.

for example, 1,1,1,1,1,20,1,1,1,1,1,1, 1, we should actually consider that his load is at 1, but the average becomes 3

If the remaining two of the three nodes are 3,3,3,3,3,3,3,3, and 3,3,3,3,3,3, we should actually expect one of the nodes to schedule 1 to the first node in front of it, but the current one does not

If filtering the extreme values, the median is better, for the trend of the catch, hma will be better?

The current result is definitely better than master in most scenarios, but I think we should add a todo here

PR #6316, opened 2023-04-13
keyspace: avoid keyspace being updated during the split

JmPotato:

Issue Number: ref #6232. In order to avoid the disruption of consistency, this restriction is added to ensure the integrity of all kinds of information during the split.

```commit-message
Prevent keyspace from being updated during the split.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


rleungx (approved):

nolouch (approved):
lgtm

PR #6332, opened 2023-04-18
resource_manage/metrics: change histogram to vector

### LINKED ISSUE #6066: metrics: consider replace resource manager related histogram with counter
## Enhancement Task

Currently, 1 resource group will result in around ~100 metrics. This will be a big problem if we are going to support 10k resource groups in one cluster. Since we only use the count/sum of these histogram, maybe we should replace the with counter to reduce the total data size.

HuSharp:

Issue Number: Close #6066

```commit-message
resource_manage/metrics: change histogram to vector
```

### Check List

Tests

### Release note

```release-note
None.
```

image


pkg/mcs/resource_manager/server/metrics.go
```diff
@@ -39,65 +37,59 @@ var (
 			Help:      "Indicate the resource manager server info, and the value is the start timestamp (s).",
 		}, []string{"version", "hash"})
 	// RU cost metrics.
-	readRequestUnitCost = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	readRequestUnitCost = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
 			Namespace: namespace,
 			Subsystem: ruSubsystem,
 			Name:      "read_request_unit",
```
glorv, line 44:
```suggestion
			Name:      "read_request_unit_sum",
```
When change histogram to counter ，should add the  `_sum` suffix to keep compatible with the previous names.

pkg/mcs/resource_manager/server/manager.go
```diff
@@ -312,25 +312,25 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
...
 			}
 			// CPU time info.
 			if consumption.TotalCpuTimeMs > 0 {
 				if consumption.SqlLayerCpuTimeMs > 0 {
 					sqlLayerRuMetrics.Add(consumption.SqlLayerCpuTimeMs * m.controllerConfig.RequestUnit.CPUMsCost)
-					sqlCPUMetrics.Observe(consumption.SqlLayerCpuTimeMs)
+					sqlCPUMetrics.Add(consumption.SqlLayerCpuTimeMs)
 				}
-				kvCPUMetrics.Observe(consumption.TotalCpuTimeMs - consumption.SqlLayerCpuTimeMs)
+				kvCPUMetrics.Add(consumption.TotalCpuTimeMs - consumption.SqlLayerCpuTimeMs)
```
bufferflies, line 333: The condition of `TotalCpuTimeMs` less than `SqlLayerCpuTimeMs` is always true.

nolouch:
@HuSharp  
I mean to test the upgrade case to check the metrics is this modification is compatible with before.

bufferflies:
Does the grafana panel need to update also?

pkg/mcs/resource_manager/server/manager.go
HuSharp: Yes, `TotalCpuTimeMs` includes `SqlLayerCpuTimeMs`.

HuSharp:
> Does the grafana panel need to update also?

No need. This is why we need to add the `_sum` suffix to keep compatible with the previous `json names`.

bufferflies (approved):

PR #6336, opened 2023-04-18
client: Count resource penalty for resource control

### LINKED ISSUE #6335: Resource control is not ideal when the request load from different resource group is unbalanced
## Enhancement Task

Consider the case that there are 3 stores, and the request load from one resource group is a hotspot workload. Then all requests from the resource group are handled on one store whereas the requests from the other resource group are distributed evenly on all stores.

As resource control on TiKV only has the local resource consumption, the local QPS ratio may meet the ratio of resource quota of different resource groups but the overall QPS ratio may not meet with skew load. To solve that, the request to one store needs to carry the resource consumption on other stores to make the overall consumption considered.

Connor1996:

Issue Number: Close #6335 

```commit-message
Count the delta resource consumption of the resource group that have completed at all stores between the previous request to this store and current request.
```

### Check List

Tests

- Unit test

Side effects

- Increased code complexity

### Release note

```release-note
Implement dmclock algorithm for resource control
```


Connor1996:
PTAL @glorv

client/resource_group/controller/controller.go
```diff
@@ -429,19 +439,57 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
...
 ) (*rmpb.Consumption, error) {
+	// record resource delta
+	c.mutex.Lock()
+	if resp.Succeed() {
+		m, ok := c.resourceDelta[resourceGroupName]
+		if !ok {
+			m = make(map[uint64]Delta)
+			c.resourceDelta[resourceGroupName] = m
+		}
+		for id, delta := range m {
```
glorv, line 480: There is no need to increase the delta for all groups, but instead you can just maintain a global non-decrease count and add the delta value after each request. And in `OnRequestWait`, the delta can be calculate by `globalCounter - storeCounter` and then set storeCounter to globalCounter, that's it.

client/resource_group/controller/controller.go
```diff
@@ -429,19 +444,60 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
...
+	defer c.mutex.Unlock()
+	m, ok := c.resourceCounter[resourceGroupName]
+	if !ok {
+		m = Counter{
+			storeCounter:  make(map[uint64]Delta),
+			globalCounter: Delta{},
+		}
+		c.resourceCounter[resourceGroupName] = m
+	}
+	delta := Delta{
```
glorv, line 468: What if StoreID is not existed before this request, I think in this case, it should be inited with the value of globalCounter, but here it is implicitly inited with 0. For new joined store, this delta value can be unexpectedly huge

Connor1996: Oh, I miss it

client/resource_group/controller/controller.go
```diff
@@ -429,19 +444,66 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
...
+	if !ok {
+		m = &Counter{
+			storeCounter:  make(map[uint64]Delta),
+			globalCounter: Delta{},
+		}
+		c.resourceCounter[resourceGroupName] = m
+	}
+
+	delta := Delta{}
+	if counter, exist := m.storeCounter[info.StoreID()]; exist {
```
glorv, line 470: storeCounter[info.StoreID()] should inited with globalCount if !exist

glorv (commented):
LGTM

client/resource_group/controller/controller.go
```diff
@@ -97,6 +97,9 @@ type ResourceGroupsController struct {
 
 	calculators []ResourceCalculator
 
+	mutex           sync.Mutex          // used for the `resourceCounter`
```
nolouch, line 100: What about using `sync.Map` directly?

Connor1996: It says 
```
// The Map type is optimized for two common use cases: (1) when the entry for a given
// key is only ever written once but read many times, as in caches that only grow,
// or (2) when multiple goroutines read, write, and overwrite entries for disjoint
// sets of keys. In these two cases, use of a Map may significantly reduce lock
// contention compared to a Go map paired with a separate Mutex or RWMutex.
```
Here doesn't fit the optimized use case, so I just use the go map

nolouch: I think the write/read is in each group, not in the map level(groups). 
BTW, it better moves the Counter to `groupCostController`.

Connor1996: Then, should we still use go map for storeCounter?

Connor1996:
PTAL again @glorv @nolouch

client/resource_group/controller/controller.go
```diff
@@ -1061,12 +1068,27 @@ func (gc *groupCostController) onRequestWait(
...
+	var penalty *rmpb.Consumption
+	if storeCounter, exist := gc.mu.storeCounter[info.StoreID()]; exist {
+		penalty = typeutil.DeepClone(gc.mu.globalCounter, func() *rmpb.Consumption { return &rmpb.Consumption{} })
+		sub(penalty, storeCounter)
+	} else {
+		penalty = &rmpb.Consumption{}
+	}
+	// More accurately, it should be reset when the request succeed. But it would cause all concurrent requests piggyback large delta which inflates penalty.
+	// So here resets it directly as failure is rare.
+	gc.mu.storeCounter[info.StoreID()] = typeutil.DeepClone(gc.mu.globalCounter, func() *rmpb.Consumption { return &rmpb.Consumption{} })
```
glorv, line 1089:
```suggestion
	*gc.mu.storeCounter[info.StoreID()] = *gc.mu.globalCounter
```

glorv: No need to allocate a new object here, just copy and overwrite is enough.

tests/integrations/mcs/resource_manager/resource_manager_test.go
```diff
@@ -493,6 +493,95 @@ func (suite *resourceManagerClientTestSuite) TestSwitchBurst() {
...
+
+	// from different group, should be zero
+	resourceGroupName = suite.initGroups[2].Name
+	req4 := controller.NewTestRequestInfo(true, 50, 1 /* store2 */)
+	resp4 := controller.NewTestResponseInfo(0, time.Duration(10), true)
+	_, penalty, err = c.OnRequestWait(suite.ctx, resourceGroupName, req4)
+	re.NoError(err)
+	re.Equal(penalty.WriteBytes, float64(0))
+	_, err = c.OnResponse(resourceGroupName, req4, resp4)
+	re.NoError(err)
```
CabinfeverB, line 582: pls add  a line `suite.cleanupResourceGroups()`

client/resource_group/controller/controller.go
```diff
@@ -1092,9 +1113,19 @@ func (gc *groupCostController) onResponse(
...
 		}
 	}
+
 	gc.mu.Lock()
+	// record the consumption of the request
 	add(gc.mu.consumption, delta)
+	// record the consumption of the request by store
+	// As the penalty is only counted when the request is completed, so here needs to calculate the write cost which is added in `BeforeKVRequest`
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(delta, req)
```
nolouch, line 1123: I think it will change the original delta, or can we ignore BeforeKV?  why not also do `add(gc.mu.storeCounter[req.StoreID()], delta)` in onRequestWait.

Connor1996: the original delta is a temporary value, it's of no use in the following

client/resource_group/controller/controller.go
```diff
@@ -465,7 +465,9 @@ type groupCostController struct {
 
 	mu struct {
 		sync.Mutex
-		consumption *rmpb.Consumption
+		consumption   *rmpb.Consumption
+		storeCounter  map[uint64]*rmpb.Consumption
```
CabinfeverB, line 469: How add adding a lock? The logic for `storeCounter` and `globalCounter ` is independent of `consumption`

Connor1996: but they almost use the lock at the same time

CabinfeverB: This a hot path. I guess two locks will help get better performance. But maybe  it's not the bottleneck currently

client/resource_group/controller/controller.go
Connor1996: Oh, I missed it

CabinfeverB (approved):

nolouch (approved):

PR #6342, opened 2023-04-19
mcs: add set handler for balancer and alloc node for default keyspace group

lhy1024:
… group

Issue Number: Ref https://github.com/tikv/pd/issues/6233

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -37,8 +37,8 @@ import (
 
 const (
 	defaultBalancerPolicy = balancer.PolicyRoundRobin
-	allocNodeTimeout      = 1 * time.Second
-	allocNodeInterval     = 10 * time.Millisecond
+	allocNodesTimeout     = 1 * time.Second
+	allocNodesInterval    = 10 * time.Millisecond
 	// TODO: move it to etcdutil
 	watchEtcdChangeRetryInterval = 1 * time.Second
 	maxRetryTimes                = 25
@@ -112,6 +112,14 @@ func (m *GroupManager) Bootstrap() error {
 
 	m.Lock()
 	defer m.Unlock()
+
+	// If the etcd client is not nil, start the watch loop.
+	if m.client != nil {
+		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
+		m.wg.Add(1)
+		go m.startWatchLoop()
+	}
+
 	// Ignore the error if default keyspace group already exists in the storage (e.g. PD restart/recover).
 	err := m.saveKeyspaceGroups([]*endpoint.KeyspaceGroup{defaultKeyspaceGroup}, false)
 	if err != nil && err != ErrKeyspaceGroupExists {
@@ -124,15 +132,15 @@ func (m *GroupManager) Bootstrap() error {
 		return err
 	}
 	for _, group := range groups {
+		if group.ID == utils.DefaultKeyspaceGroupID {
+			if len(group.Members) == 0 {
+				// The default keyspace group should have one replica at least.
+				m.wg.Add(1)
+				go m.allocNodesForDefaultKeyspaceGroup(1)
+			}
+		}
 		userKind := endpoint.StringUserKind(group.UserKind)
 		m.groups[userKind].Put(group)
-	}
-
-	// If the etcd client is not nil, start the watch loop.
-	if m.client != nil {
-		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
-		m.wg.Add(1)
-		go m.startWatchLoop()
 	}
 	return nil
 }
@@ -141,6 +149,29 @@ func (m *GroupManager) Bootstrap() error {
 func (m *GroupManager) Close() {
 	m.cancel()
 	m.wg.Wait()
+}
+
+func (m *GroupManager) allocNodesForDefaultKeyspaceGroup(replica int) {
+	defer logutil.LogPanic()
+	defer m.wg.Done()
+	ticker := time.NewTicker(retryInterval)
+	for {
+		select {
+		case <-m.ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		kg, err := m.GetKeyspaceGroupByID(utils.DefaultKeyspaceGroupID)
+		if err == nil && kg != nil && len(kg.Members) >= replica {
+			return
+		}
+		nodes, err := m.AllocNodesForKeyspaceGroup(utils.DefaultKeyspaceGroupID, replica)
+		if err == nil && len(nodes) == replica {
+			log.Info("alloc nodes for default keyspace group", zap.Reflect("nodes", nodes))
+			return
+		}
+		log.Warn("failed to alloc nodes for default keyspace group", zap.Error(err))
+	}
 }
 
 func (m *GroupManager) startWatchLoop() {
@@ -590,14 +621,19 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 
 // GetNodesNum returns the number of nodes.
 func (m *GroupManager) GetNodesNum() int {
+	if m.nodesBalancer == nil {
+		return 0
+	}
 	return m.nodesBalancer.Len()
 }
 
 // AllocNodesForKeyspaceGroup allocates nodes for the keyspace group.
 func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, replica int) ([]endpoint.KeyspaceGroupMember, error) {
-	ctx, cancel := context.WithTimeout(m.ctx, allocNodeTimeout)
+	m.Lock()
+	defer m.Unlock()
+	ctx, cancel := context.WithTimeout(m.ctx, allocNodesTimeout)
 	defer cancel()
-	ticker := time.NewTicker(allocNodeInterval)
+	ticker := time.NewTicker(allocNodesInterval)
 	defer ticker.Stop()
 	nodes := make([]endpoint.KeyspaceGroupMember, 0, replica)
 	err := m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
@@ -612,6 +648,9 @@ func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, replica int) ([]end
 		for _, member := range kg.Members {
 			exists[member.Address] = struct{}{}
 			nodes = append(nodes, member)
+		}
+		if len(exists) >= replica {
+			return nil
 		}
 		for len(exists) < replica {
 			select {
@@ -640,4 +679,36 @@ func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, replica int) ([]end
 		return nil, err
 	}
 	return nodes, nil
+}
+
+// SetNodesForKeyspaceGroup sets the nodes for the keyspace group.
+func (m *GroupManager) SetNodesForKeyspaceGroup(id uint32, nodes []string) error {
+	m.Lock()
+	defer m.Unlock()
+	return m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
+		kg, err := m.store.LoadKeyspaceGroup(txn, id)
+		if err != nil {
+			return err
+		}
+		if kg == nil {
+			return ErrKeyspaceGroupNotExists
+		}
+		members := make([]endpoint.KeyspaceGroupMember, 0, len(nodes))
+		for _, node := range nodes {
+			members = append(members, endpoint.KeyspaceGroupMember{Address: node})
+		}
+		kg.Members = members
+		return m.store.SaveKeyspaceGroup(txn, kg)
+	})
+}
+
+// IsExistNode checks if the node exists.
+func (m *GroupManager) IsExistNode(addr string) bool {
+	nodes := m.nodesBalancer.GetAll()
+	for _, node := range nodes {
+		if node == addr {
+			return true
+		}
+	}
+	return false
 }
--- a/server/apiv2/handlers/tso_keyspace_group.go
+++ b/server/apiv2/handlers/tso_keyspace_group.go
@@ -35,7 +35,8 @@ func RegisterTSOKeyspaceGroup(r *gin.RouterGroup) {
 	router.GET("", GetKeyspaceGroups)
 	router.GET("/:id", GetKeyspaceGroupByID)
 	router.DELETE("/:id", DeleteKeyspaceGroupByID)
-	router.POST("/:id/alloc", AllocNodeForKeyspaceGroup)
+	router.POST("/:id/alloc", AllocNodesForKeyspaceGroup)
+	router.POST("/:id/nodes", SetNodesForKeyspaceGroup)
 	router.POST("/:id/split", SplitKeyspaceGroupByID)
 	router.DELETE("/:id/split", FinishSplitKeyspaceByID)
 }
@@ -190,13 +191,13 @@ func FinishSplitKeyspaceByID(c *gin.Context) {
 	c.JSON(http.StatusOK, nil)
 }
 
-// AllocNodeForKeyspaceGroupParams defines the params for allocating nodes for keyspace groups.
-type AllocNodeForKeyspaceGroupParams struct {
+// AllocNodesForKeyspaceGroupParams defines the params for allocating nodes for keyspace groups.
+type AllocNodesForKeyspaceGroupParams struct {
 	Replica int `json:"replica"`
 }
 
-// AllocNodeForKeyspaceGroup allocates nodes for keyspace group.
-func AllocNodeForKeyspaceGroup(c *gin.Context) {
+// AllocNodesForKeyspaceGroup allocates nodes for keyspace group.
+func AllocNodesForKeyspaceGroup(c *gin.Context) {
 	id, err := validateKeyspaceGroupID(c)
 	if err != nil {
 		c.AbortWithStatusJSON(http.StatusBadRequest, "invalid keyspace group id")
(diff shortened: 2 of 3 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -143,6 +151,29 @@ func (m *GroupManager) Close() {
 	m.wg.Wait()
 }
 
+func (m *GroupManager) allocNodesForDefaultKeyspaceGroup(replica int) {
+	defer logutil.LogPanic()
+	defer m.wg.Done()
+	ticker := time.NewTicker(retryInterval)
+	for {
```
lhy1024, line 158: Do we need timeout for it?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -143,6 +152,30 @@ func (m *GroupManager) Close() {
...
+	ticker := time.NewTicker(allocNodesForDefaultKeyspaceGroupInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-m.ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		kg, err := m.GetKeyspaceGroupByID(utils.DefaultKeyspaceGroupID)
+		replica := m.GetNodesNum()
```
binshi-bing, line 170: GetNodesNum() returns the count of registered tso servers, correct? If yes, shall we rename replica to countOfTSONodes or numOfTSONodes or desiredReplicaCount or similar?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -143,6 +152,30 @@ func (m *GroupManager) Close() {
...
+		if err == nil && kg != nil && len(kg.Members) >= replica {
```
binshi-bing, line 171: from this line, if my understanding is correct, we want to assign a replica/member of default keyspace group to evernt tso node, but this isn't true. Currently tso service use this logic, because the kg.Members is empty. I remember we can specify the desired replica/member count (default value is 2 if it isn't specified) of a keyspace group when creating a keyspace group, so we should use that desired replica/member count instead of GetNodesNum() here.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -221,7 +256,7 @@ func (m *GroupManager) watchServiceAddrs(ctx context.Context, revision int64) (i
 			for _, event := range wresp.Events {
 				s := &discovery.ServiceRegistryEntry{}
 				if err := json.Unmarshal(event.Kv.Value, s); err != nil {
```
binshi-bing: Can't unmarshal here. see https://github.com/tikv/pd/pull/6346 Fixed bugs in tso service registry watching loop

binshi-bing: And use "revision = wresp.Header.Revision + 1" somewhere too.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +155,30 @@ func (m *GroupManager) Close() {
...
+	defer ticker.Stop()
+	for {
+		select {
+		case <-m.ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		kg, err := m.GetKeyspaceGroupByID(utils.DefaultKeyspaceGroupID)
+		replica := m.GetNodesNum()
+		if err == nil && kg != nil && len(kg.Members) >= replica {
```
lhy1024, line 171: add all nodes for the default keyspace group, because we do not support offline nodes in keyspace balancer, but we test delete tso node primary in test. If only one replica, some test will fail.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -590,14 +625,19 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 
 // GetNodesNum returns the number of nodes.
 func (m *GroupManager) GetNodesNum() int {
```
binshi-bing, line 642:
```suggestion
func (m *GroupManager) GetNodesCount() int {
```
Use count instead of number will be clearer.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -590,14 +625,19 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 
 // GetNodesNum returns the number of nodes.
```
binshi-bing, line 641:
```suggestion
// GetNodesCount returns the count of nodes.
```

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -112,6 +113,14 @@ func (m *GroupManager) Bootstrap() error {
 
 	m.Lock()
 	defer m.Unlock()
+
+	// If the etcd client is not nil, start the watch loop.
+	if m.client != nil {
+		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
```
binshi-bing, line 121:
```suggestion
		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
		m.serviceRegistryMap = make(map[string]string)
```
because of https://github.com/tikv/pd/pull/6346 being merged

binshi-bing (commented):
left some comments

pkg/keyspace/tso_keyspace_group.go
lhy1024: https://github.com/tikv/pd/pull/6342#discussion_r1172046709

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -232,6 +233,54 @@ func AllocNodeForKeyspaceGroup(c *gin.Context) {
...
+	}
+	svr := c.MustGet(middlewares.ServerContextKey).(*server.Server)
+	manager := svr.GetKeyspaceGroupManager()
+	setParams := &SetNodesForKeyspaceGroupParams{}
+	err = c.BindJSON(setParams)
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, errs.ErrBindJSON.Wrap(err).GenWithStackByCause())
+		return
+	}
+	// check keyspace group whether exist
```
binshi-bing, line 257:
```suggestion
	// check if keyspace group exists
```

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -232,6 +233,54 @@ func AllocNodeForKeyspaceGroup(c *gin.Context) {
...
+	keyspaceGroup, err := manager.GetKeyspaceGroupByID(id)
+	if err != nil || keyspaceGroup == nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, "keyspace group does not exist")
+		return
+	}
+	// check nodes whether empty
```
binshi-bing, line 263: ditto

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -232,6 +233,54 @@ func AllocNodeForKeyspaceGroup(c *gin.Context) {
...
+	if len(setParams.Nodes) == 0 {
+		c.AbortWithStatusJSON(http.StatusBadRequest, "invalid empty nodes")
+		return
+	}
+	// check nodes whether exist
```
binshi-bing, line 268: ditto

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -628,14 +669,17 @@ func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, replica int) ([]end
...
+		for len(exists) < desiredReplicaCount {
 			select {
 			case <-ctx.Done():
 				return nil
 			case <-ticker.C:
 			}
-			num := m.GetNodesNum()
-			if num < replica || num == 0 { // double check
+			countOfNodes := m.GetNodesCount()
+			if countOfNodes < desiredReplicaCount || countOfNodes == 0 { // double check
```
binshi-bing, line 682: do we still need countOfNodes == 0 when countOfNodes >= desiredReplicaCount?

binshi-bing (approved):
LGTM

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -114,6 +115,15 @@ func (m *GroupManager) Bootstrap() error {
 
 	m.Lock()
 	defer m.Unlock()
+
+	// If the etcd client is not nil, start the watch loop.
+	if m.client != nil {
+		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
```
rleungx, line 121: Can we move them to where we new group manager?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -114,6 +115,15 @@ func (m *GroupManager) Bootstrap() error {
...
+		m.serviceRegistryMap = make(map[string]string)
+		m.wg.Add(1)
+		go m.startWatchLoop()
```
rleungx, line 124: When do we use wg?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -126,17 +136,16 @@ func (m *GroupManager) Bootstrap() error {
 		return err
 	}
 	for _, group := range groups {
+		if group.ID == utils.DefaultKeyspaceGroupID {
```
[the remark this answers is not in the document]
lhy1024: Do we also need to check params for set-nodes and alloc-nodes interface?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -126,17 +136,16 @@ func (m *GroupManager) Bootstrap() error {
...
+			if len(group.Members) == 0 && m.client != nil {
```
rleungx, line 140: Can we only check m.client once?

pkg/keyspace/tso_keyspace_group.go
lhy1024: consider meeting offline node

pkg/tso/keyspace_group_manager.go
```diff
@@ -728,6 +728,9 @@ func (kgm *KeyspaceGroupManager) GetElectionMember(
 	if err != nil {
 		return nil, err
 	}
+	if am == nil {
```
rleungx, line 731: When will this happen?

lhy1024: check if it is primary after `kgm.deleteKeyspaceGroup` with default keyspace group

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -95,8 +96,9 @@ func NewKeyspaceGroupManager(ctx context.Context, store endpoint.KeyspaceGroupSt
 		client:           client,
 		tsoServiceKey:    key,
 		tsoServiceEndKey: clientv3.GetPrefixRangeEnd(key) + "/",
-		policy:           defaultBalancerPolicy,
 		groups:           groups,
+		nodesBalancer:    balancer.GenByPolicy[string](defaultBalancerPolicy),
+		serviceRegistryMap: make(map[string]string),
```
JmPotato, line 101: Do we still need this map?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+	defer ticker.Stop()
+	for {
+		select {
+		case <-m.ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		countOfNodes := m.GetNodesCount()
+		if countOfNodes < utils.KeyspaceGroupDefaultReplicaCount {
+			log.Info("the count of nodes is not enough to allocate the default keyspace group", zap.Int("count", countOfNodes))
```
JmPotato, line 164: Will this log keep flushing every second?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+					log.Error("failed to alloc default nodes for keyspace group", zap.Error(err))
+					continue
+				}
+				log.Info("alloc default nodes for keyspace group", zap.Int("count", len(nodes)))
+				group.Members = nodes
+			}
+		}
+		if !withError {
+			// all keyspace groups have equal or more than default replica count
+			return
```
JmPotato, line 187: Continue or return?

pkg/tso/keyspace_group_manager.go
rleungx: Do you mean delete the default keyspace group?

lhy1024: the default keyspace group no longer contain this member

pkg/keyspace/tso_keyspace_group.go
lhy1024: see https://github.com/tikv/pd/pull/6346

pkg/keyspace/tso_keyspace_group.go
lhy1024: I will adjust it

pkg/keyspace/tso_keyspace_group.go
lhy1024: return. It will only occur when load keyspace group. In the future, we support scale-out and update nodes in balancer, it will update always. cc @binshi-bing

binshi-bing: If we have done all the work here, then we can return. The problem is that it seems that we have a severe bug -- when utils.KeyspaceGroupDefaultReplicaCount is 2 and there are two tso nodes just registered with the third tso node upcoming, this function will proceed to load all keyspace groups and allocate 2 nodes to all of them. After the third TSO node being registered, it won't be assigned any keyspace group. This is common case when we to create a new cluster including api service and tso service, where we first create API nodes then gradually add tso nodes.

Because of this reason, I prefer to let operator manually call balance api, after all tso nodes are registered, to assign tso nodes to the keyspace groups whose member count is less than utils.KeyspaceGroupDefaultReplicaCount instead of doing this job in group manager's bootstrap.

Let's discuss this on Monday. @rleungx @JmPotato

binshi-bing (changes requested):
added one critical comment

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
 	m.wg.Wait()
 }
 
+func (m *GroupManager) allocDefaultNodesForKeyspaceGroup() {
```
binshi-bing, line 151:
```suggestion
func (m *GroupManager) allocNodesToAllKeyspaceGroups() {
```

It seems that here we try to alloc nodes to all keyspace groups, so the original name is confusing.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+	defer logutil.LogPanic()
+	defer m.wg.Done()
+	ticker := time.NewTicker(allocNodesForDefaultKeyspaceGroupInterval)
```
binshi-bing, line 154:
```suggestion
	ticker := time.NewTicker(allocNodesToKeyspaceGroupsInterval)
```
ditto

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+			log.Error("failed to load the default keyspace group", zap.Error(err))
+			continue
+		}
+		withError := false
+		for _, group := range groups {
+			if len(group.Members) < utils.KeyspaceGroupDefaultReplicaCount {
+				nodes, err := m.AllocNodesForKeyspaceGroup(group.ID, utils.KeyspaceGroupDefaultReplicaCount)
+				if err != nil {
+					withError = true
+					log.Error("failed to alloc default nodes for keyspace group", zap.Error(err))
```
binshi-bing, line 178: can we also print keyspace group id?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+					continue
+				}
+				log.Info("alloc default nodes for keyspace group", zap.Int("count", len(nodes)))
```
binshi-bing, line 181: can we also print keyspace group id?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -146,6 +148,47 @@ func (m *GroupManager) Close() {
...
+		case <-ticker.C:
+		}
+		countOfNodes := m.GetNodesCount()
+		if countOfNodes < utils.KeyspaceGroupDefaultReplicaCount {
+			log.Info("the count of nodes is not enough to allocate the default keyspace group", zap.Int("count", countOfNodes))
+			continue
+		}
+		groups, err := m.store.LoadKeyspaceGroups(utils.DefaultKeyspaceGroupID, 0)
+		if err != nil {
+			log.Error("failed to load the default keyspace group", zap.Error(err))
```
binshi-bing, line 169: we pass limit = 0 in m.store.LoadKeyspaceGroups which intends to load all keyspace group, so the "
the default keyspace group" in this log is mismatching.

binshi-bing (changes requested):

pkg/keyspace/tso_keyspace_group.go
binshi-bing: @lhy1024 , after some thoughts, let's keep your way, because we'll mostly setup 2 tso nodes for now and we can also use move keyspace group api to change the distribution.

binshi-bing (approved):
LGTM

pkg/tso/keyspace_group_manager.go
rleungx: If it actually goes here, it will panic I think.

lhy1024: it can be removed after @binshi-bing has fixed

rleungx (approved):

PR #6345, opened 2023-04-19
server: Resign as leader on exit or SIGHUP

### LINKED ISSUE #6344: Resign leadership on exit
## Development Task

When pd-server exists or is sent SIGHUP/SIGTERM it does not transfer the member leadership, which in some cases can cause long wait (30-50 sec) until a new PD leader is elected.

mjonss:
close tikv/pd#6344

Issue Number: Close #6344

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/AlekSi/gocov-xml v1.0.0
 	github.com/BurntSushi/toml v0.3.1
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
```
bufferflies, line 8: where is it used?

cmd/pd-server/main.go
```diff
@@ -224,7 +224,31 @@ func start(cmd *cobra.Command, args []string, services ...string) {
...
-		sig = <-sc
+	SignalHandler:
+		for {
+			select {
+			case sig = <-sc:
+				log.Info("Received a signal", zap.String("signal", sig.String()))
+				if svr.GetMember().IsLeader() {
+					log.Info("Transferring leadership due to signal", zap.String("signal", sig.String()))
+					err := svr.GetMember().ResignEtcdLeader(context.TODO(), svr.Name(), "")
+					if err != nil {
```
bufferflies, line 235: maybe you needs seconds to wait the pd leader has transfer also?

mjonss: Thanks for the suggestion @bufferflies , I have updated the PR. Are there any other ways of resigning (including waiting until another leader has taken over)?

coderplay:
I'm not sure if this patch is intended to solve the query latency spike issue caused by rolling restart of PD. If so, this patch may not completely solve the problem.

We conducted an experiment by simulating the rolling update process of PD using the command `/pd-ctl member leader transfer` manually. We made sure that the leadership of a PD was transferred to another PD before shutting it down. However, we still observed a 2-3 minute latency spike in P100 query latency.

Screenshot 2023-04-24 at 9 02 01 AM

cmd/pd-server/main.go
```diff
@@ -224,7 +225,36 @@ func start(cmd *cobra.Command, args []string, services ...string) {
...
+							log.Info("Transferred leadership")
+						}
+					}
+				}
+				switch sig {
+				case syscall.SIGHUP:
+					log.Info("got SIGHUP, waiting for next signal")
+				default:
+					log.Info("Exit signal handler")
+					break SignalHandler
```
coderplay, line 254: Looks like this `break` will bypass the function call of `cancel()` , will it cause any problem?

cmd/pd-server/main.go
```diff
@@ -224,7 +225,36 @@ func start(cmd *cobra.Command, args []string, services ...string) {
...
 	go func() {
-		sig = <-sc
+	SignalHandler:
+		for {
+			select {
+			case sig = <-sc:
+				log.Info("Received a signal", zap.String("signal", sig.String()))
+				if svr.GetMember().IsLeader() {
+					log.Info("Transferring leadership due to signal", zap.String("signal", sig.String()))
+					err := svr.GetMember().ResignEtcdLeader(context.TODO(), svr.Name(), "")
```
coderplay, line 235: 1. The `ResignEtcdLeader` used here does not transfer to a specified leader, it just do the resign. Can it save the etcd election timeout and leader re-election time?

2. IIRC, PD leader transfer may take several minutes in some case, and TiDB Operator even sets a wait time of up to 1500 minutes. What should we do in that case?

3. Kubernetes sends a SIGTERM signal to the process and waits for a certain number of seconds (30 by default) for it to shut down gracefully. If it does not shut down in time, the process is then killed through SIGKILL. How should we handle this situation?

mjonss: I did not manage to get the PD leader transfer to take several minutes, when testing on my laptop. I will follow up with the developer team working on PD. I was just trying to see if triggering a Resign during shutdown would help, but apparently it is not the issue here.

mjonss:
> I'm not sure if this patch is intended to solve the query latency spike issue caused by rolling restart of PD. If so, this patch may not solve the problem.
> 
> We conducted an experiment by simulating the rolling update process of PD using the command `/pd-ctl member leader transfer` manually. We made sure that the leadership of a PD was transferred to another PD WITHOUT shutting it down. However, we still observed 2-3 minute spikes in P100 TiDB query latency.
> 
> This experiment demonstrates that even with graceful shutdown, re-election still occurs and it impacts TiDB queries. This PR should only save the time for PD to realize there is no leader (etcd election timeout, by default it's 1 second).
> 
> cc: @nolouch

@coderplay Is it only a P100/long tail latency, or all TiDB queries that are affected in your experiment?

cmd/pd-server/main.go
coderplay: Oh really?  Looks like the SignalHandler block includes the line of `cancel()`.  Breaking the `SignalHandler` will skip running the `cancel()` , right?

coderplay: Can we avoid any misunderstandings caused by this, regardless of whether you are right or I am right in the end?

mjonss: I just checked, and the `cancel()` is called, I checked by adding a log between the loop and the cancel call, and it was in the output. But I agree with you that it is not the issue here.

cmd/pd-server/main.go
bufferflies: PD should grant new lease after the new pd leader compaign success ,so you can use watch api to listen the change .

coderplay (approved):

Issue #6343, opened 2023-04-19
keyspace group distribution manager failed to unmarshal service registry entry


binshi-bing:
## Enhancement Task
Issue Number: Ref #6233
There are two bugs:
1. When re-watch a range, to continue from what left by the last watch, the revision is wresp.Header.Revision + 1 instead of wresp.Header.Revision, where wresp.Header.Revision is the revision indicated in the response of the last watch. Because of this bug, it was processing the same event endless as you can see from the log below.
2. In tso service watch loop in /Users/binshi/code/pingcap/my-pd/pkg/keyspace/tso_keyspace_group.go,  If this is delete event, the json.Unmarshal(event.Kv.Value, s) will fail with the error "unexpected end of JSON input", so there is no way to get s.serviceAddr from the result of json.Unmarshal.

under /pd/tests/integrations/mcs, run "go test -run TestAPIServerForwardTestSuite/TestForwardTSOWhenPrimaryChanged tso/server_test.go -v". Though the test passed, but the log contains numberless suspicious log lines:

2023/04/19 14:40:11.285 -07:00] [WARN] [tso_keyspace_group.go:224] ["failed to unmarshal service registry entry"] [error="unexpected end of JSON input"]
[2023/04/19 14:40:11.285 -07:00] [WARN] [tso_keyspace_group.go:224] ["failed to unmarshal service registry entry"] [error="unexpected end of JSON input"]
[2023/04/19 14:40:11.285 -07:00] [WARN] [tso_keyspace_group.go:224] ["failed to unmarshal service registry entry"] [error="unexpected end of JSON input"]
...

[2023/04/19 15:14:05.499 -07:00] [ERROR] [operator_controller.go:878] ["invalid store ID"] [store-id=11]
[2023/04/19 15:14:05.499 -07:00] [INFO] [operator_controller.go:618] ["operator canceled"] [region-id=9] [takes=0s] [operator="\"add-rule-peer {add peer: store [1]} (kind:replica,region,leader, region:9(0, 0), createAt:2023-04-19 15:14:05.499451 -0700 PDT m=+18.349748751, startAt:0001-01-01 00:00:00 +0000 UTC, currentStep:0, size:1, steps:[add learner peer 18 on store 1, promote learner peer 18 on store 1 to voter, transfer leader from store 11 to store 1]，timeout:[12m0s])\""] [additional-info=]
[...]

PR #6346, opened 2023-04-19
Fixed bugs in tso service registry watching loop.

binshi-bing:

Issue Number: Ref #6343 

```commit-message
Fixed the following two bugs:
1. When re-watch a range, to continue from what left by the last watch, the revision is wresp.Header.Revision + 1 instead of wresp.Header.Revision, where wresp.Header.Revision is the revision indicated in the response of the last watch. Because of this bug, it was processing the same event endless as you can see from the log below.
2. In tso service watch loop in /Users/binshi/code/pingcap/my-pd/pkg/keyspace/tso_keyspace_group.go, If this is delete event, the json.Unmarshal(event.Kv.Value, s) will fail with the error "unexpected end of JSON input", so there is no way to get s.serviceAddr from the result of json.Unmarshal.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -76,6 +76,8 @@ type GroupManager struct {
 	// TODO: add user kind with different balancer
 	// when we ensure where the correspondence between tso node and user kind will be found
 	nodesBalancer balancer.Balancer[string]
+	// serviceRegistryMap stores the mapping from the service registry key to the service address.
+	serviceRegistryMap map[string]string
 }
 
 // NewKeyspaceGroupManager creates a Manager of keyspace group related data.
@@ -131,6 +133,7 @@ func (m *GroupManager) Bootstrap() error {
 	// If the etcd client is not nil, start the watch loop.
 	if m.client != nil {
 		m.nodesBalancer = balancer.GenByPolicy[string](m.policy)
+		m.serviceRegistryMap = make(map[string]string)
 		m.wg.Add(1)
 		go m.startWatchLoop()
 	}
@@ -169,6 +172,7 @@ func (m *GroupManager) startWatchLoop() {
 					continue
 				}
 				m.nodesBalancer.Put(s.ServiceAddr)
+				m.serviceRegistryMap[string(item.Key)] = s.ServiceAddr
 			}
 			break
 		}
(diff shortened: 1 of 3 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -219,17 +223,28 @@ func (m *GroupManager) watchServiceAddrs(ctx context.Context, revision int64) (i
...
-				s := &discovery.ServiceRegistryEntry{}
-				if err := json.Unmarshal(event.Kv.Value, s); err != nil {
-					log.Warn("failed to unmarshal service registry entry", zap.Error(err))
-				}
 				switch event.Type {
 				case clientv3.EventTypePut:
+					s := &discovery.ServiceRegistryEntry{}
+					if err := json.Unmarshal(event.Kv.Value, s); err != nil {
+						log.Warn("failed to unmarshal service registry entry",
+							zap.ByteString("event-kv-key", event.Kv.Key), zap.Error(err))
```
rleungx, line 231: How about 
```suggestion
							zap.String("event-kv-key", string(event.Kv.Key)), zap.Error(err))
```

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -219,17 +223,28 @@ func (m *GroupManager) watchServiceAddrs(ctx context.Context, revision int64) (i
 				return revision, wresp.Err()
 			}
 			for _, event := range wresp.Events {
-				s := &discovery.ServiceRegistryEntry{}
-				if err := json.Unmarshal(event.Kv.Value, s); err != nil {
```
binshi-bing, line 223: for delete event, json.Unmarshal(event.Kv.Value, s) will result in "unexpected end of JSON input" because of nil value. so it has to move to switch-case clientv3.EventTypePut:

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -219,17 +223,28 @@ func (m *GroupManager) watchServiceAddrs(ctx context.Context, revision int64) (i
...
 				switch event.Type {
 				case clientv3.EventTypePut:
+					s := &discovery.ServiceRegistryEntry{}
+					if err := json.Unmarshal(event.Kv.Value, s); err != nil {
+						log.Warn("failed to unmarshal service registry entry",
+							zap.ByteString("event-kv-key", event.Kv.Key), zap.Error(err))
+						break
+					}
 					m.nodesBalancer.Put(s.ServiceAddr)
+					m.serviceRegistryMap[string(event.Kv.Key)] = s.ServiceAddr
```
rleungx, line 235: Do we need a lock?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -219,17 +223,28 @@ func (m *GroupManager) watchServiceAddrs(ctx context.Context, revision int64) (i
...
 				case clientv3.EventTypeDelete:
-					m.nodesBalancer.Delete(s.ServiceAddr)
```
binshi-bing, line 230: ditto. the unmarshal above failed, so s.ServiceAddr actually is nil

pkg/keyspace/tso_keyspace_group.go
binshi-bing: nope. It's only accessed in the tso service entry watching loop.

binshi-bing: serviceRegistryMap is a peer structure of m.nodesBalancer which also doesn't need lock.

pkg/keyspace/tso_keyspace_group.go
binshi-bing: ok. Changed.

rleungx (approved):

JmPotato (approved):

PR #6354, opened 2023-04-21
pkg: fix tso inconsistent problem

### LINKED ISSUE #6355: TSO could be inconsistent between memory and etcd
See https://github.com/tidbcloud/pd-cse/issues/74

rleungx:

Issue Number: Close #6355.

Previously, we use txn to save timestamps. But if the timestamp is equal to or less than the previous one, we only return nil, which may cause memory to have an inconsistent state. This PR returns an error in such a case.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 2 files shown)


JmPotato (approved):

lhy1024:
lgtm

lhy1024 (approved):

PR #6356, opened 2023-04-21
mcs, tests: use TSO cluster to do the failover test

JmPotato:

Issue Number: ref #5895.

```commit-message
Use TSO cluster to do the failover test.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/integrations/tso/client_test.go
+++ b/tests/integrations/tso/client_test.go
@@ -183,7 +183,8 @@ func (suite *tsoClientTestSuite) TestRandomResignLeader() {
 
 	ctx, cancel := context.WithCancel(suite.ctx)
 	var wg sync.WaitGroup
-	wg.Add(tsoRequestConcurrencyNumber + 1)
+	checkTSO(ctx, re, &wg, suite.backendEndpoints)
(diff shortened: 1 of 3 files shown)


tests/integrations/tso/client_test.go
```diff
@@ -278,32 +272,32 @@ func TestMixedTSODeployment(t *testing.T) {
...
-					re.Less(lastTS, ts)
-					lastTS = ts
-				}
 				select {
 				case <-ctx.Done():
-					physical, logical, _ := cli.GetTS(context.Background())
-					ts = tsoutil.ComposeTS(physical, logical)
-					re.Less(lastTS, ts)
+					// Make sure the lastTS is not empty
+					re.NotEmpty(lastTS)
```
rleungx, line 289: Why not compare again?

JmPotato: `re.Less` below is to do the check, here I just want to make sure the `lastTS` is set at least once.

rleungx (approved):

PR #6362, opened 2023-04-23
mcs, tso: fix ts fallback caused by multi-primary of the same keyspace group 

binshi-bing:

Issue Number: Ref #6355 

```commit-message
Change participant election-prifix from listen-addr to advertise-listen-addr to guarantee uniqueness.
If multiple participants have the same unique name/id, they all could become the primary and cause ts fallback seen by the client.

```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


JmPotato (commented):
LGTM. What about adding a test?

rleungx (approved):
LGTM, +1 for adding a simple test.

JmPotato (approved):

PR #6373, opened 2023-04-24
mcs, tso: add API interface to obtain the TSO keyspace group member info

JmPotato:

Issue Number: ref #6232.

```commit-message
Add API interface to obtain the TSO keyspace group member info.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/tso/server/apis/v1/api.go
+++ b/pkg/mcs/tso/server/apis/v1/api.go
@@ -22,16 +22,22 @@ import (
 	"github.com/gin-contrib/gzip"
 	"github.com/gin-gonic/gin"
 	"github.com/joho/godotenv"
+	"github.com/pingcap/kvproto/pkg/tsopb"
+	"github.com/pingcap/log"
 	tsoserver "github.com/tikv/pd/pkg/mcs/tso/server"
 	"github.com/tikv/pd/pkg/mcs/utils"
+	"github.com/tikv/pd/pkg/storage/endpoint"
 	"github.com/tikv/pd/pkg/tso"
 	"github.com/tikv/pd/pkg/utils/apiutil"
 	"github.com/tikv/pd/pkg/utils/apiutil/multiservicesapi"
 	"github.com/unrolled/render"
+	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/mcs/tso/server/apis/v1/api.go
```diff
@@ -77,30 +83,64 @@ func NewService(srv *tsoserver.Service) *Service {
...
-		s.apiHandlerEngine.ServeHTTP(w, r)
-	})
+// RegisterKeyspaceGroupRouter registers the router of the TSO keyspace group handler.
+func (s *Service) RegisterKeyspaceGroupRouter() {
+	router := s.root.Group("keyspace-groups")
+	router.GET("/members", GetKeyspaceGroupMembers)
+}
+
+// KeyspaceGroupMember contains the keyspace group and its member information.
+type KeyspaceGroupMember struct {
```
binshi-bing, line 117: The information that I plan to get from every tso server includes:
1. TSO nodes. 
2. All keyspace groups; the full member/replica (distribution) list of every keyspace group; the primary of every keyspace group.

Every tso node can return the same result (global) picture with this API. The current implementation in this pr is that a KeyspaceGroupMember won't contain another member, so different node returns different partial information.

To return the required information, every tso node needs to watch tso primary addresses of all keyspace groups and watch all registered tso nodes/pods, which we can add after Haoyang completes the refactor work of the initial loading of the range and dynamic watch.

binshi-bing: Let's discuss about it so that we are on the same page.

binshi-bing (commented):
left comments.

pkg/mcs/tso/server/apis/v1/api.go
lhy1024: do we need to return its keyspaces?

binshi-bing: endpoint.KeyspaceGroup contains keyspace list, which shall be enough.

binshi-bing (approved):
LGTM

rleungx (approved):

PR #6380, opened 2023-04-25
client: add `NewClientWithKeyspaceName` for client

rleungx:

Issue Number: Ref #5895.

Due to TiDB using the string as the keyspace identity, we need to handle such a case and create a PD client through the name.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


binshi-bing (approved):
LGTM

lhy1024 (approved):

PR #6388, opened 2023-04-26
keyspace: patrol keyspace assignment before the first split

JmPotato:

Issue Number: ref #6232.

```commit-message
Patrol the keyspace assignment before the first split to make sure every keyspace has its group assignment.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/kvproto/pkg/keyspacepb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/balancer"
 	"github.com/tikv/pd/pkg/mcs/discovery"
@@ -56,6 +57,7 @@ type GroupManager struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
+
 	// the lock for the groups
 	sync.RWMutex
 	// groups is the cache of keyspace group related information.
@@ -63,7 +65,10 @@ type GroupManager struct {
 	groups map[endpoint.UserKind]*indexedHeap
 
 	// store is the storage for keyspace group related information.
-	store endpoint.KeyspaceGroupStorage
+	store interface {
+		endpoint.KeyspaceGroupStorage
+		endpoint.KeyspaceStorage
+	}
 
 	client *clientv3.Client
 
@@ -77,10 +82,21 @@ type GroupManager struct {
 	nodesBalancer balancer.Balancer[string]
 	// serviceRegistryMap stores the mapping from the service registry key to the service address.
 	serviceRegistryMap map[string]string
+
+	// patrolKeyspaceAssignmentOnce is used to patrol all keyspaces and assign them to the keyspace groups.
+	patrolKeyspaceAssignmentOnce sync.Once
 }
 
 // NewKeyspaceGroupManager creates a Manager of keyspace group related data.
-func NewKeyspaceGroupManager(ctx context.Context, store endpoint.KeyspaceGroupStorage, client *clientv3.Client, clusterID uint64) *GroupManager {
+func NewKeyspaceGroupManager(
+	ctx context.Context,
+	store interface {
+		endpoint.KeyspaceGroupStorage
+		endpoint.KeyspaceStorage
+	},
+	client *clientv3.Client,
+	clusterID uint64,
+) *GroupManager {
 	ctx, cancel := context.WithCancel(ctx)
 	key := discovery.TSOPath(clusterID)
 	groups := make(map[endpoint.UserKind]*indexedHeap)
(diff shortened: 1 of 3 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +172,31 @@ func (m *GroupManager) Close() {
...
+		config, err := m.GetKeyspaceConfigByKind(endpoint.Basic)
+		if err != nil {
+			return
+		}
+		for _, ks := range keyspaces {
+			if ks == nil {
+				continue
+			}
+			err = m.UpdateKeyspaceForGroup(endpoint.Basic, config[TSOKeyspaceGroupIDKey], ks.GetId(), opAdd)
+			if err != nil {
```
lhy1024, line 192: if meets err, then we will cannot continue to patrol other keyspaces?

JmPotato: Nice catch. I think we can not use `sync.Once` here.

lhy1024: Maybe we can put it into `allocNodesToAllKeyspaceGroups`, because they are similiar. If so, we need to change its name.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -147,7 +161,8 @@ func (m *GroupManager) Bootstrap() error {
 		m.groups[userKind].Put(group)
 	}
 
-	return nil
+	// Load all the keyspaces from the storage and assign them to the respective keyspace groups.
+	return m.patrolKeyspaceAssignmentLocked()
```
rleungx, line 165: Will it affect the time of the transfer leader?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +171,42 @@ func (m *GroupManager) Close() {
...
+	keyspaces, err := m.store.LoadRangeKeyspace(utils.DefaultKeyspaceID, 0)
+	if err != nil {
+		return err
+	}
+	config, err := m.getKeyspaceConfigByKindLocked(endpoint.Basic)
+	if err != nil {
+		return err
+	}
+	for _, ks := range keyspaces {
+		if ks == nil {
```
lhy1024, line 194: when will it be nil?

lhy1024 (approved):

pkg/keyspace/tso_keyspace_group.go
JmPotato: I still write the check to prevent panic, even though it won't be necessary.

pkg/keyspace/tso_keyspace_group.go
JmPotato: It will, so I removed this patrol to make it only run before the first split. PTAL.

rleungx (approved):

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +170,38 @@ func (m *GroupManager) Close() {
...
+			continue
+		}
+		groupID, err := strconv.ParseUint(config[TSOKeyspaceGroupIDKey], 10, 64)
+		if err != nil {
+			return err
+		}
+		err = m.updateKeyspaceForGroupLocked(endpoint.Basic, groupID, ks.GetId(), opAdd)
```
binshi-bing, line 196: How many existing keyspaces could we have? If we have too many keyspaces, it could take some time, as for every keyspace, we call saveKeyspaceGroups once. . If this is the case, can we just assign all keyspaces to the keyspace groups first, then call saveKeyspaceGroups just once to flush all 'dirty' keyspace groups?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +170,38 @@ func (m *GroupManager) Close() {
...
+		if err != nil {
+			return err
+		}
+	}
+	m.patrolKeyspaceAssignmentOnce = true
```
binshi-bing, line 201: Where do we update keyspace meta (add 'group' entry in the config field) and persistent its group assignment? Do we need to do it? There are two side effects if we don't persistent this info:
1. From the keyspace meta, you can't query its keyspace group directly.
2. After API leader switch, every split needs to go through this expensive process.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +170,38 @@ func (m *GroupManager) Close() {
...
+	keyspaces, err := m.store.LoadRangeKeyspace(utils.DefaultKeyspaceID, 0)
+	if err != nil {
+		return err
+	}
+	config, err := m.getKeyspaceConfigByKindLocked(endpoint.Basic)
+	if err != nil {
+		return err
+	}
+	for _, ks := range keyspaces {
+		if ks == nil {
```
binshi-bing, line 189: Do we need to check if ks.config map already contains the 'group' entry, i.e., already assigned keyspace group? If yes, then skip?

binshi-bing (commented):
left some comments. sorry for late review.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -156,6 +170,38 @@ func (m *GroupManager) Close() {
...
+			continue
+		}
+		groupID, err := strconv.ParseUint(config[TSOKeyspaceGroupIDKey], 10, 64)
+		if err != nil {
+			return err
+		}
+		err = m.updateKeyspaceForGroupLocked(endpoint.Basic, groupID, ks.GetId(), opAdd)
+		if err != nil {
+			return err
```
binshi-bing, line 198: SaveKeypsaceGroups transaction could fail and return here, which leaves none or some of keyspaces being assigned with keyspace groups.  Because patrolKeyspaceAssignmentOnce is still false, next split will trigger patrolKeyspaceAssignment again. For those keyspaces which have been assigned groups in the last round, because we don't persistent group assignment in keyspace meta and don't check if it's already assigned, we could pick a different keyspace group and assign this keyspace to it, which results in this keyspace being assigned to multiple groups.

binshi-bing (commented):
could have bugs.

pkg/keyspace/tso_keyspace_group.go
binshi-bing: Actually every tso restart then becoming leader could cause a keyspace being assigned to multiple keyspace groups.

PR #6390, opened 2023-04-26
etcdutil: add watch loop

### LINKED ISSUE #6391: etcdutil: add watch loop tool
## Development Task
There are some similar code with etcd watch loop:
1. https://github.com/tikv/pd/blob/400e3bd30c563bdb6b18ee7d59e56d082afff6a6/pkg/tso/keyspace_group_manager.go#L463
2. https://github.com/tikv/pd/blob/400e3bd30c563bdb6b18ee7d59e56d082afff6a6/pkg/keyspace/tso_keyspace_group.go#L299
3. https://github.com/tikv/pd/blob/400e3bd30c563bdb6b18ee7d59e56d082afff6a6/server/server.go#L1772
4. https://github.com/tikv/pd/blob/400e3bd30c563bdb6b18ee7d59e56d082afff6a6/server/keyspace_service.go#L72

lhy1024:

Issue Number: Close #6391

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/tso/keyspace_group_manager.go
```diff
@@ -305,218 +337,20 @@ func (kgm *KeyspaceGroupManager) Close() {
...
-					}
-					kgm.updateKeyspaceGroup(group)
-				case clientv3.EventTypeDelete:
-					kgm.deleteKeyspaceGroup(groupID)
-				}
-			}
-			// Retry the groups that are not initialized successfully before.
+		case <-ticker.C:
 			for id, group := range kgm.groupUpdateRetryList {
 				delete(kgm.groupUpdateRetryList, id)
```
binshi-bing, line 351: kgm.groupUpdateRetryList seems to be accessed by multiple goroutines now, and it needs to be sync.Map or to add lock.

binshi-bing: Previously, kgm.groupUpdateRetryList is only accessed by one goroutine.

pkg/tso/keyspace_group_manager.go
```diff
@@ -305,218 +334,29 @@ func (kgm *KeyspaceGroupManager) Close() {
...
-		revision = resp.Header.Revision + 1
-	}
-
-	return revision, kgs, resp.More, nil
-}
-
-// startKeyspaceGroupsMetaWatchLoop repeatedly watches any change in keyspace group membership/distribution
-// and apply the change dynamically.
-func (kgm *KeyspaceGroupManager) startKeyspaceGroupsMetaWatchLoop(revision int64) {
+func (kgm *KeyspaceGroupManager) startRetryUpdateLoop() {
```
binshi-bing, line 337: No need to add startRetryUpdateLoop(). Can we define postEventFn, similar to deleteFn and putFn, and pass it to watchLoop to execute after "for _, event := range wresp.Events {" loop and before revision = wresp.Header.Revision + 1 in LooperWatch.Watch? This can keep the generality of the watcher and also keep the exact same logic.

lhy1024: good idea

pkg/tso/keyspace_group_manager.go
```diff
@@ -305,218 +334,29 @@ func (kgm *KeyspaceGroupManager) Close() {
...
+				group, ok := value.(*endpoint.KeyspaceGroup)
+				if !ok {
+					return true
 				}
-			}
-			// Retry the groups that are not initialized successfully before.
-			for id, group := range kgm.groupUpdateRetryList {
-				delete(kgm.groupUpdateRetryList, id)
+				kgm.groupUpdateRetryList.Delete(id)
 				kgm.updateKeyspaceGroup(group)
```
binshi-bing, line 357: This is another reason to add postEventFn instead of using this retryUpdateLoop, because kgm.updateKeyspaceGroup(group) was assumed to have only one writer (goroutine) so the lock is finer which mightn't be able to accommodate this change.

pkg/tso/keyspace_group_manager.go
binshi-bing:
```
	postEventFn := func() {
	       // Retry the groups that are not initialized successfully before.
	       for id, group := range kgm.groupUpdateRetryList {
		        delete(kgm.groupUpdateRetryList, id)
		        kgm.updateKeyspaceGroup(group)
	        }
	}
```

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,227 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	putFn          func(*mvccpb.KeyValue) error
+	deleteFn       func(*mvccpb.KeyValue) error
+	opts           []clientv3.OpOption
+	loadTimeout    time.Duration
+	loadRetryTimes int
+	loadBatchSize  int64
+}
+
+// NewLoopWatcher creates a new LoopWatcher.
+func NewLoopWatcher(ctx context.Context, wg *sync.WaitGroup, client *clientv3.Client, name, key string, putFn, deleteFn func(*mvccpb.KeyValue) error, opts ...clientv3.OpOption) *LoopWatcher {
```
binshi-bing: line is too long

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,227 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+// LoopWatcher loads data from etcd and sets a watcher for it.
+type LoopWatcher struct {
+	ctx            context.Context
+	wg             *sync.WaitGroup
+	client         *clientv3.Client
+	key            string
+	name           string
+	forceLoadCh    chan struct{}
+	isLoadedCh     chan error
+	putFn          func(*mvccpb.KeyValue) error
```
binshi-bing, line 374: better add comment for putFn

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,227 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	deleteFn       func(*mvccpb.KeyValue) error
```
binshi-bing, line 375: better add comment for deleteFn

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,227 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+		}
+		for _, item := range resp.Kvs {
+			err = lw.putFn(item)
+			if err != nil {
+				log.Error("put failed in watch loop when loading", zap.String("name", lw.name), zap.String("key", lw.key), zap.Error(err))
+				continue
+			}
+		}
+		count := int64(len(resp.Kvs))
+		if count < limit {
```
binshi-bing: Can we use clientv3.GetResponse.More to judge if we have retrieved all the entries?

binshi-bing: The etcd comments only says "When passed WithLimit(limit), the number of returned keys is bounded by limit" but it doesn't mean it's done the job if count < limit here. According to another comment "// more indicates if there are more keys to return in the requested range." it seems that more is safer to use.

binshi-bing (commented):
The rest looks good to me

pkg/tso/keyspace_group_manager.go
```diff
@@ -257,20 +249,77 @@ func NewKeyspaceGroupManager(
...
+		groupID, err := endpoint.ExtractKeyspaceGroupIDFromPath(string(kv.Key))
+		if err != nil {
+			return err
+		}
+		kgm.deleteKeyspaceGroup(groupID)
+		return nil
+	}
+	postEventFn := func() error {
+		kgm.groupUpdateRetryList.Range(func(key, value interface{}) bool {
+			id, ok := key.(uint64)
```
binshi-bing, line 279: if we use postEventFn, then all accesses to groupUpdateRetryList is in the same goroutine. The sync.Map can changed back to map.

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,238 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	ctx            context.Context
+	wg             *sync.WaitGroup
+	client         *clientv3.Client
+	key            string
+	name           string
+	forceLoadCh    chan struct{}
+	isLoadedCh     chan error
+	putFn          func(*mvccpb.KeyValue) error
+	deleteFn       func(*mvccpb.KeyValue) error
+	postEventFn    func() error
```
binshi-bing, line 376: better add comment for postEventFn

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,238 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	return watchStartRevision
+}
+
+func (lw *LoopWatcher) watch(ctx context.Context, revision int64) (nextRevision int64, err error) {
+	watcher := clientv3.NewWatcher(lw.client)
+	defer watcher.Close()
+
+	for {
+	WatchChan:
+		opts := append(lw.opts, clientv3.WithRev(revision))
```
binshi-bing, line 477: Do we continuously append clientv3.WithRev(revision) to lw.opts? lw.opts will contain more and more clientv3.WithRev(revision), though eventually it will apply the least revision. Can we save the original slice before this for loop, then create a new opts by copying the original slice and append clientv3.WithRev(revision)?

binshi-bing (approved):
The rest looks good to me

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,240 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	for i := 0; i < lw.loadRetryTimes; i++ {
+		watchStartRevision, err = lw.load()
+		failpoint.Inject("loadTemporaryFail", func(val failpoint.Value) {
+			if maxFailTimes, ok := val.(int); ok && i < maxFailTimes {
+				err = errors.New("fail to read from etcd")
+				failpoint.Continue()
+			}
+		})
+		if err == nil {
+			if deadline, ok := ctx.Deadline(); ok && deadline.After(time.Now()) {
```
rleungx, line 453: What is it used for?

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,240 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+			revision = wresp.Header.Revision + 1
+		}
+	}
+}
+
+func (lw *LoopWatcher) load() (nextRevision int64, err error) {
+	startKey := lw.key
+	limit := lw.loadBatchSize
+	if limit <= 0 {
+		limit = defaultLoadBatchSize
```
rleungx, line 529: defaultLoadBatchSize is only needed in some specific scenarios?

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,240 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+
+	ctx, cancel := context.WithTimeout(lw.ctx, lw.loadTimeout)
+	defer cancel()
+	watchStartRevision := lw.initFromEtcd(ctx)
+
+	log.Info("start to watch loop", zap.String("name", lw.name), zap.String("key", lw.key))
+	for {
+		select {
+		case <-lw.ctx.Done():
+			close(lw.forceLoadCh)
```
rleungx, line 417: Do we need to close isLoadedCh?

pkg/utils/etcdutil/etcdutil.go
lhy1024: when test `delayLoad` to check whether timeout

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,240 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+			}
+		}
+		select {
+		case <-ctx.Done():
+			lw.isLoadedCh <- errors.Errorf("ctx is done before load data from etcd")
+			return watchStartRevision
+		case <-ticker.C:
+		}
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
lhy1024: it will send an error to `lw.isLoadedCh` and start watch loop

pkg/utils/etcdutil/etcdutil.go
lhy1024: when watch a lot of key

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,255 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	}
+}
+
+func (lw *LoopWatcher) load(ctx context.Context) (nextRevision int64, err error) {
+	startKey := lw.key
+	limit := lw.loadBatchSize
+	// If limit is 0, it means no limit.
+	// If limit is 1, we need to load 2 items to look for the next key.
+	if limit == 1 {
+		limit = 2
```
binshi-bing, line 544: I just noticed that we need to generally handle the start key of range scan in pagination. The boundary seems to have problem because we mostly issue put event for the last entry in the return set twice, and we actually only retrieve limit - 1 entries from the second Get Request because the first one is duplicated.

One solution might be generally let limit = limit + 1 when limit isn't zero, the for the last n - 1 batch, issue put event for the first (len - 1) entries, and issue put event for all entries for the last batch.

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,255 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+	}
+	ctx, cancel := context.WithTimeout(ctx, DefaultRequestTimeout)
+	defer cancel()
+	for {
+		opts := append(lw.opts, clientv3.WithSort(clientv3.SortByKey, clientv3.SortAscend), clientv3.WithLimit(limit))
```
binshi-bing, line 549: Just realize that using range scans with a limit, the order of the keys returned is not guaranteed, so Sort is required for pagination, though it has perf issue if the key range is too large as mentioned in https://github.com/etcd-io/etcd/issues/11001. We're good for now. No need further optimization for it.

pkg/utils/etcdutil/etcdutil.go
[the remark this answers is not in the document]
rleungx: What I mean is will etcd return an error when the key doesn't exist?

pkg/utils/etcdutil/etcdutil.go
rleungx: Then it's just for test?

pkg/utils/etcdutil/etcdutil.go
lhy1024: we don't need to worry about the performance, because the default sort is just SortByKey and SortAscend

pkg/utils/etcdutil/etcdutil.go
lhy1024: changed and I will add more test to check

pkg/utils/etcdutil/etcdutil.go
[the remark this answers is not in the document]
lhy1024: I will add more test to check

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,258 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+			return 0, err
+		}
+		for i, item := range resp.Kvs {
+			if resp.More && i == len(resp.Kvs)-1 {
+				// The last key is the start key of the next batch.
+				// To avoid to get the same key in the next load, we need to skip the last key.
+				startKey = string(item.Key)
+				continue
+			}
+			err = lw.putFn(item)
```
binshi-bing, line 568: Do we need to skip the first entry from the second page?

binshi-bing: no need. please ignore.

pkg/utils/etcdutil/etcdutil.go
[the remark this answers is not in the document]
lhy1024: etcd will not return an error when the key doesn't exist recording `TestLoadWithoutKey`

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -350,3 +353,261 @@ func InitOrGetClusterID(c *clientv3.Client, key string) (uint64, error) {
...
+		}
+		select {
+		case <-ctx.Done():
+			lw.isLoadedCh <- errors.Errorf("ctx is done before load data from etcd")
+			return watchStartRevision
+		case <-ticker.C:
+		}
+	}
+	if err != nil {
+		log.Warn("watched key doesn't exist in watch loop when loading", zap.String("name", lw.name), zap.String("key", lw.key), zap.Error(err))
```
rleungx, line 484: Then, the log is wrong.

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -433,6 +437,9 @@ func (lw *LoopWatcher) StartWatchLoop() {
 		case <-lw.ctx.Done():
 			log.Info("server is closed, exit watch loop", zap.String("name", lw.name), zap.String("key", lw.key))
 			return
+		case client := <-lw.updateClientCh:
+			log.Info("update etcd client", zap.String("name", lw.name), zap.String("key", lw.key))
+			lw.client = client
```
rleungx, line 442: Suggest not adding the test logic here.

rleungx (approved):
Overall, LGTM

rleungx (approved):
Overall, LGTM

rleungx (approved):
Overall, LGTM

rleungx (approved):
Overall, LGTM

PR #6393, opened 2023-04-27
client: metircs with const label

### LINKED ISSUE #6394: pd client metrics should be able to configure const label
## Enhancement Task
pd client metrics should be able to configure const label

zeminzhou:
pd client metrics should be able to configure const label

Issue Number: Close #6394 

```
pd client metrics should be able to configure const label
```

### Check List

Tests

- Unit test
- Integration test
1. updae tidb's pd client, and add `WithMetricsLabel` when create pd client
2. make tidb
3. run tidb cluster
4. curl http://127.0.0.1:10080/metrics
image

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/metrics.go
```diff
@@ -17,111 +17,180 @@ package pd
...
-	cmdFailedDurationPut                      = cmdFailedDuration.WithLabelValues("put")
-)
+	cmdFailedDurationLoadKeyspace = cmdDuration.WithLabelValues("load_keyspace")
+	cmdFailedDurationUpdateKeyspaceState = cmdDuration.WithLabelValues("update_keyspace_state")
+	requestDurationTSO = requestDuration.WithLabelValues("tso")
+	cmdFailedDurationGet = cmdFailedDuration.WithLabelValues("get")
+	cmdFailedDurationPut = cmdFailedDuration.WithLabelValues("put")
+}
 
-func init() {
```
ystaticy, line 124: Where do we call this function?

client/metrics.go
```diff
@@ -17,111 +17,180 @@ package pd
...
+// RegisterMetrics registers all metrics for PD client.
+func RegisterMetrics() {
```
ystaticy, line 193: Where do we call this function?

disksing (approved):

client/metrics.go
disksing: from tidb, I guess

zeminzhou: yesh, the specific location should be [here](https://github.com/tidbcloud/tidb-cse/blob/keyspace-on-release-6.6/metrics/metrics.go#L294-L296) like client-go

zeminzhou: yesh, the specific location should be [here](https://github.com/tidbcloud/tidb-cse/blob/keyspace-on-release-6.6/metrics/metrics.go#L294-L296) like client-go

JmPotato: The behavior seems to have changed since you modified the `init()` function. Will this cause any issues for those who do not wish to register explicitly?

zeminzhou: No, if users want to register `pd_client` metrics, these should call `Register` explicitly.

JmPotato: My point is that you have changed the default behavior of initializing metrics. For some libraries that are already using it, upgrading may cause some problems because they may not be aware of this change.

JmPotato (approved):

binshi-bing (approved):

rleungx:
Please sign-off the DCO.

PR #6397, opened 2023-04-28
keyspace, api2: fix the keyspace assignment patrol consistency

JmPotato:

Issue Number: ref #6232.

```commit-message
Fix the keyspace assignment patrol consistency.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -589,7 +589,15 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 		if err != nil {
 			return err
 		}
-		assigned := false
+		var (
+			assigned            = false
+			keyspaceIDsToUnlock = make([]uint32, 0, len(keyspaces))
+		)
+		defer func() {
+			for _, id := range keyspaceIDsToUnlock {
+				manager.metaLock.Unlock(id)
+			}
+		}()
 		for _, ks := range keyspaces {
 			if ks == nil {
 				continue
@@ -606,7 +614,7 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 				}
 			}
 			// Unlock the keyspace meta lock after the whole txn.
-			defer manager.metaLock.Unlock(ks.Id)
+			keyspaceIDsToUnlock = append(keyspaceIDsToUnlock, ks.Id)
 			// If the keyspace doesn't have a group ID, assign it to the default keyspace group.
 			if !slice.Contains(defaultKeyspaceGroup.Keyspaces, ks.Id) {
 				defaultKeyspaceGroup.Keyspaces = append(defaultKeyspaceGroup.Keyspaces, ks.Id)
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -574,6 +574,7 @@ func (manager *Manager) allocID() (uint32, error) {
 
 // PatrolKeyspaceAssignment is used to patrol all keyspaces and assign them to the keyspace groups.
 func (manager *Manager) PatrolKeyspaceAssignment() error {
+	// TODO: since the number of keyspaces might be large, we should consider to assign them in batches.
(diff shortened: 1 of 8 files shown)


pkg/keyspace/keyspace.go
```diff
@@ -555,3 +571,58 @@ func (manager *Manager) allocID() (uint32, error) {
 	}
 	return id32, nil
 }
+
+// PatrolKeyspaceAssignment is used to patrol all keyspaces and assign them to the keyspace groups.
+func (manager *Manager) PatrolKeyspaceAssignment() error {
+	return manager.store.RunInTxn(manager.ctx, func(txn kv.Txn) error {
```
binshi-bing, line 577: The only worry is that the transaction could be too large to accepted by etcd if we have too many keyspaces. Do we need to do it in batches?

binshi-bing: It looks good for me for now. We can revisit the comment above after we collect some rough data.

binshi-bing (approved):
LGTM for now

pkg/keyspace/keyspace.go
```diff
@@ -555,3 +571,58 @@ func (manager *Manager) allocID() (uint32, error) {
...
+			} else {
+				// If the keyspace already has a group ID, skip it.
+				_, ok := ks.Config[TSOKeyspaceGroupIDKey]
+				if ok {
+					manager.metaLock.Unlock(ks.Id)
+					continue
+				}
+			}
+			// Unlock the keyspace meta lock after the whole txn.
+			defer manager.metaLock.Unlock(ks.Id)
```
rleungx, line 609: Will it have many defer?

JmPotato: I changed it to a way that only requires one defer. PTAL.

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -164,8 +170,21 @@ func SplitKeyspaceGroupByID(c *gin.Context) {
...
 
 	svr := c.MustGet(middlewares.ServerContextKey).(*server.Server)
-	manager := svr.GetKeyspaceGroupManager()
-	err = manager.SplitKeyspaceGroupByID(id, splitParams.NewID, splitParams.Keyspaces)
+	patrolKeyspaceAssignmentState.Lock()
+	if !patrolKeyspaceAssignmentState.patrolled {
+		// Patrol keyspace assignment before splitting keyspace group.
+		manager := svr.GetKeyspaceManager()
+		err = manager.PatrolKeyspaceAssignment()
+		if err != nil {
```
lhy1024, line 178: if meet err returned, the lock will hang?

pkg/keyspace/keyspace.go
```diff
@@ -555,3 +571,66 @@ func (manager *Manager) allocID() (uint32, error) {
 	}
 	return id32, nil
 }
+
+// PatrolKeyspaceAssignment is used to patrol all keyspaces and assign them to the keyspace groups.
+func (manager *Manager) PatrolKeyspaceAssignment() error {
+	return manager.store.RunInTxn(manager.ctx, func(txn kv.Txn) error {
```
lhy1024, line 577: do we need to add txn in other operators?

server/apiv2/handlers/tso_keyspace_group.go
JmPotato: Nice catch. Fixed.

pkg/keyspace/keyspace.go
JmPotato: What do you mean?

pkg/keyspace/keyspace.go
binshi-bing: Based on the recent discussion, it looks like we need to process it in batch now.

rleungx (approved):

PR #6402, opened 2023-05-04
client: retry load keyspace meta when creating a new client

rleungx:

Issue Number: Ref #5895.

Adding retry for getting the keyspace meta in the client.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -392,7 +392,7 @@ func NewClientWithKeyspaceName(ctx context.Context, keyspace string, svrAddrs []
 func (c *client) initRetry(f func(s string) error, str string) error {
 	var err error
 	for i := 0; i < c.option.maxRetryTimes; i++ {
-		if err = f(str); err == nil {
+		if err = f(str); err == nil || strings.Contains(err.Error(), "ENTRY_NOT_FOUND") {
 			return nil
 		}
 		select {
--- a/client/client.go
+++ b/client/client.go
@@ -383,13 +383,35 @@ func NewClientWithKeyspaceName(ctx context.Context, keyspace string, svrAddrs []
 		c.cancel()
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -383,13 +383,35 @@ func NewClientWithKeyspaceName(ctx context.Context, keyspace string, svrAddrs []
...
+	if err := c.initRetry(c.loadKeyspaceMeta, keyspace); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (c *client) initRetry(f func(s string) error, str string) error {
+	var err error
+	for i := 0; i < c.option.maxRetryTimes; i++ {
+		if err = f(str); err == nil {
```
binshi-bing, line 395:
```suggestion
		if err = f(str); err == nil || strings.Contains(err.Error(), "ENTRY_NOT_FOUND") {
```

if the error contains ENTRY_NOT_FOUND, it seems that we need to immediately return success instead of retrying 30 times and each time 1 second by default. It is perfectly ok to me, if we change to function to something like loadKeyspaceMetaWithRetry which is used only for this specific LoadKeyspace rpc.

binshi-bing (commented):
left one comment

client/client.go
binshi-bing: When I look back to the original decision "ere we ignore ENTRY_NOT_FOUND error and it will set the keyspaceID to 0", I'm wondering if this is true. If the caller of PD client uses a keyspace name which can't be found in keyspace meta, could it be a human mistake or system bug somewhere? How to handle"ENTRY_NOT_FOUND" could depends on context which PD client doesn't know.

binshi-bing (approved):
LGTM

nolouch (approved):

PR #6412, opened 2023-05-06
*: fix disable failpoint 

rleungx:

Issue Number: Ref #4399.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/dashboard/adapter/manager.go
+++ b/pkg/dashboard/adapter/manager.go
@@ -105,7 +105,7 @@ func (m *Manager) updateInfo() {
 	m.isLeader = true
 
 	var err error
-	if m.members, err = cluster.GetMembers(m.srv.GetClient()); err != nil {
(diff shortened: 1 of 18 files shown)


lhy1024 (approved):

nolouch (commented):
Do you test it, is it improved?

binshi-bing (approved):
Where is the description of the issue? If there is description, shall we add one? 
Does it aim to fix the pd leader issue mentioned in on-call?

rleungx:
> Do you test it, is it improved?

No, haven't tested yet. I will update the result once I finished it. But theoretically, we need to pass the context in this way.

rleungx:
> Do you test it, is it improved?

Yes, but it seems there is no much difference for this case since TiKV doesn't use context.

PR #6413, opened 2023-05-08
minResolvedTS: Provide an API to fetch the store resolved_ts

### LINKED ISSUE #6386: Provide an API to fetch the store resolved_ts
## Development Task

TiDB is currently unable to advance the `min_resolved_ts` when any TiKV node is down, as reported in https://github.com/pingcap/tidb/issues/43386. However, PD can handle this situation by utilizing more detailed information such as leader count to exclude problematic TIKV nodes. To ensure consistent behavior across the system, an API must be provided for fetching store `resolved_ts`, which will replace the current implementation in `client-go`.

HuSharp:

Issue Number: Close #6386

```commit-message
Provide an API to fetch the store resolved_ts
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -2511,6 +2511,16 @@ func (c *RaftCluster) GetMinResolvedTS() uint64 {
 	return c.minResolvedTS
 }
 
+// GetStoreMinResolvedTS returns the min resolved ts of the store.
+func (c *RaftCluster) GetStoreMinResolvedTS(storeID uint64) uint64 {
+	c.RLock()
+	defer c.RUnlock()
+	if !c.isInitialized() {
+		return math.MaxUint64
+	}
+	return c.GetStore(storeID).GetMinResolvedTS()
+}
+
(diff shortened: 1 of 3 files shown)


server/cluster/cluster.go
```diff
@@ -2511,6 +2511,16 @@ func (c *RaftCluster) GetMinResolvedTS() uint64 {
 	return c.minResolvedTS
 }
 
+// GetStoreMinResolvedTS returns the min resolved ts of the store.
+func (c *RaftCluster) GetStoreMinResolvedTS(storeID uint64) uint64 {
+	c.RLock()
+	defer c.RUnlock()
+	if !c.isInitialized() {
```
JmPotato, line 2518: Do we also need to check `IsAvailableForMinResolvedTS` here?

HuSharp: added.

lhy1024 (approved):

JmPotato (approved):

PR #6418, opened 2023-05-08
*: move out replication and syncer

rleungx:

Issue Number: Ref #5839.

This PR mainly did the following changes:

- move replication mode to schedule
- move unsafe recovery to pkg
- move coordinator and prepare checker to schedule
- put operator related logic together
- put scheduler related logic together
- use opeartor.ClusterInfo to replace schedule.Cluster in most of the code
- move the corresponding metrics to the right place
- rename region_syncer to regionsyncer
- make all existing tests pass by using mockcluster in some places

Once we reach an agreement, I will try to split this PR into pieces.

**_Most of the jobs are done in the previous PRs. This PR will only move some packages out of server._**

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


tests/server/region_syncer/region_syncer_test.go
```diff
@@ -151,7 +151,7 @@ func TestRegionSyncer(t *testing.T) {
 		re.Equal(region.GetLeader(), r.GetLeader())
 		re.Equal(region.GetBuckets(), r.GetBuckets())
 	}
-	re.NoError(failpoint.Disable("github.com/tikv/pd/server/syncer/noFastExitSync"))
```
CabinfeverB, line 154: This issue may not be related to this PR change. The old path `github.com/tikv/pd/server/syncer/noFastExitSync` seems to be wrong? Should be `regionsyncer` other than `syncer`. So does this failpoint work?

CabinfeverB (approved):

tests/server/region_syncer/region_syncer_test.go
rleungx: The package name is `syncer` I think.

CabinfeverB: Get

lhy1024 (approved):

Issue #6142, opened 2023-03-10
Compatibility issue about multi Group TSO


nolouch:
# Compatibility issue about Group TSO
ref https://github.com/tikv/pd/issues/5895
### Problem
Async Commit use max_ts to commit. The problem is, if multi group tso is used with async commit/1pc enabled, the `commit_ts `calculation becomes complicated. We can't record only one `max_ts`any more. For example, in this case:

- Read k1 from TSO Group1, start_ts (allocated from local allocator 1) = 100, then the max_ts is updated to 100;
- Write k2 from TSO Group2, start_ts (allocated from local allocator 2) = 50, calculated commit_ts = 101 (max_ts + 1)
Commit k2 from TSO Group2
- Read k2 from TSO Group1, start_ts (allocated from local allocator 2) = 51, then the previous transaction's result is invisible.

## Resolve TS
the resolve ts formula:
```
min_ts = pd.tso()
new_resolved_ts = min(min_ts, all(lock.start_ts))
resolved_ts = min(new_resolved_ts, resolved_ts)
```

### Problem 

If there is no locks, it maybe wrong update the resolved ts, fix with:

```
min_ts = min(all group.tso())
```

ystaticy:
In keyspace scenario, there is a globally unique GC Worker to calculate the global GC safepoint.
The calculation of GC safepoint depends on three values:
1. min(all keyspace service safepoint)
2. `now`-10min
3. min(all keyspace. minstartts)
`now` in the second item above, it's always get from a unique timeline. 
`now, err := w.getOracleTime()`

But if TSO in differrent group, multiple timelines will be created.
So there is  possible options:
If TSO is in differrent group, `now` should take the minimum value of all TSO groups. It need PD provide a new interface to return `min(all TSO group nowTS)`

Subsequent plan: the calculation of GC safepoint will be separated by keyspace,but the design discussion and code review may take more time.

ystaticy:
There is another problem:
In keyspace scenario, if we use the globally unique gc safepoint.
now, gc safepoint = ts1;
We need to constrain the new tso group timeline to be later than ts1.
Otherwise, the data in the new tso group will be directly gc.

nolouch:
cc @rleungx @binshi-bing

rleungx:
native_br also needs to get the global min ts.

PR #6421, opened 2023-05-09
Provide GetMinTS API to solve the compatibility issue brought by multi-timeline tso

binshi-bing:

Issue Number: Ref #6142 

Add GetMinTS API to PD(TSO) client to get the minimal timestamp across all keyspace groups, which follows the steps below:
1. The PD(TSO) client refreshes the TSO server list first.
2. The PD(TSO) client calls GetMinTS PRC for every TSO server.
3. On every TSO server, it receives GetMinTS request, then walk through all the keyspace groups, get timestamp from every keyspace group primary, compare and get the minimal timestamp, and returns:
    a. the minimal timestamp from all keyspace groups, as primaries, served by this TSO server.
    b. The count of keyspace groups as primaries.
    c. The total count of keyspace groups of the entire TSO service.
4. The PD(TSO) client collects the results from all TSO servers, do sanity check and returns the minimal timestamp across all keyspace groups of the entire TSO service or returns error.

We also discussed two more options for further optimization which might be overkilled for now. The two more options we discussed are:
1. Read all saved timestamps of all keyspace groups from etcd. Select the minimal ts then substract a timewindow. The cons is that it could cause stuck in availability when loading all saved timestamps in a transaction. If QPS is higher than frequency of lead start/restart, it could impact availability.
2. Based on the current approach, but every keyspace group cache the last ts. If the update time of the last ts, for a keyspace group, is out of a time window, it needs to query the timestamp to pull forward the timeline. It might be overkill for now, because we won't have too many timelines from the beginning and the QPS of GetMinTS is very low (few times per day?).

```commit-message
1. Import kvproto change to introduce GetMinTS rpc in the TSO service.
6. Add server side implementation for GetMinTS rpc.
7. Add client side implementation for GetMinTS rpc.
8. Add unit test

```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/go.mod
```diff
@@ -39,3 +39,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/pingcap/kvproto => github.com/binshi-bing/kvproto v0.0.0-20230510010110-735e184fe7b2
```
lhy1024, line 43: update it, plz

go.mod
```diff
@@ -204,3 +204,5 @@ replace google.golang.org/grpc v1.54.0 => google.golang.org/grpc v1.26.0
 // kvproto at the same time. You can run `go mod tidy` to make it replaced with go-mod style specification.
 // After the PR to kvproto is merged, remember to comment this out and run `go mod tidy`.
 // replace github.com/pingcap/kvproto => github.com/$YourPrivateRepo $YourPrivateBranch
+
+replace github.com/pingcap/kvproto => github.com/binshi-bing/kvproto v0.0.0-20230510010110-735e184fe7b2
```
lhy1024, line 208: ditto

client/client.go
```diff
@@ -259,20 +259,20 @@ type serviceModeKeeper struct {
 	// triggering service mode switching concurrently.
 	sync.RWMutex
 	serviceMode     pdpb.ServiceMode
-	tsoClient       atomic.Value // *tsoClient
+	tsoClient       *tsoClient
```
lhy1024, line 262: why remove atomic?

client/client.go
```diff
@@ -714,6 +713,26 @@ func (c *client) GetLocalTS(ctx context.Context, dcLocation string) (physical in
...
+	if tsoClient == nil {
+		return 0, 0, errs.ErrClientGetMinTSO.FastGenByArgs("tso client is nil")
+	}
+
+	switch serviceMode {
+	case pdpb.ServiceMode_UNKNOWN_SVC_MODE:
+		return 0, 0, errs.ErrClientGetMinTSO.FastGenByArgs("unknown service mode")
+	case pdpb.ServiceMode_PD_SVC_MODE:
+		// If the service mode is switched to API during GetTS() call, which happens during migration,
+		// returning the default timeline should be fine.
```
lhy1024, line 727: Do we need to consider local ts?

client/tso_client.go
```diff
@@ -275,3 +281,114 @@ func (c *tsoClient) backupClientConn() (*grpc.ClientConn, string) {
...
+
+	// Get the minimal timestamp from the TSO servers/pods
+	var mutex sync.Mutex
+	resps := make([]*tsopb.GetMinTSResponse, 0)
+	wg := sync.WaitGroup{}
+	wg.Add(len(addrs))
+	for _, addr := range addrs {
+		go func(addr string) {
+			defer wg.Done()
+			resp, err := c.getMinTSFromSingleServer(ctx, addr, c.option.timeout)
```
lhy1024, line 301: if there is no primary in a tso server, do we need to wait to result for it?

pkg/mcs/tso/server/grpc_service.go
```diff
@@ -157,6 +158,18 @@ func (s *Service) Tso(stream tsopb.TSO_TsoServer) error {
...
+		return &tsopb.FindGroupByKeyspaceIDResponse{
+			Header: s.wrapErrorToHeader(tsopb.ErrorType_NOT_BOOTSTRAPPED, ErrNotStarted.Error(), 0),
+		}, nil
+	}
+
+	if request.GetHeader().GetClusterId() != s.clusterID {
+		return &tsopb.FindGroupByKeyspaceIDResponse{
+			Header: s.wrapErrorToHeader(tsopb.ErrorType_CLUSTER_MISMATCHED, ErrClusterMismatched.Error(), 0),
+		}, nil
+	}
```
JmPotato, line 171: What about extracting a function?

pkg/mcs/tso/server/grpc_service.go
```diff
@@ -199,6 +212,45 @@ func (s *Service) FindGroupByKeyspaceID(
...
+			Header: s.wrapErrorToHeader(
+				tsopb.ErrorType_NOT_BOOTSTRAPPED, ErrNotStarted.Error(), respKeyspaceGroup),
+		}, nil
+	}
+
+	if request.GetHeader().GetClusterId() != s.clusterID {
+		return &tsopb.GetMinTSResponse{
+			Header: s.wrapErrorToHeader(
+				tsopb.ErrorType_CLUSTER_MISMATCHED, ErrClusterMismatched.Error(), respKeyspaceGroup),
+		}, nil
```
JmPotato, line 232: Ditto.

client/client.go
binshi-bing: We have already used sync.RWMutex as shown by line 259. atomic.Value isn't necessary and should be removed last time.

client/tsoutil/tsoutil.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/pingcap/kvproto/pkg/pdpb"
+)
+
+// Because of the suffix, we need to shift the count before we add it to the logical part.
+func AddLogical(logical, count int64, suffixBits uint32) int64 {
+	return logical + count<<suffixBits
+}
+
+func TSLessEqual(physical, logical, thatPhysical, thatLogical int64) bool {
```
JmPotato, line 26:
```suggestion
// TSLessEqual ...
func TSLessEqual(physical, logical, thatPhysical, thatLogical int64) bool {
```

tests/integrations/tso/client_test.go
```diff
@@ -197,6 +197,37 @@ func (suite *tsoClientTestSuite) TestGetTSAsync() {
 	wg.Wait()
 }
 
+func (suite *tsoClientTestSuite) TestGetMinTS() {
```
JmPotato, line 200: Is it possible to write a test to ensure that the min ts obtained must be the smallest of all keyspace groups rather than simply comparing with the previous min ts?

binshi-bing: it's covered in this test case.

binshi-bing:
```
					// Now we check whether the returned ts is the minimum one
					// among all keyspace groups, i.e., the returned ts is
					// less than the new timestamps of all keyspace groups.
					for _, client := range suite.clients {
						physical, logical, err := client.GetTS(suite.ctx)
						suite.NoError(err)
						ts := tsoutil.ComposeTS(physical, logical)
						suite.Less(minTS, ts)
					}
```

JmPotato: Got it, my bad.

client/go.mod
binshi-bing: updated to kvproto master

go.mod
binshi-bing: updated to kvproto master

client/client.go
binshi-bing: The GetTS has passed DCLocation to GetMinTS rpc to cover local ts. It keeps back compatibility with PD, but TSO microservice doesn't support local ts.

client/tso_client.go
binshi-bing: Not really, but it does need to wait all keyspace groups in serving status. It needs to meet the following conditions.
	// Check the results. The returned minimal timestamp is valid if all the conditions are met:
	// 1. The number of responses is equal to the number of TSO servers/pods.
	// 2. The number of keyspace groups asked is equal to the number of TSO servers/pods.
	// 3. The minimal timestamp is not zero.

binshi-bing: We have discussed two more options for further optimization which might be overkilled for now. The two more options we discussed are:
1. Read all saved timestamps of all keyspace groups from etcd. Select the minimal ts then substract a timewindow. The cons is that it could cause stuck in availability when loading all saved timestamps in a transaction. If QPS is higher than frequency of lead start/restart, it could impact availability.
2. Based on the current approach, but every keyspace group cache the last ts. If the update time of the last ts, for a keyspace group, is out of a time window, it needs to query the timestamp to pull forward the timeline. It might be overkill for now.

pkg/mcs/tso/server/grpc_service.go
binshi-bing: done.

pkg/mcs/tso/server/grpc_service.go
binshi-bing: done

client/client.go
```diff
@@ -266,20 +266,20 @@ type serviceModeKeeper struct {
...
+	tsoClient       *tsoClient
 	tsoSvcDiscovery ServiceDiscovery
 }
 
-func (smk *serviceModeKeeper) close() {
-	smk.Lock()
-	defer smk.Unlock()
-	switch smk.serviceMode {
+func (k *serviceModeKeeper) close() {
+	k.Lock()
```
rleungx, line 274: Rlock is enough?

binshi-bing: It should be safer to use lock, shouldn't it? If we use RLock, even after the tsoServiceDiscovery and tsoClient have started the Close(), it will still allow access to tsoServiceDiscovery and tsoClient internal unnecessarily.

rleungx: ok

client/tsoutil/tsoutil.go
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+package tsoutil
+
+import (
+	"github.com/pingcap/kvproto/pkg/pdpb"
+)
+
+// Because of the suffix, we need to shift the count before we add it to the logical part.
+func AddLogical(logical, count int64, suffixBits uint32) int64 {
```
JmPotato, line 22:
```suggestion
// AddLogical shifts the count before we add it to the logical part.
func AddLogical(logical, count int64, suffixBits uint32) int64 {
```

JmPotato (commented):
The rest LGTM.

client/tsoutil/tsoutil.go
binshi-bing: changed

rleungx (approved):

Issue #6426, opened 2023-05-09
Lightning block at not full replicaed


nolouch:
### Problem

In concurrent import, we may have replicas that do not satisfy the replicas constraints, such as:
![image](https://github.com/tikv/pd/assets/6428910/53d8790a-ac06-4fa0-888a-193f68f25f62)

the lightning split the region during the scatter, it may cause abnormal regions. We can see the logs in pd like, it has 6 replicas :
```
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.491 +00:00] [INFO] [cluster_worker.go:145] ["alloc ids for region split"] [region-id=1902122] [peer-ids="[1902123,1902124,1902125,1902126,1902127,1902128]"] 
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.502 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.514 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.536 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.579 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.661 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:54.822 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
/var/lib/pd/log/pd-2023-05-09T06-58-33.433.log:[2023/05/09 05:41:55.144 +00:00] [WARN] [region_scatterer.go:285] ["region not replicated during scatter"] [region-id=1902122]
[...]

PR #6427, opened 2023-05-10
checker: fix unhealth region skip the rule check

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/6426

```commit-message
allow the `schedule=deny` label can do rule constraints check
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/checker_controller.go
+++ b/pkg/schedule/checker/checker_controller.go
@@ -33,6 +33,8 @@ import (
 // DefaultCacheSize is the default length of waiting list.
 const DefaultCacheSize = 1000
 
+var denyCheckersByLabelerCounter = schedule.LabelerEventCounter.WithLabelValues("checkers", "deny")
+
 // Controller is used to manage all checkers.
 type Controller struct {
 	cluster           schedule.Cluster
@@ -122,6 +124,7 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
 	if cl, ok := c.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
 		l := cl.GetRegionLabeler()
 		if l.ScheduleDisabled(region) {
+			denyCheckersByLabelerCounter.Inc()
 			return nil
 		}
 	}
--- a/pkg/schedule/checker/checker_controller.go
+++ b/pkg/schedule/checker/checker_controller.go
@@ -120,7 +120,8 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
 			c.regionWaitingList.Put(region.GetID(), nil)
 		}
 	}
-	// only consider merge checker when region label is set to "schedule=deny".
+	// skip the joint checker, split checker and rule checker when region label is set to "schedule=deny".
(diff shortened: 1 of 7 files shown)


server/cluster/cluster.go
```diff
@@ -70,6 +70,7 @@ var (
 	// WithLabelValues is a heavy operation, define variable to avoid call it every time.
 	regionUpdateCacheEventCounter = regionEventCounter.WithLabelValues("update_cache")
 	regionUpdateKVEventCounter    = regionEventCounter.WithLabelValues("update_kv")
+	regionSkipSchedulerCounter    = regionEventCounter.WithLabelValues("region_skip_scheduler")
```
rleungx, line 73: How about region_schedule_disabled?

pkg/schedule/operator_controller.go
```diff
@@ -424,14 +423,6 @@ func (oc *OperatorController) checkAddOperator(isPromoting bool, ops ...*operato
 		if op.SchedulerKind() == operator.OpAdmin || op.IsLeaveJointStateOperator() {
 			continue
 		}
-		if cl, ok := oc.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
```
rleungx, line 427: It seems we need to use the way of pausing schedulers to prevent the range from being scheduled.

nolouch: I add the tests, prevent it in `coordinator`.

rleungx (approved):

rleungx (approved):

pkg/schedule/checker/checker_controller.go
```diff
@@ -125,6 +120,14 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
 			c.regionWaitingList.Put(region.GetID(), nil)
 		}
 	}
+	// only consider merge checker when region label is set to "schedule=deny".
+	if cl, ok := c.cluster.(interface{ GetRegionLabeler() *labeler.RegionLabeler }); ok {
```
bufferflies, line 124: maybe putting it into code line 137  be better?

nolouch: This is not just for merge, but only merge is left. I change the comments.

server/cluster/coordinator.go
```diff
@@ -901,7 +901,17 @@ func (s *scheduleController) Schedule(diagnosable bool) []*operator.Operator {
...
+		for _, op := range ops {
+			if labelMgr := s.cluster.GetRegionLabeler(); labelMgr != nil {
+				if labelMgr.ScheduleDisabled(s.cluster.GetRegion(op.RegionID())) {
+					denySchedulersByLabelerCounter.Inc()
+					foundDisabled = true
+					break
+				}
+			}
+		}
+		if len(ops) > 0 && !foundDisabled {
```
bufferflies, line 914: In the origin logical code, the next interval should be min interval even if the scheduled region is disabled scheduled.

bufferflies (approved):
rest LGTM

PR #6445, opened 2023-05-11
etcdutil: fix ctx in watch loop

### LINKED ISSUE #6439: PD OOM when doing HA tests
### What did you do?

Run HA test plan

All test pass

### What did you see instead?

Screenshot 2023-05-11 at 14 30 15
Screenshot 2023-05-11 at 14 32 27
Screenshot 2023-05-11 at 14 34 33

pd-cse

lhy1024:

Issue Number: Close #6439

```commit-message
```

Before this pr, the goroutine of watch chan will exist until the watcher is closed, which means it will leak in for-loop.

We need to cancel ctx when watch chan is not needed.

And I found that `withRequireLeader` is necessary for `watch`, so I also added it.

### Check List

Tests

- Unit test
- Integration test
memory and goroutine is enough stable
image

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/etcdutil/etcdutil.go
+++ b/pkg/utils/etcdutil/etcdutil.go
@@ -461,10 +461,10 @@ func (lw *LoopWatcher) initFromEtcd(ctx context.Context) int64 {
 		watchStartRevision int64
 		err                error
 	)
-	ctx, cancel := context.WithTimeout(ctx, lw.loadTimeout)
-	defer cancel()
 	ticker := time.NewTicker(defaultLoadFromEtcdRetryInterval)
 	defer ticker.Stop()
+	ctx, cancel := context.WithTimeout(ctx, lw.loadTimeout)
+	defer cancel()
 
 	for i := 0; i < lw.loadRetryTimes; i++ {
 		failpoint.Inject("loadTemporaryFail", func(val failpoint.Value) {
--- a/pkg/utils/etcdutil/etcdutil.go
+++ b/pkg/utils/etcdutil/etcdutil.go
@@ -427,19 +427,19 @@ func (lw *LoopWatcher) StartWatchLoop() {
 	defer logutil.LogPanic()
 	defer lw.wg.Done()
 
-	ctx, cancel := context.WithTimeout(lw.ctx, lw.loadTimeout)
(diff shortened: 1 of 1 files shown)


binshi-bing (approved):
@rleungx @lhy1024 should add more details about the feature. It will be even better with short analysis and solution explanation.

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -461,6 +461,8 @@ func (lw *LoopWatcher) initFromEtcd(ctx context.Context) int64 {
 		watchStartRevision int64
 		err                error
 	)
+	ctx, cancel := context.WithTimeout(ctx, lw.loadTimeout)
+	defer cancel()
 	ticker := time.NewTicker(defaultLoadFromEtcdRetryInterval)
 	defer ticker.Stop()
```
JmPotato, line 467: Since it has a timeout, we better initialize it later.

```suggestion
	ticker := time.NewTicker(defaultLoadFromEtcdRetryInterval)
	defer ticker.Stop()
	ctx, cancel := context.WithTimeout(ctx, lw.loadTimeout)
	defer cancel()
```

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -532,11 +542,16 @@ func (lw *LoopWatcher) watch(ctx context.Context, revision int64) (nextRevision
 						log.Error("put failed in watch loop", zap.String("name", lw.name),
 							zap.String("key", lw.key), zap.Error(err))
 					}
+					log.Debug("put in watch loop", zap.String("name", lw.name),
+						zap.ByteString("key", event.Kv.Key),
```
rleungx, line 546:
```suggestion
						zap.String("key", string(event.Kv.Key)),
```

lhy1024: Why? What is the difference between them?

rleungx: Oh, they are the same.

rleungx (approved):

rleungx (approved):

lhy1024:
I will add a test for it to check memory

PR #6452, opened 2023-05-12
replication_mode: fix the state cannot switch to async while existing learner node

nolouch:

Issue Number:  ref #4399  Close https://github.com/tikv/tikv/issues/14704

```commit-message
replication_mode: fix the state cannot switch to async while existing learner node
 - skip the learner node when check the stores state
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/replication/replication_mode.go
```diff
@@ -503,6 +503,10 @@ func (m *ModeManager) checkStoreStatus() [][]uint64 {
 		if s.IsRemoved() {
 			continue
 		}
+		// learner peers do not participate in major commit or vote, so it should not count in primary/dr as a normal store.
+		if s.GetRegionCount() == s.GetLearnerCount() {
```
JmPotato, line 507: Is this comparison method confident? Do we need to check the actual status of each region?

nolouch: It's updated when the region state change, currently, I think it is believable. In addition, this does not require strict checks, because all states do not influence the data consistency, this state only let TiKV know how to decide the majority peers.

JmPotato (approved):

rleungx (approved):

PR #6479, opened 2023-05-17
client: return error if the keyspace meta cannot be found

rleungx:

Issue Number: Ref #6142.

Just as the title says. 0 is reserved for the "DEFAULT" keyspace.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -396,7 +396,7 @@ func NewClientWithKeyspaceName(
 	log.Info("[pd] create pd client with endpoints and keyspace",
 		zap.Strings("pd-address", svrAddrs), zap.String("keyspace-name", keyspaceName))
 
-	// if keyspace is empty, fall back to the legacy API
+	// if keyspace name is empty, fall back to the legacy API
 	if len(keyspaceName) == 0 {
 		return NewClientWithContext(ctx, svrAddrs, security, opts...)
 	}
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -445,7 +445,7 @@ func NewClientWithKeyspaceName(
 func (c *client) initRetry(f func(s string) error, str string) error {
 	var err error
 	for i := 0; i < c.option.maxRetryTimes; i++ {
-		if err = f(str); err == nil || strings.Contains(err.Error(), "ENTRY_NOT_FOUND") {
```
lhy1024, line 448: why remove it?

rleungx: We should set the keyspace id to 0 if there is no meta.

ystaticy (approved):

client/client.go
```diff
@@ -390,14 +390,14 @@ func createClientWithKeyspace(
...
 func NewClientWithKeyspaceName(
-	ctx context.Context, keyspace string, svrAddrs []string,
+	ctx context.Context, keyspaceName string, svrAddrs []string,
 	security SecurityOption, opts ...ClientOption,
 ) (Client, error) {
 	log.Info("[pd] create pd client with endpoints and keyspace",
-		zap.Strings("pd-address", svrAddrs), zap.String("keyspace", keyspace))
+		zap.Strings("pd-address", svrAddrs), zap.String("keyspace-name", keyspaceName))
 
 	// if keyspace is empty, fall back to the legacy API
```
binshi-bing, line 399:
```suggestion
	// if keyspace name is empty, fall back to the legacy API
```

binshi-bing (approved):

PR #6482, opened 2023-05-18
client: support use API context to create client

rleungx:

Issue Number: Ref #6142.

This PR introduces `NewClientWithAPIContext` to create pd client according to the different API version.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -395,9 +395,10 @@ func createClientWithKeyspace(
 type APIVersion int
 
 // The API versions the client supports.
+// As for V1TTL, client won't use it and we just remove it.
 const (
 	V1 APIVersion = iota
-	V1TTL
+	_
 	V2
 )
 
@@ -421,23 +422,6 @@ func (apiCtx *apiContextV1) GetAPIVersion() (version APIVersion) {
 
 // GetKeyspaceName returns the keyspace name.
 func (apiCtx *apiContextV1) GetKeyspaceName() (keyspaceName string) {
-	return ""
-}
-
-type apiContextV1TTL struct{}
-
-// NewAPIContextV1TTL creates a API context for V1TTL.
-func NewAPIContextV1TTL() APIContext {
-	return &apiContextV1TTL{}
-}
-
-// GetAPIVersion returns the API version.
-func (apiCtx *apiContextV1TTL) GetAPIVersion() (version APIVersion) {
-	return V1TTL
-}
-
-// GetKeyspaceName returns the keyspace name.
-func (apiCtx *apiContextV1TTL) GetKeyspaceName() (keyspaceName string) {
 	return ""
 }
 
@@ -470,7 +454,7 @@ func NewClientWithAPIContext(
 ) (Client, error) {
 	apiVersion, keyspaceName := apiCtx.GetAPIVersion(), apiCtx.GetKeyspaceName()
 	switch apiVersion {
-	case V1, V1TTL:
+	case V1:
 		return NewClientWithContext(ctx, svrAddrs, security, opts...)
 	case V2:
 		return newClientWithKeyspaceName(ctx, keyspaceName, svrAddrs, security, opts...)
--- a/client/client.go
+++ b/client/client.go
@@ -52,6 +52,7 @@ const (
 	// defaultKeySpaceGroupID is the default key space group id.
 	// We also reserved 0 for the keyspace group for the same purpose.
 	defaultKeySpaceGroupID = uint32(0)
+	defaultKeyspaceName    = "DEFAULT"
 )
 
 // Region contains information of a region's meta and its peers.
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -388,6 +389,79 @@ func createClientWithKeyspace(
...
+			keyspaceName = defaultKeyspaceName
+		}
+		return NewClientWithKeyspaceName(ctx, keyspaceName, svrAddrs, security, opts...)
+	default:
+		return nil, errors.Errorf("[pd] invalid API version %d", apiVersion)
+	}
+}
+
 // NewClientWithKeyspaceName creates a client with context and the specified keyspace name.
 func NewClientWithKeyspaceName(
```
ystaticy, line 466: We should avoid external call to `NewClientWithKeyspaceName `？

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
 	return c, nil
 }
 
-// NewClientWithKeyspaceName creates a client with context and the specified keyspace name.
-func NewClientWithKeyspaceName(
+// APIVersion is the API version the server and the client is using.
```
binshi-bing, line 393:
```suggestion
// APIVersion is the API version used by the server and the client.
```

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
...
+	}
+}
+
+// newClientWithKeyspaceName creates a client with context and the specified keyspace name.
+func newClientWithKeyspaceName(
 	ctx context.Context, keyspaceName string, svrAddrs []string,
 	security SecurityOption, opts ...ClientOption,
 ) (Client, error) {
 	log.Info("[pd] create pd client with endpoints and keyspace",
 		zap.Strings("pd-address", svrAddrs), zap.String("keyspace-name", keyspaceName))
```
ystaticy, line 473: How about move `if keyspaceName == ""` logic to here?
newClientWithKeyspaceName is only used for apiv2;
If someone else calls this function, there is no need to repeat the judgment logic

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
...
+// The API versions the client supports.
+const (
+	V1 APIVersion = iota
+	V1TTL
+	V2
+)
+
+// APIContext is the context for API version.
+type APIContext interface {
+	GetAPIVersionInfo() (apiVersion APIVersion, keyspaceName string)
```
iosmanthus, line 406: How about splitting this method into:
```go
type APIContext interface {
	GetKeyspaceName() string
	GetAPIVersion() APIVersion
}
```

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
 	return c, nil
 }
 
-// NewClientWithKeyspaceName creates a client with context and the specified keyspace name.
-func NewClientWithKeyspaceName(
+// APIVersion is the API version the server and the client is using.
+// See more details in https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md#kvproto
+type APIVersion int
+
+// The API versions the client supports.
```
ystaticy, line 397: @iosmanthus Is this suitable for the design of APIVersion?

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
...
+	return V1TTL, ""
+}
+
+type apiContextV2 struct {
+	keyspaceName string
+}
+
+// NewAPIContextV2 creates a API context with the specified keyspace name for V2.
+func NewAPIContextV2(keyspaceName string) APIContext {
+	return &apiContextV2{keyspaceName: keyspaceName}
```
iosmanthus, line 439: How about moving the deriving logic here:
```go
if keyspaceName == "" {
	keyspaceName = "DEFAULT"
}
return &apiContextV2{keyspaceName: keyspaceName}
```

client/client.go
rleungx: https://github.com/tikv/pd/pull/6482#discussion_r1197502311 may be better?

client/client.go
```diff
@@ -388,19 +390,87 @@ func createClientWithKeyspace(
...
-// NewClientWithKeyspaceName creates a client with context and the specified keyspace name.
-func NewClientWithKeyspaceName(
+// APIVersion is the API version the server and the client is using.
+// See more details in https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md#kvproto
+type APIVersion int
+
+// The API versions the client supports.
+const (
+	V1 APIVersion = iota
+	V1TTL
```
binshi-bing, line 400: What's the difference between V1 and V1TTL? So far, this PR doesn't reflect any differentiation. Do we intend to keep V1TTL here in case V1TTL might be different from V1 in NewClientWithAPIContext in the future?

client/client.go
binshi-bing: +1

client/client.go
rleungx: They are almost the same.

client/client.go
rleungx: I think so. /cc @iosmanthus

client/client.go
binshi-bing: nope. The red one has typo. Alternative "APIVersion is the API version that the server and the client are using."

binshi-bing (approved):
LGTM

client/client.go
iosmanthus: This reminds me that V1TTL is not designed for clients, we should remove this from the implementation. However, the value of the APIVersion should be synced with kvproto.

ystaticy (approved):

iosmanthus (approved):

Issue #6485, opened 2023-05-18
Hit connection running loop panic in tidb when do patrol check in serverless prod env


mayjiang0203:
### What did you see instead?
![image](https://github.com/tikv/pd/assets/9443637/dc98c0e6-13cd-456c-b1c1-65ac0b0dc441)

```
[...]

PR #6486, opened 2023-05-18
controller: avoid the invalid nil pointer

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/6485

```commit-message
controller: avoid the invalid nil pointer
```

### Check List

Tests

- Unit test
- Integration test
ref case https://github.com/tikv/pd/issues/6485

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/resource_group/controller/model.go
```diff
@@ -156,44 +156,59 @@ func (dsc *SQLCalculator) AfterKVRequest(consumption *rmpb.Consumption, req Requ
...
 		return custom.RRU + custom.WRU
 	}
 	return 0
 }
 
 func getRUTokenBucketSetting(group *rmpb.ResourceGroup, typ rmpb.RequestUnitType) *rmpb.TokenBucket {
+	if group == nil {
+		return nil
+	}
 	if typ == 0 {
```
JmPotato, line 172: What about also changing to `rmpb.RequestUnitType_XXX`?

client/resource_group/controller/model.go
```diff
@@ -156,44 +156,59 @@ func (dsc *SQLCalculator) AfterKVRequest(consumption *rmpb.Consumption, req Requ
...
+	case rmpb.RawResourceType_IOWriteFlow:
 		return group.RawResourceSettings.IoWrite
 	}
 	return nil
 }
 
 func add(custom1 *rmpb.Consumption, custom2 *rmpb.Consumption) {
+	if custom1 == nil || custom2 == nil {
+		return
+	}
```
JmPotato, line 211: Is it reasonable to return the non-nil one and only return nothing when both are nil?

client/resource_group/controller/model.go
```diff
@@ -156,44 +156,59 @@ func (dsc *SQLCalculator) AfterKVRequest(consumption *rmpb.Consumption, req Requ
 }
 
 func getRUValueFromConsumption(custom *rmpb.Consumption, typ rmpb.RequestUnitType) float64 {
-	if typ == 0 {
+	if custom == nil {
```
HuSharp, line 159: Does it also need to be checked in `Trickle`?

client/resource_group/controller/model.go
nolouch: I'm not getting your point. It's reasonable to return once it has nil in here I think.

nolouch:
PTAL @JmPotato @CabinfeverB @HuSharp

client/resource_group/controller/model.go
CabinfeverB: It means if custom1 is nil, we can return custom2

nolouch: There is some misunderstanding here. this function no return values, just exit the function stack.

CabinfeverB (commented):
Is it possible nil consumption occur  in`KVCalculator`?
rest LGTM.

client/resource_group/controller/model.go
CabinfeverB: OK

HuSharp (approved):

JmPotato (approved):

CabinfeverB (approved):

PR #6488, opened 2023-05-18
mcs, tso: implement GetMinTS gPRC on both API leader and PD client 

binshi-bing:

Issue Number: Ref #5895

```commit-message
implement GetMinTS gPRC on both API leader and the PD client.
```

### Check List

Tests

- Unit test (covered by the existing unittest TestGetMinTS)

### Release note

```release-note
None.
```


client/client.go
```diff
@@ -831,23 +832,28 @@ func (c *client) GetLocalTS(ctx context.Context, dcLocation string) (physical in
...
 func (c *client) GetMinTS(ctx context.Context) (physical int64, logical int64, err error) {
-	tsoClient, serviceMode := c.getServiceClientProxy()
-	if tsoClient == nil {
-		return 0, 0, errs.ErrClientGetMinTSO.FastGenByArgs("tso client is nil")
+	protoClient := c.getClient()
+	if protoClient == nil {
+		return 0, 0, errs.ErrClientGetProtoClient
 	}
 
-	switch serviceMode {
```
rleungx, line 839: Why remove it?

server/grpc_service.go
```diff
@@ -125,18 +122,160 @@ func (s *GrpcServer) GetClusterInfo(ctx context.Context, _ *pdpb.GetClusterInfoR
...
+		return &pdpb.GetMinTSResponse{
+			Header: s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN, err.Error()),
+		}, nil
+	}
+
+	var (
+		minTS *pdpb.Timestamp
+		err   error
+	)
+	if s.IsAPIServiceMode() {
```
rleungx, line 141: Should we check if the tso server has implemented this method?

server/grpc_service.go
```diff
@@ -125,18 +122,160 @@ func (s *GrpcServer) GetClusterInfo(ctx context.Context, _ *pdpb.GetClusterInfoR
...
+	var mutex sync.Mutex
+	resps := make([]*tsopb.GetMinTSResponse, 0)
+	wg := sync.WaitGroup{}
+	wg.Add(len(addrs))
+	for _, addr := range addrs {
+		go func(addr string) {
+			defer wg.Done()
+			resp, err := s.getMinTSFromSingleServer(s.ctx, dcLocation, addr)
+			if err != nil || resp == nil {
+				log.Warn("[tso] failed to get min ts from tso server",
```
rleungx, line 182:
```suggestion
				log.Warn("failed to get min ts from tso server",
```

client/client.go
binshi-bing: Since QPS is very low, decided to always let client query GetMinTS through the API leader. In this way, the logic is moved to the API leader's GetMinTS() server side implementation as shown below:
```
if s.IsAPIServiceMode() {
		minTS, err = s.GetMinTSFromTSOService(tso.GlobalDCLocation)
	} else {
		start := time.Now()
		ts, internalErr := s.tsoAllocatorManager.HandleRequest(tso.GlobalDCLocation, 1)
		if internalErr == nil {
			tsoHandleDuration.Observe(time.Since(start).Seconds())

   	         }
		minTS = &ts
	}
```

server/grpc_service.go
binshi-bing: no need, right? Two reasons:
1. The tso service GetMinTS kvproto change and server side implementation is earlier than this pr, so this is only about the startup order of the API servers and tso servers during service upgrade.
2. GetMinTS will always fail during service upgrade and there is no correctness related issue because of the validation during aggregation. Because GetMinTS is an operational request with low QPS, failure during server upgrade should be acceptable.

server/grpc_service.go
binshi-bing: changed.

rleungx (approved):

go.mod
```diff
@@ -204,3 +204,5 @@ replace google.golang.org/grpc v1.54.0 => google.golang.org/grpc v1.26.0
 // kvproto at the same time. You can run `go mod tidy` to make it replaced with go-mod style specification.
 // After the PR to kvproto is merged, remember to comment this out and run `go mod tidy`.
 // replace github.com/pingcap/kvproto => github.com/$YourPrivateRepo $YourPrivateBranch
+
+replace github.com/pingcap/kvproto => github.com/rleungx/kvproto v0.0.0-20230516051126-0db8e6f4780d
```
rleungx, line 208: https://github.com/pingcap/kvproto/pull/1116 is merged

binshi-bing: upgraded to kvproto/master.

server/grpc_service.go
```diff
@@ -125,18 +122,160 @@ func (s *GrpcServer) GetClusterInfo(ctx context.Context, _ *pdpb.GetClusterInfoR
...
+
+	return &pdpb.GetMinTSResponse{
+		Header:    s.header(),
+		Timestamp: minTS,
+	}, nil
+}
+
+// GetMinTSFromTSOService queries all tso servers and gets the minimum timestamp across
+// all keyspace groups.
+func (s *GrpcServer) GetMinTSFromTSOService(dcLocation string) (*pdpb.Timestamp, error) {
```
binshi-bing, line 166: This is refactor work. The code in this function is mainly moved from client side.

server/grpc_service.go
```diff
@@ -125,18 +122,160 @@ func (s *GrpcServer) GetClusterInfo(ctx context.Context, _ *pdpb.GetClusterInfoR
...
+	}
+
+	if minTS == nil {
+		return &pdpb.Timestamp{}, errs.ErrGetMinTS.FastGenByArgs("the tso service is not ready")
+	}
+
+	return minTS, nil
+}
+
+func (s *GrpcServer) getMinTSFromSingleServer(
```
binshi-bing, line 234: This is refactor work. The code in this function is mainly moved from client side.

server/grpc_service.go
```diff
@@ -1361,6 +1500,34 @@ func (s *GrpcServer) UpdateServiceGCSafePoint(ctx context.Context, request *pdpb
 	}, nil
 }
 
+// GetGCSafePointV2 implements gRPC PDServer.
+// Note: we need latest version of kvproto/master, but there was earlier commit https://github.com/pingcap/kvproto/pull/1111
```
binshi-bing, line 1504: @ystaticy, I need the latest version of pingcap/kvproto@master, but it includes your earlier commit https://github.com/pingcap/kvproto/pull/1111 whose server side implementation hasn't been merged, so I added the following empty methods to avoid compile error.

client/client.go
binshi-bing: I added some code back to handle compatibility issue in case of PD/API server doesn't support GetMinTS API

client/client.go
```diff
@@ -844,10 +848,37 @@ func (c *client) GetMinTS(ctx context.Context) (physical int64, logical int64, e
...
+	protoClient := c.getClient()
+	if protoClient == nil {
+		return 0, 0, errs.ErrClientGetProtoClient
+	}
+
+	resp, err := protoClient.GetMinTS(ctx, &pdpb.GetMinTSRequest{
+		Header: c.requestHeader(),
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "Unimplemented") {
```
lhy1024, line 865: Do we need to add a test for this error?

lhy1024 (approved):

client/client.go
binshi-bing: It's pretty straightforward, not necessary?

Issue #6493, opened 2023-05-19
Provide an interface to halt all scheduler and checking


JmPotato:
## Development Task

In some special situations, suspending all scheduling is a very helpful behavior, such as before performing Flashback on the TiKV cluster, so it is best to provide an interface to do this. It should be noted that the suspension of scheduling here is best able to remember and restore the old configuration, rather than simply deleting all schedulers and checkers in a crude manner.

PR #6498, opened 2023-05-22
config, cluster: add an option to halt the cluster scheduling

JmPotato:

Issue Number: ref #6493.

```commit-message
Add an option to halt the cluster scheduling.
```

### Check List

Tests

- Unit test
- Integration test

During the halt:

image

image

image

### Release note

```release-note
None.
```


server/cluster/coordinator.go
```diff
@@ -135,8 +135,7 @@ func (c *coordinator) patrolRegions() {
 			log.Info("patrol regions has been stopped")
 			return
 		}
-		if c.cluster.GetUnsafeRecoveryController().IsRunning() {
-			// Skip patrolling regions during unsafe recovery.
+		if allowed, _ := c.cluster.checkSchedulingAllowance(); !allowed {
```
nolouch, line 138: better add some metrics.

server/config/config.go
```diff
@@ -684,6 +685,10 @@ type ScheduleConfig struct {
 	// v1: which is based on the region count by rate limit.
 	// v2: which is based on region size by window size.
 	StoreLimitVersion string `toml:"store-limit-version" json:"store-limit-version,omitempty"`
+
+	// HaltScheduling is the option to halt the scheduling. Once it's on, PD will halt the scheduling,
+	// and any other scheduling configs will be ignored.
+	HaltScheduling bool `toml:"halt-scheduling" json:"halt-scheduling,string,omitempty"`
```
rleungx, line 691: Previously, I am trying to introduce a scheduling mode to cover this case. For me, it's ok to use an individual config to control it. Maybe we can name it enable-scheduling or something else.

JmPotato: I think it's best to use a configuration name with a default value of false to control the global scheduling switch, in order to avoid unexpected behaviors in scenarios that require compatibility considerations such as upgrades. Therefore, from this perspective, I think descriptions like "disable" or "halt" are more appropriate. At the same time, this global shutdown scheduling behavior should not be long-term. In addition, we already have the concept and operation of "pause" for Scheduler. So I ultimately chose the word "halt". WDYT?

server/cluster/coordinator.go
JmPotato: Added. PTAL again.

server/cluster/coordinator.go
```diff
@@ -947,7 +946,14 @@ func (s *scheduleController) AllowSchedule(diagnosable bool) bool {
 		}
 		return false
 	}
-	if s.IsPaused() || s.cluster.GetUnsafeRecoveryController().IsRunning() {
+	allowed, _ := s.cluster.checkSchedulingAllowance()
+	if !allowed {
+		if diagnosable {
+			s.diagnosticRecorder.setResultFromStatus(halted)
```
JmPotato, line 952: Please help take a look here, thx. @CabinfeverB

server/cluster/cluster.go
```diff
@@ -2733,3 +2733,25 @@ func (c *RaftCluster) GetPausedSchedulerDelayAt(name string) (int64, error) {
...
+
+var (
+	onlineUnsafeRecoveryStatus = schedulingAllowanceStatusGauge.WithLabelValues("online-unsafe-recovery")
+	haltSchedulingStatus       = schedulingAllowanceStatusGauge.WithLabelValues("halt-scheduling")
+)
+
+// checkSchedulingAllowance checks if the cluster allows scheduling.
+func (c *RaftCluster) checkSchedulingAllowance() (bool, error) {
+	// If the cluster is in the process of online unsafe recovery, it should not allow scheduling.
+	if c.GetUnsafeRecoveryController().IsRunning() {
```
binshi-bing, line 2745: Is there atomicity needed for the status check/update on c.GetUnsafeRecoveryController().IsRunning() and onlineUnsafeRecoveryStatus?

server/cluster/cluster.go
```diff
@@ -2733,3 +2733,25 @@ func (c *RaftCluster) GetPausedSchedulerDelayAt(name string) (int64, error) {
...
+		onlineUnsafeRecoveryStatus.Set(1)
+		return false, errs.ErrUnsafeRecoveryIsRunning.FastGenByArgs()
+	}
+	onlineUnsafeRecoveryStatus.Set(0)
+	// If the halt-scheduling is set, it should not allow scheduling.
+	if c.opt.IsSchedulingHalted() {
```
binshi-bing, line 2751: It sounds that once the scheduling is halted, the scheduler won't accept or initiate the new scheduling task? What will happen to the accepted/running scheduling task? Do you need to drain the running scheduling tasks or pause the long running scheduling tasks before returning false below?

server/cluster/cluster.go
```diff
@@ -2733,3 +2733,25 @@ func (c *RaftCluster) GetPausedSchedulerDelayAt(name string) (int64, error) {
...
+		haltSchedulingStatus.Set(1)
+		return false, errs.ErrSchedulingIsHalted.FastGenByArgs()
```
binshi-bing, line 2753: If this is configured to halt scheduling, returning error here seems to be inappropriate.

server/cluster/cluster.go
JmPotato: It has been protected by an internal lock of `UnsafeRecoveryController`.

server/cluster/cluster.go
JmPotato: This is used to return a corresponding error in `HandleAskSplit` and `HandleAskBatchSplit` directly.

server/cluster/cluster.go
JmPotato: Based on the current usage scenarios that require termination of scheduling, there is no such demand. We only need to ensure that no new scheduling is generated.

server/cluster/coordinator.go
CabinfeverB: This part LGTM

CabinfeverB (commented):
rest LGTM

metrics/grafana/pd.json
```diff
@@ -2340,6 +2340,113 @@
             "alignLevel": null
           }
         },
+        {
+          "aliasColors": {},
+          "bars": true,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "${DS_TEST-CLUSTER}",
+          "description": "The allowance status of the scheduling.",
```
CabinfeverB, line 2349: How about putting is near "Scheduler is running"
But it makes sense where it is now.

nolouch (approved):
lgtm

metrics/grafana/pd.json
JmPotato: I prefer leaving it here since it's more like a cluster-level status rather than the scheduler. Another reason is that if it is placed in the `Scheduler` panel, it may cause many changes to the Grafana JSON file. If it is only appended here, there will be fewer changes.

JmPotato:
@rleungx @binshi-bing PTAL, thx.

CabinfeverB (approved):

CabinfeverB (approved):

server/config/config.go
rleungx: When I work on #6553, I found that maybe it's better to use one config for both unsafe recovery or halt, so that we can decouple the dependencies between cluster and coordinator.

PR #6511, opened 2023-05-24
tso: log tso service discovery info on the client side only when the primary is changed

### LINKED ISSUE #6508: Client keeps updating the keyspace group discovery info
### What did you do?

img_v2_44c8f357-6335-4a46-92af-b94354ea762g

There should be a few logs about updating the info of the discovery.

### What did you see instead?

Too frequently and the members field keeps changing. 

master

binshi-bing:

Issue Number: Close #6508 

```commit-message
Skip logging of the tso service discovery info when the secondar list is changed,
because tso servers currently don't have consistent view of the member list due to
remote etcd being used by tso service, which results in changing member list when
the client queries the tso servers in round-robin. We need to improve the server side
so that all tso servers can return the global consistent view of the keyspace groups'
serving or membership info.
```

### Check List

Tests

- No code

```release-note
None.
```

### DIFF
--- a/client/tso_service_discovery.go
+++ b/client/tso_service_discovery.go
@@ -488,16 +488,14 @@ func (c *tsoServiceDiscovery) updateMember() error {
 		}
 	}
 
-	oldPrimary, primarySwitched, secondaryChanged :=
+	oldPrimary, primarySwitched, _ :=
 		c.keyspaceGroupSD.update(keyspaceGroup, primaryAddr, secondaryAddrs, addrs)
 	if primarySwitched {
+		log.Info("[tso] updated keyspace group service discovery info",
+			zap.String("keyspace-group-service", keyspaceGroup.String()))
 		if err := c.afterPrimarySwitched(oldPrimary, primaryAddr); err != nil {
 			return err
 		}
-	}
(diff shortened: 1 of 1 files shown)


client/tso_service_discovery.go
```diff
@@ -488,17 +488,15 @@ func (c *tsoServiceDiscovery) updateMember() error {
 		}
 	}
 
-	oldPrimary, primarySwitched, secondaryChanged :=
+	oldPrimary, primarySwitched, _ :=
```
rleungx, line 491: Do we still need to return the field if we don't use it? If we plan to use it, it's better to add a TODO here.

rleungx (approved):
The rest LGTM

rleungx (approved):
The rest LGTM

client/tso_service_discovery.go
binshi-bing: We will use it after we add two watch loops to get get all registered tso servers and the entire election member list for every keyspace group so that every tso server has the global consistent view.
When we support followProxy, this returned value will be naturally used so no need to add TODO.

JmPotato (approved):

PR #6518, opened 2023-05-25
mcs: fix the members field is null

rleungx:

Issue Number: Close #xxx

This PR fixes the issue by updating the cache. BTW, this PR also fixes a potential panic if the meta is nil.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


JmPotato (approved):

binshi-bing (approved):
Great!

pkg/keyspace/keyspace.go
```diff
@@ -390,7 +390,9 @@ func (manager *Manager) LoadKeyspaceByID(spaceID uint32) (*keyspacepb.KeyspaceMe
 		}
 		return nil
 	})
-	meta.Id = spaceID
+	if meta != nil {
```
binshi-bing, line 393: I checked the entire function. Oh, My god!

binshi-bing: Good catch!

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -162,7 +167,7 @@ func (m *GroupManager) allocNodesToAllKeyspaceGroups() {
 		}
 		groups, err := m.store.LoadKeyspaceGroups(utils.DefaultKeyspaceGroupID, 0)
 		if err != nil {
-			log.Error("failed to load the all keyspace group", zap.Error(err))
+			log.Error("failed to load the all keyspace groups", zap.Error(err))
```
binshi-bing, line 170:
```suggestion
			log.Error("failed to load all keyspace groups", zap.Error(err))
```

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -672,6 +680,7 @@ func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, desiredReplicaCount
 	if err != nil {
 		return nil, err
 	}
+	m.groups[endpoint.StringUserKind(kg.UserKind)].Put(kg)
 	log.Info("alloc nodes for keyspace group", zap.Uint32("id", id), zap.Reflect("nodes", nodes))
```
binshi-bing, line 684: here we use "id" as the field name of the keyspace group id. Shall we unify the names at the line 179 and line 684?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -672,6 +680,7 @@ func (m *GroupManager) AllocNodesForKeyspaceGroup(id uint32, desiredReplicaCount
 	if err != nil {
 		return nil, err
 	}
+	m.groups[endpoint.StringUserKind(kg.UserKind)].Put(kg)
```
binshi-bing, line 683: I see. We refresh cache here.

binshi-bing (commented):
left few comments

binshi-bing (commented):
Can we add more explanation of the root cause in the PR description? "This PR fixes the issue by updating the cache" is too short. Like "AllocNodesForKeyspaceGroup assigned the members to the default keyspace group and persistent the state but didn't update cache. Later, when we add the keyspaces and update the keyspace list of the default keyspace group, the system uses the dirty cache state (members is still empty) to overwrite the keyspace list in the store"

binshi-bing (approved):

PR #6526, opened 2023-05-26
mcs, tso: change keyspace group primary path.

binshi-bing:
Issue Number: Ref #5895

1. We need to add watch loop to every TSO server to watch the primaries of all keyspace groups. When #TSOServers > #GroupReplica, a TSO server only owns replicas of part of keyspace groups, i.e., the TSO server doesn't know the primaries of keyspace groups that the TSO server doesn't own. The partial keyspace group state view leads to slow TSO service discovery and inconsistent membership view.
2. The above watch loop needs to watch the range [/ms/{cluster_id}/tso/00000/primary, /ms/{cluster_id}/tso/99999/primary], but it overlaps with keyspace groups' timestamp path /ms/{cluster_id}/tso/{group}/{gta|lta}/.../timestamp which results in huge inefficiency (as far as I know, there is no built-in filter based on suffix).

Why do we keep the same path for Default keyspace group?
It has been deployed to prod, and the change will cause incompatability issues. Leave the change and upgrade to the future.

```commit-message
mcs, tso: change keyspace group primary path.

The path for non-default keyspace group primary election changes
from  "/ms/{cluster_id}/tso/{group}/primary" to "/ms/{cluster_id}/tso/keyspace_groups/election/{group}/primary".
Default keyspace group keeps /ms/{cluster_id}/tso/00000/primary.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/storage/endpoint/key_path.go
```diff
@@ -238,20 +237,10 @@ func KeyspaceGroupIDPath(id uint32) string {
 	return path.Join(tsoKeyspaceGroupPrefix, keyspaceGroupMembershipKey, encodeKeyspaceGroupID(id))
 }
 
-// ExtractKeyspaceGroupIDFromPath extracts keyspace group id from the given path, which contains
-// the pattern of `tso/keyspace_groups/membership/(\d{5})$`.
-func ExtractKeyspaceGroupIDFromPath(path string) (uint32, error) {
+// GetCompiledKeyspaceGroupIDRegexp returns the compiled regular expression for matching keyspace group id.
```
rleungx, line 240: Why do we need to separate it into two functions?

pkg/tso/keyspace_group_manager.go
```diff
@@ -46,9 +47,10 @@ import (
 )
 
 const (
-	// primaryElectionSuffix is the suffix of the key for keyspace group primary election
-	primaryElectionSuffix = "primary"
-	defaultRetryInterval  = 500 * time.Millisecond
+	keyspaceGroupsElectionPath = mcsutils.KeyspaceGroupsKey + "/election"
+	// leaderNameKey is the key for keyspace group primary election.
+	leaderNameKey        = "primary"
```
rleungx, line 52: How about `primaryKey`?

pkg/storage/endpoint/key_path.go
binshi-bing: performance concern. Compile the patten once and use it all the time to boost performance.

pkg/tso/keyspace_group_manager.go
binshi-bing: changed to primaryKey.

tests/integrations/mcs/tso/keyspace_group_manager_test.go
```diff
@@ -194,16 +194,24 @@ func (suite *tsoKeyspaceGroupManagerTestSuite) TestKeyspacesServedByNonDefaultKe
 						re.NotNil(am)
 
 						// Make sure every keyspace group is using the right timestamp path
-						// for loading/saving timestamp from/to etcd.
-						var timestampPath string
+						// for loading/saving timestamp from/to etcd and the right primary path
+						// for leader election.
```
rleungx, line 198: primary

rleungx (approved):
The rest LGTM

lhy1024 (approved):

tests/integrations/mcs/tso/keyspace_group_manager_test.go
binshi-bing: changed leader to primary

PR #6529, opened 2023-05-28
Fix tso server close stuck issue

### LINKED ISSUE #6304: TestMixedTSODeployment timeout sometimes
### Which jobs are failing
TestMixedTSODeployment 

### CI link
https://github.com/tikv/pd/actions/runs/4655922619/jobs/8239044048
https://github.com/tikv/pd/actions/runs/4667209181/jobs/8263220385
https://github.com/tikv/pd/actions/runs/4654377314/jobs/8236052896
https://github.com/tikv/pd/actions/runs/4637756863/jobs/8206938196

binshi-bing:

Issue Number: Close #6304
Issue Number: Ref #5895 

```commit-message
Rewrite TSO gPRC/HTTP server Close().
```

### Check List

Tests

- Unit test
Reran TestMixedTSODeployment test cases 30 times, the issue wasn't reproed. Previously, it was reproed every 3 ~ 5 times.

-->

```release-note
None.
```

### DIFF
--- a/pkg/mcs/tso/server/server.go
+++ b/pkg/mcs/tso/server/server.go
@@ -69,7 +69,6 @@ const (
 	// tsoSvcRootPathFormat defines the root path for all etcd paths used for different purposes.
 	// format: "/ms/{cluster_id}/tso".
 	tsoSvcRootPathFormat = msServiceRootPath + "/%d/" + mcsutils.TSOServiceName
-
 	// maxRetryTimesWaitAPIService is the max retry times for initializing the cluster ID.
 	maxRetryTimesWaitAPIService = 360
 	// retryIntervalWaitAPIService is the interval to retry.
@@ -106,7 +105,11 @@ type Server struct {
 	// http client
 	httpClient *http.Client
 
+	secure               bool
 	muxListener          net.Listener
+	httpListener         net.Listener
+	grpcServer           *grpc.Server
+	httpServer           *http.Server
 	service              *Service
 	keyspaceGroupManager *tso.KeyspaceGroupManager
 	// Store as map[string]*grpc.ClientConn
@@ -187,6 +190,8 @@ func (s *Server) Close() {
 	// close tso service loops in the keyspace group manager
 	s.keyspaceGroupManager.Close()
 	s.serviceRegister.Deregister()
+	s.stopHTTPServer()
+	s.stopGRPCServer()
 	s.muxListener.Close()
 	s.serverLoopCancel()
 	s.serverLoopWg.Wait()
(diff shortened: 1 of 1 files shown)


pkg/mcs/tso/server/server.go
```diff
@@ -389,83 +394,128 @@ func (s *Server) startGRPCServer(l net.Listener) {
...
-	}
-
+	log.Info("[tso] grpc server starts serving", zap.String("address", l.Addr().String()))
+	err := s.grpcServer.Serve(l)
 	if s.IsClosed() {
-		log.Info("grpc server stopped")
+		log.Info("[tso] grpc server stopped")
 	} else {
-		log.Fatal("grpc server stopped unexpectedly", errs.ZapError(serverr))
+		log.Fatal("[tso] grpc server stopped unexpectedly", errs.ZapError(err))
```
rleungx, line 402: We don't need to change it because [xxx] is mainly used for the client.

pkg/mcs/tso/server/server.go
```diff
@@ -389,83 +394,128 @@ func (s *Server) startGRPCServer(l net.Listener) {
...
-			log.Info("mux stop serving", errs.ZapError(err))
+			log.Info("[tso] mux stopped serving", errs.ZapError(err))
 		} else {
-			log.Panic("mux stop serving unexpectedly", errs.ZapError(err))
+			log.Fatal("[tso] mux stopped serving unexpectedly", errs.ZapError(err))
 		}
 	}
 }
 
+func (s *Server) stopHTTPServer() {
```
lhy1024, line 457: it is similar to stopGRPCServer. And we could use them in other micro services.

pkg/mcs/tso/server/server.go
```diff
@@ -510,8 +561,11 @@ func (s *Server) startServer() (err error) {
 		return err
 	}
 
+	serverReadyChan := make(chan struct{})
```
lhy1024, line 564: why add it?

pkg/mcs/tso/server/server.go
binshi-bing: Yes, stopHTTPServer and stopGRPCServer are similar, but there is no need to write a general function, because there are still a few differences between these two functions. stopHTTPServer and stopGRPCServer can be used in outher micro services.

pkg/mcs/tso/server/server.go
binshi-bing: removed all

pkg/mcs/tso/server/server.go
binshi-bing: To ensure tso.Server.StartServer() successfully exits right before cmux.CMux.Serve() is called. Without this, the TestServerRegisterTestSuite.TestServerPrimaryChange will fail, because it starts tso server then call tso.Server.Close() immediately. Generally, it's better to shorten the distance between tso.Server.StartServer() successfully returned and cmux.CMux.Serve() being called.

rleungx (approved):

lhy1024 (approved):

PR #6544, opened 2023-06-01
mcs, tso: delete the tso ms discovery when switching away from API mode.

### LINKED ISSUE #6543: Delete the old tso microservice discovery when switch away from API mode
## Enhancement Task
When switching from the PD mode to the API mode,  the old tso microservice discovery isn't needed anymore and all resources can be released to avoid noisy error logs when the component trying to discovery the non-existent tso microservice.

binshi-bing:

Issue Number: Close #6543 

```commit-message
When switching from the PD mode to the API mode, the old tso microservice discovery isn't needed anymore,
and all resources can be released to avoid noisy error logs when the component trying to discover
the non-existent tso microservice.
```

### Check List

Tests

- Unit test
Covered by the existing test cases.

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -621,13 +621,13 @@ func (c *client) setServiceMode(newMode pdpb.ServiceMode) {
 	oldTSOClient.Close()
 	// Replace the old TSO service discovery if needed.
 	oldTSOSvcDiscovery := c.tsoSvcDiscovery
-	if newTSOSvcDiscovery != nil {
-		c.tsoSvcDiscovery = newTSOSvcDiscovery
-		// Close the old TSO service discovery safely after both the old client
-		// and service discovery are replaced.
-		if oldTSOSvcDiscovery != nil {
-			oldTSOSvcDiscovery.Close()
-		}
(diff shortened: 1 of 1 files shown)


rleungx (approved):

client/client.go
```diff
@@ -621,13 +621,13 @@ func (c *client) setServiceMode(newMode pdpb.ServiceMode) {
 	oldTSOClient.Close()
 	// Replace the old TSO service discovery if needed.
 	oldTSOSvcDiscovery := c.tsoSvcDiscovery
-	if newTSOSvcDiscovery != nil {
-		c.tsoSvcDiscovery = newTSOSvcDiscovery
-		// Close the old TSO service discovery safely after both the old client
-		// and service discovery are replaced.
-		if oldTSOSvcDiscovery != nil {
```
binshi-bing, line 628: if newTSOSvcDiscovery != nil, it means we're switching from PD mode to API mode, the oldTSOSvcDiscovery must be nil, so this condition is always false.

JmPotato (approved):

PR #6557, opened 2023-06-06
*: extract conn and replace `GetAllocator`

rleungx:

Issue Number: Ref #5839.

Just as the title says. It should be reviewed after #6553 is merged.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/grpcutil/grpcutil.go
+++ b/pkg/utils/grpcutil/grpcutil.go
@@ -204,6 +204,7 @@ func establish(ctx context.Context, addr string, tlsConfig *TLSConfig) (*grpc.Cl
 	return cc, nil
 }
 
+// CreateClientConn creates a client connection to the given target.
 func CreateClientConn(ctx context.Context, addr string, tlsConfig *TLSConfig) *grpc.ClientConn {
 	var (
 		conn *grpc.ClientConn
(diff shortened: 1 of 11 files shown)


pkg/utils/grpcutil/grpcutil.go
```diff
@@ -160,3 +171,57 @@ func GetForwardedHost(ctx context.Context) string {
...
+			Timeout: keepaliveTimeout,
+		}),
+		grpc.WithConnectParams(grpc.ConnectParams{
+			Backoff: backoff.Config{
+				BaseDelay:  time.Second,     // Default was 1s.
+				Multiplier: 1.6,             // Default
+				Jitter:     0.2,             // Default
+				MaxDelay:   3 * time.Second, // Default was 120s.
+			},
+			MinConnectTimeout: 5 * time.Second,
```
CabinfeverB, line 196: Now that we've extracted conn, how about extracting these parameters as well

CabinfeverB (approved):
LGTM

pkg/utils/grpcutil/grpcutil.go
rleungx: sounds great

nolouch (approved):

PR #6565, opened 2023-06-07
Improve tso proxy

### LINKED ISSUE #6549: The api leader got stuck at tso requests forwarding
### What did you do?
It happened as time went when it was running in staging cluster. 

No tso forwarding stuck issue.

### What did you see instead?
The api leader got stuck at dispathTSORequest when forwarding tso requests to TSO servers.

tidbcloud/pd-cse release-6.6-keyspace 9e1e2de5194655c2e1224ce94057e7f7fb5f7eb4

### Root Cause Analysis
In short, it's about wrong usage of unbuffered channel in multiple goroutines. 

The problem is in the TSO forwarding & Dispatching framework on Server side, where many streaming process goroutines, calling Tso() gPRC api, share the same handleDispatcher() goroutine per forwarded host.

Below is the event sequence describing what happened:
1. In the first streaming process goroutine, which created the handleDispatcher() goroutine, dispatched a request then invoked the blocking call (client.gPRC stream).Recv() and waited for some time without receiving anything.
2. The handleDispatcher() goroutine failed to process the request dispatched above, then trying to pass the error through an unbuffered channel (errCh <- error)
3.  The streaming process goroutine can't move forward to reach the place to consume the error channel because of 1, so both goroutines blocked.
4.  As new Tso streaming requests came in, more streaming process goroutines were created, and they enqueued the requests to requests channel (buffered channel with 10000 capacity). Since handleDispatcher() goroutine, the consumer of requests channel, were blocked at 2, no one drained the requests in the requests channel and the max capacity of the channel was reached. Eventually there were more and more streaming process goroutines blocked at enqueuing requests to the requests channel.
5. Eventually, gPRC server can't spawn more streaming process goroutines.

**Besides the above issue in the TSO forwarding & Dispatching framework, there might have other several issues:**
1. All gPRC client streams share the same handleDispatcher() goroutine for the same forwarded host, if one gPRC client stream Send() has problem, handleDispatcher() goroutine exits, and all requests dispatched by all gPRC client streams become orphan requests.
2. When error happens in the handleDispatcher() goroutine, the error handling speed will be slowed down by blocking call of stream.Recv() in the corresponding streaming process goroutine.
3. After getting response of the batch request from tso microservice, it sequentially sends the requests to the client streams, which actually eat the benefit of batching.

binshi-bing:

Issue Number: Close #6549

```commit-message
Improve tso proxy
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -45,119 +43,159 @@ type TSODispatcher struct {
...
-	doneCh <-chan struct{},
-	errCh chan<- error,
-	tsoPrimaryWatchers ...*etcdutil.LoopWatcher) {
+// startDispatchLoop starts the dispatch loop for the forwarded host
+func (s *TSODispatcher) cleanup(
+	forwardedHost string, finalForwardErr error,
+	unprocessedRequests []Request, unprocessedReqCount int,
+) {
+	pendingReqCount := unprocessedReqCount
+	pendingRequests := unprocessedRequests[:unprocessedReqCount]
```
rleungx, line 89: Why not just use the parameter passed in the function?

pkg/mcs/tso/server/server.go
```diff
@@ -114,9 +114,6 @@ type Server struct {
 	keyspaceGroupManager *tso.KeyspaceGroupManager
 	// Store as map[string]*grpc.ClientConn
 	clientConns sync.Map
-	// tsoDispatcher is used to dispatch the TSO requests to
```
lhy1024, line 117: tso server will remove it?

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -45,119 +43,159 @@ type TSODispatcher struct {
...
+	cctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	done := make(chan struct{})
+	dl := deadline{
+		timer:  time.After(DefaultTSOProxyTimeout),
+		done:   done,
+		cancel: cancel,
+	}
+	select {
+	case tsDeadlineCh <- dl:
```
lhy1024, line 190: how about using context.WithTimeout

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -45,119 +43,159 @@ type TSODispatcher struct {
...
+	)
+	pendingRequests := make([]Request, maxMergeRequests+1)
+	pendingTSOReqCount := 0
+
+	log.Info("start the dispatch loop", zap.String("forwarded-host", forwardedHost))
+	defer func() {
+		log.Info("exiting from the dispatch loop. cleaning up the pending requests",
+			zap.String("forwarded-host", forwardedHost))
+		if forwardStream != nil {
+			forwardStream.closeSend()
```
lhy1024, line 134: Does it mean that a new stream is created for each request?

server/grpc_service.go
```diff
@@ -398,14 +367,72 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 	}
 }
 
-func (s *GrpcServer) getForwardedHost(ctx, streamCtx context.Context) (forwardedHost string, err error) {
+// forwardTSO forwards the incoming TSO requests to the TSO microservice.
+func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
```
lhy1024, line 371: Do we need to update `getGlobalTSOFromTSOServer`?

pkg/utils/tsoutil/tso_dispatcher.go
binshi-bing: It won't, because we only start one dispatch loop (this function) for all requests with the same forwarded host until there is any forwarding related error which causes the loop to exit, and we do this check and forwardStream.closeSend() in defer func() when exiting from this loop.

pkg/utils/tsoutil/tso_dispatcher.go
binshi-bing: you are right

server/grpc_service.go
binshi-bing: Before the change in this pr is proved to work as expected in dev/staging, I plan to keep the other RPCs keep the current way.

rleungx (approved):

PR #6567, opened 2023-06-07
tools: support deleting keyspace group

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -49,10 +49,9 @@ const (
 	UserKindKey = "user_kind"
 	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
 	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
-	// keyspacePatrolBatchSize is the batch size for keyspace assignment patrol.
-	// the limit of etcd txn op is 128, keyspacePatrolBatchSize need to be less than it.
+	// maxTxnOps is the batch size for operating etcd. The limit of etcd txn op is 128.
 	// See: https://github.com/etcd-io/etcd/blob/d3e43d4de6f6d9575b489dd7850a85e37e0f6b6c/server/embed/config.go#L61
-	keyspacePatrolBatchSize = 120
+	maxTxnOps = 120
 )
 
 // Config is the interface for keyspace config.
@@ -103,7 +102,6 @@ func NewKeyspaceManager(
 	cluster core.ClusterInformer,
 	idAllocator id.Allocator,
 	config Config,
-	kgm *GroupManager,
 ) *Manager {
 	return &Manager{
 		ctx:               ctx,
@@ -112,7 +110,6 @@ func NewKeyspaceManager(
 		store:             store,
 		cluster:           cluster,
 		config:            config,
-		kgm:               kgm,
 		nextPatrolStartID: utils.DefaultKeyspaceID,
 	}
 }
@@ -669,7 +666,7 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 			zap.Duration("cost", time.Since(start)),
 			zap.Uint64("patrolled-keyspace-count", patrolledKeyspaceCount),
 			zap.Uint64("assigned-keyspace-count", assignedKeyspaceCount),
-			zap.Int("batch-size", keyspacePatrolBatchSize),
+			zap.Int("batch-size", maxTxnOps),
 			zap.Uint32("current-start-id", currentStartID),
 			zap.Uint32("next-start-id", nextStartID),
 		)
@@ -688,7 +685,7 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 			if defaultKeyspaceGroup.IsSplitting() {
 				return ErrKeyspaceGroupInSplit
 			}
-			keyspaces, err := manager.store.LoadRangeKeyspace(txn, manager.nextPatrolStartID, keyspacePatrolBatchSize)
+			keyspaces, err := manager.store.LoadRangeKeyspace(txn, manager.nextPatrolStartID, maxTxnOps)
 			if err != nil {
 				return err
 			}
@@ -698,9 +695,9 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 				currentStartID = keyspaces[0].GetId()
 				nextStartID = keyspaces[keyspaceNum-1].GetId() + 1
 			}
-			// If there are less than `keyspacePatrolBatchSize` keyspaces,
+			// If there are less than `maxTxnOps`,
 			// we have reached the end of the keyspace list.
-			moreToPatrol = keyspaceNum == keyspacePatrolBatchSize
+			moreToPatrol = keyspaceNum == maxTxnOps
 			var (
 				assigned            = false
 				keyspaceIDsToUnlock = make([]uint32, 0, keyspaceNum)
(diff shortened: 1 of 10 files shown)


pkg/keyspace/keyspace.go
```diff
@@ -49,10 +49,9 @@ const (
 	UserKindKey = "user_kind"
 	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
 	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
-	// keyspacePatrolBatchSize is the batch size for keyspace assignment patrol.
-	// the limit of etcd txn op is 128, keyspacePatrolBatchSize need to be less than it.
+	// maxTxnOps is the batch size for operating etcd. The limit of etcd txn op is 128.
 	// See: https://github.com/etcd-io/etcd/blob/d3e43d4de6f6d9575b489dd7850a85e37e0f6b6c/server/embed/config.go#L61
-	keyspacePatrolBatchSize = 120
+	maxTxnOps = 120
```
lhy1024, line 54: The batch size is determined for the loaded keyspace, but in fact other operators may be included in the PatrolKeyspaceAssignment, such as SaveKeyspaceMeta and SaveKeyspaceGroup and LoadKeyspaceGroup.

tests/pdctl/keyspace/keyspace_group_test.go
```diff
@@ -120,14 +121,38 @@ func TestSplitKeyspaceGroup(t *testing.T) {
...
+		splitKeyspaces = append(splitKeyspaces, fmt.Sprintf("%d", i))
+	}
 	// split keyspace group.
 	testutil.Eventually(re, func() bool {
-		args := []string{"-u", pdAddr, "keyspace-group", "split", "0", "1", "2"}
+		args := append([]string{"-u", pdAddr, "keyspace-group", "split", "0", "1"}, splitKeyspaces...)
 		output, err := pdctl.ExecuteCommand(cmd, args...)
 		re.NoError(err)
 		return strings.Contains(string(output), "Success")
-	}, testutil.WithWaitFor(20*time.Second))
```
lhy1024, line 129: It will become unstable due to timeout

rleungx: Maybe you can check the default value. If there is a timeout. We need to investigate why instead of enlarging the wait time.

pkg/keyspace/keyspace.go
rleungx: What do you mean here?

tests/pdctl/keyspace/keyspace_group_test.go
lhy1024: Because there are more keyspaces than previous, it will take more time to prepare and split.

pkg/keyspace/keyspace.go
lhy1024: maxTxnOps is not suitable than keyspacePatrolBatchSize

rleungx: not only Patrol uses this value

tests/pdctl/keyspace/keyspace_group_test.go
rleungx: So, how long will it cost if we have 500 keyspace groups?

pkg/keyspace/keyspace.go
JmPotato, line 54:
```suggestion
	maxTxnOps = 128
```

JmPotato (commented):
Should we also reset the default keyspace group TSO to the deleted one if the latter is greater?

pkg/keyspace/keyspace.go
lhy1024: we need to ensure it is less than 128, because the txn contains other operator

pkg/keyspace/keyspace.go
```diff
@@ -698,9 +695,9 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 				currentStartID = keyspaces[0].GetId()
 				nextStartID = keyspaces[keyspaceNum-1].GetId() + 1
 			}
-			// If there are less than `keyspacePatrolBatchSize` keyspaces,
+			// If there are less than `maxTxnOps`,
```
binshi-bing, line 698: The original comment is correct
```suggestion
			// If there are less than `maxTxnOps` keyspaces,
```

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -286,33 +290,70 @@ func (m *GroupManager) GetKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGroup,
...
-			return nil
-		}
-		if kg.IsSplitting() {
-			return ErrKeyspaceGroupInSplit
+			return nil, err
 		}
-		return m.store.DeleteKeyspaceGroup(txn, id)
-	}); err != nil {
-		return nil, err
+		m.groups[endpoint.StringUserKind(kg.UserKind)].Put(kg)
```
binshi-bing, line 354: Is there any keyspace meta in memory? If yes, we need to update it too -- just like we update keyspace group meta in memory here.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -286,33 +290,70 @@ func (m *GroupManager) GetKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGroup,
...
+				keyspaceIDsToUnlock = make([]uint32, 0, maxTxnOps)
+				deleteKeyspaceCount = 0
+			)
+			defer func() {
+				for _, id := range keyspaceIDsToUnlock {
+					m.km.metaLock.Unlock(id)
+				}
+			}()
+			for _, id := range kg.Keyspaces {
+				if deleteKeyspaceCount >= maxTxnOps {
```
binshi-bing, line 323: How do we count txn ops? For every increment of deleteKeyspaceCount, given the fact that it calls m.km.store.LoadKeyspaceMeta(txn, id) and err = m.km.store.SaveKeyspaceMeta(txn, ks), it's counted two txn ops or just one (only write ops)? If it's counted two txn ops, the transaction size will be double of maxTxnOps

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -286,33 +290,70 @@ func (m *GroupManager) GetKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGroup,
 // DeleteKeyspaceGroupByID deletes the keyspace group by ID.
 func (m *GroupManager) DeleteKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGroup, error) {
```
binshi-bing, line 291: DeleteKeyspaceGroupByID() can't work alone without timeline merge logic to avoid time fallback, correct?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -286,33 +290,70 @@ func (m *GroupManager) GetKeyspaceGroupByID(id uint32) (*endpoint.KeyspaceGroup,
...
+				deleteKeyspaceCount = 0
+			)
+			defer func() {
+				for _, id := range keyspaceIDsToUnlock {
+					m.km.metaLock.Unlock(id)
+				}
+			}()
+			for _, id := range kg.Keyspaces {
+				if deleteKeyspaceCount >= maxTxnOps {
+					moreToDelete = true
```
binshi-bing, line 324:
```suggestion
					moreToDelete = deleteKeyspaceCount < len(kg.Keyspaces)
```
Otherwise, if deleteKeyspaceCount == maxTxnOps, we still need to run one more unnecessary round.

binshi-bing: And if we don't use "moreToDelete = deleteKeyspaceCount < len(kg.Keyspaces)",  actually this assignment is no need because it must already be true here.

rleungx:
I'm going to close this PR since we start using merge to solve the fallback problem which is more straightforward.

PR #6568, opened 2023-06-07
*: fix some log issues

rleungx:

Issue Number: Ref #5895.

Introduced by https://github.com/tikv/pd/issues/6390

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/utils/etcdutil/etcdutil.go
```diff
@@ -592,7 +594,6 @@ func (lw *LoopWatcher) load(ctx context.Context) (nextRevision int64, err error)
 				log.Error("run post event failed in watch loop", zap.String("name", lw.name),
 					zap.String("key", lw.key), zap.Error(err))
 			}
-			log.Info("load finished in watch loop", zap.String("name", lw.name), zap.String("key", lw.key))
```
lhy1024, line 595: Why remove it

rleungx: There is a duplicated log, you can search it.

PR #6572, opened 2023-06-08
Simplify TSO Proxy implementation by using one forward stream for one gPRC stream

### LINKED ISSUE #6549: The api leader got stuck at tso requests forwarding

binshi-bing:

Issue Number: Close #6549

```commit-message
Simplify tso proxy implementation by using one forward stream for one grpc.ServerStream.
#6565 is a longer term solution for both follower batching and tso microservice. 
It's well implemented, but just need more time to bake, and we need a short term workable solution for now.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -406,6 +406,9 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 		lastForwardedHost string
 	)
 	defer func() {
+		if forwardStream != nil {
+			forwardStream.CloseSend()
+		}
 		// cancel the forward stream
 		if cancel != nil {
 			cancel()
@@ -438,6 +441,9 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 			return errors.WithStack(ErrNotFoundTSOAddr)
 		}
 		if forwardStream == nil || lastForwardedHost != forwardedHost {
+			if forwardStream != nil {
+				forwardStream.CloseSend()
+			}
 			if cancel != nil {
 				cancel()
 			}
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -64,6 +64,7 @@ var (
 	ErrNotStarted           = status.Errorf(codes.Unavailable, "server not started")
 	ErrSendHeartbeatTimeout = status.Errorf(codes.DeadlineExceeded, "send heartbeat timeout")
 	ErrNotFoundTSOAddr      = status.Errorf(codes.NotFound, "not found tso address")
+	ErrForwardTSOTimeout    = status.Errorf(codes.DeadlineExceeded, "forward tso request timeout")
 )
 
 // GrpcServer wraps Server to provide grpc service.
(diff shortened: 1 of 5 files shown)


server/grpc_service.go
```diff
@@ -398,6 +397,120 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
...
+		}
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		if request.GetCount() == 0 {
+			err = errs.ErrGenerateTimestamp.FastGenByArgs("tso count should be positive")
+			return status.Errorf(codes.Unknown, err.Error())
+		}
+
+		forwardedHost, ok := s.GetServicePrimaryAddr(stream.Context(), utils.TSOServiceName)
```
rleungx, line 429: Do we need to handle not leader errors?

binshi-bing: handled.

lhy1024 (approved):

rleungx (approved):

Issue #6573, opened 2023-06-08
Too manly orphan peers cannot be remove


nolouch:
### What your problem?
I see a region has many learners:
```
{
  "id": 353747615,
  "start_key": "7480000000000000FFC85F698000000000FF000003016D656469FF6173746FFF726568FF6F7573652EFF636FFF6D2E61750000FD01FF5330363739316633FFFF62323364363462FF64FF633837656235FF3737FF6238333930FF396561FF36000000FF00000000F8000000FC",
  "end_key": "7480000000000000FFC85F698000000000FF000003016D656469FF6173746FFF726568FF6F7573652EFF636FFF6D2E61750000FD01FF8537323965353338FFFF65626131363466FF34FF323966393161FF6339FF3635353364FF626138FF33000000FF00000000F8000000FC",
  "epoch": {
    "conf_ver": 368,
    "version": 32492
  },
  "peers": [
    {
      "id": 353747616,
      "store_id": 350571036,
      "role_name": "Voter"
    },
    {
      "id": 353747617,
      "store_id": 350571030,
      "role_name": "Voter"
    },
    {
      "id": 353747618,
      "store_id": 350571032,
      "role_name": "Voter"
    },
    {
      "id": 391098694,
      "store_id": 350571034,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 433406127,
      "store_id": 350571134,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 522453028,
      "store_id": 350571031,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 544524983,
      "store_id": 350571762,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 556827866,
      "store_id": 111263,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 573463879,
      "store_id": 350614150,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 588797740,
      "store_id": 350571033,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 603872787,
      "store_id": 350571029,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
    },
    {
      "id": 613045451,
      "store_id": 350571028,
      "role": 1,
      "role_name": "Learner",
      "is_learner": true
[...]

PR #6574, opened 2023-06-08
checker: fix the too many orphan peers cannot be removed

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/6573 

```commit-message
rule-checker: fix the too many orphan peers that cannot be removed
- let the health peer can be removed once there exist redundant
```

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue too many peers/learners cannot be removed
```

### DIFF
--- a/pkg/schedule/checker/rule_checker_test.go
+++ b/pkg/schedule/checker/rule_checker_test.go
@@ -168,6 +168,39 @@ func (suite *ruleCheckerTestSuite) TestFixOrphanPeers() {
 	suite.Equal(uint64(4), op.Step(0).(operator.RemovePeer).FromStore)
 }
 
+func (suite *ruleCheckerTestSuite) TestFixToManyOrphanPeers() {
+	suite.cluster.AddLeaderStore(1, 1)
+	suite.cluster.AddLeaderStore(2, 1)
+	suite.cluster.AddLeaderStore(3, 1)
+	suite.cluster.AddLeaderStore(4, 1)
+	suite.cluster.AddLeaderStore(5, 1)
+	suite.cluster.AddLeaderStore(6, 1)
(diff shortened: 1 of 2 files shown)


rleungx (approved):

pkg/schedule/checker/rule_checker_test.go
```diff
@@ -168,6 +168,39 @@ func (suite *ruleCheckerTestSuite) TestFixOrphanPeers() {
...
 }
 
+func (suite *ruleCheckerTestSuite) TestFixToManyOrphanPeers() {
+	suite.cluster.AddLeaderStore(1, 1)
+	suite.cluster.AddLeaderStore(2, 1)
+	suite.cluster.AddLeaderStore(3, 1)
+	suite.cluster.AddLeaderStore(4, 1)
+	suite.cluster.AddLeaderStore(5, 1)
+	suite.cluster.AddLeaderStore(6, 1)
+	suite.cluster.AddRegionWithLearner(1, 1, []uint64{2, 3}, []uint64{4, 5, 6})
```
CabinfeverB, line 178: This comment should have nothing to do with this PR. But I want to know what happen if here is `suite.cluster.AddRegionWithLearner(1, 1, []uint64{2, 3}, []uint64{4})`, and other behaviors are same.

CabinfeverB: It seems not return operator

CabinfeverB (approved):

pkg/schedule/checker/rule_checker_test.go
nolouch: yes, because need len(`OrphanPeers`) >=2

CabinfeverB (approved):

PR #6578, opened 2023-06-09
keyspace: update keyspace config TSOKeyspaceGroupIDKey during the split

JmPotato:

Issue Number: ref #6232.

```commit-message
Update keyspace config `TSOKeyspaceGroupIDKey` during the split.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -71,6 +71,8 @@ type GroupManager struct {
 	serviceRegistryMap map[string]string
 	// tsoNodesWatcher is the watcher for the registered tso servers.
 	tsoNodesWatcher *etcdutil.LoopWatcher
+
+	km *Manager
 }
 
 // NewKeyspaceGroupManager creates a Manager of keyspace group related data.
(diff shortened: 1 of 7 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -552,6 +556,29 @@ func (m *GroupManager) SplitKeyspaceGroupByID(splitSourceID, splitTargetID uint3
...
 			return err
 		}
+		// Iterate the keyspaces to update the config key.
+		keyspaceIDsToUnlock := make([]uint32, 0, len(keyspaces))
+		defer func() {
+			for _, id := range keyspaceIDsToUnlock {
+				m.km.metaLock.Unlock(id)
+			}
+		}()
+		for _, keyspace := range keyspaces {
```
binshi-bing, line 566: SaveKeyspaceMeta() needs to be called in batches too, otherwise the len(keyspaces) in a split needs to be <= maxTxnOps which isn't practical even in dev with tens thousands of keyspaces. Adding batching in another place increases complexity again.

Let's discuss about it today. @rleungx

JmPotato:
Close since the flag `TSOKeyspaceGroupIDKey` will be ignored for now after the discussion.

JmPotato:
Close since the flag `TSOKeyspaceGroupIDKey` will be ignored for now after the discussion.

PR #6584, opened 2023-06-09
tests: make TestSplitKeyspaceGroup stable

### LINKED ISSUE #6571: TestSplitKeyspaceGroup "Condition never satisfied" randomly
### Which jobs are failing
--- FAIL: TestSplitKeyspaceGroup (44.24s)
    testutil.go:63: 
        	Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:63
        	            				/home/runner/work/pd/pd/tests/pdctl/keyspace/keyspace_group_test.go:124
        	Error:      	Condition never satisfied
        	Test:       	TestSplitKeyspaceGroup

### CI link
https://github.com/tikv/pd/actions/runs/5203268375/jobs/9385886991?pr=6565

unknown

lhy1024:

Issue Number: Close #6571

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -169,9 +169,6 @@ func (m *GroupManager) allocNodesToAllKeyspaceGroups(ctx context.Context) {
 		case <-ctx.Done():
 			log.Info("stop to alloc nodes to all keyspace groups")
 			return
-		case <-m.ctx.Done():
-			log.Info("stop to alloc nodes to all keyspace groups")
-			return
 		case <-ticker.C:
 		}
 		countOfNodes := m.GetNodesCount()
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -110,7 +110,7 @@ func NewKeyspaceGroupManager(
 }
 
 // Bootstrap saves default keyspace group info and init group mapping in the memory.
-func (m *GroupManager) Bootstrap() error {
+func (m *GroupManager) Bootstrap(ctx context.Context) error {
 	// Force the membership restriction that the default keyspace must belong to default keyspace group.
 	// Have no information to specify the distribution of the default keyspace group replicas, so just
 	// leave the replica/member list empty. The TSO service will assign the default keyspace group replica
@@ -143,7 +143,7 @@ func (m *GroupManager) Bootstrap() error {
 	// It will only alloc node when the group manager is on API leader.
 	if m.client != nil {
 		m.wg.Add(1)
-		go m.allocNodesToAllKeyspaceGroups()
+		go m.allocNodesToAllKeyspaceGroups(ctx)
 	}
 	return nil
 }
@@ -154,7 +154,7 @@ func (m *GroupManager) Close() {
 	m.wg.Wait()
 }
 
-func (m *GroupManager) allocNodesToAllKeyspaceGroups() {
+func (m *GroupManager) allocNodesToAllKeyspaceGroups(ctx context.Context) {
 	defer logutil.LogPanic()
 	defer m.wg.Done()
 	ticker := time.NewTicker(allocNodesToKeyspaceGroupsInterval)
@@ -166,6 +166,9 @@ func (m *GroupManager) allocNodesToAllKeyspaceGroups() {
 	log.Info("start to alloc nodes to all keyspace groups")
 	for {
 		select {
+		case <-ctx.Done():
+			log.Info("stop to alloc nodes to all keyspace groups")
+			return
 		case <-m.ctx.Done():
 			log.Info("stop to alloc nodes to all keyspace groups")
 			return
@@ -355,27 +358,6 @@ func (m *GroupManager) saveKeyspaceGroups(keyspaceGroups []*endpoint.KeyspaceGro
 	})
 }
 
-// saveGroupWithUpdateKeyspace will try to save the given keyspace groups into the storage.
-// It only update the keyspace field for the keyspace group.
-func (m *GroupManager) saveGroupWithUpdateKeyspace(keyspaceGroup *endpoint.KeyspaceGroup) error {
-	return m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
-		oldKG, err := m.store.LoadKeyspaceGroup(txn, keyspaceGroup.ID)
-		if err != nil {
-			return err
-		}
-		if oldKG.IsSplitting() {
-			return ErrKeyspaceGroupInSplit
-		}
-		newKG := oldKG
-		newKG.Keyspaces = keyspaceGroup.Keyspaces
-		err = m.store.SaveKeyspaceGroup(txn, newKG)
-		if err != nil {
-			return err
-		}
-		return nil
-	})
-}
-
 // GetKeyspaceConfigByKind returns the keyspace config for the given user kind.
 func (m *GroupManager) GetKeyspaceConfigByKind(userKind endpoint.UserKind) (map[string]string, error) {
 	// when server is not in API mode, we don't need to return the keyspace config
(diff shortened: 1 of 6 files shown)


lhy1024:
it is stable enough.

https://github.com/tikv/pd/actions/runs/5224016809/jobs/9431687002
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9431895862
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9432183860
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9432369885
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9432369885
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9432557465
https://github.com/tikv/pd/actions/runs/5224016809/jobs/9432742936
https://github.com/tikv/pd/actions/runs/5224583895/jobs/9432950984

tests/pdctl/keyspace/keyspace_group_test.go
```diff
@@ -96,7 +96,7 @@ func TestSplitKeyspaceGroup(t *testing.T) {
 	re.NoError(failpoint.Enable("github.com/tikv/pd/pkg/keyspace/acceleratedAllocNodes", `return(true)`))
 	re.NoError(failpoint.Enable("github.com/tikv/pd/server/delayStartServerLoop", `return(true)`))
 	keyspaces := make([]string, 0)
-	for i := 0; i < 500; i++ {
+	for i := 0; i < 129; i++ { // 128 is the default max txn ops limit in etcd.
```
binshi-bing, line 99: If we use 129 here, we'll add 129 txn ops instead of 128, right?

binshi-bing: you add 0, 1, ..., 128 keypsace here, 129 in total which will exceed the default max txn ops limit in etcd.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -352,6 +360,32 @@ func (m *GroupManager) saveKeyspaceGroups(keyspaceGroups []*endpoint.KeyspaceGro
...
+		oldKG, err := m.store.LoadKeyspaceGroup(txn, keyspaceGroup.ID)
+		if err != nil {
+			return err
+		}
+		if oldKG.IsSplitting() {
+			return ErrKeyspaceGroupInSplit
+		}
+		newKG := oldKG
+		newKG.Keyspaces = keyspaceGroup.Keyspaces
+		if oldKG.IsSplitting() {
```
binshi-bing, line 376: It has already checked that "if oldKG.IsSplitting() is true and exit on condition met" at line 371 above.
So the check here is always true.

binshi-bing: It looks like we have conflict logic here -- if oldKG.IsSplitting() is true, do we want to return ErrKeyspaceGroupInSplit, or want to set newKG.SplitState and proceed?

binshi-bing: The whole function doesn't provide CAS semantics, which is my real concern. 
Before the caller calls saveGroupWithUpdateKeyspace, the caller must have loaded the keyspace group info from the etcd, then do some calculation and update keyspace list in the memory, finally call saveGroupWithUpdateKeyspace here. If somehow the latest keyspace list loaded here is different from the original one, it will overwrite something and cause inconsistency. Two ways to solve it:
1. saveGroupWithUpdateKeyspace should check if the latest keyspace group info loaded here has the same keypace list as the original one which can pass to this function.
2. Just pass two lists to saveGroupWithUpdateKeyspace -- one keyspace list to add and one keyspace list to delete, and this function just needs to add if it doesn't exist or delete if it exists based on the latest loaded keyspace list here.

binshi-bing (changes requested):
left some comments.

tests/pdctl/keyspace/keyspace_group_test.go
lhy1024: yes, test the case which exceed the default max txn ops limit in etcd

pkg/keyspace/tso_keyspace_group.go
lhy1024: saveGroupWithUpdateKeyspace is a part of saveKeyspaceGroups, and only update keyspace list param

lhy1024: because updateKeyspaceForGroupLocked load kg from memory rather than etcd, so we should not update other params besides keyspace list.

lhy1024: keyspace check is in updateKeyspaceForGroupLocked

lhy1024:
cc @binshi-bing @JmPotato

JmPotato (approved):

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -352,6 +355,27 @@ func (m *GroupManager) saveKeyspaceGroups(keyspaceGroups []*endpoint.KeyspaceGro
 	})
 }
 
+// saveGroupWithUpdateKeyspace will try to save the given keyspace groups into the storage.
+// It only update the keyspace field for the keyspace group.
+func (m *GroupManager) saveGroupWithUpdateKeyspace(keyspaceGroup *endpoint.KeyspaceGroup) error {
```
rleungx, line 360: If we only run it on the leader node, do we still need it?

lhy1024: I think it is still necessary because we provide an external interface to set node for keyspace groups.

rleungx: But it can be only executed by the leader?

lhy1024: Yes, but `UpdateKeyspaceForGroup` is called by manager of keyspace. 

For example, when manager of keyspace call `UpdateKeyspaceForGroup`, but we exec `set node` before `updateKeyspaceForGroupLocked`, which is similiar with [failpoint.Inject](https://github.com/tikv/pd/blob/fe58ed103afc060b97f6884c71085d78b12fe5bc/pkg/keyspace/tso_keyspace_group.go#L418).

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -137,6 +140,11 @@ func (m *GroupManager) Bootstrap() error {
 		m.groups[userKind].Put(group)
 	}
 
+	// It will only alloc node when the group manager is on API leader.
+	if m.client != nil {
+		m.wg.Add(1)
+		go m.allocNodesToAllKeyspaceGroups()
```
rleungx, line 146: If we transfer the leader, will it exit?

pkg/keyspace/tso_keyspace_group.go
rleungx: We have a lock. Does it prevent this thing from happening?

tests/pdctl/keyspace/keyspace_group_test.go
binshi-bing: oh, could add this statement in comment? Without this statement, the code looks like have bug.

binshi-bing (approved):
LGTM

pkg/keyspace/tso_keyspace_group.go
lhy1024: yes, updated, PTAL

pkg/keyspace/tso_keyspace_group.go
lhy1024: Thanks, added ctx

lhy1024:
@rleungx PTAL

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -158,6 +166,9 @@ func (m *GroupManager) allocNodesToAllKeyspaceGroups() {
 	log.Info("start to alloc nodes to all keyspace groups")
 	for {
 		select {
+		case <-ctx.Done():
+			log.Info("stop to alloc nodes to all keyspace groups")
+			return
 		case <-m.ctx.Done():
```
rleungx, line 172: Do we still need it?

rleungx (approved):

PR #6585, opened 2023-06-09
Improve tso proxy reliability

binshi-bing:

Issue Number: Ref #5895

```commit-message
Improve tso proxy reliability.

1. Add protection mechanisms to TSO Proxy.
    a. Throttle the concurrency of TSO Proxy streamings. Default 5000.
    b. If TSO Proxy didn't receive the TSO request from the client for 1 hour, close the stream.
2. Optimize forceLoad lock with RW lock.
3. Enable stress test.
4. Add deadline for API leader forwarding request to TSO service.
5. Make tso response channel more safely.
6. Move tso proxy stress test away from the test suite as it has impact on other test cases.
7. Fix grpc client connection pool (server side) resource leak problem.
8. Make MaxConcurrentTSOProxyStreamings (5000 as default) and TSOProxyClientRecvTimeout (1 hour as default) configurable.
9. Add metrics tsoProxyHandleDuration, tsoProxyBatchSize and tsoProxyForwardTimeoutCounter.
    
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -80,6 +80,15 @@ type Config struct {
 	LogFileDeprecated  string `toml:"log-file" json:"log-file,omitempty"`
 	LogLevelDeprecated string `toml:"log-level" json:"log-level,omitempty"`
 
+	// MaxConcurrentTSOProxyStreamings is the maximum number of concurrent TSO proxy streaming process routines allowed.
+	// Exceeding this limit will result in an error being returned to the client when a new client starts a TSO streaming.
+	// Set this to 0 will disable TSO Proxy.
+	// Set this to the negative value to disable the limit.
+	MaxConcurrentTSOProxyStreamings int `toml:"max-concurrent-tso-proxy-streamings" json:"max-concurrent-tso-proxy-streamings"`
+	// TSOProxyClientRecvTimeout is the timeout for the TSO proxy to receive a tso request from a client via grpc TSO stream.
+	// After the timeout, the TSO proxy will close the grpc TSO stream.
+	TSOProxyClientRecvTimeout typeutil.Duration `toml:"tso-proxy-client-recv-timeout" json:"tso-proxy-client-recv-timeout"`
+
 	// TSOSaveInterval is the interval to save timestamp.
 	TSOSaveInterval typeutil.Duration `toml:"tso-save-interval" json:"tso-save-interval"`
 
@@ -218,6 +227,9 @@ const (
 	defaultDashboardAddress = "auto"
 
 	defaultDRWaitStoreTimeout = time.Minute
+
+	defaultMaxConcurrentTSOProxyStreamings = 5000
+	defaultTSOProxyClientRecvTimeout       = 1 * time.Hour
 
 	defaultTSOSaveInterval = time.Duration(defaultLeaderLease) * time.Second
 	// defaultTSOUpdatePhysicalInterval is the default value of the config `TSOUpdatePhysicalInterval`.
@@ -442,10 +454,11 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
 		}
 	}
 
+	configutil.AdjustInt(&c.MaxConcurrentTSOProxyStreamings, defaultMaxConcurrentTSOProxyStreamings)
+	configutil.AdjustDuration(&c.TSOProxyClientRecvTimeout, defaultTSOProxyClientRecvTimeout)
+
 	configutil.AdjustInt64(&c.LeaderLease, defaultLeaderLease)
-
 	configutil.AdjustDuration(&c.TSOSaveInterval, defaultTSOSaveInterval)
-
 	configutil.AdjustDuration(&c.TSOUpdatePhysicalInterval, defaultTSOUpdatePhysicalInterval)
 
 	if c.TSOUpdatePhysicalInterval.Duration > maxTSOUpdatePhysicalInterval {
@@ -1250,6 +1263,17 @@ func (c *Config) GetLeaderLease() int64 {
 // IsLocalTSOEnabled returns if the local TSO is enabled.
 func (c *Config) IsLocalTSOEnabled() bool {
 	return c.EnableLocalTSO
+}
+
+// GetMaxConcurrentTSOProxyStreamings returns the max concurrent TSO proxy streamings.
+// If the value is negative, there is no limit.
+func (c *Config) GetMaxConcurrentTSOProxyStreamings() int {
+	return c.MaxConcurrentTSOProxyStreamings
+}
+
+// GetTSOProxyClientRecvTimeout returns the TSO proxy client receive timeout.
+func (c *Config) GetTSOProxyClientRecvTimeout() time.Duration {
+	return c.TSOProxyClientRecvTimeout.Duration
 }
 
 // GetTSOUpdatePhysicalInterval returns TSO update physical interval.
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -50,26 +50,31 @@ import (
 )
 
 const (
-	heartbeatSendTimeout          = 5 * time.Second
-	maxRetryTimesRequestTSOServer = 3
-	retryIntervalRequestTSOServer = 500 * time.Millisecond
-	getMinTSFromTSOServerTimeout  = 1 * time.Second
+	heartbeatSendTimeout                   = 5 * time.Second
+	maxRetryTimesRequestTSOServer          = 3
+	retryIntervalRequestTSOServer          = 500 * time.Millisecond
+	getMinTSFromTSOServerTimeout           = 1 * time.Second
+	defaultMaxConcurrentTSOProxyStreamings = 5000
+	defaultTSOProxyClientRecvTimeout       = 1 * time.Hour
 )
 
 // gRPC errors
 var (
 	// ErrNotLeader is returned when current server is not the leader and not possible to process request.
 	// TODO: work as proxy.
-	ErrNotLeader            = status.Errorf(codes.Unavailable, "not leader")
-	ErrNotStarted           = status.Errorf(codes.Unavailable, "server not started")
-	ErrSendHeartbeatTimeout = status.Errorf(codes.DeadlineExceeded, "send heartbeat timeout")
-	ErrNotFoundTSOAddr      = status.Errorf(codes.NotFound, "not found tso address")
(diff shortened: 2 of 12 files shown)


server/grpc_service.go
```diff
@@ -513,13 +506,92 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
 
+func (s *GrpcServer) forwardTSORequestWithDeadLine(
+	ctx context.Context, request *pdpb.TsoRequest, forwardStream tsopb.TSO_TsoClient,
+) (*tsopb.TsoResponse, error) {
+	// Create a context with deadline for forwarding TSO request to TSO service.
+	ctxTimeout, cancel := context.WithTimeout(ctx, tsoutil.DefaultTSOProxyTimeout)
+	defer cancel()
+
+	// used to receive the result from doSomething function
+	tsoRespCh := make(chan *tsopbTSOResponse)
```
rleungx, line 517: Should we use the block channel?

server/grpc_service.go
```diff
@@ -513,13 +506,92 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 	}
 }
 
+func (s *GrpcServer) forwardTSORequestWithDeadLine(
```
rleungx, line 509: Please add `defer logutil.LogPanic()` for the goroutine.

tests/integrations/mcs/Makefile
```diff
@@ -32,7 +32,7 @@ test: failpoint-enable
 	$(MAKE) failpoint-disable
 
 ci-test-job:
-	CGO_ENABLED=1 go test ./... -tags deadlock -race -covermode=atomic -coverprofile=covprofile -coverpkg=$(ROOT_PATH)/... github.com/tikv/pd/tests/integrations/mcs
+	CGO_ENABLED=1 go test ./... -timeout=15m -tags deadlock -race -covermode=atomic -coverprofile=covprofile -coverpkg=$(ROOT_PATH)/... github.com/tikv/pd/tests/integrations/mcs
```
lhy1024, line 35: why change it?

server/grpc_service.go
```diff
@@ -513,13 +506,92 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
+	case <-ctxTimeout.Done():
+		return
+	default:
+	}
+
+	response, err := forwardStream.Recv()
+	if err != nil {
+		if strings.Contains(err.Error(), errs.NotLeaderErr) {
+			s.tsoPrimaryWatcher.ForceLoad()
+		}
```
lhy1024, line 562: Do we need to retry when meet this error?

tests/integrations/mcs/tso/proxy_test.go
```diff
@@ -75,17 +77,12 @@ func (s *tsoProxyTestSuite) SetupSuite() {
...
 
-	// Create some TSO client streams with the same context.
-	s.grpcClientConns, s.streams, s.cancelFuncs = createTSOStreams(re, s.ctx, s.backendEndpoints, 100, true)
-	// Create some TSO client streams with the different context.
-	grpcClientConns, streams, cancelFuncs := createTSOStreams(re, s.ctx, s.backendEndpoints, 100, false)
-	s.grpcClientConns = append(s.grpcClientConns, grpcClientConns...)
-	s.streams = append(s.streams, streams...)
-	s.cancelFuncs = append(s.cancelFuncs, cancelFuncs...)
+	// Create some TSO client streams with different context.
+	s.streams, s.cleanupFuncs = createTSOStreams(re, s.ctx, s.backendEndpoints, 200)
```
lhy1024, line 81: different context.?

lhy1024 (commented):
mostly LGTM

tests/integrations/mcs/Makefile
binshi-bing: previously mcs/tso needs over 8 mins. After I added tso_proxy_test.go, it needs a little more than 10 minutes. We might need to split mcs/tso into multiple ci chunks later. In this pr, I just simply change it to 15 minutes.

server/config/config.go
```diff
@@ -80,6 +80,14 @@ type Config struct {
 	LogFileDeprecated  string `toml:"log-file" json:"log-file,omitempty"`
 	LogLevelDeprecated string `toml:"log-level" json:"log-level,omitempty"`
 
+	// MaxConcurrentTSOProxyStreamings is the maximum number of concurrent TSO proxy streaming process routines allowed.
+	// Exceeding this limit will result in an error being returned to the client when a new client starts a TSO streaming.
+	// Set this to 0 will disable TSO Proxy.
+	// Set this to the negative value to disable the limit.
+	MaxConcurrentTSOProxyStreamings int `toml:"max-concurrent-tso-proxy-streamings" json:"max-concurrent-tso-proxy-streamings"`
+	// TSOProxyClientRecvTimeout is the timeout for the TSO proxy to receive a tso request from a client via grpc TSO stream.
+	TSOProxyClientRecvTimeout typeutil.Duration `toml:"tso-proxy-client-recv-timeout" json:"tso-proxy-client-recv-timeout"`
```
lhy1024, line 89: It needs some additional steps to add new config. If no necessary, I think it is better to not add it.

server/grpc_service.go
binshi-bing: You're right, we should use tsoRespCh := make(chan *tsopbTSOResponse, 1)

server/grpc_service.go
binshi-bing: added

server/grpc_service.go
binshi-bing: It will be retried anyway. Suppose this error happens, many forward streams will encounter this problem and report the error to the client. The client will retry to create tso stream again and send tso request, if tso server hasn't elected new primary, the NotLeaderErr will trigger force load again every 200 ms.

tests/integrations/mcs/tso/proxy_test.go
binshi-bing: Yes, inside of createTSOStreams, it will use different context to create "stream, err := grpcPDClient.Tso(cctx)" PD_TsoClient. The result is that the stream.Context() on the server side is different.

server/config/config.go
binshi-bing: What's the additional steps? Xiaoguang reviewed this pr and asked for configuration. Do we have better way to add configuration?

binshi-bing: Per offline with @lhy1024 , we need a config review to review the config change, and @rleungx should be the right one.

lhy1024 (approved):

server/config/config.go
rleungx: IMO, we don't need these configs since they will be replaced if we implement the discovery in other clients. Just using a constant is enough.

tests/integrations/mcs/tso/proxy_test.go
```diff
@@ -128,42 +123,93 @@ func (s *tsoProxyTestSuite) TestTSOProxyWorksWithCancellation() {
...
-		cancelFuncs = append(cancelFuncs, cancelFuncsTemp...)
-		s.verifyTSOProxy(streams, 50, false)
+		cleanupFuncs = append(cleanupFuncs, cleanupFuncsTemp...)
+		s.verifyTSOProxy(ctxTimeout, streams, cleanupFuncs, 50, false)
+	}
+	s.cleanupGRPCStreams(cleanupFuncs)
+	log.Info("the stress test completed.")
+
+	// Wait for the TSO Proxy to recover from the stress.
+	time.Sleep(recoverySLA)
```
rleungx, line 159: Can we make it shorter?

server/grpc_service.go
```diff
@@ -513,13 +507,101 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
 
+func (s *GrpcServer) forwardTSORequestWithDeadLine(
+	ctx context.Context, request *pdpb.TsoRequest, forwardStream tsopb.TSO_TsoClient,
+) (*tsopb.TsoResponse, error) {
+	defer logutil.LogPanic()
+	// Create a context with deadline for forwarding TSO request to TSO service.
+	ctxTimeout, cancel := context.WithTimeout(ctx, tsoutil.DefaultTSOProxyTimeout)
+	defer cancel()
+
+	tsoProxyBatchSize.Observe(float64(request.GetCount()))
```
rleungx, line 518: Do we need this metrics?

rleungx (commented):
The rest LGTM

server/grpc_service.go
binshi-bing: Yes? to be consistent with the tso dispaching, otherwise this metrics will shows no value which might cause confusing.

binshi-bing: This is exsitng metrics, so it's meaningful even just show constant value 1 instead of 0 or no value on the dashboard.

tests/integrations/mcs/tso/proxy_test.go
binshi-bing: This is just test to make the test more stable.

server/config/config.go
binshi-bing: I think we need if we can update config in the dev/staging/prod quickly without making code change, cherry-pick, hot fix and deployment.

tests/integrations/mcs/tso/proxy_test.go
rleungx: How about using something like eventually to make it as quick as possible?

server/grpc_service.go
rleungx: We don't have the batch logic here, so I think it's unnecessary.

server/grpc_service.go
binshi-bing: added

server/grpc_service.go
binshi-bing: changed

tests/integrations/mcs/tso/proxy_test.go
binshi-bing: Ok, changed to 
	// Verify the TSO Proxy can still work correctly after the stress.
		err := s.verifyTSOProxy(s.ctx, s.streams, s.cleanupFuncs, 1, false)
		return err == nil
	}, testutil.WithWaitFor(recoverySLA), testutil.WithTickInterval(500*time.Millisecond))

rleungx (approved):

Issue #6589, opened 2023-06-13
Implement the merging operation of Keyspace Group


JmPotato:
## Development Task

As we now support the split operation of Keyspace Group, it is also necessary to provide reverse merge operations, which can help us better manage the number of Keyspace Groups and do some rollback operations.

To implement the merging operation of Keyspace Group, it may contain the following process:

1. Put all merging groups into the merging state.
	a. Once the groups that need to be merged watch this change, it will stop serving any TSO request to the outside.
    b. Once the merging target groups watch this change, it will start the merging process described as follows.
3. The merging target server sends the merging gRPC requests to all group primaries which need to be merged.
4. Once a primary received the merging request, it will:
	a. Stop serving any TSO request to the outside if it has not been performed step 1.a yet.
	b. Return the current TSO as the gRPC response.
5. The merging target server keeps trying step 3 until all the primaries responded, and then:
	a. Set its current TSO to the biggest one among all the primary responses.
	b. Ask the PD leader to delete these groups from the meta and move the keyspaces.

binshi-bing:
What if the primary of a keyspace group, which has already stopped the service and responded the PD leader, transferred to another server before 4.a. Without persistent info, will the new primary of the keyspace group still serve the requests before 4.a?

binshi-bing:
To solve the above problem, can we have the following steps?

1. Once the default keyspace group watches this change, it deletes all non-default keyspace groups in etcd.
2. The keyspace group managers on tso servers watch this change and handle the DELETE events to stop serving these keyspace groups.
3.  Once the default keyspace group watches all primaries of non-default keyspace groups are gone, it reads the timestamps saved in etcd of all non-default keyspace groups, reset own timestmp with the max(all non-default persisted timestamps) + a time window.

binshi-bing:
We already have step 2 implemented. We also have part of 3 implemented and the other part has the similar logic and code.

PR #6594, opened 2023-06-13
keyspace, apiv2: implement the keyspace group merging API

JmPotato:

Issue Number: ref #6589.

```commit-message
Implement the keyspace group merging API.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -49,10 +49,10 @@ const (
 	UserKindKey = "user_kind"
 	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
 	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
-	// keyspacePatrolBatchSize is the batch size for keyspace assignment patrol.
-	// the limit of etcd txn op is 128, keyspacePatrolBatchSize need to be less than it.
+	// maxEtcdTxnOps is the batch size for operating etcd. The limit of etcd txn op is 128.
+	// We use 120 here to leave some space for other operations.
 	// See: https://github.com/etcd-io/etcd/blob/d3e43d4de6f6d9575b489dd7850a85e37e0f6b6c/server/embed/config.go#L61
-	keyspacePatrolBatchSize = 120
+	maxEtcdTxnOps = 120
 )
 
 // Config is the interface for keyspace config.
@@ -669,7 +669,7 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 			zap.Duration("cost", time.Since(start)),
 			zap.Uint64("patrolled-keyspace-count", patrolledKeyspaceCount),
 			zap.Uint64("assigned-keyspace-count", assignedKeyspaceCount),
-			zap.Int("batch-size", keyspacePatrolBatchSize),
+			zap.Int("batch-size", maxEtcdTxnOps),
 			zap.Uint32("current-start-id", currentStartID),
 			zap.Uint32("next-start-id", nextStartID),
 		)
@@ -691,7 +691,7 @@ func (manager *Manager) PatrolKeyspaceAssignment() error {
 			if defaultKeyspaceGroup.IsMerging() {
 				return ErrKeyspaceGroupInMerging
 			}
-			keyspaces, err := manager.store.LoadRangeKeyspace(txn, manager.nextPatrolStartID, keyspacePatrolBatchSize)
+			keyspaces, err := manager.store.LoadRangeKeyspace(txn, manager.nextPatrolStartID, maxEtcdTxnOps)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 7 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -743,3 +758,103 @@ func (m *GroupManager) IsExistNode(addr string) bool {
...
+				return ErrKeyspaceGroupInMerging
+			}
+			groups[kgID] = kg
+		}
+		mergeTargetKg = groups[mergeTargetID]
+		// Delete the keyspace groups in merge list and move the keyspaces in it to the target keyspace group.
+		for _, kgID := range mergeList {
+			kg := groups[kgID]
+			for _, keyspace := range kg.Keyspaces {
+				if !slice.Contains(mergeTargetKg.Keyspaces, keyspace) {
```
binshi-bing, line 799: Do we need to optimize this op with O(logn) algorithm?

binshi-bing: Probably we can put mergeTargetKg.Keyspaces into a map first, then here add keyspace into the map.
Eventually append all elements in the map to mergeTargetKg.Keyspaces.

binshi-bing: We could have long keyspace list, so it is mostly necessary.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -743,3 +758,103 @@ func (m *GroupManager) IsExistNode(addr string) bool {
...
+		return m.store.SaveKeyspaceGroup(txn, mergeTargetKg)
+	}); err != nil {
+		return err
+	}
+	// Update the keyspace group cache.
+	m.groups[endpoint.StringUserKind(mergeTargetKg.UserKind)].Put(mergeTargetKg)
+	for _, kgID := range mergeList {
+		kg := groups[kgID]
+		m.groups[endpoint.StringUserKind(kg.UserKind)].Remove(kgID)
+	}
```
binshi-bing, line 820: Does the cache update between line 816 and line 820 needs to be atomic?

JmPotato: I think it's unnecessary since the only possibility here to break the atomicity is to crash. If it crashes, the cache will bu rebuilt after rebooting.

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -237,6 +239,68 @@ func FinishSplitKeyspaceByID(c *gin.Context) {
...
+	}
+	for _, mergeID := range mergeParams.MergeList {
+		if !isValid(mergeID) {
+			c.AbortWithStatusJSON(http.StatusBadRequest, "invalid keyspace group id")
+			return
+		}
+	}
+
+	svr := c.MustGet(middlewares.ServerContextKey).(*server.Server)
+	// Split keyspace group.
```
lhy1024, line 272:
```suggestion
	// Merge keyspace group.
```

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -237,6 +239,68 @@ func FinishSplitKeyspaceByID(c *gin.Context) {
...
 }
 
+// MergeKeyspaceGroupsParams defines the params for merging the keyspace groups.
+type MergeKeyspaceGroupsParams struct {
+	MergeList []uint32 `json:"merge-list"`
+}
+
+// MergeKeyspaceGroups merges the keyspace groups in the merge list into the target keyspace group.
+func MergeKeyspaceGroups(c *gin.Context) {
+	id, err := validateKeyspaceGroupID(c)
```
lhy1024, line 249: Do we need to check whether their user kind is the same?

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -743,3 +759,116 @@ func (m *GroupManager) IsExistNode(addr string) bool {
...
+				return ErrKeyspaceGroupInSplit
+			}
+			// A keyspace group can not be split when it is in merging.
+			if kg.IsMerging() {
+				return ErrKeyspaceGroupInMerging
+			}
+			groups[kgID] = kg
+		}
+		mergeTargetKg = groups[mergeTargetID]
+		keyspaces := make(map[uint32]struct{}, len(mergeTargetKg.Keyspaces))
```
lhy1024, line 799: the capacity of map always less than needed?

pkg/keyspace/keyspace.go
```diff
@@ -49,10 +49,10 @@ const (
 	UserKindKey = "user_kind"
 	// TSOKeyspaceGroupIDKey is the key for tso keyspace group id in keyspace config.
 	TSOKeyspaceGroupIDKey = "tso_keyspace_group_id"
-	// keyspacePatrolBatchSize is the batch size for keyspace assignment patrol.
-	// the limit of etcd txn op is 128, keyspacePatrolBatchSize need to be less than it.
+	// maxEtcdTxnOps is the batch size for operating etcd. The limit of etcd txn op is 128.
```
lhy1024, line 52:
```suggestion
	// maxEtcdTxnOps is the max value for operating etcd in a txn. The default limit of etcd txn op is 128.
```

pkg/keyspace/tso_keyspace_group.go
binshi-bing: ok

binshi-bing (approved):
LGTM

server/apiv2/handlers/tso_keyspace_group.go
JmPotato: I don't think it's necessary for the time being. At present, as a split rollback, we should make it as easy as possible.

lhy1024 (approved):

Issue #6599, opened 2023-06-14
keyspace: rebalance node for keyspace group


lhy1024:
## Enhancement Task
1. online a new node when the cluster started, the members of the previously assigned keyspace group will not be updated.
2. When tso server is introduced, the machines are started at different times, for example, the first two are started first, at which time all nodes are assigned, and then the third is started, at which time the third will be empty-loaded.

lhy1024:
workaround:
1. full manual: use set-node interface
2. semi-manual: add priority for node, avoid to primary in the same machine

lhy1024:
> workaround:
> 
> 1. full manual: use set-node interface
> 2. semi-manual: add priority for node, avoid to primary in the same machine

workaround is finished

lhy1024:
After discussion with @binshi-bing, we will close it. And track auto balance in another issue.

PR #6602, opened 2023-06-14
keyspace: add priority of tso node for the keyspace group

lhy1024:

Issue Number: Ref #6599

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/utils/constant.go
+++ b/pkg/mcs/utils/constant.go
@@ -76,5 +76,5 @@ const (
 	DefaultKeyspaceGroupReplicaCount = 2
 
 	// DefaultKeyspaceGroupReplicaPriority is the default priority of a keyspace group replica.
-	DefaultKeyspaceGroupReplicaPriority = 100
+	DefaultKeyspaceGroupReplicaPriority = 0
 )
--- a/pkg/mcs/utils/constant.go
+++ b/pkg/mcs/utils/constant.go
@@ -72,9 +72,9 @@ const (
 	// foreseen future, and the former is just for extensibility in theory.
 	MaxKeyspaceGroupCountInUse = uint32(4096)
 
-	// KeyspaceGroupDefaultReplicaCount is the default replica count of keyspace group.
-	KeyspaceGroupDefaultReplicaCount = 2
+	// DefaultKeyspaceGroupReplicaCount is the default replica count of keyspace group.
+	DefaultKeyspaceGroupReplicaCount = 2
 
-	// DefaultPriority is the default priority of tso server in  keyspace group.
-	DefaultPriority = 100
+	// DefaultKeyspaceGroupReplicaPriority is the default priority of a keyspace group replica.
+	DefaultKeyspaceGroupReplicaPriority = 100
 )
--- a/pkg/mcs/utils/constant.go
+++ b/pkg/mcs/utils/constant.go
@@ -74,4 +74,7 @@ const (
 
 	// KeyspaceGroupDefaultReplicaCount is the default replica count of keyspace group.
 	KeyspaceGroupDefaultReplicaCount = 2
+
+	// DefaultPriority is the default priority of tso server in  keyspace group.
+	DefaultPriority = 100
 )
--- a/server/apiv2/handlers/tso_keyspace_group.go
+++ b/server/apiv2/handlers/tso_keyspace_group.go
@@ -327,7 +327,7 @@ func AllocNodesForKeyspaceGroup(c *gin.Context) {
 		c.AbortWithStatusJSON(http.StatusBadRequest, errs.ErrBindJSON.Wrap(err).GenWithStackByCause())
 		return
 	}
-	if manager.GetNodesCount() < allocParams.Replica || allocParams.Replica < utils.KeyspaceGroupDefaultReplicaCount {
+	if manager.GetNodesCount() < allocParams.Replica || allocParams.Replica < utils.DefaultKeyspaceGroupReplicaCount {
 		c.AbortWithStatusJSON(http.StatusBadRequest, "invalid replica, should be in [2, nodes_num]")
 		return
 	}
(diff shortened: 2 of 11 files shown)


binshi-bing (approved):
LGTM

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -347,7 +349,7 @@ func AllocNodesForKeyspaceGroup(c *gin.Context) {
 	c.JSON(http.StatusOK, nodes)
 }
 
-// SetNodesForKeyspaceGroupParams defines the params for setting nodes for keyspace groups.
+// SetNodesForKeyspaceGroupParams defines the params for setting nodes for keyspace group.
 // Notes: it should be used carefully.
 type SetNodesForKeyspaceGroupParams struct {
 	Nodes []string `json:"nodes"`
```
binshi-bing, line 355: Can we also support set priorities in  SetNodesForKeyspaceGroup API? if priorities []int is null, then set default priority for every node in the Nodes.
```suggestion
	Nodes     []string `json:"nodes"`
	Priorities []int `json:"priorities"`
```

binshi-bing: If you want to support it in the next pr, it's ok.

pkg/mcs/utils/constant.go
```diff
@@ -74,4 +74,7 @@ const (
 
 	// KeyspaceGroupDefaultReplicaCount is the default replica count of keyspace group.
 	KeyspaceGroupDefaultReplicaCount = 2
+
+	// DefaultPriority is the default priority of tso server in  keyspace group.
+	DefaultPriority = 100
```
binshi-bing, line 79: This name is too general than it suppose to be. 
```suggestion
	KeyspaceGroupReplicaDefaultPriority = uint32(100)
```

pkg/mcs/utils/constant.go
```diff
@@ -74,4 +74,7 @@ const (
 
 	// KeyspaceGroupDefaultReplicaCount is the default replica count of keyspace group.
 	KeyspaceGroupDefaultReplicaCount = 2
+
+	// DefaultPriority is the default priority of tso server in  keyspace group.
```
binshi-bing, line 78:
```suggestion
	// KeyspaceGroupReplicaDefaultPriority is the default priority of a keyspace group replica.
	// It's used in keyspace group primary weighted-election to balance primaries' distribution.
	// Among multiple replicas of a keyspace group, the higher the priority, the more likely
	// the replica is to be elected as primary.
```

binshi-bing (commented):
Left a few more comments. Though I signed off, please handle my comments before merging the pr.

server/apiv2/handlers/tso_keyspace_group.go
lhy1024: now we can support "set-node <keyspace_group_id> <tso_node_addr> [<tso_node_addr>...]" and "set-priority <keyspace_group_id> <tso_node_addr> <priority>" in pd-ctl, if only run "set-node", it means to set default priority

lhy1024: if we merge these interfaces, it will be not convenient in pd-ctl

pkg/mcs/utils/constant.go
```diff
@@ -72,6 +72,9 @@ const (
 	// foreseen future, and the former is just for extensibility in theory.
 	MaxKeyspaceGroupCountInUse = uint32(4096)
 
-	// KeyspaceGroupDefaultReplicaCount is the default replica count of keyspace group.
-	KeyspaceGroupDefaultReplicaCount = 2
+	// DefaultKeyspaceGroupReplicaCount is the default replica count of keyspace group.
+	DefaultKeyspaceGroupReplicaCount = 2
+
+	// DefaultKeyspaceGroupReplicaPriority is the default priority of a keyspace group replica.
+	DefaultKeyspaceGroupReplicaPriority = 100
```
rleungx, line 79: Why not just use 0 as the default priority? If we don't set the priority, it is empty which is 0 as the value.

lhy1024: only keep the same with leader priority

lhy1024: cc @binshi-bing

tests/pdctl/keyspace/keyspace_group_test.go
```diff
@@ -177,3 +178,107 @@ func TestExternalAllocNodeWhenStart(t *testing.T) {
...
+		testutil.Eventually(re, func() bool {
+			args := []string{"-u", pdAddr, "keyspace-group", "set-priority", defaultKeyspaceGroupID, s1.GetAddr()}
+			if p >= 0 {
+				args = append(args, strconv.Itoa(p))
+			} else {
+				args = append(args, "--", strconv.Itoa(p))
+			}
+			output, err := pdctl.ExecuteCommand(cmd, args...)
+			re.NoError(err)
+			fmt.Println(string(output))
```
rleungx, line 231: Please remove it.

tests/pdctl/keyspace/keyspace_group_test.go
```diff
@@ -177,3 +178,107 @@ func TestExternalAllocNodeWhenStart(t *testing.T) {
...
+
+	s1, tsoServerCleanup1, err := tests.StartSingleTSOTestServer(ctx, re, pdAddr, tempurl.Alloc())
+	defer tsoServerCleanup1()
+	re.NoError(err)
+	s2, tsoServerCleanup2, err := tests.StartSingleTSOTestServer(ctx, re, pdAddr, tempurl.Alloc())
+	defer tsoServerCleanup2()
+	re.NoError(err)
+	cmd := pdctlCmd.GetRootCmd()
+
+	time.Sleep(2 * time.Second)
```
rleungx, line 206: Why do we need to sleep here?

rleungx (approved):
The rest LGTM

PR #6608, opened 2023-06-15
tso: support multi-keyspace, fault injection and keyspace-name in pd-tso-bench

binshi-bing:
Issue Number: Ref #5895 

```commit-message
support multi-keyspace, fault injection and keyspace-name in pd-tso-bench
```

### Check List

Tests

// faut inject to PD/API forward
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 1 -c 1 -interval 30s -enable-fault-injection -fault-injection-rate 0.1 -use-tso-server-proxy
// fault inject to tso service
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 1 -c 1 -interval 30s -enable-fault-injection -fault-injection-rate 0.1
// Send tso requests to API leader whichs fowards requests to group 0
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 100 -c 1 -interval 10s -use-tso-server-proxy -max-send-interval-ms 100000
// Send tso requests to group 0 to tso service directly
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 100 -c 3 -interval 30s
// send tso requests to group 1
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 10 -c 3 -interval 30s -keyspace 4000
// send tso requests to group 2
./pd-tso-bench -v -duration 250000s -pd "http://serverless-cluster-pd-0.serverless-cluster-pd-peer.tidb-serverless.svc:2379/" -client 10 -c 3 -interval 30s -keyspace 2

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -244,6 +244,15 @@ func WithForwardingOption(enableForwarding bool) ClientOption {
 	}
 }
 
+// WithTSOServerProxyOption configures the client to use TSO server proxy,
+// i.e., the client will send TSO requests to the API leader (the TSO server
+// proxy) which will forward the requests to the TSO servers.
+func WithTSOServerProxyOption(useTSOServerProxy bool) ClientOption {
+	return func(c *client) {
+		c.option.useTSOServerProxy = useTSOServerProxy
+	}
+}
+
 // WithMaxErrorRetry configures the client max retry times when connect meets error.
 func WithMaxErrorRetry(count int) ClientOption {
 	return func(c *client) {
@@ -586,6 +595,11 @@ func (c *client) Close() {
 func (c *client) setServiceMode(newMode pdpb.ServiceMode) {
 	c.Lock()
 	defer c.Unlock()
+
+	if c.option.useTSOServerProxy && newMode == pdpb.ServiceMode_API_SVC_MODE {
+		// If we are using TSO server proxy, we always use PD_SVC_MODE.
+		newMode = pdpb.ServiceMode_PD_SVC_MODE
+	}
 
 	if newMode == c.serviceMode {
 		return
--- a/client/client.go
+++ b/client/client.go
@@ -244,15 +244,6 @@ func WithForwardingOption(enableForwarding bool) ClientOption {
 	}
 }
 
-// WithTSOServerProxyOption configures the client to use TSO server proxy,
-// i.e., the client will send TSO requests to the API leader (the TSO server
-// proxy) which will forward the requests to the TSO servers.
-func WithTSOServerProxyOption(useTSOServerProxy bool) ClientOption {
-	return func(c *client) {
-		c.option.useTSOServerProxy = useTSOServerProxy
-	}
-}
-
 // WithMaxErrorRetry configures the client max retry times when connect meets error.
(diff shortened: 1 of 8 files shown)


lhy1024 (approved):

client/client.go
```diff
@@ -244,6 +244,15 @@ func WithForwardingOption(enableForwarding bool) ClientOption {
 	}
 }
 
+// WithTSOServerProxyOption configures the client to use TSO server proxy,
+// i.e., the client will send TSO requests to the API leader (the TSO server
+// proxy) which will forward the requests to the TSO servers.
+func WithTSOServerProxyOption(useTSOServerProxy bool) ClientOption {
```
rleungx, line 250: Is it only used for testing?

binshi-bing: nope. If tso service discovery has problem, the new client can use this option to bypass tso service discovery too.

binshi-bing: the implementation has production quality.

binshi-bing: To let tool to support "tso server proxy", let me know if you have better idea.

rleungx: > nope. If tso service discovery has problem, the new client can use this option to bypass tso service discovery too.

If so, how do we set it? BTW, I only find it is used in the tso bench.

binshi-bing: provide a potential solution for any caller to use it. Just like we have provided followerProxy option

binshi-bing: it's mostly used by tool. Are you saying this reason isn't strong enough to add this option and if we add this option it could be abused by the client users?

binshi-bing: Do you have any better idea to let pd-tso-bench to send tso requests to  api leader instead of tso service directly? Or are you saying that we shouldn't let new PD client to talk to API leader to get TS? In my opinion, this is something that we let the new PD client to provide a legacy client to to talk to PD/API leader to get TS.

rleungx: IMO, if we provide such an option, anyone can use it. On the other hand, other components that use the pd client aren't going to support it for now. if we are going to test forward, we can call the API that the real scene will call, or use an old version of pd client.

binshi-bing: so I haven't get your point. What's your concern? Are you saying that if pd-tso-bench is the only one who needs this option then we shouldn't add it to pollute PD interface?

rleungx: Yes

binshi-bing: make sense. We can sync further on this and see if adding comment is sufficient or if there is better way.

binshi-bing: remove use tso server proxy option from this pr for now

rleungx (approved):

Issue #5892, opened 2023-02-01
Found few typos in the comments


binshi-bing:
### What did you do?

Read tikv/pd code base.

### What did you expect to see?
No typos in codebase and generated documentation.

### What did you see instead?
Saw two typos in the comments in server/config/service_middleware_config.go and server/join/join.go

### What version of PD are you using (`pd-server -V`)?
the latest code (6ca9a33) in master branch

binshi-bing:
use this very simple case to practice development cycle.

binshi-bing:
How do I assign this issue to me?

PR #6610, opened 2023-06-15
typo: fix some typo

HuSharp:

Issue Number: Ref #5892

```commit-message
fix some typos
```

### Release note

```release-note
None.
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -563,7 +563,7 @@ type ScheduleConfig struct {
 	// SplitMergeInterval is the minimum interval time to permit merge after split.
 	SplitMergeInterval typeutil.Duration `toml:"split-merge-interval" json:"split-merge-interval"`
 	// SwitchWitnessInterval is the minimum interval that allows a peer to become a witness again after it is promoted to non-witness.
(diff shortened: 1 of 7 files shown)


lhy1024 (approved):

server/config/config.go
```diff
@@ -563,7 +563,7 @@ type ScheduleConfig struct {
 	// SplitMergeInterval is the minimum interval time to permit merge after split.
 	SplitMergeInterval typeutil.Duration `toml:"split-merge-interval" json:"split-merge-interval"`
 	// SwitchWitnessInterval is the minimum interval that allows a peer to become a witness again after it is promoted to non-witness.
-	SwitchWitnessInterval typeutil.Duration `toml:"switch-witness-interval" json:"swtich-witness-interval"`
```
rleungx, line 566: Will it have a compatibility issue?

HuSharp: This is a feature that has no GA and is only used within pd...
But I think it is possible to meet a compatibility issue

nolouch (approved):

server/config/config.go
nolouch: not GA we can fix it directly.

PR #6617, opened 2023-06-15
mcs, tso: support weighted-election for TSO keyspace group primary election

### LINKED ISSUE #6616: Support weighted-election for TSO keyspace group primary election.
## Development Task
Issue Number: Ref #5895 

Used to balance keyspace group primaries' distribution.
Among multiple replicas of a keyspace group, the higher the priority, the more likely the replica is to be elected as primary.

It's the foundation for scale-out and re-balance tso service.

binshi-bing:

Issue Number: Close #6616 

```commit-message
Add the tso server registry watch loop in tso's keyspace group manager.
re-distribute TSO keyspace group primaries according to their replica priorities
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/keyspace_group_manager.go
+++ b/pkg/tso/keyspace_group_manager.go
@@ -571,7 +571,10 @@ func (kgm *KeyspaceGroupManager) primaryPriorityCheckLoop() {
 					case <-ctx.Done():
 					default:
 						member.ResetLeader()
-						log.Info("reset primary", zap.Uint32("keyspace-group-id", kg.ID))
+						log.Info("reset primary",
+							zap.String("local-address", kgm.tsoServiceID.ServiceAddr),
+							zap.Uint32("keyspace-group-id", kg.ID),
+							zap.Int("local-priority", localPriority))
 					}
 				} else {
 					log.Warn("no need to reset primary as the replicas with higher priority are offline",
--- a/pkg/tso/keyspace_group_manager.go
+++ b/pkg/tso/keyspace_group_manager.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"path"
 	"regexp"
@@ -27,6 +28,7 @@ import (
 	"time"
 
 	perrors "github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/election"
@@ -55,6 +57,10 @@ const (
 	// mergingCheckInterval is the interval for merging check to see if the keyspace groups
 	// merging process could be moved forward.
 	mergingCheckInterval = 5 * time.Second
+	// defaultPrimaryPriorityCheckInterval is the default interval for checking if the priorities
+	// of the primaries on this TSO server/pod have changed. A goroutine will periodically check
+	// do this check and re-distribute the primaries if necessary.
+	defaultPrimaryPriorityCheckInterval = 10 * time.Second
 )
 
 type state struct {
@@ -153,6 +159,40 @@ func (s *state) getKeyspaceGroupMetaWithCheck(
 		mcsutils.DefaultKeyspaceGroupID, nil
 }
 
+func (s *state) getNextPrimaryToReset(
+	groupID int, localAddress string,
+) (member ElectionMember, kg *endpoint.KeyspaceGroup, localPriority, nextGroupID int) {
+	s.RLock()
+	defer s.RUnlock()
+
+	groupSize := len(s.ams)
+	groupID %= groupSize
+	for j := 0; j < groupSize; groupID, j = (groupID+1)%groupSize, j+1 {
+		am := s.ams[groupID]
+		kg := s.kgs[groupID]
+		if am != nil && kg != nil && am.GetMember().IsLeader() {
+			maxPriority := math.MinInt32
+			localPriority := math.MaxInt32
+			for _, member := range kg.Members {
+				if member.Priority > maxPriority {
+					maxPriority = member.Priority
+				}
+				if member.Address == localAddress {
+					localPriority = member.Priority
+				}
+			}
+
+			if localPriority < maxPriority {
+				// return here and reset the primary outside of the critical section
+				// as resetting the primary may take some time.
+				return am.GetMember(), kg, localPriority, (groupID + 1) % groupSize
+			}
+		}
+	}
+
+	return nil, nil, 0, groupID
+}
+
 // kgPrimaryPathBuilder builds the path for keyspace group primary election.
 // default keyspace group: "/ms/{cluster_id}/tso/00000/primary".
 // non-default keyspace group: "/ms/{cluster_id}/tso/keyspace_groups/election/{group}/primary".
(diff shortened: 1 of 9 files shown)


pkg/tso/keyspace_group_manager.go
```diff
@@ -367,20 +518,63 @@ func (kgm *KeyspaceGroupManager) Initialize() error {
...
+					if _, ok := aliveTSONodes[member.Address]; ok {
+						resetLeader = true
+						break
+					}
+				}
+				if resetLeader {
+					select {
+					case <-ctx.Done():
+					default:
+						member.ResetLeader()
```
rleungx, line 565: Will it always reset? Should we use transfer a member with the highest priority instead of reset?

pkg/tso/keyspace_group_manager.go
```diff
@@ -367,20 +518,63 @@ func (kgm *KeyspaceGroupManager) Initialize() error {
...
+						log.Info("reset primary", zap.Uint32("keyspace-group-id", kg.ID))
```
rleungx, line 566: How about also logging the priority?

pkg/tso/keyspace_group_manager.go
binshi-bing: In my understanding, we can't transfer leader in non-embedded etcd, though the EmbeddedEtcdMember can transferLeader.

pkg/tso/keyspace_group_manager.go
binshi-bing: done.

pkg/tso/keyspace_group_manager.go
binshi-bing: Currently after member.ResetLeader(), the outside campaignLeader will sleep 25 ms in average before it recampaigns leader, so another replica with

pkg/tso/keyspace_group_manager.go
```diff
@@ -288,6 +345,100 @@ func NewKeyspaceGroupManager(
...
+
+	putFn := func(kv *mvccpb.KeyValue) error {
+		s := &discovery.ServiceRegistryEntry{}
+		if err := json.Unmarshal(kv.Value, s); err != nil {
+			log.Warn("failed to unmarshal service registry entry",
+				zap.String("event-kv-key", string(kv.Key)), zap.Error(err))
+			return err
+		}
+		kgm.tsoNodes.Store(s.ServiceAddr, struct{}{})
+		kgm.serviceRegistryMap[string(kv.Key)] = s.ServiceAddr
```
rleungx, line 401: Can we only use one map?

rleungx (approved):
The rest LGTM

pkg/tso/keyspace_group_manager.go
```diff
@@ -148,6 +154,40 @@ func (s *state) getKeyspaceGroupMetaWithCheck(
...
 
+func (s *state) getNextPrimaryToReset(
+	groupID int, localAddress string,
+) (member ElectionMember, kg *endpoint.KeyspaceGroup, localPriority, nextGroupID int) {
+	s.RLock()
+	defer s.RUnlock()
+
+	groupSize := len(s.ams)
+	groupID %= groupSize
+	for j := 0; j < groupSize; groupID, j = (groupID+1)%groupSize, j+1 {
```
lhy1024, line 165: why not foreach `s.ams` or `s.kgs`?

pkg/tso/keyspace_group_manager.go
```diff
@@ -233,14 +277,23 @@ type KeyspaceGroupManager struct {
...
+	primaryPathBuilder           *kgPrimaryPathBuilder
+	primaryPriorityCheckInterval time.Duration
+
+	// tsoNodes is the registered tso servers.
+	tsoNodes sync.Map // store as map[string]struct{}
+	// serviceRegistryMap stores the mapping from the service registry key to the service address.
+	// Note: it is only used in tsoNodesWatcher.
+	serviceRegistryMap map[string]string
+	// tsoNodesWatcher is the watcher for the registered tso servers.
+	tsoNodesWatcher *etcdutil.LoopWatcher
```
lhy1024, line 296: They are similiar to GroupManager in pkg/keyspace/tso_ketyspace_group.go

pkg/tso/keyspace_group_manager.go
binshi-bing: Can't. Two mappings are needed here. tsoNodes is basic a set of ServiceAddr() for retrieving or checking the existence of a tso server;  serviceRegistryMap is the mapping from etcd key to value to handle etcd delete event which only has the key and save the trouble from parsing address from the etcd key.

pkg/tso/keyspace_group_manager.go
binshi-bing: because "foreach s.ams and s.kgs" brings O(n^2) time complexity to adjust all primary locations based on their priorities. This function is called every 10 seconds, and each time it only finds the next primary which needs to adjust location then exit the loop. For better efficiency, when scanning the array each time, it starts from the position left by the last scan.

pkg/tso/keyspace_group_manager.go
binshi-bing: Yes, it's similar, but refactor seems to be unnecessary and overkill due to two reasons:
1. pkg/keyspace/tso_ketyspace_group.go uses nodesBalancer balancer.Balancer[string], where here tsoNodes sync.Map is used.
2. Both places need to pass so many parameters to etcdutil.NewLoopWatcher.


func (kgm *KeyspaceGroupManager) InitializeTSOServerWatchLoop() error {
	tsoServiceEndKey := clientv3.GetPrefixRangeEnd(kgm.tsoServiceKey) + "/"

        ...

	kgm.tsoNodesWatcher = etcdutil.NewLoopWatcher(
		kgm.ctx,
		&kgm.wg,
		kgm.etcdClient,
		"tso-nodes-watcher",
		kgm.tsoServiceKey,
		putFn,
		deleteFn,
		func() error { return nil },
		clientv3.WithRange(tsoServiceEndKey),
	)

	kgm.wg.Add(1)
	go kgm.tsoNodesWatcher.StartWatchLoop()

	if err := kgm.tsoNodesWatcher.WaitLoad(); err != nil {
		log.Error("failed to load the registered tso servers", errs.ZapError(err))
		return err
	}

	return nil
}

pkg/tso/keyspace_group_manager.go
binshi-bing: BTW, in the second step mentioned above "add priority field to the leadership edcd value and use transaction to ensure the replica with higher priority will become the primary/leader", this will add expensive transaction in a very common case in groups' primary election, which seems to be a concern to me.

pkg/tso/keyspace_group_manager.go
```diff
@@ -153,6 +159,40 @@ func (s *state) getKeyspaceGroupMetaWithCheck(
 		mcsutils.DefaultKeyspaceGroupID, nil
 }
 
+func (s *state) getNextPrimaryToReset(
+	groupID int, localAddress string,
+) (member ElectionMember, kg *endpoint.KeyspaceGroup, localPriority, nextGroupID int) {
+	s.RLock()
+	defer s.RUnlock()
+
+	groupSize := len(s.ams)
```
lhy1024, line 168: we should add a comment

lhy1024 (approved):

pkg/tso/keyspace_group_manager.go
binshi-bing: handled.

	// Both s.ams and s.kgs are arrays with the fixed size defined by the const value MaxKeyspaceGroupCountInUse.
	groupSize := int(mcsutils.MaxKeyspaceGroupCountInUse)

PR #6622, opened 2023-06-16
api: support record the caller IP when forwarding

### LINKED ISSUE #6595: Audit log should record the GetRegions requests
## Enhancement Task
Now, we have supported the audit log for some HTTP requests like placement rule operations and region label operations. But there are still some APIs such as `/regions` and  `/stats/region` that needs support. These APIs might cost too many resources of PD and their execution time may be long. We only record them after they finish, which cannot help us debug the problem sometimes.

### LINKED ISSUE #6598: Support getting real source address of HTTP requests
## Enhancement Task
The current implementation of HTTP request forwarding doesn't add `X-Forwarded-For` header to the request. Once the PD follower receives the request and forwards it to the leader, the leader can only know it coming from the follower. But we need to know the real source IP of the request.

rleungx:

Issue Number: Close #6598, close #6595.

The PR adds `X-Forwarded-For` into the HTTP header when the PD follower redirects the requests to the leader.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/apiutil/serverapi/middleware.go
+++ b/pkg/utils/apiutil/serverapi/middleware.go
@@ -30,6 +30,7 @@ import (
 const (
 	PDRedirectorHeader    = "PD-Redirector"
 	PDAllowFollowerHandle = "PD-Allow-follower-handle"
+	ForwardedForHeader    = "X-Forwarded-For"
 )
 
 type runtimeServiceValidator struct {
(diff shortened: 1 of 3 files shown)


disksing (approved):

CabinfeverB (approved):

nolouch (approved):

hihihuhu:
@rleungx could we cherry-pick this to release-6.5? thanks, this could be really useful to debug, with this, it would be much easier for us to find out that the tidb-dashboard calls the expensive `/pd/api/v1/regions` api every minutes

rleungx:
Also need https://github.com/tikv/pd/pull/6958

PR #6625, opened 2023-06-16
tso, tests: implement the keyspace group merge checker

JmPotato:

Issue Number: ref #6589.

```commit-message
Implement the keyspace group merge checker.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/keyspace_group_manager.go
+++ b/pkg/tso/keyspace_group_manager.go
@@ -34,6 +34,7 @@ import (
 	"github.com/tikv/pd/pkg/mcs/discovery"
 	mcsutils "github.com/tikv/pd/pkg/mcs/utils"
 	"github.com/tikv/pd/pkg/member"
+	"github.com/tikv/pd/pkg/slice"
 	"github.com/tikv/pd/pkg/storage/endpoint"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/apiutil"
@@ -41,6 +42,7 @@ import (
 	"github.com/tikv/pd/pkg/utils/logutil"
 	"github.com/tikv/pd/pkg/utils/memberutil"
 	"github.com/tikv/pd/pkg/utils/tsoutil"
+	"github.com/tikv/pd/pkg/utils/typeutil"
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/mvcc/mvccpb"
 	"go.uber.org/zap"
@@ -50,6 +52,9 @@ const (
 	keyspaceGroupsElectionPath = mcsutils.KeyspaceGroupsKey + "/election"
 	// primaryKey is the key for keyspace group primary election.
 	primaryKey = "primary"
+	// mergingCheckInterval is the interval for merging check to see if the keyspace groups
+	// merging process could be moved forward.
+	mergingCheckInterval = 5 * time.Second
 )
 
 type state struct {
@@ -384,12 +389,9 @@ func (kgm *KeyspaceGroupManager) Close() {
 }
 
 func (kgm *KeyspaceGroupManager) isAssignedToMe(group *endpoint.KeyspaceGroup) bool {
-	for _, member := range group.Members {
-		if member.Address == kgm.tsoServiceID.ServiceAddr {
-			return true
-		}
-	}
-	return false
+	return slice.AnyOf(group.Members, func(i int) bool {
+		return group.Members[i].Address == kgm.tsoServiceID.ServiceAddr
+	})
 }
 
 // updateKeyspaceGroup applies the given keyspace group. If the keyspace group is just assigned to
@@ -414,6 +416,12 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
 		// it should resign the election membership now.
 		kgm.exitElectionMembership(group)
 		return
+	}
+
+	// If this host owns a replica of the keyspace group which is the merge target, it should
+	// run the merging checker.
+	if group.IsMergeTarget() {
+		go kgm.mergingChecker(kgm.ctx, group.ID, group.MergeState.MergeList)
 	}
 
 	// If this host is already assigned a replica of this keyspace group, i.e., the election member
@@ -738,6 +746,10 @@ func (kgm *KeyspaceGroupManager) HandleTSORequest(
 	if err != nil {
 		return pdpb.Timestamp{}, curKeyspaceGroupID, err
 	}
+	err = kgm.checkTSOMerge(curKeyspaceGroupID)
+	if err != nil {
+		return pdpb.Timestamp{}, curKeyspaceGroupID, err
+	}
 	ts, err = am.HandleRequest(dcLocation, count)
 	return ts, curKeyspaceGroupID, err
 }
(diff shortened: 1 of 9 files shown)


pkg/tso/keyspace_group_manager.go
```diff
@@ -410,6 +412,12 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
 		return
 	}
 
+	// If this host owns a replica of the keyspace group which is the merge target, it should
+	// run the merging checker.
+	if group.IsMergeTarget() {
+		go kgm.mergingChecker(kgm.ctx, group.ID, group.MergeState.MergeList)
```
rleungx, line 418: Will it create multiple goroutines?

pkg/tso/keyspace_group_manager.go
```diff
@@ -881,3 +893,172 @@ func (kgm *KeyspaceGroupManager) finishSplitKeyspaceGroup(id uint32) error {
...
+// make sure the newly merged TSO keep consistent with the original ones.
+func (kgm *KeyspaceGroupManager) mergingChecker(ctx context.Context, mergeTargetID uint32, mergeList []uint32) {
+	log.Info("start to merge the keyspace group",
+		zap.Uint32("merge-target-id", mergeTargetID),
+		zap.Any("merge-list", mergeList))
+	checkTicker := time.NewTicker(mergingCheckInterval)
+	defer checkTicker.Stop()
+	// Prepare the merge map.
+	mergeMap := make(map[uint32]struct{}, len(mergeList))
+	for _, id := range mergeList {
```
rleungx, line 937: Do we need to limit the length of the merge list?

pkg/tso/keyspace_group_manager.go
```diff
@@ -881,3 +893,172 @@ func (kgm *KeyspaceGroupManager) finishSplitKeyspaceGroup(id uint32) error {
...
+		// TODO: support the Local TSO Allocator.
+		allocator, err := am.GetAllocator(GlobalDCLocation)
+		if err != nil {
+			log.Error("failed to get the allocator",
+				zap.Uint32("merge-target-id", mergeTargetID),
+				zap.Any("merge-list", mergeList),
+				zap.Error(err))
+			continue
+		}
+		err = allocator.SetTSO(
```
rleungx, line 1026: Do the timestamps of other groups need to be deleted?

pkg/tso/keyspace_group_manager.go
JmPotato: I added a check to prevent this. PTAL.

pkg/tso/keyspace_group_manager.go
JmPotato: Since it's possible these keyspace groups being split out again, it's necessary to keep them to make sure the TSO consistency.

binshi-bing (commented):
add some comments.

pkg/tso/keyspace_group_manager.go
rleungx: If we merge groups, the merged groups will use the maximum timestamp of all. If we want to split out again, the split group will use the maximum timestamp as the baseline. Why is it related to consistency?

pkg/tso/keyspace_group_manager.go
```diff
@@ -881,3 +894,172 @@ func (kgm *KeyspaceGroupManager) finishSplitKeyspaceGroup(id uint32) error {
...
+			}
+		}
+		if len(mergeMap) > 0 {
+			continue
+		}
+		// All the keyspace group primaries in the merge list are gone,
+		// update the newly merged TSO to make sure it is greater than the original ones.
+		var mergedTS time.Time
+		for _, id := range mergeList {
+			ts, err := kgm.tsoSvcStorage.LoadTimestamp(am.getKeyspaceGroupTSPath(id))
```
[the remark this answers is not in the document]
JmPotato: Nice idea. I will implement this after this PR is merged.

pkg/tso/keyspace_group_manager.go
JmPotato: I see. I didn't consider the TSO synchronization process of split before. You're right. It's reasonable to delete them.

pkg/tso/keyspace_group_manager.go
```diff
@@ -881,3 +894,172 @@ func (kgm *KeyspaceGroupManager) finishSplitKeyspaceGroup(id uint32) error {
 	kgm.kgs[id] = splitGroup
 	return nil
 }
+
+func (kgm *KeyspaceGroupManager) finishMergeKeyspaceGroup(id uint32) error {
```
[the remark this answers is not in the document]
binshi-bing: By "the keyspace meta's config map still has it's group data which will block this keyspace to be split again in PatrolKeyspaceAssignment", I meant in the following code, if the keyspace already has a group ID, we simply skip the only place during split to set the group in the existing code base (w/o #6578), which has the problem if splitTargetID in re-split is different from the existing group of the keyspace. I agree that "which will block this keyspace to be split again" in my statement isn't accurate.

#6578 can fix the inconsistency between keyspace meta and keyspace group's keyspace list, but it also has problem, because SaveKeyspaceMeta() needs to be called in batches too, otherwise the len(keyspaces) in a split needs to be <= maxTxnOps which isn't practical even in dev with tens thousands of keyspaces. Adding batching in another place increases complexity again. 

We need to simplify the split/merge logic and I have some thoughts about it. Let's discuss at at the first thing of Tuesday. @JmPotato  @rleungx @lhy1024 


```
func (manager *Manager) PatrolKeyspaceAssignment() error {
...
			for _, ks := range keyspaces {
				if ks == nil {
					continue
				}
				patrolledKeyspaceCount++
				manager.metaLock.Lock(ks.Id)
				if ks.Config == nil {
					ks.Config = make(map[string]string, 1)
				} else if _, ok := ks.Config[TSOKeyspaceGroupIDKey]; ok {
					// If the keyspace already has a group ID, skip it.
					manager.metaLock.Unlock(ks.Id)
					continue
				}
                                 ...
                                 ks.Config[TSOKeyspaceGroupIDKey] = toString(utils.DefaultKeyspaceGroupID)
				err = manager.store.SaveKeyspaceMeta(txn, ks)

```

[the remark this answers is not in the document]
binshi-bing: After syncing offline, we'll downgrade 'group' field in keyspace meta to a flag indicating if a keyspace is a legacy keyspace and if it needs to be put into default keyspace group. With this design change, #6578 isn't necessary.

binshi-bing (approved):
LGTM

pkg/tso/keyspace_group_manager.go
```diff
@@ -410,9 +416,26 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
 		return
 	}
 
+	oldAM, oldGroup := kgm.getKeyspaceGroupMeta(group.ID)
+	// If this host owns a replica of the keyspace group which is the merge target,
+	// it should run the merging checker when the merge state first time changes.
+	if !oldGroup.IsMergeTarget() && group.IsMergeTarget() {
```
JmPotato, line 422: Should also dispatch the checker when creating a new keyspace group with merge state.

JmPotato:
@rleungx @lhy1024 PTAL, thx.

pkg/tso/keyspace_group_manager.go
```diff
@@ -410,9 +416,26 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroup(group *endpoint.KeyspaceGro
...
+		ctx, cancel := context.WithCancel(kgm.ctx)
+		kgm.mergeCheckerCancelMap[group.ID] = cancel
```
rleungx, line 424: Will it be called in multiple places in the same time?

JmPotato: Yes, it will be called in each assigned replica and keep running until the merging is finished by the primary node. This is to prevent leader changing to make sure the merge could eventually be done by every node.

rleungx: So does the map have concurrent read/write operations?

JmPotato: Since the `updateKeyspaceGroup` will only be called when initializing and the etcd watching event coming, it will always be called sequentially. So it won't have concurrent read/write I think.

rleungx: The watch will start first, does it have a time window that https://github.com/tikv/pd/pull/6625/files#diff-5b6926e1bf5eeb6e97e1b27cb6eb8156d3d06ba9e72fb027f613c407f3674c68R364 happens at the same time?

rleungx (approved):

PR #6633, opened 2023-06-19
schedule: fix location label in placement rule not work well with rule checker

### LINKED ISSUE #6637: Placement rule in SQL isn't compatible with rule checker when using location labels
### What did you do?

Consider we have a cluster (three zones, each zone has a host, and each host has multiple instances)

zone1 --- host1 --- ap --- (kv1, kv2)
zone2 --- host2 --- ap --- (kv3, kv4)
zone3 --- host3 --- tp --- (kv5, kv6)

If we create the following placement policy: 
```
create placement policy p leader_constraints="[+type=ap]" follower_constraint={+type=ap:1, +type=tp:1}", survival_preferences="[host]";
```

The region peer won't be distributed in the same host.

### What did you see instead?

There will be three rules in PD, and each of them has a location label setting. However, each rule's count is 1, and the location label only takes effect in one rule. 

So, we still can find two replicas of a region in a host.

master

rleungx:

Issue Number: Close #6637

This PR is going to use the region store as the source instead of rule stores.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/rule_checker.go
+++ b/pkg/schedule/checker/rule_checker.go
@@ -410,6 +410,9 @@ func (c *RuleChecker) fixBetterLocation(region *core.RegionInfo, rf *placement.R
 		}
 	}
 
+	if len(coLocationStores) == 0 {
+		return nil, nil
+	}
 	newStore, filterByTempState := strategy.SelectStoreToImprove(coLocationStores, oldStore)
 	if newStore == 0 {
 		log.Debug("no replacement store", zap.Uint64("region-id", region.GetID()))
--- a/pkg/schedule/checker/rule_checker.go
+++ b/pkg/schedule/checker/rule_checker.go
@@ -391,18 +391,19 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
 }
 
 func (c *RuleChecker) fixBetterLocation(region *core.RegionInfo, rf *placement.RuleFit) (*operator.Operator, error) {
-	if len(rf.Rule.LocationLabels) == 0 || rf.Rule.Count <= 1 {
+	if len(rf.Rule.LocationLabels) == 0 {
 		return nil, nil
 	}
 
 	isWitness := rf.Rule.IsWitness && c.isWitnessEnabled()
 	// If the peer to be moved is a witness, since no snapshot is needed, we also reuse the fast failover logic.
 	strategy := c.strategy(region, rf.Rule, isWitness)
-	ruleStores := c.getRuleFitStores(rf)
-	oldStore := strategy.SelectStoreToRemove(ruleStores)
+	regionStores := c.cluster.GetRegionStores(region)
+	oldStore := strategy.SelectStoreToRemove(regionStores)
 	if oldStore == 0 {
 		return nil, nil
 	}
+	ruleStores := c.getRuleFitStores(rf)
 	newStore, filterByTempState := strategy.SelectStoreToImprove(ruleStores, oldStore)
 	if newStore == 0 {
 		log.Debug("no replacement store", zap.Uint64("region-id", region.GetID()))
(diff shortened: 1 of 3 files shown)


pkg/schedule/checker/rule_checker.go
```diff
@@ -391,18 +391,19 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
...
 	// If the peer to be moved is a witness, since no snapshot is needed, we also reuse the fast failover logic.
 	strategy := c.strategy(region, rf.Rule, isWitness)
-	ruleStores := c.getRuleFitStores(rf)
-	oldStore := strategy.SelectStoreToRemove(ruleStores)
+	regionStores := c.cluster.GetRegionStores(region)
+	oldStore := strategy.SelectStoreToRemove(regionStores)
 	if oldStore == 0 {
 		return nil, nil
 	}
+	ruleStores := c.getRuleFitStores(rf)
```
nolouch, line 406: I think here the `rule store` is a sub-set of the`region stores`? maybe directly get all stores which match the rule constraint.

pkg/schedule/checker/rule_checker.go
```diff
@@ -391,19 +391,26 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
...
 	strategy := c.strategy(region, rf.Rule, isWitness)
-	ruleStores := c.getRuleFitStores(rf)
-	oldStore := strategy.SelectStoreToRemove(ruleStores)
+	regionStores := c.cluster.GetRegionStores(region)
+	oldStore := strategy.SelectStoreToRemove(regionStores)
 	if oldStore == 0 {
 		return nil, nil
 	}
-	newStore, filterByTempState := strategy.SelectStoreToImprove(ruleStores, oldStore)
+	var newStores []*core.StoreInfo
```
nolouch, line 406:
```suggestion
	var coLocationStores []*core.StoreInfo
```

nolouch (commented):
rest lgtm

pkg/schedule/checker/rule_checker.go
```diff
@@ -391,19 +391,26 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
...
-	oldStore := strategy.SelectStoreToRemove(ruleStores)
+	regionStores := c.cluster.GetRegionStores(region)
+	oldStore := strategy.SelectStoreToRemove(regionStores)
 	if oldStore == 0 {
 		return nil, nil
 	}
-	newStore, filterByTempState := strategy.SelectStoreToImprove(ruleStores, oldStore)
+	var coLocationStores []*core.StoreInfo
+	for _, s := range regionStores {
+		if s.GetID() != oldStore && placement.MatchLabelConstraints(s, rf.Rule.LabelConstraints) {
```
CabinfeverB, line 408: Why check `s.GetID() != oldStore`? we swap oldstore in `SelectStoreToImprove`

CabinfeverB (approved):

nolouch (approved):
lgtm

pkg/schedule/checker/rule_checker.go
```diff
@@ -391,19 +391,26 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
...
 	}
-	newStore, filterByTempState := strategy.SelectStoreToImprove(ruleStores, oldStore)
+	var coLocationStores []*core.StoreInfo
+	for _, s := range regionStores {
+		if placement.MatchLabelConstraints(s, rf.Rule.LabelConstraints) {
+			coLocationStores = append(coLocationStores, s)
+		}
+	}
+
+	newStore, filterByTempState := strategy.SelectStoreToImprove(coLocationStores, oldStore)
```
bufferflies, line 413: what's happen if coLocationStores is nil.

rleungx: add check in both `SelectStoreToImprove` and `SelectStoreToFix`

PR #6646, opened 2023-06-20
keyspace, apiv2: support to split keyspace group with the keyspace ID range

JmPotato:

Issue Number: ref #6232.

```commit-message
Support to split keyspace group with the keyspace ID range.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace_test.go
+++ b/pkg/keyspace/keyspace_test.go
@@ -393,7 +393,7 @@ func (suite *keyspaceTestSuite) TestPatrolKeyspaceAssignment() {
 	re.NotNil(defaultKeyspaceGroup)
 	re.NotContains(defaultKeyspaceGroup.Keyspaces, uint32(111))
 	// Patrol the keyspace assignment.
-	err = suite.manager.PatrolKeyspaceAssignment()
+	err = suite.manager.PatrolKeyspaceAssignment(0, 0)
 	re.NoError(err)
 	// Check if the keyspace is attached to the default group.
 	defaultKeyspaceGroup, err = suite.manager.kgm.GetKeyspaceGroupByID(utils.DefaultKeyspaceGroupID)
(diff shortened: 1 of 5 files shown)


server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -186,10 +189,15 @@ func SplitKeyspaceGroupByID(c *gin.Context) {
 		c.AbortWithStatusJSON(http.StatusBadRequest, "invalid keyspace group id")
 		return
 	}
-	if len(splitParams.Keyspaces) == 0 {
+	if len(splitParams.Keyspaces) == 0 && splitParams.StartKeyspaceID == 0 && splitParams.EndKeyspaceID == 0 {
```
rleungx, line 192: If we only give the end key, will it move all keyspace to a new group?

JmPotato: Yes, it will move all keyspaces in the source group less than the end keyspace ID to the new group.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -509,7 +509,10 @@ func (m *GroupManager) UpdateKeyspaceGroup(oldGroupID, newGroupID string, oldUse
 
 // SplitKeyspaceGroupByID splits the keyspace group by ID into a new keyspace group with the given new ID.
 // And the keyspaces in the old keyspace group will be moved to the new keyspace group.
-func (m *GroupManager) SplitKeyspaceGroupByID(splitSourceID, splitTargetID uint32, keyspaces []uint32) error {
+func (m *GroupManager) SplitKeyspaceGroupByID(
```
lhy1024, line 512: Do we also need to support it in pd-ctl?

JmPotato: Yes, I will implement it in a new PR after this one is merged.

binshi-bing (approved):

server/apiv2/handlers/tso_keyspace_group.go
rleungx: Do we need to make sure the default keyspace 0 won't be moved?

JmPotato: If I remember correctly, the result of the last discussion was that keyspace 0 is just an ordinary keyspace ID, and there should be no problem without special treatment?

pkg/keyspace/keyspace_test.go
```diff
@@ -435,6 +435,43 @@ func (suite *keyspaceTestSuite) TestPatrolKeyspaceAssignmentInBatch() {
 	}
 }
 
+func (suite *keyspaceTestSuite) TestPatrolKeyspaceAssignmentWithRange() {
```
rleungx, line 438: How about also testing maxEtcdTxnOps? If we use a txn, will maxEtcdTxnOps affect only one kind of operation or all kinds of operations?

rleungx (approved):

PR #6660, opened 2023-06-24
checker: fix the conflict between tiflash learner and location labels

### LINKED ISSUE #6662: rule-checker does not consider the engine label constraint
## Bug Report

https://github.com/tikv/pd/pull/6633 broke the constrain that we can only move peer between the stores that satisfy the rule, as a result, `rule-checker` will move the learner peer away to a TiKV store while keep adding rule peer in the TiFlash stores. This will cause a TiFlash replica of a region never to be replicated well while bringing many redundant replicas to other TiKV nodes.

b4e2acdf-68ad-42b4-86e8-540e759e45b9

In monitoring, its phenomenon may be as shown in the figure above, with a large number of `add-rule-peer` and `move-to-better-location` at the same time. And in the log, you can see that the two act in the same region successively.

rleungx:

Issue Number: Close #6662.

This PR only chooses the source store that is contained in the rule.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


JmPotato (approved):

nolouch (approved):
lgtm

Issue #6659, opened 2023-06-23
PD/API forward streaming is expensive 


binshi-bing:
## Enhancement Task
[profile004.pdf](https://github.com/tikv/pd/files/11850234/profile004.pdf)
[2023/06/23 16:23:04.959 +00:00] [INFO] [trace.go:152] ["trace[529495468] linearizableReadLoop"] [detail="{readStateIndex:52252203; appliedIndex:52252203; }"] [duration=565.10434ms] [start=2023/06/23 16:23:04.394 +00:00] [end=2023/06/23 16:23:04.959 +00:00] [steps="[\"trace[529495468] 'read index received'  (duration: 565.085558ms)\",\"trace[529495468] 'applied index is now lower than readState.Index'  (duration: 3.208µs)\"]"]
[2023/06/23 16:23:04.962 +00:00] [WARN] [util.go:163] ["apply request took too long"] [took=567.79578ms] [expected-duration=100ms] [prefix="read-only range "] [request="key:\"/pd/7187976276065784319/gc/safe_point\" "] [response="range_response_count:1 size:79"] []
[2023/06/23 16:23:04.962 +00:00] [INFO] [trace.go:152] ["trace[1260142690] range"] [detail="{range_begin:/pd/7187976276065784319/gc/safe_point; range_end:; response_count:1; response_revision:49649625; }"] [duration=567.883329ms] [start=2023/06/23 16:23:04.394 +00:00] [end=2023/06/23 16:23:04.962 +00:00] [steps="[\"trace[1260142690] 'agreement among raft nodes before linearized reading'  (duration: 567.767596ms)\"]"]
[2023/06/23 16:23:04.966 +00:00] [WARN] [util.go:163] ["apply request took too long"] [took=572.019237ms] [expected-duration=100ms] [prefix="read-only range "] [request="key:\"/pd/7187976276065784319/gc/safe_point\" "] [response="range_response_count:1 size:79"] []
[2023/06/23 16:23:04.966 +00:00] [INFO] [trace.go:152] ["trace[96759071] range"] [detail="{range_begin:/pd/7187976276065784319/gc/safe_point; range_end:; response_count:1; response_revision:49649625; }"] [duration=572.104726ms] [start=2023/06/23 16:23:04.394 +00:00] [end=2023/06/23 16:23:04.966 +00:00] [steps="[\"trace[96759071] 'agreement among raft nodes before linearized reading'  (duration: 571.971197ms)\"]"]
[...]

PR #6664, opened 2023-06-24
mcs, tso: fix expensive async forwardTSORequest() and its timeout mechanism.

binshi-bing:

Issue Number: Ref #6659 

```commit-message
Fix expensive async forwardTSORequest() and its timeout mechanism.

In order to handle the timeout case for forwardStream send/recv, the existing logic is to create 
context.withTimeout(forwardCtx,...) for every request, then start a new goroutine "forwardTSORequest", 
which is very expensive as shown by the profiling in #6659. 

This change create a watchDeadline routine per forward stream and reuse it for all the forward requests
in which forwardTSORequest is called synchronously. Compared to the existing logic, the new change
is much cheaper and the latency is much stable.  

```

### Check List

Tests

- Unit test
Covered by the exiting unit test.

### Release note

```release-note
None.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -406,22 +406,26 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 	var (
 		server            = &tsoServer{stream: stream}
 		forwardStream     tsopb.TSO_TsoClient
-		cancel            context.CancelFunc
+		forwardCtx        context.Context
+		cancelForward     context.CancelFunc
 		lastForwardedHost string
 	)
 	defer func() {
 		s.concurrentTSOProxyStreamings.Add(-1)
-		// cancel the forward stream
-		if cancel != nil {
-			cancel()
+		if cancelForward != nil {
+			cancelForward()
 		}
 	}()
+
 	maxConcurrentTSOProxyStreamings := int32(s.GetMaxConcurrentTSOProxyStreamings())
 	if maxConcurrentTSOProxyStreamings >= 0 {
 		if newCount := s.concurrentTSOProxyStreamings.Add(1); newCount > maxConcurrentTSOProxyStreamings {
 			return errors.WithStack(ErrMaxCountTSOProxyRoutinesExceeded)
 		}
 	}
+
+	tsDeadlineCh := make(chan *tsoutil.TSDeadline, 1)
+	go tsoutil.WatchTSDeadline(stream.Context(), tsDeadlineCh)
 
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


server/grpc_service.go
```diff
@@ -406,23 +406,27 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
 	}()
+
 	maxConcurrentTSOProxyStreamings := int32(s.GetMaxConcurrentTSOProxyStreamings())
 	if maxConcurrentTSOProxyStreamings >= 0 {
 		if newCount := s.concurrentTSOProxyStreamings.Add(1); newCount > maxConcurrentTSOProxyStreamings {
 			return errors.WithStack(ErrMaxCountTSOProxyRoutinesExceeded)
 		}
 	}
 
+	tsDeadlineCh := make(chan *tsoutil.TSDeadline, 1)
```
lhy1024, line 427: how about using `context.withTimeout(forwardCtx,...)`

binshi-bing: We can't use context.withTimeout() per grpc.ServerStream, because here we want to add timeout mechanism to watch every request forwarded by the forward stream instead of the entire grpc.ServerStream's deadline. The existing logic use context.withTimeout() then asynchronously call forwardTSORequest which is expensive and the latency varies in a wider range due to synchronization across multiple goroutines. This change aims to fix this problem as described in the PR description "What is changed and how does it work?" which is also pasted below:

_In order to handle the timeout case for forwardStream send/recv, the existing logic is to create 
context.withTimeout(forwardCtx,...) for every request, then start a new goroutine "forwardTSORequest", 
which is very expensive as shown by the profiling in #6659. 

This change create a watchDeadline routine per forward stream and reuse it for all the forward requests
in which forwardTSORequest is called synchronously. Compared to the existing logic, the new change
is much cheaper and the latency is much stable._

server/grpc_service.go
```diff
@@ -502,37 +508,39 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
-	defer logutil.LogPanic()
-	// Create a context with deadline for forwarding TSO request to TSO service.
-	ctxTimeout, cancel := context.WithTimeout(ctx, tsoutil.DefaultTSOProxyTimeout)
-	defer cancel()
-
-	tsoProxyBatchSize.Observe(float64(request.GetCount()))
+	done := make(chan struct{})
+	dl := tsoutil.NewTSDeadline(tsoutil.DefaultTSOProxyTimeout, done, cancelForward)
+	select {
+	case tsDeadlineCh <- dl:
```
lhy1024, line 520: It seems only check whether timeout is in L520, if it is stuck in L526, what will we do?

rleungx (approved):

lhy1024 (approved):

PR #6673, opened 2023-06-26
operator: Merge operators canceled together

### LINKED ISSUE #6607: Merge operators better to be canceled together
## Enhancement Task
Screenshot 2023-06-14 at 16 38 25
Screenshot 2023-06-14 at 16 58 49

At 05:51:30, PD creates two merge operators that are going to merge 515 and 127. But due to "cannot demote leader peer", the operator of 515 is canceled, and the operator of 127 still leaves which will continue sending merge region step to TiKV until timeout. It also affects the merge limit since it will occupy one before its timeout.

At 05:51:47, PD creates another two merge operators that are going to merge 515 and 079. This operation is successfully executed.

So we might find in TiKV, there are two different kinds of operators related to merge 515.

It's better to cancel two merge operators together so that they won't occupy the merge-schedule-limit and the merge checker will create the operators for them soon. Otherwise, it will block merge speed if the timeout is long.

HuSharp:

Issue Number: Close #6607

```commit-message
cancel two merge operators together so that they won't occupy the merge-schedule-limit.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024 (approved):

pkg/schedule/operator/operator_controller.go
```diff
@@ -550,11 +550,28 @@ func (oc *Controller) removeOperatorLocked(op *Operator) bool {
...
+		if op.Kind()&OpMerge != 0 {
+			oc.removeRelatedOperator(op)
+		}
+
 		return true
 	}
 	return false
 }
 
+func (oc *Controller) removeRelatedOperator(op *Operator) {
```
rleungx, line 563: Do we need to bury the operator?

HuSharp: fixed.

pkg/schedule/operator/operator.go
```diff
@@ -63,6 +64,8 @@ var (
 	ExceedStoreLimit CancelReasonType = "exceed store limit"
 	// ExceedWaitLimit is the cancel reason when the operator exceeds the waiting queue limit.
 	ExceedWaitLimit CancelReasonType = "exceed wait limit"
+	// RelatedRegionID is the cancel reason when the operator is related to another region.
+	RelatedRegionID CancelReasonType = "related region id"
```
rleungx, line 68: "related region id" is not a good reason, how about something like "related merge region" or else?

pkg/schedule/operator/operator_controller.go
```diff
@@ -580,6 +583,19 @@ func (oc *Controller) removeOperatorLocked(op *Operator) bool {
...
 
+func (oc *Controller) removeRelatedOperatorLocked(op *Operator) {
+	relatedID, _ := strconv.ParseUint(op.AdditionalInfos[string(RelatedRegionID)], 10, 64)
+	if relatedOp := oc.operators[relatedID]; relatedOp != nil && relatedOp.Status() != CANCELED {
+		log.Info("operator canceled related region",
+			zap.Uint64("region-id", op.RegionID()),
+			zap.String("additional-info", relatedOp.GetAdditionalInfo()),
+			zap.Duration("takes", relatedOp.RunningTime()))
+		relatedOp.Cancel(RelatedRegionID)
+		oc.removeOperatorLocked(relatedOp)
```
rleungx, line 594: Do we need to reorder these two like others do?

pkg/schedule/operator/operator_controller.go
```diff
@@ -558,6 +558,9 @@ func (oc *Controller) RemoveOperator(op *Operator, reasons ...CancelReasonType)
 				zap.Reflect("operator", op))
 		}
 		oc.buryOperator(op)
+		if op.Kind()&OpMerge != 0 {
+			oc.removeRelatedOperatorLocked(op)
```
rleungx, line 562: Need a lock?

pkg/schedule/operator/operator.go
HuSharp: For the two operators of hot-region and other cases, maybe we also need to cancel together, so I hope to unify cancel unified reason.
what do you think?

rleungx: Yes, but it should be readable first.

pkg/schedule/operator/operator_controller.go
HuSharp: I don't get what you mean, can u give me more info...?

rleungx: change the sequence of these two lines

bufferflies (approved):

pkg/schedule/operator/operator_controller.go
```diff
@@ -575,11 +575,27 @@ func (oc *Controller) removeOperatorLocked(op *Operator) bool {
...
 		oc.ack(op)
+		if op.Kind()&OpMerge != 0 {
+			oc.removeRelatedOperator(op)
+		}
 		return true
 	}
 	return false
 }
 
+func (oc *Controller) removeRelatedOperator(op *Operator) {
```
JmPotato, line 586:
```suggestion
func (oc *Controller) removeRelatedMergeOperator(op *Operator) {
```

pkg/schedule/operator/operator_controller.go
```diff
@@ -575,11 +575,27 @@ func (oc *Controller) removeOperatorLocked(op *Operator) bool {
...
+	relatedID, _ := strconv.ParseUint(op.AdditionalInfos[string(RelatedMergeRegion)], 10, 64)
+	if relatedOp := oc.operators[relatedID]; relatedOp != nil && relatedOp.Status() != CANCELED {
+		log.Info("operator canceled related region",
```
JmPotato, line 589:
```suggestion
		log.Info("operator canceled related merge region",
```

JmPotato (changes requested):
Let's keep the term "merge" mentioned all around to ensure there will be no ambiguity.

pkg/schedule/operator/operator_controller.go
HuSharp: Okay, This name is because for the two operators of hot-region and other cases, we may need to cancel together as well. So I hope to unify name for removing related func.
But for now we just need to be specific.

HuSharp: done

HuSharp: done

JmPotato (approved):

PR #6713, opened 2023-06-29
mcs: use patch method in keyspace group

lhy1024:

Issue Number: Ref #6233

replace `POST("/:id/nodes")` with `PATCH("/:id")`

replace `POST("/:id/priority"` with `PATCH("/:id/*node")`

```commit-message
```

### Check List

Tests

- Unit test

Side effects

- modify by HTTP will be more difficult because we need to manually escape the node address in the URL.

### Release note

```release-note
None.
```


server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -40,9 +40,9 @@ func RegisterTSOKeyspaceGroup(r *gin.RouterGroup) {
 	router.GET("", GetKeyspaceGroups)
 	router.GET("/:id", GetKeyspaceGroupByID)
 	router.DELETE("/:id", DeleteKeyspaceGroupByID)
+	router.PATCH("/:id", SetNodesForKeyspaceGroup)          // only to support set nodes
+	router.PATCH("/:id/*node", SetPriorityForKeyspaceGroup) // only to support set priority
```
rleungx, line 44: why use *?

binshi-bing: Why do we change it? The current pd-ctl command "./pd-ctl keyspace-group set-priority 1 "http://pd-tso-server-0.tso-service.tidb-serverless.svc:2379" -- 100" seems good enough to me. With this change, how does the pd-ctl command look like?

binshi-bing (commented):
left one comment.

server/apiv2/handlers/tso_keyspace_group.go
lhy1024: > Why do we change it? The current pd-ctl command "./pd-ctl keyspace-group set-priority 1 "http://pd-tso-server-0.tso-service.tidb-serverless.svc:2379" -- 100" seems good enough to me. With this change, how does the pd-ctl command look like?

This pr doesn't affect pd-ctl

lhy1024: > Why do we change it? The current pd-ctl command "./pd-ctl keyspace-group set-priority 1 "http://pd-tso-server-0.tso-service.tidb-serverless.svc:2379" -- 100" seems good enough to me. With this change, how does the pd-ctl command look like?

It is a suggestion from @rleungx, which make url more restful.

POST means create and PATCH means partially update.

binshi-bing: fine-grained update is addressed by GraphQL, isn't it?

binshi-bing (approved):
LGTM

rleungx (approved):

Issue #6719, opened 2023-06-29
Frequently new a timer may cost too much memory


rleungx:
## Enhancement Task
After we upgrade, PD consumes more memory compared to before.
<img width="1440" alt="ePWuRRzpCj" src="https://github.com/tikv/pd/assets/35896542/e0332c70-a1fe-4214-918b-217462fa4012">
For each request, we will create a deadline that contains a timer. We should improve it in this kind of situation.

lhy1024:
Besides TSDeadline, time.After also cost too much memory
![image](https://github.com/tikv/pd/assets/19542290/2ab238a5-5a83-4e39-842e-daaad1b33d3a)

It is unrecommended to use time.After in for-select

PR #6720, opened 2023-06-29
*: fix memory leak introduced by timer.After

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/6719

1. use sync.Pool to maintain TSDeadline

2. replace time.After with NewTimer and timer.Stop, or ticker
- for loop retry use ticker, if the loop is broken, we will stop it by deferring or checking each branch

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

only bench with forward `c.serviceModeUpdateCb(pdpb.ServiceMode_PD_SVC_MODE)`

`~/pingcap/tiup/bin/tiup-playground v6.6.0  --pd.micro '[{"api":true, "amount":3},{"tso":true,"amount":3}]' --pd.binpath ~/pd-cse/bin/pd-server --db.binpath ~/tidb-cse/bin/tidb-server --kv 3 --db 1 --tiflash 0 --db.port 14000 --host 192.168.8.88`

`./pd-tso-bench -v -duration 250000s -client 10 -c 1 -interval 3s -pd 192.168.8.88:2379`

Before this pr

71f02e8f-43b9-4cc7-8121-663d5408b5c4

After this pr
image

### Release note

```release-note
None.
```


pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -208,8 +214,10 @@ func NewTSDeadline(
 	done chan struct{},
 	cancel context.CancelFunc,
 ) *TSDeadline {
+	timer := timerPool.Get().(*time.Timer)
+	timer.Reset(timeout)
 	return &TSDeadline{
```
binshi-bing, line 219: How about we use sync.Pool for TSDeadline and a timer is always attached to a TSDeadline and can be reused and reset every time? In this way, we use sync.Pool for both TSDeadline and Timer in fact.

binshi-bing (commented):
left one comment. The reset LGTM

pkg/utils/tsoutil/tso_dispatcher.go
lhy1024: It will make code complex and test show that `done` and `cancel` don't cost some memory.

binshi-bing: You're right. Timer is the main issue and we can keep smaller change.

binshi-bing (approved):
LGTM

client/tso_dispatcher.go
```diff
@@ -172,19 +194,25 @@ func (c *tsoClient) tsCancelLoop() {
...
 			for {
 				select {
 				case d := <-tsDeadlineCh:
 					select {
-					case <-d.timer:
+					case <-d.timer.C:
 						log.Error("[tso] tso request is canceled due to timeout", zap.String("dc-location", dc), errs.ZapError(errs.ErrClientGetTSOTimeout))
 						d.cancel()
+						timerPool.Put(d.timer) // it's safe to put the timer back to the pool
+						continue
```
rleungx, line 208: Do we need continue?

lhy1024: seem L212 is also able to be removed?

client/tso_dispatcher.go
```diff
@@ -138,12 +138,34 @@ func (c *tsoClient) updateTSODispatcher() {
...
 }
 
+// NewTSDeadline creates a new TSDeadline.
+func NewTSDeadline(
+	timeout time.Duration,
+	done chan struct{},
+	cancel context.CancelFunc,
+) *TSDeadline {
+	timer := timerPool.Get().(*time.Timer)
+	timer.Reset(timeout)
```
rleungx, line 161: Do we need to reset here?

lhy1024: yes, timer needs to be Reset.
> Reset stops a ticker and resets its period to the specified duration. The next tick will arrive after the new period elapses
https://pkg.go.dev/time@master#Ticker.Reset

server/grpc_service.go
```diff
@@ -633,14 +635,16 @@ func (s *tsoServer) Recv(timeout time.Duration) (*pdpb.TsoRequest, error) {
 		request, err := s.stream.Recv()
 		requestCh <- &pdpbTSORequest{request: request, err: err}
 	}()
+	timer := time.NewTimer(timeout)
```
rleungx, line 638: How to choose between a ticker and a timer? I find in this PR, some places use tickers, and others use timers.

lhy1024: IMO, if need to retry some times, ticker is more suitable, if only to use once, timer is more suitable.

rleungx: Yes, but seem we mixed them.

client/tso_dispatcher.go
rleungx: Then, do we need to remove it?

server/grpc_service.go
lhy1024: Which ones do you think need to be modified?

rleungx: like PatrolRegions maybe?

lhy1024: `PatrolRegions` and `runScheduler` use timer and timer.Reset, I will replace them with ticker

client/tso_dispatcher.go
lhy1024: ok, I will remove them

client/tso_dispatcher.go
```diff
@@ -138,12 +138,43 @@ func (c *tsoClient) updateTSODispatcher() {
...
+func NewTSDeadline(
+	timeout time.Duration,
+	done chan struct{},
+	cancel context.CancelFunc,
+) *TSDeadline {
+	timer := timerPool.Get().(*time.Timer)
+	// Stop the timer if it's not stopped.
+	if !timer.Stop() {
+		select {
+		case <-timer.C: // try to drain from the channel
```
rleungx, line 164: Will it cause latency?

client/tso_dispatcher.go
```diff
@@ -172,19 +203,23 @@ func (c *tsoClient) tsCancelLoop() {
...
+		tsDeadlineCh := make(chan *TSDeadline, 1)
 		c.tsDeadline.Store(dcLocation, tsDeadlineCh)
-		go func(dc string, tsDeadlineCh <-chan deadline) {
+		go func(dc string, tsDeadlineCh <-chan *TSDeadline) {
 			for {
 				select {
 				case d := <-tsDeadlineCh:
 					select {
-					case <-d.timer:
+					case <-d.timer.C:
```
rleungx, line 213: no need to stop?

client/tso_dispatcher.go
```diff
@@ -389,6 +426,15 @@ tsoBatchLoop:
 		if maxBatchWaitInterval >= 0 {
 			tbc.adjustBestBatchSize()
 		}
+		// Stop the timer if it's not stopped.
+		if !streamLoopTimer.Stop() {
+			select {
+			case <-streamLoopTimer.C: // try to drain from the channel
```
rleungx, line 432: ditto

client/tso_dispatcher.go
lhy1024: Don't worry. It will go default

client/tso_dispatcher.go
lhy1024: we have recieve from timer.C, don't need to stop again.

server/grpc_service.go
rleungx: oh, I noticed that we need some intervals to be configurable.

binshi-bing (approved):
LGTM

client/tso_dispatcher.go
```diff
@@ -138,12 +139,27 @@ func (c *tsoClient) updateTSODispatcher() {
 	})
 }
 
-type deadline struct {
-	timer  <-chan time.Time
+// TSDeadline is used to watch the deadline of each tso request.
+type TSDeadline struct {
```
JmPotato, line 143: Why change this struct to be exported? Since it's an internal struct, I prefer to keep it not exported.

client/tso_dispatcher.go
```diff
@@ -138,12 +139,27 @@ func (c *tsoClient) updateTSODispatcher() {
...
+	timer  *time.Timer
 	done   chan struct{}
 	cancel context.CancelFunc
 }
 
+// NewTSDeadline creates a new TSDeadline.
+func NewTSDeadline(
```
JmPotato, line 150: Ditto.

server/grpc_service.go
lhy1024: revert it and add comments

rleungx (approved):

JmPotato (approved):

Issue #6756, opened 2023-07-05
Provide an interface to merge all keyspace groups into the default one


JmPotato:
Ref #6589.

## Enhancement Task

The current merge operation only supports manually specifying multiple keyspace groups to merge. It would be nice to provide an interface to merge all keyspace groups into the default one easily, which should be a convenient method to do the rollback online.

PR #6757, opened 2023-07-05
keyspace, tso, apiv2: impl the interface to merge all keyspace groups into the default

JmPotato:

Issue Number: ref #6756.

```commit-message
Impl the interface to merge all keyspace groups into the default.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -1051,6 +1051,7 @@ func (m *GroupManager) MergeAllIntoDefaultKeyspaceGroup() error {
 			}
 			// Wait for the merge to finish.
 			ctx, cancel := context.WithTimeout(m.ctx, time.Minute)
+			ticker := time.NewTicker(time.Second)
 		checkLoop:
 			for {
 				select {
@@ -1060,8 +1061,9 @@ func (m *GroupManager) MergeAllIntoDefaultKeyspaceGroup() error {
 						zap.Int("batch-size", len(groupsToMerge)),
 						zap.Int("merged-group-num", mergedGroupNum))
 					cancel()
+					ticker.Stop()
 					return nil
-				case <-time.After(time.Second):
+				case <-ticker.C:
 					kg, err := m.GetKeyspaceGroupByID(utils.DefaultKeyspaceGroupID)
 					if err != nil {
 						log.Error("failed to check the default keyspace group merge state",
@@ -1070,6 +1072,7 @@ func (m *GroupManager) MergeAllIntoDefaultKeyspaceGroup() error {
 							zap.Int("merged-group-num", mergedGroupNum),
 							zap.Error(err))
 						cancel()
+						ticker.Stop()
 						return err
 					}
 					if !kg.IsMergeTarget() {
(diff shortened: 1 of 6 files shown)


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -1013,6 +1013,80 @@ func (m *GroupManager) FinishMergeKeyspaceByID(mergeTargetID uint32) error {
 	return nil
 }
 
+// MergeAllIntoDefaultKeyspaceGroup merges all other keyspace groups into the default keyspace group.
+func (m *GroupManager) MergeAllIntoDefaultKeyspaceGroup() error {
+	mergedGroupNum := 0
+	for i := 0; i < int(endpoint.UserKindCount); i++ {
+		userKind := endpoint.UserKind(i)
```
rleungx, line 1020: We didn't consider the user kind in other methods right now.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -1013,6 +1013,80 @@ func (m *GroupManager) FinishMergeKeyspaceByID(mergeTargetID uint32) error {
...
+			for {
+				select {
+				case <-ctx.Done():
+					log.Info("cancel merging all keyspace groups into the default one",
+						zap.Int("index", idx),
+						zap.Int("batch-size", len(groupsToMerge)),
+						zap.Int("merged-group-num", mergedGroupNum))
+					cancel()
+					return nil
+				case <-time.After(time.Second):
```
lhy1024, line 1064: It maybe better to use time.Ticker to avoid leak

pkg/keyspace/tso_keyspace_group.go
JmPotato: It's just used to iterate the whole `groups` map here, maybe we could simply keep it.

pkg/keyspace/tso_keyspace_group.go
JmPotato: Addressed.

lhy1024 (approved):

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -1013,6 +1013,85 @@ func (m *GroupManager) FinishMergeKeyspaceByID(mergeTargetID uint32) error {
...
+			groupsToMerge = append(groupsToMerge, group.ID)
+			if len(groupsToMerge) < maxBatchSize && idx < groups.Len()-1 {
+				continue
+			}
+			log.Info("merge keyspace groups into the default one",
+				zap.Int("index", idx),
+				zap.Int("batch-size", len(groupsToMerge)),
+				zap.Int("merged-group-num", mergedGroupNum))
+			// Reach the batch size, merge them into the default keyspace group.
+			if err := m.MergeKeyspaceGroups(utils.DefaultKeyspaceGroupID, groupsToMerge); err != nil {
```
rleungx, line 1045: Is there possible that some of the groups are merged and the rest are not?

JmPotato: Yes, it's possible if the merging batch failed halfway. And it's better to let the API caller determine whether we should retry or do something else. IMO, maybe we could introduce the retry mechanism in `pd-ctl` later to ensure the merging could be finished as much as possible.

rleungx: How about printing the merged/unmerged count in this situation?

JmPotato: Added, PTAL.

rleungx (approved):

PR #6758, opened 2023-07-05
tso: allow mergedTS to be zero in mergingChecker

JmPotato:

Issue Number: ref #6589.

```commit-message
Since it's possible that a keyspace group is to be deleted and merged before its TSO is initialized,
we should allow `mergedTS` to be zero in `mergingChecker`. This PR allows this case and only
block the merging when loading the TSO meets the error.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/tso/keyspace_group_manager.go
```diff
@@ -1268,51 +1268,62 @@ func (kgm *KeyspaceGroupManager) mergingChecker(ctx context.Context, mergeTarget
...
+			zap.Uint32("merge-target-id", mergeTargetID),
+			zap.Any("merge-list", mergeList))
 		// All the keyspace group primaries in the merge list are gone,
-		// update the newly merged TSO to make sure it is greater than the original ones.
+		// calculate the newly merged TSO to make sure it is greater than the original ones.
 		var mergedTS time.Time
 		for _, id := range mergeList {
 			ts, err := kgm.tsoSvcStorage.LoadTimestamp(am.getKeyspaceGroupTSPath(id))
-			if err != nil || ts == typeutil.ZeroTime {
+			if err != nil {
```
rleungx, line 1281: Will the merge use the wrong ts if the source has not been initialized?

JmPotato: I didn't think of such a case. Do you have any specific concerns?

JmPotato: I added a test case for this kind of situation. PTAL.

rleungx (approved):

lhy1024 (approved):

PR #6766, opened 2023-07-06
resource_manager: ignore ru metrics for background request

HuSharp:

Issue Number: Ref #5851

```commit-message
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```


client/resource_group/controller/controller.go
```diff
@@ -457,6 +462,11 @@ func (c *ResourceGroupsController) OnRequestWait(
 		failedRequestCounter.WithLabelValues(resourceGroupName).Inc()
 		return nil, nil, err
 	}
+	jobTypes := gc.meta.BackgroundSettings.JobTypes
+	if JobTypesContains(jobTypes, info.RequestSource()) {
+		// If the resource group has background jobs, we should not wait for it.
```
CabinfeverB, line 467:
```suggestion
		// If the resource group has background jobs, we should not record consumption and wait for it.
```

client/resource_group/controller/controller.go
```diff
@@ -469,6 +479,11 @@ func (c *ResourceGroupsController) OnResponse(
 		log.Warn("[resource group controller] resource group name does not exist", zap.String("name", resourceGroupName))
 		return &rmpb.Consumption{}, nil
 	}
+	jobTypes := tmp.(*groupCostController).meta.BackgroundSettings.JobTypes
+	if JobTypesContains(jobTypes, req.RequestSource()) {
+		// If the resource group has background jobs, we should not wait for it.
```
CabinfeverB, line 484:
```suggestion
		// If the resource group has background jobs, we should not record consumption.
```

HuSharp:
@glorv also PTAL, thx!

pkg/mcs/resourcemanager/server/grpc_service.go
```diff
@@ -194,6 +195,9 @@ func (s *Service) AcquireTokenBuckets(stream rmpb.ResourceManager_AcquireTokenBu
 				resourceGroupName string
 				*rmpb.Consumption
 			}{resourceGroupName, req.GetConsumptionSinceLastRequest()}
+			if isBackground {
```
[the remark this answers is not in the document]
HuSharp: > So are you going to send two request of a resource group has both background and foreground ru consumptions?

Yes, they will both use `AcquireTokenBuckets` func

>  need to mark the ru consumptions of foreground/background

Okay, I will also update the metrics panel.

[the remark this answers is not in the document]
glorv: I mean how do you distinct between foreground and background requests in `consumptionDispatcher`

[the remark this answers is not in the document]
HuSharp: Oh, I didn't add it because I ignored the panel before. I will differentiate by type.

client/resource_group/controller/controller.go
```diff
@@ -1044,6 +1055,22 @@ func (gc *groupCostController) collectRequestAndConsumption(selectTyp selectType
...
 
+// If the resource group has background jobs, we should not record consumption and wait for it.
+func (gc *groupCostController) isBackgroundRequest(requestResource string) bool {
+	gc.metaLock.Lock()
+	defer gc.metaLock.Unlock()
+
+	if gc.meta.BackgroundSettings != nil {
+		jobTypes := gc.meta.BackgroundSettings.JobTypes
+		for _, jobType := range jobTypes {
+			if strings.Contains(requestResource, jobType) {
```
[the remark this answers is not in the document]
HuSharp: done.

pkg/mcs/resourcemanager/server/grpc_service.go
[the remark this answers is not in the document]
HuSharp: added, and I checked it in metrics.

image

client/resource_group/controller/controller.go
```diff
@@ -335,6 +336,10 @@ func (c *ResourceGroupsController) tryGetResourceGroup(ctx context.Context, name
 		return gc, nil
 	}
 	// Initialize the resource group controller.
+	// TODO: remove after test
+	if group.BackgroundSettings == nil {
```
glorv, line 340: remove debug code

client/resource_group/controller/controller.go
```diff
@@ -1044,6 +1055,28 @@ func (gc *groupCostController) collectRequestAndConsumption(selectTyp selectType
...
+		for _, jobType := range jobTypes {
+			jobTypesMap[jobType] = struct{}{}
+		}
+
+		parts := strings.Split(requestResource, "_")
+		for _, part := range parts {
+			if _, ok := jobTypesMap[part]; ok {
+				return true
+			}
+		}
```
glorv, line 1075:
```suggestion
		if len(requestResource) == 0 || len(gc.meta.BackgroundSettings.JobTypes) == 0 {
				return false;
		}
		parts := strings.Split(requestResource, "_")
		return slices.Contains(gc.meta.BackgroundSettings.JobTypes, parts[len(parts)-1])
```

go.mod
```diff
@@ -205,3 +205,4 @@ replace google.golang.org/grpc v1.54.0 => google.golang.org/grpc v1.26.0
 // kvproto at the same time. You can run `go mod tidy` to make it replaced with go-mod style specification.
 // After the PR to kvproto is merged, remember to comment this out and run `go mod tidy`.
 // replace github.com/pingcap/kvproto => github.com/$YourPrivateRepo $YourPrivateBranch
+replace github.com/pingcap/kvproto => github.com/HuSharp/kvproto v0.0.0-20230710073438-f629a3cd43e9
```
glorv, line 208: remove this patch and update the version

glorv (commented):
rest LGTM

client/resource_group/controller/controller.go
HuSharp: Check for all parts due to this https://github.com/pingcap/tidb/pull/44950/files#diff-b1e739aabb5ea3cbc8cac2e5df62c017e733983caf39b090e094f48091c9eadbR263

pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -104,6 +104,22 @@ var (
 			Name:      "available_ru",
 			Help:      "Counter of the available RU for all resource groups.",
 		}, []string{resourceGroupNameLabel})
+
+	// Background RU cost metrics.
+	backgroundReadRequestUnitCost = prometheus.NewCounterVec(
```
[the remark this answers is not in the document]
HuSharp: previous ru consumption is like: `sum(rate(resource_manager_resource_unit_read_request_unit_sum{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster"}[1m]))`
if add new label will change promptsql to: 
`sum(rate(resource_manager_resource_unit_read_request_unit_sum{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster", type = "label_foreground"}[1m]))`

For the previous promptSQL will include both foreground and background, 
When old tidb is used with new PD server, extra background RUs will be calculated

go.mod
HuSharp: Okay, when TiKV and PD all get merge label, I will update kvproto.

client/resource_group/controller/controller.go
glorv: you should depend on nothing but the explict_source_type in the request source. Any component uses the background resource control should explicitly set the right session variable or default value.

HuSharp: Okay, I will fix tidb also.

pkg/mcs/resourcemanager/server/metrics.go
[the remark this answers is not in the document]
nolouch: We should not change the before behavior about the metrics, especially if it may be relaid by the billing.

client/resource_group/controller/controller.go
```diff
@@ -1044,6 +1056,22 @@ func (gc *groupCostController) collectRequestAndConsumption(selectTyp selectType
 	return req
 }
 
+// If the resource group has background jobs, we should not record consumption and wait for it.
+func (gc *groupCostController) isBackgroundRequest(requestResource string) bool {
+	gc.metaLock.Lock()
+	defer gc.metaLock.Unlock()
+
+	if gc.meta.BackgroundSettings != nil {
+		jobTypes := gc.meta.BackgroundSettings.JobTypes
```
nolouch, line 1065: I think we can cache  job type, it's changed with low frequency, so we can make a quick path to check it.

nolouch (commented):
cc

client/resource_group/controller/controller.go
```diff
@@ -469,6 +474,9 @@ func (c *ResourceGroupsController) OnResponse(
 		log.Warn("[resource group controller] resource group name does not exist", zap.String("name", resourceGroupName))
 		return &rmpb.Consumption{}, nil
 	}
+	if tmp.(*groupCostController).isBackgroundRequest(req.RequestSource()) {
```
nolouch, line 477: Maybe we label the req as background before `OnRequestWait`? then we can directly judge it by `RequestInfo`. like adding an interface about judge background first, and setting the `RequestInfo` as background, then `OnRequestWait`, `OnResponds`, or others can share the result.

HuSharp: You can pass settings to client-go?

nolouch: Let the client-go have a way to know the type (background/foreground) of the request. and then we no need to check inside the `OnResponds`, we an directly bypass `OnResponds`.

HuSharp: done. add `IsBackgroundRequest` to check in client-go, PTAL, thx!
https://github.com/tikv/client-go/pull/872/files#diff-34812587e42268a5d84933be286601fc58fdff17ab336b09ba75455ffe5ed447R111

client/resource_group/controller/controller.go
```diff
@@ -472,6 +475,26 @@ func (c *ResourceGroupsController) OnResponse(
...
+// IsBackgroundRequest If the resource group has background jobs, we should not record consumption and wait for it.
+func (c *ResourceGroupsController) IsBackgroundRequest(ctx context.Context,
+	resourceGroupName, requestResource string) bool {
+	gc, err := c.tryGetResourceGroup(ctx, resourceGroupName)
+	if err != nil {
+		failedRequestCounter.WithLabelValues(resourceGroupName).Inc()
+		return false
+	}
+
+	if bg := gc.meta.BackgroundSettings; bg != nil {
```
nolouch, line 487: need lock?

pkg/mcs/resourcemanager/server/metrics.go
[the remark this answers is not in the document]
nolouch: maybe add a type as we disscussed?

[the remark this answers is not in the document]
HuSharp: Okay

client/resource_group/controller/controller.go
```diff
@@ -472,6 +475,28 @@ func (c *ResourceGroupsController) OnResponse(
...
+// IsBackgroundRequest If the resource group has background jobs, we should not record consumption and wait for it.
+func (c *ResourceGroupsController) IsBackgroundRequest(ctx context.Context,
+	resourceGroupName, requestResource string) bool {
+	gc, err := c.tryGetResourceGroup(ctx, resourceGroupName)
+	if err != nil {
+		failedRequestCounter.WithLabelValues(resourceGroupName).Inc()
+		return false
+	}
+
+	gc.metaLock.Lock()
```
glorv, line 487: Is Rwlock better here as this call can in the hot path

client/resource_group/controller/controller.go
```diff
@@ -472,6 +475,28 @@ func (c *ResourceGroupsController) OnResponse(
...
+	defer gc.metaLock.Unlock()
+	if bg := gc.meta.BackgroundSettings; bg != nil {
+		if len(requestResource) == 0 || len(bg.JobTypes) == 0 {
+			return false
+		}
+		if idx := strings.LastIndex(requestResource, "_"); idx != -1 {
```
glorv, line 493: Should also implement the fallback to default logic here

HuSharp: Because of this pr https://github.com/tikv/client-go/pull/890, I think currently it's work fine...Any scenarios I haven't considered?

HuSharp: https://github.com/tikv/tikv/pull/15100/files#diff-4fc0c7882a408c1e08b80a597c1d551dd92301f7f6da911b5993e41ac82a265eR244 You mean just like that?

glorv: Yes. That means if a non-default resource group's config is not explicitly set, you should uses the default group's rule to check it.

client/resource_group/controller/controller.go
```diff
@@ -472,6 +475,43 @@ func (c *ResourceGroupsController) OnResponse(
...
+	resourceGroupName, requestResource string) bool {
+	gc, err := c.tryGetResourceGroup(ctx, resourceGroupName)
+	if err != nil {
+		failedRequestCounter.WithLabelValues(resourceGroupName).Inc()
+		return false
+	}
+
+	gc.metaLock.RLock()
+	defer gc.metaLock.RUnlock()
+	return c.checkBackgroundSettings(ctx, gc.meta.BackgroundSettings, requestResource)
```
glorv, line 489:
```suggestion
	return c.checkBackgroundSettings(ctx, gc.getMeta().BackgroundSettings, requestResource)
```
BTW, I think it's better to replace gc.metaLock with a atomic pointer here since it only protects a single field.

HuSharp: This must be a better way to do it! But I would prefer to set a `getBackgroundSettings` function, since `getMeta` is not a `Rwlock` function. What do u think? :)
```
func (gc *groupCostController) getBackgroundSettings() *rmpb.BackgroundSettings {
	gc.metaLock.RLock()
	defer gc.metaLock.RUnlock()
	return gc.meta.BackgroundSettings
}
```

glorv: you can just replace the Lock in getMeta with RLock instead

HuSharp: Okay

nolouch (approved):
lgtm, plz update the proto.

HuSharp:
@glorv @CabinfeverB friendly ping~

glorv (commented):
LGTM

PR #6771, opened 2023-07-06
schedule: refactor diagnostic manager

rleungx:

Issue Number: Ref #5839.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/diagnostic/diagnostic_manager.go
+++ b/pkg/schedule/diagnostic/diagnostic_manager.go
@@ -26,21 +26,13 @@ import (
 type Manager struct {
 	config              *config.PersistOptions
 	schedulerController map[string]*schedulers.ScheduleController
-	schedulerRecorders  map[string]*schedulers.DiagnosticRecorder
 }
 
 // NewManager creates a new Manager.
 func NewManager(schedulerController map[string]*schedulers.ScheduleController, config *config.PersistOptions) *Manager {
-	schedulerRecorders := make(map[string]*schedulers.DiagnosticRecorder)
-	for name := range schedulers.DiagnosableSummaryFunc {
-		if _, ok := schedulerController[name]; ok {
-			schedulerRecorders[name] = schedulerController[name].GetDiagnosticRecorder()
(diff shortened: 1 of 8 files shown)


pkg/schedule/diagnostic/diagnostic_manager.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2022 TiKV Project Authors.
```
CabinfeverB, line 1:
```suggestion
// Copyright 2023 TiKV Project Authors.
```

pkg/schedule/diagnostic/diagnostic_manager.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+// GetDiagnosticResult gets the diagnostic result of the scheduler.
+func (d *Manager) GetDiagnosticResult(name string) (*schedulers.DiagnosticResult, error) {
+	if !d.config.IsDiagnosticAllowed() {
+		return nil, errs.ErrDiagnosticDisabled
+	}
+
+	scheduler, isSchedulerExisted := d.schedulerController[name]
```
CabinfeverB, line 45: I think there's going to be a data race. We handle schedulers in coordinator with lock, but here is no lock.

CabinfeverB (commented):
Can we modify title?  because moving `scheduler_controller` from `/pkg/schedue` is also important

pkg/schedule/diagnostic/diagnostic_manager.go
rleungx: Yes, add a lock, and I will remove it in the next PR.

pkg/schedule/diagnostic/diagnostic_manager.go
rleungx: I think it's just a movement, maybe we don't need to change it?

JmPotato (approved):

nolouch (approved):

PR #6776, opened 2023-07-07
*: decouple the scheduler and checker interfaces

rleungx:

Issue Number: Ref #5839.

The PR makes the interface usage more clear. `SchedulerCluster` represents the methods that the schedulers need. `CheckerCluster` represents the methods that the checkers need. `SharedCluster` represents the methods that both schedulers and checkers need. The definitions are the same as config.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


rleungx:
@nolouch @JmPotato PTAL

JmPotato (approved):

nolouch (approved):
lgtm

PR #6806, opened 2023-07-13
*: remove unnecessary metrics

### LINKED ISSUE #6805: The cancel reason of operator should be shown on the monitor
## Enhancement Task
We do have these metrics but we don't add them to the monitor, if we cannot edit the metrics, it's hard to know the actual reason.

rleungx:

Issue Number: Close #6805

Unify the `operatorWaitCounter` and `operatorCounter`.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/operator/operator_controller.go
+++ b/pkg/schedule/operator/operator_controller.go
@@ -126,7 +126,7 @@ func (oc *Controller) Dispatch(region *core.RegionInfo, source string, recordOpS
 				recordOpStepWithTTL(op.RegionID())
 			}
 			if oc.RemoveOperator(op, Succeed) {
-				operatorWaitCounter.WithLabelValues(op.Desc(), "promote-success").Inc()
+				operatorCounter.WithLabelValues(op.Desc(), "promote-success").Inc()
 				oc.PromoteWaitingOperator()
 			}
 			if time.Since(op.GetStartTime()) < FastOperatorFinishTime {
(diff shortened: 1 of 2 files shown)


pkg/schedule/operator/operator_controller.go
```diff
@@ -126,7 +126,7 @@ func (oc *Controller) Dispatch(region *core.RegionInfo, source string, recordOpS
 				recordOpStepWithTTL(op.RegionID())
 			}
 			if oc.RemoveOperator(op, Succeed) {
-				operatorWaitCounter.WithLabelValues(op.Desc(), "promote-success").Inc()
+				operatorCounter.WithLabelValues(op.Desc(), "promote-success").Inc()
```
JmPotato, line 129: Could we pre-initialize these label values inside the operator struct to reduce the cost here?

JmPotato (commented):
The rest LGTM.

pkg/schedule/operator/operator_controller.go
rleungx: Here we need to use `op.Desc()` which cannot be pre-initialized.

JmPotato: Maybe we could initialize it when creating the operator like:

```go
func NewOperator(desc, brief string, regionID uint64, regionEpoch *metapb.RegionEpoch, kind OpKind, approximateSize int64, steps ...OpStep) *Operator {
	return &Operator{
		desc:                  desc,
		promoteSuccessCounter: operatorCounter.WithLabelValues(desc, "promote-success")
	}
}
```

rleungx: hmm, what is the difference?

JmPotato: Oops, I misunderstood the initialization process here, it's the same if we initialize the metrics each time we create a new operator here.

JmPotato (approved):

nolouch (approved):

PR #6815, opened 2023-07-17
*: remove hot region check (#90)

### LINKED ISSUE #5535: SplitAndScatter will not work when region is hot
### What did you do?

BR uses SplitAndScatter gRPC interface of PD to split region and scatter regions.

Split and scatter successfully.

### What did you see instead?

add log
img_v2_aaab6fe9-7d77-495d-8761-feb021cfa89g
Then see
img_v2_2797fe4c-fd99-40c6-b2cb-bcc14258810g

master

rleungx:

Issue Number: Close #5535.

Skip the hot region check when splitting. 
cherry-pick https://github.com/tidbcloud/pd-cse/pull/90.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/splitter/region_splitter.go
+++ b/pkg/schedule/splitter/region_splitter.go
@@ -147,6 +147,9 @@ func (r *RegionSplitter) groupKeysByRegion(keys [][]byte) map[uint64]*regionGrou
 		if !r.checkRegionValid(region) {
 			continue
 		}
+		if bytes.Equal(region.GetStartKey(), key) {
+			continue
+		}
 		log.Info("found region",
 			zap.Uint64("region-id", region.GetID()),
 			logutil.ZapRedactByteString("key", key))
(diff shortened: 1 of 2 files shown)


pkg/schedule/splitter/region_splitter.go
```diff
@@ -147,6 +147,9 @@ func (r *RegionSplitter) groupKeysByRegion(keys [][]byte) map[uint64]*regionGrou
 		if !r.checkRegionValid(region) {
 			continue
 		}
+		if bytes.Equal(region.GetStartKey(), key) {
```
lhy1024, line 150: Do we need to check end key?

rleungx: No, we don't need it.

CabinfeverB: If key is strart key of a region, does TiKV report error?

CabinfeverB (commented):
LGTM. but I find scatter will check whether region is hot too, do we need also skip?

rleungx:
> LGTM. but I find scatter will check whether region is hot too, do we need also skip?

IMO, it's not related to SplitAndScatter, maybe we can skip it after we meet the problems.

CabinfeverB (approved):

PR #6823, opened 2023-07-18
scheduler: calculate the probability to move peer

### LINKED ISSUE #6824: scheduler: replace schedulePeerPr with dynamic function
## Enhancement Task
Now we use a const value `schedulePeerPr` to get the probability of schedule the hot peer.
It is not suitable for multiple replica region.

lhy1024:

Issue Number: Close #6824

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/hot_region_test.go
```diff
@@ -203,120 +204,6 @@ func TestHotWriteRegionScheduleByteRateOnly(t *testing.T) {
 	checkHotWriteRegionScheduleByteRateOnly(re, true /* enable placement rules */)
 }
 
-func TestSplitIfRegionTooHot(t *testing.T) {
```
lhy1024, line 206: only move test

pkg/schedule/schedulers/hot_region_test.go
```diff
@@ -398,20 +286,6 @@ func checkHotWriteRegionScheduleByteRateOnly(re *require.Assertions, enablePlace
 	clearPendingInfluence(hb.(*hotScheduler))
 	tc.SetHotRegionScheduleLimit(int(opt.GetHotRegionScheduleLimit()))
 
-	for i := 0; i < 20; i++ {
```
lhy1024, line 401: It was used to check leader-schedule-limit. so remove it

pkg/schedule/schedulers/hot_region.go
```diff
@@ -152,6 +176,22 @@ func (h *baseHotScheduler) prepareForBalance(rw statistics.RWType, cluster sche.
 			prepare(regionWrite, constant.LeaderKind)
 			prepare(regionWrite, constant.RegionKind)
 			h.updateWriteTime = time.Now()
+			// calculate the probability to move peer or transfer leader
+			total, count := 0.0, 0.0
+			for _, store := range h.stLoadInfos[writePeer] {
+				for _, peer := range store.HotPeers {
+					total += float64(len(peer.GetStores()))
```
bufferflies, line 183: Do you needs to filter some unhealty peer like pending peer or down peer?

pkg/schedule/schedulers/hot_region.go
```diff
@@ -152,6 +176,22 @@ func (h *baseHotScheduler) prepareForBalance(rw statistics.RWType, cluster sche.
...
+					count++
+				}
+			}
+			replica := total / count
+			if replica <= 1 {
+				replica = float64(cluster.GetSchedulerConfig().GetMaxReplicas())
+			}
+			h.probabilityToMovePeer = (replica - 1) / replica
+			failpoint.Inject("setProbabilityToMovePeer", func(val failpoint.Value) {
```
bufferflies, line 192: It is too ugly to pass  by using

pkg/schedule/schedulers/hot_region.go
lhy1024: added `IsRegionReplicated`

pkg/schedule/schedulers/hot_region_v2_test.go
```diff
@@ -139,13 +144,15 @@ func TestHotWriteRegionScheduleWithRevertRegionsDimFirst(t *testing.T) {
 	operatorutil.CheckTransferPeer(re, ops[1], operator.OpHotRegion, 5, 2)
 	re.True(hb.searchRevertRegions[writePeer])
 	clearPendingInfluence(hb)
+	re.NoError(failpoint.Disable("github.com/tikv/pd/pkg/schedule/schedulers/setProbabilityToMovePeer"))
```
bufferflies, line 147: this failpoint use is strange

bufferflies (commented):
rest lgtm

pkg/schedule/schedulers/hot_region_v2_test.go
lhy1024: Do you have a good idea about it?

lhy1024: There doesn't seem to be a good way to remove the failpoint here, in the past we used to use a uniform global variable to control the scheduling of only write peers or only write leaders, now this is calculated in real time based on configuration

bufferflies (approved):

pkg/schedule/schedulers/hot_region.go
```diff
@@ -365,17 +386,16 @@ func (h *hotScheduler) balanceHotReadRegions(cluster sche.SchedulerCluster) []*o
 
 func (h *hotScheduler) balanceHotWriteRegions(cluster sche.SchedulerCluster) []*operator.Operator {
 	// prefer to balance by peer
-	s := h.r.Intn(100)
-	switch {
-	case s < int(schedulePeerPr*100):
+	// try to move peer
+	if h.r.Float64() < h.probabilityToMovePeer {
```
nolouch, line 390: what is the probability effect, I feel that the benefits are not obvious. and in the case of 5 replicas + 2 tiflash, the leader may not have many chances to be scheduled?

lhy1024: In the past, the probability of scheduling was based on the assumption of three replicas to calculate 2/3 = 0.66, it is not accurate for the probability under multiple replicas.

lhy1024:
close it after discussion with @nolouch

PR #6825, opened 2023-07-20
schedule: fix the attempt mechanism

### LINKED ISSUE #5201: checker/priority_inspector: the Attempt will be lost when the region's priority equals to last
## Bug Report

In the function `addOrRemoveRegion` in [server/schedule/checker/priority_inspector.go](server/schedule/checker/priority_inspector.go), it says the `Attempt` will increase if the region's priority equals to the last,
However, it seems the `Attempt` will be reset on L109 with init value from L108  when the region's priority do not change, should we return during L106-L107? 
[https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/priority_inspector.go#L98-L109](https://github.com/tikv/pd/blob/c8775b6176a03eb25cb1622fc46ba1d496872df7/server/schedule/checker/priority_inspector.go#L98-L109)

rleungx:

Issue Number: Close #5201.

Fix the logic of `Attempt` as the comment says.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/priority_inspector.go
+++ b/pkg/schedule/checker/priority_inspector.go
@@ -104,9 +104,11 @@ func (p *PriorityInspector) addOrRemoveRegion(priority int, regionID uint64) {
 			e := entry.Value.(*RegionPriorityEntry)
 			e.Attempt++
 			e.Last = time.Now()
+			p.queue.Put(priority, e)
+		} else {
+			entry := NewRegionEntry(regionID)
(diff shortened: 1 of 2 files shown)


nolouch (commented):
need test?

bufferflies (approved):
plz, add unit test

JmPotato (approved):

PR #6826, opened 2023-07-20
mcs: init cluster for scheduler

rleungx:

Issue Number: Ref #5839.

We need a cluster to init scheduler controller, and the cluster should implement the shared cluster interfaces.

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None.
```


JmPotato (approved):

pkg/mcs/scheduling/server/config.go
```diff
@@ -170,3 +196,210 @@ func (c *Config) Validate() error {
...
+// IsPlacementRulesCacheEnabled returns if the placement rules cache is enabled.
+func (o *PersistConfig) IsPlacementRulesCacheEnabled() bool {
+	return false
+}
+
+// SetPlacementRulesCacheEnabled sets if the placement rules cache is enabled.
+func (o *PersistConfig) SetPlacementRulesCacheEnabled(b bool) {}
+
+// SetEnableWitness sets if the witness is enabled.
+func (o *PersistConfig) SetEnableWitness(b bool) {}
```
lhy1024, line 405: Empty function?

pkg/mcs/scheduling/server/server.go
```diff
@@ -410,10 +510,62 @@ func (s *Server) startServer() (err error) {
 		log.Error("failed to register the service", zap.String("service-name", utils.SchedulingServiceName), errs.ZapError(err))
 		return err
 	}
+
 	atomic.StoreInt64(&s.isRunning, 1)
 	return nil
 }
 
+func (s *Server) waitAPIServiceReady() error {
```
lhy1024, line 518: How about moving waitAPIServiceReady and isAPIServiceReady to util.go?

pkg/mcs/scheduling/server/config.go
rleungx: Yes, since we don't need them.

lhy1024 (approved):

Issue #6645, opened 2023-06-20
scheduler: hot region scheduler doesn't work in v2 policy


lhy1024:
### What did you do?

sysbench oltp read write

### What did you expect to see?

balance workload

### What did you see instead?

unbalance until change policy to v1
```
scheduler config balance-hot-region-scheduler set rank-formula-version v1
scheduler config balance-hot-region-scheduler set strict-picking-store false
```
![image](https://github.com/tikv/pd/assets/19542290/6005ade1-4681-437b-a54b-4dc660f6f557)

### What version of PD are you using (`pd-server -V`)?
6.5.3

PR #6827, opened 2023-07-20
scheduler: make hot v2 more suitable small hot region

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/6645

At the master, when the load difference between the high and low nodes is large, the hot peer needs to be larger than 20% of the diff or 2% of the low node.

In this pr, we introduce the 20th node to be compared with 2% of the low node and take the smaller of these values to avoid unscheduling when the load is all composed of small hotspots.

And we cache `filterHotpeers` to avoid redundancy calculations.

### Check List

Tests

- Unit test
- Integration test

1. test in some small hot regions
master after evict leader
image
image

this pr after evict leader
image
image

2. contrast to 7.2, there is no rollback of tpcc loads for this pr
image

### Release note

```release-note
None.
```


pkg/schedule/schedulers/hot_region_v2.go
```diff
@@ -262,6 +268,7 @@ func (bs *balanceSolver) getScoreByPriorities(dim int, rs *rankV2Ratios) int {
 		// maxBetterRate may be less than minBetterRate, in which case a positive fraction cannot be produced.
 		minNotWorsenedRate = -bs.getMinRate(dim)
 		minBetterRate = math.Min(minBalancedRate*rs.perceivedRatio, lowRate*rs.minHotRatio)
+		minBetterRate = math.Min(minBetterRate, topnRate)
```
nolouch, line 271: why no need for the`pre-balanced state`? and should maxBetterRate be influenced by it?

pkg/schedule/schedulers/hot_region.go
```diff
@@ -869,44 +876,52 @@ func (bs *balanceSolver) checkSrcHistoryLoadsByPriorityAndTolerance(current, exp
...
+			return secondSort[i].GetLoad(bs.secondPriority) > secondSort[j].GetLoad(bs.secondPriority)
+		})
+	}
+	if len(hotPeers) >= topnPosition {
+		storeID := storeLoad.GetID()
+		bs.nthHotPeer[storeID][bs.firstPriority] = firstSort[topnPosition-1]
+		bs.nthHotPeer[storeID][bs.secondPriority] = secondSort[topnPosition-1]
+	}
+	if len(hotPeers) > bs.maxPeerNum {
+		union := bs.sortHotPeers(firstSort, secondSort)
```
nolouch, line 908:
```suggestion
		union := bs.unionHotPeers(firstSort, secondSort)
```

pkg/schedule/schedulers/hot_region_v2.go
lhy1024: In this branch, it means that there is a big difference between a high store and a low store.
Previously, we limited the scheduling by `minBalancedRate*rs.perceivedRatio` and  `lowRate*rs.minHotRatio` to avoid scheduling a region that is too small.
However, these values are not flexible enough for the scenario involved in the issue, which is a hotspot store composed of small regions, so we use topn for evaluation.

nolouch (approved):
lgtm

Issue #6715, opened 2023-06-29
[dr-autosync] switch to backup dc during sync_recover mode, flashback panic due to no leader for one region


mayjiang0203:
### What did you do?
Switch to backup dc during sync_recover mode, do online recovery, and flashback in the new backup cluster.

### What did you see instead?
[root@tiup-0 ~]# tiup ctl:v6.5.3 tikv --pd  pd3-peer.e2e-dr-auto-sync-5r-func-tps-1744499-1-8:2379 flashback -v 442490139984003200
Starting component `ctl`: /root/.tiup/components/ctl/v6.5.3/ctl /root/.tiup/components/ctl/v6.5.3/ctl tikv --pd pd3-peer.e2e-dr-auto-sync-5r-func-tps-1744499-1-8:2379 flashback -v 442490139984003200
thread 'flashback' panicked at 'called `Result::unwrap()` on an `Err` value: Other("[components/pd_client/src/util.rs:878]: invalid store ID 0, not found")', cmd/tikv-ctl/src/main.rs:820:62
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(9), ...)', cmd/tikv-ctl/src/main.rs:945:6
Error: exit status 101

Above error cause by following region.
![img_v2_599e8c06-0966-4051-b8df-416ff202291g](https://github.com/tikv/pd/assets/9443637/35046a0f-8928-4902-805d-c7a5b9b71385)

### What version of PD are you using (`pd-server -V`)?
sh-4.2# ./pd-server -V
Release Version: v6.5.3-hotfix0627
Edition: Community
Git Commit Hash: 22e04cea8095edbac23ff9354796070105e734be
Git Branch: heads/refs/tags/v6.5.3-hotfix0627
UTC Build Time:  2023-06-27 03:13:08

Seems it caused by cluster change to async after pd allocate id for region split and before region split started in tikv.
```
[2023/06/28 23:25:30.964 +08:00] [INFO] [cluster_worker.go:145] ["alloc ids for region split"] [region-id=3583] [peer-ids="[3584,3585,3586,3587,3588,3589,3590]"]
[...]

PR #6869, opened 2023-08-01
unsafe recovery: Add newly created empty regions output for unsafe recovery

v01dstar:

Issue Number: Ref #6715

```commit-message
In DR auto sync scenario, it is possible that a empty range hole is left in the secondary data center 
when the primary data center is totally unavailable and the replication mode is "sync recover" (super 
rare). In this case, unsafe recovery + flashback is not able to recover the cluster. The fallback 
is BR + PiTR.  While, to help the people faster identify whether the fallback is needed, we need to add 
empty regions info in the output of unsafe recovery.

If there are empty regions created, use BR+PiTR. If there are no empty regions, use flashback.
```

### Check List

Tests

- No code

Code changes

N/A

Side effects

N/A

Related changes

N/A

### Release note

```release-note
None.
```


disksing (approved):

pkg/unsaferecovery/unsafe_recovery_controller.go
```diff
@@ -136,6 +136,7 @@ type Controller struct {
 	// exposed to the outside for testing
 	AffectedTableIDs    map[int64]struct{}
 	affectedMetaRegions map[uint64]struct{}
+	newlyCreatedRegions map[uint64]struct{}
```
disksing, line 139: need make map in `reset()`

v01dstar:
This PR only needs to be (auto) cherry-picked by release 7.2.

I created a manual cherry-pick PR for release 6.5 #6877

Connor1996 (approved):
LGTM

PR #6873, opened 2023-08-01
mcs: add log `delete tso primary`

lhy1024:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/server.go
```diff
@@ -1861,6 +1861,7 @@ func (s *Server) initTSOPrimaryWatcher() {
 		return nil
 	}
 	deleteFn := func(kv *mvccpb.KeyValue) error {
+		log.Info("delete tso primary")
```
rleungx, line 1864: print the primary?

PR #6919, opened 2023-08-09
config: sync store config in time

### LINKED ISSUE #6918: Sync store may block the process of leader exit
### What did you do?

Transfer PD leader

It should be finished in a short time.

### What did you see instead?

Cost about 3 min to finish

img_v2_e647dfae-9587-4d8b-b046-1d1093b3f2eg

It will try to connect to the tombstone store and doesn't have any timeout mechanism. And the store list is only updated when fail to request all stores. So If we offline a store, even if we do remove-tombstone, it could still be on the list.
```
// runSyncConfig runs the job to sync tikv config.
func (c *RaftCluster) runSyncConfig() {
	defer logutil.LogPanic()
	defer c.wg.Done()

	ticker := time.NewTicker(time.Minute)
	defer ticker.Stop()
	stores := c.GetStores()

	syncConfig(c.storeConfigManager, stores)
	for {
		select {
		case <-c.ctx.Done():
			log.Info("sync store config job is stopped")
			return
		case <-ticker.C:
			if !syncConfig(c.storeConfigManager, stores) {
				stores = c.GetStores() // update until all stores fail 
			}
		}
	}
}
```

```
func syncConfig(manager *config.StoreConfigManager, stores []*core.StoreInfo) bool {
	for index := 0; index < len(stores); index++ {
		// filter out the stores that are tiflash
		store := stores[index]
		if core.IsStoreContainLabel(store.GetMeta(), core.EngineKey, core.EngineTiFlash) {
			continue
		}

		// filter out the stores that are not up.
		if !(store.IsPreparing() || store.IsServing()) {
			continue
		}
		// it will try next store if the current store is failed.
		address := netutil.ResolveLoopBackAddr(stores[index].GetStatusAddress(), stores[index].GetAddress())
		if err := manager.ObserveConfig(address); err != nil {
			storeSyncConfigEvent.WithLabelValues(address, "fail").Inc()
			log.Debug("sync store config failed, it will try next store", zap.Error(err))
			continue
		}
		storeSyncConfigEvent.WithLabelValues(address, "succ").Inc()
		// it will only try one store.
		return true
	}
	return false
}
```

```
// GetConfig returns the store config from TiKV.
func (s TiKVConfigSource) GetConfig(statusAddress string) (*StoreConfig, error) {
	url := fmt.Sprintf("%s://%s/config", s.schema, statusAddress)
	resp, err := s.client.Get(url)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return nil, err
	}
	var cfg StoreConfig
	if err := json.Unmarshal(body, &cfg); err != nil {
		return nil, err
	}
	return &cfg, nil
}
```

v6.1.0

bufferflies:

Issue Number: Close #6918

```commit-message
add timeout context for observer tikv config to avoid wait too long
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
sync store config in time
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -453,7 +453,7 @@ func (c *RaftCluster) syncStoreConfig(stores []*core.StoreInfo) (synced bool, sw
 	for index := 0; index < len(stores); index++ {
 		select {
 		case <-c.ctx.Done():
-			log.Info("sync store config job is stopped")
+			log.Info("stop sync store config job due to server shutdown")
 			return
 		default:
 		}
@@ -535,7 +535,7 @@ func (c *RaftCluster) fetchStoreConfigFromTiKV(ctx context.Context, statusAddres
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, bytes.NewBuffer(nil))
 	if err != nil {
 		cancel()
-		return nil, fmt.Errorf("failed to create http request: %w", err)
+		return nil, fmt.Errorf("failed to create store config http request: %w", err)
 	}
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -15,6 +15,7 @@
 package cluster
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -450,6 +451,12 @@ func (c *RaftCluster) runStoreConfigSync() {
 //   - `switchRaftV2` is true if the config of tikv engine is change to raft-kv2.
 func (c *RaftCluster) syncStoreConfig(stores []*core.StoreInfo) (synced bool, switchRaftV2 bool) {
 	for index := 0; index < len(stores); index++ {
+		select {
+		case <-c.ctx.Done():
+			log.Info("sync store config job is stopped")
+			return
+		default:
+		}
 		// filter out the stores that are tiflash
 		store := stores[index]
 		if store.IsTiFlash() {
@@ -462,8 +469,11 @@ func (c *RaftCluster) syncStoreConfig(stores []*core.StoreInfo) (synced bool, sw
 		}
 		// it will try next store if the current store is failed.
 		address := netutil.ResolveLoopBackAddr(stores[index].GetStatusAddress(), stores[index].GetAddress())
-		switchRaftV2, err := c.observeStoreConfig(address)
+		switchRaftV2, err := c.observeStoreConfig(c.ctx, address)
 		if err != nil {
+			// delete the store if it is failed and retry next store.
+			stores = append(stores[:index], stores[index+1:]...)
+			index--
 			storeSyncConfigEvent.WithLabelValues(address, "fail").Inc()
 			log.Debug("sync store config failed, it will try next store", zap.Error(err))
 			continue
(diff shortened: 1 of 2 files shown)


server/cluster/cluster.go
```diff
@@ -521,12 +531,20 @@ func (c *RaftCluster) fetchStoreConfigFromTiKV(statusAddress string) (*config.St
 	} else {
 		url = fmt.Sprintf("%s://%s/config", "http", statusAddress)
 	}
-	resp, err := c.httpClient.Get(url)
+	ctx, cancel := context.WithTimeout(ctx, clientTimeout)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, bytes.NewBuffer(nil))
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("failed to create http request: %w", err)
```
JmPotato, line 538:
```suggestion
		return nil, fmt.Errorf("failed to create store config http request: %w", err)
```

server/cluster/cluster_test.go
```diff
@@ -1386,6 +1388,49 @@ func TestStoreConfigUpdate(t *testing.T) {
 	}
 }
 
+func TestRemoveUnhealthyStore(t *testing.T) {
```
JmPotato, line 1391: I think this test is unnecessary.

server/cluster/cluster.go
```diff
@@ -450,6 +451,12 @@ func (c *RaftCluster) runStoreConfigSync() {
 //   - `switchRaftV2` is true if the config of tikv engine is change to raft-kv2.
 func (c *RaftCluster) syncStoreConfig(stores []*core.StoreInfo) (synced bool, switchRaftV2 bool) {
 	for index := 0; index < len(stores); index++ {
+		select {
+		case <-c.ctx.Done():
+			log.Info("sync store config job is stopped")
```
JmPotato, line 456: What about being different with line 442?

JmPotato (commented):
The rest are LGTM.

server/cluster/cluster.go
bufferflies: it may wait too long if there are many unhealthy store.

server/cluster/cluster.go
```diff
@@ -521,12 +531,20 @@ func (c *RaftCluster) fetchStoreConfigFromTiKV(statusAddress string) (*config.St
 	} else {
 		url = fmt.Sprintf("%s://%s/config", "http", statusAddress)
 	}
-	resp, err := c.httpClient.Get(url)
+	ctx, cancel := context.WithTimeout(ctx, clientTimeout)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, bytes.NewBuffer(nil))
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("failed to create store config http request: %w", err)
```
rleungx, line 538: Do we need to update the store instead of always trying them?

server/cluster/cluster_test.go
bufferflies: removed

server/cluster/cluster.go
bufferflies: no, it will remove this store if the function return err in line 475.

rleungx: Will line 475 delete all stores?

bufferflies: only delete the bad store.

server/cluster/cluster.go
```diff
@@ -450,6 +451,12 @@ func (c *RaftCluster) runStoreConfigSync() {
 //   - `switchRaftV2` is true if the config of tikv engine is change to raft-kv2.
 func (c *RaftCluster) syncStoreConfig(stores []*core.StoreInfo) (synced bool, switchRaftV2 bool) {
 	for index := 0; index < len(stores); index++ {
+		select {
+		case <-c.ctx.Done():
+			log.Info("pause sync store config job due to server shutdown")
```
JmPotato, line 456:
```suggestion
			log.Info("stop sync store config job due to server shutdown")
```

server/cluster/cluster.go
rleungx: But if the PD has a network partition with all TiKV within the sync period, once it recovers, will it have a problem?

bufferflies: I think no, pd will refresh all store state if all the previous stores are bad.

rleungx: make sense

server/cluster/cluster.go

rleungx (approved):

JmPotato (approved):

PR #6955, opened 2023-08-14
scheduling: watch the respective scheduler config and trim the prefix

JmPotato:

Issue Number: ref #5839.

```commit-message
- Make scheduling service watch the respective scheduler config.
- Trim the key prefix inside the watcher.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/config/config.go
+++ b/pkg/mcs/scheduling/server/config/config.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 	"unsafe"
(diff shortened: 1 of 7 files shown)


rleungx:
Can the schedulers run normally now?

JmPotato:
> Can the schedulers run normally now?

Currently `Coordinator` directly reads the scheduler config from the storage, so the sync results here can not be used. We may need to do some modifications after this PR to handle these cases:

- Handle the compatibility with the old PD scheduling.
- Determine the new path to store the scheduling-service-created config data.
- Unify the way `Coordinator` fetches scheduler config whether from storage or watcher.

So currently I'd like to keep this PR to only implement the synchronization logic, WDYT?

lhy1024 (approved):

rleungx (approved):

PR #6965, opened 2023-08-16
Scatter: make peer scatter logic same with the leader

### LINKED ISSUE #6962: Scatter peer should consider the group
## Bug Report

Right now, the scatter region consists of two parts: peer and leader. But the behavior is different. The scatter peer is according to the count of all groups, however, the scatter leader will use the count of a single group. It might cause the leader to be not balanced due to the wrong usage of TiDB. See https://github.com/pingcap/tidb/issues/46135.

https://github.com/tikv/pd/blob/70e9ffc3eda3718ba59d85bae1f2b5ef66e9bf63/pkg/schedule/scatter/region_scatterer.go#L453-L461

https://github.com/tikv/pd/blob/70e9ffc3eda3718ba59d85bae1f2b5ef66e9bf63/pkg/schedule/scatter/region_scatterer.go#L514-L524

bufferflies:

Issue Number: Close #6962

```commit-message
In past, PD conside peer distribution in the different group influenece by using `TotalCountByStore` , but not include the leader distribution. The max used situation is partition table. 
After this pr, TIDB call scatter api will use same group not different. 

ref:
https://github.com/tikv/pd/pull/3422
https://github.com/pingcap/tidb/pull/46156

```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```


pkg/schedule/scatter/region_scatterer.go
```diff
@@ -340,9 +326,15 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 				// It is both sourcePeer and targetPeer itself, no need to select.
 				continue
 			}
+			sourceStore := r.cluster.GetStore(peer.GetStoreId())
+			if sourceStore == nil {
+				log.Error("failed to get the store", zap.Uint64("store-id", peer.GetStoreId()), errs.ZapError(errs.ErrGetSourceStore))
+				continue
+			}
+			filters[filterLen-1] = filter.NewPlacementSafeguard(r.name, r.cluster.GetSharedConfig(), r.cluster.GetBasicCluster(), r.cluster.GetRuleManager(), region, sourceStore, oldFit)
```
rleungx, line 334: Why not just use append?

bufferflies: can't use filter，this filter should be updated when peer is changed.

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -332,6 +313,11 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 	selectedStores := make(map[uint64]struct{}, len(region.GetPeers()))                   // selected StoreID set
 	leaderCandidateStores := make([]uint64, 0, len(region.GetPeers()))                    // StoreID allowed to become Leader
 	scatterWithSameEngine := func(peers map[uint64]*metapb.Peer, context engineContext) { // peers: StoreID -> Peer
+		filterLen := len(context.filterFuncs) + 2
```
lhy1024, line 316: we need to add some comments about `+2`?

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -340,9 +326,15 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
...
+			sourceStore := r.cluster.GetStore(peer.GetStoreId())
+			if sourceStore == nil {
+				log.Error("failed to get the store", zap.Uint64("store-id", peer.GetStoreId()), errs.ZapError(errs.ErrGetSourceStore))
+				continue
+			}
+			filters[filterLen-1] = filter.NewPlacementSafeguard(r.name, r.cluster.GetSharedConfig(), r.cluster.GetBasicCluster(), r.cluster.GetRuleManager(), region, sourceStore, oldFit)
 			for {
-				candidates := r.selectCandidates(region, oldFit, peer.GetStoreId(), selectedStores, context)
-				newPeer := r.selectStore(group, peer, peer.GetStoreId(), candidates, context)
+				filters[filterLen-2] = filter.NewExcludedFilter(r.name, nil, selectedStores)
```
CabinfeverB, line 336: I think this filter is same for all peers

pkg/schedule/scatter/region_scatterer.go
CabinfeverB, line 334: But we don't change region info, so this filter can be used again.
Maybe u need to update `region`

bufferflies: but the source store is different

pkg/schedule/scatter/region_scatterer.go
bufferflies: selectedStores is different .

CabinfeverB: selectedStores is a pointer

pkg/schedule/scatter/region_scatterer.go
CabinfeverB: Get, but I think it's nice to put it in `selectNewPeer` like before

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -432,69 +426,46 @@ func isSameDistribution(region *core.RegionInfo, targetPeers map[uint64]*metapb.
...
 	}
+
+	var newPeer *metapb.Peer
+	minCount := uint64(math.MaxUint64)
+	sourceHit := uint64(math.MaxUint64)
 	for _, store := range stores {
-		storeCount := context.selectedPeer.TotalCountByStore(store.GetID())
+		storeCount := context.selectedPeer.Get(store.GetID(), group)
+		if store.GetID() == peer.GetId() {
+			sourceHit = storeCount
```
rleungx, line 449: How about using `originStorePeerCount?

pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -533,48 +532,11 @@ func TestSelectedStoreGC(t *testing.T) {
 	re.False(ok)
 }
 
-// TestRegionFromDifferentGroups test the multi regions. each region have its own group.
```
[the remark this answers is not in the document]
bufferflies: I think it's meaness, we don't need to support the balance between the different group .

pkg/schedule/scatter/region_scatterer.go
bufferflies: fixed as originStorePickedCound.

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -363,7 +355,7 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 	// FIXME: target leader only considers the ordinary stores, maybe we need to consider the
 	// special engine stores if the engine supports to become a leader. But now there is only
 	// one engine, tiflash, which does not support the leader, so don't consider it for now.
-	targetLeader := r.selectAvailableLeaderStore(group, region, leaderCandidateStores, r.ordinaryEngine)
+	targetLeader, leaderHit := r.selectAvailableLeaderStore(group, region, leaderCandidateStores, r.ordinaryEngine)
```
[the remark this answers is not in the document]
bufferflies: fixed as leaderStorePickedCount.

pkg/schedule/scatter/region_scatterer.go
bufferflies: good point，I will change it

pkg/schedule/scatter/region_scatterer.go
bufferflies: yes, fixed

pkg/schedule/scatter/region_scatterer_test.go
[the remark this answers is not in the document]
rleungx: Does the current test cover this case?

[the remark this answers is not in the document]
bufferflies: not cover this case. The behavior is not correct .

CabinfeverB (approved):
Which branches do we need to cherry pick

rleungx (approved):

bufferflies:
> Which branches do we need to cherry pick

will follow https://github.com/pingcap/tidb/pull/46156.

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -432,69 +426,52 @@ func isSameDistribution(region *core.RegionInfo, targetPeers map[uint64]*metapb.
...
 		}
 	}
+
+	var newPeer *metapb.Peer
+	minCount := uint64(math.MaxUint64)
+	originStorePickedCount := uint64(math.MaxUint64)
 	for _, store := range stores {
-		storeCount := context.selectedPeer.TotalCountByStore(store.GetID())
+		storeCount := context.selectedPeer.Get(store.GetID(), group)
+		if store.GetID() == peer.GetId() {
```
CabinfeverB, line 454: if store.GetID() == peer.GetStoreId() {

PR #6985, opened 2023-08-24
mcs: add region stats to scheduling cluster

rleungx:

Issue Number: Ref #5839.

The region heartbeat handling relies on the region stats and coordinator.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


rleungx:
@JmPotato PTAL

JmPotato (approved):

Issue #5273, opened 2022-07-06
Identify regions of unknown size as empty-region and undersized-region.


HunDunDM:
### What did you do?

Restart TiKV.

### What did you expect to see?

TiKV needs to re-count the region-size information. During this period, the reported `approximate-size` and `approximate-keys` are 0, and they should not be counted as empty-region and undersized-region.

This is just a monitoring issue, PD will not merge these regions.

### What did you see instead?

<img width="704" alt="image" src="https://user-images.githubusercontent.com/19789302/177482843-7a04c54d-55d1-43eb-896a-6525b32536dd.png">

### What version of PD are you using (`pd-server -V`)?

[master](https://github.com/tikv/pd/commit/3abc703fa00cef164cd75426c7590d59bd0649c1)

rleungx:
Does https://github.com/tikv/pd/pull/7009 also solve it? @CabinfeverB

CabinfeverB:
solved by #7009

PR #7009, opened 2023-08-30
statistics: fix empty region count when resuming

### LINKED ISSUE #7008: Statistics: EmptyRegion count is incorrect after the cluster is paused and resumed
## Bug Report

img_v2_7d85c37a-cdb7-4f53-84d9-08b47492324g

### What did you do?
as title

EmptyRegion is very few
### What did you see instead?
EmptyRegion equals region number

CabinfeverB:

Issue Number: Close #7008 #5273

We only save `meta` when save regions, so the approximate size of the original region is zero when restarting and `Inherit` is useless.

```commit-message
```

### Check List

Tests

image

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/statistics/region.go
+++ b/pkg/statistics/region.go
@@ -59,7 +59,7 @@ func (s *RegionStats) Observe(r *core.RegionInfo) {
 	approximateKeys := r.GetApproximateKeys()
 	approximateSize := r.GetApproximateSize()
 	approximateKvSize := r.GetApproximateKvSize()
-	if approximateSize <= core.EmptyRegionApproximateSize {
+	if approximateSize == core.EmptyRegionApproximateSize {
 		s.EmptyCount++
 	}
 	s.StorageSize += approximateSize
--- a/pkg/statistics/region.go
+++ b/pkg/statistics/region.go
@@ -62,7 +62,9 @@ func (s *RegionStats) Observe(r *core.RegionInfo) {
 	if approximateSize == core.EmptyRegionApproximateSize {
 		s.EmptyCount++
 	}
-	s.StorageSize += approximateSize
+	if !r.IsEmptyRegion() {
+		s.StorageSize += approximateSize
+	}
 	s.UserStorageSize += approximateKvSize
 	s.StorageKeys += approximateKeys
 	leader := r.GetLeader()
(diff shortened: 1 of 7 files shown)


pkg/statistics/region_collection.go
```diff
@@ -198,7 +198,9 @@ func (r *RegionStatistics) Observe(region *core.RegionInfo, stores []*core.Store
 			return false
 		}(),
 		LearnerPeer: len(region.GetLearners()) > 0,
-		EmptyRegion: region.GetApproximateSize() <= core.EmptyRegionApproximateSize,
+		// When cluster resumes, the region size may be not initialized, but region heartbeat is send.
```
rleungx, line 201: If the region size is not initialized, will it inherit from the original one?

rleungx (approved):

pkg/statistics/region.go
```diff
@@ -59,7 +59,7 @@ func (s *RegionStats) Observe(r *core.RegionInfo) {
 	approximateKeys := r.GetApproximateKeys()
 	approximateSize := r.GetApproximateSize()
 	approximateKvSize := r.GetApproximateKvSize()
-	if approximateSize <= core.EmptyRegionApproximateSize {
+	if approximateSize == core.EmptyRegionApproximateSize {
 		s.EmptyCount++
 	}
 	s.StorageSize += approximateSize
```
Connor1996, line 65: approximateSize == 1 means empty, should we add it to storage size?

CabinfeverB: Do u mean current storage size? If so, it is calculated from the used size of store

Connor1996: I mean 
```go
if approximateSize > core.EmptyRegionApproximateSize {
   s.StorageSize += approximateSize
}
```

pkg/statistics/region_collection.go
```diff
@@ -198,15 +198,17 @@ func (r *RegionStatistics) Observe(region *core.RegionInfo, stores []*core.Store
...
+		// So use `==` here.
+		EmptyRegion: region.GetApproximateSize() == core.EmptyRegionApproximateSize,
```
Connor1996, line 203: can we extract a function `IsEmptyRegion()` for `regionInfo`

Connor1996 (approved):
LGTM

PR #7013, opened 2023-08-30
mcs: extract the common part as base server

rleungx:

Issue Number: Ref #5839.

There is too many duplicated code in the different services. Maintaining one copy will help us reduce the probability of making mistakes.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/server/server.go
+++ b/pkg/mcs/server/server.go
@@ -1,0 +1,164 @@
+// Copyright 2023 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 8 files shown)


pkg/mcs/server/server.go
```diff
@@ -0,0 +1,164 @@
...
+		startTimestamp: time.Now().Unix(),
+	}
+}
+
+// Context returns the context of server.
+func (bs *BaseServer) Context() context.Context {
+	return bs.ctx
+}
+
+// GetDelegateClient returns grpc client connection talking to the forwarded host
```
lhy1024, line 64:
```suggestion
// GetDelegateClient returns grpc client connection talking to the forwarded host.
```

lhy1024 (approved):

JmPotato (approved):

pkg/mcs/server/server.go
rleungx: Maybe in another PR.

PR #7023, opened 2023-08-31
mcs: forward region requests to scheduling server

rleungx:

Issue Number: Ref #5839.

This PR is used to forward region requests to the scheduling server and let it handle the scheduling directly, the operator will be brought through forwarded heartbeat response to API server. Then send it to TiKV.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


rleungx:
/test build
--

pkg/core/region.go
```diff
@@ -194,6 +195,54 @@ func RegionFromHeartbeat(heartbeat *pdpb.RegionHeartbeatRequest, opts ...RegionC
 	return region
 }
 
+// RegionFromForward constructs a Region from forwarded region heartbeat.
+func RegionFromForward(heartbeat *schedulingpb.RegionHeartbeatRequest, opts ...RegionCreateOption) *RegionInfo {
```
lhy1024, line 199: RegionFromForward and RegionFromHeartbeat seem to be different only `regionKvSize := heartbeat.GetApproximateKvSize() / units.MiB` here.

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -348,3 +349,82 @@ func (c *Cluster) HandleStoreHeartbeat(heartbeat *schedulingpb.StoreHeartbeatReq
...
+	if err := c.processRegionHeartbeat(region); err != nil {
+		return err
+	}
+
+	c.coordinator.GetOperatorController().Dispatch(region, operator.DispatchFromHeartBeat, c.coordinator.RecordOpStepWithTTL)
+	return nil
+}
+
+// processRegionHeartbeat updates the region information.
+func (c *Cluster) processRegionHeartbeat(region *core.RegionInfo) error {
```
lhy1024, line 364: Maybe we can reduce some of the duplicate functions, and of course, we can do it in the next pr.

server/cluster/cluster_test.go
```diff
@@ -3633,7 +3633,7 @@ func TestInterval(t *testing.T) {
 func waitAddLearner(re *require.Assertions, stream mockhbstream.HeartbeatStream, region *core.RegionInfo, storeID uint64) *core.RegionInfo {
```
lhy1024, line 3633: Do we need to add some tests with API mode to test forward?

lhy1024 (commented):
the rest LGTM

server/cluster/cluster_test.go
rleungx: added

pkg/core/region.go
rleungx: request type is different

lhy1024 (approved):

JmPotato (approved):

PR #7031, opened 2023-09-01
tso: sync timestamp once before the first serving

### LINKED ISSUE #7015: TestMixedTSODeployment failed again
### CI link
https://github.com/tikv/pd/actions/runs/6023573258/job/16340590837
https://github.com/tikv/pd/actions/runs/6022180459/job/16336412853

JmPotato:

Issue Number: Close #7015.

```commit-message
Sync the timestamp once before the first serving to make sure that the TSO is the latest one from the storage,
which could prevent the potential fallback caused by the rolling update of the mixed PD and TSO service.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/keyspace_group_manager.go
+++ b/pkg/tso/keyspace_group_manager.go
@@ -78,7 +78,7 @@ type state struct {
 	// deletedGroups is the cache of deleted keyspace group related information.
 	deletedGroups map[uint32]struct{}
 	// requestedGroups is the cache of requested keyspace group related information.
-	// Once a group receives its first TSO request, it will be added to this map.
+	// Once a group receives its first TSO request and pass the check, it will be added to this map.
 	requestedGroups map[uint32]struct{}
 }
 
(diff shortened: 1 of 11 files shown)


rleungx (approved):

lhy1024 (commented):
TestTSOProxyTestSuite failed

pkg/tso/keyspace_group_manager.go
```diff
@@ -146,6 +153,43 @@ func (s *state) getDeletedGroupNum() int {
...
+	s.RLock()
+	_, ok := s.requestedGroups[groupID]
+	s.RUnlock()
+	if ok {
+		return nil
+	}
+	s.Lock()
+	defer s.Unlock()
+	// Double check if the keyspace group has been marked as requested.
+	if _, ok := s.requestedGroups[groupID]; ok {
```
rleungx, line 183: Is it necessary to double check it?

tests/integrations/tso/client_test.go
```diff
@@ -436,7 +436,7 @@ func TestMixedTSODeployment(t *testing.T) {
 
 	ctx1, cancel1 := context.WithCancel(context.Background())
 	var wg sync.WaitGroup
-	checkTSO(ctx1, re, &wg, backendEndpoints, pd.WithAllowTSOFallback() /* It's expected that the timestamp fallback happens here */)
```
rleungx, line 439: Do we still need WithAllowTSOFallback?

JmPotato: Removed.

pkg/tso/keyspace_group_manager.go
JmPotato: I think this may happen in the case of high concurrency. Double checking can reduce the possibility of etcd transaction conflicts in sync timestamp.

JmPotato: Just to avoid the unnecessary operation as much as possible on the TSO hot path.

rleungx (approved):

lhy1024 (approved):

PR #7032, opened 2023-09-01
mcs: use separate participant

rleungx:

Issue Number: Ref #5839.

Previously, we used tso participant for all micro-services which is not suitable. This PR distinguishes them by using different proto definitions.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/member/participant.go
```diff
@@ -131,27 +146,36 @@ func (m *Participant) GetLeaderID() uint64 {
...
+func (m *Participant) setLeader(member participant) {
 	m.leader.Store(member)
 	m.lastLeaderUpdatedTime.Store(time.Now())
 }
 
 // unsetLeader unsets the member's leader.
 func (m *Participant) unsetLeader() {
-	m.leader.Store(&tsopb.Participant{})
+	var leader participant
+	switch m.serviceName {
```
JmPotato, line 170: Will generic type eliminate these switches?

JmPotato (approved):

pkg/member/participant.go
rleungx: Maybe rewrite it later.

JmPotato (approved):

rleungx:
@lhy1024 PTAL

pkg/member/participant.go
```diff
@@ -22,28 +22,42 @@ import (
...
 	"github.com/tikv/pd/pkg/utils/etcdutil"
 	"go.etcd.io/etcd/clientv3"
 	"go.uber.org/zap"
 )
 
 type leadershipCheckFunc func(*election.Leadership) bool
 
+type participant interface {
+	GetName() string
+	GetId() uint64
```
lhy1024, line 41: How about GetID

lhy1024 (approved):

pkg/member/participant.go
rleungx: No, it's generated by protobuf.

PR #7047, opened 2023-09-05
mcs: sync store state to scheduling service

rleungx:

Issue Number: Ref #5839.

This PR synchronizes the changed state to the scheduling service because the heartbeat only reports the statistics.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/meta/watcher.go
+++ b/pkg/mcs/scheduling/server/meta/watcher.go
@@ -1,0 +1,117 @@
+// Copyright 2023 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


pkg/mcs/scheduling/server/meta/watcher.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return w, nil
+}
+
+func (w *Watcher) initializeStoreWatcher() error {
+	putFn := func(kv *mvccpb.KeyValue) error {
+		store := &metapb.Store{}
+		if err := proto.Unmarshal(kv.Value, store); err != nil {
+			log.Warn("failed to unmarshal store entry",
+				zap.String("event-kv-key", string(kv.Key)), zap.Error(err))
```
JmPotato, line 76: Do we need to call `ExtractStoreIDFromPath` to get the store ID here?

JmPotato (approved):
The reset LGTM.

lhy1024 (approved):

pkg/mcs/scheduling/server/meta/watcher.go
rleungx: Do you mean to add it to the log?

Issue #7073, opened 2023-09-11
pd incorrectly report "scatter region failed" in normal scatter reigon scenario


seiya-annie:
### What did you do?

open titan
lightning  import data to tidb cluster

### What did you expect to see?
import successfully
### What did you see instead?
```
[...]

PR #7075, opened 2023-09-11
scatter: fix the unexpected error code

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/7073

```commit-message
scatter: fix the unexpected error code
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/scatter/region_scatterer.go
+++ b/pkg/schedule/scatter/region_scatterer.go
@@ -296,7 +296,7 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 	for _, peer := range region.GetPeers() {
 		store := r.cluster.GetStore(peer.GetStoreId())
 		if store == nil {
-			return nil, errs.ErrGetSourceStore.FastGenByArgs(fmt.Sprintf("store not found, peer: %v", peer))
+			return nil, errs.ErrGetSourceStore.FastGenByArgs(fmt.Sprintf("store not found, peer: %v, region id: %d", peer, region.GetID()))
 		}
 		if engineFilter.Target(r.cluster.GetSharedConfig(), store).IsOK() {
 			ordinaryPeers[peer.GetStoreId()] = peer
@@ -384,6 +384,7 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 			targetPeers[peer.GetStoreId()] = peer
 		}
 		r.Put(targetPeers, region.GetLeader().GetStoreId(), group)
+		log.Debug("fail to create scatter region operator", errs.ZapError(err))
 		return nil, errs.ErrCreateOperator.FastGenByArgs(fmt.Sprintf("failed to create scatter region operator for region %v", region.GetID()))
 	}
 	if op != nil {
--- a/pkg/schedule/scatter/region_scatterer.go
+++ b/pkg/schedule/scatter/region_scatterer.go
@@ -284,10 +284,10 @@ func (r *RegionScatterer) Scatter(region *core.RegionInfo, group string, skipSto
 		return nil, errors.Errorf("region %d is hot", region.GetID())
 	}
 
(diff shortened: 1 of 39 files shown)


server/grpc_service.go
```diff
@@ -1710,13 +1710,6 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
 		return nil, err
 	}
 
-	if op == nil {
```
nolouch, line 1713: Should be seen as a normal situation

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -296,7 +296,7 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 	for _, peer := range region.GetPeers() {
 		store := r.cluster.GetStore(peer.GetStoreId())
 		if store == nil {
-			return nil
+			return nil, errs.ErrGetSourceStore.FastGenByArgs(fmt.Sprintf("store not found, peer: %v", peer))
```
bufferflies, line 299: how about add region id for debug? In this case, should we need to tell users wait to fix the fault peer .

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -384,8 +384,7 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 			targetPeers[peer.GetStoreId()] = peer
 		}
 		r.Put(targetPeers, region.GetLeader().GetStoreId(), group)
-		log.Debug("fail to create scatter region operator", errs.ZapError(err))
-		return nil
+		return nil, errs.ErrCreateOperator.FastGenByArgs(fmt.Sprintf("failed to create scatter region operator for region %v", region.GetID()))
```
CabinfeverB, line 387: I prefer to print the error for debug

pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -140,6 +146,7 @@ func scatter(re *require.Assertions, numStores, numRegions uint64, useRules bool
 		re.LessOrEqual(float64(count), 1.1*float64(numRegions)/float64(numStores))
 		re.GreaterOrEqual(float64(count), 0.9*float64(numRegions)/float64(numStores))
 	}
+	re.GreaterOrEqual(noNeedMoveNum, 0)
```
CabinfeverB, line 149: It seems always be true

CabinfeverB (commented):
REST LGTM

server/grpc_service.go
```diff
@@ -1709,18 +1709,13 @@ func (s *GrpcServer) ScatterRegion(ctx context.Context, request *pdpb.ScatterReg
...
-		}, nil
-	}
-
-	if !rc.GetOperatorController().AddOperator(op) {
-		return &pdpb.ScatterRegionResponse{
-			Header: s.wrapErrorToHeader(pdpb.ErrorType_UNKNOWN,
-				"operator cancelled because store limit exceeded"),
-		}, nil
+	if op != nil {
+		if !rc.GetOperatorController().AddOperator(op) {
```
rleungx, line 1713: I realize that if the operator cannot be added, the store limit exceeded is just one of the reasons.

pkg/schedule/scatter/region_scatterer_test.go
nolouch: yes, here only verify `noNeedMoveNum` is normal situation and no error reported.

server/grpc_service.go
nolouch: What other reasons?

rleungx: If checkAddOperator fails, it will also return false.

nolouch: I changed the error info. PTAL @rleungx

rleungx (approved):

CabinfeverB (approved):

PR #7076, opened 2023-09-11
scheduling: sync schedulers from the API server

JmPotato:

Issue Number: ref #5839. Should merge after #7082.

```commit-message
- Sync the schedulers from the API server.
- Dynamically reload the scheduler config.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/cluster.go
+++ b/pkg/mcs/scheduling/server/cluster.go
@@ -261,7 +261,7 @@ func (c *Cluster) waitSchedulersInitialized() {
 	ticker := time.NewTicker(time.Millisecond * 100)
 	defer ticker.Stop()
 	for {
-		if c.coordinator.CheckSchedulersInitialized() {
+		if c.coordinator.AreSchedulersInitialized() {
 			return
 		}
 		select {
--- a/pkg/mcs/scheduling/server/cluster.go
+++ b/pkg/mcs/scheduling/server/cluster.go
@@ -208,26 +208,34 @@ func (c *Cluster) UpdateScheduler() {
 			s, err := schedulers.CreateScheduler(
 				scheduler.Type,
 				c.coordinator.GetOperatorController(),
-				// NOTICE: currently the scheduler config stored in this storage won't be used,
-				// we need to figure out a way to unify this storage and persisted config.
 				c.storage,
 				schedulers.ConfigSliceDecoder(scheduler.Type, scheduler.Args),
 				schedulersController.RemoveScheduler,
 			)
-			name := s.GetName()
-			if existed, _ := schedulersController.IsSchedulerExisted(name); existed {
-				continue
-			}
 			if err != nil {
 				log.Error("failed to create scheduler",
-					zap.String("scheduler-name", name), zap.Strings("scheduler-args", scheduler.Args), errs.ZapError(err))
+					zap.String("scheduler-type", scheduler.Type),
+					zap.Strings("scheduler-args", scheduler.Args),
+					errs.ZapError(err))
+				continue
+			}
+			name := s.GetName()
+			if existed, _ := schedulersController.IsSchedulerExisted(name); existed {
+				log.Info("scheduler has already existed, skip adding it",
+					zap.String("scheduler-name", name),
+					zap.Strings("scheduler-args", scheduler.Args))
+				continue
 			}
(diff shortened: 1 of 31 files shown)


pkg/mcs/scheduling/server/cluster.go
```diff
@@ -170,6 +176,85 @@ func (c *Cluster) SwitchAPIServerLeader(new pdpb.PDClient) bool {
 	return c.apiServerLeader.CompareAndSwap(old, new)
 }
 
+// UpdateScheduler listens on the schedulers updating notifier and manage the scheduler creation and deletion.
+func (c *Cluster) UpdateScheduler() {
+	defer logutil.LogPanic()
+
+	notifier := make(chan struct{}, 1)
+	c.persistConfig.SetSchedulersUpdatingNotifier(notifier)
```
rleungx, line 184: If schedulers are initialized and we add/remove a scheduler immediately before setting the notifier, will the notifier lose the update?

pkg/schedule/coordinator.go
```diff
@@ -481,6 +499,8 @@ func (c *Coordinator) InitSchedulers(needRun bool) {
 	if err := c.cluster.GetSchedulerConfig().Persist(c.cluster.GetStorage()); err != nil {
 		log.Error("cannot persist schedule config", errs.ZapError(err))
 	}
+
+	c.markSchedulersInitialized()
```
rleungx, line 503: Is it necessary for PD mode?

JmPotato: It's only used in the scheduling service, but I think leaving the mark here is okay, which is not an expensive operation.

pkg/mcs/scheduling/server/cluster.go
JmPotato: Fixed. PTAL.

tests/integrations/mcs/scheduling/server_test.go
```diff
@@ -115,3 +119,30 @@ func (suite *serverTestSuite) TestPrimaryChange() {
...
+	re := suite.Require()
+	tc, err := tests.NewTestSchedulingCluster(suite.ctx, 1, suite.backendEndpoints)
+	re.NoError(err)
+	defer tc.Destroy()
+	tc.WaitForPrimaryServing(re)
+	schedulersController := tc.GetPrimaryServer().GetCluster().GetCoordinator().GetSchedulersController()
+	re.Len(schedulersController.GetSchedulerNames(), 5)
+	re.Nil(schedulersController.GetScheduler(schedulers.EvictLeaderName))
+	// Add a new scheduler through the API server.
+	err = suite.pdLeader.GetServer().GetHandler().AddEvictLeaderScheduler(1)
```
rleungx, line 133: How about adding the evict leader for two stores?

pkg/schedule/schedulers/grant_leader.go
```diff
@@ -178,6 +178,24 @@ func (s *grantLeaderScheduler) EncodeConfig() ([]byte, error) {
 	return EncodeConfig(s.conf)
 }
 
+func (s *grantLeaderScheduler) ReloadConfig() error {
```
lhy1024, line 181: Can we reduce the duplication in ReloadConfig?

pkg/schedule/coordinator.go
```diff
@@ -91,19 +94,34 @@ func NewCoordinator(ctx context.Context, cluster sche.ClusterInformer, hbStreams
...
+
+// markSchedulersInitialized marks the scheduler initialization is finished.
+func (c *Coordinator) markSchedulersInitialized() {
+	c.Lock()
+	defer c.Unlock()
+	c.schedulersInitialized = true
+}
+
+// CheckSchedulersInitialized returns whether the schedulers have been initialized.
+func (c *Coordinator) CheckSchedulersInitialized() bool {
```
lhy1024, line 121: How about `AreSchedulersInitialized`?

pkg/schedule/schedulers/grant_leader.go
JmPotato: Different config structures need to have different implementation, I think it's OK to have some duplication here.

rleungx (approved):

lhy1024 (approved):

lhy1024 (commented):
TestSchedulerSync is failed

PR #7082, opened 2023-09-12
mcs/scheduling: register the scheduler handler in API service mode

JmPotato:

Issue Number: ref #5839.

```commit-message
To ensure that the HTTP API handler could be initialized properly for each scheduler in the API service mode,
this PR updates the scheduler controller and PD server handler to support initialize the HTTP handler only.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/server.go
+++ b/pkg/mcs/scheduling/server/server.go
@@ -47,7 +47,6 @@ import (
 	"github.com/tikv/pd/pkg/schedule"
 	"github.com/tikv/pd/pkg/schedule/hbstream"
 	"github.com/tikv/pd/pkg/schedule/schedulers"
-	"github.com/tikv/pd/pkg/storage"
 	"github.com/tikv/pd/pkg/storage/endpoint"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/apiutil"
@@ -433,7 +432,7 @@ func (s *Server) startServer() (err error) {
 
 func (s *Server) startCluster(context.Context) error {
 	s.basicCluster = core.NewBasicCluster()
-	err := s.startWatcher(s.storage)
+	err := s.startWatcher()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/mcs/scheduling/server/server.go
```diff
@@ -453,12 +454,12 @@ func (s *Server) stopCluster() {
...
 }
 
-func (s *Server) startWatcher() (err error) {
+func (s *Server) startWatcher(storage storage.Storage) (err error) {
 	s.metaWatcher, err = meta.NewWatcher(s.Context(), s.GetClient(), s.clusterID, s.basicCluster)
 	if err != nil {
 		return err
 	}
-	s.configWatcher, err = config.NewWatcher(s.Context(), s.GetClient(), s.clusterID, s.persistConfig)
+	s.configWatcher, err = config.NewWatcher(s.Context(), s.GetClient(), s.clusterID, s.persistConfig, storage)
```
rleungx, line 462: How about directly using `s.storage`?

rleungx:
How about also adding a test to validate if pd-ctl can work normally?

JmPotato:
> How about also adding a test to validate if pd-ctl can work normally?

I would like to add it in #7076 after merging this PR to test the scheduler with `pd-ctl` as well.

tests/integrations/mcs/scheduling/config_test.go
```diff
@@ -128,39 +135,81 @@ func (suite *configTestSuite) TestSchedulerConfigWatch() {
...
+				return len(evictLeaderCfg.StoreIDWithRanges) == 2
+			}
+		}
+		return false
 	})
+	// Validate the updated scheduler's config.
+	re.Contains(evictLeaderCfg.StoreIDWithRanges, uint64(1))
+	re.Contains(evictLeaderCfg.StoreIDWithRanges, uint64(2))
 	// Remove the scheduler.
 	err = suite.pdLeaderServer.GetServer().GetHandler().RemoveScheduler(schedulers.EvictLeaderName)
```
rleungx, line 206: What about deleting it one by one?

pkg/schedule/schedulers/scheduler_controller.go
```diff
@@ -117,6 +126,50 @@ func (c *Controller) ResetSchedulerMetrics() {
...
+	c.Lock()
+	defer c.Unlock()
+
+	name := scheduler.GetName()
+	if _, ok := c.schedulerHandlers[name]; ok {
+		return errs.ErrSchedulerExisted.FastGenByArgs()
+	}
+
+	c.schedulerHandlers[name] = scheduler
+	c.cluster.GetSchedulerConfig().AddSchedulerCfg(scheduler.GetType(), args)
```
rleungx, line 140: Do we need AddSchedulerCfg?

rleungx (approved):

server/handler.go
```diff
@@ -236,19 +236,24 @@ func (h *Handler) AddScheduler(name string, args ...string) error {
 		return err
 	}
 	log.Info("create scheduler", zap.String("scheduler-name", s.GetName()), zap.Strings("scheduler-args", args))
-	if !h.s.IsAPIServiceMode() {
+	if h.s.IsAPIServiceMode() {
```
lhy1024, line 239: Will we not need to forward create-scheduler and remove-scheduler requests to the scheduler server

lhy1024 (approved):

server/handler.go
JmPotato: No, the meta info should all be managed in the API server, and the scheduling service will watch them.

PR #7089, opened 2023-09-13
*: decouple operator handler and server

lhy1024:

1. move operator handler from server to pkg/operator/handler
2. move error from server to pkg/errs
3. add a interface for pd server and scheduling server

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


go.mod
```diff
@@ -145,7 +145,7 @@ require (
 	github.com/pelletier/go-toml/v2 v2.0.1 // indirect
 	github.com/petermattis/goid v0.0.0-20211229010228-4d14c490ee36 // indirect
 	github.com/pingcap/tipb v0.0.0-20220718022156-3e2483c20a9e // indirect
-	github.com/pkg/errors v0.9.1 // indirect
+	github.com/pkg/errors v0.9.1
```
rleungx, line 148: Don't import it

lhy1024: why

rleungx: We use pingcap/errors

lhy1024: Maybe because move operator error from store to pkg/errs?

pkg/errs/error.go
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package errs
+
+import "github.com/pingcap/errors"
+
+var (
+	// ErrOperatorNotFound is error info for operator not found.
+	ErrOperatorNotFound = errors.New("operator not found")
```
rleungx, line 21: What is the difference between errors.go and errno.go?

lhy1024: This pr just moves the code to decouple without changing it, do you think the two should be merged?

server/handler.go
```diff
@@ -50,44 +45,41 @@ import (
...
+	*Server
+}
+
+func (s *server) GetCoordinator() *schedule.Coordinator {
+	return s.GetRaftCluster().GetCoordinator()
+}
+
+func (s *server) GetCluster() sche.SharedCluster {
+	return s.GetRaftCluster()
+}
```
rleungx, line 61: Why do we need it?

pkg/errs/error.go
rleungx: If you want to move it to pkg/errs, IMO, we need to.

rleungx (approved):

JmPotato (approved):

PR #7097, opened 2023-09-14
mcs: update store stats

rleungx:

Issue Number: Ref #5839.

Make scheduler server can update the store statistics.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/cluster.go
+++ b/pkg/mcs/scheduling/server/cluster.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/mcs/scheduling/server/cluster.go
```diff
@@ -348,3 +348,28 @@ func (c *Cluster) HandleStoreHeartbeat(heartbeat *schedulingpb.StoreHeartbeatReq
 	c.hotStat.CheckReadAsync(statistics.NewCollectUnReportedPeerTask(storeID, regions, interval))
 	return nil
 }
+
+// RunUpdateStoreStats updates store stats periodically.
+func (c *Cluster) RunUpdateStoreStats() {
+	defer logutil.LogPanic()
+
+	ticker := time.NewTicker(9 * time.Millisecond)
```
lhy1024, line 356: 9 * time.Millisecond?

rleungx: keep the same as previous and less than the interval of store heartbeat.

rleungx: Keep the same as previous.

JmPotato (approved):

lhy1024 (approved):

PR #7122, opened 2023-09-20
*: fix sync isolation level to default placement rule

### LINKED ISSUE #7121: Isolation level is not synchronize to default rule when setting with pd-ctl
### What did you do?

Use pd-ctl to set location labels and isolation level

Both two configurations in the placement rule are updated

### What did you see instead?

Only location label is updated

rleungx:

Issue Number: Close #7121.

This PR will sync the isolation level to the default rule when setting config with HTTP API or pd-ctl.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


bufferflies (approved):

JmPotato (commented):
What if we update the `isolation-level`? Will it be set to the rule also?

rleungx:
> What if we update the `isolation-level`? Will it be set to the rule also?

Do you mean `TestUpdateDefaultReplicaConfig`?

JmPotato (approved):

PR #7128, opened 2023-09-21
schedulers: fix the grant-leader-scheuler store pause/resume

JmPotato:

Issue Number: ref #5839.

```commit-message
The grant-leader-scheduler should also check the store pause/resume after reloading the config.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


rleungx (approved):

lhy1024 (approved):
LGTM, do we need to add some test about it?

PR #7090, opened 2023-09-13
mcs: make scheduling server support operator http interface 

lhy1024:
### What problem does this PR solve?

Issue Number: Ref #5839

### What is changed and how does it work?

```commit-message
```

It can be reviewed after #7113 and #7120

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

PR #7131, opened 2023-09-21
mcs: make scheduling server support checker and scheduler http interface

lhy1024:

```commit-message
```
It can be reviewed after #7090 and #7169
### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -120,12 +119,15 @@ func NewService(srv *scheserver.Service) *Service {
 func (s *Service) RegisterSchedulersRouter() {
 	router := s.root.Group("schedulers")
 	router.GET("", getSchedulers)
+	router.GET("/diagnostic/:name", getDiagnosticResult)
```
rleungx, line 122: We don't need it for now.

pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -120,12 +119,15 @@ func NewService(srv *scheserver.Service) *Service {
...
+	router.POST("/:name", pauseOrResumeScheduler)
```
rleungx, line 123: Shall we split it into two APIs?

tests/integrations/mcs/scheduling/api_test.go
```diff
@@ -159,10 +159,17 @@ func (suite *apiTestSuite) TestAPIForward() {
 	pauseArgs, err := json.Marshal(input)
 	suite.NoError(err)
 	err = testutil.CheckPostJSON(testDialClient, fmt.Sprintf("%s/%s", urlPrefix, "checker/merge"), pauseArgs,
-		testutil.StatusOK(re), testutil.WithoutHeader(re, apiutil.PDRedirectorHeader))
```
rleungx, line 162: Why change it?

lhy1024: Because we should allow to redirect pause/resume checker

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: why? It can pass tests

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: No, we only support one-to-one forwarding at this time.

rleungx: What I mean is since we use the new framework for scheduling service, we should make it more standard. The previous one is not good enough, so we should make it more clear.

pkg/mcs/scheduling/server/apis/v1/api.go
rleungx: This feature is not widely used, to reduce the problem, I think we can hold it instead of expose it.

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: If we want to achieve one-to-two forwarding, i.e. post `scheduler/{name}` is at the api server and forwards to post `scheduler/{name}/pause` and post `scheduler/{name}/resume`, we may need to add a check for body to the match rule, which I'll try to add.

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: I think it's best to keep forwarding, one reason is that it will make matching rules simple, and another reason is to avoid us missing an interface or creating incompatibility issues

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: If we want to achieve one-to-two forwarding, i.e. post `scheduler/{name}` is at the api server and forwards to post `scheduler/{name}/pause` and post `scheduler/{name}/resume`, we may need to add a check for body to the match rule.

rleungx: We should make the new API more clear.

rleungx: If you don't want to change it and since it is only used by forwarding now, I'm ok with the current status. But in the future, it's better to separate these actions.

lhy1024: ok, I will add TODO about it.

rleungx (approved):

JmPotato (approved):

PR #7132, opened 2023-09-21
scheduler: make `evict-slow-trend` scheduler can manually modify recovery duration.

LykxSassinator:

Issue Number: Ref #7156  , Ref https://github.com/tikv/tikv/issues/15271

```commit-message
With this pr, users can manually modify the minimal recovery time when encountering an I/O jitter case.

That is, only when the jitter is disappear and the recovery time reach this limit, 
can the given slow node be mark with normal for balancing leaders to it.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/schedule/schedulers/evict_slow_trend.go
```diff
@@ -55,10 +60,27 @@ type evictSlowTrendSchedulerConfig struct {
 	// Last chosen candidate for eviction.
 	lastEvictCandidate slowCandidate
 
+	RecoveryDurationGap uint64 `json:"recovery-duration"`
```
nolouch, line 63: add comments, and it's unit is secs.

nolouch (approved):
rest lgtm

nolouch (commented):
need a reference to the issue.

LykxSassinator:
> need a reference to the issue.

Got it, I'll supplement one issue for it.

pkg/schedule/schedulers/evict_slow_trend.go
```diff
@@ -162,9 +193,50 @@ func (conf *evictSlowTrendSchedulerConfig) clearAndPersist(cluster sche.Schedule
...
+	router.HandleFunc("/list", h.ListConfig).Methods(http.MethodGet)
+	return router
+}
+
+func (handler *evictSlowTrendHandler) UpdateConfig(w http.ResponseWriter, r *http.Request) {
+	var input map[string]interface{}
+	if err := apiutil.ReadJSONRespondError(handler.rd, w, r.Body, &input); err != nil {
+		return
+	}
+	recoveryDurationGapFloat, ok := input["recovery-duration"].(float64)
```
bufferflies, line 217: why not convert to uint64 directly? BTW, Does the input needs to check also.

pkg/schedule/schedulers/evict_slow_trend.go
```diff
@@ -162,9 +193,50 @@ func (conf *evictSlowTrendSchedulerConfig) clearAndPersist(cluster sche.Schedule
...
+	if ok {
+		recoveryDurationGap := (uint64)(recoveryDurationGapFloat)
+		prevRecoveryDurationGap := atomic.LoadUint64(&handler.config.RecoveryDurationGap)
+		atomic.StoreUint64(&handler.config.RecoveryDurationGap, recoveryDurationGap)
+		log.Info("evict-slow-trend-scheduler update 'recovery-duration' - unit: s", zap.Uint64("prev", prevRecoveryDurationGap), zap.Uint64("cur", recoveryDurationGap))
+	}
```
bufferflies, line 223: you should handler the `not ok` case. user may think this behavior  is successful but the config not update.

pkg/schedule/schedulers/evict_slow_trend.go
LykxSassinator: IMO, the check is necessary. The given `body` might contain unexpected errors, which should be skipped.
And the conversion refers to:
https://github.com/tikv/pd/blob/edf1f95c2ef08dbc9af1c1f38a8ed5fdda469726/pkg/schedule/schedulers/evict_leader.go#L382.

bufferflies (approved):

PR #7139, opened 2023-09-22
tso: improve skipping sync condition and add more logs

### LINKED ISSUE #7138: SyncTimestamp might be skipped unexpectedly
## Bug Report

#7031 introduced a new path to skip `SyncTimestamp` in some cases, which might be triggered unexpectedly. For example, if it's the first time that a `timestampOracle` is created, both `last` and `lastSavedTime` could be zero and cause the sync to return early.

https://github.com/tikv/pd/blob/301b917365bb071652b297cf5cbc8e2c0e0e6ddd/pkg/tso/tso.go#L158-L175

JmPotato:

Issue Number: Close #7138.

```commit-message
Only skip the sync if the following conditions are met:

1. The timestamp in memory has been initialized.
2. The last saved timestamp in etcd is not zero.
3. The last saved timestamp in memory is not zero.
4. The last saved timestamp in etcd is equal to the last saved timestamp in memory.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/tso.go
+++ b/pkg/tso/tso.go
@@ -180,7 +180,9 @@ func (t *timestampOracle) SyncTimestamp(leadership *election.Leadership) error {
 		last != typeutil.ZeroTime &&
 		lastSavedTime != typeutil.ZeroTime &&
 		typeutil.SubRealTimeByWallClock(last, lastSavedTime) == 0 {
-		log.Info("skip sync timestamp", logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0))
+		log.Info("skip sync timestamp",
+			logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
+			zap.Time("last", last), zap.Time("last-saved", lastSavedTime))
 		t.metrics.skipSyncEvent.Inc()
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/tso/tso.go
```diff
@@ -167,9 +168,11 @@ func (t *timestampOracle) SyncTimestamp(leadership *election.Leadership) error {
...
 	}
 	lastSavedTime := t.getLastSavedTime()
-	// If `lastSavedTime` is not zero, it means that the `timestampOracle` has already been initialized
-	// before, so we could safely skip the sync if `lastSavedTime` is equal to `last`.
-	if lastSavedTime != typeutil.ZeroTime && typeutil.SubRealTimeByWallClock(lastSavedTime, last) == 0 {
+	// If `last` and `lastSavedTime` are the same and both not zero, it means that the
+	// timestamp has been synchronized before. We could skip the synchronization safely.
+	if last != typeutil.ZeroTime && lastSavedTime != typeutil.ZeroTime &&
+		typeutil.SubRealTimeByWallClock(lastSavedTime, last) == 0 {
+		log.Info("skip sync timestamp", zap.Uint32("keyspace-group-id", t.keyspaceGroupID))
```
rleungx, line 175: Will the keyspace group id be printed in non serverless environment?

JmPotato: I want to retain this field to help us better debug, but in order not to trouble users in the OP environment so much, I have hidden the keyspace-related words. Do you think it is okay?

rleungx: How about using CondUint32

JmPotato: Nice suggestion. Addressed.

rleungx (approved):

lhy1024 (approved):

PR #7157, opened 2023-09-26
*: use syncutil lock

rleungx:

Issue Number: Ref #4399

Replace sync.RWMutex/Mutex with syncutil which will be changed to deadlock one in tests.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/balancer/round_robin.go
+++ b/pkg/balancer/round_robin.go
@@ -15,13 +15,14 @@
 package balancer
 
 import (
-	"sync"
 	"sync/atomic"
+
+	"github.com/tikv/pd/pkg/utils/syncutil"
 )
 
 // RoundRobin is a balancer that selects nodes in a round-robin fashion.
 type RoundRobin[T uint32 | string] struct {
-	sync.RWMutex
+	syncutil.RWMutex
 	nodes  []T
 	exists map[T]struct{}
 	next   uint32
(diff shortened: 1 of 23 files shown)


pkg/balancer/round_robin.go
```diff
@@ -15,13 +15,14 @@
...
-	"sync"
 	"sync/atomic"
+
+	"github.com/tikv/pd/pkg/utils/syncutil"
 )
 
 // RoundRobin is a balancer that selects nodes in a round-robin fashion.
 type RoundRobin[T uint32 | string] struct {
-	sync.RWMutex
+	syncutil.RWMutex
```
lhy1024, line 25: What difference are there?

rleungx: If we run `make test`, it will use the deadlock RWMutex.

lhy1024 (approved):

HuSharp (commented):
I noticed a few `sync.RWMutex`s have not been changed, does this make sense?
For example, in `keyspaceGroupSvcDiscovery`

rleungx:
> I noticed a few `sync.RWMutex`s have not been changed, does this make sense? For example, in `keyspaceGroupSvcDiscovery`

They are in client module

HuSharp (approved):
LGTM!

HuSharp:
> > I noticed a few `sync.RWMutex`s have not been changed, does this make sense? For example, in `keyspaceGroupSvcDiscovery`
> 
> They are in client module

get

PR #7158, opened 2023-09-27
pd-ctl: add keyspace commands

AmoebaProtozoa:

Add Keyspace related commands to pd-ctl, so they can be accessed via pd-ctl, which gives better help information than using curl to access RESTful api.

#### Create:
```
Usage:
  pd-ctl keyspace create <keyspace-name> [flags]

Flags:
      --config strings   keyspace configs for the new keyspace
                         specify as comma separated key value pairs, e.g. --config k1=v1,k2=v2
```
#### Show:
```
Usage:
  pd-ctl keyspace show id <id> [flags]
  pd-ctl keyspace show name <name> [flags]
```
#### List:
```
Usage:
  pd-ctl keyspace list [flags]

Flags:
      --limit string        The maximum number of keyspace metas to return. If not set, no limit is posed.
      --page_token string   The keyspace id of the scan start. If not set, scan from keyspace/keyspace group with id 0
```
#### Update State:
```
Usage:
  pd-ctl keyspace update-state <keyspace-name> <state> [flags]
```
#### Update Config:
```
Usage:
  pd-ctl keyspace update-config <keyspace-name> [flags]

Flags:
      --remove strings   keys to remove from keyspace config
                         specify as comma separated keys, e.g. --remove k1,k2
      --update strings   kv pairs to upsert into keyspace config
                         specify as comma separated key value pairs, e.g. --update k1=v1,k2=v2
```

Issue Number: Ref #4399 

### Check List

Tests

- Unit tests

### Release note

```release-note
None.
```


tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -15,33 +15,216 @@
...
+		cmd.Printf("Failed to create the keyspace: %s\n", err)
+		return
+	}
+	cmd.Println(resp)
+}
+
+func newUpdateKeyspaceConfigCommand() *cobra.Command {
+	r := &cobra.Command{
+		Use:   "update-config <keyspace-name> <config-patch>",
+		Short: "update keyspace config, a json merge patch is expected",
```
rleungx, line 139: Is it more convenient to use JSON or use key-value list? BTW, do we need a force overwrite option?

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -15,33 +15,216 @@
...
 }
 
+func newShowKeyspaceCommand() *cobra.Command {
+	r := &cobra.Command{
+		Use:   "show [flags]",
+		Short: "show keyspace metadata specified by keyspace name/id, if both provided, name will be used",
+		Run:   showKeyspaceCommandFunc,
+	}
+	r.Flags().String(nmID, "", "keyspace id")
+	r.Flags().String(nmName, "", "keyspace name")
```
rleungx, line 58: I prefer to use the subcommand here so that we can avoid setting them at the same time.

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -15,33 +15,216 @@
...
 // NewKeyspaceCommand returns a keyspace subcommand of rootCmd.
 func NewKeyspaceCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "keyspace [command] [flags]",
-		Short: "show keyspace information",
-		Run:   showKeyspaceCommandFunc,
+		Use:   "keyspace <command> [flags]",
+		Short: "keyspace commands",
 	}
+	cmd.AddCommand(newShowKeyspaceCommand())
```
rleungx, line 43: Do we need a unit test?

tools/pd-ctl/pdctl/command/keyspace_command.go
AmoebaProtozoa: done

tools/pd-ctl/pdctl/command/keyspace_command.go
AmoebaProtozoa: I have changed the json to comma separated kv pairs.

tools/pd-ctl/pdctl/command/keyspace_command.go
AmoebaProtozoa: unit tests added.

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -15,33 +15,278 @@
...
+			if _, exist := configPatch[key]; exist {
+				cmd.PrintErrf("Failed to update keyspace config: key %s is specified multiple times\n", key)
+				return
+			}
+			configPatch[key] = nil
+		}
+	}
+	updateFlags, err := cmd.Flags().GetStringSlice(nmUpdate)
+	if err != nil {
+		cmd.Println("Failed to parse flag: ", err)
```
rleungx, line 189: I find in this PR, some use Println and the others use PrintErrln.

AmoebaProtozoa: Ok, that one is fixed.
The description of `PrintErrln` and `Println` state that one print to defined output and the other prints to defined err output, so I use `PrintErrln` whenever theres an error and `Println` for regular response.

But it seems that underlying implementations of `PrintErrln` and `Println` are the same, which is quite different from `Print` vs `PrintErr`, and they are indeed used quite interchangeably in other commands of pd-ctl.
https://github.com/spf13/cobra/blob/4dd4b25de38418174a6e859e8a32eaccca32dccc/command.go#L1369

AmoebaProtozoa: It's now fixed.

In general, I use `Println` for response, and `PrintErrln` for any error occurred.

I believe the version of cobra we used does not really make difference between PrintErrln and Println, despite the description of them, resulting in that other places of pd-ctl use them interchangeably (they have the same implementation anyway). 

[This is fixed in newer version of cobra](https://github.com/spf13/cobra/pull/894), and we probably need to do some cleanup when we update the reference.

rleungx (approved):

ystaticy (approved):

nolouch (approved):

PR #7163, opened 2023-09-27
*: prevent store config updating periodically

rleungx:

Issue Number: Ref #5839.

This PR reduces the time for unnecessary updating store config and adds more logs to help us debug the problem.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/config/watcher.go
+++ b/pkg/mcs/scheduling/server/config/watcher.go
@@ -119,6 +119,7 @@ func (cw *Watcher) initializeConfigWatcher() error {
 				zap.String("event-kv-key", string(kv.Key)), zap.Error(err))
 			return err
 		}
+		log.Info("update scheduling config", zap.Reflect("new", cfg))
 		cw.AdjustScheduleCfg(&cfg.Schedule)
 		cw.SetClusterVersion(&cfg.ClusterVersion)
 		cw.SetScheduleConfig(&cfg.Schedule)
@@ -146,6 +147,7 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 	prefixToTrim := cw.schedulerConfigPathPrefix + "/"
 	putFn := func(kv *mvccpb.KeyValue) error {
 		name := strings.TrimPrefix(string(kv.Key), prefixToTrim)
+		log.Info("update scheduler config", zap.Reflect("name", string(kv.Value)))
 		err := cw.storage.SaveSchedulerConfig(name, kv.Value)
 		if err != nil {
 			log.Warn("failed to save scheduler config",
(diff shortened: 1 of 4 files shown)


pkg/mcs/scheduling/server/rule/watcher.go
```diff
@@ -163,12 +165,14 @@ func (rw *Watcher) initializeRuleWatcher() error {
...
 		// Since the PD API server will validate the rule before saving it to etcd,
 		// so we could directly save the string rule in JSON to the storage here.
+		log.Info("update placement rule", zap.String("key", string(kv.Key)), zap.String("value", string(kv.Value)))
 		return rw.ruleStore.SaveRule(
 			strings.TrimPrefix(string(kv.Key), prefixToTrim),
 			string(kv.Value),
 		)
 	}
 	deleteFn := func(kv *mvccpb.KeyValue) error {
+		log.Info("delete placement rule", zap.String("key", string(kv.Key)))
```
lhy1024, line 175: Will it print a lot of logs?

rleungx: If we frequently add/remove the rule, it will, like the current status of the audit log. However, we cannot change the log level for microservices. Once it is supported, I can change it to the debug level if it's too noisy. Right now, it is more convenient for us to test.

server/cluster/cluster.go
```diff
@@ -441,11 +441,14 @@ func (c *RaftCluster) runStoreConfigSync() {
 				log.Warn("store config persisted failed", zap.Error(err))
 			}
 		}
+		isLastSynced = synced
 		// Update the stores if the synchronization is not completed.
 		if !synced {
 			stores = c.GetStores()
-		} else if err := c.opt.Persist(c.storage); err != nil {
-			log.Warn("store config persisted failed", zap.Error(err))
+		} else if !isLastSynced && synced {
```
JmPotato, line 448: This seems to cause the config only to be saved once forever.

rleungx: Oh, nice catch, let me think about how to fix it.

JmPotato: `observeStoreConfig` will check if the config fetched is equal to the old one, maybe we could use that as a flag to check if the persistence is necessary.

rleungx: Updated. PTAL

JmPotato (approved):

pkg/mcs/scheduling/server/config/watcher.go
```diff
@@ -146,6 +147,7 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 	prefixToTrim := cw.schedulerConfigPathPrefix + "/"
 	putFn := func(kv *mvccpb.KeyValue) error {
 		name := strings.TrimPrefix(string(kv.Key), prefixToTrim)
+		log.Info("update scheduler config", zap.Reflect("name", string(kv.Value)))
```
bufferflies, line 150: why not use zap.String?

lhy1024 (approved):

pkg/mcs/scheduling/server/config/watcher.go
rleungx: updated

PR #7183, opened 2023-10-09
mcs: support history hot region in scheduling server

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/server.go
```diff
@@ -419,6 +466,18 @@ func (s *Server) startServer() (err error) {
 		return err
 	}
 
+	s.encryptionKeyManager, err = encryption.NewManager(s.GetClient(), &s.cfg.Security.Encryption)
+	if err != nil {
+		return err
+	}
+
+	h := handler.NewHandler(&svr{s})
+	s.hotRegionStorage, err = storage.NewHotRegionsStorage(
```
rleungx, line 475: Will it make the scheduling server stateful？

lhy1024: It will create a leveldb in scheduling server. 
If hope to store it in api server, it means that we need to pull all hot region info from scheduling server per 10 minutes
https://docs.pingcap.com/zh/tidb/stable/information-schema-tidb-hot-regions-history

rleungx: Scheduling server should be stateless.

lhy1024:
After discussion, this pr will be suspended.

PR #7190, opened 2023-10-11
mcs: support forwarding split and scatter request

rleungx:

Issue Number: Ref #5839.

Before this PR, the API service was responsible for scatter and split. After this PR, the request will be forwarded to the scheduling service. Once we implement the scheduling service discovery, we can directly send requests to the scheduling service.

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -191,6 +185,48 @@ func (s *Service) StoreHeartbeat(ctx context.Context, request *schedulingpb.Stor
 	return &schedulingpb.StoreHeartbeatResponse{Header: &schedulingpb.ResponseHeader{ClusterId: s.clusterID}}, nil
 }
 
+// SplitRegions split regions by the given split keys
+func (s *Service) SplitRegions(ctx context.Context, request *schedulingpb.SplitRegionsRequest) (*schedulingpb.SplitRegionsResponse, error) {
```
lhy1024, line 189: Do we need to add some tests about forward requests?

rleungx: Yes, I have added a TODO in the PR description. Will do it later.

lhy1024, line 189: BTW, we seem to need to support `AskSplit`, `AskBatchSplit`, and `GetOperator` with grpc method too.

rleungx: We don't need AskSplit/AskBatchSplit. But GetOperator is necessary.

lhy1024: `HandleAskBatchSplit` will call `GetMergeChecker()`

rleungx: I think we can move the split cache out, WDYT?

rleungx: Oh, you are right.

rleungx: Will do it in the next PR

JmPotato (approved):
LGTM. Please update the kvproto first.

server/grpc_service.go
```diff
@@ -2175,7 +2283,7 @@ func (s *GrpcServer) SplitRegions(ctx context.Context, request *pdpb.SplitRegion
 }
 
 // SplitAndScatterRegions split regions by the given split keys, and scatter regions.
-// Only regions which splited successfully will be scattered.
+// Only regions which split successfully will be scattered.
 // scatterFinishedPercentage indicates the percentage of successfully splited regions that are scattered.
 func (s *GrpcServer) SplitAndScatterRegions(ctx context.Context, request *pdpb.SplitAndScatterRegionsRequest) (*pdpb.SplitAndScatterRegionsResponse, error) {
```
lhy1024, line 2288: Do we need to forward this functions?

lhy1024 (commented):
the rest LGTM

server/grpc_service.go
rleungx: AFAIK, this API is not used for now.

lhy1024 (approved):

PR #7202, opened 2023-10-13
replication_mode: use placement to determin canSync and hasMajority

### LINKED ISSUE #7201: dr-autosync: `canSync` and `hasMajority` can be calculated wrong when placement rule config is complex
### What did you do?
deploy a dr-autosync cluster, setup 3 voters in primary dc, 2 voters in dr dc. the tikv-servers in dr dc splits into 2 different logic zones and use placement rule to place 1 voter in each logic zone.

kill all tikv-servers in a logic zone, the dr-autosync state should not switch to async_wait or async, as there is still 1 voter in dr zone.

### What did you see instead?
the state is switched to async_wait and then async.

master / v6.5.5

disksing:
Issue Number: Close #7201

- determine `canSync` and `hasMajority` flags based on placement rule configuration.

```commit-message
```

### Check List

Tests
- Unit test

Code changes

`primary-replicas` and `dr-replicas` configurations are not deprecated.

Related changes

### Release note

```release-note
None.
```


nolouch (approved):
lgtm

CabinfeverB (approved):

PR #7205, opened 2023-10-16
scheduling/watcher, storage: store rules into the unified storage

### LINKED ISSUE #7200: mcs: support placement rules in scheduling server
## Development Task

Now, we can watch the change of placement rules with `ruleWatcher`, but we don't use it in the scheduling server with `GetRuleStorage`, which should be supported.

BTW, after the change of placement rules, we also need to call `AddSuspectKeyRange`, which is similar to `/config/rule [post]`, `/config/rule/{group}/{id} [delete]`, `/config/rule_group [post]`, `/config/rule_group/{id} [delete]`

JmPotato:

Issue Number: Close #7200.

```commit-message
Store rules into the unified storage rather than the isolated implementation.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


lhy1024 (approved):

pkg/mcs/scheduling/server/server.go
```diff
@@ -479,7 +479,7 @@ func (s *Server) startWatcher() (err error) {
 	if err != nil {
 		return err
 	}
-	s.ruleWatcher, err = rule.NewWatcher(s.Context(), s.GetClient(), s.clusterID)
+	s.ruleWatcher, err = rule.NewWatcher(s.Context(), s.GetClient(), s.clusterID, s.storage)
```
lhy1024, line 482: BTW, after the change of placement rules, we also need to call `AddSuspectKeyRange`, which is similar to `/config/rule [post]`, `/config/rule/{group}/{id} [delete]`, `/config/rule_group [post]`, `/config/rule_group/{id} [delete]`

JmPotato: I will implement this in a separate PR.

rleungx (approved):

pkg/mcs/scheduling/server/server.go
JmPotato: Implemented in https://github.com/tikv/pd/pull/7213.

Issue #7185, opened 2023-10-10
Placement cannot add learner with the rule


nolouch:
### What did you do?

region: 
```
[root@tc-pd-0 /]# ./pd-ctl region 10894
{
  "id": 10894,
  "start_key": "7480000000000003FFBD00000000000000F8",
  "end_key": "7480000000000003FFBE00000000000000F8",
  "epoch": {
    "conf_ver": 95,
    "version": 478
  },
  "peers": [
    {
      "role_name": "Voter",
      "id": 10895,
      "store_id": 7
    },
    {
      "role_name": "Voter",
      "id": 10898,
      "store_id": 33
    },
    {
      "role_name": "Voter",
      "id": 10899,
      "store_id": 16
    }
  ],
  "leader": {
    "role_name": "Voter",
    "id": 10895,
    "store_id": 7
  },
  "cpu_usage": 0,
  "written_bytes": 0,
  "read_bytes": 0,
  "written_keys": 0,
  "read_keys": 0,
  "approximate_size": 1,
  "approximate_keys": 0
}

```

ruleFit: 

```
[root@tc-pd-0 /]# curl http://127.0.0.1:2379/pd/api/v1/config/rules/region/10894/detail
{
  "rule-fits": [
    {
      "rule": {
        "group_id": "TiDB_DDL_957",
        "id": "table_rule_957_0",
        "index": 40,
        "start_key": "7480000000000003ffbd00000000000000f8",
        "end_key": "7480000000000003ffbe00000000000000f8",
        "role": "voter",
        "is_witness": false,
        "count": 3,
        "label_constraints": [
          {
            "key": "engine",
            "op": "notIn",
            "values": [
              "tiflash"
            ]
          }
        ],
        "version": 5,
        "create_timestamp": 1696924085
      },
      "peers": [
        {
          "id": 10895,
          "store_id": 7
        },
        {
          "id": 10898,
          "store_id": 33
        },
        {
          "id": 10899,
          "store_id": 16
        }
      ],
      "peers-different-role": null,
      "isolation-score": 0,
      "witness-score": 0
    },
    {
      "rule": {
        "group_id": "TiDB_DDL_957",
        "id": "table_rule_957_1",
        "index": 40,
        "start_key": "7480000000000003ffbd00000000000000f8",
        "end_key": "7480000000000003ffbe00000000000000f8",
        "role": "learner",
        "is_witness": false,
        "count": 3,
        "label_constraints": [
          {
            "key": "disk",
[...]

PR #7219, opened 2023-10-17
rule_checker: fix the issue of not being able to achieve the better RegionFit

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/7185

```commit-message
fix the issue of not being able to achieve the better RegionFit
- try to replace the existing peer in another rulefit
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/rule_checker.go
+++ b/pkg/schedule/checker/rule_checker.go
@@ -60,6 +60,7 @@ var (
 	ruleCheckerReplaceOfflineCounter              = checkerCounter.WithLabelValues(ruleChecker, "replace-offline")
 	ruleCheckerAddRulePeerCounter                 = checkerCounter.WithLabelValues(ruleChecker, "add-rule-peer")
 	ruleCheckerNoStoreAddCounter                  = checkerCounter.WithLabelValues(ruleChecker, "no-store-add")
+	ruleCheckerNoStoreThenTryReplace              = checkerCounter.WithLabelValues(ruleChecker, "no-store-then-try-replace")
 	ruleCheckerNoStoreReplaceCounter              = checkerCounter.WithLabelValues(ruleChecker, "no-store-replace")
 	ruleCheckerFixPeerRoleCounter                 = checkerCounter.WithLabelValues(ruleChecker, "fix-peer-role")
 	ruleCheckerFixLeaderRoleCounter               = checkerCounter.WithLabelValues(ruleChecker, "fix-leader-role")
@@ -185,7 +186,7 @@ func (c *RuleChecker) isWitnessEnabled() bool {
 func (c *RuleChecker) fixRulePeer(region *core.RegionInfo, fit *placement.RegionFit, rf *placement.RuleFit) (*operator.Operator, error) {
 	// make up peers.
 	if len(rf.Peers) < rf.Rule.Count {
-		return c.addRulePeer(region, rf)
+		return c.addRulePeer(region, fit, rf)
 	}
 	// fix down/offline peers.
 	for _, peer := range rf.Peers {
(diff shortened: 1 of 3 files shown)


nolouch:
ptal @CabinfeverB @rleungx

pkg/schedule/checker/rule_checker_test.go
```diff
@@ -1583,3 +1586,142 @@ func (suite *ruleCheckerTestSuite) TestTiFlashLocationLabels() {
...
+	for _, store := range stores.GetStores() {
+		suite.cluster.PutStore(store)
+	}
+
+	testCases := []struct {
+		region string
+		rules  []string
+		opStr  string
+	}{
+		{"111_leader,211,311", []string{"3/voter//", "3/learner/type=read/"}, "replace-rule-swap-fit-peer {mv peer: store [111] to"},
```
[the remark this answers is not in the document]
CabinfeverB: And IMO, no need to remove the old peer, just add voter and demote old peer.

pkg/schedule/checker/rule_checker.go
```diff
@@ -229,6 +230,25 @@ func (c *RuleChecker) addRulePeer(region *core.RegionInfo, rf *placement.RuleFit
...
+		// issue: https://github.com/tikv/pd/issues/7185
+		for _, p := range region.GetPeers() {
+			s := c.cluster.GetStore(p.GetStoreId())
+			if placement.MatchLabelConstraints(s, rf.Rule.LabelConstraints) {
+				oldPeerRuleFit := fit.GetRuleFit(p.GetId())
+				if oldPeerRuleFit == nil || !oldPeerRuleFit.IsSatisfied() || oldPeerRuleFit == rf {
+					continue
+				}
+				ruleCheckerNoStoreThenTryReplace.Inc()
+				op, err := c.replaceUnexpectRulePeer(region, oldPeerRuleFit, fit, p, "swap-fit")
```
CabinfeverB, line 243: ref https://github.com/tikv/pd/pull/7219/files#r1367687550, maybe another op?

nolouch: I want to reuse the old function and make it right first, maybe improve it after this pr.

pkg/schedule/checker/rule_checker_test.go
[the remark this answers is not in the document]
nolouch: > `mv peer: store [111] to` pls add target into op desc?

The target is random here.

CabinfeverB (approved):

pkg/schedule/checker/rule_checker_test.go
```diff
@@ -1583,3 +1586,142 @@ func (suite *ruleCheckerTestSuite) TestTiFlashLocationLabels() {
...
+			leader = peer
+			regionMeta.Id = peer.Id - 1
+		}
+	}
+	return core.NewRegionInfo(&regionMeta, leader)
+}
+
+// example: "3/voter/zone=zone1+zone2,rack=rack2/zone,rack,host"
+// count role constraints location_labels
+func makeRule(def string) *placement.Rule {
```
bufferflies, line 1663: how about use this method `https://github.com/tikv/pd/blob/4b7e79209c11f35e1e326c1864b944c1c18d941e/pkg/schedule/placement/fit_test.go#L80`

pkg/schedule/checker/rule_checker_test.go
```diff
@@ -1583,3 +1586,142 @@ func (suite *ruleCheckerTestSuite) TestTiFlashLocationLabels() {
...
+				}
+				stores.SetStore(core.NewStoreInfoWithLabel(id, labels).Clone(core.SetLastHeartbeatTS(now), core.SetStoreState(metapb.StoreState_Up)))
+			}
+		}
+	}
+	return stores
+}
+
+// example: "1111_leader,1234,2111_learner"
+func makeRegion(def string) *core.RegionInfo {
```
bufferflies, line 1641: ditto 'https://github.com/HunDunDM/pd/blob/4b7e79209c11f35e1e326c1864b944c1c18d941e/pkg/schedule/placement/fit_test.go#L58-L76'

bufferflies (approved):
rest LGTM

pkg/schedule/checker/rule_checker_test.go
nolouch: yes, it's a copy of that because not in the same package, do not let the test have dependence first. maybe extract to unit package later.

PR #7222, opened 2023-10-18
replication mode: fix wrong available store list

disksing:
### What problem does this PR solve?
Issue Number: Close #7221 #7218

### What is changed and how does it work?
As now `canSync` and `hasMajority` are calculated using placement rules, we can contain stores that have only learners in avalable stores list.

This PR also fix a minor bug that relates to available stores id comparsion.

```commit-message
```

### Check List

Tests
- Unit test

Related changes
- Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

PR #7223, opened 2023-10-18
dr: fix store learner check 

### LINKED ISSUE #7218: dr: cannot add new stores in primary
### What did you do?

test dr-auto-sync

1. down dr 
2. add new stores in primary 

tikv can add new peers in new stores 

### What did you see instead?

tikv cannot add new peers and log print "drop msg"

v6.5.4

lhy1024:

Issue Number: Close #7218

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024:
it is duplicated with #7222

PR #7278, opened 2023-10-26
mcs: support config http interface in scheduling server

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/api/config.go
```diff
@@ -60,7 +62,36 @@ func newConfHandler(svr *server.Server, rd *render.Render) *confHandler {
 // @Router   /config [get]
 func (h *confHandler) GetConfig(w http.ResponseWriter, r *http.Request) {
 	cfg := h.svr.GetConfig()
-	cfg.Schedule.MaxMergeRegionKeys = cfg.Schedule.GetMaxMergeRegionKeys()
+	if h.svr.IsAPIServiceMode() {
+		b, err := h.GetSchedulingServerConfig("config/schedule")
```
rleungx, line 66: How about just calling config??

server/api/config.go
```diff
@@ -60,7 +62,24 @@ func newConfHandler(svr *server.Server, rd *render.Render) *confHandler {
...
+		}
+		var configSchedulingServer config.Config
+		err = json.Unmarshal(b, &configSchedulingServer)
+		if err != nil {
+			h.rd.JSON(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+		cfg.Schedule = configSchedulingServer.Schedule
+		cfg.Replication = configSchedulingServer.Replication
+		// TODO: will we support config/store?
```
lhy1024, line 79: TODO: will we support config/store?

tests/server/config/config_test.go
```diff
@@ -74,3 +82,517 @@ func TestRateLimitConfigReload(t *testing.T) {
...
+}
+
+func createTTLUrl(url string, ttl int) string {
+	return fmt.Sprintf("%s/pd/api/v1/config?ttlSecond=%d", url, ttl)
+}
+
+func (suite *configTestSuite) TestConfigTTL() {
+	env := tests.NewSchedulingTestEnvironment(suite.T())
+	// FIXME: enable this test in two modes after ttl config is supported.
+	env.RunTestInPDMode(suite.checkConfigTTL)
```
lhy1024, line 505: https://github.com/tikv/pd/issues/7296

tests/server/config/config_test.go
```diff
@@ -74,3 +82,517 @@ func TestRateLimitConfigReload(t *testing.T) {
...
+	postData, err = json.Marshal(mergeConfig)
+	suite.NoError(err)
+	err = tu.CheckPostJSON(testDialClient, createTTLUrl(urlPrefix, 10), postData, tu.StatusOK(re))
+	suite.NoError(err)
+	suite.assertTTLConfigItemEqaul(cluster, "enable-tikv-split-region", true)
+}
+
+func (suite *configTestSuite) TestTTLConflict() {
+	env := tests.NewSchedulingTestEnvironment(suite.T())
+	// FIXME: enable this test in two modes after ttl config is supported.
```
lhy1024, line 567: https://github.com/tikv/pd/issues/7296

pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -126,6 +127,15 @@ func (s *Service) RegisterAdminRouter() {
 	router.DELETE("cache/regions/:id", deleteRegionCacheByID)
 }
 
+// RegisterConfigRouter registers the router of the config handler.
+func (s *Service) RegisterConfigRouter() {
+	router := s.root.Group("config")
+	router.GET("", getConfig)
+	router.GET("/schedule", getScheduleConfig)
+	router.GET("/replicate", getReplicationConfig)
+	router.GET("/store", getStoreConfig)
```
rleungx, line 136: I was wondering if we need these APIs, I think they are barely used.

lhy1024: Because api server has the same interface. Maybe we also call "/config" in api server for these interfaces?

rleungx: I think it's better, WDYT?

lhy1024: I think we can cancel "/config/schedule" and "/config/replicate", which are in "/config". But we should keep "config/store", otherwise we have no way to get store config, which is not in "/config".

lhy1024: Or we can put store config into "/config" and add a new struct for it. How about it?
```
type ConfigPayload{
   Schedule
   Replication
   StoreConfig
   ...
}
```

rleungx: OK, make sense to me.

lhy1024: Or only remove "/config/store", I found there is no "/config/store" in api server

rleungx (approved):

JmPotato (approved):

PR #7279, opened 2023-10-26
mcs: support admin/cache http interface in scheduling server

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/apis/v1/api.go
+++ b/pkg/mcs/scheduling/server/apis/v1/api.go
@@ -15,7 +15,6 @@
 package apis
 
 import (
-	"fmt"
 	"net/http"
 	"strconv"
 	"sync"
@@ -26,6 +25,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/joho/godotenv"
 	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/errs"
 	scheserver "github.com/tikv/pd/pkg/mcs/scheduling/server"
 	mcsutils "github.com/tikv/pd/pkg/mcs/utils"
 	sche "github.com/tikv/pd/pkg/schedule/core"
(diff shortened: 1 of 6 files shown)


pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -121,6 +121,9 @@ func NewService(srv *scheserver.Service) *Service {
 func (s *Service) RegisterAdminRouter() {
 	router := s.root.Group("admin")
 	router.PUT("/log", changeLogLevel)
+	cacheRouter := router.Group("cache/regions")
```
rleungx, line 124: How about only cache or just not create a new group?

lhy1024: Do you mean `cacheRouter := s.root.Group("admin/cache/regions")`

rleungx: cacheRouter := s.root.Group("cache") or just use router.DELETE("cache/regions", xxx)

tests/integrations/mcs/scheduling/api_test.go
```diff
@@ -218,3 +220,57 @@ func (suite *apiTestSuite) TestAPIForward() {
 		testutil.WithHeader(re, apiutil.ForwardToMicroServiceHeader, "true"))
 	re.NoError(err)
 }
+
+func TestAdminRegionCache(t *testing.T) {
```
rleungx, line 224: How about also checking if both caches can be cleaned in microservice mode?

lhy1024: TestAdminRegionCacheForward tests it by "/pd/api/v1/admin/cache/region"

rleungx: What I mean is if we only call the API server's interface, both caches can be cleaned in microservice mode.

rleungx (approved):

nolouch (approved):

Issue #7300, opened 2023-11-02
Tracking issue for the HTTP SDK


JmPotato:
## Development Task

As https://github.com/pingcap/tidb/issues/35319 mentioned, providing a unified PD HTTP API SDK has become an urgent need. With this means, it could be easier for us to control and observe the use of the PD HTTP API in general.

PR #7304, opened 2023-11-02
client: introduce the HTTP client

JmPotato:

Issue Number: ref #7300.

```commit-message
Introduce the HTTP client.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/http/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+package http
+
+import "time"
+
+// NOTICE: the structures below are copied from the PD API definitions.
+// Please make sure the consistency if any change happens to the PD API.
+
+// RegionInfo stores the information of one region.
+type RegionInfo struct {
```
rleungx, line 23: Is it possible to share the same definition with api package?

JmPotato: Because we cannot reference each other between the tikv/pd and tikv/pd/client packages, it probably won't work if we don't open a separate package to define these types.

rleungx (approved):

client/http/client.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	return nil
+}
+
+// GetRegionByID gets the region info by ID.
+func (c *client) GetRegionByID(ctx context.Context, regionID uint64) (*RegionInfo, error) {
+	var region RegionInfo
+	err := c.request(ctx, "GetRegionByID", RegionByID(regionID), &region)
```
nolouch, line 175: TiDB has a metric like `tidb_server_pd_api_execution_duration`.  before using this `sdk`, It should also be considered to pass the metric into the client for compatibility.

nolouch (commented):
rest lgtm

client/http/types.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// RegionsInfo stores the information of regions.
+type RegionsInfo struct {
+	Count   int64        `json:"count"`
+	Regions []RegionInfo `json:"regions"`
+}
+
+// Merge merges two RegionsInfo together and returns a new one.
+func (ri *RegionsInfo) Merge(other *RegionsInfo) *RegionsInfo {
```
CabinfeverB, line 78: Where it is used?

client/http/client.go
```diff
@@ -0,0 +1,280 @@
...
+// Close closes the HTTP client.
+func (c *client) Close() {
+	if c.cli != nil {
+		c.cli.CloseIdleConnections()
+	}
+	log.Info("[pd] http client closed")
+}
+
+func (c *client) pdAddr() string {
+	// TODO: support the customized PD address selection strategy.
```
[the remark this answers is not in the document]
JmPotato: I implemented it to make it the same as https://github.com/pingcap/tidb/blob/6e8df186f51b90d57e29875b03afe317298bc949/pkg/store/helper/helper.go#L900-L906.

client/http/client.go
JmPotato: Done.

client/http/types.go
JmPotato: After checking, I found that it was not used. Removed.

JmPotato: It will be used here:

```shell
pkg/executor/infoschema_reader.go:1609:37: allRegionsInfo.Merge undefined (type *"github.com/tikv/pd/client/http".RegionsInfo has no field or method Merge)
```

JmPotato: https://github.com/pingcap/tidb/blob/6e8df186f51b90d57e29875b03afe317298bc949/pkg/executor/infoschema_reader.go#L1609

client/http/client.go
JmPotato: Local test with replaced PD HTTP client:

image

CabinfeverB (approved):

client/http/types.go
CabinfeverB: It's better to add a unit test for this function

PR #7305, opened 2023-11-02
*: Improve region forward

rleungx:

Issue Number: Ref #5839.

Improve the region forwarding logic.

```commit-message
```

### Check List

Tests

- Manual test

### Release note

```release-note
None.
```


server/forward.go
```diff
@@ -0,0 +1,504 @@
...
+	case tsDeadlineCh <- dl:
+	case <-forwardCtx.Done():
+		return nil, forwardCtx.Err()
+	}
+
+	start := time.Now()
+	resp, err := s.forwardTSORequest(forwardCtx, request, forwardStream)
+	close(done)
+	if err != nil {
+		if strings.Contains(err.Error(), errs.NotLeaderErr) {
```
CabinfeverB, line 207: How about add a `handleErr` func?

CabinfeverB (approved):

server/forward.go
rleungx: I don't change the logic here, maybe we can keep it as it is.

server/forward.go
```diff
@@ -0,0 +1,504 @@
...
+		// Successfully stored the connection we created.
+		return newConn, nil
+	}
+	// Loaded a connection created/stored by another goroutine, so close the one we created
+	// and return the one we loaded.
+	newConn.Close()
+	return conn.(*grpc.ClientConn), nil
+}
+
+func (s *GrpcServer) getForwardedHost(ctx, streamCtx context.Context, serviceName ...string) (forwardedHost string, err error) {
```
lhy1024, line 366:
```suggestion
func (s *GrpcServer) getForwardedHost(ctx, streamCtx context.Context, serviceNames ...string) (forwardedHost string, err error) {
```

rleungx: It's just an optional parameter in stead of plural.

lhy1024 (approved):

PR #7321, opened 2023-11-06
mcs: solve forward stream error

### LINKED ISSUE #7320: The error of forwarding TSO should be handled
## Bug Report

https://github.com/tikv/pd/blob/ab8bf7b7a62d43abd9a8d5213fc3b5855472bd9e/server/grpc_service.go#L2626-L2653
https://github.com/tikv/pd/blob/ab8bf7b7a62d43abd9a8d5213fc3b5855472bd9e/server/grpc_service.go#L2635

The error of sending the request is not properly handled. We should do the error handling in case there is something wrong when forwarding the request to get TSO.

lhy1024:

Issue Number: Close #7320

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -2634,7 +2634,10 @@ func (s *GrpcServer) getGlobalTSOFromTSOServer(ctx context.Context) (pdpb.Timest
 		}
 		err := forwardStream.Send(request)
 		if err != nil {
-			return pdpb.Timestamp{}, err
+			s.tsoClientPool.Lock()
+			delete(s.tsoClientPool.clients, forwardedHost)
+			s.tsoClientPool.Unlock()
+			continue
 		}
 		ts, err = forwardStream.Recv()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


server/grpc_service.go
```diff
@@ -2632,7 +2632,13 @@ func (s *GrpcServer) getGlobalTSOFromTSOServer(ctx context.Context) (pdpb.Timest
 		if err != nil {
 			return pdpb.Timestamp{}, err
 		}
-		forwardStream.Send(request)
+		err := forwardStream.Send(request)
+		if err != nil {
+			s.tsoClientPool.Lock()
+			delete(s.tsoClientPool.clients, forwardedHost)
```
zeminzhou, line 2638: Do we need to do close？

rleungx (approved):

server/grpc_service.go
lhy1024: I  think so. We need to avoid that there are always errors.

JmPotato (approved):

PR #7371, opened 2023-11-15
client/http: implement more HTTP APIs

JmPotato:

Issue Number: ref #7300.

```commit-message
- Implement more HTTP APIs.
- Use consts more in `Rule` structure.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -47,8 +47,13 @@ type Client interface {
 	GetRegionsByStoreID(context.Context, uint64) (*RegionsInfo, error)
 	GetHotReadRegions(context.Context) (*StoreHotPeersInfos, error)
 	GetHotWriteRegions(context.Context) (*StoreHotPeersInfos, error)
+	GetRegionStatusByKey(context.Context, []byte, []byte) (*RegionStats, error)
 	GetStores(context.Context) (*StoresInfo, error)
+	GetPlacementRulesByGroup(context.Context, string) ([]*Rule, error)
+	SetPlacementRule(context.Context, *Rule) error
+	DeletePlacementRule(context.Context, string, string) error
 	GetMinResolvedTSByStoresIDs(context.Context, []uint64) (uint64, map[uint64]uint64, error)
+	AccelerateSchedule(context.Context, []byte, []byte) error
 	Close()
 }
 
@@ -154,7 +159,7 @@ func (c *client) execDuration(name string, duration time.Duration) {
 // it consistent with the current implementation of some clients (e.g. TiDB).
 func (c *client) requestWithRetry(
 	ctx context.Context,
-	name, uri string,
+	name, uri, method string,
 	res interface{},
 ) error {
 	var (
@@ -163,7 +168,7 @@ func (c *client) requestWithRetry(
 	)
 	for idx := 0; idx < len(c.pdAddrs); idx++ {
 		addr = c.pdAddrs[idx]
-		err = c.request(ctx, name, addr, uri, res)
+		err = c.request(ctx, name, fmt.Sprintf("%s%s", addr, uri), method, res)
 		if err == nil {
 			break
 		}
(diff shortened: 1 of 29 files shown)


client/http/client.go
```diff
@@ -47,8 +47,13 @@ type Client interface {
 	GetRegionsByStoreID(context.Context, uint64) (*RegionsInfo, error)
 	GetHotReadRegions(context.Context) (*StoreHotPeersInfos, error)
 	GetHotWriteRegions(context.Context) (*StoreHotPeersInfos, error)
+	GetRegionStatusByKey(context.Context, []byte, []byte) (*RegionStats, error)
```
HuSharp, line 50: Does it mean by KeyRange?

client/http/client.go
```diff
@@ -289,23 +293,57 @@ func (c *client) GetHotReadRegions(ctx context.Context) (*StoreHotPeersInfos, er
...
+	err := c.requestWithRetry(ctx, "GetPlacementRulesByGroup", PlacementRulesByGroup(group), http.MethodGet, &rules)
+	if err != nil {
+		return nil, err
+	}
+	return rules, nil
+}
+
+// SetPlacementRule sets the placement rule.
+func (c *client) SetPlacementRule(ctx context.Context, rule *Rule) error {
+	return c.requestWithRetry(ctx, "SetPlacementRule", PlacementRule, http.MethodPost, nil)
```
HuSharp, line 339: Does this function seem to be invalid without using `rule`?

client/http/client.go
```diff
@@ -335,3 +373,8 @@ func (c *client) GetMinResolvedTSByStoresIDs(ctx context.Context, storeIDs []uin
 	}
 	return resp.MinResolvedTS, resp.StoresMinResolvedTS, nil
 }
+
+// AccelerateSchedule accelerates the scheduling of the regions within the given key range.
+func (c *client) AccelerateSchedule(ctx context.Context, startKey, endKey []byte) error {
```
HuSharp, line 378: ditto

client/http/client.go
JmPotato: Nice catch. I think I should also add some tests.

client/http/client.go
```diff
@@ -154,7 +159,7 @@ func (c *client) execDuration(name string, duration time.Duration) {
 // it consistent with the current implementation of some clients (e.g. TiDB).
 func (c *client) requestWithRetry(
```
rleungx, line 160: Is it possible to customize the retry?

JmPotato: We will provide the function of custom backoffer in the future, but at present, the interface we are about to replace in TiDB uses this simple retry method.

client/http/client.go
```diff
@@ -289,23 +311,70 @@ func (c *client) GetHotReadRegions(ctx context.Context) (*StoreHotPeersInfos, er
...
+		return nil, err
+	}
+	return rules, nil
+}
+
+// SetPlacementRule sets the placement rule.
+func (c *client) SetPlacementRule(ctx context.Context, rule *Rule) error {
+	ruleJSON, err := json.Marshal(rule)
+	if err != nil {
+		return errors.Trace(err)
```
HuSharp, line 364: not sure why needs `errors.Trace`?

JmPotato: One is to be consistent with the original HTTP call code of TiDB, and the other is that because it is a general library, adding trace information can help us better locate where the call is wrong I think.

client/http/client.go
rleungx: I still prefer to leave it to the caller even if the retry method is simple.

JmPotato: Make sense to me though I'd like to implement it in an independent PR later.

rleungx (approved):

HuSharp (approved):
LGTM!

PR #7372, opened 2023-11-15
api: support mcs api for members

HuSharp:

Issue Number: Ref #7519

```commit-message
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```


server/api/micro_service.go
```diff
@@ -0,0 +1,103 @@
...
+
+	h.rd.JSON(w, http.StatusInternalServerError, "not support service")
+}
+
+// @Tags     members
+// @Summary  Delete the member of the cluster for the specified service.
+// @Produce  json
+// @Success  200  {string}  string  "removed, pd: {service}"
+// @Router   /ms/members/{service} [delete]
+func (h *microServiceHandler) DeleteMember(w http.ResponseWriter, r *http.Request) {
```
rleungx, line 89: When do we need to use it?

HuSharp: Currently not in use, tidb-operator uses process shutdown for deletion🤔

If there is no usage scenario, I can remove this func...

HuSharp: removed.

server/api/micro_service.go
```diff
@@ -0,0 +1,82 @@
...
+		rd:  rd,
+	}
+}
+
+// @Tags     members
+// @Summary  Get all members of the cluster for the specified service.
+// @Produce  json
+// @Success  200  {object}  []string
+// @Router   /ms/members/{service} [get]
+func (h *microServiceHandler) GetMembers(w http.ResponseWriter, r *http.Request) {
```
rleungx, line 32: What if it is in PD mode?

HuSharp: will return `not support micro service`

tests/integrations/mcs/members/member_test.go
```diff
@@ -0,0 +1,111 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package members_test
+
+import (
+	"context"
+	testing2 "testing"
```
rleungx, line 19: How about just testing?

server/api/micro_service.go
```diff
@@ -0,0 +1,91 @@
...
+// @Produce  json
+// @Success  200  {object}  pdpb.Member
+// @Router   /ms/leader/{service} [get]
+func (h *microServiceHandler) GetLeader(w http.ResponseWriter, r *http.Request) {
+	if !h.svr.IsAPIServiceMode() {
+		h.rd.JSON(w, http.StatusServiceUnavailable, "not support micro service")
+		return
+	}
+	if service := mux.Vars(r)["service"]; len(service) > 0 {
+		leader, _, err := discovery.GetMCSPrimary(service, r.URL.Query().Get("keyspace_id"), h.svr.GetClient())
```
rleungx, line 77: How about making keyspace id as an optional parameter?

HuSharp: Currently, keyspace id is an optional para, which will be checked in `getMCSPrimaryPath`

HuSharp:
@rleungx @JmPotato @lhy1024 @CabinfeverB PTAL, thx! :)

server/api/micro_service.go
```diff
@@ -0,0 +1,90 @@
...
+				var entry discovery.ServiceRegistryEntry
+				if err = entry.Deserialize(keyValue.Value); err != nil {
+					log.Info("deserialize failed", zap.String("key", string(keyValue.Key)), zap.Error(err))
+				}
+				apis = append(apis, entry.ServiceAddr)
+			}
+		}
+		h.rd.JSON(w, http.StatusOK, apis)
+		return
+	}
```
CabinfeverB, line 60: How about telling the user to add service name.

server/api/micro_service.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			h.rd.JSON(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+		if resps == nil {
+			h.rd.JSON(w, http.StatusNotFound, fmt.Sprintf("no members for %s", service))
+			return
+		}
+
+		var apis []string
```
CabinfeverB, line 48:
```suggestion
		var addrs []string
```

server/api/router.go
```diff
@@ -295,6 +295,10 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	registerFunc(apiRouter, "/leader/resign", leaderHandler.ResignLeader, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
 	registerFunc(apiRouter, "/leader/transfer/{next_leader}", leaderHandler.TransferLeader, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
 
+	msHandler := newMicroServiceHandlerHandler(svr, rd)
+	registerFunc(apiRouter, "/ms/members/{service}", msHandler.GetMembers, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(apiRouter, "/ms/leader/{service}", msHandler.GetLeader, setMethods(http.MethodGet), setAuditBackend(prometheus))
```
CabinfeverB, line 300: I have a question, is it better to use primary or leader here？

CabinfeverB: I found we call it as primary in grafana. 
image
PTAL @rleungx

server/api/micro_service.go
HuSharp: done.

server/api/router.go
rleungx: prefer to use primary

HuSharp: replaced

server/api/micro_service.go
JmPotato, line 1: Suggest putting the microservice API into the v2 package.

HuSharp: moved.

pkg/mcs/discovery/discover.go
```diff
@@ -35,3 +45,77 @@ func Discover(cli *clientv3.Client, clusterID, serviceName string) ([]string, er
...
+	primaryPath, err := getMCSPrimaryPath(name, client, keyspaceGroupIDKey)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return election.GetLeader(client, primaryPath)
+}
+
+// GetMembers returns all the members of the specified service name.
+func GetMembers(name string, client *clientv3.Client) (*clientv3.TxnResponse, error) {
```
JmPotato, line 73: Can we handle the corresponding data structure internally in this function instead of directly returning an etcd TxnResponse?

HuSharp: Good catch, some stale implementation 0 0

server/api/router.go
HuSharp: It seems we don't need this function right now, so let's delete it for now

pkg/mcs/discovery/discover.go
```diff
@@ -35,3 +43,37 @@ func Discover(cli *clientv3.Client, clusterID, serviceName string) ([]string, er
...
+		if !resps.Succeeded {
+			return nil, errs.ErrEtcdTxnConflict.FastGenByArgs()
+		}
+
+		var addrs []string
+		for _, resp := range resps.Responses {
+			for _, keyValue := range resp.GetResponseRange().GetKvs() {
+				var entry ServiceRegistryEntry
+				if err = entry.Deserialize(keyValue.Value); err != nil {
+					log.Error("deserialize failed", zap.String("key", string(keyValue.Key)), zap.Error(err))
```
JmPotato, line 69:
```suggestion
					log.Error("try to deserialize service registry entry failed", zap.String("key", string(keyValue.Key)), zap.Error(err))
```

JmPotato (approved):
The rest LGTM.

CabinfeverB (approved):

PR #7397, opened 2023-11-20
client/http: encapsulate rule-related PD HTTP interfaces

JmPotato:

Issue Number: ref #7300.

```commit-message
Expand the PD HTTP interfaces to include more encapsulation, particularly focusing on the rule-related interfaces.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -42,6 +42,7 @@ const (
 
 // Client is a PD (Placement Driver) HTTP client.
 type Client interface {
+	/* Meta-related interfaces */
 	GetRegionByID(context.Context, uint64) (*RegionInfo, error)
 	GetRegionByKey(context.Context, []byte) (*RegionInfo, error)
 	GetRegions(context.Context) (*RegionsInfo, error)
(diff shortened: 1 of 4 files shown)


client/http/client.go
```diff
@@ -51,11 +52,24 @@ type Client interface {
 	GetHotWriteRegions(context.Context) (*StoreHotPeersInfos, error)
 	GetRegionStatusByKeyRange(context.Context, []byte, []byte) (*RegionStats, error)
 	GetStores(context.Context) (*StoresInfo, error)
+	/* Rule-related interfaces */
```
rleungx, line 55: We have two groups of API for the placement rule, maybe we can reduce it later.

HuSharp (approved):
LGTM!

client/http/client.go
JmPotato: Yes, after replacing the existing TiDB code, we can reduce the exposed interfaces here.

rleungx (approved):

PR #7297, opened 2023-11-01
mcs: support region http interface in scheduling server

lhy1024:
### What problem does this PR solve?

Issue Number: Ref https://github.com/tikv/pd/issues/5839

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

PR #7403, opened 2023-11-21
mcs: fix api wrongly send to scheduling server

rleungx:

Issue Number: Ref #5839.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


lhy1024 (approved):

server/api/admin.go
```diff
@@ -61,7 +61,7 @@ func (h *adminHandler) DeleteRegionCache(w http.ResponseWriter, r *http.Request)
 		return
 	}
 	rc.DropCacheRegion(regionID)
-	if h.svr.IsAPIServiceMode() {
+	if h.svr.IsServiceIndependent(utils.SchedulingServiceName) {
```
lhy1024, line 64: we also need to add judge in `pkg/utils/apiutil/serverapi/middleware.go` with ` h.s.IsServiceIndependent(rule.targetServiceName) `
image

disksing (approved):

server/api/admin.go
lhy1024: I have added it in #7297

PR #7428, opened 2023-11-23
*: Use http.NoBody instead of nil in http.NewRequest calls

lhy1024:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


HuSharp (approved):
maybe can instead of `nil` in `client/http/client.go` as well :)

rleungx (approved):

PR #7448, opened 2023-11-27
keyspace: add option to skip splitting rawkv regions

### LINKED ISSUE #7444: Add option to skip splitting rawKV regions when creating new keyspace.
## Enhancement Task

Currently, for every new keyspace created, two sets of region bounds are created to sure different keyspaces' rawKV region and txnKV does not overlap.

But in most scenarios, only txnKV is used, resulting in empty rawKV regions. There should be a option to skip creating rawKV regions when creating new keyspace.

AmoebaProtozoa:

Issue Number: Close #7444

Added an option to skip splitting rawkv regions.

```commit-message
```

### Check List

Tests

- Unit test

Code changes

### Release note

```release-note
None.
```


server/config/config.go
```diff
@@ -855,6 +856,8 @@ type KeyspaceConfig struct {
 	WaitRegionSplit bool `toml:"wait-region-split" json:"wait-region-split"`
 	// WaitRegionSplitTimeout indicates the max duration to wait region split.
 	WaitRegionSplitTimeout typeutil.Duration `toml:"wait-region-split-timeout" json:"wait-region-split-timeout"`
+	// DisableRawKVRegionSplit indicates whether to skip raw kv region split.
+	DisableRawKVRegionSplit bool `toml:"disable-raw-kv-region-split" json:"disable-raw-kv-region-split,string"`
```
lhy1024, line 860: how about using `Enable...` as the config name?

pkg/keyspace/util.go
```diff
@@ -196,7 +204,7 @@ func getRegionLabelID(id uint32) string {
 }
 
 // MakeLabelRule makes the label rule for the given keyspace id.
-func MakeLabelRule(id uint32) *labeler.LabelRule {
+func MakeLabelRule(id uint32, skipRaw bool) *labeler.LabelRule {
```
lhy1024, line 207: Do we need to export it?

server/config/config.go
AmoebaProtozoa: done, I have change it to simply `SplitRawKVRegion`

pkg/keyspace/keyspace.go
```diff
@@ -325,65 +328,62 @@ func (manager *Manager) splitKeyspaceRegion(id uint32, waitRegionSplit bool) (er
...
+			if tickerInterval != newInterval {
+				tickerInterval = newInterval
+				ticker.Reset(newInterval)
+			}
+			regionsInfo := manager.cluster.GetBasicCluster().RegionsInfo
+			totalSplit := 0 // number of split keys that have been split.
+			var region *core.RegionInfo
+			for _, splitKey := range splitKeys {
+				region = regionsInfo.GetRegionByKey(splitKey)
+				if region != nil && bytes.Equal(region.GetStartKey(), splitKey) {
```
JmPotato, line 370: Do we need to check if the region and key being compared each time are duplicated?

AmoebaProtozoa: I don't think that's necessary, the `totalSplit` reset to 0 every tick anyway, so just counting the number should work.

JmPotato: If some Regions never split for some reason, will this check last forever?

AmoebaProtozoa: there is a timer controlling the timeout, if it take too long to split, it will throw ErrRegionSplitTimeout
https://github.com/tikv/pd/blob/a7dd9fa89c8c72da1f6bbb096a356f23694de390/pkg/keyspace/keyspace.go#L378

JmPotato (approved):

PR #7453, opened 2023-11-27
client/http: add GetHistoryHotRegions interface and more tests

JmPotato:

Issue Number: ref #7300.

```commit-message
- Add `GetHistoryHotRegions` interface.
- Add more tests.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -52,6 +52,9 @@ type Client interface {
 	GetHistoryHotRegions(context.Context, *HistoryHotRegionsRequest) (*HistoryHotRegions, error)
 	GetRegionStatusByKeyRange(context.Context, *KeyRange) (*RegionStats, error)
 	GetStores(context.Context) (*StoresInfo, error)
+	/* Config-related interfaces */
+	GetScheduleConfig(context.Context) (map[string]interface{}, error)
+	SetScheduleConfig(context.Context, map[string]interface{}) error
 	/* Rule-related interfaces */
 	GetAllPlacementRuleBundles(context.Context) ([]*GroupBundle, error)
 	GetPlacementRuleBundleByGroup(context.Context, string) (*GroupBundle, error)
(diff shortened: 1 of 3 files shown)


client/http/client.go
```diff
@@ -191,20 +192,31 @@ func (c *client) execDuration(name string, duration time.Duration) {
 	c.executionDuration.WithLabelValues(name).Observe(duration.Seconds())
 }
 
+// HeaderOption configures the HTTP header.
+type HeaderOption func(header http.Header)
+
+// WithAllowFollowerHandle sets the header field to allow a PD follower to handle this request.
+func WithAllowFollowerHandle(allow bool) HeaderOption {
```
rleungx, line 199: How about:
```suggestion
func WithAllowFollowerHandle() HeaderOption {
```

client/http/client.go
```diff
@@ -191,20 +192,31 @@ func (c *client) execDuration(name string, duration time.Duration) {
...
+	return func(header http.Header) {
+		header.Set("PD-Allow-follower-handle", fmt.Sprintf("%t", allow))
```
CabinfeverB, line 201: It's better to format it as `PD-Allow-Follower-Handle`. Also, `CanonicalMIMEHeaderKey` has helped us do this, but I think it's clear if we do this.

client/http/client.go
```diff
@@ -191,20 +195,31 @@ func (c *client) execDuration(name string, duration time.Duration) {
 	c.executionDuration.WithLabelValues(name).Observe(duration.Seconds())
 }
 
+// HeaderOption configures the HTTP header.
+type HeaderOption func(header http.Header)
+
+// WithAllowFollowerHandle sets the header field to allow a PD follower to handle this request.
+func WithAllowFollowerHandle() HeaderOption {
+	return func(header http.Header) {
+		header.Set("PD-Allow-Follower-Handle", "true")
```
HuSharp, line 204: Do we need to use const for `PD-Allow-Follower-Handle`?

JmPotato: Since there is only one constant right now, I think we can change it later when we introduce more.

rleungx (approved):

CabinfeverB (approved):

PR #7462, opened 2023-11-28
client/http: implement the marshaler interfaces for Rule/RuleOp

JmPotato:

Issue Number: ref #7300.

```commit-message
Implement the marshaler interfaces for `Rule` and `RuleOP` to make sure we could set/get the correct start/end key.

Ref https://github.com/pingcap/tidb/blob/46d4231c8b0ade353b98572e7c2a015bddf940f4/pkg/ddl/placement/rule.go#L76-L91.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/http/client.go
```diff
@@ -356,6 +357,18 @@ func (c *client) GetRegionsByStoreID(ctx context.Context, storeID uint64) (*Regi
 	return &regions, nil
 }
 
+// GetRegionsReplicatedStateByKeyRange gets the regions replicated state info by key range.
+func (c *client) GetRegionsReplicatedStateByKeyRange(ctx context.Context, keyRange *KeyRange) (string, error) {
```
HuSharp, line 361: Does this function annotation need to emphasize keys format?

client/http/client.go
```diff
@@ -399,10 +412,10 @@ func (c *client) GetHistoryHotRegions(ctx context.Context, req *HistoryHotRegion
 
 // GetRegionStatusByKeyRange gets the region status by key range.
 // The keys in the key range should be encoded in the UTF-8 bytes format.
-func (c *client) GetRegionStatusByKeyRange(ctx context.Context, keyRange *KeyRange) (*RegionStats, error) {
+func (c *client) GetRegionStatusByKeyRange(ctx context.Context, keyRange *KeyRange, onlyCount bool) (*RegionStats, error) {
```
HuSharp, line 415: I noticed some missing introductions in `func (h *statsHandler) GetRegionStatus(w http.ResponseWriter, r *http.Request) {`, such as explaining the role of `count`, can you update the comment?

client/http/codec.go
```diff
@@ -0,0 +1,121 @@
...
+				}
+			}
+			break
+		}
+	}
+	return buf, nil
+}
+
+// keyToKeyHexStr converts a raw key to a hex string after encoding.
+func rawKeyToKeyHexStr(key []byte) string {
```
HuSharp, line 104:
```suggestion
// rawKeyToKeyHexStr converts a raw key to a hex string after encoding.
func rawKeyToKeyHexStr(key []byte) string {
```

HuSharp (approved):
LGTM!

Issue #7468, opened 2023-11-29
TestScheduleTestSuite/TestDisable is unstable


lhy1024:
### Which jobs are failing
```
2023-11-29T01:42:11.6662377Z --- FAIL: TestScheduleTestSuite (103.50s)
2023-11-29T01:42:11.6662663Z     --- FAIL: TestScheduleTestSuite/TestDisable (26.11s)
2023-11-29T01:42:11.6662936Z         testutil.go:265: start to run test in pd mode
2023-11-29T01:42:11.6663205Z         testutil.go:269: finish to run test in pd mode
2023-11-29T01:42:11.6663459Z         testutil.go:274: start to run test in api mode
2023-11-29T01:42:11.6663618Z         testutil.go:64: 
2023-11-29T01:42:11.6664122Z             	Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:64
2023-11-29T01:42:11.6664822Z             	            				/home/runner/work/pd/pd/tests/server/api/scheduler_test.go:686
2023-11-29T01:42:11.6665527Z             	            				/home/runner/work/pd/pd/tests/server/api/scheduler_test.go:604
2023-11-29T01:42:11.6666067Z             	            				/home/runner/work/pd/pd/tests/testutil.go:279
2023-11-29T01:42:11.6666613Z             	            				/home/runner/work/pd/pd/tests/testutil.go:260
2023-11-29T01:42:11.6667373Z             	            				/home/runner/work/pd/pd/tests/server/api/scheduler_test.go:568
2023-11-29T01:42:11.6667633Z             	Error:      	Condition never satisfied
2023-11-29T01:42:11.6667935Z             	Test:       	TestScheduleTestSuite/TestDisable
2023-11-29T01:42:11.6668017Z FAIL
```

### CI link
https://github.com/tikv/pd/actions/runs/7026828089/job/19120231295?pr=7465

lhy1024:
It failed in https://github.com/tikv/pd/blob/4356aebb55f453cf91116a22461f4ab375c9ce11/tests/server/api/scheduler_test.go#L604 https://github.com/tikv/pd/blob/4356aebb55f453cf91116a22461f4ab375c9ce11/tests/server/api/scheduler_test.go#L684-L690

This is api https://github.com/tikv/pd/blob/4356aebb55f453cf91116a22461f4ab375c9ce11/pkg/schedule/handler/handler.go#L862

We adjust it according to the scheduling config
[...]

PR #7470, opened 2023-11-29
tests: make TestScheduleTestSuite stable

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/7468

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/api/scheduler_test.go
+++ b/tests/server/api/scheduler_test.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/metapb"
-	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	sc "github.com/tikv/pd/pkg/schedule/config"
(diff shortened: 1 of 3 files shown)


tests/server/api/scheduler_test.go
```diff
@@ -43,6 +42,14 @@ func TestScheduleTestSuite(t *testing.T) {
 	suite.Run(t, new(scheduleTestSuite))
 }
 
+func (suite *scheduleTestSuite) SetupSuite() {
+	suite.NoError(failpoint.Enable("github.com/tikv/pd/pkg/dashboard/adapter/skipDashboardLoop", `return(true)`))
```
rleungx, line 46: Why does it fix this problem?

lhy1024: https://github.com/tikv/pd/issues/7468#issuecomment-1831199529

Yes, dashboard update pd config, which overwrite the config.

rleungx: But why updating the pd server config can overwrite the scheduler config?

Issue #7431, opened 2023-11-24
Activate Follower-Handle


CabinfeverB:
## Development Task

PD serves as the central node and provides services for the cluster. In the past, PD leaders have been carrying all the services and follower resources have not been utilized.

- [x] Region API
### Summary
An important function of PD is to provide routing information for data replicas. For example, GetRegion is a hot path in TiDB cluster. The pd provides the sync region mechanism. The pd leader synchronizes the region with the changed meta information to a follower.  So we can handle the region requeset in PD follower server.

#### Sync region
The pd provides the sync region mechanism. The pd leader synchronizes the region with the changed meta information to a follower. 
And the followers hold basicCluster which has region tree.
#### Follower forward mechanism
When enable-forwarding is enabled (the default value in TiDB is false), the pd periodically checks the connectivity and health status of the pd leader. If false, pd client will try to send request with pd-forwarded-hostto followers and followers will forward the request to the leader.
#### Region Cache in TiKV client-go
The region cache deletes a region item in either of the following situations:
1. GC triggered by TTL
2. When reloading region, insert new region item and delete old region
3. client-go will do ScanRegions for all regions caused by GC. It will ignore the region cache and request to PD.
Therefore, if a region is unavailable in client-Go, the client does not delete the old region first and then find a new region

### design
There may be a delay in synchronizing region data to followers. As a result, the followers data is not up-to-date. To reduce potential retries on the associated path, callers need to select followers or leaders according to the usage scenario. 
Specifically, if the caller wants to obtain a new region first time, it can do so from the followers (we can assume that most of the region information in the followers is up to date). If the caller needs to update the region information, we think that this time must get the latest, then this time from the leader.
According to above additional knowledge, this transformation is very easy to do in client-go.

##### Client-go Impl
1. In findRegionByKey, if region is nil, the region cache can get region with follower enabled optin. Otherwise, no option.
- [x] [client-go/pull/1079](https://github.com/tikv/client-go/pull/1079)
2. We should check epoch in region cache.
- [x] [client-go/pull/1072](https://github.com/tikv/client-go/pull/1072)

##### PD Client Implement
1. Only do BuildForwardContext, if we need to forward request to the leader.  https://github.com/tikv/pd/pull/7471
2. For the GetRegion interface, add FollowerHandleOption in parameters.  https://github.com/tikv/pd/pull/7465
[...]

PR #7489, opened 2023-12-04
client: Introduce `ServiceClient`

CabinfeverB:

Issue Number: ref #7431 ref #7576

This PR introduce the interface `ServiceClient` which can get the gRPC client of a specific gRPC server in a PD cluster. It can send request to the follower including follower-handle or forward.
And also a balancer is introduced.
Some tools have been added to client to make it possible to run integration tests in the client package. The reason for adding it to the client package is that ServiceClient should be invisible to the caller, so it cannot be tested well in `tests/integrations`.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/grpcutil/grpcutil.go
```diff
@@ -75,6 +77,39 @@ func BuildForwardContext(ctx context.Context, addr string) context.Context {
...
+// BuildFollowerHandleContext creates a context with follower handle metadata information.
+// It is used in client side.
+func BuildFollowerHandleContext(ctx context.Context) context.Context {
+	md := metadata.Pairs(FollowerHandleMetadataKey, "")
+	return metadata.NewOutgoingContext(ctx, md)
+}
+
+// GetFollowerHandleEnable returns the forwarded host in metadata.
+// Only used for test.
+func GetFollowerHandleEnable(ctx context.Context, f func(context.Context) (metadata.MD, bool)) bool {
```
rleungx, line 96:
```suggestion
func IsFollowerHandleEnabled(ctx context.Context, f func(context.Context) (metadata.MD, bool)) bool {
```

client/pd_service_discovery.go
```diff
@@ -97,6 +100,237 @@ type ServiceDiscovery interface {
...
+}
+
+// BuildGRPCContext implements ServiceClient.
+func (c *pdServiceClient) BuildGRPCContext(ctx context.Context, toLeader bool) context.Context {
+	if c == nil {
+		return ctx
+	}
+	if c.IsLeader() {
+		return ctx
+	}
```
rleungx, line 158: merge with line 153?

client/testutil/check_env_linux.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 TiKV Project Authors.
```
rleungx, line 1:
```suggestion
// Copyright 2023 TiKV Project Authors.
```

client/pd_service_discovery.go
```diff
@@ -97,6 +100,237 @@ type ServiceDiscovery interface {
...
+}
+
+func (c *pdServiceBalancer) next() {
+	c.now = c.now.next
+}
+
+func (c *pdServiceBalancer) get() (ret ServiceClient) {
+	c.mu.Lock()
+	i := 0
+	defer c.mu.Unlock()
```
rleungx, line 319: exchange these two?

client/pd_service_discovery.go
```diff
@@ -97,6 +100,237 @@ type ServiceDiscovery interface {
...
+	if c.now == nil {
+		return nil
+	}
+	for ; i < c.totalNode; i++ {
+		if c.now.Available() {
```
rleungx, line 324: If it's not available, will it be excluded in the next round?

client/pd_service_discovery.go
```diff
@@ -97,6 +100,237 @@ type ServiceDiscovery interface {
...
+	// GetAddress returns the address information of the PD server.
+	GetAddress() string
+	// GetClientConn returns the gRPC connection of the service client
+	GetClientConn() *grpc.ClientConn
+	// BuildGRPCContext builds a context object with a gRPC context.
+	// ctx: the original context object.
+	// mustLeader: whether must send to leader.
+	BuildGRPCContext(ctx context.Context, mustLeader bool) context.Context
+	// IsLeader returns whether the target PD server is leader.
+	IsLeader() bool
```
rleungx, line 114: What if the leader changes?

CabinfeverB: If leader changes, it will create new ServiceClient and store it in `leader`

client/pd_service_discovery.go
CabinfeverB: No. I don't remove unavailable nodes from the queue. The simple approach is taken now and can be optimized in the future if needed

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+// GetAddress implements ServiceClient.
+func (c *pdServiceClient) GetAddress() string {
+	if c == nil {
+		return ""
+	}
+	return c.addr
+}
+
+// BuildGRPCContext implements ServiceClient.
+func (c *pdServiceClient) BuildGRPCContext(ctx context.Context, toLeader bool) context.Context {
```
rleungx, line 152: What is the difference between toLeader and isLeader?

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+// pdServiceBalancerNode is a balancer node for PD service.
+// It extends the pdServiceClient and adds additional fields for the next polling client in the chain.
+type pdServiceBalancerNode struct {
+	ServiceClient
+	next *pdServiceBalancerNode
+}
+
+// pdServiceBalancer is a load balancer for PD service clients.
+// It is used to balance the request to all servers and manage the connections to multiple PD service nodes.
+type pdServiceBalancer struct {
```
rleungx, line 280: How about supporting multiple policies? And make the round-robin a common library?

CabinfeverB:
@JmPotato PTAL~

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+// ServiceClient is an interface that defines a set of operations for a raw PD gRPC client to specific PD server.
+type ServiceClient interface {
+	// GetAddress returns the address information of the PD server.
+	GetAddress() string
+	// GetClientConn returns the gRPC connection of the service client
+	GetClientConn() *grpc.ClientConn
+	// BuildGRPCContext builds a context object with a gRPC context.
+	// ctx: the original context object.
+	// mustLeader: whether must send to leader.
+	BuildGRPCContext(ctx context.Context, mustLeader bool) context.Context
```
JmPotato, line 112: According to the name, it's hard to tell what this build is for. Suggest using a more specified name like `BuildGRPCTargetContext`.

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+	IsLeader() bool
+	// Available returns if the network or other availability for the current service client is available.
+	Available() bool
+	// NeedRetry checks if client need to retry based on the PD server error response.
+	// And It will mark the client as unavailable if the pd error shows the follower can't handle request.
+	NeedRetry(*pdpb.Error, error) bool
+}
+
+var _ ServiceClient = (*pdServiceClient)(nil)
+var _ ServiceClient = (*pdServiceAPIClient)(nil)
```
JmPotato, line 123:
```suggestion
var (
	_ ServiceClient = (*pdServiceClient)(nil)
	_ ServiceClient = (*pdServiceAPIClient)(nil)
)
```

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+type pdServiceClient struct {
+	addr       string
+	conn       *grpc.ClientConn
+	isLeader   bool
+	leaderAddr string
+
+	networkFailure atomic.Bool
+}
+
+func newPDServiceClient(addr, leaderAddr string, conn *grpc.ClientConn, isLeader bool) *pdServiceClient {
```
JmPotato, line 134: Maybe using the interface as the returned value is better.

```suggestion
func newPDServiceClient(addr, leaderAddr string, conn *grpc.ClientConn, isLeader bool) ServiceClient {
```

client/pd_service_discovery.go
```diff
@@ -97,6 +100,234 @@ type ServiceDiscovery interface {
...
+// IsLeader implements ServiceClient.
+func (c *pdServiceClient) IsLeader() bool {
+	if c == nil {
+		return false
+	}
+	return c.isLeader
+}
+
+// NetworkAvailable implements ServiceClient.
+func (c *pdServiceClient) Available() bool {
```
JmPotato, line 171: Since the judgment condition is whether there is a network error, would `Reachable` be a better name?

client/pd_service_discovery.go
JmPotato: I am a bit confused here because it seems that there is forwarding logic in the internal `ServiceClient`, and now we have introduced an additional layer of the load balancer. Will these two things conflict with each other?

client/pd_service_discovery.go
CabinfeverB: The function of the `ServiceClient` is to interact with a PD server, which may be a follower or leader. When a client sends a request to a server, `mutLeader` is true if the request is forwarded to the leader. Otherwise, it is sent to the server to process the request, and `mutLeader` is false.

client/pd_service_discovery.go
CabinfeverB: > I am a bit confused here because it seems that there is forwarding logic in the internal `ServiceClient`, and now we have introduced an additional layer of the load balancer. Will these two things conflict with each other?

There will be no conflict. In the past, followers were randomly selected for forwarding, but now it is in turn

CabinfeverB: > How about supporting multiple policies? And make the round-robin a common library?

I've thought about it. Since it is observed that there is now only one Round-Robin balancer in `/pkg/balancer`, I haven't implemented multiple policies yet

client/pd_service_discovery.go
CabinfeverB: `pdServiceAPIClient` checks the network and availability

client/go.mod
```diff
@@ -19,6 +20,8 @@ require (
 	google.golang.org/grpc v1.54.0
 )
 
+require google.golang.org/genproto/googleapis/rpc v0.0.0-20231106174013-bbf56f31fb17 // indirect
```
rleungx, line 23: need format

client/pd_service_discovery.go
rleungx: How about targetIsLeader or something else?

rleungx (approved):
The rest LGTM.

client/testutil/check_env.go
JmPotato, line 1: Maybe the name `check_env_dummy.go` is better.

CabinfeverB: Also renamed the file in /pkg/utils/tempurl. PTAL

JmPotato (approved):

PR #7490, opened 2023-12-04
client/http: introduce caller ID into the HTTP client

JmPotato:

Issue Number: ref #7300.

```commit-message
Introduce caller ID into the HTTP client.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -32,6 +32,7 @@ import (
 )
 
 const (
+	defaultCallerID    = "pd-http-client"
 	httpScheme         = "http"
 	httpsScheme        = "https"
 	networkErrorStatus = "network error"
@@ -145,7 +146,7 @@ func NewClient(
 	pdAddrs []string,
 	opts ...ClientOption,
 ) Client {
-	c := &client{inner: &clientInner{}}
+	c := &client{inner: &clientInner{}, callerID: defaultCallerID}
 	// Apply the options first.
 	for _, opt := range opts {
 		opt(c)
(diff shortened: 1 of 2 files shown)


CabinfeverB (approved):

client/http/client.go
```diff
@@ -196,13 +218,19 @@ func (c *client) execDuration(name string, duration time.Duration) {
 	c.executionDuration.WithLabelValues(name).Observe(duration.Seconds())
 }
 
+// Header key definition constants.
+const (
+	pdAllowFollowerHandleKey = "PD-Allow-Follower-Handle"
+	componentSignatureKey    = "component"
```
niubell, line 224: We can have a more intuitive name like 'CallerID', and it's a filed in http header, "X-Caller-ID" maybe a better name according HTTP RFC docs @JmPotato

client/http/client.go
```diff
@@ -239,6 +267,8 @@ func (c *client) request(
 	logFields := []zap.Field{
 		zap.String("name", name),
 		zap.String("url", url),
+		zap.String("method", method),
+		zap.String("caller-id", c.callerID),
```
niubell, line 271: Do we log the caller id on the PD server side?

client/http/client.go
JmPotato: Using `component` here is to reuse https://github.com/tikv/pd/blob/master/pkg/utils/requestutil/request_info.go#L52, but you're right about the naming, we could create a new filed to distinguish with it.

client/http/client.go
JmPotato: It will appear in the audit log also.

PR #7508, opened 2023-12-07
client/http: support tidb-operator SDK

HuSharp:

Issue Number: Ref #7300

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```


client/http/client.go
```diff
@@ -493,6 +679,56 @@ func (c *client) GetStores(ctx context.Context) (*StoresInfo, error) {
...
+	err = c.requestWithRetry(ctx,
+		"SetStoreLabels", StoreLabelByID(storeID),
+		http.MethodPost, bytes.NewBuffer(data), nil)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (c *client) DeleteStore(ctx context.Context, storeID uint64) error {
```
nolouch, line 719: It's better to group the Interface by severity. some interfaces are only used in operators and it's critical. Once user use this function, should let the user know it's critical when using that. like the group the op interface to the `MaintianceOpClient` etc.

PR #7554, opened 2023-12-15
client/http, tests: implement retry mechanism based on the leader and follower

JmPotato:

Issue Number: ref #7300.

```commit-message
- Implement retry mechanism based on the leader and follower.
- Move method definitions into a separate file. 
- Use a sturct `requestInfo` to gather the parameters.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/http/client.go
```diff
@@ -158,40 +154,40 @@ func (ci *clientInner) requestWithRetry(
 
 func (ci *clientInner) doRequest(
 	ctx context.Context,
-	callerID, name, url, method string,
-	body []byte, res interface{},
-	respHandler respHandleFunc, headerOpts ...HeaderOption,
+	addr string, reqInfo *requestInfo,
+	headerOpts ...HeaderOption,
 ) error {
+	url := reqInfo.getURL(addr)
```
rleungx, line 160: How about using variables for other reqInfo fields?

rleungx (approved):
The rest LGTM

client/http/client.go
```diff
@@ -339,508 +230,178 @@ func (c *client) request(
...
+	if len(newPDAddrs) == 0 {
+		log.Error("[pd] http client get empty member addresses")
+		return
 	}
-	return c.requestWithRetry(ctx,
-		"SetPlacementRule", PlacementRule,
-		http.MethodPost, ruleJSON, nil)
+	ci.setPDAddrs(newPDAddrs)
+	ci.setLeaderAddrIdx(newLeaderAddrIdx)
+	log.Info("[pd] http client update members info",
```
HuSharp, line 282: Should this log level be info? will print every minute

JmPotato: I added a condition to ensure that it only prints when members change. PTAL.

JmPotato: image

client/http/client.go
```diff
@@ -23,265 +23,172 @@ import (
...
-	httpScheme         = "http"
-	httpsScheme        = "https"
-	networkErrorStatus = "network error"
-
-	defaultTimeout = 30 * time.Second
+	// defaultCallerID marks the default caller ID of the PD HTTP client.
+	defaultCallerID = "pd-http-client"
+	// defaultInnerCallerID marks the default caller ID of the inner PD HTTP client.
+	// It's used to distinguish the requests sent by the inner client via some internal logic.
+	defaultInnerCallerID = "pd-http-client-inner"
```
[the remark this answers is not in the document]
JmPotato: According to the current implementation, in a component (such as TiDB), although we may duplicate multiple `Client` structures with different caller IDs through `WithCallerID`, they will always share the same underlying inner client. Therefore, if we want to further annotate different inner client sources, we may need to abandon this design of sharing the underlying inner client.

client/http/client.go
```diff
@@ -341,520 +248,160 @@ func (c *client) request(
...
-
-// SetPlacementRuleBundles sets the placement rule bundles.
-// If `partial` is false, all old configurations will be over-written and dropped.
-func (c *client) SetPlacementRuleBundles(ctx context.Context, bundles []*GroupBundle, partial bool) error {
-	bundlesJSON, err := json.Marshal(bundles)
-	if err != nil {
-		return errors.Trace(err)
+	oldPDAddrs, oldLeaderAddrIdx := ci.getPDAddrs()
+	ci.setPDAddrs(newPDAddrs, newLeaderAddrIdx)
+	if len(oldPDAddrs) != len(newPDAddrs) || oldLeaderAddrIdx != newLeaderAddrIdx {
```
CabinfeverB, line 300: I suggest that we can directly compare the address of the leader and output the address of the leader directly in the log

client/http/client.go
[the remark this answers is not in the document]
CabinfeverB: I got it. Moreover, maybe we can add a field to show the creator of inner client.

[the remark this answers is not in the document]
JmPotato: Good idea. I can do it in an independent PR later.

CabinfeverB (commented):
rest LGTM

client/http/client.go
JmPotato: Done. PTAL.

CabinfeverB (approved):

PR #7577, opened 2023-12-19
client: rename follower handle var

CabinfeverB:

Issue Number: ref #7431

We need to clearly define the range of APIs that a follower handle can handle

```commit-message
```

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/client.go
```diff
@@ -707,12 +707,12 @@ func (c *client) UpdateOption(option DynamicOption, value interface{}) error {
 			return errors.New("[pd] invalid value type for EnableTSOFollowerProxy option, it should be bool")
 		}
 		c.option.setEnableTSOFollowerProxy(enable)
-	case EnableFollowerHandle:
+	case EnableFollowerHandleRegionRequest:
```
rleungx, line 710: If we have more and more requests, will the option become bloated in the future?

CabinfeverB: It'll be more, but I don't think it'll be bloated. If there are more options, we can provide a function to turn all options on or off, but we must be able to control them separately

lhy1024 (approved):

CabinfeverB:
After discussion, because there is only region api at present, and the variables of pd client are generally not directly exposed to users, we can not change it for the time being

Issue #7298, opened 2023-11-02
Enhance the integration test of CI


JmPotato:
Although we already had some code simulation integration tests inside the PD repositories, there is still a lack of further integration tests, such as black box testing with other components. Having more varieties of integration test cases integrated into the CI would help us identify more breaking changes and bugs in advance.

PR #7578, opened 2023-12-19
tests/realcluster: support local real cluster test

HuSharp:

Issue Number: ref #7298

```commit-message
tests/realcluster: support local real cluster test
```
Currently, we download the binaries via the ci repo [PingCAP-QE/ci@main/pipelines/tikv/pd/latest/pull_integration_realcluster_test.groovy#L57-](https://github.com/PingCAP-QE/ci/blob/main/pipelines/tikv/pd/latest/pull_integration_realcluster_test.groovy?rgh-link-date=2023-12-21T02%3A18%3A21Z#L57-) L65 in pd/bin to download the binary.

This approach reduces the cost of learning for developers. However, these packages can only be downloaded by the ci, which means that it is not possible to run locally.

To solve this problem, I would like to load the packages via tiup instead of downloading them by self.

- Because for example, tiflash is a bit tricky to download packages
[PingCAP-QE/ci@main/scripts/artifacts/download_pingcap_artifact.sh#L101](https://github.com/PingCAP-QE/ci/blob/main/scripts/artifacts/download_pingcap_artifact.sh?rgh-link-date=2023-12-21T02%3A18%3A21Z#L101)

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```


tests/integrations/realtiup/deploy.sh
```diff
@@ -15,9 +15,14 @@ curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh
 $TIUP_BIN_DIR update playground
 
 cd ../../..
-# Run TiUP
-$TIUP_BIN_DIR playground nightly --kv 3 --tiflash 1 --db 1 --pd 3 --without-monitor \
-	--pd.binpath ./bin/pd-server --kv.binpath ./bin/tikv-server --db.binpath ./bin/tidb-server --tiflash.binpath ./bin/tiflash --tag pd_test \
-	> $CUR_PATH/playground.log 2>&1 &
+if [ ! -d "bin" ] || [ ! -e "bin/tikv-server" ]; then
```
JmPotato, line 18: Shouldn't we check all needed binaries rather than just the TiKV?

HuSharp: okay, I can add them

HuSharp: image
added

tests/integrations/realcluster/deploy.sh
```diff
@@ -15,9 +15,15 @@ curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh
...
-	--pd.binpath ./bin/pd-server --kv.binpath ./bin/tikv-server --db.binpath ./bin/tidb-server --tiflash.binpath ./bin/tiflash --tag pd_test \
-	> $CUR_PATH/playground.log 2>&1 &
+if [ ! -d "bin" ] || [ ! -e "bin/tikv-server" ] && [ ! -e "bin/tidb-server" ] && [ ! -e "bin/pd-server" ] && [ ! -e "bin/tiflash" ]; then
+	color-green "check binaries..."
+	$TIUP_BIN_DIR playground nightly --kv 3 --tiflash 1 --db 1 --pd 3 --without-monitor --tag pd_test \
+		> $CUR_PATH/playground.log 2>&1 &
+else
+  color-green "using existing binaries..."
+	$TIUP_BIN_DIR playground nightly --kv 3 --tiflash 1 --db 1 --pd 3 --without-monitor \
+		--pd.binpath ./bin/pd-server --kv.binpath ./bin/tikv-server --db.binpath ./bin/tidb-server --tiflash.binpath ./bin/tiflash --tag pd_test \
```
CabinfeverB, line 25: Does the user need to move bin to the current bin path?

HuSharp: Currently, we download the binaries via the `ci` repo https://github.com/PingCAP-QE/ci/blob/main/pipelines/tikv/pd/latest/pull_integration_realcluster_test.groovy#L57- L65 in `pd/bin` to download the binary.

This approach reduces the cost of learning for developers. However, these packages can only be downloaded by the ci, which means that it is not possible to run locally.

To solve this problem, I would like to load the packages via tiup instead of downloading them by self.

- Because for example, tiflash is a bit tricky to download packages
https://github.com/PingCAP-QE/ci/blob/main/scripts/artifacts/download_pingcap_artifact.sh#L101
- maybe I can give more information in annotation, or do you have any advice? :)

CabinfeverB: I see. Maybe we provider some flags to specify bin paths for other components?

HuSharp: I'm not sure we need to support this feature
I think both methods start the test in essentially the same way?
- cp binary file to directory
- specific binpath flags

cc @JmPotato

CabinfeverB: Yes, It's both OK to me.

JmPotato: I think we could keep the current implementation since it's most used in the CI environment.

JmPotato (approved):

CabinfeverB (approved):

PR #7589, opened 2023-12-20
tests: fix all errors detected by testifylint and enable it

JmPotato:

Issue Number: ref #4813.

```commit-message
Fix all errors detected by testifylint and enable it.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/http/types_test.go
```diff
@@ -45,7 +45,7 @@ func TestMergeRegionsInfo(t *testing.T) {
 	}
 	regionsInfo := regionsInfo1.Merge(regionsInfo2)
 	re.Equal(int64(2), regionsInfo.Count)
-	re.Equal(2, len(regionsInfo.Regions))
+	re.Len(regionsInfo.Regions, 2)
```
HuSharp, line 48: seems remains `re.Equal`

pkg/schedule/checker/merge_checker_test.go
```diff
@@ -441,6 +443,7 @@ func (suite *mergeCheckerTestSuite) TestMatchPeers() {
 }
 
 func (suite *mergeCheckerTestSuite) TestStoreLimitWithMerge() {
+	re := suite.Require()
```
lhy1024, line 446: why replace suite with re?

JmPotato: Using `Require` will call `FailNow()` to fatal the test as soon as possible rather than calling `Fail()`.

nolouch (approved):

HuSharp (approved):

PR #7595, opened 2023-12-20
mcs: add a switch to dynamically enable scheduling service

rleungx:

Issue Number: Ref #5839.

This PR adds a switch to dynamically enable scheduling service. If we only upgrade the scheduling service, the logic could be different from the logic inside the api service. To avoid the changes of different logic bringing some troubles, we use a config to let the users control it by themselves.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


CabinfeverB (approved):
LGTM. But I think there's a little bit of learning cost to using this configuration

CabinfeverB:
I test it by pd-ctl 

Before
image

> config set enable-dynamic-switch "false"

After 
image

nolouch (approved):

rleungx:
change to `enable-scheduling-fallback`

CabinfeverB (approved):

PR #7600, opened 2023-12-20
rule: split txn to multi batch

### LINKED ISSUE #7599: rule: cannot load with large txn
### What did you do?

test with a lot of rules.

can load

### What did you see instead?

cannot load with large txn

lhy1024:

Issue Number: Close #7599

```commit-message
```
1. cancel txn in loading
2. use batch to avoid that txn is too large

### Check List

Tests

- Unit test
- Manual test
image

### Release note

```release-note
None.
```


pkg/schedule/placement/rule_manager.go
```diff
@@ -492,30 +493,56 @@ func (m *RuleManager) TryCommitPatch(patch *RuleConfigPatch) error {
 }
 
 func (m *RuleManager) savePatch(p *ruleConfig) error {
```
lhy1024, line 495: it could meet "duplicate key given in txn request", where etcd not allow this case.

rleungx (approved):

CabinfeverB (approved):

PR #7605, opened 2023-12-21
ctl: replace cluster info with HTTP SDK

HuSharp:

Issue Number: Ref #7300, wait for https://github.com/tikv/pd/pull/7635 merged

```commit-message
```
- want to replace all pd-ctl
- support cluster command firstly

```bash
$ ./pd-ctl cluster
{
  "id": 7316377368062736100,
  "max_peer_count": 3
}

$ ./pd-ctl --pd http://127.0.0.1:52719 cluster
{
  "id": 7316426302202361416,
  "max_peer_count": 3
}
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/http/types.go
```diff
@@ -24,6 +24,14 @@ import (
 	"github.com/pingcap/kvproto/pkg/pdpb"
 )
 
+// ClusterStatus saves some cluster state information.
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
```
JmPotato, line 28:
```suggestion
```

tools/pd-ctl/pdctl/command/cluster_command.go
```diff
@@ -47,19 +43,21 @@ func NewClusterStatusCommand() *cobra.Command {
 }
 
 func showClusterCommandFunc(cmd *cobra.Command, args []string) {
-	r, err := doRequest(cmd, clusterPrefix, http.MethodGet, http.Header{})
+	cli := pd.NewClient(pdControlCallerID, getEndpoints(cmd))
```
JmPotato, line 46: Since the internal of pd-ctl will definitely share the same PD HTTP Client instance in the end, can we initialize and pass it to each subcommand from a more central position? For example, as a singleton global variable.

go.mod
```diff
@@ -207,6 +208,8 @@ require (
 
 replace google.golang.org/grpc v1.54.0 => google.golang.org/grpc v1.26.0
 
+replace github.com/tikv/pd/client => ./client
```
rleungx, line 211: Will it have a cycle dependency problem?

tools/pd-ctl/pdctl/command/cluster_command.go
HuSharp: pd-ctl supports passing the parameter pdAddrs, which means it's a way to set `NewClient` dynamically.

It might be possible to support global functions reading addrs instead of global variables.

go.mod
HuSharp: Do you think it's a better way to support a seperate gomod for pd-ctl?

CabinfeverB: I think it seems feasible, but I'm not sure if it's a best practice

HuSharp:
@rleungx @CabinfeverB friendly ping :)

go.mod
HuSharp: I am not sure whether have better way as well...

CabinfeverB (approved):

go.mod
rleungx: How about using a separate go mod for pd-ctl like other tools?

HuSharp: supported by https://github.com/tikv/pd/pull/7635

HuSharp:
FYI, after https://github.com/tikv/pd/pull/7635 merged, we do not need to update the server's gomod to import client

@JmPotato @CabinfeverB @rleungx PTAL, thx!

client/http/api.go
```diff
@@ -73,6 +73,8 @@ const (
 	PProfGoroutine = "/pd/api/v1/debug/pprof/goroutine"
 	// Others
 	MinResolvedTSPrefix = "/pd/api/v1/min-resolved-ts"
+	ClusterPrefix       = "/pd/api/v1/cluster"
+	ClusterStatusPrefix = "/pd/api/v1/cluster/status"
```
JmPotato, line 77: `XXXPrefix` here means there will be more content attached after this URI, it seems these two URIs are completed, so `Cluster` and `ClusterStatus` would be more suitable.

tools/pd-ctl/pdctl/command/global.go
```diff
@@ -34,12 +35,25 @@ const (
 	pingPrefix        = "pd/api/v1/ping"
 )
 
+// PDCli is a pd HTTP client
+var PDCli pd.Client
```
JmPotato, line 39: What about using a `sync.Once` here?

tools/pd-ctl/pdctl/ctl.go
```diff
@@ -86,11 +92,14 @@ func GetRootCmd() *cobra.Command {
 				return err
 			}
 
-			if err := command.InitHTTPSClient(CAPath, certPath, keyPath); err != nil {
+			if err := command.InitHTTPSClient(addrs, CAPath, certPath, keyPath); err != nil {
 				rootCmd.Println(err)
 				return err
 			}
+		} else {
+			command.SetNewPDClient(strings.Split(addrs, ","))
```
JmPotato, line 100: What about setting the PD HTTP client unconditionally before we finish the replacement?

client/http/api.go
HuSharp: get it

tools/pd-ctl/pdctl/command/global.go
HuSharp: because we support `-i` which can reset client addrs, maybe `sync.Once` is not suitable

tools/pd-ctl/pdctl/ctl.go
HuSharp: we can simplify it after replacing dialClient

JmPotato (approved):

PR #7610, opened 2023-12-22
chore: supply a script to update tidb-dashboard

baurine:

Issue Number: Close https://github.com/tikv/pd/issues/7612

Because the tidb-dashboard repo may abandon to use `release-version` file to express its version, so we need to adjust the scripts to get tidb-dashbboard version.

This PR supplies a new script `scripts/update-dashobard`, when we need to update the tidb-dashobard in PD side, we just need to run the following command:

```
$ ./scripts/update-dashboard <new_dashboard_version>
```

Then commit and push the changes.

That ok, it simplifies the multiple steps before.

This script will persist a dashboard version in a file named `scripts/dashboard-version`, so now we can explicitly know which dashboard version this PD embeds, while we don't know before.

This PR also delete the `scripts/embed-dashboard-ui.ps1` and `build.ps1` files, which I think they are un-used anymore.

### Release note

```release-note
None.
```


scripts/dashboard-version
```diff
@@ -0,0 +1,3 @@
+# This file is updated by running scripts/update-dashboard.sh
+# Don't edit it manullay
+2023.12.18.1
```
baurine, line 3: tidb-dashboard version will change to `x.y.z` (likes `7.6.0`) style in the future, same as TiDB/TiKV/PD.

rleungx (approved):

mornyx (approved):

HuSharp (approved):

Issue #7614, opened 2023-12-25
The PD API returned the wrong message


arturiamu:
### What did you do?
Send create scheduler request to pd through pd api

### What did you expect to see?
api returns correct information

### What did you see instead?
When creating a new evict-leader-scheduler, if the passed store_id does not exist, the pd api will still print ""The scheduler is created."" after returning "[PD:core:ErrStoreNotFound]store 999 not found".

### What version of PD are you using (`pd-server -V`)?
v7.5.0

### scene
```
❯ curl http://127.0.0.1:2379/pd/api/v1/schedulers
[
  "balance-region-scheduler",
  "balance-leader-scheduler",
  "balance-witness-scheduler",
  "balance-hot-region-scheduler",
  "transfer-witness-leader-scheduler"
]
❯ curl -H "Content-Type: application/json" -X POST -d '{"name": "evict-leader-scheduler", "store_id": 999}' http://127.0.0.1:2379/pd/api/v1/schedulers
"[PD:core:ErrStoreNotFound]store 999 not found"
"The scheduler is created."
```

rleungx:
Thanks for your feedback, would you like to send a patch to fix this problem?

arturiamu:
After my simple investigation, I found that the cause of the problem was: after the program returned an error message to the user in the `addEvictOrGrant` method, the "addEvictOrGrant" method did not return an error message to the caller `CreateScheduler`, resulting in `CreateScheduler` The method continues to return "The scheduler is created" information to the user, and since "grant-leader-scheduler" and "evict-leader-scheduler" share the "addEvictOrGrant" method, the problem also occurs in "evict-leader-scheduler".

And I'd be happy to fix this

rleungx:
Got it, you can raise a PR for it.

JackL9u:
The issue no longer exists.

```
> curl http://127.0.0.1:2379/pd/api/v1/schedulers
[
  "balance-region-scheduler",
  "balance-leader-scheduler",
  "balance-hot-region-scheduler",
  "evict-slow-store-scheduler"
]
> curl -H "Content-Type: application/json" -X POST -d '{"name": "evict-leader-scheduler", "store_id": 999}' http://127.0.0.1:2379/pd/api/v1/schedulers
"[PD:core:ErrStoreNotFound]store 999 not found"
```

PR #7618, opened 2023-12-25
api: Fix the return error of pd api in "grant-leader-scheduler" and "grant-leader-scheduler"

arturiamu:

Issue Number: ref https://github.com/tikv/pd/issues/7614

1.In the addEvictOrGrant method, a success message is returned to the user
2.Add a return statement ending method to the schedulers.GrantLeaderName and schedulers.EvictLeaderName branches of the CreateScheduler method to prevent the user from returning a success message at the end of the method

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/server/api/scheduler.go
+++ b/server/api/scheduler.go
@@ -144,10 +144,8 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
 
 	case schedulers.GrantLeaderName:
 		h.addEvictOrGrant(w, input, schedulers.GrantLeaderName)
-		return
 	case schedulers.EvictLeaderName:
 		h.addEvictOrGrant(w, input, schedulers.EvictLeaderName)
-		return
 	case schedulers.ShuffleLeaderName:
 		if err := h.AddShuffleLeaderScheduler(); err != nil {
 			h.r.JSON(w, http.StatusInternalServerError, err.Error())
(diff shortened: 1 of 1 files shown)


server/api/scheduler.go
```diff
@@ -213,7 +215,9 @@ func (h *schedulerHandler) addEvictOrGrant(w http.ResponseWriter, input map[stri
 	err := h.AddEvictOrGrant(storeID, name)
 	if err != nil {
 		h.r.JSON(w, http.StatusInternalServerError, err.Error())
+		return
 	}
+	h.r.JSON(w, http.StatusOK, "The scheduler is created.")
```
rleungx, line 220: Is the original one useless?

arturiamu: Since this method does not return an error, after returning user error information to the user, the outer layer will also return "The scheduler is created."

arturiamu: after the program returned an error message to the user in the addEvictOrGrant method, the "addEvictOrGrant" method did not return an error message to the caller CreateScheduler, resulting in CreateScheduler The method continues to return "The scheduler is created" information to the user, and since "grant-leader-scheduler" and "evict-leader-scheduler" share the "addEvictOrGrant" method, the problem also occurs in "evict-leader-scheduler".

rleungx: What I mean is can we remove line 206?

rleungx:
Please sign off the DCO. See https://github.com/tikv/pd/blob/master/CONTRIBUTING.md#signing-off-the-commit

server/api/scheduler.go
HuSharp: Maybe can modify the function `addEvictOrGrant` to return error, and then modify the function `CreateScheduler` to return the error in a uniform location?

Issue #7506, opened 2023-12-07
For some Gauge metrics items, `max` should be used


CabinfeverB:
## Enhancement Task
There are one tikv. I killed PD and restarted PD, and put a store.
I see

![image](https://github.com/tikv/pd/assets/23399268/01fd7655-8c75-4b32-910a-a935e0fa4c4e)
![image](https://github.com/tikv/pd/assets/23399268/52c9f9e9-1f06-4657-8142-618eedbec10c)

when PD killed, the prometheus metrics will not be reset, so `sum` maybe wrong

PR #7619, opened 2023-12-25
metrics: add instance var and improve conccurrency metrics

### LINKED ISSUE #7583: Use instance label wisely to optimize the presentation of Grafana
## Enhancement Task
After the microservices were implemented, the instance variable was removed from Grafana so that metrics for different instances could be displayed on the same page.
But this brings some issues:
1. The pd server does not reset metrics immediately after panic occurs. As a result, the monitored items using the sum function become larger such as #7506.
2. We support gRPC forwarding, which also causes the QPS we see to be higher than the actual one.
3. We will support the follower handle of gRPC and HTTP. We need to monitor QPS according to the instance.

CabinfeverB:

Issue Number: close #7583

add isntance var
add instance check for gRPC and HTTP
use `max_over_time` to  keep conccurrency metrics same with qps

```commit-message
```

### Check List

Tests

image
image
image
image

conccurrency
image

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


metrics/grafana/pd.json
```diff
@@ -132,7 +132,7 @@
       "tableColumn": "",
       "targets": [
         {
-          "expr": "sum(pd_cluster_status{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", type=\"storage_capacity\"})",
+          "expr": "max(pd_cluster_status{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", type=\"storage_capacity\"})",
```
rleungx, line 135: If the original leader panics, and then offline a store, will the monitor always show the original value?

CabinfeverB: Got it. This PR will not close #7506

rleungx (approved):

CabinfeverB:
cc @nolouch @HuSharp

HuSharp (approved):

PR #7620, opened 2023-12-25
api: Fix the return error of pd api in "grant-leader-scheduler" and "evict-leader-scheduler"

arturiamu:

Issue Number: ref https://github.com/tikv/pd/issues/7614

1.In the addEvictOrGrant method, a success message is returned to the user
2.Add a return statement ending method to the schedulers.GrantLeaderName and schedulers.EvictLeaderName branches of the CreateScheduler method to prevent the user from returning a success message at the end of the method

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


arturiamu:
Due to the DCO problem in [pr 7618](https://github.com/tikv/pd/pull/7618) , in order to facilitate the merge, a new merge request was submitted.

server/api/scheduler.go
```diff
@@ -144,8 +144,10 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
 
 	case schedulers.GrantLeaderName:
 		h.addEvictOrGrant(w, input, schedulers.GrantLeaderName)
+		return
 	case schedulers.EvictLeaderName:
 		h.addEvictOrGrant(w, input, schedulers.EvictLeaderName)
```
HuSharp, line 149: Maybe can modify the function `addEvictOrGrant` to return error, and then modify the function `CreateScheduler` to return the error in a uniform location?

arturiamu: Yes, I have also considered this method, but I noticed that many different types of errors will be generated in the addEvictOrGrant method. However, there is no definition of the related "missing store id" error in the errs package.

Now I added the "missing store id" error type definition, return the error in addEvictOrGrant and handle it uniformly in the handleAddEvictOrGrantErr method

HuSharp: If possible, could the function `CreateScheduler` be unified to return error as well? :)

pkg/errs/errno.go
```diff
@@ -217,6 +217,7 @@ var (
 	ErrServerNotStarted       = errors.Normalize("server not started", errors.RFCCodeText("PD:server:ErrServerNotStarted"))
 	ErrRateLimitExceeded      = errors.Normalize("rate limit exceeded", errors.RFCCodeText("PD:server:ErrRateLimitExceeded"))
 	ErrLeaderFrequentlyChange = errors.Normalize("leader %s frequently changed, leader-key is [%s]", errors.RFCCodeText("PD:server:ErrLeaderFrequentlyChange"))
+	ErrMissingStoreId         = errors.Normalize("missing store id", errors.RFCCodeText("PD:server:ErrMissingStoreId"))
```
HuSharp, line 220: maybe can include miss store id?

arturiamu: Sorry, I don't quite understand what you mean

HuSharp: Oh, just like `store id %d`, you can check the above `ErrLeaderFrequentlyChange` usage

server/api/scheduler.go
arturiamu: Do you mean not to use the handleAddEvictOrGrantErr method and return it directly in CreateScheduler?

HuSharp: yes, can we handle error in `CreateScheduler`?

arturiamu: If error handling is performed directly in CreateScheduler and returned, it will cause code redundancy in the GrantLeaderName branch and the EvictLeaderName branch.

HuSharp: - maybe can delete function `addEvictOrGrant` to unify 2 branch and pass scheduler-name:
`case schedulers.GrantLeaderName, schedulers.EvictLeaderName:`

- what I want is to delete the redundant error handle
```
	var serverError error

	switch name {
	case schedulers.BalanceLeaderName:
		serverError = h.AddBalanceLeaderScheduler()		
	case schedulers.BalanceWitnessName:
		serverError = h.AddBalanceWitnessScheduler()
	...
	}
	if serverError != nil {
		...
	}
		
```

PR #7667, opened 2024-01-04
client/http: do not retry the 4XX code returned by leader

JmPotato:

Issue Number: ref #7300.

```commit-message
When the PD leader returns a 404 status code, it no longer continues to retry but returns directly.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -147,6 +147,7 @@ func (ci *clientInner) requestWithRetry(
 	headerOpts ...HeaderOption,
 ) error {
 	var (
+		statusCode             int
 		err                    error
 		addr                   string
 		pdAddrs, leaderAddrIdx = ci.getPDAddrs()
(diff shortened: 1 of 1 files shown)


client/http/client.go
```diff
@@ -147,16 +147,18 @@ func (ci *clientInner) requestWithRetry(
...
 	)
 	// Try to send the request to the PD leader first.
 	if leaderAddrIdx != -1 {
 		addr = pdAddrs[leaderAddrIdx]
-		err = ci.doRequest(ctx, addr, reqInfo, headerOpts...)
-		if err == nil {
-			return nil
+		statusCode, err = ci.doRequest(ctx, addr, reqInfo, headerOpts...)
+		// No need to retry if there is no error or the status code is 404 returned by the PD leader.
+		if err == nil || statusCode == http.StatusNotFound {
```
CabinfeverB, line 160: Also for StatusInternalServerError and StatusBadRequest?

rleungx: IMO, only the latter one is no need.

rleungx: BTW, we also have API which will return the 403 error.

CabinfeverB: Or we discuss what situations need retries?

JmPotato: We can first exclude some obvious scenarios that do not require retrying, and then supplement more later.

CabinfeverB (approved):

HuSharp (approved):

Issue #7576, opened 2023-12-19
Enhance ServiceDiscovery to improve client high availability and load balance


CabinfeverB:
## Enhancement Task
Currently, the `ServiceDiscovery` is used in the pd gRPC client, and it is not directly exposed to the users. Now, we have encapsulated an HTTP SDK, which also requires service discovery functionality. Therefore, we need to optimize the abstraction of ServiceDiscovery to be used by both the pd gRPC client and the pd HTTP client.

Additionally, ServiceDiscovery also needs load balancing and better maintenance of the health status of leaders and followers, among other things. This aspect also requires optimization.

Moreover, after implementing the scheduling microservice, the related requests can be routed directly to the scheduling server through service discovery, instead of going through the API server. This reduces the resource cost on the API server as well.

PR #7668, opened 2024-01-04
client: use ServiceDiscovery to update member in HTTP client

CabinfeverB:

Issue Number: ref #7576

use ServiceDiscovery to update member in HTTP client

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/http/client.go
```diff
@@ -58,8 +57,8 @@ type clientInner struct {
 	cancel context.CancelFunc
 
 	sync.RWMutex
```
JmPotato, line 59: Is this lock still needed?

client/http/client.go
```diff
@@ -382,19 +270,36 @@ func WithLoggerRedirection(logLevel, fileName string) ClientOption {
 	return func(c *client) {}
 }
 
+// NewClientWithServiceDiscovery creates a PD HTTP client with the given PD service discovery.
+func NewClientWithServiceDiscovery(
+	source string,
+	sd pd.ServiceDiscovery,
+	opts ...ClientOption,
+) Client {
+	ctx, cancel := context.WithCancel(context.Background())
```
JmPotato, line 279: Should use a context passed through.

client/http/client.go
```diff
@@ -147,33 +117,18 @@ func (ci *clientInner) requestWithRetry(
...
-		addr                   string
-		pdAddrs, leaderAddrIdx = ci.getPDAddrs()
+		err        error
+		statusCode int
 	)
-	// Try to send the request to the PD leader first.
-	if leaderAddrIdx != -1 {
-		addr = pdAddrs[leaderAddrIdx]
+	clients := ci.sd.GetAllServiceClients()
+	for _, cli := range clients {
```
JmPotato, line 124: Can we maintain the original logic of requesting the leader first and then the followers? This is to reduce the unnecessary redirect as much as possible.

client/pd_service_discovery.go
```diff
@@ -140,20 +148,32 @@ var (
...
+	isLeader    bool
+	leaderAddr  string
 
 	networkFailure atomic.Bool
 }
 
-func newPDServiceClient(addr, leaderAddr string, conn *grpc.ClientConn, isLeader bool) ServiceClient {
+func newPDServiceClient(addr, leaderAddr string, tlsCfg *tls.Config, conn *grpc.ClientConn, isLeader bool) ServiceClient {
+	var httpAddress string
+	if strings.HasPrefix(addr, httpScheme) {
```
JmPotato, line 162: What if the address starts with `http` but still has a TLS config passed in? In this case, replacing the scheme is also needed.

client/pd_service_discovery.go
```diff
@@ -732,6 +780,21 @@ func (c *pdServiceDiscovery) GetServiceClient() ServiceClient {
 	return leaderClient
 }
 
+// GetAllServiceClients implments ServiceDiscovery
+func (c *pdServiceDiscovery) GetAllServiceClients() []ServiceClient {
```
JmPotato, line 784: Maybe return leader and follower addresses respectively could help with the above comment.

client/pd_service_discovery.go
CabinfeverB: Good catch. It seems that we didn't handle this case well before.

client/pd_service_discovery.go
CabinfeverB: It's both OK to me. My idea is that we only need one `for loop`, so I just return all clients together

client/http/client.go
```diff
@@ -382,19 +267,36 @@ func WithLoggerRedirection(logLevel, fileName string) ClientOption {
...
-	c := &client{inner: newClientInner(source), callerID: defaultCallerID}
+	ctx, cancel := context.WithCancel(context.Background())
+	c := &client{inner: newClientInner(ctx, cancel, source), callerID: defaultCallerID}
 	// Apply the options first.
 	for _, opt := range opts {
 		opt(c)
 	}
-	c.inner.setPDAddrs(pdAddrs, -1)
-	c.inner.init()
+	sd := pd.NewDefaultPDServiceDiscovery(ctx, cancel, pdAddrs, c.inner.tlsConf)
```
rleungx, line 298: why not just use newPDServiceDiscovery?

CabinfeverB: OK. I didn't think it was a good idea to pass in a lot of nil params before.

CabinfeverB: I forgot that `option` is unexported, so I think we can use `NewDefaultPDServiceDiscovery ` to avoid to input some unconcerned params.

client/http/client_test.go
```diff
@@ -101,77 +65,6 @@ func TestCallerID(t *testing.T) {
 	c.Close()
 }
 
-func TestRedirectWithMetrics(t *testing.T) {
```
CabinfeverB, line 104: move it into integration test.

client/http/client.go
JmPotato: Forgot about the logic. Then let's keep it the same.

client/http/client.go
```diff
@@ -464,16 +365,19 @@ func (c *client) request(ctx context.Context, reqInfo *requestInfo, headerOpts .
...
 }
 
-// UpdateMembersInfo updates the members info of the PD cluster in the inner client.
-// Exported for testing.
-func (c *client) UpdateMembersInfo() {
-	c.inner.updateMembersInfo(c.inner.ctx)
+// requestChecker is used to check the HTTP request sent by the client.
+type requestChecker struct {
+	checker func(req *http.Request) error
 }
```
JmPotato, line 371: What about just:

```suggestion
type requestChecker func(req *http.Request) error
```

client/http/client.go
```diff
@@ -464,16 +372,17 @@ func (c *client) request(ctx context.Context, reqInfo *requestInfo, headerOpts .
 		headerOpts...)
 }
 
-// UpdateMembersInfo updates the members info of the PD cluster in the inner client.
-// Exported for testing.
-func (c *client) UpdateMembersInfo() {
-	c.inner.updateMembersInfo(c.inner.ctx)
+// requestChecker is used to check the HTTP request sent by the client.
+type requestChecker func(req *http.Request) error
```
CabinfeverB, line 376: cc @HuSharp

client/retry/backoff.go
```diff
@@ -50,6 +51,7 @@ func (bo *Backoffer) Exec(
 	)
 	for {
 		err = fn()
+		fmt.Println(err)
```
HuSharp, line 54: redundant info?

client/http/client.go
```diff
@@ -151,35 +120,24 @@ func (ci *clientInner) requestWithRetry(
...
-		// Try to send the request to the other PD followers.
-		for idx := 0; idx < len(pdAddrs); idx++ {
-			if idx == leaderAddrIdx {
-				continue
-			}
-			addr = ci.pdAddrs[idx]
+		for _, cli := range clients {
+			addr := cli.GetHTTPAddress()
 			statusCode, err = ci.doRequest(ctx, addr, reqInfo, headerOpts...)
+			fmt.Println("statusCode", statusCode)
```
HuSharp, line 131: ditto

client/http/client.go
HuSharp: Is it just used in the test?

HuSharp: Is it just used in the test?

CabinfeverB: Yes

client/pd_service_discovery.go
```diff
@@ -732,6 +792,21 @@ func (c *pdServiceDiscovery) GetServiceClient() ServiceClient {
...
+	ret := make([]ServiceClient, 0)
```
HuSharp, line 797: Is it better to specify `make(client, 1)` or another larger value? perhaps the number of clients in the main scenario is greater than 0.

client/http/client.go
HuSharp: maybe better to add some comments for `NewHTTPClientWithRequestChecker`

client/http/client.go
```diff
@@ -151,35 +120,23 @@ func (ci *clientInner) requestWithRetry(
...
+				return err
 			}
-			log.Debug("[pd] request follower addr failed",
-				zap.String("source", ci.source), zap.Int("idx", idx), zap.String("addr", addr), zap.Error(err))
+			log.Debug("[pd] request addr failed",
+				zap.String("source", ci.source), zap.Bool("is-leader", cli.IsConnectedToLeader()), zap.String("addr", addr), zap.Error(err))
 		}
 		return err
 	}
+	fmt.Println("reqInfo.bo", reqInfo.bo == nil)
```
HuSharp, line 139: ditto

JmPotato (approved):

HuSharp (approved):

PR #7669, opened 2024-01-04
metrics: convert delta() usage to rate(), except for directional negative

### LINKED ISSUE #4489: `delta()` should not be used in counter

lhy1024:

Issue Number: Close #4489

We need to change the metrics for operator, etcd, and scheduler from delta() to rate(), and multiply by 60 to ensure consistency before and after the modification. 

For hot region and balance regions, we should not make changes because these metrics are directional and have positive or negative values.
```commit-message
```

### Check List

Tests

- Unit test
- Integration test
#### operator
- Before
image
image

- After
image
image

#### scheduler
- Before
image
image
- After 
image
image

#### etcd
- Before
image
- After
image

#### heartbeat
- Before
image
image
- After
image
image

### Release note

```release-note
None.
```


metrics/grafana/pd.json
```diff
@@ -11369,7 +11369,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(delta(pd_scheduler_region_heartbeat{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", store=~\"$store\", type=\"report\", status=\"err\"}[1m])) by (address, store)",
+              "expr": "sum(rate(pd_scheduler_region_heartbeat{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", store=~\"$store\", type=\"report\", status=\"err\"}[1m])) by (address, store)",
```
CabinfeverB, line 11372: need change L11279 also？

lhy1024: I think we don't change it. We have multiple 60 times in L11279, so we should use opm.

CabinfeverB: I mean that we can change it and don't multiply 60. So all expressions will use `ops`

metrics/grafana/pd.json
```diff
@@ -2692,7 +2692,7 @@
           },
           "yaxes": [
             {
-              "format": "opm",
+              "format": "ops",
```
rleungx, line 2695: Is it possible to use opm since the store limit is based on minutes?

CabinfeverB: I think it's both OK, but the units need to be the same for at least one plate.

CabinfeverB (approved):

metrics/grafana/pd.json
```diff
@@ -11179,7 +11179,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(delta(pd_scheduler_region_heartbeat{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", store=~\"$store\", type=\"report\", status=\"ok\"}[1m])) by (address, store)",
+              "expr": "sum(rate(pd_scheduler_region_heartbeat{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", store=~\"$store\", type=\"report\", status=\"ok\"}[1m])*60) by (address, store)",
```
CabinfeverB, line 11182: In TiKV side, the time window is 2m. Do we need keep same with TiKV?

rleungx (approved):

CabinfeverB (approved):

nolouch (approved):

Issue #7499, opened 2023-12-06
Reduce the impact of etcd leader on the availability of PD leader


JmPotato:
We have met several cases showing that even if the PD leader can no longer provide services, the etcd leader does not switch, resulting in the entire cluster being unable to provide services and affecting the high availability of PD, ref #7251, https://github.com/pingcap/tidb/issues/48204 and https://github.com/pingcap/tidb/issues/48206.

Because our implementation adopts the design of a PD leader elected based on the etcd leader, we need to determine more clearly whether the current unavailable scenario requires switching the etcd leader rather than just the PD leader to achieve the higher availability.

JmPotato:
Close with #7737.

PR #7670, opened 2024-01-05
lease: check etcd leader healthy by KeepAliveOnce

HuSharp:

Issue Number: Ref #7499

```commit-message
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/election/leadership.go
+++ b/pkg/election/leadership.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
+	"github.com/tikv/pd/pkg/cache"
 	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
@@ -97,6 +98,24 @@ func (ls *Leadership) setLease(lease *lease) {
 	ls.lease.Store(lease)
 }
 
+// GetUnHealthyTimesNum is used to get the unHealthy times of the leader lease within `DefaultTTL`.
+func (ls *Leadership) GetUnHealthyTimesNum() int {
+	l := ls.lease.Load()
+	if l == nil {
+		return 0
+	}
+	return l.(*lease).GetUnHealthyTimesNum()
+}
+
+// ResetUnHealthyTimesNum is used to reset the unHealthy times of the leader lease.
+func (ls *Leadership) ResetUnHealthyTimesNum(ctx context.Context) {
+	l := ls.lease.Load()
+	if l == nil {
+		return
+	}
+	l.(*lease).ResetUnHealthyTimes(ctx)
+}
+
 // GetClient is used to get the etcd client.
 func (ls *Leadership) GetClient() *clientv3.Client {
 	if ls == nil {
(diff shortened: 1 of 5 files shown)


pkg/election/lease.go
```diff
@@ -165,11 +213,22 @@ func (l *lease) keepAliveWorker(ctx context.Context, interval time.Duration) <-c
...
+				})
+
 				res, err := l.lease.KeepAliveOnce(ctx1, leaseID)
 				if err != nil {
 					log.Warn("lease keep alive failed", zap.String("purpose", l.Purpose), zap.Time("start", start), errs.ZapError(err))
 					return
 				}
+
+				if time.Since(start) < unhealthyTimesRecordTimeout {
+					l.removeUnHealthyTimesLock(start)
```
JmPotato, line 230: Could we only add the unhealthy time when it's greater than `unhealthyTimesRecordTimeout` rather than adding it first and then deleting it?

server/server.go
```diff
@@ -1781,6 +1783,13 @@ func (s *Server) campaignLeader() {
 				log.Info("etcd leader changed, resigns pd leadership", zap.String("old-pd-leader-name", s.Name()))
 				return
 			}
+			// check healthy status of etcd leader.
+			if s.member.GetLeadership().GetUnHealthyTimesNum() > unhealthyLeaderLeaseTimes {
```
JmPotato, line 1787: What about doing this check at the beginning of the tick to make sure we could resign the etcd leader as soon as possible?

pkg/election/lease.go
HuSharp: Yes, we can. But it will rely on upper layer timeout mechanism which reaches 3s(or greater)

add and then remove will be more fine-grained.

But I am not sure keeping fine-grained under 3s is necessary

server/server.go
HuSharp: I think 50ms maybe can be ignored

pkg/election/leadership.go
```diff
@@ -97,6 +98,24 @@ func (ls *Leadership) setLease(lease *lease) {
 	ls.lease.Store(lease)
 }
 
+// GetUnHealthyTimesNum is used to get the unHealthy times of the leader lease within `DefaultTTL`.
+func (ls *Leadership) GetUnHealthyTimesNum() int {
+	l := ls.lease.Load()
+	if l == nil {
+		return 0
+	}
```
JmPotato, line 106: What about using `ls.getLease()` directly?

pkg/election/leadership.go
```diff
@@ -97,6 +98,24 @@ func (ls *Leadership) setLease(lease *lease) {
...
+	return l.(*lease).GetUnHealthyTimesNum()
+}
+
+// ResetUnHealthyTimesNum is used to reset the unHealthy times of the leader lease.
+func (ls *Leadership) ResetUnHealthyTimesNum(ctx context.Context) {
+	l := ls.lease.Load()
+	if l == nil {
+		return
+	}
```
JmPotato, line 115: Ditto.

server/server.go
```diff
@@ -1781,6 +1783,14 @@ func (s *Server) campaignLeader() {
 				log.Info("etcd leader changed, resigns pd leadership", zap.String("old-pd-leader-name", s.Name()))
 				return
 			}
+			// check healthy status of etcd leader.
+			if s.member.GetLeadership().GetUnHealthyTimesNum() >= unhealthyLeaderLeaseTimes {
+				if err := s.member.ResignEtcdLeader(ctx, s.member.Name(), ""); err != nil {
```
JmPotato, line 1788: Better add a log here.

pkg/election/lease.go
```diff
@@ -32,6 +34,10 @@ const (
 	revokeLeaseTimeout = time.Second
 	requestTimeout     = etcdutil.DefaultRequestTimeout
 	slowRequestTime    = etcdutil.DefaultSlowRequestTime
+
+	unhealthyTimesRecordTimeout = 1 * time.Second
+	unhealthyTTLGCInterval      = 5 * time.Second
```
JmPotato, line 39: I prefer using a more conservative timeout value, e.g. the lease timeout of 3 seconds, since as long as the lease keeps alive successfully inside a lease timeout, it means the etcd leader is just fine. Based on this, manually controlling the GC might be a better option so we can clear the cache each time the lease is kept alive.

HuSharp:
Fixed by https://github.com/tikv/pd/pull/7737.

PR #7678, opened 2024-01-08
mcs: fix the prepare checker is directly skipped

### LINKED ISSUE #7671: There will be lots of balance leader operators after enabling scheduling service
## Enhancement Task
Screen Shot 2024-01-05 at 3 31 00 PM

It's not expected if the original cluster is balanced. From the log, it's indeed imbalanced.
Screen Shot 2024-01-05 at 3 33 29 PM

rleungx:

Issue Number: Close #7671.

It is because the scheduling service won't sync the region from either local storage or other PD. When it is started, the region tree will be empty and skip the prepare checker directly. The same phenomenon is also happening when using pd-recover to recover the pd cluster. It will wrongly send many operators especially balance leaders to TiKV. But this PR doesn't handle the pd-recover case.

```commit-message
```

### Check List

Tests

- Manual test

Using tiup playground to create a cluster and restart the scheduling service.
```
[2024/01/08 12:59:08.094 +08:00] [INFO] [prepare_checker.go:68] ["not loaded from storage region number is satisfied, finish prepare checker"] [not-from-storage-region=61] [total-region=61]
[2024/01/08 12:59:08.094 +08:00] [INFO] [coordinator.go:390] ["coordinator has finished cluster information preparation"]
[2024/01/08 12:59:08.094 +08:00] [INFO] [coordinator.go:400] ["coordinator starts to run schedulers"]
```

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/cluster.go
```diff
@@ -452,7 +452,8 @@ func (c *Cluster) runUpdateStoreStats() {
 func (c *Cluster) runCoordinator() {
 	defer logutil.LogPanic()
 	defer c.wg.Done()
-	c.coordinator.RunUntilStop()
+	// force wait for 1 minute to make prepare checker won't be directly skipped
```
lhy1024, line 455: Is minute enough?

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -452,7 +452,8 @@ func (c *Cluster) runUpdateStoreStats() {
...
+	c.coordinator.RunUntilStop(time.Minute)
```
lhy1024, line 456: Do we need to manual test about unnecessary balance leader schedule?

pkg/mcs/scheduling/server/cluster.go
rleungx: The heartbeat interval is 1 minute.

HuSharp: do we need to use a constant variable?

pkg/mcs/scheduling/server/cluster.go
rleungx: I think we can test it in dev env?

lhy1024 (approved):

nolouch (approved):

Issue #7673, opened 2024-01-08
add index failed with error '503 Service Unavailable' when kill pdleader


Lily2025:
### What did you do?
1、run add index or import into 
2、kill pdleader or inject pd leader network partition last for 1mins

### What did you expect to see?
add index or import into can success

### What did you see instead?
add index failed with error '503 Service Unavailable' when kill pdleader

add index failed at 2024-01-05 17:44:10 
Error 1105 (HY000): request pd http api failed with status: '503 Service Unavailable'

tidb logs：
[tidb-0-503.txt](https://github.com/tikv/pd/files/13856002/tidb-0-503.txt)
[tidb-1-503.txt](https://github.com/tikv/pd/files/13856003/tidb-1-503.txt)

### What version of PD are you using (`pd-server -V`)?
./pd-server -V
 Release Version: v7.6.0-alpha
Edition: Community
Git Commit Hash: c3ad361486ef29fb1a9001f5374788ad741d5616
Git Branch: heads/refs/tags/v7.6.0-alpha
UTC Build Time:  2024-01-05 11:36:37
2024-01-06T15:09:48.760+0800

PR #7680, opened 2024-01-08
client/http, retry: support backoffer in HTTP client

JmPotato:

Issue Number: ref #7300, #7673.

```commit-message
Support backoffer in HTTP client.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -293,8 +293,7 @@ func (ci *clientInner) updateMembersInfo(ctx context.Context) {
 		WithName(getMembersName).
 		WithURI(membersPrefix).
 		WithMethod(http.MethodGet).
-		WithResp(&membersInfo).
-		WithBackoffer(nil))
+		WithResp(&membersInfo))
 	if err != nil {
 		log.Error("[pd] http client get members info failed", zap.String("source", ci.source), zap.Error(err))
 		return
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -146,10 +146,12 @@ func (ci *clientInner) requestWithRetry(
 	reqInfo *requestInfo,
 	headerOpts ...HeaderOption,
 ) error {
+	var (
+		statusCode int
+		err        error
+	)
 	execFunc := func() error {
 		var (
-			statusCode             int
-			err                    error
 			addr                   string
 			pdAddrs, leaderAddrIdx = ci.getPDAddrs()
 		)
@@ -169,7 +171,7 @@ func (ci *clientInner) requestWithRetry(
 				continue
 			}
 			addr = ci.pdAddrs[idx]
-			_, err = ci.doRequest(ctx, addr, reqInfo, headerOpts...)
+			statusCode, err = ci.doRequest(ctx, addr, reqInfo, headerOpts...)
 			if err == nil || noNeedRetry(statusCode) {
 				break
 			}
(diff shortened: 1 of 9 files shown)


client/http/client.go
```diff
@@ -413,6 +421,13 @@ func (c *client) WithRespHandler(
 	return &newClient
 }
 
+// WithBackoffer sets and returns a new client with the given backoffer.
+func (c *client) WithBackoffer(bo *retry.Backoffer) Client {
```
HuSharp, line 425: Do we need to use a default backoff?

JmPotato: We could leave this as an option to the caller. If the caller needs one, it could initialize a backoffer by itself.

HuSharp: get it. Please add a test in client_test

client/http/interface.go
```diff
@@ -92,6 +93,8 @@ type Client interface {
 	// Additionally, it is important for the caller to handle the content of the response body properly
 	// in order to ensure that it can be read and marshaled correctly into `res`.
 	WithRespHandler(func(resp *http.Response, res interface{}) error) Client
+	// WithBackoffer sets and returns a new client with the given backoffer.
+	WithBackoffer(*retry.Backoffer) Client
```
lance6716, line 97: I prefer caller use `ClientOption` in `NewClient` to set it.

lance6716 (commented):
rest lgtm

client/http/interface.go
JmPotato: Using `WithBackoffer` here allows us to flexibly use different `Backoffer`s for different scenarios and interfaces like:

```go
cli := NewClient()
cli.WithBackoffer(bo1).GetRegion()
cli.WithBackoffer(bo2).GetStore()
```

If we use`ClientOption` for initialization, we may need to create a new `Client` every time we want to use a different `Backoffer`.

lance6716 (approved):

client/retry/backoff.go
```diff
@@ -48,6 +49,8 @@ func (bo *Backoffer) Exec(
 		after := time.NewTimer(currentInterval)
 		select {
 		case <-ctx.Done():
+			after.Stop()
+			return errors.Trace(ctx.Err())
```
HuSharp, line 53: Do we need to reset the backoffer here?

JmPotato: Nice catch. Fixed.

HuSharp (approved):

client/http/client.go
```diff
@@ -287,7 +293,8 @@ func (ci *clientInner) updateMembersInfo(ctx context.Context) {
 		WithName(getMembersName).
 		WithURI(membersPrefix).
 		WithMethod(http.MethodGet).
-		WithResp(&membersInfo))
+		WithResp(&membersInfo).
+		WithBackoffer(nil))
```
rleungx, line 297: If no need to back off, how about just removing WithBackoffer?

client/retry/backoff.go
```diff
@@ -18,65 +18,103 @@ import (
...
-		defer after.Stop()
+) (err error) {
+	defer bo.resetBackoff()
+	for {
+		err = fn()
+		if err == nil {
+			break
+		}
+		currentInterval := bo.nextInterval()
+		after := time.NewTimer(currentInterval)
```
rleungx, line 50: Is it necessary to create a timer every time?

rleungx (approved):

Issue #7251, opened 2023-10-25
pd can not select new leader which lead qps drop to zero when inject pdleader io delay 500ms last for 5mins


Lily2025:
### What did you do?
1、run workload
2、inject pdleader io delay 500ms last for 5mins
apiVersion: chaos-mesh.org/v1alpha1
kind: IOChaos
metadata:
  name: kv-timeout-data
  namespace: testbed-xxx
spec:
  action: latency
  mode: one
  selector:
    namespaces:
      - testbed-xxx
    labelSelectors:
      statefulset.kubernetes.io/pod-name: tc-pd-0
  volumePath: /var/lib/pd
  path: "/var/lib/pd/data/**/*"
  delay: "500ms"
  percent: 100
  duration: "300s"

### What did you expect to see?
1、pd can  select new leader when inject pdleader io delay 500ms last for 5mins
2、qps can recover within 2mins when inject pdleader io delay 500ms

### What did you see instead?
pd can not select new leader when inject pdleader io delay 500ms last for 5mins
write qps drop to zero
![3bf7daa4-ac99-4ae7-8651-52b59d4ea66e](https://github.com/tikv/pd/assets/84712107/5c53b413-8925-403c-8571-1a16dec97b8f)
![46775ab0-f274-434b-a4b8-b61f1b71e352](https://github.com/tikv/pd/assets/84712107/85b900dd-1b7b-45e2-9df0-84e9d0e34e03)

### What version of PD are you using (`pd-server -V`)?
./pd-server -V
 Release Version: v6.5.0-nightly
Edition: Community
Git Commit Hash: 77d6f5b6785eb4e153495bf1c32b165019ae119e
Git Branch: heads/refs/tags/v6.5.0-nightly
UTC Build Time:  2023-10-07 10:21:09
2023-10-19T09:33:53.650+0800

HuSharp:
### Situation

![img_v3_024k_d7092279-28a0-464a-be0e-36b77cb7bd1g](https://github.com/tikv/pd/assets/53859786/93e39af8-db2e-430c-86fd-4a2d1695c4d0)

It looks like pd0 kept trying to elect leader timed out for five minutes. Finally, pd2 is elected.
- It looks like pd0 is running for leader because the apply is slow (io delay).
- etcd leader was updated at the end of pd0's campaign. 
- etcd updated leader because heartbeat is slow.
- The qps drop to zero continue time depends on how long it took for etcd to change leader.

[...]

PR #7682, opened 2024-01-09
server/*: add trend to check etcd healthy

HuSharp:

Issue Number: Ref #7251

```commit-message
```
- introduce the `trend` struct which has 2 windows, related to [slow node detection](https://docs.google.com/presentation/d/1TsCp2NcR8vgKs7heslApVEkVPEBhGSULyRMerdtt7kg/edit#slide=id.g1c6e79701f9_0_107)
	- small windows duration is `20s`, big windows duration is `60s`
	- `rate = (small_windows_avg - big_windows_avg - unSenseVal)/small_windows_avg`
	- rate indicates the input's changing rate
	- The average value is from fsync which is recorded in etcd's server side
	- unSenseVal related to the environment, which is calculated by all average fsync val
- event been triggered by 2 aspects: etcd leader changed & rate less than point
	- put the check rate at the campaign leader
	- record rate every 1 seconds
- We have 4 formulas:
	- (A-B)/B >= m, (rate)
	- Sa = A*x, Sb = nB*x, (n is a multiple of window big for window small, x is small window duration)
	- Sb = δ + Sa
	-  δ = (big_dur-small_dur)*base (base determined by environment)
	We can get -> δ/Sb = δ/(δ+Sa) <= (m+1-n) => (big_dur-small_dur) * base / (base * bid_dur +spike * t) <= (m+1-n)
- for now, we want to monitor spike/base = 50, which means needs to grow 100 times what before
	- we can get: (big_dur-small_dur)*b / (small_dur*b + 100 * t * b) <= (m+1-n)/n
	- for now: big_dur=60, small_dur=20, n=3, s=50, and spike times be 3:  40b/(60b+50 * 3 * b) <= (m-2) -> m=2.19
	- I set n = 3. and when n is fixed increasing m means increasing the difficulty of triggering

rate graph
image
fsync graph
image

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/etcdutil/etcdutil.go
+++ b/pkg/utils/etcdutil/etcdutil.go
@@ -365,7 +365,7 @@ func (checker *HealthyChecker) inspector(ctx context.Context) {
 			for _, ep := range lastEps {
 				// add new trend
 				if _, ok := checker.allRecordUrls[ep]; !ok {
-					checker.allRecordUrls[ep] = trend.NewTrend(fmt.Sprintf("%s-%s", checker.name, ep), 20)
+					checker.allRecordUrls[ep] = trend.NewTrend(fmt.Sprintf("%s-%s", checker.name, ep), 20*time.Second)
 					etcdAsyncVals[ep] = trend.NewAsyncFromEtcd()
 				}
 			}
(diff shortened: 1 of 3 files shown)


HuSharp:
Fixed by https://github.com/tikv/pd/pull/7737.

PR #7691, opened 2024-01-10
mcs: add option to control redirect http request

### LINKED ISSUE #7690: Support using an option to decide if the request should be redirected to scheduling service
## Enhancement Task
It's hard to know if the rule is right since we can only get the rules from the scheduling service when it is enabled. We also need the real rules on the PD side so that we can know if there is a problem in synchronization through comparison.

lhy1024:

Issue Number: Close #7690

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/apiutil/serverapi/middleware.go
+++ b/pkg/utils/apiutil/serverapi/middleware.go
@@ -202,7 +202,7 @@ func (h *redirector) ServeHTTP(w http.ResponseWriter, r *http.Request, next http
 			return
 		}
 		clientUrls = append(clientUrls, targetAddr)
-		// Add a header to the response, it is used to check whether the request is forwarded to the micro service
+		// Add a header to the response, it is used to mark whether the request has been forwarded to the micro service.
 		w.Header().Add(apiutil.XForwardedToMicroServiceHeader, "true")
 	} else {
 		leader := h.s.GetMember().GetLeader()
--- a/pkg/utils/apiutil/serverapi/middleware.go
+++ b/pkg/utils/apiutil/serverapi/middleware.go
@@ -19,7 +19,6 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/pingcap/failpoint"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
 	mcsutils "github.com/tikv/pd/pkg/mcs/utils"
@@ -119,6 +118,9 @@ func (h *redirector) matchMicroServiceRedirectRules(r *http.Request) (bool, stri
 	if len(h.microserviceRedirectRules) == 0 {
 		return false, ""
 	}
+	if r.Header.Get(apiutil.XForbiddenForwardToMicroServiceHeader) == "true" {
+		return false, ""
+	}
 	// Remove trailing '/' from the URL path
 	// It will be helpful when matching the redirect rules "schedulers" or "schedulers/{name}"
 	r.URL.Path = strings.TrimRight(r.URL.Path, "/")
@@ -200,11 +202,8 @@ func (h *redirector) ServeHTTP(w http.ResponseWriter, r *http.Request, next http
 			return
 		}
 		clientUrls = append(clientUrls, targetAddr)
-		// Add a header to the response, this is not a failure injection
-		// it is used for testing, to check whether the request is forwarded to the micro service
-		failpoint.Inject("checkHeader", func() {
-			w.Header().Set(apiutil.ForwardToMicroServiceHeader, "true")
-		})
+		// Add a header to the response, it is used to check whether the request is forwarded to the micro service
+		w.Header().Add(apiutil.XForwardedToMicroServiceHeader, "true")
 	} else {
 		leader := h.s.GetMember().GetLeader()
 		if leader == nil {
--- a/tools/pd-ctl/pdctl/command/config_command.go
+++ b/tools/pd-ctl/pdctl/command/config_command.go
@@ -77,7 +77,7 @@ func NewShowConfigCommand() *cobra.Command {
 	sc.AddCommand(NewShowClusterVersionCommand())
 	sc.AddCommand(newShowReplicationModeCommand())
 	sc.AddCommand(NewShowServerConfigCommand())
-	sc.Flags().Bool(flagFromAPIServer, true, "read data from api server rather than micor service")
+	sc.Flags().Bool(flagFromAPIServer, false, "read data from api server rather than micor service")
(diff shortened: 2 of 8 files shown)


tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -437,6 +438,7 @@ func NewPlacementRulesCommand() *cobra.Command {
 	show.Flags().String("id", "", "rule id")
 	show.Flags().String("region", "", "region id")
 	show.Flags().Bool("detail", false, "detailed match info for region")
+	show.Flags().Bool(flagRedirectMicroService, true, "allow to redirect to micro service")
```
rleungx, line 441: What if the flag is misused in non-micro service env?

pkg/utils/apiutil/serverapi/middleware.go
```diff
@@ -119,6 +119,9 @@ func (h *redirector) matchMicroServiceRedirectRules(r *http.Request) (bool, stri
 	if len(h.microserviceRedirectRules) == 0 {
 		return false, ""
 	}
+	if r.URL.Query().Get("redirect_mcs") == "false" {
```
rleungx, line 122: redirect_to_ms?

tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -27,24 +27,26 @@ import (
...
+	clusterVersionPrefix     = "pd/api/v1/config/cluster-version"
+	rulesPrefix              = "pd/api/v1/config/rules"
+	rulesBatchPrefix         = "pd/api/v1/config/rules/batch"
+	rulePrefix               = "pd/api/v1/config/rule"
+	ruleGroupPrefix          = "pd/api/v1/config/rule_group"
+	ruleGroupsPrefix         = "pd/api/v1/config/rule_groups"
+	replicationModePrefix    = "pd/api/v1/config/replication-mode"
+	ruleBundlePrefix         = "pd/api/v1/config/placement-rule"
+	pdServerPrefix           = "pd/api/v1/config/pd-server"
+	flagRedirectMicroService = "redirect_to_mcs"
```
rleungx, line 49: Shall we make it easier to use?

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -343,6 +343,48 @@ func (suite *configTestSuite) checkConfig(cluster *pdTests.TestCluster) {
...
+	re.NoError(failpoint.Disable("github.com/tikv/pd/pkg/dashboard/adapter/skipDashboardLoop"))
+}
+
+func (suite *configTestSuite) checkConfigForwardControl(cluster *pdTests.TestCluster) {
+	re := suite.Require()
+	leaderServer := cluster.GetLeaderServer()
+	pdAddr := leaderServer.GetAddr()
+	cmd := ctl.GetRootCmd()
+
+	args := []string{"-u", pdAddr, "config", "show", "--from_api_server"}
```
rleungx, line 359: How do we know if this flag is effective?

pkg/utils/apiutil/serverapi/middleware.go
```diff
@@ -200,11 +202,8 @@ func (h *redirector) ServeHTTP(w http.ResponseWriter, r *http.Request, next http
 			return
 		}
 		clientUrls = append(clientUrls, targetAddr)
-		// Add a header to the response, this is not a failure injection
-		// it is used for testing, to check whether the request is forwarded to the micro service
-		failpoint.Inject("checkHeader", func() {
-			w.Header().Set(apiutil.ForwardToMicroServiceHeader, "true")
-		})
+		// Add a header to the response, it is used to mark whether the request has been forwarded to the micro service.
+		w.Header().Add(apiutil.XForwardedToMicroServiceHeader, "true")
```
CabinfeverB, line 206: Can we not put this header on every forward request? This reduces network traffic. For example, when the `len(r.Header.Get(apiutil.XForbiddenForwardToMicroServiceHeader)) > 0`, that's the same as making an extra mark, similar to failpoint.

CabinfeverB, line 206: Or can we can keep failpoint?  If `len(r.Header.Get(apiutil.XForbiddenForwardToMicroServiceHeader)) > 0` or failpoint enabled, we add the header into resp.

lhy1024: If `len(r.Header.Get(apiutil.XForbiddenForwardToMicroServiceHeader)) > 0` , code will not reach here

CabinfeverB: If it is as you say, then the new check in `pdctl/global.go` is also unreachable

tools/pd-ctl/tests/config/config_test.go
lhy1024: I have added more tests to check

pkg/utils/apiutil/serverapi/middleware.go
lhy1024: But if we use curl or pdctl to call api, how do we tell if it's coming from the scheduling server or the api server?

rleungx (approved):

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -343,6 +343,220 @@ func (suite *configTestSuite) checkConfig(cluster *pdTests.TestCluster) {
...
+		re.Len(ruleManager.GetAllRules(), 2)
+	}
+
+	// show placement rules
+	testRules(true /*isFromAPIServer*/, "show")
+	testRules(false /*isFromAPIServer*/, "show")
+
+	// load placement rules
+	testRules(false /*isFromAPIServer*/, "load", "--out="+fname)
+	testRules(false /*isFromAPIServer*/, "load", "--out="+fname)
```
rleungx, line 545: What is the difference?

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -343,6 +343,220 @@ func (suite *configTestSuite) checkConfig(cluster *pdTests.TestCluster) {
...
+
+	// show placement rules group
+	testRules(true /*isFromAPIServer*/, "rule-group", "show", placement.DefaultGroupID)
+	testRules(false /*isFromAPIServer*/, "rule-group", "show", placement.DefaultGroupID)
```
rleungx, line 549: If we need to test both true and false, why do we need this flag?

CabinfeverB (approved):

PR #7697, opened 2024-01-11
pd-ctl: add `region rule` to get the detail of region fit rule

### LINKED ISSUE #7690: Support using an option to decide if the request should be redirected to scheduling service

lhy1024:

Issue Number: Close #7690

```commit-message
```

### Check List

Tests

- Unit test
```
./pd/bin/pd-ctl -u http://127.0.0.1:59935 region rules 54                         ~ 
{
  "rule-fits": [
    {
      "rule": {
        "group_id": "pd",
        "id": "default",
        "start_key": "",
        "end_key": "",
        "role": "voter",
        "is_witness": false,
        "count": 3
      },
      "peers": [
        {
          "id": 55,
          "store_id": 1
        },
        {
          "id": 147,
          "store_id": 3
        },
        {
          "id": 249,
          "store_id": 2
        }
      ],
      "peers-different-role": null,
      "isolation-score": 0,
      "witness-score": 0
    }
  ],
  "orphan-peers": null
}
```

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/region_command.go
+++ b/tools/pd-ctl/pdctl/command/region_command.go
@@ -48,6 +48,7 @@ var (
 	regionsKeyspacePrefix   = "pd/api/v1/regions/keyspace"
 	regionIDPrefix          = "pd/api/v1/region/id"
 	regionKeyPrefix         = "pd/api/v1/region/key"
+	regionRulePrefix        = "pd/api/v1/config/rules/region"
 )
 
 // NewRegionCommand returns a region subcommand of rootCmd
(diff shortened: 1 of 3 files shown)


tools/pd-ctl/pdctl/command/region_command.go
```diff
@@ -48,6 +48,7 @@ var (
 	regionsKeyspacePrefix   = "pd/api/v1/regions/keyspace"
 	regionIDPrefix          = "pd/api/v1/region/id"
 	regionKeyPrefix         = "pd/api/v1/region/key"
+	regionRulePrefix        = "pd/api/v1/config/rules/region"
```
rleungx, line 51: Does it belong to region or config?

lhy1024: I'm not sure too. How about `config placement-rules region <region-id>` or `region rules <region-id>`

lhy1024:
it is duplicated with https://github.com/tikv/pd/pull/7697
`config placement-rules show --region <region-id> --detail`

PR #7729, opened 2024-01-18
etcdutil: remove client when etcd server is unhealthy

HuSharp:

Issue Number: Ref #7730

```commit-message
```
### Question summary:
- We have a etcd endpoints updated mechanism which will be balanced in etcd client's balance mechanism
- When the new PD leader sends an etcd request:
	- Write requests can only be handled by the etcd leader, at this time, the etcd leader does not have io hang problem, so there is no problem to write, and it can be elected as the PD leader successfully.
	- Read requests can be handled by all nodes and default to ReadIndex, at this point if the request is sent to a machine that has been injected with io hang, the read index is likely to get stuck and fail with a timeout because the application can't advance normally.

### PR Summary:
- When meeting `frequently error`, will remove client endpoints 
- client endpoints will be added after healthy check

### TODO
using trend to update ref #7682

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/server.go
```diff
@@ -1684,6 +1689,12 @@ func (s *Server) campaignLeader() {
 			log.Info(fmt.Sprintf("campaign %s leader meets error due to txn conflict, another PD/API server may campaign successfully", s.mode),
 				zap.String("campaign-leader-name", s.Name()))
 		} else {
+			// for frequently changed etcd leader, we should set client status to unhealthy
+			if strings.Contains(err.Error(), "frequently changed") {
```
lhy1024, line 1693: only this one kind of error?

pkg/utils/etcdutil/etcdutil.go
```diff
@@ -440,6 +458,13 @@ func (checker *healthyChecker) update() {
 			}
 			continue
 		}
+
+		if status, ok := checker.isUnHealthy.Load(ep); ok && status.(bool) {
+			log.Info("some etcd server is unhealthy", zap.String("endpoint", ep))
+			// set unhealthy etcd client to healthy
+			checker.isUnHealthy.Delete(ep)
```
lhy1024, line 465: If we remove `isUnHealthy` here, will we receive "frequently changed" error again？

HuSharp:
Since we need to get the state of all the pd's and then temporarily remove them in current PD own `etcd client checker`, I'll implement it in the trend.

Close this pr now.

server/server.go
HuSharp: will being checked by trend

pkg/utils/etcdutil/etcdutil.go
HuSharp: trend check will make sure this node not participate in campaigning leader

PR #7734, opened 2024-01-19
ctl: clone tls config to avoid reusing connection

### LINKED ISSUE #7739: pctl: `release-7.6` ctl failed to get stores using tls
### What did you do?
when using pdctl stores with tls, will return error 
img_v3_0278_76f83778-08ad-4909-8998-24726a4d4c4g

return normal stores information

### What did you see instead?
return error

release 7.6

HuSharp:

Issue Number: Close #7739

```commit-message
```

- There is a read goroutine inside http transport that detects the closure of the socket. When another request is made before the readLoop actually detects the closure, the socket closes and an EOF is read.
- When they have the same tls they will shared connection
- release 7.6 does update member at startup. 
- Since get stores are slow, update member is done when the request is sent and not yet returned.
- Then the member returns faster and sends the request to readloop to read the return and then closes the socket, and stores encounters an eof.

img_v3_0278_13250d7e-4d0e-4f82-847c-1e771ccf671g

can check error log
update member
img_v3_0278_57ae34aa-87b1-42c5-bcde-37e5eb5f467g
store url
img_v3_0278_33deb940-1b00-4d3f-aa8d-ad580064d01g
they are using URL in the same time
img_v3_0278_95c0ef3f-a3ad-4291-9df9-dbce55d5f93g
img_v3_0278_b1d54304-3c6b-4b8b-bad6-640bf11e197g

### Check List

Tests

### Release note

```release-note
ctl: clone tls config to avoid reusing connection
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/global.go
+++ b/tools/pd-ctl/pdctl/command/global.go
@@ -69,7 +69,7 @@ func InitHTTPSClient(pdAddrs, caPath, certPath, keyPath string) error {
 			&http.Transport{TLSClientConfig: tlsConfig}, pdControlCallerID),
 	}
 
-	SetNewPDClient(strings.Split(pdAddrs, ","), pd.WithTLSConfig(tlsConfig))
+	SetNewPDClient(strings.Split(pdAddrs, ","), pd.WithTLSConfig(tlsConfig.Clone()))
 
 	return nil
 }
--- a/tools/pd-ctl/tests/cert_opt.sh
+++ b/tools/pd-ctl/tests/cert_opt.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+cert_dir="$2"
+
+function generate_certs() {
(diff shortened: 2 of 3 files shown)


tools/pd-ctl/pdctl/command/global.go
```diff
@@ -69,7 +69,7 @@ func InitHTTPSClient(pdAddrs, caPath, certPath, keyPath string) error {
 			&http.Transport{TLSClientConfig: tlsConfig}, pdControlCallerID),
```
JmPotato, line 69: Should we clone here also to avoid the same problem appearing somewhere later?

HuSharp: In fact, our pd-ctl doesn't support concurrency, so I think that's probably not necessary.

CabinfeverB (approved):

HuSharp:
closed by #7741

Issue #7730, opened 2024-01-18
Enhance the detection mechanism for the unhealthy etcd node


JmPotato:
Part of #7499.

In various tests, we always observe continuous unavailability when injecting IO latency and other chaos into the PD leader node., like #6291. Upon further investigation of the logs, we discovered that the detection and eviction of unhealthy nodes are not always accurate. As a result, problematic etcd nodes can persistently impact our requests due to the round-robin balancer used by the etcd client. Especially during the leader switch, this problem could cause the PD leader unable to be stabilized and prolong the election, which affects the availability a lot.

```shell
2024-01-17 12:42:26 | {"level":"INFO","namespace":"endless-ha-test-tps-6150264-1-163","pod":"tc-pd-1","container":"pd","log":"[etcdutil.go:314] [\"update endpoints\"] [num-change=3->2] [last-endpoints=\"[http://tc-pd-2.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-0.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-1.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379]\"] [endpoints=\"[http://tc-pd-2.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-0.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379]\"]"} | 
2024-01-17 12:41:33 | {"pod":"tc-pd-0","container":"pd","level":"INFO","log":"[etcdutil.go:314] [\"update endpoints\"] [num-change=2->3] [last-endpoints=\"[http://tc-pd-2.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-0.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379]\"] [endpoints=\"[http://tc-pd-2.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-0.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379,http://tc-pd-1.tc-pd-peer.endless-ha-test-tps-6150264-1-163.svc:2379]\"]","namespace":"endless-ha-test-tps-6150264-1-163"} |  
[...]

PR #7737, opened 2024-01-19
etcdutil: consider the latency while patrolling the healthy endpoints

JmPotato:

Issue Number: ref #7730, #7499.

```commit-message
Consider the latency while patrolling the healthy endpoints to reduce the effect of slow nodes.
Now, there are the following strategies to select and remove unhealthy endpoints:

- Choose only the healthy endpoint within the lowest acceptable latency range.
- The evicted endpoint can only rejoin if it is selected again for three consecutive times.
```

### Check List

Tests

- Unit test
- Integration test

Inject the etcd leader IO latency like:

image

image

The duration of QPS being affected is reduced, and it may not necessarily hit rock bottom completely. It will only be impacted by the switch of the etcd leader.

image

Before:

image

Etcd leader changing will be finished in just one term.

image

Before:

image

Endpoint updating will be stabilized and only occur during IO hang injection and recovery points.

image

image

Before:

img_v3_0276_d4aef8d7-fb9e-4ae5-a0de-638c0905dccg

### Release note

```release-note
Enhance PD availability in the event of IO hang causing node unavailability.
```

### DIFF
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -10220,7 +10220,7 @@
           ],
           "thresholds": [],
           "timeRegions": [],
-          "title": "Endpoint Health State",
+          "title": "Endpoint health state",
           "tooltip": {
             "shared": true,
             "sort": 0,
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -10072,13 +10072,195 @@
           "dashLength": 10,
           "dashes": false,
           "datasource": "${DS_TEST-CLUSTER}",
+          "description": "The time consumed of etcd endpoint health check in .99",
+          "editable": true,
+          "error": false,
+          "fill": 1,
+          "fillGradient": 0,
+          "grid": {},
+          "gridPos": {
+            "h": 8,
+            "w": 24,
+            "x": 0,
+            "y": 53
+          },
+          "hiddenSeries": false,
+          "id": 1607,
+          "legend": {
+            "alignAsTable": true,
+            "avg": false,
+            "current": true,
+            "hideEmpty": true,
+            "hideZero": true,
+            "max": true,
+            "min": false,
+            "rightSide": true,
+            "show": true,
+            "sort": "current",
+            "sortDesc": false,
+            "total": false,
+            "values": true
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "options": {
+            "alertThreshold": true
+          },
+          "paceLength": 10,
+          "percentage": false,
+          "pluginVersion": "8.5.27",
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.99, sum(rate(pd_server_etcd_endpoint_latency_seconds_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=~\"$tidb_cluster.*\", source=\"server-etcd-client\"}[30s])) by (instance, endpoint, le))",
+              "intervalFactor": 2,
+              "legendFormat": "{{instance}} -> {{endpoint}}",
+              "range": true,
+              "refId": "A",
+              "step": 4
+            }
+          ],
+          "thresholds": [],
+          "timeRegions": [],
+          "title": "99% Endpoint health check latency",
+          "tooltip": {
+            "msResolution": false,
+            "shared": true,
+            "sort": 0,
+            "value_type": "cumulative"
+          },
+          "type": "graph",
+          "xaxis": {
+            "mode": "time",
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "s",
+              "logBase": 1,
+              "show": true
+            },
+            {
+              "format": "short",
+              "logBase": 1,
+              "show": true
+            }
+          ],
+          "yaxis": {
+            "align": false
+          }
+        },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "${DS_TEST-CLUSTER}",
+          "description": "The state of the endpoint health.",
+          "fill": 1,
+          "fillGradient": 0,
+          "gridPos": {
+            "h": 8,
+            "w": 24,
+            "x": 0,
+            "y": 61
+          },
+          "hiddenSeries": false,
+          "id": 1110,
+          "legend": {
+            "alignAsTable": true,
+            "avg": false,
+            "current": true,
+            "hideEmpty": true,
+            "hideZero": true,
+            "max": false,
+            "min": false,
+            "rightSide": true,
+            "show": true,
+            "total": false,
+            "values": true
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "options": {
+            "alertThreshold": true
+          },
+          "paceLength": 10,
+          "percentage": false,
+          "pluginVersion": "8.5.27",
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "stack": false,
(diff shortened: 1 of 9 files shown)


pkg/utils/etcdutil/health_checker.go
```diff
@@ -152,25 +163,133 @@ func (checker *healthChecker) patrol(ctx context.Context) []string {
...
+	var (
+		healthyCount = len(healthyProbes)
+		pickedEps    = make([]string, 0, healthyCount)
+	)
+	if healthyCount == 0 {
+		return pickedEps
+	}
+	factor := int(DefaultRequestTimeout / DefaultSlowRequestTime)
+	for i := 0; i < factor; i++ {
+		minLatency, maxLatency := DefaultSlowRequestTime*time.Duration(i), DefaultSlowRequestTime*time.Duration(i+1)
```
lhy1024, line 221: Do we need to add more comments here?

pkg/utils/etcdutil/health_checker.go
```diff
@@ -152,25 +163,133 @@ func (checker *healthChecker) patrol(ctx context.Context) []string {
...
+		pickedEps = checker.pickEps(healthProbes)
+	)
+	if len(pickedEps) > 0 {
+		checker.updateEvictedEps(lastEps, pickedEps)
+		pickedEps = checker.filterEps(pickedEps)
+	}
+	return lastEps, pickedEps, !typeutil.AreStringSlicesEquivalent(lastEps, pickedEps)
+}
+
+// Divide the acceptable latency range into several parts, and pick the endpoints which are in the first acceptable latency range.
```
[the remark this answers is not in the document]
JmPotato: Currently, the latency is sufficient for us to improve availability.

pkg/utils/etcdutil/health_checker.go
JmPotato: Added more comments. PTAL.

pkg/utils/etcdutil/health_checker.go
```diff
@@ -152,25 +163,147 @@ func (checker *healthChecker) patrol(ctx context.Context) []string {
...
+					lastHealth: start,
 				})
-				return
+			}
+			// Send the probe result to the channel.
+			probeCh <- healthProbe{
+				ep:      ep,
+				healthy: healthy,
+				took:    time.Since(start),
 			}
```
HuSharp, line 182: seems like can remove the `healthy` check?
we can only pass `healthy` healthProbe in probech?

HuSharp (commented):
Do we need to add some metrics to indicate healthy status?

pkg/utils/etcdutil/health_checker.go
JmPotato: Addressed.

JmPotato:
> Do we need to add some metrics to indicate healthy status?

Added. PTAL.

HuSharp:
> > Do we need to add some metrics to indicate healthy status?
> 
> Added. PTAL.

can you paste the metric pic in the pr description :)

metrics/grafana/pd.json
```diff
@@ -10066,6 +10066,202 @@
...
           }
         },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PB0C37C6A9EC17FDC"
```
HuSharp, line 10076: what does this mean?

pkg/utils/etcdutil/health_checker.go
```diff
@@ -137,40 +157,160 @@ func resetClientEndpoints(client *clientv3.Client, endpoints ...string) {
...
+// become available before adding them back to the client.
+func (checker *healthChecker) filterEps(eps []string) []string {
+	pickedEps := make([]string, 0, len(eps))
+	for _, ep := range eps {
+		if count, ok := checker.evictedEps.Load(ep); ok {
+			if count.(int) < pickedCountThreshold {
+				continue
+			}
+			checker.evictedEps.Delete(ep)
+			log.Info("add evicted etcd endpoint back",
```
HuSharp, line 306: do we need to add `source` in log as well?

HuSharp (approved):
rest LGTM!

pkg/utils/etcdutil/metrics.go
```diff
@@ -16,14 +16,32 @@ package etcdutil
...
+	typeLabel     = "type"
+	endpointLabel = "endpoint"
+)
+
+var (
+	etcdStateGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "pd",
+			Subsystem: "server",
+			Name:      "etcd_client",
```
HuSharp, line 30: maybe
```suggestion
			Name:      "etcd_client_state",
```

JmPotato: It already exists in several versions before, so I prefer not to modify it to maintain backward compatibility.

HuSharp: Oh, I forgot it!

pkg/utils/etcdutil/health_checker.go
```diff
@@ -137,46 +160,178 @@ func resetClientEndpoints(client *clientv3.Client, endpoints ...string) {
...
+	close(probeCh)
+	var (
+		lastEps   = checker.client.Endpoints()
+		pickedEps = checker.pickEps(probeCh)
+	)
+	if len(pickedEps) > 0 {
+		checker.updateEvictedEps(lastEps, pickedEps)
+		pickedEps = checker.filterEps(pickedEps)
+	}
+	return lastEps, pickedEps, !typeutil.AreStringSlicesEquivalent(lastEps, pickedEps)
```
JmPotato, line 218: Key modifications. You can specifically review this section.

pkg/utils/etcdutil/health_checker.go
```diff
@@ -137,46 +160,178 @@ func resetClientEndpoints(client *clientv3.Client, endpoints ...string) {
...
+}
+
+// Filter out the endpoints that are in evictedEps and have not been continuously picked
+// for `pickedCountThreshold` times still, this is to ensure the evicted endpoints truly
+// become available before adding them back to the client.
+func (checker *healthChecker) filterEps(eps []string) []string {
+	pickedEps := make([]string, 0, len(eps))
+	for _, ep := range eps {
+		if count, ok := checker.evictedEps.Load(ep); ok {
+			if count.(int) < pickedCountThreshold {
```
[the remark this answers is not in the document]
JmPotato: If the final picked endpoint is empty, we will not use it to replace the last:

https://github.com/JmPotato/pd/blob/17097f35dc235c94f63ff395956e7e5bad1da80b/pkg/utils/etcdutil/health_checker.go#L118-L132

pkg/utils/etcdutil/health_checker.go
```diff
@@ -137,46 +160,178 @@ func resetClientEndpoints(client *clientv3.Client, endpoints ...string) {
...
+	//   - [1s, 2s)
+	//   - [2s, 3s) -> {C}
+	//   - ...
+	//  - [9s, 10s)
+	// Then the picked endpoints will be {A, B} and if C is in the last used endpoints, it will be evicted later.
+	factor := int(DefaultRequestTimeout / DefaultSlowRequestTime)
+	for i := 0; i < factor; i++ {
+		minLatency, maxLatency := DefaultSlowRequestTime*time.Duration(i), DefaultSlowRequestTime*time.Duration(i+1)
+		for probe := range probeCh {
+			if minLatency <= probe.took && probe.took < maxLatency {
```
rleungx, line 248: What will happen if probe.took less than the minLatency?

pkg/utils/etcdutil/health_checker.go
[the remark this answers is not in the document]
rleungx: Then, we might need to reset all, right?

pkg/utils/etcdutil/health_checker.go
JmPotato: `minLatency` starts from zero. So this would not happen.

pkg/utils/etcdutil/health_checker.go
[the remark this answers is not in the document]
JmPotato: Do you mean we should reset all endpoints when there are no available endpoints to improve availability, considering the situation can't get any worse?

rleungx (approved):

PR #7738, opened 2024-01-19
mcs, schedule: use txn in label rule manager

lhy1024:

Issue Number: Ref #5839

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/labeler/labeler.go
+++ b/pkg/schedule/labeler/labeler.go
@@ -90,9 +90,11 @@ func (l *RegionLabeler) checkAndClearExpiredLabels() {
 		}
 		if len(rule.Labels) == 0 {
 			err = l.DeleteLabelRuleLocked(key)
-			deleted = true
+			if err == nil {
+				deleted = true
+			}
 		} else {
-			err = l.SetLabelRuleLocked(rule)
+			err = l.SaveLabelRuleLocked(rule)
 		}
 		if err != nil {
 			log.Error("failed to save rule expired label rule", zap.String("rule-key", key), zap.Error(err))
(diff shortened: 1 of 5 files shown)


pkg/schedule/labeler/labeler.go
```diff
@@ -88,11 +89,10 @@ func (l *RegionLabeler) checkAndClearExpiredLabels() {
 			continue
 		}
 		if len(rule.Labels) == 0 {
-			err = l.storage.DeleteRegionRule(key)
-			delete(l.labelRules, key)
+			err = l.DeleteLabelRuleLocked(key)
 			deleted = true
```
rleungx, line 93: If there is an error, `deleted` is still true?

pkg/schedule/labeler/labeler.go
```diff
@@ -88,11 +89,10 @@ func (l *RegionLabeler) checkAndClearExpiredLabels() {
...
 		} else {
-			err = l.storage.SaveRegionRule(key, rule)
+			err = l.SetLabelRuleLocked(rule)
```
rleungx, line 95: I prefer not to change the unnecessary code, it does one more check and memory modification.

pkg/schedule/labeler/labeler.go
```diff
@@ -124,7 +124,7 @@ func (l *RegionLabeler) loadRules() error {
 		return err
 	}
 	for _, d := range toDelete {
-		if err = l.storage.DeleteRegionRule(d); err != nil {
+		if err = l.DeleteLabelRuleLocked(d); err != nil {
```
rleungx, line 127: ditto, previous, it didn't change the memory.

lhy1024: we should keep storage and memory consistent

rleungx: The rule won't be add into memory.

rleungx (approved):

CabinfeverB (approved):

Issue #7703, opened 2024-01-12
Make bench tools more powerful


rleungx:
## Enhancement Task
To simulate the cases better, we should support more features for our bench tools, such as changing the bench parameters dynamically, etc.

PR #7745, opened 2024-01-22
tools: support tls and refactor

rleungx:

Issue Number: Ref #7703.

```commit-message
```

### Check List

Tests

- Manual test 

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-heartbeat-bench/main.go
+++ b/tools/pd-heartbeat-bench/main.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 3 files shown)


tools/pd-heartbeat-bench/main.go
```diff
@@ -293,24 +294,37 @@ func (rs *Regions) update(cfg *config.Config, options *config.Options, indexes [
...
 	}
 	// update interval
 	for _, region := range rs.regions {
 		region.Interval.StartTimestamp = region.Interval.EndTimestamp
 		region.Interval.EndTimestamp = region.Interval.StartTimestamp + regionReportInterval
 	}
-	for _, region := range updatedRegionsMap {
+	for _, i := range reportRegions {
+		region := rs.regions[i]
+		if r, exist := updatedRegionsMap[i]; exist {
```
CabinfeverB, line 306: What do `L306-L308`  mean

rleungx: replace by the region which is updated above

PR #7746, opened 2024-01-22
labeler: keep compatibility when meeting duplicate rules.

### LINKED ISSUE #7747: ticdc ci report `etcdserver: duplicate key given in txn request`
## Flaky Test

after #7738, we handle `/config/region-label/rules` with txn, if there are duplicate keys in the request, it will report this kind of error.

so we could need to keep compatibility about deleting and setting rules.

But I am not sure whether we need to do the same compatibility thing about placement rules.

lhy1024:

Issue Number: Close #7747

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/labeler/labeler.go
+++ b/pkg/schedule/labeler/labeler.go
@@ -266,15 +266,24 @@ func (l *RegionLabeler) DeleteLabelRuleLocked(id string) error {
 
 // Patch updates multiple region rules in a batch.
 func (l *RegionLabeler) Patch(patch LabelRulePatch) error {
+	// setRulesMap is used to solve duplicate entries in DeleteRules and SetRules.
+	// Note: We maintain compatibility with the previous behavior, which is to process DeleteRules before SetRules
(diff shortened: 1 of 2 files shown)


pkg/schedule/labeler/labeler.go
```diff
@@ -272,9 +272,20 @@ func (l *RegionLabeler) Patch(patch LabelRulePatch) error {
 		}
 	}
 
+	// Handling duplicate entries in DeleteRules and SetRules
+	// Note: We maintain compatibility with the previous behavior, which is to process DeleteRules before SetRules
+	// If there are duplicate rules, we will prioritize SetRules
+	deleteRules := make(map[string]struct{})
+	for _, ruleID := range patch.DeleteRules {
```
JmPotato, line 279: Iterate and build the set rules first could eliminate one traversal. We just need to skip the rules in set later while building the delete ruels.

rleungx (approved):

JmPotato (approved):

PR #7754, opened 2024-01-24
mcs: add store and region interface in scheduling server

lhy1024:

```commit-message
```

1. move store and region response to another pkg
2. add store and region interface in the scheduling server

### Check List

Tests

- Unit test
- Manual test
image
image
image
image

### Release note

```release-note
None.
```


pkg/response/store.go
```diff
@@ -0,0 +1,150 @@
...
+
+const (
+	// DisconnectedName is the name when store is disconnected.
+	DisconnectedName = "Disconnected"
+	// DownStateName is the name when store is down.
+	DownStateName = "Down"
+)
+
+// BuildStoreInfoResp builds a storeInfo response.
+func BuildStoreInfoResp(opt *sc.ScheduleConfig, store *core.StoreInfo) *StoreInfo {
```
rleungx, line 91: How about just `BuildStoreInfo`?

rleungx (approved):

pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -174,9 +177,19 @@ func (s *Service) RegisterOperatorsRouter() {
...
+	router := s.root.Group("stores")
+	router.GET("", getAllStores)
+	router.GET("/:id", getStoreByID)
+}
+
 // RegisterRegionsRouter registers the router of the regions handler.
 func (s *Service) RegisterRegionsRouter() {
 	router := s.root.Group("regions")
+	router.GET("", getAllRegions)
+	router.GET("/:id", getRegionByID)
```
CabinfeverB, line 191: Will we need to support GetRegionByKey in the future? If so, I think it's best to change the path to be consistent with api server?

lhy1024: I think the interface may not be supported. The main purpose of this pr is to determine if the region and store are successfully synchronized between the api server and the scheduling server. cc @rleungx

CabinfeverB: rest LGTM

CabinfeverB (approved):

pkg/mcs/scheduling/server/apis/v1/api.go
rleungx: The main purpose of this PR is to debug. If we do need it later, we can add it.

PR #7779, opened 2024-01-30
etcdutil: return original enpoints when all are evicted

JmPotato:

Issue Number: ref #7730, ref #7499.

```commit-message
Return the originally picked endpoints directly if all are evicted to gain better availability.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


HuSharp (commented):
Can you provide more information in the log message `etcd client is closed, exit update endpoint goroutine` to distinguish the situations in syncer and inspector?

HuSharp (approved):

rleungx (approved):

Issue #7782, opened 2024-02-01
Enhance go leak check


HuSharp:
## Enhancement Task
for example, we will meet goroutine leak which top stack is `runtime_pollWait` which resulted from `dashboard`
```bash
Goroutine 1362 in state IO wait, with internal/poll.runtime_pollWait on top of the stack:
goroutine 1362 [IO wait]:
internal/poll.runtime_pollWait(0x14dc55908, 0x72)
	/opt/homebrew/opt/go/libexec/src/runtime/netpoll.go:343 +0xa0
internal/poll.(*pollDesc).wait(0x14008d8e580?, 0x0?, 0x0)
	/opt/homebrew/opt/go/libexec/src/internal/poll/fd_poll_runtime.go:84 +0x28
internal/poll.(*pollDesc).waitRead(...)
	/opt/homebrew/opt/go/libexec/src/internal/poll/fd_poll_runtime.go:89
internal/poll.(*FD).Accept(0x14008d8e580)
	/opt/homebrew/opt/go/libexec/src/internal/poll/fd_unix.go:611 +0x250
net.(*netFD).accept(0x14008d8e580)
	/opt/homebrew/opt/go/libexec/src/net/fd_unix.go:172 +0x28
net.(*TCPListener).accept(0x140088d1840)
	/opt/homebrew/opt/go/libexec/src/net/tcpsock_posix.go:152 +0x28
net.(*TCPListener).Accept(0x140088d1840)
	/opt/homebrew/opt/go/libexec/src/net/tcpsock.go:315 +0x2c
github.com/pingcap/tidb-dashboard/pkg/tidb.(*proxy).run(0x14008d9e1e0, {0x104f79778?, 0x14007fd5c20})
	/Users/pingcap/go/pkg/mod/github.com/pingcap/tidb-dashboard@v0.0.0-20240111062855-41f7c8011953/pkg/tidb/proxy.go:227 +0x37c
created by github.com/pingcap/tidb-dashboard/pkg/tidb.(*Forwarder).Start in goroutine 1296
	/Users/pingcap/go/pkg/mod/github.com/pingcap/tidb-dashboard@v0.0.0-20240111062855-41f7c8011953/pkg/tidb/forwarder.go:57 +0x1d4
```

besides we will meet goroutine leak error which top stack is `runtime_pollWait` as well, but the root case is `go.etcd.io/etcd/pkg/transport.timeoutConn.Read` which is different with `dashboard`
```bash
[...]

PR #7786, opened 2024-02-01
server, allocator_manager: close grpc conn when close server

HuSharp:

Issue Number: Ref #7782, tested by #7777

```commit-message
server: close grpc conn when close server
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/mcs/resourcemanager/server/server.go
```diff
@@ -226,6 +226,14 @@ func (s *Server) Close() {
 		}
 	}
 
+	s.BaseServer.GetClientConns().Range(func(key, value any) bool {
```
rleungx, line 229: How about moving it to pkg/mcs/server/server.go?

HuSharp: extracted

rleungx (approved):

JmPotato (approved):

PR #7787, opened 2024-02-01
etcdutil_test: close  goroutine leak when block in accept

HuSharp:

Issue Number: Ref #7782, tested by #7777

```commit-message
etcdutil_test: close goroutine leak when block in accept
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/utils/etcdutil/etcdutil_test.go
```diff
@@ -307,59 +309,80 @@ func checkEtcdWithHangLeader(t *testing.T) error {
...
+				writeNum, errWrite := dst.Write(buffer[:readNum])
+				if errWrite != nil {
+					return errWrite
+				}
+				if readNum != writeNum {
+					return io.ErrShortWrite
+				}
+			}
+			if errRead != nil {
+				return err
```
rleungx, line 382: It might return nil?

HuSharp: yes, thx for the catch! fixed

rleungx (approved):

JmPotato (approved):

PR #7791, opened 2024-02-01
tools: improve flow update

rleungx:

Issue Number: Ref #7703.

```commit-message
```

### Check List

Tests

- Manual test

After this PR:

Screenshot 2024-02-02 at 17 39 31

### Release note

```release-note
None.
```


tools/pd-heartbeat-bench/main.go
```diff
@@ -46,9 +47,12 @@ import (
 )
 
 const (
-	bytesUnit            = 8 * units.MiB
-	keysUint             = 8 * units.KiB
-	queryUnit            = 1 * units.KiB
+	bytesUnit            = 128
```
lhy1024, line 50: When we use it, we need to note that item of tikv heartbeat is the cumulative, which is used for hotspot determination is the rate.

For example, if the hotKeysUint is 256, the calculated rate is 25.6 for reads and 4.2 for writes, which is lower than the threshold for the hot side, so it won't be treated as a hot region.

server/api/operator.go
```diff
@@ -101,6 +101,21 @@ func (h *operatorHandler) GetOperators(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// @Tags     operator
+// @Summary  Cancel all pending operators.
+// @Produce  json
+// @Success  200  {string}  string  "All pending operators are canceled."
+// @Failure  500  {string}  string  "PD server failed to proceed the request."
+// @Router   /operators [delete]
+func (h *operatorHandler) DeleteOperators(w http.ResponseWriter, r *http.Request) {
+	if err := h.RemoveOperators(); err != nil {
```
CabinfeverB, line 111: also use reason `operator.AdminStop`?

CabinfeverB (commented):
Add some tests for API?

server/api/operator.go
rleungx: Any better idea?

CabinfeverB (approved):

PR #7800, opened 2024-02-04
ci: putting tools tests in a separate chunk

### LINKED ISSUE #7805: ci: putting tools tests in a separate chunk
## Bug Report

tools ci not run

### What did you do?
run pr

run tools ci in pr

### What did you see instead?
not run tools ci

master

HuSharp:

Issue Number: Close #7805

```commit-message
```
because https://github.com/tikv/pd/blob/93f89f1bb2e0b6cd9e4e11a5d61cb32a0c0fd38a/scripts/ci-subtask.sh#L29 not contains tools after tools has own go mod

we can put tools tests in a separate chunk

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


.github/workflows/pd-tests.yaml
```diff
@@ -23,9 +23,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        worker_id: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]
```
CabinfeverB, line 26: Do we need to modify it? To do so, you need to change the github Settings related to the required CI

HuSharp: - I found that tools is the most time-consuming task, so I am checking the reason now.
- after putting tools tests in a separate chunk, the other unit test time is just 1-3 min when there are 10 chunks, so there are many redundant chunks...

> you need to change the github Settings related to the required CI

I am not sure about that, maybe you are right, we can rethink it in another pr

scripts/ci-subtask.sh
```diff
@@ -1,13 +1,19 @@
...
 # ./ci-subtask.sh <TOTAL_TASK_N> <TASK_INDEX>
 
 ROOT_PATH=../../
 
-if [[ $2 -gt 10 ]]; then
+if [[ $2 -gt 9 ]]; then
+    # run tools tests
+    if [[ $2 -eq 10 ]]; then
+        cd ./tools && make ci-test-job && cd .. && cat ./covprofile >> covprofile || exit 1
+        exit
```
JmPotato, line 11: Are the two `exit` duplicated?

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -545,7 +545,10 @@ func (suite *configTestSuite) checkConfigForwardControl(cluster *pdTests.TestClu
 }
 
 func (suite *configTestSuite) TestPlacementRules() {
-	suite.env.RunTestInTwoModes(suite.checkPlacementRules)
+	// use a new environment to avoid affecting other tests
+	env := pdTests.NewSchedulingTestEnvironment(suite.T())
+	env.RunTestInTwoModes(suite.checkPlacementRules)
+	env.Cleanup()
```
JmPotato, line 551: What about moving `NewSchedulingTestEnvironment` into `SetupTest` from `SetupSuite` to make sure each test uses a different env?

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -611,7 +614,10 @@ func (suite *configTestSuite) checkPlacementRules(cluster *pdTests.TestCluster)
 }
 
 func (suite *configTestSuite) TestPlacementRuleGroups() {
-	suite.env.RunTestInTwoModes(suite.checkPlacementRuleGroups)
+	// use a new environment to avoid being affected by other tests
+	env := pdTests.NewSchedulingTestEnvironment(suite.T())
+	env.RunTestInTwoModes(suite.checkPlacementRuleGroups)
+	env.Cleanup()
```
JmPotato, line 620: Ditto.

tools/pd-ctl/tests/config/config_test.go
```diff
@@ -688,7 +694,10 @@ func (suite *configTestSuite) checkPlacementRuleGroups(cluster *pdTests.TestClus
 }
 
 func (suite *configTestSuite) TestPlacementRuleBundle() {
-	suite.env.RunTestInTwoModes(suite.checkPlacementRuleBundle)
+	// use a new environment to avoid being affected by other tests
+	env := pdTests.NewSchedulingTestEnvironment(suite.T())
+	env.RunTestInTwoModes(suite.checkPlacementRuleBundle)
+	env.Cleanup()
```
JmPotato, line 700: Ditto.

tools/pd-ctl/tests/scheduler/scheduler_test.go
```diff
@@ -483,7 +483,8 @@ func (suite *schedulerTestSuite) checkScheduler(cluster *pdTests.TestCluster) {
 		pdTests.MustPutStore(re, cluster, store)
 	}
 	re.Equal("5.2.0", leaderServer.GetClusterVersion().String())
-	// After upgrading, we should not use query.
+	// After upgrading, we can use query.
+	expected1["write-leader-priorities"] = []any{"query", "byte"}
```
JmPotato, line 487: Why changed?

tools/pd-ctl/tests/tso/tso_test.go
```diff
@@ -36,7 +36,7 @@ func TestTSO(t *testing.T) {
 
 	// tso command
 	ts := "395181938313123110"
-	args := []string{"-u", "127.0.0.1", "tso", ts}
+	args := []string{"tso", ts}
```
JmPotato, line 39: We could keep this to test the case that even if the address given is invalid the command still works.

tools/pd-ctl/tests/scheduler/scheduler_test.go
HuSharp: It seems that after 5.2.0 defaultPrioritiesConfig is the query and byte.

https://github.com/tikv/pd/blob/cd0ffbafaa1adf9b3bc4d8cbe90e993c79f99e2b/pkg/schedule/schedulers/hot_region_config.go#L47

tools/pd-ctl/tests/config/config_test.go
HuSharp: I think setup just only once in a suite 0 0

tools/pd-ctl/tests/tso/tso_test.go
HuSharp: get it

tools/pd-ctl/tests/config/config_test.go
JmPotato: `SetupSuite` only runs once in the same suite, while `SetupTest` runs each time before a test function in a suite.

https://github.com/stretchr/testify/blob/v1.8.4/suite/interfaces.go#L21

HuSharp: get it

tools/Makefile
```diff
@@ -32,3 +32,15 @@ tidy:
...
 	git diff go.mod go.sum | cat
 	git diff --quiet go.mod go.sum
+
+ci-test-job: failpoint-enable
+	CGO_ENABLED=1 go test ./... -v -tags deadlock -race -cover || { $(MAKE) failpoint-disable && exit 1; }
+	$(MAKE) failpoint-disable
+
+failpoint-enable:
+	cd $(ROOT_PATH) && $(MAKE) failpoint-enable
+	go mod tidy
```
rleungx, line 42: Why do we need go mod tidy here?

HuSharp: copy from another Makefile... seems like assume It's possible that enabling these failpoints might also involve adding dependencies which need `go mod tidy`

HuSharp: This hypothesis doesn't seem to exist for the moment, so let's remove it for now

rleungx (approved):

PR #7806, opened 2024-02-05
*: fix context usage when watch etcd

### LINKED ISSUE #7807: Memory leak caused by watch
## Bug Report

Screenshot 2024-02-05 at 15 42 26

https://github.com/tikv/pd/blob/12d75c778d9360510686d5df8a4a9a6cfbdf0fac/pkg/mcs/metastorage/server/grpc_service.go#L105-L151

It won't exit even if the client is gone. watch global config also has this problem.

rleungx:

Issue Number: Close #7807

```commit-message
```

### Check List

Tests

- Manual test 

Use playground to start 3 TiDB, and kill one by one, the watch goroutines are reduced.

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/metastorage/server/grpc_service.go
+++ b/pkg/mcs/metastorage/server/grpc_service.go
@@ -86,7 +86,7 @@ func (s *Service) Watch(req *meta_storagepb.WatchRequest, server meta_storagepb.
 	if err := s.checkServing(); err != nil {
 		return err
 	}
-	ctx, cancel := context.WithCancel(s.ctx)
+	ctx, cancel := context.WithCancel(server.Context())
 	defer cancel()
 	options := []clientv3.OpOption{}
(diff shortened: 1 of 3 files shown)


pkg/mcs/metastorage/server/grpc_service.go
```diff
@@ -86,7 +86,7 @@ func (s *Service) Watch(req *meta_storagepb.WatchRequest, server meta_storagepb.
 	if err := s.checkServing(); err != nil {
 		return err
 	}
-	ctx, cancel := context.WithCancel(s.ctx)
+	ctx, cancel := context.WithCancel(server.Context())
```
CabinfeverB, line 89: Please also modify `WatchGlobalConfig`

CabinfeverB (approved):

HuSharp (approved):

PR #7809, opened 2024-02-06
server: pause scheduling immediately

rleungx:

Issue Number: Ref #5839.

We use a watch mechanism to sync the configuration to the scheduling server, so there might be a delay. In the online recovery scene, we need to pause scheduling immediately. In this PR, we drop all region heartbeat responses from the scheduling server once the `IsSchedulingHalted` is changed to true.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```

### DIFF
--- a/server/forward.go
+++ b/server/forward.go
@@ -32,6 +32,7 @@ import (
 	"github.com/tikv/pd/pkg/utils/grpcutil"
 	"github.com/tikv/pd/pkg/utils/logutil"
 	"github.com/tikv/pd/pkg/utils/tsoutil"
+	"github.com/tikv/pd/server/cluster"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


JmPotato (approved):

server/forward.go
```diff
@@ -262,6 +263,10 @@ func forwardRegionHeartbeatToScheduling(forwardStream schedulingpb.Scheduling_Re
 			errCh <- errors.WithStack(err)
 			return
 		}
+		// TODO: find a better way to halt scheduling immediately.
+		if rc.GetOpts().IsSchedulingHalted() {
+			continue
```
CabinfeverB, line 268: One problem is that the operator generated by the scheduling server still exists in memory. Is it possible to cancel all the scheduling server operators after the online recovery is complete so that the scheduling can be resumed as soon as possible?

rleungx: It depends. IMO, if those operators still take effect, cancel all is not the best way. But if the region is changed, we may need to cancel. So I prefer to leave it for now.

CabinfeverB (approved):

PR #7814, opened 2024-02-08
mcs: wait until API server is ready

rleungx:

Issue Number: Ref #5839.

TSO and scheduling server should keep waiting until the API server is ready.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/utils/util.go
+++ b/pkg/mcs/utils/util.go
@@ -123,7 +123,7 @@ func WaitAPIServiceReady(s server) error {
 	)
 	ticker := time.NewTicker(RetryIntervalWaitAPIService)
 	defer ticker.Stop()
-	for i := 0; i < MaxRetryTimesWaitAPIService; i++ {
+	for {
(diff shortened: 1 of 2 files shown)


pkg/mcs/utils/util.go
```diff
@@ -135,10 +135,6 @@ func WaitAPIServiceReady(s server) error {
 		case <-ticker.C:
 		}
 	}
-	if err != nil {
-		log.Warn("failed to check api server ready", errs.ZapError(err))
-	}
```
HuSharp, line 140: How about print warn log every 60 seconds or another better time?
I'm worried that all are debug logs won't be easy to find root case such as when the wrong api addr is set.

rleungx: make sense

HuSharp (approved):

nolouch (approved):

PR #7833, opened 2024-02-21
client: enabling TSO Follower Proxy with TSO service should fail 

JmPotato:

Issue Number: ref #5836 and pingcap/tidb#51194.

```commit-message
Since the TSO service does not support the TSO Follower Proxy, enabling it in this case should fail.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -702,6 +702,9 @@ func (c *client) UpdateOption(option DynamicOption, value any) error {
 			return err
 		}
 	case EnableTSOFollowerProxy:
+		if c.getServiceMode() != pdpb.ServiceMode_PD_SVC_MODE {
+			return errors.New("[pd] tso follower proxy is only supported in PD service mode")
+		}
 		enable, ok := value.(bool)
 		if !ok {
 			return errors.New("[pd] invalid value type for EnableTSOFollowerProxy option, it should be bool")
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -702,6 +702,9 @@ func (c *client) UpdateOption(option DynamicOption, value any) error {
 			return err
 		}
 	case EnableTSOFollowerProxy:
+		if c.getServiceMode() != pdpb.ServiceMode_PD_SVC_MODE {
```
rleungx, line 705: Will MaxTSOBatchWaitInterval also need it?

JmPotato: No, `MaxTSOBatchWaitInterval` is supported since it only relies on the client implementation.

tests/integrations/tso/client_test.go
```diff
@@ -269,10 +269,6 @@ func (suite *tsoClientTestSuite) TestDiscoverTSOServiceWithLegacyPath() {
 // TestGetMinTS tests the correctness of GetMinTS.
 func (suite *tsoClientTestSuite) TestGetMinTS() {
 	re := suite.Require()
-	if !suite.legacy {
-		suite.waitForAllKeyspaceGroupsInServing(re)
```
CabinfeverB, line 273: Why removing this

CabinfeverB (commented):
rest LGTM

tests/integrations/tso/client_test.go
JmPotato: It has been set in `SetupSuite`.

CabinfeverB (approved):

rleungx (approved):

PR #7847, opened 2024-02-22
client: support timeout in http client

CabinfeverB:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -42,8 +42,9 @@ const (
 	httpsScheme          = "https"
 	networkErrorStatus   = "network error"
 
-	defaultMembersInfoUpdateInterval = time.Minute
-	defaultTimeout                   = 30 * time.Second
+	// The value of defaultTimeout is the maximum value.
+	// If you want to shorten timeout, declare it when creating the client
+	defaultTimeout = 300 * time.Second
 )
 
 // respHandleFunc is the function to handle the HTTP response.
@@ -77,7 +78,7 @@ func newClientInner(ctx context.Context, cancel context.CancelFunc, source strin
 func (ci *clientInner) init(sd pd.ServiceDiscovery) {
 	// Init the HTTP client if it's not configured.
 	if ci.cli == nil {
-		ci.cli = &http.Client{Timeout: defaultTimeout}
+		ci.cli = &http.Client{}
 		if ci.tlsConf != nil {
 			transport := http.DefaultTransport.(*http.Transport).Clone()
 			transport.TLSClientConfig = ci.tlsConf
@@ -172,6 +173,7 @@ func (ci *clientInner) doRequest(
 		body        = reqInfo.body
 		res         = reqInfo.res
 		respHandler = reqInfo.respHandler
+		cancel      context.CancelFunc
 	)
 	logFields := []zap.Field{
 		zap.String("source", source),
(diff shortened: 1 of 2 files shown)


client/http/request_info.go
```diff
@@ -88,11 +89,14 @@ type requestInfo struct {
...
 	respHandler respHandleFunc
 	bo          *retry.Backoffer
+	timeout     time.Duration
 }
 
 // newRequestInfo creates a new request info.
 func newRequestInfo() *requestInfo {
-	return &requestInfo{}
+	return &requestInfo{
+		timeout: defaultTimeout,
```
JmPotato, line 98: What do you think about setting a default timeout for the client and passing it through? Leave the`requestInfo` always empty to customize.

rleungx: +1

CabinfeverB: Do you mean that it is better to set default timeout for `http.Client`?
I can have a try, but we should replace `innter client` to do it.
The reason why I made this change is because:
1. I just want to initialize an `http.Client`
2. The timeout value of `http.Client` used in TiDB is 0
3. The timeout value of `http.Client` used in client-go is 30s

client/http/client.go
```diff
@@ -181,6 +183,8 @@ func (ci *clientInner) doRequest(
 		zap.String("caller-id", callerID),
 	}
 	log.Debug("[pd] request the http url", logFields...)
+	ctx, cancel = context.WithTimeout(ctx, reqInfo.timeout)
```
niubell, line 186: Set a timeout for each request may not be a good idea. How about just supporting only one way to set timeout: init timeout in client with WithTimeout function, and timeout mechanism also through Timeout in http.Client?

CabinfeverB: Got it. My initial idea was to create as few http.clients as possible, but I've found that this is not a good way to modify it.

client/http/client.go
```diff
@@ -342,6 +367,13 @@ func (c *client) WithBackoffer(bo *retry.Backoffer) Client {
 	return &newClient
 }
 
+// WithTimeout sets and returns a new client with a new `http.Client` whose timeout is the given one.
+func (c *client) WithTimeout(dur time.Duration) Client {
+	newClient := *c
+	newClient.inner = c.inner.clone(withInnerTimeout(dur))
+	return &newClient
+}
```
JmPotato, line 375: The initial purpose of introducing `clientInner` is to allow safe sharing of the same inner core, even with multiple `Client` interfaces. This ensures that common components such as service discovery are reused effectively. However, this constraint appears to be violated. What about just keeping the timeout option as an initialization parameter instead of a dynamic variable?

JmPotato: Or we could make the timeout an upper layer's behavior rather than the HTTP client configuration.

CabinfeverB: Currently, Common components don't rely on `clientInner(http.Client)`. So I think it's safe to clone the `clientInner`.  However, I do not consider this parameter to be a dynamic variable because it creates a new `clientInner` without affecting the previous `clientInner`.

In fact, the reason of why I do this is because the timeout of `http. Client` in the store helper in TiDB is 0.  However, the timeout value of http. Client in client-go is 30 seconds.  Setting a default timeout for `http.Client` eliminates the need to set a timeout in context each time for most requests.

PR #7858, opened 2024-02-29
mcs: add config api for tso service

rleungx:

Issue Number: Ref #5839.

This PR adds a config API so that TiDB can use query to get the config of the TSO server.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/tso/server/apis/v1/api.go
+++ b/pkg/mcs/tso/server/apis/v1/api.go
@@ -95,7 +95,8 @@ func NewService(srv *tsoserver.Service) *Service {
 	}
 	s.RegisterAdminRouter()
 	s.RegisterKeyspaceGroupRouter()
-	s.RegisterHealth()
+	s.RegisterHealthRouter()
+	s.RegisterConfigRouter()
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


lhy1024 (approved):

pkg/mcs/tso/server/apis/v1/api.go
```diff
@@ -248,3 +255,13 @@ func GetKeyspaceGroupMembers(c *gin.Context) {
...
 	c.IndentedJSON(http.StatusOK, members)
 }
+
+// @Tags     config
+// @Summary  Get full config.
+// @Produce  json
+// @Success  200  {object}  server.Config
+// @Router   /config [get]
+func getConfig(c *gin.Context) {
+	svr := c.MustGet(multiservicesapi.ServiceContextKey).(*tsoserver.Service)
```
JmPotato, line 265: Is it possible to panic here?

rleungx: seem we need to use Get?

rleungx: But I think it barely happens.

JmPotato (approved):

Issue #7868, opened 2024-03-01
Add Heartbeat trace statistics help analyze performance bottlenecks


nolouch:
## Development Task
Example:
![image](https://github.com/tikv/pd/assets/6428910/f1b183b3-7339-4f4a-a2dc-1b2ed8ceec70)

we can found most of the time spent in `UpdateSubTree`

PR #7871, opened 2024-03-04
*: add region heartbeat duration breakdown metrics

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/7868

```commit-message
*: add region heartbeat duration breakdown metrics
- add a tracer during the heartbeat process
- statistic the lock wait time in RegionsInfo 
```

### Check List

Tests

- Unit test
- Integration test

image

- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/core/metrics.go
```diff
@@ -0,0 +1,245 @@
...
+	"go.uber.org/zap"
+)
+
+var (
+	// HeartbeatBreakdownHandleDurationSum is the summary of the processing time of handle the heartbeat stage.
+	HeartbeatBreakdownHandleDurationSum = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "pd",
+			Subsystem: "core",
+			Name:      "heartbeat_breakdown_handle_duration_seconds_sum",
```
JmPotato, line 30: Do we need to add a field to indicate it's the region's heartbeat?

pkg/core/metrics.go
```diff
@@ -0,0 +1,245 @@
...
+	h.startTime = now
+	h.lastCheckTime = now
+}
+
+func (h *regionHeartbeatProcessTracer) OnPreCheckFinished() {
+	now := time.Now()
+	h.preCheckDuration = now.Sub(h.lastCheckTime)
+	h.lastCheckTime = now
+	preCheckDurationSum.Add(h.preCheckDuration.Seconds())
+	preCheckCount.Inc()
```
JmPotato, line 164: Is it possible to flush these metrics asynchronously? It appears that setting the metrics in the middle of processing may affect the performance.

nolouch: I think it's ok because all of it use atomic. and we can dynamically disable the trace by the config `EnableHeartbeatBreakdownMetrics`

nolouch:
ptal @JmPotato @rleungx

pkg/core/region.go
```diff
@@ -1653,6 +1695,42 @@ func (r *RegionsInfo) GetRegionSizeByRange(startKey, endKey []byte) int64 {
 	return size
 }
 
+// metrics default poll interval
+const magicCount = 15 * time.Second
```
rleungx, line 1699: How about just DefaultPollInterval?

rleungx (commented):
BTW, will it cause too much lock contention?

nolouch:
@rleungx  This PR does not introduce additional locking and unlocking processes. All trace operations are added within the existing locks. Since metrics are directly atomic operations, I think it's okay, but it may increase some CPU time (holding locks). However, this CPU time should be relatively small compared to other parts.

rleungx (approved):

nolouch:
ptal @JmPotato

JmPotato (approved):

JmPotato:
Should we conduct a benchmark test on this feature to assess any potential impact on performance?

HuSharp:
why not add metrics in `grafana/pd.json`?

Issue #7894, opened 2024-03-07
Support Backoff mechanism for pd client(gRPC)


CabinfeverB:

rleungx:
tracked by #8047.

PR #7895, opened 2024-03-07
retry, client: support to log after execing some times

CabinfeverB:

Issue Number: ref #7894

```commit-message
```

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/retry/backoff.go
```diff
@@ -16,12 +16,27 @@ package retry
...
 	"github.com/pingcap/failpoint"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
 )
 
+// Option is used to customize the backoffer.
+type Option func(*Backoffer)
+
+// WithLogTimes sets the number of retries required to print a log.
+func WithLogTimes(logTimes int) Option {
```
rleungx, line 34: I think it should belong to the log component?

CabinfeverB: But the log component doesn't know how many times the backoffer has done?
How about providing a function param that the caller defines? like func `WithActionAfterTimes(times int, fn func()) Option`

rleungx: For the caller, I think it's different to know how to set this value, it's affected by the backoff policy. If we must do it, I prefer something like withMinLogInterval.

rleungx (approved):

client/retry/backoff_test.go
```diff
@@ -119,13 +119,28 @@ func TestBackofferWithLog(t *testing.T) {
...
+	bo := InitialBackoffer(time.Millisecond*10, time.Millisecond*100, time.Millisecond*1000, withMinLogInterval(time.Millisecond*100))
 	err := bo.Exec(ctx, testFn)
 	re.ErrorIs(err, errTest)
 
-	m1 := lg.Message()
-	rfc := `["call PD API failed and retrying"] [api=testFn] [retry-time=10] [error=test]`
-	re.Contains(m1, rfc)
+	ms := lg.Messages()
+	len1 := len(ms)
+	re.Len(ms, 9)
```
HuSharp, line 129: Can you add more information to explain why care 5 and 13?(such as (1+2+4+8)%10=5)

client/retry/backoff.go
```diff
@@ -16,15 +16,30 @@ package retry
...
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 )
 
 const maxRecordErrorCount = 20
 
+// Option is used to customize the backoffer.
+type Option func(*Backoffer)
+
+// withMinLogInterval sets the mininum log interval for retrying.
```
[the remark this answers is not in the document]
HuSharp: maybe can add more comments/change this function name, It seems unrelated to the **minimum**?

client/retry/backoff_test.go
```diff
@@ -107,3 +110,91 @@ func TestBackoffer(t *testing.T) {
...
+	re.Contains(ms[len1], rfc)
+}
+
+var errTest = errors.New("test")
+
+func testFn() error {
+	return errTest
+}
+
+// testingWriter is a WriteSyncer that writes to the the messages.
```
HuSharp, line 153:
```suggestion
// testingWriter is a WriteSyncer that writes the messages.
```

HuSharp (approved):

PR #7903, opened 2024-03-11
client: add status http interface

rleungx:

Issue Number: Ref #7300.

Add status API which TiDB uses it.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/types.go
+++ b/client/http/types.go
@@ -34,6 +34,15 @@ type ClusterState struct {
 	ReplicationStatus string    `json:"replication_status"`
 }
 
+// State is the status of PD server.
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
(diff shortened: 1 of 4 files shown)


client/http/types.go
```diff
@@ -34,6 +34,15 @@ type ClusterState struct {
 	ReplicationStatus string    `json:"replication_status"`
 }
 
+// State is the status of PD server.
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
```
HuSharp, line 38: Maybe it would be better to associate to the original type place?

rleungx: done

HuSharp (approved):

JmPotato (approved):

PR #7907, opened 2024-03-12
client: print more info when leader disconnect

### LINKED ISSUE #7906: Log more info when client can not connect to leader
## Enhancement Task
as title

CabinfeverB:

Issue Number: Close #7906

should be merged after #7911

```commit-message
```

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/pd_service_discovery.go
+++ b/client/pd_service_discovery.go
@@ -32,7 +32,6 @@ import (
 	"github.com/tikv/pd/client/errs"
 	"github.com/tikv/pd/client/grpcutil"
 	"github.com/tikv/pd/client/retry"
-	"go.uber.org/multierr"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -1023,7 +1022,7 @@ func (c *pdServiceDiscovery) switchLeader(url string) (change bool, err error) {
 		c.leader.Store(leaderClient)
 		// Run callbacks
 		if c.tsoGlobalAllocLeaderUpdatedCb != nil {
-			err = multierr.Append(err, c.tsoGlobalAllocLeaderUpdatedCb(url))
+			return change, c.tsoGlobalAllocLeaderUpdatedCb(url)
 		}
 		for _, cb := range c.leaderSwitchedCbs {
 			cb()
--- a/client/pd_service_discovery.go
+++ b/client/pd_service_discovery.go
@@ -1015,6 +1015,9 @@ func (c *pdServiceDiscovery) switchLeader(url string) (change bool, err error) {
 	}
 	if change {
 		// Set PD leader and Global TSO Allocator (which is also the PD leader)
+		// Note: Even if the connection is not established, the leader is still updated.
+		// The reason is that the leader has transferred to another PD server, and the client should
+		// update the member information. Meanwhile, the client must know the newest leader for follower proxy.
 		leaderClient := newPDServiceClient(url, url, newConn, true)
 		c.leader.Store(leaderClient)
 		// Run callbacks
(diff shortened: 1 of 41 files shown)


client/pd_service_discovery.go
```diff
@@ -1020,32 +1021,41 @@ func (c *pdServiceDiscovery) updateURLs(members []*pdpb.Member) {
...
-	newConn, err := c.GetOrCreateGRPCConn(addr)
+	var newConn *grpc.ClientConn
+	newConn, err = c.GetOrCreateGRPCConn(addr)
 	// If gRPC connect is created successfully or leader is new, still saves.
-	if addr != oldLeader.GetAddress() || newConn != nil {
-		// Set PD leader and Global TSO Allocator (which is also the PD leader)
-		leaderClient := newPDServiceClient(addr, addr, c.tlsCfg, newConn, true)
-		c.leader.Store(leaderClient)
+	if addr != oldLeader.GetAddress() {
+		change = true
```
rleungx, line 1039: It might fail to connect to the leader?

CabinfeverB: Yes, but the leader does change and it will affect the follower proxy

client/pd_service_discovery.go
```diff
@@ -990,30 +991,41 @@ func (c *pdServiceDiscovery) updateURLs(members []*pdpb.Member) {
...
+	}
+	if err == nil {
+		change = true
+		log.Info("[pd] successfully connected to leader", zap.String("leader", url))
+	} else {
+		log.Warn("[pd] failed to connect leader", zap.String("leader", url), errs.ZapError(err))
+	}
+	if change {
 		// Set PD leader and Global TSO Allocator (which is also the PD leader)
 		leaderClient := newPDServiceClient(url, url, newConn, true)
```
HuSharp, line 1018: Does `newConn` will be nill?

CabinfeverB: Even if the connection is not established, the leader is still updated.
The reason is that the leader has transferred to another PD server, and the client should update the member information. Meanwhile, the client must know the newest leader for follower proxy.

client/pd_service_discovery.go
```diff
@@ -990,30 +991,44 @@ func (c *pdServiceDiscovery) updateURLs(members []*pdpb.Member) {
...
 		c.leader.Store(leaderClient)
-	}
-	// Run callbacks
-	if c.tsoGlobalAllocLeaderUpdatedCb != nil {
-		if err := c.tsoGlobalAllocLeaderUpdatedCb(url); err != nil {
-			return true, err
+		// Run callbacks
+		if c.tsoGlobalAllocLeaderUpdatedCb != nil {
+			err = multierr.Append(err, c.tsoGlobalAllocLeaderUpdatedCb(url))
```
rleungx, line 1025: Why not just return the error?

CabinfeverB: I am not sure if it is necessary to return both connection establishment and call back errors

rleungx: The log already print it.

client/pd_service_discovery.go
```diff
@@ -991,30 +991,46 @@ func (c *pdServiceDiscovery) updateURLs(members []*pdpb.Member) {
...
 		return false, nil
 	}
 
-	newConn, err := c.GetOrCreateGRPCConn(url)
+	var newConn *grpc.ClientConn
+	newConn, err = c.GetOrCreateGRPCConn(url)
 	// If gRPC connect is created successfully or leader is new, still saves.
-	if url != oldLeader.GetURL() || newConn != nil {
+	if url != oldLeader.GetURL() {
+		change = true
```
JmPotato, line 1007: This check and introducing `change` seem unnecessary due to line 999.

CabinfeverB: If the client continues to fail to establish a connection with the leader, there will be situations where `oldURL` == `url` and `err != nil``. So it's not always `change = true` after line 999

Issue #7908, opened 2024-03-12
Monitors RU consumption to maximize short-term usage of MAX RU.


nolouch:
## Problem
I have a group:
![image](https://github.com/tikv/pd/assets/6428910/a59d50b3-58b6-4bf9-bfa2-8fa6d87a54e9)

When I only run workload A:
```
sysbench --mysql-user=user1 --mysql-host=10.2.12.53 --mysql-port=32335 --table-size=10000000  oltp_read_only run --time=1000 --report-interval=1 --threads=4  --mysql-db=test --tables=32 --rate=2
```
RU Avg is OK because the workload is stable.
![image](https://github.com/tikv/pd/assets/6428910/250c2ab9-d54c-4ffb-a990-4a55203f83d3)

But When I manually run the big query SQL like:

```
select count(*) from test.sbtest1;
```

The monitoring was not very accurate(on-avg), which led me to mistakenly believe that I was far away from triggering RC control. 
![image](https://github.com/tikv/pd/assets/6428910/67e1746c-b596-4cd3-b387-705c5e3af1bc)

But from the slow query: 
![image](https://github.com/tikv/pd/assets/6428910/04b9e2c2-844e-43b7-a0f1-21ae6a28bcff)
it cost 8000+ RU, and which let some query need wait in the RC queue

## Development Task

enhance the observability.

PR #7936, opened 2024-03-18
resource_manager: record the max RU per second

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/7908

When I only run workload A:
```
sysbench --mysql-user=user1 --mysql-host=10.2.12.53 --mysql-port=32335 --table-size=10000000  oltp_read_only run --time=1000 --report-interval=1 --threads=4  --mysql-db=test --tables=32 --rate=2
```
RU Avg is OK because the workload is stable.
image

But When I manually run the big query SQL like:

```
select count(*) from test.sbtest1;
```

The monitoring was not very accurate(on-avg), which led me to mistakenly believe that I was far away from triggering RC control. 
image

But from the slow query: 
image
it cost 8000+ RU, and which let some query need wait in the RC queue

```commit-message
resource_manager: record the max RU per second 
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -363,8 +363,6 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 	recordMaxTicker := time.NewTicker(tickPerSecond)
 	defer recordMaxTicker.Stop()
 	maxPerSecTrackers := make(map[string]*maxPerSecCostTracker)
-	rruSum := make(map[string]float64)
-	wruSum := make(map[string]float64)
 	for {
 		select {
 		case <-ctx.Done():
@@ -394,14 +392,19 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 				readRequestCountMetrics  = requestCount.WithLabelValues(name, name, readTypeLabel)
 				writeRequestCountMetrics = requestCount.WithLabelValues(name, name, writeTypeLabel)
 			)
+			t, ok := maxPerSecTrackers[name]
+			if !ok {
+				t = newMaxPerSecCostTracker(name, defaultCollectIntervalSec)
+				maxPerSecTrackers[name] = t
+			}
+			t.CollectConsumption(consumption)
+
 			// RU info.
 			if consumption.RRU > 0 {
 				rruMetrics.Add(consumption.RRU)
-				rruSum[name] += consumption.RRU
 			}
 			if consumption.WRU > 0 {
 				wruMetrics.Add(consumption.WRU)
-				wruSum[name] += consumption.WRU
 			}
 			// Byte info.
 			if consumption.ReadBytes > 0 {
@@ -448,8 +451,6 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 					availableRUCounter.DeleteLabelValues(r.name, r.name, r.ruType)
 					delete(m.consumptionRecord, r)
 					delete(maxPerSecTrackers, r.name)
-					delete(rruSum, r.name)
-					delete(wruSum, r.name)
 					readRequestUnitMaxPerSecCost.DeleteLabelValues(r.name)
 					writeRequestUnitMaxPerSecCost.DeleteLabelValues(r.name)
 				}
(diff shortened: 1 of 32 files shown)


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -41,7 +42,8 @@ const (
 	defaultConsumptionChanSize = 1024
 	metricsCleanupInterval     = time.Minute
 	metricsCleanupTimeout      = 20 * time.Minute
-	metricsAvailableRUInterval = 30 * time.Second
+	metricsAvailableRUInterval = 1 * time.Second
+	defaultCollectIntervalSec  = 20
```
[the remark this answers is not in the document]
nolouch: it's better a little bigger than that. so I choose 20 secs.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -437,21 +446,94 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
...
+	if deltaRRU > t.maxPerSecRRU {
+		t.maxPerSecRRU = deltaRRU
+	}
+	if deltaWRU > t.maxPerSecWRU {
+		t.maxPerSecWRU = deltaWRU
+	}
+	t.cnt++
+	// flush to metrics in every flushPeriod.
+	if t.cnt%t.flushPeriod == 0 {
+		t.rruMaxMetrics.Set(t.maxPerSecRRU)
```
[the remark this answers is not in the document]
nolouch: I think it's ok,  it may be slightly larger, but we do not require precision.

CabinfeverB (approved):

HuSharp (commented):
Do we need to update `metrics.json`?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -437,21 +447,94 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
...
+			names := make([]string, 0, len(m.groups))
+			for name := range m.groups {
+				names = append(names, name)
 			}
 			m.RUnlock()
+			for _, name := range names {
+				if maxPerSecTrackers[name] == nil {
+					maxPerSecTrackers[name] = newMaxPerSecCostTracker(name, defaultCollectIntervalSec)
+				}
+				maxPerSecTrackers[name].Observe(rruSum[name], wruSum[name])
```
glorv, line 488:
```suggestion
				if maxPerSecTrackers[name] == nil && (rruSum[name] + wruSum[name] > 0.0) {
					maxPerSecTrackers[name] = newMaxPerSecCostTracker(name, defaultCollectIntervalSec)
				}
				if maxPerSecTrackers[name] != nil {
					maxPerSecTrackers[name].Observe(rruSum[name], wruSum[name])
				}
```
Maybe skip inactive groups here is better

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -357,6 +360,11 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 	defer cleanUpTicker.Stop()
 	availableRUTicker := time.NewTicker(metricsAvailableRUInterval)
 	defer availableRUTicker.Stop()
+	recordMaxTicker := time.NewTicker(tickPerSecond)
+	defer recordMaxTicker.Stop()
+	maxPerSecTrackers := make(map[string]*maxPerSecCostTracker)
+	rruSum := make(map[string]float64)
+	wruSum := make(map[string]float64)
```
glorv, line 367: what about move rruSum, wruSum into maxPerSecCostTracker, no need to maintain 3 maps.

glorv (commented):
rest LGTM

pkg/mcs/resourcemanager/server/manager.go
nolouch: It may be inactive for a period of time, and active for a period of time. To reflect this trend of change, the cost is acceptable

HuSharp (approved):

nolouch:
ptal @glorv

glorv (commented):
LGTM

Issue #7849, opened 2024-02-23
(pdms)tpcc continues to report a lot of Error 8027 after disable/enable ms mode


mayjiang0203:
### What did you do?
Enable and disable pd ms mode by edit tc.

### What did you expect to see?
No error report by workload.

### What did you see instead?
```
root@benchtoolset-0:/go# go-tpc tpcc run -D tpcc --host tc-tidb.testbed-oltp-hm-bm8rb -P4000 --warehouses 10 -T 10 --time 36000m --ignore-error '2013,1213,1105,1205,8022,8028,9004,9007,1062' --user root --password '' --interval '10s'
[2024-02-22 09:32:28] execute run failed, err Error 8027: Information schema is out of date: schema failed to update in 1 lease, please make sure TiDB can connect to TiKV
[2024-02-22 09:32:28] execute run failed, err Error 8027: Information schema is out of date: schema failed to update in 1 lease, please make sure TiDB can connect to TiKV
[2024-02-22 09:32:32] execute run failed, err Error 8027: Information schema is out of date: schema failed to update in 1 lease, please make sure TiDB can connect to TiKV
[Current] DELIVERY - Takes(s): 0.9, Count: 1, TPM: 69.2, Sum(ms): 118.8, Avg(ms): 119.5, 50th(ms): 121.6, 90th(ms): 121.6, 95th(ms): 121.6, 99th(ms): 121.6, 99.9th(ms): 121.6, Max(ms): 121.6
[Current] NEW_ORDER - Takes(s): 9.9, Count: 22, TPM: 133.2, Sum(ms): 13133.6, Avg(ms): 595.5, 50th(ms): 21.0, 90th(ms): 31.5, 95th(ms): 3892.3, 99th(ms): 9126.8, 99.9th(ms): 9126.8, Max(ms): 9126.8
[Current] NEW_ORDER_ERR - Takes(s): 9.9, Count: 2, TPM: 12.1, Sum(ms): 13516.4, Avg(ms): 6710.9, 50th(ms): 5100.3, 90th(ms): 8589.9, 95th(ms): 8589.9, 99th(ms): 8589.9, 99.9th(ms): 8589.9, Max(ms): 8589.9
[...]

PR #7937, opened 2024-03-18
client/tso: fix the bug that TSO may hang when switching mode

JmPotato:

Issue Number: ref #7849.

```commit-message
Because the old `(*tsoClient).dispatchRequest` only checked the request's context,
this could potentially cause a `tsoRequest` to be sent to a batch channel that has already
finished `revokePendingTokenRequest`, thus being left there forever without returning.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -802,7 +802,11 @@ func (c *client) dispatchTSORequestWithRetry(req *tsoRequest) error {
 		err       error
 	)
 	for i := 0; i < dispatchRetryCount; i++ {
-		// Get the tsoClient for each retry, as it may be initialized or switched during the process.
+		// Do not delay for the first time.
+		if i > 0 {
+			time.Sleep(dispatchRetryDelay)
+		}
+		// Get the tsoClient each time, as it may be initialized or switched during the process.
 		tsoClient := c.getTSOClient()
 		if tsoClient == nil {
 			err = errs.ErrClientGetTSO.FastGenByArgs("tso client is nil")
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -782,23 +782,39 @@ func (c *client) GetLocalTSAsync(ctx context.Context, dcLocation string) TSFutur
...
+		tsoClient := c.getTSOClient()
+		if tsoClient == nil {
+			err = errs.ErrClientGetTSO.FastGenByArgs("tso client is nil")
+			continue
 		}
+		retryable, err = tsoClient.dispatchRequest(req)
+		if !retryable {
+			break
+		}
+		time.Sleep(dispatchRetryDelay)
```
CabinfeverB, line 815: No need wait for the last retry

CabinfeverB (commented):
rest LGTM

client/client.go
```diff
@@ -782,23 +782,39 @@ func (c *client) GetLocalTSAsync(ctx context.Context, dcLocation string) TSFutur
...
+	var (
+		retryable bool
+		err       error
+	)
+	for i := 0; i < dispatchRetryCount; i++ {
+		// Get the tsoClient for each retry, as it may be initialized or switched during the process.
+		tsoClient := c.getTSOClient()
+		if tsoClient == nil {
+			err = errs.ErrClientGetTSO.FastGenByArgs("tso client is nil")
+			continue
```
CabinfeverB, line 809: And do we need to wait some time for this case？
Maybe we can wait in L805 except first time

CabinfeverB (approved):

HuSharp (approved):

Issue #7930, opened 2024-03-18
Sort out the unit tests


JmPotato:
## Enhancement Task

At present, PD's unit tests are doped with some tests that will start the actual etcd/PD cluster. We should classify them as integrated tests and move them to the `tests` directory.

In addition, it is also necessary to provide a clearer `make` command to run the different tests, which helps us quickly verify any changes locally.

PR #7947, opened 2024-03-19
tests/etcdutil: reduce `TestRandomKillEtcd` test time

lhy1024:

Issue Number: ref #7930

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/testutil/testutil.go
+++ b/pkg/utils/testutil/testutil.go
@@ -104,6 +104,7 @@ func InitTempFileLogger(level string) (fname string) {
 	return fname
 }
 
+// GenerateTestDataConcurrently generates test data concurrently.
 func GenerateTestDataConcurrently(count int, f func(int)) {
 	var wg sync.WaitGroup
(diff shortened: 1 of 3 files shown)


HuSharp (approved):

pkg/utils/testutil/testutil.go
```diff
@@ -101,3 +103,24 @@ func InitTempFileLogger(level string) (fname string) {
...
+		go func() {
+			defer wg.Done()
+			for i := range tasks {
+				f(i)
+			}
+		}()
+	}
+	for i := 0; i < count; i++ {
+		tasks <- i
+	}
```
JmPotato, line 123: Why not just pass the `i` as a goroutine function parameter above?

lhy1024: Because we need to divide tasks to different goroutines

JmPotato (approved):

Issue #7969, opened 2024-03-22
Make Unit Test Great!


HuSharp:
## Development Task

- The unstable CI has been a hinder to our daily development. Sometimes we have to run it again and again, and again ... to merge a PR.
- When run `make basic-test` in local machine, there will be a lot of strange problems, as well as running extremely slow :(
> Need `5:44.15 total`  for Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GH, 144 core 72 CPU
> but local mac(m1pro 16G) needs > 15min

We use a wrong way to parallel. Some test cases are not side effect free, for example:

- the enable and disable of the failpoint
- modification of the configuration
- change of the global variable in a single process.
etc...

### Add back parallel by run a single test case in a isolated OS process #7952

If the test cases with side effect run parallelly with others, it could cause some test fail unexpectedly.
When they're in a single OS process, make them parallel is not a big deal.

#### print function time

when using ./pd-ut run --junitfile xxx, can get junitfile as XML file to get every function cost time.

### Furthermore

- handling the coverage.txt & integration with the CI script #7991
- classify some real integrated tests and move them to the tests directory #7930 
- support cache mechanism to accelerate bin build #8069
- change required ci https://github.com/ti-community-infra/configs/pull/1048
- support running integration test #8200
- support parallel parameter #8186
- unify tests HTTP client #8199
- support an alloc server to alloc URL #8203

### Related

- Ref TiDB https://github.com/pingcap/tidb/issues/30822
- Enhance the integration test of CI  https://github.com/tikv/pd/issues/7298

#### Local Test

Since we run parallel now, the CI runs faster than before, on my server computer (Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GH, 144 core 72 CPU):
`reduce from 6min -> 1min`

for make ut:
```bash
time make ut
building task finish, parallelism=144, count=641, takes=10.154869881s
run all tasks takes 50.559951196s
> make ut  327.15s user 48.61s system 615% cpu 1:01.03 total
```

for make basic-test
```bash
time make basic-test
CGO_ENABLED=1 go test -tags tso_function_test,deadlock -timeout 20m -race
> make basic-test 1176.46s user 169.41s system 473% cpu 5:44.15 total
```

#### CI Result
| Before | After |
|--------|-------|
[...]

PR #7998, opened 2024-03-28
tests/server: reduce test time for heartbeat msg cannot be triggered

HuSharp:

Issue Number: Ref #7969

```commit-message
```

### Performance
can reduce test time from 5s+1min(wait for keepalive) to 5s

### Why change?

Previous test want to mock client send region heartbeat to server, and then server will return msg to client.

But Previous test only can be triggered by keepalive https://github.com/tikv/pd/blob/f51f9134558e793e7ea08fc32cef672c7afa37b7/pkg/schedule/hbstream/heartbeat_streams.go#L154
So I changed region id to 0 which will send error msg https://github.com/tikv/pd/blob/5c58dddc66feb343e19d7f792852bb5b6184d9dd/server/grpc_service.go#L1279-L1284
and then client will get msg by https://github.com/tikv/pd/blob/f51f9134558e793e7ea08fc32cef672c7afa37b7/pkg/schedule/hbstream/heartbeat_streams.go#L127

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


tests/server/cluster/cluster_test.go
```diff
@@ -753,20 +753,19 @@ func TestConcurrentHandleRegion(t *testing.T) {
...
 		re.NoError(err)
-		regionID, err := id.Alloc()
-		re.NoError(err)
 		peer := &metapb.Peer{Id: peerID, StoreId: store.GetId()}
 		regionReq := &pdpb.RegionHeartbeatRequest{
 			Header: testutil.NewRequestHeader(clusterID),
 			Region: &metapb.Region{
-				Id:    regionID,
+				// mock error msg to trigger stream.Recv()
+				Id:    0,
```
rleungx, line 761: Will it skip some processes of handle region heartbeat?

HuSharp: yes, but this `regionReq` wants to ensure the client can get the server's response.

- only call `SendScheduleCommand` will send right msg by `oc.hbStreams.SendMsg(region, cmd)`
- `DispatchFromHeartBeat` will not trigger `SendScheduleCommand`

rleungx: I don't remember what is the purpose of this test. Does it aim to test the concurrent problem or race problem?

HuSharp: Yes, Seems like testing(including race) for concurrent running `HandleRegionHeartbeat`

HuSharp: I think core logic for this test is here 
https://github.com/tikv/pd/pull/7998/files#diff-4f23e8da76548a94e6779db8719386a8eabb85ad56deca5e6271a8e162652baaR789-R818

rleungx: If the heartbeat process exits earlier, will it still take effect?

HuSharp: I think this part is just like a `ping` which wants to ensure the connection between the client and server.
So I think the msg of the response is unimportant :)

And I have added some breakpoints to ensure concurrent region heartbeat has been handled.

HuSharp:
@rleungx @lhy1024 friendly ping :)

HuSharp:
@rleungx  @lhy1024  friendly ping :)

PR #8010, opened 2024-04-01
schedule: add check action when poll the opeators from opNotifierQueue

### LINKED ISSUE #7992: placement-rule: it can't transfer leader to the target store
### What did you do?
There are two data centers, one is dc1 and the other is dc2. At first we forced the leader at dc2 by placement-rule , and we switched the leader to dc1 by another placement-rule. But it left two region leader in dc2 at last.

All region leaders were transfered to dc2.

### What did you see instead?
There are two region leaders left in dc2 at last.

Release Version: v6.5.4-20231107-65de8fc

TonsnakeLin:

Issue Number: Close #7992 

```
It add light check for polling the opeators from Controller::opNotifierQueue. 
For example, check whether the region has disappeared，check whether the version of region epoch has changed for merge-region operator.
It has the chance to cancel the operators from Controller::opNotifierQueue execept timeout. We can only cancel the operator by timeout mechanism before.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/operator/operator_controller.go
+++ b/pkg/schedule/operator/operator_controller.go
@@ -459,7 +459,7 @@ func (oc *Controller) checkAddOperator(isPromoting bool, ops ...*Operator) (bool
 }
 
 // checkOperatorLightly checks whether the ops can be dispatched in Controller::pollNeedDispatchRegion.
-// The operators can't be dispatched for some scenarios, such as region disapeared, region changed ...
+// The operators can't be dispatched for some scenarios, such as region disappeared, region changed ...
 // `region` is the target region of `op`.
 func (oc *Controller) checkOperatorLightly(op *Operator) (*core.RegionInfo, CancelReasonType) {
 	region := oc.cluster.GetRegion(op.RegionID())
--- a/pkg/schedule/operator/operator_controller.go
+++ b/pkg/schedule/operator/operator_controller.go
@@ -471,7 +471,7 @@ func (oc *Controller) checkOperatorLightly(op *Operator) (*core.RegionInfo, Canc
 	// It may be suitable for all kinds of operator but not merge-region.
 	// But to be cautions, it only takes effect on merge-region currently.
 	// If the version of epoch is changed, the region has been splitted or merged, and the key range has been changed.
-	// The changing for conf_version of epoch doesn't modify the region key range, skipt it.
+	// The changing for conf_version of epoch doesn't modify the region key range, skip it.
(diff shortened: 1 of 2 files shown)


disksing (approved):

pkg/schedule/operator/operator_controller.go
```diff
@@ -455,6 +458,27 @@ func (oc *Controller) checkAddOperator(isPromoting bool, ops ...*Operator) (bool
...
+	if region == nil {
+		operatorCounter.WithLabelValues(op.Desc(), "not-found").Inc()
+		return nil, RegionNotFound
+	}
+
+	// It may be suitable for all kinds of operator but not merge-region.
+	// But to be cautions, it only takes effect on merge-region currently.
+	// If the version of epoch is changed, the region has been splitted or merged, and the key range has been changed.
+	// The changing for conf_version of epoch doesn't modify the region key range, skipt it.
+	if (op.Kind()&OpMerge != 0) && region.GetRegionEpoch().GetVersion() != op.RegionEpoch().GetVersion() {
```
rleungx, line 475: Do we need to compare the epoch like checkStaleOperator?

TonsnakeLin: `checkStaleOperator` only check the ConfVersion of region epoch which  maybe not resolve the issue https://github.com/tikv/pd/issues/7992. For example, "operator add split-region 258 --policy="approximate" will change the version of region 258, but not for ConfVersion.
Another reason is I don't know why limits the checkStaleOperator only to `DispatchFromHeartBeat` source https://github.com/tikv/pd/blob/master/pkg/schedule/operator/operator_controller.go#L120

rleungx: Yes, what I mean here is that is it enough to only check the version? Or we also need to check both the version and conf version?

TonsnakeLin: At begining, I checked both the version and conf-version. But I think the conf-version changed doesn't measn the key-range changed, the merge operator can still be executed sucessfully. At last I check only the version of RegionEpoch.
um... I'm a little not sure whether to consider conf-version.
Could you give me some suggestion ? @disksing @rleungx , Thank you!!

rleungx: BTW, I think there will be a problem if we use `!=` since the operator's epoch won't change after creating.

rleungx (approved):

PR #7898, opened 2024-03-07
Optimize heartbeat - async process statistics

nolouch:
### What problem does this PR solve?

Issue Number: Ref https://github.com/tikv/pd/issues/7897

### What is changed and how does it work?

```commit-message
Optimize heartbeat process
- Introduce `SyncRunner` and `AsyncRunner`
- Async the statics updates on subtree
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

Before:
image
image

After:
image
image

The duration of the pending state on tikvs has been reduced from 10 minutes to 3 minutes

### Release note

```release-note
None.
```

Issue #7897, opened 2024-03-07
Enhance heartbeat processing capabilities


nolouch:
#### Improve the diagnosis process and implement hierarchical processing for Heartbeat logic
- [x] https://github.com/tikv/pd/issues/7868
- [x] Async the logic not the relative metadata process
    - [x] Introduce a common async method with the resource limiter. https://github.com/tikv/pd/pull/8024
    - [x] async the statistical processing 
         - https://github.com/tikv/pd/pull/8053
         - https://github.com/tikv/pd/pull/8052

#### Optimize Lock and Reduce lock holding time

- [x] https://github.com/tikv/pd/issues/8136
- [x] Reduce operator controller locking competition https://github.com/tikv/pd/pull/8032
- [x] https://github.com/tikv/pd/pull/8104

#### Improve the memory alloc

- once there are 10M regions, the GC and memory issues will be significant.
    - https://github.com/tikv/pd/pull/8182
    - https://github.com/tikv/pd/pull/8180
    - https://github.com/tikv/pd/pull/8181
    - https://github.com/tikv/pd/pull/8188
    - https://github.com/tikv/pd/issues/8207
    - https://github.com/tikv/pd/pull/8213
    - https://github.com/tikv/pd/pull/8164
    

### Scalability

- [TBD] Proposal
    - only metadata(lookup) part, statics part not consider

PR #8024, opened 2024-04-03
pkg/ratelimit: introduce an executor that can run with a limiter

nolouch:

Issue Number: Ref https://github.com/tikv/pd/issues/7897

part of https://github.com/tikv/pd/pull/7898

```commit-message
pkg/ratelimit: Introduce an executor that can run with a limiter
- async runner can run the task with the limiter, the task will parallel run.
- sync runner used to keep the behavior like before
```

### Check List

Tests

- Unit test

```release-note
None.
```


nolouch:
PTAL @rleungx @HuSharp

pkg/ratelimit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+// Task is a task to be run.
+type Task struct {
+	Ctx         context.Context
+	Opts        TaskOpts
+	f           func(context.Context)
+	submittedAt time.Time
+}
+
+// ErrMaxWaitingTasksExceeded is returned when the number of waiting tasks exceeds the maximum.
+var ErrMaxWaitingTasksExceeded = errors.New("max waiting tasks exceeded")
```
CabinfeverB, line 43: Not used?

pkg/ratelimit/concurrency_limiter.go
```diff
@@ -84,3 +98,66 @@ func (l *concurrencyLimiter) getMaxConcurrency() uint64 {
...
+	if tt.released {
+		return
+	}
+	if tt.limiter.current == 0 {
+		panic("release token more than acquire")
+	}
+	tt.released = true
+	tt.limiter.current--
+	if len(tt.limiter.queue) < int(tt.limiter.limit) {
+		tt.limiter.queue <- tt
```
CabinfeverB, line 161: Is it expected to block?

pkg/ratelimit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+
+// Runner is the interface for running tasks.
+type Runner interface {
+	RunTask(ctx context.Context, opt TaskOpts, f func(context.Context)) error
+}
+
+// Task is a task to be run.
+type Task struct {
+	Ctx         context.Context
+	Opts        TaskOpts
```
CabinfeverB, line 37: Is it better to use pointer?

pkg/ratelimit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+	s.processPendingTasks()
+	select {
+	case s.taskChan <- task:
+	default:
+		s.pendingMu.Lock()
+		defer s.pendingMu.Unlock()
+		if len(s.pendingTasks) > 0 {
+			maxWait := time.Since(s.pendingTasks[0].submittedAt)
+			if maxWait > s.maxPendingDuration {
+				return errors.New("max pending duration exceeded")
```
CabinfeverB, line 146: use ErrMaxWaitingTasksExceeded

pkg/ratelimit/concurrency_limiter.go
nolouch: once the running task number reaches the limit, the `Acquire` blocking until a running task released. which means `release` will no block because there is a goroutine wait.

pkg/ratelimit/runner.go
nolouch: I think it's ok now

pkg/ratelimit/concurrency_limiter.go
```diff
@@ -84,3 +98,58 @@ func (l *concurrencyLimiter) getMaxConcurrency() uint64 {
...
+// Release releases the token.
+func (tt *TaskToken) Release() {
+	tt.limiter.mu.Lock()
+	defer tt.limiter.mu.Unlock()
+	if tt.released {
+		return
+	}
+	tt.released = true
+	tt.limiter.current--
+	if len(tt.limiter.queue) < int(tt.limiter.limit) {
```
rleungx, line 152: The `current` and the `queue` share the same limit?

pkg/ratelimit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+					token, err := task.Opts.Limit.Acquire(context.Background())
+					if err != nil {
+						log.Error("failed to acquire semaphore", zap.String("task-name", task.Opts.TaskName), zap.Error(err))
+						continue
+					}
+					go s.run(task.Ctx, task.f, token)
+				} else {
+					go s.run(task.Ctx, task.f, nil)
+				}
+			case <-s.stopChan:
```
rleungx, line 94: How about adding a log here?

pkg/ratelimit/concurrency_limiter.go
nolouch: yes.

CabinfeverB (approved):

PR #7991, opened 2024-03-27
*: support generate the test coverage file and using it in ci

HuSharp:
### What problem does this PR solve?

Issue Number: Ref #7969

need merge https://github.com/ti-community-infra/configs/pull/1048/files firstly

### add coverprofile for pd-ut
```bash
$ ./bin/pd-ut run pkg/audit --coverprofile xxx
2024/03/27 17:37:18 maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined
building task finish, parallelism=16, count=3, takes=2.976902167s
run all tasks takes 1.032669083s
```

```bash
support_coverprofile* $ cat xxx
mode: set
github.com/tikv/pd/pkg/audit/audit.go:44.58,45.37 1 1
github.com/tikv/pd/pkg/audit/audit.go:45.37,46.29 1 1
github.com/tikv/pd/pkg/audit/audit.go:46.29,48.4 1 1
github.com/tikv/pd/pkg/audit/audit.go:50.2,50.14 1 1
github.com/tikv/pd/pkg/audit/audit.go:59.48,61.2 1 0
github.com/tikv/pd/pkg/audit/audit.go:83.96,89.2 1 1
github.com/tikv/pd/pkg/audit/audit.go:92.81,94.9 2 1
github.com/tikv/pd/pkg/audit/audit.go:94.9,96.3 1 0
github.com/tikv/pd/pkg/audit/audit.go:97.2,98.9 2 1
github.com/tikv/pd/pkg/audit/audit.go:98.9,100.3 1 1
github.com/tikv/pd/pkg/audit/audit.go:101.2,102.13 2 1
github.com/tikv/pd/pkg/audit/audit.go:113.46,118.2 1 1
github.com/tikv/pd/pkg/audit/audit.go:121.68,123.9 2 1
github.com/tikv/pd/pkg/audit/audit.go:123.9,125.3 1 1
github.com/tikv/pd/pkg/audit/audit.go:126.2,127.13 2 1
```

```bash
support_coverprofile* $ go tool cover --func=xxx
github.com/tikv/pd/pkg/audit/audit.go:44:       Match                           100.0%
github.com/tikv/pd/pkg/audit/audit.go:59:       ProcessBeforeHandler            0.0%
github.com/tikv/pd/pkg/audit/audit.go:83:       NewPrometheusHistogramBackend   100.0%
github.com/tikv/pd/pkg/audit/audit.go:92:       ProcessHTTPRequest              87.5%
[...]

PR #8038, opened 2024-04-08
tests/rule: make TestLargeRules more stable

HuSharp:

Issue Number: Ref #7969

tested by #7969
Because below code will add default rule after the previous test finished, we need to skip default rule for large rule test to make it stable
https://github.com/tikv/pd/blob/7e95abff62564dbe683ee2359aa69769cab4dc0d/tests/server/api/rule_test.go#L62-L78 will 

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/api/rule_test.go
+++ b/tests/server/api/rule_test.go
@@ -1261,6 +1261,13 @@ func (suite *ruleTestSuite) postAndCheckRuleBundle(urlPrefix string, bundle []pl
 		err = tu.CheckGetJSON(testDialClient, urlPrefix+"/config/placement-rule", nil,
 			tu.StatusOK(re), tu.ExtractJSON(re, &respBundle))
 		re.NoError(err)
+		// skip default rule check
+		for i := range respBundle {
+			if respBundle[i].ID == placement.DefaultGroupID {
+				respBundle = append(respBundle[:i], respBundle[i+1:]...)
+				break
+			}
+		}
 		if len(respBundle) != len(bundle) {
 			return false
 		}


rleungx:
Is there any error stack?

HuSharp:
> Is there any error stack?

This condition will never be met, I printed `respBundle` which has one more default value `{GroupID: "pd", ID: "default", Role: "voter", Count: 3}  than resp (this default value is added by `TearDownTest`)
(https://github.com/tikv/pd/blob/7e95abff62564dbe683ee2359aa69769cab4dc0d/tests/server/api/rule_test.go#L1271-L1273)

tests/server/api/rule_test.go
```diff
@@ -1261,6 +1261,13 @@ func (suite *ruleTestSuite) postAndCheckRuleBundle(urlPrefix string, bundle []pl
 		err = tu.CheckGetJSON(testDialClient, urlPrefix+"/config/placement-rule", nil,
 			tu.StatusOK(re), tu.ExtractJSON(re, &respBundle))
 		re.NoError(err)
+		// skip default rule check
```
rleungx, line 1264: Will it affect `TestDeleteAndUpdate`?

HuSharp:
> > will add default rule after the previous test finished
> 
> Why does it happen? Does it share the same environment?

yes, this suite test build env firstly and then using `TearDownTest` to clean up placement rules

https://github.com/tikv/pd/blob/7e18a69734f2ae391516625ec0230c05b363f5c5/tests/server/api/rule_test.go#L51-L56

HuSharp:
/hold

meet other issues need to be fixed

HuSharp:
Root cause is 
- ruleTestSuite has run some heavy load tests and reusing the same environment
- we run `suiteTest` just by `1 CPU` resulting in the cluster maybe meeting leader transfer during the whole suite test
- I have allocated suiteTest more resources to avoid unnecessary error https://github.com/tikv/pd/pull/7991/files#diff-d1de7e6511458d33a769c0c501ec0ea45494dd9d2e830dc315e8fe623eb34b76R594-R598

So, we can close this pr now :)

PR #8046, opened 2024-04-09
cluster: print down peer when the related store is not disconnected

AndreMouche:
This PR will print down peer when the related store is not disconnected

Issue Number: Ref #4399  Close [#16279@tikv](https://github.com/tikv/tikv/issues/16279)

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/cluster/cluster.go
```diff
@@ -1003,6 +1003,19 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo, tracer cor
 		cluster.HandleStatsAsync(c, region)
 	}
 	tracer.OnAsyncHotStatsFinished()
+	// log down peers when the related store is not disconnected.
+	for _, p := range region.GetDownPeers() {
```
nolouch, line 1007: It will trigger every time in a heartbeat once a store is down....

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,30 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
+
+// logDownPeerOnConnectedStore logs the down peers on connected stores.
+// It is used to help users to know the down peer status.
+// TODO: It is not a good way to log during process region heartbeat, it is a temporary solution.
+// region: the region which has down peers
+// stores: the stores that the peers of the region belong to.
+func logDownPeerOnConnectedStore(region *core.RegionInfo, stores []*core.StoreInfo) {
+	if len(region.GetDownPeers()) == 0 {
+		return
+	}
```
nolouch, line 442: How about:

```
needLog := true
	if !store.IsUp() {
	   needLog = false
	   break
	}
}

if needLog {
 log.Warn("region has down peer on connected store",
			zap.Uint64("region-id", region.GetID()),
			zap.Strings("down-peers", p GetDownPeers()),
}
```

nolouch: I think `len(region.GetDownPeers()) == disconnectStoresNum` is hard to reach.

nolouch (approved):

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,24 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
 }
+
+// logDownPeerWithNoDisconnectedStore logs down peers on connected stores.
+// It won't log down peer when any store of the replica is disconnected which is
+// used to avoid too many logs when a store is disconnected.
+// TODO: it's not a good way to log down peer during process region heartbeat, we should handle it in another way.
+// region: the region which has down peer
+// stores: all stores that the region has peer on them
+func logDownPeerWithNoDisconnectedStore(region *core.RegionInfo, stores []*core.StoreInfo) {
+	for _, store := range stores {
```
nolouch, line 441: if len(stores) is 0, do we need also skip?

AndreMouche: It would be strange if there is a down-peer without a store, so it is necessary to log down peers in this situation I think.

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,24 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
+		if store.IsDisconnected() {
+			return
+		}
+	}
+	for _, p := range region.GetDownPeers() {
+		log.Warn("region has down peer on connected store",
```
rleungx, line 447: Will it print a lot?

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,24 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
+			zap.Uint64("region-id", region.GetID()),
+			zap.Uint64("down-peer", p.Peer.Id),
+			zap.Uint64("down seconds", p.DownSeconds),
```
rleungx, line 450:
```suggestion
			zap.Uint64("down-seconds", p.DownSeconds),
```

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,24 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
+func logDownPeerWithNoDisconnectedStore(region *core.RegionInfo, stores []*core.StoreInfo) {
+	for _, store := range stores {
+		if store.IsDisconnected() {
+			return
+		}
+	}
+	for _, p := range region.GetDownPeers() {
+		log.Warn("region has down peer on connected store",
+			zap.Uint64("region-id", region.GetID()),
+			zap.Uint64("down-peer", p.Peer.Id),
```
rleungx, line 449: p.GetPeer().GetId()?

pkg/statistics/region_collection.go
```diff
@@ -428,3 +430,24 @@ func notIsolatedStoresWithLabel(stores []*core.StoreInfo, label string) [][]*cor
...
+			zap.Uint64("down seconds", p.DownSeconds),
+			zap.Uint64("store-id", p.Peer.StoreId))
```
rleungx, line 451: ditto

rleungx (approved):

wuhuizuo:
/hold

a ci task failed.

PR #8052, opened 2024-04-10
*: optimize heartbeat process with concurrent runner - part 2

nolouch:

Issue Number: Ref https://github.com/tikv/pd/issues/7897

```commit-message
Optimize heartbeat process
- Split the statics updates on the subtree
```

### Check List

Tests

- Unit test
- Integration test

Before:
image
image

After:
image
image

The duration of the pending state on tikvs has been reduced from 10 minutes to 3 minutes

### Release note

```release-note
None.
```


nolouch:
PTAL @rleungx @JmPotato @HuSharp

server/cluster/cluster.go
```diff
@@ -1065,10 +1065,22 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
...
+		if overlaps, err = c.core.CheckAndPutRootTree(ctx, region); err != nil {
 			tracer.OnSaveCacheFinished()
 			return err
 		}
+		ctx.TaskRunner.RunTask(
+			ctx,
+			ratelimit.TaskOpts{
+				TaskName: "UpdateSubTree",
+				Limit:    ctx.Limiter,
+			},
```
JmPotato, line 1077: Could we provide a function to do the initialization work?

pkg/ratelimit/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var (
+	RunnerTaskMaxWaitingDuration = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "pd",
+			Subsystem: "ratelimit",
+			Name:      "runner_task_max_waiting_duration_seconds",
+			Help:      "The duration of tasks waiting in the runner.",
+		}, []string{"name"})
```
JmPotato, line 28: "name" could be reused by defining a const.

pkg/core/region.go
```diff
@@ -1004,6 +1004,162 @@ func (r *RegionsInfo) AtomicCheckAndPutRegion(ctx *MetaProcessContext, region *R
...
+				}
+			}
+		}
+	}
+	for _, peer := range region.GetMeta().GetPeers() {
+		storeID := peer.GetStoreId()
+		collectFromItemSlice(r.leaders, storeID)
+		collectFromItemSlice(r.followers, storeID)
+		collectFromItemSlice(r.learners, storeID)
+		collectFromItemSlice(r.witnesses, storeID)
```
JmPotato, line 1158: Is there a possibility that the following situation exists: I detected an overlap in the leaders tree and added `item.RegionInfo`, then, before I check other role trees, something like a leader transfer occurs. Will I not be able to get the latest `item.RegionInfo` because of this? Will there be any side effects?

rleungx (approved):

JmPotato (approved):

PR #8053, opened 2024-04-10
*: optimize heartbeat process with concurrent runner - part 1

nolouch:

Issue Number:  Ref https://github.com/tikv/pd/issues/7897

Part of https://github.com/tikv/pd/pull/8052

```commit-message
*: Optimize the heartbeat process with the concurrent runner 
- add a switch about the runner
- move some tasks out, not block the heartbeat process
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/cluster.go
```diff
@@ -81,6 +89,9 @@ func NewCluster(parentCtx context.Context, persistConfig *config.PersistConfig,
 		storage:           storage,
 		clusterID:         clusterID,
 		checkMembershipCh: checkMembershipCh,
+
+		taskRunner:           ratelimit.NewAsyncRunner("heartbeat-async-task-runner", time.Minute),
```
HuSharp, line 93: Should we put all the tasknames into a constant list?

pkg/core/region.go
```diff
@@ -711,20 +714,51 @@ func (r *RegionInfo) isRegionRecreated() bool {
...
+		d = log.Debug
+		i = log.Info
+		debug, info = d, i
 	}
 	// Save to storage if meta is updated.
 	// Save to cache if meta or leader is updated, or contains any down/pending peer.
-	return func(region, origin *RegionInfo) (saveKV, saveCache, needSync bool) {
+	return func(ctx context.Context, region, origin *RegionInfo) (saveKV, saveCache, needSync bool) {
+		taskRunner, ok := ctx.Value(ctxutil.TaskRunnerKey).(ratelimit.Runner)
+		limiter, _ := ctx.Value(ctxutil.LimiterKey).(*ratelimit.ConcurrencyLimiter)
```
JmPotato, line 734: If the task runner and limiter are set into the context definitely, why not make them the function parameters instead of retrieving them from the context every time?

server/cluster/cluster.go
```diff
@@ -1032,43 +1072,72 @@ func (c *RaftCluster) processRegionHeartbeat(region *core.RegionInfo, tracer cor
...
+				ratelimit.TaskOpts{
+					TaskName: "SaveRegionToKV",
+					Limit:    c.hbConcurrencyLimiter,
+				},
+				func(ctx context.Context) {
+					// If there are concurrent heartbeats from the same region, the last write will win even if
+					// writes to storage in the critical area. So don't use mutex to protect it.
+					// Not successfully saved to storage is not fatal, it only leads to longer warm-up
+					// after restart. Here we only log the error then go on updating cache.
+					for _, item := range overlaps {
```
HuSharp, line 1124: why move `for range` info `saveKV`?

pkg/core/region.go
nolouch: Using context can reduce the number of function parameters. Sometimes some parameters are not required, such as when the limiter can be nil for testing or lighter tasks.

pkg/cluster/cluster.go
```diff
@@ -55,8 +56,13 @@ func HandleOverlaps(c Cluster, overlaps []*core.RegionInfo) {
...
-func Collect(c Cluster, region *core.RegionInfo, stores []*core.StoreInfo, hasRegionStats bool) {
+func Collect(c Cluster, region *core.RegionInfo, hasRegionStats bool) {
 	if hasRegionStats {
-		c.GetRegionStats().Observe(region, stores)
+		// get region again from root tree. make sure the observed region is the latest.
+		region = c.GetBasicCluster().GetRegion(region.GetID())
+		if region == nil {
+			return
+		}
+		c.GetRegionStats().Observe(region, c.GetBasicCluster().GetRegionStores(region))
```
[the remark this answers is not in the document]
nolouch: BaiscCluster was created after initialized, and do not changs.

server/cluster/cluster.go
nolouch: Combine it into a single task to execute simultaneously, which also helps reduce issues with parallel execution order.

nolouch:
PTAL @HuSharp @JmPotato

pkg/core/region.go
```diff
@@ -711,20 +714,51 @@ func (r *RegionInfo) isRegionRecreated() bool {
...
+		// print log asynchronously
+		if ok {
+			debug = func(msg string, fields ...zap.Field) {
+				taskRunner.RunTask(
+					ctx,
+					ratelimit.TaskOpts{
+						TaskName: "Log",
+						Limit:    limiter,
+					},
+					func(_ context.Context) {
```
HuSharp, line 744: Do we need the context parameter?

nolouch: It's according to the task logic.

JmPotato (approved):

HuSharp (approved):

pkg/core/region.go
```diff
@@ -711,20 +713,50 @@ func (r *RegionInfo) isRegionRecreated() bool {
...
 	// Save to cache if meta or leader is updated, or contains any down/pending peer.
-	return func(region, origin *RegionInfo) (saveKV, saveCache, needSync bool) {
+	return func(ctx *MetaProcessContext, region, origin *RegionInfo) (saveKV, saveCache, needSync bool) {
+		taskRunner := ctx.TaskRunner
+		limiter := ctx.Limiter
+		// print log asynchronously
+		debug, info := d, i
+		if taskRunner != nil {
+			debug = func(msg string, fields ...zap.Field) {
+				taskRunner.RunTask(
```
rleungx, line 736: Is the error needed?

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -81,6 +91,9 @@ func NewCluster(parentCtx context.Context, persistConfig *config.PersistConfig,
 		storage:           storage,
 		clusterID:         clusterID,
 		checkMembershipCh: checkMembershipCh,
+
+		taskRunner:           ratelimit.NewConcurrentRunner(hbConcurrentRunner, time.Minute),
```
rleungx, line 95: Will it cause the memory increase if there are many hb requests?

nolouch: it depends on the number of waiting queue closures (tasks). we suppose each costs 128 bytes, 1M is about 128MB. but it's released very quickly I think.

pkg/core/region.go
nolouch: I will add metrics about this error in next PR.

rleungx (approved):

PR #8060, opened 2024-04-12
*: fix client log

rleungx:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/tso_service_discovery.go
+++ b/client/tso_service_discovery.go
@@ -465,7 +465,7 @@ func (c *tsoServiceDiscovery) updateMember() error {
 	oldGroupID := c.GetKeyspaceGroupID()
 	if oldGroupID != keyspaceGroup.Id {
 		log.Info("[tso] the keyspace group changed",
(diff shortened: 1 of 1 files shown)


client/tso_service_discovery.go
```diff
@@ -465,7 +465,7 @@ func (c *tsoServiceDiscovery) updateMember() error {
 	oldGroupID := c.GetKeyspaceGroupID()
 	if oldGroupID != keyspaceGroup.Id {
 		log.Info("[tso] the keyspace group changed",
-			zap.Uint32("keyspace-id", keyspaceGroup.Id),
+			zap.Uint32("keyspace-id", keyspaceID),
```
HuSharp, line 468: Is it necessary to use "keyspace-id-in-request" as in the code above?

rleungx: It's unnecessary because it might printed after finishing the request.

HuSharp (approved):

Issue #8100, opened 2024-04-19
server/api: add the api the show the regions in subtree by type


nolouch:
## Development Task
help to debug and check the information.

PR #8101, opened 2024-04-19
server/api: add the api to show the regions in subtree by type

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/8100

```commit-message
server/api: add the api to show the regions in subtree by type
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/core/region.go
```diff
@@ -1368,6 +1368,57 @@ func (r *RegionsInfo) GetStoreRegions(storeID uint64) []*RegionInfo {
...
+			regions = witnesses.scanRanges()
+		}
+	case PendingPeerInSubTree:
+		if pendingPeers, ok := r.pendingPeers[storeID]; ok {
+			regions = pendingPeers.scanRanges()
+		}
+	case AllInSubTree:
+		r.st.RUnlock()
+		return r.GetStoreRegions(storeID)
+	}
```
JmPotato, line 1417: Maybe add a fallback logic to avoid forgetting to update the logic here when adding more types in the future.

tools/pd-ctl/pdctl/command/region_command.go
```diff
@@ -496,6 +497,26 @@ func showRegionWithStoreCommandFunc(cmd *cobra.Command, args []string) {
...
+		typ = 3
+	case "witness":
+		typ = 4
+	case "pending":
+		typ = 5
+	default:
+		cmd.Println("unknown type")
+		return
+	}
+	prefix += "?type=" + strconv.Itoa(typ)
```
rleungx, line 519: Will string be more readable than int?

rleungx (approved):

JmPotato (approved):

Issue #8047, opened 2024-04-10
Introduce the backoff mechanism for gRPC client


CabinfeverB:
## Development Task

- Refactor `tsoClient` to make adopting the later changes easier.
	- [x] #8121
	- [x] #8134
- Introduce the backoff mechanism into the critical paths.
	- [X] #8123
	- [ ] `GetTS` and `GetLocalTS`.
	- [ ] Member update.
- Smooth the TSO internal dispatch to reduce the failure rate.
	- [x] #7997

PR #8121, opened 2024-04-23
client/tso: organize the methods of TSO dispatcher

JmPotato:

Issue Number: ref #8047.

```commit-message
Organize the methods of the TSO dispatcher.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/tso_dispatcher.go
+++ b/client/tso_dispatcher.go
@@ -22,35 +22,15 @@ import (
 	"sync"
 	"time"
 
-	"github.com/opentracing/opentracing-go"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/client/errs"
-	"github.com/tikv/pd/client/grpcutil"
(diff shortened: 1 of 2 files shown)


client/tso_dispatcher.go
```diff
@@ -115,6 +86,17 @@ func (c *tsoClient) dispatchRequest(request *tsoRequest) (bool, error) {
 	return false, nil
 }
 
+func (c *tsoClient) closeTSODispatcher() {
```
rleungx, line 89: Shall we move the tsoClient-related function to tso_client.go?

JmPotato: Currently, because all data structures are inside `tsoClient`, it is impossible to strictly distinguish them according to files. The organization logic here is to put as much dispatcher-related code as possible in `tso_dispatcher.go`, and then move some common methods to `tso_client.go`. Perhaps we can further abstract the data structure in the future and put all dispatcher-related logic into a separate structure.

rleungx (approved):

Issue #8135, opened 2024-04-29
Make simulator Great Again!


HuSharp:
## Development Task

- Currently, we have many components to simulate the interaction with pd, `pd-api-bench`, `pd-tso-bench`, `pd-heartbeat-bench`, `pd-simulator`. We can integrate them step by step, with lower learning costs and easier interaction testing of various pd's.
- After that we can provide pd the same specification machine to deploy the real pd under the big cluster, simulate the request number of TiDB/TiKV + TiKV heartbeat heartbeat report + all kinds of scheduling for the region and so on.

### mock TiKV data to archive the following goals:
- can balance region...
- can support three forms node: meta-node/simulator(simple raft)/real-tikv
- support unified interface to interact with real-tikv

### mock TiDB&TiKV API

existed tool usages like

`/bin/pd-api-bench -http-cases GetRegionStatus-1+1,GetMinResolvedTS-1+1 -client 1 -debug true`

###	mock TiKV heartbeat/TSO

existed tool usages like
`./bin/pd-heartbeat-bench --config=./tools/pd-heartbeat-bench/config-template.toml --pd-endpoints http://127.0.0.1:2379`

### Others

- mock single PD-server
- use tiup

PR #8140, opened 2024-04-30
tools: support pd dev

HuSharp:

Issue Number: Ref #8135

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


nolouch:
I think for those tools, perhaps we can learn something from `[Carge Custom Commands]`(https://doc.rust-lang.org/cargo/reference/external-tools.html#custom-subcommands).

PR #8153, opened 2024-05-08
api: remove set node replica check

rleungx:

Issue Number: Ref #8154

```commit-message
```

### Check List

Tests

- Manual test

```
$ bin/pd-ctl -u http://127.0.0.1:32975 -i
» keyspace-group set-node 0 http://127.0.0.1:34019
Success! null
```

### Release note

```release-note
None.
```

### DIFF
--- a/server/apiv2/handlers/tso_keyspace_group.go
+++ b/server/apiv2/handlers/tso_keyspace_group.go
@@ -453,11 +453,6 @@ func SetNodesForKeyspaceGroup(c *gin.Context) {
 		c.AbortWithStatusJSON(http.StatusBadRequest, "keyspace group does not exist")
 		return
 	}
-	// check if nodes is less than default replica count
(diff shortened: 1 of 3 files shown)


server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -453,11 +453,6 @@ func SetNodesForKeyspaceGroup(c *gin.Context) {
 		c.AbortWithStatusJSON(http.StatusBadRequest, "keyspace group does not exist")
 		return
 	}
-	// check if nodes is less than default replica count
-	if len(setParams.Nodes) < utils.DefaultKeyspaceGroupReplicaCount {
```
lhy1024, line 457: Do we need to check whether it is zero?

rleungx: I think no need

lhy1024 (approved):

JmPotato (approved):

PR #8158, opened 2024-05-09
*: use a separate runner for updating subtree

rleungx:

Issue Number: Ref #7897 

```commit-message
```

### Check List

Tests

- Manual test 

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/context.go
+++ b/pkg/core/context.go
@@ -23,19 +23,21 @@ import (
 // MetaProcessContext is a context for meta process.
 type MetaProcessContext struct {
 	context.Context
-	Tracer     RegionHeartbeatProcessTracer
-	TaskRunner ratelimit.Runner
-	LogRunner  ratelimit.Runner
+	Tracer           RegionHeartbeatProcessTracer
+	TaskRunner       ratelimit.Runner
+	StatisticsRunner ratelimit.Runner
+	LogRunner        ratelimit.Runner
 }
 
 // NewMetaProcessContext creates a new MetaProcessContext.
 // used in tests, can be changed if no need to test concurrency.
 func ContextTODO() *MetaProcessContext {
 	return &MetaProcessContext{
-		Context:    context.TODO(),
-		Tracer:     NewNoopHeartbeatProcessTracer(),
-		TaskRunner: ratelimit.NewSyncRunner(),
-		LogRunner:  ratelimit.NewSyncRunner(),
+		Context:          context.TODO(),
+		Tracer:           NewNoopHeartbeatProcessTracer(),
+		TaskRunner:       ratelimit.NewSyncRunner(),
+		StatisticsRunner: ratelimit.NewSyncRunner(),
+		LogRunner:        ratelimit.NewSyncRunner(),
 		// Limit default is nil
 	}
 }
--- a/pkg/core/context.go
+++ b/pkg/core/context.go
@@ -23,21 +23,21 @@ import (
 // MetaProcessContext is a context for meta process.
 type MetaProcessContext struct {
 	context.Context
-	Tracer           RegionHeartbeatProcessTracer
-	TaskRunner       ratelimit.Runner
-	StatisticsRunner ratelimit.Runner
-	LogRunner        ratelimit.Runner
+	Tracer     RegionHeartbeatProcessTracer
+	TaskRunner ratelimit.Runner
+	MiscRunner ratelimit.Runner
+	LogRunner  ratelimit.Runner
 }
 
 // NewMetaProcessContext creates a new MetaProcessContext.
 // used in tests, can be changed if no need to test concurrency.
 func ContextTODO() *MetaProcessContext {
 	return &MetaProcessContext{
-		Context:          context.TODO(),
-		Tracer:           NewNoopHeartbeatProcessTracer(),
-		TaskRunner:       ratelimit.NewSyncRunner(),
-		StatisticsRunner: ratelimit.NewSyncRunner(),
-		LogRunner:        ratelimit.NewSyncRunner(),
+		Context:    context.TODO(),
+		Tracer:     NewNoopHeartbeatProcessTracer(),
+		TaskRunner: ratelimit.NewSyncRunner(),
+		MiscRunner: ratelimit.NewSyncRunner(),
+		LogRunner:  ratelimit.NewSyncRunner(),
 		// Limit default is nil
 	}
 }
--- a/pkg/mcs/scheduling/server/cluster.go
+++ b/pkg/mcs/scheduling/server/cluster.go
@@ -569,6 +569,7 @@ func (c *Cluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 	taskRunner, statisticsRunner, logRunner = syncRunner, syncRunner, syncRunner
 	if c.persistConfig.GetScheduleConfig().EnableHeartbeatConcurrentRunner {
 		taskRunner = c.heartbeatRunner
+		statisticsRunner = c.statisticsRunner
 		logRunner = c.logRunner
 	}
 	ctx := &core.MetaProcessContext{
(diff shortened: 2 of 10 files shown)


pkg/ratelimit/runner.go
```diff
@@ -76,8 +76,8 @@ func NewConcurrentRunner(name string, limiter *ConcurrencyLimiter, maxPendingDur
 		limiter:            limiter,
 		maxPendingDuration: maxPendingDuration,
 		taskChan:           make(chan *Task),
-		pendingTasks:       make([]*Task, 0, initialCapacity),
-		failedTaskCount:    RunnerTaskFailedTasks.WithLabelValues(name),
+		pendingNormalTasks: make([]*Task, 0, initialCapacity),
+		pendingMetaTasks:   make([]*Task, 0, initialCapacity),
```
nolouch, line 80:
```suggestion
		pendingHighPriorityTasks:   make([]*Task, 0, initialCapacity),
```

pkg/ratelimit/metrics.go
```diff
@@ -31,25 +31,41 @@ var (
...
 )
 
 func init() {
 	prometheus.MustRegister(RunnerTaskMaxWaitingDuration)
-	prometheus.MustRegister(RunnerTaskPendingTasks)
-	prometheus.MustRegister(RunnerTaskFailedTasks)
+	prometheus.MustRegister(RunnerPendingTasks)
+	prometheus.MustRegister(RunnerFailedTasks)
+	prometheus.MustRegister(RunnerTaskExecutionDuration)
+	prometheus.MustRegister(RunnerSucceededTasks)
```
JmPotato, line 70: Should we also add metrics to tell the task number of different priorities?

pkg/ratelimit/runner.go
```diff
@@ -144,16 +158,25 @@ func (cr *ConcurrentRunner) run(ctx context.Context, task func(context.Context),
 func (cr *ConcurrentRunner) processPendingTasks() {
 	cr.pendingMu.Lock()
 	defer cr.pendingMu.Unlock()
-	for len(cr.pendingTasks) > 0 {
-		task := cr.pendingTasks[0]
+	if len(cr.pendingHighPriorityTasks) > 0 {
+		task := cr.pendingHighPriorityTasks[0]
+		select {
+		case cr.taskChan <- task:
```
nolouch, line 164: Here, I have some confusion. Even though we use two queues to distinguish priority, we dispatch the tasks one by one, and they run concurrently. It is not very relevant to the priority because we cannot control the priority during the runtime process. The main difference lies in the elimination strategy of pending tasks. So, if we split the runner and let each runner have its own elimination strategy for pending tasks, would it be clearer?

rleungx: What do you mean about "control the priority during the runtime process"?

nolouch: like unified-pool.

pkg/ratelimit/runner.go
```diff
@@ -133,8 +142,13 @@ func (cr *ConcurrentRunner) Start() {
 	}()
 }
 
-func (cr *ConcurrentRunner) run(ctx context.Context, task func(context.Context), token *TaskToken) {
-	task(ctx)
+func (cr *ConcurrentRunner) run(task *Task, token *TaskToken) {
+	start := time.Now()
+	defer func() {
```
nolouch, line 147: no need defer

pkg/ratelimit/runner.go
```diff
@@ -165,34 +188,48 @@ func (cr *ConcurrentRunner) Stop() {
...
+		// We use the max task number to prevent the OOM issue.
+		// It occupies around 1.5GB memory when there is 20000000 pending task.
+		if len(cr.pendingHighPriorityTasks) > maxHighPriorityTaskNum {
+			RunnerFailedTasks.WithLabelValues(cr.name, task.name).Inc()
+			return ErrMaxWaitingTasksExceeded
 		}
 		task.submittedAt = time.Now()
-		cr.pendingTasks = append(cr.pendingTasks, task)
-		cr.pendingTaskCount[task.name]++
+		cr.pendingHighPriorityTasks = append(cr.pendingHighPriorityTasks, task)
```
nolouch, line 212: if only one task is submitted, will it always be in the pending queue because you have not sent it to the channel?

pkg/ratelimit/runner.go
rleungx: > Will it lead to normal priority task accumulation 
No, the normal priority task will be restricted by the maximum wait time.

> I think splitting it into two runners and isolating the light task and heavy task is more reasonable.
Split into two runners will also compete for the same resource which may slow down the speed of executing the high priority tasks

nolouch: > Split into two runners will also compete for the same resource which may slow down the speed of executing the high-priority tasks.

why?

rleungx: Even though the task is low priority, it might also acquire the lock.

nolouch: they only share the root tree lock for get region sometime, I think impact won't be significant.

rleungx: The reason why we need to distinguish the priority is that we want these high-priority tasks can be executed ASAP.

rleungx: Split runners won't take effect.

JmPotato: Considering that the current implementation method is to process normal tasks only after all high-priority tasks are completed, will this lead to starvation? Do we need to change it to a probabilistic selection method based on task priority before each execution?

server/cluster/cluster.go
```diff
@@ -1121,7 +1122,7 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
 	tracer.OnCollectRegionStatsFinished()
 	if c.storage != nil {
 		if saveKV {
-			ctx.TaskRunner.RunTask(
+			ctx.StatisticsRunner.RunTask(
```
rleungx, line 1125: `SaveRegionToKV` is put into the statistics runner, the runner name might be strange. But I cannot come up with a better name.

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -560,17 +565,18 @@ func (c *Cluster) HandleRegionHeartbeat(region *core.RegionInfo) error {
 	if c.persistConfig.GetScheduleConfig().EnableHeartbeatBreakdownMetrics {
 		tracer = core.NewHeartbeatProcessTracer()
 	}
-	var taskRunner, logRunner ratelimit.Runner
-	taskRunner, logRunner = syncRunner, syncRunner
+	var taskRunner, statisticsRunner, logRunner ratelimit.Runner
+	taskRunner, statisticsRunner, logRunner = syncRunner, syncRunner, syncRunner
 	if c.persistConfig.GetScheduleConfig().EnableHeartbeatConcurrentRunner {
-		taskRunner = c.heartbeatRunnner
+		taskRunner = c.heartbeatRunner
```
nolouch, line 571: need to add in here?

```
statisticsRunner = c.statisticsRunner
```

nolouch (commented):
rest lgtm

pkg/mcs/scheduling/server/cluster.go
rleungx: nice catch

nolouch (approved):

pkg/core/context.go
```diff
@@ -23,19 +23,21 @@ import (
 // MetaProcessContext is a context for meta process.
 type MetaProcessContext struct {
 	context.Context
-	Tracer     RegionHeartbeatProcessTracer
-	TaskRunner ratelimit.Runner
-	LogRunner  ratelimit.Runner
+	Tracer           RegionHeartbeatProcessTracer
+	TaskRunner       ratelimit.Runner
+	StatisticsRunner ratelimit.Runner
+	LogRunner        ratelimit.Runner
```
JmPotato, line 29: Could you please help add some comments here to explain the different responsibilities of these runners?

pkg/ratelimit/runner.go
```diff
@@ -48,9 +51,9 @@ type Runner interface {
 type Task struct {
 	ctx         context.Context
 	submittedAt time.Time
-	opts        *TaskOpts
 	f           func(context.Context)
 	name        string
+	retained    bool
```
JmPotato, line 56: This naming is vague without looking at the actual logic, maybe a line of comment would be helpful.

server/cluster/cluster.go
JmPotato: How about just calling it `MiscRunner`?

JmPotato (approved):

PR #8186, opened 2024-05-17
tools/ut: add a parallel parameter

HuSharp:

can execute
```bash
./bin/pd-ut run --race tests --parallel 100
```

Issue Number: Ref #7969

```commit-message
add parallel parameter
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


nolouch (approved):

tools/pd-ut/ut.go
```diff
@@ -694,7 +701,8 @@ func buildTestBinaryMulti(pkgs []string) error {
 		packages = append(packages, path.Join(modulePath, pkg))
 	}
 
-	p := strconv.Itoa(buildParallel)
+	// We use 2 * parallel for `go build` to make it faster.
```
okJiang, line 704: One question, why do we choose  `2 * parallel` to execute `go build`?

tools/pd-ut/ut.go
```diff
@@ -118,10 +118,17 @@ func main() {
...
-	// Get the correct count of CPU if it's in docker.
-	p = runtime.GOMAXPROCS(0)
-	// We use 2 * p for `go build` to make it faster.
-	buildParallel = p * 2
+	if parallelStr == "" {
+		// Get the correct count of CPU if it's in docker.
+		parallel = runtime.GOMAXPROCS(0)
+	} else {
+		var err error
+		parallel, err = strconv.Atoi(parallelStr)
```
okJiang, line 126: Do we need check if the `parallel<=runtime.GOMAXPROCS(0)`

tools/pd-ut/ut.go
HuSharp: background:

When building the test binaries, by default it is the number of cores([pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies](https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies)):

> -p n
the number of programs, such as build commands or
test binaries, that can be run in parallel.
The default is GOMAXPROCS, normally the number of CPUs available.

When running the 'building binaries stage' of make ut, the CPU usage is not high with the default value. So I think it's a better idea to increase the parallelism.

> why do we choose 2 * parallel to execute go build? not 1.5 or 3.
Choice 2 wasn't much thought or testing, and I think 2 is fine for the time being :) Or do you have any ideas?

tools/pd-ut/ut.go
HuSharp: There is no need, we can support parallel greater than procs. It depends on user

tools/pd-ut/ut.go
okJiang: > Or do you have any ideas?

I have no idea about that. I thought 2 had any special meaning before

tools/pd-ut/ut.go
HuSharp: make sense, and add a hint when parallel was set greater than procs :)
https://github.com/tikv/pd/pull/8186/commits/f42fc5e450d1eaa05d21d9b7cefbd291712dabd1#diff-d1de7e6511458d33a769c0c501ec0ea45494dd9d2e830dc315e8fe623eb34b76R132

okJiang (approved):

PR #8132, opened 2024-04-28
core: check stats healthy by reference

nolouch:
### What problem does this PR solve?

Issue Number: Close https://github.com/tikv/pd/issues/8131 Ref https://github.com/tikv/pd/issues/7897

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

bench config:
```
round = 100000

store-count = 400
region-count = 10000000
leader-update-ratio = 0.0
epoch-update-ratio = 0.0
space-update-ratio = 0.0
flow-update-ratio = 0.0
no-update-ratio = 0.0
```

Before 
image

After

image

### Release note

```release-note
None.
```

PR #8201, opened 2024-05-20
core/region: fix the bug causing incorrect reference count after a cache reset

JmPotato:

Issue Number: ref #7897, #8132.

```commit-message
https://github.com/tikv/pd/pull/8132 introduced a count reference to ensure the consistency between
the root tree and subtrees, however, the reference count was not handled correctly during the cache
reset process, which may lead to incorrect reference counts in later heartbeat processing.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


nolouch (approved):
lgtm

rleungx (approved):

PR #8203, opened 2024-05-21
tools/ut: add alloc server for `pd-ut`

HuSharp:

Issue Number: Ref #7969

When using larger concurrency for `pd-ut`, there are problems with `address already in use`

```bash
[PD:etcd:ErrStartEtcd]listen tcp [127.0.0.1:37561](http://127.0.0.1:37561/): bind: address already in use: listen tcp [127.0.0.1:37561](http://127.0.0.1:37561/): bind: address already in use
```

because there is a `URL alloc` for init etcd cluster config that assigns a local port number.

https://github.com/tikv/pd/blob/8e844d8483655ca3cd902220c372621c6c14cb43/pkg/utils/tempurl/tempurl.go#L32-L42

After changing to pd-ut, our tests run concurrently and are unaware of the URL used by the other tests, so there may be a conflict if there are many concurrent tests.

You can check by script below

```bash
for i in {1..200}
do
netstat -n | awk '/^tcp/ {++S[$NF]} END {for(a in S) print a, S[a]}'
sleep 1
done
```

```output
...
ESTABLISHED 205
TIME_WAIT 5905
ESTABLISHED 205
TIME_WAIT 5922
ESTABLISHED 205
TIME_WAIT 5929
...
```
So we can support a alloc server for `pd-ut` to ensure the URL is unique.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/member/member_test.go
+++ b/tests/server/member/member_test.go
@@ -289,7 +289,7 @@ func TestMoveLeader(t *testing.T) {
 	re := require.New(t)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	cluster, err := tests.NewTestCluster(ctx, 3)
+	cluster, err := tests.NewTestCluster(ctx, 5)
 	defer cluster.Destroy()
 	re.NoError(err)
 
@@ -298,7 +298,7 @@ func TestMoveLeader(t *testing.T) {
 	cluster.WaitLeader()
 
 	var wg sync.WaitGroup
-	wg.Add(3)
+	wg.Add(5)
 	for _, s := range cluster.GetServers() {
 		go func(s *tests.TestServer) {
 			defer wg.Done()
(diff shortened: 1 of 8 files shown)


tests/server/member/member_test.go
```diff
@@ -289,7 +289,7 @@ func TestMoveLeader(t *testing.T) {
 	re := require.New(t)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	cluster, err := tests.NewTestCluster(ctx, 5)
+	cluster, err := tests.NewTestCluster(ctx, 3)
```
lhy1024, line 292: why reduce it?

lhy1024 (approved):

tests/server/member/member_test.go
HuSharp: redundant code, removed.

nolouch (approved):

PR #7565, opened 2023-12-18
client/http: require source mark when initializing the http client

JmPotato:
### What problem does this PR solve?

Issue Number: ref #7300.

### What is changed and how does it work?

```commit-message
When creating a client, it is required to pass in the `source` parameter to further distinguish the source of logs and requests.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

PR #4491, opened 2021-12-22
api, pdctl: add HTTP Component signature

CabinfeverB:
### What problem does this PR solve?
close #4490  and related to #3111

This PR is going to support HTTP component signature

### What is changed and how it works?
It adds a `RoundTripper` for adding a component signature header to every request and applies it to pdctl. (other components outside PD could add header by themselves)

For the test, this PR creates a test server and registers a temp handler for `/pd/api/v1/health`(just for test) and sends the request by health pdctl cmd.

### Check List

Tests

- Unit test

### Release note

```release-note
add HTTP Component signature
```

PR #8214, opened 2024-05-24
tools/ctl: add caller ID for `pd-ctl`

HuSharp:

Issue Number: Ref #7300

After replacing HTTP with SDK #7490, we need to add the `calledID` explicitly to keep consistent with #4491

And source ID can be consistent with caller ID, #7565 

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/tests/global_test.go
+++ b/tools/pd-ctl/tests/global_test.go
@@ -16,10 +16,12 @@ package tests
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"testing"
 
+	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	"github.com/stretchr/testify/require"
 	"github.com/tikv/pd/pkg/utils/apiutil"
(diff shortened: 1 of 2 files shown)


HuSharp:
@okJiang PTAL, thx!

tools/pd-ctl/tests/global_test.go
```diff
@@ -16,33 +16,46 @@ package tests
...
 	"go.uber.org/zap"
 )
 
-const pdControlCallerID = "pd-ctl"
-
 func TestSendAndGetComponent(t *testing.T) {
 	re := require.New(t)
 	handler := func(context.Context, *server.Server) (http.Handler, apiutil.APIServiceGroup, error) {
 		mux := http.NewServeMux()
+		mux.HandleFunc("/pd/api/v1/cluster", func(w http.ResponseWriter, r *http.Request) {
```
HuSharp, line 40: For now, only cluster command is replaced by PD HTTP SDK

tools/pd-ctl/tests/global_test.go
```diff
@@ -16,33 +16,46 @@ package tests
...
+			callerID := apiutil.GetCallerIDOnHTTP(r)
+			for k := range r.Header {
+				log.Info("header", zap.String("key", k))
+			}
+			log.Info("caller id", zap.String("caller-id", callerID))
```
JmPotato, line 45:
```suggestion
```

okJiang (approved):
Please add a comment to explain why we need two handle for testing. Rest lgtm

PR #8216, opened 2024-05-27
apiutil/middleware: add retry logic for obtaining PD leader in redirector

### LINKED ISSUE #8142: add index or lightning or import into failed with error “Error 1105 (HY000): request pd http api failed with status: '503 Service Unavailable” when kill pd leader
### What did you do?
1、add index
2、kill pd leader

add index can success

### What did you see instead?
add index failed with error “Error 1105 (HY000): request pd http api failed with status: '503 Service Unavailable” 

./pd-server -V
 Release Version: v8.2.0-alpha
Edition: Community
Git Commit Hash: 1679dbca25b3483d1375c7e747da27e99ad77360
Git Branch: heads/refs/tags/v8.2.0-alpha
UTC Build Time:  2024-04-30 11:39:01

JmPotato:

Issue Number: close #8142.

```commit-message
Add retry logic to improve PD HTTP request forwarding success rate during PD leader switch.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/apiutil/serverapi/middleware.go
+++ b/pkg/utils/apiutil/serverapi/middleware.go
@@ -18,7 +18,9 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
+	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
 	"github.com/tikv/pd/pkg/errs"
 	mcsutils "github.com/tikv/pd/pkg/mcs/utils"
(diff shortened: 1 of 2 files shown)


rleungx (approved):
LGTM, but I was thinking that compared with retry in pd, will it be more reasonable to send a request to the leader directly and backoff on the client side?

JmPotato:
> LGTM, but I was thinking that compared with retry in pd, will it be more reasonable to send a request to the leader directly and backoff on the client side?

I agree with this point; client retries should be the most straightforward choice. However, considering that this PR only takes effect during the interval of PD leader re-election, a small internal retry attempt at this time would be lighter than having the client receive an error and then retry. The two are not in conflict.

pkg/utils/apiutil/serverapi/middleware.go
```diff
@@ -233,3 +234,38 @@ func (h *redirector) ServeHTTP(w http.ResponseWriter, r *http.Request, next http
 	client := h.s.GetHTTPClient()
 	apiutil.NewCustomReverseProxies(client, urls).ServeHTTP(w, r)
 }
+
+const (
+	backoffMaxDelay = 3 * time.Second
```
HuSharp, line 239: LGTM! a nitpick :) 

Do we need to consider changing this time if the client sets a timeout for the request?
such as `&http.Client{Timeout: 2 * time.Second}`

HuSharp (approved):

lhy1024:
Does it need to be picked to other branches?

JmPotato:
> Does it need to be picked to other branches?

Let's observe the effect of this change for a period of time before cherry-picking.

PR #8223, opened 2024-05-29
client/http: handle 503/leader/primary change to schedule member check

JmPotato:

Issue Number: ref #8142, ref #7300.

```commit-message
- Schedule a member change check when the HTTP status code is 503 or receives a leader/primary change error.
- Only return the latest error in backoffer to ensure the logic of successful retry.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/client.go
+++ b/client/http/client.go
@@ -120,10 +120,23 @@ func (ci *clientInner) requestWithRetry(
 	headerOpts ...HeaderOption,
 ) error {
 	var (
+		serverURL  string
+		isLeader   bool
 		statusCode int
 		err        error
+		logFields  = append(reqInfo.logFields(),
+			zap.String("source", ci.source),
+			zap.String("server-url", serverURL),
+			zap.Bool("is-leader", isLeader),
+			zap.Int("status-code", statusCode),
+			zap.Error(err))
 	)
 	execFunc := func() error {
+		defer func() {
+			// Handle some special status codes to increase the success rate of the following requests.
+			ci.handleHTTPStatusCode(statusCode)
+			log.Debug("[pd] http request finished", logFields...)
+		}()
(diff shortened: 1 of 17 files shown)


client/http/client.go
```diff
@@ -120,28 +120,41 @@ func (ci *clientInner) requestWithRetry(
...
 				continue
 			}
-			statusCode, err = ci.doRequest(ctx, url, reqInfo, headerOpts...)
+			statusCode, err = ci.doRequest(ctx, serverURL, reqInfo, headerOpts...)
 			if err == nil || noNeedRetry(statusCode) {
 				return err
 			}
-			log.Debug("[pd] request url failed",
-				zap.String("source", ci.source), zap.Bool("is-leader", cli.IsConnectedToLeader()), zap.String("url", url), zap.Error(err))
+			log.Debug("[pd] http request url failed", logFields...)
```
niubell, line 157: Should we use `log.Error` instead of `log.Debug` for this http failed log?

JmPotato: I prefer not to since the error will be returned to the caller in most cases. Additionally, outputting this internal log could be overwhelming if the QPS is high.

okJiang (approved):

client/http/client.go
```diff
@@ -120,28 +120,41 @@ func (ci *clientInner) requestWithRetry(
...
 ) error {
 	var (
+		serverURL  string
+		isLeader   bool
 		statusCode int
 		err        error
+		logFields  = append(reqInfo.logFields(),
+			zap.String("source", ci.source),
+			zap.String("server-url", serverURL),
+			zap.Bool("is-leader", isLeader),
```
HuSharp, line 130: The variable will always be the initial status.

PR #8227, opened 2024-05-30
client/retry: only return the latest error in backoffer

### LINKED ISSUE #8142: add index or lightning or import into failed with error “Error 1105 (HY000): request pd http api failed with status: '503 Service Unavailable” when kill pd leader

JmPotato:

Issue Number: ref #8142.

```commit-message
Due to the return of historical errors causing the client's retry logic to fail,
and since we currently do not need to obtain all errors during retries, this PR
removes `multierr` from backoffer and add tests to ensure the correctness of the retry logic.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix the bug where HTTP client retry logic is ineffective.
```


HuSharp (approved):

client/retry/backoff.go
```diff
@@ -148,18 +137,25 @@ func InitialBackoffer(base, max, total time.Duration, opts ...Option) *Backoffer
 	return bo
 }
 
-// SetRetryableChecker sets the retryable checker.
-func (bo *Backoffer) SetRetryableChecker(checker func(err error) bool) {
+// SetRetryableChecker sets the retryable checker, `overwrite` flag is used to indicate whether to overwrite the existing checker.
+func (bo *Backoffer) SetRetryableChecker(checker func(err error) bool, overwrite bool) {
```
okJiang, line 141: Isn't `overwrite` over-designed? Because now there is only one call to

client/retry/backoff.go
```diff
@@ -69,18 +66,13 @@ func (bo *Backoffer) Exec(
...
 	)
 	fnName := getFunctionName(fn)
 	for {
 		err = fn()
 		bo.attempt++
-		if bo.attempt < maxRecordErrorCount {
-			// multierr.Append will ignore nil error.
-			allErrors = multierr.Append(allErrors, err)
-		}
 		if !bo.isRetryable(err) {
```
okJiang, line 76: I think don't need to put `err != nil` into retryable checker function. Just put it before calling `isRetryable` is enough. Each error need to be check `err != nil`.

okJiang (approved):

rleungx (approved):

PR #8234, opened 2024-05-31
pkg: remove old duplicated task

rleungx:

Issue Number: ref #7897.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/core/region.go
```diff
@@ -754,18 +753,18 @@ func GenerateRegionGuideFunc(enableLog bool) RegionGuideFunc {
 		if logRunner != nil {
 			debug = func(msg string, fields ...zap.Field) {
 				logRunner.RunTask(
-					ctx.Context,
+					region.GetID(),
```
nolouch, line 756: It's better not to pass exclusive attributes to the runner, other types of tasks can also use it.

pkg/ratelimit/runner.go
```diff
@@ -42,16 +42,16 @@ const (
 
 // Runner is the interface for running tasks.
 type Runner interface {
-	RunTask(ctx context.Context, name string, f func(context.Context), opts ...TaskOption) error
+	RunTask(id, name string, f func(), opts ...TaskOption) error
```
nolouch, line 45: I think `RunTask(id uint64, name string, f func(), opts ...TaskOption) error` is ok. `fmt.Sprint` has an allocation cost.

rleungx: But the same id may have different tasks. If we don't use fmt.Sprint, then we need a map[uint64]map[string]struct.

pkg/ratelimit/runner.go
```diff
@@ -188,22 +193,25 @@ func (cr *ConcurrentRunner) RunTask(ctx context.Context, name string, f func(con
 
 	pendingTaskNum := len(cr.pendingTasks)
 	if pendingTaskNum > 0 {
+		if t, ok := cr.pendingRegionTasks[task.id]; ok {
+			t.f = f
```
okJiang, line 197: Can you add a comment to explain the function will be overwrite and it doesn't matter

rest lgtm

pkg/ratelimit/runner.go
okJiang: If the key is different, the bench will be different

nolouch (approved):

okJiang (approved):

pkg/ratelimit/runner.go
```diff
@@ -42,16 +42,16 @@ const (
...
-	RunTask(ctx context.Context, name string, f func(context.Context), opts ...TaskOption) error
+	RunTask(id uint64, name string, f func(), opts ...TaskOption) error
 	Start()
 	Stop()
 }
 
 // Task is a task to be run.
 type Task struct {
-	ctx         context.Context
+	id          uint64
```
okJiang, line 52: can use taskID directly

rleungx: I think it's ok and use less memory.

pkg/ratelimit/runner.go
```diff
@@ -61,16 +61,23 @@ type Task struct {
 var ErrMaxWaitingTasksExceeded = errors.New("max waiting tasks exceeded")
 
 // ConcurrentRunner is a simple task runner that limits the number of concurrent tasks.
```
JmPotato, line 63: It's a mismatched comment line.

JmPotato (approved):

PR #8259, opened 2024-06-05
scheduler: add batch config for evict leader scheduler

### LINKED ISSUE #8265: Support customizing the batch size for evict leader scheduler
## Enhancement Task
The default value is 3 which is slow in a large cluster. We need to support changing the default value to accelerate the speed of evicting leader.

rleungx:

Issue Number: Close #8265.

Refactor the evict leader lock and support setting the batch config.

```commit-message
```

### Check List

Tests

- Unit test

Testing using batch size 10 and default 3 (leader-schedule-limit is set to 2048)
Screenshot 2024-06-07 at 16 03 43
Screenshot 2024-06-07 at 16 03 50
Screenshot 2024-06-07 at 16 03 56
Screenshot 2024-06-07 at 16 04 09

### Release note

```release-note
None.
```


pkg/schedule/schedulers/evict_leader.go
```diff
@@ -163,6 +153,109 @@ func (conf *evictLeaderSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRa
...
+
+func (conf *evictLeaderSchedulerConfig) delete(id uint64) (any, error) {
+	conf.Lock()
+	keyRanges := conf.StoreIDWithRanges[id]
+	succ, last := conf.removeStoreLocked(id)
+	var resp any
+
+	if !succ {
+		conf.Unlock()
+		return resp, errs.ErrScheduleConfigNotExist.FastGenByArgs()
```
okJiang, line 235: Is it better to put this error into `removeStoreLocked` instead with returning `succ`?

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -163,6 +153,109 @@ func (conf *evictLeaderSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRa
...
+func (conf *evictLeaderSchedulerConfig) update(id uint64, newRanges []core.KeyRange, batch int) error {
+	conf.Lock()
+	defer conf.Unlock()
+	if id != 0 {
+		conf.StoreIDWithRanges[id] = newRanges
+	}
+	conf.Batch = batch
+	err := conf.persistLocked()
+	if err != nil {
+		conf.removeStoreLocked(id)
```
okJiang, line 222: Do we need to check `id != 0`?

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -163,6 +153,109 @@ func (conf *evictLeaderSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRa
...
+		conf.resetStoreLocked(id, keyRanges)
+		conf.Unlock()
+		return resp, err
+	}
+	if !last {
+		conf.Unlock()
+		return resp, nil
+	}
+	conf.Unlock()
+	if err := conf.removeSchedulerCb(EvictLeaderName); err != nil {
```
okJiang, line 249: Is it better to `removeSchedulerCb` inside the `removeStoreLocked`? Is it a heavy operation that you don't want to put it into a xxxxLocked function? (But I think updateConfig/deleteConfig are low frequency operations 🤔)

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -360,35 +426,50 @@ func (handler *evictLeaderHandler) UpdateConfig(w http.ResponseWriter, r *http.R
...
-	var id uint64
-	idFloat, ok := input["store_id"].(float64)
-	if ok {
+	var (
+		exist     bool
+		err       error
+		id        uint64
+		newRanges []core.KeyRange
+	)
+	idFloat, ok1 := input["store_id"].(float64)
```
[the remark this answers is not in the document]
rleungx: It's just a bool to indicate if there is a value.

pkg/schedule/schedulers/evict_leader.go
rleungx: No, there is a lock problem, if we put it into removeStoreLocked, it might have a deadlock issue.

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -360,35 +426,50 @@ func (handler *evictLeaderHandler) UpdateConfig(w http.ResponseWriter, r *http.R
...
+		if err != nil {
+			handler.rd.JSON(w, http.StatusInternalServerError, err.Error())
+			return
 		}
-		handler.config.RUnlock()
-		args = append(args, strconv.FormatUint(id, 10))
+	}
+
+	batch := handler.config.getBatch()
+	batchFloat, ok := input["batch"].(float64)
```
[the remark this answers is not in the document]
rleungx: To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:

- bool, for JSON booleans
- float64, for JSON numbers
- string, for JSON strings
- []interface{}, for JSON arrays
- map[string]interface{}, for JSON objects
- nil for JSON null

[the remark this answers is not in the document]
okJiang: Thank you, I understand.

```
func TestUnmarshal(t *testing.T) {
	str := `{
	"id": 1
}`
	b := []byte(str)
	var data any
	err := json.Unmarshal(b, &data)
	require.NoError(t, err)

	id, ok := data.(map[string]interface{})["id"].(float64)
	require.True(t, ok)
	require.Equal(t, float64(1), id)

	_, ok = data.(map[string]interface{})["id"].(int)
	require.False(t, ok)
}

```

pkg/schedule/schedulers/evict_leader.go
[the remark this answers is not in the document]
okJiang: Yes, `userInputId` is just a variable to show that "there is a store_id value"

pkg/schedule/schedulers/evict_leader.go
okJiang: Sure, you are right. `removeSchedulerCb` also acquires lock resources internally. My idea is that as long as `removeStoreLocked` is successful and all deletions are completed (last=true), then `removeSchedulerCb` should be executed. However, the logic here seems a bit confusing, making it difficult to modify. It may be necessary to reorganize and refactor this part. But it's not related to this PR.

okJiang (approved):
LGTM

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -166,6 +154,108 @@ func (conf *evictLeaderSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRa
...
+	err := conf.persistLocked()
+	if err != nil && id != 0 {
+		_, _ = conf.removeStoreLocked(id)
+	}
+	return err
+}
+
+func (conf *evictLeaderSchedulerConfig) delete(id uint64) (any, error) {
+	conf.Lock()
+	keyRanges := conf.StoreIDWithRanges[id]
```
lhy1024, line 230: Do we need to check whether `keyRanges` is nil?

HuSharp: maybe we can move `removeStoreLocked`  before `keyRanges := conf.StoreIDWithRanges[id]` to help check config existence?

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -166,6 +154,108 @@ func (conf *evictLeaderSchedulerConfig) getKeyRangesByID(id uint64) []core.KeyRa
...
+func (conf *evictLeaderSchedulerConfig) update(id uint64, newRanges []core.KeyRange, batch int) error {
+	conf.Lock()
+	defer conf.Unlock()
+	if id != 0 {
+		conf.StoreIDWithRanges[id] = newRanges
+	}
+	conf.Batch = batch
+	err := conf.persistLocked()
+	if err != nil && id != 0 {
+		_, _ = conf.removeStoreLocked(id)
```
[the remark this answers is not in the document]
rleungx: I think no need

lhy1024 (approved):

nolouch (approved):
lgtm

PR #8262, opened 2024-06-06
api: supports `/regions/key` by hex key

### LINKED ISSUE #8261: supports `/regions/key` by hex key
## Enhancement Task

sometimes, the user uses the API `/regions/key` without pd-ctl. it's not useful because the key should specified with the raw key.

HuSharp:

Issue Number: Close #8261

```commit-message
api: supports `/regions/key` by hex key
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


HuSharp:
PTAL, thx! @yongman

server/api/region.go
```diff
@@ -176,6 +176,25 @@ func (h *regionsHandler) ScanRegions(w http.ResponseWriter, r *http.Request) {
...
+		}
+		startKey = string(keyBytes)
+
+		keyBytes, err = hex.DecodeString(endKey)
+		if err != nil {
+			h.rd.JSON(w, http.StatusBadRequest, err.Error())
+			return
+		}
+		endKey = string(keyBytes)
+	}
```
JmPotato, line 196: Can we extract this part into a common function since it's used multiple times in the file?

JmPotato (commented):
Can we add this new change to the PD HTTP SDK as well?

yongman (approved):

HuSharp:
> Can we add this new change to the PD HTTP SDK as well?

The `GetRegionsByKeyRange` interface of the PD HTTP SDK only supports bytes, as does `GetRegionByKey`. Do we need hex support for both interfaces?

pkg/utils/apiutil/apiutil.go
```diff
@@ -336,6 +336,22 @@ func ParseKey(name string, input map[string]any) ([]byte, string, error) {
 	return returned, rawKey, nil
 }
 
+// ParseHexKeys decodes hexadecimal keys to bytes if the format is "hex".
+func ParseHexKeys(format string, keys ...*string) error {
```
okJiang, line 340: Returning parsed key to user is better than returning result by input parameter implicitly.
```suggestion
func ParseHexKeys(format string, keys []string) ([]string, error) {
```

okJiang, line 340: Can add some unit tests for this function

okJiang (commented):
rest lgtm

pkg/utils/apiutil/apiutil.go
HuSharp: want to reduce redundant code :(

okJiang: This suggestion aim to avoid to changing data in the input implicitly, and each update operation, I think, should re-assign to the old value. Let's see what other reviewers think.

HuSharp: Perhaps there is no need to provide a function. Because of the multiple arguments and the error check which will make code more redundant.

HuSharp: cc @JmPotato

JmPotato: This involves converting between `[]byte` and `string`. I prefer a more performant approach; a benchmark may help.

HuSharp: added

pkg/utils/apiutil/apiutil.go
```diff
@@ -336,6 +336,22 @@ func ParseKey(name string, input map[string]any) ([]byte, string, error) {
 	return returned, rawKey, nil
 }
 
+// ParseHexKeys decodes hexadecimal keys to bytes if the format is "hex".
+func ParseHexKeys(format string, keys []string) (hexStrings []string, err error) {
```
okJiang, line 340:
```suggestion
func ParseHexKeys(format string, keys []string) (hexBytes [][]byte, err error) {
```

returning bytes can reduce some converting

okJiang (approved):

nolouch (approved):

pkg/utils/apiutil/apiutil.go
```diff
@@ -336,6 +336,22 @@ func ParseKey(name string, input map[string]any) ([]byte, string, error) {
...
+	if format != "hex" {
+		return keys, nil
+	}
```
rleungx, line 343: Shall we move it out of this function?

pkg/utils/apiutil/apiutil.go
HuSharp: changed in https://github.com/tikv/pd/pull/8262/commits/3f515a6d39dad824fb9499492988c6dd7af65dd4

JmPotato (approved):

pkg/utils/apiutil/apiutil.go
rleungx: Since the function is named ParseHexKeys, I think it has already be in hex format before calling this function.

PR #8266, opened 2024-06-06
grpc_service: add `ManuallyChangePeer` and `ManuallyChangeMultiPeers`.

LykxSassinator:

Issue Number: Ref https://github.com/tikv/tikv/issues/15292 proto: https://github.com/pingcap/kvproto/pull/1250

```commit-message
Make `grpc_service` support manually add / remove peers by `ManuallyChangePeer` and `ManuallyChangeMultiPeers`.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Add extra apis `ManuallyChangePeer` and `ManuallyChangeMultiPeers` for `grpc_service` to support manually add / remove peers.
```


client/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/tikv/pd/client
 
 go 1.21
 
+replace github.com/pingcap/kvproto => github.com/LykxSassinator/kvproto v0.0.0-20240605160713-4dfe4594d6ca
```
LykxSassinator, line 5: Will be cleared after https://github.com/pingcap/kvproto/pull/1250 is merged.

go.mod
```diff
@@ -6,6 +6,7 @@ go 1.21
 // kvproto at the same time. You can run `go mod tidy` to make it replaced with go-mod style specification.
 // After the PR to kvproto is merged, remember to comment this out and run `go mod tidy`.
 // replace github.com/pingcap/kvproto => github.com/$YourPrivateRepo $YourPrivateBranch
+replace github.com/pingcap/kvproto => github.com/LykxSassinator/kvproto v0.0.0-20240605160713-4dfe4594d6ca
```
LykxSassinator, line 9: ditto.

tools/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/tikv/pd/tools
 go 1.21
 
 replace (
+	github.com/pingcap/kvproto => github.com/LykxSassinator/kvproto v0.0.0-20240605160713-4dfe4594d6ca
```
LykxSassinator, line 6: ditto.

server/grpc_service.go
```diff
@@ -3000,6 +3001,78 @@ func (s *GrpcServer) GetExternalTimestamp(ctx context.Context, request *pdpb.Get
 	}, nil
 }
 
+// ManuallyChangePeer implements gRPC PDServer.
+func (s *GrpcServer) ManuallyChangePeer(ctx context.Context, request *pdpb.ChangePeerRequest) (*pdpb.ChangePeerResponse, error) {
```
overvenus, line 3005: Any input about the RPC name? @nolouch @rleungx

client/client.go
```diff
@@ -133,6 +134,11 @@ type RPCClient interface {
 	// SetExternalTimestamp sets external timestamp
 	SetExternalTimestamp(ctx context.Context, timestamp uint64) error
 
+	// ManuallyChangePeer manually changes the peer of a region.
+	ManuallyChangePeer(ctx context.Context, regionID uint64, changeType eraftpb.ConfChangeType, peer *metapb.Peer) error
```
rleungx, line 138: Is it necessary? Why not let TiKV handle it? If we are going to use RPC to solve the snapshot problem, we still can not ensure that the operator is executed successfully.

server/grpc_service.go
nolouch: Can we only report the purpose then PD generate operators. like call interface as `ReportDamagedPeer .

client/client.go
LykxSassinator: It's used to make the operation on `add peer` / `remove peer` managed by PD uniformly.

If let TiKV handle it directly, the scheduling on `add peer` or `remove peer` will not be perceived by PD, and the check of the scheduler cannot be validated by the `store limit` and other checking mechanisms.

server/grpc_service.go
LykxSassinator: Do u mean using `GetDamagedRegionsId`?

Not appropriate. The `DamagedRegionsId` just records the abnormal region, and it's used to delete all of the peers of this region.

But what we need now is "add / remove" a single peer.

nolouch: I mean the TiKV tells PD which peer is damaged or needs to be cleaned. then PD will log it(better with the source) and add an operator to add/remove it.

Do not directly guide PD on how to do it, tell PD your purpose, and then PD makes the decisions. it's prevents some unknow source from abusing it.

LykxSassinator: It just needs a simple way to let PD trigger the operator on `add` or `remove` peer here. And we've already have Restful API used for `add` & `remove` peer, gRPC API is intended to be used as the same purpose here.

LykxSassinator:
This pr can be closed as this RPC is not necessary.

Issue #8267, opened 2024-06-06
rc: group change will meet data race


HuSharp:
## Bug Report

```
==================
WARNING: DATA RACE
Write at 0x00c0014f5b30 by goroutine 513:
  github.com/tikv/pd/pkg/mcs/resourcemanager/server.(*Manager).Init()
      /data/nvme0n1/husharp/proj/pd/pkg/mcs/resourcemanager/server/manager.go:132 +0x5d1
  github.com/tikv/pd/pkg/mcs/resourcemanager/server.(*Manager).Init-fm()
      <autogenerated>:1 +0x47
  github.com/tikv/pd/server.(*Server).campaignLeader()
      /data/nvme0n1/husharp/proj/pd/server/server.go:1772 +0x1742
  github.com/tikv/pd/server.(*Server).leaderLoop()
      /data/nvme0n1/husharp/proj/pd/server/server.go:1699 +0x1444
  github.com/tikv/pd/server.(*Server).startServerLoop.func1()
      /data/nvme0n1/husharp/proj/pd/server/server.go:658 +0x33

Previous read at 0x00c0014f5b30 by goroutine 29900015:
  github.com/tikv/pd/pkg/mcs/resourcemanager/server.(*Manager).backgroundMetricsFlush()
      /data/nvme0n1/husharp/proj/pd/pkg/mcs/resourcemanager/server/manager.go:480 +0x12ed
  github.com/tikv/pd/pkg/mcs/resourcemanager/server.(*Manager).Init.func4()
      /data/nvme0n1/husharp/proj/pd/pkg/mcs/resourcemanager/server/manager.go:180 +0x4f

```

### What did you do?
run pd-server

### What did you expect to see?
no data race

### What did you see instead?
meet data race

### What version of PD are you using (`pd-server -V`)?
494c0e956b622a9b3cec8b98180b942cbbe5a0f3

Issue #8263, opened 2024-06-06
schedule: `operator.check` will meet datarace


lhy1024:
### What did you do?

run a server which built with `-race`

### What did you expect to see?

no datarace

### What did you see instead?

```
WARNING: DATA RACE
Read at 0x00c059c28fc8 by goroutine 2253766:
  github.com/tikv/pd/pkg/schedule/operator.(*Operator).Check()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/operator/operator.go:382 +0x544
  github.com/tikv/pd/pkg/schedule/operator.(*Controller).pollNeedDispatchRegion()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/operator/operator_controller.go:262 +0x144
  github.com/tikv/pd/pkg/schedule/operator.(*Controller).PushOperators()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/operator/operator_controller.go:281 +0x2e
  github.com/tikv/pd/pkg/schedule.(*Coordinator).drivePushOperator()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/coordinator.go:412 +0x224
  github.com/tikv/pd/pkg/schedule.(*Coordinator).Run.func5()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/coordinator.go:491 +0x33

Previous write at 0x00c059c28fc8 by goroutine 37567:
  sync/atomic.CompareAndSwapInt64()
      /usr/local/go/src/runtime/race_amd64.s:316 +0xb
  sync/atomic.CompareAndSwapInt64()
      <autogenerated>:1 +0x1a
  github.com/tikv/pd/pkg/schedule/operator.(*Controller).Dispatch()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/pkg/schedule/operator/operator_controller.go:145 +0x104
  github.com/tikv/pd/server/cluster.(*RaftCluster).HandleRegionHeartbeat()
      /home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/pd/server/cluster/cluster_worker.go:66 +0x5c6  github.com/tikv/pd/server.(*GrpcServer).RegionHeartbeat()
[...]

PR #8270, opened 2024-06-07
tests/realcluster: using real pd with race

HuSharp:

Issue Number: Ref #7298

running with data race

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


lhy1024 (approved):

HuSharp:
Ref #8263, #8267

Issue #8281, opened 2024-06-13
Improve the observability of the TSO server handle


JmPotato:
## Enhancement Task

At present, although we have recorded metrics such as TSO server handle duration, it only includes the time taken for successful TSO requests. Therefore, it cannot reflect service anomalies, which affects troubleshooting. We should provide more to help diagnose the status of TSO service.

PR #8282, opened 2024-06-13
server, metrics: let server TSO handle duration including failed requests

JmPotato:

Issue Number: ref #8281.

```commit-message
- Delete the 99.999% percentile data because it does not correspond with TiDB and is too tail-end, which can easily mislead.
- Emphasize PD server/client in the panel title.
- Add corresponding 90/99/99.9% percentile data on the client handle duration for easier comparison.
- The PD server TSO handle duration now includes the failed requests, directly reflecting TSO HA anomalies in the monitoring data.
```

### Check List

Tests

- Unit test
- Integration test
- Mannual test

image

### Release note

```release-note
Let PD server TSO handle duration including failed requests.
```

### DIFF
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -10634,19 +10634,12 @@
               "step": 2
             },
             {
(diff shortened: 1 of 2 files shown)


rleungx (approved):

AndreMouche (approved):
LGTM

PR #8300, opened 2024-06-17
api: client and server support `BatchScanRegions`

### LINKED ISSUE #8307: Support `BatchScanRegions` which loads non-continuous regions in one RPC
## Development Task

Sometimes, the client need to locate non-continuous key ranges. Current PD has no good solution for this requirement.

There are 2 workaround:

1. Use [`GetRegion`](https://github.com/tikv/pd/blob/c1d422ec4924799a5ff7da6606644959e339f1c0/client/client.go#L81) to load region one by one, check if the ranges are drained eachtime.
2. Use [`ScanRegion`](https://github.com/tikv/pd/blob/c1d422ec4924799a5ff7da6606644959e339f1c0/client/client.go#L92) to load `(ranges[0].start, ranges[len(ranges)-1].end)`, which can cause unexpected traffics (imagine the ranges[0] is the first row and ranges[len(ranges)-1] is the last row of a large table).

Thus a interface `BatchScanRegions` is required.

It returns multi regions like `ScanRegion`, but accept a key range slice parameters. The returned regions are guaranteed to cover the given ranges if limitation is not reached.

you06:

Issue Number: close #8307 ref pingcap/tidb#53850

Add an interface `BatchScanRegions` which support scanning regions with non-continuous ranges.

```commit-message
Add `BatchScanRegions` interface for pd-client.
```

### Check List

Tests

- Unit test
- Manual test
  - When loading non-continuous regions in TiDB's coprocessor, this PR optimize the duration from 81.5ms to 2.7ms.

Code changes

Side effects

Related changes

### Release note

```release-note
support `BatchScanRegions` for pd-client.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -17,7 +17,9 @@ package pd
 import (
 	"context"
 	"crypto/tls"
+	"encoding/hex"
 	"fmt"
+	"net/url"
 	"runtime/trace"
 	"strings"
 	"sync"
@@ -85,11 +87,18 @@ type RPCClient interface {
 	GetPrevRegion(ctx context.Context, key []byte, opts ...GetRegionOption) (*Region, error)
 	// GetRegionByID gets a region and its leader Peer from PD by id.
 	GetRegionByID(ctx context.Context, regionID uint64, opts ...GetRegionOption) (*Region, error)
+	// Deprecated: use BatchScanRegions instead.
 	// ScanRegions gets a list of regions, starts from the region that contains key.
-	// Limit limits the maximum number of regions returned.
+	// Limit limits the maximum number of regions returned. It returns all the regions in the given range if limit <= 0.
 	// If a region has no leader, corresponding leader will be placed by a peer
 	// with empty value (PeerID is 0).
 	ScanRegions(ctx context.Context, key, endKey []byte, limit int, opts ...GetRegionOption) ([]*Region, error)
+	// BatchScanRegions gets a list of regions, starts from the region that contains key.
+	// Limit limits the maximum number of regions returned. It returns all the regions in the given ranges if limit <= 0.
+	// If a region has no leader, corresponding leader will be placed by a peer
+	// with empty value (PeerID is 0).
+	// The returned regions are flattened, even there are key ranges located in the same region, only one region will be returned.
+	BatchScanRegions(ctx context.Context, keyRanges []KeyRange, limit int, opts ...GetRegionOption) ([]*Region, error)
 	// GetStore gets a store from PD by store id.
 	// The store may expire later. Caller is responsible for caching and taking care
 	// of store change.
@@ -335,6 +344,38 @@ type SecurityOption struct {
 	SSLCABytes   []byte
 	SSLCertBytes []byte
 	SSLKEYBytes  []byte
+}
+
+// KeyRange defines a range of keys in bytes.
+type KeyRange struct {
+	StartKey []byte
+	EndKey   []byte
+}
+
+// NewKeyRange creates a new key range structure with the given start key and end key bytes.
+// Notice: the actual encoding of the key range is not specified here. It should be either UTF-8 or hex.
+//   - UTF-8 means the key has already been encoded into a string with UTF-8 encoding, like:
+//     []byte{52 56 54 53 54 99 54 99 54 102 50 48 53 55 54 102 55 50 54 99 54 52}, which will later be converted to "48656c6c6f20576f726c64"
+//     by using `string()` method.
+//   - Hex means the key is just a raw hex bytes without encoding to a UTF-8 string, like:
+//     []byte{72, 101, 108, 108, 111, 32, 87, 111, 114, 108, 100}, which will later be converted to "48656c6c6f20576f726c64"
+//     by using `hex.EncodeToString()` method.
+func NewKeyRange(startKey, endKey []byte) *KeyRange {
+	return &KeyRange{startKey, endKey}
+}
+
+// EscapeAsUTF8Str returns the URL escaped key strings as they are UTF-8 encoded.
+func (r *KeyRange) EscapeAsUTF8Str() (startKeyStr, endKeyStr string) {
+	startKeyStr = url.QueryEscape(string(r.StartKey))
+	endKeyStr = url.QueryEscape(string(r.EndKey))
+	return
+}
+
+// EscapeAsHexStr returns the URL escaped key strings as they are hex encoded.
+func (r *KeyRange) EscapeAsHexStr() (startKeyStr, endKeyStr string) {
+	startKeyStr = url.QueryEscape(hex.EncodeToString(r.StartKey))
+	endKeyStr = url.QueryEscape(hex.EncodeToString(r.EndKey))
+	return
 }
 
 // NewClient creates a PD client.
@@ -1094,6 +1135,7 @@ func (c *client) ScanRegions(ctx context.Context, key, endKey []byte, limit int,
 	if serviceClient == nil {
 		return nil, errs.ErrClientGetProtoClient
 	}
+	//nolint:staticcheck
 	resp, err := pdpb.NewPDClient(serviceClient.GetClientConn()).ScanRegions(cctx, req)
 	failpoint.Inject("responseNil", func() {
 		resp = nil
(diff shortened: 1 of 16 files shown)


client/client.go
```diff
@@ -337,6 +343,12 @@ type SecurityOption struct {
 	SSLKEYBytes  []byte
 }
 
+// KeyRange represents a range of keys.
+type KeyRange struct {
```
JmPotato, line 347: We have the same definition in `client/http` pkg, is it possible to reuse it?

https://github.com/tikv/pd/blob/c1d422ec4924799a5ff7da6606644959e339f1c0/client/http/types.go#L47

client/client.go
```diff
@@ -1113,6 +1126,74 @@ func (c *client) ScanRegions(ctx context.Context, key, endKey []byte, limit int,
 	return handleRegionsResponse(resp), nil
 }
 
+func (c *client) BatchScanRegions(ctx context.Context, ranges []KeyRange, limit int, opts ...GetRegionOption) ([]*Region, error) {
+	if span := opentracing.SpanFromContext(ctx); span != nil && span.Tracer() != nil {
+		span = span.Tracer().StartSpan("pdclient.ScanRegions", opentracing.ChildOf(span.Context()))
```
JmPotato, line 1131:
```suggestion
		span = span.Tracer().StartSpan("pdclient.BatchScanRegions", opentracing.ChildOf(span.Context()))
```

server/grpc_service.go
```diff
@@ -1622,11 +1628,89 @@ func (s *GrpcServer) ScanRegions(ctx context.Context, request *pdpb.ScanRegionsR
...
+	}
+	res := rc.BatchScanRegions(keyRanges, int(limit))
+	regions := make([]*pdpb.Region, 0, len(res))
+	for _, r := range res {
+		leader := r.GetLeader()
+		if leader == nil {
+			leader = &metapb.Peer{}
+		}
+		var buckets *metapb.Buckets
+		if !*followerHandle && rc.GetStoreConfig().IsEnableRegionBucket() && needBucket {
```
JmPotato, line 1696: You can extract this check outside of the for loop.

client/client.go
you06: Can I simply made the `startKey` and `endKey` of this struct public? If so, we can certainly reuse it.

you06: I moved the `KeyRange` struct from `client/http` pkg to `client` pkg(unless there will be cycling imports).
To avoid breaking compatibility, an alias is made in `client/http` pkg.

pkg/core/basic_cluster.go
```diff
@@ -140,6 +141,12 @@ type KeyRanges struct {
 	krs []*KeyRange
 }
 
+func NewKeyRangesWithSize(size int) *KeyRanges {
```
JmPotato, line 144:
```suggestion
// NewKeyRangesWithSize ...
func NewKeyRangesWithSize(size int) *KeyRanges {
```

JmPotato (approved):

client/client.go
```diff
@@ -1113,6 +1154,74 @@ func (c *client) ScanRegions(ctx context.Context, key, endKey []byte, limit int,
...
+	}
+	pbRanges := make([]*pdpb.KeyRange, 0, len(ranges))
+	for _, r := range ranges {
+		pbRanges = append(pbRanges, &pdpb.KeyRange{StartKey: r.StartKey, EndKey: r.EndKey})
+	}
+	req := &pdpb.BatchScanRegionsRequest{
+		Header:      c.requestHeader(),
+		NeedBuckets: options.needBuckets,
+		Ranges:      pbRanges,
+		Limit:       int32(limit),
```
HuSharp, line 1183: Why not use the `int` type directly?

client/client.go
```diff
@@ -90,6 +92,12 @@ type RPCClient interface {
 	// If a region has no leader, corresponding leader will be placed by a peer
 	// with empty value (PeerID is 0).
 	ScanRegions(ctx context.Context, key, endKey []byte, limit int, opts ...GetRegionOption) ([]*Region, error)
+	// BatchScanRegions gets a list of regions, starts from the region that contains key.
+	// Limit limits the maximum number of regions returned.
```
HuSharp, line 96: can you add a comment to indicate when limit <= 0 means no limit?

client/client.go
```diff
@@ -1113,6 +1154,74 @@ func (c *client) ScanRegions(ctx context.Context, key, endKey []byte, limit int,
...
+		NeedBuckets: options.needBuckets,
+		Ranges:      pbRanges,
+		Limit:       int32(limit),
+	}
+	serviceClient, cctx := c.getRegionAPIClientAndContext(scanCtx, options.allowFollowerHandle && c.option.getEnableFollowerHandle())
+	if serviceClient == nil {
+		return nil, errs.ErrClientGetProtoClient
+	}
+	resp, err := pdpb.NewPDClient(serviceClient.GetClientConn()).BatchScanRegions(cctx, req)
+	failpoint.Inject("responseNil", func() {
```
HuSharp, line 1190: Have you used this failpoint?

pkg/core/region.go
```diff
@@ -1833,6 +1833,42 @@ func (r *RegionsInfo) ScanRegions(startKey, endKey []byte, limit int) []*RegionI
...
+		if lastRegion != nil {
+			if lastRegion.Contains(keyRange.EndKey) {
+				continue
+			} else if lastRegion.Contains(keyRange.StartKey) {
+				keyRange.StartKey = lastRegion.GetEndKey()
+			}
+		}
+		if limit > 0 && len(res) >= limit {
+			return res
+		}
```
HuSharp, line 1856: Do we need to change the order?
```suggestion
		if limit > 0 && len(res) >= limit {
			return res
		}
		if lastRegion != nil {
			if lastRegion.Contains(keyRange.EndKey) {
				continue
			} else if lastRegion.Contains(keyRange.StartKey) {
				keyRange.StartKey = lastRegion.GetEndKey()
			}
		}
```

HuSharp:
And please create an issue to remove `do-not-merge/needs-linked-issue` label :)

pkg/core/region.go
you06: It's on purposem, if the last region within limitation can contains more given ranges, we should not return it earlier.

client/client.go
you06: Nope, I just copy it from `ScanRegions`, and I also can't find the usage in client-go and tidb, remove it.

pkg/core/region.go
HuSharp: I am sorry for not getting the point, why can the order changed to reject `a region within limitation`? 0 0 
- Because this change seems to just return earlier with greater than limitation condition.

Are you expecting the parameter's value to be changed in place?

you06: You're right, the ranges contained by the last region should be decided by the client.

HuSharp (approved):
waiting for https://github.com/pingcap/kvproto/pull/1251

pkg/core/region.go
```diff
@@ -1833,6 +1833,42 @@ func (r *RegionsInfo) ScanRegions(startKey, endKey []byte, limit int) []*RegionI
 	return res
 }
 
+// BatchScanRegions scans regions in given key pairs, returns at most `limit` regions.
+// limit <= 0 means no limit.
+// The given key pairs should be non-overlapping.
+func (r *RegionsInfo) BatchScanRegions(keyRanges *KeyRanges, limit int) []*RegionInfo {
+	r.t.RLock()
+	defer r.t.RUnlock()
```
rleungx, line 1841: Will it hold the lock for a long time?

you06: Searching in memory is fast anyway, and there is a limit parameter which should be set properly by client([default 128](https://github.com/tikv/client-go/blob/ac8fa1d73a0c55e8526a1ae9b88670b1b64970e6/internal/locate/region_cache.go#L78) in client-go).

rleungx: Ok, if it holds the lock for a long time, It might affect the heartbeat process.

client/client.go
```diff
@@ -1103,6 +1145,7 @@ func (c *client) ScanRegions(ctx context.Context, key, endKey []byte, limit int,
 		if protoClient == nil {
 			return nil, errs.ErrClientGetProtoClient
 		}
+		//lint:ignore staticcheck
```
HuSharp, line 1148: maybe can use `//nolint:staticcheck`

you06: let me try, the golang ci lint cannot run in my laptop.

HuSharp: I have run on my laptop, you need to change this code snippet with `nolint:staticcheck` as well :)

https://github.com/tikv/pd/blob/7e18a69734f2ae391516625ec0230c05b363f5c5/tools/pd-api-bench/cases/cases.go#L355

Issue #8305, opened 2024-06-18
Enhance log desensitization


JmPotato:
Ref #2852 and https://github.com/pingcap/tidb/issues/51306.

To better provide log desensitization functionality, we need to introduce the following enhancements and improvements for PD:

- [x] Complete the unprocessed sensitive information logs.
- [x] Provide sensitive data mark with `‹..›` like [tidb_redact_log](https://docs.pingcap.com/tidb/dev/system-variables#tidb_redact_log).
	- [x] #8306
- [ ] Update documents.
	- [ ] https://github.com/pingcap/docs-cn/pull/17835

PR #8306, opened 2024-06-18
*: add mark function for log desensitization

JmPotato:

Issue Number: ref #8305.

```commit-message
Provide sensitive log redact marker `‹..›` to customize log desensitization.
```

### Check List

Tests

- Unit test
- Integration test
- Manual test

image

### Release note

```release-note
Provide sensitive log redact marker `‹..›` to customize log desensitization.
```


okJiang (approved):

pkg/utils/configutil/configutil.go
```diff
@@ -79,8 +79,25 @@ func (m *ConfigMetaData) CheckUndecoded() error {
...
+	// for clarity. For instance, "k<ey" would be redacted as "<k<<ey>" if the mark is "<>".
+	RedactInfoMark string            `toml:"redact-info-mark" json:"redact-info-mark"`
+	Encryption     encryption.Config `toml:"encryption" json:"encryption"`
+}
+
+// Adjust adjusts the security configuration.
+func (sc *SecurityConfig) Adjust() error {
+	AdjustBool(&sc.RedactInfoLog, false)
+	markLength := len(sc.RedactInfoMark)
+	if markLength > 0 && markLength != 2 {
```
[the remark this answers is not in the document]
JmPotato: It can be left empty.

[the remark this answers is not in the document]
okJiang: Oh, in this case, markLength != 0 is better ~

pkg/utils/logutil/log.go
```diff
@@ -88,22 +92,61 @@ func LogPanic() {
...
+		curRedactType.Store(redactOFF)
+		return
+	}
+	markLength := len(redactInfoMark)
+	if markLength == 0 || markLength != 2 {
+		curRedactType.Store(redactON)
+		return
+	}
+	curRedactType.Store(redactWithMarker)
+	redactMarkLeft.Store(rune(redactInfoMark[0]))
```
rleungx, line 131: can the mark be everything?

JmPotato: I can only be `‹›`.

pkg/utils/logutil/log.go
```diff
@@ -92,52 +94,111 @@ func LogPanic() {
...
+	case RedactInfoLogMark:
+		return json.Marshal("MARK")
+	default:
+	}
+	return json.Marshal(false)
+}
+
+const invalidRedactInfoLogTypeErrMsg = `the "redact-info-log" value is invalid; it should be either false, true, or "MARK"`
+
+// UnmarshalJSON implements the `json.Marshaler` interface	to ensure the compatibility.
```
okJiang, line 123:
```suggestion
// UnmarshalJSON implements the `json.Marshaler` interface to ensure the compatibility.
```

pkg/utils/logutil/log_test.go
```diff
@@ -33,99 +35,158 @@ func TestStringToZapLogLevel(t *testing.T) {
...
+	re := require.New(t)
+	// JSON unmarshal.
+	jsonUnmarshalTestCases := []struct {
+		jsonStr string
+		expect  RedactInfoLogType
+	}{
+		{`false`, RedactInfoLogOFF},
+		{`true`, RedactInfoLogON},
+		{`"MARK"`, RedactInfoLogMark},
+		{`"OTHER"`, RedactInfoLogON},
```
okJiang, line 48: Why `OTHER` should be ON?

pkg/utils/logutil/log.go
```diff
@@ -88,22 +94,120 @@ func LogPanic() {
...
+func (t *RedactInfoLogType) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err == nil {
+		switch strings.ToUpper(s) {
+		case "FALSE", "OFF", "":
```
okJiang, line 128: The behavior here seems different than before, do you need to add it to the release note?

pkg/utils/logutil/log.go
```diff
@@ -88,22 +94,120 @@ func LogPanic() {
...
+			*t = RedactInfoLogOFF
+		case "MARK":
+			*t = RedactInfoLogMark
+		default:
+			*t = RedactInfoLogON
```
okJiang, line 133: ditto

pkg/utils/logutil/log_test.go
```diff
@@ -33,52 +35,172 @@ func TestStringToZapLogLevel(t *testing.T) {
...
+		re.Equal(tc.expect, string(b))
+	}
+	// TOML unmarshal.
+	tomlTestCases := []struct {
+		tomlStr string
+		expect  RedactInfoLogType
+	}{
+		{`redact-info-log = false`, RedactInfoLogOFF},
+		{`redact-info-log = true`, RedactInfoLogON},
+		{`redact-info-log = "MARK"`, RedactInfoLogMark},
```
rleungx, line 85: Should we use marker instead of mark?

conf/config.toml
```diff
@@ -33,14 +33,11 @@
 # key-path = ""
 ## A CN which must be provided by a client
 # cert-allowed-cn = ["example.com"]
-## Whether or not to enable redact log.
+## Whether to enable the log redaction. It can be the following values:
+##   - false: disable redact log.
+##   - true: enable redact log, which will replace the sensitive information with "?".
+##   - "MARK": enable redact log, which will use ‹› to enclose the sensitive information.
```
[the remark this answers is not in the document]
JmPotato: `‹` is different from `<`, the former is called [guillemet](https://en.wikipedia.org/wiki/Guillemet). This is intend to use `‹` rather than `<`.

pkg/utils/logutil/log.go
```diff
@@ -88,22 +94,120 @@ func LogPanic() {
...
 
-// SetRedactLog sets enabledRedactLog
-func SetRedactLog(enabled bool) {
-	enabledRedactLog.Store(enabled)
+const (
+	leftMark  = '‹'
+	rightMark = '›'
+)
+
+func redactInfo(input string) string {
```
okJiang, line 195: `redactInfo` just wraps the input with `< >`, it doesn't allow configuring specific strings to replace values in the logs. This is different from the logic in your code two weeks ago, is this what you expected?

okJiang (changes requested):

pkg/utils/logutil/log_test.go
JmPotato: I consider that if the user uses a string other than true/false and it is not `MARK`, at least the enabled state should be retained. Of course, this can also be discussed, and I remain open to this.

pkg/utils/logutil/log.go
JmPotato: I won't expose the "FALSE" or "OFF" (so does "TRUE" or "ON") to the user in the release note and documentation. This is only to ensure the robustness.

pkg/utils/logutil/log.go
JmPotato: Yes, after discussion with PM, this is the only behavior when `MARK` is enabled.

okJiang: After checking the [tidb_redact_log](https://docs.pingcap.com/tidb/dev/system-variables#tidb_redact_log), it really is like that. I find it a bit strange, does it have a desensitizing effect?

pkg/utils/logutil/log_test.go
JmPotato: Nice catch. It should be `MARKER` as [TiDB does](https://docs.pingcap.com/tidb/stable/system-variables#tidb_redact_log). I will fix it.

pkg/utils/logutil/log_test.go
okJiang: Just reach an agreement with PM~

pkg/utils/logutil/log.go
JmPotato: The purpose of this feature is to allow the cloud or customers to customize how to handle sensitive information. As for the output here in PD, sensitive information will not actually be desensitized; it is up to the user to decide how to handle this sensitive information.

okJiang: lgtm

conf/config.toml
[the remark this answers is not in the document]
JmPotato: I have added comments to emphasize this.

pkg/utils/logutil/log_test.go
JmPotato: I modified the behavior to be consistent with TiDB, only allowing it to be true/false/"MARKER". PTAL.

okJiang (approved):

rleungx (approved):

HuSharp (approved):
nit: Can you add to the pr description the log display for different log settings?

JmPotato:
> nit: Can you add to the pr description the log display for different log settings?

Pasted. PTAL.

/approve

niubell (approved):

PR #8301, opened 2024-06-17
*: enable error check for some files

okJiang:
### What problem does this PR solve?

Issue Number: Ref https://github.com/tikv/pd/issues/1919

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Code changes

- Has the configuration change
- Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Possible performance regression
- Increased code complexity
- Breaking backward compatibility

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

PR #8308, opened 2024-06-19
test: add test for clone unmarshal

HuSharp:

Issue Number: Ref #4399

issue relies on 
https://github.com/tikv/pd/pull/8301/commits/76b49327ff7842027631642ad1d3746c5b14a02a#diff-5cf563058fdae91ee59093d5721fde21b45888669570a5fdfd54bbe502ce3e03L93-L95

We have this logic for `clone`
https://github.com/tikv/pd/blob/13174b5d4cab4d958f0b4ea2718ea7bb74992bc7/client/http/types.go#L366-L373

This is because startKey is json:"-" which means cannot be Unmarshal from json, We need to take care of `Clone` method.
https://github.com/tikv/pd/blob/13174b5d4cab4d958f0b4ea2718ea7bb74992bc7/client/http/types.go#L346-L349

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/http/types_test.go
+++ b/client/http/types_test.go
@@ -308,3 +308,30 @@ func mustMarshalAndUnmarshalRuleOp(re *require.Assertions, ruleOp *RuleOp) *Rule
 	re.NoError(err)
 	return newRuleOp
 }
+
+// startKey and endKey are json:"-" which means cannot be Unmarshal from json
+// We need to take care of `Clone` method.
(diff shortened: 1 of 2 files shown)


HuSharp:
@okJiang PTAL, thx!

rleungx (approved):

JmPotato (approved):

Issue #5845, opened 2023-01-10
metrics: unify '-' and '_' in schedulers and checkers


lhy1024:
## Enhancement Task
There are mixed '-' and '_' in schedulers and checkers
```
replicaCheckerName = "replica-checker"
replicaChecker     = "replica_checker"
```

lhy1024:
how about unified with '_'? @nolouch @rleungx

nolouch:
lgtm

okJiang:
According to the conclusion in this [PR](https://github.com/tikv/pd/issues/8379), this issue can be closed.

> For monitoring, it will be used as a filter rule_checker/replica_checker/merge_checker , temporarily converted from xxxx-checker to xxxx_checker when needed to maintain user experience compatibility. If it is determined that there is no compatibility problem, then change it

okJiang:
Please help to close the issue~ cc @rleungx @lhy1024

PR #8316, opened 2024-06-21
checker: unify checker name and checker type

okJiang:

Issue Number: ~~Ref https://github.com/tikv/pd/issues/5845~~ Ref https://github.com/tikv/pd/issues/8379

```commit-message
* Use underscore format to represent a checker
* use dash format to represent a checker in monitoring
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/schedule/config/name.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+type CheckerSchedulerName string
```
okJiang, line 17: In a later PR, the scheduler will also use this type.

okJiang: Why using one type, not checkerName and schedulerName separately?

It is because I found they sometimes mix.
https://github.com/tikv/pd/blob/41e63768870fd86aaf78f499503a601249a7db39/pkg/schedule/filter/counter.go#L77-L93

pkg/schedule/filter/counter.go
```diff
@@ -83,8 +83,8 @@ var scopes = [scopeLen]string{
 
 	"evict-leader-scheduler",
 	"region-scatter",
-	"replica-checker",
-	"rule-checker",
+	"replica_checker",
```
rleungx, line 86: Can we unify the `-` and `_`?

okJiang: scheduler will be unified in the next pr~

rleungx: The monitor might also need to be changed.

okJiang: Do you mean the checker's monitor?

rleungx: I'm not sure about that, maybe scheduler or maybe checker

okJiang: The scheduler name has not been changed.

The checker monitor is no need to change, after I checked. They are all in the form of xxx_checker.

rleungx: Just a reminder

pkg/schedule/checker/rule_checker.go
```diff
@@ -404,7 +403,7 @@ func (c *RuleChecker) allowLeader(fit *placement.RegionFit, peer *metapb.Peer) b
 	if s == nil {
 		return false
 	}
-	stateFilter := &filter.StoreStateFilter{ActionScope: "rule-checker", TransferLeader: true}
+	stateFilter := &filter.StoreStateFilter{ActionScope: "rule_checker", TransferLeader: true}
```
rleungx, line 406: Will it affect the monitor?

pkg/schedule/checker/rule_checker.go
```diff
@@ -108,7 +107,7 @@ func NewRuleChecker(ctx context.Context, cluster sche.CheckerCluster, ruleManage
 
 // GetType returns RuleChecker's Type
 func (*RuleChecker) GetType() string {
-	return ruleCheckerName
+	return config.RuleCheckerName.String()
```
rleungx, line 110: After this PR, the type will be changed from xx-xx to xx_xx.

pkg/schedule/checker/rule_checker.go
okJiang: image

Scope is by group, so it will not affect the monitor. It will generate a new name after the new version goes live, but it won't have any actual impact.

image

pkg/schedule/checker/rule_checker.go
okJiang: https://github.com/tikv/pd/pull/8316/files#r1650478721

pkg/schedule/checker/rule_checker.go
rleungx: Will it become xxx_checker-store which mix use dash and underscore?

okJiang: Indeed. We want to use underscores in the code and hyphens for monitoring, right?

Waiting for confirmation~

rleungx: IMO, it's better to use hyphens in monitor.

pkg/schedule/operator/metrics.go
```diff
@@ -82,10 +85,16 @@ var (
...
 	prometheus.MustRegister(operatorCounter)
 	prometheus.MustRegister(operatorDuration)
 	prometheus.MustRegister(operatorSizeHist)
 	prometheus.MustRegister(storeLimitCostCounter)
 }
+
+// IncOperatorLimitCounter increases the counter of operator meeting limit.
+func IncOperatorLimitCounter(typ, name string) {
+	operatorLimitCounter.WithLabelValues(
+		strings.ReplaceAll(typ, "_", "-"), name).Inc()
```
rleungx, line 99: Is it possible not to replace it every time?

okJiang: Is it for performance reasons? In that case, we still need to cache one more xxxx-xxxx for monitoring, which brings us back to the issue.

okJiang: How about using a map to do this?

okJiang: https://github.com/tikv/pd/pull/8316/commits/aadb26d8fdc83a93e7f6b7843866f63c94118262

plugin/scheduler_example/evict_leader.go
```diff
@@ -37,14 +37,11 @@ import (
 	"github.com/unrolled/render"
 )
 
-const (
-	// EvictLeaderName is evict leader scheduler name.
-	EvictLeaderName        config.CheckerSchedulerName = "user-evict-leader-scheduler"
-	noStoreInSchedulerInfo                             = "No store in user-evict-leader-scheduler-config"
-)
+const noStoreInSchedulerInfo = "No store in user-evict-leader-scheduler-config"
```
rleungx, line 40: no need to define a constant

rleungx:
BTW, this PR is too large to review. Is it possible to split them?

okJiang:
> BTW, this PR is too large to review. Is it possible to split them?

After the ci pass, I will try it. But I still have to say, splitting is really tough😭 @rleungx

Issue #1919, opened 2019-11-10
let errcheck pass in ci


zhouqiang-cl:
## PCP Issue

- difficult: easy
- score: 50
- min score required: 0
- Recommended Skills: Go

## Description
we disable the  errorcheck in #1956, if you run `GO111MODULE=on errcheck ./...` will got:
```
client/client.go:278:11:	cc.Close()
client/client_test.go:114:17:	s.srv.PutStore(context.Background(), &pdpb.PutStoreRequest{Header: newHeader(s.srv), Store: store})
client/client_test.go:295:46:	s.srv.GetRaftCluster().HandleRegionHeartbeat(region3)
cmd/pd-server/main.go:70:16:	defer log.Sync()
cmd/pd-server/main.go:133:10:	log.Sync()
pkg/apiutil/apiutil.go:120:10:	rd.JSON(w, http.StatusInternalServerError, "nil error")
pkg/apiutil/apiutil.go:125:10:	rd.JSON(w, errCode.Code().HTTPCode(), errcode.NewJSONFormat(errCode))
pkg/apiutil/apiutil.go:127:10:	rd.JSON(w, http.StatusInternalServerError, err.Error())
pkg/btree/btree.go:746:13:	fmt.Fprintf(w, "%sNODE:%v, %v\n", strings.Repeat("  ", level), n.items, n.indices)
pkg/etcdutil/etcdutil_test.go:64:14:	os.RemoveAll(cfg.Dir)
pkg/testutil/testutil.go:65:14:	os.RemoveAll(dataDir)
server/api/admin.go:44:12:	h.rd.JSON(w, http.StatusBadRequest, err.Error())
server/api/admin.go:48:11:	h.rd.JSON(w, http.StatusOK, nil)
server/api/admin.go:59:12:	h.rd.JSON(w, http.StatusBadRequest, "invalid tso value")
server/api/admin.go:64:12:	h.rd.JSON(w, http.StatusBadRequest, "invalid tso value")
server/api/admin.go:70:13:	h.rd.JSON(w, http.StatusInternalServerError, err.Error())
server/api/admin.go:72:13:	h.rd.JSON(w, http.StatusForbidden, err.Error())
server/api/admin.go:75:11:	h.rd.JSON(w, http.StatusOK, "success")
server/api/admin_test.go:77:16:	res.Body.Close()
server/api/cluster.go:36:11:	h.rd.JSON(w, http.StatusOK, h.svr.GetCluster())
[...]

PR #8317, opened 2024-06-21
*: enable errcheck for server/*.go except server/api

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/1919

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


cmd/pd-server/main.go
```diff
@@ -295,6 +299,6 @@ func start(cmd *cobra.Command, args []string, services ...string) {
 }
 
 func exit(code int) {
-	log.Sync()
+	log.Sync() // nolint:errcheck
```
rleungx, line 302:
```suggestion
	_ = log.Sync()
```

pkg/replication/replication_mode.go
```diff
@@ -232,46 +232,44 @@ func (m *ModeManager) loadDRAutoSync() error {
 	return nil
 }
 
-func (m *ModeManager) drSwitchToAsyncWait(availableStores []uint64) error {
+func (m *ModeManager) drSwitchToAsyncWait(availableStores []uint64) {
```
rleungx, line 235: Why change it?

pkg/replication/replication_mode.go
```diff
@@ -232,46 +232,44 @@ func (m *ModeManager) loadDRAutoSync() error {
...
-		return err
+		return
 	}
 	m.drAutoSync = dr
 	log.Info("switched to async_wait state", zap.String("replicate-mode", modeDRAutoSync))
-	return nil
 }
 
-func (m *ModeManager) drSwitchToAsync(availableStores []uint64) error {
+func (m *ModeManager) drSwitchToAsync(availableStores []uint64) {
```
rleungx, line 253: ditto

pkg/replication/replication_mode.go
okJiang: The caller always ignored this error, and all errors were logged within the function, so I removed it.

rleungx: But the tests will check the return value of `drSwitchToAsync`.

okJiang: Got it. I will revert it for the test

cmd/pd-server/main.go
rleungx: yes

pkg/mcs/registry/registry.go
```diff
@@ -82,14 +82,20 @@ func (r *ServiceRegistry) InstallAllRESTHandler(srv bs.Server, h map[string]http
 	for name, builder := range r.builders {
 		serviceName := createServiceName(prefix, name)
 		if l, ok := r.services[serviceName]; ok {
-			l.RegisterRESTHandler(h)
-			log.Info("restful API service already registered", zap.String("prefix", prefix), zap.String("service-name", name))
+			if err := l.RegisterRESTHandler(h); err != nil {
+				log.Error("register REST handler failed", zap.String("prefix", prefix), zap.String("service-name", name), zap.Error(err))
```
rleungx, line 86:
```suggestion
				log.Error("register restful API service failed", zap.String("prefix", prefix), zap.String("service-name", name), zap.Error(err))
```

pkg/storage/endpoint/gc_safe_point.go
```diff
@@ -100,7 +100,9 @@ func (se *StorageEndpoint) LoadMinServiceGCSafePoint(now time.Time) (*ServiceSaf
 		}
 
 		if ssp.ExpiredAt < now.Unix() {
-			se.Remove(key)
+			if err := se.Remove(key); err != nil {
```
rleungx, line 103: Do we need to return the error here?

pkg/syncer/server.go
```diff
@@ -282,7 +282,10 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
 					RegionLeaders: leaders,
 					Buckets:       buckets,
 				}
-				s.limit.WaitN(ctx, resp.Size())
+				if err := s.limit.WaitN(ctx, resp.Size()); err != nil {
+					log.Error("failed to wait rate limit", errs.ZapError(err))
+					return err
```
rleungx, line 287: Do we need to return?

pkg/mcs/resourcemanager/server/server.go
```diff
@@ -213,7 +215,9 @@ func (s *Server) Close() {
 	}
 
 	log.Info("closing resource manager server ...")
-	s.serviceRegister.Deregister()
+	if err := s.serviceRegister.Deregister(); err != nil {
+		log.Error("failed to deregister the service", errs.ZapError(err))
```
HuSharp, line 219: Do you think we should indicate which micro-service? such as
```suggestion
		log.Error("micro-service resource manager failed to deregister the service", errs.ZapError(err))
```

pkg/syncer/server.go
okJiang: I think it's possible to return an error here.

After checking the source code, there are two types of errors that WaitN returns, `fmt.Errorf("rate: Wait(n=%d) exceeds limiter's burst %d", n, burst)` and context errors.

- For the former. If `resp.Size()` exceeds the burst, then continuing to loop will still result in an error because `resp.Size` will keep increasing. In this case, the error won't be thrown, but adding logs will keep printing out, and the client may never receive `resp`. Unless we can guarantee that within `maxSyncRegionBatchSize`, resp.Size will never exceed the burst (20MiB).
- For the latter, it doesn't matter, because there is a `select ctx.Done()` inside the loop.

If we can't guarantee that the information of maxSyncRegionBatchSize regions is definitely less than 20MiB, it's best to check if resp.Size is close to 20MiB before continuing at L275.

pkg/mcs/resourcemanager/server/server.go
okJiang: The L217 prints `closing resource manager server ...`, so it doesn't matter.

HuSharp: got

pkg/storage/endpoint/gc_safe_point.go
okJiang: fixed in https://github.com/tikv/pd/pull/8317/commits/c01b053b3bf93405d746c7db75db873345257ca3

cmd/pd-server/main.go
JmPotato: Using `_ = log.Sync()` might be better to me.

pkg/mcs/resourcemanager/server/grpc_service.go
```diff
@@ -49,7 +49,7 @@ type dummyRestService struct{}
 
 func (dummyRestService) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	w.WriteHeader(http.StatusNotImplemented)
-	w.Write([]byte("not implemented"))
+	w.Write([]byte("not implemented")) // nolint:errcheck
```
JmPotato, line 52: Ditto. Prefer using `_ = w.Write([]byte("not implemented"))`.

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -53,7 +53,7 @@ type dummyRestService struct{}
 
 func (dummyRestService) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	w.WriteHeader(http.StatusNotImplemented)
-	w.Write([]byte("not implemented"))
+	w.Write([]byte("not implemented")) // nolint:errcheck
```
JmPotato, line 56: Ditto.

pkg/mcs/registry/registry.go
```diff
@@ -82,14 +82,20 @@ func (r *ServiceRegistry) InstallAllRESTHandler(srv bs.Server, h map[string]http
...
+				log.Info("restful API service already registered", zap.String("prefix", prefix), zap.String("service-name", name))
+			}
 			continue
 		}
 		l := builder(srv)
 		r.services[serviceName] = l
-		l.RegisterRESTHandler(h)
-		log.Info("restful API service registered successfully", zap.String("prefix", prefix), zap.String("service-name", name))
+		if err := l.RegisterRESTHandler(h); err != nil {
+			log.Error("register restful API service failed", zap.String("prefix", prefix), zap.String("service-name", name), zap.Error(err))
```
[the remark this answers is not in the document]
okJiang: The line numbers of log will be automatically differentiated, or do you have any suggestions on how to modify the log content?

okJiang:
Can you take a look again? Thanks! @JmPotato @rleungx @HuSharp

cmd/pd-server/main.go
```diff
@@ -231,7 +235,7 @@ func start(cmd *cobra.Command, args []string, services ...string) {
 		log.Fatal("initialize logger error", errs.ZapError(err))
 	}
 	// Flushing any buffered log entries
-	defer log.Sync()
+	defer log.Sync() //nolint:errcheck
```
rleungx, line 238: Please also change it.

pkg/mcs/resourcemanager/server/grpc_service.go
```diff
@@ -228,6 +228,8 @@ func (s *Service) AcquireTokenBuckets(stream rmpb.ResourceManager_AcquireTokenBu
 			log.Debug("finish token request from", zap.String("resource-group", resourceGroupName))
 			resps.Responses = append(resps.Responses, resp)
 		}
-		stream.Send(resps)
+		if err := stream.Send(resps); err != nil {
+			return errors.WithStack(err)
+		}
```
rleungx, line 233: I'm not sure whether it is on purpose.

nolouch: indeed forgot to consider error handling previously.

rleungx (approved):

JmPotato (approved):

niubell (approved):

Issue #8343, opened 2024-06-27
Interfering logs in resource controller


nolouch:
### What did you see instead?
![image](https://github.com/tikv/pd/assets/6428910/850b56ff-0ace-440f-91dd-a6df4a027052)
some of the quota limitation errors in the logs are false positives. Only the errors within the red boxes are real.

PR #8344, opened 2024-06-27
client/controller: record context error and add slowlog about token bucket

nolouch:

Issue Number: Close https://github.com/tikv/pd/issues/8343

```commit-message
client/controller: record context error and add slowlog about token bucket
- record low process start time, and log it if it's too slow
- record the context error
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```

### DIFF
--- a/client/resource_group/controller/controller.go
+++ b/client/resource_group/controller/controller.go
@@ -140,7 +140,7 @@ type ResourceGroupsController struct {
 	calculators []ResourceCalculator
 
 	// When a signal is received, it means the number of available token is low.
-	lowTokenNotifyChan chan NotifyMsg
+	lowTokenNotifyChan chan notifyMsg
 	// When a token bucket response received from server, it will be sent to the channel.
 	tokenResponseChan chan []*rmpb.TokenBucketResponse
 	// When the token bucket of a resource group is updated, it will be sent to the channel.
@@ -182,7 +182,7 @@ func NewResourceGroupController(
 		clientUniqueID:        clientUniqueID,
 		provider:              provider,
 		ruConfig:              ruConfig,
-		lowTokenNotifyChan:    make(chan NotifyMsg, 1),
+		lowTokenNotifyChan:    make(chan notifyMsg, 1),
 		tokenResponseChan:     make(chan []*rmpb.TokenBucketResponse, 1),
 		tokenBucketUpdateChan: make(chan *groupCostController, maxNotificationChanLen),
 		opts:                  opts,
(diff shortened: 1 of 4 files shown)


client/resource_group/controller/limiter.go
```diff
@@ -218,7 +224,8 @@ func (lim *Limiter) Reserve(ctx context.Context, waitDuration time.Duration, now
 	select {
 	case <-ctx.Done():
 		return &Reservation{
-			ok: false,
+			ok:  false,
+			err: ctx.Err(),
```
HuSharp, line 228: Can we use `err` instead of `ok`? That would add `ErrClientResourceGroupThrottled` to Err.

client/resource_group/controller/controller.go
```diff
@@ -542,13 +543,16 @@ func (c *ResourceGroupsController) sendTokenBucketRequests(ctx context.Context,
 		if err != nil {
 			// Don't log any errors caused by the stopper canceling the context.
 			if !errors.ErrorEqual(err, context.Canceled) {
-				log.L().Sugar().Infof("[resource group controller] token bucket rpc error: %v", err)
+				log.Info("[resource group controller] token bucket rpc error", zap.Error(err))
```
HuSharp, line 546: using error?
```suggestion
				log.Error("[resource group controller] token bucket rpc error", zap.Error(err))
```

HuSharp, line 546: using error?
```suggestion
				log.Error("[resource group controller] token bucket rpc error", zap.Error(err))
```

client/resource_group/controller/controller.go
```diff
@@ -853,7 +857,7 @@ func (gc *groupCostController) updateRunState() {
 	}
 	*gc.run.consumption = *gc.mu.consumption
 	gc.mu.Unlock()
-	logControllerTrace("[resource group controller] update run state", zap.Any("request-unit-consumption", gc.run.consumption))
+	logControllerTrace("[resource group controller] update run state", zap.String("name", gc.name), zap.Any("request-unit-consumption", gc.run.consumption))
```
HuSharp, line 860: Can you help change this log in `applyBasicConfigForRUTokenCounters`?
`log.Info("[resource group controller] resource token bucket enter degraded mode", zap.String("resource-group", gc.name), zap.String("type", rmpb.RequestUnitType_name[int32(typ)]))`
to
`log.Info("[resource group controller] resource token bucket enter degraded mode", zap.String("name", gc.name), zap.String("type", rmpb.RequestUnitType_name[int32(typ)]))`

client/resource_group/controller/limiter.go
nolouch: `ok` is only used for can if reserved enough tokens, but the error may include others like ctx.Error(). 

let's keep min changes in this PR first, I don't want to bring potential risks to other branches by picking.

HuSharp (approved):

client/resource_group/controller/limiter.go
```diff
@@ -88,6 +88,11 @@ type Limiter struct {
 	metrics *limiterMetricsCollection
 }
 
+// NotifyMsg is a message to notify the low token state.
+type NotifyMsg struct {
+	StartTime time.Time
+}
```
JmPotato, line 94: Both the struct and its field appear unnecessary to be exported.

Issue #5691, opened 2022-11-09
Hot statistic and hot scheduler issue list


lhy1024:
## Performance
- [ ] https://github.com/tikv/pd/issues/5692
- [ ] https://github.com/tikv/pd/issues/3725
- [ ] speed up to create operator in big cluster

## Statistic
- [x] #4042
- [ ] https://github.com/tikv/pd/issues/2497
- [ ] https://github.com/tikv/pd/issues/4481
- [ ] https://github.com/tikv/pd/issues/3158
- [ ] https://github.com/tikv/pd/issues/3084

## Why no hot-scheduler operator?
- [X] https://github.com/tikv/pd/issues/4529
- [ ] https://github.com/tikv/tikv/issues/13461
- [ ] #3098
- [ ] https://github.com/tikv/pd/issues/4041

## Why has hot-scheduler operator?
- [x] https://github.com/tikv/pd/issues/3919
- [ ] https://github.com/tikv/pd/issues/5559
- [ ] https://github.com/tikv/pd/issues/3298

## Reduce schedule cost
- [ ] https://github.com/tikv/pd/issues/5465
- [ ] https://github.com/tikv/pd/issues/3852
- [ ] https://github.com/tikv/pd/issues/2999
- [ ] https://github.com/tikv/pd/issues/3842
- [ ] https://github.com/tikv/pd/issues/662

## Dimension selection
- [ ] #5718
- [ ] https://github.com/tikv/pd/issues/4612
- [ ] https://github.com/tikv/pd/issues/4513
- [X] https://github.com/tikv/pd/issues/3869
- [x] https://github.com/tikv/pd/issues/1432

## Dimension conflict
- [x] https://github.com/tikv/pd/issues/5446
- [X] https://github.com/tikv/pd/issues/4332
- [x] https://github.com/tikv/pd/issues/4949
- [x] https://github.com/tikv/pd/issues/4628

## Expectation and variance
- [ ] https://github.com/tikv/pd/issues/2168
- [ ] https://github.com/tikv/pd/issues/5545
- [x] https://github.com/tikv/pd/issues/3830

## Load base split
- [ ] https://github.com/tikv/pd/issues/5267
- [ ] hot scheduler can create split operator

## Metrics, logs and pd-ctl 
- [x] https://github.com/tikv/pd/issues/5643
- [ ] https://github.com/tikv/pd/issues/4350
- [ ] https://github.com/tikv/pd/issues/4993
[...]

PR #8345, opened 2024-06-27
scheduler: add more hot scheduler comments and replace negative rank

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/5691

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/hot_region.go
```diff
@@ -1242,7 +1245,7 @@ func (bs *balanceSolver) calcProgressiveRankV1() {
 		// If the first priority is better, the progressiveRank is -3.
```
okJiang, line 1245:
```suggestion
		// If the first priority is better, the progressiveRank is 3.
```

pkg/schedule/schedulers/hot_region.go
```diff
@@ -1225,12 +1228,12 @@ func (bs *balanceSolver) isUniformSecondPriority(store *statistics.StoreLoadDeta
...
 // See the comments of `solution.progressiveRank` for more about progressive rank.
-// | ↓ firstPriority \ secondPriority → | isBetter | isNotWorsened | Worsened |
-// |   isBetter                         | -4       | -3            | -1 / 0   |
-// |   isNotWorsened                    | -2       | 1             | 1        |
-// |   Worsened                         | 0        | 1             | 1        |
+// | ↓ firstPriority \ secondPriority → | isBetter | isNotWorsened | Worsened  |
+// |   isBetter                         | 4        | 3             | 1         |
+// |   isNotWorsened                    | 2        | -1            | -1        |
+// |   Worsened                         | 0        | -1            | -1        |
 func (bs *balanceSolver) calcProgressiveRankV1() {
```
okJiang, line 1235: How about using variables to represent instead of magic numbers?

lhy1024: It is the score to use comparing.

okJiang: Yes, it is. But it is still being used in many places for making judgments, which indeed can be hard to understand in the code, so more comments have to be added to explain it. Like these

https://github.com/tikv/pd/blob/cbba49ab32ddd82d86194abd8cf9c25d1db0c667/pkg/schedule/schedulers/hot_region.go#L1373-L1393

I feel like it's really hard to name, so I'm just suggesting something and not rushing to solve it. :)

pkg/schedule/schedulers/hot_region_rank_v1.go
```diff
@@ -29,9 +29,9 @@ func initRankV1(r *balanceSolver) *rankV1 {
 }
 
 // isAvailable returns the solution is available.
-// The solution should have no revertRegion and progressiveRank < 0.
+// The solution should have no revertRegion and progressiveRank > 0.
 func (*rankV1) isAvailable(s *solution) bool {
-	return s.progressiveRank < 0
+	return s.progressiveRank > 0 && s.revertRegion == nil
```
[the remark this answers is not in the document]
lhy1024: keep consistent with comments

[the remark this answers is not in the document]
lhy1024: However, it seems that v1 does not enable this feature.

pkg/statistics/store_hot_peers_infos.go
```diff
@@ -166,49 +165,37 @@ func summaryStoresLoadByEngine(
...
 			for i := range peerLoadSum {
 				peerLoadSum[i] += peer.GetLoad(i)
 			}
 			hotPeers = append(hotPeers, peer.Clone())
 		}
-		{
-			// Metric for debug.
-			// TODO: pre-allocate gauge metrics
-			ty := "byte-rate-" + rwTy.String() + "-" + kind.String()
-			hotPeerSummary.WithLabelValues(ty, fmt.Sprintf("%v", id)).Set(peerLoadSum[utils.ByteDim])
```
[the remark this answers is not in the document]
lhy1024: Yeah, we haven't been using this metrics

rleungx (approved):

pkg/schedule/schedulers/hot_region_rank_v1.go
```diff
@@ -29,9 +29,10 @@ func initRankV1(r *balanceSolver) *rankV1 {
 }
 
 // isAvailable returns the solution is available.
-// The solution should have no revertRegion and progressiveRank < 0.
+// The solution should progressiveRank > 0.
```
HuSharp, line 32: Does the comment in `type solution struct` about `progressiveRank` also need to be changed?

lhy1024: good catch

pkg/schedule/schedulers/hot_region.go
```diff
@@ -409,9 +411,9 @@ type solution struct {
 
 	// progressiveRank measures the contribution for balance.
 	// The smaller the rank, the better this solution is.
```
HuSharp, line 413:
```suggestion
	// The bigger the rank, the better this solution is.
```

HuSharp (approved):

pkg/schedule/schedulers/hot_region_config.go
```diff
@@ -38,7 +38,6 @@ import (
 )
 
 const (
-
```
lhy1024, line 41: Because modify config file, PTAL cc @niubell

niubell (approved):

PR #8347, opened 2024-07-01
client: clean up the tls config for client

rleungx:

Issue Number: Ref #4399

There are some useless configs which are exposed to the user.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/tlsutil/tlsconfig.go
+++ b/client/tlsutil/tlsconfig.go
@@ -47,20 +47,11 @@ import (
 type TLSInfo struct {
 	CertFile           string
 	KeyFile            string
-	CAFile             string // TODO: deprecate this in v4
 	TrustedCAFile      string
-	ClientCertAuth     bool
-	CRLFile            string
 	InsecureSkipVerify bool
-
-	SkipClientSANVerify bool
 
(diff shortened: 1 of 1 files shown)


client/tlsutil/tlsconfig.go
```diff
@@ -47,21 +47,12 @@ import (
 type TLSInfo struct {
 	CertFile           string
 	KeyFile            string
-	CAFile             string // TODO: deprecate this in v4
```
nolouch, line 50: Does it work if there is no CAfile? like pdctl no specified the `--cacert`.

rleungx: https://github.com/tikv/pd/pull/8347/files#diff-91dc9b32b58d11d334b55b1e6fdc4495c5d28d42f125320b8611d43238799bbcR160 still works. It will be assigned to TrustedCAFile.

okJiang (approved):

nolouch (approved):

HuSharp (approved):

niubell (approved):

PR #8367, opened 2024-07-05
metrics: use second as the unit of latency

rleungx:

Issue Number: Ref #4399

These metrics are not displayed in the Grafana.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/tso_stream.go
+++ b/client/tso_stream.go
@@ -139,7 +139,7 @@ func (s *pdTSOStream) processRequests(
 		}
 		return
 	}
-	tsoBatchSendLatency.Observe(float64(time.Since(batchStartTime)))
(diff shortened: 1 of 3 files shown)


JmPotato (approved):

lhy1024 (approved):

client/tso_stream.go
```diff
@@ -139,7 +139,7 @@ func (s *pdTSOStream) processRequests(
 		}
 		return
 	}
-	tsoBatchSendLatency.Observe(float64(time.Since(batchStartTime)))
+	tsoBatchSendLatency.Observe(time.Since(batchStartTime).Seconds())
```
niubell, line 142: If the time delta is less than 1 second, will we got zero as the value of Observe?Is it expected?

Issue #8358, opened 2024-07-03
BatchScanRegions need to return an error to tell client regions exist hole for retry


okJiang:
## Enhancement Task

ref 

- https://github.com/tikv/client-go/pull/1377
- https://github.com/pingcap/tidb/blob/1480bb127a9363e31e8b2e9f3808289ea54d9d28/br/pkg/streamhelper/regioniter.go#L103
- https://github.com/pingcap/tidb/blob/1480bb127a9363e31e8b2e9f3808289ea54d9d28/br/pkg/restore/split/split.go#L40

Tristan1900:
@okJiang Looks like after the proto change, TiKV code `pd_client/src/util#check_resp_header` breaks because of not handling the added error type. Curious if there is a fix already opened or in progress? Or who should be good to contact to fix it?  Thanks!

okJiang:
> @okJiang Looks like after the proto change, TiKV code `pd_client/src/util#check_resp_header` breaks because of not handling the added error type. Curious if there is a fix already opened or in progress? Or who should be good to contact to fix it? Thanks!

If you don't set the added parameter(contain_all_key_range), it will not return the added error type(REGIONS_NOT_CONTAIN_ALL_KEY_RANGE). So can you explain which problem do you meet?

Tristan1900:
@okJiang hmm, from my side it's a compiler error that `check_resp_header` did not exhaustively handle all the enum type
<img width="1486" alt="image" src="https://github.com/user-attachments/assets/f7b48e06-4516-4040-acfd-e76b29dd0ab9">

maybe I somehow use it wrong? I added a panic to bypass the complier to not blocking me but I don't think that's the correct way to do so asking for help.

okJiang:
> @okJiang hmm, from my side it's a compiler error that `check_resp_header` did not exhaustively handle all the enum type. BTW I did not make any changes on pd_client, I'm working on the backup side
> 
[...]

PR #8375, opened 2024-07-08
client: report error when scan region encounter hole region

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8358

ref https://github.com/pingcap/kvproto/pull/1252

```commit-message
client: report error when scan region encounter hole region
- add a input parameter(OutputMustContainAllKeyRange) in the BatchScanRegionsRequest
- when the new param enable and find the result doesn't contain all key range(input), it will return an error to user
- add Merge() method to merge the continuous KeyRanges
- pull out the scanRegion function and add unit tests for it
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
client: report error when scan region encounter hole region
```


okJiang:
cc @lance6716 @you06 @rleungx

pkg/core/basic_cluster.go
```diff
@@ -97,7 +99,7 @@ type RegionSetInformer interface {
 	GetAdjacentRegions(region *RegionInfo) (*RegionInfo, *RegionInfo)
 	ScanRegions(startKey, endKey []byte, limit int) []*RegionInfo
 	GetRegionByKey(regionKey []byte) *RegionInfo
-	BatchScanRegions(keyRanges *KeyRanges, limit int) []*RegionInfo
+	BatchScanRegions(keyRanges *KeyRanges, limit int, outputMustContainAllKeyRange bool) ([]*RegionInfo, error)
```
lance6716, line 102: I hope it's 

```golang
BatchScanRegions(keyRanges *KeyRanges, limit int, scanOptions ...BatchScanRegionsOptions) ([]*RegionInfo, error)
```

So in future we don't need to change the signature.

rleungx: +1

okJiang: lgtm

okJiang: But this is an internal interface and it won't change the way it's called externally. @lance6716 

Grpc interface is here https://github.com/pingcap/kvproto/pull/1252

okJiang: For internal interfaces, I prefer to wait for more options to emerge before tweaking to avoid over-optimization. @rleungx

lance6716: Although it's an internal interface, if PD developers will add more options in future, using `...BatchScanRegionsOptions` can save your time of editing the signature of interface and implementation.

server/grpc_service.go
```diff
@@ -1680,7 +1680,10 @@ func (s *GrpcServer) BatchScanRegions(ctx context.Context, request *pdpb.BatchSc
 		}
 		keyRanges.Append(reqRange.StartKey, reqRange.EndKey)
 	}
-	res := rc.BatchScanRegions(keyRanges, int(limit))
+	res, err := rc.BatchScanRegions(keyRanges, int(limit), request.OutputMustContainAllKeyRange)
+	if err != nil {
```
lance6716, line 1684: I'm not sure this error check relies too much on the behaviour of `rc.BatchScanRegions`. What if it will return another type of error (not ErrorType_REGIONS_NOT_CONTAIN_ALL_KEY_RANGE) in future?

pkg/core/region.go
```diff
@@ -1826,37 +1827,75 @@ func (r *RegionsInfo) ScanRegions(startKey, endKey []byte, limit int) []*RegionI
...
+		if len(keyRange.EndKey) > 0 && len(region.GetStartKey()) > 0 &&
+			bytes.Compare(region.GetStartKey(), keyRange.EndKey) >= 0 {
+			return false
+		}
+		if exceedLimit() {
+			return false
+		}
+		if len(lastRegion.GetEndKey()) > 0 && len(region.GetStartKey()) > 0 &&
+			bytes.Compare(region.GetStartKey(), lastRegion.GetEndKey()) > 0 {
+			err = multierr.Append(err, errors.Errorf(
```
lance6716, line 1878: if we pass `outputMustContainAllKeyRange` into this function, we can return `false` to `regionTree.scanRange` to fail fast. Also I think a large error message may exceed gRPC message size limit.

okJiang: > if we pass outputMustContainAllKeyRange into this function, we can return false to regionTree.scanRange to fail fast.

Yeah, it's a trade-off. I chose not to pass in any extra parameters to avoid making this function too complicated, but that does mean it can't fail fast. Waiting to hear what other reviewers think.

> Also it can avoid a large error message exceeding gRPC message size limit.

This function won't cause any issues with gRPC transmission, so there's no problem. Regions with issues will be dropped at the upper level.

server/grpc_service.go
okJiang: In that case, we only have to do some more complicated error handling :(

pkg/core/basic_cluster.go
rleungx: option will help us reduce the modification of the interface

okJiang:
Could you take a look again? @lance6716 @rleungx @you06

server/grpc_service.go
lance6716: In future the developer may forget to change this line. So I sightly recommend at present we add more strict check before converting errors.

lance6716 (commented):
rest lgtm

client/client.go
```diff
@@ -214,8 +214,9 @@ func WithSkipStoreLimit() RegionsOption {
 
 // GetRegionOp represents available options when getting regions.
 type GetRegionOp struct {
-	needBuckets         bool
-	allowFollowerHandle bool
+	needBuckets                  bool
+	allowFollowerHandle          bool
+	outputMustContainAllKeyRange bool
```
rleungx, line 219: Do we need it?

okJiang: Yes?

It has been input in L1179

rleungx: Oh, I was mislead by batchScanRegionsOptions.

tools/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/tikv/pd/tools
 go 1.21
 
 replace (
+	github.com/pingcap/kvproto => github.com/okjiang/kvproto v0.0.0-20240708081830-174112807173
```
rleungx, line 6: remove it?

pkg/core/basic_cluster.go
```diff
@@ -97,7 +99,29 @@ type RegionSetInformer interface {
 	GetAdjacentRegions(region *RegionInfo) (*RegionInfo, *RegionInfo)
 	ScanRegions(startKey, endKey []byte, limit int) []*RegionInfo
 	GetRegionByKey(regionKey []byte) *RegionInfo
-	BatchScanRegions(keyRanges *KeyRanges, limit int) []*RegionInfo
+	BatchScanRegions(keyRanges *KeyRanges, opts ...BatchScanRegionsOptionFunc) ([]*RegionInfo, error)
```
rleungx, line 102: This is an internal method, maybe bool is enough.

okJiang: You didn't say that before, did you? 😭
https://github.com/tikv/pd/pull/8375#discussion_r1670120502

rleungx (commented):
LGTM

pkg/core/basic_cluster.go
rleungx: Both are ok to me.

pkg/core/region.go
```diff
@@ -1826,37 +1827,86 @@ func (r *RegionsInfo) ScanRegions(startKey, endKey []byte, limit int) []*RegionI
...
+		if len(keyRange.EndKey) > 0 && len(region.GetStartKey()) > 0 &&
+			bytes.Compare(region.GetStartKey(), keyRange.EndKey) >= 0 {
+			return false
+		}
+		if exceedLimit() {
+			return false
+		}
+		if len(lastRegion.GetEndKey()) > 0 && len(region.GetStartKey()) > 0 &&
+			bytes.Compare(region.GetStartKey(), lastRegion.GetEndKey()) > 0 {
+			err = multierr.Append(err, errs.ErrRegionNotAdjacent.FastGen(
```
you06, line 1884: Can this function accept parameter `outputMustContainAllKeyRange` and fast fail and return here?

you06 (commented):
This PR LGTM.

My only concern is whether also add this hole check for `ScanRegions`, though it's deprecated, but we still need to keep the compatibility with PD server without `BatchScanRegions` due to lightning needs to support earlier version PD server...

see pingcap/tidb#54341

There are some [compatibility code](https://github.com/tikv/client-go/blob/d73cc1ed6503925dfc7226e8d5677ceb4c2fd6f1/internal/locate/region_cache.go#L2200-L2202) in the current client-go, I know it's evil, but we need to keep a hole check function in the future client-go if `ScanRegions` keeps with possible holes.

Check holes for `ScanRegions` only in the client-side is also OK to me.

okJiang:
> Check holes for `ScanRegions` only in the client-side is also OK to me.

I prefer this~ Because `ScanRegions` has been marked deprecated.

pkg/core/region.go
okJiang: I will try to add it~

okJiang: I added the fast fail mechanism. PTAL again  @you06 cc @lance6716 
https://github.com/tikv/pd/pull/8375/commits/53f1e62aa3d55cedced9df810932d2485ef347f1

pkg/core/basic_cluster_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+package core
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMerge(t *testing.T) {
```
JmPotato, line 23:
```suggestion
func TestMergeKeyRanges(t *testing.T) {
```

JmPotato (approved):

JmPotato:
Please update the kvproto.

rleungx (approved):

PR #8376, opened 2024-07-08
tools/simulator: resolve simulator race

HuSharp:

Issue Number: Ref #8135

```commit-message
```

### Check List

Tests

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-simulator/simulator/client.go
+++ b/tools/pd-simulator/simulator/client.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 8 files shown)


tools/pd-simulator/simulator/client.go
```diff
@@ -50,7 +51,7 @@ type Client interface {
 }
 
 const (
-	pdTimeout             = time.Second
+	pdTimeout             = 3 * time.Second
```
lhy1024, line 54: Why change it?

HuSharp: to keep consistent with https://github.com/tikv/pd/blob/8215f3088c7373512909d60c9d0ae7d9a577091b/client/option.go#L27

lhy1024 (approved):

rleungx (approved):

PR #8387, opened 2024-07-11
rg/controller: use the default resource group if the requested one doesn't exist

### LINKED ISSUE #8388: Support fallback to the default resource group when the specified one does not exist
Part of https://github.com/pingcap/tidb/issues/54434 and https://github.com/pingcap/tidb/issues/54453.

To prevent interrupting the current connection using a dropped resource group, we should support fallback to the default resource group when the specified one does not exist.

JmPotato:

Issue Number: close #8388.

```commit-message
Fallback to use the default resource group if the requested one doesn't exist.
```

### Check List

Tests

- Unit test
- Integration test
- Manual test

image

### Release note

```release-note
None.
```


client/resource_group/controller/controller.go
```diff
@@ -433,6 +435,11 @@ func (c *ResourceGroupsController) tryGetResourceGroup(ctx context.Context, name
 		return nil, err
 	}
 	if group == nil {
+		// If the resource group does not exist and it's not the default resource group and fallback is true,
+		// try again with the default resource group.
+		if fallback && name != defaultResourceGroupName {
```
nolouch, line 440: Here path is after calling gRPC, there may many gRPC calls according to the request. maybe we need to rewrite it as default in the topper level.

JmPotato: I modified the logic here:

- When fallback is triggered, the default group controller will be used as an alias.
- When the actual group creation event is watched, it will replace the default.
- The cleanup that runs every 5 minutes will delete the fallback group controller.

client/resource_group/controller/controller.go
```diff
@@ -356,22 +357,27 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 						if err = proto.Unmarshal(item.Kv.Value, group); err != nil {
 							continue
 						}
-						if item, ok := c.groupsController.Load(group.Name); ok {
-							gc := item.(*groupCostController)
+						if gc, ok := c.loadGroupController(group.Name); ok {
```
nolouch, line 360: Should we reset the tombstone flag in here if the user deletes it and creates it again?

tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -402,8 +403,9 @@ func (suite *resourceManagerClientTestSuite) TestResourceGroupController() {
 		CPUMsCost:        1,
 	}
 
-	controller, _ := controller.NewResourceGroupController(suite.ctx, 1, cli, cfg, controller.EnableSingleGroupByKeyspace())
```
HuSharp, line 405: why remove `EnableSingleGroupByKeyspace`?

client/resource_group/controller/controller.go
JmPotato: Nice catch. Fixed.

tests/integrations/mcs/resourcemanager/resource_manager_test.go
JmPotato: Enabling it will prevent the controller from starting the watch channel, this test needs it.

client/resource_group/controller/controller.go
```diff
@@ -572,12 +597,16 @@ func (c *ResourceGroupsController) OnRequestWait(
...
 ) (*rmpb.Consumption, error) {
-	tmp, ok := c.groupsController.Load(resourceGroupName)
+	gc, ok := c.loadGroupController(resourceGroupName)
 	if !ok {
+		// If the resource group does not exist, use the default resource group.
+		if resourceGroupName != defaultResourceGroupName {
+			return c.OnResponse(defaultResourceGroupName, req, resp)
+		}
 		log.Warn("[resource group controller] resource group name does not exist", zap.String("name", resourceGroupName))
 		return &rmpb.Consumption{}, nil
```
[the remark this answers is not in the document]
HuSharp: When does it go to this logic to print this log? Is it possible to **NOT** have the default group?

HuSharp (approved):

nolouch (approved):
lgtm

PR #8402, opened 2024-07-16
rg/controller: optimize the fallback mechanism of the controller

### LINKED ISSUE #8388: Support fallback to the default resource group when the specified one does not exist

JmPotato:

Issue Number: close #8388.

```commit-message
This PR implements a more robust fallback method, which uses an independent default group controller for each tombstone group.
```

### Check List

Tests

- Unit test
- Integration test

Support multiple dynamic changes:

No interruption even after the GC cleanup:

### Release note

```release-note
None.
```


client/resource_group/controller/controller.go
```diff
@@ -476,14 +483,52 @@ func (c *ResourceGroupsController) tryGetResourceGroup(ctx context.Context, name
...
+		log.Warn("[resource group controller] create default resource group cost controller for tombstone failed",
+			zap.String("name", name), zap.Error(err))
+		// Directly delete the resource group controller if the default group controller cannot be created.
+		c.groupsController.Delete(name)
+		return
+	}
+	gc.tombstone.Store(true)
+	c.groupsController.Store(name, gc)
+	resourceGroupStatusGauge.DeleteLabelValues(name, name)
+	resourceGroupStatusGauge.WithLabelValues(name, defaultResourceGroupName).Set(1)
```
nolouch, line 527: Here are two names for compatibility. maybe the old `name` in the future. so it should be the same currently.

nolouch: How about setting the tombstone group to state 2. and then finally delete in `cleanUpResourceGroup`.

nolouch (commented):
rest lgtm

HuSharp (approved):

nolouch (approved):

PR #8405, opened 2024-07-16
schedule: fix split-merge-interval update

### LINKED ISSUE #8404: schedule: if aba modification occurs, split-merge-interval will not take effect
### What did you do?

test merge and split region

1. start a cluster while split-merge-interval is 1s
3. merge empty regions
4. set split-merge-interval to 1h
5. split some regions
6. again set split-merge-interval to 1s

merge empty region

### What did you see instead?

image

lhy1024:

Issue Number: Close #8404

```commit-message
```

### Check List

Tests

- Unit test
- Manual test

test will failed with master branch
```
--- FAIL: TestMergeCheckerTestSuite (1.51s)
    --- FAIL: TestMergeCheckerTestSuite/TestBasic (1.51s)
        /home/lhy1024/pd/pkg/schedule/checker/merge_checker_test.go:261: 
            	Error Trace:	/home/lhy1024/pd/pkg/schedule/checker/merge_checker_test.go:261
            	Error:      	Expected value not to be nil.
            	Test:       	TestMergeCheckerTestSuite/TestBasic
FAIL
```

### Release note

```release-note
None.
```


rleungx (approved):

pkg/schedule/checker/merge_checker.go
```diff
@@ -88,13 +90,16 @@ func (m *MergeChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		return nil
 	}
 
+	// update the split cache.
+	// It must be called before the following merge checker logic.
+	m.splitCache.UpdateTTL(m.conf.GetSplitMergeInterval())
```
rleungx, line 95: BTW, will it have a risk? If the heartbeat is not send in time?

pkg/schedule/checker/merge_checker_test.go
```diff
@@ -56,6 +56,7 @@ func TestMergeCheckerTestSuite(t *testing.T) {
 
 func (suite *mergeCheckerTestSuite) SetupTest() {
 	cfg := mockconfig.NewTestOptions()
+	gcInterval = 100 * time.Millisecond
```
okJiang, line 59: Need revert it

okJiang (approved):
lgtm

pkg/schedule/checker/merge_checker_test.go
lhy1024: `gcInterval` is 1 minute by default, which is too long for test.

pkg/schedule/checker/merge_checker.go
lhy1024: At the same time, we wait until 90% of the heartbeats are collected before checking.

rleungx: There still be 10% and merge check may create lots of scheduling?

pkg/schedule/checker/merge_checker_test.go
lhy1024: get

pkg/schedule/checker/merge_checker.go
lhy1024: The behavior is similar in this case. For example, if split-merge-interval is set to 1s.
In the past, it would pass the recently started check and then run `UpdateTTL`
Now, it will run `UpdateTTL` first and then pass the recently started check.

Perhaps we can add a longer time before recently started check, such as `expireTime := m.startTime.Add(m.conf.GetSplitMergeInterval()+5*time.Minute)`

nolouch (approved):

rleungx (approved):

PR #8409, opened 2024-07-17
*: enable gocritic.exitAfterDefer

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/4322

ref https://www.quasilyte.dev/blog/post/log-fatal-vs-log-panic/

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/pd_service_discovery_test.go
```diff
@@ -107,10 +107,10 @@ func newTestServer(isLeader bool) *testServer {
 func (s *testServer) run() {
 	lis, err := net.Listen("tcp", s.addr)
 	if err != nil {
-		log.Fatalf("failed to serve: %v", err)
+		log.Panicf("failed to serve: %v", err)
```
HuSharp, line 110: maybe can `log.Panic(zap.error())` ?

HuSharp (approved):

cmd/pd-server/main.go
```diff
@@ -254,7 +254,7 @@ func start(cmd *cobra.Command, args []string, services ...string) {
 
 	err = join.PrepareJoinCluster(cfg)
 	if err != nil {
-		log.Fatal("join meet error", errs.ZapError(err))
+		log.Panic("join meet error", errs.ZapError(err))
```
rleungx, line 257: I have some questions:
1. The behavior is different from the previous 
2. Will it be printed out if we set the log level to fatal?
3. What if the panic has a recover hook?

okJiang: 1. Yes
2. Yes, the log will be printed when log level is fatal
3. If exist a recover hook, the panic will be caught.

> from: https://www.quasilyte.dev/blog/post/log-fatal-vs-log-panic/
> The log.Panic vs log.Fatal is essentially panic vs os.Exit(1). The latter will not let deferred calls to run.

okJiang:
Since some users might rely on the "fatal" keyword to define alarm rules, we won't make any changes to this.

Issue #8379, opened 2024-07-09
Standardize the use of scheduler name/type and checker name/type


okJiang:
# Target

- Use a unified naming convention for schedulers/checkers
- Standardize the concepts of name and type of scheduler and converge their usage

# Naming summary

- xxxx_checker this underscore format only appears in the checker, scheduler does not. It is recommended to unify it into the format of xxxx-checker
- Xx-y-scheduler (schedulerName) and xx-yy (schedulerType) are unified into xx-y-scheduler. Reason
  - The interaction with users uses the former. If the latter is used, it will destroy the consistency of user usage or generate a lot of additional work (do a lot of converts) to maintain user consistency, increasing complexity
  - The former has more usage scenarios in the code than the latter
  - The former will be persistent, the latter will not
  - Unified with xxxx-checker
- Operator desc
  - Checker uses xxxx-checker.
  - Scheduler uses xxxx-yyyy. It will be replaced with xxxx-yyy-scheduler. It affects the printing of logs

# checker

- xxxx_checker(checkerName)
- xxxx-checker(checkerType)

# scheduler

- schedulerName is in the format of xx-y-scheduler
- schedulerType is in the format of xx-yy
- In addition, there are two special ones
```golang
    // HotRegionName is balance hot region scheduler name.
    HotRegionName = "balance-hot-region-scheduler"
    // HotRegionType is balance hot region scheduler type.
    HotRegionType          = "hot-region"
        // ScatterRangeType is scatter range scheduler type
    ScatterRangeType = "scatter-range"
    // ScatterRangeName is scatter range scheduler name
    ScatterRangeName = "scatter-range"
```

- xx-yy-scheduler(schedulerName)
- xx-yy(schedulerType)

# Deep scheduler name/type

For schedulers, there is indeed a difference between name and type, but in the current code, these two are completely confused, making it difficult for readers to understand them. This section aims to completely distinguish them and give a clear definition. First

## name

There are three ways to name a name

### WithXxxxxxName

These three schedulers allow you to specify their names at creation time.
- balanceLeaderScheduler, not used yet
- balanceRegionScheduler, will specify the name as scatter-range-region
- balanceWitnessScheduler, will specify the name as scatter-range-leader

### ScatterRange

Scatter-range-scheduler name is named by the key-range entered by the user, named scatter-range- {key-range} . Later changed to scatter-range-scheduler- {key-range}

### Other

The names of other schedulers can be directly equivalent to type because they are constant

## Name vs type

[...]

PR #8416, opened 2024-07-18
scheduler: Replace the input parameter of AddScheduler with types.CheckerSchedulerType.

okJiang:

```commit-message
1. Replace the input parameter of AddScheduler with types.CheckerSchedulerType.
2. Simplified the logic of AddScheduler by using a temporary map.
3. Remove scope's redundant definition and use types.CheckerSchedulerType.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


server/api/scheduler.go
```diff
@@ -137,64 +107,39 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
...
+				h.r.JSON(w, http.StatusInternalServerError, err.Error())
+				return
+			}
+			log.Info("update scheduler", zap.String("scheduler-name", name), zap.Uint64("store-id", uint64(storeID)))
 			h.r.JSON(w, http.StatusOK, "The scheduler has been applied to the store.")
 			return
 		}
-	case schedulers.ShuffleLeaderName:
-		if err := h.AddShuffleLeaderScheduler(); err != nil {
+		if err != nil && !errors.ErrorEqual(err, errs.ErrSchedulerNotFound.FastGenByArgs()) {
```
okJiang, line 129:
```suggestion
		if err != nil {
```

do not need check

pkg/schedule/type/type.go
```diff
@@ -70,3 +70,46 @@ const (
 	// LabelScheduler is label scheduler name.
 	LabelScheduler CheckerSchedulerType = "label-scheduler"
 )
+
+// SchedulerTypeCompatibleMap temporarily exists for compatibility.
```
rleungx, line 74: I suggest adding more explanation about when we need to consider the compatibility.

okJiang: I don't think there is a need to explain in the code now. Its current role is only to do a layer of conversion when the user adds a scheduler. If it is used in more places in the future, or permanently exists in the code, I will add a more accurate description to it.

okJiang: ping @rleungx

rleungx: I just wondering when we need to consider the compatibility.

rleungx (approved):

pkg/schedule/type/type.go
okJiang: This is an intermediate pr, it may be deleted in the subsequent pr, if it stays, I will add more comments

server/handler.go
```diff
@@ -186,7 +185,8 @@ func (h *Handler) GetAllRequestHistoryHotRegion(request *HistoryHotRegionsReques
 }
 
 // AddScheduler adds a scheduler.
-func (h *Handler) AddScheduler(name string, args ...string) error {
+func (h *Handler) AddScheduler(tp types.CheckerSchedulerType, args ...string) error {
+	name := types.SchedulerTypeCompatibleMap[tp]
```
okJiang, line 189: This line may be removed in the subsequent pr.

okJiang:
Could you take a look? @HuSharp @JmPotato

server/api/scheduler.go
```diff
@@ -81,48 +85,21 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
...
-			return
+	tp, ok := types.SchedulerStr2Type[name]
+	if !ok {
+		h.r.JSON(w, http.StatusBadRequest, "unknown scheduler")
+		return
+	}
+	var args []string
+	collector := func(v string) {
+		if args == nil {
+			args = []string{}
```
HuSharp, line 96: maybe can
```suggestion
```

HuSharp (commented):
rest LGTM

server/api/scheduler.go
okJiang: Do you mean this?

```go
        args := []string{}
	collector := func(v string) {
		args = append(args, v)
	}
```

This is incompatible with before because when there are no args, args should be a null pointer, otherwise unit tests will fail.

HuSharp (approved):

Issue #8389, opened 2024-07-12
TestReloadLabel is flaky


okJiang:
### Which jobs are failing

```
--- FAIL: TestReloadLabel (339.90s)
panic: exit status 2 [recovered]
	panic: exit status 2

goroutine 357 [running]:
testing.tRunner.func1.2({0x1ad2aa0, 0xc00055bb40})
	/usr/local/go/src/testing/testing.go:1545 +0x3f7
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1548 +0x716
panic({0x1ad2aa0?, 0xc00055bb40?})
	/usr/local/go/src/runtime/panic.go:920 +0x270
github.com/tikv/pd/tests/integrations/realcluster.restartTiUP()
	/home/jenkins/agent/workspace/tikv/pd/pull_integration_realcluster_test/pd/tests/integrations/realcluster/reboot_pd_test.go:31 +0xb2
github.com/tikv/pd/tests/integrations/realcluster.TestReloadLabel(0xc0001a3d40)
	/home/jenkins/agent/workspace/tikv/pd/pull_integration_realcluster_test/pd/tests/integrations/realcluster/reboot_pd_test.go:75 +0x4d8
testing.tRunner(0xc0001a3d40, 0x1bcf890)
	/usr/local/go/src/testing/testing.go:1595 +0x262
created by testing.(*T).Run in goroutine 1
	/usr/local/go/src/testing/testing.go:1648 +0x846
FAIL	github.com/tikv/pd/tests/integrations/realcluster	340.028s
````

### CI link

https://do.pingcap.net/jenkins/blue/organizations/jenkins/tikv%2Fpd%2Fpull_integration_realcluster_test/detail/pull_integration_realcluster_test/290/pipeline/

okJiang:
I only seem to see "start to restart TiUP" printed out. I didn't see any logs from the running restart, so I can't pinpoint where the exit 2 occurred.

PR #8427, opened 2024-07-22
test: add Stdout to TestReloadLabel

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/8389#issuecomment-2242055199

```commit-message
```

### Check List

Tests

complete test

### Release note

```release-note
None.
```


tests/integrations/realcluster/reboot_pd_test.go
```diff
@@ -26,6 +27,8 @@ import (
 func restartTiUP() {
 	log.Info("start to restart TiUP")
 	cmd := exec.Command("make", "deploy")
+	cmd.Stdout = os.Stdout
```
rleungx, line 30: What is the current output?

okJiang: `exec.Command` creates a child process, so it will output on the child process's terminal.

HuSharp (approved):

rleungx (approved):

Issue #7963, opened 2024-03-21
checker: make patrol region sooner


lhy1024:
## Enhancement Task

![image](https://github.com/tikv/pd/assets/19542290/f46c4d9e-bb9f-4b83-aa11-b05f21c760e2)

We are now using one thread to scan all the regions and when the number of regions is high, it will always have more time without scheduling generation.

To enhance efficiency, we might consider implementing concurrent scanning or establishing a cache to prioritize the scanning of previously encountered regions that can generate operators but the key range is limited because of limit.

PR #8428, opened 2024-07-22
schedule: move the logic of patrol region from coordinator to checkers

lhy1024:

Issue Number: Ref #7963

```commit-message
```

1. move the logic of patrol region from coordinator to checkers
2. introduce PatrolRegionContext
3. modify some comments

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/checker/checker_controller.go
```diff
@@ -181,6 +330,41 @@ func (c *Controller) RemovePriorityRegions(id uint64) {
 	c.priorityInspector.RemovePriorityRegion(id)
 }
 
+// CheckSuspectRanges would pop one suspect key range group
+// The regions of new version key range and old version key range would be placed into
+// the suspect regions map
+func (c *Controller) CheckSuspectRanges() {
```
HuSharp, line 336: Do we need to check `if c == nil ` in `func (c *Coordinator) checkSuspectRanges() {`?

pkg/schedule/checker/checker_controller.go
```diff
@@ -67,10 +84,119 @@ func NewController(ctx context.Context, cluster sche.CheckerCluster, conf config
...
+		key, regions = c.checkRegions(key)
+		if len(regions) == 0 {
+			continue
+		}
+		// Updates the label level isolation statistics.
+		c.cluster.UpdateRegionsLabelLevelStats(regions)
+		if len(key) == 0 {
+			c.patrolRegionContext.roundUpdateMetrics()
+		}
+		failpoint.Inject("break-patrol", func() {
```
HuSharp, line 128: It looks like the failpoint name convention is `Camel case`, can we modify it?

HuSharp (commented):
rest LGTM

pkg/schedule/checker/checker_controller.go
lhy1024: It is always initialized in NewCoordinator

HuSharp: Got it, just noticed the `c==nil` check in `GetPatrolRegionsDuration`. Seems like its an API called.

HuSharp (approved):

pkg/schedule/checker/checker_controller.go
```diff
@@ -233,3 +417,39 @@ func (c *Controller) GetPauseController(name string) (*PauseController, error) {
 		return nil, errs.ErrCheckerNotFound.FastGenByArgs()
 	}
 }
+
+// PatrolRegionContext is used to store the context of patrol regions.
```
rleungx, line 421: Do we still need it?

lhy1024: it will be used in the next pr, it may be more suitable to move it to the next pr?

rleungx: Should the fields be part of the checker controller? Is it necessary to wrap it?

lhy1024: I think it is needed. It will contains more value and config https://github.com/tikv/pd/blob/ae0778f4d80df70f415f8bf60d87b39601fc6471/pkg/schedule/checker/checker_controller.go#L465

pkg/schedule/checker/checker_controller.go
```diff
@@ -67,10 +84,119 @@ func NewController(ctx context.Context, cluster sche.CheckerCluster, conf config
...
 		pendingProcessedRegions: pendingProcessedRegions,
 		suspectKeyRanges:        cache.NewStringTTL(ctx, time.Minute, 3*time.Minute),
+		patrolRegionContext:     &PatrolRegionContext{},
+	}
+}
+
+// PatrolRegions is used to scan regions.
+// The checkers will check these regions to decide if they need to do some operations.
+func (c *Controller) PatrolRegions() {
+	c.patrolRegionContext.init(c.cluster)
```
rleungx, line 94:
```suggestion
        patrolInterval := cluster.GetCheckerConfig().GetPatrolRegionInterval()
	c.patrolRegionContext.init(patrolInterval)
```

pkg/schedule/checker/checker_controller.go
```diff
@@ -67,10 +84,119 @@ func NewController(ctx context.Context, cluster sche.CheckerCluster, conf config
...
+		failpoint.Inject("breakPatrol", func() {
+			failpoint.Break()
+		})
+	}
+}
+
+func (c *Controller) updateTickerIfNeeded(ticker *time.Ticker) {
+	// Note: we reset the ticker here to support updating configuration dynamically.
+	newInterval := c.cluster.GetCheckerConfig().GetPatrolRegionInterval()
+	if c.patrolRegionContext.interval != newInterval {
```
rleungx, line 137: Why not keep it the same as before, just reset it every time?

pkg/schedule/checker/checker_controller.go
```diff
@@ -181,6 +330,41 @@ func (c *Controller) RemovePriorityRegions(id uint64) {
...
+	for {
+		select {
+		case <-c.ctx.Done():
+			return
+		case <-ticker.C:
+			keyRange, success := c.PopOneSuspectKeyRange()
+			if !success {
+				continue
+			}
+			limit := 1024
```
rleungx, line 348: use a constant

rleungx (approved):

pkg/schedule/checker/checker_controller.go
rleungx: init doesn't use it.

pkg/schedule/checker/checker_controller.go
lhy1024: Maybe move it to patrolRegionContext?

pkg/schedule/checker/checker_controller.go
```diff
@@ -233,3 +417,39 @@ func (c *Controller) GetPauseController(name string) (*PauseController, error) {
 		return nil, errs.ErrCheckerNotFound.FastGenByArgs()
 	}
 }
+
+// PatrolRegionContext is used to store the context of patrol regions.
+type PatrolRegionContext struct {
+	// config
+	interval time.Duration
+	// status
+	patrolRoundStartTime time.Time
```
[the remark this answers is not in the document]
lhy1024: I tried to add a new commit, what do you think? Is it better to move both `updateTickerIfNeeded` and `roundUpdateMetrics` to the `patrolRegion` function, would that make it too long?

PR #1733, opened 2019-09-05
*: supports http handler for each scheduler

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve? 

- [x] Supports the HTTP handler router
- [x] Compatible with the previous option to create Scheduler
- [x] Add the handler for the scheduler if needed
- [x] Compatible with the previous option to delete Scheduler
- [x] Test

### Check List 

Tests 

 - Unit test

PR #8440, opened 2024-07-25
scheduler: GetType() returns types.CheckerSchedulerType directly

okJiang:

```commit-message
- GetType() returns types.CheckerSchedulerType directly
- keep the compatibility in `PersistOptions` and `PersistConfig“
- wrap `operator.OperatorLimitCounter.WithLabelValues().Inc()` to func `IncOperatorLimitCounter` 
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ ifeq ($(RUN_CI), 1)
 	RELEASE_VERSION := None
 endif
 
-LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
+LDFLAGS = -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDBuildTS=$(shell date -u '+%Y-%m-%d %I:%M:%S')"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDGitHash=$(shell git rev-parse HEAD)"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDGitBranch=$(shell git rev-parse --abbrev-ref HEAD)"
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ ifeq ($(RUN_CI), 1)
 	RELEASE_VERSION := None
 endif
 
-LDFLAGS = -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
+LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDBuildTS=$(shell date -u '+%Y-%m-%d %I:%M:%S')"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDGitHash=$(shell git rev-parse HEAD)"
 LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDGitBranch=$(shell git rev-parse --abbrev-ref HEAD)"
--- a/pkg/schedule/checker/replica_checker.go
+++ b/pkg/schedule/checker/replica_checker.go
@@ -61,6 +61,7 @@ func (*ReplicaChecker) Name() string {
 	return types.ReplicaChecker.String()
 }
 
+// GetType return ReplicaChecker's type.
 func (*ReplicaChecker) GetType() types.CheckerSchedulerType {
 	return types.ReplicaChecker
 }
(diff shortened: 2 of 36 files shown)


pkg/schedule/schedulers/evict_leader.go
```diff
@@ -395,7 +382,7 @@ func scheduleEvictLeaderOnce(name, typ string, cluster sche.SchedulerCluster, co
 		for _, t := range targets {
 			targetIDs = append(targetIDs, t.GetID())
 		}
-		op, err := operator.CreateTransferLeaderOperator(typ, cluster, region, target.GetID(), targetIDs, operator.OpLeader)
+		op, err := operator.CreateTransferLeaderOperator(name, cluster, region, target.GetID(), targetIDs, operator.OpLeader)
```
okJiang, line 385: Using name instead of type for desc is ok

pkg/schedule/checker/replica_checker.go
```diff
@@ -61,6 +61,10 @@ func (*ReplicaChecker) Name() string {
 	return types.ReplicaChecker.String()
 }
 
+func (*ReplicaChecker) GetType() types.CheckerSchedulerType {
```
JmPotato, line 64:
```suggestion
// GetType ...
func (*ReplicaChecker) GetType() types.CheckerSchedulerType {
```

pkg/schedule/operator/metrics.go
```diff
@@ -82,10 +84,14 @@ var (
...
-	prometheus.MustRegister(OperatorLimitCounter)
+	prometheus.MustRegister(operatorLimitCounter)
 	prometheus.MustRegister(OperatorExceededStoreLimitCounter)
 	prometheus.MustRegister(operatorCounter)
 	prometheus.MustRegister(operatorDuration)
 	prometheus.MustRegister(operatorSizeHist)
 	prometheus.MustRegister(storeLimitCostCounter)
 }
+
+func IncOperatorLimitCounter(typ types.CheckerSchedulerType, kind OpKind) {
```
JmPotato, line 95:
```suggestion
// IncOperatorLimitCounter ...
func IncOperatorLimitCounter(typ types.CheckerSchedulerType, kind OpKind) {
```

pkg/schedule/checker/replica_checker.go
okJiang: Filed an issue https://github.com/tikv/pd/issues/8458#issue-2437331188

And fixed it https://github.com/tikv/pd/pull/8440/commits/0d2676fc4c3f82a3a9a8123e9c2d2529aaa2e3b5

pkg/schedule/config/config_provider.go
```diff
@@ -49,9 +50,9 @@ type SchedulerConfigProvider interface {
 	SetSchedulingAllowanceStatus(bool, string)
 	GetStoresLimit() map[uint64]StoreLimitConfig
 
-	IsSchedulerDisabled(string) bool
-	AddSchedulerCfg(string, []string)
-	RemoveSchedulerCfg(string)
+	IsSchedulerDisabled(types.CheckerSchedulerType) bool
+	AddSchedulerCfg(types.CheckerSchedulerType, []string)
```
nolouch, line 54: we can consider removing this interface.  here is some background:

- v2.x all scheduler configs are stored in the global config `Config.SchedulerConfig` 
- v3.x let the scheduler has it's own config, that's store in another key/value in etcd https://github.com/tikv/pd/pull/1733/

To ensure compatibility, both configurations are retained, and we can now clean up the old one. However, there is one problem that needs to be solved: the default scheduler state needs special handling.

okJiang: It is better to do it in a new pr. I filed an issue to track it. https://github.com/tikv/pd/issues/8474

okJiang: BTW, do you mean all this interface or these two function?

Makefile
```diff
@@ -60,7 +60,7 @@ ifeq ($(RUN_CI), 1)
 	RELEASE_VERSION := None
 endif
 
-LDFLAGS += -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
+LDFLAGS = -X "$(PD_PKG)/pkg/versioninfo.PDReleaseVersion=$(RELEASE_VERSION)"
```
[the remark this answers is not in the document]
okJiang: The local LDFLAGS variable is inherited here, and since I don't know where this variable is set locally, an error is reported here.

```shell
# In my local env
echo $LDFLAGS  
-L/opt/homebrew/opt/protobuf@3/lib
```

img_v3_02d5_bb52df28-2704-43a7-915f-90a93dedf6bg

So, I think it does not need inherited here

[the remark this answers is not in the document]
rleungx: We should inherit LDFLAGS and keep it the same as the other components.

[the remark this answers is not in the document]
okJiang: ok, I reverted it

nolouch (approved):

JmPotato (approved):

PR #8445, opened 2024-07-26
*: enlarge the default value of `max-merge-region-size`.

### LINKED ISSUE #8484: region: larger region size for the compatibility to larger clusters.
## Development Task
This issue is a linked issue to https://github.com/tikv/tikv/issues/17309

LykxSassinator:
This pr is used to enlarge the region size from default value `96MB` to `256MB`, compatible to the requirement of the stability of large cluster.

Issue Number: close #8484, ref https://github.com/tikv/tikv/issues/17309

```commit-message
This pr is used to enlarge the region size from default value `96MB` to `256MB`, compatible to the requirement of the stability of large cluster.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Enlarge the region size from `96MB` to `256MB` as default.
```

### DIFF
--- a/pkg/schedule/config/config.go
+++ b/pkg/schedule/config/config.go
@@ -27,10 +27,15 @@ import (
 
 const (
 	// DefaultMaxReplicas is the default number of replicas for each region.
-	DefaultMaxReplicas            = 3
-	defaultMaxSnapshotCount       = 64
-	defaultMaxPendingPeerCount    = 64
-	defaultMaxMergeRegionSize     = 96
+	DefaultMaxReplicas         = 3
+	defaultMaxSnapshotCount    = 64
+	defaultMaxPendingPeerCount = 64
+	// defaultMaxMergeRegionSize is the default maximum size of region when regions can be merged.
+	// After https://github.com/tikv/tikv/issues/17309, the default value is enlarged from 20 to 54,
+	// to make it compatible with the default value of region size of tikv. To avoid triggering
+	// too many merge operations when upgrading, the old value is kept in the configuration file.
+	oldDefaultMaxMergeRegionSize  = 20
+	defaultMaxMergeRegionSize     = 54
 	defaultLeaderScheduleLimit    = 4
 	defaultRegionScheduleLimit    = 2048
 	defaultWitnessScheduleLimit   = 4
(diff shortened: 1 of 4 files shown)


pkg/schedule/config/config.go
```diff
@@ -30,7 +30,7 @@ const (
 	DefaultMaxReplicas            = 3
 	defaultMaxSnapshotCount       = 64
 	defaultMaxPendingPeerCount    = 64
-	defaultMaxMergeRegionSize     = 20
+	defaultMaxMergeRegionSize     = 96
```
nolouch, line 33: We need to add more comments about which version changed and the related issue.

rleungx: Will it cause lots of merge operations?

zhangjinpeng87: +1, 96 is too aggressive, it might cause a lot of merging after upgrading. We increase the default region size from 96MB to 256MB in TiKV side, it is 2.67x larger than before. But here 20 -> 96 is nearly 5x larger than before, probably 40MB is better.

LykxSassinator: How about using `54MB`(=> `20MB * 256 / 96`) as the recommended value ?

LykxSassinator: > Will it cause lots of merge operations when upgrading?

Nope, the old value of `max-merge-region-size` is already persisted in the cluster and it will be reloaded after upgrading. Only the newly created cluster will use the new default value.

rleungx: Oh, right. But for those existing clusters, do we still use the old max-merge-region-size and let the user change the config? If so, then LGTM.

LykxSassinator: IMO, it should be.

Since each user's cluster may have a unique self-defined `max-merge-region-size`, manual adjustments should be checked and executed if the user requires it.

rleungx (approved):

zhangjinpeng87 (approved):
LGTM

niubell (approved):

PR #8455, opened 2024-07-29
*: replace part of path with filepath

rleungx:

Issue Number: Ref #8475

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/grpcutil/grpcutil_test.go
+++ b/pkg/utils/grpcutil/grpcutil_test.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"os"
 	"os/exec"
-	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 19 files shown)


pkg/utils/grpcutil/grpcutil_test.go
```diff
@@ -41,9 +41,9 @@ func TestToTLSConfig(t *testing.T) {
 
 	re := require.New(t)
 	tlsConfig := TLSConfig{
-		KeyPath:  path.Join(certPath, "pd-server-key.pem"),
-		CertPath: path.Join(certPath, "pd-server.pem"),
-		CAPath:   path.Join(certPath, "ca.pem"),
+		KeyPath:  filepath.Join(certPath, "pd-server-key.pem"),
```
HuSharp, line 44: Why change only `a part of` it? It looks like there's still quite a bit of `path` used

rleungx: some of them are etcd keys which doesn't rely on os

HuSharp: got it.

HuSharp (approved):

tools/pd-ut/ut.go
```diff
@@ -746,13 +746,13 @@ func buildTestBinaryMulti(pkgs []string) ([]byte, error) {
 
 	// go test --exec=xprog --tags=tso_function_test,deadlock -vet=off --count=0 $(pkgs)
 	// workPath just like `/pd/tests/integrations`
-	xprogPath := path.Join(workDir, "bin/xprog")
+	xprogPath := filepath.Join(workDir, "bin/xprog")
```
HuSharp, line 749: If we consider different operating systems, maybe`bin/xprog` also need to be changed?

nolouch (approved):

Issue #8432, opened 2024-07-23
client: Support TSO RPC Parallelizing


MyonKeminta:
## Development Task

- The main tracking issue in TiDB repo: https://github.com/pingcap/tidb/issues/54960

We used to find that in some OLTP workloads where the QPS is high and the queries are simple, the TSO Wait duration usually become a significant portion of the total duration of queries. In TiDB, TSO loading is already made concurrent with some other works such as compiling. In cases that the queries are simple, it would be hard to further optimize it by making it concurrent with more phases of the SQL execution. But we found a practical way to optimize it is to do it from the TSO client.

Currently, a TSO client object has a goroutine that collects `GetTS` (and `GetTSAsync`) calls (`tsoRequest`s) as a batch, send it to PD, wait for the response, and dispatch the results to these `tsoRequests`, serially. As a result, each `GetTS` calls may need to spend up to 1x TSO RPC time to wait for being collected to the next batch.

Considering the case that PD's TSO allocator is not the bottle neck and can deal with more TSO requests (so that the majority part of TSO RPC's time cost is on the network), we find that it's possible to start collecting the next batch and send it before receiving the response of the previous batch. So that each `GetTS` call needs to wait for less time to be batched, and gets a shorter total duration.

So this is an approach that reduces the duration of `GetTS` & `GetTSAsync - Wait` at the expense of higher TSO RPC OPS and higher pressure to PD. It's not suitable to be enabled by default, but we can provide such an option when the TSO Wait duration becomes a problem.

### Subtasks

Side changes:

- [x] Merge the two `xxxTSOStream` types  so that the error handling and metrics reporting logic for PD server deployment and TSO service deployment can be reused. https://github.com/tikv/pd/pull/8433
- [x] Split sending and receiving part of `tsoStream` into separated goroutines
- [ ] Let `tsoDispatcher` support batching according to estimated TSO RPC duration from `tsoStream`

PR #8483, opened 2024-08-02
client: Make tsoStream receives asynchronously

MyonKeminta:

Issue Number: Ref #8432

```commit-message
client: Make tsoStream receives asynchronously. This makes it possible to allow the tsoDispatcher send multiple requests and wait for their responses concurrently. 
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/tso_batch_controller.go
```diff
@@ -19,53 +19,87 @@ import (
...
+func (tbc *tsoBatchController) fetchPendingRequests(ctx context.Context, tsoRequestCh <-chan *tsoRequest, tokenCh chan struct{}, maxBatchWaitInterval time.Duration) (errRet error) {
+	var tokenAcquired bool
+	defer func() {
+		if errRet != nil {
+			// Something went wrong when collecting a batch of requests. Release the token and cancel collected requests
+			// if any.
+			if tokenAcquired {
+				tokenCh <- struct{}{}
+			}
+			if tbc.collectedRequestCount > 0 {
```
JmPotato, line 59: I think we could put this check inside `finishCollectedRequests` to make it more general.

client/tso_dispatcher.go
```diff
@@ -91,24 +99,27 @@ func newTSODispatcher(
...
+	tsoRequestCh := make(chan *tsoRequest, maxBatchSize*2)
 	failpoint.Inject("shortDispatcherChannel", func() {
-		tsoBatchController = newTSOBatchController(
-			make(chan *tsoRequest, 1),
-			maxBatchSize,
-		)
+		tsoRequestCh = make(chan *tsoRequest, 1)
 	})
+
+	tokenCh := make(chan struct{}, 64)
```
JmPotato, line 107: Better add a const to explain the underlying meaning of this number `64`.

client/tso_stream.go
```diff
@@ -172,51 +178,252 @@ func (s tsoTSOStreamAdapter) Recv() (tsoRequestResult, error) {
...
+// we instead use the `batchedRequestsNotifier` as the abstraction, and accepts generic type instead of dynamic interface
+// type.
 type tsoStream struct {
 	serverURL string
 	// The internal gRPC stream.
 	//   - `pdpb.PD_TsoClient` for a leader/follower in the PD cluster.
 	//   - `tsopb.TSO_TsoClient` for a primary/secondary in the TSO cluster.
 	stream grpcTSOStreamAdapter
+	// An identifier of the tsoStream object for metrics reporting and diagnosing.
+	streamID string
```
JmPotato, line 202: Is it possible to attach this ID to each TSO request so we can know which stream a TSO request is dispatched to and put some debug info around it as well?

client/tso_batch_controller.go
MyonKeminta: Looks not necessary as `finishCollectedRequests` is internally a loop on `collectedRequestCount`. I'll just remove this condition.

client/tso_stream.go
MyonKeminta: I added it to the `tsoRequest` but it's for now not used. Do have any idea about how to add the new debug info you mentioned?

cfzjywxk:
@JmPotato @lhy1024 PTAL, thx

client/tso_stream.go
```diff
@@ -172,51 +178,254 @@ func (s tsoTSOStreamAdapter) Recv() (tsoRequestResult, error) {
...
+	streamStateIdle int32 = iota
+	streamStateSending
+	streamStateClosing
+)
+
+var streamIDAlloc atomic.Int32
+
+const invalidStreamID = "<invalid>"
+
+// TODO: Pass a context?
```
lhy1024, line 227: Will we pass it?

client/tso_stream.go
JmPotato: I was thinking about whether we could add fields to `TsoRequest`, so that the client side could also get the stream ID that processes its own TSO request, but considering that we would need to modify kvproto, it's okay if we don't do this in this PR.

client/tso_dispatcher.go
```diff
@@ -91,24 +99,29 @@ func newTSODispatcher(
...
+	tsoRequestCh := make(chan *tsoRequest, maxBatchSize*2)
 	failpoint.Inject("shortDispatcherChannel", func() {
-		tsoBatchController = newTSOBatchController(
-			make(chan *tsoRequest, 1),
-			maxBatchSize,
-		)
+		tsoRequestCh = make(chan *tsoRequest, 1)
 	})
+
+	// A large-enough capacity to hold maximum concurrent RPC requests. In our design, the concurrency is at most 16.
```
lhy1024, line 107: Will it be changed in the future?

MyonKeminta: I don't have such plan. As the more concurrent RPC there is, the less we can benefit from each additional concurrent RPC, it's not a good idea to allow too many concurrent RPCs for each stream. In our current plan, 16 is the maximum.

client/tso_stream_test.go
```diff
@@ -0,0 +1,333 @@
+// Copyright 2024 TiKV Project Authors.
```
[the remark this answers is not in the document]
MyonKeminta: Do you mean mocking the internal implementation of the stream and benching the overhead of the `tsoStream` layer?

client/tso_stream.go
MyonKeminta: I forgot this. I'll have a try.

client/tso_batch_controller.go
```diff
@@ -19,53 +19,85 @@ import (
...
+			// if any.
+			if tokenAcquired {
+				tokenCh <- struct{}{}
+			}
+			tbc.finishCollectedRequests(0, 0, 0, invalidStreamID, errRet)
+		}
+	}()
+
+	// Wait until BOTH the first request and the token have arrived.
+	// TODO: `tbc.collectedRequestCount` should never be non-empty here. Consider do assertion here.
```
lhy1024, line 64: There is also a TODO

client/tso_stream_test.go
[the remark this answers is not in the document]
lhy1024: I'd like to add a bench to compare the performance change before and after introducing this pr.

client/tso_batch_controller.go
MyonKeminta: This assertion should be true in my understanding. Is it true? If so, I can change it to an assertion here.

The assignment statement in the next line exists before my change. If I understood it right, it didn't make much sense here, since if it's non-zero and we set it zero, it might mean that some requests will be dropped without notifying their caller and leave them waiting forever.

client/tso_stream.go
```diff
@@ -172,51 +178,255 @@ func (s tsoTSOStreamAdapter) Recv() (tsoRequestResult, error) {
...
+	// Check if the stream is closing or closed, in which case no more requests should be put in.
+	// Note that the prevState should be restored very soon, as the receiver may check
+	prevState := s.state.Swap(streamStateSending)
+	switch prevState {
+	case streamStateIdle:
+		// Expected case
+	case streamStateClosing:
+		s.state.Store(prevState)
+		err := s.GetRecvError()
+		log.Info("sending to closed tsoStream", zap.Error(err))
```
JmPotato, line 273: Better attach the stream ID to all logs produced by a `*tsoStream`.

client/tso_dispatcher.go
```diff
@@ -271,55 +303,87 @@ tsoBatchLoop:
...
+		err = td.processRequests(stream, dc, batchController, done)
 		// If error happens during tso stream handling, reset stream and run the next trial.
-		if err != nil {
+		if err == nil {
+			// If the request is started successfully, the `batchController` will be put back to the pool when the
+			// request is finished (either successful or not). In this case, set the `batchController` to nil so that
+			// another one will be fetched from the pool.
+			// Otherwise, the `batchController` won't be processed in other goroutines concurrently, and it can be
+			// reused in the next loop safely.
+			batchController = nil
```
JmPotato, line 339: The comment above said we have to set this `batchController` to nil either successful or not, do we need to set it to nil even if the `err` is not nil?

MyonKeminta:
@lhy1024 I've done some benchmarks with the stream implemented by channels internally to show the overhead of `tsoStream` and `tsoDispatcher`. Please see if it's acceptable.

Before this PR: (code in #8618 , which is branched from the merge-base of this PR and master)

```text
goos: linux
goarch: amd64
pkg: github.com/tikv/pd/client
cpu: Intel(R) Xeon(R) Silver 4214R CPU @ 2.40GHz
BenchmarkTSODispatcherHandleRequests-28           140149              7508 ns/op
BenchmarkTSOStreamSendRecv-28                    2788069               425.8 ns/op
PASS
ok      github.com/tikv/pd/client       2.796s
```

This PR (bff9eae):

```text
goos: linux
goarch: amd64
pkg: github.com/tikv/pd/client
cpu: Intel(R) Xeon(R) Silver 4214R CPU @ 2.40GHz
BenchmarkTSODispatcherHandleRequests-28           119170              9254 ns/op
BenchmarkTSOStreamSendRecv-28                     676719              1647 ns/op
PASS
ok      github.com/tikv/pd/client       3.276s
```

There are several microseconds of additional overhead in this PR.

Please see if this is acceptable.

If you are good with it, let's consider merging #8618 before this PR.

client/tso_dispatcher.go
MyonKeminta: I updated the comment. Hope it's more clear now.

JmPotato (approved):

client/tso_stream.go
```diff
@@ -172,51 +178,255 @@ func (s tsoTSOStreamAdapter) Recv() (tsoRequestResult, error) {
...
+			// There's an unfinished request, cancel it, otherwise it will be blocked forever.
+			currentReq.callback(tsoRequestResult{}, currentReq.reqKeyspaceGroupID, finishWithErr)
+		}
+
+		s.stoppedWithErr.Store(&finishWithErr)
+		s.cancel()
+		for !s.state.CompareAndSwap(streamStateIdle, streamStateClosing) {
+			switch state := s.state.Load(); state {
+			case streamStateIdle, streamStateSending:
+				// streamStateSending should switch to streamStateIdle very quickly. Spin until successfully setting to
```
lhy1024, line 338: Spin？

lhy1024 (approved):
the rest LGTM

client/tso_stream.go
MyonKeminta: Yes. Considering that there won't be any blocking operation in `streamStateSending` state and the state won't last long, I think spinning here is better than writing another syncing operation here.

PR #8495, opened 2024-08-06
tso: skip resetting leader when resetting tso allocator

rleungx:

Issue Number: Ref #8477.

This PR supports skipping resetting the leadership when resetting the TSO allocator which will be used in the next PR. The PD might reset the TSO allocator leader and let the TSO service serve TSO requests. In this case, the PD leader should not be affected.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/tso/allocator_manager.go
```diff
@@ -1132,13 +1132,13 @@ func (am *AllocatorManager) HandleRequest(ctx context.Context, dcLocation string
 
 // ResetAllocatorGroup will reset the allocator's leadership and TSO initialized in memory.
 // It usually should be called before re-triggering an Allocator leader campaign.
-func (am *AllocatorManager) ResetAllocatorGroup(dcLocation string) {
+func (am *AllocatorManager) ResetAllocatorGroup(dcLocation string, skipResetLeader bool) {
```
lhy1024, line 1135: When is skipResetLeader true?

rleungx: I add a PR description.

lhy1024 (approved):

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -181,6 +181,9 @@ func (m *GroupManager) allocNodesToAllKeyspaceGroups(ctx context.Context) {
 			log.Info("the raftcluster is closed, stop to alloc nodes to all keyspace groups")
 			return
 		case <-ticker.C:
+			if m.nodesBalancer.Len() == 0 {
```
okJiang, line 184: `GetNodesCount()` is better

okJiang (approved):

PR #8497, opened 2024-08-06
DNM Support BalaceRegion

CalvinNeo:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/server/api/region.go
+++ b/server/api/region.go
@@ -128,6 +128,18 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
 	h.rd.JSON(w, http.StatusOK, state)
 }
 
+func (h *regionsHandler) BalanceRegion(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	rawStartKey := vars["startKey"]
+	rawEndKey := vars["endKey"]
+	state, err := h.Handler.BalanceRegion(rawStartKey, rawEndKey)
+	if err != nil {
+		h.rd.JSON(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	h.rd.JSON(w, http.StatusOK, state)
+}
+
 type regionsHandler struct {
 	*server.Handler
 	svr *server.Server
--- a/server/api/router.go
+++ b/server/api/router.go
@@ -284,6 +284,7 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	registerFunc(clusterRouter, "/regions/split", regionsHandler.SplitRegions, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
 	registerFunc(clusterRouter, "/regions/range-holes", regionsHandler.GetRangeHoles, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(clusterRouter, "/regions/replicated", regionsHandler.CheckRegionsReplicated, setMethods(http.MethodGet), setQueries("startKey", "{startKey}", "endKey", "{endKey}"), setAuditBackend(prometheus))
+	registerFunc(clusterRouter, "/regions/balance", regionsHandler.BalanceRegion, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
 
 	registerFunc(apiRouter, "/version", newVersionHandler(rd).GetVersion, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(apiRouter, "/status", newStatusHandler(svr, rd).GetPDStatus, setMethods(http.MethodGet), setAuditBackend(prometheus))
--- a/Makefile
+++ b/Makefile
@@ -243,12 +243,14 @@ SUBMODULES := $(filter $(shell find . -iname "go.mod" -exec dirname {} \;),\
 test: install-tools
 	# testing all pkgs...
 	@$(FAILPOINT_ENABLE)
+	echo ${TEST_PKGS}
 	CGO_ENABLED=1 go test -tags tso_function_test,deadlock -timeout 20m -race -cover $(TEST_PKGS) || { $(FAILPOINT_DISABLE); exit 1; }
 	@$(FAILPOINT_DISABLE)
 
 basic-test: install-tools
 	# testing basic pkgs...
 	@$(FAILPOINT_ENABLE)
+	echo ${TEST_PKGS}
 	go test $(BASIC_TEST_PKGS) || { $(FAILPOINT_DISABLE); exit 1; }
 	@$(FAILPOINT_DISABLE)
 
--- a/pkg/schedule/handler/handler.go
+++ b/pkg/schedule/handler/handler.go
@@ -1261,8 +1261,155 @@ func (h *Handler) SplitRegions(ctx context.Context, rawSplitKeys []any, retryLim
 	return s, nil
 }
 
+type StoreRegionSet struct {
+	ID          int64
+	Info 		StoreInfo
+	RegionIDSet map[int64]bool
+}
+
+type MigrationOp struct {
+	FromStore int64
+	ToStore   int64
+	Regions   map[int64]interface{}
+}
+
+func PickRegions(n int, fromStore *StoreRegionSet, toStore *StoreRegionSet) *MigrationOp {
+	o := MigrationOp{
+		FromStore: fromStore.ID,
+		ToStore:   toStore.ID,
+		Regions:   make(map[int64]interface{}),
+	}
+	for r, removed := range fromStore.RegionIDSet {
+		if n == 0 {
+			break
+		}
+		if removed {
+			continue
+		}
+		if _, exist := toStore.RegionIDSet[r]; !exist {
+			o.Regions[r] = nil
+			fromStore.RegionIDSet[r] = true
+		}
+		n--
+	}
+	return &o
+}
+
+func MigrationPlan(stores []*StoreRegionSet) ([]int, []int, []*MigrationOp) {
+	totalRegionCount := 0
+	for _, store := range stores {
+		totalRegionCount += len(store.RegionIDSet)
+	}
+	for _, store := range stores {
+		percentage := 100 * float64(len(store.RegionIDSet)) / float64(totalRegionCount)
+		log.Info("store region dist",
+			zap.Int64("store-id", store.ID),
+			zap.Int("num-region", len(store.RegionIDSet)),
+			zap.String("percentage", fmt.Sprintf("%.2f%%", percentage)))
+	}
+	avr := totalRegionCount / len(stores)
+	remainder := totalRegionCount % len(stores)
+	// sort TiFlash stores by region count in descending order
+	slices.SortStableFunc(stores, func(lhs, rhs *StoreRegionSet) int {
+		return -cmp.Compare(len(lhs.RegionIDSet), len(rhs.RegionIDSet))
+	})
+	expectedCount := []int{}
+	for i := 0; i < remainder; i++ {
+		expectedCount = append(expectedCount, avr+1)
+	}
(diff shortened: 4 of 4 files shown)


server/api/router.go
```diff
@@ -284,6 +284,8 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	registerFunc(clusterRouter, "/regions/split", regionsHandler.SplitRegions, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
 	registerFunc(clusterRouter, "/regions/range-holes", regionsHandler.GetRangeHoles, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(clusterRouter, "/regions/replicated", regionsHandler.CheckRegionsReplicated, setMethods(http.MethodGet), setQueries("startKey", "{startKey}", "endKey", "{endKey}"), setAuditBackend(prometheus))
+	registerFunc(clusterRouter, "/regions/balance-keyrange", regionsHandler.BalanceKeyrange, setMethods(http.MethodPost), setAuditBackend(localLog, prometheus))
```
rleungx, line 287: Should we config it through the scheduler API?

CalvinNeo: What kind of configuration could be done to this scheduler? I think its behavior is already determined when creating, so we either wait it finished or timeout, or we can just delete it.

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,488 @@
...
+	StartKey   []byte             `json:"start_key"`
+	EndKey     []byte             `json:"end_key"`
+	Ops        []*migrationOp     `json:"ops"`
+	Operators  []*operatorWrapper `json:"operators"`
+	Running    []*operatorWrapper `json:"running"`
+	TotalCount int                `json:"total"`
+	StartTime  time.Time
+}
+
+func computeCandidateStores(requiredLabels []*metapb.StoreLabel, stores []*metapb.Store, regions []*core.RegionInfo) []*storeRegionSet {
```
rleungx, line 162: We need to decide who can be a candidate for each region according to the constraint. Not all regions share the same stores.

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,488 @@
...
+				toStore.RegionIDSet[r] = true
+				fromStore.RegionIDSet[r] = false
+				break
+			}
+		}
+	}
+	return &o
+}
+
+func buildMigrationPlan(stores []*storeRegionSet) ([]int, []int, []*migrationOp, int) {
```
rleungx, line 79: We also need to make store balance under constraint. `expectedCount` is not accurate.

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+type operatorWrapper struct {
+	Operator  *operator.Operator `json:"operators"`
+	Region    *core.RegionInfo
+	FromStore uint64
+	Peer      *metapb.Peer
+}
+type migrationPlan struct {
+	ErrorCode  uint64             `json:"error_code"`
```
rleungx, line 152: What is json used for?

pkg/schedule/schedulers/base_scheduler.go
```diff
@@ -146,3 +146,10 @@ func (s *BaseScheduler) IsDefault() bool {
 	}
 	return false
 }
+
+// IsFinished implements the Scheduler interface.
+func (*BaseScheduler) IsFinished() bool {
```
rleungx, line 151: If it's only used in test and log, please remove it.

server/api/region.go
```diff
@@ -130,6 +131,47 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
 	h.rd.JSON(w, http.StatusOK, state)
 }
 
+// @Tags     region
+// @Summary
```
JaySon-Huang, line 135:
```suggestion
// @Summary Add a scheduler that try best to balance the region peer distribution among store for a given range [startKey, endKey]
```

server/api/region.go
```diff
@@ -130,6 +131,47 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
...
+// @Accept   json
+// @Param    start_key  body  string  true  "Regions start key, hex encoded"
+// @Param    end_key    body  string  true  "Regions end key, hex encoded"
+// @Param    batch_size  body  string  true  "Maximum operators scheduled in one"
+// @Param    timeout    body  string  true  "Timeout time in milliseconds"
+// @Param    required_labels    body  array  true  "Only the stores with these labels would be scheduled"
```
JaySon-Huang, line 141:
```suggestion
```
User only specify the key-range, this API would follow the `<region,zone,host>` constraint for rebalancing the peer among stores

JaySon-Huang (commented):
Let's first define the API interface of adding scheduler and checking scheduler status
@CalvinNeo @rleungx @joey-yez

server/api/region.go
```diff
@@ -130,6 +131,47 @@ func (h *regionsHandler) CheckRegionsReplicated(w http.ResponseWriter, r *http.R
 	h.rd.JSON(w, http.StatusOK, state)
 }
 
+// @Tags     region
+// @Summary
+// @Accept   json
+// @Param    start_key  body  string  true  "Regions start key, hex encoded"
+// @Param    end_key    body  string  true  "Regions end key, hex encoded"
+// @Param    batch_size  body  string  true  "Maximum operators scheduled in one"
+// @Param    timeout    body  string  true  "Timeout time in milliseconds"
```
JaySon-Huang, line 140: I think using "seconds" instead of "milliseconds" is more reasonable.
What's more, there is a "timeout" param in the "/admin/unsafe/remove-failed-stores", which unit is "seconds".

pkg/schedule/operator/kind.go
```diff
@@ -49,6 +49,8 @@ const (
 	OpWitnessLeader
 	// Include witness transfer.
 	OpWitness
+	// Include keyrange balancer
+	OpKeyrange
```
JaySon-Huang, line 53: Can we reuse the "OpRange" instead of adding "OpKeyrange"?

CalvinNeo: This is for scatter range scheduler. IMO, different schedulers do not interfere with each other.

pkg/schedule/schedulers/base_scheduler.go
CalvinNeo: fixed

server/api/region.go
CalvinNeo: Fixed

server/api/region.go
CalvinNeo: Yes, maybe we can follow the pattern of scatter region

pkg/schedule/schedulers/balance_keyrange.go
CalvinNeo: Then I will adopt the region-wise strategy.

pkg/schedule/schedulers/balance_keyrange.go
CalvinNeo: It is not implemented yet, but I would like if to be marshaled as a string and printed into log, so we can see the detailed migration plan.

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,488 @@
...
+	})
+	expectedCount := []int{}
+	for range remainder {
+		expectedCount = append(expectedCount, avr+1)
+	}
+	for i := remainder; i < len(stores); i++ {
+		expectedCount = append(expectedCount, avr)
+	}
+
+	senders := []int{}
```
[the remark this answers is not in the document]
CalvinNeo: Some regions are scheduled out from sender to receiver. We can remove it from return value afterwards.

pkg/schedule/operator/operator_controller.go
```diff
@@ -418,6 +418,16 @@ func (oc *Controller) PromoteWaitingOperator() {
 	}
 }
 
+// GetWopCount returns the count in the waiting list of a kind of scheduler.
+func (oc *Controller) GetWopCount(kind string) uint64 {
```
bufferflies, line 422: How about using this function to check the concurrent operators ? `s.OpController.OperatorCount`

pkg/schedule/operator/operator_controller.go
```diff
@@ -418,6 +418,16 @@ func (oc *Controller) PromoteWaitingOperator() {
...
+	return oc.wopStatus.getCount(kind)
+}
+
+// GetSchedulerMaxWaitingOperator returns the waiting queue side.
+func (oc *Controller) GetSchedulerMaxWaitingOperator() uint64 {
```
bufferflies, line 427:
```suggestion
func (oc *Controller) GetSchedulerMaxWaitingOperatorLimit() uint64 {
```

pkg/schedule/schedulers/metrics.go
```diff
@@ -342,4 +346,6 @@ var (
 	transferWitnessLeaderCounter              = transferWitnessLeaderCounterWithEvent("schedule")
 	transferWitnessLeaderNewOperatorCounter   = transferWitnessLeaderCounterWithEvent("new-operator")
 	transferWitnessLeaderNoTargetStoreCounter = transferWitnessLeaderCounterWithEvent("no-target-store")
+
+	balanceKeyrangeScheduleCounter = balanceKeyrangeScheduleCounterWithEvent("schedule")
```
bufferflies, line 350: how about adding new metrics for `new-operator`?

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,502 @@
...
+	sche "github.com/tikv/pd/pkg/schedule/core"
+	"github.com/tikv/pd/pkg/schedule/operator"
+	"github.com/tikv/pd/pkg/schedule/plan"
+	"github.com/tikv/pd/pkg/schedule/types"
+)
+
+type storeRegionSet struct {
+	ID   uint64
+	Info *metapb.Store
+	// If the region still exists durgin migration
```
bufferflies, line 40: what‘s the meaning about `durgin`?

pkg/schedule/schedulers/balance_keyrange.go
```diff
@@ -0,0 +1,502 @@
...
+	// When a pd cluster is bootstrapping, it should not create this scheduler.
+	// However, in order to prevent a wrong operation which could also be from outside pd, we reject a schedule of a universal key-range.
+	if len(s.conf.Range.StartKey) == 0 && len(s.conf.Range.EndKey) == 0 {
+		log.Debug("Invalid keyrange", zap.ByteString("startKey", s.conf.Range.StartKey), zap.ByteString("endStartKey", s.conf.Range.EndKey), zap.Any("plan", s.migrationPlan))
+		doShutdown()
+		return []*operator.Operator{}, []plan.Plan{}
+	}
+
+	oc := s.BaseScheduler.OpController
+	rangeChanged := true
```
bufferflies, line 414: Is it reasonable to update the key range?

pkg/schedule/schedulers/balance_keyrange.go
JaySon-Huang: Does it means we may generate an operator that move some Region out of the given range to keep the "store balance"?

server/api/router.go
niubell: It's better to router/implement the new shceduler type in CreateScheduler with a uniform way.

PR #8505, opened 2024-08-08
checker: use ttl cache

### LINKED ISSUE #8500: Duration of analyze partition table increased more than 50x
### What did you do?

1. create partition table
> CREATE TABLE `t1` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `k` int(11) NOT NULL DEFAULT '0',
  `c` char(120) NOT NULL DEFAULT '',
  `pad` char(60) NOT NULL DEFAULT '',
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_INCREMENT=100279
PARTITION BY HASH (`id`) PARTITIONS 1000 
2. insert 100000 rows data
3. analyze table t1

Analyze finishes in a shorter time.
According to previous test experience, it is within 100s. 
The most recent successful test was 37s.

### What did you see instead?
"analyze time: "=1929.67098683
Sometimes it can be as long as 4,000 seconds . 
A lot of time is spent on: `merging global stats for test.t1 columns`
The performance degradation is very serious

After testing, it was confirmed that this PR(https://github.com/tikv/pd/pull/8377) caused it

rleungx:

Issue Number: Close #8500

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/checker/checker_controller.go
```diff
@@ -88,7 +88,7 @@ type Controller struct {
 
 // NewController create a new Controller.
 func NewController(ctx context.Context, cluster sche.CheckerCluster, conf config.CheckerConfigProvider, ruleManager *placement.RuleManager, labeler *labeler.RegionLabeler, opController *operator.Controller) *Controller {
-	pendingProcessedRegions := cache.NewDefaultCache(DefaultPendingRegionCacheSize)
+	pendingProcessedRegions := cache.NewIDTTL(ctx, time.Minute, 3*time.Minute)
```
lhy1024, line 91: sure, previous `suspectRegions` is `*cache.TTLUint64`

lhy1024: Do we need to add a test for this issue?

pkg/schedule/checker/checker_controller.go
```diff
@@ -327,16 +327,15 @@ func (c *Controller) GetRuleChecker() *RuleChecker {
...
-		pendingRegions = append(pendingRegions, item.Key)
-	}
-	return pendingRegions
+	return c.pendingProcessedRegions.GetAllID()
 }
 
 // AddPendingProcessedRegions adds the pending processed region into the cache.
 func (c *Controller) AddPendingProcessedRegions(ids ...uint64) {
 	for _, id := range ids {
+		if c.pendingProcessedRegions.Len() > DefaultPendingRegionCacheSize {
```
lhy1024, line 336: why to add it?

rleungx: To prevent the cache from becoming too large.

pkg/schedule/checker/checker_controller.go
rleungx: It's not just the problem of suspectRegions. The waiting list also has the same problem previously.

pkg/schedule/checker/checker_controller.go
lhy1024: If so, some regions, such as those from `checkSuspectRanges` or `AccelerateRegionsScheduleInRange`, may not be put successfully.

Should we only skip adding to `pendingProcessedRegions` in `tryAddOperators` when `pendingProcessedRegions` is full?

pkg/schedule/checker/checker_controller.go
lhy1024: ok, then do we need to add a test for this issue?

HuSharp: why 3 mins?

rleungx: The original setting is 3 min.

pkg/schedule/checker/checker_controller.go
rleungx: done

lhy1024 (approved):

HuSharp (approved):

PR #8516, opened 2024-08-12
scheduler: remove old scheduler name

okJiang:

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/diagnostic_recorder.go
```diff
@@ -23,6 +23,7 @@ import (
 	sc "github.com/tikv/pd/pkg/schedule/config"
 	"github.com/tikv/pd/pkg/schedule/operator"
 	"github.com/tikv/pd/pkg/schedule/plan"
+	types "github.com/tikv/pd/pkg/schedule/type"
```
rleungx, line 26: How about renaming it to types?

pkg/schedule/schedulers/diagnostic_recorder.go
```diff
@@ -131,11 +132,11 @@ func (d *DiagnosticRecorder) GetLastResult() *DiagnosticResult {
 			}
 		} else if firstStatus == Pending {
 			// This is used to handle pending status because of reach limit in `IsScheduleAllowed`
-			resStr = fmt.Sprintf("%s reach limit", d.schedulerName)
+			resStr = fmt.Sprintf("%v reach limit", d.schedulerType)
 		}
 	}
 	return &DiagnosticResult{
-		Name:      d.schedulerName,
+		Name:      string(d.schedulerType),
```
rleungx, line 139: What is the difference between `string(d.schedulerType)` and `d.schedulerType.String()`?

okJiang: no difference

pkg/schedule/schedulers/diagnostic_recorder.go
okJiang: sgtm

pkg/schedule/schedulers/diagnostic_recorder.go
rleungx: Then shall we keep consistent with other places?

okJiang: updated

rleungx (approved):

lhy1024 (approved):

niubell (approved):

PR #8523, opened 2024-08-14
tso: fix an invalid err != nil check

### LINKED ISSUE #8524: An invalid err != nil check may cause the merged TS inconsistent
## Bug Report

https://github.com/tikv/pd/blob/1c1cd99ba73eaf1fd80a03310221b37ec48cf7d1/pkg/tso/keyspace_group_manager.go#L1365-L1384

Since the `err` in the loop is a local variable, so the check at line 1382 will always be false.

JmPotato:

Issue Number: close #8524.

https://github.com/tikv/pd/blob/1c1cd99ba73eaf1fd80a03310221b37ec48cf7d1/pkg/tso/keyspace_group_manager.go#L1365-L1384

Since the `err` in the loop is a local variable, so the check at line 1382 will always be false before this fix.

```commit-message
Use the continue label to retry the whole merge loop correctly.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/keyspace_group_manager.go
+++ b/pkg/tso/keyspace_group_manager.go
@@ -1302,6 +1302,7 @@ func (kgm *KeyspaceGroupManager) mergingChecker(ctx context.Context, mergeTarget
 		mergeMap[id] = struct{}{}
 	}
 
+mergeLoop:
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


lhy1024 (approved):
We also need to pick it to other branches and pd-cse.

okJiang (commented):
It is better to use `continue xxxxLable` in Line 1376 instead of `break`

rleungx (approved):

okJiang (approved):

Issue #8474, opened 2024-08-01
deprecate `SchedulerConfigProvider` interface, the old scheduler config


okJiang:
## Enhancement Task

We can consider removing this interface. here is some background:

- v2.x all scheduler configs are stored in the global config `Config.SchedulerConfig`
- v3.x let the scheduler has it's own config, that's store in another key/value in etcd https://github.com/tikv/pd/pull/1733

To ensure compatibility, both configurations are retained, and we can now clean up the old one. However, there is one problem that needs to be solved: the default scheduler state needs special handling.

okJiang:
So we can deprecate these two

https://github.com/tikv/pd/blob/3ce31ef4b3e81cbf77271cc29384454f8b3d0a8a/pkg/schedule/config/config.go#L274-L278

niubell:
Do we also have the specified user docs to deprecate? If there is really a v2 user (the possibility is extremely low), how to migrate to the new version and configuration smoothly?

okJiang:
> Do we also have the specified user docs to deprecate?

no, I didn't find

> If there is really a v2 user (the possibility is extremely low), how to migrate to the new version and configuration smoothly?

Currently, the code has been made compatible, so users can upgrade directly from v2 to higher versions. But after it is deprecated, v2 cannot be upgraded to master, but clusters with v3+ can all be directly upgraded to master.

okJiang:
Approved by @easonn7

easonn7:
Both schedulers-v2 and schedulers-payload are legacy remnants from version 2 and earlier. It is recommended to delete them entirely, including the code and the show command display.

PR #8567, opened 2024-08-23
scheduler: add disable to independent config

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/8474

Before deprecating `scheduler-v2`, we need to save the `disable` field of the default scheduler to the etcd `xxxx/scheduler_config` path. This pr does not modify the use of `disable`. `disable` only is added to/deleted from the storage. So it will not affect the existing logic.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

### Release note

```release-note
None.
```


pkg/schedule/schedulers/scheduler_controller.go
```diff
@@ -205,6 +214,11 @@ func (c *Controller) AddScheduler(scheduler Scheduler, args ...string) error {
 	c.wg.Add(1)
 	go c.runScheduler(s)
 	c.schedulers[s.Scheduler.GetName()] = s
+	if err := scheduler.SetDisable(false); err != nil {
+		log.Error("can not update scheduler status", zap.String("scheduler-name", name),
+			errs.ZapError(err))
+		return err
```
okJiang, line 220:
```suggestion
		// No need to return here, we still use the scheduler config to control the `disable`.
```

pkg/schedule/schedulers/scheduler.go
```diff
@@ -47,6 +47,16 @@ type Scheduler interface {
...
+	// IsDiable returns if the scheduler is disabled, it only works for default schedulers.
+	// - BalanceRegionScheduler
+	// - BalanceLeaderScheduler
+	// - BalanceHotRegionScheduler
+	// - EvictSlowStoreScheduler
+	IsDisable() bool
+	// SetDisable sets the scheduler's disable, it only works for default schedulers.
+	SetDisable(bool) error
+	// IsDefault returns if the scheduler is a default scheduler.
+	IsDefault() bool
```
okJiang, line 59: need an ut

pkg/schedule/schedulers/balance_leader.go
```diff
@@ -541,3 +539,18 @@ func (l *balanceLeaderScheduler) createOperator(solver *solver, collector *plan.
 	op.SetAdditionalInfo("targetScore", strconv.FormatFloat(solver.targetScore, 'f', 2, 64))
 	return op
 }
+
+// IsDiable implements the Scheduler interface.
+func (l *balanceLeaderScheduler) IsDisable() bool {
+	return l.conf.isDisable()
+}
```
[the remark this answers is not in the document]
okJiang: base_scheduler.go implemented it.

okJiang:
ptal~ @rleungx @nolouch

pkg/schedule/schedulers/balance_leader.go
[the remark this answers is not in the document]
okJiang: If we want to implement it once, we need to put `isDisable` and `setDisable` into `schedulerConfig` interface. Do you think you need to do this?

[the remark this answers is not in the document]
rleungx: WDYT? Is it possible that we might customize it?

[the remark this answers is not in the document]
okJiang: In this pr, we only need to make a customized implementation for defaultScheduler. I think this is acceptable and only needs to be implemented four times.

If follow what you said, it does only need to be implemented once, but I feel that `setDisable` should not be placed in the `schedulerConfig` interface, because only `defaultScheduler` can be `setDisabled`. But if you insist, I can also modify it

pkg/schedule/schedulers/balance_leader.go
```diff
@@ -52,8 +51,7 @@ const (
 )
 
 type balanceLeaderSchedulerConfig struct {
-	syncutil.RWMutex
-	schedulerConfig
+	baseDefaultSchedulerConfig
```
rleungx, line 54: I suggest moveing `Disabled` here.

okJiang: If this, we should implement `isDisable` and `setDisable` four times.

okJiang: friendly ping~

rleungx: Just like other config?

pkg/schedule/schedulers/base_scheduler.go
```diff
@@ -114,3 +119,12 @@ func (s *BaseScheduler) GetName() string {
...
 }
+
+// IsDiable implements the Scheduler interface.
+func (*BaseScheduler) IsDisable() bool { return false }
+
+// SetDisable implements the Scheduler interface.
+func (*BaseScheduler) SetDisable(bool) error { return nil }
+
+// IsDefault returns if the scheduler is a default scheduler.
+func (*BaseScheduler) IsDefault() bool { return false }
```
nolouch, line 130: How about:
```
type hasDefaulSchedulerConfig interface {
  func IsDisable() bool
  func SetDisable(disabled bool) error
}

func (s *BaseScheduler) IsDefault() bool { 
    _, ok := s.conf.(hasDefaulSchedulerConfig)
    return ok
}

func (s *BaseScheduler) SetDisable(v bool) error {
  if c, ok := s.conf.(hasDefaulSchedulerConfig);ok {
     c.SetDisable(v)
  }
  return nil
}

func (*BaseScheduler) IsDisable() bool { 
    if c, ok := s.conf.(hasDefaulSchedulerConfig);ok {
          return c.IsDisable()
    }
     return false 
}
```

tools/pd-ctl/tests/scheduler/scheduler_test.go
```diff
@@ -621,6 +621,7 @@ func (suite *schedulerTestSuite) checkHotRegionSchedulerConfig(cluster *pdTests.
 		"min-hot-key-rate":        float64(10),
 		"min-hot-query-rate":      float64(10),
 		"src-tolerance-ratio":     1.05,
+		"disabled":                false,
```
rleungx, line 624: Is it allowed to be set through `scheduler config xxx`?

nolouch (approved):

tools/pd-ctl/tests/scheduler/scheduler_test.go
okJiang: In current pr, it is allowed. Do we need mask it for users?

rleungx: I prefer to make it only for internal use purpose.

okJiang: https://github.com/tikv/pd/pull/8567/commits/2765b8ce679d5ea6bec11bebafe1c850ff74a750

https://github.com/tikv/pd/pull/8567/commits/590f2c1486c6d909efa5eaf53032dc623eed024c

done

rleungx (approved):

niubell (approved):

PR #8569, opened 2024-08-26
pd-heartbeat-bench: add default values

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/7703

```commit-message
```

After #7791, we remove the default config in the code to test a single variable.

But in some scenarios, we're more concerned with simulating a large cluster's status, so I put the default value back in the configuration file.

It is more friendly to guide the user to test.

### Check List

Tests

- Unit test
 

### Release note

```release-note
None.
```


HuSharp (approved):

okJiang (approved):

tools/pd-heartbeat-bench/config-template.toml
```diff
@@ -1,14 +1,15 @@
...
 
 replica = 3
 
 leader-update-ratio = 0.06
-epoch-update-ratio = 0.0
-space-update-ratio = 0.0
-flow-update-ratio = 0.0
+epoch-update-ratio = 0.04
+space-update-ratio = 0.15
+flow-update-ratio = 0.35
```
okJiang, line 12: Can you explain why it should be `0.04/0.15/0.35`?

okJiang (changes requested):

tools/pd-heartbeat-bench/config-template.toml
lhy1024: Copy from older configure

okJiang (approved):

lhy1024:
PTAL @rleungx

okJiang (approved):

PR #8596, opened 2024-09-04
*: move key_path.go

okJiang:

The place that access the keypath is not required to import endpoint.

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


lhy1024:
Maybe we will replace etcd with other backend, I don't think it is a good idea to move keypath to etcdutil.

okJiang:
> Maybe we will replace etcd with other backend, I don't think it is a good idea to move keypath to etcdutil.

Do you have any better suggestion? In current situation, I think it is ok to move into etcdutil.

okJiang:
> > Maybe we will replace etcd with other backend, I don't think it is a good idea to move keypath to etcdutil.
> 
> Do you have any better suggestion? In current situation, I think it is ok to move into etcdutil.

How about `utils/keypath`?

lhy1024 (commented):
I think utils/keypath might be better, and see what others suggest.

okJiang:
> I think utils/keypath might be better, and see what others suggest.

What do you think? @rleungx @JmPotato

rleungx:
utils/keypath LGTM

rleungx (approved):

JmPotato (approved):

lhy1024 (approved):

PR #8601, opened 2024-09-09
chore(dashboard): update TiDB Dashboard to v6.5.11-9d2b0ce6 [release-6.5]

baurine:

Issue Number: ref #8611.

Update TiDB Dashboard to [v6.5.11-9d2b0ce6](https://github.com/pingcap/tidb-dashboard/releases/tag/v6.5.11-9d2b0ce6).

### Release note

```release-note
None
```

### DIFF
--- a/pkg/dashboard/adapter/config.go
+++ b/pkg/dashboard/adapter/config.go
@@ -35,6 +35,7 @@ func GenDashboardConfig(srv *server.Server) (*config.Config, error) {
 	dashboardCfg.PublicPathPrefix = cfg.Dashboard.PublicPathPrefix
 	dashboardCfg.EnableTelemetry = cfg.Dashboard.EnableTelemetry
 	dashboardCfg.EnableExperimental = cfg.Dashboard.EnableExperimental
+	dashboardCfg.DisableCustomPromAddr = cfg.Dashboard.DisableCustomPromAddr
 	if dashboardCfg.ClusterTLSConfig, err = cfg.Security.ToTLSConfig(); err != nil {
 		return nil, err
 	}
--- a/pkg/dashboard/adapter/config.go
+++ b/pkg/dashboard/adapter/config.go
@@ -15,9 +15,16 @@
 package adapter
 
 import (
+	"os"
+	"strings"
+
 	"github.com/pingcap/tidb-dashboard/pkg/config"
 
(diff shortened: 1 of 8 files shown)


mornyx:
This is a 6.5 cherry-pick of https://github.com/tikv/pd/pull/8581, PTAL, thx!
/cc @lhy1024 @rleungx @niubell

pkg/dashboard/adapter/config.go
```diff
@@ -35,6 +35,7 @@ func GenDashboardConfig(srv *server.Server) (*config.Config, error) {
 	dashboardCfg.PublicPathPrefix = cfg.Dashboard.PublicPathPrefix
 	dashboardCfg.EnableTelemetry = cfg.Dashboard.EnableTelemetry
 	dashboardCfg.EnableExperimental = cfg.Dashboard.EnableExperimental
+	dashboardCfg.DisableCustomPromAddr = cfg.Dashboard.DisableCustomPromAddr
```
rleungx, line 38: Will the config be lost when we upgrade?

mornyx: > Will the config be lost when we upgrade?

This configuration already exists in the `master` and `release-6.5` branches of tidb-dashboard and will not be lost during a normal upgrade.

rleungx: What if we set it to true in 6.5.11, then the cluster is upgraded to 7.1.x which doesn't contain this PR?

mornyx: Oh, I get what you mean. This change will be cherry-picked to all LTS branches later. (At least for now only 6.5.x customers explicitly need to configure this.)

mornyx: If an upgrade occurs during this period, this configuration will indeed lose its effect.

rleungx: After losing it, if we upgrade again to a higher version with this config, the config value will be changed to false.

mornyx: I think this is acceptable and I will doc this behavior. It is expected that this configuration will not work after upgrading to a version that does not support this feature.

rleungx (approved):

HuSharp (approved):

PR #8618, opened 2024-09-12
client: Add benchmark for tsoStream and tsoDispatcher

### LINKED ISSUE #8432: client: Support TSO RPC Parallelizing

MyonKeminta:

Issue Number: Ref #8432

```commit-message
client: Add benchmark for tsoStream and tsoDispatcher
```

This PR adds some benchmarks that might be helpful for checking if there's any regression when working on #8432 .

### Current result:

```text
goos: linux
goarch: amd64
pkg: github.com/tikv/pd/client
cpu: Intel(R) Xeon(R) Silver 4214R CPU @ 2.40GHz
BenchmarkTSODispatcherHandleRequests-28           140149              7508 ns/op
BenchmarkTSOStreamSendRecv-28                    2788069               425.8 ns/op
PASS
ok      github.com/tikv/pd/client       2.796s
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/tso_dispatcher_test.go
+++ b/client/tso_dispatcher_test.go
@@ -39,11 +39,11 @@ func (m *mockTSOServiceProvider) getOption() *option {
 	return m.option
 }
 
-func (m *mockTSOServiceProvider) getServiceDiscovery() ServiceDiscovery {
+func (*mockTSOServiceProvider) getServiceDiscovery() ServiceDiscovery {
 	return NewMockPDServiceDiscovery([]string{mockStreamURL}, nil)
 }
 
-func (m *mockTSOServiceProvider) updateConnectionCtxs(ctx context.Context, _dc string, connectionCtxs *sync.Map) bool {
+func (*mockTSOServiceProvider) updateConnectionCtxs(ctx context.Context, _dc string, connectionCtxs *sync.Map) bool {
 	_, ok := connectionCtxs.Load(mockStreamURL)
 	if ok {
 		return true
--- a/client/tso_dispatcher_test.go
+++ b/client/tso_dispatcher_test.go
@@ -1,0 +1,107 @@
+// Copyright 2024 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


lhy1024 (commented):
please fix ci

lhy1024 (approved):

MyonKeminta:
Already included in https://github.com/tikv/pd/pull/8483 .

PR #8632, opened 2024-09-14
*: reset config if the input is invalid 

### LINKED ISSUE #8619: scheduler: there is error not handled in evict-leader-scheduler
### What did you do?

add an evict-leader-scheduler with error params

report an error and it should be successful next time when I use the correct params

### What did you see instead?

it failed again when I used correct params

it shows `Failed! [500] "[PD:core:ErrPauseLeaderTransfer]store 75 is paused for leader transfer"`

v6.5.10

rleungx:

Issue Number: Close #8619.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/schedulers/evict_leader.go
+++ b/pkg/schedule/schedulers/evict_leader.go
@@ -408,6 +408,7 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	batchFloat, ok := input["batch"].(float64)
 	if ok {
 		if batchFloat < 1 || batchFloat > 10 {
+			handler.config.delete(id)
 			handler.rd.JSON(w, http.StatusBadRequest, "batch is invalid, it should be in [1, 10]")
 			return
 		}
@@ -417,6 +418,7 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	ranges, ok := (input["ranges"]).([]string)
 	if ok {
 		if !inputHasStoreID {
+			handler.config.delete(id)
 			handler.rd.JSON(w, http.StatusInternalServerError, errs.ErrSchedulerConfig.FastGenByArgs("id"))
 			return
 		}
@@ -426,6 +428,7 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 
 	newRanges, err = getKeyRanges(ranges)
 	if err != nil {
+		handler.config.delete(id)
 		handler.rd.JSON(w, http.StatusInternalServerError, err.Error())
 		return
 	}
--- a/tools/pd-ctl/tests/scheduler/scheduler_test.go
+++ b/tools/pd-ctl/tests/scheduler/scheduler_test.go
@@ -813,6 +813,60 @@ func (suite *schedulerTestSuite) checkSchedulerDiagnostic(cluster *pdTests.TestC
 	checkSchedulerDescribeCommand("balance-leader-scheduler", "normal", "")
 }
 
+func (suite *schedulerTestSuite) TestEvictLeaderScheduler() {
+	suite.env.RunTestInPDMode(suite.checkEvictLeaderScheduler)
+}
+
+func (suite *schedulerTestSuite) checkEvictLeaderScheduler(cluster *pdTests.TestCluster) {
+	re := suite.Require()
+	pdAddr := cluster.GetConfig().GetClientURL()
+	cmd := ctl.GetRootCmd()
+
+	stores := []*metapb.Store{
+		{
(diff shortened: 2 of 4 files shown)


pkg/schedule/schedulers/evict_leader.go
```diff
@@ -103,6 +103,12 @@ func (conf *evictLeaderSchedulerConfig) removeStoreLocked(id uint64) (bool, erro
 	return false, errs.ErrScheduleConfigNotExist.FastGenByArgs()
 }
 
+func (conf *evictLeaderSchedulerConfig) removeStore(id uint64) {
+	conf.Lock()
+	defer conf.Unlock()
+	conf.removeStoreLocked(id)
```
lhy1024: ci failed

tools/pd-ctl/tests/scheduler/scheduler_test.go
```diff
@@ -813,6 +813,60 @@ func (suite *schedulerTestSuite) checkSchedulerDiagnostic(cluster *pdTests.TestC
 	checkSchedulerDescribeCommand("balance-leader-scheduler", "normal", "")
 }
 
+func (suite *schedulerTestSuite) TestEvictLeaderScheduler() {
+	// FIXME: API mode may have the problem
```
lhy1024, line 817: Will we fix it in another PR?

lhy1024 (approved):
the rest LGTM

tools/pd-ctl/tests/scheduler/scheduler_test.go
```diff
@@ -813,6 +813,60 @@ func (suite *schedulerTestSuite) checkSchedulerDiagnostic(cluster *pdTests.TestC
...
+	pdTests.MustPutRegion(re, cluster, 1, 1, []byte("a"), []byte("b"))
+	output, err := tests.ExecuteCommand(cmd, []string{"-u", pdAddr, "scheduler", "add", "evict-leader-scheduler", "2"}...)
+	re.NoError(err)
+	re.Contains(string(output), "Success!")
+	output, err = tests.ExecuteCommand(cmd, []string{"-u", pdAddr, "scheduler", "add", "evict-leader-scheduler", "1"}...)
+	re.NoError(err)
+	re.Contains(string(output), "Success!")
+	output, err = tests.ExecuteCommand(cmd, []string{"-u", pdAddr, "scheduler", "remove", "evict-leader-scheduler"}...)
+	re.NoError(err)
+	re.Contains(string(output), "Success!")
```
lhy1024, line 861: Maybe we need to check
```
testutil.Eventually(re, func() bool { // wait for removed scheduler to be synced to scheduling server.
		output := mustExec(re, cmd, []string{"-u", pdAddr, "scheduler", "config", "evict-leader-scheduler"}, nil)
		return strings.Contains(output, "[404] scheduler not found")
	})
```

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -408,6 +415,7 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	batchFloat, ok := input["batch"].(float64)
 	if ok {
 		if batchFloat < 1 || batchFloat > 10 {
+			handler.config.removeStore(id)
```
niubell, line 418: For invalid config parameters, It's better to validate it, do nothing, and return 4xx + error message to let the users know it, it's not classic and unsafe to reset memory content when meeting invalid inputs.

rleungx: It's a rollback operation since the memory state has been changed in line 407.

tools/pd-ctl/tests/scheduler/scheduler_test.go
rleungx: seem not work

pkg/schedule/schedulers/evict_leader.go
niubell: Got it, make sense

okJiang, line 418: 1. Do we allow the input to not contain the store id?
2. How about put L407(pauseLeaderTransferIfStoreNotExist) behind all checks over?

rleungx: 1. Yes, we allow only changing batch size.
2. range relies on `exist`, only batch can be moved, but no much difference I think.

okJiang: How about adding a defer after L411? For now, it is so easy to forget to `removeStore`.

```go
defer func () {
    if err != nil {
         ....
    }
}()
```

okJiang (approved):

PR #8633, opened 2024-09-14
client: Support parallel TSO RPC requests on single dispatcher loop

### LINKED ISSUE #8432: client: Support TSO RPC Parallelizing

MyonKeminta:

Issue Number: Close #8432

```commit-message
client: Support parallel TSO RPC requests on single dispatcher loop

This commit supports handling multiple TSO RPC concurrently in one single dispatcher loop to reduce the expected time that each GetTS call spent on waiting the next batch.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/tso_dispatcher.go
```diff
@@ -578,5 +650,98 @@ func (td *tsoDispatcher) compareAndSwapTS(
...
+	// because that MaxTSOBatchWaitInterval is just enabled. In this case, disable concurrent TSO RPC requests
+	// immediately, because MaxTSOBatchWaitInterval and concurrent RPC requests has opposite purpose.
+	immediatelyUpdate := td.rpcConcurrency > 1 && maxBatchWaitInterval > 0
+
+	// Allow always updating for test purpose.
+	failpoint.Inject("tsoDispatcherAlwaysCheckConcurrency", func() {
+		immediatelyUpdate = true
+	})
+
+	if !immediatelyUpdate && now.Sub(td.lastCheckConcurrencyTime) < dispatcherCheckRPCConcurrencyInterval {
```
JmPotato, line 698: Don't we need to update `td.lastCheckConcurrencyTime` here also?

client/tso_stream.go
```diff
@@ -228,6 +232,8 @@ var streamIDAlloc atomic.Int32
 
 const invalidStreamID = "<invalid>"
 
+const maxPendingRequestsInTSOStream = 64
```
JmPotato, line 235: Please merge with line 233.

client/tso_dispatcher.go
```diff
@@ -542,32 +619,27 @@ func (td *tsoDispatcher) doneCollectedRequests(tbc *tsoBatchController, physical
...
 				zap.Uint32("new-group-id", curTSOInfo.respKeyspaceGroupID))
 		}
 		// The TSO we get is a range like [largestLogical-count+1, largestLogical], so we save the last TSO's largest logical
 		// to compare with the new TSO's first logical. For example, if we have a TSO resp with logical 10, count 5, then
-		// all TSOs we get will be [6, 7, 8, 9, 10]. lastTSOInfo.logical stores the logical part of the largest ts returned
+		// all TSOs we get will be [6, 7, 8, 9, 10]. latestTSOInfo.logical stores the logical part of the largest ts returned
 		// last time.
-		if tsoutil.TSLessEqual(physical, firstLogical, lastTSOInfo.physical, lastTSOInfo.logical) {
+		if tsoutil.TSLessEqual(curTSOInfo.physical, firstLogical, lastTSOInfo.physical, lastTSOInfo.logical) {
 			log.Panic("[tso] timestamp fallback",
```
JmPotato, line 638: Shall we print out the stream ID here?

client/tso_dispatcher.go
MyonKeminta: It's updated at line 701. We don't update it if it's skipped, otherwise it may never be checked

lhy1024 (approved):

PR #8636, opened 2024-09-18
client: Support both sync and async interface for tsoStream, to avoid potential performane regression when concurrent RPC is not enabled

MyonKeminta:

Issue Number: 
 * Ref #8432 
 * Ref https://github.com/pingcap/tidb/issues/56103

```commit-message
client: Support both sync and async interface for tsoStream, to avoid potential performane regression when concurrent RPC is not enabled
```

The issue https://github.com/pingcap/tidb/issues/56103 reports a performance regression introduced by PR https://github.com/pingcap/tidb/pull/56093 , which introduces #8483 to TiDB.

It's reasonable that this PR does have risk to introduce performance regression in non-current mode (which would be default after finishing #8432). However:

* After I rerun the failed bisect task twice, both of these reruns failed to reproduce the regression.
* According to the previously done microbench (mentioned [here](https://github.com/tikv/pd/pull/8483#issuecomment-2345496796)), the additional overhead introduced by that PR should not be that much.

So I think the performance regression report looks quite doubtful.

Anyway, I still prepared this PR: the `tsoStream` provides both sync and async interface. When concurrent RPC is not enabled, the sync interface will be used, making the logic the same as that before merging #8483 .

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/tso_dispatcher.go
+++ b/client/tso_dispatcher.go
@@ -328,7 +328,7 @@ tsoBatchLoop:
 		case td.tsDeadlineCh <- dl:
 		}
 		// processRequests guarantees that the collected requests could be finished properly.
-		tbcConsumed, err := td.processRequests(stream, dc, batchController, done, false)
+		tbcConsumed, err := td.processRequests(stream, dc, batchController, done, td.useAsyncStream())
 		// If error happens during tso stream handling, reset stream and run the next trial.
 		if tbcConsumed {
 			// The function `processRequests` *consumed* the batchController.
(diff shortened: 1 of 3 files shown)


client/tso_dispatcher.go
```diff
@@ -328,10 +328,10 @@ tsoBatchLoop:
 		case td.tsDeadlineCh <- dl:
 		}
 		// processRequests guarantees that the collected requests could be finished properly.
-		err = td.processRequests(stream, dc, batchController, done)
+		tbcConsumed, err := td.processRequests(stream, dc, batchController, done, td.useAsyncStream())
```
lhy1024, line 331: This means that if we don't turn on the newly introduced feature, it will always work in a synchronized way, do I understand correctly?

MyonKeminta: Yes. But let's hold this PR for now as the performance regression is currently very likely to be a misreport due to the unstable test environment.

Issue #4373, opened 2021-11-24
Provide PD self-protection mechanism


rleungx:
## Background
We have met many cases which crush down the PD leader or make the PD leader overloaded. Under these circumstances, PD cannot provide important services such as TSO. It has a large impact on high availability and performance. 

In the current implementation, PD doesn't have any self-protection mechanism. In other words, any component which accesses etcd directly or PD API may lead to the above things happening, like TiDB, TiFlash, TiCDC, etc. Even we manually request to PD through the raw HTTP API or pd-ctl may cause PD unavailable.

## Development task
- [ ] https://github.com/tikv/pd/issues/4349
- [x] https://github.com/tikv/pd/issues/4207
- [x] https://github.com/tikv/pd/issues/5739

Some optional tasks:
- [x] https://github.com/tikv/pd/issues/2971
- [x] https://github.com/tikv/pd/issues/3111
- [x] https://github.com/tikv/pd/issues/4105
- [x] https://github.com/tikv/pd/issues/3383

## Testing
TODO

PR #8653, opened 2024-09-24
pkg: delete config when QPS and concurrency are both deleted

rleungx:

Issue Number: Ref #4373.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/ratelimit/limiter.go
+++ b/pkg/ratelimit/limiter.go
@@ -41,7 +41,10 @@ type limiter struct {
 }
 
 func newLimiter() *limiter {
-	return &limiter{}
+	lim := &limiter{
+		concurrency: NewConcurrencyLimiter(0),
+	}
+	return lim
 }
 
 func (l *limiter) getConcurrencyLimiter() *ConcurrencyLimiter {
@@ -57,7 +60,7 @@ func (l *limiter) getRateLimiter() *RateLimiter {
 }
 
 func (l *limiter) isEmpty() bool {
-	return l.concurrency == nil && l.rate == nil
+	return (l.concurrency == nil || l.concurrency.limit == 0) && l.rate == nil
 }
 
 func (l *limiter) getQPSLimiterStatus() (limit rate.Limit, burst int) {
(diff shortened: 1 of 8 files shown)


pkg/ratelimit/option.go
```diff
@@ -78,7 +78,11 @@ func UpdateDimensionConfig(cfg *DimensionConfig) Option {
 			return InAllowList
 		}
 		lim, _ := l.limiters.LoadOrStore(label, newLimiter())
-		return lim.(*limiter).updateDimensionConfig(cfg)
+		status := lim.(*limiter).updateDimensionConfig(cfg)
+		if status&QPSDeleted != 0 && status&ConcurrencyDeleted != 0 {
+			l.limiters.Delete(label)
```
lhy1024, line 83: What will happen if this is not removed?

rleungx: leave a config in output that is meaningless

lhy1024 (approved):

okJiang (approved):

rleungx:
I made more changes. PTAL @lhy1024 @okJiang

pkg/ratelimit/option.go
```diff
@@ -89,6 +83,6 @@ func InitLimiter() Option {
 			return InAllowList
 		}
 		l.limiters.LoadOrStore(label, newLimiter())
-		return ConcurrencyChanged
+		return LimiterNotChanged
```
okJiang, line 86: Has there been a change here, as expected?

pkg/ratelimit/limiter.go
```diff
@@ -89,47 +82,51 @@ func (l *limiter) getConcurrencyLimiterStatus() (limit uint64, current uint64) {
...
-		return QPSDeleted
+		return LimiterNotChanged
 	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	if l.rate != nil {
+		if limit <= eps || burst < 1 {
+			l.rate = nil
+			if l.isEmpty() {
+				return LimiterDeleted
```
okJiang, line 116: Should this be `LimiterNoChanged`?

pkg/ratelimit/option.go
rleungx: Yes, the limit has not been changed when we initialize the limiter.

pkg/ratelimit/limiter.go
rleungx: No, LimiterDeleted indicates that we should remove the corresponding label from the config.

lhy1024 (approved):

okJiang (approved):

PR #8702, opened 2024-10-14
statistics: add gc in hot peer cache

### LINKED ISSUE #8698: Region may not be removed in hot peer cache after merging
## Bug Report

If the region is not completely cooled down and the merge happens, it might be left in the hot peer cache and never be cleaned up. The merged region will never send a heartbeat again, resulting in checkExpiredTask not working.

lhy1024:

Issue Number: Close #8698

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/statistics/hot_peer_cache.go
```diff
@@ -115,6 +116,7 @@ func (f *HotPeerCache) UpdateStat(item *HotPeerStat) {
 		return
 	}
 	f.incMetrics(item.actionType, item.StoreID)
+	f.removeExpiredItems()
```
rleungx, line 119: Will it be called every time we receive a region heartbeat?

lhy1024: L554-L556

pkg/statistics/hot_cache.go
```diff
@@ -80,7 +80,7 @@ func (w *HotCache) CheckReadAsync(task func(cache *HotPeerCache)) bool {
 func (w *HotCache) RegionStats(kind utils.RWType, minHotDegree int) map[uint64][]*HotPeerStat {
 	ret := make(chan map[uint64][]*HotPeerStat, 1)
 	collectRegionStatsTask := func(cache *HotPeerCache) {
-		ret <- cache.RegionStats(minHotDegree)
+		ret <- cache.PeerStats(minHotDegree)
```
rleungx, line 83: Why change the name, it's not consistent with other places.

lhy1024: Okay, I'll keep the TODO and change them in another PR uniform.

rleungx (approved):

okJiang (approved):

Issue #8593, opened 2024-09-04
introduce callerID for grpc client and apply it for GetRegions


okJiang:
## Enhancement Task

`GetRegion(s)` is a hot path and an accident-prone area in PD. In order to better troubleshoot `GetRegion(s)` in the event of an incident, we decided to record callerID information for it to make it easier to know who was frantically calling `GetRegion(s)` in the meantime.

okJiang:
I would like to put the `CallerID` and `CallerComponent` information into the pd header, e.g. https://github.com/okJiang/kvproto/commit/9835bbc4473a0698f62fff971cad6ba43274ce85. `CallerID/CallerComponent` can be passed in via `ClientOption` when the client is created, and should also be allowed to be passed in temporarily each time the interface is called. The priority of `ClientOption` should be `GetRegion(xxx, xxx, ...ClientOption)` > `CreateClient(xxx, xxx, ...ClientOption)`.

There is a problem, in the client side, there are very many kinds of options, such as `GetStoreOp`, `RegionsOp`, `GetRegionOp`, due to the client interface in a function can only have one kind of option exists, I tend to unify these `Option` as `ClientOption`, all following the rules in the previous paragraph.

What do you think? @rleungx @JmPotato @lhy1024 @nolouch

JmPotato:
What do you think about this approach to chain calling like the HTTP client? Instead of modifying the function signature, we can return a new client with the `CallerID` set. We also need to ensure that multiple references to the same client work correctly.

https://github.com/pingcap/tidb/blob/26443dab89ba5b1a866de3c3e44e2f51468bc435/pkg/store/helper/helper.go#L108

okJiang:
@JmPotato I think that's a good way to solve this problem as well 👍

okJiang:
TiDB uses the client-go to call GetRegion()

TiCDC both uses the client-go and pd-client(client.ScanRegion())

BR both uses client-go(RegionCache.LocateKey()) and pd-client(client.GetRegion())

Lightning uses pd-client(client.GetRegionByID(), client.ScanRegion())

Since client-go(RegionCache) is a separate repository, we need to additionally identify ourselves when we create the RegionCache so that the RegionCache knows who it is when it calls pd-client.

PR #8704, opened 2024-10-15
grpc: add caller info to client and wrap innerClient

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/8593

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/caller/constant.go
```diff
@@ -0,0 +1,46 @@
...
+	// Component represents the caller component.
+	Component string
+)
+
+// nolint:exported
+const (
+	Unknown   ID = "unknown"
+	Tidb      ID = "tidb"
+	Pd        ID = "pd"
+	Tikv      ID = "tikv"
```
JmPotato, line 29: I prefer using a more formal name.

```suggestion
	TiDB      ID = "tidb"
	PD        ID = "pd"
	TiKV      ID = "tikv"
```

client/caller/constant.go
```diff
@@ -0,0 +1,46 @@
...
+	TiFlash   ID = "tiflash"
+	Lightning ID = "lightning"
+	Dumpling  ID = "dumpling"
+	Br        ID = "br"
+	Dm        ID = "dm"
+	Cdc       ID = "cdc"
```
JmPotato, line 35: Ditto.

client/client.go
```diff
@@ -141,6 +141,11 @@ type RPCClient interface {
 	// SetExternalTimestamp sets external timestamp
 	SetExternalTimestamp(ctx context.Context, timestamp uint64) error
 
+	// WithCallerID returns a new RPCClient with the specified caller ID.
```
JmPotato, line 144: What about doing a simple explanation here to explain the differences between the called ID and caller component?

client/client.go
```diff
@@ -212,22 +217,10 @@ func (k *serviceModeKeeper) close() {
...
-	updateTokenConnectionCh chan struct{}
+	callerID        caller.ID
+	callerComponent caller.Component
 
-	ctx    context.Context
-	cancel context.CancelFunc
-	wg     sync.WaitGroup
-	tlsCfg *tls.Config
-	option *option
+	inner *innerClient
```
JmPotato, line 223: A nested struct without introducing the field `inner` may reduce the code changes. But I'm not sure whether the pointer here still works, maybe we can have a try.

client/caller/constant.go
okJiang: For example?

client/client.go
okJiang: - Caller ID can be understood as a binary file; it is a process.  
- Caller component refers to the components within the process.

client/client.go
okJiang: > But I'm not sure whether the pointer here still works, maybe we can have a try.

I tried it before and indeed encountered some issues, so I wrapped the inner client.

client/client.go
JmPotato: I mean write them into the comments.

client/client.go
okJiang: And I think wrapping `innerclient` helps to reduce the amount of code in client.go, as the code in client.go is a bit excessive.

client/client.go
```diff
@@ -141,6 +141,13 @@ type RPCClient interface {
 	// SetExternalTimestamp sets external timestamp
 	SetExternalTimestamp(ctx context.Context, timestamp uint64) error
 
+	// WithCallerID returns a new RPCClient with the specified caller ID.
+	// Caller ID can be understood as a binary file; it is a process.
+	WithCallerID(callerID caller.ID) RPCClient
+	// WithCallerComponent returns a new RPCClient with the specified caller component.
+	// Caller component refers to the components within the process.
+	WithCallerComponent(callerComponent caller.Component) RPCClient
```
nolouch, line 149: Can we provide an automatic method to ingest the CallerID and Component? Like:

```
package processutil

import "os"

// GetCallerID returns the name of the currently running process
func GetCallerID() string {
	return os.Args[0]
}


func GetTheComponent() string {
	// Get the program counter for the calling function
	pc, _, _, ok := runtime.Caller(1)
	if !ok {
		return "unknown"
	}

	// Retrieve the full function name, including the package path
	fullFuncName := runtime.FuncForPC(pc).Name()

	// Find the last slash, which separates the package path from the function name
	lastSlash := strings.LastIndex(fullFuncName, "/")
	if lastSlash == -1 {
		// No slash, use the last period instead, which separates the package and function
		lastSlash = strings.LastIndex(fullFuncName, ".")
	}
	// Extract the package name
	return fullFuncName[:lastSlash]
}

```

okJiang: Very good suggestion👍, I will consider if there are any flaws

JmPotato: Perhaps the HTTP client could also adopt this method.

okJiang: I have added these two functions to this pr. In the default situation, `CreateClient` will set the `callerID` and `callerComponent` by these two functions. I removed `callerID` input from `CreateClient` and keep the `callerComponent` to allow caller set component.

And I expose the upperLayer to allow user to set it. For example, client-go can set it to catch the real caller in TiDB.

How do you think about these changes?

client/client.go
```diff
@@ -440,192 +453,72 @@ func newClientWithKeyspaceName(
...
 	clientCtx, clientCancel := context.WithCancel(ctx)
 	c := &client{
-		keyspaceID:              nullKeyspaceID,
-		updateTokenConnectionCh: make(chan struct{}, 1),
-		ctx:                     clientCtx,
-		cancel:                  clientCancel,
-		svrUrls:                 svrAddrs,
-		tlsCfg:                  tlsCfg,
-		option:                  newOption(),
+		callerID:        caller.GetCallerID(),
```
nolouch, line 456: why not move to innerClient, and make callerComponent as `ClientOption`?

nolouch (commented):
rest lgtm

client/client.go
okJiang: What you said is exactly what I thought at the beginning. You can see it [here](https://github.com/tikv/pd/issues/8593#issuecomment-2354705326)

The `ClientOption` is not actually universal; if we want to apply it across all interfaces, we need to further organize the various options. Therefore, I adopted @JmPotato 's suggestion. In my view, I can accept both implementation methods.

okJiang: Yes, I intend to require the client to set it.

okJiang:
ptal @JmPotato @nolouch 

The conflicts are too much... Let me fix them after your approve😭

client/inner_client.go
```diff
@@ -0,0 +1,244 @@
...
+
+func (c *innerClient) getOrCreateGRPCConn() (*grpc.ClientConn, error) {
+	cc, err := c.pdSvcDiscovery.GetOrCreateGRPCConn(c.pdSvcDiscovery.getLeaderURL())
+	if err != nil {
+		return nil, err
+	}
+	return cc, err
+}
+
+func (c *innerClient) dispatchTSORequestWithRetry(ctx context.Context) TSFuture {
```
nolouch, line 211: Why have these changes?

client/client.go
```diff
@@ -1597,10 +1437,34 @@ func (c *client) respForErr(observer prometheus.Observer, start time.Time, err e
...
+func (c *client) WithCallerComponent(callerComponent caller.Component) RPCClient {
+	newClient := *c
+	newClient.callerComponent = callerComponent
+	return &newClient
+}
+
+// adjustCallerComponent returns the caller component if it is empty, it
+// is the upper layer of the pd client.
+func adjustCallerComponent(callerComponent caller.Component) caller.Component {
+	if len(callerComponent) != 0 {
```
JmPotato, line 1458: Better trim it first before the check.

okJiang: Trim space? https://github.com/tikv/pd/pull/8704/commits/de83c86db37420405deea689e68894c9d39fc2be

JmPotato (approved):

PR #8517, opened 2024-08-12
*: support dynamic tso service

rleungx:
### What problem does this PR solve?

Issue Number: ref #8477

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

PR #8720, opened 2024-10-16
*: move tso to independent thread

rleungx:

Issue Number: Ref #8477.

This is a part of #8517.  

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -203,7 +203,9 @@ func (k *serviceModeKeeper) close() {
 	defer k.Unlock()
 	switch k.serviceMode {
 	case pdpb.ServiceMode_API_SVC_MODE:
-		k.tsoSvcDiscovery.Close()
+		if k.tsoSvcDiscovery != nil {
+			k.tsoSvcDiscovery.Close()
+		}
 		fallthrough
 	case pdpb.ServiceMode_PD_SVC_MODE:
 		if k.tsoClient != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -203,14 +203,10 @@ func (k *serviceModeKeeper) close() {
 	defer k.Unlock()
 	switch k.serviceMode {
 	case pdpb.ServiceMode_API_SVC_MODE:
-		if k.tsoSvcDiscovery != nil {
-			k.tsoSvcDiscovery.Close()
-		}
+		k.tsoSvcDiscovery.Close()
 		fallthrough
 	case pdpb.ServiceMode_PD_SVC_MODE:
-		if k.tsoClient != nil {
-			k.tsoClient.close()
-		}
+		k.tsoClient.close()
 	case pdpb.ServiceMode_UNKNOWN_SVC_MODE:
 	}
 }
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -17,6 +17,7 @@ package cluster
 import (
 	"context"
 	"encoding/json"
+	errorspkg "errors"
 	"fmt"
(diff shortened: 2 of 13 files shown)


client/client.go
```diff
@@ -203,7 +203,9 @@ func (k *serviceModeKeeper) close() {
 	defer k.Unlock()
 	switch k.serviceMode {
 	case pdpb.ServiceMode_API_SVC_MODE:
-		k.tsoSvcDiscovery.Close()
+		if k.tsoSvcDiscovery != nil {
```
JmPotato, line 206: What about moving this check into `Close()`?

server/cluster/cluster.go
```diff
@@ -390,25 +400,85 @@ func (c *RaftCluster) checkServices() {
 	}
 }
 
+// checkTSOService checks the TSO service.
+func (c *RaftCluster) checkTSOService() {
+	if !c.isAPIServiceMode {
```
JmPotato, line 405:
```suggestion
	if c.isAPIServiceMode {
		return
	}
```

server/cluster/cluster.go
```diff
@@ -390,25 +400,85 @@ func (c *RaftCluster) checkServices() {
...
+		if c.member.IsLeader() {
+			if err := c.startTSOJobs(); err != nil {
+				// If there is an error, need to wait for the next check.
```
JmPotato, line 408: Do we need log here?

server/cluster/cluster.go
```diff
@@ -390,25 +400,85 @@ func (c *RaftCluster) checkServices() {
...
+				return
+			}
+		} else {
+			// leader exits, reset the allocator group
+			if err := c.stopTSOJobs(); err != nil {
+				// If there is an error, need to wait for the next check.
```
JmPotato, line 414: Ditto.

server/cluster/cluster.go
```diff
@@ -314,11 +322,13 @@ func (c *RaftCluster) Start(s Server) error {
 	if err != nil {
 		return err
 	}
+	c.checkTSOService()
```
okJiang, line 325: Can we put this func next to `checkSchedulingService`?

okJiang (commented):
rest lgtm

server/server.go
```diff
@@ -1715,29 +1714,6 @@ func (s *Server) campaignLeader() {
 	s.member.KeepLeader(ctx)
 	log.Info(fmt.Sprintf("campaign %s leader ok", s.mode), zap.String("campaign-leader-name", s.Name()))
 
-	if !s.IsAPIServiceMode() {
```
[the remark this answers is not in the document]
rleungx: it still works

server/cluster/cluster.go
rleungx: prefer to check it immediately

nolouch (approved):

server/cluster/cluster.go
```diff
@@ -390,25 +400,76 @@ func (c *RaftCluster) checkServices() {
 	}
 }
 
+// checkTSOService checks the TSO service.
+func (c *RaftCluster) checkTSOService() {
```
okJiang, line 404: Seems we do not need this wrap function now. We can call `startTSOJobs` directly.

rleungx: for the upcomming PR

okJiang (approved):

PR #8738, opened 2024-10-22
server: refactor the independent service check

rleungx:

Issue Number: ref #8477.

This is a part of https://github.com/tikv/pd/pull/8517.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/forward.go
+++ b/server/forward.go
@@ -133,66 +133,82 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 			err = errs.ErrGenerateTimestamp.FastGenByArgs("tso count should be positive")
 			return status.Error(codes.Unknown, err.Error())
 		}
-
-		forwardedHost, ok := s.GetServicePrimaryAddr(stream.Context(), constant.TSOServiceName)
-		if !ok || len(forwardedHost) == 0 {
-			tsoStreamErr = errors.WithStack(ErrNotFoundTSOAddr)
+		forwardCtx, cancelForward, forwardStream, lastForwardedHost, tsoStreamErr, err = s.handleTSOForwarding(forwardCtx, forwardStream, stream, server, request, tsDeadlineCh, lastForwardedHost, cancelForward)
+		if tsoStreamErr != nil {
 			return tsoStreamErr
 		}
-		if forwardStream == nil || lastForwardedHost != forwardedHost {
-			if cancelForward != nil {
-				cancelForward()
-			}
-
-			clientConn, err := s.getDelegateClient(s.ctx, forwardedHost)
-			if err != nil {
-				tsoStreamErr = errors.WithStack(err)
-				return tsoStreamErr
-			}
-			forwardStream, forwardCtx, cancelForward, err = createTSOForwardStream(stream.Context(), clientConn)
-			if err != nil {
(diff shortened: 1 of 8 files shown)


server/server.go
```diff
@@ -1417,6 +1417,18 @@ func (s *Server) GetRaftCluster() *cluster.RaftCluster {
...
 }
 
+// IsServiceIndependent returns whether the service is independent.
+func (s *Server) IsServiceIndependent(name string) bool {
+	if s.mode == APIServiceMode && !s.IsClosed() {
+		// TODO: remove it after we support tso discovery
+		if name == constant.TSOServiceName {
+			return true
+		}
+		return s.cluster.IsServiceIndependent(name)
```
lhy1024, line 1427: Do we need to check whether cluster is nil?

server/forward.go
```diff
@@ -133,66 +133,82 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
 			err = errs.ErrGenerateTimestamp.FastGenByArgs("tso count should be positive")
 			return status.Error(codes.Unknown, err.Error())
 		}
-
-		forwardedHost, ok := s.GetServicePrimaryAddr(stream.Context(), constant.TSOServiceName)
-		if !ok || len(forwardedHost) == 0 {
-			tsoStreamErr = errors.WithStack(ErrNotFoundTSOAddr)
+		forwardCtx, cancelForward, forwardStream, lastForwardedHost, tsoStreamErr, err = s.handleTSOForwarding(forwardCtx, forwardStream, stream, server, request, tsDeadlineCh, lastForwardedHost, cancelForward)
```
lhy1024, line 136: There are so many incoming and returning parameters here

lhy1024 (commented):
the rest LGTM

server/forward.go
rleungx: Yes. But I don't find a better way to avoid it.

server/server.go
rleungx: no need

server/forward.go
okJiang: Is there any benefit after wraping `handleTSOForwarding`?

rleungx: for upcoming PR

server/forward.go
```diff
@@ -133,66 +133,82 @@ func (s *GrpcServer) forwardTSO(stream pdpb.PD_TsoServer) error {
...
-		}
+	response := &pdpb.TsoResponse{
+		Header: &pdpb.ResponseHeader{
+			ClusterId: tsopbResp.GetHeader().GetClusterId(),
+			Error:     pdpbErr,
+		},
+		Count:     tsopbResp.GetCount(),
+		Timestamp: tsopbResp.GetTimestamp(),
+	}
+	if server != nil {
```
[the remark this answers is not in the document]
rleungx: We receive both `stream pdpb.PD_TsoServer` and `server *tsoServer` as the parameters.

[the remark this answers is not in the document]
okJiang: `Server` seems that will not be `nil`?🤔

lhy1024 (approved):

okJiang (approved):

PR #8742, opened 2024-10-23
*: enable more linters

rleungx:

Issue Number: ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


.golangci.yml
```diff
@@ -18,6 +18,17 @@ linters:
...
+    - asasalint
+    - asciicheck
+    - bidichk
+    - durationcheck
+    - gocheckcompilerdirectives
+    - gochecksumtype
+    - makezero
+    - protogetter
+    - reassign
+    - zerologlint
```
okJiang, line 30: Seems we don't use `zerolog`, if we don't use it, I think we don't have to add this linter. Less linters have less check time.

rleungx: nice catch

okJiang (approved):

lhy1024 (approved):
so big

PR #8744, opened 2024-10-25
server: implement dynamic forward 

rleungx:

Issue Number: Ref #8477.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -412,14 +412,14 @@ func (c *RaftCluster) checkTSOService() {
 		if c.opt.GetMicroServiceConfig().IsTSODynamicSwitchingEnabled() {
 			servers, err := discovery.Discover(c.etcdClient, strconv.FormatUint(c.clusterID, 10), constant.TSOServiceName)
 			if err != nil || len(servers) == 0 {
-				if err := c.startTSOJobs(); err != nil {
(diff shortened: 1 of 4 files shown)


nolouch (approved):

okJiang (commented):
can we add any tests for it?

rleungx:
> can we add any tests for it?

This PR doesn't change too much logic when the config value is false. I will add some tests in the next few PRs if we support changing the configuration.

server/cluster/cluster.go
```diff
@@ -403,11 +403,30 @@ func (c *RaftCluster) checkSchedulingService() {
...
+				log.Info("TSO is provided by TSO server")
+				if !c.IsServiceIndependent(constant.TSOServiceName) {
+					c.SetServiceIndependent(constant.TSOServiceName)
+				}
+			}
+		}
 		return
 	}
 
 	if err := c.startTSOJobs(); err != nil {
```
okJiang, line 429: How about renaming it to `startTSOJobsIfNeed`? When `isAPIServiceMode=false`, we will call `startTSOJobs` every time. Starting jobs too much times is strange.

okJiang (commented):
rest lgtm

okJiang (approved):

PR #8749, opened 2024-10-28
api: support to query whether pd has loaded region

### LINKED ISSUE #8426: Alleviate the latency impact when reloading PD
## Enhancement Task
Right now, when we need to reload PD, we need to switch the leader at least twice. The region's cache of PD will be invalid, and we must load them from the persisted storage. If the region synchronizes process is not finished, the leader information is lost since we don't persist it and that information must be accomplished through the region heartbeat, which might cause the GetRegion request to get an unexpected result. Finally, it will trigger the backoff mechanism which results in the latency increase.

### LINKED ISSUE #8748: Improve availability during pd rolling restart
## Enhancement Task

After a PD instance is restarted, it may take some time to load region information. If a PD leader doesn't have all up-to-date region information, it may cause some problems.

When updating/upgrading PD, the TiDB Operator restarts PD instances one by one, e.g. if the current PD instance is ready, the next one will be restarted. But, this ready doesn't include region information sync.

A case with problems:
1. All PD instances run for a while and with all region information synced
2. PD-2 is the leader, and an updating operation is triggered
3. TiDB Operator calls PD API to transfer leader from PD-2 to PD-1
4. TiDB Operator restarts PD-2 and waits for PD-2 to be ready (but without the additional wait for region information sync) 
5. TiDB Operator calls PD API to transfer leader from PD-1 to PD-2
6. As the region information in PD-2 is not synced, problems happen

At this point, the problem arises that PD-2 is elected as the leader, but it can't provide services related to region query until regions are loaded, it can only provide tso services.

The current workaround is to wait a while after the pd rolls reboot until the load region is complete and then let it be the leader
Typically, for a 10 million cluster, five to ten minutes is enough.

But in the longer term, it is more flexible for pd to provide an interface to query if the load region is complete.

lhy1024:

Issue Number: Close #8748, close #8426

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/storage/storage.go
```diff
@@ -77,9 +77,10 @@ type coreStorage struct {
 	Storage
 	regionStorage endpoint.RegionStorage
 
-	useRegionStorage int32
-	regionLoaded     bool
-	mu               syncutil.Mutex
+	useRegionStorage        int32
+	regionLoadedFromDefault bool
+	regionLoadedFromStorage bool
```
rleungx, line 82: Why do we need to distinguish these two kinds?

lhy1024: I want to avoid problems after ps.useRegionStorage switch

server/api/status.go
```diff
@@ -39,11 +40,13 @@ func newStatusHandler(svr *server.Server, rd *render.Render) *statusHandler {
 // @Success  200  {object}  versioninfo.Status
 // @Router   /status [get]
```
lhy1024, line 41: I am not sure whether it is suitable. ref https://github.com/tikv/pd/issues/8748#issuecomment-2440939642

rleungx: How about /member or /health?

pkg/versioninfo/versioninfo.go
```diff
@@ -27,10 +27,11 @@ import (
...
 type Status struct {
-	BuildTS        string `json:"build_ts"`
-	Version        string `json:"version"`
-	GitHash        string `json:"git_hash"`
-	StartTimestamp int64  `json:"start_timestamp"`
+	BuildTS          string `json:"build_ts"`
+	Version          string `json:"version"`
+	GitHash          string `json:"git_hash"`
+	StartTimestamp   int64  `json:"start_timestamp"`
+	AreRegionsLoaded bool   `json:"are_regions_loaded"`
```
rleungx, line 34: prefer to use loaded

server/api/status.go
lhy1024: /members return all member info according etcd.ListEtcdMembers
/health is to query the status of each node through the /ping interface to return information about all the nodes, if we don't add a new api, we won't be able to get the result of whether the load region is complete or not.

pkg/versioninfo/versioninfo.go
lhy1024: > Loaded bool   `json:"loaded"`

?

rleungx: yes

server/api/status.go
rleungx: status is more like some fixed information to me, at least for the current definition.

lhy1024: > put it under /status, which is appropriate from the name, but currently status returns compiled information.

Yes, as I said in the issue.

pkg/storage/storage.go
okJiang: How about using string? 

`regionLoadedFrom`: "", "etcd", "leveldb"

pkg/storage/storage.go
```diff
@@ -133,17 +134,22 @@ func TryLoadRegionsOnce(ctx context.Context, s Storage, f func(region *core.Regi
...
-	defer ps.mu.Unlock()
-	if !ps.regionLoaded {
+	if !ps.regionLoadedFromStorage {
 		if err := ps.regionStorage.LoadRegions(ctx, f); err != nil {
 			return err
 		}
-		ps.regionLoaded = true
+		ps.regionLoadedFromStorage = true
 	}
 	return nil
```
okJiang, line 154: Seems we can use `*coreStorage.LoadRegions()` directly?

pkg/storage/storage.go
lhy1024: I don't think so, we just have these two ways now

pkg/storage/storage.go
lhy1024: It doesn't check ps.regionLoadedFromStorage

pkg/storage/storage.go
```diff
@@ -77,9 +77,10 @@ type coreStorage struct {
 	Storage
 	regionStorage endpoint.RegionStorage
 
-	useRegionStorage int32
-	regionLoaded     bool
-	mu               syncutil.Mutex
+	useRegionStorage        int32
```
niubell, line 80: Is that safe to switch between using or not using RegionStore in the master branch or the latest releases now?

lhy1024: defaultUseRegionStorage is true and only check switch when it was just elected leader.

pkg/storage/storage.go
```diff
@@ -73,13 +73,21 @@ func NewRegionStorageWithLevelDBBackend(
...
+)
+
 type coreStorage struct {
 	Storage
 	regionStorage endpoint.RegionStorage
 
-	useRegionStorage int32
-	regionLoaded     bool
-	mu               syncutil.Mutex
+	useRegionStorageFlag int32
```
rleungx, line 88: use bool?

server/api/status.go
```diff
@@ -44,6 +45,7 @@ func (h *statusHandler) GetPDStatus(w http.ResponseWriter, _ *http.Request) {
 		GitHash:        versioninfo.PDGitHash,
 		Version:        versioninfo.PDReleaseVersion,
 		StartTimestamp: h.svr.StartTimestamp(),
+		RegionLoaded:   storage.AreRegionsLoaded(h.svr.GetStorage()),
```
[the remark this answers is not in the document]
niubell: Hi Tema, thanks for your advice
Q1: A general field will be helpful to expand different scenarios, but we found that currently there is only RegionLoaded is needed. If there are other situations in the future, we tend to add new fields with more accurate meanings.
Q2: Standalone PD or Scheduling Service will expose this API(especially the RegionLoaded field), others such as TSO Service don't.
Q3: IMO, it's good we can use the same interface to query service status, but the fields may be different. The benefits of unification may not be obvious, but it is indeed a better way.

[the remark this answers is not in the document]
niubell: PTAL @csuzhangxc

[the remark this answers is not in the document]
csuzhangxc: I agree with Tema, for TiDB Operator, it's better to use a general API (like `/ready`) to check the status in most cases. If this API can't meet some other special cases in the future, then we can consider to add/use some other APIs.

For this RegionLoaded case, in fact, TiDB Operator just wants to know when it can restart the next instance safely.

lhy1024:
@rleungx @JmPotato @okJiang PTAL

server/api/status.go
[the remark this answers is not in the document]
niubell: > @niubell can we land this one and open a separate task for a more general /ready endpoint?
@Tema Agreed, we will merge this PR first. After the general /ready proposal for all components is finalized, PD will also add the /ready API, cc @zhangjinpeng87 @csuzhangxc

okJiang (approved):

pkg/versioninfo/versioninfo.go
```diff
@@ -31,6 +31,7 @@ type Status struct {
 	Version        string `json:"version"`
 	GitHash        string `json:"git_hash"`
 	StartTimestamp int64  `json:"start_timestamp"`
+	RegionLoaded   bool   `json:"region_loaded"`
```
rleungx, line 34: What if the operator/tiup uses this API and PD doesn't provide this field? Will it block the upgrade process?

lhy1024: After discussion, we will use /ready api directly.

lhy1024:
PTAL @rleungx @okJiang

server/api/router.go
```diff
@@ -333,6 +333,7 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 	healthHandler := newHealthHandler(svr, rd)
 	registerFunc(apiRouter, "/health", healthHandler.GetHealthStatus, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(apiRouter, "/ping", healthHandler.Ping, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(apiRouter, "/ready", healthHandler.Ready, setMethods(http.MethodGet), setAuditBackend(prometheus))
```
rleungx, line 336: Shall we use v2?

server/apiv2/router.go
```diff
@@ -56,5 +56,6 @@ func NewV2Handler(_ context.Context, svr *server.Server) (http.Handler, apiutil.
 	handlers.RegisterKeyspace(root)
 	handlers.RegisterTSOKeyspaceGroup(root)
 	handlers.RegisterMicroService(root)
+	handlers.RegisterReadyRouter(root)
```
rleungx, line 59: Do we need a group?

rleungx (approved):

rleungx:
Let's wait for @okJiang to give an approve.

okJiang (approved):

Issue #8683, opened 2024-10-10
Using golang to deploy PD cluster in real cluster test


okJiang:
## Enhancement Task

All current real cluster test cases share a single PD cluster.

1. Different clusters may be required for different test cases to improve the stability of the tests.
2. Some test cases may require multiple PD clusters.

Using golang to deploy PDs is better for test code implementation and maintenance.

PR #8752, opened 2024-10-30
tests: add ms cluster case

rleungx:

Issue Number: ref #8683

```commit-message
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/integrations/realcluster/cluster.go
+++ b/tests/integrations/realcluster/cluster.go
@@ -53,9 +53,9 @@ func (s *clusterSuite) SetupSuite() {
 	for _, match := range matches {
 		require.NoError(t, runCommand("rm", "-rf", match))
 	}
-	s.startRealCluster(t)
+	s.startCluster(t)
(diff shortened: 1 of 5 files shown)


lhy1024 (approved):

tests/integrations/realcluster/cluster.go
```diff
@@ -165,3 +182,58 @@ func waitTiupReady(t *testing.T, tag string) {
 	}
 	require.Failf(t, "TiUP is not ready", "tag: %s", tag)
 }
+
+type msClusterSuite struct {
```
okJiang, line 186: seems we can add a field `ms bool` to `clusterSuite` rather than create a new structure.

okJiang (commented):
rest lgtm

tests/integrations/realcluster/cluster.go
rleungx: good point

okJiang (approved):

Issue #8756, opened 2024-10-31
Evict-leader-scheduler dosen't work when creating it with the same store-id repeatedly


okJiang:
### What did you do?

create evict-leader-scheduler with the same store-id repeatly

### What did you expect to see?

- it works well

### What did you see instead?

- balance-leader-scheduler conflict with evict-leader-scheduler

okJiang:
This issue affects <8.3

okJiang:
this issue is fixed

PR #8623, opened 2024-09-13
scheduler: replace pauseLeader with two flags and add source filter to transferIn

lhy1024:
### What problem does this PR solve?

Issue Number: Close #8621

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
Before this pr:
image
After this pr:
image

### Release note

```release-note
None.
```

PR #8757, opened 2024-10-31
scheduler: add test for creating evict-leader-scheduler twice

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8756

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


okJiang:
Grant-leader exists bug. It will not work in some specific condition... I have to find it firstly. So this pr hold

okJiang:
> Grant-leader exists bug. It will not work in some specific conditions... I have to find it firstly. So this pr hold

The issue has been fixed by https://github.com/tikv/pd/pull/8623/files#diff-a220672e83f3e5fa65303ae268e69c0e4bcabebd6d3b5d1c81898504f9ac5095R496-R503.

Assuming there are three nodes: store0, store1, and store2, we set up `scheduler add grant-leader-scheduler store0`.

Then, when the balance-leader-scheduler selects a target, it cannot be store0, but it may choose store1 or store2. Let's assume target=store1. Ultimately, when creating the transfer-leader-operator for the target, there will be no filtering on the source, which leads to the creation of a transfer-leader-operator from store0 to store1.

This is a grant-leader-scheduler problem. So we do not need to pick it to previous versions.

lhy1024 (approved):

tests/integrations/realcluster/scheduler_test.go
```diff
@@ -201,3 +203,72 @@ func (s *schedulerSuite) TestRegionLabelDenyScheduler() {
...
+
+	var i int
+	evictLeader := func() {
+		re.NoError(pdHTTPCli.CreateScheduler(ctx, types.EvictLeaderScheduler.String(), uint64(region1.Leader.StoreID)))
+		// if the second evict leader scheduler cause the pause-leader-filter
+		// disable, the balance-leader-scheduler need some time to transfer
+		// leader. See details in https://github.com/tikv/pd/issues/8756.
+		if i == 1 {
+			time.Sleep(3 * time.Second)
+		}
```
JmPotato, line 226: I think we should cover this kind of waiting inside the `testutil.Eventually` below.

JmPotato (approved):

PR #8768, opened 2024-11-04
test: add ut to test the input when creating evict-leader scheduler

okJiang:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/api/scheduler.go
```diff
@@ -113,11 +113,16 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
 			return
 		}
 	case types.GrantLeaderScheduler, types.EvictLeaderScheduler:
-		storeID, ok := input["store_id"].(float64)
+		_, ok := input["store_id"]
```
rleungx, line 116: Why?

okJiang: `_, ok := input["store_id"]` represents the input("store_id") doesn't exist.

`storeID, ok := input["store_id"].(float64)` represents the input("store_id") can not be convert to float.

rleungx: Is it necessary?

okJiang: I think we can put it here. It won't be bad at all.

rleungx (approved):

nolouch (approved):

PR #8786, opened 2024-11-07
*: reduce raft cluster lock and replace some place with atomic

### LINKED ISSUE #8785: *: split raft cluster lock

lhy1024:

Issue Number: Close #8785

```commit-message
```

### Check List

Tests

- Unit test
### Release note

```release-note
None.
```


server/cluster/cluster.go
```diff
@@ -2319,25 +2318,19 @@ func (c *RaftCluster) loadMinResolvedTS() {
...
 	if !c.isInitialized() {
 		return math.MaxUint64
 	}
-	return c.minResolvedTS
+	return c.minResolvedTS.Get()
 }
 
 // GetStoreMinResolvedTS returns the min resolved ts of the store.
 func (c *RaftCluster) GetStoreMinResolvedTS(storeID uint64) uint64 {
-	c.RLock()
```
lhy1024, line 2339: it is not necessary

okJiang (commented):
Why not use atomic.Value?

lhy1024:
> Why not use atomic.Value?

Sure, at first I thought it was possible that we'd need to manually lock and unlock to control more of the code, but now it seems that atomic is sufficient for these three places

JmPotato (approved):

Issue #8582, opened 2024-08-28
etcd key path is so difficult to read


okJiang:
## Enhancement Task

update them to string rather than several function nesting. In this case it is straightforward to use the function to generate the final key path yet, without any change in usage, but we can clearly see the format of each path.

https://github.com/tikv/pd/blob/master/pkg/storage/endpoint/key_path.go

PR #8789, opened 2024-11-07
etcd/path: change leaderPath/dcLocationPath/memberPath/allocalIDPath to absolute path

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/8582

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/member/member.go
```diff
@@ -488,17 +484,9 @@ func (m *EmbeddedEtcdMember) SetMemberDeployPath(id uint64) error {
 	return nil
 }
 
-func (m *EmbeddedEtcdMember) getMemberGitHashPath(id uint64) string {
```
nolouch, line 491: Do we need to add a test to check if the changes produce the same results as before? For example, we could output the path from the old version as a dataset for the new code test.

okJiang: All I can think of is to verify the consistency of the paths through manual testing. There doesn't seem to be any good testing methods to automate🤔

nolouch: Maybe we can use fixed testdata from your manual testing.

nolouch (approved):
lgtm!

tests/integrations/realcluster/download_integration_test_binaries.sh
```diff
@@ -28,6 +28,8 @@ tidb_download_url="${file_server_url}/download/builds/pingcap/tidb/${tidb_sha1}/
 tikv_download_url="${file_server_url}/download/builds/pingcap/tikv/${tikv_sha1}/centos7/tikv-server.tar.gz"
 tiflash_download_url="${file_server_url}/download/builds/pingcap/tiflash/${branch}/${tiflash_sha1}/centos7/tiflash.tar.gz"
 
+ETCD_VERSION=v3.5.9
```
rleungx, line 31: why 3.5.9?

okJiang: No special meaning

okJiang: It is used to download etcdctl

tests/integrations/realcluster/etcd_key_test.go
```diff
@@ -0,0 +1,146 @@
...
+		"/pd//leader",
+		"/pd//member//binary_version",
+		"/pd//member//deploy_path",
+		"/pd//member//git_hash",
+		"/pd//raft",
+		"/pd//raft/min_resolved_ts",
+		"/pd//raft/r/",
+		"/pd//raft/s/",
+		"/pd//raft/status/raft_bootstrap_time",
+		"/pd//region_label/keyspaces/",
```
rleungx, line 68: Can you explain the difference between `//` and `/`?

okJiang: To ensure consistency in each test, I removed the numerical parts in the tests.

okJiang: image

pkg/utils/keypath/key_path_v2.go
```diff
@@ -0,0 +1,101 @@
...
+	GroupID     uint32 // only used for tso keyspace group
+}
+
+// Prefix returns the parent directory of the given path.
+func Prefix(str string) string {
+	return path.Dir(str)
+}
+
+// LeaderPath returns the leader path.
+func LeaderPath(p *MsParam) string {
```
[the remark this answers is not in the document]
okJiang: I prefer to keeping this way. Because it is only called in a few places, and it is wrapped by its upper layer yet.

tests/integrations/realcluster/download_integration_test_binaries.sh
rleungx: It should be 3.5.15?

okJiang: They are all ok for me

okJiang: updated

pkg/utils/keypath/key_path_v2.go
```diff
@@ -0,0 +1,101 @@
...
+		if p.GroupID == 0 {
+			return fmt.Sprintf(msTsoDefaultLeaderPathFormat, ClusterID())
+		}
+		return fmt.Sprintf(msTsoKespaceLeaderPathFormat, ClusterID(), p.GroupID)
+	}
+	return fmt.Sprintf(msLeaderPathFormat, ClusterID(), p.ServiceName)
+}
+
+// DCLocationPath returns the dc-location path.
+func DCLocationPath(p *MsParam, memberID uint64) string {
```
rleungx, line 65: Can we just remove them?

okJiang: Considering the server side code of Local TSO is not removed yet, it is better to keep them now, and remove them later

tests/integrations/realcluster/etcd_key_test.go
rleungx: I'm afraid it's hard to maintain later.

okJiang: Listing them one by one is somewhat challenging, but it does have two main benefits:  
1. It adds tests for them.  
2. Through testing, we also clearly understand in which scenarios different keys will be generated.

Moreover, I really haven't thought of any better way to achieve it. - -

tests/integrations/realcluster/etcd_key_test.go
```diff
@@ -0,0 +1,146 @@
...
+		"/pd//rule_group/tiflash",
+		"/pd//rules/-c",
```
rleungx, line 70: What is it?

okJiang: I think it is the default rule key

https://github.com/tikv/pd/blob/be4a366da715ff3eeb628d98248068364be8ef91/pkg/schedule/placement/rule_manager.go#L43-L46

https://github.com/tikv/pd/blob/be4a366da715ff3eeb628d98248068364be8ef91/pkg/schedule/placement/rule_manager.go#L134

 that has been encoded by

https://github.com/tikv/pd/blob/be4a366da715ff3eeb628d98248068364be8ef91/pkg/schedule/placement/rule.go#L104-L107


And after removing the number from the key, it is `/pd//rules/-c`

okJiang:
```golang
func TestEncodeToString(t *testing.T) {
	require.Equal(t, hex.EncodeToString([]byte("pd")), "7064")
	require.Equal(t, hex.EncodeToString([]byte("default")), "64656661756c74")
}
```

okJiang: ping @rleungx

pkg/member/participant.go
```diff
@@ -87,10 +85,9 @@ func (m *Participant) InitInfo(p participant, rootPath string, leaderName string
 	m.member = p
 	m.memberValue = string(data)
 	m.rootPath = rootPath
```
rleungx, line 87: Do we still need it?

rleungx, line 87: Do we still need it?

okJiang: Yes, this issue does not close. `getMemberLeaderPriorityPath` still uses `rootPath`.

I will replace them one by one and try to remove `rootPath`

rleungx (approved):

Issue #8802, opened 2024-11-12
Eliminate all code related to Local TSO


JmPotato:
As we're working on introducing #8690 to the PD client, the code reuse of stream batch-related work is necessary in all kinds of terms. However, some legacy code from the Local TSO feature prevents us from resuing and simplifying the code. Considering that this feature will not be GA anymore due to the current implementation of microservice TSO and its code has been lacking maintenance for a long time, this issue proposes eliminating all code related to Local TSO.

This should bring us these pros:

- Simplify the TSO code of both the PD server and PD client.
- Allow us to abstract the PD client infrastructure further.

okJiang:
LGTM

lhy1024:
Great

PR #8810, opened 2024-11-14
*: remove the Local TSO test cases

JmPotato:

Issue Number: ref #8802.

```commit-message
Remove the Local TSO test cases.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


tests/server/tso/tso_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 TiKV Project Authors.
+// Copyright 2024 TiKV Project Authors.
```
rleungx, line 1: Why change it?

rleungx (approved):

tests/server/tso/tso_test.go
JmPotato: The original content of `tso_test.go` has been modified, and the test cases from `global_tso_test.go` have been added, so I simply consider it a new file.

okJiang (approved):

Issue #8690, opened 2024-10-11
Support batch processing with the GetRegion stream call


JmPotato:
We have encountered several issues related to the `GetRegion` call and region cache updates, such as #2843 and #8426. At present, a highly feasible solution to alleviate this type of problem is to change `GetRegion` from a unary gRPC call to a stream call that supports batch processing, which should reduce the CPU pressure on PD in scenarios such as cache expiration and error retries strom.

- [x] Implement the gRPC stream version of `GetRegion` which supports handling various queries, such as `GetPrevRegion` and `GetRegionByID`.
- [x] Implement the batch processing logic within the PD client.
- [ ] Integrate the new `GetRegion` method into the client-go/TiDB.
- [ ] Conduct the benchmark tests with the previous version to evaluate the QPS/CPU usage performance.

PR #8813, opened 2024-11-14
*: independent the client option

JmPotato:

Issue Number: ref #8690.

```commit-message
Independent the client option to reduce the number of files under the client root directory.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/opt/option.go
+++ b/client/opt/option.go
@@ -1,4 +1,4 @@
-// Copyright 2021 TiKV Project Authors.
+// Copyright 2024 TiKV Project Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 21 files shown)


okJiang (approved):

client/opt/option.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package pd
+package opt
```
rleungx, line 15: Will it have a compatibility issue?

JmPotato: Yes, when upgrading the downstream components, manual modification of the relevant code is required.

rleungx (approved):

PR #8833, opened 2024-11-20
client: separate the metrics package

JmPotato:

Issue Number: ref #8690.

```commit-message
Separate the client metrics package.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/metrics/metrics.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// RequestDurationTSO records the durations of the successful TSO requests.
+	RequestDurationTSO prometheus.Observer
+	// RequestFailedDurationTSO records the durations of the failed TSO requests.
+	RequestFailedDurationTSO prometheus.Observer
+)
+
+func initCmdDurations() {
+	// WithLabelValues is a heavy operation, define variable to avoid call it every time.
+	CmdDurationTSOWait = cmdDuration.WithLabelValues("wait")
```
okJiang, line 204: We can init them in L150-L199 directly

JmPotato: It is by design that we have a function to do this so we can call `InitAndRegisterMetrics` to reinitialize them with a client option.

rleungx (approved):

okJiang (approved):

PR #8849, opened 2024-11-25
grpc/service: Refactor middleware

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8850

```commit-message
Organize the middleware code and handle them centrally.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -1,0 +1,151 @@
+// Copyright 2024 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/kvproto/pkg/pdpb"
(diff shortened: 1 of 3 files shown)


server/middleware.go
```diff
@@ -0,0 +1,205 @@
...
+	"google.golang.org/grpc"
+)
+
+type request interface {
+	GetHeader() *pdpb.RequestHeader
+}
+
+type forwardFn func(ctx context.Context, client *grpc.ClientConn, request any) (any, error)
+
+var forwardFns = map[string]forwardFn{
```
rleungx, line 35: I think it's more complicated.

okJiang: Isn't it easier to manage after the modifications compared to the previous definition scattered across each function? Could you elaborate on your viewpoint?

rleungx: There are too many hardcode things and you still need to add them one by one.

okJiang: From the current implementation, it seems that hardcoding cannot be avoided (it was hardcoded before as well). Based on the existing changes, where do you think it has become more complex?

rleungx: We don't need to care about the function name for now.

okJiang: If we need to maintain it, we have to pay attention to the function names, right? I think we should know which functions support a certain middleware and which do not.
> don't need to care about the function name

Or do you mean that all functions should support the forward middleware, so we don't need to worry about the function names? If that's the case, I don't think the impact is significant; placing a map here doesn't imply that we need to be overly concerned about function names, and it doesn't add much burden. If we didn't have this map, we would still need to define a `forwardFn` inside the function when adding new functions, which makes no difference.

If we still hold different views, perhaps we should listen to the opinions of other members. cc @JmPotato @nolouch @lhy1024

okJiang:
I will fix the problem from ci after discussion

okJiang:
Since there are no obvious benefits, this issue will be closed for now.

PR #8856, opened 2024-11-26
client: introduce circuit breaker for region calls

Tema:

Issue Number: Ref https://github.com/tikv/pd/issues/8678, https://github.com/tikv/rfcs/pull/115

This PR implements circuit breaker from https://github.com/tikv/rfcs/pull/115 and integrates it into http and grpc PD clients for get region metadata calls, but disabled by default.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/circuit_breaker/circuit_breaker.go
```diff
@@ -0,0 +1,296 @@
...
+	"time"
+
+	"github.com/pingcap/log"
+)
+
+// ErrOpenState is returned when the CircuitBreaker is open or half-open with pending requests.
+var ErrOpenState = errors.New("circuit breaker is open")
+
+// Overloading is a type describing service return value
+type Overloading int
```
rleungx, line 33: How about just using bool?

client/http/interface.go
```diff
@@ -541,6 +543,11 @@ func (c *client) GetReplicateConfig(ctx context.Context) (map[string]any, error)
 	return config, nil
 }
 
+// UpdateCircuitBreakerSettings updates config for circuit breaker
```
rleungx, line 546: Let's focus on gRPC calls in this PR.

client/circuit_breaker/circuit_breaker.go
```diff
@@ -0,0 +1,296 @@
...
+	"go.uber.org/zap"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pingcap/log"
+)
+
+// ErrOpenState is returned when the CircuitBreaker is open or half-open with pending requests.
+var ErrOpenState = errors.New("circuit breaker is open")
```
rleungx, line 30: How about moving it to errs/errno.go?

client/circuit_breaker/circuit_breaker.go
```diff
@@ -0,0 +1,296 @@
...
+// CircuitBreaker is a state machine to prevent sending requests that are likely to fail.
+type CircuitBreaker[T any] struct {
+	config *Settings
+	name   string
+
+	mutex sync.Mutex
+	state *State[T]
+
+	successCounter  prometheus.Counter
+	failureCounter  prometheus.Counter
```
rleungx, line 73: When will it be used?

rleungx:
> @rleungx can you help to take a look at https://github.com/tikv/pd/actions/runs/12043529055/job/33580779593?pr=8856 as I don't see why it marks it as FAILED while as there is no FAIL in the tests.

```
=== RUN   TestCircuitBreaker_OpenState_Not_Enough_QPS
    circuit_breaker_test.go:58: 
        	Error Trace:	/home/runner/work/pd/pd/client/circuit_breaker/circuit_breaker_test.go:58
        	Error:      	Not equal: 
        	            	expected: circuit_breaker.StateType(0)
        	            	actual  : *circuit_breaker.State[int](&circuit_breaker.State[int]{stateType:0, cb:(*circuit_breaker.CircuitBreaker[int])(0xc00012e7e0), end:time.Time{wall:0xc1c9c8f10d7e21ae, ext:30011969707, loc:(*time.Location)(0xd37300)}, pendingCount:0x0, successCount:0x0, failureCount:0x0})
        	Test:       	TestCircuitBreaker_OpenState_Not_Enough_QPS
--- FAIL: TestCircuitBreaker_OpenState_Not_Enough_QPS (0.00s)
```

client/circuit_breaker/circuit_breaker.go
rleungx: I don't find where it is used in the current codebase.

client/inner_client.go
```diff
@@ -245,3 +251,12 @@ func (c *innerClient) dispatchTSORequestWithRetry(ctx context.Context) tso.TSFut
 	}
 	return req
 }
+
+func isOverloaded(err error) cb.Overloading {
```
rleungx, line 255: How about just
```suggestion
func isOverloaded(err error) bool {
```

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,290 @@
...
+// The implementation represents a state machine effectively
+func (s *State[T]) onRequest(cb *CircuitBreaker[T]) (*State[T], error) {
+	var now = time.Now()
+	switch s.stateType {
+	case StateClosed:
+		if s.end.Before(now) {
+			// ErrorRateWindow is over, let's evaluate the error rate
+			total := s.failureCount + s.successCount
+			observedErrorRatePct := s.failureCount * 100 / total
+			if s.cb.config.ErrorRateThresholdPct > 0 && total >= uint32(s.cb.config.ErrorRateWindow.Seconds())*s.cb.config.MinQPSForOpen && observedErrorRatePct >= s.cb.config.ErrorRateThresholdPct {
```
rleungx, line 229:
```suggestion
			// ErrorRateWindow is over, let's evaluate the error rate
			if s.cb.config.ErrorRateThresholdPct <= 0 {
			        return cb.newState(now, StateClosed), nil
			}
			total := s.failureCount + s.successCount
			observedErrorRatePct := s.failureCount * 100 / total
			if total >= uint32(s.cb.config.ErrorRateWindow.Seconds())*s.cb.config.MinQPSForOpen && observedErrorRatePct >= s.cb.config.ErrorRateThresholdPct {
```

client/client.go
```diff
@@ -755,7 +766,10 @@ func (c *client) GetPrevRegion(ctx context.Context, key []byte, opts ...opt.GetR
 	if serviceClient == nil {
 		return nil, errs.ErrClientGetProtoClient
 	}
-	resp, err := pdpb.NewPDClient(serviceClient.GetClientConn()).GetPrevRegion(cctx, req)
+	resp, err := c.inner.regionMetaCircuitBreaker.Execute(func() (*pdpb.GetRegionResponse, cb.Overloading, error) {
+		resp, err := pdpb.NewPDClient(serviceClient.GetClientConn()).GetPrevRegion(cctx, req)
+		return resp, isOverloaded(err), err
```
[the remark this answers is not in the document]
okJiang: Got it👍

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// onRequest transitions the state to the next state based on the current state and the previous requests results
+// All state transitions happens at the request evaluation time only
+// The implementation represents a state machine effectively
+func (s *State[T]) onRequest(cb *CircuitBreaker[T]) (*State[T], error) {
+	var now = time.Now()
+	switch s.stateType {
+	case StateClosed:
+		if s.end.Before(now) {
```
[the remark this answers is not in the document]
okJiang: > Each state is bound by time. When circuit breaker moves to a new state, it defines sets the deadline (end time) for this state.

Yes, but I mean during one state, we only need to check it once. For example:

- After `StateHalfOpen` switches to `StateClosed`. Before it switch to next state, we only need to check it once. Once the condition `s.end.Before(now)` is met, we do not need check it again, because after that time, each request will meet it?

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,280 @@
...
+				log.Error("Circuit breaker tripped. Starting to fail all requests",
+					zap.String("name", cb.name),
+					zap.Uint32("observedErrorRatePct", observedErrorRatePct),
+					zap.String("config", fmt.Sprintf("%+v", cb.config)))
+				cb.fastFailCounter.Inc()
+				return cb.newState(now, StateOpen), errs.ErrCircuitBreakerOpen
+			}
+			// the error threshold is not breached or there were not enough requests to evaluate it,
+			// continue in the closed state and allow all requests
+			return cb.newState(now, StateClosed), nil
```
okJiang, line 239: Do we need to return a new state🤔? It means if the fact error rate is not met `ErrorRateThresholdPct` during one `ErrorRateWindow`, we will move the window into next window. Isn't the range of movement too large? Can the window be moved in one second?

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,280 @@
...
+// All state transitions happens at the request evaluation time only
+// The implementation represents a state machine effectively
+func (s *State[T]) onRequest(cb *CircuitBreaker[T]) (*State[T], error) {
+	var now = time.Now()
+	switch s.stateType {
+	case StateClosed:
+		if s.end.Before(now) {
+			// ErrorRateWindow is over, let's evaluate the error rate
+			total := s.failureCount + s.successCount
+			observedErrorRatePct := s.failureCount * 100 / total
```
okJiang, line 228: Please check `total=0`

client/circuitbreaker/circuit_breaker.go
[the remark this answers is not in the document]
okJiang: Maybe the root problem is https://github.com/tikv/pd/pull/8856/files/a95b198e3f3ba8355def44b1210237d67c9e9a19#r1865433614 mentioned.

client/circuitbreaker/circuit_breaker_test.go
```diff
@@ -0,0 +1,205 @@
...
+		<-started[i]
+	}
+	assertFastFail(cb, re)
+	re.Equal(StateHalfOpen, cb.state.stateType)
+	for i := range ended {
+		waited[i] <- true
+		<-ended[i]
+	}
+	assertSucceeds(cb, re)
+	re.Equal(StateClosed, cb.state.stateType)
```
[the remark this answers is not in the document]
okJiang: Can you add a case to test two threads call `Execute` during one state(StateOpen), just like this case. I misunderstand them. I think we need them all.

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,290 @@
...
+	// even if the circuit breaker already moved to a new state while the request was in progress,
+	// it is still ok to update the old state, but it is not relevant anymore
+	state.onResult(overloaded)
+}
+
+func (cb *CircuitBreaker[T]) emitMetric(overloaded Overloading, err error) {
+	if err == nil {
+		cb.successCounter.Inc()
+	} else {
+		cb.failureCounter.Inc()
```
nolouch, line 176: It's a bit confusing because it has a different meaning in the State.

client/circuitbreaker/circuit_breaker.go
```diff
@@ -0,0 +1,290 @@
...
+	MinQPSForOpen uint32
+	// Defines how long to track errors before evaluating error_rate_threshold.
+	ErrorRateWindow time.Duration
+	// Defines how long to wait after circuit breaker is open before go to half-open state to send a probe request.
+	CoolDownInterval time.Duration
+	// Defines how many subsequent requests to test after cooldown period before fully close the circuit.
+	HalfOpenSuccessCount uint32
+}
+
+// AlwaysOpenSettings is a configuration that never trips the circuit breaker.
```
nolouch, line 55: never trips = AlwaysCloseSettings?

rleungx (approved):

client/circuitbreaker/circuit_breaker.go
nolouch: So should it be named `AlwaysCloseSettings`?

client/circuitbreaker/circuit_breaker.go
okJiang: > The time consists of windows back to back. Windows are static, they are not moving. When one window is over we evaluate the error rate, then we move to the next state with a new window.

OK, my previous idea was a movable window, but I think the current implementation can also meet the needs. LGTM

client/circuitbreaker/circuit_breaker.go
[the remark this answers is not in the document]
okJiang: This comment has been resolved by https://github.com/tikv/pd/pull/8856#discussion_r1875457345

client/circuitbreaker/circuit_breaker_test.go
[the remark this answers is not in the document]
okJiang: Just test count. Do not need to assert `State`.

okJiang (commented):
Except https://github.com/tikv/pd/pull/8856#discussion_r1875517116. Rest lgtm

okJiang (commented):
LGTM, but wait for @nolouch last review

nolouch (approved):

PR #8869, opened 2024-12-04
api: return not found when region doesn't exist

### LINKED ISSUE #8868: pd-ctl return error result when region doesn't exist
### What did you do?
query a region by id

Image

### What did you see instead?
Image

lhy1024:

Issue Number: Close #8868

```commit-message
```

### Check List

Tests

- Unit test
Before:
image
After:
image

### Release note

```release-note
None.
```


server/api/region.go
```diff
@@ -99,6 +103,10 @@ func (h *regionHandler) GetRegion(w http.ResponseWriter, r *http.Request) {
 	}
 
 	regionInfo := rc.GetRegionByKey(paramsByte[0])
+	if regionInfo == nil {
+		h.rd.JSON(w, http.StatusNotFound, errs.ErrRegionNotFound.FastGenByArgs().Error())
```
rleungx, line 107: 404 is expected?

lhy1024: Which status code do you think is more reasonable?

lhy1024:
tidb, tikv, tiup do not call `region/key` or `region/id`

server/api/region.go
rleungx: I was wondering if a 0 is valid request.

JmPotato (commented):
My concern here is that some of the downstream components may rely on this behavior to retry (when the result is empty), will this break the compatibility?

lhy1024:
> My concern here is that some of the downstream components may rely on this behavior to retry (when the result is empty), will this break the compatibility?

this api is not used by tidb, tikv, tiup, it is used by pd-ctl

server/api/region.go
lhy1024: why is 0 valid request?

rleungx: 0 is invalid, but here you return 404 instead of 400.

lhy1024: got it

rleungx (approved):

JmPotato (approved):

PR #8874, opened 2024-12-05
client: introduce an independent region client interface

JmPotato:

Issue Number: ref #8690.

```commit-message
Introduce an independent region client interface.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -16,9 +16,7 @@ package pd
 
 import (
 	"context"
-	"encoding/hex"
 	"fmt"
-	"net/url"
 	"runtime/trace"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -141,6 +96,7 @@ type RPCClient interface {
 	//     on your needs.
 	WithCallerComponent(callerComponent caller.Component) RPCClient
 
+	region.Client
```
rleungx, line 99: How about creating a client that contains the interfaces related to router functionality?

JmPotato: You mean using the term "router" rather than "region"?

rleungx (approved):

okJiang (approved):

Issue #4039, opened 2021-08-26
Why not support setting store_state to Tombstone?


qiaohao9:
### What did you do?

1. phsycally destroyed tikv data_dir
2. store state is set to `offline`
3. set store to `Tombstone` 

```shell
$ curl -X POST http://10.1.20.221:2379/pd/api/v1/store/5/state?state=Tombstone
```

### What did you expect to see?

the store is set to `Tombstone` successfully.

### What did you see instead?

```shell
[root@localhost ~]# curl -X POST http://10.1.20.221:2379/pd/api/v1/store/5/state?state=Tombstone
"invalid state Tombstone"
```

why only support `Up` and `Offline`, it's confused me in the AskTUG [手动删除tidb-data下面的pd和tikv目录，如何恢复](https://asktug.com/t/topic/68469/17)

https://github.com/tikv/pd/blob/20b00a7c4933e307c5423da0c5047250259e19e9/server/api/store.go#L223-L229

But pd can set state to tombstone actually

https://github.com/tikv/pd/blob/20b00a7c4933e307c5423da0c5047250259e19e9/server/cluster/cluster.go#L1040-L1075

### What version of PD are you using (`pd-server -V`)?

```text
[root@localhost ~]# tiup ctl:v5.1.1 pd -V
Starting component `ctl`: /root/.tiup/components/ctl/v5.1.1/ctl pd -V
Release Version: v5.1.1
Edition: Community
Git Commit Hash: 7cba1912b317a533e18b16ea2ba9a14ed2891129
Git Branch: heads/refs/tags/v5.1.1
UTC Build Time:  2021-07-23 01:16:35
```

nolouch:
There is an issue with it: https://github.com/tikv/pd/issues/3076
we support physically destroyed to replace it.

PR #8879, opened 2024-12-06
Add force option to "pd-ctl store delete" command

### LINKED ISSUE #8878: Add force option to pd-ctl store delete
## Enhancement Task
There is no way to specify `force` in pd-ctl.
https://download.pingcap.com/pd-api-doc.html#store__storeid__delete

This option is useful in case of node failure, etc.
https://ask.pingcap.com/t/removing-a-completely-offline-tikv-node-from-the-cluster/4318/7

shunki-fujita:

Issue Number: Close #8878

Add `--force` option to `pd-ctl store delete` command.
Set status to Tombstone directly.
https://download.pingcap.com/pd-api-doc.html#store__storeid__delete

```commit-message
$ ./bin/pd-ctl store delete --help
delete the store

Usage:
  pd-ctl store delete <store_id> [flags]
  pd-ctl store delete [command]

Available Commands:
  addr        delete store by its address

Flags:
  -f, --force   Set status to Tombstone directly
  -h, --help    help for delete

Global Flags:
      --cacert string   path of file that contains list of trusted SSL CAs
      --cert string     path of file that contains X509 certificate in PEM format
      --key string      path of file that contains X509 key in PEM format
  -u, --pd string       address of PD (default "http://127.0.0.1:2379")

Use "pd-ctl store delete [command] --help" for more information about a command.
$
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/store_command.go
+++ b/tools/pd-ctl/pdctl/command/store_command.go
@@ -60,20 +60,22 @@ func NewStoreCommand() *cobra.Command {
 // NewDeleteStoreByAddrCommand returns a subcommand of delete
 func NewDeleteStoreByAddrCommand() *cobra.Command {
 	d := &cobra.Command{
-		Use:   "addr <address>",
+		Use:   "addr <address> [flag]",
 		Short: "delete store by its address",
 		RunE:  deleteStoreCommandByAddrFunc,
 	}
+	d.Flags().BoolP("force", "f", false, "Set status to Tombstone directly")
 	return d
 }
 
 // NewDeleteStoreCommand return a delete subcommand of storeCmd
 func NewDeleteStoreCommand() *cobra.Command {
 	d := &cobra.Command{
-		Use:   "delete <store_id>",
+		Use:   "delete <store_id> [flag]",
 		Short: "delete the store",
 		Run:   deleteStoreCommandFunc,
 	}
+	d.Flags().BoolP("force", "f", false, "Set status to Tombstone directly")
 	d.AddCommand(NewDeleteStoreByAddrCommand())
 	return d
 }
@@ -319,6 +321,9 @@ func deleteStoreCommandFunc(cmd *cobra.Command, args []string) {
 		return
 	}
 	prefix := fmt.Sprintf(storePrefix, args[0])
+	if force, _ := cmd.Flags().GetBool("force"); force {
+		prefix += "?force=true"
+	}
 	_, err := doRequest(cmd, prefix, http.MethodDelete, http.Header{})
 	if err != nil {
 		cmd.Printf("Failed to delete store %s: %s\n", args[0], err)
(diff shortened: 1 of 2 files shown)


rleungx:
@shunki-fujita Thanks for your PR. But we have encountered many issues that misuse this API, so we prevent doing that anymore, see https://github.com/tikv/pd/issues/4039.

shunki-fujita:
@rleungx 
Thank you for your reply!

The `--force` option sets `physical_destroyed=true` rather than directly setting store state to `Tombstone`.
https://github.com/tikv/pd/pull/3388

Even so, is it better not to implement this option?

(By the way, since it is not directly set to Tombstone, the help message might not be accurate.)
https://download.pingcap.com/pd-api-doc.html#store__storeid__delete

rleungx:
> better

The reason why we use `physical_destroyed` is that we might need to add the store with the same address back in some scenarios, e.g., such as k8s env and we are sure that the previous store will never come back. `physical_destroyed` will let us no need to wait the damaged store finish offline before adding it's replacement. 
As for API docs, I believe that it's outdated. Let me fix it.

rleungx:
> Thank you for your reply!
The --force option sets physical_destroyed=true rather than directly setting store state to Tombstone.
https://github.com/tikv/pd/pull/3388
Even so, is it better not to implement this option?
(By the way, since it is not directly set to Tombstone, the help message might not be correct.)
https://download.pingcap.com/pd-api-doc.html#store__storeid__delete> 

The reason why we use `physical_destroyed` is that we might need to add the store with the same address back in some scenarios, e.g., such as k8s env and we are sure that the previous store will never come back. `physical_destroyed` will let us no need to wait the damaged store finish offline before adding it's replacement. As for API docs, I believe that it's outdated. Let me fix it.

shunki-fujita:
@rleungx 
> The reason why we use physical_destroyed is that we might need to add the store with the same address back in some scenarios, e.g., such as k8s env and we are sure that the previous store will never come back. physical_destroyed will let us no need to wait the damaged store finish offline before adding it's replacement. 

I encountered exactly that scenario, so I created this PR.
If implementing it in `pd-ctl` is considered risky, I will execute it via the API instead.

rleungx:
Both pd-ctl command and api are dangrous. If you use force option, it will set the store state to offline, and it won't block you to add a replacement node. 

Here is an example:
if the damaged store id is 5 which **never come back including the data**. When calling the API with force option equal to true, the store 5 will be offline state. At this moment, you can add a replacement node, assume that the new store id is 6. 
We might have two stores with the same address but different id in the cluster. After the store 5 finish offline, it will become tombstone state.

shunki-fujita:
@rleungx
Thank you for the advice.

> If the store 5 cannot be tombstone, it might have some other issues which we need to fix it instead of directly setting it to tombstone state.

I see,
I will conduct a verification separately.

PR #8885, opened 2024-12-09
cluster: fix tso fallback due raft cluster did not stop tso service (part2)

### LINKED ISSUE #8889: pd client panics with "[tso] timestamp fallback"
### What did you do?

run jepson test

no panic

### What did you see instead?

pd client panics with "[tso] timestamp fallback"

8.5

lhy1024:

Issue Number: Close #8889 ref #8477

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


server/cluster/cluster.go
```diff
@@ -329,6 +329,19 @@ func (c *RaftCluster) Start(s Server) error {
 		return err
 	}
 	c.checkTSOService()
+	defer func() {
+		// We need to try to stop tso jobs when the cluster is not running.
+		// Ref: https://github.com/tikv/pd/issues/8836
+		if !c.running {
+			c.stopTSOJobsIfNeeded()
```
rleungx, line 336: Shall we check the error instead of running?

server/cluster/cluster.go
```diff
@@ -329,6 +329,19 @@ func (c *RaftCluster) Start(s Server) error {
 		return err
 	}
 	c.checkTSOService()
+	defer func() {
+		// We need to try to stop tso jobs when the cluster is not running.
+		// Ref: https://github.com/tikv/pd/issues/8836
+		if !c.running {
```
lhy1024, line 335: I think it is better to check `runing` rather than error. Firstly,  we need to call `stopTSOJobsIfNeeded` whenever we return early, regardless of whether error is nil or not. Secondly, it prevents us from modifying the code later to return the error without assigning a value to it.

rleungx (approved):

okJiang (approved):

lhy1024:
/cherry-pick-release-8.5

PR #8902, opened 2024-12-11
*: unify the error definition

rleungx:

Issue Number: Ref #8922.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/errs/errno.go
+++ b/pkg/errs/errno.go
@@ -140,8 +140,69 @@ var (
 
 // checker errors
 var (
-	ErrCheckerNotFound   = errors.Normalize("checker not found", errors.RFCCodeText("PD:checker:ErrCheckerNotFound"))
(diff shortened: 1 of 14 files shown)


pkg/errs/errno.go
```diff
@@ -140,8 +140,69 @@ var (
...
+)
+
+// scatter errors
+var (
+	ErrEmptyRegion = errors.Normalize("empty region", errors.RFCCodeText("PD:scatter:ErrEmptyRegion"))
+)
+
+// keyspace errors
+var (
+	// ErrKeyspaceNotFound is used to indicate target keyspace does not exist.
```
okJiang, line 160: Are these comments neccesary?

rleungx: Just remain the same as before.

lhy1024 (approved):

okJiang (approved):

PR #8903, opened 2024-12-11
client/batch: allow tokenCh of batch controller to be nil

JmPotato:

Issue Number: ref #8690.

```commit-message
Allow `tokenCh` of batch controller be nil.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/pkg/batch/batch_controller.go
+++ b/client/pkg/batch/batch_controller.go
@@ -66,7 +66,7 @@ func (bc *Controller[T]) FetchPendingRequests(ctx context.Context, requestCh <-c
 		if errRet != nil {
 			// Something went wrong when collecting a batch of requests. Release the token and cancel collected requests
 			// if any.
-			if tokenAcquired {
+			if tokenAcquired && tokenCh != nil {
 				tokenCh <- struct{}{}
 			}
 			bc.FinishCollectedRequests(bc.finisher, errRet)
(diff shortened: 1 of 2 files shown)


client/pkg/batch/batch_controller.go
```diff
@@ -88,17 +91,19 @@ func (bc *Controller[T]) FetchPendingRequests(ctx context.Context, requestCh <-c
...
-			return ctx.Err()
-		case req := <-requestCh:
-			// Start to batch when the first request arrives.
-			bc.pushRequest(req)
-			// A request arrives but the token is not ready yet. Continue waiting, and also allowing collecting the next
-			// request if it arrives.
-			continue
-		case <-tokenCh:
-			tokenAcquired = true
+		if tokenCh != nil {
```
bufferflies, line 94: when the token channel is nil in product branch? I see only test maybe use nil.

JmPotato: I will use it in a later PR.

MyonKeminta (commented):
Rest LGTM

MyonKeminta (approved):

nolouch (approved):

client/pkg/batch/batch_controller.go
```diff
@@ -60,13 +60,18 @@ func NewController[T any](maxBatchSize int, finisher FinisherFunc[T], bestBatchO
 // It returns nil error if everything goes well, otherwise a non-nil error which means we should stop the service.
 // It's guaranteed that if this function failed after collecting some requests, then these requests will be cancelled
 // when the function returns, so the caller don't need to clear them manually.
+// `tokenCh` is an optional parameter:
```
rleungx, line 63: Shall we use something like
```
FetchPendingRequests(ctx context.Context, requestCh <-chan T, maxBatchWaitInterval time.Duration, tokenCh ...chan struct{}) (errRet error) {
```

JmPotato: It looks good to me, but I prefer keeping the current way to force the caller to decide whether `tokenCh` is needed by explicitly passing it in.

rleungx (approved):

PR #8912, opened 2024-12-12
scheduler: consider leader score when evict leader

### LINKED ISSUE #8895: Evicting slow stores may trigger redundant scheduling
## Enhancement Task

There are many redundant scheduling after evict-slow-store has been triaggered.

Image

Image

Image

lhy1024:

Issue Number: Close #8895

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/evict_leader.go
```diff
@@ -385,6 +378,22 @@ func scheduleEvictLeaderOnce(r *rand.Rand, name string, cluster sche.SchedulerCl
 	return ops
 }
 
+func createOperatorWithSort(name string, cluster sche.SchedulerCluster, candidates *filter.StoreCandidates, region *core.RegionInfo) (*operator.Operator, error) {
```
okJiang, line 381: How about renaming it to `CreateTransferLeaderOperatorToLowestScoreStore` and moving it to operator.go?

pkg/schedule/filter/comparer.go
```diff
@@ -40,6 +40,24 @@ func RegionScoreComparer(conf config.SharedConfigProvider) StoreComparer {
 	}
 }
 
+// LeaderScoreComparer creates a StoreComparer to sort store by leader
+// score.
+func LeaderScoreComparer(conf config.SchedulerConfigProvider) StoreComparer {
```
okJiang, line 45: Can you add an unit test for this function?

okJiang (approved):

pkg/schedule/filter/comparer.go
```diff
@@ -40,6 +41,25 @@ func RegionScoreComparer(conf config.SharedConfigProvider) StoreComparer {
 	}
 }
 
+// LeaderScoreComparer creates a StoreComparer to sort store by leader
+// score.
+func LeaderScoreComparer(conf config.SharedConfigProvider) StoreComparer {
+	leaderSchedulePolicy := conf.GetLeaderSchedulePolicy()
+	return func(a, b *core.StoreInfo) int {
+		// TODO: we should use the real time delta data to calculate the score.
+		sa := a.LeaderScore(leaderSchedulePolicy, 0)
```
bufferflies, line 50: The leader score is not very accurate, so it leads to the leader count of the lowest score goes up too much. How about considering the the running operators influence？

lhy1024: done

pkg/schedule/operator/create_operator.go
```diff
@@ -80,10 +80,9 @@ func CreateRemovePeerOperator(desc string, ci sche.SharedCluster, kind OpKind, r
 }
 
 // CreateTransferLeaderOperator creates an operator that transfers the leader from a source store to a target store.
-func CreateTransferLeaderOperator(desc string, ci sche.SharedCluster, region *core.RegionInfo, targetStoreID uint64, targetStoreIDs []uint64, kind OpKind) (*Operator, error) {
+func CreateTransferLeaderOperator(desc string, ci sche.SharedCluster, region *core.RegionInfo, targetStoreID uint64, kind OpKind) (*Operator, error) {
 	return NewBuilder(desc, ci, region, SkipOriginJointStateCheck).
 		SetLeader(targetStoreID).
-		SetLeaders(targetStoreIDs).
```
rleungx, line 86: Why remove it?

lhy1024: We do not need `targetStoreIDs`, which is used in `evict_leader.go` before.

rleungx: It's an optimization, why we don't need it?

lhy1024: `SetLeaders` checks learner and unhealthy  peer. https://github.com/tikv/pd/blob/master/pkg%2Fschedule%2Foperator%2Fbuilder.go#L301-L301

`SetLeader` also check these. https://github.com/tikv/pd/blob/master/pkg%2Fschedule%2Foperator%2Fbuilder.go#L284-L284

SetLeaders sort target stores according to store id. This PR sort target stores according to score.

rleungx: targetLeaderStoreIDs is used previously, but removed by this PR?

lhy1024: Yes, we use it in evict leader scheduler previously to select targets. It is replaced with this pr.

rleungx: We allow multiple targets in op step before, this PR changes it.

lhy1024: It only used in evict leader, after this pr there is no other scheduler using it. So I remove it. 
If we will use it in the future, I will recover it.

lhy1024: It was O(n) previously and it is O(nlogn) now. 
Considering that the number of sorts is usually two (three replicas) or four (five replicas), the difference is not too big.

rleungx: See https://github.com/tikv/tikv/issues/10602

lhy1024:
As @rleungx said, https://github.com/tikv/tikv/pull/11063 let tikv pick the fastest store to transfer leader to and may lead to leader unbalanced when some store is slow.

PR #8919, opened 2024-12-16
keypath: remove rootPath from storage and change all paths to absolute path

### LINKED ISSUE #8582: etcd key path is so difficult to read

okJiang:

Issue Number: Close #8582 

- Organize all keypaths and change them to absolute paths
- Remove rootpath from the underlying etcd interface of PD

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


server/gc_service.go
```diff
@@ -255,7 +251,7 @@ func (s *GrpcServer) loadRangeFromEtcd(startKey, endKey string) ([]string, []str
 	keys := make([]string, 0, len(resp.Kvs))
 	values := make([]string, 0, len(resp.Kvs))
 	for _, item := range resp.Kvs {
-		keys = append(keys, strings.TrimPrefix(strings.TrimPrefix(string(item.Key), s.rootPath), "/"))
+		keys = append(keys, string(item.Key), "/")
```
okJiang, line 254: Here is an error.

JmPotato (approved):

pkg/utils/keypath/region_test.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Sprintf("/pd/0/raft/r/%020d", id)
+	}
+	rand.New(rand.NewSource(time.Now().Unix()))
+	for range 1000 {
+		id := rand.Uint64()
+		re.Equal(f(id), RegionPath(id))
+	}
+}
+
+// func regionPath(regionID uint64) string {
```
rleungx, line 38: no need anymore?

pkg/tso/util_test.go
```diff
@@ -71,24 +71,3 @@ func TestExtractKeyspaceGroupIDFromKeyspaceGroupMembershipPath(t *testing.T) {
 		re.Error(err)
 	}
 }
-
-func TestExtractKeyspaceGroupIDFromKeyspaceGroupPrimaryPath(t *testing.T) {
```
rleungx, line 75: Why remove it?

pkg/storage/kv/kv_test.go
```diff
@@ -95,8 +93,8 @@ func testRange(re *require.Assertions, kv Base) {
 		limit      int
 		expect     []string
 	}{
-		{start: "", end: "z", limit: 100, expect: sortedKeys},
-		{start: "", end: "z", limit: 3, expect: sortedKeys[:3]},
+		{start: "a", end: "z", limit: 100, expect: sortedKeys},
```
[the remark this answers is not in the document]
okJiang: The key can not be empty now.

pkg/utils/keypath/region_test.go
okJiang: Yes, if this change is ok for reviewer, I can remove these codes.

pkg/storage/kv/kv_test.go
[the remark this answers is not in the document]
rleungx: Then we need to prevent it from being empty.

pkg/storage/kv/etcd_kv.go
```diff
@@ -70,11 +70,8 @@ func (kv *etcdKVBase) Load(key string) (string, error) {
 func (kv *etcdKVBase) LoadRange(key, endKey string, limit int) ([]string, []string, error) {
 	var OpOption []clientv3.OpOption
 	// If endKey is "\x00", it means to scan with prefix.
+	// If the prefix is empty, it means to scan all keys.
```
rleungx, line 73:
```suggestion
	// If the key is empty and endKey is "\x00", it means to scan all keys.
```

rleungx (approved):

Issue #8920, opened 2024-12-16
Add gRPC received command rate


okJiang:
## Enhancement Task

If the request is stuck during internal grpc server, we can't know the real request count, like [this](https://github.com/tikv/pd/issues/8882#issuecomment-2527032295). So we want to add this metrics.

PR #8921, opened 2024-12-16
grafana: Add `gRPC Received commands rate` panel

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8920

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code
image

### Release note

```release-note
None.
```


metrics/grafana/pd.json
```diff
@@ -8937,6 +8937,115 @@
...
+          "links": [],
+          "nullPointMode": "null as zero",
+          "paceLength": 10,
+          "pointradius": 5,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "targets": [
+            {
+              "expr": "sum(rate(grpc_server_started_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}[1m])) by (instance, grpc_method)",
```
rleungx, line 8981: How about also adding the error rate?

okJiang: Sure

okJiang: I added it. But btw, it can't catch the error hide in the header, like this https://github.com/tikv/pd/blob/2d970a619a8917c35d306f401326141481c133e0/server/grpc_service.go#L2071

metrics/grafana/pd.json
```diff
@@ -8937,6 +8937,226 @@
...
+          },
+          "paceLength": 10,
+          "pluginVersion": "7.5.17",
+          "pointradius": 5,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "targets": [
+            {
+              "expr": "sum(rate(grpc_server_handled_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", grpc_type=\"unary\", grpc_code!=\"OK\"}[1m])) by (grpc_method)",
```
bufferflies, line 9098: How about adding one expr as unary total about ignoring grpc_code?

okJiang: `grpc_server_handled_total` only records the count after handling. If we want to record the receive msg count, we can use `grpc_server_msg_received_total`, but it does not record `grpc_code` label. So we can't record the error rate of the stream, no grpc native metrics we can use.

metrics/grafana/pd.json
rleungx: Please add a comment or raise an issue for further improvement.

bufferflies (approved):

okJiang:
ping @rleungx @lhy1024

rleungx (approved):

Issue #8882, opened 2024-12-09
A large volume of HTTP requests combined with slow response time will significantly increase the memory usage of the PD


okJiang:
### What did you do?

In a 1 million region cluster, simultaneously start 600,000 ScanRegions requests to PD.

### What did you expect to see?

Memory is maintained within an acceptable range.

### What did you see instead?

Memory usage has increased by several times.

![Image](https://github.com/user-attachments/assets/c3fec2b9-2a11-4d69-9173-6084fde2687f)

### Extral

API Duration is 1s

![Image](https://github.com/user-attachments/assets/1793387b-98ac-4412-b325-c3b3ec2ddc2c)

In fact, PD only received over 200 requests.

okJiang:
The reason for the increase in memory is a bug in gRPC. https://github.com/grpc/grpc-go/issues/7261

It is unreasonable to make such large requests to such a massive cluster. This issue is temporarily noted for record-keeping. Perhaps significant improvements can be achieved after implementing https://github.com/tikv/pd/issues/8678.

okJiang:
Maybe we should set `maxConcurrentStreams` in grpc to limit the goroutines. Ref https://github.com/grpc/grpc-go/blob/66ba4b264d26808cb7af3c86eee66e843472915e/server.go#L1023-L1042.

Issue #4480, opened 2021-12-20
 Tracking issue for PD own service self-protection


CabinfeverB:
## Enhancement

This issue is tracking PD's own service self-protection mechanism.

The self-protection mechanism (refer to #4373) consists of two parts the standard use of etcd and the own service of PD itself. PD provides services for other components in two ways: HTTP API and gRPC. But now PD provides gRPC service by embedding etcd which is lack of a method to use gRPC interceptor, so we will support HTTP self-protection firstly.

We will develop two features to implement the self-protection mechanism. Such as k8s, hold a handler chain to process middleware. So PD will hold the rate limit handler and auditing handler. To gather some information used by the above handlers, it will also hold an information-gathering handler.

### Subtasks

- [x]  Implementing the information-gathering handler
  - [x] #4490 
  - [x] #4494 
  - [x] #4538
  - [x] #4681
- [x] Support HTTP QPS rate limit
  - [x] #4666
  - [x] #4500
- [x] Support HTTP audit 
  - [x] #4601
  - [x] #4602
  - [x] #4608
- [x] Config
  - [x] #4761
  - [x] #4869
- [ ] Let other components add `component` header
- [ ] Edit docs
### TODO
- [ ] support gRPC self-protection

lhy1024:
@CabinfeverB Can we close it?

CabinfeverB:
Closed since most tasks has been finished.

PR #8977, opened 2025-01-06
grpc: set MaxConcurrentStreams to avoid sudden traffic spikes that lead to PD OOM

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8882, ref https://github.com/tikv/pd/issues/4480

Added `MaxConcurrentStreams` to limit the request concurrency. This is a self-protection mechanism of PD. Once the number of concurrent requests reaches the limit, gRPC will wait for the ongoing requests to finish and allocate resources to the waiting requests. Therefore, in this situation, the request time may slow down, but it will provide better robustness.

PS: This parameter cannot be modified during runtime, so we cannot support changing it by pd-ctl. If you think it should be configurable, please comment.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -38,6 +38,7 @@ import (
 
 	"github.com/tikv/pd/pkg/errs"
 	rm "github.com/tikv/pd/pkg/mcs/resourcemanager/server"
+	"github.com/tikv/pd/pkg/memory"
 	sc "github.com/tikv/pd/pkg/schedule/config"
 	"github.com/tikv/pd/pkg/utils/configutil"
 	"github.com/tikv/pd/pkg/utils/grpcutil"
@@ -243,6 +244,9 @@ const (
 	defaultGCTunerThreshold           = 0.6
 	minGCTunerThreshold               = 0
 	maxGCTunerThreshold               = 0.9
+	// If concurrentStreams reaches 600k, the memory usage is about 40GB. To
+	// protect the server from OOM, we set the default concurrency to 10k per GB.
+	defaultConcurrencyPerGB = 10000
 
 	defaultWaitRegionSplitTimeout   = 30 * time.Second
 	defaultCheckRegionSplitInterval = 50 * time.Millisecond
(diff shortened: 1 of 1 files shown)


server/config/config.go
```diff
@@ -243,6 +244,9 @@ const (
 	defaultGCTunerThreshold           = 0.6
 	minGCTunerThreshold               = 0
 	maxGCTunerThreshold               = 0.9
+	// If concurrentStreams reaches 600k, the memory usage is about 40GB. To
```
rleungx, line 247: Is there any tests to support the conclusion?

okJiang: This conclusion comes from a practical case where the cluster contains millions of regions.

Let me elaborate. There are many factors that affect memory usage, primarily depending on the resource consumption during the request processing. For example, if the request is to retrieve simple metadata, its memory usage should be very low, and the response speed will be quite fast. For requests like `ScanRegion`, when the number of regions is small, the response speed is faster, and the backlog of requests will also decrease, resulting in relatively lower memory usage. However, if the number of regions is large, memory usage will increase, response speed will slow down, and the backlog will grow. At the same time, if there is a very high volume of traffic, the rise in concurrent backlogs can easily lead to an OOM situation.

What I want to say is, should we consider this sudden surge in traffic as an **anomaly**? If so, we can add this parameter to protect PD. Do you know in what normal circumstances PD would experience such high traffic? For example, 16GB PD and 160k requests.

rleungx: CMIIW, it only limits the concurrency for one connection, not total concurrency on the server side.

okJiang: In simple terms, should we consider this sudden surge in traffic as an anomaly? If so, I think we can set this parameter to protect PD. Do you know in what normal circumstances PD would experience such high traffic? For example, 16GB PD and 160k requests.

okJiang: > CMIIW, it only limits the concurrency for one connection, not total concurrency on the server side.

It serves all the request when both enabling gRPC and HTTP😂

okJiang:
No more update and comment, close it now.

server/config/config.go
lhy1024: Is it applicable to unary and stream?

okJiang: Yes. This is a gRPC/HTTP2 transport-level limit, so it applies to both unary RPCs and streaming RPCs.

- Unary: each in-flight unary call occupies one stream until the call finishes.
- Stream: each open client/server/bidi stream occupies one stream for its whole lifetime.
- Scope: it limits concurrent streams per client connection (server transport), not the total concurrency across all clients.

So it can help on both paths, but for streaming it only limits the number of open streams, not the number of messages within one stream. I think this thread is about clarifying the behavior, so no code change is needed here.

okJiang: Rechecked: yes, it applies to both unary RPCs and streaming RPCs, because this limit is enforced at the gRPC/HTTP2 stream layer per client connection.

- Unary RPC: one in-flight call uses one stream until it completes.
- Streaming RPC: one open client/server/bidi stream holds one stream for its lifetime.
- It does not limit the total concurrency across multiple client connections.

So the knob is effective for both unary and stream requests, but for streaming it limits the number of open streams rather than the number of messages sent on one stream. No code change is needed for this clarification.

PR #8979, opened 2025-01-08
server, core: implement the query region gRPC server

JmPotato:

Issue Number: ref #8690.

```commit-message
Implement the query region gRPC server interface.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/core/region.go
```diff
@@ -1464,6 +1464,97 @@ func (r *RegionsInfo) GetStoreRegions(storeID uint64) []*RegionInfo {
...
+// TODO: benchmark the performance of `QueryRegions`.
+// QueryRegions searches RegionInfo from regionTree by keys and IDs in batch.
+func (r *RegionsInfo) QueryRegions(
+	keys, prevKeys [][]byte, ids []uint64, needBuckets bool,
+) ([]uint64, []uint64, map[uint64]*pdpb.RegionResponse) {
+	// Iterate the region keys to find the regions.
+	regions := r.getRegionsByKeys(keys)
+	// Assert the returned regions count matches the input keys.
+	if len(regions) != len(keys) {
+		panic("returned regions count mismatch with the input keys")
```
rleungx, line 1476: Shall we return an error instead of panic?

pkg/core/region.go
```diff
@@ -1464,6 +1464,97 @@ func (r *RegionsInfo) GetStoreRegions(storeID uint64) []*RegionInfo {
 	return regions
 }
 
+// TODO: benchmark the performance of `QueryRegions`.
+// QueryRegions searches RegionInfo from regionTree by keys and IDs in batch.
+func (r *RegionsInfo) QueryRegions(
```
rleungx, line 1469: Will we pass both keys and prevKeys at the same time?

pkg/core/region.go
JmPotato: My idea is to set such panics at the initial stage of implementing this feature to help with testing. Once the feature needs to be released, I will change them to regular errors.

pkg/core/region.go
JmPotato: Yes, if the `GetRegion` and `GetPrevRegion` are batched in the same request.

server/grpc_service.go
```diff
@@ -1563,6 +1563,58 @@ func (s *GrpcServer) GetRegionByID(ctx context.Context, request *pdpb.GetRegionB
...
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return errors.WithStack(err)
+		}
+
+		// TODO: add forwarding logic.
+
+		if s.IsClosed() {
```
rleungx, line 1589: Do we need to check if cluster is started?

pkg/core/region.go
```diff
@@ -1464,6 +1464,97 @@ func (r *RegionsInfo) GetStoreRegions(storeID uint64) []*RegionInfo {
...
+				DownPeers:    region.GetDownPeers(),
+				PendingPeers: region.GetPendingPeers(),
+			}
+			if needBuckets {
+				regionResp.Buckets = region.GetBuckets()
+			}
+			regionsByID[id] = regionResp
+		}
+	}
+	return keyIDMap, prevKeyIDMap, regionsByID
```
rleungx, line 1510: What are those maps used for?

JmPotato: It helps the client map the keys passed in with their corresponding region IDs.

rleungx (approved):

pkg/core/region.go
```diff
@@ -1464,6 +1464,97 @@ func (r *RegionsInfo) GetStoreRegions(storeID uint64) []*RegionInfo {
...
+}
+
+// getRegionsByKeys searches RegionInfo from regionTree by keys.
+func (r *RegionsInfo) getRegionsByKeys(keys [][]byte) []*RegionInfo {
+	r.t.RLock()
+	defer r.t.RUnlock()
+	return r.tree.searchByKeys(keys)
```
nolouch, line 1517: I wonder if having too many keys will affect the lock's performance.

JmPotato: Batch parallel search is an optimization within the plan, and I want to implement it after completing an MVP.

nolouch: we need to prevent the behavior of holding the lock for too long.

nolouch (approved):
lgtm

rleungx:
/hold 
Wait for updating kvproto.

PR #8983, opened 2025-01-08
pd-ctl: print warn information when set max-replica to less than current

### LINKED ISSUE #8959: double confirm or print warn information when set max-replica to less than 3
## Enhancement Task
when we set the `max-replica` less than 3( mostly may be a mistake and carries high risk) with pd-ctl:

```
 tiup ctl:v7.5.2 pd config set max-replicas 2
Starting component ctl: /home/tidb/.tiup/components/ctl/v7.5.2/ctl pd config set max-replicas 2
Success!
```

As you may know, Setting max-replica to less than 3 may be a mistake and carries high risk, and I think we would better print some warning information( or let the customer do a second confirm) to alert the customer that this behavior carries high risk.

lhy1024:

Issue Number: Close #8959

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/config_command.go
+++ b/tools/pd-ctl/pdctl/command/config_command.go
@@ -372,9 +372,8 @@ func postConfigDataWithPath(cmd *cobra.Command, key, value, path string) error {
 	val, err := strconv.ParseFloat(value, 64)
 	if err != nil {
 		val = value
-	}
-	if key == "max-replicas" {
-		checkMaxReplicas(cmd, val)
+	} else if key == "max-replicas" {
+		checkMaxReplicas(cmd, val.(float64))
 	}
 	data[key] = val
 	reqData, err := json.Marshal(data)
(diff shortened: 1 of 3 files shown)


tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -374,6 +375,12 @@ func postConfigDataWithPath(cmd *cobra.Command, key, value, path string) error {
 		val = value
 	}
 	data[key] = val
+	if key == "max-replicas" {
+		replica, err := strconv.ParseInt(value, 10, 64)
+		if err == nil && replica < sc.DefaultMaxReplicas {
+			cmd.Println("Setting max-replica to less than 3 may be a mistake and carries high risk. Please confirm the setting.")
```
rleungx, line 381: I think the original issue wants an interactive way to confirm the operation instead of just printing a log.

okJiang (approved):

tools/pd-ctl/pdctl/command/config_command.go
lhy1024: After the discussion with @niubell , we will not add an interactive way in pd-ctl for now.

rleungx: How about not using 3?

tools/pd-ctl/pdctl/command/config_command.go
```diff
@@ -386,6 +389,30 @@ func postConfigDataWithPath(cmd *cobra.Command, key, value, path string) error {
 	return nil
 }
 
+func checkMaxReplicas(cmd *cobra.Command, value any) {
+	newReplica, ok := value.(float64)
```
rleungx, line 393: seem line 372 already checked it?

lhy1024: If err is not nil, val will be the type of string.

rleungx: Is that correct?

rleungx (approved):

PR #8984, opened 2025-01-09
*: no need to distinguish PD mode

rleungx:

Issue Number: ref #8477

This PR does the following changes:
1. use `IsKeyspaceGroupEnabled` to replace `IsPDServiceMode`, they are equivalent
2. merge pd mode and pd service mode
3. refactor tests to make them clearer
4. add a TODO to explain the current situation

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -131,7 +131,7 @@ type Server interface {
 	GetMembers() ([]*pdpb.Member, error)
 	ReplicateFileToMember(ctx context.Context, member *pdpb.Member, name string, data []byte) error
 	GetKeyspaceGroupManager() *keyspace.GroupManager
-	IsPDServiceMode() bool
+	IsKeyspaceGroupEnabled() bool
 	GetSafePointV2Manager() *gc.SafePointV2Manager
 }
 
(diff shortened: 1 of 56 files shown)


server/grpc_service.go
```diff
@@ -271,8 +271,8 @@ func (s *GrpcServer) GetClusterInfo(context.Context, *pdpb.GetClusterInfoRequest
 	}, nil
 }
 
-// GetMinTS implements gRPC PDServer. In PD mode, it simply returns a timestamp.
-// In PD service mode, it queries all tso servers and gets the minimum timestamp across
+// GetMinTS implements gRPC PDServer. In normal PD, it simply returns a timestamp.
```
lhy1024, line 274: normal PD?

server/cluster/cluster.go
```diff
@@ -404,7 +404,7 @@ func (c *RaftCluster) Start(s Server, bootstrap bool) (err error) {
 }
 
 func (c *RaftCluster) checkSchedulingService() {
-	if c.isPDServiceMode {
+	if c.isKeyspaceGroupEnabled {
```
lhy1024, line 407: If we only use pd service and scheduling service, is there keyspace group? Keyspace group similiar be about tso only.

rleungx: see https://github.com/tikv/pd/pull/8984/commits/2e8acc2a95e51cebc76dd9b59806132143ac8801

lhy1024 (approved):

rleungx:
@niubell @okJiang PTAL

PR #8994, opened 2025-01-14
client: fix pd client metrics registration

Tema:

Issue Number: ref https://github.com/tikv/pd/issues/8678

@JmPotato if there is a code referencing metrics from the global context like in https://github.com/tikv/client-go/pull/1543#discussion_r1906267748 then they could consume metrics created in the `metrics.init()` method. However, when latter the global metrics are recreated, reinitialized and registered with `metrics.InitAndRegisterMetrics()` the metrics consumed before are abandoned and not reported to prometheus. This PR allows to register metrics consumers, so that if they are consumed before `metrics.InitAndRegisterMetrics()` in invoked, they are reinitilized and registered later.

cc: @rleungx 

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/metrics/metrics.go
```diff
@@ -26,6 +27,26 @@ var initialized int32
 func init() {
 	initMetrics(prometheus.Labels{})
 	initCmdDurations()
+	initRegisteredConsumers()
+}
+
+var mutex sync.Mutex
+var consumersInitializers []func()
```
JmPotato, line 34: What about using an `atomic.Value` to store `[]func()`?

JmPotato: Your consideration is thorough. My initial intention was to merge the lock on line 33 with line 34 into one structure, making it look more elegant.

JmPotato: I still prefer using a single structure, you could use a customed structure rather than `atomic.Value`.

JmPotato (approved):
Overall LGTM. My final question is whether a `CircuitBreaker` has a lifecycle? If we have a scenario where the old one is deprecated and a new `CircuitBreaker` is created, do we need to remove the corresponding consumer?

rleungx (approved):

PR #8996, opened 2025-01-14
*: rename file and clean up

rleungx:

Issue Number: ref #5766

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


client/inner_client.go
```diff
@@ -71,15 +71,8 @@ func (c *innerClient) setServiceMode(newMode pdpb.ServiceMode) {
 	if newMode == c.serviceMode {
 		return
 	}
-	log.Info("[pd] changing service mode",
-		zap.String("old-mode", c.serviceMode.String()),
-		zap.String("new-mode", newMode.String()))
 	c.resetTSOClientLocked(newMode)
-	oldMode := c.serviceMode
 	c.serviceMode = newMode
-	log.Info("[pd] service mode changed",
```
lhy1024, line 80: Do we need to keep one of the two logs?

pkg/utils/keypath/key_path.go
```diff
@@ -321,7 +321,7 @@ func svcRootPath(svcName string) string {
 	return path.Join(constant.MicroserviceRootPath, c, svcName)
 }
 
-// LegacyRootPath returns the root path of legacy pd service.
+// LegacyRootPath returns the root path of legacy PD.
```
lhy1024, line 324: Do we need to replace legacy with non-microservice env?

lhy1024 (commented):
The rest LGTM

pkg/utils/keypath/key_path.go
rleungx: LegacyRootPath uses "legacy", I just keep using it.

client/inner_client.go
lhy1024:
```
if newMode != pdpb.ServiceMode_PD_SVC_MODE {
log.Info("[pd] changing from non-microservice to microservice")
}else{
log.Info("[pd] changing from microservice to non-microservice")
}
```

lhy1024: how about it?

rleungx: Is it similar to line 87 and line 91?

lhy1024: resetTSOClientLocked and setServiceMode may be called at different places?

rleungx: done

lhy1024 (approved):

JmPotato (approved):

niubell (approved):

PR #9005, opened 2025-01-17
schedule: impl balance range scheduler

### LINKED ISSUE #9006: scheduler: implement balance range scheduler algorithm

bufferflies:

Issue Number: Close #9006

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```


pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+
+func (s *balanceRangeScheduler) prepare(cluster sche.SchedulerCluster, opInfluence operator.OpInfluence, job *balanceRangeSchedulerJob) (*balanceRangeSchedulerPlan, error) {
+	krs := core.NewKeyRanges(job.Ranges)
+	scanRegions, err := cluster.BatchScanRegions(krs)
+	if err != nil {
+		return nil, err
+	}
+	filters := s.filters
+	if job.Engine == tiflash {
+		filters = append(filters, filter.NewEngineFilter(balanceRangeName, filter.SpecialEngines))
```
rleungx, line 388: use types.BalanceRangeScheduler?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+}
+
+type storeInfo struct {
+	store *core.StoreInfo
+	score int64
+}
+
+func (s *balanceRangeScheduler) prepare(cluster sche.SchedulerCluster, opInfluence operator.OpInfluence, job *balanceRangeSchedulerJob) (*balanceRangeSchedulerPlan, error) {
+	krs := core.NewKeyRanges(job.Ranges)
+	scanRegions, err := cluster.BatchScanRegions(krs)
```
rleungx, line 382: Do we need to separate krs into multiple batches if it is a big range?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -70,13 +77,59 @@ type balanceRangeSchedulerConfig struct {
...
+func (conf *balanceRangeSchedulerConfig) begin(job *balanceRangeSchedulerJob) bool {
+	conf.Lock()
+	defer conf.Unlock()
+	if job.Status != pending {
+		return false
+	}
+	now := time.Now()
+	job.Start = &now
+	job.Status = running
+	if err := conf.save(); err != nil {
```
rleungx, line 99: If persist fails, do we still regard it as a running job?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+func (s *balanceRangeScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
+	balanceRangeCounter.Inc()
+	job := s.conf.peek()
+	if job == nil {
+		balanceRangeNoJobCounter.Inc()
+		return nil, nil
+	}
+
+	opInfluence := s.OpController.GetOpInfluence(cluster.GetBasicCluster(), operator.WithRangeOption(job.Ranges))
+	plan, err := s.prepare(cluster, opInfluence, job)
```
rleungx, line 240: For every time we schedule, it will scan the range from the beginning?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -70,13 +77,59 @@ type balanceRangeSchedulerConfig struct {
...
+		return false
+	}
+	return true
+}
+
+func (conf *balanceRangeSchedulerConfig) peek() *balanceRangeSchedulerJob {
+	conf.RLock()
+	defer conf.RLock()
+	for _, job := range conf.jobs {
+		if job.Status == finished {
```
rleungx, line 125: Where do we change the status to finished?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+	targetInfluence := p.opInfluence.GetStoreInfluence(p.targetStoreID())
+	targetInf := p.job.Role.getStoreInfluence(targetInfluence)
+	if targetInf < 0 {
+		targetInf = -targetInf
+	}
+	targetScore := p.targetScore + targetInf + p.tolerate
+
+	shouldBalance := sourceScore >= targetScore
+	if !shouldBalance && log.GetLevel() <= zap.DebugLevel {
+		log.Debug("skip balance ",
```
rleungx, line 479:
```suggestion
		log.Debug("skip balance",
```

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+	}
+	return shouldBalance
+}
+
+// Role is the role of the region.
+type Role int
+
+const (
+	leader Role = iota
+	// include leader + voter
```
rleungx, line 500: What does it mean?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+	default:
+		return "unknown"
+	}
+}
+
+// engine is the engine of the store.
+type engine int
+
+const (
+	tiKV engine = iota
```
rleungx, line 523: We have EngineTiFlash/EngineTiKV

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+	fit          *placement.RegionFit
+	averageScore int64
+	job          *balanceRangeSchedulerJob
+	opInfluence  operator.OpInfluence
+	tolerate     int64
+}
+
+type storeInfo struct {
+	store *core.StoreInfo
+	score int64
```
rleungx, line 377: How about using scoreMap to replace it?

pkg/schedule/schedulers/balance_range.go
bufferflies: fixed, it needs to roll back.

pkg/schedule/schedulers/balance_range.go
bufferflies: good catch

pkg/schedule/schedulers/balance_range.go
bufferflies: yes, the info on the regions and stores in the given ranges may be updated.

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		}
+	}
+	return nil, nil
+}
+
+// transferPeer selects the best store to create a new peer to replace the old peer.
+func (s *balanceRangeScheduler) transferPeer(plan *balanceRangeSchedulerPlan, dstStores []*core.StoreInfo) *operator.Operator {
+	excludeTargets := plan.region.GetStoreIDs()
+	if plan.job.Role == leader {
+		excludeTargets = make(map[uint64]struct{})
```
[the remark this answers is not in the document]
bufferflies: good idea

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: fixed

pkg/schedule/operator/operator_controller.go
```diff
@@ -828,14 +829,33 @@ func (oc *Controller) GetHistory(start time.Time) []OpHistory {
 	return history
 }
 
+// OpInfluenceOption is used to filter the region.
+// returns true if the region meets the condition, it will ignore this region in the influence calculation.
+// returns false if the region does not meet the condition, it will calculate the influence of this region.
+type OpInfluenceOption func(region *core.RegionInfo) bool
+
+// WithRangeOption returns an OpInfluenceOption that filters the region by the label.
```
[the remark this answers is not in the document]
bufferflies: rename by the key ranges

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +227,332 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		log.Debug("select region", zap.String("scheduler", s.GetName()), zap.Uint64("region-id", plan.region.GetID()))
+		// Skip hot regions.
+		if cluster.IsRegionHot(plan.region) {
+			log.Debug("region is hot", zap.String("scheduler", s.GetName()), zap.Uint64("region-id", plan.region.GetID()))
+			balanceRangeHotCounter.Inc()
+			continue
+		}
+		// Check region leader
+		if plan.region.GetLeader() == nil {
+			log.Warn("region have no leader", zap.String("scheduler", s.GetName()), zap.Uint64("region-id", plan.region.GetID()))
```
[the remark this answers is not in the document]
bufferflies: fixed

pkg/schedule/schedulers/balance_range.go
rleungx: Will it affect the other requests like get region or heartbeat if the range is large?

pkg/schedule/schedulers/balance_range.go
rleungx: If all ranges is balanced, we still need to wait for timeout?

pkg/schedule/schedulers/balance_range.go
bufferflies: removed

pkg/schedule/schedulers/balance_range.go
bufferflies: yes in this pr. I will add other conditions to check whether the distribution is balanced in next PR.

pkg/schedule/schedulers/balance_range.go
bufferflies: yes, the operator of scanning regions acquires Rlock which shares with the region heartbeat.  Maybe we can reuse the previous distribution by interval duration, not every scheduler.

pkg/schedule/schedulers/balance_range.go
bufferflies: fixed

pkg/schedule/schedulers/balance_range.go
bufferflies: It doesn't need to separate the ranges by many times because the duration of the lock is the same

rleungx: If we acquire the lock multiple times and for every time we hold the lock for a short time, the other places can also get the lock. But If we hold the lock for a long time, it may cause other places starvation.

bufferflies: ok, I will split the ranges.

pkg/schedule/schedulers/balance_range.go
```diff
@@ -74,9 +79,60 @@ type balanceRangeSchedulerJob struct {
...
+func (conf *balanceRangeSchedulerConfig) begin(job *balanceRangeSchedulerJob) bool {
+	conf.Lock()
+	defer conf.Unlock()
+	if job.Status != pending {
+		return false
+	}
+	now := time.Now()
+	job.Start = &now
+	job.Status = running
+	if err := conf.save(); err != nil {
```
okJiang, line 97: It is not recommended that
1. modify the input param
2. call conf.save() to persist the modified input struct

I prefer more explicit implement. For example, we can use `jobs map[JobID]balanceRangeSchedulerJob` and input `JobID` here rather than Job.

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +230,286 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		role := job.Role
+		iop := role.getStoreInfluence(opInfluence.GetStoreInfluence(sources[i].GetID()))
+		jop := role.getStoreInfluence(opInfluence.GetStoreInfluence(sources[j].GetID()))
+		iScore := scoreMap[sources[i].GetID()]
+		jScore := scoreMap[sources[j].GetID()]
+		return iScore+iop > jScore+jop
+	})
+
+	averageScore := int64(0)
+	if len(sources) != 0 {
```
okJiang, line 423: This has been checked in L390. Maybe we can remove this.

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +230,286 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+	sourceInfluence := p.opInfluence.GetStoreInfluence(p.sourceStoreID())
+	sourceInf := p.job.Role.getStoreInfluence(sourceInfluence)
+	if sourceInf < 0 {
+		sourceInf = -sourceInf
+	}
+	sourceScore := p.sourceScore - sourceInf - p.tolerate
+
+	targetInfluence := p.opInfluence.GetStoreInfluence(p.targetStoreID())
+	targetInf := p.job.Role.getStoreInfluence(targetInfluence)
+	if targetInf < 0 {
```
okJiang, line 467: targetInf is int64?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +230,286 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		targetInf = -targetInf
+	}
+	targetScore := p.targetScore + targetInf + p.tolerate
+
+	shouldBalance := sourceScore >= targetScore
```
okJiang, line 472: Can you add some comments to explan the principle? Why is the score calculated this way? What is the score difference from the balanceRegionScheduler? How are conflicts between different balance schedulers handled?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +230,286 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
 	return s
 }
 
+// Schedule schedules the balance key range operator.
+func (s *balanceRangeScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
+	balanceRangeCounter.Inc()
+	job := s.conf.peek()
+	if job == nil {
+		balanceRangeNoJobCounter.Inc()
```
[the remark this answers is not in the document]
bufferflies: It will count if all the jobs are finished.

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: yes, if the store has some remove-peer operators

pkg/schedule/schedulers/balance_range.go
bufferflies: The `balanceRegionScheduler` and  `balanceLeaderScheduler` only consider the global regions not the given key ranges, so the store status can calculate their score. But the `balance-range-scheduler` is different, the given key ranges should set it

pkg/schedule/schedulers/balance_range.go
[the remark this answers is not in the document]
okJiang: How about using `RunningJobCounter` rather than `NoJobCounter`?

pkg/schedule/schedulers/balance_range.go
okJiang: So we need not check `targetInf < 0`?

pkg/schedule/schedulers/balance_range.go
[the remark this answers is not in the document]
bufferflies: if `conf.peek()` is nil, it indicates that all the jobs in the configuration are finished, so the metrics just tell DBA that there is no unable job in this scheduler.

pkg/schedule/schedulers/balance_range.go
bufferflies: I must consider the scenario involving `removing-peer `, so the `targeting` is negative, the target score is small. This causes too many peers to be added to this store.

okJiang (approved):

pkg/schedule/schedulers/balance_range.go
```diff
@@ -74,9 +80,60 @@ type balanceRangeSchedulerJob struct {
...
+	Start   *time.Time      `json:"start,omitempty"`
+	Finish  *time.Time      `json:"finish,omitempty"`
+	Create  time.Time       `json:"create"`
 	Status  JobStatus       `json:"status"`
 }
 
+func (conf *balanceRangeSchedulerConfig) begin(index int) *balanceRangeSchedulerJob {
+	conf.Lock()
+	defer conf.Unlock()
+	job := conf.jobs[index]
```
rleungx, line 92: Seems we have already had the job, why do we need to get it again?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -74,9 +80,60 @@ type balanceRangeSchedulerJob struct {
...
+		job.Status = pending
+		job.Start = nil
+	}
+	return job
+}
+
+func (conf *balanceRangeSchedulerConfig) finish(index int) *balanceRangeSchedulerJob {
+	conf.Lock()
+	defer conf.Unlock()
+	job := conf.jobs[index]
```
rleungx, line 110: ditto

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +232,314 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		jop := role.getStoreInfluence(opInfluence.GetStoreInfluence(sources[j].GetID()))
+		iScore := scoreMap[sources[i].GetID()]
+		jScore := scoreMap[sources[j].GetID()]
+		return iScore+iop > jScore+jop
+	})
+
+	averageScore := int64(0)
+	averageScore = totalScore / int64(len(sources))
+
+	tolerate := int64(float64(len(scanRegions)) * adjustRatio)
```
rleungx, line 452:
```suggestion
	tolerantSizeRatio := int64(float64(len(scanRegions)) * adjustRatio)
```

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +232,314 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+			zap.Int64("influence-target-score", targetScore),
+			zap.Int64("average-region-size", p.averageScore),
+			zap.Int64("tolerate", p.tolerate),
+		)
+	}
+	return shouldBalance
+}
+
+// Role is the role of the region.
+type Role int
```
rleungx, line 521: put definitions and functions together?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -218,16 +580,35 @@ func NewRole(role string) Role {
...
-		return "unknown"
+		return nil
+	}
+}
+
+func (r Role) getStoreInfluence(influence *operator.StoreInfluence) int64 {
+	switch r {
+	case leader:
+		return influence.LeaderCount
+	case follower:
```
rleungx, line 606: Will it conflict with balance region?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -218,16 +580,35 @@ func NewRole(role string) Role {
 	}
 }
 
-func (r Role) String() string {
+func (r Role) getPeers(region *core.RegionInfo) []*metapb.Peer {
```
rleungx, line 583: I think it should be a method of region instead of role.

pkg/schedule/schedulers/balance_range.go
```diff
@@ -218,16 +580,35 @@ func NewRole(role string) Role {
...
-		return "learner"
+		learners := region.GetLearners()
+		return learners
 	default:
-		return "unknown"
+		return nil
+	}
+}
+
+func (r Role) getStoreInfluence(influence *operator.StoreInfluence) int64 {
```
rleungx, line 602: ditto

rleungx: should be influence' method

pkg/schedule/schedulers/balance_range.go
```diff
@@ -167,6 +232,314 @@ func newBalanceRangeScheduler(opController *operator.Controller, conf *balanceRa
...
+		log.Debug("skip balance",
+			zap.String("scheduler", scheduler),
+			zap.Uint64("region-id", p.region.GetID()),
+			zap.Uint64("source-store", p.sourceStoreID()),
+			zap.Uint64("target-store", p.targetStoreID()),
+			zap.Int64("origin-source-score", p.sourceScore),
+			zap.Int64("origin-target-score", p.targetScore),
+			zap.Int64("influence-source-score", sourceScore),
+			zap.Int64("influence-target-score", targetScore),
+			zap.Int64("average-region-size", p.averageScore),
```
rleungx, line 513: region size?

pkg/schedule/schedulers/balance_range.go
bufferflies: @rleungx   the functions must ensure that jobs list must contains the given job.

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: yes，the next MR will solve it

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
bufferflies: done

pkg/schedule/schedulers/balance_range.go
```diff
@@ -463,7 +463,7 @@ func (s *balanceRangeScheduler) prepare(cluster sche.SchedulerCluster, opInfluen
 		averageScore:     averageScore,
 		job:              job,
 		opInfluence:      opInfluence,
-		tolerate:         tolerate,
+		tolerate:         tolerantSizeRatio,
```
rleungx, line 466: tolerantSizeRatio

rleungx (approved):
Overall LGTM

PR #8939, opened 2024-12-19
client/router: implement the query region gRPC client

JmPotato:
### What problem does this PR solve?

Issue Number: ref #8690.

### What is changed and how does it work?

```commit-message
Implement the router stream update logic.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

PR #9014, opened 2025-01-20
client/sd: unify the service discovery callbacks within a struct

JmPotato:

Issue Number: ref #8690.

```commit-message
Unify the service discovery callbacks within a struct.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/servicediscovery/callbacks.go
+++ b/client/servicediscovery/callbacks.go
@@ -1,0 +1,104 @@
+// Copyright 2025 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package servicediscovery
+
+import (
+	"sync"
+
+	"github.com/pingcap/kvproto/pkg/pdpb"
+)
+
+type leaderSwitchedCallbackFunc func(string) error
+
+// serviceCallbacks contains all the callback functions for service discovery events
+type serviceCallbacks struct {
+	sync.RWMutex
+	// serviceModeUpdateCb will be called when the service mode gets updated
+	serviceModeUpdateCb func(pdpb.ServiceMode)
+	// leaderSwitchedCbs will be called after the leader switched
+	leaderSwitchedCbs []leaderSwitchedCallbackFunc
+	// membersChangedCbs will be called after there is any membership change in the
+	// leader and followers
+	membersChangedCbs []func()
+}
+
+func newServiceCallbacks() *serviceCallbacks {
+	return &serviceCallbacks{
+		leaderSwitchedCbs: make([]leaderSwitchedCallbackFunc, 0),
+		membersChangedCbs: make([]func(), 0),
+	}
+}
+
+func (c *serviceCallbacks) setServiceModeUpdateCallback(cb func(pdpb.ServiceMode)) {
+	c.Lock()
+	defer c.Unlock()
+	c.serviceModeUpdateCb = cb
+}
+
+func (c *serviceCallbacks) addServingURLSwitchedCallback(cb leaderSwitchedCallbackFunc) {
+	c.Lock()
+	defer c.Unlock()
+	c.leaderSwitchedCbs = append(c.leaderSwitchedCbs, cb)
+}
+
+func (c *serviceCallbacks) addServiceURLsSwitchedCallback(cb func()) {
+	c.Lock()
+	defer c.Unlock()
+	c.membersChangedCbs = append(c.membersChangedCbs, cb)
+}
+
+func (c *serviceCallbacks) callServiceModeUpdateCallback(mode pdpb.ServiceMode) {
+	c.RLock()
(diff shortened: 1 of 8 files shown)


client/servicediscovery/service_discovery.go
```diff
@@ -980,23 +961,18 @@ func (c *serviceDiscovery) switchLeader(url string) (bool, error) {
 		return false, nil
 	}
 
-	newConn, err := c.GetOrCreateGRPCConn(url)
+	newConn, _ := c.GetOrCreateGRPCConn(url)
```
rleungx, line 964: Why omit the error?

client/servicediscovery/service_discovery.go
```diff
@@ -570,7 +555,7 @@ func (c *serviceDiscovery) updateServiceModeLoop() {
 		failpoint.Return()
 	})
 	failpoint.Inject("usePDServiceMode", func() {
-		c.serviceModeUpdateCb(pdpb.ServiceMode_PD_SVC_MODE)
+		c.callbacks.callServiceModeUpdateCallback(pdpb.ServiceMode_PD_SVC_MODE)
```
rleungx, line 558: Is it possible to make the name shorter?

rleungx (approved):

okJiang:
> @JmPotato: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-integration-realcluster-test	[65ed86d](https://github.com/tikv/pd/commit/65ed86dc473747e669fbaab547c33afa6153e8f0)	[link](https://do.pingcap.net/jenkins/job/tikv/job/pd/job/pull_integration_realcluster_test/1791/display/redirect)	true	`/test pull-integration-realcluster-test`
> [Full PR test history](https://prow.tidb.net/pr-history?org=tikv&repo=pd&pr=9014). [Your PR dashboard](https://prow.tidb.net/pr?query=is%3Apr%20state%3Aopen%20author%3AJmPotato).

https://github.com/tikv/pd/blob/093b84112e19a2ad54ca3282726147bc7198b8f2/tests/integrations/realcluster/cluster.go#L141-L152

In the test report, it download the binaries when deploy ms cluster, which is unexpected🤔, maybe it is a tiup bug.

client/servicediscovery/callbacks.go
```diff
@@ -0,0 +1,104 @@
...
+	c.serviceModeUpdateCb = cb
+}
+
+func (c *serviceCallbacks) addServingURLSwitchedCallback(cb leaderSwitchedCallbackFunc) {
+	c.Lock()
+	defer c.Unlock()
+	c.leaderSwitchedCbs = append(c.leaderSwitchedCbs, cb)
+}
+
+func (c *serviceCallbacks) addServiceURLsSwitchedCallback(cb func()) {
```
okJiang, line 56: How about this?
```suggestion
func (c *serviceCallbacks) addMembersChangedCallback(cb func()) {
```

client/servicediscovery/callbacks.go
```diff
@@ -0,0 +1,104 @@
...
+	"sync"
+
+	"github.com/pingcap/kvproto/pkg/pdpb"
+)
+
+type leaderSwitchedCallbackFunc func(string) error
+
+// serviceCallbacks contains all the callback functions for service discovery events
+type serviceCallbacks struct {
+	sync.RWMutex
```
okJiang, line 27: Is this lock necessary?

client/servicediscovery/service_discovery.go
```diff
@@ -126,14 +126,16 @@ type ServiceDiscovery interface {
 	// CheckMemberChanged immediately check if there is any membership change among the leader/followers
 	// in a quorum-based cluster or among the primary/secondaries in a primary/secondary configured cluster.
 	CheckMemberChanged() error
+	// ExecuteAndAddServingURLSwitchedCallback executes the callback once and adds it to the callback list then.
+	ExecuteAndAddServingURLSwitchedCallback(cb leaderSwitchedCallbackFunc)
```
okJiang, line 130: Why we need a `ExecuteAndAddServingURLSwitchedCallback` interface?

If we need execute some functions, I think we can execute them directly.

client/servicediscovery/callbacks.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2025 TiKV Project Authors.
```
okJiang, line 1: How about moving this file to pkg?

client/servicediscovery/callbacks.go
JmPotato: This lock will be used in the next PR.

client/servicediscovery/service_discovery.go
JmPotato: `serviceDiscovery` is exposed as an interface, the caller cannot access those internal callbacks.

client/servicediscovery/callbacks.go
JmPotato: Its code is highly related with the service discovery, `pkg` is for those components which are more common.

client/servicediscovery/service_discovery.go
okJiang: Because I find it is called in only one place. So maybe `Execute` is an unnecessary abstraction.

JmPotato: It will be used in the next PR also, this method is useful if we want to integrate a sub-client into the client, an immediate execution of the callback with the lastest leader URL is necessary, e.g., https://github.com/tikv/pd/pull/8939/files#diff-c2cd9790c2f82b378808c8f7326f85b46ab36aa388c95250865571df80a4679bR193.

okJiang (approved):

PR #9015, opened 2025-01-20
*: check file header

rleungx:

Issue Number: ref #4322

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/btree/btree_generic_test.go
+++ b/pkg/btree/btree_generic_test.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2022 Google Inc.
+// Copyright 2022 TiKV Project Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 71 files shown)


pkg/btree/btree_generic_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014-2022 Google Inc.
+// Copyright 2022 TiKV Project Authors.
```
lhy1024, line 1: Will we modify it?

rleungx: just reorder to make detection easier

pkg/utils/grpcutil/grpcutil_test.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2019 TiKV Project Authors.
```
bufferflies, line 1: Is it better to use 2025 for the new license if the previous header is blank?

bufferflies (commented):
rest lgmt

pkg/utils/grpcutil/grpcutil_test.go
bufferflies: By the way, I found the new  issued years of different files are different.

rleungx: I use the file created time, just the same as others.

JmPotato (approved):

lhy1024 (approved):

PR #9048, opened 2025-02-08
cache: lazy gc in ttl

### LINKED ISSUE #9047: goroutine leak: goroutine increase during the long-term operation
## Bug Report
The count of pd goroutine continue increase in long-term cluster

### What did you do?
Trigger condition: remove scheduler 

 ```
#/bin/bash

for i in $(seq 1 50); do
tiup ctl:v6.5.6 pd -u  http://127.0.0.1:2379 scheduler remove balance-leader-scheduler
sleep 10
tiup ctl:v6.5.6 pd -u  http://127.0.0.1:2379 scheduler add balance-leader-scheduler
sleep 20
done
```

Image

root cause: https://github.com/tikv/pd/blob/77d6f5b6785eb4e153495bf1c32b165019ae119e/server/cluster/coordinator.go#L696-L700

The count of goroutine is steady

v6.5.6

### PR DESCRIPTION
**Pull request by bufferflies**

Issue Number: Close #9047  

```commit-message
```

### Check List

Tests

- Unit test

6.5.6 fixed 
image

Code changes

- Has persistent data change

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/cache/ttl.go
+++ b/pkg/cache/ttl.go
@@ -49,8 +49,6 @@ func newTTL(ctx context.Context, gcInterval time.Duration, duration time.Duratio
 		ttl:        duration,
 		gcInterval: gcInterval,
 	}
-
-	go c.doGC()
 	return c
 }
 
@@ -63,7 +61,9 @@ func (c *ttlCache) put(key any, value any) {
 func (c *ttlCache) putWithTTL(key any, value any, ttl time.Duration) {
 	c.Lock()
 	defer c.Unlock()
-
+	if len(c.items) == 0 {
+		go c.doGC()
+	}
 	c.items[key] = ttlCacheItem{
 		value:  value,
 		expire: time.Now().Add(ttl),
@@ -162,6 +162,11 @@ func (c *ttlCache) doGC() {
 						delete(c.items, key)
 					}
 				}
+			}
+			if len(c.items) == 0 {
+				c.Unlock()
+				log.Debug("TTL GC items is empty exit")
+				return
 			}
 			c.Unlock()
 			log.Debug("TTL GC items", zap.Int("count", count))
--- a/pkg/cache/ttl.go
+++ b/pkg/cache/ttl.go
@@ -16,6 +16,7 @@ package cache
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/cache/ttl.go
```diff
@@ -63,7 +61,9 @@ func (c *ttlCache) put(key any, value any) {
 func (c *ttlCache) putWithTTL(key any, value any, ttl time.Duration) {
 	c.Lock()
 	defer c.Unlock()
-
+	if len(c.items) == 0 {
+		go c.doGC()
```
okJiang, line 65: There is a possibility that when `len(items) == 0`, the old goroutine has not exited, but a new put request arrives, which can lead to multiple `goGC()` goroutines existing for the same structure.

bufferflies: there is one lock，the previous must exit when the item queue is empty

okJiang: It is possible during the gaps in gcInterval.

okJiang: I understand now. The deletion can only be done through goGC, and the situation I mentioned will not occur.

okJiang (approved):

pkg/cache/ttl.go
bufferflies: you mean that the previous gc has deleted all items but not exit?

pkg/cache/ttl.go
```diff
@@ -163,6 +163,11 @@ func (c *ttlCache) doGC() {
 					}
 				}
 			}
+			if len(c.items) == 0 {
```
rleungx, line 166: How about `remove`?

bufferflies: it doesn't matter, one gc interval will exit.

pkg/cache/ttl.go
okJiang: 1. remove all items(not through goGC)
2. put one item(a new goGC())
3. old goGC() tick

This situation can lead to unexpected goroutines.

pkg/cache/ttl.go
bufferflies: good catch, I have added one atomic flag to check the goroutine exist.

rleungx (approved):

PR #9055, opened 2025-02-11
*: fix several bugs in QueryRegion

JmPotato:

Issue Number: ref #8690.

```commit-message
Fix several bugs in QueryRegion, including:

- Perform a deep copy of the region results from a gRPC response if used multiple times.
- Store the `QueryRegion` stream when it is not nil.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/clients/router/client.go
+++ b/client/clients/router/client.go
@@ -23,6 +23,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/opentracing/opentracing-go"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


pkg/core/region.go
```diff
@@ -1516,11 +1516,14 @@ func (r *RegionsInfo) getRegionsByKeys(keys [][]byte) []*RegionInfo {
 	regions := make([]*RegionInfo, 0, len(keys))
 	// Split the keys into multiple batches, and search each batch separately.
 	// This is to avoid the lock contention on the `regionTree`.
+	var results []*RegionInfo
 	for _, batch := range splitKeysIntoBatches(keys) {
 		r.t.RLock()
-		results := r.tree.searchByKeys(batch)
+		results = r.tree.searchByKeys(batch)
+		for _, region := range results {
```
rleungx, line 1523: What is the difference?

JmPotato: `result` is pre-defined to avoid being created multiple times during the batches.

`region` should be retrieved again by ID rather than just using the one from the tree.

rleungx (approved):

nolouch (approved):

client/clients/router/client.go
```diff
@@ -66,15 +67,21 @@ func ConvertToRegion(res regionResponse) *Region {
 		return nil
 	}
 
+	// Deep-copy here is necessary since the data inside the protobuf message can be reused,
+	// such like that if a []byte field in the protobuf message is reused, which might be changed
```
nolouch, line 71: why `RegionFromHeartbeat` works well?

nolouch (approved):
lgtm

PR #9078, opened 2025-02-13
client: skip if circuit breaker is disabled

rleungx:

Issue Number: ref #8678.

After this PR, if we disable the circuit breaker, the metrics won't be recorded.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


bufferflies (approved):

lhy1024 (approved):

rleungx:
> @rleungx what is the rationale for this change? The metrics are useful to decide what percent error to use to enable circuit breaker.

IMO, there is a lock for each request, and not everyone uses this feature. So I intend to disable it most of the time.

PR #9082, opened 2025-02-18
client: support dynamic start/stop of the router client

JmPotato:

Issue Number: ref #8690.

```commit-message
Support dynamic start and stop of the router client.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


client/opt/option.go
```diff
@@ -96,17 +103,14 @@ func (o *Option) SetMaxTSOBatchWaitInterval(interval time.Duration) error {
...
 	return nil
 }
 
 // SetEnableFollowerHandle set the Follower Handle option.
 func (o *Option) SetEnableFollowerHandle(enable bool) {
-	old := o.GetEnableFollowerHandle()
-	if enable != old {
-		o.dynamicOptions[EnableFollowerHandle].Store(enable)
-	}
+	o.dynamicOptions[EnableFollowerHandle].CompareAndSwap(!enable, enable)
```
rleungx, line 113: !enable and enable are always different?

JmPotato: The goal is to ensure that a swap occurs only when the old value differs from the new value. For instance, if `enable` is `true`, the swap will only take place if the old value of `EnableFollowerHandle` is `false`.

rleungx (approved):

tests/integrations/client/router_client_test.go
```diff
@@ -0,0 +1,352 @@
...
+	ctx, cancel := context.WithCancel(suite.ctx)
+	defer cancel()
+
+	wg := sync.WaitGroup{}
+	suite.dispatchConcurrentRequests(ctx, re, &wg)
+	// Concurrently enable and disable the router client.
+	for _, enabled := range []bool{!suite.routerClientEnabled, suite.routerClientEnabled} {
+		err := suite.client.UpdateOption(opt.EnableRouterClient, enabled)
+		re.NoError(err)
+	}
```
okJiang, line 350: How about adding a `sleep` and switching it more times?

okJiang (approved):

PR #9092, opened 2025-02-21
mcs: support batch allocating ids

### LINKED ISSUE #9085: Batch alloc ID for scheduling service
## Enhancement Task

https://github.com/tikv/pd/blob/daa23f832fb95ecda6c622beeaec3af30fd3f4a9/pkg/mcs/scheduling/server/grpc_service.go#L291-L311

For each split request, we need to call `AllocID` several times. When there are a large number of split requests, like import data scenarios, it costs too much resources. We can support batch alloc ID to reduce the call times.

Here is a profile:
Image

rleungx:

Issue Number: Close #9085.

kvproto PR: https://github.com/pingcap/kvproto/pull/1296

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


ystaticy (approved):

pkg/id/id.go
```diff
@@ -94,19 +94,21 @@ func NewAllocator(params *AllocatorParams) Allocator {
...
+func (alloc *allocatorImpl) Alloc(count uint32) (uint64, uint32, error) {
 	alloc.mu.Lock()
 	defer alloc.mu.Unlock()
 
-	if alloc.base == alloc.end {
-		if err := alloc.rebaseLocked(true); err != nil {
-			return 0, err
+	for range count {
+		if alloc.base == alloc.end {
+			if err := alloc.rebaseLocked(true); err != nil {
```
bufferflies, line 103: It may generate many etcd requests.

pkg/id/id.go
```diff
@@ -94,19 +94,21 @@ func NewAllocator(params *AllocatorParams) Allocator {
...
+			}
 		}
-	}
 
-	alloc.base++
+		alloc.base++
+	}
 
-	return alloc.base, nil
+	return alloc.base, count, nil
```
bufferflies, line 111: the result of count can be skipped

pkg/id/id.go
rleungx: Only when `alloc.base == alloc.end`, it will rebase.

pkg/core/region_test.go
```diff
@@ -839,15 +839,15 @@ func newRegionInfoIDRandom(idAllocator id.Allocator) *RegionInfo {
 	// Randomly select a peer as the leader.
 	leaderIdx := mrand.Intn(peerNum)
 	for i := range peerNum {
-		id, _ := idAllocator.Alloc()
+		id, _, _ := idAllocator.Alloc(1)
```
JmPotato, line 842: A lot of places are using `idAllocator.Alloc(1)`, what about providing a function named `AllocOne()`?

pkg/id/id.go
JmPotato: What about just calculating if the rebase is necessary rather than passively waiting for the loop to do it?

pkg/id/id.go
rleungx: The Alloc needs the `count` as the return value in the current implementation. We do have another way that does not add count in the response but I prefer to leave it.

pkg/core/region_test.go
rleungx: Then we need to add one more interface.

JmPotato (approved):

Issue #8978, opened 2025-01-06
Refactor GC related PD APIs


MyonKeminta:
## Development Task

This is the majority part of the refactoring task https://github.com/pingcap/tidb/issues/58720.

A set of new APIs for GC will be implemented which is expected to provide a more comprehensible semantics and recude potential misuses. This will also be the basis for implementing a planned feature name *protected snapshots*.

### Subtasks

Note that *protected snapshots* won't be included in this phase and will be another separated work after finishing the refactoring.

- [x] Add new RPC calls to kvproto https://github.com/pingcap/kvproto/pull/1288
- [x] Expose the if-then-else transaction interface of etcd to `kv.Base` layer, parallel to the existing `kv.Base.RunInTxn`. https://github.com/tikv/pd/pull/9016
- [ ] Add implementation new RPC interfaces. (may need to further breakdown into muitiple subtasks  later)
    - [x] Basic API set.
        - Including:
            - AdvanceGCSafePoint
            - AdvanceTxnSafePoint
            - SetGCBarrier
            - DeleteGCBarrier
            - GetGCState
            - GetGlobalGCState
        - [x] storage/endpoint layer https://github.com/tikv/pd/pull/9109
        - `GCStateManager` layer 
            - [x] https://github.com/tikv/pd/pull/9169
            - [x] https://github.com/tikv/pd/pull/9134
        - [x] gRPC layer https://github.com/tikv/pd/pull/9189
    - [ ] Global GC barriers support https://github.com/tikv/pd/pull/9340
        - [x] kvproto: https://github.com/pingcap/kvproto/pull/1315
    - [ ] Support controlling disabling physical GC operations on TiKV (required by BR, lower priority)
    - [ ] Support watching global GC state changes (including all keyspaces; *maybe* required by keyspace-level-GC for running GC in TiKV, lower priority, necessary when the amount of keyspaces in a single cluster grows too high)
- [ ] Implement the new set of methods on `pd.Client` interface type.
    - [x] Basic API set https://github.com/tikv/pd/pull/9203
    - [ ] Global GC barriers support
    - [ ] Support controlling disabling physical GC operations on TiKV (lower priority)
- [ ] New pd-ctl commands `gc-state` for replacing the current `service-gc-safepoint` command.
- [ ] Mark the old API deprecated. https://github.com/tikv/pd/pull/9292
- [ ] Make keyspace level GC enabled by default
    - [x] (Next gen only) https://github.com/tikv/pd/pull/9399

[...]

PR #9109, opened 2025-02-27
storage/endpoint: Add a new component `GCStateStorage` and `GCStateProvider` to storage/endpoints as the new storage layer for GC related metadata

MyonKeminta:

Issue Number: Ref #8978

```commit-message
Add a new component `GCStateStorage` and `GCStateProvider` to storage/endpoints as the new storage layer for GC related metadata.
This is part of the GC Interface Refactor task. The `GCSafePointStorage` is planned to be deprecated, and the GCStateStorage will be the replacement. It's currently not used, and will be used when we refactor the upper layer (`GCSafePointManager`->`GCStateManager`, the gRPC layer, etc.).
```

### Check List

Tests

- Unit test

Code changes

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,556 @@
...
+	gcSafePoint, err := strconv.ParseUint(value, 16, 64)
+	if err != nil {
+		return 0, errs.ErrStrconvParseUint.Wrap(err).GenWithStackByArgs()
+	}
+	return gcSafePoint, nil
+}
+
+// keyspaceGCSafePoint is the data structure when writing
+type keyspaceGCSafePoint struct {
+	KeyspaceID uint32 `json:"keyspace_id"`
```
rleungx, line 243: What is the ID used for?

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+	txnSafePoint, err := strconv.ParseUint(value, 10, 64)
+	if err != nil {
+		return 0, errs.ErrStrconvParseUint.Wrap(err).GenWithStackByArgs()
+	}
+	return txnSafePoint, err
+}
+
+// loadJSON loads a specific key from the StorageEndpoint, and parses it as JSON into type T.
+func loadJSON[T any](se *StorageEndpoint, key string) (T, error) {
```
rleungx, line 285: Shall we move loadJSON and loadJSONByPrefix to endpoint.go?

pkg/storage/endpoint/gc_states.go
MyonKeminta: Nothing but for keeping the data format compatible with the exisiting design of the "GC API V2". I forgot to finish the comments of this type. I'll fix it.

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,556 @@
...
+	return minKey, minMinStartTS, nil
+}
+
+// RunInGCStateTransaction runs a transaction for updating GC states or read a batch of GC states.
+// The atomicity is guaranteed by a "revision" key. Any non-empty write caused by the transaction will increase the
+// revision.
+// In the transaction, reads can be performed on the GCStateProvider as usual, while writes should only be performed
+// through the GCStateWriteBatch.
+func (p GCStateProvider) RunInGCStateTransaction(f func(wb *GCStateWriteBatch) error) error {
+	revisionKey := keypath.GCStateRevisionPath()
```
rleungx, line 409: Is it possible to use the revision provided by etcd?

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,556 @@
...
+			Key:    revisionKey,
+			OpType: kv.RawTxnOpPut,
+			Value:  nextRevision,
+		})
+	}
+
+	txn := p.storage.CreateRawTxn()
+	result, err := txn.If(condition).Then(ops...).Commit()
+	if err != nil {
+		return errors.AddStack(err)
```
rleungx, line 451: How about `return errs.ErrEtcdTxnInternal.Wrap(err).GenWithStackByArgs()`

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+	if !result.Succeeded {
+		return errs.ErrEtcdTxnConflict.GenWithStackByArgs()
+	}
+
+	if len(ops) != len(result.Responses) {
```
rleungx, line 457: If this happens, do we need to rollback or manually fix it?

pkg/storage/endpoint/gc_states.go
MyonKeminta: Yes but hard to provide it over the abstracted `kv.Base` interface.

MyonKeminta: And it needs a single centrialized key anyway.

pkg/storage/endpoint/gc_states.go
MyonKeminta: If I understanding is correct, etcd shoudl guarantee this to be true. I expect that this path is never reachable. If this happen, I think it should be a bug in etcd or etcd-client and it doesn't look possible to automatically fix it in a simple way. And as the `Commit` returns no error here, it seems we no longer have chance to roll it back atomically. I think if this does happen, we may have no choice but to manually recover it.

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,541 @@
+// Copyright 2024 TiKV Project Authors.
```
ystaticy, line 1:
```suggestion
// Copyright 2025 TiKV Project Authors.
```

pkg/utils/keypath/absolute_key_path.go
```diff
@@ -61,10 +61,18 @@ const (
 	// "%08d" adds extra padding to make encoded ID ordered.
 	// Encoded ID can be decoded directly with strconv.ParseUint. Width of the
 	// padded keyspaceID is 8 (decimal representation of uint24max is 16777215).
-	gcSafePointPathFormat        = "/pd/%d/gc/safe_point/"                       // "/pd/{cluster_id}/gc/safe_point"
-	gcSafePointServicePathFormat = "/pd/%d/gc/safe_point/service/%s"             // "/pd/{cluster_id}/gc/safe_point/service/{service_id}"
-	gcSafePointV2PathFormat      = "/pd/%d/keyspaces/gc_safe_point/%08d"         // "/pd/{cluster_id}/keyspaces/gc_safe_point/{keyspace_id}"
-	serviceSafePointV2PathFormat = "/pd/%d/keyspaces/service_safe_point/%08d/%s" // "/pd/{cluster_id}/keyspaces/service_safe_point/{keyspace_id}/{service_id}"
+	gcStateRevisionPathFormat     = "/pd/%d/gc/gc_state_revision"         // "/pd/{cluster_id}/gc/gc_state_revision"
+	gcSafePointPathFormat         = "/pd/%d/gc/safe_point"                // "/pd/{cluster_id}/gc/safe_point"
```
ystaticy, line 65: How about naming this variable globalGCSafePointPathFormat directly? Other developers might have questions about the relationship between gcSafePointPathFormat and keyspaceGCSafePointPathFormat: keyspaceGCSafePointPathFormat.

pkg/utils/keypath/safe_point.go
```diff
@@ -18,19 +18,74 @@ import (
 	"fmt"
 )
 
-// GCSafePointPath returns the GC safe point key path.
+// GCStateRevisionPath returns the key path for storing the revision of GC state data.
+func GCStateRevisionPath() string {
+	return fmt.Sprintf(gcStateRevisionPathFormat, ClusterID())
+}
+
+// GCSafePointPath returns the key path of the global (NullKeyspace) GC safe point.
```
ystaticy, line 26: How about explicitly introducing and explaining the relationship between global and NullKeyspace in the code comments? This can help everyone understand its meaning clearly.

pkg/storage/endpoint/gc_states.go
MyonKeminta: Time went too fast before I noticed it😕

pkg/storage/endpoint/gc_states_test.go
```diff
@@ -0,0 +1,748 @@
...
+	"math/rand"
+	"strconv"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"golang.org/x/exp/slices"
```
rleungx, line 30: use "slices"

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	if len(ops) != len(result.Responses) {
+		return errors.Errorf("unexpected number of results: %d != %d", len(ops), len(result.Responses))
+	}
+	return nil
+}
+
+// CompatibleLoadAllServiceGCSafePoints returns all services GC safe points with their etcd key.
+func (p GCStateProvider) CompatibleLoadAllServiceGCSafePoints() ([]string, []*ServiceSafePoint, error) {
```
rleungx, line 443: Is it for only test purpose?

rleungx (commented):
The rest LGTM

pkg/storage/endpoint/gc_states.go
MyonKeminta: No. It will be needed by the HTTP API (`server/api/service_gc_safepoint.go`) when we abandoned the old code in `GCSafePointStorage`.

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+type ServiceSafePoint struct {
+	ServiceID string
+	ExpiredAt int64
+	SafePoint uint64
+
+	// Note than when marshalled into JSON, omitting KeyspaceID stands for the NullKeyspace (0xffffffff),
+	// rather than KeyspaceID = 0 which is the ID of the default keyspace.
+	// Special marshalling / unmarshalling methods are given for handling this field in a non-default way.
+	//
+	// The purpose is to make the code (for keyspaced and non-keyspaced/global GC) unified while keeping the
```
ystaticy, line 79: How about we consistently use 'unified' in the comments to represent the GC of Null keyspace, and use 'Null keyspace' to represent the situation of 'unconfigured keyspace'? Let's avoid using terms like 'non-keyspaced' and 'global GC' or other expressions. This will help standardize the definitions of these concepts.

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+
+// LoadGCSafePoint loads current GC safe point of the given keyspaceID from storage.
+func (p GCStateProvider) LoadGCSafePoint(keyspaceID uint32) (uint64, error) {
+	if keyspaceID == constant.NullKeyspaceID {
+		return p.loadGCSafePointForUnifiedGC()
+	}
+	return p.loadGCSafePointForKeyspaceLevelGC(keyspaceID)
+}
+
+// loadGCSafePointForUnifiedGC loads the GC safe point of the global GC.
```
ystaticy, line 260:
```suggestion
// loadGCSafePointForUnifiedGC loads the GC safe point of the unified GC.
```

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+
+// SetGCSafePoint sets the GC safe point for the given keyspace.
+func (wb *GCStateWriteBatch) SetGCSafePoint(keyspaceID uint32, gcSafePoint uint64) error {
+	if keyspaceID == constant.NullKeyspaceID {
+		return wb.setGCSafePointForUnifiedGC(gcSafePoint)
+	}
+	return wb.setGCSafePointForKeyspaceLevelGC(keyspaceID, gcSafePoint)
+}
+
+// setGCSafePointForUnifiedGC sets the GC safe point for global GC (NullKeyspace).
```
ystaticy, line 478:
```suggestion
// setGCSafePointForUnifiedGC sets the GC safe point for unified GC (NullKeyspace).
```

pkg/storage/endpoint/gc_states.go
MyonKeminta: It seems some comments are missed when I update it. I'll check it again.

pkg/utils/keypath/gc_states.go
```diff
@@ -0,0 +1,97 @@
...
+
+// ServiceSafePointV2Prefix is the path prefix of all service safe point that belongs to a specific keyspace.
+// Can be used to retrieve keyspace's service safe point at once.
+func ServiceSafePointV2Prefix(keyspaceID uint32) string {
+	return ServiceSafePointV2Path(keyspaceID, "")
+}
+
+// GCSafePointV2Prefix is the path prefix to all gc safe point v2.
+func GCSafePointV2Prefix() string {
+	return Prefix(GCSafePointV2Path(0))
```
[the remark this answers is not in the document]
ystaticy: It seems that he intends to retrieve the first key in order when getting/watching all keyspace-related data for traversal. However, the naming of this function suggests that it is meant to obtain the prefix of etcd keys. The naming needs to be revised, and keyspace ID = 0 should not be used.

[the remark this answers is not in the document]
MyonKeminta: This came from https://github.com/tikv/pd/pull/8919 and I didn't touch it. It shows up as new code because the file is renamed (previously safe_point.go).
cc @okJiang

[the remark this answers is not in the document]
MyonKeminta: It seems this code uses the `Prefix` funciton to trim out the keyspace ID 0 so that only the prefix without keyspace ID part is returned.
But it seems it doesn't keep the trailing slash '/', which looks not consistent with other functions in the same file.
@okJiang PTAL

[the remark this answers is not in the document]
okJiang: @ystaticy When Prefix is called, the input parameters do not take effect, meaning that regardless of whether 0 or 100 is passed, it will return Prefix. This is a way of reusing the Format, as we don't want to maintain a separate path specifically for the Prefix. For now, there doesn't seem to be a better solution.

FYI, you can use `git move` to keep your change when you rename file. @MyonKeminta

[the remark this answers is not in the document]
okJiang: > It seems this code uses the `Prefix` funciton to trim out the keyspace ID 0 so that only the prefix without keyspace ID part is returned. But it seems it doesn't keep the trailing slash '/', which looks not consistent with other functions in the same file. @okJiang PTAL

This is indeed a mistake, and I will try to write a more robust test to ensure the compatibility of the code ASAP.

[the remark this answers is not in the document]
MyonKeminta: Ok.
I don't plan to fix it in this PR as it's out of the scope. cc @ystaticy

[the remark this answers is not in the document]
okJiang: > I don't plan to fix it in this PR as it's out of the scope.

Agree

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+	if repr.KeyspaceID != nil {
+		s.KeyspaceID = *repr.KeyspaceID
+	} else {
+		s.KeyspaceID = constant.NullKeyspaceID
+	}
+	return nil
+}
+
+var _ json.Marshaler = &ServiceSafePoint{}
+var _ json.Unmarshaler = &ServiceSafePoint{}
```
JmPotato, line 139:
```suggestion
var （
	_ json.Marshaler = (*ServiceSafePoint)(nil)
	_ json.Unmarshaler = (*ServiceSafePoint)(nil)
)
```

ystaticy (approved):

pkg/storage/endpoint/gc_states_test.go
JmPotato, line 1: Perhaps add some test cases for `(*GCBarrier).IsExpired()`?

JmPotato (approved):

JmPotato (approved):

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,526 @@
...
+// GetGCStateProvider returns an GCStateProvider for reading and writing GC state data.
+// GCStateProvider provides direct read/write to GC related metadata without maintaining the consistency or invariants.
+// It's only designed for internal use of GC. Avoid using it for other purposes.
+func (se *StorageEndpoint) GetGCStateProvider() GCStateProvider {
+	return newGCStateProvider(se)
+}
+
+// GCStateProvider is a stateless wrapper over StorageEndpoint that provides methods for reading/writing GC states.
+// It can be dangerous to misuse GC related operations. As an explicit wrapper, it hides the GC related methods away
+// from the StorageEndpoint type and the Storage interface, making it less likely to be misused unintentionally when
```
okJiang, line 237: Do we need to ensure the order inside `GCStateProvider`? For example, check `GCBarrier >= TxnSafePoint` when we set GCBarrier or set TxnSafePoint?

pkg/storage/endpoint/gc_states_test.go
```diff
@@ -0,0 +1,782 @@
...
+	testData := []uint64{0, 1, 2, 233, 2333, 23333333333, math.MaxUint64}
+	keyspaceIDs := []uint32{constant.NullKeyspaceID, 0, 1000}
+
+	for _, keyspaceID := range keyspaceIDs {
+		res, err := provider.LoadGCSafePoint(keyspaceID)
+		re.NoError(err)
+		re.Equal(uint64(0), res)
+		for _, gcSafePoint := range testData {
+			// For checking physical data representation.
+			expectedKey := "/pd/0/gc/safe_point"
```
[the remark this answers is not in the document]
MyonKeminta: It's intentional written by literal to reduce the possibility of undetected mistakes.

pkg/storage/endpoint/gc_states.go
```diff
@@ -0,0 +1,524 @@
...
+// Passing nil to `expirationTime` means the barrier never expires.
+func NewGCBarrier(barrierID string, barrierTS uint64, expirationTime *time.Time) *GCBarrier {
+	// Round up the expirationTime.
+	if expirationTime != nil {
+		rounded := expirationTime.Truncate(time.Second)
+		if rounded.Before(*expirationTime) {
+			rounded = rounded.Add(time.Second)
+		}
+		*expirationTime = rounded
+	}
```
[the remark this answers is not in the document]
MyonKeminta: Perhaps you meant `.Add(time.Second - time.Nanosecond).Truncate(time.Second)`? It sounds a good idea.
Golang does't allow taking address from an rvalue expression though... which I don't know why

pkg/storage/endpoint/gc_states.go
MyonKeminta: No.  As said in the comments at the beginning of the `gc_states.go` file, this layer is only responsible for providing methods for read and write. The upper layer `GCStateManager`, which will come in the next PR, is responsible for maintaining the invariants.

pkg/utils/keypath/absolute_key_path.go
```diff
@@ -61,10 +61,19 @@ const (
 	// "%08d" adds extra padding to make encoded ID ordered.
 	// Encoded ID can be decoded directly with strconv.ParseUint. Width of the
 	// padded keyspaceID is 8 (decimal representation of uint24max is 16777215).
-	gcSafePointPathFormat        = "/pd/%d/gc/safe_point"                        // "/pd/{cluster_id}/gc/safe_point"
-	gcSafePointServicePathFormat = "/pd/%d/gc/safe_point/service/%s"             // "/pd/{cluster_id}/gc/safe_point/service/{service_id}"
-	gcSafePointV2PathFormat      = "/pd/%d/keyspaces/gc_safe_point/%08d"         // "/pd/{cluster_id}/keyspaces/gc_safe_point/{keyspace_id}"
-	serviceSafePointV2PathFormat = "/pd/%d/keyspaces/service_safe_point/%08d/%s" // "/pd/{cluster_id}/keyspaces/service_safe_point/{keyspace_id}/{service_id}"
+	gcStateRevisionPathFormat          = "/pd/%d/gc/gc_state_revision"         // "/pd/{cluster_id}/gc/gc_state_revision"
```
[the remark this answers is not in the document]
MyonKeminta: I'm afraid the phrase "GC state revision" doesn't make sense clearly.

PR #9116, opened 2025-03-05
*: switch to rand v2

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/core/region_option.go
```diff
@@ -396,8 +395,7 @@ func SetSource(source RegionSource) RegionCreateOption {
 
 // RandomKindReadQuery returns query stat with random query kind, only used for unit test.
 func RandomKindReadQuery(queryRead uint64) *pdpb.QueryStats {
-	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
lhy1024, line 399: We don't need to create new one after v2?

lhy1024 (approved):

bufferflies (approved):

PR #9117, opened 2025-03-05
metrics: add region request metrics which record caller info

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/8593

```commit-message
Add two panels Success Region Request(1% sample) and Failed Region Request in Grpc 
```

### Check List

Tests

- Unit test
- Integration test

image

Related changes

### Release note

```release-note
None.
```


server/metrics.go
```diff
@@ -177,6 +180,14 @@ var (
...
 			Buckets:   prometheus.ExponentialBuckets(0.0005, 2, 13),
 		})
+
+	regionRequestCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "pd",
+			Subsystem: "server",
+			Name:      "region_request_cnt",
+			Help:      "Counter of region request.",
+		}, []string{"request", "caller_id", "caller_component", "err_msg"})
```
rleungx, line 190: /Can we make it more simple, e.g., request-component-module?

rleungx: `ScanRegions-tidb-server->tidb-tikv-driver` is kind of confusing and not readable.

okJiang: request-component-module LGTM

However, there are `caller_id` and caller_component in the current code. If we use `request-component-module` here, it will cause confusion with the term `component`.

> ScanRegions-tidb-server->tidb-tikv-driver is kind of confusing and not readable.

It is indeed easy to get confused, and I am still thinking of a better format. Since we chose to use a hyphen as a connector in the component, continuing to use a hyphen in Grafana might look a bit strange.

image

Maybe we should name the component using camel case? Such as `RegionCache`

client/pkg/caller/caller.go
```diff
@@ -31,8 +31,27 @@ type (
...
+	InterceptedPDClient Component = "intercepted-pd-client"
+
+	// TiDB component
+	Pitr             Component = "pitr"
+	Ddl              Component = "ddl"
+	ImportInto       Component = "import-into"
+	TikvHandler      Component = "tikv-handler"
+	GcWorker         Component = "gc-worker"
+	GcJob            Component = "gc-job"
+	DistributedGcJob Component = "distributed-gc-job"
```
okJiang, line 54: These names can be combined with these two PR to review. https://github.com/pingcap/tidb/pull/59911/files and https://github.com/tikv/client-go/pull/1516

server/metrics.go
okJiang: What do you think? cc @lhy1024 @JmPotato @bufferflies

client/pkg/caller/caller.go
```diff
@@ -31,8 +31,27 @@ type (
...
 	TestID ID = "test"
+
 	// TestComponent is used for test.
 	TestComponent Component = "test"
+
+	// client-go component
+	CodecPDClient       Component = "codec-pd-client"
+	RegionCache         Component = "region-cache"
+	StoreCache          Component = "store-cache"
+	Oracles             Component = "oracle"
```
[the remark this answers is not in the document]
okJiang: These two pr https://github.com/tikv/pd/pull/9117/files/ebd66e5b0042353ad6402f6239ab049a438c21a0#r1982716822

And does this your comment say? https://github.com/tikv/client-go/pull/1516#discussion_r1969149999

server/metrics.go
```diff
@@ -199,4 +211,23 @@ func init() {
 	prometheus.MustRegister(apiConcurrencyGauge)
 	prometheus.MustRegister(forwardFailCounter)
 	prometheus.MustRegister(forwardTsoDuration)
+	prometheus.MustRegister(regionRequestCounter)
+}
+
+func incRegionRequestCounter(method string, header *pdpb.RequestHeader, err *pdpb.Error) {
```
bufferflies, line 217: Hash function is little heavy, and we don't need to report all the request just see the the source. So we can use sample to report the normal request to decrease the usage of hash function.

okJiang: This can indeed lighten the load; for example, we can collect metrics with a probability of one percent. Even with tens of thousands of requests, we can still identify the sources of these large requests.

If there are no objections from others, I will make the changes today.

okJiang: Done~ https://github.com/tikv/pd/pull/9117/commits/7c00c9e9024d5236efbf4adb15d3aa8ace494eb7

client/pkg/caller/caller.go
```diff
@@ -28,11 +28,31 @@ type (
 	Component string
 )
 
+// nolint: exported
```
JmPotato, line 31: Is this necessary?

okJiang: Yes, otherwise it would require a comment for each variable

okJiang:
ping @bufferflies @JmPotato @rleungx

server/metrics.go
```diff
@@ -199,4 +213,28 @@ func init() {
...
+	if err != nil {
+		errMsg = err.Type.String()
+	}
+	if callerID == "" {
+		callerID = "unknown"
+	}
+	if callerComponent == "" {
+		callerComponent = "unknown"
+	}
+	regionRequestCounter.WithLabelValues(method, callerID, callerComponent, errMsg).Inc()
```
bufferflies, line 239: if the error is nil, the inc one is wrong; it should multi * 100.

bufferflies (commented):
rest lgtm

server/metrics.go
```diff
@@ -199,4 +213,28 @@ func init() {
...
 	prometheus.MustRegister(forwardFailCounter)
 	prometheus.MustRegister(forwardTsoDuration)
+	prometheus.MustRegister(regionRequestCounter)
+}
+
+func incRegionRequestCounter(method string, header *pdpb.RequestHeader, err *pdpb.Error) {
+	if err == nil && rand.Intn(100) != 0 {
+		// sample 1% region requests to avoid high cardinality
+		return
+	}
```
JmPotato, line 223: Instead of sampling with a specific probability, can we record the data into a counter and periodically flush it to Prometheus?

okJiang: Recording to the counter also requires calculation via a hash map, and we need to ensure it's thread-safe, which likewise introduces considerable overhead.

server/metrics.go
JmPotato: `request-id-component-msg` is clear enough to me.

server/metrics.go
JmPotato, line 239: Apart from the `errMsg` field, does caching the `Counter` composed of the first three fields help improve performance? `WithLabelValues` is not considered a very lightweight operation.

bufferflies (approved):

PR #9126, opened 2025-03-10
test: add etcd key compatibility test

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/9127

This PR introduces a new etcd key compatibility test, ~~with its test data based on release 8.5. It is only for reviewing the code of this test, so I have not enabled it. In the next PR, I will fix the changes to the etcd keys from release 8.5 to master and provide source explanations for those changes.~~ I has updated the test data using master code, and enable this test.

Code Overview: I used failpoint to collect the operation of keys during the test and saved the results to a file. If the results from the new code do not match those in the file, it indicates that the operations on the keys have changed during testing. This compels us to pay attention to changes in the keys.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Please refer to the test results [here](https://github.com/tikv/pd/pull/9125); they are based on test data generated from the release-8.5 code.

### Release note

```release-note
None.
```


pkg/utils/etcdutil/etcdutil.go
```diff
@@ -167,6 +170,44 @@ func EtcdKVGet(c *clientv3.Client, key string, opts ...clientv3.OpOption) (*clie
 	return resp, nil
 }
 
+// InjectFailToCollectTestEtcdKey injects the failpoint to collect the key for testing.
+func InjectFailToCollectTestEtcdKey(key, op string) {
+	failpoint.Inject("CollectEtcdKey", func(val failpoint.Value) {
+		file := val.(string)
+		fmt.Println("collect etcd key", file, key, op)
```
lhy1024, line 177: Do we need to output it?

tools/pd-ut/testdata/group.1.etcdkey
```diff
@@ -0,0 +1,75 @@
+ get
```
lhy1024, line 1: Are these files necessary? How will we maintain them?

nolouch: How about adding a script to generate it?

okJiang: It will be print after execute `./scripts/ci-subtask.sh {group_id}`. So we can see it in the github action 
image

BTW: the code is in https://github.com/okJiang/pd/blob/7025e1f6144f5f6e40546246e97b452e11d1329f/tools/pd-ut/ut.go#L419. If diff exists, it will also print https://github.com/okJiang/pd/blob/7025e1f6144f5f6e40546246e97b452e11d1329f/tools/pd-ut/ut.go#L421-L437


> How will we maintain them?

If the ci failed, we can copy the log to the xxxx.etcdkey

pkg/utils/etcdutil/etcdutil.go
okJiang: done

lhy1024 (approved):

Issue #9132, opened 2025-03-11
The etcd key has been modified incorrectly.


okJiang:
## Bug Report

- gcSafePointV2Prefix: "/pd/%d/keyspaces/gc_safe_point/" -> "/pd/%d/keyspaces/gc_safe_point"
- servicePath: "/ms/{cluster_id}/{service_name}/registry/" -> "/ms/{cluster_id}/{service_name}/registry"
- watch key
    - "/pd/{cluster_id}/tso/keyspace_groups/membership" -> "/pd/{cluster_id}/tso/keyspace_groups/membership/"
    - "/pd/{cluster_id}/region_label" -> "/pd/{cluster_id}/region_label/"
    - "/pd/{cluster_id}/scheduler_config" -> "/pd/{cluster_id}/scheduler_config/"
- (Important) schedulerConfigPathFormat: "/pd/{cluster_id}/**scheduler_config**/{scheduler_name}" -> "/pd/{cluster_id}/**schedule**/{scheduler_name}"
- Error LoadTimestamp: 
    - before: "/pd/{cluster_id}/timestamp" and "/ms/{cluster_id}/tso/{group_id}/gta/timestamp"
    - after: "/pd/{cluster_id}" and "/ms/{cluster_id}/tso/{group_id}/gta", The suffix was incorrectly removed.

The above conclusion is based on testing of https://github.com/tikv/pd/pull/9131. To meet the code freeze deadline, we should prioritize fixing this issue.

### What version of PD are you using (`pd-server -V`)?

master

PR #9133, opened 2025-03-11
keypath: fix incorrect key

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9132

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

ref https://github.com/tikv/pd/pull/9131

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/config/watcher.go
+++ b/pkg/mcs/scheduling/server/config/watcher.go
@@ -215,6 +215,9 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 	cw.schedulerConfigWatcher = etcdutil.NewLoopWatcher(
 		cw.ctx, &cw.wg,
 		cw.etcdClient,
+		// NOTE: schedulerConfigPathPrefix is "/pd/{cluster_id}/scheduler_config" before.
+		// Now it is "/pd/{cluster_id}/scheduler_config/". I think This has no impact.
+		// If it needs to be fixed, I can update it.
 		"scheduling-scheduler-config-watcher", cw.schedulerConfigPathPrefix,
 		func([]*clientv3.Event) error { return nil },
 		putFn, deleteFn,
--- a/pkg/mcs/scheduling/server/config/watcher.go
+++ b/pkg/mcs/scheduling/server/config/watcher.go
@@ -42,10 +42,6 @@ type Watcher struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	// configPath is the path of the configuration in etcd:
-	//  - Key: /pd/{cluster_id}/config
-	//  - Value: configuration JSON.
-	configPath string
 	// schedulerConfigPathPrefix is the path prefix of the scheduler configuration in etcd:
 	//  - Key: /pd/{cluster_id}/scheduler_config/{scheduler_name}
 	//  - Value: configuration JSON.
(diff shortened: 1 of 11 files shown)


pkg/mcs/scheduling/server/config/watcher.go
```diff
@@ -215,6 +215,9 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 	cw.schedulerConfigWatcher = etcdutil.NewLoopWatcher(
 		cw.ctx, &cw.wg,
 		cw.etcdClient,
+		// NOTE: schedulerConfigPathPrefix is "/pd/{cluster_id}/scheduler_config" before.
+		// Now it is "/pd/{cluster_id}/scheduler_config/". I think This has no impact.
+		// If it needs to be fixed, I can update it.
```
okJiang, line 220: These comment will be removed after review

JmPotato: I think it's best to remain consistent with what it was before.

pkg/mcs/scheduling/server/rule/watcher.go
```diff
@@ -238,6 +238,9 @@ func (rw *Watcher) initializeRegionLabelWatcher() error {
 	rw.labelWatcher = etcdutil.NewLoopWatcher(
 		rw.ctx, &rw.wg,
 		rw.etcdClient,
+		// NOTE: regionLabelPathPrefix is "/pd/{cluster_id}/region_label" before.
+		// Now it is "/pd/{cluster_id}/region_label/". I think This has no impact.
+		// If it needs to be fixed, I can update it.
```
JmPotato, line 243: Ditto.

pkg/tso/keyspace_group_manager_test.go
```diff
@@ -109,15 +109,15 @@ func (suite *keyspaceGroupManagerTestSuite) TestDeletedGroupCleanup() {
 	suite.applyEtcdEvents(re, []*etcdEvent{generateKeyspaceGroupPutEvent(1, []uint32{1}, []string{svcAddr})})
 	// Check if the TSO key is created.
 	testutil.Eventually(re, func() bool {
-		ts, err := mgr.storage.LoadTimestamp(keypath.Prefix(keypath.TimestampPath(1)))
+		ts, err := mgr.storage.LoadTimestamp(keypath.TimestampPath(1))
```
lhy1024, line 112: Do we need to add some other test to avoid the similiar problems?

pkg/mcs/scheduling/server/config/watcher.go
okJiang: Done

pkg/tso/keyspace_group_manager_test.go
okJiang: I think https://github.com/tikv/pd/pull/9126 will be a stronger test to avoid this.

okJiang: BTW, the reason that we do not detect the problem of `LoadTimestamp` is the non-timestamp key has been continue in this. So it will not cause any actual damage. 😂

https://github.com/tikv/pd/blob/e245c1fb30c9ccac07d30f813b2bbd486f31c058/pkg/storage/endpoint/tso.go#L56-L60

pkg/mcs/scheduling/server/config/watcher.go
okJiang: The previous commit had issues, so I submitted a new one. https://github.com/tikv/pd/pull/9133/commits/6b837249ae6d06cdd366b9a91eee929457eaf573

JmPotato (approved):

PR #9137, opened 2025-03-12
mcs: fix the compatibility issue of batch allocating IDs

### LINKED ISSUE #9138: ID will be duplicated if the scheduling service upgrade first
## Bug Report

Previously, we only ensured compatibility issues when PD was upgraded first. If the scheduling server completed the upgrade first but PD has not been upgraded yet, there will be duplicate IDs.

1. Assume that the current PD base ID is 10. 
2. The upgraded scheduling server receives the `AskBatchSplit` request. Assume that the request count is 1. 
3. But the PD server has not been upgraded at this time, it is sent to the PD server for processing as a non-batch request. The base ID returned to the caller is 11, the region ID is 11, and the peer IDs are 12, 13, and 14 respectively.
4. After the PD server upgrade is completed, the scheduling server receives the `AskBatchSplit` request again. 
5. Assuming that the request count is 1, it is processed as a batch request. And the base ID returned to the caller is `15-4=11`, the region ID is 11, and the peer IDs are 12, 13, and 14. 

So the bug is triggered that the region ID is the same but the range is different.

rleungx:

Issue Number: Close #9138.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/grpc_service.go
+++ b/pkg/mcs/scheduling/server/grpc_service.go
@@ -288,28 +288,56 @@ func (s *Service) AskBatchSplit(_ context.Context, request *schedulingpb.AskBatc
 	splitIDs := make([]*pdpb.SplitID, 0, splitCount)
 	recordRegions := make([]uint64, 0, splitCount+1)
 
-	id, count, err := c.AllocID(splitCount * (1 + uint32(len(request.Region.Peers))))
+	requestIDCount := splitCount * (1 + uint32(len(request.Region.Peers)))
+	id, count, err := c.AllocID(requestIDCount)
 	if err != nil {
 		return nil, err
 	}
-	curID := id - uint64(count)
-	for range splitCount {
-		newRegionID := curID
-		curID++
 
-		peerIDs := make([]uint64, len(request.Region.Peers))
-		for j := 0; j < len(peerIDs); j++ {
-			peerIDs[j] = curID
-			curID++
+	// If the count is not equal to the requestIDCount, it means that the
+	// PD doesn't support allocating IDs in batch. We need to allocate IDs
+	// for each region.
+	if requestIDCount != count {
+		// use old way to split region
+		for range splitCount {
+			newRegionID, _, err := c.AllocID(1)
+			if err != nil {
+				return nil, errs.ErrSchedulerNotFound.FastGenByArgs()
+			}
+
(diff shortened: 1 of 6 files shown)


tests/integrations/mcs/scheduling/server_test.go
```diff
@@ -833,3 +833,101 @@ func (suite *serverTestSuite) TestBatchSplit() {
...
+	re.Len(allocatedIDs, 40)
+
+	// simulate the new version of PD
+	re.NoError(failpoint.Disable("github.com/tikv/pd/server/allocIDOnce"))
+
+	resp, err = grpcPDClient.AskBatchSplit(suite.ctx, req)
+	re.NoError(err)
+	re.Empty(resp.GetHeader().GetError())
+	for _, id := range resp.GetIds() {
+		_, ok := allocatedIDs[id.NewRegionId]
```
okJiang, line 922: How about checking `NewRegionId > all allocatedIDs`?

rleungx: add a check in grpc_service.go

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -298,47 +299,39 @@ func (s *Service) AskBatchSplit(_ context.Context, request *schedulingpb.AskBatc
 	// PD doesn't support allocating IDs in batch. We need to allocate IDs
 	// for each region.
 	if requestIDCount != count {
-		// use old way to split region
-		for range splitCount {
-			newRegionID, _, err := c.AllocID(1)
+		for range requestIDCount - 1 {
+			id, _, err = c.AllocID(1)
```
okJiang, line 303: Can we guarantee that the IDs returned from consecutive requests are sequential? What if another client makes an ID request in the meantime?

okJiang: Maybe we can save these ids and handle them together later.

rleungx: I plan to revert to the original implementation. /cc @bufferflies

rleungx: I have added a concurrency test.

bufferflies (approved):

okJiang (approved):

PR #9148, opened 2025-03-18
client: add GetLeaders method

### LINKED ISSUE #9147: Client support get the etcd and etcd leader of the pd cluster
## Enhancement Task
Consider add `GetLeaders` method to the client.

disksing:

Issue Number: Close #9147

```commit-message
client: add GetLeaders method
```

### Check List

Tests

### Release note

```release-note
None.
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -490,7 +490,7 @@ func (c *client) GetAllMembers(ctx context.Context) ([]*pdpb.Member, error) {
 }
 
 // GetLeaders gets the leader and etcd leader from PD.
-func (c *client) GetLeaders(ctx context.Context) (*pdpb.Member, *pdpb.Member, error) {
+func (c *client) GetLeaders(ctx context.Context) (leader *pdpb.Member, etcdLeader *pdpb.Member, err error) {
 	start := time.Now()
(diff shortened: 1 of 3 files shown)


JmPotato (approved):

client/client.go
```diff
@@ -56,6 +56,8 @@ type GlobalConfigItem struct {
 type RPCClient interface {
 	// GetAllMembers gets the members Info from PD
 	GetAllMembers(ctx context.Context) ([]*pdpb.Member, error)
+	// GetLeader gets the current leader and etcd leader from PD
+	GetLeaders(ctx context.Context) (leader *pdpb.Member, etcdLeader *pdpb.Member, err error)
```
rleungx, line 60: If, in some places, we both need members and leaders, then we need to call two RPCs?

disksing: yes. here we try to not break capatibility of old method.

rleungx (approved):

client/client.go
bufferflies: how about define one response struct for this if we needs the informations of the follower?

Issue #9154, opened 2025-03-19
Region will not be synced if the channel buffer is full


okJiang:
## Bug Report

https://github.com/tikv/pd/blob/e8050f72d00db4b57dcac1306c1de8e3cc7fb9b5/server/cluster/cluster.go#L1293-L1298

### What did you do?

Too large heartbeat qps leads to `changedRegions` full, and the region miss syncing.

### What did you expect to see?

Keep syncing

### What did you see instead?

Region miss in the PD follower.

okJiang:
<img width="954" alt="Image" src="https://github.com/user-attachments/assets/377702dd-b116-4b08-abb3-ca9909762d0d" />

Added logging and monitoring, and the issue has been reproduced. This is a scenario of a BR restore with a 3M region and 200T of data, which will generate a large number of heartbeats in a short period of time.

okJiang:
Although https://github.com/tikv/pd/pull/9159 can prevent sync tasks from being discarded, it may lead to a significant backlog of tasks. To address this issue more effectively, we need to greatly enhance the consumption capacity of the syncer to avoid task accumulation.

For example, we can directly enable multiple streams for synchronization for each follower.

okJiang:
> Although [#9159](https://github.com/tikv/pd/pull/9159) can prevent sync tasks from being discarded, it may lead to a significant backlog of tasks. To address this issue more effectively, we need to greatly enhance the consumption capacity of the syncer to avoid task accumulation.
> 
> For example, we can directly enable multiple streams for synchronization for each follower.

All is in https://github.com/tikv/pd/pull/9159

okJiang:
If pending tasks accumulate for too long, newly arrived tasks will be regarded as failures. Therefore, we need to set retained=true for the sync runner. 

<img width="1484" alt="Image" src="https://github.com/user-attachments/assets/93229808-480f-4e6e-bf88-f29f5241de59" />

PR #9159, opened 2025-03-19
syncer: use task runner to sync region

okJiang:

~~This PR only fixes the bug; there will still be some delay between the follower and leader regions. In the future, we will increase the concurrency of consumers to alleviate the delay between the follower and leader under these conditions.~~

Issue Number: Close https://github.com/tikv/pd/issues/9154

- use task runner to sync region
- add a case to test sync region task backlog
- enlarge the sync batch 100 -> 1000
- concurrently synchronize with different followers to increase consumption capacity.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix a bug that the region of follower synced from leader may miss during large of region heartbeat.
```


okJiang:
/hold 

I will try to test in a real cluster.

server/cluster/cluster.go
```diff
@@ -1291,10 +1301,13 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
...
 
 	if saveKV || needSync {
-		select {
-		case c.changedRegions <- region:
-		default:
-		}
+		ctx.SyncRegionRunner.RunTask(
+			regionID,
+			ratelimit.SyncRegionToFollower,
+			func(context.Context) {
```
lhy1024, line 1307: Do we need to add a unit test for it?

lhy1024 (approved):
LGTM

pkg/syncer/server.go
```diff
@@ -346,36 +348,44 @@ func (s *RegionSyncer) bindStream(name string, stream ServerStream) {
...
-				log.Info("region syncer delete the stream", zap.String("stream", name))
-			}
-			s.mu.Unlock()
-		}
+		}(name, sender)
+	}
+
+	go func() {
+		wg.Wait()
+		failed.Range(func(key, value any) bool {
```
lhy1024, line 380: ci failed

pkg/syncer/server.go
```diff
@@ -129,38 +143,25 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
...
+					break loop
 				}
-				buckets = append(buckets, bucket)
-				leaders = append(leaders, region.GetLeader())
-				s.history.record(region)
 			}
 			regions := &pdpb.SyncRegionResponse{
 				Header:        &pdpb.ResponseHeader{ClusterId: keypath.ClusterID()},
-				Regions:       requests,
+				Regions:       slices.Clone(requests),
```
lhy1024, line 160: why add clone

lhy1024 (commented):
The rest LGTM

pkg/syncer/server.go
okJiang: This is the code I added when I tried to call broadcast asynchronously, but I ultimately did not use the asynchronous broadcast. I will revert them.

server/cluster/cluster.go
```diff
@@ -1291,10 +1304,13 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
...
+				c.changedRegions <- region
```
nolouch, line 1311: maybe we need a metric about `len(c.changedRegions)`

nolouch (commented):
rest lgtm

server/cluster/cluster.go
okJiang: Sorry, I don't understand. What is its purpose?

nolouch (approved):

PR #9134, opened 2025-03-11
gc: Implement compatibility methods and `GetAllKeyspacesGCStates` method to `GCStateManager`

MyonKeminta:
### What problem does this PR solve?

Issue Number: Ref #8978 

### What is changed and how does it work?

```commit-message
Implement compatibility methods and `GetAllKeyspacesGCStates` method to `GCStateManager`, which is planned to be the replacement of `SafePointManager`.
```

Blocked by:

* ~~https://github.com/tikv/pd/pull/9109~~
* https://github.com/tikv/pd/pull/9169

~~This PR is branched from #9109 which is not merged yet. If you want to view the diff in this PR without being disturbed by changes from #9109, please see: https://github.com/MyonKeminta/pd/compare/m/gc-state-storage...MyonKeminta:pd:m/gc-state-manager~~

Part of this PR is split out to https://github.com/tikv/pd/pull/9169 . Please merge #9169 first. To see changes in this PR based on #9169, see: https://github.com/MyonKeminta/pd/compare/m/gc-state-manager-partial...MyonKeminta:pd:m/gc-state-manager

### Check List

Tests

- Unit test

Code changes

Side effects

- Possible performance regression
- Increased code complexity

Related changes

### Release note

```release-note
None.
```

PR #9169, opened 2025-03-24
gc: Add `GCStateManager` which is planned to be the replacement of `SafePointManager`

MyonKeminta:

Issue Number: Ref #8978

```commit-message
Add `GCStateManager` which is planned to be the replacement of `SafePointManager`.

This is part of the refactor task #8978 . The new module is not yet used until the next PR updates the gRPC service interfaces.
```

This PR is split out from #9134 to reduce the single PR size a little.

### Check List

Tests

- Unit test

Code changes

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+		return constant.NullKeyspaceID, nil
+	}
+
+	keyspaceMeta, err := m.keyspaceManager.LoadKeyspaceByID(keyspaceID)
+	if err != nil {
+		return 0, err
+	}
+	if keyspaceMeta.Config[keyspace.GCManagementType] != keyspace.KeyspaceLevelGC {
+		if isUserAPI {
+			// The user API is expected to always work. Operate on the state of global GC instead.
```
ystaticy, line 133: it should be `unified GC` ?

pkg/gc/gc_state_manager_test.go
```diff
@@ -0,0 +1,824 @@
...
+	err = kgm.Bootstrap(context.Background())
+	re.NoError(err)
+	err = keyspaceManager.Bootstrap()
+	re.NoError(err)
+
+	// keyspaceID 0 exists automatically after bootstrapping.
+
+	ks1, err := keyspaceManager.CreateKeyspace(&keyspace.CreateKeyspaceRequest{
+		Name:       "ks1",
+		Config:     map[string]string{"gc_management_type": "global_gc"},
```
ystaticy, line 105: it should be `unified_gc`  ?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+			return constant.NullKeyspaceID, nil
+		}
+		// Internal API should never be called on keyspaces without keyspace level GC. They won't perform any active
+		// GC operation and will be managed by the global GC.
```
ystaticy, line 137: Ditto.

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+	// Note that in old versions, there wasn't the concept of txn safe point. The step to update the service safe
+	// point of "gc_worker" is somewhat just like the current procedure of advancing the txn safe point, the most
+	// important purpose of which is to find the actual GC safe point that's safe to use.
+	downgradeCompatibleMode := false
+
+	var oldTxnSafePoint uint64
+	newTxnSafePoint := target
+	minBlocker := target
+	var blockingBarrier *endpoint.GCBarrier
+	var blockingMinStartTSOwner *string
```
JmPotato, line 245:
```suggestion
	var (
		minBlocker              = target
		newTxnSafePoint         = target
		oldTxnSafePoint         uint64
		blockingBarrier         *endpoint.GCBarrier
		blockingMinStartTSOwner *string
	)
```

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+			return err1
+		}
+
+		for _, barrier := range barriers {
+			if keyspaceID == constant.NullKeyspaceID && barrier.BarrierID == keypath.GCWorkerServiceSafePointID {
+				downgradeCompatibleMode = true
+				continue
+			}
+
+			if barrier.IsExpired(now) {
```
JmPotato, line 269: I noticed that this comparison occurs during/after the etcd transcation. If certain disk failures cause the read/write operation to take an extremely long time, would this affect the judgment of the barrier time?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+	if err != nil {
+		return GCState{}, err
+	}
+
+	return result, nil
+}
+
+// saturatingDuration returns a duration calculated by multiplying the given `ratio` and `base`, truncated within the
+// range [0, math.MaxInt64] to avoid negative value and overflowing.
+func saturatingDuration(ratio int64, base time.Duration) time.Duration {
```
okJiang, line 536: You don't use this function in this pr. How about removing it now until you use it?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+func NewGCStateManager(store endpoint.GCStateProvider, cfg config.PDServerConfig, keyspaceManager *keyspace.Manager) *GCStateManager {
+	return &GCStateManager{gcMetaStorage: store, cfg: cfg, keyspaceManager: keyspaceManager}
+}
+
+// redirectKeyspace checks the given keyspaceID, and returns the actual keyspaceID to operate on.
+func (m *GCStateManager) redirectKeyspace(keyspaceID uint32, isUserAPI bool) (uint32, error) {
+	// Regard it as NullKeyspaceID if the given one is invalid (exceeds the valid range of keyspace id), no matter
+	// whether it exactly matches the NullKeyspaceID.
+	if keyspaceID & ^constant.ValidKeyspaceIDMask != 0 {
+		return constant.NullKeyspaceID, nil
```
okJiang, line 124: How about adding a warning log here?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+		return
+	}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	return m.advanceGCSafePointImpl(keyspaceID, target, false)
+}
+
+func (m *GCStateManager) advanceGCSafePointImpl(keyspaceID uint32, target uint64, compatible bool) (oldGCSafePoint uint64, newGCSafePoint uint64, err error) {
```
okJiang, line 162: Where is compatible=true?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+	//
+	// In old versions, every time TiDB performs GC, it updates the service safe point of "gc_worker" new txn safe
+	// point.
+	// Note that in old versions, there wasn't the concept of txn safe point. The step to update the service safe
+	// point of "gc_worker" is somewhat just like the current procedure of advancing the txn safe point, the most
+	// important purpose of which is to find the actual GC safe point that's safe to use.
+	downgradeCompatibleMode := false
+
+	var oldTxnSafePoint uint64
+	newTxnSafePoint := target
```
okJiang, line 242: It seems that this assignment is meaningless

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+		}
+
+		// Txn safe point never decreases.
+		newTxnSafePoint = max(oldTxnSafePoint, minBlocker)
+
+		if downgradeCompatibleMode {
+			err1 = wb.SetGCBarrier(keyspaceID, endpoint.NewGCBarrier(keypath.GCWorkerServiceSafePointID, newTxnSafePoint, nil))
+			if err1 != nil {
+				return err1
+			}
```
okJiang, line 312: I am not sure. Do we need reset `downgradeCompatibleMode=false` here?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+		if blockingBarrier == nil && blockingMinStartTSOwner == nil {
+			panic("unreachable")
+		}
+		if newTxnSafePoint == minBlocker {
+			log.Info("txn safe point advancement is being blocked",
+				zap.Uint64("oldTxnSafePoint", oldTxnSafePoint), zap.Uint64("target", target),
+				zap.Uint64("newTxnSafePoint", newTxnSafePoint), zap.String("blocker", blockerDesc),
+				zap.Bool("downgradeCompatibleMode", downgradeCompatibleMode))
+		} else {
+			log.Info("txn safe point advancement unable to be blocked by the minimum blocker",
```
okJiang, line 337: If I understand you correctly, when newTxnSafePoint != minBlocker, newTxnSafePoint = oldTxnSafePoint. So is this better?

```suggestion
			log.Info("txn safe point can't be advanced due to the minimum blocker is less than old txn safe point",
```

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+				zap.Uint64("oldTxnSafePoint", oldTxnSafePoint), zap.Uint64("target", target),
+				zap.Uint64("newTxnSafePoint", newTxnSafePoint), zap.String("blocker", blockerDesc),
+				zap.Uint64("minBlockerTS", minBlocker), zap.Bool("downgradeCompatibleMode", downgradeCompatibleMode))
+		}
+	} else if newTxnSafePoint > oldTxnSafePoint {
```
okJiang, line 342: Do we need the `else`?
```suggestion
	}
    if newTxnSafePoint > oldTxnSafePoint {
```

pkg/gc/gc_state_manager.go
okJiang: Is this same as `txn safe point is remaining unchanged`?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,567 @@
...
+	} else if blockingMinStartTSOwner != nil {
+		blockerDesc = fmt.Sprintf("TiDBMinStartTS { Key: %+q, MinStartTS: %d }", *blockingMinStartTSOwner, newTxnSafePoint)
+	}
+
+	if newTxnSafePoint != target {
+		if blockingBarrier == nil && blockingMinStartTSOwner == nil {
+			panic("unreachable")
+		}
+		if newTxnSafePoint == minBlocker {
+			log.Info("txn safe point advancement is being blocked",
```
okJiang, line 332: Why we not print log once `minBlocker` update? Such a log is more accurate and clear

pkg/gc/gc_state_manager_test.go
MyonKeminta: This should be updated when the renaming is done. As I see, the constant definitions in pkg/keyspace/keyspace.go is still not updated.
I'll add TODO comment here.

pkg/gc/gc_state_manager.go
MyonKeminta: I didn't get it. Can you explain the bad case you are concerning about in detail?
In my consideration, it's safe that the actual expiration time is a little longer than the specified expiration time, and guarantee a GC barrier to be expired once it's regarded expired (delete immediately in the currrent transaction if `IsExpired` returns true). It should not affect the correctness or safety when a transaction runs too long.

pkg/gc/gc_state_manager.go
MyonKeminta: Oh..
I will move it to the next PR where it's used

pkg/gc/gc_state_manager.go
MyonKeminta: Not yet included in this PR.
See: https://github.com/tikv/pd/pull/9134

pkg/gc/gc_state_manager.go
MyonKeminta: For keyspaces where the keyspace-level GC is not enabled, this might constantly happen and should not be treated as a warning.

pkg/gc/gc_state_manager.go
MyonKeminta: Looks not necessary as the transaction will exit soon. And the variable is also needed in the logs later.

pkg/gc/gc_state_manager.go
MyonKeminta: In this log I emphasized the fact that there is something that's expected to block the txn safe point at a smaller value, but the txn safe point has exceeded it already. In this case, the major risk is that some blocker isn't effective, instead of txn safe point not  being advanced.

pkg/gc/gc_state_manager.go
MyonKeminta: Yes

pkg/gc/gc_state_manager.go
MyonKeminta: It's possible that `minBlocker` is updated multiple times when iterating over all potential blockers, producing too much log. We only care about the final result here.

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+}
+
+// getGCStateInTransaction gets all properties in GC states within a context of gcMetaStorage.RunInGCStateTransaction.
+// This read only and won't write anything to the GCStateWriteBatch. It still receives a write batch to ensure
+// it's running in a transactional context.
+func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.GCStateWriteBatch) (GCState, error) {
+	result := GCState{
+		KeyspaceID: keyspaceID,
+	}
+	if keyspaceID != constant.NullKeyspaceID {
```
ystaticy, line 480: "In this code, why do we set result.IsKeyspaceLevel = true when checking if keyspaceID != constant.NullKeyspaceID?"

MyonKeminta: If it's not keyspace level GC, the call to this function must have been redirected to the null keyspace. I'll add comments to explain this later.

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+//
+//  4. (weakened) For each GC barrier `b`, each advancement of the txn safe point should not push it to a new value
+//     that is larger than `b.BarrierTS` (`t' <= max{t, min(b.BarrierTS for b in B)}` where `B` is the set of GC
+//     barriers).
+//
+// TODO: Explicitly state the versions that GCStateManager starts to be functional and the old APIs/concepts/terms is
+//       deprecated when these work are all done.
+
+// GCStateManager is the manager for all kinds of states of TiKV's GC for MVCC data.
+// nolint:revive
```
rleungx, line 105: Why not remove it?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+type GCStateManager struct {
+	lock            syncutil.RWMutex
```
rleungx, line 107:
```suggestion
        syncutil.RWMutex
```

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+
+	return
+}
+
+// AdvanceTxnSafePoint tries to advance the txn safe point to the given target.
+//
+// Returns a struct AdvanceTxnSafePointResult, which contains the old txn safe point, the target, and the new
+// txn safe point it finally made it to advance to. If there's something blocking the txn safe point from being
+// advanced to the given target, it may finally be advanced to a smaller value or remains the previous value, in which
+// case the BlockerDescription field of the AdvanceTxnSafePointResult will be set to a non-empty string describing
```
rleungx, line 208: Do blocker and barrier have the same meaning? If so, how about unifying them?

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	var result GCState
+	err = m.gcMetaStorage.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
+		var err1 error
+		result, err1 = m.getGCStateInTransaction(keyspaceID, wb)
+		return err1
+	})
+
+	if err != nil {
```
rleungx, line 528: return result, err

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+type AdvanceTxnSafePointResult struct {
+	OldTxnSafePoint    uint64
+	Target             uint64
+	NewTxnSafePoint    uint64
+	BlockerDescription string
+}
+
+// GCState represents the GC state of a keyspace, and additionally its keyspaceID and whether the keyspace-level GC is
+// enabled in this keyspace.
+// nolint:revive
```
rleungx, line 545: How about removing it?

pkg/gc/gc_state_manager.go
MyonKeminta: The linter suggests removing the `GC` prefix which is repeative with the packagename `gc`, but the name `StateManager` looks making nonsense to me.

pkg/gc/gc_state_manager.go
MyonKeminta: Is this the recommended style? It makes the `Lock` and `Unlock` methods exported and visible in external packages.

pkg/gc/gc_state_manager.go
MyonKeminta: Apart from the GC barriers, The *blocker* can also be a *TiDB min start TS* from an earlier version of TiDB that didn't follow up the refactoring. It need to be considered for compatibility concerns.

pkg/gc/gc_state_manager.go
rleungx: It makes sense. How about changing it to mu so that we don't need to write lock.Lock.

pkg/gc/gc_state_manager.go
```diff
@@ -0,0 +1,552 @@
...
+		blockerDesc = fmt.Sprintf("TiDBMinStartTS { Key: %+q, MinStartTS: %d }", *blockingMinStartTSOwner, newTxnSafePoint)
+	}
+
+	if newTxnSafePoint != target {
+		if blockingBarrier == nil && blockingMinStartTSOwner == nil {
+			panic("unreachable")
+		}
+		if newTxnSafePoint == minBlocker {
+			log.Info("txn safe point advancement is being blocked",
+				zap.Uint64("oldTxnSafePoint", oldTxnSafePoint), zap.Uint64("target", target),
```
rleungx, line 334:
```suggestion
				zap.Uint64("old-txn-safe-point", oldTxnSafePoint), zap.Uint64("target", target),
```

ystaticy (approved):

JmPotato (approved):

PR #9171, opened 2025-03-26
tso: enhance timestamp persistency with strong leader consistency

### LINKED ISSUE #9157: TSO over-generated problem in an extreme case
## Bug Report

Before #8123, the PD client did not proactively update the TSO stream in response to events such as a PD leader switch. This typically did not pose a problem. However, due to the bug found in #9051, there is an extreme case where a stale PD leader may continue to advance and generate excessive TSO requests. This could result in the new PD leader generating a timestamp that falls back, potentially compromising the consistency of the cluster.

Considering that the current client version has been refactored a lot since #8123, the cost of picking related client changes is relatively high. Therefore, we are considering addressing this issue on the server side. By verifying leadership before saving the timestamp to etcd, we can establish a robust safeguard to prevent this problem from occurring as the excess TSO will not exceed the next start timestamp loaded by the new PD leader.

JmPotato:

Issue Number: close #9157.

```commit-message
This commit enhances the timestamp storage mechanism by adding a leadership validation check when saving TSO timestamps to etcd.
This prevents a former leader from incorrectly persisting timestamps after losing leadership.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix a bug that could cause the former PD leader to incorrectly persist timestamps after losing its leadership.
```

### DIFF
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/election"
+	"github.com/tikv/pd/pkg/errs"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/keypath"
 	"github.com/tikv/pd/pkg/utils/typeutil"
@@ -77,7 +78,7 @@ func (se *StorageEndpoint) SaveTimestamp(groupID uint32, ts time.Time, leadershi
 	log.Info("saving timestamp to the storage", logFilds...)
 	// Fast path to return error if the leadership has not been granted yet.
 	if len(leadership.GetLeaderValue()) == 0 {
-		return errors.New("leadership has not been granted yet, leader value is empty")
+		return errors.Errorf("%s due to leadership has not been granted yet", errs.NotLeaderErr)
 	}
 	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
 		// Ensure the current server is leader by reading and comparing the leader value.
@@ -87,7 +88,7 @@ func (se *StorageEndpoint) SaveTimestamp(groupID uint32, ts time.Time, leadershi
 		}
 		if expected := leadership.GetLeaderValue(); leaderValue != expected {
 			log.Error("leader value does not match", append(logFilds, zap.String("current-leader-value", leaderValue))...)
-			return errors.Errorf("leader value does not match, current: %s, expected: %s", leaderValue, expected)
+			return errors.Errorf("%s due to leader value does not match, current: %s, expected: %s", errs.NotLeaderErr, leaderValue, expected)
 		}
 
 		value, err := txn.Load(keypath.TimestampPath(groupID))
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -76,7 +76,8 @@ func (se *StorageEndpoint) SaveTimestamp(groupID uint32, ts time.Time, leadershi
 		zap.String("expected-leader-value", leadership.GetLeaderValue()),
 	}
 	log.Info("saving timestamp to the storage", logFilds...)
-	// Fast path to return error if the leadership has not been granted yet.
+	// The PD leadership or TSO primary will always be granted first before the TSO timestamp window is saved.
+	// So we here check whether the leader value is filled to see if the requirement is met.
 	if len(leadership.GetLeaderValue()) == 0 {
 		return errors.Errorf("%s due to leadership has not been granted yet", errs.NotLeaderErr)
 	}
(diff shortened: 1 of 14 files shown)


pkg/storage/endpoint/tso.go
```diff
@@ -70,9 +72,31 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+// SaveTimestamp saves the timestamp to the storage. The leadership is used to check if the current server is leader
+// before saving the timestamp to ensure a strong consistency for persistence of the TSO timestamp window.
+func (se *StorageEndpoint) SaveTimestamp(groupID uint32, ts time.Time, leadership *election.Leadership) error {
+	logFilds := []zap.Field{
+		zap.Uint32("group-id", groupID),
+		zap.Time("ts", ts),
+		zap.String("leader-key", leadership.GetLeaderKey()),
+		zap.String("expected-leader-value", leadership.GetLeaderValue()),
+	}
+	log.Info("saving timestamp to the storage", logFilds...)
```
okJiang, line 84: I am not sure the frequency

pkg/storage/endpoint/tso.go
```diff
@@ -70,9 +72,31 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+	// Fast path to return error if the leadership has not been granted yet.
+	if len(leadership.GetLeaderValue()) == 0 {
+		return errors.Errorf("%s due to leadership has not been granted yet", errs.NotLeaderErr)
```
okJiang, line 87: Do we need to add a test case for this one?

pkg/storage/endpoint/tso.go
```diff
@@ -70,9 +72,31 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+	}
 	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
+		// Ensure the current server is leader by reading and comparing the leader value.
+		leaderValue, err := txn.Load(leadership.GetLeaderKey())
+		if err != nil {
+			return err
+		}
+		if expected := leadership.GetLeaderValue(); leaderValue != expected {
+			log.Error("leader value does not match", append(logFilds, zap.String("current-leader-value", leaderValue))...)
+			return errors.Errorf("%s due to leader value does not match, current: %s, expected: %s", errs.NotLeaderErr, leaderValue, expected)
```
okJiang, line 97: ditto

pkg/storage/endpoint/tso.go
JmPotato: By default it's `tso-save-interval`, which is 3 seconds.

pkg/storage/endpoint/tso.go
JmPotato: Already added in `storage_tso_test.go`.

okJiang (approved):

pkg/storage/endpoint/tso.go
```diff
@@ -70,9 +72,31 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+	if len(leadership.GetLeaderValue()) == 0 {
+		return errors.Errorf("%s due to leadership has not been granted yet", errs.NotLeaderErr)
+	}
 	return se.RunInTxn(context.Background(), func(txn kv.Txn) error {
+		// Ensure the current server is leader by reading and comparing the leader value.
+		leaderValue, err := txn.Load(leadership.GetLeaderKey())
+		if err != nil {
+			return err
+		}
+		if expected := leadership.GetLeaderValue(); leaderValue != expected {
```
nolouch, line 95: An additional check is needed here. Why can’t we confirm if the leadership matches the etcd persisted data or is empty?

JmPotato: We verified whether `leadership.GetLeaderValue()` is empty at line 86. This check is necessary because all TSO persistence operations must occur only after the PD leadership or TSO primary has been successfully retrieved.

nolouch (approved):

pkg/storage/endpoint/tso.go
```diff
@@ -70,9 +72,32 @@ func (se *StorageEndpoint) LoadTimestamp(prefix string) (time.Time, error) {
...
+	logFilds := []zap.Field{
+		zap.Uint32("group-id", groupID),
+		zap.Time("ts", ts),
+		zap.String("leader-key", leadership.GetLeaderKey()),
+		zap.String("expected-leader-value", leadership.GetLeaderValue()),
+	}
+	log.Info("saving timestamp to the storage", logFilds...)
+	// The PD leadership or TSO primary will always be granted first before the TSO timestamp window is saved.
+	// So we here check whether the leader value is filled to see if the requirement is met.
+	if len(leadership.GetLeaderValue()) == 0 {
```
rleungx, line 87: Do we need to reset the leaderValue once the leader steps down?

rleungx (approved):

PR #9173, opened 2025-04-01
config: increase the default lease

### LINKED ISSUE #9156: Increase the default lease configuration
## Enhancement Task
It isn't recommended to create a lease with short TTL (e.g. < 5s), because obviously it may be affected by either server-side or client-side delay.
We should change the default value of the lease.

rleungx:

Issue Number: Close #9156.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -172,7 +172,7 @@ func NewConfig() *Config {
 }
 
 const (
-	defaultLeaderLease             = int64(3)
+	defaultLeaderLease             = int64(5)
 	defaultCompactionMode          = "periodic"
 	defaultAutoCompactionRetention = "1h"
 	defaultQuotaBackendBytes       = typeutil.ByteSize(8 * units.GiB) // 8GB


nolouch:
Could you add more comments about the reason?

lhy1024:
> Could you add more comments about the reason?

It seems to be from the discussion of etcd community

rleungx:
> Could you add more comments about the reason?

Add a description, PTAL.

bufferflies:
I read the [ahrtr](https://github.com/ahrtr) answer about the TTL, he suggests the TTL should be less than 5s.
```
You need to analyze the trace of the write/put operation, to figure out which step consumed most of the time.
disk I/O and network connectivity are important to etcd, this looks likes an environment issue which you need to resolve.
As mentioned previously, it isn't recommended to create lease with short TTL (e.g. < 5s), because obviously it may be affected by either server side or client side delay.
```

JmPotato:
> As mentioned previously, it isn't recommended to create lease with short TTL (e.g. < 5s)

I think what he means here is better setting TTL greater than or equal to 5 seconds.

niubell (approved):

Issue #6067, opened 2023-03-01
Clean up the archived keyspace


zeminzhou:
## Development Task

When the state of a keyspace is changed to the archived state, the data corresponding to this keyspace needs to be cleaned up.

- [ ] Periodically retrieve archived keyspaces.
- [ ] A tikv client is used to clean up the data corresponding to the keyspace on tikv.
- [ ] Clean up the data corresponding to the keyspace on tikv.
- [ ] Clean up the PlacementRule corresponding to the keyspace.
- [ ] Clean up the RegionLabelRule corresponding to the keyspace.
- [ ] Clean up the ETCD Path corresponding to the keyspace.
- [ ] Change keyspace state to TOMBSTONE.

lhy1024:
After consulting with @AmoebaProtozoa, there is now a specialized tool for serverless cleanup, so we are not moving forward for now.

PR #9182, opened 2025-04-07
[WIP] Keyspace cleaning

AmoebaProtozoa:

Issue Number: Ref #6067 

This PR added the basic keyspace cleaning framework and some configurable options to keyspace module.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/gc.go
+++ b/pkg/keyspace/gc.go
@@ -1,0 +1,189 @@
+// Copyright 2023 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


pkg/keyspace/gc.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2023 TiKV Project Authors.
```
ystaticy, line 1: 2025

pkg/keyspace/gc.go
```diff
@@ -0,0 +1,189 @@
...
+	manager *Manager
+	member  *member.EmbeddedEtcdMember
+
+	nextKeyspaceID uint32
+	ticker         *time.Ticker
+	// gcLifeTime specifies how long should we keep archived keyspace.
+	gcLifeTime time.Duration
+}
+
+// newGCWorker returns a newGCWorker.
```
ystaticy, line 50: Introduce the concept of 'archiving' instead of calling it GC to prevent other developers from confusing this GC with MVCC GC

ystaticy: Additionally, do not use the concept of 'GC' when archiving a keyspace, and the naming of safepoint variables should include a prefix to indicate what operation the safepoint corresponds to.

AmoebaProtozoa: ok, sounds good, will use `cleaning` & `archiving` instead of `gc`

PR #9183, opened 2025-04-08
pkg: return error if regions not found

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


xhebox (commented):
I remembered that I want a default `REPLICATED` state.

Did we modify the logic that we need an error for wrong region ranges?

rleungx:
> I remembered that I want a default `REPLICATED` state.
> 
> Did we modify the logic that we need an error for wrong region ranges?

What is the purpose of that if the given range is wrong?

PR #9196, opened 2025-04-14
feat(router): allow the follower to handle QueryRegion

JmPotato:

Issue Number: ref #8690.

```commit-message
Implement Follower Handle functionality for QueryRegion.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/client/clients/router/client.go
+++ b/client/clients/router/client.go
@@ -547,10 +547,8 @@ batchLoop:
 		// Step 3: Dispatch the router requests to the stream connection.
 		// TODO: timeout handling if the stream takes too long to process the requests.
 		err = c.processRequests(stream)
-		if err != nil {
-			if !c.handleProcessRequestError(ctx, streamURL, err) {
-				return
-			}
+		if err != nil && !c.handleProcessRequestError(ctx, streamURL, err) {
+			return
 		}
 	}
 }
--- a/client/clients/router/client.go
+++ b/client/clients/router/client.go
@@ -519,6 +519,7 @@ batchLoop:
 			allowFollowerHandle := c.option.GetEnableFollowerHandle()
 			if allowFollowerHandle {
 				// We need to ensure all requests in a same batch allow to be handled by the follower.
+				// IMPROVE: separate into the follower and leader handle batches.
 				c.batchController.IterCollectedRequests(func(req *Request) bool {
 					if !req.options.AllowFollowerHandle {
 						allowFollowerHandle = false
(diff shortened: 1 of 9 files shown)


client/clients/router/client.go
```diff
@@ -367,22 +397,53 @@ func (c *Cli) updateConnection(ctx context.Context) {
 	cc, url := c.getLeaderClientConn()
 	if cc == nil || len(url) == 0 {
 		log.Warn("[router] got an invalid leader client connection", zap.String("url", url))
-		return
 	}
 	if c.conCtxMgr.Exist(url) {
```
rleungx, line 401: If the url is invalid, do we still need to check Exist?

client/clients/router/client.go
```diff
@@ -367,22 +397,53 @@ func (c *Cli) updateConnection(ctx context.Context) {
...
+			}
+			// Store the stream connection context if it is successfully created.
+			if stream != nil {
+				c.conCtxMgr.Store(ctx, url, stream)
+				log.Info("[router] successfully established the router stream connection", zap.String("url", url))
+			}
+		}
+	} else {
+		// GC all the follower router stream connections.
+		c.conCtxMgr.GC(func(url string) bool {
```
rleungx, line 438: Do we need the handle the scale in/out cases?

client/clients/router/client.go
```diff
@@ -446,8 +507,25 @@ batchLoop:
 				continue batchLoop
 			default:
 			}
+			// Check whether allow the follower to handle this batch of requests.
+			allowFollowerHandle := c.option.GetEnableFollowerHandle()
+			if allowFollowerHandle {
+				// We need to ensure all requests in a same batch allow to be handled by the follower.
+				c.batchController.IterCollectedRequests(func(req *Request) bool {
+					if !req.options.AllowFollowerHandle {
```
rleungx, line 515: What if the AllowFollowerHandle changed when IterCollectedRequests?

client/clients/router/client.go
JmPotato: Refined this check.

client/clients/router/client.go
JmPotato: Nice catch. Added.

client/clients/router/client.go
JmPotato: I added one last check before choosing the stream, PTAL.

rleungx (approved):

tests/integrations/client/router_client_test.go
okJiang, line 1: The client in this test does not SetEnableFollowerHandle.

okJiang (commented):
rest lgtm

okJiang (approved):

PR #9199, opened 2025-04-15
server: remove logic that depends on join file

### LINKED ISSUE #9190: "Etcd cluster ID mismatch" upon PD restart if first peer moved out of the cluster
## Bug Report

A PD that joined the cluster using the `join` mode will have a local `join` file in the data directory. In the future every restart for this PD will use the addresses in the join file as the initial cluster, and query them one by one to verify that the remote cluster ID matches what's in the current PD's snap file: 

https://github.com/tikv/pd/blob/master/server/join/join.go#L96

https://github.com/tikv/pd/blob/aa28f66d6f8e583d9ad5e16046b34d77d83aac9e/pkg/utils/etcdutil/etcdutil.go#L89

If the first PD in this join list got removed from the cluster and joined another cluster, the current (joined) PD's future restart will all fail because the first PD to verify the cluster ID now belongs to another cluster with a different ID.

### What did you do?

1. Create two cluster configs for PD:
```
# pd-cluster1.toml
initial-cluster-token = "cluster1"

# pd-cluster2.toml
initial-cluster-token = "cluster2"
```
2. Start a playground with 3 PDs using the cluster 1 config:
```
➜  ~ tiup playground --pd 3 --pd.config ./pd-cluster1.toml
```
3. Create a join PD:
```
➜  ~ tiup playground scale-out --pd 1
```
4. Delete the first PD from the cluster, clear data directory and restart it using cluster 2's config as a one-node cluster:
```
➜  ~ curl localhost:2382/pd/api/v1/members/id/4319256102737695738 -X DELETE
"removed, pd: 4319256102737695738"
➜  ~ rm -rf /Users/yzhan/.tiup/data/Ui6ZDT5/pd-0/*
➜  ~ /Users/yzhan/.tiup/components/pd/v8.5.1/pd-server --name=pd-0 --data-dir=/Users/yzhan/.tiup/data/Ui6ZDT5/pd-0/data --peer-urls=http://127.0.0.1:2380 --advertise-peer-urls=http://127.0.0.1:2380 --client-urls=http://127.0.0.1:2379 --advertise-client-urls=http://127.0.0.1:2379 --log-file=/Users/yzhan/.tiup/data/Ui6ZDT5/pd-0/pd.log --config=/Users/yzhan/pd-cluster2.toml --initial-cluster=pd-0=http://127.0.0.1:2380
```
5. Kill and restart the join PD
```
➜  ~ /Users/yzhan/.tiup/components/pd/v8.5.1/pd-server --name=pd-3 --data-dir=/Users/yzhan/.tiup/data/Ui6ZDT5/pd-3/data --peer-urls=http://127.0.0.1:59274 --advertise-peer-urls=http://127.0.0.1:59274 --client-urls=http://127.0.0.1:59275 --advertise-client-urls=http://127.0.0.1:59275 --log-file=/Users/yzhan/.tiup/data/Ui6ZDT5/pd-3/pd.log --config=/Users/yzhan/pd-cluster1.toml --join=http://127.0.0.1:2380,http://127.0.0.1:2381,http://127.0.0.1:2383
```
Join PD restarts successfully.
### What did you see instead?
Join PD fails to start due to cluster ID mismatch because pd-0 now belongs to another cluster:
```
[2025/04/11 16:29:30.853 -07:00] [FATAL] [main.go:289] ["run server failed"] [error="Etcd cluster ID mismatch, expect 7882946083602988861, got 4506588655196452383"] [stack="main.start\n\t/Users/pingcap/workspace/bp-pd-release-darwin-arm64-4l5sm-build-binaries/source/pd/cmd/pd-server/main.go:289\nmain.createServerWrapper\n\t/Users/pingcap/workspace/bp-pd-release-darwin-arm64-4l5sm-build-binaries/source/pd/cmd/pd-server/main.go:194\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/pingcap/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:987\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/pingcap/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/pingcap/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039\nmain.main\n\t/Users/pingcap/workspace/bp-pd-release-darwin-arm64-4l5sm-build-binaries/source/pd/cmd/pd-server/main.go:71\nruntime.main\n\t/usr/local/go1.23/src/runtime/proc.go:272"]
```
v8.5.1

yzhan1:
Deletes logic that creates and reads from the join file to prevent start-up issues when the first peer in the join list is removed from the cluster.

Issue Number: Close #9190 

PDs joining an existing cluster will not create a local join file that persists the point-in-time cluster members and will not read from that join file during restarts.

The join file is not needed anyways since later restarts can just read the existing topology from the local etcd snap file.

The file was used to restart a PD that joined successfully but failed the first start. This PR addresses that issue by detecting the failed to start member using the `MemberList` response and add it to the initial cluster parameter.

```commit-message
server: remove logic that depends on join file
```

### Check List

Tests

- Unit test
- Integration test
  - Tested manually following steps in the issue
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


rleungx:
The join file seems to resolve the operator issues: https://github.com/pingcap/tidb-operator/issues/126. I'm not sure if it is still a problem right now.

yzhan1:
@rleungx @nolouch I read that issue and PR but it's not very clear to me how we got into that error state where the new PD seems half-joined the cluster without a name. Also unclear on how using the join file can help with that situation.

If somehow we got into that state and PD cannot join later, my understanding is that it will:
1. Not have a member directory
2. Be restarted using similar join arguments
which should allow it to be restarted and joined successfully.

Please correct me if I'm wrong.

If we know that we need the join file for sure, then maybe the other way to fix this is to have PD delete the join file once it joins successfully? Since after that it should just read the member directory for data.

rleungx:
Consider such a case:
1. We scale out a new PD member, assume pd-3.
2. `PrepareJoinCluster` succeeds, however, the server isn't started due to some error.
3. At that time, there is no member file.
4. Then we restart that PD and call `MemberList` API to generate the `InitialCluster`, and we get the pd-3 with the empty name according to the comment: "**name is the human-readable name of the member. If the member is not started, the name will be an empty string.**"
5. Finally, the error happens.

> If we know that we need the join file for sure, then maybe the other way to fix this is to have PD delete the join file once it joins successfully? Since after that it should just read the member directory for data.

After I discussed with @nolouch, we think it's safe to delete the file after it joined successfully.

yzhan1:
@rleungx Actually even if we delete the join file after it joined successfully, it still will happen for this case:
1. PD B tries to join but failed half-way
2. PD A moved out of the cluster
3. PD B now can start but will hit the wrong cluster ID exception since it will call PD A to verify the cluster ID

yzhan1:
I think a better way to handle both cases would be:
1. We have a cluster with `pd-1`, `pd-2`, `pd-3`
2. `pd-4` tries to join the cluster with `--join=pd-1,pd-2,pd-3 --peer-url=pd-4`, but fails
3. Now we have a PD member with no name but it has the peer url `pd-4`
4. Failed PD tries to restart with the same parameteres `--join=pd-1,pd-2,pd-3 --peer-url=pd-4` and join again with no member directory
5. Here https://github.com/tikv/pd/blame/748d6c8321b51a99d787e340dcbb20da629a76fa/server/join/join.go#L139 we loop through the member list and find out there's a no-namer PD with the same peer url `pd-4`
6. We exit the loop and change the config with `--initial-cluster=pd-1,pd-2,pd-3,pd-4 --mode=existing`, exit the prepare join func and skip adding member again

In this way we don't need the join file, which will become stale as time goes, and also we can handle all 3 cases:
1. Successful join without member directory
2. Re-join after member added but start failed
3. Restart a joined node using local member directory

@rleungx WDYT?

rleungx:
LGTM, WDYT? @JmPotato @nolouch

JmPotato:
> 5\. `748d6c8`/server/join/join.go#L139 (blame)

LGTM.

JmPotato:
> LGTM, WDYT? @JmPotato @nolouch

LGTM.

yzhan1:
Thanks for confirming. Will update the PR.

yzhan1:
@rleungx @JmPotato @nolouch PTAL at the PR. Thanks!

rleungx:
Sorry for the late reply. I will take a look tomorrow.

server/join/join.go
```diff
@@ -190,11 +191,16 @@ func PrepareJoinCluster(cfg *config.Config) error {
 				listSucc = true
 			}
 			if len(n) == 0 {
-				log.Error("there is an abnormal joined member in the current member list",
-					zap.Uint64("id", memb.ID),
-					zap.Strings("peer-urls", memb.PeerURLs),
-					zap.Strings("client-urls", memb.ClientURLs))
-				return errors.Errorf("there is a member %d that has not joined successfully", memb.ID)
+				if joinedFailedToStart && isSameSlice(memb.PeerURLs, strings.Split(cfg.AdvertisePeerUrls, ",")) {
```
rleungx, line 194: Do we need to check isSameSlice again?

server/join/join.go
```diff
@@ -218,8 +224,19 @@ func PrepareJoinCluster(cfg *config.Config) error {
 		return errors.WithStack(err)
 	}
 
-	err = os.WriteFile(filePath, []byte(cfg.InitialCluster), privateFileMode)
-	return errors.WithStack(err)
+	return nil
+}
+
+func isSameSlice(a, b []string) bool {
```
rleungx, line 230: How about moving it to pkg/slice and renaming it to slice.EqualWithoutOrder?

yzhan1: Done

server/join/join.go
yzhan1: Probably not, since we alreay know that at this point there can be only one member without a name and it's the node itself. Otherwise the for loop above would have already exited with error.

rleungx: After I rethink it, maybe it's better to use slice.EqualWithoutOrder here to avoid more than one PD with an empty name.

rleungx (commented):
The rest LGTM.

server/join/join.go
```diff
@@ -51,6 +50,8 @@ var listMemberRetryTimes = 20
 //
 //	and it is not a member of cluster, join does MemberAdd, it returns an
 //	error if PD tries to join itself, missing data or join a duplicated PD.
+//	If previously the node joined but failed to start, it will attempt to
+//	start as the same membeer again.
```
JmPotato, line 54:
```suggestion
//	start as the same member again.
```

tests/server/join/join_fail/join_fail_test.go
```diff
@@ -44,3 +46,36 @@ func TestFailedPDJoinInStep1(t *testing.T) {
...
+	})
+	re.NoError(err)
+	re.NoError(pd2.Run())
+	re.NotEmpty(cluster.WaitLeader())
+
+	// Check that the new PD has joined the cluster.
+	members, err := client.MemberList(ctx)
+	re.NoError(err)
+	re.Len(members.Members, 2)
+	re.Equal(cluster.GetServer("pd1").GetClusterID(), pd2.GetClusterID())
```
JmPotato, line 80: How about transferring the leadership to the newly added member to ensure it can serve as well?

JmPotato (commented):
The rest LGTM.

rleungx (approved):

JmPotato (approved):

PR #9201, opened 2025-04-15
mcs: remove unnecessary config

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/resourcemanager/server/config.go
```diff
@@ -71,8 +68,6 @@ type Config struct {
 	ListenAddr          string `toml:"listen-addr" json:"listen-addr"`
 	AdvertiseListenAddr string `toml:"advertise-listen-addr" json:"advertise-listen-addr"`
 	Name                string `toml:"name" json:"name"`
-	DataDir             string `toml:"data-dir" json:"data-dir"` // TODO: remove this after refactoring
```
lhy1024, line 74: Will removing these configurations cause compatibility issues?

tests/testutil.go
```diff
@@ -171,7 +169,6 @@ func NewSchedulingTestServer(ctx context.Context, cfg *sc.Config) (*scheduling.S
 	}
 	cleanup := func() {
 		s.Close()
-		os.RemoveAll(cfg.DataDir)
```
lhy1024, line 174: Does it leave files behind after the test?

pkg/mcs/resourcemanager/server/config.go
rleungx: AFAIK, the deploy tool doesn't use the data-dir in the start script.

okJiang (approved):

JmPotato (approved):

niubell (approved):

Issue #9222, opened 2025-04-21
Pending task may be kept forever when there is no task coming


okJiang:
### What did you do?

Stop load after a period of peak traffic.

### What did you expect to see?

the pending task can reset to zero.

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/a45ae77d-d6b9-4ed1-8d3c-eb44bde36b06" />

### What did you see instead?

the pending task stop reducing

<img width="1196" alt="Image" src="https://github.com/user-attachments/assets/a6052c84-8008-41f7-b883-39dd79733319" />

PR #9226, opened 2025-04-21
runner: set capacity to Task channel

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9222

If the program processPendingTask and handle `<-ticker.C` in the same time

https://github.com/tikv/pd/blob/f86063f6063813ddd6a2ec23d5b645170049b722/pkg/ratelimit/runner.go#L171

https://github.com/tikv/pd/blob/f86063f6063813ddd6a2ec23d5b645170049b722/pkg/ratelimit/runner.go#L133

, the task may be skipped due to taskChan block.

https://github.com/tikv/pd/blob/f86063f6063813ddd6a2ec23d5b645170049b722/pkg/ratelimit/runner.go#L170-L175

So, we can give a capacity to the task channel to avoid this situation.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

image

- No code

### Release note

```release-note
None.
```


pkg/ratelimit/runner_test.go
```diff
@@ -82,7 +82,7 @@ func TestConcurrentRunner(t *testing.T) {
 		for i := 1; i < 11; i++ {
 			regionID := uint64(i)
 			if i == 10 {
-				regionID = 4
+				regionID = 6
```
lhy1024, line 85: why change it？

lhy1024 (approved):

pkg/ratelimit/runner_test.go
okJiang: To pass this ut. Capacity changed the pendingTask index. To ensure the task is not consumed as much as possible, I have increased the region ID a bit more.

nolouch:
I think the task will be handled well next time, even if you have the capacity,  the same problem will still exist.

nolouch (approved):

PR #9227, opened 2025-04-21
syncer: set retained=true to syncRegionRunner

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9154

```commit-message
syncer: set retained=true to syncRegionRunner
```

### Check List

Tests

- Unit test
- Integration test

If the pending task exceeds 1 minute, the task will not fail.

image

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1306,6 +1306,7 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
 			func(context.Context) {
 				c.changedRegions <- region
 			},
+			ratelimit.WithRetained(true),
 		)
 	}
 	return nil


server/cluster/cluster.go
```diff
@@ -1306,6 +1306,7 @@ func (c *RaftCluster) processRegionHeartbeat(ctx *core.MetaProcessContext, regio
 			func(context.Context) {
 				c.changedRegions <- region
 			},
+			ratelimit.WithRetained(true),
```
bufferflies, line 1309: Would you be able to handle this error in the function?

okJiang: What do you mean?

nolouch (approved):

PR #9237, opened 2025-04-23
keyspace: support creating keyspace by id

### LINKED ISSUE #9318: Support creating keyspace with the given ID

rleungx:

Issue Number: Close #9318.

We need to use ID to create a keyspace in the next-gen TiDB. The name is useless, but I keep it the same as the ID for compatibility issues.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/keyspace/keyspace.go
```diff
@@ -280,6 +290,83 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	return keyspace, nil
 }
 
+// CreateKeyspaceByID create a keyspace meta with given config and save it to storage.
+// This function is used by next-gen TiDB.
```
JmPotato, line 294: Better not to refer this directly in our codebase IMO.

pkg/keyspace/keyspace.go
```diff
@@ -280,6 +290,83 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
...
+		log.Warn("[keyspace] failed to save keyspace before split",
+			zap.Uint32("keyspace-id", keyspace.GetId()),
+			zap.String("keyspace-name", keyspace.GetName()),
+			zap.Error(err),
+		)
+		return nil, err
+	}
+	// Split keyspace region.
+	err = manager.splitKeyspaceRegion(id, manager.config.ToWaitRegionSplit())
+	if err != nil {
```
JmPotato, line 335: Better add a warn log here.

rleungx: The error will be returned finally.

pkg/keyspace/keyspace.go
rleungx: ok, removed

AmoebaProtozoa (approved):

JmPotato (approved):

lhy1024 (approved):

pkg/keyspace/keyspace.go
```diff
@@ -280,6 +290,81 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	return keyspace, nil
 }
 
+// CreateKeyspaceByID create a keyspace meta with given config and save it to storage.
+func (manager *Manager) CreateKeyspaceByID(request *CreateKeyspaceByIDRequest) (*keyspacepb.KeyspaceMeta, error) {
+	if request.ID == nil {
```
glorv, line 295: Can you explain the reason of using `*uint32` instead of `uint32` for the ID field?

PR #9275, opened 2025-04-28
scheduler: evict hot leader first

### LINKED ISSUE #8949: Consider evict leader with flow first
## Enhancement Task

When TiKV encounters an IO jitter issue, we have provided a slow store detection mechanism to reduce the impact of latency. Currently, we are using a random method to choose the region that will be transferred. But we can improve it by choosing the region according to its flow.

rleungx:

Issue Number: Close #8949

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


mittalrishabh:
@rleungx we also have few ideas on how to handle hot regions especially in stale reads/follower reads environment. We should collaborate on this feature.

rleungx:
@Tema @mittalrishabh We are going to resolve #8949. This PR is just a draft now and is used for internal testing.

PR #9279, opened 2025-04-28
statistics: use Loads/StoreStats and HistoryLoads instead of []float64 and [][]float64

okJiang:

Issue Number: Ref #4399 

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


bufferflies (approved):

pkg/statistics/store_load.go
```diff
@@ -145,17 +145,28 @@ func (s *StoreSummaryInfo) SetEngineAsTiFlash() {
...
+// Loads is a vector that contains different dimensions of loads.
+type Loads [utils.DimLen]float64
+
+// HistoryLoads is a circular buffer of loads. The reason for using
+// [utils.DimLen][]float64 instead of [][utils.DimLen]float64 is to make it easier
+// for the `checkHistoryLoadsByPriority` function to handle.
+type HistoryLoads [utils.DimLen][] /* buffer size */ float64
+
+// StoreStats is a vector that contains different dimensions of store stat.
+type StoreStats [utils.StoreStatCount]float64
```
rleungx, line 157: I think it's a little bit confusing between this variable and Loads. e.g., `GetStoresLoads` returns `StoreStats`.

okJiang: Do you have any suggestions on this? They had this confusion before. I originally intended to use StoreLoads, but I found that there is also a concept called StoreStatCount.

PR #9293, opened 2025-05-07
server: optimize store topology weight memory

### LINKED ISSUE #9288: Improve the memory usage for `buildTopology`
## Enhancement Task

Image

We need to improve its memory usage, especially when the cluster has a large number of regions and labels.

rleungx:

Issue Number: Close #9288.

```
$ benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/tikv/pd/pkg/core
cpu: Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz
                                        │   old.txt    │              new.txt               │
                                        │    sec/op    │   sec/op     vs base               │
GetStoreTopoWeight/Small-10-Stores-72     16.457µ ± 5%   5.190µ ± 6%  -68.47% (p=0.002 n=6)
GetStoreTopoWeight/Medium-100-Stores-72    84.98µ ± 6%   35.44µ ± 4%  -58.29% (p=0.002 n=6)
GetStoreTopoWeight/Large-1000-Stores-72    475.9µ ± 3%   241.3µ ± 3%  -49.29% (p=0.002 n=6)
geomean                                    87.31µ        35.40µ       -59.45%

                                        │    old.txt     │                  new.txt                  │
                                        │      B/op      │     B/op       vs base                    │
GetStoreTopoWeight/Small-10-Stores-72       5.555Ki ± 0%   0.000Ki ±  0%  -100.00% (p=0.002 n=6)
GetStoreTopoWeight/Medium-100-Stores-72     16.34Ki ± 0%    0.00Ki ±  0%  -100.00% (p=0.002 n=6)
GetStoreTopoWeight/Large-1000-Stores-72   67368.000 ± 0%     3.000 ± 33%  -100.00% (p=0.002 n=6)
geomean                                     18.14Ki                       ?                      ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean

                                        │   old.txt   │                 new.txt                 │
                                        │  allocs/op  │  allocs/op   vs base                    │
GetStoreTopoWeight/Small-10-Stores-72      65.00 ± 0%     0.00 ± 0%  -100.00% (p=0.002 n=6)
GetStoreTopoWeight/Medium-100-Stores-72    435.0 ± 0%      0.0 ± 0%  -100.00% (p=0.002 n=6)
GetStoreTopoWeight/Large-1000-Stores-72   3.140k ± 0%   0.000k ± 0%  -100.00% (p=0.002 n=6)
geomean                                    446.1                     ?                      ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Performance**
  * Improved store topology weight calculation efficiency with reduced memory churn during topology updates.
  * Added object pooling to reduce allocations when handling many labeled stores.

* **Bug Fixes**
  * Improved handling for partially valid location labels.
  * Preserved empty store label values (for example, `zone: ""`).
  * Ensured safe results (including no panics) when all eligible stores are tombstoned or when many distinct label keys are provided.

* **Tests**
  * Added unit tests and benchmarks for topology building and store topology weight behavior.


bufferflies (approved):

pkg/core/store_topology.go
```diff
@@ -40,26 +44,126 @@ func GetStoreTopoWeight(store *StoreInfo, stores []*StoreInfo, locationLabels []
...
 
-func buildTopology(s *StoreInfo, stores []*StoreInfo, locationLabels []string, count int) (map[string]any, []string, float64, bool) {
-	topology := make(map[string]any)
+var (
+	storeLabelPool = sync.Pool{
+		New: func() any {
+			return &metapb.StoreLabel{}
+		},
+	}
+)
```
JmPotato:
```suggestion
var storeLabelPool = sync.Pool{
	New: func() any {
		return &metapb.StoreLabel{}
	},
}
```

pkg/core/store_topology.go
```diff
@@ -15,22 +15,26 @@
...
+
 	"github.com/pingcap/kvproto/pkg/metapb"
 
 	"github.com/tikv/pd/pkg/slice"
 )
 
 // GetStoreTopoWeight calculates the topology weight of a store based on its labels and the labels of other stores.
 func GetStoreTopoWeight(store *StoreInfo, stores []*StoreInfo, locationLabels []string, count int) float64 {
-	topology, validLabels, sameLocationStoreNum, isMatch := buildTopology(store, stores, locationLabels, count)
+	topology := getTopology()
```
JmPotato: Why not put this inside `buildTopology`?

pkg/tso/keyspace_group_manager.go
```diff
@@ -548,6 +548,7 @@ func (kgm *KeyspaceGroupManager) InitializeGroupWatchLoop() error {
 			}
 		})
 		kgm.updateKeyspaceGroup(group)
+		kgm.SetModRevision(uint64(kv.ModRevision))
```
lhy1024: This publishes the transaction revision before the whole watch response has been applied. `LoopWatcher` invokes `putFn` once per event and calls `postEventsFn` only after all events; keyspace-group create/split/merge paths can modify multiple group keys in one etcd transaction. After the first PUT, a concurrent `FindGroupByKeyspaceID` can therefore observe revision R while a later group from the same transaction is still old, and accept that partial state as revision R.

I reproduced this deterministically by committing two group PUTs in one transaction and blocking the second update with the existing `delayBeforeCheckingElectionMember` failpoint:

```go
txnRevision := uint64(txnResp.Header.Revision)
    mgr.RLock()
    defer mgr.RUnlock()
    return mgr.modRevision >= txnRevision && mgr.kgs[secondGroupID] == nil
})
```

Could we keep revision publication in `postEventsFn`, after the complete batch has been applied, and add a regression test for this ordering?

lhy1024: This still cannot establish the correct freshness watermark for an initial load. `LoopWatcher.load` calls `putFn` only for surviving KVs, so their maximum `kv.ModRevision` cannot represent newer revisions carried by deleted keys. A newly started node can load a snapshot taken after a deletion but advertise an older revision and reject a client that already observed the deletion until another group write occurs.

A targeted test reproduces the gap:

```go
group0Resp, _ := client.Put(ctx, group0Path, group0Value)
_, _ = client.Put(ctx, deletedGroupPath, deletedGroupValue)
deleteResp, _ := client.Delete(ctx, deletedGroupPath)
require.NoError(t, mgr.Initialize())
require.Equal(t, uint64(group0Resp.Header.Revision), mgr.modRevision)
require.Less(t, mgr.modRevision, uint64(deleteResp.Header.Revision))
```

The load and watch paths need one consistent snapshot/watermark contract; updating from each surviving KV is not sufficient. Since this is independent of topology pooling, could we remove it from this PR and address it separately?

pkg/core/store_topology.go
```diff
@@ -40,26 +43,130 @@ func GetStoreTopoWeight(store *StoreInfo, stores []*StoreInfo, locationLabels []
...
+
+// putSortedLabels puts the label pairs back to pool.
+func putSortedLabels(pairs *LabelPairs) {
+	for _, label := range pairs.pooledLabels {
+		label.Key = ""
+		label.Value = ""
+		storeLabelPool.Put(label)
+	}
+
+	pairs.pairs = pairs.pairs[:0]
```
lhy1024: Could we clear the pointer slots before returning this object to the pool? Reslicing to zero does not clear the backing array scanned by the GC, so a pooled `LabelPairs` can keep the last real `StoreLabel` objects and their strings reachable, including stale tail entries after the configured label count shrinks.

I verified the current behavior with this focused check:

```go
label := &metapb.StoreLabel{Key: "zone", Value: "z1"}
pairs := &LabelPairs{pairs: []*metapb.StoreLabel{label}}
putSortedLabels(pairs)
require.Same(t, label, pairs.pairs[:cap(pairs.pairs)][0])
```

This retention is bounded and `sync.Pool` may discard it during GC, so I am mainly asking from a memory-hygiene perspective. Would `clear(pairs.pairs[:cap(pairs.pairs)])` plus dropping unusually large buffers be preferable here?

pkg/core/store_topology.go
```diff
@@ -40,26 +43,130 @@ func GetStoreTopoWeight(store *StoreInfo, stores []*StoreInfo, locationLabels []
...
 	return weight / sameLocationStoreNum
 }
 
+var storeLabelPool = sync.Pool{
+	New: func() any {
+		return &metapb.StoreLabel{}
+	},
+}
+
+// LabelPairs is pre-allocated buffer for sorting labels.
```
lhy1024: Could we avoid exporting `LabelPairs` and pooling temporary protobuf labels altogether? `updateTopology` can iterate `locationLabels`, exact-match each key against `store.GetLabels()`, and use `""` for a missing value. A stack-backed `[16]int` counter with a fallback for longer configurations would also remove `labelCountPool`, leaving only the topology map pool.

I compared such a prototype against this implementation for complete labels, mixed missing/empty labels, 17 location labels, and all-tombstone inputs; every target store produced the same weight. The important compatibility detail is to preserve exact key comparison (`label.Key == key`) rather than call `GetLabelValue`, which is case-insensitive.

Would that smaller design be sufficient for the allocation goal? If zero allocations for arbitrarily long label lists is a hard requirement, we could retain only the counter-slice pool.

tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -475,8 +478,9 @@ func TestSwitchModeDuringWorkload(t *testing.T) {
 						continue
 					}
 					if switched.Load() {
-						if !rgController.IsDegraded() {
-							atomic.AddInt64(&okAfter, 1)
+						atomic.AddInt64(&okAfter, 1)
```
lhy1024: Is it intentional that `okAfter` now counts successful requests even while the controller remains degraded? Previously the eventual assertion required ten successful, non-degraded requests. Since `degradedAfter` is only logged, this version can pass even if the controller never recovers from degraded mode after the switch. Could we keep the per-request timeout but preserve an assertion that non-degraded requests eventually succeed?

lhy1024 (approved):
Thanks for addressing the review feedback and removing the unrelated changes. The current implementation looks good to me.

Before merging, could you please rerun the benchmarks against the current head and update the benchstat results in the PR description? The implementation has changed substantially since the existing report: the benchmark now lives in `pkg/core`, and the label/counter pools have been removed in favor of direct label lookup and a stack-backed counter. The current description still reports `github.com/tikv/pd/server/cluster` and shows about a 21% `sec/op` regression, so fresh `sec/op`, `B/op`, and `allocs/op` numbers would make the final CPU/memory trade-off clear.

PR #9324, opened 2025-05-15
keyspace: allow `CreateKeyspaceByID` supporting name param

rleungx:

Issue Number: ref #9318

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -111,7 +111,10 @@ type CreateKeyspaceRequest struct {
 type CreateKeyspaceByIDRequest struct {
 	// ID of the keyspace to be created.
 	// Using an existing ID will result in error.
-	ID     *uint32
+	ID *uint32
+	// Name of the keyspace to be created.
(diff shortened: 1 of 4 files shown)


pkg/keyspace/keyspace.go
```diff
@@ -111,7 +111,10 @@ type CreateKeyspaceRequest struct {
 type CreateKeyspaceByIDRequest struct {
 	// ID of the keyspace to be created.
 	// Using an existing ID will result in error.
-	ID     *uint32
+	ID *uint32
+	// Name of the keyspace to be created.
+	// Using an existing name will result in error.
+	Name   string
```
lhy1024, line 117: Do we need to add a test for existed name?

lhy1024 (approved):

pkg/keyspace/keyspace.go
rleungx: done

JmPotato (approved):

PR #9328, opened 2025-05-15
scheduler: increase the default recovery time

### LINKED ISSUE #9329: Increase the default `RecoveryDurationGap`
## Enhancement Task

Previously, we set it to 600s, which is not suitable. We should adjust it to 900s (15m) to satisfy the SLA of EBS.

rleungx:

Issue Number: Close #9329

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/evict_slow_trend.go
```diff
@@ -40,7 +40,7 @@ import (
 const (
 	alterEpsilon               = 1e-9
 	minReCheckDurationGap      = 120 // default gap for re-check the slow node, unit: s
-	defaultRecoveryDurationGap = 600 // default gap for recovery, unit: s.
+	defaultRecoveryDurationGap = 900 // default gap for recovery, unit: s.
```
overvenus, line 43: Let's set it to 1800. It should be enough time for TiKV disk failover.

rleungx:
> NIT: please add necessary annotations to clarify this change, such as: "Based on the SLA level reflected by AWS EBS."
> 
> This will help provide better context for the modification.

Sure, the linked issue clarifies it. I will add the comment inside the code.

pkg/schedule/schedulers/evict_slow_trend.go
rleungx: ok

pkg/schedule/schedulers/evict_slow_trend.go
```diff
@@ -38,9 +38,11 @@ import (
...
 
 const (
-	alterEpsilon               = 1e-9
-	minReCheckDurationGap      = 120 // default gap for re-check the slow node, unit: s
-	defaultRecoveryDurationGap = 600 // default gap for recovery, unit: s.
+	alterEpsilon          = 1e-9
+	minReCheckDurationGap = 120 // default gap for re-check the slow node, unit: s
+	// We use 1800 seconds as the default gap for recovery, which is 30 minutes.
+	// This is based on the SLA level reflected by AWS EBS. And we can adjust it later if needed.
+	defaultRecoveryDurationGap = 1800 // default gap for recovery, unit: s.
```
lhy1024, line 45: Do we need to pick it to other branches?

lhy1024 (approved):
Do we need

pkg/schedule/schedulers/evict_slow_trend.go
rleungx: Yes

Issue #9296, opened 2025-05-08
Tracking issue for the Keyspace Service Limit


JmPotato:
# Background

**Service Limit** will serve as the first layer boundary for resource control, ensuring that each TiDB Service (virtual cluster) in a multi-tenant (also known as multi-keyspace) environment operates within its allocated resources, preventing tasks from affecting each other, and achieving resource isolation and service stability. A Service Limit can be applied to multiple TiDB instances, and within a single Service Limit, users can continue to create different resource groups.

This is the tracking issue for the Keyspace Service Limit, which aims to:

- Implement RU/s limits for each TiDB Service to ensure fair resource usage.
- Supports multiple resource groups and sets priorities to meet business isolation and flexible scheduling.
- Provide firendly observability to ensure service experience.

# Tasks

- [ ] Support resource group management at the keyspace level.
- [ ] Implement the basic RU/s service limit for the resource groups within the same keyspace.
- [ ] Support modify the service limit for a keyspace dynamically.
- [ ] Implement a priority allocation mechanism for RU resources among different resource groups under the same service limit.
- [ ] Data collection, aggregation, and observability.

PR #9331, opened 2025-05-19
feat(resourcemanager): implement keyspace manager for resource groups

JmPotato:

Issue Number: ref #9296.

```commit-message
Add keyspace-level resource group management.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -15,14 +15,11 @@
 package server
 
 import (
-	"context"
 	"encoding/json"
 	"math"
 	"sort"
-	"time"
 
 	"github.com/gogo/protobuf/proto"
-	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
 	rmpb "github.com/pingcap/kvproto/pkg/resource_manager"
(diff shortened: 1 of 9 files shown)


pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -0,0 +1,461 @@
+// Copyright 2022 TiKV Project Authors.
```
JmPotato, line 1:
```suggestion
// Copyright 2025 TiKV Project Authors.
```

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -118,6 +83,28 @@ func (m *Manager) GetBasicServer() bs.Server {
...
+	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
+	if krgm == nil {
+		krgm = newKeyspaceResourceGroupManager(keyspaceID, m)
+		m.Lock()
+		m.krgms[keyspaceID] = krgm
+		m.Unlock()
+	}
+	return krgm
+}
+
```
lhy1024, line 101: To avoid a race where two goroutines might both see `krgm == nil` and create it, should we move the m.Lock() call to before the initial if krgm == nil check?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -346,215 +310,20 @@ func (m *Manager) persistLoop(ctx context.Context) {
...
-}
-
-// FlushMetrics and set the maxPerSecRRU and maxPerSecWRU to the metrics.
-func (t *maxPerSecCostTracker) FlushMetrics() {
-	if t.lastRRUSum == 0 && t.lastWRUSum == 0 {
-		t.lastRRUSum = t.rruSum
-		t.lastWRUSum = t.wruSum
+func (m *Manager) dispatchConsumption(keyspaceID uint32, item *consumptionItem) {
+	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
+	if krgm == nil {
```
lhy1024, line 325: When will it be nill and do we need to add a log?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -0,0 +1,461 @@
...
+	defer krgm.RUnlock()
+	res := make([]*ResourceGroup, 0, len(krgm.groups))
+	for _, group := range krgm.groups {
+		if !inlcudeDefault && group.Name == reservedDefaultGroupName {
+			continue
+		}
+		res = append(res, group.Clone(true))
+	}
+	return res
+}
```
lhy1024, line 243: This function is named with `Mutable`, it returns a clone copy?

lhy1024 (approved):

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -0,0 +1,451 @@
...
+					zap.Int("index", idx),
+					zap.Error(err))
+			}
+		}
+		krgm.RUnlock()
+	}
+}
+
+// TODO: add keyspace name lable to the metrics.
+func (krgm *keyspaceResourceGroupManager) backgroundMetricsFlush(ctx context.Context) {
```
glorv, line 272: Is it better to share a single background routine for the backgroundMetricsFlush for all keyspaces? if there are thousands of keyspaces while each only contains few resource groups, to many background routines may have a bad effect on performance.

rleungx (approved):

Issue #8165, opened 2024-05-10
The last moments of offline store and online store progress are not displayed correctly


lhy1024:
### What did you do?

online 2 new store and offline them

### What did you expect to see?

offline store and online store progress are correct

### What did you see instead?

they stay here until pd restart

![image](https://github.com/tikv/pd/assets/19542290/5d0b76e0-6df7-4ea4-bdec-ed497b7abd67)

okJiang:
Is progress the only problem, and does `Tombstone Stores(2)` in abnormal stores need to become `Offline Stores(2)` as well?

lhy1024:
> ~Is progress the only problem, and does `Tombstone Stores(2)` in abnormal stores need to become `Offline Stores(2)` as well?~

No, store will become offline store firstly and then become tombostone if offline is finished.

okJiang:
This issue arises because Grafana's progress gauge can only display the last data. In the code, when the scale ends, we directly delete the corresponding metrics. In this case, Grafana does not show a zero value but instead displays the last valid value.

Therefore, I plan to set the corresponding metric to 100% when the scale ends and wait for 1 minute before deleting the corresponding metric. This way, Grafana will be able to display correctly.

PR #9333, opened 2025-05-20
core: Refactor checkstore and progress manager

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/8165

- Converge the progress calculation logic into the progress manager and progress indicator
- Each time new data is added, the indicator will automatically calculate the progress
- Progress related metrics will be updated in the progress manager
- Refactoring checkStore using switch
- Added GC logic in progress manager

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test

The speed change is smooth
image

After the online and offline are completed, the progress is 100
image

### Release note

```release-note
None.
```


pkg/progress/indicator.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return pi
+}
+
+func (p *progressIndicator) adjustWindowLength(dur time.Duration) {
+	if dur < minSpeedCalculationWindow {
+		dur = minSpeedCalculationWindow
+	} else if dur > maxSpeedCalculationWindow {
+		dur = maxSpeedCalculationWindow
+	}
```
JmPotato, line 103: Do we need to put two warn logs here as well?

pkg/progress/progress.go
```diff
@@ -15,229 +15,272 @@
...
+		failpoint.Inject("gcExpiredTime", func(val failpoint.Value) {
+			if s, ok := val.(string); ok {
+				var err error
+				exactExpiredTime, err = time.ParseDuration(s)
+				if err != nil {
+					panic(err)
+				}
+			}
+		})
+		if p.completeAt.Before(time.Now().Add(-exactExpiredTime)) {
```
JmPotato, line 134: Perhaps call `time.Now()` outside the loop.

okJiang:
ping @JmPotato @rleungx

pkg/progress/indicator.go
```diff
@@ -0,0 +1,168 @@
...
+		updateInterval:      updateInterval,
+		currentWindowLength: 1,
+	}
+
+	return pi
+}
+
+func (p *progressIndicator) adjustWindowLength(dur time.Duration) {
+	if dur < minSpeedCalculationWindow {
+		log.Warn("The window duration is too small, to ensure enough data to calculate progress, "+
```
rleungx: Is the log necessary?

server/cluster/cluster_test.go
```diff
@@ -534,16 +535,16 @@ func TestRemovingProcess(t *testing.T) {
...
 	// In above we delete 5 region from store 1, the total count of region in store 1 is 20.
 	// process = 5 / 20 = 0.25
-	re.Equal(0.25, p)
+	re.Equal(0.25, p.ProgressPercent)
 	// Each region is 100MB, we use more than 1s to move 5 region.
-	// speed = 5 * 100MB / 20s = 25MB/s
-	re.Equal(25.0, cs)
-	// left second = 15 * 100MB / 25s = 60s
-	re.Equal(60.0, l)
+	// speed = 5 * 100MB / 10s = 50MB/s
```
rleungx, line 546: Why change it?

server/cluster/cluster.go
```diff
@@ -1766,24 +1777,24 @@ func (c *RaftCluster) isStorePrepared() bool {
 }
 
 func (c *RaftCluster) checkStores() {
-	var offlineStores []*metapb.Store
-	var upStoreCount int
-	stores := c.GetStores()
+	c.UpdateAllStoreStatus()
```
rleungx: Is it duplicated with `runUpdateStoreStats`?

server/cluster/cluster.go
```diff
@@ -1766,24 +1777,24 @@ func (c *RaftCluster) isStorePrepared() bool {
...
+		offlineStores []*metapb.Store
+		upStoreCount  int
+		stores        = c.GetStores()
+	)
 
 	for _, store := range stores {
-		isUp, isOffline := c.checkStore(store, stores)
-		if isUp {
+		isInOffline := c.checkStore(store, stores)
+		if store.IsUp() && store.IsLowSpace(c.opt.GetLowSpaceRatio()) {
```
rleungx, line 1789: Why do we need to check the low space?

server/cluster/cluster.go
```diff
@@ -1766,24 +1777,24 @@ func (c *RaftCluster) isStorePrepared() bool {
...
 			upStoreCount++
 		}
-		if isOffline {
+		if isInOffline {
 			offlineStores = append(offlineStores, store.GetMeta())
 		}
 	}
 
-	if len(offlineStores) == 0 {
```
rleungx, line 1791: Why remove it?

server/cluster/cluster.go
```diff
@@ -1792,77 +1803,66 @@ func (c *RaftCluster) checkStores() {
...
+				zap.Stringer("store", store.GetMeta()),
+				zap.Duration("down-time", store.DownTime()),
 				errs.ZapError(err))
+		} else {
+			log.Info("auto gc the tombstone store success", zap.Stringer("store", store.GetMeta()), zap.Duration("down-time", store.DownTime()))
 		}
-	} else {
-		isOffline = true
+	}
+	if c.isStorePrepared() {
```
rleungx: Why change to store prepared?

server/cluster/cluster.go
```diff
@@ -1679,16 +1686,16 @@ func (c *RaftCluster) UpStore(storeID uint64) error {
...
 		zap.Uint64("store-id", storeID),
 		zap.String("store-address", newStore.GetAddress()))
-	err := c.setStore(newStore, options...)
-	if err == nil {
-		if exist {
-			// persist the store limit
-			_ = c.SetStoreLimit(storeID, storelimit.AddPeer, limiter[storelimit.AddPeer])
-			_ = c.SetStoreLimit(storeID, storelimit.RemovePeer, limiter[storelimit.RemovePeer])
-		}
-		c.resetProgress(storeID, store.GetAddress())
```
rleungx, line 1697: no need to reset?

pkg/progress/indicator.go
okJiang: No, I think it's optional.

server/cluster/cluster_test.go
okJiang: Because previous code set the progress in `RemoveStore`. For now, we always calculate the progress in the `checkStore`. So the number of calculations is twice before:
1. RemoveStore
2. checkStores
3. checkStores

Now it is once:
1. checkStores
2. checkStores

server/cluster/cluster.go
okJiang: It will be set to 100% in next `checkStores`, and gced by `progress.Manager.GC()`

server/cluster/cluster.go
okJiang: If len(offlineStores) == 0, it will do nothing. So it is unnecessary.

server/cluster/cluster.go
okJiang: It is from here
image

server/cluster/cluster.go
okJiang: It is from previous code, and I added them to one function `isStorePrepared`.


image

image

server/cluster/cluster.go
okJiang: I didn't notice it. It's meant to pass certain tests, I'll figure out how to modify it

server/cluster/cluster.go
okJiang: Can we just remove this check?🤔

server/cluster/cluster.go
rleungx: We can skip the following lines?

okJiang: If len(offlineStores) == 0, it will not print the warning log.

server/cluster/cluster.go
okJiang: Can we keep it? The frequency here is much lower than runUpdateStoreStats and will not affect the performance too much, but it can ensure that the region information in checkStore is the latest.

pkg/progress/indicator_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"math"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestProgressIndicator_UpdateProgress(t *testing.T) {
```
rleungx:
```suggestion
func TestProgressIndicatorUpdateProgress(t *testing.T) {
```

pkg/progress/indicator.go
rleungx: How about removing it or changing the log level?

server/cluster/cluster.go
rleungx: It's strange to update the statistics in two places. The interval of runUpdateStoreStats is 9ms, which is enough for progress update?

server/cluster/cluster.go
```diff
@@ -1792,77 +1803,66 @@ func (c *RaftCluster) checkStores() {
...
-			}
-		}
-		return
-	}
-
-	if store.IsPreparing() {
-		if store.GetUptime() >= c.opt.GetMaxStorePreparingTime() || c.GetTotalRegionCount() < core.InitClusterRegionThreshold {
+	switch store.GetNodeState() {
+	case metapb.NodeState_Preparing:
+		timeToServe := store.GetUptime() >= c.opt.GetMaxStorePreparingTime() ||
```
rleungx:
```suggestion
		readyToServe := store.GetUptime() >= c.opt.GetMaxStorePreparingTime() ||
```

server/cluster/cluster.go
```diff
@@ -1753,6 +1757,13 @@ func (c *RaftCluster) setStore(store *core.StoreInfo, opts ...core.StoreCreateOp
 }
 
 func (c *RaftCluster) isStorePrepared() bool {
+	// nolint:staticcheck
+	if c.schedulingController.IsPrepared() {
```
rleungx: There is a duplicate judgment. Why not keep the original implementation?

rleungx (approved):

PR #9336, opened 2025-05-21
feat(resourcemanager): load resource groups in the background

### LINKED ISSUE #9335: Load resource group will greatly affect availability of PD
## Enhancement Task
Currently, we need to load the resource group meta before the leader is ready to serve. If there are thousands of keyspaces, once the leader is changed, we must wait a long time before the resource group is ready. It's unreasonable. We should initialize the resource group meta asynchronously.

Some context related to the issue:
https://github.com/tikv/pd/blob/fa843982dd03f6abd640a01d4695e8e93d34dcc8/server/server.go#L1716-L1721

https://github.com/tikv/pd/blob/fa843982dd03f6abd640a01d4695e8e93d34dcc8/pkg/mcs/resourcemanager/server/manager.go#L128-L156

JmPotato:

Issue Number: close #9335, ref #9296.

```commit-message
Improves resource manager initialization by loading keyspace resource groups asynchronously in a background goroutine instead of blocking the leader startup process.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -40,10 +40,10 @@ import (
 )
 
 const (
-	persistLoopInterval        = 1 * time.Minute
+	persistLoopInterval        = time.Minute
 	metricsCleanupInterval     = time.Minute
 	metricsCleanupTimeout      = 20 * time.Minute
-	metricsAvailableRUInterval = 1 * time.Second
+	metricsAvailableRUInterval = time.Second
 	defaultCollectIntervalSec  = 20
 	tickPerSecond              = time.Second
 )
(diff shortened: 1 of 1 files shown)


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -141,9 +141,17 @@ func (m *Manager) Init(ctx context.Context) error {
...
 	// Load keyspace resource groups from the storage.
-	if err := m.loadKeyspaceResourceGroups(); err != nil {
-		return err
-	}
+	// This operation is potentially time-consuming when there are many keyspaces and resource groups.
+	// So we run it in a new goroutine to prevent it from blocking the leader startup.
+	go func() {
+		start := time.Now()
+		log.Info("start to load keyspace resource groups in the background")
+		if err := m.loadKeyspaceResourceGroups(start); err != nil {
```
glorv, line 149: This behavior is different from the old one if the background task failed(maybe due to IO errors). Silently omit the error may cause undefined behavior in some cases.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -141,9 +141,17 @@ func (m *Manager) Init(ctx context.Context) error {
 	}
 
 	// Load keyspace resource groups from the storage.
-	if err := m.loadKeyspaceResourceGroups(); err != nil {
-		return err
-	}
+	// This operation is potentially time-consuming when there are many keyspaces and resource groups.
+	// So we run it in a new goroutine to prevent it from blocking the leader startup.
```
Tema, line 145: If we move it to the background, could you please add reporting of resource group loading to the /ready API as well.

PR #9337, opened 2025-05-21
client: add retry mechanism for GetTS

### LINKED ISSUE #8630: Caller don't need to care about new errors in GetTS
```
tidb lightning encountered error: [Lightning:KV:ErrCreateKVClient]create kv client error: rpc error: code = Unknown desc = [PD:tso:ErrGenerateTimestamp]generate timestamp failed, requested pd is not leader of cluster
```

https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fpull_lightning_integration_test/detail/pull_lightning_integration_test/2225/pipeline

Please auto-retry or internally forward to leader for these errors. I have requested it before https://github.com/tikv/pd/issues/8443

rleungx:

Issue Number: Close #8630

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


client/client.go
```diff
@@ -516,10 +516,28 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
 // GetTS implements the TSOClient interface.
 func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err error) {
-	resp := c.GetTSAsync(ctx)
-	return resp.Wait()
+	for range maxTSORetryTimes {
+		resp := c.GetTSAsync(ctx)
+		if physical, logical, err = resp.Wait(); err != nil {
+			if errs.IsLeaderChange(err) {
+				// If the leader changes, we need to retry.
+				time.Sleep(retryInterval)
```
JmPotato, line 533: What about using `select` to check the context as well?

Tema: Can we skip sleep on the first retry to reduce the stall on graceful leader transfer?

client/client.go
```diff
@@ -516,10 +516,28 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
+)
+
 // GetTS implements the TSOClient interface.
 func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err error) {
-	resp := c.GetTSAsync(ctx)
-	return resp.Wait()
+	for range maxTSORetryTimes {
+		resp := c.GetTSAsync(ctx)
+		if physical, logical, err = resp.Wait(); err != nil {
+			if errs.IsLeaderChange(err) {
```
Tema: This condition covers lot of errors including `NoLeader` when PD is overloaded and can't stabilize the leader. The 10 retries per second from each SELECT can trigger a request storm which PD won't be able to recover.  Please consider backoff with reduced number of attempts or/and cover this PD call with circuit breaker like GetRegion method has.

rleungx: TSO requests are processed asynchronously. It will be batched and then sent to PD, so I think here the backoff is unnecessary.

Tema: Right. I also checked if that could lead to connection churn, but this does not look like the case.

client/client.go
```diff
@@ -519,10 +519,37 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
+					if retryCount != 0 {
+						// If the leader changes, we need to retry.
+						// For the first time, we retry immediately to avoid impacting the latency.
+						time.Sleep(retryInterval)
+					}
+					continue
+				}
+			}
+			break
+		}
```
JmPotato: This can reduce one level of nesting.

```suggestion
			if !errs.IsLeaderChange(err) {
				break
			}
			select {
			case <-ctx.Done():
				return 0, 0, errs.ErrClientGetTSO.Wrap(ctx.Err()).GenWithStackByCause()
			default:
				if retryCount != 0 {
					// If the leader changes, we need to retry.
					// For the first time, we retry immediately to avoid impacting the latency.
					time.Sleep(retryInterval)
				}
				continue
			}
		}
```

client/client.go
```diff
@@ -519,10 +519,37 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
+		if physical, logical, err = resp.Wait(); err != nil {
+			if errs.IsLeaderChange(err) {
+				select {
+				case <-ctx.Done():
+					return 0, 0, errs.ErrClientGetTSO.Wrap(ctx.Err()).GenWithStackByCause()
+				default:
+					if retryCount != 0 {
+						// If the leader changes, we need to retry.
+						// For the first time, we retry immediately to avoid impacting the latency.
+						time.Sleep(retryInterval)
```
JmPotato: Using `<-time.After(interval)` is more suitable within a select statement, like:

```go
var interval time.Duration
if retryCount != 0 {
	interval = retryInterval
}
select {
case <-ctx.Done():
	return 0, 0, errs.ErrClientGetTSO.Wrap(ctx.Err()).GenWithStackByCause()
case <-time.After(interval):
	continue
}
```

client/client.go
```diff
@@ -519,10 +519,39 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
+			// If the leader changes, we need to retry.
+			// For the first time, we retry immediately to avoid impacting the latency.
+			var interval time.Duration
+			if retryCount != 0 {
+				interval = retryInterval
+			}
+			select {
+			case <-ctx.Done():
+				return 0, 0, errs.ErrClientGetTSO.Wrap(ctx.Err()).GenWithStackByCause()
+			case <-time.After(interval):
```
MyonKeminta, line 548: ~~In performance sensitive situation, maybe reusing a single timer created by `NewTimer` is better than using `time.After` for every retry?~~
Just found that Go 1.23 no longer recommend preferring `NewTimer`. Ignore this.

client/client.go
```diff
@@ -519,10 +519,39 @@ func (c *client) GetLocalTSAsync(ctx context.Context, _ string) tso.TSFuture {
...
+	retryInterval = 100 * time.Millisecond
+)
+
 // GetTS implements the TSOClient interface.
 func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err error) {
-	resp := c.GetTSAsync(ctx)
-	return resp.Wait()
+	var retryCount int
+	for retryCount = range maxTSORetryTimes {
+		resp := c.GetTSAsync(ctx)
```
MyonKeminta, line 533: Is it that `GetTSAsync` currently still doesn't have built-in retrying?

client/metrics/metrics.go
```diff
@@ -173,6 +175,16 @@ func initMetrics(constLabels prometheus.Labels) {
...
 		})
 
+	TSORetryCount = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace:   "pd_client",
+			Subsystem:   "request",
+			Name:        "tso_retry",
+			ConstLabels: constLabels,
+			Help:        "Counter of retry time for TSO request.",
+			Buckets:     []float64{1, 2, 4, 8, 10, 12, 16, 20},
```
MyonKeminta:
```suggestion
			Buckets:     []float64{1, 2, 4, 8, 12, 16, 20},
```

Maybe this is better, otherwise the bucket 8-10, 10-12 is smaller than previous bucket 4-8 ...

client/client.go
rleungx: Yes, for GetTSAsync, the caller retry will be much easier.

MyonKeminta (approved):

rleungx:
@okJiang PTAL

client/metrics/metrics.go
```diff
@@ -173,6 +175,16 @@ func initMetrics(constLabels prometheus.Labels) {
...
 		})
 
+	TSORetryCount = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace:   "pd_client",
+			Subsystem:   "request",
+			Name:        "tso_retry",
+			ConstLabels: constLabels,
+			Help:        "Counter of retry time for TSO request.",
+			Buckets:     []float64{1, 2, 4, 8, 12, 16, 20},
```
okJiang, line 188: Can we add more buckets?

okJiang (commented):
rest lgtm

client/metrics/metrics.go
rleungx: The max retry time is 20 now.

PR #9340, opened 2025-05-22
gc: global GC barriers support

tiancaiamao:

Issue Number: ref #8978

Subtask of Refactor GC related PD APIs:   **Global GC barriers support**

- Add new grpc API `SetGlobalGCBarrier` and `DeleteGlobalGCBarrier`
- The global GC barriers will also block `AdvanceTxnSafePoint` just like the keyspace level GC barriers

```commit-message
```

### Check List

Tests

- [x] Unit test
- [x] Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Add grpc API for global GC barriers support
```


pkg/errs/errno.go
```diff
@@ -539,10 +539,11 @@ var (
...
+	ErrGCSafePointExceedsTxnSafePoint      = errors.Normalize("trying to update GC safe point to a too large value that exceeds the txn safe point, current value: %v, given: %v, current txn safe point: %v", errors.RFCCodeText("PD:gc:ErrGCSafePointExceedsTxnSafePoint"))
+	ErrDecreasingTxnSafePoint              = errors.Normalize("trying to update txn safe point to a smaller value, current value: %v, given: %v", errors.RFCCodeText("PD:gc:ErrDecreasingTxnSafePoint"))
+	ErrGCBarrierTSBehindTxnSafePoint       = errors.Normalize("trying to set a GC barrier on ts %d which is already behind the txn safe point %d", errors.RFCCodeText("PD:gc:ErrGCBarrierTSBehindTxnSafePoint"))
+	ErrReservedGCBarrierID                 = errors.Normalize("trying to set a GC barrier with a barrier ID that is reserved: %v", errors.RFCCodeText("PD:gc:ErrReservedGCBarrierID"))
+	ErrGlobalGCBarrierTSBehindTxnSafePoint = errors.Normalize("trying to set a global GC barrier on ts %d which is already behind the txn safe point %d", errors.RFCCodeText("PD:gc:ErrGlobalGCBarrierTSBehindTxnSafePoint"))
```
MyonKeminta: As txn safe points are keyspace scope and global GC barriers are global scope, there's no such a txn safe point in the context. How about explicitly explain which keyspace it is, like:
```
trying to set a global GC barrier on ts %d which is already behind the txn safe point %d of keyspace id=%d
```
and add the keyspace id to the arguments when generating this error

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
 	}
 }
 
+// SetGlobalGCBarrier sets a global GC barrier.
+// Global GC barrier takes effect globally, every keyspace should also consider global barriers.
```
MyonKeminta: Can you consider enrich this comment with detailed descriptions about its behavior under either sucessful cases or failed cases, guarantees on the result, etc., like how `SetGCBarrier` did?

pkg/storage/endpoint/gc_states.go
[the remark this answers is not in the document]
tiancaiamao: test added

server/gc_service.go
```diff
@@ -532,3 +531,82 @@ func (s *GrpcServer) GetAllKeyspacesGCStates(ctx context.Context, request *pdpb.
 		GcStates: gcStatesPb,
 	}, nil
 }
+
+// SetGlobalGCBarrier sets a global GC barrier.
+func (s *GrpcServer) SetGlobalGCBarrier(ctx context.Context, request *pdpb.SetGlobalGCBarrierRequest) (*pdpb.SetGlobalGCBarrierResponse, error) {
```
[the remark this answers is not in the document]
tiancaiamao: In another PR https://github.com/tikv/pd/pull/9361

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
...
+		}
+		if barrierTS < txnSafePoint {
+			return errs.ErrGlobalGCBarrierTSBehindTxnSafePoint.GenWithStackByArgs(barrierTS, txnSafePoint)
+		}
+
+		err2 = wb.SetGlobalGCBarrier(newBarrier)
+		return err2
+	})
+	if err != nil {
+		log.Error("failed to set Global GC barrier",
```
MyonKeminta:
```suggestion
		log.Error("failed to set global GC barrier",
```

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
...
+			zap.String("barrier-id", barrierID),
+			zap.Uint64("barrier-ts", barrierTS),
+			zap.Duration("ttl", ttl), zap.Error(err))
+		return nil, err
+	}
+	log.Info("Global GC barrier set",
```
MyonKeminta:
```suggestion
	log.Info("global GC barrier set",
```

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
...
+		var err1 error
+		deletedBarrier, err1 = m.gcMetaStorage.LoadGlobalGCBarrier(barrierID)
+		if err1 != nil {
+			return err1
+		}
+		return wb.DeleteGlobalGCBarrier(barrierID)
+	})
+
+	if err != nil {
+		log.Error("failed to delete Global GC barrier",
```
MyonKeminta:
```suggestion
		log.Error("failed to delete global GC barrier",
```

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
...
+			zap.String("barrier-id", barrierID), zap.Error(err))
+		return nil, err
+	}
+
+	if deletedBarrier == nil {
+		log.Info("deleting a not-existing Global GC barrier",
```
MyonKeminta:
```suggestion
		log.Info("deleting a not-existing global GC barrier",
```

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
...
+			zap.String("barrier-id", barrierID))
+	} else {
+		log.Info("Global GC barrier deleted",
```
MyonKeminta:
```suggestion
		log.Info("global GC barrier deleted",
```

pkg/gc/gc_state_manager.go
```diff
@@ -417,6 +442,114 @@ func (*GCStateManager) logAdvancingTxnSafePoint(keyspaceID uint32, result Advanc
 	}
 }
 
+// SetGlobalGCBarrier sets a global GC barrier.
+// Global GC barrier takes effect globally, every keyspace should also consider global barriers.
+func (m *GCStateManager) SetGlobalGCBarrier(_ context.Context, barrierID string, barrierTS uint64, ttl time.Duration, now time.Time) (*endpoint.GlobalGCBarrier, error) {
```
MyonKeminta: It's not important thought, I prefer moving the global GC barrier releated functions after the non-global ones, for a concern about ordering code from basic to extended

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,91 @@ var (
...
+		res.ExpirationTime = expirationTime
+	}
+	return res
+}
+
+// ToServiceSafePoint converts the global GCBarrier to a synonymous ServiceSafePoint for storing physically.
+// This method should never be used unless handling the physical data, which needs to be compatible with service safe
+// points.
+// This method is public only for keeping some old HTTP API compatible.
+func (b *GlobalGCBarrier) ToServiceSafePoint() *ServiceSafePoint {
```
MyonKeminta: It doesn't make sense to convert the global GC barrier to service safe points. Please give it standalone (un)marshalling implementation.
The reason why the non-global GC barrier uses service safe point data structure internally is to ensure the compatibility of its data format. Global GC barrier as a brand-new feature doesn't need to be compatible with anything in the data storage.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,91 @@ var (
 	_ json.Unmarshaler = (*ServiceSafePoint)(nil)
 )
 
+// GlobalGCBarrier represents a global GC barrier.
+// It looks like a GCBarrier now but the struct might change with the code evolve.
```
MyonKeminta, line 125: A more important reason is to distinguish from GC barrier by the type system, avoiding potential misuse between them in code.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,91 @@ var (
...
+
+type globalGCBarrierDecoder struct {
+	barriers []*GlobalGCBarrier
+}
+
+func (d *globalGCBarrierDecoder) decode(v *ServiceSafePoint) {
+	if v != nil {
+		d.barriers = append(d.barriers, globalGCBarrierFromServiceSafePoint(v))
+	}
+}
```
MyonKeminta: For the same reason mentioned above, please avoid reusing `ServiceSafePoint`.

pkg/storage/endpoint/gc_states.go
```diff
@@ -308,19 +403,49 @@ func (p GCStateProvider) LoadGCBarrier(keyspaceID uint32, barrierID string) (*GC
...
+	serviceSafePoint, err := loadJSON[*ServiceSafePoint](p.storage, key)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	barrier := globalGCBarrierFromServiceSafePoint(serviceSafePoint)
+	return barrier, nil
+}
+
+// LoadGlobalGCBarriers loads all global GC barriers.
+func (p GCStateProvider) LoadGlobalGCBarriers() ([]*GlobalGCBarrier, error) {
```
MyonKeminta:
```suggestion
func (p GCStateProvider) LoadAllGlobalGCBarriers() ([]*GlobalGCBarrier, error) {
```

pkg/gc/gc_state_manager.go
MyonKeminta, line 1: There's still something missing: As said in our design document *Compatibility / UpdateServiceGCSafePoint* section, call with service ID "native_br" need to be redirected to global GC barrier. Similar to that, you can see how "gc_worker" is redirected to AdvanceTxnSafePoint in the current existing code, in the method `CompatibleUpdateServiceGCSafePoint`.

pkg/errs/errno.go
tiancaiamao: > there's no such a txn safe point in the context

No, the global GC barrier works globally, so it compares with all txn safe points, and  **any txn safe point** that is greater than the global GC barrier prevent updating the global GC barrier.  @MyonKeminta

MyonKeminta: I mean, in the context of accessing the global GC barrier, there is not a corresponding txn safe point **in this context** here. When you find a txn safe point in some keyspace is too large and you say "global GC barrier ts is behind the txn safe point", one cannot know **which** txn safe point is causing the problem, unless you explicitly tell clearify which keyspace it came from. So I suggest you add the keyspace information to the error messate.

tiancaiamao: OK, I'll add the keyspace information.

pkg/storage/endpoint/gc_states.go
tiancaiamao: Done

pkg/gc/gc_state_manager.go
tiancaiamao: Updated

pkg/errs/errno.go
```diff
@@ -541,12 +541,13 @@ var (
...
```
MyonKeminta: Keyspace is still missing in the error message

pkg/gc/gc_state_manager.go
```diff
@@ -781,3 +830,139 @@ type GCState struct {
...
+		for _, keyspaceMeta := range allKeyspaces {
+			if keyspaceMeta.State != keyspacepb.KeyspaceState_ENABLED {
+				continue
+			}
+			txnSafePoint, err1 := m.gcMetaStorage.LoadTxnSafePoint(keyspaceMeta.Id)
+			if err1 != nil {
+				return err1
+			}
+			if barrierTS < txnSafePoint {
+				return errs.ErrGlobalGCBarrierTSBehindTxnSafePoint.GenWithStackByArgs(barrierTS, txnSafePoint, keyspaceMeta.Name)
```
MyonKeminta: Currently in PD's logs about GC, we always uses keyspace ID instead of keyspace name. It's ok that you include the keyspace name in the error message, but keyspace ID is still necessary. Otherwise it may need more mental cost to map keyspace names and IDs when investigating PD's logs.
Please consider using keyspace ID, or include both name and ID.

pkg/gc/gc_state_manager.go
```diff
@@ -781,3 +830,139 @@ type GCState struct {
...
+			}
+		}
+
+		// NOTE, allKeyspaces by LoadRangeKeyspace() do not contain the null keyspace!
+		txnSafePoint, err2 := m.gcMetaStorage.LoadTxnSafePoint(constant.NullKeyspaceID)
+		if err2 != nil {
+			return err2
+		}
+		if barrierTS < txnSafePoint {
+			return errs.ErrGlobalGCBarrierTSBehindTxnSafePoint.GenWithStackByArgs(barrierTS, txnSafePoint, "null keyspace")
```
MyonKeminta: `"null keyspace"` is not a keyspace name. Consider:
* non-null: `"{ id: %d, name: \"%s\" }"`
* null: `<null>`

pkg/gc/gc_state_manager.go
```diff
@@ -731,7 +758,29 @@ func (m *GCStateManager) CompatibleUpdateServiceGCSafePoint(serviceID string, ne
...
+		} else {
+			_, err = m.deleteGlobalGCBarrierImpl(context.Background(), serviceID)
+		}
+		if err != nil && !errors.Is(err, errs.ErrGlobalGCBarrierTSBehindTxnSafePoint) {
+			return nil, false, err
+		}
+		// The atomicity between setting/deleting GC barrier and loading the txn safe point is not guaranteed here.
+		// It doesn't matter much whether it's atomic, but it's important to ensure LoadTxnSafePoint happens *AFTER*
+		// setting/deleting global GC barrier.
+		txnSafePoint, err := m.gcMetaStorage.LoadTxnSafePoint(keyspaceID)
```
MyonKeminta: For global GC barrier, you should return the max txn safe point among all keyspaces, otherwise the return value is not sufficient to notify the caller whether it's sucessful to set the global GC barrier.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
...
 )
 
+// TimeOptional is time.Time or nil
+type TimeOptional struct {
+	*time.Time
+}
+
+func (t *TimeOptional) int64() int64 {
+	if t.Time != nil {
+		return t.UnixNano()
```
MyonKeminta: It's not necessary to use unix nano, for the following reasons:
* We don't have such high precision. For GC barrier, the precision is second level, and we round up when storing it to avoid it expires earlier than expected.
* As (non-global) GC barriers are already implemented, we can keep the style of the two consistent. The fact that we don't need to keep global GC barriers compatible with service safe points doesn't necessarily mean that we need to make it too difference with non-global GC barriers.
* The representable range becomes smaller, making some `now + TTL` values that were valid for non-global GC barriers imrepresentable for global GC barriers. At least you need to truncate these too large-values to MaxInt64, but you didn't. Even if you did it, it still causes some inconstent behavior between global and non-globla GC barriers.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
...
+}
+
+// GlobalGCBarrier represents a global GC barrier.
+// It looks like a GCBarrier now but the struct might change with the code evolve.
+// A more important reason is to distinguish from GC barrier by the type system,
+// avoiding potential misuse between them in code.
+type GlobalGCBarrier struct {
+	BarrierID string       `json:"barrier_id"`
+	BarrierTS uint64       `json:"barrier_ts"`
+	ExpiredAt TimeOptional `json:"expired_at"`
```
MyonKeminta: The service safe points introduces many negative examples of namings. `ExpiredAt` looks like describing something that has already been expired, and records a past time at which the expiration happened.
As part of the new-designed GC, please keep the naming style consistent with `GCBarrier`, instead of `ServiceGCSafePoint`.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
...
+
+// NewGlobalGCBarrier creates a new GlobalGCBarrier. The given expirationTime will be rounded up to the next second if it's
+// not in integral seconds.
+// Passing nil to `expirationTime` means the barrier never expires.
+func NewGlobalGCBarrier(barrierID string, barrierTS uint64, expirationTime *time.Time) *GlobalGCBarrier {
+	// Round up the expirationTime.
+	if expirationTime != nil {
+		rounded := expirationTime.Add(time.Second - time.Nanosecond).Truncate(time.Second)
+		*expirationTime = rounded
+	}
```
MyonKeminta, line 142: The rounding-up used in GC bariers is for handling the fact that the storage precision is in seconds level, and it doesn't make sense if you store it in nanoseconds precision.
However, please consider keep the rounding up and change the precision back to seconds.

tiancaiamao: Changed back to seconds

tiancaiamao:
The CI is itself unstable, unrelated the change in this PR.
ref https://github.com/tikv/pd/issues/9461

pkg/gc/gc_state_manager.go
```diff
@@ -868,47 +868,59 @@ func (m *GCStateManager) SetGlobalGCBarrier(ctx context.Context, barrierID strin
...
+	}
+	// NOTE, allKeyspaces by LoadRangeKeyspace() do not contain the null keyspace!
+	txnSafePoint, err3 := m.gcMetaStorage.LoadTxnSafePoint(constant.NullKeyspaceID)
+	if err3 != nil {
+		err = err3
+		return
+	}
+	if txnSafePoint > maxSafePoint {
+		maxSafePoint = txnSafePoint
+		keyspaceID = constant.NullKeyspaceID
```
MyonKeminta: You may return the null keyspace ID with a non-null keyspace name retrieved in the previous loop.

pkg/gc/gc_state_manager.go
```diff
@@ -868,47 +868,59 @@ func (m *GCStateManager) SetGlobalGCBarrier(ctx context.Context, barrierID strin
...
-		if err2 != nil {
-			return err2
+		maxTxnSafePoint, keyspaceName, keyspaceID, err := m.getAllKeyspacesMaxTxnSafePoint()
+		if err != nil {
+			return err
 		}
-		if barrierTS < txnSafePoint {
-			return errs.ErrGlobalGCBarrierTSBehindTxnSafePoint.GenWithStackByArgs(barrierTS, txnSafePoint, "null keyspace")
+		if barrierTS < maxTxnSafePoint {
+			return errs.ErrGlobalGCBarrierTSBehindTxnSafePoint.GenWithStackByArgs(barrierTS, maxTxnSafePoint, keyspaceName, keyspaceID)
```
MyonKeminta: Consider print `"{ name: \"%s\", id:  %d}"` for non null keyspace and something like `"<null>"` or `"<null_keyspace>"` for null keyspace, so that the error format becomes `"... ehind the txn safe point %d of keyspace %s"` and you format it like either 

```go
...GenWithStackByArgs(barrierTS, maxTxnSafePoint, fmt.Sprintf("{ name: \"%s\", id: %d }", keyspaceName, keyspaceID))
```

or

```go
...GenWithStackByArgs(barrierTS, maxTxnSafePoint, "<null_keyspace>")
```

This avoids ambiguity between the representation of null and not-null keyspaces, and avoids the problem that the null keyspace doesn't have a proper widely recognized name (though we believe the proper name should be an empty string).

MyonKeminta: By the way, formatting using `%q` might be better as it handles the case that the name contains special characters, and automatically wrapped with quotes:

```go
fmt.Sprintf("{ name: %q, id: %d }", keyspaceName, keyspaceID)
```

pkg/gc/gc_state_manager.go
```diff
@@ -781,3 +830,151 @@ type GCState struct {
...
+		return nil, errs.ErrInvalidArgument.GenWithStackByArgs("barrierID", barrierID)
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	return m.setGlobalGCBarrierImpl(ctx, barrierID, barrierTS, ttl, now)
+}
+
+func (m *GCStateManager) getAllKeyspacesMaxTxnSafePoint() (maxSafePoint uint64, keyspaceName string, keyspaceID uint32, err error) {
```
MyonKeminta: Please comment that this must be called within a transaction. Better accepting a dummy `WriteBatch` to force it to be called in a transaction, like how `getGCStateInTransaction` did.

MyonKeminta: Avoid using the word *safe point* here. Don't forget one of  a reason for our refactor is to clearify at which phase the safe point value suitable for what functionality, leading to the design that GC safe point and txn safe point is clearly distinguished, and explicitly state it when possible.
Here please name it `maxTxnSafePoint`.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
 	_ json.Unmarshaler = (*ServiceSafePoint)(nil)
 )
 
+// TimeOptional is time.Time or nil
+type TimeOptional struct {
+	*time.Time
+}
+
+func (t *TimeOptional) int64() int64 {
```
MyonKeminta:
```suggestion
func (t *TimeOptional) unixSeconds() int64 {
```

While the name `int64` doesn't well-define what it exactly returns, the name `unixSeconds` looks a better choice.

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
...
+}
+
+// MarshalJSON implements json.Marshaler for TimeOptional
+func (t *TimeOptional) MarshalJSON() ([]byte, error) {
+	return json.Marshal(t.int64())
+}
+
+// UnmarshalJSON implements json.Marshaler for TimeOptional
+func (t *TimeOptional) UnmarshalJSON(b []byte) error {
+	var expireAt int64
```
MyonKeminta: With in the `TimeOptional` which is a wrapper on general nullalble times, the name `expireAt` isn't proper to occur here.

```suggestion
	var unixSeconds int64
```

pkg/gc/gc_state_manager.go
tiancaiamao: Done

pkg/gc/gc_state_manager.go
tiancaiamao: fixed

pkg/gc/gc_state_manager.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
```diff
@@ -567,6 +580,20 @@ func (s *gcStateManagerTestSuite) getAllGCBarriers(keyspaceID uint32) []*endpoin
...
 }
 
+func (s *gcStateManagerTestSuite) getGlobalGCBarrier(barrierID string) *endpoint.GlobalGCBarrier {
+	re := s.Require()
+	barrier, err := s.manager.gcMetaStorage.LoadGlobalGCBarrier(barrierID)
+	re.NoError(err)
+	return barrier
+}
+
+func (s *gcStateManagerTestSuite) getGlobalGCBarriers() []*endpoint.GlobalGCBarrier {
```
MyonKeminta:
```suggestion
func (s *gcStateManagerTestSuite) getAllGlobalGCBarriers() []*endpoint.GlobalGCBarrier {
```

pkg/gc/gc_state_manager.go
```diff
@@ -781,3 +835,157 @@ type GCState struct {
...
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	return m.setGlobalGCBarrierImpl(ctx, barrierID, barrierTS, ttl, now)
+}
+
+// getAllKeyspacesMaxTxnSafePoint must be called inside a transaction,
+// The WriteBatch parameter in function signature is deliberate to the call safe, do not pass nil.
+func (m *GCStateManager) getAllKeyspacesMaxTxnSafePoint(_ *endpoint.GCStateWriteBatch) (maxTxnSafePoint uint64, keyspaceName string, keyspaceID uint32, err error) {
```
MyonKeminta, line 878: As this function was not correctly implemented by mistake previously, how about add a unit test for this function?

pkg/gc/gc_state_manager_test.go
```diff
@@ -849,6 +876,247 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
...
+	re.Error(err)
+	re.ErrorIs(err, errs.ErrGlobalGCBarrierTSBehindTxnSafePoint)
+	re.Empty(s.getGlobalGCBarriers())
+	// BarrierTS exactly equals to txn safe point is allowed.
+	b, err = s.manager.SetGlobalGCBarrier(context.Background(), "b6", 60, time.Hour, now)
+	re.NoError(err)
+	expected = endpoint.NewGlobalGCBarrier("b6", 60, ptime(now.Add(time.Hour)))
+	re.Equal(expected, b)
+	re.Len(s.getGlobalGCBarriers(), 1)
+	re.Equal(expected, s.getGlobalGCBarrier("b6"))
```
MyonKeminta: Multi-keyspace cases are not covered. Consider cover the case that multiple keyspaces (including the null keyspace) has different txn safe points.

pkg/gc/gc_state_manager_test.go
```diff
@@ -849,6 +876,247 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
...
+
+	// Clear.
+	_, err = s.manager.DeleteGlobalGCBarrier(context.Background(), "b6")
+	re.NoError(err)
+
+	// Global barriers take effect on all keyspaces.
```
MyonKeminta:
```suggestion
	// Global GC barriers take effect on all keyspaces.
```

pkg/gc/gc_state_manager_test.go
```diff
@@ -849,6 +876,247 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
 	re.Empty(res.BlockerDescription)
 }
 
+func (s *gcStateManagerTestSuite) TestGlobalGCBarriers() {
+	re := s.Require()
+
+	now := time.Date(2025, 03, 06, 11, 50, 30, 0, time.Local)
+	re.Empty(s.getGlobalGCBarriers())
+
+	keyspaceID := s.keyspacePresets.manageable[0]
```
MyonKeminta: `manageable[0]` is always the null keyspace. And this doesn't match our test pattern in this file.

For most of the cases in the test where different states in different keyspaces is not relavant, consider replace your calls to `AdvanceTxnSafePoint` to
```go
for _, id := range s.keyspacePresets.manageable {
    s.manager.AdvanceTxnSafePoint(...)
}
```

pkg/gc/gc_state_manager.go
MyonKeminta: Looks it's still not using the maximum among all keyspaces.

pkg/gc/gc_state_manager_test.go
```diff
@@ -1118,6 +1386,37 @@ func (s *gcStateManagerTestSuite) TestServiceGCSafePointCompatibility() {
 	re.NoError(err)
 	re.Equal(uint64(15), res.NewTxnSafePoint)
 	re.Empty(res.BlockerDescription)
+
+	// Test CompatibleUpdateServiceGCSafePoint for native_br
+	// delete svc1's service safepoint before test
+	// gcworker's service safepoint cannot be deleted because it's mapping to txn safe point rather than a barrier
```
MyonKeminta:
```suggestion
	// gc_worker's service safepoint cannot be deleted because it's mapping to txn safe point rather than a barrier
```

pkg/gc/gc_state_manager_test.go
```diff
@@ -849,6 +876,247 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
...
+	ks2 := s.keyspacePresets.manageable[1]
+	_, err = s.manager.SetGlobalGCBarrier(context.Background(), "b1", 200, time.Hour, now)
+	re.NoError(err)
+	re.Empty(s.getAllGCBarriers(ks1))
+	re.Empty(s.getAllGCBarriers(ks2))
+	res, err = s.manager.AdvanceTxnSafePoint(ks2, 300, now)
+	re.NoError(err)
+	re.Equal(uint64(200), res.NewTxnSafePoint)
+	re.Contains(res.BlockerDescription, "BarrierID: \"b1\"")
+}
```
MyonKeminta, line 1230: Please also consider covering the case that global GC barriers occurs with non-global GC barriers and/or TiDB min start ts, and check the txn safe point is always blocked on the minimal one.

pkg/gc/gc_state_manager_test.go
```diff
@@ -1118,6 +1386,37 @@ func (s *gcStateManagerTestSuite) TestServiceGCSafePointCompatibility() {
...
+	re.NoError(err)
+	re.Equal(uint64(32), res.NewTxnSafePoint)
+	re.Contains(res.BlockerDescription, `BarrierID: "native_br"`)
+	_, allSsp, err = s.provider.CompatibleLoadAllServiceGCSafePoints()
+	re.NoError(err)
+	re.Len(allSsp, 2)
+	re.Equal("gc_worker", allSsp[0].ServiceID)
+	re.Equal(uint64(32), allSsp[1].SafePoint)
+	re.Equal("native_br", allSsp[1].ServiceID)
+	re.Equal(uint64(32), allSsp[1].SafePoint)
```
MyonKeminta: Also check:
* Advancing txn safe point of other keyspaces can also be blocked
* The result can be read out by using `getGlobalGCBarrier`/`getAllGlobalGCBarriers`, but not visible by `getGCBarrier`/`getAllGCBarriers`
* `CompatibleUpdateServiceGCSafePoint` on `native_br` cannot succeed if any of the keyspace has larger txn safe point than the given service safe point value (and the test should cover both null keyspace or non-null keyspace. note that null keyspace is not completely equivalent or symmetric to other keyspaces).

pkg/storage/endpoint/gc_states_test.go
```diff
@@ -91,6 +92,39 @@ func TestGCBarriersConversions(t *testing.T) {
 	}
 }
 
+func TestGlobalGCBarriersConversions(t *testing.T) {
```
MyonKeminta:
```suggestion
func TestGlobalGCBarriersMarshalling(t *testing.T) {
```

The test `TestGCBarriersConversions` actually means converting between GC barriers and the underlying storage type `ServiceSafePoint`, which is not part of global GC barrier.

tests/server/gc/gc_test.go
```diff
@@ -251,4 +251,103 @@ func TestGCOperations(t *testing.T) {
...
+		resp, err := grpcPDClient.SetGlobalGCBarrier(ctx, req)
+		re.NoError(err)
+		re.NotNil(resp.Header)
+		re.Nil(resp.Header.Error)
+		re.Equal("b2", resp.GetNewBarrierInfo().GetBarrierId())
+		re.Equal(uint64(24), resp.GetNewBarrierInfo().GetBarrierTs())
+		re.Equal(math.MaxInt64, int(resp.GetNewBarrierInfo().GetTtlSeconds()))
+	}
+
+	// Failed to set a global GC barrier (below txn safe point)
```
MyonKeminta: Also consider cover the case that there exist different txn safe points in multiple keyspaces

tests/server/gc/gc_test.go
```diff
@@ -251,4 +251,103 @@ func TestGCOperations(t *testing.T) {
 		re.Greater(gcState.GetGcBarriers()[0].GetTtlSeconds(), int64(3500))
 		re.Less(gcState.GetGcBarriers()[0].GetTtlSeconds(), int64(3601))
 	}
+
+	// Global GC Barrier API
+	{
+		req := &pdpb.SetGlobalGCBarrierRequest{
```
MyonKeminta, line 277:
```suggestion
		// Successfully sets a global GC barrier
		req := &pdpb.SetGlobalGCBarrierRequest{
```

pkg/gc/gc_state_manager.go
tiancaiamao: I made a terrible mistable by changing the wrong code branch.
I should have update the `case keypath.NativeBRServiceSafePointID` branch to using getAllKeyspacesMaxTxnSafePoint, but I mistakenly change the default branch.

pkg/gc/gc_state_manager.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
tiancaiamao: done

pkg/gc/gc_state_manager_test.go
```diff
@@ -859,98 +886,412 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
...
 		re.Equal(uint64(5), res.OldTxnSafePoint)
 		re.Equal(uint64(10), res.NewTxnSafePoint)
 		re.Empty(res.BlockerDescription)
-		res, err = s.manager.AdvanceTxnSafePoint(keyspaceID, 15, time.Now())
+		res, err = s.manager.AdvanceTxnSafePoint(keyspaceID, 15, now)
 		re.NoError(err)
 		re.Equal(uint64(10), res.OldTxnSafePoint)
 		re.Equal(uint64(10), res.NewTxnSafePoint)
 		re.Equal(uint64(15), res.Target)
+		re.Contains(res.BlockerDescription, "BarrierID: \"b1\"")
```
MyonKeminta: How about ensuring it's recorded as a `GlobalGCBarrier` instead of other type:
```suggestion
		re.Contains(res.BlockerDescription, "GlobalGCBarrier { BarrierID: \"b1\"")
```

pkg/gc/gc_state_manager_test.go
```diff
@@ -859,98 +886,412 @@ func (s *gcStateManagerTestSuite) TestGCBarriers() {
...
+	re.Equal(uint64(220), res.NewTxnSafePoint)
+	re.Contains(res.BlockerDescription, "BarrierID: \"b3\"")
+
+	// Clear.
+	_, err = s.manager.DeleteGlobalGCBarrier(context.Background(), "b1")
+	re.NoError(err)
+	_, err = s.manager.DeleteGCBarrier(ks1, "b2")
+	re.NoError(err)
+	_, err = s.manager.DeleteGCBarrier(ks2, "b3")
+	re.NoError(err)
```
MyonKeminta: The clean-up is not necessary.

pkg/gc/gc_state_manager_test.go
```diff
@@ -1128,6 +1469,66 @@ func (s *gcStateManagerTestSuite) TestServiceGCSafePointCompatibility() {
...
+	_, _, err = s.manager.CompatibleUpdateServiceGCSafePoint("svc1", 0, -1, now)
+	re.NoError(err)
+
+	// CompatibleUpdateServiceGCSafePoint on native_br cannot succeed if any of the
+	// keyspace has larger txn safe point than the given service safe point value
+	_, err = s.manager.AdvanceTxnSafePoint(constant.NullKeyspaceID, 25, time.Now())
+	re.NoError(err)
+	_, updated, err = s.manager.CompatibleUpdateServiceGCSafePoint("native_br", 16, math.MaxInt64, now)
+	re.NoError(err)
+	re.False(updated) // the call failed, but this is not an error
```
MyonKeminta: The failure should be able to be checked by comparing the returned `minSsp` and the specified value (16). You should assert `minSsp > 16` here, or further, just assert `minSsp = 25` which is the maximum txn safe point at that time.

pkg/gc/gc_state_manager_test.go
```diff
@@ -1128,6 +1469,66 @@ func (s *gcStateManagerTestSuite) TestServiceGCSafePointCompatibility() {
...
+	_, err = s.manager.AdvanceTxnSafePoint(2, 22, time.Now())
+	re.NoError(err)
+	re.False(updated)
```
MyonKeminta: Did you mean to check the `NewTxnSafePoint`?

pkg/gc/gc_state_manager_test.go
```diff
@@ -1128,6 +1469,66 @@ func (s *gcStateManagerTestSuite) TestServiceGCSafePointCompatibility() {
...
+	_, updated, err = s.manager.CompatibleUpdateServiceGCSafePoint("native_br", 20, math.MaxInt64, now)
+	re.NoError(err)
+	re.False(updated)
```
MyonKeminta: Ditto

pkg/gc/gc_state_manager_test.go
```diff
@@ -1321,6 +1722,32 @@ func (s *gcStateManagerTestSuite) TestGetGCState() {
...
+	re.Equal(uint64(0), txnSafePoint)
+
+	// change the value and check again
+	for i, keyspaceID := range s.keyspacePresets.manageable {
+		_, err = s.manager.AdvanceTxnSafePoint(keyspaceID, uint64(i+1), time.Now())
+		re.NoError(err)
+	}
+	err = s.provider.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
+		var err1 error
+		txnSafePoint, _, _, err1 = s.manager.getAllKeyspacesMaxTxnSafePoint(wb)
```
MyonKeminta: How about also try to check the other two return values?

pkg/storage/endpoint/gc_states.go
```diff
@@ -428,6 +526,22 @@ func (p GCStateProvider) CompatibleLoadAllServiceGCSafePoints() ([]string, []*Se
...
+
+	for i, barrier := range barriers {
+		keys = append(keys, keys1[i])
+		ssps = append(ssps, &ServiceSafePoint{
+			ServiceID:  barrier.BarrierID,
+			SafePoint:  barrier.BarrierTS,
+			ExpiredAt:  barrier.ExpirationTime.unixSeconds(),
+			KeyspaceID: constant.NullKeyspaceID,
+		})
+	}
```
MyonKeminta: No reason to make it visible when reading service safe points. What's more, except the special ID `native_br`, there's no way to guarantee there are no collision between global GC barrier IDs and non-global ones (service IDs). Please remove it.
And what's more *2, `GCStateProvider` as the storage layer is not designed to handle any detailed special compatibility behaviors, except for the service safe points <-> (non global) GC barriers conversion which is regarded as the internal mechanism of storage of GC barriers.

tests/server/gc/gc_test.go
```diff
@@ -243,12 +243,129 @@ func TestGCOperations(t *testing.T) {
 			continue
 		}
 		re.Equal(gcState.KeyspaceScope.KeyspaceId != constant.NullKeyspaceID, gcState.GetIsKeyspaceLevelGc())
-		re.Equal(uint64(15), gcState.GetTxnSafePoint())
+		// re.Equal(uint64(15), gcState.GetTxnSafePoint())
```
MyonKeminta: Why is this commented out?

MyonKeminta (approved):

pkg/storage/endpoint/gc_states.go
```diff
@@ -120,6 +120,79 @@ var (
 	_ json.Unmarshaler = (*ServiceSafePoint)(nil)
 )
 
+// TimeOptional is time.Time or nil
+type TimeOptional struct {
```
rleungx: How about moving it to pkg/utils/typeutil?

pkg/gc/gc_state_manager.go
```diff
@@ -694,7 +722,9 @@ func (m *GCStateManager) GetAllKeyspacesGCStates() (map[uint32]GCState, error) {
 //     the minimal service safe point. Returns a simulated service safe point whose serviceID starts with
 //     "tidb_min_start_ts_" to simulate the minimal service safe point. It may actually be either a GC barrier or
 //     a *TiDB min start ts*.
-//  2. If the given serviceID is anything else, it internally calls SetGCBarrier or DeleteGCBarrier, depending on
+//  2. If the given serviceID is "native_br" and keyspaceID is NullKeyspaceID, it is mapping to global GC barriers.
```
rleungx: Will the global GC barriers only be used by native br?

MyonKeminta: It's the only existing usage known to us that needs to block GC globally. `native_br` is treated as a spacial service ID of service safe points in serverless. This logic provides a path to keep compatible with their existing behavior.

MyonKeminta:
```suggestion
//  2. If the given serviceID is "native_br" and keyspaceID is NullKeyspaceID, it is mapped to global GC barriers.
```

rleungx: I would like to know if we can make it more general so that we can easily extend the usage.

tiancaiamao: Sure, you can call the `SetGlobalGCBarrier` API directly to create global GC barriers, and global GC barriers blocks the GC among all keyspaces. `SetGlobalGCBarrier` is new introduced API, feel free to use it.

Here the `CompatibleUpdateServiceGCSafePoint` is aimed to be compatible with old API behavior, and we have special requirement for "native_br" to be compatible with old API.

pkg/storage/endpoint/gc_states.go
tiancaiamao: done

rleungx (approved):

pkg/gc/gc_state_manager.go
```diff
@@ -346,6 +348,26 @@ func (m *GCStateManager) advanceTxnSafePointImpl(keyspaceID uint32, target uint6
...
+		globals, err2 := m.gcMetaStorage.LoadAllGlobalGCBarriers()
+		if err2 != nil {
+			return err2
+		}
+		for _, barrier := range globals {
+			if barrier.IsExpired(now) {
+				err1 = wb.DeleteGlobalGCBarrier(barrier.BarrierID)
+				if err1 != nil {
+					return err1
+				}
```
JmPotato, line 361: Would it be better to collect only the expired barriers here and delete them in a batch outside the loop?

tiancaiamao: It *is* a batch. @JmPotato 
This function is called inside a `RunInGCStateTransaction` function, and wb means `write batch`.

https://github.com/tikv/pd/pull/9340/files/4b2cb6e49c17e94e0ac6f5fa994fefc8224d7c9d#diff-78d46211e6cfafde196a36eb7ed67f85df599b946f1cea39e148f885a621d3a2L291

Issue #9351, opened 2025-05-26
Upgrade Etcd to v3.6.0


okJiang:
### Describe your feature request related problem

Etcd v3.6.0 has been released. https://etcd.io/blog/2025/announcing-etcd-3.6/

Its performance has been significantly improved 

> In this release, we reduced average memory consumption by at least 50%
> Compared to v3.5, etcd v3.6 delivers an average performance improvement of approximately 10% in both read and write throughput

In addition, this [feature](https://github.com/etcd-io/etcd/pull/14066) is helpful for PD. Currently, the middleware on the PD server side exists in each interface, which is very difficult to manage. The middleware implemented by PD itself cannot cover the interface of the etcd server.

### Describe the feature you'd like

Upgrade Etcd to v3.6.0 

### Teachability, Documentation, Adoption, Migration Strategy

As https://etcd.io/blog/2025/announcing-etcd-3.6/#upgrade-issue mentioned, 

> Upgrade issue
This section highlights a common issue [issues/19557](https://github.com/etcd-io/etcd/issues/19557) in the etcd v3.5 to v3.6 upgrade that may cause the upgrade process to fail. For a complete upgrade guide, refer to [Upgrade etcd from v3.5 to v3.6](https://etcd.io/docs/v3.6/upgrades/upgrade_3_6/).
> 
> The issue was introduced in etcd v3.5.1, and resolved in v3.5.20.
> 
> Key takeaway: users are required to first upgrade to etcd v3.5.20 (or a higher patch version) before upgrading to etcd v3.6.0; otherwise, the upgrade may fail.

We need upgrade to etcd v3.5.20 firstly.

rleungx:
In our case, it seems we are not affected by the issue.

okJiang:
> In our case, it seems we are not affected by the issue.

Indeed, usually, PD will only have three nodes. However, users may still scale out PD.

PR #9352, opened 2025-05-26
dep: upgrade etcd to 3.5.20

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/9351

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```


lhy1024 (approved):

go.mod
```diff
@@ -52,10 +52,10 @@ require (
 	github.com/syndtr/goleveldb v1.0.1-0.20190318030020-c3a204f8e965
 	github.com/unrolled/render v1.0.1
 	github.com/urfave/negroni v0.3.0
-	go.etcd.io/etcd/api/v3 v3.5.15
-	go.etcd.io/etcd/client/pkg/v3 v3.5.15
-	go.etcd.io/etcd/client/v3 v3.5.15
-	go.etcd.io/etcd/server/v3 v3.5.15
+	go.etcd.io/etcd/api/v3 v3.5.20
```
rleungx, line 55: Does it mean we should upgrade PD to this version first? So we need to wait for at least one release version, then we can upgrade to v3.6?

okJiang: Maybe we should wait a LTS version 🤔

okJiang:
The ci failed because timeout during running `regionTestSuite`(exceed 5min).

I think it is because this [commit](https://github.com/etcd-io/etcd/pull/19167) added a `wg.Wait()` inside etcd.Stop(). So it need to consume more time(10+s for each test case) to exit etcd server.

Maybe we have some operation that caused this. I don't find the reason yet.

okJiang:
The demand is not strong enough

PR #9354, opened 2025-05-26
feat(resourcemanager): add service limit for keyspace

JmPotato:

Issue Number: ref #9296.

```commit-message
Add a keyspace-level service limiter to control the maximum request rate for each keyspace.
The service limit is an unburstable RU per second rate limit that applies across all resource groups within a keyspace.
```

### Check List

Tests

- Unit test
- Integration test

With a resource group rg1 whose RU is 5000 and is non-burstable, set the service limit to 10000 RU > 5000 RU:

image

image

Set the service limit to 2500 RU:

image

image

rg1 5000 RU + rg2 2500 RU with 2500 RU service limit.

image

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/apis/v1/api.go
+++ b/pkg/mcs/resourcemanager/server/apis/v1/api.go
@@ -267,10 +267,6 @@ func (s *Service) setKeyspaceServiceLimit(c *gin.Context) {
 		c.String(http.StatusBadRequest, err.Error())
 		return
 	}
-	if keyspaceIDValue == nil {
-		c.String(http.StatusNotFound, fmt.Sprintf("keyspace not found with name: %s", keyspaceName))
-		return
-	}
 	var req KeyspaceServiceLimitRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
 		c.String(http.StatusBadRequest, err.Error())
@@ -291,16 +287,13 @@ func (s *Service) setKeyspaceServiceLimit(c *gin.Context) {
 //	@Param		keyspace_name	path		string	true	"Keyspace name"
 //	@Success	200				{string}	json	format	of	rmserver.serviceLimiter
 //	@Failure	400				{string}	error
+//	@Failure	404				{string}	error
 //	@Router		/config/keyspace/service-limit/{keyspace_name} [get]
 func (s *Service) getKeyspaceServiceLimit(c *gin.Context) {
 	keyspaceName := c.Param("keyspace_name")
 	keyspaceIDValue, err := s.manager.GetKeyspaceIDByName(c, keyspaceName)
 	if err != nil {
 		c.String(http.StatusBadRequest, err.Error())
-		return
-	}
-	if keyspaceIDValue == nil {
-		c.String(http.StatusNotFound, fmt.Sprintf("keyspace not found with name: %s", keyspaceName))
 		return
 	}
 	keyspaceID := keyspaceIDValue.GetValue()
--- a/pkg/mcs/resourcemanager/server/apis/v1/api.go
+++ b/pkg/mcs/resourcemanager/server/apis/v1/api.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/gin-contrib/cors"
@@ -36,15 +37,14 @@ import (
 	"github.com/tikv/pd/pkg/utils/reflectutil"
 )
 
-// APIPathPrefix is the prefix of the API path.
-const APIPathPrefix = "/resource-manager/api/v1/"
+const apiPathPrefix = "/resource-manager/api/v1/"
 
 var (
 	apiServiceGroup = apiutil.APIServiceGroup{
 		Name:       "resource-manager",
 		Version:    "v1",
 		IsCore:     false,
-		PathPrefix: APIPathPrefix,
+		PathPrefix: apiPathPrefix,
 	}
 )
 
@@ -78,7 +78,7 @@ func NewService(srv *rmserver.Service) *Service {
 	apiHandlerEngine.GET("metrics", utils.PromHandler())
 	apiHandlerEngine.GET("status", utils.StatusHandler)
 	pprof.Register(apiHandlerEngine)
-	endpoint := apiHandlerEngine.Group(APIPathPrefix)
+	endpoint := apiHandlerEngine.Group(apiPathPrefix)
 	endpoint.Use(multiservicesapi.ServiceRedirector())
 	s := &Service{
 		manager:          manager,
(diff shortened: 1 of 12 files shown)


pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -0,0 +1,137 @@
...
+	if elapsed < 0 {
+		return
+	}
+	// Add tokens based on the configured rate and the burst limit.
+	krl.AvailableTokens = math.Min(
+		krl.AvailableTokens+krl.ServiceLimit*elapsed,
+		krl.ServiceLimit*serviceLimiterBurstFactor,
+	)
+	// Update the last update time.
+	krl.LastUpdate = now
```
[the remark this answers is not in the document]
JmPotato: `refillTokensLocked` is widely used in the unit tests, moving this line out will cause us to manually update `LastUpdate` in the test.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -107,6 +107,24 @@ func (m *Manager) GetStorage() endpoint.ResourceGroupStorage {
 	return m.storage
 }
 
+// GetKeyspaceServiceLimiter returns the service limit of the keyspace.
+func (m *Manager) GetKeyspaceServiceLimiter(keyspaceID uint32) *serviceLimiter {
+	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
```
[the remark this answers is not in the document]
JmPotato: Considering that a GET request should be idempotent, I did not modify the behavior of `GetKeyspaceServiceLimit`, but for `SetKeyspaceServiceLimit`, now if the corresponding keyspace manager does not exist, it will be created.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -116,6 +119,25 @@ func (m *Manager) GetStorage() endpoint.ResourceGroupStorage {
...
+		return nil
+	}
+	return krgm.getServiceLimiter().Clone()
+}
+
+// SetKeyspaceServiceLimit sets the service limit of the keyspace.
+func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float64) {
+	// If the keyspace is not found, create a new keyspace resource group manager.
+	krgm := m.getOrCreateKeyspaceResourceGroupManager(keyspaceID)
+	if krgm == nil {
```
glorv, line 135: Seems this is an unreachable path as `getOrCreateKeyspaceResourceGroupManager` should never fail. If it can actually retrun `nil` value, I think you should return an error as `SetKeyspaceServiceLimit` is not successful here.

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -239,3 +245,69 @@ func (s *Service) setControllerConfig(c *gin.Context) {
...
+	c.String(http.StatusOK, "Success!")
+}
+
+// GetKeyspaceServiceLimit
+//
+//	@Tags		ResourceManager
+//	@Summary	Get the service limit of the keyspace. If the keyspace name is empty, it will return the service limit of the null keyspace.
+//	@Param		keyspace_name	path		string	true	"Keyspace name"
+//	@Success	200				{string}	json	format	of	rmserver.serviceLimiter
+//	@Failure	400				{string}	error
```
rleungx, line 293: 404 also needed

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -239,3 +245,69 @@ func (s *Service) setControllerConfig(c *gin.Context) {
...
+//	@Router		/config/keyspace/service-limit/{keyspace_name} [get]
+func (s *Service) getKeyspaceServiceLimit(c *gin.Context) {
+	keyspaceName := c.Param("keyspace_name")
+	keyspaceIDValue, err := s.manager.GetKeyspaceIDByName(c, keyspaceName)
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+	if keyspaceIDValue == nil {
```
rleungx, line 302: BTW, It seems this situation will never happen?

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -99,6 +99,12 @@ func (s *Service) RegisterRouter() {
 	configEndpoint.DELETE("/group/:name", s.deleteResourceGroup)
 	configEndpoint.GET("/controller", s.getControllerConfig)
 	configEndpoint.POST("/controller", s.setControllerConfig)
+	// Without keyspace name, it will get/set the service limit of the null keyspace.
```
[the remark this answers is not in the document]
JmPotato: I don't want to make the null keyspace too special within the system; also it's useful for the test's convience.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -391,10 +411,50 @@ func (m *Manager) getKeyspaceNameByID(ctx context.Context, id uint32) (string, e
 		return "", fmt.Errorf("got an empty keyspace name by id %d", id)
 	}
 	// Update the cache.
+	m.Lock()
```
rleungx, line 414: How about using a separate function to avoid forgetting to update one of them?

pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (krl *serviceLimiter) refillTokensLocked(now time.Time) {
+	// No limit configured, do nothing.
+	if krl.ServiceLimit <= 0 {
+		return
+	}
+	// Calculate the elapsed time since the last update.
+	elapsed := now.Sub(krl.LastUpdate).Seconds()
+	if elapsed < 0 {
```
rleungx, line 81: Do we need a log here?

pkg/mcs/resourcemanager/server/apis/v1/api.go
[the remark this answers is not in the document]
glorv: +1

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -391,10 +411,51 @@ func (m *Manager) getKeyspaceNameByID(ctx context.Context, id uint32) (string, e
 		return "", fmt.Errorf("got an empty keyspace name by id %d", id)
 	}
 	// Update the cache.
-	m.keyspaceNameLookup[id] = loadedName
+	m.updateKeyspaceNameLookup(id, loadedName)
```
lhy1024, line 414: Is it possible to change the keyspace name? If keyspace is modified or deleted, do we need to update it?

JmPotato: Currently, once a keyspace is created, its ID and name cannot be changed. Therefore, we don’t need to consider cache misses at this stage.

JmPotato: Currently, once a keyspace is created, its ID and name cannot be changed. Therefore, we don’t need to consider cache misses at this stage.

Issue #9366, opened 2025-05-28
Hot cache won't expired even if the leader is changed


rleungx:
## Bug Report

We only update the peer if there is a change from the region heartbeat. 
https://github.com/tikv/pd/blob/master/pkg/statistics/hot_peer_cache.go#L139

However, if the leader changes, there is no check for hot read cache until the store heartbeat comes. We should also update the cache to avoid using the stale information of the hot leader.

PR #9372, opened 2025-05-30
hot_region: check leader during handle expired item

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9366

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/statistics/hot_peer_cache.go
+++ b/pkg/statistics/hot_peer_cache.go
@@ -157,6 +157,9 @@ func (f *HotPeerCache) CollectExpiredItems(region *core.RegionInfo) []*HotPeerSt
 
 // CheckLeaderChange checks whether the leader of the region has changed.
 func (f *HotPeerCache) CheckLeaderChange(region *core.RegionInfo) []*HotPeerStat {
+	if f.kind == utils.Write {
+		return nil
+	}
 	regionID := region.GetID()
 	newLeader := region.GetLeader().GetStoreId()
 	items := make([]*HotPeerStat, 0)
@@ -167,12 +170,14 @@ func (f *HotPeerCache) CheckLeaderChange(region *core.RegionInfo) []*HotPeerStat
 				if item != nil {
 					item.actionType = utils.Update
 					item.isLeader = false
+					item.lastTransferLeaderTime = time.Now()
 					items = append(items, item)
 				}
 				item = f.getOldHotPeerStat(regionID, newLeader)
 				if item != nil {
 					item.actionType = utils.Update
 					item.isLeader = true
+					item.lastTransferLeaderTime = time.Now()
 					items = append(items, item)
 				}
 			}
--- a/pkg/statistics/hot_peer_cache_test.go
+++ b/pkg/statistics/hot_peer_cache_test.go
@@ -97,7 +97,9 @@ func checkFlow(cache *HotPeerCache, region *core.RegionInfo, peers []*metapb.Pee
 	reportInterval := region.GetInterval()
 	interval := reportInterval.GetEndTimestamp() - reportInterval.GetStartTimestamp()
 	res = append(res, cache.CollectExpiredItems(region)...)
-	res = append(res, cache.CheckLeaderChange(region)...)
+	if cache.kind == utils.Read {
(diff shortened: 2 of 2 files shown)


pkg/statistics/hot_peer_cache.go
```diff
@@ -140,8 +142,8 @@ func (f *HotPeerCache) CollectExpiredItems(region *core.RegionInfo) []*HotPeerSt
 	regionID := region.GetID()
 	items := make([]*HotPeerStat, 0)
 	if ids, ok := f.storesOfRegion[regionID]; ok {
-		for storeID := range ids {
-			if region.GetStorePeer(storeID) == nil {
+		for storeID, isLeader := range ids {
```
rleungx: Will `justTransferLeader` be invalid?

okJiang: Made some changes, ptal again https://github.com/tikv/pd/pull/9372/commits/9e2197461dbe023cdd61601881de49e5b9858d81

pkg/cluster/cluster.go
```diff
@@ -51,8 +51,16 @@ func HandleStatsAsync(c Cluster, region *core.RegionInfo) {
...
+	checkLeaderChangeTask := func(cache *statistics.HotPeerCache) {
+		leaderChangedStats := cache.CheckLeaderChange(region)
+		for _, stat := range leaderChangedStats {
+			cache.UpdateStat(stat)
+		}
+	}
+
 	c.GetHotStat().CheckWriteAsync(checkExpiredTask)
 	c.GetHotStat().CheckReadAsync(checkExpiredTask)
+	c.GetHotStat().CheckReadAsync(checkLeaderChangeTask)
```
okJiang, line 63: Should we check it for Write?

rleungx: Is it possible to reuse checkWritePeerTask?

okJiang: I'm not sure if need to do this check with `write`.

okJiang: I think `justTransferLeader` handles the leader change case in `CheckPeerFlow`, so it shouldn't run `checkLeaderChangeTask` on Write anymore. 

Maybe we should strip out the relevant logic from `CheckPeerFlow`. But the code(`CheckPeerFlow`) is kind of crappy right now, and I don't really think of a better way to abstract it away.

pkg/statistics/hot_peer_cache.go
```diff
@@ -153,6 +155,37 @@ func (f *HotPeerCache) CollectExpiredItems(region *core.RegionInfo) []*HotPeerSt
 	return items
 }
 
+// CheckLeaderChange checks whether the leader of the region has changed.
+func (f *HotPeerCache) CheckLeaderChange(region *core.RegionInfo) []*HotPeerStat {
```
lhy1024, line 159: Do we need to add a unit for it?

pkg/cluster/cluster.go
lhy1024: It may need to be checked with write.

okJiang: @lhy1024 Did you see this reply? https://github.com/tikv/pd/pull/9372#discussion_r2125892035

pkg/statistics/hot_peer_cache.go
okJiang: I think it has been added in [hot_peer_cache_test.go](https://github.com/tikv/pd/pull/9372/files#diff-18a77c47ea41755636f2b09a6b8cb7fd9624586d42e7180335ba0507254fa3f0)? 🤔

pkg/cluster/cluster.go
lhy1024: You are right. We only need to solve hot read change by store heartbeat.

lhy1024 (approved):

pkg/cluster/cluster.go
okJiang: Agreed. I will keep this PR scoped to the hot read change triggered by store heartbeat only. The current patch already limits `CheckLeaderChange` to the read path, so there is no further code change to make for this thread.

PR #9377, opened 2025-06-04
metrics: record evict store status which is slow

### LINKED ISSUE #9391: Record the evict slow store status
## Enhancement Task
It would be better for diagnostics if we had metrics to record the current eviction status.

rleungx:

Issue Number: Close #9391

```commit-message
```

### Check List

Tests

- Unit test
- Manual test

Screenshot 2025-06-13 at 10 50 28

Screenshot 2025-06-13 at 11 02 33

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/schedulers/metrics.go
+++ b/pkg/schedule/schedulers/metrics.go
@@ -106,11 +106,11 @@ var (
 			Buckets:   prometheus.ExponentialBuckets(1, 2, 30),
 		}, []string{"type", "rw", "dim"})
 
-	storeSlowEvictedStatusGauge = prometheus.NewGaugeVec(
+	evictedSlowStoreStatusGauge = prometheus.NewGaugeVec(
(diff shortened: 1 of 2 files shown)


pkg/schedule/schedulers/metrics.go
```diff
@@ -106,6 +106,14 @@ var (
 			Buckets:   prometheus.ExponentialBuckets(1, 2, 30),
 		}, []string{"type", "rw", "dim"})
 
+	evictedSlowStoreStatusGauge = prometheus.NewGaugeVec(
```
okJiang, line 109: Do we need to add a panel to grafana?

rleungx: I think it's unnecessary. Usually, this metrics will be empty.

LykxSassinator (approved):

lhy1024 (approved):

PR #9381, opened 2025-06-05
feat(resourcemanager): add ruTracker for monitoring keyspace RU/s

JmPotato:

Issue Number: ref #9296.

```commit-message
This PR introduces a `ruTracker` to use a time-aware EMA algorithm to sample and calculate the real-time RU/s for each resource group.
It it critical to provide an accurate sampled RU/s for the priority scheduling under service limit, so this is a prerequisite work.
```

An illustration for the formual `e^{-β·Δt}`, which desribe the relation between decay factor and time delta:

### Check List

Tests

- Unit test
- Integration test

Test with:

- rg1: burstable 5000 RU/s
- rg2: non-burstable 2500 RU/s

截屏2025-06-05 17 42 25

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -254,8 +254,12 @@ func (krgm *keyspaceResourceGroupManager) getOrCreateRUTracker(name string) *ruT
 	rt := krgm.getRUTracker(name)
 	if rt == nil {
 		krgm.Lock()
-		rt = newRUTracker(defaultRUTrackerTimeConstant)
-		krgm.ruTrackers[name] = rt
+		// Double check the RU tracker is not created by other goroutine.
+		rt = krgm.ruTrackers[name]
+		if rt == nil {
+			rt = newRUTracker(defaultRUTrackerTimeConstant)
+			krgm.ruTrackers[name] = rt
+		}
 		krgm.Unlock()
 	}
 	return rt
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"math"
 	"sort"
+	"time"
 
 	"github.com/gogo/protobuf/proto"
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -240,3 +249,83 @@ func (krgm *keyspaceResourceGroupManager) getServiceLimiter() *serviceLimiter {
 	defer krgm.RUnlock()
 	return krgm.sl
 }
+
+func (krgm *keyspaceResourceGroupManager) getOrCreateRUTracker(name string) *ruTracker {
+	rt := krgm.getRUTracker(name)
+	if rt == nil {
```
bufferflies, line 258: It will set twice in the race.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -240,3 +249,83 @@ func (krgm *keyspaceResourceGroupManager) getServiceLimiter() *serviceLimiter {
...
+	// For example, if τ = 5s, then the decay factor calculated by e^{-β·Δt} will be 0.5 when Δt = 5s,
+	// which means the weight of the "old data" is 0.5 when the elapsed time is 5s.
+	beta           float64
+	lastSampleTime time.Time
+	lastEMA        float64
+}
+
+func newRUTracker(timeConstant time.Duration) *ruTracker {
+	return &ruTracker{
+		beta: math.Log(2) / timeConstant.Seconds(),
```
bufferflies, line 297: How about defining a new constant variable？

JmPotato: I don't think it's necessary since showing the formual is more readable and easy to understand.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -240,3 +252,91 @@ func (krgm *keyspaceResourceGroupManager) getServiceLimiter() *serviceLimiter {
...
+	defer rt.Unlock()
+	// If `dur` is not greater than 0, skip this record.
+	if dur <= 0 {
+		return
+	}
+	// Calculate the average RU/s within the `dur`.
+	ruPerSec := math.Max(0, totalRU) / dur.Seconds()
+	// If the last sample time is not set, set the last EMA directly.
+	if rt.lastSampleTime.IsZero() {
+		rt.lastEMA = ruPerSec
```
rleungx, line 316: Do we need to check minSampledRUPerSec here?

JmPotato: `ruPerSec` is calculated from actual consumption, so it is unlikely to be an extremely small value. I don't think it's necessary to perform additional checks here.

PR #9382, opened 2025-06-05
metrics(resourcemanager): use an embedded metrics collection

JmPotato:

Issue Number: ref #9296.

```commit-message
Using a global map during local integration tests may cause potential data races.
This PR replaces the global map with an embedded collection to prevent this issue.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager_test.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager_test.go
@@ -255,11 +255,6 @@ func TestGetResourceGroupList(t *testing.T) {
 	re.Len(groups, 4)
 	groups = krgm.getResourceGroupList(false, false)
 	re.Len(groups, 3)
-
-	names := krgm.getResourceGroupNames(true)
-	re.Len(names, 4)
-	names = krgm.getResourceGroupNames(false)
-	re.Len(names, 3)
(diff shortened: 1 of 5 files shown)


pkg/mcs/resourcemanager/server/keyspace_manager_test.go
```diff
@@ -255,11 +255,6 @@ func TestGetResourceGroupList(t *testing.T) {
 	re.Len(groups, 4)
 	groups = krgm.getResourceGroupList(false, false)
 	re.Len(groups, 3)
-
-	names := krgm.getResourceGroupNames(true)
-	re.Len(names, 4)
-	names = krgm.getResourceGroupNames(false)
-	re.Len(names, 3)
```
lhy1024, line 262: why remove it?

lhy1024 (approved):
the rest LGTM

pkg/mcs/resourcemanager/server/keyspace_manager_test.go
JmPotato: Because `getResourceGroupNames` method is removed.

rleungx (approved):

PR #9393, opened 2025-06-10
mcs: init default resource group name when access new keyspace

lhy1024:

Issue Number: ref https://github.com/tikv/pd/issues/9296.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -157,6 +157,20 @@ func (m *Manager) getKeyspaceResourceGroupManager(keyspaceID uint32) *keyspaceRe
 	return m.krgms[keyspaceID]
 }
 
+func (m *Manager) accessKeyspaceResourceGroupManager(keyspaceID uint32, groupName string) (*keyspaceResourceGroupManager, error) {
```
lhy1024, line 160: Do we need to add it to `GetKeyspaceServiceLimiter`?

glorv: No need, we can return "Not found" error if the keyspace is not inited.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -296,52 +310,52 @@ func (m *Manager) AddResourceGroup(grouppb *rmpb.ResourceGroup) error {
...
 	}
 	return krgm.modifyResourceGroup(grouppb)
 }
 
 // DeleteResourceGroup deletes a resource group.
 func (m *Manager) DeleteResourceGroup(keyspaceID uint32, name string) error {
-	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
-	if krgm == nil {
-		return errs.ErrKeyspaceNotExists.FastGenByArgs(keyspaceID)
+	krgm, err := m.accessKeyspaceResourceGroupManager(keyspaceID, name)
```
glorv: NIT: "default" group can't be deleted, so maybe there is not need to call `accessKeyspaceResourceGroupManager`

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -314,34 +329,34 @@ func (m *Manager) DeleteResourceGroup(keyspaceID uint32, name string) error {
 
 // GetResourceGroup returns a copy of a resource group.
 func (m *Manager) GetResourceGroup(keyspaceID uint32, name string, withStats bool) *ResourceGroup {
-	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
-	if krgm == nil {
+	krgm, err := m.accessKeyspaceResourceGroupManager(keyspaceID, name)
+	if err != nil {
```
rleungx, line 340: The error is ignored?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -157,6 +157,20 @@ func (m *Manager) getKeyspaceResourceGroupManager(keyspaceID uint32) *keyspaceRe
...
 }
 
+func (m *Manager) accessKeyspaceResourceGroupManager(keyspaceID uint32, groupName string) (*keyspaceResourceGroupManager, error) {
+	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
+	if krgm == nil && groupName == DefaultResourceGroupName {
+		// For the default resource group, if the keyspace manager doesn't exist yet
+		// and the group name is the default resource group name, we try to get or create it.
+		krgm = m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, true)
+		return krgm, nil
+	}
```
JmPotato: What about:

```suggestion
	var krgm *keyspaceResourceGroupManager
	if groupName == DefaultResourceGroupName {
		// For the default resource group, if the keyspace manager doesn't exist yet
		// and the group name is the default resource group name, we try to get or create it.
		krgm = m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, true)
	} else {
		krgm = m.getKeyspaceResourceGroupManager(keyspaceID)
	}
```

tests/integrations/mcs/resourcemanager/api_test.go
```diff
@@ -201,6 +201,59 @@ func (suite *resourceManagerAPITestSuite) TestResourceGroupAPI() {
 	}
 }
 
+func (suite *resourceManagerAPITestSuite) Test() {
```
JmPotato: The naming should be more specific.

tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -973,7 +973,7 @@ func (suite *resourceManagerClientTestSuite) TestBasicResourceGroupCURD() {
 				re.NoError(err)
 				re.Contains(dresp, "Success!")
 				_, err = cli.GetResourceGroup(suite.ctx, g.Name)
-				re.EqualError(err, fmt.Sprintf("get resource group %v failed, rpc error: code = Unknown desc = resource group not found", g.Name))
+				re.EqualError(err, fmt.Sprintf("get resource group %v failed, rpc error: code = Unknown desc = [PD:resourcemanager:ErrGroupNotExists]the %v resource group does not exist", g.Name, g.Name))
```
JmPotato, line 976: Why has the error message here changed? Can it be kept the same as before?

lhy1024: replace `errors.New("resource group not found")` with `errs.ErrResourceGroupNotExists.FastGenByArgs(req.ResourceGroupName)`

pkg/mcs/resourcemanager/server/grpc_service.go
```diff
@@ -107,7 +110,10 @@ func (s *Service) ListResourceGroups(_ context.Context, req *rmpb.ListResourceGr
 	if err := s.checkServing(); err != nil {
 		return nil, err
 	}
-	groups := s.manager.GetResourceGroupList(ExtractKeyspaceID(req.GetKeyspaceId()), req.WithRuStats)
+	groups, err := s.manager.GetResourceGroupList(ExtractKeyspaceID(req.GetKeyspaceId()), req.WithRuStats)
+	if err != nil {
```
rleungx, line 114: Do we need a status code?

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -223,6 +233,7 @@ func (s *Service) deleteResourceGroup(c *gin.Context) {
 	keyspaceID := rmserver.ExtractKeyspaceID(keyspaceIDValue)
 	if err := s.manager.DeleteResourceGroup(keyspaceID, c.Param("name")); err != nil {
 		c.String(http.StatusNotFound, err.Error())
```
JmPotato: Should this error also use `errs.ErrResourceGroupNotExists`?

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -200,7 +206,11 @@ func (s *Service) getResourceGroupList(c *gin.Context) {
 		return
 	}
 	keyspaceID := rmserver.ExtractKeyspaceID(keyspaceIDValue)
-	groups := s.manager.GetResourceGroupList(keyspaceID, withStats)
+	groups, err := s.manager.GetResourceGroupList(keyspaceID, withStats)
+	if err != nil {
+		c.String(http.StatusNotFound, err.Error())
```
JmPotato: Ditto.

lhy1024: It will return group list rather some group

pkg/mcs/resourcemanager/server/apis/v1/api.go
lhy1024: done

rleungx: seem DeleteResourceGroup will return ErrKeyspaceNotExists

lhy1024: > seem DeleteResourceGroup will return ErrKeyspaceNotExists

so replace it with `c.String(http.StatusNotFound, errs.ErrResourceGroupNotExists.FastGenByArgs(groupName).Error())`

rleungx: Shall we change the internal error? DeleteResourceGroup also has an error in the storage layer.

D3Hunter:
@lhy1024 please don't force push a reviewing PR, it's harder to review incrementally

@JmPotato ptal

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -176,6 +176,12 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
 	if name == DefaultResourceGroupName {
 		return errs.ErrDeleteReservedGroup
 	}
+	krgm.RLock()
+	if _, ok := krgm.groups[name]; !ok {
+		krgm.RUnlock()
+		return errs.ErrResourceGroupNotExists.FastGenByArgs(name)
+	}
+	krgm.RUnlock()
```
JmPotato:
```suggestion
	krgm.RLock()
	_, ok := krgm.groups[name]
	krgm.RUnlock()
	if !ok {
		return errs.ErrResourceGroupNotExists.FastGenByArgs(name)
	}
```

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -313,38 +329,38 @@ func (m *Manager) DeleteResourceGroup(keyspaceID uint32, name string) error {
...
-	return krgm.getResourceGroupList(withStats, true)
+	return krgm.getResourceGroupList(withStats, true), nil
 }
 
-func (m *Manager) getRUTracker(keyspaceID uint32, name string) *ruTracker {
-	krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
-	if krgm == nil {
-		return nil
+func (m *Manager) getRUTracker(keyspaceID uint32, name string) (*ruTracker, error) {
+	krgm, err := m.accessKeyspaceResourceGroupManager(keyspaceID, DefaultResourceGroupName)
```
JmPotato: I don't think we need to change this to `accessKeyspaceResourceGroupManager`, `getRUTracker` wil only be called when the consumption info is handled, which means the resource group definitely exist. Even if it doesn't, the nil case is enough to skip the later process. Introducing `DefaultResourceGroupName` here is also werid.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -496,8 +512,13 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
...
 			resourceGroupName := consumptionInfo.resourceGroupName
 			// TODO: maybe we need to distinguish background ru.
-			if rg := m.GetMutableResourceGroup(keyspaceID, resourceGroupName); rg != nil {
+			if rg, err := m.GetMutableResourceGroup(keyspaceID, resourceGroupName); rg != nil {
 				rg.UpdateRUConsumption(consumptionInfo.Consumption)
+			} else {
+				log.Error("failed to get mutable resource group",
+					zap.Uint32("keyspace-id", keyspaceID),
+					zap.String("resource-group-name", resourceGroupName),
+					zap.Error(err))
```
JmPotato: I feel that this log may not be very necessary. If the corresponding group is deleted during the workload, the reporting frequency might become very frequent due to the load. Perhaps deleting it or changing it to debug would be more appropriate.

JmPotato (commented):
The rest LGTM.

PR #9415, opened 2025-06-17
keyspace, tso: add system keyspace for next-gen

### LINKED ISSUE #9414: next-gen: add reserved keyspaces
## Development Task

Keyspace ID is allocated in PD, we better reserve some range of ID and name as reserved one,  at least for SYSTEM keyspace in next-gen version, :
- Reserve keyspace ID range [uint32.Max - 1024, uint32.Max) for internal usage
- Reserve SYSTEM keyspace name
- SYSTEM keyspace is created automatically when nextgen PD start, no need any special config, and its ID is fixed to 1.

lhy1024:

Issue Number: Close #9414

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/id/id.go
```diff
@@ -27,11 +27,14 @@ import (
...
 	"github.com/tikv/pd/pkg/utils/syncutil"
 	"github.com/tikv/pd/pkg/utils/typeutil"
+	"github.com/tikv/pd/pkg/versioninfo/kerneltype"
 )
 
 type label string
 
 const (
+	// ReservedKeyspaceIDEnd is the end of reserved keyspace IDs.
+	ReservedKeyspaceIDEnd = 1024
```
rleungx: Please add a comment to explain why we reserve these keyspace ID.

pkg/errs/errno.go
```diff
@@ -242,6 +242,8 @@ var (
 	ErrExceedMaxEtcdTxnOps = errors.Normalize("exceed max etcd txn operations", errors.RFCCodeText("PD:keyspace:ErrExceedMaxEtcdTxnOps"))
 	// ErrModifyDefaultKeyspace is used to indicate that default keyspace cannot be modified.
 	ErrModifyDefaultKeyspace = errors.Normalize("cannot modify default keyspace's state", errors.RFCCodeText("PD:keyspace:ErrModifyDefaultKeyspace"))
+	// ErrModifySystemKeyspace is used to indicate that system keyspace cannot be modified.
+	ErrModifySystemKeyspace = errors.Normalize("cannot modify system keyspace's state", errors.RFCCodeText("PD:keyspace:ErrModifySystemKeyspace"))
```
JmPotato, line 246: Just for reference, how about the name `ErrModifyReservedKeyspace`?

pkg/id/id.go
```diff
@@ -27,11 +27,15 @@ import (
...
 	"github.com/tikv/pd/pkg/utils/typeutil"
+	"github.com/tikv/pd/pkg/versioninfo/kerneltype"
 )
 
 type label string
 
 const (
+	// ReservedKeyspaceIDEnd is the end of reserved keyspace IDs.
+	// Reserve keyspace ID range [1, 1024] for internal usage.
+	ReservedKeyspaceIDEnd = uint64(1024)
```
JmPotato: Is it more appropriate to use reserved IDs at the end, such as [uint32.Max - 1025, uint32.Max), rather than starting from 1?

pkg/id/id.go
```diff
@@ -27,11 +29,18 @@ import (
 	"github.com/tikv/pd/pkg/utils/keypath"
 	"github.com/tikv/pd/pkg/utils/syncutil"
 	"github.com/tikv/pd/pkg/utils/typeutil"
+	"github.com/tikv/pd/pkg/versioninfo/kerneltype"
 )
 
 type label string
 
 const (
+	// ReservedKeyspaceIDCount is the reserved count for keyspace id.
```
lhy1024: Do we need to provide uint32.Max as the end of non-next-gen for keyspaces?

lhy1024: It will be fixed in another PR.

pkg/id/id_test.go
```diff
@@ -73,3 +76,66 @@ func testAllocator(re *require.Assertions, allocator Allocator) {
 		re.Equal(i, id)
 	}
 }
+
+// TestIDAllocationEndValue tests the start value for different label allocators.
```
JmPotato: The comment is incorrect, we only test the keyspace label allocator.

pkg/keyspace/keyspace.go
```diff
@@ -714,6 +731,12 @@ func (manager *Manager) UpdateKeyspaceStateByID(id uint32, newState keyspacepb.K
 		)
 		return nil, errs.ErrModifyDefaultKeyspace
 	}
+	if kerneltype.IsNextGen() && id == constant.SystemKeyspaceID {
```
JmPotato: Maybe we can extract these two checks into one function:

```go
func checkReservedID() error {
	...
}
```

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -668,6 +676,11 @@ func buildSplitKeyspaces(
 			// keyspace in the default keyspace group.
 			continue
 		}
+		if kerneltype.IsNextGen() && keyspace == constant.SystemKeyspaceID {
```
JmPotato: Ditto.

pkg/id/id_test.go
lhy1024: I forgot to update it, which shows even "start value"

pkg/id/id.go
```diff
@@ -90,6 +100,14 @@ func NewAllocator(params *AllocatorParams) Allocator {
 	if allocator.step == 0 {
 		allocator.step = defaultAllocStep
 	}
+	var effectiveEnd uint64
+	effectiveEnd = math.MaxUint64
```
rleungx: uint32?

lhy1024: Its type should be the same as `base` and `end`.

pkg/mcs/utils/constant/constant.go
```diff
@@ -47,6 +47,11 @@ const (
 	// 0 is reserved for default keyspace with the name "DEFAULT". It's initialized when PD bootstraps
 	// and reserved for users who haven't been assigned keyspace.
 	DefaultKeyspaceID = uint32(0)
+
+	// SystemKeyspaceID is the system keyspace ID.
+	SystemKeyspaceID = uint32(1)
```
rleungx: We still use 1 for system keyspace?

lhy1024: It may be more suitable to use uint32.Max - 1024 or uint32.Max - 1  or uint32.Max cc @D3Hunter @JmPotato

pkg/id/id.go
rleungx: What I mean is do we need to use `MaxKeyspaceID` instead of `math.MaxUint64` when label is KeyspaceLabel

lhy1024: It will be fixed in another PR to be convenient to cherry-pick. https://github.com/tikv/pd/pull/9423

pkg/keyspace/keyspace.go
```diff
@@ -149,32 +149,15 @@ func NewKeyspaceManager(
...
-	defaultKeyspaceMeta := &keyspacepb.KeyspaceMeta{
-		Id:             constant.DefaultKeyspaceID,
-		Name:           constant.DefaultKeyspaceName,
-		State:          keyspacepb.KeyspaceState_ENABLED,
-		CreatedAt:      now,
-		StateChangedAt: now,
-	}
-
-	config, err := manager.kgm.GetKeyspaceConfigByKind(endpoint.Basic)
+	err := manager.initReserveKeyspace(constant.DefaultKeyspaceID, constant.DefaultKeyspaceName)
```
rleungx, line 152: Do we still need it?

pkg/keyspace/keyspace.go
```diff
@@ -205,6 +188,34 @@ func (manager *Manager) Bootstrap() error {
...
+	// Split Keyspace Region for default/system keyspace.
+	if err := manager.splitKeyspaceRegion(id, false); err != nil {
+		return err
+	}
+	now := time.Now().Unix()
+	meta := &keyspacepb.KeyspaceMeta{
+		Id:             id,
+		Name:           name,
+		State:          keyspacepb.KeyspaceState_ENABLED,
+		CreatedAt:      now,
```
rleungx, line 201: The `CreatedAt` is no longer the same as `CreateKeyspaceRequest`.

pkg/keyspace/keyspace.go
lhy1024: Maybe we can handle the keyspace that the next gen doesn't need anymore in another pr.

pkg/keyspace/keyspace.go
lhy1024: What problems will it cause?

PR #9427, opened 2025-06-20
feat(resourcemanager): implement fill rate conciliation based on priority

JmPotato:

Issue Number: ref #9296.

```commit-message
Implement the fill rate conciliation based on resource group's priority. This ensure the RU priority scheduling within the service limit.
```

### Check List

Tests

- Unit test
- Integration test

Case 1: Two Resource Groups With the Same Priority

- rg1 RU_PER_SEC=2500 PRIORITY=MEDIUM BURSTABLE=ON
- rg2 RU_PER_SEC=5000 PRIORITY=MEDIUM BURSTABLE=OFF

Case 2: Two Resource Groups With Different Priorities

- rg1 RU_PER_SEC=2500 PRIORITY=MEDIUM BURSTABLE=ON
- rg2 RU_PER_SEC=5000 PRIORITY=HIGH BURSTABLE=OFF

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -251,6 +251,10 @@ func (krgm *keyspaceResourceGroupManager) setServiceLimit(serviceLimit float64) 
 	krgm.RUnlock()
 	// Set the new service limit to the limiter.
 	sl.setServiceLimit(serviceLimit)
+	// Cleanup the overrides if the service limit is set to 0.
+	if serviceLimit <= 0 {
+		krgm.cleanupOverrides()
+	}
 }
 
 func (krgm *keyspaceResourceGroupManager) getServiceLimiter() *serviceLimiter {
@@ -451,20 +455,21 @@ func (krgm *keyspaceResourceGroupManager) conciliateFillRates() {
 		if totalRUDemand >= remainingServiceLimit {
 			// If the basic RU demand already exceeds the remaining service limit, then we can only try to meet the basic needs of
 			// the resource groups through proportional allocation of the fill rate setting.
+			basicCapacity := remainingServiceLimit
 			if totalBasicRUDemand >= remainingServiceLimit {
 				for _, group := range queue {
 					basicRUDemand := basicRUDemandMap[group.Name]
 					// Allocate the remaining service limit proportionally based on basic demand.
 					overrideFillRate := math.Min(
-						remainingServiceLimit*basicRUDemand/totalBasicRUDemand,
+						basicCapacity*basicRUDemand/totalBasicRUDemand,
 						group.getFillRateSetting(), // Should not exceed the original fill rate setting.
 					)
 					// Do not allow the resource group to consume extra tokens in this case,
 					// so the override burst limit is set to the same as the override fill rate.
 					group.overrideFillRateAndBurstLimit(overrideFillRate, int64(overrideFillRate))
+					// Deduct the basic capacity from the remaining service limit.
+					remainingServiceLimit -= overrideFillRate
 				}
-				// After allocation, all remaining service limit is consumed.
-				remainingServiceLimit = 0
 			} else {
 				// If the basic RU demand can be fully met, we can further satisfy burst RU demands.
 				burstCapacity := remainingServiceLimit - totalBasicRUDemand
@@ -528,4 +533,12 @@ func (krgm *keyspaceResourceGroupManager) getPriorityQueues() [][]*ResourceGroup
 		n++
 	}
 	return priorityQueues[:n]
+}
+
+func (krgm *keyspaceResourceGroupManager) cleanupOverrides() {
+	krgm.RLock()
+	for _, group := range krgm.groups {
+		group.overrideFillRateAndBurstLimit(-1, -1)
+	}
+	krgm.RUnlock()
 }
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -346,3 +346,113 @@ func (rt *ruTracker) getRUPerSec() float64 {
 	defer rt.RUnlock()
 	return rt.lastEMA
 }
+
+// conciliateFillRates is used to conciliate the fill rate of each resource group.
+// Under the service limit, there might be multiple resource groups with different
+// priorities consuming the RU at the same time. In this case, we need to conciliate
+// the fill rate of the resource group to ensure the priority, for example, there are
+// three resource groups with priorities 1, 2, and 3, and the service limit is 50 RU/s:
+// - Group A: priority 1, fill rate 30 RU/s
+// - Group B: priority 2, fill rate 20 RU/s
+// - Group C: priority 3, fill rate 10 RU/s
+// In this case, the fill rate of the resource group should be 30 RU/s, 20 RU/s, and 0 RU/s,
+// respectively. The conciliation algorithm is as follows:
+//  1. Sort the resource groups by the priority.
+//  2. Start from the highest priority resource group, get the real-time RU/s of each.
+//  3. Calculate the total real-time RU/s of the resource groups within the same priority.
+//     a. If the total real-time RU/s is greater than the service limit,
+//     allocate the service limit to all resource groups based on the proportion of RU/s.
+//     b. If the total real-time RU/s is less or equal to the service limit,
+//     allocate the needed RU/s to the resource group respectively.
+//  4. Calculate the remaining service limit, and repeat the process for the next priority
+//     until the service limit is allocated.
+func (krgm *keyspaceResourceGroupManager) conciliateFillRates() {
+	serviceLimiter := krgm.getServiceLimiter()
+	// No need to conciliate if the service limit is not set or is 0.
+	if serviceLimiter == nil || serviceLimiter.ServiceLimit == 0 {
+		return
+	}
+	priorityQueues := krgm.getPriorityQueues()
+	if len(priorityQueues) == 0 {
+		return
+	}
+	remainingServiceLimit := serviceLimiter.ServiceLimit
+	for _, queue := range priorityQueues {
+		if len(queue) == 0 {
+			continue
+		}
+		totalRUPerSec := 0.0 // The total real-time RU/s of all resource groups in the current priority.
+		ruPerSecMap := make(map[string]float64, len(queue))
+		for _, group := range queue {
+			ruTracker := krgm.getRUTracker(group.Name)
+			// Not found the RU tracker, skip this group.
+			if ruTracker == nil {
+				continue
+			}
+			ruPerSec := ruTracker.getRUPerSec()
+			totalRUPerSec += ruPerSec
+			ruPerSecMap[group.Name] = ruPerSec
+		}
+		if totalRUPerSec > remainingServiceLimit {
+			// If the total real-time RU/s is greater than the service limit,
+			// allocate the service limit to all resource groups based on the proportion of RU/s.
+			for _, group := range queue {
+				ruPerSec := ruPerSecMap[group.Name]
+				overrideFillRate := math.Min(
+					remainingServiceLimit*ruPerSec/totalRUPerSec,
+					group.getFillRate(), // Should not exceed the fill rate setting.
+				)
+				group.overrideFillRate(overrideFillRate)
+			}
+			remainingServiceLimit = 0
+		} else {
+			// If the total real-time RU/s is less or equal to the service limit,
+			// allocate the needed RU/s to the resource group respectively.
+			for _, group := range queue {
+				// By setting the override fill rate to -1 to allow the resource group to consume
+				// as many RUs as they originally need according to its fill rate setting. This is
+				// to prevent the true demand from being suppressed due to setting `overrideFillRate`,
+				// thereby wasting idle resources of the Service Limit.
+				group.overrideFillRate(-1)
+			}
+			// Although this layer does not set `overrideFillRate`, it still needs to deduct the actual
+			// RU consumption from `remainingFillRate` to reflect the concept of priority, so that the
+			// available service limit share decreases level by level.
+			remainingServiceLimit -= totalRUPerSec
+		}
+	}
+}
+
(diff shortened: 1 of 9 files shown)


pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,127 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+		totalRUDemand := 0.0
+		ruDemandMap := make(map[string]float64, len(queue))
+		for _, group := range queue {
+			ruTracker := krgm.getRUTracker(group.Name)
+			// Not found the RU tracker, skip this group.
+			if ruTracker == nil {
+				continue
+			}
+			// The RU demand exceeding the fill rate setting should be ignored.
+			ruDemand := math.Min(ruTracker.getRUPerSec(), group.getFillRateSetting())
```
glorv: for a burstable group, the ru demand can be higher than the ru fill rate if there is still enough available ru

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,127 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+			remainingServiceLimit -= totalRUDemand
+		}
+	}
+}
+
+func (krgm *keyspaceResourceGroupManager) getPriorityQueues() [][]*ResourceGroup {
+	priorityQueues := make([][]*ResourceGroup, maxPriority+1)
+	krgm.RLock()
+	for _, group := range krgm.groups {
+		if group.Name == DefaultResourceGroupName {
```
glorv: Why bypass the default group? It should also be controlled especially when there are high priority groups.

glorv (commented):
rest LGTM

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,182 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+// This ensures:
+// - Higher priority groups get resources first
+// - Within same priority, resources are allocated proportionally based on actual demand
+// - Basic needs are prioritized over burst needs
+// - No group exceeds its configured limits
+// - Inactive groups (no RU consumption) are skipped to avoid unnecessary computation
+func (krgm *keyspaceResourceGroupManager) conciliateFillRates() {
+	serviceLimiter := krgm.getServiceLimiter()
+	// No need to conciliate if the service limit is not set or is 0.
+	if serviceLimiter == nil || serviceLimiter.ServiceLimit == 0 {
```
glorv, line 401: Is it possible the ServiceLimit is reset from a >0 value to 0, then we should also remove all the override fillrate here?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -307,37 +308,37 @@ func newRUTracker(timeConstant time.Duration) *ruTracker {
...
+	var dur time.Duration
+	if !rt.lastSampleTime.IsZero() {
+		dur = now.Sub(rt.lastSampleTime)
+	}
+	rt.lastSampleTime = now
 	// If `dur` is not greater than 0, skip this record.
 	if dur <= 0 {
 		return
 	}
+	durSeconds := dur.Seconds()
```
rleungx, line 328: Do we need to skip the duration that is less than a second?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,185 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+		return
+	}
+	remainingServiceLimit := serviceLimiter.ServiceLimit
+	for _, queue := range priorityQueues {
+		if len(queue) == 0 {
+			continue
+		}
+		// If the remaining service limit is 0, set the fill rate of all the remaining resource groups to 0 directly.
+		// Only apply this to active resource groups (those with RU tracker and actual RU consumption).
+		if remainingServiceLimit == 0 {
```
rleungx, line 415: How about moving out `remainingServiceLimit == 0` and directly return after override to 0?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,185 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+					overrideFillRate := math.Min(
+						remainingServiceLimit*basicRUDemand/totalBasicRUDemand,
+						group.getFillRateSetting(), // Should not exceed the original fill rate setting.
+					)
+					// Do not allow the resource group to consume extra tokens in this case,
+					// so the override burst limit is set to the same as the override fill rate.
+					group.overrideFillRateAndBurstLimit(overrideFillRate, int64(overrideFillRate))
+				}
+				// After allocation, all remaining service limit is consumed.
+				remainingServiceLimit = 0
```
rleungx: If `group.getFillRateSetting()` is less than `remainingServiceLimit*basicRUDemand/totalBasicRUDemand`, it might still have `remainingServiceLimit`?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,185 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+		}
+
+		totalRUDemand := totalBasicRUDemand + totalBurstRUDemand
+		// If the total real-time RU demand is greater than or equal to the remaining service limit,
+		// we need to divide the remaining service limit proportionally within the same priority level.
+		if totalRUDemand >= remainingServiceLimit {
+			// If the basic RU demand already exceeds the remaining service limit, then we can only try to meet the basic needs of
+			// the resource groups through proportional allocation of the fill rate setting.
+			if totalBasicRUDemand >= remainingServiceLimit {
+				for _, group := range queue {
```
glorv: Here you may need first sort the groups by normalizedRUDemand(RUDemand/FillRate) and handle groups by this order. So if some groups' normalizedRUDemand is smaller than totalBasicRUDemand/totalBasicRUDemand, that means it's required ru is smaller than its proportion, then we can retain this unused RUs to other needed groups.
Example,  say total limit is 1000 there are 3 groups with (ru_per_sec, ru_required) values (1000, 100), (1000, 400), (1000, 800). The ideal granted override limit should be 100, 400, 500. But in the algorithm in this PR, the override limit will be 100, 333, 333.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +347,185 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+					// Deduct the burst capacity from the remaining service limit.
+					remainingServiceLimit -= overrideBurstLimit
+				}
+			}
+		} else {
+			// If the total real-time RU demand is less than the remaining service limit, no need to divide the service limit,
+			// just set the override fill rate to -1 to allow the resource group to consume as many RUs as they originally
+			// need according to its fill rate setting.
+			for _, group := range queue {
+				group.overrideFillRateAndBurstLimit(-1, -1)
```
glorv, line 442: In the client implementation, if ru is burst, the ru consumption is unlimited. So we should avoid letting resource group to be burst. Thus, we better override the fillrate of burstable group with something like the service limit and make it not burstable.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: Since `durSeconds` is a float64, I don't think we need to skip it.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: Nice catch, fixed.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: The `remainingServiceLimit` can change during iteration at each level; for example, it won't reach zero until higher levels have exhausted it.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: Nice catch, added actual deduction logic.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: I think I can address this issue with a separate PR after merging this one.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -251,6 +251,11 @@ func (krgm *keyspaceResourceGroupManager) setServiceLimit(serviceLimit float64)
 	krgm.RUnlock()
 	// Set the new service limit to the limiter.
 	sl.setServiceLimit(serviceLimit)
+	// Cleanup the overrides if the service limit is set to 0.
+	if serviceLimit > 0 {
+		return
+	}
+	krgm.cleanupOverrides()
```
glorv:
```suggestion
	if serviceLimit <= 0 {
		krgm.cleanupOverrides()
	}
```

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +352,228 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+			fillRateSetting := group.getFillRateSetting()
+			burstLimitSetting := group.getBurstLimitSetting()
+			// Calculate the basic RU demand of the resource group.
+			// Basic demand is the minimum of real-time RU consumption and configured fill rate.
+			basicRUDemand := math.Min(ruPerSec, fillRateSetting)
+			totalBasicRUDemand += basicRUDemand
+			basicRUDemandMap[group.Name] = basicRUDemand
+			// Calculate the burst RU demand of the resource group.
+			// Burst demand is the excess consumption beyond the fill rate setting.
+			burstRUDemand := math.Max(0, ruPerSec-fillRateSetting)
```
glorv: If the group is not burstable, we should set burstDemand to 0

pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -373,15 +380,19 @@ func newGaugeMetrics(keyspaceName, groupName string) *gaugeMetrics {
...
 
 func (m *gaugeMetrics) setGroup(group *ResourceGroup) {
 	ru := math.Max(group.getRUToken(), 0)
 	m.availableRUCounter.Set(ru)
 	m.priorityResourceGroupConfigGauge.Set(group.getPriority())
-	m.ruPerSecResourceGroupConfigGauge.Set(group.getFillRate())
-	m.ruCapacityResourceGroupConfigGauge.Set(group.getBurstLimit())
+	m.ruPerSecResourceGroupConfigGauge.Set(group.getFillRateSetting())
+	m.ruCapacityResourceGroupConfigGauge.Set(group.getBurstLimitSetting())
+	m.overrideFillRateGauge.Set(group.getOverrideFillRate())
```
glorv: Is it better to delete the metric if the override rate is -1?

glorv (commented):
rest LGTM

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -79,6 +79,13 @@ type GroupTokenBucket struct {
...
 }
 
+func (gtb *GroupTokenBucket) getFillRate() float64 {
+	if gtb.overrideFillRate >= 0 {
+		return gtb.overrideFillRate
+	}
+	return float64(gtb.Settings.GetFillRate())
+}
+
 func (gtb *GroupTokenBucket) getFillRateSetting() float64 {
```
rleungx, line 89: Can we keep only one of them?

JmPotato:
Please take another look, thx~ @lhy1024

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +351,273 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+	remainingServiceLimit float64,
+) float64 {
+	remainingServiceLimit -= di.totalBasicRUDemand
+	// The remaining service limit is the burst capacity.
+	burstCapacity := remainingServiceLimit
+	for _, group := range queue {
+		burstRUDemand := di.burstRUDemandMap[group.Name]
+		fillRateSetting := group.getFillRate(true)
+		// Allocate the remaining service limit proportionally based on burst demand.
+		overrideBurstLimit := fillRateSetting + burstCapacity*burstRUDemand/di.totalBurstRUDemand
```
lhy1024, line 596: Do we need to check whether `di.totalBurstRUDemand` is zero?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -346,3 +351,273 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+func (krgm *keyspaceResourceGroupManager) cleanupOverrides() {
+	krgm.RLock()
+	groups := make([]*ResourceGroup, 0, len(krgm.groups))
+	for _, group := range krgm.groups {
+		groups = append(groups, group)
+	}
+	krgm.RUnlock()
+	// Cleanup the overrides for all the resource groups without holding the lock.
+	for _, group := range groups {
+		group.overrideFillRateAndBurstLimit(-1, -1)
```
lhy1024, line 621: Do we need to use a const value for `-1`?

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: No, if we reach here, it means `totalRUDemand >= remainingServiceLimit` and `di.totalBasicRUDemand < remainingServiceLimit`, which indicates that `di.totalBurstRUDemand` must be non-zero.

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: I don't think this is necessary.

lhy1024 (approved):

PR #9429, opened 2025-06-20
client: Add a helper function for determining whether a keyspace has keyspace level GC enabled

MyonKeminta:

Issue Number: Ref #8978

```commit-message
client: Add a helper function for determining whether a keyspace has keyspace level GC enabled
```

### Check List

Tests

- Unit test

Code changes

\-

Side effects

\-

Related changes

\-

### Release note

```release-note
None.
```


client/keyspace_client.go
```diff
@@ -27,6 +27,15 @@ import (
...
 )
 
+const (
+	// KeyspaceConfigGCManagementType is the key for GC management type in keyspace config.
+	KeyspaceConfigGCManagementType = "gc_management_type"
+	// KeyspaceConfigGCManagementTypeKeyspaceLevel is the value representing keyspace level GC in keyspace config.
+	KeyspaceConfigGCManagementTypeKeyspaceLevel = "keyspace_level"
+	// KeyspaceConfigGCManagementTypeUnified is the value representing unified GC in keyspace config.
+	KeyspaceConfigGCManagementTypeUnified = "unified"
+)
```
JmPotato, line 37: Do these constants need to be exported?

JmPotato (commented):
The rest LGTM.

client/keyspace_client.go
MyonKeminta: Currently not. I'm considering if later there's some new cases where these constants are really needed (e.g., writing the configuration for a new keyspace), we can use this directly, without having to re-define them or modify these code to export it.
🤔 if you mind it, it's ok to make them private for now.

JmPotato (approved):

PR #9435, opened 2025-06-23
gc,metrics: add metrics for GC barrier

tiancaiamao:

Issue Number: ref https://github.com/pingcap/tidb/issues/58720

The new GC API `SetBarrier` is added, but there is still no metrics for it.

### Check List

Tests

- Unit test
- Integration test
- [X] Manual test (add detailed scripts or steps below)

image

- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -510,7 +510,8 @@ func (m *GCStateManager) setGCBarrierImpl(keyspaceID uint32, barrierID string, b
 		zap.String("barrier-id", barrierID), zap.Uint64("barrier-ts", barrierTS), zap.Duration("ttl", ttl),
 		zap.Stringer("new-gc-barrier", newBarrier))
 
-	gcBarrierGauge.WithLabelValues(strconv.Itoa(int(keyspaceID)), barrierID).Set(float64(newBarrier.BarrierTS))
+	barrierTSMilliSec := newBarrier.BarrierTS >> 18
+	gcBarrierGauge.WithLabelValues(strconv.Itoa(int(keyspaceID)), barrierID).Set(float64(barrierTSMilliSec))
 
 	return newBarrier, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/gc/gc_state_manager.go
```diff
@@ -360,6 +368,11 @@ func (m *GCStateManager) advanceTxnSafePointImpl(keyspaceID uint32, target uint6
 	if err != nil {
 		return AdvanceTxnSafePointResult{}, err
 	}
+	if len(barrierExpired) > 0 {
+		for _, barrier := range barrierExpired {
+			gcBarrierGauge.DeleteLabelValues(strconv.Itoa(int(barrier.keyspaceID)), barrier.barrierID)
+		}
+	}
```
MyonKeminta, line 375: Is it necessary to record a complete list of GC barriers? How about only showing the minimal one?
And another potential issue is that, according to my experiment, grafana losts the precision of timestamps, which are 64bit integers. It's even worse that the `Set` function of the gauge in golang accepts fload64 values instead of integers, making the precision losing happens even before being uploaded to prometheus.
As the precision of timestamps are quite important, I'm afraid it's not a good idea to upload the timestamp itself. It's impreciseness may cause misleading in problem investigations. Instead, please consider extracting the physical part from the ts, and upload only the physical. People who need the precise values should search in the logs.

tiancaiamao: I assume there won't be too many barriers.
Record the min barrier ts is not as useful, we have the TxnSafePoint metrics to see whether it's blocked.
And from the barrier here, I want to see what barrier is blocking TxnSafePoint.

For example, if we record all the barriers, it looks like:

```
                                                                                 --------  barrier4
           --------------- barrier1
                           -------------------- barrier2
                      -------------- barrier3
```

If we have just the minimal barrier TS, it become:

```
                                                                                 --------
           ------ 
                                              ----------
                      -------------- 
```

The range is incorrect, and hard to read epecally when we want to check historial metrics.

Issue #9430, opened 2025-06-20
TestRetryGetTSNotLeader is unstable


lhy1024:
### CI link
https://github.com/tikv/pd/actions/runs/15771236477/job/44459189501?pr=9393

lhy1024:
https://github.com/tikv/pd/actions/runs/15824985807/job/44603181272?pr=9436

lhy1024:
https://github.com/tikv/pd/actions/runs/15842064029/job/44656482613?pr=9436

PR #9438, opened 2025-06-24
tests: make TestRetryGetTSNotLeader stable

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/9430

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


tests/integrations/tso/client_test.go
```diff
@@ -673,36 +679,38 @@ func TestRetryGetTSNotLeader(t *testing.T) {
...
+	for range 10 {
 		err = pdLeader.ResignLeader()
 		re.NoError(err)
+		leaderName = pdCluster.WaitLeader()
+		re.NotEmpty(leaderName)
+		pdLeader = pdCluster.GetServer(leaderName)
 	}
 
+	// Make sure the lastTS is not empty
+	testutil.Eventually(re, func() bool {
```
rleungx: Why do we need Eventually?

lhy1024: Yes, it will make the test stable. According to my local test, if only one comparison is done, lastTs could still be 0.

tests/integrations/tso/client_test.go
```diff
@@ -673,36 +679,37 @@ func TestRetryGetTSNotLeader(t *testing.T) {
...
+				continue
 			}
+			ts := tsoutil.ComposeTS(physical, logical)
+			re.Less(lastTS, ts)
+			lastTS = ts
 		}
 	}(pdClient)
 
-	for range 5 {
+	for range 10 {
```
rleungx: How about reverting this change to reduce the test time?

bufferflies (approved):

PR #9441, opened 2025-06-24
keyspace: ensure keyspace id is less than 0xFFFFFF

lhy1024:

Issue Number: Close #xxx

```commit-message
```

Valid keyspace id range is [0, 0xFFFFFF](uint24max, or 16777215)

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/id/id.go
+++ b/pkg/id/id.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/mcs/utils/constant"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
 	"github.com/tikv/pd/pkg/utils/keypath"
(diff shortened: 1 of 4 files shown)


pkg/id/id.go
```diff
@@ -43,12 +44,11 @@ const (
 	// reservedKeyspaceIDCount is the reserved count for keyspace id.
 	reservedKeyspaceIDCount = uint64(1024)
 	// reservedKeyspaceIDStart is the start id for reserved keyspace id.
-	// The reserved keyspace id range is [uint32.Max - 1024, uint32.Max)
-	reservedKeyspaceIDStart = uint64(math.MaxUint32) - reservedKeyspaceIDCount
+	// The reserved keyspace id range is [0xFFFFFF - 1024, 0xFFFFFF)
```
D3Hunter, line 47: please add comment for why keyspaceID uses 24 bits

lhy1024: done

rleungx (approved):

pkg/id/id.go
lhy1024: add comments in MaxValidKeyspaceID and spaceIDBitSizeMax

PR #9444, opened 2025-06-25
*: move some keyspace constant to pkg `constant`

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/9414

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/utils/constant/constant.go
+++ b/pkg/mcs/utils/constant/constant.go
@@ -99,6 +99,11 @@ const (
 
 // only for next gen
 const (
+	// ReservedKeyspaceIDCount is the reserved count for keyspace id.
+	ReservedKeyspaceIDCount = uint64(1024)
+	// ReservedKeyspaceIDStart is the start id for reserved keyspace id.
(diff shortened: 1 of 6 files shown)


pkg/mcs/utils/constant/constant.go
```diff
@@ -99,6 +99,11 @@ const (
 
 // only for next gen
 const (
+	// ReservedKeyspaceIDCount is the reserved count for keyspace id.
+	ReservedKeyspaceIDCount = uint64(1024)
```
rleungx: Maybe it is not the right place. Non-MS may also use keyspace.

lhy1024: Do we also need to move https://github.com/tikv/pd/blob/a300e0126a2ec8c224da315cf93f57d8b45e0bd7/pkg/mcs/utils/constant/constant.go#L44-L60 to pkg/keyspace/constant.go ?

rleungx: LGTM

JmPotato (approved):

rleungx (approved):

Issue #9359, opened 2025-05-27
Enhance Slow Store Scheduler


okJiang:
### Describe your feature request related problem

Assume a store in the cluster is experiencing network jitter. This jitter causes region leaders in the store to be re-campaigned by raft on other stores, resulting in leader transfers. However, PD is unable to detect the network jitter on the TiKV store, because PD can recieve the store heartbeat in this network status. Consequently, the PD's balance leader scheduler schedules the leader back to the problematic TiKV, leading to increased latency in the cluster.

To address the above problem, we need to add a new network status feedback mechanism to PD, and avoid to transfer leaders to the problematic TiKV node.

okJiang:
## Task List

- [x] kvproto, https://github.com/pingcap/kvproto/pull/1324

- [x] pd 
    - [x] https://github.com/tikv/pd/pull/9451
    - [x] https://github.com/tikv/pd/pull/9702
    - [x] https://github.com/tikv/pd/pull/9716
- [x] tikv
    - [x] https://github.com/tikv/tikv/pull/18798
    - [x] https://github.com/tikv/tikv/pull/18525
    - [x] https://github.com/tikv/tikv/pull/18715


## Found bug

- [x] Recovery gap duration in PD should be calculated after tikv recovery https://github.com/tikv/pd/pull/9710
- [x] After PD leader transfer, network slow store will not be recovered https://github.com/tikv/pd/pull/9716
- [x] Should allow user set inspect_network_interval to zero. https://github.com/tikv/tikv/pull/18947
- [x] When two network/disk slow stores exist at the same time, avoid the brief rebalance during the intermediate switching. https://github.com/tikv/pd/pull/9739

okJiang:
Cherry-pick v8.5.5:

PD

- [x] https://github.com/tikv/pd/pull/9971
    - [x] https://github.com/tikv/pd/pull/8888
    - [x] https://github.com/tikv/pd/pull/9451
    - [x] https://github.com/tikv/pd/pull/9702
    - [x] https://github.com/tikv/pd/pull/9710
    - [x] https://github.com/tikv/pd/pull/9716
    - [x] https://github.com/tikv/pd/pull/9730
    - [x] https://github.com/tikv/pd/pull/9739
    - [x] https://github.com/tikv/pd/pull/9377
    - [x] https://github.com/tikv/pd/pull/9941

TiKV

- [x] https://github.com/tikv/tikv/pull/19151
[...]

PR #9451, opened 2025-06-26
scheduler: implement network slow store scheduler

okJiang:

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/schedulers/evict_slow_store.go
+++ b/pkg/schedule/schedulers/evict_slow_store.go
@@ -40,13 +40,16 @@ import (
 
 const (
 	slowStoreEvictThreshold = 100
-	// For network slow store, its slow scores include the scores from each tikv.
-	// These scores should meet the following conditions and then will be
-	// considered as a slow store:
-	// 1. One network slow score is greater than networkSlowStoreIssueThreshold.
-	// 2. After removing the maximum value, if the number of stores with a score
-	//    greater than networkSlowStoreFluctuationThreshold is greater than or equal
-	//    to 2, it is considered a slow store.
+	// For network slow store, its slow scores include the scores from each tikv(If the score
+	// is equal to 1, it will not report to PD). These scores should meet the following
+	// conditions and then will be considered as a slow store:
+	// 1. At least one network slow score is greater than networkSlowStoreIssueThreshold.
+	// 2. A network with a slow score exceeding networkSlowStoreIssueThreshold is considered
+	//    a problem network. Since it is impossible to directly identify which side of the
+	//    network has a problem, the score is removed from the stores on both sides.
+	// 3. After removing the problem network slow score, if the number of scores are greater
+	//    than or equal to networkSlowStoreFluctuationThreshold is greater than or equal
+	//    to 2 in one store, it is considered a slow store.
 	networkSlowStoreIssueThreshold       = 95 // Threshold for detecting network issues with a specific store
 	networkSlowStoreFluctuationThreshold = 10 // Threshold for detecting network fluctuations with other stores
 	slowStoreRecoverThreshold            = 1


pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +344,122 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		if countScoreGreaterThan(networkSlowScores, networkSlowStoreSecondThreshold) < 2 {
+			log.Info("network slow store is not slow enough, skip",
+				zap.Uint64("store-id", store.GetID()),
+				zap.Any("network-slow-score", store.GetNetworkSlowScore()))
+			continue
+		}
+
+		if len(networkSlowStores) >= int(s.conf.MaxNetworkSlowStore) {
+			failpoint.InjectCall("evictSlowStoreTriggerLimit")
+			slowStoreTriggerLimitGauge.WithLabelValues(strconv.FormatUint(store.GetID(), 10), string(networkSlowStore)).Set(1)
```
okJiang:
```suggestion
			slowStoreTriggerLimitGauge.WithLabelValues(strconv.FormatUint(store.GetID(), 10), string(networkSlowStore)).Inc()
```

okJiang (commented):
self review

.golangci.yml
```diff
@@ -242,7 +242,7 @@ linters:
             - ""
         - name: var-naming
           severity: warning
-          disabled: false
```
rleungx: Why disable it?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +344,122 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		}
+
+		networkSlowScores := store.GetNetworkSlowScores()
+		for storeID := range networkSlowStores {
+			delete(networkSlowScores, storeID)
+		}
+		if len(networkSlowScores) == 0 {
+			continue
+		}
+		if countScoreGreaterThan(networkSlowScores, networkSlowStoreFirstThreshold) < 1 {
```
rleungx: `countScoreGreaterThan() < 1` looks strange.

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -38,8 +38,25 @@ import (
...
-	slowStoreRecoverThreshold = 1
+	slowStoreEvictThreshold = 100
+	// For network slow store, its slow scores include the scores from each tikv.
+	// These scores should meet the following conditions and then will be
+	// considered as a slow store:
+	// 1. One network slow score is greater than networkSlowStoreFirstThreshold.
+	// 2. After removing the maximum value, if the number of stores with a score
+	//    greater than networkSlowStoreSecondThreshold is greater than or equal
+	//    to 2, it is considered a slow store.
+	networkSlowStoreFirstThreshold  = 95
```
rleungx: Can we use a more meaningful name?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +344,122 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
 // Schedule implements the Scheduler interface.
 func (s *evictSlowStoreScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
 	evictSlowStoreCounter.Inc()
+	s.scheduleNetworkSlowStore(cluster)
```
rleungx: Do we need to add a switch to disable it?

okJiang: When `MaxNetworkSlowStore` is 0, it is disable?

rleungx: If so, we can skip the useless logic inside scheduleNetworkSlowStore.

pkg/response/store.go
```diff
@@ -67,6 +67,7 @@ type StoreStatus struct {
 	WitnessCount       int                `json:"witness_count,omitempty"`
 	PendingPeerCount   int                `json:"pending_peer_count,omitempty"`
 	SlowScore          uint64             `json:"slow_score,omitempty"`
+	NetworkSlowScore   map[uint64]uint64  `json:"network_slow_score,omitempty"`
```
bufferflies: If the cluster is big,  the JSON of these properties may be too long. Most case, we don't need it

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -50,8 +67,10 @@ type evictSlowStoreSchedulerConfig struct {
 	lastSlowStoreCaptureTS time.Time
 	isRecovered            bool
 	// Duration gap for recovering the candidate, unit: s.
-	RecoveryDurationGap uint64   `json:"recovery-duration"`
-	EvictedStores       []uint64 `json:"evict-stores"`
+	RecoveryDurationGap uint64               `json:"recovery-duration"`
+	EvictedStores       []uint64             `json:"evict-stores"`
+	MaxNetworkSlowStore uint64               `json:"max-network-slow-store"`
+	NetworkSlowStores   map[uint64]time.Time `json:"network-slow-stores"`
```
bufferflies: How about adding more comments to describe this config?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +350,113 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
 // Schedule implements the Scheduler interface.
 func (s *evictSlowStoreScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
 	evictSlowStoreCounter.Inc()
+	if len(s.conf.getNetworkSlowStores()) != 0 || s.conf.getMaxNetworkSlowStore() != 0 {
```
rleungx: Why use `||`?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +350,113 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+				zap.Uint64("store-id", storeID),
+				zap.Error(err))
+		}
+		s.conf.Unlock()
+	}
+
+	// try to recover the network slow store if it is normal.
+	for storeID, startTime := range networkSlowStores {
+		store := cluster.GetStore(storeID)
+		if store == nil || store.IsRemoved() {
```
rleungx: If a store is removing, we can also delete it.

pkg/schedule/schedulers/evict_slow_store.go
rleungx: Shall we expose the config? If we expose it, how do we set it?

pkg/response/store.go
okJiang: Yes, we only send non-1 scores to PD. Scores equal to 1 don't need to be sent.

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +350,113 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+
+		// Check condition 1: At least one score >= networkSlowStoreIssueThreshold (95)
+		// This identifies a store with network issues to a specific peer
+		maxScoreStoreID := findMaxScoreStore(networkSlowScores, networkSlowStoreIssueThreshold)
+		if maxScoreStoreID == 0 {
+			continue
+		}
+
+		// Remove the max score store for condition 2
+		delete(networkSlowScores, maxScoreStoreID)
```
rleungx: Consider that we have removed the networkSlowScores from networkSlowScores, should we continue removing the max score?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +350,113 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		// Check condition 2: At least 2 scores >= networkSlowStoreFluctuationThreshold (10) after removing max
+		// This confirms the store has network fluctuations with other peers
+		if countScoresAboveThreshold(networkSlowScores, networkSlowStoreFluctuationThreshold) < 2 {
+			log.Info("network slow store is not slow enough, skip",
+				zap.Uint64("store-id", store.GetID()),
+				zap.Any("network-slow-score", store.GetNetworkSlowScores()))
+			continue
+		}
+
+		if len(networkSlowStores) >= int(s.conf.MaxNetworkSlowStore) {
```
rleungx: getMaxNetworkSlowStore

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Must be uppercase to persist. This is not exposed via pd-ctl

pkg/schedule/schedulers/evict_slow_store.go
okJiang: If a slow store already exists and has been detected by PD, and the user adjusts MaxNetworkSlowStore to 0, should we allow the slow store to recover?

pkg/response/store.go
bufferflies: Where does it implement?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -389,3 +537,42 @@ func newEvictSlowStoreScheduler(opController *operator.Controller, conf *evictSl
...
+	for _, score := range scores {
+		sum += score
+	}
+
+	return sum / uint64(len(scores))
+}
+
+// findMaxScoreStore finds the store with the highest score that meets the threshold
+// Returns 0 if no store meets the threshold
+func findMaxScoreStore(scores map[uint64]uint64, threshold uint64) uint64 {
```
bufferflies: How about moving these three functions to the store struct?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +350,113 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		}
+
+		if len(networkSlowStores) >= int(s.conf.MaxNetworkSlowStore) {
+			failpoint.InjectCall("evictSlowStoreTriggerLimit")
+			slowStoreTriggerLimitGauge.WithLabelValues(strconv.FormatUint(store.GetID(), 10), string(networkSlowStore)).Inc()
+			continue
+		}
+		log.Info("detected network slow store, start to pause scheduler",
+			zap.Uint64("store-id", store.GetID()),
+			zap.Any("network-slow-score", store.GetNetworkSlowScores()))
```
bufferflies: It may be too large if you print all slow score

pkg/schedule/schedulers/evict_slow_store.go
bufferflies: And can we change it by the api? If we can, we need to lock the NetworkSlowStores in scheduleNetworkSlowStore scope.

pkg/response/store.go
okJiang: It will be in tikv's next pr

pkg/schedule/schedulers/evict_slow_store.go
okJiang: updated to `[]uint64`

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Done

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Replace `MaxNetworkSlowStore` with a switch, and we do not need to consider multiple NetworkSlowStores now

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Its each get/set should be locked 🤔

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -38,20 +39,41 @@ import (
...
 	// Last timestamp of the chosen slow store for eviction.
-	lastSlowStoreCaptureTS time.Time
-	isRecovered            bool
+	lastSlowStoreCaptureTS     time.Time
+	isRecovered                bool
+	networkSlowStoreCaptureTSs map[uint64]time.Time
 	// Duration gap for recovering the candidate, unit: s.
-	RecoveryDurationGap uint64   `json:"recovery-duration"`
-	EvictedStores       []uint64 `json:"evict-stores"`
+	RecoveryDurationGap     uint64   `json:"recovery-duration"`
```
bufferflies: How about renaming this variable with the suffix sec, such as RecoverySec? and add one comment for the default value

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,151 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+
+		if calculateAvgScore(networkSlowScores) <= slowStoreRecoverThreshold && uint64(time.Since(startTime).Seconds()) >= recoveryGap {
+			s.conf.deleteNetworkSlowStore(storeID, cluster)
+		}
+	}
+
+	// If there are some slow stores recovery, we can pause leader transfer from existing slow stores.
+	if len(pausedNetworkSlowStores) < defaultMaxNetworkSlowStore && len(networkSlowStoreCaptureTSs) > 0 {
+		var slowStoreID uint64
+		for storeID := range networkSlowStoreCaptureTSs {
```
bufferflies, line 487: How about using `slowStoreID=networkSlowStoreCaptureTSs[0]`?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,151 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+				fluctuationCount++
+			}
+		}
+
+		// At least 2 scores >= networkSlowStoreFluctuationThreshold (10) after removing
+		// problematic network. This confirms the store has network fluctuations with other peers
+		if fluctuationCount >= 2 {
+			isSlowStore = true
+		}
+		if !isSlowStore {
```
bufferflies: Why not removing this slowStore if it's not slow?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,151 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		if len(networkSlowScores) <= 2 {
+			continue
+		}
+
+		var (
+			fluctuationCount int
+			isSlowStore      bool
+		)
+		// If the store's network from all other stores are problematic, it must be a slow store.
+		if len(problematicNetwork[store.GetID()]) >= len(stores)-1-len(networkSlowStoreCaptureTSs) {
```
bufferflies: Does it need to exclude some unhealthy stores, such as disconnected/down stores?

okJiang: I think as long as tikv can create a connection internally, we can continue to run according to the current logic. If tikv can no longer create a connection, then tikv will not return the score of the corresponding store.

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Did you mean remove it from `problematicNetwork`?

okJiang: In fact, at this point we still cannot completely determine that it is definitely not a slow store. Because there may be such a situation:

Assume store 1, store 2 and store 3 is a slow store, its score may be:

- store 1 -> store 2: 100 (will be filtered)
- store 1 -> store 3: 100 (will be filtered)
- store 1 -> store 4: 5
- store 1 -> store 5: 5

- store 2 -> store 1: 100 (will be filtered)
- store 2 -> store 3: 100 (will be filtered)
- store 2 -> store 4: 5
- store 2 -> store 5: 5

And another normal store(store 4) may be:

- store 4 -> store 1: 100
- store 4 -> store 2: 100
- store 4 -> store 3: 100
- store 4 -> store 5: 1

The store 1 and store 2 can't be view as slow stores now. If we remove them from `problematicNetwork` here, the score (4 ->1, 4-> 2) will not be filtered and it will lead to store 4 mistakenly identified as a slow node

pkg/schedule/schedulers/evict_slow_store.go
okJiang: `networkSlowStoreCaptureTSs[0]` is a Time 😂

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Even if we modify this in the code, we can no longer modify the json field. Do you think it is necessary to modify it?

bufferflies: Yes, you'll need to let the user know about this configuration unit through no-code.

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -102,6 +133,77 @@ func (conf *evictSlowStoreSchedulerConfig) evictStore() uint64 {
...
+		log.Warn("failed to persist evict slow store config",
+			zap.Uint64("store-id", storeID),
+			zap.Error(err))
+		conf.PausedNetworkSlowStores = oldValue
+	}
+	return err
+}
+
+func (conf *evictSlowStoreSchedulerConfig) deleteNetworkSlowStore(storeID uint64, cluster sche.SchedulerCluster) {
+	cluster.ResumeLeaderTransfer(storeID, constant.In)
```
bufferflies: How about doing it after saving success? Do we need to check the store in the evict-leader scheduler?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,151 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
 // Schedule implements the Scheduler interface.
 func (s *evictSlowStoreScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
 	evictSlowStoreCounter.Inc()
+	if s.conf.getEnableNetworkSlowStore() {
```
bufferflies, line 431: It means that the network scheduler has a higher priority than the disk if they both happen.

okJiang: I think it doesn't has priority. If they happen at the same time, even if store A is a NetworkSlowStore, and then store A is view as a DiskSlowStore too, it will be evicted leader yet.

pkg/schedule/schedulers/evict_slow_store.go
okJiang: I think we need a code that can set `pauseLeaderTransfer` multiple times. I will file an another issue to add it.

okJiang: https://github.com/tikv/pd/issues/9673

I will fix it ASAP

pkg/schedule/schedulers/evict_slow_store.go
bufferflies: The network will occupy the EvictedStores, so the diskSlowScheduler can't run due to EvictedStores is not nil

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,153 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		}
+		s.conf.addNetworkSlowStore(slowStoreID, cluster)
+	}
 
+	stores := cluster.GetStores()
+
+	// Find potential slow stores
+	problematicNetwork := make(map[uint64]map[uint64]struct{})
+	for _, store := range stores {
+		if _, exist := networkSlowStoreCaptureTSs[store.GetID()]; exist {
```
rleungx: Do we need a lock for networkSlowStoreCaptureTSs?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -102,15 +133,86 @@ func (conf *evictSlowStoreSchedulerConfig) evictStore() uint64 {
...
+		conf.PausedNetworkSlowStores = oldPausedNetworkSlowStores
+		// If failed to persist, we still need to remove the store from the networkSlowStoreCaptureTSs.
+	}
+}
+
+func (conf *evictSlowStoreSchedulerConfig) addNetworkSlowStore(storeID uint64, cluster sche.SchedulerCluster) {
+	log.Info("detected network slow store, start to pause scheduler",
+		zap.Uint64("store-id", storeID))
+	conf.networkSlowStoreCaptureTSs[storeID] = time.Now()
+	if err := cluster.PauseLeaderTransfer(storeID, constant.In); err != nil {
```
rleungx, line 185: What if the networkSlowStoreCaptureTSs has been set, but the pause leader failed?

okJiang: It will still be considered a slow store and will continue to try to pause it at the next schedule. Ref L465-L473

Image

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Seems you mistake this. Network will not occupy `EvictedStores`. There are two different fields

Image

I will add some comments to `EvictedStores`

pkg/schedule/schedulers/evict_slow_store.go
okJiang: I think, after https://github.com/tikv/pd/pull/9451/commits/3d1c14858243e1122bbc2b39599f05dab6ceb253, we don't need the lock. Every operation of `networkSlowStoreCaptureTSs` is in `scheduleNetworkSlowStore`

okJiang: Maybe lock will be needed after supporting modification of `EnableNetworkSlowStore` configuration, but not now.

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Done

okJiang:
ping @bufferflies @rleungx

pkg/schedule/schedulers/evict_slow_store.go
rleungx: Shall we persist first?

bufferflies (approved):

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,153 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		for storeID := range networkSlowStoreCaptureTSs {
+			delete(networkSlowScores, storeID)
+		}
+
+		if calculateAvgScore(networkSlowScores) <= slowStoreRecoverThreshold && uint64(time.Since(startTime).Seconds()) >= recoveryGap {
+			s.conf.deleteNetworkSlowStore(storeID, cluster)
+		}
+	}
+
+	// If there are some slow stores recovery, we can pause leader transfer from existing slow stores.
```
rleungx: Why does recovery need a pause leader transfer?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -308,7 +426,153 @@ func (s *evictSlowStoreScheduler) IsScheduleAllowed(cluster sche.SchedulerCluste
...
+		for storeID := range networkSlowStoreCaptureTSs {
+			slowStoreID = storeID
+			break
+		}
+		s.conf.addNetworkSlowStore(slowStoreID, cluster)
+	}
 
+	stores := cluster.GetStores()
+
+	// Find potential slow stores
```
rleungx: How about using a separate function?

pkg/schedule/schedulers/evict_slow_store.go
okJiang: @rleungx Regardless of whether persistence is successful or not, should we execute Resume?

pkg/schedule/schedulers/evict_slow_store.go
okJiang: When a slow store is converted from a slow store to a normal store, we should directly select a store from the candidates that are already slow stores to perform pause leader transfer

pkg/schedule/schedulers/evict_slow_store.go
okJiang: Done some refactors https://github.com/tikv/pd/pull/9451/commits/2294b75cb0660e73a7ae3e0490b4150c08f8334c

pkg/schedule/schedulers/evict_slow_store.go
rleungx: If it failed to persist, will the metrics be inconsistent?

okJiang: Yes, it seems that we still need to modify the memory state after the persistence is successful.

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -38,20 +39,42 @@ import (
...
 
 const (
-	slowStoreEvictThreshold   = 100
-	slowStoreRecoverThreshold = 1
+	slowStoreEvictThreshold = 100
+	// For network slow store, its slow scores include the scores from each tikv.
+	// These scores should meet the following conditions and then will be
+	// considered as a slow store:
+	// 1. One network slow score is greater than networkSlowStoreIssueThreshold.
+	// 2. After removing the maximum value, if the number of stores with a score
```
rleungx: The comment should be changed.

rleungx (commented):
Overall LGTM

pkg/schedule/schedulers/evict_slow_store.go
okJiang: done https://github.com/tikv/pd/pull/9451/commits/d520d153ce5e21f6de17cb05f389cad0a822ad32

okJiang: updated https://github.com/tikv/pd/pull/9451/commits/d520d153ce5e21f6de17cb05f389cad0a822ad32

Issue #9458, opened 2025-06-27
GetTS will slow down due to retry


rleungx:
## Bug Report

We should return immediately if there is no error. However, it doesn't break the loop, which results in high TSO latency. See

https://github.com/tikv/pd/blob/master/client/client.go#L539-L566

PR #9459, opened 2025-06-27
client: add unit test for get ts

rleungx:

Issue Number: ref #9458

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


client/client.go
```diff
@@ -539,17 +539,13 @@ func (c *client) GetTS(ctx context.Context) (physical int64, logical int64, err
 	for retryCount = range maxRetries {
 		resp := c.GetTSAsync(ctx)
 		physical, logical, err = resp.Wait()
-		if err == nil {
+		if !errs.IsLeaderChange(err) {
```
rleungx, line 542: err == nil will be checked in this function

okJiang (approved):

PR #9467, opened 2025-07-01
gc: add a test TestUpdateServiceSafePointToGCBarrier

tiancaiamao:

Issue Number: ref #8978

Add a test case to cover the compatibility of old `UpdateServiceSafePoint` API.
That API is deprecated, but we still need to confirm it's behavior is expected after introducing new API.

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


tiancaiamao:
The old `NewSafePointManager` related code will be cleaned up, so this test code will also vanish with the cleanup of old API.
We can not test the compatibility in this way, so close it.

PR #9468, opened 2025-07-02
client: Remove public v2 client and redirect call to v2 internally if keyspace is configured

### LINKED ISSUE #8978: Refactor GC related PD APIs

MyonKeminta:
… 

Issue Number: Close #8978

```commit-message
client: Remove public v2 client and redirect call to v2 internally if keyspace is configured

Removes the GC V2 related public APIs from PD client to avoid being used by new code. This is a temporary approach, and #9292 will give a complete deprecation.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```


client/gc_client.go
```diff
@@ -91,50 +89,6 @@ func (c *client) UpdateServiceSafePointV2(ctx context.Context, keyspaceID uint32
 	return resp.GetMinSafePoint(), nil
 }
 
-// WatchGCSafePointV2 watch gc safe point change.
-func (c *client) WatchGCSafePointV2(ctx context.Context, revision int64) (chan []*pdpb.SafePointEvent, error) {
```
D3Hunter: i think this can be kept? there is no v1 API for this

tests/integrations/client/gc_api_v2_client_test.go
D3Hunter, line 1: why remove the test? we only need to change the API, so the v2 path can be covered

client/gc_client.go
MyonKeminta: Is there any cases that it's actually used?

D3Hunter: not by tidb. not sure if serverless use it.

if needed, i think we can remove it in other PR when you do refactor, this PR we can focus on redirect the other 2 API

lhy1024 (approved):

PR #9475, opened 2025-07-02
resource group: use inner keyspace id rather outside

lhy1024:

Issue Number:  ref https://github.com/tikv/pd/issues/9296

```commit-message
```

use inner keyspace id rather outside to avoid tidb forget pass keyspace id when putResourceGroup

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


client/resource_manager_client.go
```diff
@@ -159,6 +159,9 @@ func (c *client) putResourceGroup(ctx context.Context, metaGroup *rmpb.ResourceG
 	if err != nil {
 		return "", err
 	}
+	metaGroup.KeyspaceId = &rmpb.KeyspaceIDValue{
```
rleungx: Can we add some checks to prevent this thing from happening again?

tools/pd-ut/testdata/group.9.etcdkey
```diff
@@ -89,6 +89,7 @@ resource_group/keyspace/service_limits/ get
 resource_group/keyspace/service_limits/ save
 resource_group/keyspace/settings/ get
 resource_group/keyspace/settings/ watch
+resource_group/keyspace/settings//cross_modify_test save
```
rleungx: Can we remove it? Otherwise, as the tests become more and more, the key will increase.

okJiang (approved):

PR #9478, opened 2025-07-03
maintenance: implement PD maintenance endpoints to serialize TiKV maintenance operations

### LINKED ISSUE #9477: Implementation for rfc-118
## Development Task
Please find design details in https://github.com/tikv/rfcs/blob/master/text/0118-pd-maintenance-endpoints.md

HaoW30:

Issue Number: Close https://github.com/tikv/pd/issues/9477
  This commit implements the maintenance endpoints feature as specified in [RFC-0118](https://github.com/tikv/rfcs/blob/master/text/0118-pd-maintenance-endpoints.md)
  to prevent Raft quorum loss during TiKV maintenance operations.

  Key features:
  - Add atomic maintenance task storage operations using etcd transactions
  - Implement four HTTP endpoints: PUT, GET, GET by type, DELETE
  - Add pdctl CLI commands for maintenance task management
  - Integrate Prometheus metrics for monitoring
  - Ensure single maintenance task constraint with proper conflict handling

Tests

- [x] Unit test

- Integration test
- No code

Code changes

- [x]  Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/storage/endpoint/maintenance.go
+++ b/pkg/storage/endpoint/maintenance.go
@@ -1,0 +1,239 @@
+// Copyright 2025 TiKV Project Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package endpoint provides storage operations for maintenance tasks.
+// This implementation follows the design specified in RFC-0118:
+// https://github.com/tikv/rfcs/blob/master/text/0118-pd-maintenance-endpoints.md
+//
+// The storage layer provides atomic operations to ensure only one maintenance
+// task can be active at a time, preventing Raft quorum loss during TiKV maintenance.
+
+package endpoint
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/tikv/pd/pkg/storage/kv"
+	"github.com/tikv/pd/pkg/utils/keypath"
+)
+
+// MaintenanceTask represents a maintenance task stored in etcd.
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
+type MaintenanceTask struct {
+	Type           string `json:"type"`            // The type of maintenance task
+	ID             string `json:"id"`              // Unique identifier for the task
+	StartTimestamp int64  `json:"start_timestamp"` // Unix timestamp when the task started
+	Description    string `json:"description"`     // Optional description of the task
+}
+
+// MaintenanceStorage defines the interface for maintenance task storage operations.
+type MaintenanceStorage interface {
+	// LoadMaintenanceTask loads a maintenance task by type from storage.
+	LoadMaintenanceTask(taskType string) (*MaintenanceTask, error)
+	// LoadAllMaintenanceTasks loads all maintenance tasks from storage.
+	LoadAllMaintenanceTasks(f func(k, v string)) error
+	// TryStartMaintenanceTaskAtomic tries to start a maintenance task atomically.
+	TryStartMaintenanceTaskAtomic(ctx context.Context, task *MaintenanceTask) (bool, *MaintenanceTask, error)
+	// TryEndMaintenanceTaskAtomic tries to end a maintenance task atomically.
+	TryEndMaintenanceTaskAtomic(ctx context.Context, taskType, taskID string) (bool, *MaintenanceTask, error)
+}
+
+var _ MaintenanceStorage = (*StorageEndpoint)(nil)
+
+// LoadMaintenanceTask loads a maintenance task by type from storage.
+func (se *StorageEndpoint) LoadMaintenanceTask(taskType string) (*MaintenanceTask, error) {
+	key := keypath.MaintenanceTaskPath(taskType)
+	val, err := se.Load(key)
+	if err != nil {
+		return nil, err
+	}
+	if val == "" {
+		return nil, nil
+	}
+	var task MaintenanceTask
+	if err := json.Unmarshal([]byte(val), &task); err != nil {
+		return nil, err
+	}
+	return &task, nil
+}
+
+// LoadAllMaintenanceTasks loads all maintenance tasks from storage.
+func (se *StorageEndpoint) LoadAllMaintenanceTasks(f func(k, v string)) error {
+	return se.loadRangeByPrefix(keypath.MaintenanceTaskPathPrefix(), f)
+}
+
+// TryStartMaintenanceTaskAtomic tries to start a maintenance task atomically.
+// Returns (true, nil) if success, (false, existingTask, nil) if conflict, (false, nil, err) if error.
+func (se *StorageEndpoint) TryStartMaintenanceTaskAtomic(_ context.Context, task *MaintenanceTask) (bool, *MaintenanceTask, error) {
+	// Use a single atomic transaction with proper etcd pattern
+	rawTxn := se.CreateRawTxn()
+
+	// Maintenance lock key to enforce single task type constraint
+	lockKey := keypath.MaintenanceTaskPath("__maintenance_lock__")
+	// Task key for storing the actual task data
+	taskKey := keypath.MaintenanceTaskPath(task.Type)
+
+	// Condition: check if no maintenance task is running (lock doesn't exist)
+	cond := kv.RawTxnCondition{
+		Key:     lockKey,
+		CmpType: kv.RawTxnCmpNotExists,
+	}
+
+	// Then: put both the lock and the task
+	val, err := json.Marshal(task)
+	if err != nil {
+		return false, nil, err
+	}
+
+	// Put the maintenance lock with "task_type/task_id"
+	lockOp := kv.RawTxnOp{
+		Key:    lockKey,
+		OpType: kv.RawTxnOpPut,
+		Value:  task.Type + "/" + task.ID, // Store "task_type/task_id" as the lock value
+	}
+
+	// Put the actual task
+	taskOp := kv.RawTxnOp{
+		Key:    taskKey,
+		OpType: kv.RawTxnOpPut,
+		Value:  string(val),
+	}
+
+	// Else: get all tasks to see what's currently running
+	getAllOp := kv.RawTxnOp{
+		Key:    keypath.MaintenanceTaskPathPrefix(),          // Start key
+		OpType: kv.RawTxnOpGetRange,                          // Get range of keys
+		EndKey: keypath.MaintenanceTaskPathPrefix() + "\xff", // End key (prefix + max byte)
+		Limit:  100,                                          // Reasonable limit
+	}
+
+	// Execute the transaction: if no maintenance running, put lock and task; else get all tasks
+	resp, err := rawTxn.If(cond).Then(lockOp, taskOp).Else(getAllOp).Commit()
+	if err != nil {
+		return false, nil, err
+	}
+
+	// Check if the transaction succeeded (lock and task were put)
+	if resp.Succeeded {
+		return true, nil, nil
+	}
+
+	// Transaction failed, meaning there was an existing lock
+	// Get all tasks and find the one that's currently running
+	if len(resp.Responses) > 0 && len(resp.Responses[0].KeyValuePairs) > 0 {
+		var existingTask *MaintenanceTask
+		taskCount := 0
+
+		for _, kv := range resp.Responses[0].KeyValuePairs {
+			// Skip the lock key itself
+			if kv.Key == lockKey {
+				continue
+			}
+
+			// Parse the task
+			var task MaintenanceTask
+			if err := json.Unmarshal([]byte(kv.Value), &task); err == nil {
+				existingTask = &task
+				taskCount++
+			}
+		}
+
+		// Assert that there should be only one task
+		if taskCount > 1 {
+			// This shouldn't happen - multiple tasks running simultaneously
+			return false, nil, fmt.Errorf("multiple maintenance tasks running simultaneously: found %d tasks", taskCount)
+		}
+
+		// Return the single task that's running
+		if existingTask != nil {
+			return false, existingTask, nil
+		}
+	}
+
+	return false, nil, nil
+}
+
+// TryEndMaintenanceTaskAtomic tries to end a maintenance task atomically.
+// Returns (true, nil) if success, (false, existingTask, nil) if conflict, (false, nil, err) if error.
+func (se *StorageEndpoint) TryEndMaintenanceTaskAtomic(_ context.Context, taskType, taskID string) (bool, *MaintenanceTask, error) {
+	// Use a single atomic transaction
+	rawTxn := se.CreateRawTxn()
+
+	lockKey := keypath.MaintenanceTaskPath("__maintenance_lock__")
+	taskKey := keypath.MaintenanceTaskPath(taskType)
+
+	// Condition: check if the lock value matches "task_type/task_id"
+	expectedLockValue := taskType + "/" + taskID
+	cond := kv.RawTxnCondition{
+		Key:     lockKey,
+		CmpType: kv.RawTxnCmpEqual,
+		Value:   expectedLockValue,
+	}
+
+	// Then: delete both the lock and the task
+	lockDeleteOp := kv.RawTxnOp{
+		Key:    lockKey,
+		OpType: kv.RawTxnOpDelete,
+	}
+
+	taskDeleteOp := kv.RawTxnOp{
+		Key:    taskKey,
+		OpType: kv.RawTxnOpDelete,
+	}
+
+	// Else: get all tasks to check what's currently running
+	getAllOp := kv.RawTxnOp{
+		Key:    keypath.MaintenanceTaskPathPrefix(),          // Start key
+		OpType: kv.RawTxnOpGetRange,                          // Get range of keys
+		EndKey: keypath.MaintenanceTaskPathPrefix() + "\xff", // End key (prefix + max byte)
+		Limit:  100,                                          // Reasonable limit
+	}
+
+	// Execute the transaction: if lock matches "task_type/task_id", delete both; else get all tasks
+	resp, err := rawTxn.If(cond).Then(lockDeleteOp, taskDeleteOp).Else(getAllOp).Commit()
+	if err != nil {
+		return false, nil, err
+	}
+
+	// Check if the transaction succeeded (deletes were executed)
+	if resp.Succeeded {
+		return true, nil, nil
+	}
+
+	// Transaction failed, meaning we got all tasks
+	// Check if there's a task running for the requested type
+	if len(resp.Responses) > 0 && len(resp.Responses[0].KeyValuePairs) > 0 {
+		for _, kv := range resp.Responses[0].KeyValuePairs {
+			// Skip the lock key itself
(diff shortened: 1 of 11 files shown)


HaoW30:
Will update docs onece this PR is approved

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+// MaintenanceStorage defines the interface for maintenance task storage operations.
+type MaintenanceStorage interface {
+	// LoadMaintenanceTask loads a maintenance task by type from storage.
+	LoadMaintenanceTask(taskType string) (*MaintenanceTask, error)
+	// LoadAllMaintenanceTasks loads all maintenance tasks from storage.
+	LoadAllMaintenanceTasks(f func(k, v string)) error
+	// TryStartMaintenanceTaskAtomic tries to start a maintenance task atomically.
+	TryStartMaintenanceTaskAtomic(ctx context.Context, task *MaintenanceTask) (bool, *MaintenanceTask, error)
+	// TryEndMaintenanceTaskAtomic tries to end a maintenance task atomically.
+	TryEndMaintenanceTaskAtomic(ctx context.Context, taskType, taskID string) (bool, *MaintenanceTask, error)
```
bufferflies: How about a unified TryDeleteTaskAtomic as the same as the client?

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+		Key:    taskKey,
+		OpType: kv.RawTxnOpPut,
+		Value:  string(val),
+	}
+
+	// Else: get all tasks to see what's currently running
+	getAllOp := kv.RawTxnOp{
+		Key:    keypath.MaintenanceTaskPathPrefix(),          // Start key
+		OpType: kv.RawTxnOpGetRange,                          // Get range of keys
+		EndKey: keypath.MaintenanceTaskPathPrefix() + "\xff", // End key (prefix + max byte)
```
SerjKol80, line 130: Will this skip all keys with prefix ````keypath.MaintenanceTaskPathPrefix() + "\xff"````?
It looks like if you set '0x00' in EndKey, the Prefix Get will be used instead of Raneg Get: https://github.com/tikv/pd/blob/f1b0c68e817505930e21a6ffb3c0022292a3c767/pkg/storage/kv/etcd_kv.go#L341-L346

But to be honest, that is pretty tricky and not readable. My suggestion is while you are here just add support of ````RawTxnOpGetPrefix````in this const: https://github.com/tikv/pd/blob/f1b0c68e817505930e21a6ffb3c0022292a3c767/pkg/storage/kv/kv.go#L37-L42

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+				existingTask = &task
+				taskCount++
+			}
+		}
+
+		// Assert that there should be only one task
+		if taskCount > 1 {
+			// This shouldn't happen - multiple tasks running simultaneously
+			return false, nil, fmt.Errorf("multiple maintenance tasks running simultaneously: found %d tasks", taskCount)
+		}
```
SerjKol80: What do you think about printing all found tasks? More context will help in debugging if we ever face this issue.

SerjKol80 (commented):
How do different components (ie. k8s-operator) suppose to distinguish hit own operations from others'? Is it supposed that it needs to encode his ID into taskID?

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+	StartTimestamp int64  `json:"start_timestamp"` // Unix timestamp when the task started
+	Description    string `json:"description"`     // Optional description of the task
+}
+
+// MaintenanceStorage defines the interface for maintenance task storage operations.
+type MaintenanceStorage interface {
+	// LoadMaintenanceTask loads a maintenance task by type from storage.
+	LoadMaintenanceTask(taskType string) (*MaintenanceTask, error)
+	// LoadAllMaintenanceTasks loads all maintenance tasks from storage.
+	LoadAllMaintenanceTasks(f func(k, v string)) error
```
bufferflies, line 55: Why need to input one function? It's better to return all tasks and hide the function in the implementation.

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// TryStartMaintenanceTaskAtomic tries to start a maintenance task atomically.
+// Returns (true, nil) if success, (false, existingTask, nil) if conflict, (false, nil, err) if error.
+func (se *StorageEndpoint) TryStartMaintenanceTaskAtomic(_ context.Context, task *MaintenanceTask) (bool, *MaintenanceTask, error) {
+	// Use a single atomic transaction with proper etcd pattern
+	rawTxn := se.CreateRawTxn()
+
+	// Maintenance lock key to enforce single task type constraint
+	lockKey := keypath.MaintenanceTaskPath("__maintenance_lock__")
```
bufferflies: how about defining one constant variable or function?

pkg/storage/endpoint/maintenance.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// LoadAllMaintenanceTasks loads all maintenance tasks from storage.
+func (se *StorageEndpoint) LoadAllMaintenanceTasks(f func(k, v string)) error {
+	return se.loadRangeByPrefix(keypath.MaintenanceTaskPathPrefix(), f)
+}
+
+// TryStartMaintenanceTaskAtomic tries to start a maintenance task atomically.
+// Returns (true, nil) if success, (false, existingTask, nil) if conflict, (false, nil, err) if error.
+func (se *StorageEndpoint) TryStartMaintenanceTaskAtomic(_ context.Context, task *MaintenanceTask) (bool, *MaintenanceTask, error) {
```
bufferflies, line 88: I think all these actions must happen in the pd leader so that we can use the memory lock instead of the etcd txn api to make the modification aotmic .

tools/pd-ctl/pdctl/command/maintenance_command.go
```diff
@@ -0,0 +1,186 @@
...
+			cmd.Println(resp)
+		},
+	}
+	return cmd
+}
+
+// extractExistingTaskInfo tries to extract existing task information from error response
+func extractExistingTaskInfo(errorMsg string) string {
+	// Look for JSON content in the error message
+	start := strings.Index(errorMsg, "{")
```
bufferflies: It seems strange to decode the error message manually.

pkg/utils/keypath/absolute_key_path.go
```diff
@@ -58,6 +58,9 @@ const (
 	regionLablePathFormat   = "/pd/%d/region_label/%s" // "/pd/{cluster_id}/region_label/{label_id}"
 	regionLabelPrefixFormat = "/pd/%d/region_label/"   // "/pd/{cluster_id}/region_label/"
 
+	// Maintenance task path format
+	maintenanceTaskPathFormat = "/pd/%d/maintenance/%s" // "/pd/{cluster_id}/maintenance/{task_type}"
```
SerjKol80, line 62: Why do we need {cluster_id} parameter if PD instance work only with a single cluster? In what scenarios we expect to use different ids?

pkg/storage/endpoint/maintenance.go
HaoW30: @bufferflies 
This is essentially a case of batch processing vs. streaming processing.

In general, for range-based queries like LoadAllMaintenanceTasks, streaming is often the better choice. By passing in a processor function, we can avoid unnecessary memory copies and enable short-circuiting when appropriate. This approach is also consistent with the existing codebase, as seen in [this example](https://github.com/tikv/pd/blob/6bdfbd398bce2e0cf2a3753a180470d12737699d/pkg/storage/endpoint/util.go#L63).

Although the difference is minimal in the current design, I prefer to stick with the streaming approach for better extensibility and as a good coding practice.

pkg/utils/keypath/absolute_key_path.go
HaoW30: @SerjKol80 I think this is a more general question for PingCap folks. 

From my understanding, this is related to multi-cluster support—allowing multiple clusters to share the same etcd infrastructure. In this context, the cluster ID acts like a namespace identifier to isolate data between clusters.
https://github.com/tikv/pd/blob/6bdfbd398bce2e0cf2a3753a180470d12737699d/tests/integrations/realcluster/cluster_id_test.go#L40-L63
It may also be required for BR and recovery operations.
https://github.com/tikv/pd/blob/6bdfbd398bce2e0cf2a3753a180470d12737699d/tools/pd-recover/main.go#L114-L163

pkg/storage/endpoint/maintenance.go
HaoW30: Renamed

pkg/storage/endpoint/maintenance.go
HaoW30: @bufferflies Good point. I actually considered both approaches during implementation. While using an in-memory lock combined with separate etcd operations can achieve similar results to an etcd transaction, I chose to go with the etcd transaction for the following reasons:

1. **Built-in support**: PD already provides robust support for etcd transactions through storage/kv/etcd_kv.go, so there's no need to implement anything new. We can simply build on top of the atomicity guarantees etcd already provides.
2. **True atomicity**: etcd ensures true atomicity, whereas the in-memory lock approach requires careful review of corner cases. It also removes the burden of checking for leadership and explicitly wrapping updates with LeaderTxn() to ensure they go through the PD leader.
3. **Better extensibility**: While in-memory locking might seem simpler at first—possibly saving one etcd write—it introduces extensibility issues. If we later expand support to multiple types of maintenance tasks and allow independent updates to each type, etcd transactions can elegantly handle this without requiring multiple locks for each type of task_type.
4. **Future-proofing follower handling**: According to the current design, maintenance task tracking serves as a system of record. There’s no fundamental reason this endpoint couldn’t be handled by followers in the future (via PD-Allow-Follower-Handle). Using etcd transactions aligns well with that potential evolution, making the transition easier and more robust.
5. **Low performance sensitivity**: Since maintenance task registration is expected to be a pretty low-frequency operation, I’m not particularly concerned about the performance difference between the two approaches.

That said, if you have strong reasons for preferring the in-memory lock solution, I’m happy to discuss further and explore alternatives.

pkg/storage/endpoint/maintenance.go
HaoW30: Will do once we finalize the solution for ensuring atomicity.

tools/pd-ctl/pdctl/command/maintenance_command.go
HaoW30: Yes, the reason I added this was to provide a more human-readable error message in the pd-ctl command response. That’s why I included decoding of the responses.

Let me know if you feel this is too cumbersome, I can update it to return the raw HTTP response directly as the pd-ctl result instead.

pkg/storage/endpoint/maintenance.go
HaoW30: This is an unlikely error scenario. If it does occur, the user can simply issue another GET request to retrieve all running tasks.

pkg/storage/endpoint/maintenance.go
HaoW30: I think it should be fine, it only leaves out the keys that's lexicographically equal to or greater than prefix + "\xff".

\xff is not a printable ASCII char. 
In UTF-8, "\xff" is not valid on its own, but Go will treat it as a single byte in a string. So in this context, it’s safe and intentional.

Only if you actually have a key that starts with prefix + "\xff", which is unlikely, unless you explicitly used that value as the first char of a task type name.
Normally, all your keys will fall under prefix + [0x00 - 0xFE].

HaoW30:
@bufferflies gentle ping... thanks!

tools/pd-ctl/pdctl/command/maintenance_command.go
bufferflies: I think it's better to return the raw http response directly

pkg/storage/endpoint/maintenance.go
bufferflies: What I mean is that you can define a constant, such as LockName = "__maintenance_lock__", and then use it here.

tools/pd-ctl/pdctl/command/maintenance_command.go
HaoW30: will do

HaoW30:
> How do different components (ie. k8s-operator) suppose to distinguish hit own operations from others'? Is it supposed that it needs to encode his ID into taskID?

That’s entirely up to the orchestrator (k8s operator). Yes, I think ideally, the operator should include its identifier in the task type or task ID.

pkg/utils/keypath/absolute_key_path.go
Tema: Yeah, it is mostly historic. There were namespaces in plain TiKV, which are not used in TiDB bundle. Basically there is a single namespace per cluster in TiDB world.
Also till 8.0 there were no mTLS for PD/etcd data endpoint, so providing clusterID is a good safety check that you talk to intended cluster.

tools/pd-ctl/pdctl/command/maintenance_command.go
HaoW30: Updated to return raw http response. Please take another look.

pkg/storage/endpoint/maintenance.go
HaoW30: Done

pkg/storage/endpoint/maintenance.go
bufferflies: No, thank you for your analysis. You can add a comment to note the reason in this PR or the code.

server/apiv2/handlers/maintenance.go
```diff
@@ -0,0 +1,203 @@
...
+
+// GetMaintenanceTask gets information about the ongoing maintenance task.
+// @Tags     maintenance
+// @Summary  Get information about the ongoing maintenance task.
+// @Produce  json
+// @Success  200  {array}  endpoint.MaintenanceTask
+// @Failure  404  {string}  string  "No maintenance task is running."
+// @Failure  500  {string}  string  "PD server failed to proceed the request."
+// @Router   /maintenance [get]
+func GetMaintenanceTask(c *gin.Context) {
```
bufferflies: it should be renamed as `GetMaintenanceTasks` if this returns all the task.

server/apiv2/handlers/maintenance.go
```diff
@@ -0,0 +1,203 @@
...
+		return
+	}
+
+	// Update metrics
+	maintenance.SetMaintenanceTaskInfo(taskType, taskID, 1)
+
+	c.JSON(http.StatusOK, "Maintenance task started successfully.")
+}
+
+// GetMaintenanceTask gets information about the ongoing maintenance task.
```
bufferflies:
```suggestion
// GetMaintenanceTask gets information about the ongoing maintenance tasks.
```

server/apiv2/handlers/maintenance.go
```diff
@@ -0,0 +1,203 @@
...
+// @Tags     maintenance
+// @Summary  Get information about the ongoing maintenance task.
```
bufferflies:
```suggestion
// @Summary  Get information about the ongoing maintenance tasks.
```

server/apiv2/handlers/maintenance.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	task := &endpoint.MaintenanceTask{
+		Type:           taskType,
+		ID:             taskID,
+		StartTimestamp: time.Now().Unix(),
+		Description:    description,
+	}
+
+	ok, existingTask, err := svr.GetStorage().TryStartMaintenanceTaskAtomic(c.Request.Context(), task)
```
bufferflies, line 82: Do we need to print some logs to debug  this inner status if something's wrong and it's helpful to record all the behavior

server/apiv2/handlers/maintenance.go
```diff
@@ -0,0 +1,203 @@
...
+// @Summary  Delete a maintenance task.
+// @Param    task_type  path  string  true  "The type of maintenance task"
+// @Param    task_id    path  string  true  "Unique identifier for the task"
+// @Produce  json
+// @Success  200  {string}  string  "Maintenance task deleted successfully."
+// @Failure  404  {string}  string  "No maintenance task is running for this type."
+// @Failure  409  {object}  map[string]interface{}  "Task ID does not match the current task."
+// @Failure  500  {string}  string  "PD server failed to proceed the request."
+// @Router   /maintenance/{task_type}/{task_id} [delete]
+func DeleteMaintenanceTask(c *gin.Context) {
```
bufferflies, line 177: ditto

tools/pd-ctl/pdctl/command/maintenance_command.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			resp, err := doRequest(cmd, path, http.MethodGet, http.Header{})
+			if err != nil {
+				cmd.Printf("Failed to get maintenance task: %s\n", err)
+				return
+			}
+
+			// Print the raw response
+			cmd.Println(resp)
```
bufferflies, line 100: Does it need to decode the timestamp to a readable time string?

tools/pd-ctl/pdctl/command/maintenance_command.go
```diff
@@ -0,0 +1,125 @@
...
+func newMaintenanceSetCommand() *cobra.Command {
+	var desc string
+	cmd := &cobra.Command{
+		Use:   "set <task_type> <task_id>",
+		Short: "Start a maintenance task",
+		Args:  cobra.ExactArgs(2),
+		Run: func(cmd *cobra.Command, args []string) {
+			taskType := args[0]
+			taskID := args[1]
+			path := fmt.Sprintf("pd/api/v2/maintenance/%s/%s", taskType, taskID)
```
bufferflies: how about defining this as maintenancePrefixURI?

tools/pd-ctl/pdctl/command/maintenance_command.go
HaoW30: I’d like to keep the raw timestamp in the HTTP response, as it’s easier for the k8s operator to handle.
As for the pdctl command response, I think it’s fine to return the raw Unix epoch—it’s straightforward for a human user to decode.
If we were to decode the timestamp for the user, we’d be reverting to the old approach of parsing the HTTP response and handling all the potential error scenarios.

server/apiv2/handlers/maintenance.go
HaoW30: Updated

server/apiv2/handlers/maintenance.go
HaoW30: Done

server/apiv2/handlers/maintenance.go
HaoW30: Done

server/apiv2/handlers/maintenance.go
HaoW30: Added more log prints for TryStartMaintenanceTaskAtomic and TryDeleteMaintenanceTaskAtomic in pkg/storage/endpoint/maintenance.go

tools/pd-ctl/pdctl/command/maintenance_command.go
HaoW30: Done

HaoW30:
@bufferflies comments addressed, please take another look, thanks!

bufferflies (approved):

okJiang (approved):

HaoW30:
@okJiang could you please help to trigger the rerun of the ci tests again?

Tema (approved):

Tema (approved):

PR #9499, opened 2025-07-07
mcs, test: fix `IsKeyspaceServing` and `TestKeyspacesServedByDefaultK…

lhy1024:
…eyspaceGroup`

Issue Number: Ref https://github.com/tikv/pd/issues/9414

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/keyspace/util.go
```diff
@@ -269,3 +258,44 @@ func (hp *indexedHeap) Remove(id uint32) *endpoint.KeyspaceGroup {
...
+}
+
+func isProtectedKeyspaceID(id uint32) bool {
+	if kerneltype.IsNextGen() {
+		return uint64(id) >= constant.ReservedKeyspaceIDStart
+	}
+	return id == constant.DefaultKeyspaceID
+}
+
+func isProtectedKeyspaceName(name string) bool {
```
rleungx, line 296: We only exclude two keyspaces through name, but exclude more through ID.

pkg/keyspace/constant/constant.go
```diff
@@ -18,6 +18,9 @@ const (
 	// DefaultKeyspaceName is the name reserved for default keyspace.
 	DefaultKeyspaceName = "DEFAULT"
 
+	// ZeroKeyspaceID is the start keyspace for keyspace id.
+	ZeroKeyspaceID = uint32(0)
```
JmPotato: What about naming `StartKeyspaceID`?

JmPotato (approved):

PR #9510, opened 2025-07-09
ci: close fail-fast in pd tests chunks

lhy1024:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


.github/workflows/pd-tests.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
   chunks:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
```
bufferflies, line 26: It may bring some resource loads of the GitHub action

okJiang (approved):

.github/workflows/pd-tests.yaml
MyonKeminta: In the current situation in which the fail rate is high, fail-fast causes many sub-jobs canceled and rerun from the beginning after a `/retest`, then it has another chance to be failed or canceled. Before we fix the unstable tests, it only wastes more resources and significantly increases the failure rate of `/retest`.

Issue #9357, opened 2025-05-27
Reduce the /metrics result size


nolouch:
## Enhancement Task
In the large cluster, we found scrape the PD metrics will timeout:

```
time curl --tlsv1.2  "https://xxx:2379/metrics" > scrape.txt

% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dload  Upload   Total   Spent    Left  Speed
100  165M    0  165M    0     0  10.7M      0 --:--:--  0:00:15 --:--:-- 35.0M

real	0m15.433s
user	0m0.157s
sys	0m0.438s
```
exceed default timeout 15s. and each pull will cost many network flow. 

```
ll metrics.txt
➜  Downloads ll metrics.txt
-rw-r--r--@ 1 nolouch  staff   183M  5 27 14:58 metrics.txt
```

I write a scriple analyzed the top30 metrics in the result:

```
Top 30 Metrics by Occurrence:

pd_schedule_filter: 781310 lines
pd_scheduler_balance_direction: 292851 lines
pd_scheduler_hot_pending: 280922 lines
pd_service_audit_handling_seconds_bucket: 54444 lines
pd_scheduler_store_status: 17574 lines
pd_scheduler_handle_store_heartbeat_duration_seconds_bucket: 15480 lines
pd_scheduler_handle_region_heartbeat_duration_seconds_bucket: 14280 lines
pd_scheduler_hot_region_direction: 11896 lines
pd_scheduler_hot_region: 11202 lines
pd_hotspot_status: 8256 lines
pd_hotcache_status: 7224 lines
pd_scheduler_hot_peers_summary: 6232 lines
pd_scheduler_region_heartbeat_latency_seconds_bucket: 6188 lines
pd_service_audit_handling_seconds_sum: 4537 lines
pd_service_audit_handling_seconds_count: 4537 lines
pd_scheduler_hot_pending_sum: 2916 lines
pd_scheduler_region_heartbeat: 1864 lines
pd_schedule_store_limit_cost: 1508 lines
pd_hbstream_region_message: 1269 lines
pd_cluster_store_limit: 1034 lines
grpc_server_handling_seconds_bucket: 840 lines
grpc_server_handled_total: 743 lines
pd_scheduler_handle_store_heartbeat_duration_seconds_sum: 516 lines
pd_scheduler_handle_store_heartbeat_duration_seconds_count: 516 lines
pd_cluster_placement_status: 481 lines
[...]

PR #9535, opened 2025-07-14
metrics: clean up metrics

rleungx:

Issue Number: ref #9357

This PR makes the following changes:
- Remove some unnecessary metrics
- Remove some useless monitors
- Remove dc field for TSO

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/statistics/hot_cache.go
+++ b/pkg/statistics/hot_cache.go
@@ -28,11 +28,6 @@ import (
 
 const chanMaxLength = 6000000
 
-var (
-	readTaskMetrics  = hotCacheFlowQueueStatusGauge.WithLabelValues(utils.Read.String())
-	writeTaskMetrics = hotCacheFlowQueueStatusGauge.WithLabelValues(utils.Write.String())
-)
-
 // HotCache is a cache hold hot regions.
 type HotCache struct {
 	ctx        context.Context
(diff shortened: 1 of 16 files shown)


pkg/schedule/checker/checker_controller.go
```diff
@@ -55,13 +55,6 @@ const (
 	patrolRegionChanLen      = MaxPatrolScanRegionLimit
 )
 
-var (
-	// WithLabelValues is a heavy operation, define variable to avoid call it every time.
-	pendingProcessedRegionsGauge = regionListGauge.WithLabelValues("pending_processed_regions")
```
lhy1024: pending_processed_regions maybe helpful

pkg/statistics/hot_cache.go
```diff
@@ -28,11 +28,6 @@ import (
 
 const chanMaxLength = 6000000
 
-var (
```
lhy1024, line 31: I am not sure whether we need it.

lhy1024 (commented):
The rest LGTM about schedule metrics

pkg/schedule/checker/checker_controller.go
rleungx: When do we use it?

rleungx: add it back

lhy1024 (approved):

JmPotato (approved):

PR #9537, opened 2025-07-14
metrics: reduce balance metrics

rleungx:

Issue Number: ref #9357.

```commit-message
```

### Check List

Tests

- Manual test
Screenshot 2025-07-14 at 17 24 57

### Release note

```release-note
None.
```


pkg/schedule/schedulers/balance_leader.go
```diff
@@ -565,7 +565,8 @@ func (s *balanceLeaderScheduler) createOperator(solver *solver, collector *plan.
 		balanceLeaderNewOpCounter,
 	)
 	op.FinishedCounters = append(op.FinishedCounters,
-		balanceDirectionCounter.WithLabelValues(s.GetName(), solver.sourceMetricLabel(), solver.targetMetricLabel()),
+		balanceDirectionCounter.WithLabelValues(s.GetName(), solver.sourceMetricLabel(), "out"),
+		balanceDirectionCounter.WithLabelValues(s.GetName(), solver.targetMetricLabel(), "in"),
```
okJiang, line 569: Doesn't this mean that the number of metrics records does not decrease, but increases?

rleungx: No, the number of metrics will decrease.

okJiang (approved):

PR #9561, opened 2025-07-17
api: fix some wrong metrics and add bucket version

### LINKED ISSUE #9557: Region API doesn't have bucket version
## Enhancement Task
We need to investigate whether the bucket is updated successfully sometimes. Both TiDB and TiKV track the bucket version numbers in the log. However, we don't have a way to find the bucket version of a specified region on the PD side, which is inconvenient for troubleshooting problems. 

Also, we don't have any logs either.

bufferflies:

Issue Number: Close #9557

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/response/region_easyjson.go
+++ b/pkg/response/region_easyjson.go
@@ -19,7 +19,185 @@ var (
 	_ easyjson.Marshaler
 )
 
-func easyjson75d7afa0DecodeGithubComTikvPdServerApi(in *jlexer.Lexer, out *RegionInfo) {
+func easyjson75d7afa0DecodeGithubComTikvPdPkgResponse(in *jlexer.Lexer, out *ReplicationStatus) {
+	isTopLevel := in.IsStart()
+	if in.IsNull() {
+		if isTopLevel {
+			in.Consumed()
+		}
+		in.Skip()
+		return
+	}
+	in.Delim('{')
+	for !in.IsDelim('}') {
+		key := in.UnsafeFieldName(false)
+		in.WantColon()
+		if in.IsNull() {
+			in.Skip()
+			in.WantComma()
+			continue
+		}
+		switch key {
+		case "state":
+			out.State = string(in.String())
+		case "state_id":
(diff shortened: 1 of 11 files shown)


pkg/response/region_easyjson.go
```diff
@@ -139,6 +317,8 @@ func easyjson75d7afa0DecodeGithubComTikvPdServerApi(in *jlexer.Lexer, out *Regio
 			out.ApproximateSize = int64(in.Int64())
 		case "approximate_keys":
 			out.ApproximateKeys = int64(in.Int64())
+		case "approximate_kv_size":
```
lhy1024, line 320: After https://github.com/tikv/pd/pull/8403, we added `ApproximateKvSize` but no call easyjson. Do we need to create another PR to pick it to other branches?

tests/server/api/region_test.go
```diff
@@ -409,7 +409,9 @@ func (suite *regionTestSuite) checkRegion(cluster *tests.TestCluster) {
 		core.SetWrittenBytes(100*units.MiB),
 		core.SetWrittenKeys(1*units.MiB),
 		core.SetReadBytes(200*units.MiB),
-		core.SetReadKeys(2*units.MiB))
+		core.SetReadKeys(2*units.MiB),
```
lhy1024, line 412: Do we need to add a test for bucket version?

lhy1024 (commented):
the rest LGTM

tests/server/api/region_test.go
bufferflies: This test case verifies the bucket version if we marshal the bucket version.

pkg/response/region_easyjson.go
bufferflies: I don't think so; it means that nobody uses this information now. If someone needs it, we can cp this later.

rleungx (approved):

PR #9577, opened 2025-07-23
feat(resourcemanager): invalidate the burstable groups by overriding to service limit

JmPotato:

Issue Number: ref #9296.

```commit-message
By overriding the burst limit with the service limit for burstable groups to enhance the constraint imposed by the service limit.
```

### Check List

Tests

- Unit test
- Integration test
- Mannual test

image

image

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/resource_group.go
+++ b/pkg/mcs/resourcemanager/server/resource_group.go
@@ -165,6 +165,10 @@ func (rg *ResourceGroup) overrideFillRateLocked(new float64) {
 func (rg *ResourceGroup) getBurstLimit(ignoreOverride ...bool) int64 {
 	rg.RLock()
 	defer rg.RUnlock()
+	return rg.getBurstLimitLocked(ignoreOverride...)
+}
+
+func (rg *ResourceGroup) getBurstLimitLocked(ignoreOverride ...bool) int64 {
 	if len(ignoreOverride) > 0 && ignoreOverride[0] {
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/resource_group.go
```diff
@@ -252,13 +258,12 @@ func (rg *ResourceGroup) RequestRU(
...
 ) *rmpb.GrantedRUTokenBucket {
 	rg.Lock()
-	defer rg.Unlock()
-
 	if rg.RUSettings == nil || rg.RUSettings.RU.Settings == nil {
 		return nil
 	}
 	// First, try to get tokens from the resource group.
 	tb, trickleTimeMs := rg.RUSettings.RU.request(now, requiredToken, targetPeriodMs, clientUniqueID)
+	rg.Unlock()
```
JmPotato: This is to fix the deadlock caused by line 273.

https://github.com/tikv/pd/pull/9577/files#diff-626c8564c95f18abf54adf61e8b99ec1457d48f587ee7b91d7969bfccceb2748R273

glorv: Line 274 updates the `lastLimitedTokens` of rg, so we should still hold the lock to avoid race. Better change L273 instead of release the lock too early.

JmPotato: Fixed.

glorv (commented):
LGTM

rleungx (approved):

PR #9587, opened 2025-07-25
api: fix get region count not correct 

bufferflies:
### What problem does this PR solve?

Issue Number: Close #9586

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

PR #9588, opened 2025-07-25
core: add new interface api to return the store region count by the special key range 

### LINKED ISSUE #9589: optimize the `fetchAllRegions`
## Development Task
The `fetchAllRegions` action will acquire the region lock. If the key range is big, it will take a lot of time and affect the region tree qps

bufferflies:

Issue Number: Close #9589

```commit-message
Replacing FetchAllRegion by GetStoreRegionCount which only needs one lock 
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

Related changes

### Release note

```release-note
None.
```


pkg/core/region_tree_test.go
```diff
@@ -368,6 +368,45 @@ func TestRandomRegionDiscontinuous(t *testing.T) {
...
+			StoreId: 3,
+			Id:      i,
+			Role:    metapb.PeerRole_Learner,
+		}
+	}
+
+	regions.CheckAndPutRegion(NewTestRegionInfo(1, 1, []byte("a"), []byte("c"), WithAddPeer(voterFn()), WithAddPeer(learnerFn())))
+	regions.CheckAndPutRegion(NewTestRegionInfo(2, 1, []byte("e"), []byte("g"), WithAddPeer(voterFn()), WithAddPeer(learnerFn())))
+	regions.CheckAndPutRegion(NewTestRegionInfo(3, 1, []byte("g"), []byte("i"), WithAddPeer(voterFn()), WithAddPeer(learnerFn())))
+	for _, endKey := range [][]byte{[]byte("b"), []byte("c"), []byte("d"), []byte("e"), []byte("f"), []byte("")} {
```
lhy1024: Do we need some hole cases such as [c,e], [b,d]

lhy1024 (commented):
the rest LGTM

pkg/core/region_tree_test.go
bufferflies: yes, wait for the pr https://github.com/tikv/pd/pull/9587

bufferflies: done

pkg/core/region.go
```diff
@@ -1986,33 +1972,43 @@ func (r *RegionInfo) GetWriteLoads() []float64 {
 	}
 }
 
+// GetStoreRegionCountByRule returns the number of regions in given store that overlap with the range [startKey, endKey) for a specific rule.
+func (r *RegionsInfo) GetStoreRegionCountByRule(storeID uint64, startKey, endKey []byte, rule Rule) int {
```
rleungx: How about splitting it into 3 functions?

bufferflies: done

PR #9595, opened 2025-07-28
client: fix tso service discovery

### LINKED ISSUE #9593: TSO service discovery may lose some TSO servers when updating primary
## Bug Report

We use `len(t.urls) == 0 || t.failureCount == len(t.urls)` to decide if we need to update urls according to the service discovery.
https://github.com/tikv/pd/blob/830a43945e096baa6e17b9582666e98bcdba775b/client/servicediscovery/tso_service_discovery.go#L586-L608

The failure count only increases when the request fails.
https://github.com/tikv/pd/blob/830a43945e096baa6e17b9582666e98bcdba775b/client/servicediscovery/tso_service_discovery.go#L432-L441

If the first TSO is registered and ready to serve, which means there are no errors for the request. The rest of the TSO servers won't be added to the urls.

rleungx:

Issue Number: Close #9593.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/servicediscovery/tso_service_discovery.go
+++ b/client/servicediscovery/tso_service_discovery.go
@@ -557,9 +557,6 @@ func (c *tsoServiceDiscovery) findGroupByKeyspaceID(
 }
 
 func (c *tsoServiceDiscovery) getTSOServer(sd ServiceDiscovery) (string, error) {
-	c.Lock()
-	defer c.Unlock()
-
 	var (
 		urls []string
 		err  error
@@ -569,6 +566,8 @@ func (c *tsoServiceDiscovery) getTSOServer(sd ServiceDiscovery) (string, error) 
 		return "", err
 	}
 
+	c.Lock()
+	defer c.Unlock()
 	t := c.tsoServerDiscovery
 	failpoint.Inject("serverReturnsNoTSOAddrs", func() {
 		if len(t.urls) == 0 {


client/servicediscovery/tso_service_discovery.go
```diff
@@ -583,28 +565,29 @@ func (c *tsoServiceDiscovery) getTSOServer(sd ServiceDiscovery) (string, error)
...
-		urls, err = sd.(*serviceDiscovery).discoverMicroservice(tsoService)
-		if err != nil {
-			return "", err
-		}
+	urls, err = sd.(*serviceDiscovery).discoverMicroservice(tsoService)
+	if err != nil {
+		return "", err
+	}
+
+	if len(t.urls) == 0 {
```
JmPotato: Perhaps put the check inisde the failpoint?

client/servicediscovery/tso_service_discovery.go
```diff
@@ -583,28 +565,29 @@ func (c *tsoServiceDiscovery) getTSOServer(sd ServiceDiscovery) (string, error)
 		err  error
 	)
 	t := c.tsoServerDiscovery
```
JmPotato: Now this `t` is assigned across multiple lines in advance, which looks somewhat strange.

rleungx: Assign it after calling discoverMicroservice now.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -582,29 +564,31 @@ func (c *tsoServiceDiscovery) getTSOServer(sd ServiceDiscovery) (string, error)
 		urls []string
 		err  error
 	)
+	urls, err = sd.(*serviceDiscovery).discoverMicroservice(tsoService)
```
lhy1024, line 564: There will be more requests about `discoverMicroservice`. Will it introduce performance problems?

lhy1024, line 564:
```suggestion
	var (
		urls []string
		err  error
	)
	urls, err = sd.(*serviceDiscovery).discoverMicroservice(tsoService)
	if err != nil {
		return "", err
	}
	c.Lock()
	defer c.Unlock()
```

lhy1024: Maybe we could reduce lock time.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -431,15 +416,12 @@ func (c *tsoServiceDiscovery) updateMember() error {
 	if len(tsoServerURL) > 0 {
 		keyspaceGroup, err = c.findGroupByKeyspaceID(keyspaceID, tsoServerURL, UpdateMemberTimeout)
 		if err != nil {
-			if c.countFailure() {
-				log.Error("[tso] failed to find the keyspace group",
-					zap.Uint32("keyspace-id-in-request", keyspaceID),
-					zap.String("tso-server-url", tsoServerURL),
-					errs.ZapError(err))
-			}
+			log.Error("[tso] failed to find the keyspace group",
```
lhy1024, line 419: Will it print more logs?

client/servicediscovery/tso_service_discovery.go
rleungx: Otherwise, we can't update the tso members.

Issue #9455, opened 2025-06-27
Global Admission Controller return Nan tokens


guo-shaoge:
### What did you do?
The TiFlash local bucket continuously requests tokens from PD (GAC) to refill itself. 
However, since the number of tokens returned by PD (GAC) is **NaN**, tiflash cannot refill its local bucket. 
As a result, the TiFlash local bucket eventually runs out of tokens, causing queries to be completely throttled and stuck.

Following is the tiflash log that checks tokens returned from PD is not greater than zero.
![Image](https://github.com/user-attachments/assets/fa15c5b4-b5e8-4928-865b-902cfb129942)

Following is the tiflash log that log the PD resp, which show the tokens is Nan.
![Image](https://github.com/user-attachments/assets/ac2c2878-27f2-47b6-8f78-c26fe2dd4599)

### What version of PD are you using (`pd-server -V`)?
v7.5.3

guo-shaoge:
This issue has occured twice, we have to fix it as soon as possible.

guo-shaoge:
Here are a few workarounds for this issue:

1. Upgrade to v8.5.4 or higher(which include this pr: https://github.com/tikv/pd/pull/9751): Although the root cause of the NaN issue has not been identified, logic has been added to the PD (Placement Driver) code to immediately reset the resource group token slot when this problem is encountered. Therefore, the issue of TiFlash hanging will no longer occur.
2. Disable TiFlash resource control: Set `profiles.default.enable_resource_control` to false. This supports hot reloading, so you do not need to restart TiFlash.
3. Re-create the Resource Control: Drop the problematic resource control and create a new one. However, please note that this issue may recur.

PR #9597, opened 2025-07-29
fix(resourcemanager): add validation for token assignment anomaly

JmPotato:

Issue Number: ref #9455.

```commit-message
Add safeguard checks and warnings during token assignment to handle potential invalid token values.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/token_buckets.go
+++ b/pkg/mcs/resourcemanager/server/token_buckets.go
@@ -150,7 +150,11 @@ type tokenSlot struct {
 
 // GroupTokenBucketState is the running state of TokenBucket.
 type GroupTokenBucketState struct {
-	Tokens float64 `json:"tokens,omitempty"`
+	Tokens      float64    `json:"tokens,omitempty"`
+	LastUpdate  *time.Time `json:"last_update,omitempty"`
+	Initialized bool       `json:"initialized"`
+
+	resourceGroupName string
 	// ClientUniqueID -> TokenSlot
 	tokenSlots                 map[uint64]*tokenSlot
 	clientConsumptionTokensSum float64
@@ -173,8 +177,6 @@ type GroupTokenBucketState struct {
 	// means the burst limit is overridden.
 	overrideBurstLimit int64
 
-	LastUpdate  *time.Time `json:"last_update,omitempty"`
-	Initialized bool       `json:"initialized"`
 	// settingChanged is used to avoid that the number of tokens returned is jitter because of changing fill rate.
 	settingChanged      bool
 	lastCheckExpireSlot time.Time
@@ -198,6 +200,7 @@ func (gts *GroupTokenBucketState) clone() *GroupTokenBucketState {
 		Tokens:                     gts.Tokens,
 		LastUpdate:                 lastUpdate,
 		Initialized:                gts.Initialized,
+		resourceGroupName:          gts.resourceGroupName,
 		tokenSlots:                 tokenSlots,
 		overrideFillRate:           gts.overrideFillRate,
 		overrideBurstLimit:         gts.overrideBurstLimit,
@@ -330,7 +333,7 @@ func (gtb *GroupTokenBucket) calcRateAndBurstLimit(ratio float64) (fillRate uint
 }
 
 // NewGroupTokenBucket returns a new GroupTokenBucket
-func NewGroupTokenBucket(tokenBucket *rmpb.TokenBucket) *GroupTokenBucket {
+func NewGroupTokenBucket(resourceGroupName string, tokenBucket *rmpb.TokenBucket) *GroupTokenBucket {
 	if tokenBucket == nil || tokenBucket.Settings == nil {
 		return &GroupTokenBucket{}
 	}
@@ -338,6 +341,7 @@ func NewGroupTokenBucket(tokenBucket *rmpb.TokenBucket) *GroupTokenBucket {
 		Settings: tokenBucket.GetSettings(),
 		GroupTokenBucketState: GroupTokenBucketState{
 			Tokens:             tokenBucket.GetTokens(),
+			resourceGroupName:  resourceGroupName,
 			tokenSlots:         make(map[uint64]*tokenSlot),
 			overrideFillRate:   -1,
 			overrideBurstLimit: -1,
@@ -419,7 +423,8 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
 }
 
 // request requests tokens from the corresponding slot.
-func (gtb *GroupTokenBucket) request(now time.Time,
+func (gtb *GroupTokenBucket) request(
+	now time.Time,
 	requiredToken float64,
 	targetPeriodMs, clientUniqueID uint64,
 ) (*rmpb.TokenBucket, int64) {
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -419,21 +423,47 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
 	}
 	res, trickleDuration := slot.assignSlotTokens(requiredToken, targetPeriodMs)
 	// Update bucket to record all tokens.
 	gtb.Tokens -= slot.lastTokenCapacity - slot.tokenCapacity
 	slot.lastTokenCapacity = slot.tokenCapacity
-
+	// Verify whether the allocated token is invalid, such as negative values, math.Inf, or math.NaN.
+	if res.Tokens <= 0 || math.IsInf(res.Tokens, 0) || math.IsNaN(res.Tokens) {
+		// Output detailed information to help locate the problem.
+		log.Warn("assigned token is invalid",
```
guo-shaoge: how about error level? we have alters that will notify us as soon as possible.

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -419,21 +423,47 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+			Settings: &rmpb.TokenLimitSettings{BurstLimit: burstLimit},
+			Tokens:   0.0,
+		}, 0
 	}
 	res, trickleDuration := slot.assignSlotTokens(requiredToken, targetPeriodMs)
 	// Update bucket to record all tokens.
 	gtb.Tokens -= slot.lastTokenCapacity - slot.tokenCapacity
 	slot.lastTokenCapacity = slot.tokenCapacity
-
+	// Verify whether the allocated token is invalid, such as negative values, math.Inf, or math.NaN.
```
lhy1024: Good job. However, we need some way to trace where we meet Nan. How about adding some logs when meeting Nan.

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -419,21 +423,47 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+	if res.Tokens <= 0 || math.IsInf(res.Tokens, 0) || math.IsNaN(res.Tokens) {
```
lhy1024: Does it also need to log when res.Tokens is less than zero?

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -419,21 +423,47 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+		// Output detailed information to help locate the problem.
+		log.Warn("assigned token is invalid",
+			zap.Time("now", now),
+			zap.String("resource-group-name", gtb.resourceGroupName),
+			zap.Uint64("client-unique-id", clientUniqueID),
+			zap.Uint64("target-period-ms", targetPeriodMs),
+			zap.Float64("required-token", requiredToken),
+			zap.Float64("assigned-tokens", res.Tokens),
+			zap.Int("slot-len", len(gtb.tokenSlots)),
+			zap.Float64("fill-rate", float64(slot.fillRate)),
```
guo-shaoge: slot.fillRate -> `ts.settings.GetFillRate()`?

guo-shaoge: how about print `slot.setting.String()`?

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -419,21 +423,47 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+			zap.Int64("burst-limit", slot.burstLimit),
+			zap.Float64("require-tokens-sum", slot.requireTokensSum),
+			zap.Float64("token-capacity", slot.tokenCapacity),
```
nolouch: I suspect it's because this field turned into 'nan', but it's hard for me to guess how it happened. Can we add a separate log specifically for when this field becomes 'nan'?

nolouch (approved):

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -418,43 +428,88 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+	logFields []zap.Field,
+) {
+	// Add some common fields to the log fields to help debug.
+	logFields = append(logFields,
+		append(
+			slot.logFields(),
+			zap.String("resource-group-name", gtb.resourceGroupName),
+			zap.String("settings", gtb.Settings.String()),
+			zap.Float64("tokens", gtb.Tokens),
+			zap.Int("slot-len", len(gtb.tokenSlots)),
```
guo-shaoge: also log `gtb.clientConsumptionTokensSum`?

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -418,43 +428,88 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
 	gtb.balanceSlotTokens(clientUniqueID, requiredToken, elapseTokens)
 }
 
+func (gtb *GroupTokenBucket) inspectAnomalies(
+	tb *rmpb.TokenBucket,
+	slot *tokenSlot,
+	logFields []zap.Field,
+) {
+	// Add some common fields to the log fields to help debug.
+	logFields = append(logFields,
```
guo-shaoge: only setup logFileds when need to Reset?

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -418,43 +428,87 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
 	if !ok {
-		return &rmpb.TokenBucket{Settings: &rmpb.TokenLimitSettings{BurstLimit: burstLimit}}, 0
+		return &rmpb.TokenBucket{
+			Settings: &rmpb.TokenLimitSettings{BurstLimit: burstLimit},
+			Tokens:   0.0,
+		}, 0
 	}
 	res, trickleDuration := slot.assignSlotTokens(requiredToken, targetPeriodMs)
+	// Inspect the group token bucket and the assigned token result to catch any anomalies.
+	gtb.inspectAnomalies(res, slot, []zap.Field{
```
lhy1024: When we find nan, should we return nil? Just like yesterday's implementation.

Image

pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -418,43 +428,87 @@ func (gtb *GroupTokenBucket) updateTokens(now time.Time, burstLimit int64, clien
...
+		zap.Float64("assigned-tokens", res.Tokens),
+	})
 	// Update bucket to record all tokens.
 	gtb.Tokens -= slot.lastTokenCapacity - slot.tokenCapacity
 	slot.lastTokenCapacity = slot.tokenCapacity
-
 	return res, trickleDuration
 }
 
 func (ts *tokenSlot) assignSlotTokens(requiredToken float64, targetPeriodMs uint64) (*rmpb.TokenBucket, int64) {
```
lhy1024, line 499: Do we need to add a check such as `targetPeriodMs == 0 || ts.fillRate <= 0`?

pkg/mcs/resourcemanager/server/token_buckets.go
JmPotato: Nice catch. I'll fix it!

pkg/mcs/resourcemanager/server/token_buckets.go
JmPotato: If the bad things happen, it will be included in the logs currently we have.

lhy1024 (approved):

Issue #9556, opened 2025-07-16
Address vulnerabilities in multiple third-party dependencies


JmPotato:
## Bug Report

<img width="1716" height="614" alt="Image" src="https://github.com/user-attachments/assets/5d13f69d-a9e3-4a5e-80b5-0a74eb4ed76e" />

### What version of PD are you using (`pd-server -V`)?

release-8.5 https://github.com/tikv/pd/commit/83f10e83e8fbe5184d77679efd4d362615bfc203

rleungx:
Do we also need to upgrade negroni for master?

PR #9600, opened 2025-07-30
deps(release-7.5): update to fix the vulnerabilities

JmPotato:

Issue Number: ref #9556.

```commit-message
Update the third-party dependencies to address vulnerabilities.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/coreos/go-semver v0.3.0
 	github.com/docker/go-units v0.4.0
 	github.com/elliotchance/pie/v2 v2.1.0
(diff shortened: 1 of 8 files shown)


client/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/tikv/pd/client
 
-go 1.21
+go 1.23.0
```
rleungx: Will it break https://github.com/PingCAP-QE/artifacts/tree/main/packages#whats-the-golang-version-the-builders-are-using?

lhy1024 (approved):

PR #9601, opened 2025-07-30
mcs: refactor election member interface (part 2)

rleungx:

Issue Number: ref #5766.

```commit-message
```

### Check List

Tests

- Unit test
### Release note

```release-note
None.
```

### DIFF
--- a/pkg/member/member.go
+++ b/pkg/member/member.go
@@ -118,8 +118,8 @@ func (m *Member) Client() *clientv3.Client {
 	return m.client
 }
 
-// IsLeader returns whether the server is PD leader or not by checking its leadership's lease and leader info.
-func (m *Member) IsLeader() bool {
+// IsServing returns whether the server is PD leader or not by checking its leadership's lease and leader info.
+func (m *Member) IsServing() bool {
 	return m.leadership.Check() && m.GetLeader().GetMemberId() == m.member.GetMemberId()
 }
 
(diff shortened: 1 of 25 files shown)


pkg/member/member.go
```diff
@@ -272,9 +272,9 @@ func (m *Member) WatchLeader(ctx context.Context, leader *pdpb.Member, revision
 	m.unsetLeader()
 }
 
-// ResetLeader is used to reset the PD member's current leadership.
+// Resign is used to reset the PD member's current leadership.
```
lhy1024, line 275: reset? resign?

pkg/member/participant.go
```diff
@@ -240,9 +240,9 @@ func (p *Participant) WatchLeader(ctx context.Context, primary participant, revi
 	p.unsetPrimary()
 }
 
-// ResetLeader is used to reset the participant's current leadership.
+// Resign is used to reset the participant's current leadership.
```
lhy1024, line 243: ditto

lhy1024 (commented):
the rest LGTM

pkg/member/member.go
rleungx: Resign

lhy1024:
```suggestion
// Resign is used to resign the PD member's current leadership.
```

Issue #9596, opened 2025-07-29
Accelerate testing speed by reducing clusters


lhy1024:
## Enhancement Task
- [x] https://github.com/tikv/pd/pull/9550
- [x] https://github.com/tikv/pd/pull/9581

PR #9603, opened 2025-07-30
tests: use temp dir to ensure test isolation

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/9596

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/cluster.go
+++ b/tests/cluster.go
@@ -16,8 +16,10 @@ package tests
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tests/cluster.go
```diff
@@ -83,7 +85,16 @@ var zapLogOnce sync.Once
 
 // NewTestServer creates a new TestServer.
 func NewTestServer(ctx context.Context, cfg *config.Config, services []string) (*TestServer, error) {
-	//  disable the heartbeat async runner in test
+	// use temp dir to ensure test isolation.
+	if cfg.DataDir == "" || strings.HasPrefix(cfg.DataDir, "default.") {
+		dirPattern := fmt.Sprintf("pd-server-test-%s-", cfg.Name)
+		tempDir, err := os.MkdirTemp("", dirPattern)
```
rleungx: What if we have already set the data dir in the config parameter?

lhy1024: So we only use temp dir in this func when data dir is empty or default prefix

bufferflies (approved):

tests/cluster.go
bufferflies: where is used default prefix?

lhy1024: > where is used default prefix?

https://github.com/tikv/pd/blob/aa75cbbc452906ff205cdde430907423e01e9119/server/config/config.go#L363

rleungx:
```suggestion
		tempDir, err := os.MkdirTemp("", "pd_tests")
```

lhy1024: done

rleungx (approved):

PR #9606, opened 2025-07-31
fix(resourcemanager): avoid the default group from occupying unreasonable service limit

JmPotato:

Issue Number: ref #9296.

```commit-message
This PR ensures that, during the calculation of `proportionalFillRate`, the unlimited fill rate of the default resource group does not consume nearly the entire service limit quota.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -117,38 +117,21 @@ func (krgm *keyspaceResourceGroupManager) initDefaultResourceGroup() {
 	if ok {
 		return
 	}
-
-	var (
-		fillRate     uint64 = UnlimitedRate
-		burstLimit   int64  = UnlimitedBurstLimit
-		serviceLimit        = krgm.getServiceLimiter().getServiceLimit()
-	)
-	// If there is a service limit set, ensure the default resource group has a reasonable fill rate initially.
-	if serviceLimit > 0 {
-		fillRate = uint64(serviceLimit)
-		// Busrt limit will be handled by the `invalidateBurstability` method during the actual initialization.
-	}
 	defaultGroup := &ResourceGroup{
 		Name: DefaultResourceGroupName,
 		Mode: rmpb.GroupMode_RUMode,
 		RUSettings: &RequestUnitSettings{
 			RU: &GroupTokenBucket{
 				Settings: &rmpb.TokenLimitSettings{
-					FillRate:   fillRate,
-					BurstLimit: burstLimit,
+					FillRate:   UnlimitedRate,
+					BurstLimit: UnlimitedBurstLimit,
 				},
 			},
 		},
 		Priority: middlePriority,
 	}
 	if err := krgm.addResourceGroup(defaultGroup.IntoProtoResourceGroup(krgm.keyspaceID)); err != nil {
-		log.Warn("init default group failed",
-			zap.Uint32("keyspace-id", krgm.keyspaceID),
-			zap.Uint64("fill-rate", fillRate),
-			zap.Int64("burst-limit", burstLimit),
-			zap.Float64("service-limit", serviceLimit),
-			zap.Error(err),
-		)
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -117,21 +117,38 @@ func (krgm *keyspaceResourceGroupManager) initDefaultResourceGroup() {
...
+	}
 	defaultGroup := &ResourceGroup{
 		Name: DefaultResourceGroupName,
 		Mode: rmpb.GroupMode_RUMode,
 		RUSettings: &RequestUnitSettings{
 			RU: &GroupTokenBucket{
 				Settings: &rmpb.TokenLimitSettings{
-					FillRate:   UnlimitedRate,
-					BurstLimit: UnlimitedBurstLimit,
+					FillRate:   fillRate,
```
glorv: Shall we not change the default group's configuration but only change the override limit?  User can see this config in the  information_schema table `resource_groups`, we'd better keep its value `ULIMITED` by default.

glorv: BTW, from the offline discussion, we'd like to treat the default group as with fill_rate=0, burstable. So it's priority is like to be lower than the same priority but higher than lower priority group.

JmPotato: Please take another look at this PR. I achieved equivalent logic by handling the default group configuration specially during the conciliate process, ensuring that the default group configuration remains unchanged while also minimizing disruption to existing code paths.

lhy1024 (approved):

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -524,11 +525,21 @@ func (krgm *keyspaceResourceGroupManager) calculateDemandInfo(queue []*ResourceG
 		ruPerSec := ruTracker.getRUPerSec()
 		fillRateSetting := group.getFillRate(true)
 		burstLimitSetting := float64(group.getBurstLimit(true))
+		// Handle specifically if the fill rate and burst limit settings of the default resource group are set to unlimited (i.e., default settings).
+		// This ensures that, during the calculation of `proportionalFillRate`, the unlimited fill rate does not consume nearly the entire service limit quota.
+		if group.Name == DefaultResourceGroupName &&
+			fillRateSetting == UnlimitedRate &&
+			burstLimitSetting == UnlimitedBurstLimit {
+			// Override the fill rate setting of the default resource group to the service limit.
+			fillRateSetting = serviceLimit
```
glorv, line 534:
```suggestion
			fillRateSetting = 0
			burstLimitSetting = serviceLimit
```
override the fill rate to 0. This means `default` group has the lowest priority.

glorv (commented):
rest LGTM

pkg/mcs/resourcemanager/server/keyspace_manager.go
JmPotato: I have few questions about this change. First, a non-zero postive burst limit means it's not burstable, so this can't work with the current implementaion.

https://github.com/tikv/pd/blob/3f6597985720e786648e214e2897cd795a87c2c4/client/resource_group/controller/controller.go#L953-L955

Second, if you mean assigning the default resource group the lowest priority during conciliation, we could simply ensure it has a maximum `normalizedRUDemand` during sorting. Otherwise, a zero fill rate would prevent it from accumulating any RUs.

glorv: Oh, I misunderstand the burstLimitSetting. 

For the second question, if there is not enough ru for none burstable ru, then it's expected behavior that the default group will receive no ru. We should only allocate ru for the default group if there is available burstable ru. `serviceLimit` is still too large for the default group as most normal group's ru should be smaller than it.

PR #9623, opened 2025-08-04
mcs: reduce some unnecessary logs

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Manual test

```
[2025/08/04 14:30:53.601 +08:00] [INFO] [watcher.go:189] ["update scheduler config"] [name=balance-region-scheduler] [value="{\"disabled\":false,\"ranges\":[{\"end-key\":\"\",\"start-key\":\"\"}]}"]
[2025/08/04 14:30:53.601 +08:00] [ERROR] [etcdutil.go:594] ["put failed in watch loop"] [error="scheduler balance-region-scheduler is not existed"] [revision=9] [name=scheduling-scheduler-config-watcher] [watch-key
=/pd/7534614101345316697/scheduler_config] [event-kv-key=/pd/7534614101345316697/scheduler_config/balance-region-scheduler]
[2025/08/04 14:30:53.602 +08:00] [INFO] [watcher.go:189] ["update scheduler config"] [name=balance-region-scheduler] [value="{\"disabled\":false,\"ranges\":[{\"end-key\":\"\",\"start-key\":\"\"}]}"]
[2025/08/04 14:30:53.602 +08:00] [ERROR] [etcdutil.go:594] ["put failed in watch loop"] [error="scheduler balance-region-scheduler is not existed"] [revision=25] [name=scheduling-scheduler-config-watcher] [watch-ke
y=/pd/7534614101345316697/scheduler_config] [event-kv-key=/pd/7534614101345316697/scheduler_config/balance-region-scheduler]
```

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/scheduling/server/config/watcher.go
+++ b/pkg/mcs/scheduling/server/config/watcher.go
@@ -17,6 +17,7 @@ package config
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -29,6 +30,7 @@ import (
 
 	"github.com/pingcap/log"
 
+	"github.com/tikv/pd/pkg/errs"
 	sc "github.com/tikv/pd/pkg/schedule/config"
 	"github.com/tikv/pd/pkg/schedule/schedulers"
 	"github.com/tikv/pd/pkg/storage"
@@ -198,7 +200,16 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 		}
 		// Ensure the scheduler config could be updated as soon as possible.
 		if sc := cw.getSchedulersController(); sc != nil {
-			return sc.ReloadSchedulerConfig(name)
+			err1 := sc.ReloadSchedulerConfig(name)
+			if errors.Is(err1, errs.ErrSchedulerNotFound) {
+				// Mostly it is caused by the coordinator doesn't start running.
+				// But to prevent the scheduler is truly not found, we still add a debug log here.
+				log.Debug("failed to reload scheduler config, scheduler not found",
+					zap.String("scheduler-name", name),
+					zap.Error(err1))
+				return nil
+			}
+			return err1
 		}
 		return nil
 	}
--- a/pkg/schedule/schedulers/config.go
+++ b/pkg/schedule/schedulers/config.go
@@ -102,6 +102,6 @@ func (b *baseDefaultSchedulerConfig) setDisable(disabled bool) error {
 	b.Lock()
 	defer b.Unlock()
 	b.Disabled = disabled
-	log.Info("set scheduler disable", zap.Bool("disabled", disabled))
+	log.Debug("set scheduler disable", zap.Bool("disabled", disabled))
 	return b.save()
 }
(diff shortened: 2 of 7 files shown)


pkg/mcs/scheduling/server/config/watcher.go
```diff
@@ -198,7 +200,16 @@ func (cw *Watcher) initializeSchedulerConfigWatcher() error {
 		}
 		// Ensure the scheduler config could be updated as soon as possible.
 		if sc := cw.getSchedulersController(); sc != nil {
-			return sc.ReloadSchedulerConfig(name)
+			err1 := sc.ReloadSchedulerConfig(name)
+			if errors.Is(err1, errs.ErrSchedulerNotFound) {
```
lhy1024, line 204: How about checking whether coordinator starts running?

rleungx: We don't have such an interface now. Besides, adding it inside the watcher looks strange.

lhy1024: https://github.com/tikv/pd/blob/4d498809b238adb5829c0e6b0f7f7a580d9970f7/pkg/mcs/scheduling/server/cluster.go#L727-L729

rleungx: I think they are two different things.

lhy1024: Can we add an interface to detect whether the coordinator or schedulerController is running? I think IsPrepared is sufficient.
```
if errors.Is(err1, errs.ErrSchedulerNotFound) && !cluster.IsPrapared(){
return nil
} else {
log.Error(...)
}
```
I think this might be better. By using IsPrepared, we can reduce unnecessary log printing, and if ErrSchedulerNotFound really occurs, we can find it in the log on time. We have too many debug-level logs, so it is impossible to keep them enabled all the time.

pkg/schedule/schedulers/config.go
```diff
@@ -102,6 +102,6 @@ func (b *baseDefaultSchedulerConfig) setDisable(disabled bool) error {
 	b.Lock()
 	defer b.Unlock()
 	b.Disabled = disabled
-	log.Info("set scheduler disable", zap.Bool("disabled", disabled))
+	log.Debug("set scheduler disable", zap.Bool("disabled", disabled))
```
Tema, line 105: I feel it is pretty important to print INFO in case of scheduler was actually disabled. Maybe print it if b.Disabled is true and disabled is false?

rleungx: Usually, we don't use config to set the scheduler to disable directly. If we remove a scheduler, it won't print any log after restarting PD, either.

okJiang: There is really no need to print this log, you can see the comments here

https://github.com/tikv/pd/blob/a1d89da95263b0407c6fb5d1c01d86f5e2329b45/pkg/schedule/schedulers/scheduler_controller.go#L186-L194

okJiang (approved):

pkg/mcs/scheduling/server/config/watcher.go
rleungx: Even if IsPrapared is true, it doesn't mean the scheduler is initialized.

lhy1024 (approved):

PR #9626, opened 2025-08-06
server/api: Return pseudo service safe point for gc_worker to provide compabitility for some existing tests of CDC

MyonKeminta:

Issue Number: Ref #8978

```commit-message
Make the service safe point HTTP API return a pseudo service safe point for "gc_worker". 
```

There's some existing integration tests of CDC that assumes the minimal service safe point returned by the command `pd-ctl service-gc-safepoint` should be the lower bound of readable snapshots. However, as new GC states API is adopted, the result of this command becomes 1-to-1 mapped from GC barriers,  and the thing whose semantics matches those tests' requirement becomes the txn safe point. As new GC states related HTTP APIs and pd-ctl commands are not yet implemented, we provide this compatibility logic to make them work.

### Check List

Tests

- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

\-

### Release note

```release-note
None.
```

### DIFF
--- a/server/api/service_gc_safepoint.go
+++ b/server/api/service_gc_safepoint.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"math"
 	"net/http"
 	"time"
 
@@ -23,6 +24,7 @@ import (
 
 	"github.com/tikv/pd/pkg/keyspace/constant"
 	"github.com/tikv/pd/pkg/storage/endpoint"
+	"github.com/tikv/pd/pkg/utils/keypath"
 	"github.com/tikv/pd/server"
 )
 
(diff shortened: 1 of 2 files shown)


server/api/service_gc_safepoint.go
```diff
@@ -63,11 +65,31 @@ func (h *serviceGCSafepointHandler) GetGCSafePoint(w http.ResponseWriter, _ *htt
...
+		ssps = append(ssps, equivalentSsp)
+		if equivalentSsp.ServiceID == keypath.GCWorkerServiceSafePointID {
+			gcWorkerSsp = equivalentSsp
+		}
+	}
+	if gcWorkerSsp == nil {
+		// Generate a pseudo service safe point for GC worker. While GC worker's service safe point won't exist after
+		// the new GC states API is adopted, some existing tests relies on the fact to pass: the minimal service safe
+		// point indicates the lower bound of the timestamp that is safe to read. Now, this value becomes the txn safe
+		// point. We generate a pseudo service safe point with the value of txn safe point to provide the compatibility.
```
rleungx, line 80: Will it be removed later if we introduce the new pd-ctl cmd?

MyonKeminta: If necessary, we can remove the service-gc-safepoint API entirely. But I think it doesn't make sense to remove the compatibility logic while keeping the command.

MyonKeminta:
@tiancaiamao The problem is feedback from CDC team due to test failure. From the perspective of a component that's using old APIs or commands, the "gc_worker"'s service safe point disappears as we are developing the new API, and it comes back again after this PR. Also considering the fact that the mapping between "gc_worker"'s service safe point and the txn safe point is also adopted in the `CompatibleUpdateServiceGCSafePoint`, it shouldn't have extra impact.
Also, of course, I'll take care whether there's any further feedback.

JmPotato (approved):

Issue #9481, opened 2025-07-03
TestStores unstable and panic


okJiang:
### Which jobs are failing

```
    panic.go:262: test panicked: runtime error: invalid memory address or nil pointer dereference
        goroutine 217646 [running]:
        runtime/debug.Stack()
        	/opt/hostedtoolcache/go/1.23.10/x64/src/runtime/debug/stack.go:26 +0x67
        github.com/stretchr/testify/suite.failOnPanic(0xc003ef3520, {0x49300e0, 0x7174f20})
        	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:89 +0x5b
        github.com/stretchr/testify/suite.Run.func1.1()
        	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:188 +0x365
        panic({0x49300e0?, 0x7174f20?})
        	/opt/hostedtoolcache/go/1.23.10/x64/src/runtime/panic.go:791 +0x132
        github.com/tikv/pd/pkg/core.(*StoreInfo).Clone(0x0, {0xc0054831a8, 0x4, 0x0?})
        	/home/runner/work/pd/pd/pkg/core/store.go:111 +0x92
        github.com/tikv/pd/tests.MustPutStore(0xc007ff1ec0, 0xc00d26f680, 0xc0052b0aa0)
        	/home/runner/work/pd/pd/tests/testutil.go:234 +0xc34
        github.com/tikv/pd/tests/integrations/mcs/scheduling_test.(*apiTestSuite).checkStores(0xc00067a550, 0xc00d26f680)
        	/home/runner/work/pd/pd/tests/integrations/mcs/scheduling/api_test.go:661 +0x6e5
        github.com/tikv/pd/tests.(*SchedulingTestEnvironment).RunTestInMicroserviceEnv(0xc00065b9e0, 0xc007ba5508)
        	/home/runner/work/pd/pd/tests/testutil.go:397 +0x324
        github.com/tikv/pd/tests/integrations/mcs/scheduling_test.(*apiTestSuite).TestStores(0xc00067a550)
        	/home/runner/work/pd/pd/tests/integrations/mcs/scheduling/api_test.go:617 +0x70
        reflect.Value.call({0xc0012c5ac0?, 0xc000091df0?, 0x13?}, {0x4cbac66, 0x4}, {0xc007ba5ea8, 0x1, 0x1?})
[...]

PR #9634, opened 2025-08-08
testutil: fix LastHeartbeat assignment and streamline PutMetaStore call

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9481

`NodeState_Removed` store may be removed from memory due to its `LastHeartbeat` is zero

https://github.com/tikv/pd/blob/a1d89da95263b0407c6fb5d1c01d86f5e2329b45/server/cluster/cluster.go#L1856-L1861

```commit-message
test: fix flaky test that store may panic during test
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/testutil.go
+++ b/tests/testutil.go
@@ -227,11 +227,10 @@ func MustPutStore(re *require.Assertions, tc *TestCluster, store *metapb.Store) 
 		// Wait for the raft cluster on the leader to be bootstrapped.
 		return raftCluster != nil && raftCluster.IsRunning()
 	})
-	re.NoError(raftCluster.PutMetaStore(store))
-	ts := store.GetLastHeartbeat()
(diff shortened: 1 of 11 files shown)


tests/testutil.go
```diff
@@ -227,6 +227,9 @@ func MustPutStore(re *require.Assertions, tc *TestCluster, store *metapb.Store)
 		// Wait for the raft cluster on the leader to be bootstrapped.
 		return raftCluster != nil && raftCluster.IsRunning()
 	})
+	if store.LastHeartbeat == 0 {
```
bufferflies, line 230: How about deleting some repeated code if it exists?

lhy1024 (approved):

PR #9654, opened 2025-08-12
resource_manager; fix the bug that ru tokens is not limit when available service limit is 0

### LINKED ISSUE #9651: resourcemanager: service limit can be limited occationally
### What did you do?

Image

There is no user configured resource group and the service limit is set to 10k. When running a stable workload that can consume all the ru, I found the ru consumption rate can exceed the service limit sometimes.

glorv:

Issue Number: Close #9651

- Fix the bug that ru tokens is not limit when service limiter's available tokens is 0
- Add a gauge metrics for service limit
- Also track some extended metrics for the default resource group.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -530,7 +530,7 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 				if err != nil {
 					continue
 				}
-				setServiceLimitMetrics(keyspaceName, krgm.getServiceLimiter().GetServiceLimit())
+				setOrRemoveServiceLimitMetrics(keyspaceName, krgm.getServiceLimiter().GetServiceLimit())
 
 				for _, group := range krgm.getResourceGroupList(true, true) {
 					groupName := group.Name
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -530,14 +530,12 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 				if err != nil {
 					continue
 				}
+				setServiceLimitMetrics(keyspaceName, krgm.getServiceLimiter().GetServiceLimit())
```
JmPotato: Should we only record service limits that are greater than or equal to zero?

pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -91,6 +91,13 @@ func (krl *serviceLimiter) setServiceLimit(newServiceLimit float64) {
 	}
 }
 
+// GetServiceLimit return the service limit value of this keyspace.
+func (krl *serviceLimiter) GetServiceLimit() float64 {
```
JmPotato, line 95:
```suggestion
func (krl *serviceLimiter) GetServiceLimit() float64 {
		if krl == nil {
				return 0.0
		}
```

JmPotato (approved):
LGTM. Please fix the statics check.

PR #9663, opened 2025-08-13
tsoutil: optimize TSO proxy request processing to improve stability and performance

### LINKED ISSUE #9662: Enhance TSO Request Processing to Address Blocking Issues and Improve Reliability
### Problem Statement:

In our deployment scenario, TiKV serves as the metadata storage for a large-scale distributed file system with thousands of clients concurrently requesting TSO timestamps through PD followers.

The current implementation in `PDProtoRequest.postProcess()` sequentially sent responses directly to the gRPC stream of thousands of clients using `r.stream.Send(response)`. 
Under this high-concurrency workload, when some clients were slow to process responses or experienced network latency, this direct sending mechanism would block the TSO dispatcher goroutine.

This blocking behavior had severe consequences:
1. It caused cascading delays throughout the TSO processing pipeline
2. Led to increased latency for all TSO proxy requests

### Solution:

We introduced a buffered channel for each TSO stream to resolve these issues:

1. Added a buffered channel `tsoRespCh` with capacity 2 in the gRPC service for each TSO stream
2. Modified `PDProtoRequest.postProcess()` to send responses to the channel instead of directly to the stream
3. Implemented a select statement with a default case that drops responses when the channel is full, preventing blocking

These enhancements maintain the non-blocking benefits while improving reliability and observability, especially for our high-concurrency deployment with thousands of concurrent clients.

felixYing:

Issue Number: Close #9662

We introduced a buffered channel for each TSO stream to resolve these issues:

1. Added a block channel (or a buffered channel tsoRespCh with capacity 2) in the gRPC service for each TSO stream
2. Modified PDProtoRequest.postProcess() to send responses to the channel instead of directly to the stream
3. Implemented a select statement with a default case that drops responses when the tso goroutine gone( or channel is full), preventing blocking

```commit-message
```

before this pr
image

after this pr
image

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/tsoutil/tso_request.go
+++ b/pkg/utils/tsoutil/tso_request.go
@@ -15,9 +15,11 @@
 package tsoutil
 
 import (
+	"go.uber.org/zap"
 	"google.golang.org/grpc"
 
 	"github.com/pingcap/kvproto/pkg/pdpb"
+	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/keyspace"
 	"github.com/tikv/pd/pkg/keyspace/constant"
@@ -45,15 +47,17 @@ type PDProtoRequest struct {
 	clientConn    *grpc.ClientConn
 	request       *pdpb.TsoRequest
 	stream        pdpb.PD_TsoServer
+	tsoRespCh     chan *pdpb.TsoResponse
 }
 
 // NewPDProtoRequest creates a PDProtoRequest and returns as a Request
-func NewPDProtoRequest(forwardedHost string, clientConn *grpc.ClientConn, request *pdpb.TsoRequest, stream pdpb.PD_TsoServer) Request {
+func NewPDProtoRequest(forwardedHost string, clientConn *grpc.ClientConn, request *pdpb.TsoRequest, stream pdpb.PD_TsoServer, tsoRespCh chan *pdpb.TsoResponse) Request {
 	tsoRequest := &PDProtoRequest{
 		forwardedHost: forwardedHost,
 		clientConn:    clientConn,
 		request:       request,
 		stream:        stream,
+		tsoRespCh:     tsoRespCh,
 	}
 	return tsoRequest
 }
@@ -93,9 +97,11 @@ func (r *PDProtoRequest) postProcess(countSum, physical, firstLogical int64) (in
 			Logical:  firstLogical + countSum,
 		},
 	}
-	// Send back to the client.
-	if err := r.stream.Send(response); err != nil {
-		return countSum, err
+	select {
+	case r.tsoRespCh <- response:
+	default:
+		// If the channel is full, we drop the response to avoid blocking. and return nil error
+		log.Warn("tso response channel is full, drop the response", zap.Uint32("count", count), zap.Int64("physical", physical), zap.Int64("firstLogical", firstLogical))
 	}
 	return countSum, nil
 }
--- a/pkg/utils/tsoutil/tso_request.go
+++ b/pkg/utils/tsoutil/tso_request.go
@@ -15,6 +15,8 @@
 package tsoutil
 
 import (
+	"context"
+
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 
@@ -38,7 +40,7 @@ type Request interface {
 	// count defines the count of timestamps to retrieve.
 	process(forwardStream stream, count uint32) (tsoResp, error)
 	// postProcess sends the response back to the sender of the request
-	postProcess(countSum, physical, firstLogical int64) (int64, error)
+	postProcess(ctx context.Context, countSum, physical, firstLogical int64) (int64, error)
 }
 
 // PDProtoRequest wraps the request and stream channel in the PD grpc service
@@ -86,7 +88,7 @@ func (r *PDProtoRequest) process(forwardStream stream, count uint32) (tsoResp, e
 }
 
 // postProcess sends the response back to the sender of the request
-func (r *PDProtoRequest) postProcess(countSum, physical, firstLogical int64) (int64, error) {
+func (r *PDProtoRequest) postProcess(ctx context.Context, countSum, physical, firstLogical int64) (int64, error) {
 	count := r.request.GetCount()
 	countSum += int64(count)
 	response := &pdpb.TsoResponse{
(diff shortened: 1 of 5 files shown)


pkg/utils/tsoutil/tso_request.go
```diff
@@ -154,9 +158,11 @@ func (r *PDProtoRequest) postProcess(countSum, physical, firstLogical int64) (in
...
 		},
 	}
-	// Send back to the client.
-	if err := r.stream.Send(response); err != nil {
-		return countSum, err
+	select {
+	case r.tsoRespCh <- response:
+	default:
+		// If the channel is full, we drop the response to avoid blocking. and return nil error
+		log.Warn("tso response channel is full, drop the response", zap.Uint32("count", count), zap.Int64("physical", physical), zap.Int64("firstLogical", firstLogical))
```
JmPotato: I'm a bit unclear about the point here. Previously, we could handle multiple requests from different clients simultaneously, though each client responded back in a blocking, sequential manner. In the current PR, if any response sent through `tsoRespCh` from a client is delayed or blocked, wouldn't that cause all other responses to be unexpectedly dropped if we have > 2 requests processing at the same time?

felixYing: I originally thought that a client sends requests and receives responses serially within a stream. We might need to re-evaluate the capacity of tsoRespCh here.

The intention of this PR is to prevent individual clients from causing all TSO requests to slow down, and if processRequests exceeds 3s, NewTSDeadline will reset the entire dispatch goroutine.

If clients could limit the number of pending TSO requests per TSO stream, that would be a better solution.

felixYing: Or in func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error:

1. Set the capacity of tsoRespCh to 100
2. If len(tsoRespCh) > 50 is detected, temporarily suspend receiving new TSO requests.

JmPotato: After re-reading the code, I withdraw my previous doubts. Your understanding is correct: each Tso's gRPC corresponds to a separate client stream, and because the client sends TSO requests in a request-response manner, each `tsoRespCh` can only correspond to one client's request at the same time. Therefore, it seems we can design it as a block channel without considering buffer size, since we can fully utilize the existing asynchronous mechanism by using each Tso stream to send back the response for the current stream.

felixYing: yeah, I see in the comments for checkTSORPCConcurrency, when EnableTSOFollowerProxyis enabled, the concurrency level of the TSO stream is set to 1. However, I still have a question: ​​Why is a blocking channel preferred here, rather than a channel with a buffer size of 1?​

felixYing: @JmPotato could add the ok-to-test label?

pkg/utils/tsoutil/tso_request.go
```diff
@@ -154,9 +158,11 @@ func (r *PDProtoRequest) postProcess(countSum, physical, firstLogical int64) (in
 			Logical:  firstLogical + countSum,
 		},
 	}
-	// Send back to the client.
-	if err := r.stream.Send(response); err != nil {
-		return countSum, err
+	select {
+	case r.tsoRespCh <- response:
+	default:
```
JmPotato: Why still keeping this default branch if it’s already a blocked channel?

server/grpc_service.go
```diff
@@ -561,9 +563,17 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 				return errors.WithStack(err)
 			}
 
-			tsoRequest := tsoutil.NewPDProtoRequest(forwardedHost, clientConn, request, stream)
+			tsoRequest := tsoutil.NewPDProtoRequest(forwardedHost, clientConn, request, stream, tsoRespCh)
 			// don't pass a stream context here as dispatcher serves multiple streams
 			tsoRequestProxyCtx = s.tsoDispatcher.DispatchRequest(s.ctx, tsoRequest, s.pdProtoFactory, s.tsoPrimaryWatcher)
+			select {
+			case response := <-tsoRespCh:
```
JmPotato, line 571: We may need to add metrics here to track and measure response latency from the channel.

pkg/utils/tsoutil/tso_request.go
felixYing: The following scenarios may cause the tso_dispatcher goroutine to block:

1. When the PD follower takes more than 3 seconds to obtain TSO from the PD leader, then sends responses to all requesters' tsoRespCh channels
2. WatchTSDeadline detects a 3-second timeout and calls the tsoDispatcher's cancel method
3. In the client's TSO stream goroutine, while waiting for tsoRespCh, it first receives a cancel signal, causing the client TSO stream goroutine to exit
4. . However, the tsoDispatcher's goroutine is still waiting to send responses to tsoRespCh, which causes it to block at this point.

3 code in func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {

```go

			tsoRequestProxyCtx = s.tsoDispatcher.DispatchRequest(s.ctx, tsoRequest, s.pdProtoFactory, s.tsoPrimaryWatcher)
			select {
			case response := <-tsoRespCh:
				if err = stream.Send(response); err != nil {
					return errors.WithStack(err)
				}
			case <-tsoRequestProxyCtx.Done():
				return errors.WithStack(context.Cause(tsoRequestProxyCtx))
			}
```

pkg/utils/tsoutil/tso_request.go
okJiang: Since here are PD followers, I think it is safe to start a new goroutine for each client to send responses. The clients will not interfere with each other and there is no need to worry about responses being dropped.

pkg/utils/tsoutil/tso_request.go
JmPotato: Then what about watching the same context here rather than just noop?

felixYing: You mean adding tsoRequestProxyQueue.ctx to NewPDProtoRequest as well?
And then also watch this ctx in PDProtoRequest.postProcess?

```go
select {
case r.tsoRespCh <- response:
    // Response sent successfully
case <-r.dispatcherCtx.Done():
   log.Warn("tso dispatcher done, drop the response")
}
```

felixYing: or add ctx in 
func (r *TSOProtoRequest) postProcess(dispatcherCtx context.Context, countSum, physical, firstLogical int64) (int64, error) {

server/grpc_service.go
felixYing: done

pkg/utils/tsoutil/tso_request.go
okJiang: cc @felixYing @JmPotato

JmPotato (commented):
The overall LGTM. Can you add at least one test case to cover the latest changes?

Issue #9628, opened 2025-08-06
Improve the performance for the region tree


bufferflies:
## Enhancement Task
Currently, all operators of the region tree share the same RWLock. All the region queries will be affected if one operator needs more time. 
Such as TABLE_STORAGE_STATS memory table

PR #9664, opened 2025-08-14
statistics: scan region using limit

bufferflies:

Issue Number: Ref #9628

```commit-message
StatsRegion use limit to scan the region tree, avoid to lock the tree too long one time 
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -58,7 +58,6 @@ import (
 	"github.com/tikv/pd/pkg/schedule/keyrange"
 	"github.com/tikv/pd/pkg/schedule/labeler"
 	"github.com/tikv/pd/pkg/schedule/placement"
-	"github.com/tikv/pd/pkg/slice"
 	"github.com/tikv/pd/pkg/statistics"
 	"github.com/tikv/pd/pkg/statistics/utils"
 	"github.com/tikv/pd/pkg/storage"
@@ -68,6 +67,7 @@ import (
 	"github.com/tikv/pd/pkg/utils/apiutil"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
 	"github.com/tikv/pd/pkg/utils/keypath"
+	"github.com/tikv/pd/pkg/utils/keyutil"
 	"github.com/tikv/pd/pkg/utils/logutil"
 	"github.com/tikv/pd/pkg/utils/netutil"
 	"github.com/tikv/pd/pkg/utils/syncutil"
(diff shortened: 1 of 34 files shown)


rleungx (approved):

server/cluster/cluster.go
```diff
@@ -119,6 +119,9 @@ const (
 	miscTaskRunner       = "misc-async"
 	logTaskRunner        = "log-async"
 	syncRegionTaskRunner = "sync-region-async"
+
+	// defaultScanRegionsLimit is the default limit for scanning regions.
+	defaultScanRegionsLimit = 100
```
lhy1024: Why choose 100? Do we need to add a bench?

server/cluster/cluster.go
```diff
@@ -2224,8 +2239,18 @@ func (c *RaftCluster) GetHotRegionStatusByRange(startKey, endKey []byte, engine
...
+		regions := c.ScanRegions(startKey, endKey, defaultScanRegionsLimit)
+		for _, region := range regions {
+			stats.Observe(region, nil, opts...)
+		}
+		if len(regions) < defaultScanRegionsLimit {
+			break
+		}
+		startKey = regions[len(regions)-1].GetEndKey()
+	}
+	return stats
```
lhy1024: They are similiar to L2186-2198, could we reduce duplicate code?

server/cluster/cluster.go
hawkingrei: +1, Can we add a variable to control it?

bufferflies: removed it by using the core.ScanRegionLimit.

server/cluster/cluster.go
bufferflies: good suggestion, fixed

pkg/core/region.go
```diff
@@ -47,8 +47,9 @@ import (
 
 const (
 	randomRegionMaxRetry = 10
-	scanRegionLimit      = 1000
-	batchSearchSize      = 16
+	// ScanRegionLimit is the minimum number of regions to scan in a region scan request.
```
lhy1024:
```suggestion
	// ScanRegionLimit is the default limit for the number of regions to scan in a region scan request.
```

bufferflies: fixed

lhy1024: Where is the new commit?

bufferflies: forgot to submit new commit,

lhy1024 (approved):

PR #9669, opened 2025-08-18
feat(resourcemanager): migrate service limiter to GCRA-based rate limit

### LINKED ISSUE #9668: Service limit suffer from short-term RU spikes due to frequent token requests
### What did you do?

Run workload with service limit set.

Under no circumstances should the service limit be exceeded.

### What did you see instead?

If a large number of KV request retries occur within a period of time, the service limit may be exceeded.

Image

Image

Image

### Root cause

This happens because `refillTokensLocked` allows accumulation up to `ServiceLimit × serviceLimiterBurstFactor` and `applyServiceLimit` can grant all available tokens in a single request.

A typical trigger scenario is when the same resource group issues multiple RU requests in quick succession within a short time window. If the token bucket has accumulated multiple seconds’ worth of quota, all of it can be consumed almost instantly, causing an instantaneous RU usage higher than the configured ServiceLimit.

Even with `serviceLimiterBurstFactor = 5.0` intended for a 5s request interval, all accumulated tokens can be consumed in less than one second, resulting in observed spikes (e.g., 12K RU/s when limit is 10K).

JmPotato:

Issue Number: close #9668, ref #9296.

```commit-message
Replace token bucket logic (`AvailableTokens`) with Generic Cell Rate Algorithm (GCRA).
This should provide a more accurate and consistent service rate limiting mechanism while maintaining persistence and API compatibility at higher layers.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/metrics.go
+++ b/pkg/mcs/resourcemanager/server/metrics.go
@@ -39,6 +39,9 @@ const (
 	keyspaceNameLabel         = "keyspace_name"
 	fillRateLabel             = "fill_rate"
 	burstLimitLabel           = "burst_limit"
+	serviceLimitLabel         = "service_limit"
+	tatLabel                  = "tat"
+	slackLabel                = "slack"
 
 	// Labels for the config.
 	ruPerSecLabel   = "ru_per_sec"
@@ -169,7 +172,7 @@ var (
 			Subsystem: ruSubsystem,
 			Name:      "service_limit",
 			Help:      "Gauge of the total RU limit of specific keyspace.",
-		}, []string{keyspaceNameLabel})
+		}, []string{keyspaceNameLabel, typeLabel})
 )
 
 type metrics struct {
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -438,7 +438,7 @@ func (gtb *GroupTokenBucket) inspectAnomalies(
 ) bool {
 	var errMsg string
 	// Verify whether the allocated token is invalid, such as negative values, math.Inf, or math.NaN.
-	if tb.Tokens <= 0 || math.IsInf(tb.Tokens, 0) || math.IsNaN(tb.Tokens) {
```
lhy1024, line 441: Is it expected to remove an equal zero judge?

pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -134,41 +131,45 @@ func (krl *serviceLimiter) applyServiceLimit(
...
+	if now.After(krl.TAT) {
+		base = now
 	}
-
-	// If the requested tokens is greater than the available tokens, grant all available tokens.
-	if requestedTokens > krl.AvailableTokens {
-		limitedTokens = math.Max(0, krl.AvailableTokens)
-		// TODO: allow the loan to decrease the allocation at a smooth rate.
-		krl.AvailableTokens = 0
+	// TODO: smooth the burst window in a cold start scenario.
```
lhy1024, line 146: If the physical clock jumps, will it also meet a problem?

pkg/mcs/resourcemanager/server/token_buckets.go
JmPotato: Yes. Otherwise, once the service limit is reached, the granted token will always be zero, causing the token bucket to reset unexpectedly.

pkg/mcs/resourcemanager/server/service_limit.go
JmPotato: According to the current implementation, yes, but considering this is a rather extreme case, we might think of a separate solution for it later. For now, I don’t plan to handle this situation in this PR.

pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -423,11 +426,16 @@ func (m *gaugeMetrics) setSampledRUPerSec(ruPerSec float64) {
 	m.sampledRequestUnitPerSecGauge.Set(ruPerSec)
 }
 
-func setOrRemoveServiceLimitMetrics(keyspaceName string, limit float64) {
+func setServiceLimitMetrics(keyspaceName string, sl *serviceLimiter) {
+	limit, tat, slack := sl.getStates()
 	if limit > 0 {
-		serviceLimit.WithLabelValues(keyspaceName).Set(limit)
+		serviceLimit.WithLabelValues(keyspaceName, serviceLimitLabel).Set(limit)
+		serviceLimit.WithLabelValues(keyspaceName, tatLabel).Set(float64(tat.UnixNano() / int64(time.Millisecond)))
```
lhy1024: It will be difficult to understand, how about using `tat.Sub(now).Millisecond()`

glorv (commented):
overall LGTM

could you add the test result and the reference of the algarithm?

PR #9671, opened 2025-08-19
core: caller can't scan too much regions in one lock even if limit equals -1 

### LINKED ISSUE #9674: API: avoid to scan too much regions in one lock
## Enhancement Task
If one caller uses a big limit, it may take the lock too long, and affect the other short requests, such as GetRegion/GetRegionByID

bufferflies:

Issue Number: Close #9674, Ref #9628

```commit-message
The caller can only scan the fixed regions in one lock.
```

### Check List

Tests

- Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -2018,13 +2018,13 @@ func (r *RegionsInfo) GetRegionCount(startKey, endKey []byte) int {
 	return r.tree.GetCountByRange(startKey, endKey)
 }
 
-func (r *RegionsInfo) scanRegionsInner(startKey, endKey []byte, limit int, fns ...scanRegionFunc) []*RegionInfo {
+func (r *RegionsInfo) scanRegionsInner(startKey, endKey []byte, limit int, fns ...scanRegionFunc) ([]*RegionInfo, error) {
 	var res []*RegionInfo
 	for {
 		scanLimit := limit - len(res)
 		regions, err := r.scanRegionsOnce(startKey, endKey, scanLimit, fns...)
 		if err != nil {
-			break
+			return nil, err
 		}
 		res = append(res, regions...)
 		if limit >= 0 && len(res) >= limit {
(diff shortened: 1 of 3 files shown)


okJiang (commented):
It is better to update `BatchScanRegions` together

bufferflies:
> It is better to update `BatchScanRegions` together

good idea, done

okJiang (approved):

pkg/core/region.go
```diff
@@ -2018,25 +2018,36 @@ func (r *RegionsInfo) GetRegionCount(startKey, endKey []byte) int {
...
-	defer r.t.RUnlock()
+func (r *RegionsInfo) scanRegionsInner(startKey, endKey []byte, limit int, fns ...scanRegionFunc) []*RegionInfo {
 	var res []*RegionInfo
-	r.tree.scanRange(startKey, func(region *RegionInfo) bool {
-		if len(endKey) > 0 && bytes.Compare(region.GetStartKey(), endKey) >= 0 {
-			return false
+	for {
+		scanLimit := limit - len(res)
+		regions, err := r.scanRegionsOnce(startKey, endKey, scanLimit, fns...)
+		if err != nil {
```
rleungx, line 2026: The error is ignored.

pkg/core/region.go
```diff
@@ -2018,25 +2018,36 @@ func (r *RegionsInfo) GetRegionCount(startKey, endKey []byte) int {
...
+			break
 		}
-		if limit > 0 && len(res) >= limit {
-			return false
+		res = append(res, regions...)
+		if limit >= 0 && len(res) >= limit {
+			break
```
rleungx, line 2031: Why break here?

bufferflies: The count of the regions exceeds than the given limit

pkg/core/region.go
bufferflies: the error means that this scan met the hole range if it forbid no any hole

pkg/core/region.go
rleungx: If the count of regions equals to limit, will it also break?

bufferflies: yes

pkg/core/region.go
rleungx: Will the caller need this error?

pkg/core/region.go
rleungx: Oh, I have a misunderstanding about batch size and limit.

pkg/core/region.go
bufferflies: yes, good catch ,fixed

rleungx (approved):

PR #9675, opened 2025-08-20
store: refactor leader transfer pause logic to use atomic operations

okJiang:

Issue Number: Close https://github.com/tikv/pd/issues/9673

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/store.go
+++ b/pkg/core/store.go
@@ -17,6 +17,7 @@ package core
 import (
 	"math"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/docker/go-units"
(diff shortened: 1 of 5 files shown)


pkg/core/store.go
```diff
@@ -57,10 +58,10 @@ const (
 type StoreInfo struct {
 	meta *metapb.Store
 	*storeStats
-	pauseLeaderTransferIn  bool // not allow to be used as target of transfer leader
-	pauseLeaderTransferOut bool // not allow to be used as source of transfer leader
-	slowStoreEvicted       bool // this store has been evicted as a slow store, should not transfer leader to it
-	slowTrendEvicted       bool // this store has been evicted as a slow store by trend, should not transfer leader to it
+	pauseLeaderTransferIn  atomic.Int64 // not allow to be used as target of transfer leader
+	pauseLeaderTransferOut atomic.Int64 // not allow to be used as source of transfer leader
```
lhy1024, line 62: why not using atomic.bool

pkg/core/store_option.go
```diff
@@ -115,9 +115,9 @@ func PauseLeaderTransfer(d constant.Direction) StoreCreateOption {
 	return func(store *StoreInfo) {
```
lhy1024, line 115: Previously, we could call PauseLeaderTransfer multiple times and get the same result, but now its behavior has changed.

pkg/core/store_option.go
```diff
@@ -127,9 +127,9 @@ func ResumeLeaderTransfer(d constant.Direction) StoreCreateOption {
...
```
bufferflies, line 127: The assignment of StoreCreateOption should be changed. It is not always successful if there's a conflict

pkg/core/store.go
okJiang: After adding network slow store scheduler, it may be called multiple times. ref https://github.com/tikv/pd/pull/9451/files/dbd5a79cce34520d0e40a01033285784822ca2df#r2284518807

pkg/core/store_option.go
okJiang: Yes, Is there any logic behind this behavior?

Issue #9008, opened 2025-01-17
metrics: add grafana details of balance range scheduler


bufferflies:

bufferflies:
closed by https://github.com/tikv/pd/pull/9676

PR #9676, opened 2025-08-20
scheduler: stop balance range job if it's balanced

### LINKED ISSUE #9484: scheduler: finish job when the key range of the job range has been balanced
## Development Task
When the key range of the job range has been balanced, the balance-range scheduler can stop the running job and then run the next waiting job, not wait util the timeout finished

bufferflies:

Issue Number: Close #9484, #9006，#9008

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/schedulers/balance_range.go
+++ b/pkg/schedule/schedulers/balance_range.go
@@ -46,6 +46,7 @@ import (
 var (
 	defaultJobTimeout = 30 * time.Minute
 	reserveDuration   = 7 * 24 * time.Hour
+	reportInterval    = 10 * time.Second
 )
 
 type balanceRangeSchedulerHandler struct {
@@ -377,7 +378,7 @@ type balanceRangeScheduler struct {
 
 func (s *balanceRangeScheduler) report() {
 	now := time.Now()
-	if now.Sub(s.lastReportTime) < 10*time.Second {
+	if now.Sub(s.lastReportTime) < reportInterval {
 		return
 	}
 	for storeID, score := range s.scoreMap {


pkg/schedule/schedulers/balance_range.go
```diff
@@ -372,43 +372,114 @@ type balanceRangeScheduler struct {
...
+	balanceRangeJobGauge.WithLabelValues().Set(float64(s.job.JobID))
+	s.lastReportTime = now
 }
 
 // ServeHTTP implements the http.Handler interface.
 func (s *balanceRangeScheduler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	s.handler.ServeHTTP(w, r)
 }
 
+func (s *balanceRangeScheduler) isBalanced() bool {
```
rleungx, line 402: If there are two stores far from the expected score, but the diff is the same. Do we still think it is balanced?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -372,43 +372,114 @@ type balanceRangeScheduler struct {
 	expectScoreMap map[uint64]float64
 	job            *balanceRangeSchedulerJob
 	solver         *solver
+	lastReportTime time.Time
+}
+
+func (s *balanceRangeScheduler) report() {
+	now := time.Now()
+	if now.Sub(s.lastReportTime) < 10*time.Second {
```
lhy1024: how about using const value

pkg/schedule/schedulers/balance_range.go
```diff
@@ -372,43 +372,114 @@ type balanceRangeScheduler struct {
...
+}
+
 // IsScheduleAllowed checks if the scheduler is allowed to schedule new operators.
 func (s *balanceRangeScheduler) IsScheduleAllowed(cluster sche.SchedulerCluster) bool {
 	allowed := s.OpController.OperatorCount(operator.OpRange) < cluster.GetSchedulerConfig().GetRegionScheduleLimit()
 	if !allowed {
 		operator.IncOperatorLimitCounter(s.GetType(), operator.OpRange)
+		return false
 	}
 	if err := s.conf.gc(); err != nil {
```
rleungx, line 427: Shall we move the following code to the Schedule function?

pkg/schedule/schedulers/balance_range.go
```diff
@@ -372,43 +372,114 @@ type balanceRangeScheduler struct {
...
-	if job != nil {
-		if job.Status == pending {
-			if err := s.conf.begin(index); err != nil {
-				return false
-			}
-			km := cluster.GetKeyRangeManager()
-			km.Append(job.Ranges)
+	// all jobs are completed
+	if job == nil {
+		return false
```
rleungx, line 439: Will the scheduler be deleted if all jobs are completed?

bufferflies: no, the user can query the job status history

pkg/schedule/schedulers/balance_range.go
bufferflies: It doesn't happen, the diff is the math.abs(diff-epect)

pkg/schedule/schedulers/balance_range.go
bufferflies: good suggestion, I will move this codes to new function

pkg/schedule/schedulers/balance_range.go
bufferflies: done

PR #9684, opened 2025-08-22
feat(resourcemanager): add trickle-time enforcement to service limiter

### LINKED ISSUE #9668: Service limit suffer from short-term RU spikes due to frequent token requests

JmPotato:

Issue Number: close #9668, ref #9296.

```commit-message
Enhance `serviceLimiter.applyServiceLimit` to return both limited tokens and a minimum trickle time, ensuring client-side rate does not exceed service limit.
```

### Check List

Tests

- Unit test
- Integration test

image

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/service_limit.go
+++ b/pkg/mcs/resourcemanager/server/service_limit.go
@@ -159,7 +159,7 @@ func (krl *serviceLimiter) applyServiceLimit(
 	}
 
 	// Calculate a minimum trickle time to ensure the granted tokens' rate in client won't exceed the service limit.
-	minTrickleTimeMs = int64(limitedTokens * 1000.0 / krl.getServiceLimitLocked())
+	minTrickleTimeMs = int64(math.Round(limitedTokens * 1000.0 / krl.getServiceLimitLocked()))
 
 	return limitedTokens, minTrickleTimeMs
 }
--- a/pkg/mcs/resourcemanager/server/service_limit.go
+++ b/pkg/mcs/resourcemanager/server/service_limit.go
@@ -91,13 +91,16 @@ func (krl *serviceLimiter) setServiceLimit(newServiceLimit float64) {
 	}
 }
 
-// GetServiceLimit return the service limit value of this keyspace.
-func (krl *serviceLimiter) GetServiceLimit() float64 {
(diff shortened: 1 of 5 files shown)


pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -130,35 +133,35 @@ func (krl *serviceLimiter) refillTokensLocked(now time.Time) {
...
+	} else {
+		// If the requested tokens is greater than the available tokens, grant all available tokens.
 		limitedTokens = math.Max(0, krl.AvailableTokens)
-		// TODO: allow the loan to decrease the allocation at a smooth rate.
 		krl.AvailableTokens = 0
 	}
 
-	return limitedTokens
+	// Calculate a minimum trickle time to ensure the granted tokens' rate in client won't exceed the service limit.
+	minTrickleTimeMs = int64(limitedTokens * 1000.0 / krl.getServiceLimitLocked())
```
lhy1024: If `limitedTokens * 1000.0 / krl.getServiceLimitLocked()` is 9.9999, minTrickleTimeMs will be 9 rather than 10. How about using `math.Ceil`?

JmPotato: Nice catch. I changed it to use `math.Round` here.

lhy1024 (approved):

PR #9710, opened 2025-09-01
scheduler: slow score reaches 1, then wait for the recoveryGap before releasing the slow store.

okJiang:

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/schedulers/evict_slow_store.go
+++ b/pkg/schedule/schedulers/evict_slow_store.go
@@ -69,9 +69,12 @@ type evictSlowStoreSchedulerConfig struct {
 
 	cluster *core.BasicCluster
 	// Last timestamp of the chosen slow store for eviction.
-	lastSlowStoreCaptureTS     time.Time
-	isRecovered                bool
-	networkSlowStoreCaptureTSs map[uint64]time.Time
+	lastSlowStoreCaptureTS time.Time
+	isRecovered            bool
+	// The value should be nil when store is slow. Once the store's
+	// score becomes 1, the store prepares to recover and record the
+	// timestamp.
+	networkSlowStoreCaptureTSs map[uint64]*time.Time
 	// Duration gap for recovering the candidate, unit: s.
 	RecoverySec uint64 `json:"recovery-duration"`
 	// EvictedStores is only used by disk slow store scheduler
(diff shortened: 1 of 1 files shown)


pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -172,7 +175,7 @@ func (conf *evictSlowStoreSchedulerConfig) addNetworkSlowStoreLocked(storeID uin
 	oldValue := conf.PausedNetworkSlowStores
 	conf.PausedNetworkSlowStores = append(conf.PausedNetworkSlowStores, storeID)
 	oldCaptureTS, ok := conf.networkSlowStoreCaptureTSs[storeID]
-	conf.networkSlowStoreCaptureTSs[storeID] = time.Now()
+	conf.networkSlowStoreCaptureTSs[storeID] = nil
```
rleungx: Why set to nil?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -481,22 +488,33 @@ func (s *evictSlowStoreScheduler) tryRecoverNetworkSlowStores(cluster sche.Sched
...
+		avgScore := calculateAvgScore(networkSlowScores)
+
+		// Recover the slow store only if the recoveryGap time is
+		// continuously less than or equal to slowStoreRecoverThreshold.
+		if avgScore > slowStoreRecoverThreshold {
+			storesStillSlow = append(storesStillSlow, storeID)
+			continue
+		}
+
+		if startTime == nil {
```
rleungx, line 511: If we use nil, startTime is not a capture ts, it should be a recovery ts.

pkg/schedule/schedulers/evict_slow_store.go
okJiang: `nil` represents it is a slow store, but not recover in tikv
not `nil` represents it is a slow store, already recover in tikv, and ready to recover in pd.

pkg/schedule/schedulers/evict_slow_store.go
okJiang: I can rename it.

okJiang: Done rename https://github.com/tikv/pd/pull/9710/commits/aa792678ef69691837a5d1662635ab12544b9e4d

PR #9716, opened 2025-09-03
scheduler: try to evict leader from network slow store

okJiang:

In clusters with thousands of regions, leader election is slow and cannot complete leader election within ten minutes. The data shows that as the number of regions increases to more than 10,000, there will always be some regions with slower leader election. If network jitter persists, it will affect the cluster, so it is necessary to directly evict the leaders of the slow store.

However, if the slow store experiences only a very brief network jitter (only a few seconds), it is unnecessary to evict the leader. To address these two situations, we will introduce an algorithm.

- update the default value of `enable-network-slow-store`
- try to evict leader from network slow store if it meet the following requirement:
    - Before the tikv slow score reaches 1, this store's slow score triggers `networkSlowStoreEvictThreshold` 6 times
- Fix a bug: After PD leader transfer, network slow store will not be recovered

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -52,6 +52,18 @@ const (
...
 	networkSlowStoreIssueThreshold       = 95 // Threshold for detecting network issues with a specific store
 	networkSlowStoreFluctuationThreshold = 10 // Threshold for detecting network fluctuations with other stores
+
+	// Why 99? If the network is normal within the 10-second store heartbeat, the
+	// score will drop from 100 to: 100 - (100/10(min)/(60s/10s)) = 100 - (10/6) < 99. In
+	// other words, if the network is completely normal within 10 seconds, the score will
+	// be less than 99.
+	networkSlowStoreEvictThreshold = 99
+
+	// Why 20? The scheduler triggers every 3 seconds, and store heartbeat reports a score
```
rleungx: 3s？

okJiang: Sorry, I misremembered the interval. . . It seems I need to modify the trigger condition.😂

okJiang: I moved the logic to store heartbeat. This way we can accurately know whether the store has network jitter within 10 seconds. https://github.com/tikv/pd/pull/9716/commits/57688fc8e1bfdc3c63da9dfddd14060b9b51162f

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -64,61 +69,70 @@ const (
 	networkSlowStore slowStoreType = "network"
 )
 
+type networkSlowStoreStatus struct {
```
rleungx: If there is only one field, why do we need to wrap it?

server/cluster/cluster.go
```diff
@@ -2526,3 +2527,18 @@ func (c *RaftCluster) SetServiceIndependent(name string) {
...
+// other words, if the network is completely normal within 10 seconds, the score will
+// be less than 99.
+const networkSlowStoreEvictThreshold = 99
+
+func (c *RaftCluster) adjustNetworkSlowStore(storeID uint64) {
+	if c.GetAvgNetworkSlowScore(storeID) >= networkSlowStoreEvictThreshold {
+		c.TriggerNetworkSlowEvict(storeID)
+	}
+	if c.GetAvgNetworkSlowScore(storeID) <= 1 {
+		c.ResetTriggerNetworkSlowEvict(storeID)
```
rleungx, line 2547: Please also add a comment to explain how long the count will be reset.

rleungx (approved):

pkg/core/store.go
```diff
@@ -160,23 +162,24 @@ func (s *StoreInfo) Feedback(e float64) {
...
+		leaderSize:                     s.leaderSize,
+		regionSize:                     s.regionSize,
+		pendingPeerCount:               s.pendingPeerCount,
+		lastPersistTime:                s.lastPersistTime,
+		leaderWeight:                   s.leaderWeight,
+		regionWeight:                   s.regionWeight,
+		limiter:                        s.limiter,
+		minResolvedTS:                  s.minResolvedTS,
+		lastAwakenTime:                 s.lastAwakenTime,
+		triggeredNetworkSlowEvictCount: s.triggeredNetworkSlowEvictCount,
```
rleungx: How about networkSlowTriggers?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -129,13 +141,11 @@ func (conf *evictSlowStoreSchedulerConfig) persistLocked(updateFn func()) error
 func (conf *evictSlowStoreSchedulerConfig) getStores() []uint64 {
 	conf.RLock()
 	defer conf.RUnlock()
-	return conf.EvictedStores
+
+	return append(conf.EvictedStores, conf.EvictedNetworkSlowStores...)
```
bufferflies, line 145: Is it possible that there are some overlapping stores in this array?

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -273,13 +299,13 @@ func (handler *evictSlowStoreHandler) updateConfig(w http.ResponseWriter, r *htt
 	}
 	recoveryDurationGapFloat, inputRecoveryDuration := input["recovery-duration"].(float64)
 	if input["recovery-duration"] != nil && !inputRecoveryDuration {
-		handler.rd.JSON(w, http.StatusInternalServerError, errors.New("invalid argument for 'recovery-duration'").Error())
+		handler.rd.JSON(w, http.StatusInternalServerError, perrors.New("invalid argument for 'recovery-duration'").Error())
```
bufferflies: The http code should be StatusBadRequest not a server problem

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -273,13 +299,13 @@ func (handler *evictSlowStoreHandler) updateConfig(w http.ResponseWriter, r *htt
...
 		return
 	}
 
 	batchFloat, inputBatch := input["batch"].(float64)
 	if input["batch"] != nil && !inputBatch {
-		handler.rd.JSON(w, http.StatusInternalServerError, errors.New("invalid argument for 'batch'").Error())
+		handler.rd.JSON(w, http.StatusInternalServerError, perrors.New("invalid argument for 'batch'").Error())
```
bufferflies: ditto

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -381,22 +408,28 @@ func (s *evictSlowStoreScheduler) ReloadConfig() error {
...
 	evictStore := s.conf.evictStore()
 	if evictStore != 0 {
-		return cluster.SlowStoreEvicted(evictStore)
+		errs = append(errs, cluster.SlowStoreEvicted(evictStore))
 	}
 	for _, storeID := range s.conf.getPausedNetworkSlowStores() {
-		// only support one network slow store now
-		return cluster.PauseLeaderTransfer(storeID, constant.In)
+		s.conf.networkSlowStoreRecoverStartAts[storeID] = nil
+		errs = append(errs, cluster.PauseLeaderTransfer(storeID, constant.In))
```
bufferflies: It will append the nil into the errs array even if the error is nil

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -129,13 +141,11 @@ func (conf *evictSlowStoreSchedulerConfig) persistLocked(updateFn func()) error
 func (conf *evictSlowStoreSchedulerConfig) getStores() []uint64 {
 	conf.RLock()
 	defer conf.RUnlock()
-	return conf.EvictedStores
+
+	return append(conf.EvictedStores, conf.EvictedNetworkSlowStores...)
 }
 
-func (conf *evictSlowStoreSchedulerConfig) getKeyRangesByID(id uint64) []keyutil.KeyRange {
-	if conf.evictStore() != id {
```
bufferflies, line 136: Why delete it?

pkg/schedule/schedulers/evict_slow_store.go
okJiang: No, `SlowStoreEvicted` only allows one marker.

pkg/schedule/schedulers/evict_slow_store.go
okJiang:
```
// Join returns an error that wraps the given errors.
// Any nil error values are discarded.
// Join returns nil if every value in errs is nil.
```

pkg/schedule/schedulers/evict_slow_store.go
okJiang: `storeID` is from `evictLeaderStoresConf.getStores`. Do not need check it again. And `evictStore` do not represents all evicted store, it only represent disk evicted store.

pkg/schedule/schedulers/evict_slow_store.go
bufferflies: Yes, the logic is right but the the errors array length has changed, it waste the memory here

okJiang: Prepare is only called once scheduler creating. So I think it is acceptable

okJiang: But I can modify it to better. That is no problem.

okJiang: done

Issue #9707, opened 2025-08-29
Tracking issue for Next-Gen TiDB


rleungx:
## Development Task
This issue is used to track the development or bug fix progress of PD when adapting to the new architecture of TiDB.

PR #9717, opened 2025-09-03
feat(metering): integrate RU metering and flush to object storage

JmPotato:

Issue Number: ref #9707.

```commit-message
- Introduced a new **metering writer** (`pkg/metering`) that collects RU usage data and periodically flushes it to object storage via `pingcap/metering_sdk`.
- Implemented a **RU collector** (`ruCollector`) in the resource manager to track resource usage (OLTP/OLAP RU, cross-AZ traffic) per keyspace.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -503,7 +511,9 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 			if err != nil {
 				continue
 			}
-			m.metrics.recordConsumption(consumptionInfo, keyspaceName, m.controllerConfig, time.Now())
+			consumptionInfo.keyspaceName = keyspaceName
```
rleungx, line 514: How about adding the keyspaceName before sending to the channel?

JmPotato: `getKeyspaceNameByID` has potential disk read overhead. Since sending to the channel is on the path of the RU request, I prefer to do this asynchronously in a separate goroutine like now. WDYT?

rleungx: ok, make sense

pkg/metering/writer.go
```diff
@@ -0,0 +1,183 @@
...
+		if len(records) == 0 {
+			continue
+		}
+		meteringData := &common.MeteringData{
+			SelfID:    mw.id,
+			Timestamp: ts,
+			Category:  category,
+			Data:      records,
+		}
+		if mw.inner == nil {
```
rleungx: How about return directly?

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -0,0 +1,112 @@
...
+	"sync"
+
+	"github.com/pingcap/metering_sdk/common"
+
+	"github.com/tikv/pd/pkg/metering"
+)
+
+const (
+	resourceManagerCategory = "resource-manager"
+	ruMeteringVersion       = "1"
```
rleungx, line 27: v1 maybe better?

pkg/metering/config.go
```diff
@@ -0,0 +1,44 @@
...
+package metering
+
+import (
+	"github.com/pingcap/errors"
+	"github.com/pingcap/metering_sdk/storage"
+)
+
+// Config represents the configuration for metering.
+type Config struct {
+	Type   storage.ProviderType `toml:"type" json:"type"`
```
djshow832, line 24: I suggest `json:"provider-type"` to be more specific.

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -0,0 +1,130 @@
...
+			meteringDataOltpRUField:              ruMetering.oltpMeteringValue(),
+			meteringDataOlapRUField:              ruMetering.olapMeteringValue(),
+			meteringDataCrossAZTrafficBytesField: ruMetering.crossAZTrafficBytesMeteringValue(),
+		})
+	}
+	c.RUnlock()
+	// Clear the data after reading.
+	c.Lock()
+	c.keyspaceRUMetering = make(map[string]*ruMetering)
+	c.Unlock()
```
djshow832: Why not:
```go
c.Lock()
keyspaceRUMetering := c.keyspaceRUMetering
c.keyspaceRUMetering = make(map[string]*ruMetering)
c.Unlock()

records := make([]map[string]any, 0, len(keyspaceRUMetering))
for keyspaceName, ruMetering := range keyspaceRUMetering {
...
}
```

pkg/metering/writer.go
```diff
@@ -0,0 +1,191 @@
...
+
+	ticker := time.NewTicker(meteringFlushInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-mw.ctx.Done():
+			log.Info("metering writer loop exits")
+			return
+		case <-ticker.C:
```
djshow832: I'm afraid the time is not accurate enough in extreme situations.
E.g. The ticker starts at 00:00:59:999.
In the first round, getting the collector is blocked a little and `time.Now()` gets 00:02:00:000, the written minute is 00:02:00.
In the second round, it's fast enough and `time.Now()` gets 00:02:59:999, the written minute is also 00:02:00.
The second round overwrites the first found and the metering data is missing for 1 minute.

pkg/metering/writer.go
```diff
@@ -0,0 +1,191 @@
...
+	if mw == nil {
+		return
+	}
+	mw.wg.Add(1)
+	go mw.meteringLoop()
+	log.Info("metering writer started")
+}
+
+// Stop stops the metering writer
+func (mw *Writer) Stop() {
```
djshow832, line 118: The metering data after the last ticker is not recorded, is it necessary to flush before stopping?

.github/workflows/check.yaml
```diff
@@ -15,6 +15,6 @@ jobs:
           go-version: '1.23'
       - name: Make Check
         run: |
-          SWAGGER=1 make build
```
JmPotato, line 18: Bypass this error:

```shell
2025/09/08 14:57:27 ParseComment error in file /Users/kaijikikou/go/pkg/mod/github.com/alibabacloud-go/darabonba-openapi/v2@v2.0.11/client/client.go :missing required param comment parameters "config - config contains the necessary information to create a client"
make: *** [swagger-spec] Error 1
```

Will be fixed later after merging this PR.

pkg/metering/writer.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		meteringData := &common.MeteringData{
+			SelfID:    mw.id,
+			Timestamp: ts,
+			Category:  category,
+			Data:      records,
+		}
+		recordCount := len(records)
+		// TODO: write with pagination if needed.
+		if err := mw.inner.Write(mw.ctx, meteringData); err != nil {
```
djshow832: The mw.ctx may be done in the last round, so it's useless to flush it after ctx.Done().

djshow832 (approved):

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -0,0 +1,127 @@
...
+	meteringUnitRU    = "RU"
+	meteringUnitBytes = "Bytes"
+
+	meteringDataVersionField             = "version"
+	meteringDataClusterIDField           = "cluster_id"
+	meteringDataSourceNameField          = "source_name"
+	meteringDataOltpRUField              = "oltp_ru"
+	meteringDataOlapRUField              = "olap_ru"
+	meteringDataCrossAZTrafficBytesField = "cross_az_traffic_bytes"
+)
```
okJiang, line 39: Just a suggestion. Is it better to define them in `sdk` repo?

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -0,0 +1,127 @@
...
+	return &ruCollector{
+		keyspaceRUMetering: make(map[string]*ruMetering),
+	}
+}
+
+// Category returns the category of the collector.
+func (*ruCollector) Category() string { return resourceManagerCategory }
+
+// Collect collects the RU metering data.
+func (c *ruCollector) Collect(data any) {
```
okJiang, line 102: Why not input `consumptionItem` directly?

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -0,0 +1,127 @@
...
+	rm, ok := c.keyspaceRUMetering[consumption.keyspaceName]
+	if !ok {
+		rm = &ruMetering{}
+		c.keyspaceRUMetering[consumption.keyspaceName] = rm
+	}
+	rm.add(consumption)
+}
+
+// Flush flushes the RU metering data.
+func (c *ruCollector) Flush() []map[string]any {
```
okJiang: How about this

```suggestion
func (c *ruCollector) GetAndClear() []map[string]any {
```

pkg/metering/config.go
```diff
@@ -0,0 +1,45 @@
...
+	if len(c.Type) == 0 {
+		c.Type = storage.ProviderTypeS3
+	}
+	if len(c.Region) == 0 {
+		return errors.New("region is required for the metering config")
+	}
+	if len(c.Bucket) == 0 {
+		return errors.New("bucket is required for the metering config")
+	}
+	if c.Type == storage.ProviderTypeS3 && len(c.RoleARN) == 0 {
```
okJiang: Currently, we only need to support role arn, right? Is the long-term plan to support all defined provider types? I think this could be slightly encapsulated, with authentication information in a separate structure, and each provider also having a separate structure.

And you can comment the supported range

pkg/metering/writer.go
```diff
@@ -0,0 +1,216 @@
...
+func (mw *Writer) RegisterCollector(collector Collector) {
+	if mw == nil {
+		return
+	}
+	mw.Lock()
+	defer mw.Unlock()
+	if mw.collectors == nil {
+		mw.collectors = make(map[string]Collector)
+	}
+	mw.collectors[collector.Category()] = collector
```
okJiang, line 142: Is this over-engineering? Will we have more collector types?

pkg/metering/writer.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// meteringLoop runs the background loop to flush metering data periodically.
+func (mw *Writer) meteringLoop() {
+	defer logutil.LogPanic()
+	defer mw.wg.Done()
+
+	now := time.Now()
+	// Truncate to the nearest minute and add the interval to get the next flush time.
+	next := now.Truncate(flushInterval).Add(flushInterval)
```
okJiang, line 154: What's the purpose?

pkg/metering/writer.go
```diff
@@ -0,0 +1,216 @@
...
+		meteringData := &common.MeteringData{
+			SelfID:    mw.id,
+			Timestamp: ts,
+			Category:  category,
+			Data:      records,
+		}
+		recordCount := len(records)
+		start := time.Now()
+		ctx, cancel := context.WithTimeout(ctx, flushTimeout)
+		// TODO: write with pagination if needed.
```
okJiang, line 202: Will it be completed at this milestone? The data may be very large.

pkg/mcs/resourcemanager/server/metering.go
JmPotato: `Collector` is designed as an generic interface, and later we will add another collector, so we need to have the function signature unified.

pkg/mcs/resourcemanager/server/metering.go
JmPotato: In my opinion, the term *flush* in the context of data persistence implies both retrieving and clearing operations.

pkg/metering/config.go
JmPotato: Very fair suggestion, but at the moment we haven’t worked out a fully refined configuration recommendation with the SDK provider, so for now we just wrote it in a rough-and-ready way to simply “get it running.”

pkg/metering/writer.go
JmPotato: Yes, later we will have an another collector to report the store backup related metering data from the store heartbeat.

pkg/metering/writer.go
JmPotato: This is the conclusion after discussion with the metering team: we need to write metering data at the minute level and ensure it carries the correct timestamp, so we need to perform a round-like operation on the time to guarantee accuracy.

pkg/metering/writer.go
JmPotato: This may depend on whether we encounter scenarios requiring pagination in subsequent tests. For now, I think the amount of RU statistics data at the minute level is only related to the number of keyspaces, and in the short term we may not encounter a data volume that requires pagination. However, adding it may be necessary in the future, so I left a TODO as a reminder.

pkg/mcs/resourcemanager/server/metering.go
JmPotato: Nice suggestion, have feedbacked to the SDK team.

pkg/mcs/resourcemanager/server/metering.go
okJiang: But in fact this function does not include persistence, which is implemented in the upper layer

pkg/metering/config.go
okJiang: But this should be a public configuration, should we give it more consideration?

pkg/metering/writer.go
okJiang: Please add more related comments.

pkg/metering/writer.go
```diff
@@ -0,0 +1,231 @@
...
+		err := mw.inner.Write(ctx, meteringData)
```
glorv, line 203: Shall we backoff some time (maybe in another routine) instead of directly dropping the request and returning an error? Or is the write already implemented backoff logic?

JmPotato: After reviewing the SDK code, I can confirm that the current implementation does not include any retry logic, so you are correct. A separate PR will be proposed to address this later. In this case, a producer–consumer queue model might be a good approach to avoid blocking the flush while still ensuring the writes are performed.

PR #9720, opened 2025-09-04
scheduler: graceful shutdown implement

### LINKED ISSUE #9719: Graceful Shutdown of TiKV pods when SIGTERM is sent to the pod
## Development Task
PD should handle the graceful shutdown of TiKV when TiKV receives the SIGTERM signal. ref https://github.com/tikv/tikv/issues/17221

hujiatao0:
add test for scheduler

Issue Number: Close #9719 

```commit-message
Add an is_stopping status to the StoreHeartbeat message. When TiKV receives a SIGTERM, it sets this flag. This change adds a new evict-stopping-store-scheduler to PD, which is analogous to the evict-slow-store-scheduler. It proactively transfers leaders away from nodes by inspecting the is_stopping status from store heartbeats.
```

### Check List

Tests

- Unit test

Code changes

- Added a new evict-stopping-store scheduler, it inspects the is_stopping status from storeheartbeat and evict leaders for the Tikv node if is_stopping is true.

Side effects

- Increased code complexity

Related changes

- PR to update [`pingcap/kvproto`](https://github.com/pingcap/kvproto): https://github.com/pingcap/kvproto/pull/1345
- PR to update [`pingcap/tikv`](https://github.com/pingcap/tikv):https://github.com/tikv/tikv/pull/18930

### Release note

```release-note
None.
```


server/grpc_service.go
```diff
@@ -927,6 +931,57 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+							log.Error("redirect scheduler update failed", zap.Error(err))
+						}
+					}
+				}
+			}
+		} else {
+			// try to remove the dedicated scheduler every time when the state is not stopping.
+			// avoid the case that the last StoreHeartbeat of graceful shutdown process failed and can't remove the dedicated scheduler
+			if exist, err := h.IsSchedulerExisted(gracefulName); err == nil && exist {
+				if ids, err := h.GetEvictLeaderStoreIDs(gracefulName); err == nil {
```
rleungx: Looks strange. It's a gRPC request, but we call the HTTP API internally?

hujiatao0: Yes, it's a bit strange in its implementation. However, I haven't been able to find a more direct interface to change the scheduler configuration from the gRPC side. It seems the current code structure only allows for external changes to the scheduler config via an HTTP API request. Is there a better way to call it?

okJiang: Since we have already saved the `isStopping` information in the store info, I think we only need to add a new simple internal scheduler to judge it.
```
If it is Stopping
    add the store_id to evict_leader scheduler;
otherwise
    remove the store_id from evict_leader scheduler.
```

The current implementation in store heartbeat is too complex

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -52,7 +52,9 @@ type evictLeaderSchedulerConfig struct {
 
 	StoreIDWithRanges map[uint64][]keyutil.KeyRange `json:"store-id-ranges"`
 	// Batch is used to generate multiple operators by one scheduling
-	Batch             int `json:"batch"`
+	Batch int `json:"batch"`
+	// SchedulerName allows a dedicated scheduler instance name while keeping the type as evict-leader-scheduler.
+	SchedulerName     string `json:"scheduler_name"`
```
rleungx: We have a name field inside the base scheduler. Can we reuse it like WithBalanceLeaderName?

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
 		}, nil
 	}
 
+	log.Info("store heartbeat", zap.Bool("is_stopping", request.GetStats().GetIsStopping()))
```
rleungx: How about removing it?

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+	// When the store is stopping, manage a dedicated instance named "graceful-shutdown-scheduler".
+	// When it is not stopping, try to remove the dedicated instance if it exists.
+	{
+		isStopping := request.GetStats().GetIsStopping()
+		h := s.GetHandler()
+		gracefulName := "graceful-shutdown-scheduler"
+
+		rc := s.GetRaftCluster()
+		if rc == nil {
+			log.Error("raft cluster is nil when managing graceful shutdown scheduler")
```
rleungx: Please remove the log, and we do have `notBootstrappedHeader` for this case.

rleungx: BTW, we have already checked the rc at line 915

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+			return &pdpb.StoreHeartbeatResponse{Header: wrapHeader()}, nil
+		}
+		sc := rc.GetCoordinator().GetSchedulersController()
+
+		if isStopping {
```
rleungx:
```suggestion
		if request.GetStats().GetIsStopping() {
```

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
 	}
 
+	log.Info("store heartbeat", zap.Bool("is_stopping", request.GetStats().GetIsStopping()))
+
+	// Dedicate evict-leader scheduler for graceful shutdown.
+	// When the store is stopping, manage a dedicated instance named "graceful-shutdown-scheduler".
+	// When it is not stopping, try to remove the dedicated instance if it exists.
+	{
+		isStopping := request.GetStats().GetIsStopping()
+		h := s.GetHandler()
```
rleungx: move it to the if block

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+			scheduler := sc.GetScheduler(gracefulName)
+			if scheduler == nil {
+				// Create scheduler first
+				payload := map[string]any{
+					"name":           types.EvictLeaderScheduler.String(),
+					"scheduler_name": gracefulName,
+				}
+				if body, err := json.Marshal(payload); err == nil {
+					if err := h.AddSchedulerWithDecoder(types.EvictLeaderScheduler, schedulers.ConfigJSONDecoder(body)); err == nil {
+						scheduler = sc.GetScheduler(gracefulName)
```
rleungx: If the scheduler is nil, seem we can return directly?

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+					} else {
+						log.Error("add scheduler failed", zap.Error(err))
+					}
+				}
+			}
+			if scheduler != nil {
+				if evictScheduler, ok := scheduler.Scheduler.(interface {
```
rleungx: How about abstracting some functions to make the code clear? There are too many if...else...

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+					AddStore(uint64) error
+					RemoveStore(uint64) error
+					EvictStoreIDs() []uint64
+				}); ok {
+					ids := evictScheduler.EvictStoreIDs()
+					log.Info("get evict leader store ids", zap.Any("ids", ids))
+					hasCurrent := slices.Contains(ids, storeID)
+					if hasCurrent {
+						err := evictScheduler.RemoveStore(storeID)
```
rleungx: When to remove the store? What if the leader still exists?

server/grpc_service.go
```diff
@@ -927,6 +931,84 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
...
+			scheduler := sc.GetScheduler(gracefulName)
+			if scheduler != nil {
+				if evictScheduler, ok := scheduler.Scheduler.(interface {
+					AddStore(uint64) error
+					RemoveStore(uint64) error
+					EvictStoreIDs() []uint64
+				}); ok {
+					ids := evictScheduler.EvictStoreIDs()
+					log.Info("get evict leader store ids", zap.Any("ids", ids))
+					hasCurrent := slices.Contains(ids, storeID)
```
rleungx: Do we allow multiple stores to shut down at the same time?

server/grpc_service.go
okJiang: cc @rleungx

server/grpc_service.go
hujiatao0: > If the scheduler is nil, seem we can return directly?

When a TiKV instance is in the is_stopping state, PD needs to check for an evict-leader scheduler. If no such scheduler is found for that node to perform leader transfers, then one needs to be added.

server/grpc_service.go
hujiatao0: > When to remove the store? What if the leader still exists?

The timing for a store removal is controlled by TiKV. When all leaders on a TiKV node have been transferred, or when the internal graceful_shutdown_timeout is triggered, it will immediately set the is_stopping status to false and immediately send a StoreHeartbeat. Upon receiving this heartbeat, PD will check if the store is in a non-stopping state and, if so, will attempt to remove any dedicated graceful shutdown evict-leader scheduler that may exist on the node.

hujiatao0: > When to remove the store? What if the leader still exists?

The current design of graceful shutdown is to transfer all leaders whenever possible. However, there is a timeout mechanism, and if it's triggered, the shutdown process will be initiated directly.

server/grpc_service.go
hujiatao0: > Do we allow multiple stores to shut down at the same time?

Currently, users can ensure that two or more TiKV instances will not be shut down at the same time. This logic is designed to identify the deletion of the corresponding store on the scheduler, ensuring that PD can accurately process the relevant store when exchanging StoreHeartbeat messages with multiple TiKV instances.

server/grpc_service.go
hujiatao0: > Since we have already saved the `isStopping` information in the store info, I think we only need to add a new simple internal scheduler to judge it.
> 
> ```
> If it is Stopping
>     add the store_id to evict_leader scheduler;
> otherwise
>     remove the store_id from evict_leader scheduler.
> ```
> 
> The current implementation in store heartbeat is too complex

The current design requires a dedicated evict-leader scheduler with a specific name. This is to avoid conflicts between evict-leader schedulers added by users and the add/remove logic used during the graceful shutdown operation.

okJiang: You can use `scheduleEvictLeaderBatch` in the new scheduler(just like `evict-slow-store-scheduler`), which does not conflict with evict-leader-scheduler.

hujiatao0: > You can use `scheduleEvictLeaderBatch` in the new scheduler(just like `evict-slow-store-scheduler`), which does not conflict with evict-leader-scheduler.

Ok, I see. This logic of reusing the slow store scheduler seems to be simpler than the previous approach. And please review the new updated commit. I've implemented a similar logic by simplifying the evict-leader scheduler's call, which involves adding or deleting the store ID based on the is_stopping status. It seems the main difference from the comments is that a scheduler like evict-slow-store would always exist in PD, while this current implementation involves creating and deleting the scheduler during the graceful shutdown cycle. The functional verification test is currently no problem, is this implementation also feasible?

okJiang: If you can get the approval of two reviewers, it will be fine. Could you please give your opinion? cc @lhy1024 @bufferflies

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,367 @@
...
+	"github.com/tikv/pd/pkg/core"
+	sche "github.com/tikv/pd/pkg/schedule/core"
+	"github.com/tikv/pd/pkg/schedule/operator"
+	"github.com/tikv/pd/pkg/schedule/plan"
+	"github.com/tikv/pd/pkg/schedule/types"
+	"github.com/tikv/pd/pkg/utils/keyutil"
+)
+
+const (
+	defaultStoppingStoreRecoverySec = 30 // shorter than slow store
```
rleungx: Do we need it?

hujiatao0: we may not need it, i just think 

> Do we need it?

This probably isn't needed. I kept this section for now because it is copied from slow store scheduler and doesn't have any real impact after a direct TiKV shutdown. It can be cleaned up in a future commit when the unit tests are added.

server/grpc_service.go
```diff
@@ -64,6 +64,7 @@ const (
 	defaultGRPCDialTimeout        = 3 * time.Second
 
 	gRPCServiceName = "pdpb.PD"
+
```
hbisheng: the extra empty lines in this file should be removed

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,307 @@
...
+	return nil
+}
+
+// CleanConfig implements the Scheduler interface.
+func (s *evictStoppingStoreScheduler) CleanConfig(cluster sche.SchedulerCluster) {
+	s.cleanupEvictLeader(cluster)
+}
+
+func (s *evictStoppingStoreScheduler) prepareEvictLeader(cluster sche.SchedulerCluster, storeID uint64) error {
+	if err := cluster.SlowStoreEvicted(storeID); err != nil {
```
hbisheng: Here we are reusing the slow store logic. We should probably add a note to the `slowStoreEvicted` field in `StoreInfo`. 
```
pkg/core/store.go
type StoreInfo struct {
	slowStoreEvicted       bool         // this store has been evicted as a slow/stopping store, should not transfer leader to it
```

hujiatao0: > Here we are reusing the slow store logic. We should probably add a note to the `slowStoreEvicted` field in `StoreInfo`.
> 
> ```
> pkg/core/store.go
> type StoreInfo struct {
> 	slowStoreEvicted       bool         // this store has been evicted as a slow/stopping store, should not transfer leader to it
> ```

ok

okJiang:
please fix the conflict @hujiatao0

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,306 @@
...
+			operator.IncOperatorLimitCounter(s.GetType(), operator.OpLeader)
+		}
+		return allowed
+	}
+	return true
+}
+
+// Schedule implements the Scheduler interface.
+func (s *evictStoppingStoreScheduler) Schedule(cluster sche.SchedulerCluster, _ bool) ([]*operator.Operator, []plan.Plan) {
+	evictLeaderCounter.Inc()
```
rleungx: Shall we use a new Counter?

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,306 @@
...
+	s.scheduleStoppingStore(cluster)
+	return s.schedulerEvictLeader(cluster), nil
```
rleungx, line 275: How about merging these two steps?

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,306 @@
...
+	// If there is only one stopping store, evict leaders from that store.
+	log.Info("detected stopping store, start to evict leaders", zap.Uint64("store-id", stoppingStore.GetID()))
+	err := s.prepareEvictLeader(cluster, stoppingStore.GetID())
+	if err != nil {
+		log.Info("prepare for evicting leader failed", zap.Error(err), zap.Uint64("store-id", stoppingStore.GetID()))
+		return
+	}
+	// Record the stopping store evicted status.
+	storeIDStr := strconv.FormatUint(stoppingStore.GetID(), 10)
+	evictedSlowStoreStatusGauge.WithLabelValues(storeIDStr, string(gracefulShutdownStore)).Set(1)
```
rleungx: ditto

pkg/schedule/schedulers/evict_stopping_store.go
rleungx: If we are going to reuse the logic, it's better to change the name to a general one.

pkg/schedule/schedulers/evict_stopping_store.go
hujiatao0: > How about merging these two steps?

The logic for separating these two steps references the implementation in the slow store scheduler. Their processing logic seems quite different, so would it be better to keep them as two distinct steps?

pkg/core/store_option.go
```diff
@@ -134,9 +134,9 @@ func ResumeLeaderTransfer(d constant.Direction) StoreCreateOption {
 	}
 }
 
-// SlowStoreEvicted marks a store as a slow store and prevents transferring
+// SlowStoppingStoreEvicted marks a store as a slow store or stopping store and prevents transferring
 // leader to the store
-func SlowStoreEvicted() StoreCreateOption {
+func SlowStoppingStoreEvicted() StoreCreateOption {
 	return func(store *StoreInfo) {
 		store.slowStoreEvicted.Add(1)
```
rleungx, line 141: Is it possible to use a new field?

pkg/schedule/schedulers/evict_stopping_store.go
rleungx: Still slow?

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+
+	if stoppingStore == nil {
+		return
+	}
+
+	// If there is only one stopping store, evict leaders from that store.
+	log.Info("detected stopping store, start to evict leaders", zap.Uint64("store-id", stoppingStore.GetID()))
+	err := s.prepareEvictLeader(cluster, stoppingStore.GetID())
```
rleungx: How about using a variable for store id? You don't need to call it every time.

rleungx:
test failed

hujiatao0:
> test failed

all tests passed

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,307 @@
...
+		rd:     render.New(render.Options{IndentJSON: true}),
+	}
+	router := mux.NewRouter()
+	router.HandleFunc("/config", h.updateConfig).Methods(http.MethodPost)
+	router.HandleFunc("/list", h.listConfig).Methods(http.MethodGet)
+	return router
+}
+
+func (handler *evictStoppingStoreHandler) updateConfig(w http.ResponseWriter, _ *http.Request) {
+	handler.rd.JSON(w, http.StatusOK, "Config updated.")
```
rleungx, line 163: Do we support updating the batch?

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,307 @@
...
+	stoppingStoreID := stoppingStore.GetID()
+	log.Info("detected stopping store, start to evict leaders", zap.Uint64("store-id", stoppingStoreID))
+	err := s.prepareEvictLeader(cluster, stoppingStoreID)
+	if err != nil {
+		log.Info("prepare for evicting leader failed", zap.Error(err), zap.Uint64("store-id", stoppingStoreID))
+		return
+	}
+	// Record the stopping store evicted status.
+	storeIDStr := strconv.FormatUint(stoppingStoreID, 10)
+	evictedStoppingStoreStatusGauge.WithLabelValues(storeIDStr, string(gracefulShutdownStore)).Set(1)
```
rleungx: When do we reset it?

hujiatao0: it will be set 0 when cleanup

pkg/schedule/schedulers/evict_stopping_store.go
hujiatao0: I think it's ok to support it and I added the logic and related tests

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,344 @@
...
+	return conf.persistLocked(func() {
+		conf.EvictedStores = []uint64{id}
+	})
+}
+
+func (conf *evictStoppingStoreSchedulerConfig) clearEvictedAndPersist() (oldID uint64, err error) {
+	oldID = conf.evictStore()
+	conf.Lock()
+	defer conf.Unlock()
+	if oldID > 0 {
```
okJiang, line 106: Seems do not need check this

pkg/schedule/schedulers/evict_stopping_store.go
```diff
@@ -0,0 +1,344 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (s *evictStoppingStoreScheduler) cleanupEvictLeader(cluster sche.SchedulerCluster) {
+	evictStoppingStore, err := s.conf.clearEvictedAndPersist()
+	if err != nil {
+		log.Info("evict-stopping-store-scheduler persist config failed", zap.Uint64("store-id", evictStoppingStore))
```
okJiang: It is better to keep the consistency between storage and memory.

```suggestion
		log.Warn("evict-stopping-store-scheduler persist config failed", zap.Uint64("store-id", evictStoppingStore))
        return
```

pkg/schedule/schedulers/metrics.go
```diff
@@ -106,6 +106,14 @@ var (
 			Help:      "Store evicted status due to slow",
 		}, []string{"store", "slow_type"})
 
+	evictedStoppingStoreStatusGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "pd",
+			Subsystem: "scheduler",
+			Name:      "evicted_stopping_store_status",
+			Help:      "Store evicted status due to stopping",
+		}, []string{"store", "stopping_type"})
```
okJiang: No need `stopping_type` now.

```suggestion
		}, []string{"store"})
```

pkg/schedule/config/config.go
```diff
@@ -590,6 +590,7 @@ var defaultSchedulersInit = func() SchedulerConfigs {
 	}
 	if !kerneltype.IsNextGen() {
 		defaultSchedulers = append(defaultSchedulers, SchedulerConfig{Type: types.SchedulerTypeCompatibleMap[types.EvictSlowStoreScheduler]})
+		defaultSchedulers = append(defaultSchedulers, SchedulerConfig{Type: types.SchedulerTypeCompatibleMap[types.EvictStoppingStoreScheduler]})
```
okJiang, line 593: Do not enable it in next-gen?

okJiang (commented):
rest lgtm

pkg/schedule/config/config.go
hujiatao0: Yes, we are not planning to support graceful shutdown in next-gen for now.

hujiatao0: > Do not enable it in next-gen?

Yes, we are not planning to support graceful shutdown in next-gen for now.

pkg/schedule/schedulers/evict_stopping_store.go
hujiatao0: > Seems do not need check this

Removing this check will cause some tests to fail.

niubell (approved):

PR #9721, opened 2025-09-04
config: disable scheduling fallback

rleungx:

Issue Number: ref #5839.

When we deploy the microservice, PD might use fewer resources than before. We should not allow scheduling fallback to PD automatically to avoid PD overload.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -249,7 +249,7 @@ const (
 	minCheckRegionSplitInterval     = 1 * time.Millisecond
 	maxCheckRegionSplitInterval     = 100 * time.Millisecond
 
-	defaultEnableSchedulingFallback  = true
+	defaultEnableSchedulingFallback  = false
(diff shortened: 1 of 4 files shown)


JmPotato (approved):

server/config/config.go
```diff
@@ -249,7 +249,7 @@ const (
 	minCheckRegionSplitInterval     = 1 * time.Millisecond
 	maxCheckRegionSplitInterval     = 100 * time.Millisecond
 
-	defaultEnableSchedulingFallback  = true
+	defaultEnableSchedulingFallback  = false
```
lhy1024, line 252: If we upgrade from old version, `enable-scheduling-fallback` will still be true. Is this expected?

lhy1024 (commented):
the rest LGTM

PR #9739, opened 2025-09-15
store: replace slow store eviction flags with atomic counters

okJiang:

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/store.go
+++ b/pkg/core/store.go
@@ -24,7 +24,6 @@ import (
 	"github.com/docker/go-units"
 	"go.uber.org/zap"
 
-	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/log"
@@ -62,8 +61,8 @@ type StoreInfo struct {
 	*storeStats
 	pauseLeaderTransferIn  atomic.Int64 // not allow to be used as target of transfer leader
 	pauseLeaderTransferOut atomic.Int64 // not allow to be used as source of transfer leader
-	slowStoreEvicted       bool         // this store has been evicted as a slow store, should not transfer leader to it
-	slowTrendEvicted       bool         // this store has been evicted as a slow store by trend, should not transfer leader to it
+	slowStoreEvicted       atomic.Int64 // this store has been evicted as a slow store, should not transfer leader to it
+	slowTrendEvicted       atomic.Int64 // this store has been evicted as a slow store by trend, should not transfer leader to it
 	leaderCount            int
 	regionCount            int
 	learnerCount           int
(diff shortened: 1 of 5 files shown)


pkg/core/store.go
```diff
@@ -119,8 +118,6 @@ func (s *StoreInfo) Clone(opts ...StoreCreateOption) *StoreInfo {
 	store := &StoreInfo{
 		meta:                typeutil.DeepClone(s.meta, StoreFactory),
 		storeStats:          s.storeStats,
-		slowStoreEvicted:    s.slowStoreEvicted,
```
bufferflies, line 122: why not need to copy the slowStoreEvicted?

pkg/core/store.go
```diff
@@ -204,12 +203,12 @@ func (s *StoreInfo) AllowLeaderTransferOut() bool {
 
 // EvictedAsSlowStore returns if the store should be evicted as a slow store.
 func (s *StoreInfo) EvictedAsSlowStore() bool {
-	return s.slowStoreEvicted
+	return s.slowStoreEvicted.Load() > 0
```
rleungx, line 206: Do we need to revert the order changes from the last PR?

okJiang: No need

pkg/errs/errno.go
```diff
@@ -131,7 +131,6 @@ var (
 	ErrStoreDestroyed         = errors.Normalize("store %v has been physically destroyed", errors.RFCCodeText("PD:core:ErrStoreDestroyed"))
 	ErrStoreUnhealthy         = errors.Normalize("store %v is unhealthy", errors.RFCCodeText("PD:core:ErrStoreUnhealthy"))
 	ErrStoreServing           = errors.Normalize("store %v has been serving", errors.RFCCodeText("PD:core:ErrStoreServing"))
-	ErrSlowStoreEvicted       = errors.Normalize("store %v is evicted as a slow store", errors.RFCCodeText("PD:core:ErrSlowStoreEvicted"))
```
lhy1024, line 134: Is it a breaking change?

pkg/core/store.go
```diff
@@ -62,8 +61,8 @@ type StoreInfo struct {
 	*storeStats
 	pauseLeaderTransferIn  atomic.Int64 // not allow to be used as target of transfer leader
 	pauseLeaderTransferOut atomic.Int64 // not allow to be used as source of transfer leader
-	slowStoreEvicted       bool         // this store has been evicted as a slow store, should not transfer leader to it
-	slowTrendEvicted       bool         // this store has been evicted as a slow store by trend, should not transfer leader to it
+	slowStoreEvicted       atomic.Int64 // this store has been evicted as a slow store, should not transfer leader to it
+	slowTrendEvicted       atomic.Int64 // this store has been evicted as a slow store by trend, should not transfer leader to it
```
lhy1024, line 65: Do we need to rename them?

lhy1024 (commented):
the rest LGTM

pkg/core/store.go
okJiang: copy them in L138

pkg/errs/errno.go
okJiang: Are there any external components that depend on this error?

pkg/core/store.go
okJiang: I can't find a better name

PR #9749, opened 2025-09-17
client: signal RG GC goroutines to exit properly

### LINKED ISSUE #9745: Goroutine leak from controller.(*groupCostController).handleTokenBucketUpdateEvent
## Bug Report

We noticed an issue where RG goroutines don't get cleaned up properly when the traffic ramp down:

```
goroutine 5042535512 [select, 1540 minutes]:

github.com/tikv/pd/client/resource_group/controller.(*groupCostController).handleTokenBucketUpdateEvent(0xc023ad4000?, {0x679abb0, 0xc002803040}, {0xc1d66f62a0?, 0xc0e3235560?, 0x9662f80?})
      .../resource_group/controller/controller.go:720 +0x1ea
created by github.com/tikv/pd/client/resource_group/controller.(*ResourceGroupsController).Start.func1 in goroutine 1064
      .../resource_group/controller/controller.go:720 +0x1ea
created by github.com/tikv/pd/client/resource_group/controller.(*ResourceGroupsController).Start.func1 in goroutine 1064
      .../resource_group/controller/controller.go:251 +0x906
```

We saw thousands of `tikv:8252` errors during that window and every time this happens, the TiDB server's goroutine goes up by 100K, and all goroutines are blocked on the above code. This leads to high CPU usage on the TiDB node.

### What did you do?

1. Create a RG group with minimal quota `RU_PER_SEC=1`
2. Send 10K QPS using this RG group

TiDB server's goroutine should be stable. When the traffic ramp down it should recover to lower level.

### What did you see instead?
A lot of goroutines from RG package sticked around and didn't get cleaned up.

trace is provided for v7.1.0 but the same can be reproduced on our v8.5.2 cluster.

yzhan1:

Issue Number: Close #9745

Adds a new `cancel` channel to signal GC goroutines when they can be stopped, to prevent goroutine leaks on client side.

```commit-message
client: signal RG GC goroutines to exit properly
```

### Check List

Tests

- Unit test
- Integration test
Manually tested using TiDB build with the same traffic which generates 600 extra goroutines stuck in `handleTokenBucketUpdateEvent`. After the fix only 1 goroutine in `handleTokenBucketUpdateEvent`.
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/client/resource_group/controller/controller.go
+++ b/client/resource_group/controller/controller.go
@@ -874,7 +874,9 @@ type tokenCounter struct {
 		setupNotificationCh        <-chan time.Time
 		setupNotificationThreshold float64
 		setupNotificationTimer     *time.Timer
-		cancelCh                   chan struct{}
+		// cancelCh is used to cancel the notification to unblock gc goroutines.
+		// It's needed since setupNotificationTimer.Stop() does not close the setupNotificationCh channel.
+		cancelCh chan struct{}
 	}
 
 	lastDeadline time.Time
@@ -1044,7 +1046,7 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 			ch := counter.notify.setupNotificationCh
 			cancel := counter.notify.cancelCh
 			counter.notify.mu.Unlock()
-			if ch == nil {
+			if ch == nil || cancel == nil {
 				continue
 			}
 			select {
@@ -1056,7 +1058,8 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 				counter.notify.mu.Unlock()
 				counter.limiter.SetupNotificationThreshold(threshold)
 			case <-cancel:
-				return
+				// move on to the next resource token counter since this one is cancelled.
+				continue
 			case <-ctx.Done():
 				return
 			}
@@ -1068,7 +1071,7 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 			ch := counter.notify.setupNotificationCh
 			cancel := counter.notify.cancelCh
 			counter.notify.mu.Unlock()
-			if ch == nil {
+			if ch == nil || cancel == nil {
 				continue
 			}
 			select {
@@ -1080,7 +1083,8 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 				counter.notify.mu.Unlock()
 				counter.limiter.SetupNotificationThreshold(threshold)
 			case <-cancel:
-				return
+				// move on to the next request unit token counter since this one is cancelled.
+				continue
 			case <-ctx.Done():
 				return
 			}
--- a/client/resource_group/controller/controller.go
+++ b/client/resource_group/controller/controller.go
@@ -876,7 +876,8 @@ type tokenCounter struct {
 		setupNotificationTimer     *time.Timer
 		// cancelCh is used to cancel the notification to unblock gc goroutines.
 		// It's needed since setupNotificationTimer.Stop() does not close the setupNotificationCh channel.
-		cancelCh chan struct{}
+		cancelCh <-chan struct{}
+		cancelFn context.CancelFunc
 	}
 
 	lastDeadline time.Time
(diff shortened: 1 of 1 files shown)


client/resource_group/controller/controller.go
```diff
@@ -1289,6 +1297,10 @@ func initCounterNotify(counter *tokenCounter) {
 		counter.notify.setupNotificationTimer.Stop()
 		counter.notify.setupNotificationTimer = nil
 		counter.notify.setupNotificationCh = nil
+		if counter.notify.cancelCh != nil {
+			close(counter.notify.cancelCh)
+			counter.notify.cancelCh = nil
```
okJiang, line 1310: Seems do not need set `nil`. If a channel is `nil`, it will be blocked.

client/resource_group/controller/controller.go
```diff
@@ -874,6 +874,7 @@ type tokenCounter struct {
 		setupNotificationCh        <-chan time.Time
 		setupNotificationThreshold float64
 		setupNotificationTimer     *time.Timer
+		cancelCh                   chan struct{}
```
okJiang: I think it is better to add detailed comments to this. The logic here is too complicated.

client/resource_group/controller/controller.go
```diff
@@ -1074,6 +1079,8 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 				threshold := counter.notify.setupNotificationThreshold
 				counter.notify.mu.Unlock()
 				counter.limiter.SetupNotificationThreshold(threshold)
+			case <-cancel:
```
okJiang, line 1086: We should do the same behavior as `<-ch`?

yzhan1: I don't think so. When the goroutine is cancelled it means we already override everything (timer/channels/notification threshold) inside `modifyTokenCounter`, and `modifyTokenCounter` will add a new goroutine to take care of the updates, and the previous goroutine can just exit without doing anything

client/resource_group/controller/controller.go
yzhan1: Updated this so that if it's `nil` we skip the iteration, as we do for `setupNotificationCh`

yzhan1:
@okJiang @JmPotato updated based on comments. PTAL, thanks!

client/resource_group/controller/controller.go
```diff
@@ -874,6 +874,9 @@ type tokenCounter struct {
 		setupNotificationCh        <-chan time.Time
 		setupNotificationThreshold float64
 		setupNotificationTimer     *time.Timer
+		// cancelCh is used to cancel the notification to unblock gc goroutines.
+		// It's needed since setupNotificationTimer.Stop() does not close the setupNotificationCh channel.
+		cancelCh chan struct{}
```
bufferflies: How about using the cancel context ？

yzhan1: @bufferflies PTAL. LMK if that's what you meant by using cancel context. The main context received in the function is for shutting down the whole module, so probably not the best to use

okJiang (approved):

bufferflies (commented):
how about add one test for this case

lhy1024:
Thanks for the fix! It looks like this PR has become stale: the controller code on master has since been split into `global_controller.go` and `group_controller.go`, so this change now needs a rebase and to be ported to the current `group_controller.go` path.

Could you also add a small regression test for this leak case? For example: start `handleTokenBucketUpdateEvent` with a long notification timer, reset/stop  the notification via `initCounterNotify`, and verify the goroutine exits without relying on parent context cancellation.

One small cleanup suggestion: when the timer fires normally, `cancelFn/cancelCh` seem to remain set. It may be better to clear them in the `<-ch` path as well, ideally via a shared cleanup helper.

lhy1024:
Thanks @yzhan1 for the original fix.

Since this PR has been inactive for a while and now conflicts with the current controller split, I opened #10724 as a replacement PR. It ports this fix to the current `group_controller.go` path and adds regression tests for the notification timer cancellation case.

I kept the commit author as @yzhan1 and preserved the original Signed-off-by. The replacement PR also clears the notification state when the timer fires normally, following the earlier review discussion.

Could reviewers continue the review on #10724?
/cc @JmPotato @bufferflies @okJiang

lhy1024:
It is replace with https://github.com/tikv/pd/pull/10724 Thanks for the fix again!

PR #9774, opened 2025-09-23
feat(metering): add storage size metering and keyspace name caching

JmPotato:

Issue Number: ref #9707.

```commit-message
- Introduced storage size metering:  
  - Added `storageSizeCollector` to periodically record row-based and column-based storage usage per keyspace.  
  - Integrated with cluster metering writer for reporting.  
- Extended `RegionInfo` and statistics with `ApproximateColumnarKvSize` tracking.  
- Added keyspace name caching and lookup in `KeyspaceManager` for efficient resolution.  
- Extracted common metering utilities (`NewRUValue`, `NewBytesValue`, constants for source/fields) into `pkg/metering/utils.go`.  
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/metering.go
+++ b/server/cluster/metering.go
@@ -68,8 +68,8 @@ func (*storageSizeCollector) Category() string { return storageSizeCollectorCate
 func (c *storageSizeCollector) Collect(data any) {
 	c.Lock()
 	defer c.Unlock()
-	info := data.([]*storageSizeInfo)
-	for _, info := range info {
+	infoList := data.([]*storageSizeInfo)
+	for _, info := range infoList {
 		c.keyspaceStorageSize[info.keyspaceName] = info
 	}
 }


pkg/core/region.go
```diff
@@ -63,26 +63,27 @@ func errRegionIsStale(region *metapb.Region, origin *metapb.Region) error {
...
+	pendingPeers              []*metapb.Peer
+	term                      uint64
+	cpuUsage                  uint64
+	writtenBytes              uint64
+	writtenKeys               uint64
+	readBytes                 uint64
+	readKeys                  uint64
+	approximateSize           int64
+	approximateKvSize         int64
+	approximateColumnarKvSize int64
```
okJiang: Add comment to describe the unit?

pkg/keyspace/util.go
```diff
@@ -161,6 +161,31 @@ func MakeLabelRule(id uint32) *labeler.LabelRule {
 	}
 }
 
+// ParseKeyspaceIDFromLabelRule parses the keyspace ID from the label rule.
+// It will return the keyspace ID and a boolean indicating whether the label
+// rule is a keyspace label rule.
+func ParseKeyspaceIDFromLabelRule(rule *labeler.LabelRule) (uint32, bool) {
```
okJiang, line 168: Can add an unit test for this.

pkg/keyspace/keyspace.go
```diff
@@ -96,6 +97,8 @@ type Manager struct {
 	kgm *GroupManager
 	// nextPatrolStartID is the next start id of keyspace assignment patrol.
 	nextPatrolStartID uint32
+	// cached keyspace name for each keyspace ID.
+	keyspaceNameLookup sync.Map // store as ID(uint32) -> name(string)
```
okJiang, line 101: I don't see keyspace name deleted or updated. Is that expected?

JmPotato: Currently, we do not have logic for modifying ID and name or deleting keyspace.

server/cluster/cluster.go
```diff
@@ -410,6 +414,7 @@ func (c *RaftCluster) Start(s Server, bootstrap bool) (err error) {
 	go c.runUpdateStoreStats()
 	go c.startGCTuner()
 	go c.startProgressGC()
+	go c.runStorageSizeCollector(s.GetMeteringWriter(), c.regionLabeler, s.GetKeyspaceManager())
```
rleungx, line 417: Do we need to disable it in op environment?

pkg/keyspace/util.go
```diff
@@ -161,6 +161,31 @@ func MakeLabelRule(id uint32) *labeler.LabelRule {
...
+	if rule.RuleType != labeler.KeyRange {
```
rleungx: There could be other features using label rules.

server/cluster/metering.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Category returns the category of the collector.
+func (*storageSizeCollector) Category() string { return storageSizeCollectorCategory }
+
+// Collect collects the row-based and column-based storage size data.
+func (c *storageSizeCollector) Collect(data any) {
+	c.Lock()
+	defer c.Unlock()
+	info := data.([]*storageSizeInfo)
```
okJiang:
```suggestion
	infos := data.([]*storageSizeInfo)
```

server/cluster/cluster.go
```diff
@@ -2545,3 +2550,79 @@ func (c *RaftCluster) adjustNetworkSlowStore(storeID uint64) {
...
+		keyspaceName, err := keyspaceManager.GetKeyspaceNameByID(keyspaceID)
+		if err != nil {
+			// TODO: improve the observability of this error.
+			return true
+		}
+		// Make the region bounds.
+		regionBoundsMap[keyspaceName] = keyspace.MakeRegionBound(keyspaceID)
+		return true
+	})
+	log.Info("iterated the region bounds of all keyspaces",
```
okJiang, line 2609: Is this log necessary

JmPotato: It's mainly used to log the duration cost of previous iteration.

server/cluster/cluster.go
```diff
@@ -2545,3 +2550,79 @@ func (c *RaftCluster) adjustNetworkSlowStore(storeID uint64) {
...
+		zap.Duration("cost", time.Since(start)),
+		zap.Int("count", len(regionBoundsMap)))
+
+	start = time.Now()
+	storageSizeInfoList := make([]*storageSizeInfo, 0, len(regionBoundsMap))
+	// Observe the region stats of each keyspace.
+	for keyspaceName, regionBounds := range regionBoundsMap {
+		regionStats := c.GetRegionStatsByRange(regionBounds.TxnLeftBound, regionBounds.TxnRightBound)
```
rleungx, line 2617: Will it affect performance?

JmPotato: Scanning by keyspace is essentially the first level of batching, and then during the internal scan in `GetRegionStatsByRange` it will further batch in groups of 1000. So my understanding is that as long as the lock on the region meta tree is not held for a long time, the performance impact won’t be significant. Moreover, since this scan runs relatively infrequently, if later testing shows it has a higher impact on performance, we can reduce its frequency, since this metering does not require highly real-time accuracy.

PR #9777, opened 2025-09-24
gc: Optimize GetAllKeyspacesGCStates for scenarios where there are too many keyspaces

MyonKeminta:

Issue Number: Ref #8978

```commit-message
This PR supports loading keyspaces by smaller batches in `GetAllKeyspacesGCStates`. This avoids it througing errors due to etcd txn size limit when there are too many keyspaces.
Also makes `GetAllKeyspacesGCStates` support cancelling.
This is not the ultimate solution for tracking GC states changes. The best way is to implement a streaming API that watches the changes of the GC states. However, this PR provides a simpler fix to make it able to run.
```

### Check List

Tests

- Unit test

Code changes

\-

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -664,27 +664,34 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32) (GCState, error) {
 // must be fetched AFTER the beginning of the current invocation, and it never reuses the result of invocations that
 // started earlier than the current one.
 func (m *GCStateManager) GetAllKeyspacesGCStates(ctx context.Context) (map[uint32]GCState, error) {
-	return m.allKeyspacesGCStatesSingleFlight.Do(ctx, func() (map[uint32]GCState, error) {
-		result, err := m.getAllKeyspacesGCStatesImpl()
+	return m.allKeyspacesGCStatesSingleFlight.Do(ctx, func(execCtx context.Context) (map[uint32]GCState, error) {
+		result, err := m.getAllKeyspacesGCStatesImpl(execCtx)
 		failpoint.Inject("onGetAllKeyspacesGCStatesFinish", func() {})
 		return result, err
 	})
 }
 
-func (m *GCStateManager) getAllKeyspacesGCStatesImpl() (map[uint32]GCState, error) {
+func (m *GCStateManager) getAllKeyspacesGCStatesImpl(ctx context.Context) (map[uint32]GCState, error) {
 	failpoint.InjectCall("onGetAllKeyspacesGCStatesStart")
 
-	// TODO: Handle the case that there are too many keyspaces and loading them at once is not suitable.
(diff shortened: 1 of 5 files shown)


pkg/gc/gc_state_manager.go
```diff
@@ -664,39 +664,66 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32) (GCState, error) {
...
+		mutexLocked = true
+	}
+	unlock := func() {
+		m.mu.Unlock()
+		mutexLocked = false
 	}
-	m.mu.Lock()
-	defer m.mu.Unlock()
+
+	ensureUnlocked := func() {
```
rleungx, line 687: Why do we need it?

MyonKeminta: To ensure the mutex is unlocked if the function is aborted abnormally (e.g., panicking), or more return path is introduced in the future.

pkg/utils/syncutil/ordered_single_flight.go
```diff
@@ -23,6 +23,60 @@ type task[TResult any] struct {
...
+	// the execution will be canceled by calling `execCancel`.
+	refState   atomic.Int64
+	execCtx    context.Context
+	execCancel func()
+}
+
+const sealedFlag int64 = 1 << 62
+
+func newTask[TResult any]() *task[TResult] {
+	ctx, cancel := context.WithCancel(context.Background())
```
rleungx, line 39: Do we need to pass a context instead of using a background context?

MyonKeminta: This is to make it able to be canceled if all calls waiting for the result is canceled, but it won't be canceled immediately if the call that triggers the execution of the inner function is canceled while there are still other calls waiting for the result. If it should be an externally provided context, which should it be? Sounds like it should be a context that is canceled when the PD server process exits, if there is one.

pkg/keyspace/keyspace.go
```diff
@@ -950,28 +983,73 @@ func (manager *Manager) PatrolKeyspaceAssignment(startKeyspaceID, endKeyspaceID
...
+	currentBatch []*keyspacepb.KeyspaceMeta
+	currentIndex int
+	isDrained    bool
+	err          error
+}
+
+func newKeyspaceIterator(manager *Manager, startID uint32) *Iterator {
+	return &Iterator{
+		manager: manager,
+		startID: startID,
```
rleungx: Where is it used?

MyonKeminta: It's mistakenly left unused. As we don't need it for now, I removed it.
But I just noticed that there's no test that directly covers the `IterateKeyspaces` itself.. it's indirectly covered by the tests of GC states though.
Maybe I'd better add some?

rleungx (approved):

rleungx:
BTW, I suggest using a separate goroutine to get GC states periodically in the future.

PR #9781, opened 2025-09-26
mcs: fix panic of store not found

rleungx:

Issue Number: ref #5839.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```


pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -207,6 +207,7 @@ func (s *Service) RegionBuckets(stream schedulingpb.Scheduling_RegionBucketsServ
 			// As TiKV report buckets just after the region heartbeat, for new created region, PD may receive buckets report before the first region heartbeat is handled.
 			// So we should not return error here.
 			log.Warn("the store of the bucket in region is not found ", zap.Uint64("region-id", buckets.GetRegionId()))
+			continue
```
bufferflies: The address can be nil if the store is nil

rleungx: Yes, it won't get the address if it is nil.

lhy1024 (approved):

bufferflies (approved):

PR #9783, opened 2025-09-26
feat(metering): add DFS stats collector for write requests and bytes

JmPotato:

Issue Number: ref #9707.

```commit-message
Add `dfs-stats` collector to track DFS write requests and written bytes per keyspace/component.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -417,7 +417,7 @@ func (c *RaftCluster) Start(s Server, bootstrap bool) (err error) {
 	go c.startGCTuner()
 	go c.startProgressGC()
 	go c.runStorageSizeCollector(s.GetMeteringWriter(), c.regionLabeler, s.GetKeyspaceManager())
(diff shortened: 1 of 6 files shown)


server/cluster/cluster.go
```diff
@@ -2626,3 +2629,68 @@ func (c *RaftCluster) collectStorageSize(
...
+			} else {
+				keyspaceName, err = keyspaceManager.GetKeyspaceNameByID(scope.GetKeyspaceId())
+				if err != nil {
+					continue
+				}
+			}
+			dfsStatsInfoList = append(dfsStatsInfoList, &dfsStatsInfo{
+				keyspaceName: keyspaceName,
+				dfsStats:     item,
+			})
```
JmPotato: The implementation here has an issue; incremental logic needs to be handled.

rleungx (approved):

PR #9801, opened 2025-10-09
controller: Split controller.go to global_controller.go and group_controller.go

okJiang:

Issue Number: Ref #4399

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ dev-basic: build check basic-test
 
 BUILD_FLAGS ?=
 BUILD_TAGS ?=
+TEST_TAGS ?=
 BUILD_CGO_ENABLED := 0
 BUILD_TOOL_CGO_ENABLED := 0
 BUILD_GOEXPERIMENT ?=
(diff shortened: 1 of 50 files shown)


client/resource_group/controller/group_controller.go
```diff
@@ -0,0 +1,672 @@
+// Copyright 2023 TiKV Project Authors.
```
JmPotato: Should be upated?

```suggestion
// Copyright 2025 TiKV Project Authors.
```

client/resource_group/controller/group_controller_test.go
```diff
@@ -0,0 +1,222 @@
+// Copyright 2023 TiKV Project Authors.
```
JmPotato: Ditto.

client/resource_group/controller/group_controller.go
okJiang: done

Issue #9806, opened 2025-10-10
Stabilize nextgen testing


rleungx:
### Which jobs are failing

ref #9707.

- [x] TestAdvanceGCSafePointBasic
- [x] TestAdvanceTxnSafePointBasic
- [x] TestGCBarriers
- [x] TestGetGCState
- [x] TestRedirectKeyspace
- [x] TestGetAllKeyspacesMaxTxnSafePoint
- [x] TestKeyspaceGroupTestSuite/TestKeyspaceGroupSplit
- [x] TestKeyspaceGroupManagerTestSuite/TestDefaultMembershipRestriction
- [x] TestBuildSplitKeyspaces
- [x] TestIDAllocationEndValue
- [x] TestKeyspaceTestSuite/TestLoadRangeKeyspace
- [x] TestKeyspaceTestSuite/TestUpdateKeyspaceConfig
- [x] TestValidateID
- [ ] TestIsDefault
- [ ] TestAddScheduler
- [ ] TestPersistScheduler
- [ ] TestRemoveScheduler
- [ ] TestRegionLabel
- [ ] TestUpdateKeyspaceGCManagementType
- [ ] TestGetKeyspaceMetaByID
- [x] clientStatelessTestSuite/TestLoadKeyspace
- [x] clientStatelessTestSuite/TestGetAllKeyspaces
- [x] clientStatefulTestSuite/TestIsKeyspaceUsingKeyspaceLevelGC
- [x] keyspaceTestSuite/TestCreateLoadKeyspace
- [x] keyspaceTestSuite/TestCreateLoadKeyspaceByID
- [x] keyspaceTestSuite/TestUpdateKeyspaceConfig
- [x] TestGCOperations
- [x] TestRegionPlacementRule
- [x] TestServerTestSuite/TestPrimaryChange
- [x] TestGetTSOImmediately
- [x] TestTwiceSplitKeyspaceGroup
- [x] TestTSOPath

### CI link
https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/9802/pull-unit-test-next-gen/1976466510004097024

okJiang:
## Flaky Test
### Which jobs are failing
```text
panic: test timed out after 5m0s
running tests:
[FAIL]  pkg/statistics/utils TestRemove
github.com/tikv/pd/pkg/statistics/utils.TestMain(...)
/home/runner/work/pd/pd/pkg/statistics/utils/kind_test.go:31
```
[...]

PR #9807, opened 2025-10-10
tests: fix `TestDefaultMembershipRestriction`

rleungx:

Issue Number: ref #9806.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/tso/keyspace_group_manager_test.go
+++ b/pkg/tso/keyspace_group_manager_test.go
@@ -45,6 +45,7 @@ import (
 	"github.com/tikv/pd/pkg/utils/testutil"
 	"github.com/tikv/pd/pkg/utils/tsoutil"
 	"github.com/tikv/pd/pkg/utils/typeutil"
+	"github.com/tikv/pd/pkg/versioninfo/kerneltype"
 )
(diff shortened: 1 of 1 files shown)


pkg/tso/keyspace_group_manager_test.go
```diff
@@ -468,8 +469,10 @@ func (suite *keyspaceGroupManagerTestSuite) TestGetKeyspaceGroupMetaWithCheck()
 	re.Nil(kg)
 }
 
-// TestDefaultMembershipRestriction tests the restriction of default keyspace always
+// TestDefaultMembershipRestriction tests the restriction of bootstrap keyspace always
 // belongs to default keyspace group.
+// In Classic mode: protects keyspace 0 (default keyspace)
+// In NextGen mode: protects keyspace 16777214 (system keyspace), allows keyspace 0 to move freely
```
lhy1024, line 475: Do we need to test system keyspace for next-gen?

rleungx: Yes, we can.

Issue #8997, opened 2025-01-14
it maybe failed to offline store when it is preparing


lhy1024:
### What did you do?

offline store when it is preparing in test

### What did you expect to see?

store is offline

### What did you see instead?

```
2025-01-14T08:43:50.8117961Z [2025/01/14 08:43:27.876 +00:00] [WARN] [cluster.go:1474] ["store has been offline"] [store-id=1] [store-address=tikv1] [physically-destroyed=false]
2025-01-14T08:43:50.8119445Z [2025/01/14 08:43:28.483 +00:00] [WARN] [tso.go:297] ["clock offset"] [] [jet-lag=466.470615ms] [prev-physical=2025/01/14 08:43:28.017 +00:00] [now=2025/01/14 08:43:28.483 +00:00] [update-physical-interval=50ms]
2025-01-14T08:43:50.8120896Z [2025/01/14 08:43:28.620 +00:00] [INFO] [cluster.go:1678] ["store has changed to serving"] [store-id=1] [store-address=tikv1]
2025-01-14T08:43:50.8121788Z [2025/01/14 08:43:29.101 +00:00] [INFO] [cluster.go:2407] ["store limit changed"] [store-id=1] [type=remove-peer] [rate-per-min=100000000]
2025-01-14T08:43:50.8122932Z [2025/01/14 08:43:29.341 +00:00] [INFO] [audit.go:128] ["audit log"] [service-info="{ServiceLabel:DeleteStore, Method:HTTP/1.1/DELETE:/pd/api/v1/store/2, CallerID:anonymous, IP:127.0.0.1, Port:52560, StartTime:2025-01-14 08:43:29 +0000 UTC, URLParam:{}, BodyParam:}"]
2025-01-14T08:43:50.8124241Z [2025/01/14 08:43:29.342 +00:00] [WARN] [cluster.go:1474] ["store has been offline"] [store-id=2] [store-address=tikv2] [physically-destroyed=false]
2025-01-14T08:43:50.8125240Z [2025/01/14 08:43:29.362 +00:00] [WARN] [tso.go:297] ["clock offset"] [] [jet-lag=234.966935ms] [prev-physical=2025/01/14 08:43:29.127 +00:00] [now=2025/01/14 08:43:29.361 +00:00] [update-physical-interval=50ms]
```
store 1 is changed to serving after offline

lhy1024:
We check store state in `runNodeStateCheckJob` and try to set store from preparing to serving.
If the code run in below order
[...]

Issue #9380, opened 2025-06-05
`TestConcurrentStoreStats` is flaky


rleungx:
### Which jobs are failing
```
    testutil.go:68: 
        	Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:68
        	            				/home/runner/work/pd/pd/server/cluster/cluster_test.go:3900
        	Error:      	Condition never satisfied
        	Test:       	TestConcurrentStoreStats
--- FAIL: TestConcurrentStoreStats (20.32s)
FAIL
```

### CI link
https://github.com/tikv/pd/actions/runs/15438943768/job/43451895510?pr=9377

glorv:
another failed case: https://github.com/tikv/pd/actions/runs/15751742731/job/44399896815?pr=9410

okJiang:
https://github.com/tikv/pd/actions/runs/16826473356/job/47664020182?pr=9634

rleungx:
closed by #9821

PR #9821, opened 2025-10-13
cluster: fix store state change problem (part 2)

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/8997 #9380

```commit-message
```

Ensure all changes are in the lock. And fix the judgement about up store count.

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster_test.go
+++ b/server/cluster/cluster_test.go
@@ -3921,6 +3921,78 @@ func TestConcurrentStoreStats(t *testing.T) {
 	wg.Wait()
 }
 
+func TestCheckStoresUpCountWithLowSpace(t *testing.T) {
+	re := require.New(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	_, opt, err := newTestScheduleConfig()
+	re.NoError(err)
+	cluster := newTestRaftCluster(ctx, mockid.NewIDAllocator(), opt, storage.NewStorageWithMemoryBackend())
+	cluster.coordinator = schedule.NewCoordinator(ctx, cluster, nil)
+	cluster.ruleManager = placement.NewRuleManager(ctx, storage.NewStorageWithMemoryBackend(), cluster, cluster.GetOpts())
+	if opt.IsPlacementRulesEnabled() {
+		err := cluster.ruleManager.Initialize(opt.GetMaxReplicas(), opt.GetLocationLabels(), opt.GetIsolationLevel(), false)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	// Add stores with low space to the cluster.
+	storeCount := uint64(3)
+	for i := uint64(1); i <= storeCount; i++ {
+		store := &metapb.Store{
+			Id:            i,
+			Address:       fmt.Sprintf("mock://tikv-%d:%d", i, i),
+			StatusAddress: fmt.Sprintf("mock://tikv-%d:%d", i, i+1),
+			Version:       "2.0.0",
+			DeployPath:    getTestDeployPath(i),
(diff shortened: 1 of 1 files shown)


server/cluster/cluster.go
```diff
@@ -1826,21 +1827,27 @@ func (c *RaftCluster) checkStores() {
...
-	c.storeStateLock.Lock(uint32(storeID))
-	defer c.storeStateLock.Unlock(uint32(storeID))
 	switch store.GetNodeState() {
 	case metapb.NodeState_Preparing:
 		readyToServe := store.GetUptime() >= c.opt.GetMaxStorePreparingTime() ||
 			c.GetTotalRegionCount() < core.InitClusterRegionThreshold
 		if !readyToServe && (c.IsPrepared() || (c.IsServiceIndependent(constant.SchedulingServiceName) && c.isStorePrepared())) {
 			kr := keyutil.NewKeyRange("", "")
-			threshold = c.getThreshold(stores, store, &kr)
+			threshold = c.getThreshold(c.GetStores(), store, &kr)
```
rleungx, line 1850: Is it necessary?

server/cluster/cluster.go
```diff
@@ -1886,7 +1893,11 @@ func (c *RaftCluster) checkStore(store *core.StoreInfo, stores []*core.StoreInfo
 		}
 	}
 	c.progressManager.UpdateProgress(store, regionSize, threshold)
-	return
+	// When store is preparing or serving, we think it is in up.
+	// `checkStore` only may change store state from preparing to serving.
+	// So we don't need to get store again.
+	isInUp = store.IsUp() && store.IsLowSpace(c.opt.GetLowSpaceRatio())
```
rleungx: Is it right?

server/cluster/cluster.go
lhy1024: Ensure that the latest stores are always used for calculations.

server/cluster/cluster.go
lhy1024: I also find this line of code confusing, just keeping it the same as before.

lhy1024: I see. It is a mistake from https://github.com/tikv/pd/pull/9333/files#r2127899721

rleungx:
Need a test case if we are going to cp.

lhy1024:
> Need a test case if we are going to cp.

Is TestConcurrentStoreStats enough?

rleungx:
> > Need a test case if we are going to cp.
> 
> Is TestConcurrentStoreStats enough?

The up state needs to be checked.

server/cluster/cluster.go
```diff
@@ -1886,7 +1893,11 @@ func (c *RaftCluster) checkStore(store *core.StoreInfo, stores []*core.StoreInfo
 		}
 	}
 	c.progressManager.UpdateProgress(store, regionSize, threshold)
-	return
+	// When store is preparing or serving, we think it is in up.
+	// `checkStore` only may change store state from preparing to serving.
+	// So we don't need to get store again.
+	isInUp = store.IsUp() && !store.IsLowSpace(c.opt.GetLowSpaceRatio())
```
bufferflies, line 1899: why adding the low space check?

lhy1024: from here 

Image

lhy1024: ref https://github.com/tikv/pd/pull/9821#discussion_r2425707809

lhy1024:
> > > Need a test case if we are going to cp.
> > 
> > 
> > Is TestConcurrentStoreStats enough?
> 
> The up state needs to be checked.

OK, I added it.

Issue #9825, opened 2025-10-14
There are also forward problems in scheduling, resource manager and pd


lhy1024:
It is similiar to https://github.com/tikv/pd/issues/9797

PR #9836, opened 2025-10-15
server, api: fix forward logic

lhy1024:

Issue Number: Close https://github.com/tikv/pd/issues/9825

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/api/api_test.go
+++ b/tests/server/api/api_test.go
@@ -69,7 +69,7 @@ func TestReconnect(t *testing.T) {
 	re.NotEmpty(leader)
 	for name, s := range cluster.GetServers() {
 		if name != leader {
-			res, err := tests.TestDialClient.Get(s.GetConfig().AdvertiseClientUrls + "/pd/api/v1/version")
+			res, err := tests.TestDialClient.Get(s.GetConfig().AdvertiseClientUrls + "/pd/api/v1/members")
 			re.NoError(err)
 			res.Body.Close()
 			re.Equal(http.StatusOK, res.StatusCode)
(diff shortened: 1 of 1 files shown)


tests/server/api/api_test.go
```diff
@@ -1032,6 +1032,134 @@ func TestRemovingProgress(t *testing.T) {
...
+	resp, err := tests.TestDialClient.Do(req)
+	re.NoError(err)
+	defer resp.Body.Close()
+	re.NoError(err)
+	re.NotEqual(followerName, resp.Header.Get(apiutil.XPDHandleHeader), "POST /config request should not be redirected")
+	testutil.Eventually(re, func() bool {
+		var followerCfg config.Config
+		err = testutil.ReadGetJSON(re, tests.TestDialClient, followerURL, &followerCfg)
+		re.NoError(err)
+		fmt.Println(followerCfg.Replication.MaxReplicas)
```
rleungx: remove it?

pkg/utils/apiutil/apiutil.go
```diff
@@ -65,6 +65,8 @@ const (
 	XRealIPHeader = "X-Real-Ip"
 	// XCallerIDHeader is used to mark the caller ID.
 	XCallerIDHeader = "X-Caller-ID"
+	// XPDHandleHeader is used to mark whether this request is handled by the PD.
+	XPDHandleHeader = "X-PD-Handle-By"
```
rleungx, line 69: Only for tests?

lhy1024: Yes. For now, it's only being used for testing. If similar issues occur with other interfaces in the future, we'll be able to identify which PD returned the response.

rleungx: Then, how about using a failpoint?

tests/server/api/api_test.go
```diff
@@ -69,7 +69,7 @@ func TestReconnect(t *testing.T) {
 	re.NotEmpty(leader)
 	for name, s := range cluster.GetServers() {
 		if name != leader {
-			res, err := tests.TestDialClient.Get(s.GetConfig().AdvertiseClientUrls + "/pd/api/v1/version")
+			res, err := tests.TestDialClient.Get(s.GetConfig().AdvertiseClientUrls + "/pd/api/v1/members")
```
rleungx, line 72: How about using another API? member is kind of special.

lhy1024: What's special here?

PR #9851, opened 2025-10-21
api: introduce `/readyz` and `/livez` for clearer health check

### LINKED ISSUE #9846: Introduce `/readyz` and `/livez` for clearer health checks
## Motivation

The current `/ready` endpoint is designed with a specific purpose: to check if a PD follower is ready to accept a leader transfer. It primarily verifies if the regions have been loaded. It returns `200 OK` if a leader transfer is possible and `500 Internal Server Error` otherwise.

If we plan to expand this api, it will confuse the users. Let us consider these two cases:

- Scenario 1:
  - livez (process is alive): true
  - readyz (pod is ready for traffic, e.g., election complete): true
  - Current `/ready` (ready for leader promotion, regions loaded): false (regions are still loading)

- Scenario 2:
  - livez (process is alive): true
  - readyz (pod is not ready, e.g., leader election in progress): false
  - Current `/ready` (ready for leader promotion, regions loaded): true

In both cases, the single `/ready` endpoint provides a signal that can be misinterpreted depending on the context (internal leader transfer vs. external service readiness).

To resolve this ambiguity and align with industry best practices, we propose introducing a new, more comprehensive health check endpoint.

## Proposal

To resolve this, we propose introducing a new set of health check endpoints, livez and readyz, aligning with Kubernetes and etcd conventions ([KEP-4331](https://github.com/kubernetes/enhancements/tree/master/keps/sig-etcd/4331-livez-readyz)), and eventually deprecating the existing `/ready` endpoint.

1. Introduce a New /readyz Endpoint

- `GET /readyz`
  - The endpoint will return 200 OK only if all underlying readiness checks pass. It acts as a logical AND for all sub-checks.
  - If any sub-check fails (e.g., leader_promotion is false), the main endpoint will return a non-200 status code.

- `GET /readyz/<check_name>`
  - This allows for checking a specific, granular readiness condition.
  - To maintain backward compatibility for the leader transfer use case, we will introduce `GET /readyz/leader_promotion`. This endpoint will replicate the exact logic of the old GET /ready endpoint.

- `GET /readyz?exclude=<check_name>`
  - The primary use case is for Kubernetes readiness probes, where we want to know if the pod is ready to serve traffic, but we might want to exclude the leader_promotion check. A request to `GET /readyz?exclude=leader_promotion` would be suitable for this.

2. Introduce `/livez` for Liveness Checks

- GET /livez
  - This endpoint will simply check if the process is alive

3. Deprecate `/ready`

- For a period, it can be maintained as an alias for `/readyz/leader_promotion` to ensure a smooth transition for existing users.

lhy1024:

Issue Number: Close #9846

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
image

### Release note

```release-note
None.
```

### DIFF
--- a/server/apiv2/handlers/ready.go
+++ b/server/apiv2/handlers/ready.go
@@ -100,7 +100,11 @@ func (reg *checkRegistry) installHandlers(r *gin.RouterGroup, basePath string) {
 				c.String(http.StatusServiceUnavailable, "[-]%s failed: %v\n", checkName, err)
 				return
 			}
-			c.String(http.StatusOK, "[+]%s ok\n", checkName)
+			c.Status(http.StatusOK)
+			if _, verbose := c.GetQuery("verbose"); verbose {
+				fmt.Fprintf(c.Writer, "[+]%s ok\n", checkName)
+			}
+			fmt.Fprintf(c.Writer, "ok\n")
 		})
 	}
 }
@@ -141,11 +145,11 @@ func (reg *checkRegistry) createRootHandler(allChecks []string) gin.HandlerFunc 
 			return
 		}
 
+		c.Status(http.StatusOK)
 		if _, verbose := c.GetQuery("verbose"); verbose {
-			c.String(http.StatusOK, output.String())
-		} else {
-			c.String(http.StatusOK, "ok\n")
+			fmt.Fprint(c.Writer, output.String())
 		}
+		fmt.Fprintf(c.Writer, "ok\n")
 	}
 }
 
--- a/tests/server/apiv2/handlers/ready_test.go
+++ b/tests/server/apiv2/handlers/ready_test.go
@@ -135,7 +135,7 @@ func TestReadyCheckAPI(t *testing.T) {
 			targetURL:          leaderURL,
 			path:               "/livez?verbose",
 			expectCode:         http.StatusOK,
-			expectBodyContains: []string{"[+]etcd-serializable-read ok"},
+			expectBodyContains: []string{"[+]etcd-serializable-read ok", "ok\n"},
 		},
 		{
 			name:      "livez: unhealthy due to serializable read error",
@@ -157,7 +157,7 @@ func TestReadyCheckAPI(t *testing.T) {
 			targetURL:          leaderURL,
 			path:               "/readyz?verbose",
 			expectCode:         http.StatusOK,
-			expectBodyContains: []string{"[+]leader-promotion ok", "[+]etcd-data-corruption ok"},
+			expectBodyContains: []string{"[+]leader-promotion ok", "[+]etcd-data-corruption ok", "ok\n"},
 		},
 		{
 			name:      "readyz: unhealthy when regions are not loaded (leader-promotion fails)",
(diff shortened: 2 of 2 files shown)


server/apiv2/handlers/ready.go
```diff
@@ -15,17 +15,227 @@
...
+	readyzRegistry.register("leader-promotion", checkLeaderPromotion)
+	readyzRegistry.register("etcd-data-corruption", checkEtcdDataCorruption)
+	readyzRegistry.register("etcd-serializable-read", checkEtcdSerializableRead)
+	readyzRegistry.register("etcd-linearizable-read", checkEtcdLinearizableRead)
+	readyzRegistry.register("etcd-non-learner", checkEtcdNonLearner)
+	readyzRegistry.installHandlers(r, "/readyz")
+
+	// --- Backward Compatibility ---
+	// The old /ready endpoint is now an alias for the leader-promotion readiness check.
+	r.GET("ready", Ready)
```
rleungx, line 55: break compatibility?

lhy1024: Why? It can pass `TestReadyAPI`, which is not modified.

server/apiv2/handlers/ready.go
```diff
@@ -15,17 +15,227 @@
...
+	livezRegistry.installHandlers(r, "/livez")
+
+	// --- Readiness Checks ---
+	readyzRegistry := NewCheckRegistry()
+	readyzRegistry.register("leader-promotion", checkLeaderPromotion)
+	readyzRegistry.register("etcd-data-corruption", checkEtcdDataCorruption)
+	readyzRegistry.register("etcd-serializable-read", checkEtcdSerializableRead)
+	readyzRegistry.register("etcd-linearizable-read", checkEtcdLinearizableRead)
+	readyzRegistry.register("etcd-non-learner", checkEtcdNonLearner)
+	readyzRegistry.installHandlers(r, "/readyz")
```
liubog2008, line 51: Which checker is for kubernetes readiness probe?

lhy1024: only `readyz` is enough. If you don't want to check the leader-promotion, you may use `readyz?exclude=leader-promotion`

liubog2008: `RegionIsLoaded` is needed, and a stable checker is needed to avoid a new unnecessary checker is added into `/readyz`

server/apiv2/handlers/ready.go
```diff
@@ -15,17 +15,227 @@
...
+func (reg *checkRegistry) installHandlers(r *gin.RouterGroup, basePath string) {
+	checkNames := make([]string, 0, len(reg.checks))
+	for name := range reg.checks {
+		checkNames = append(checkNames, name)
+	}
+	sort.Strings(checkNames) // ensure consistent order
+
+	rootHandler := reg.createRootHandler(checkNames)
+	r.GET(basePath, rootHandler)     // for /readyz and /livez endpoints
+	r.GET(basePath+"/", rootHandler) // also handle trailing slash
```
rleungx: etcd will return 404

server/apiv2/handlers/ready.go
```diff
@@ -15,17 +15,227 @@
...
+func NewCheckRegistry() *checkRegistry {
+	return &checkRegistry{
+		checks: make(map[string]checkFunction),
+	}
+}
+
+// register adds a new check to the registry.
+func (reg *checkRegistry) register(name string, check checkFunction) {
+	if _, exists := reg.checks[name]; exists {
+		log.Fatal("check function is already registered", zap.String("name", name))
```
rleungx: Is it fatal necessary?

server/apiv2/handlers/ready.go
lhy1024: ok

server/apiv2/handlers/ready.go
lhy1024: Do you have any ideas? Error or Warning?

rleungx: warning is enough, or simply ignore?

okJiang (approved):

PR #1361, opened 2018-12-06
*: Add shuffle hot region scheduler

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

### What problem does this PR solve? 
Mainly use this schedule in the test. Used to test whether TiKV will be affected under poor hotspot scheduling.

### What is changed and how it works?
 move hot region and transfer the leader to the newly added hot peer

### Check List 

Tests 

 - Unit test
 - Manual test

Code changes
 - Add an API interfaces

PR #9862, opened 2025-10-22
client: updating keyspace group must check the previous version

### LINKED ISSUE #6770: Round-robin updateMember may cause the timestamp to fallback
## Bug Report

We use a round-robin way to request the TSO server with `FindGroupByKeyspaceID` gRPC call.

https://github.com/tikv/pd/blob/b176a0a573a6633f0185cc8e2225af6b6371485c/client/tso_service_discovery.go#L418-L441

However, if the requested TSO server satisfies this kind of case, it may cause the timestamp to fallback.

- TSO node A initialized group 1.
- TSO node B has not initialized group 1 yet.

If we do `updateMember` quickly while requesting the TSO, it may cause the keyspace group in the request to change between the default group and group 1, resulting in the timestamp fallback.

bufferflies:

Issue Number: Close #6770

[Merged]kv proto: [#1361](https://github.com/pingcap/kvproto/pull/1361)

```commit-message
'FindGroupByKeyspaceID' must carry the current keyspace group version, and the server needs to check the memory
state isn't stale, and then return the server keyspace group info.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/errors.toml
+++ b/errors.toml
@@ -556,6 +556,11 @@ error = '''
 keyspace group %v is in split state
 '''
 
+["PD:keyspace:ErrKeyspaceGroupModRevisionStale"]
+error = '''
+keyspace group mod revision is stale
+'''
+
 ["PD:keyspace:ErrKeyspaceGroupNotEnoughReplicas"]
 error = '''
 not enough replicas in the keyspace group
@@ -579,11 +584,6 @@ keyspace group %v is not in split state
 ["PD:keyspace:ErrKeyspaceGroupPrimaryNotFound"]
 error = '''
 primary of keyspace group does not exist
-'''
-
-["PD:keyspace:ErrKeyspaceGroupModReVisionStale"]
-error = '''
-keyspace group mod revision is stale
 '''
 
 ["PD:keyspace:ErrKeyspaceGroupWithEmptyKeyspace"]
--- a/pkg/tso/keyspace_group_manager_test.go
+++ b/pkg/tso/keyspace_group_manager_test.go
@@ -731,7 +731,7 @@ func (suite *keyspaceGroupManagerTestSuite) runTestLoadKeyspaceGroupsAssignment(
 	re *require.Assertions,
 	numberOfKeyspaceGroupsToAdd int,
 	loadKeyspaceGroupsBatchSize int64, // set to 0 to use the default value
-	probabilityAssignToMe int,         // percentage of assigning keyspace groups to this host/pod
+	probabilityAssignToMe int, // percentage of assigning keyspace groups to this host/pod
 ) {
 	expectedGroupIDs := []uint32{}
 	mgr := suite.newUniqueKeyspaceGroupManager(loadKeyspaceGroupsBatchSize)


client/servicediscovery/tso_service_discovery.go
```diff
@@ -66,16 +66,27 @@ type keyspaceGroupSvcDiscovery struct {
 	secondaryURLs []string
 	// urls are the primary/secondary serving URL
 	urls []string
+
+	// version is used to avoid updating stale info
```
ystaticy: How about adding a comment to clarify where the version is sourced from and briefly explain its usage?

client/servicediscovery/tso_service_discovery.go
```diff
@@ -160,6 +171,7 @@ func NewTSOServiceDiscovery(
 		primaryURL:    "",
 		secondaryURLs: make([]string, 0),
 		urls:          make([]string, 0),
+		version:       0,
```
ystaticy: When the TSO client starts up, does it use version=0 for the initial request?

Furthermore, if this initial request is routed to a TSO service with a lower version, is that behavior expected?

client/servicediscovery/tso_service_discovery.go
bufferflies: done

client/servicediscovery/tso_service_discovery.go
bufferflies: No, this version just works for requesting the keyspace group. Most updates don't relate to this client's keyspace ID, so it doesn't need to be updated.

bufferflies: The higher version of the server side doesn't mean that the client's keyspace group has changed.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -66,16 +66,31 @@ type keyspaceGroupSvcDiscovery struct {
 	secondaryURLs []string
 	// urls are the primary/secondary serving URL
 	urls []string
+
+	// version is used to avoid updating stale info
+	// this version is provided by etcd watcher on the server side
+	// and increases monotonically.
+	// If the new version is less than the current version, it means that the requested tso service hasn't applied the
```
ystaticy: How about clarifying the source of both `the new version` and `the current version`?

pkg/tso/keyspace_group_manager.go
```diff
@@ -91,6 +91,8 @@ type state struct {
 	// Once a group receives its first TSO request and pass the certain check, it will be added to this map.
 	// Being merged will cause the group to be removed from this map eventually if the merge is successful.
 	requestedGroups map[uint32]struct{}
+
+	modVersion uint64
```
ystaticy: Please add comments for `modVersion uint64`.

client/servicediscovery/tso_service_discovery.go
bufferflies: done

pkg/tso/keyspace_group_manager.go
bufferflies: done

pkg/tso/keyspace_group_manager.go
```diff
@@ -515,12 +538,22 @@ func (kgm *KeyspaceGroupManager) InitializeGroupWatchLoop() error {
...
 		}
+		failpoint.Inject("SkipKeyspaceWatch", func(val failpoint.Value) {
+			addr, ok := val.(string)
+			if ok && addr == kgm.electionNamePrefix {
+				failpoint.Return(nil)
+			}
+		})
+		if len(event) > 0 {
+			last := event[len(event)-1]
+			kgm.SetModVersion(uint64(last.Kv.ModRevision))
```
rleungx: Shall we use the kv mod revision or the revision in the response?

bufferflies: No, the response header revision is the latest version of the global KV. It may bring some probs: 
1. It updates stale if it lists all keyspaces and updates the cache one by one; the revision between the first kv and last kv is the same. If we update the revision at first and the second kv does not update. The cache is stale.
2. if no update in the etcd key, the header revision is always the same, we can't know the kv is the latest.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -413,12 +428,16 @@ func (c *tsoServiceDiscovery) updateMember() error {
 
 	keyspaceID := c.GetKeyspaceID()
 	var keyspaceGroup *tsopb.KeyspaceGroup
+	var version uint64
```
rleungx: How about unifying the naming, using modRevision and curModRevision?

PR #9865, opened 2025-10-22
Add 'store' label to metric pd_cluster_status in 8.5.

### LINKED ISSUE #9855: Add 'store' label to metric pd_cluster_status.
## Enhancement Task

Currently gauge metric pd_cluster_status is aggregated across all stores. This is pretty inconvenient, because if you have long running deployment across TiKV component this metric indicate problematic store for a long period of time and it will not be possible to distinguish different tikv stores contributed to this metric.

What is suggested.
Instead of emitting counts across all stores [here](https://github.com/tikv/pd/blob/aa5b0e965087ebdd2c1d843eff5a9043960d45a6/pkg/statistics/store_collection.go#L185-L207), we will emit the same gauge metrics with additional label 'store' with values either 0 or 1 in [observe() method](https://github.com/tikv/pd/blob/aa5b0e965087ebdd2c1d843eff5a9043960d45a6/pkg/statistics/store_collection.go#L80-L118). This will allow to get more detailed view from the metric. And aggregation across all values of 'store' label will let you to get the same metrics as today.

SerjKol80:

Issue Number: Close #9855 

```commit-message
Add 'store' label to metric pd_cluster_status in 8.5.
```

### Check List

Tests

- Unit test

Code changes

- Code now emit Gauge metric pd_cluster_status for each store separately and adds label 'store' containing storeId. Previously code aggregated this metric across all stores and then emitted one time.

Side effects

- metric pd_cluster_status now needs to be aggregated in metic system across all stores if you need value across whole cluster.

Related changes

None

### Release note

```release-note
We added 'store' label to metric pd_cluster_status. This new label contains store ID for which datapoint is emitted. How if you still need the value from this metric for the whole cluster, then you need to aggregate it across all stores
```

### DIFF
--- a/pkg/statistics/metrics.go
+++ b/pkg/statistics/metrics.go
@@ -47,7 +47,7 @@ var (
 			Subsystem: "cluster",
 			Name:      "status",
 			Help:      "Status of the cluster.",
-		}, []string{"type"})
+		}, []string{"type", "store"})
 
 	placementStatusGauge = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
--- a/pkg/statistics/store_collection.go
+++ b/pkg/statistics/store_collection.go
@@ -29,29 +29,31 @@ import (
 const (
 	unknown   = "unknown"
 	labelType = "label"
+
+	clusterStatusStoreUpCount           = "store_up_count"
+	clusterStatusStoreDisconnectedCount = "store_disconnected_count"
+	clusterStatusStoreSlowCount         = "store_slow_count"
+	clusterStatusStoreDownCount         = "store_down_count"
+	clusterStatusStoreUnhealthCount     = "store_unhealth_count"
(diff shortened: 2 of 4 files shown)


SerjKol80:
@lhy1024 
I don't think cherry-pick works here, because this part of code was refactored significantly, thus will have a lot of conficts.
Instead I implemented the simial logic in ````master```` branch: https://github.com/tikv/pd/pull/9898

SerjKol80:
@lhy1024 this feature is now merged to master: https://github.com/tikv/pd/pull/9898
I cherry-picked it to this PR now.
Would you please take a look?

SerjKol80:
@lhy1024 @bufferflies Thank you for looking at this PR. How can I trigger merging?

bufferflies:
bugfix cp：https://github.com/tikv/pd/pull/9945

Issue #6892, opened 2023-08-03
`TestEtcdScaleInAndOut` failed


rleungx:
### Which jobs are failing

```
--- FAIL: TestEtcdScaleInAndOut (0.45s)
    etcdutil_test.go:310: 
        	Error Trace:	/home/runner/work/pd/pd/pkg/utils/etcdutil/etcdutil_test.go:310
        	Error:      	Received unexpected error:
        	            	[PD:etcd:ErrEtcdMemberRemove]etcdserver: server stopped: etcdserver: server stopped
        	            	github.com/pingcap/errors.AddStack
        	            		/home/runner/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/errors.go:174
        	            	github.com/pingcap/errors.(*Error).GenWithStackByCause
        	            		/home/runner/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/normalize.go:307
        	            	github.com/tikv/pd/pkg/utils/etcdutil.RemoveEtcdMember
        	            		/home/runner/work/pd/pd/pkg/utils/etcdutil/etcdutil.go:124
        	            	github.com/tikv/pd/pkg/utils/etcdutil.TestEtcdScaleInAndOut
        	            		/home/runner/work/pd/pd/pkg/utils/etcdutil/etcdutil_test.go:309
        	            	testing.tRunner
        	            		/opt/hostedtoolcache/go/1.20.1/x64/src/testing/testing.go:1576
        	            	runtime.goexit
        	            		/opt/hostedtoolcache/go/1.20.1/x64/src/runtime/asm_amd64.s:1598
        	Test:       	TestEtcdScaleInAndOut
```

### CI link

https://github.com/tikv/pd/actions/runs/5748065431/job/15580210773?pr=6891

lhy1024:
It met `server stopped` before we removed this server from the cluster.

There is only a log with it `2023-08-03T07:55:34.9662606Z [2023/08/03 07:54:49.387 +00:00] [WARN] [retry_interceptor.go:62] ["retrying of unary invoker failed"] [target=endpoint://client-aa84e8be-f741-46c2-b8ce-334a8bf853e9/127.0.0.1:33107] [attempt=0] [error="rpc error: code = Unavailable desc = etcdserver: server stopped"]`

lhy1024:
```
[...]

PR #9870, opened 2025-10-23
tests: fix `TestEtcdScaleInAndOut`

rleungx:

Issue Number: ref #6892.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/utils/etcdutil/etcdutil_test.go
+++ b/pkg/utils/etcdutil/etcdutil_test.go
@@ -228,10 +228,16 @@ func TestEtcdScaleInAndOut(t *testing.T) {
 	defer etcd2.Close()
 	checkMembers(re, client2, []*embed.Etcd{etcd1, etcd2})
 
-	// scale in etcd1
-	_, err = RemoveEtcdMember(client1, uint64(etcd1.Server.ID()))
+	// Create a client connected to etcd2 to perform the removal
+	cfg2 := etcd2.Config()
+	client3, err := CreateEtcdClient(nil, cfg2.ListenClientUrls)
 	re.NoError(err)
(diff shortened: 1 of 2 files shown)


pkg/utils/etcdutil/etcdutil_test.go
```diff
@@ -228,10 +228,16 @@ func TestEtcdScaleInAndOut(t *testing.T) {
...
-	_, err = RemoveEtcdMember(client1, uint64(etcd1.Server.ID()))
+	// Create a client connected to etcd2 to perform the removal
+	cfg2 := etcd2.Config()
+	client3, err := CreateEtcdClient(nil, cfg2.ListenClientUrls)
 	re.NoError(err)
-	checkMembers(re, client2, []*embed.Etcd{etcd2})
+	defer client3.Close()
+
+	// scale in etcd1 using client3 (connected to etcd2)
+	_, err = RemoveEtcdMember(client3, uint64(etcd1.Server.ID()))
```
okJiang, line 238: Can we use client2 directly?

tests/dashboard/service_test.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !without_dashboard
```
lhy1024, line 15: Why add it?

rleungx: TiDB X doesn't have a dashboard, so there is no need to run these tests.

pkg/utils/etcdutil/etcdutil_test.go
rleungx: client2 connects to etcd1

PR #9746, opened 2025-09-17
feat(resourcemanager): refactor token bucket slot allocation

JmPotato:
### What problem does this PR solve?

Issue Number: Close #9605.

### What is changed and how does it work?

```commit-message
- Replaced `requireTokensSum` and `clientConsumptionTokensSum` with `ruTracker` for per-slot RU demand tracking.  
- Refactored `balanceSlotTokens` to fairly allocate tokens across slots based on RU demand and over-demand proportion.
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

### Release note

```release-note
Fix the issue of uneven RU allocation rates among different TiDB instances within the same resource group.
```

PR #9887, opened 2025-10-29
feat(resourcemanager): refine token distribution logic with weighted consumption bias

### LINKED ISSUE #9605: resourcemanager: slot token limit can be too low when the ru request rate is not even
### What did you do?
When the ru consumption rate of different tidb instances are not evenly distributed, even when the ru consumption rate is far beyond the group fill rate, due to the current ru distributing strategy, the tidb with the maximum ru consuming rate can result with the min ru fill rate and cause high ru wait duration.

https://github.com/tikv/pd/blob/aa75cbbc452906ff205cdde430907423e01e9119/pkg/mcs/resourcemanager/server/token_buckets.go#L296-L307

This current distributing strategy that only based on the history consumption rate works really bad in the uneven scenario.

JmPotato:

Issue Number: Close #9605.

```commit-message
Introduced `defaultConsumptionBiasWeight` to adjust how client consumption history influences token assignments.
```

### Check List

Tests

- Unit test
- Integration test

Before:

before

After:

after

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/token_buckets.go
+++ b/pkg/mcs/resourcemanager/server/token_buckets.go
@@ -26,13 +26,14 @@ import (
 )
 
 const (
-	defaultRefillRate         = 10000
-	defaultModeratedBurstRate = 10000
-	defaultInitialTokens      = 10 * 10000
-	defaultReserveRatio       = 0.5
-	defaultLoanCoefficient    = 2
-	maxAssignTokens           = math.MaxFloat64 / 1024 // assume max client connect is 1024
-	slotExpireTimeout         = 10 * time.Minute
+	defaultRefillRate            = 10000
+	defaultModeratedBurstRate    = 10000
+	defaultInitialTokens         = 10 * 10000
+	defaultReserveRatio          = 0.5
+	defaultLoanCoefficient       = 2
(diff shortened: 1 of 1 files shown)


pkg/mcs/resourcemanager/server/token_buckets.go
```diff
@@ -26,13 +26,14 @@ import (
...
-	maxAssignTokens           = math.MaxFloat64 / 1024 // assume max client connect is 1024
-	slotExpireTimeout         = 10 * time.Minute
+	defaultRefillRate            = 10000
+	defaultModeratedBurstRate    = 10000
+	defaultInitialTokens         = 10 * 10000
+	defaultReserveRatio          = 0.5
+	defaultLoanCoefficient       = 2
+	maxAssignTokens              = math.MaxFloat64 / 1024 // assume max client connect is 1024
+	slotExpireTimeout            = 10 * time.Minute
+	defaultConsumptionBiasWeight = 0.75
```
rleungx, line 36: Is there any reason to set it to 0.75?

JmPotato: This PR is no longer a suitable solution. Let's focus our review on https://github.com/tikv/pd/pull/9746 instead.

Issue #9895, opened 2025-10-31
"watch request" log flood


okJiang:
## Enhancement Task

Starting form v8.1, the PD leader node is flooded with messages like:

```
{"level":"INFO","time":"2025/09/05 01:49:57.122 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/settings","range-end":"resource_group/settingt","start-revision":2929620745}
{"level":"INFO","time":"2025/09/05 01:49:57.219 +00:00","caller":"grpc_service.go:2173","message":"update service GC safe point","service-id":"backup-stream-pitr2-343690334","expire-at":1757044197,"safepoint":460596657291543244}
{"level":"INFO","time":"2025/09/05 01:49:57.221 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/settings","range-end":"resource_group/settingt","start-revision":1876548480}
{"level":"INFO","time":"2025/09/05 01:49:57.221 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/controller","range-end":"resource_group/controlles","start-revision":2435631273}
{"level":"INFO","time":"2025/09/05 01:49:57.221 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/settings","range-end":"resource_group/settingt","start-revision":2669002567}
{"level":"INFO","time":"2025/09/05 01:49:57.322 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/settings","range-end":"resource_group/settingt","start-revision":2929914284}
{"level":"INFO","time":"2025/09/05 01:49:57.322 +00:00","caller":"grpc_service.go:99","message":"watch request","key":"resource_group/settings","range-end":"resource_group/settingt","start-revision":2929621484}
```

```
sudo cat pd-2025-09-11T16-23-44.243.log |wc -l
3524452
sudo cat pd-2025-09-11T16-23-44.243.log |grep "watch request"|wc -l
731503

```

Here's the details:
```
On the PD node:
 sudo cat pd-2025-09-25T07-02-06.543.log|grep -i "watch request" |wc -l
1385145


On a TiKV node:
sudo cat tikv.log |grep "pd meta client creating watch stream"|wc -l
74
sudo cat tikv.log |grep "required revision has been compacted"|wc -l
1
sudo cat tikv.log |grep "failed to watch resource groups"|wc -l
46
```
Most of the TiKV nodes didn't have these keywords

[...]

PR #9896, opened 2025-10-31
service: add log for watch

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/9895

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/metastorage/server/grpc_service.go
+++ b/pkg/mcs/metastorage/server/grpc_service.go
@@ -110,7 +110,7 @@ func (s *Service) Watch(req *meta_storagepb.WatchRequest, server meta_storagepb.
 			return nil
 		case res := <-watchChan:
 			if res.Err() != nil {
-				log.Error("watch channel failed", zap.Error(res.Err()))
+				log.Warn("watch channel failed", zap.Error(res.Err()))
 				var resp meta_storagepb.WatchResponse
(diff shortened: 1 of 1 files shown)


pkg/mcs/metastorage/server/grpc_service.go
```diff
@@ -110,6 +110,7 @@ func (s *Service) Watch(req *meta_storagepb.WatchRequest, server meta_storagepb.
 			return nil
 		case res := <-watchChan:
 			if res.Err() != nil {
+				log.Error("watch channel failed", zap.Error(res.Err()))
```
bufferflies: Is it necessary to use the error level log?  If yes,  you should explain how to solve it?

okJiang: Updated it to `warn`

pkg/mcs/metastorage/server/grpc_service.go
lhy1024, line 86: Will we print this error?

bufferflies (approved):

pkg/mcs/metastorage/server/grpc_service.go
okJiang: This error is before "watch request". I think it is not related to this issue.

PR #9898, opened 2025-10-31
statistics: Add 'store' label to metric pd_cluster_status.

### LINKED ISSUE #9855: Add 'store' label to metric pd_cluster_status.

SerjKol80:

Issue Number: Close #9855 

```commit-message
Add 'store' label to metric pd_cluster_status.
```

### Check List

Tests
- Unit test

Code changes
- Metrics only

Side effects
- metric pd_cluster_status now needs to be aggregated in metic system across all stores if you need value across whole cluster.

Related changes
N/A

### Release note

```release-note
metric "pd_cluster_status" now has additional label "store" containing ID of the store.
```

### DIFF
--- a/pkg/statistics/metrics.go
+++ b/pkg/statistics/metrics.go
@@ -47,7 +47,7 @@ var (
 			Subsystem: "cluster",
 			Name:      "status",
 			Help:      "Status of the cluster.",
-		}, []string{"type", "engine"})
+		}, []string{"type", "engine", "store"})
 
 	placementStatusGauge = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
--- a/pkg/statistics/store_collection.go
+++ b/pkg/statistics/store_collection.go
@@ -30,116 +30,89 @@ import (
 const (
 	unknown   = "unknown"
 	labelType = "label"
-)
 
-var (
-	// tikv status counters.
-	tikvUpCounter          = clusterStatusGauge.WithLabelValues("store_up_count", "tikv")
-	tikvDiconnectedCounter = clusterStatusGauge.WithLabelValues("store_disconnected_count", "tikv")
-	tikvDownCounter        = clusterStatusGauge.WithLabelValues("store_down_count", "tikv")
-	tikvUnhealthCounter    = clusterStatusGauge.WithLabelValues("store_unhealth_count", "tikv")
-	tikvOfflineCounter     = clusterStatusGauge.WithLabelValues("store_offline_count", "tikv")
-	tikvTombstoneCounter   = clusterStatusGauge.WithLabelValues("store_tombstone_count", "tikv")
-	tikvLowSpaceCounter    = clusterStatusGauge.WithLabelValues("store_low_space_count", "tikv")
-	tikvPreparingCounter   = clusterStatusGauge.WithLabelValues("store_preparing_count", "tikv")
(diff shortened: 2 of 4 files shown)


pkg/statistics/store_collection_test.go
```diff
@@ -76,22 +76,7 @@ func TestStoreStatistics(t *testing.T) {
 		ObserveHotStat(store, storesStats)
 	}
 	stats := storeStats.stats
-	tikvStats := stats.engineStatistics[core.EngineTiKV]
```
lhy1024, line 79: Do we need to add a new test for it?

pkg/statistics/metrics.go
```diff
@@ -47,7 +47,7 @@ var (
 			Subsystem: "cluster",
 			Name:      "status",
 			Help:      "Status of the cluster.",
-		}, []string{"type", "engine"})
+		}, []string{"type", "engine", "store"})
```
lhy1024, line 50: Do we need to update `metrics/grafana/pd.json`?

pkg/statistics/store_collection_test.go
SerjKol80: @lhy1024 
Since we don't have aggregation logic anymore, then the test was removed for that part.
So, the only thing that might be tested in a new logic is if actual metric is emitted. I looked through existing tests and didn't find any tests validating actual metric emission. Thus, no new tests.

pkg/statistics/metrics.go
SerjKol80: @lhy1024 
I don't think so. All queries for ````pd_cluster_status```` metrics already aggregate that metric with sum(). Thus, there should no be any changes there. The description of that metrics should be updated, but I wasn't able to find doc describing metrics in this repo.

lhy1024 (approved):

SerjKol80:
@lhy1024 
Thank you. Would you initiate the merge. It looks like I don't have permission.

lhy1024:
@bufferflies PTAL

Tema (approved):

bufferflies (approved):

Tema:
@lhy1024 @bufferflies  could you please approve 4 pending workflows to run ^

pkg/statistics/store_collection.go
```diff
@@ -19,6 +19,7 @@ import (
 	"strconv"
 
 	"github.com/pingcap/kvproto/pkg/metapb"
```
lhy1024, line 21: plz fix ci

PR #9901, opened 2025-11-04
feat(keyspace): add cached keyspace state and enforce enabled-state access checks

JmPotato:

Issue Number: ref #9707.

```commit-message
- Introduced `keyspaceStateLookup` cache in `Manager` to store and retrieve keyspace states efficiently.
- Added `GetKeyspaceStateByID` and `GetEnabledKeyspaceNameByID` methods:
  - The former fetches the keyspace state from cache or persistent storage.
  - The latter ensures that only enabled keyspaces can be referenced by ID.
- Updated cluster logic (`collectStorageSize`, `collectDFSStats`) to use the enabled-only name lookup, preventing operations on disabled or archived keyspaces.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -849,6 +849,13 @@ func (manager *Manager) GetKeyspaceNameByID(id uint32) (string, error) {
 // GetKeyspaceStateByID gets the keyspace state by ID, which will try to get it from the cache first.
 // If not found, it will try to get it from the storage.
 func (manager *Manager) GetKeyspaceStateByID(id uint32) (keyspacepb.KeyspaceState, error) {
+	if id == constant.NullKeyspaceID {
+		return keyspacepb.KeyspaceState_DISABLED, nil
+	}
+	state, ok := manager.keyspaceStateLookup.Load(id)
+	if ok {
+		return state.(keyspacepb.KeyspaceState), nil
+	}
 	var loadedState keyspacepb.KeyspaceState
 	// If the keyspace state is not in the cache, try to get it from the storage.
 	meta, err := manager.LoadKeyspaceByID(id)


pkg/keyspace/keyspace.go
```diff
@@ -97,8 +97,9 @@ type Manager struct {
 	kgm *GroupManager
 	// nextPatrolStartID is the next start id of keyspace assignment patrol.
 	nextPatrolStartID uint32
-	// cached keyspace name for each keyspace ID.
-	keyspaceNameLookup sync.Map // store as ID(uint32) -> name(string)
+	// cached keyspace meta info for each keyspace ID.
+	keyspaceNameLookup  sync.Map // store as ID(uint32) -> name(string)
+	keyspaceStateLookup sync.Map // store as ID(uint32) -> state(keyspacepb.KeyspaceState)
```
rleungx, line 102: Will the state be accessed frequently?

JmPotato: According to the implementation of this PR, we will check the state of the corresponding keyspace each time metering data is reported, so the access frequency of this state map will be the same as the metering reporting interval, and it will be accessed up to N times per minute (N depends on the number of keyspaces).

pkg/keyspace/keyspace.go
```diff
@@ -843,6 +846,34 @@ func (manager *Manager) GetKeyspaceNameByID(id uint32) (string, error) {
 	return actual.(string), nil
 }
 
+// GetKeyspaceStateByID gets the keyspace state by ID, which will try to get it from the cache first.
+// If not found, it will try to get it from the storage.
+func (manager *Manager) GetKeyspaceStateByID(id uint32) (keyspacepb.KeyspaceState, error) {
+	var loadedState keyspacepb.KeyspaceState
+	// If the keyspace state is not in the cache, try to get it from the storage.
+	meta, err := manager.LoadKeyspaceByID(id)
+	if err != nil {
```
rleungx: What if the meta exists, but err is not nil?

JmPotato: After checking the code of `(*Manager).LoadKeyspaceByID` and `(*StorageEndpoint).LoadKeyspaceMeta`, there is no such case.

rleungx: The txn itself may have an error?

JmPotato: After some consideration, I changed it to only check whether meta is nil. PTAL.

pkg/keyspace/keyspace.go
```diff
@@ -843,6 +846,35 @@ func (manager *Manager) GetKeyspaceNameByID(id uint32) (string, error) {
 	return actual.(string), nil
 }
 
+// GetKeyspaceStateByID gets the keyspace state by ID, which will try to get it from the cache first.
+// If not found, it will try to get it from the storage.
+func (manager *Manager) GetKeyspaceStateByID(id uint32) (keyspacepb.KeyspaceState, error) {
+	var loadedState keyspacepb.KeyspaceState
+	// If the keyspace state is not in the cache, try to get it from the storage.
```
okJiang, line 860: Seems you do not use the cache?

JmPotato: Oops, fixed.

okJiang (approved):

PR #9902, opened 2025-11-04
client: fix stale dns issue when retrying

### LINKED ISSUE #9894: Connection won't be re-established if DNS has a delay
### What did you do?

We have a cache to manage the mapping between URL and gRPC connection:
https://github.com/tikv/pd/blob/e2ce37be70eb478137bd22b34843c5fff8481d28/client/servicediscovery/tso_service_discovery.go#L124

But if the DNS has a delay, the gRPC connection won't be re-established. If the pod IP can be reused, there is a possibility that the request is sent to the wrong place. We have encountered this issue in the serverless environment, which causes the tso request to be sent to the tso server, which is not the primary. And the client keeps trying at the same connection, which is wrong until the TiDB is restarted.

rleungx:

Issue Number: Close #9894.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/pkg/utils/grpcutil/grpcutil.go
+++ b/client/pkg/utils/grpcutil/grpcutil.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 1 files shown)


client/clients/tso/client.go
```diff
@@ -198,11 +198,12 @@ func (c *Cli) getTSOLeaderClientConn() (*grpc.ClientConn, string) {
 	if len(url) == 0 {
 		log.Fatal("[tso] the tso leader should exist")
 	}
-	cc, ok := c.svcDiscovery.GetClientConns().Load(url)
-	if !ok {
-		return nil, url
+	cc, err := c.svcDiscovery.GetOrCreateGRPCConn(url)
```
bufferflies, line 201: https://github.com/tikv/pd/blob/f86063f6063813ddd6a2ec23d5b645170049b722/client/clients/router/client.go#L333-L337
I think this code also needs to change

client/clients/tso/dispatcher.go
```diff
@@ -371,11 +371,15 @@ func (td *tsoDispatcher) handleProcessRequestError(
...
-	conCtxMgr.Release(streamURL)
 	// Update the member list to ensure the latest topology is used before the next batch.
 	svcDiscovery := td.provider.getServiceDiscovery()
+	// Release this stream from the manager due to error.
+	conCtxMgr.Release(streamURL)
+	if errs.ShouldRedial(err) {
+		svcDiscovery.RemoveClientConn(streamURL)
+	}
 	if errs.IsLeaderChange(err) {
+		metrics.TSOLeaderChangedCounter.Inc()
```
bufferflies, line 382: What if this err is `NotPrimaryErr`?

bufferflies (commented):
rest lgtm

client/clients/tso/client.go
rleungx: I can raise another PR to change it.

client/clients/tso/dispatcher.go
rleungx: Do you mean the naming issue?

bufferflies: No, I mean this counter should not include `notLeaderError` also `NotPrimaryErr`

bufferflies (approved):

PR #9919, opened 2025-11-11
cluster: fix tiflash cannot offline

### LINKED ISSUE #9933: TiFlash cannot be deleted if the replica is not enough
## Bug Report

Image

We need to skip TiFlash nodes when `checkReplicaBeforeOfflineStore`.

rleungx:

Issue Number: Close #9933

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1556,9 +1556,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 }
 
 func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
-	upStores := c.getUpStores( /* isInclude */ func(store *core.StoreInfo) bool {
-		return store.IsTiKV()
-	})
+	upStores := c.getUpStores()
 	expectUpStoresNum := len(upStores) - 1
 	if expectUpStoresNum < c.opt.GetMaxReplicas() {
 		return errs.ErrStoresNotEnough.FastGenByArgs(storeID, expectUpStoresNum, c.opt.GetMaxReplicas())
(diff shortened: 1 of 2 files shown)


server/cluster/cluster.go
```diff
@@ -1556,7 +1556,9 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 }
 
 func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
-	upStores := c.getUpStores()
+	upStores := c.getUpStores(func(store *core.StoreInfo) bool {
```
lhy1024:
```suggestion
	upStores := c.getUpStores(/*isExcluded */func(store *core.StoreInfo) bool {
```

server/cluster/cluster.go
```diff
@@ -1590,9 +1592,12 @@ func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
 	return nil
 }
 
-func (c *RaftCluster) getUpStores() []uint64 {
+func (c *RaftCluster) getUpStores(isInclude ...func(store *core.StoreInfo) bool) []uint64 {
```
JmPotato: It seems that currently there’s only the usage of `IsTiKV()`. How about writing the logic directly inside the function instead of introducing the `isInclude` parameter?

rleungx: done

server/cluster/cluster.go
```diff
@@ -1592,10 +1590,11 @@ func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
 	return nil
 }
 
-func (c *RaftCluster) getUpStores(isInclude ...func(store *core.StoreInfo) bool) []uint64 {
+// getUpStores gets all up TiKV stores.
+func (c *RaftCluster) getUpStores() []uint64 {
```
lhy1024, line 1594: How about `getUpTiKVs`

rleungx: Actually, I prefer not to introduce the TiKV/TiFlash concept as much as possible.

PR #9920, opened 2025-11-11
build(response): integrate easyjson generation

JmPotato:

Issue Number: ref #4399.

```commit-message
- Regenerated easyjson code for `pkg/response/region.go` to support for new `approximate_columnar_kv_size` field in `RegionInfo`.
- Introduced `generate-easyjson` Makefile target and integrated it into build deps.
```

### Check List

Tests

- Unit test
- Integration test
- Manual test

image

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/response/region_test.go
+++ b/pkg/response/region_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
+	"github.com/pingcap/kvproto/pkg/replication_modepb"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -218,6 +219,10 @@ check-plugin:
 	@echo "checking plugin..."
 	cd ./plugin/scheduler_example && $(MAKE) evictLeaderPlugin.so && rm evictLeaderPlugin.so
 
+generate-easyjson: install-tools
```
rleungx, line 222: How about adding it to make check?

JmPotato: It's added into `pre-build` which will be called during `make check`.

rleungx (approved):

PR #9923, opened 2025-11-12
schedule: fix tiflash hot scheduling

### LINKED ISSUE #9924: TiFlash expectation of hot scheduling is wrong
## Bug Report

Image

The expectation is lower than all TiFlash nodes.

rleungx:

Issue Number: Close #9924

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/unsaferecovery/unsafe_recovery_controller_test.go
+++ b/pkg/unsaferecovery/unsafe_recovery_controller_test.go
@@ -2013,8 +2013,8 @@ func TestSelectLeader(t *testing.T) {
 			Value: core.EngineTiFlash,
 		},
 	}
-	re.True(stores[5].IsTiFlashWrite())
 	core.SetStoreLabels(labels)(stores[5])
+	re.True(stores[5].IsTiFlashWrite())
 	for _, store := range stores {
 		cluster.PutStore(store)
 	}


pkg/statistics/store_load.go
```diff
@@ -109,8 +109,10 @@ func SummaryStoreInfos(stores []*core.StoreInfo) map[uint64]*StoreSummaryInfo {
 	infos := make(map[uint64]*StoreSummaryInfo, len(stores))
 	for _, store := range stores {
 		info := &StoreSummaryInfo{
-			StoreInfo:  store,
-			isTiFlash:  store.IsTiFlash(),
+			StoreInfo: store,
+			// Only mark TiFlash write nodes, not compute nodes.
+			// Compute nodes don't store data and shouldn't be included in hot region scheduling.
+			isTiFlash:  store.IsTiFlashWrite(),
```
lhy1024:
```suggestion
			isTiFlashWrite:  store.IsTiFlashWrite(),
```

lhy1024: Will it affect the exp of tikv?

lhy1024: Perhaps we need to modify `IsTiFlash()` to change it from a Boolean value to an enum variable for `TiKV`, `TiFlashWrite`, and `TiFlashCompute`.

rleungx: I have removed this flag and use the store label to check.

pkg/statistics/collector.go
```diff
@@ -41,6 +41,7 @@ func (tikvCollector) engine() string {
 }
 
 func (tikvCollector) filter(info *StoreSummaryInfo, kind constant.ResourceKind) bool {
+	// Exclude all TiFlash nodes (both Write and Compute)
 	if info.IsTiFlash() {
```
lhy1024: Do we need to remove `IsTiFlash()` in `StoreSummaryInfo`?

rleungx: Why?

pkg/schedule/handler/handler.go
```diff
@@ -1017,7 +1017,7 @@ func (h *Handler) GetHotStores() (*HotStoreStats, error) {
 	for _, store := range stores {
 		id := store.GetID()
 		if loads, ok := storesLoads[id]; ok {
-			if store.IsTiFlash() {
+			if store.IsTiFlashWrite() {
```
JmPotato, line 1023: Will this change affect the legacy architecture?

pkg/statistics/collector.go
lhy1024: Only concern whether need to modify for other `IsTiFlash()`

pkg/schedule/handler/handler.go
rleungx: IsTiFlashWrite returns true if the store is TiFlash write node or TiFlash classic node.

lhy1024 (approved):

Issue #9737, opened 2025-09-15
Split resource manager to a independent service


okJiang:
### Describe your feature request related problem

Currently, TiDB's Resource Control functionality is relatively mature, implementing resource isolation and traffic control through Resource Group and Request Unit (RU) mechanisms. Although the Resource Control service itself has relatively light computational load, running as an internal module of PD presents the following key issues:
1. Lack of fault isolation: Failures or exceptions in other PD services may affect the stability of Resource Control
2. Service priority conflicts: During resource constraints, it cannot guarantee that Resource Control gets sufficient computational resources
3. Cannot independently upgrade the Resource Control service
4. Limited future expansion: As Resource Control's importance increases in the next-generation architecture, it needs stronger independence and scalability

### Describe the feature you'd like

Based on the strategic importance of Resource Control in future architecture, and referencing the existing microservice experience of TSO and Scheduling, we plan to split Resource Control into an independent microservice to achieve:
1. Fault isolation: Resource Control failures don't affect other PD services, and failures in other services don't affect Resource Control
2. Backward compatibility: Maintain complete compatibility with existing Resource Control functionality, ensuring smooth migration

okJiang:
Quick progress sync and rollout plan update for reviewers.

Updated on 2026-03-23.

## Goal and Constraints
- Goal: move RM metadata writes to PD (for both classic mode and microservice mode).
- Compatibility constraint: keep `redirector.NewHandler`; do not remove it directly.
- Incremental rule: keep responsibility boundaries clear and keep follow-up fixes split when they are not required for the routing cutover.

## Current Status
- The overall migration/spec split is tracked in `WORK.md` on the `codex/rm-followup-wip` branch.
- PR1 is merged:
  - https://github.com/tikv/pd/pull/10227
  - Scope: introduce write roles (`LegacyAll / PDMetaOnly / RMTokenOnly`) and manager/keyspace write gates.
- PR2 is merged:
  - https://github.com/tikv/pd/pull/10234
  - Scope: add settings-only apply path and token bucket settings deep-copy fix.
- PR3 is merged:
  - https://github.com/tikv/pd/pull/10246
  - Scope:
    - add PD standalone metadata HTTP handler;
    - route metadata config APIs locally in PD via redirector, while keeping non-metadata routes forwarded to RM.
- PR5 is merged:
  - https://github.com/tikv/pd/pull/10255
  - Scope:
    - split client metadata/token connections;
[...]

PR #9963, opened 2025-11-28
mcs: redirect resource manager requests to resource manager service

okJiang:

Issue Number: Ref https://github.com/tikv/pd/issues/9737

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/server/resource_group_proxy_service.go
+++ b/server/resource_group_proxy_service.go
@@ -150,7 +150,7 @@ func (s *resourceGroupProxyServer) AcquireTokenBuckets(stream resource_manager.R
 		})
 	}
 	var wg sync.WaitGroup
-	wg.Add(2)
+	wg.Add(1)
 
 	// client -> server
 	go func() {
@@ -175,31 +175,24 @@ func (s *resourceGroupProxyServer) AcquireTokenBuckets(stream resource_manager.R
 	}()
 
 	// server -> client
-	go func() {
-		defer logutil.LogPanic()
-		defer wg.Done()
-		for {
-			out, err := delegateStream.Recv()
-			if err != nil {
-				if err == io.EOF {
-					// The server has finished sending messages.
-					break
-				}
-				reportErr(err)
-				return
+	for {
+		out, err := delegateStream.Recv()
+		if err != nil {
+			if err == io.EOF {
+				// The server has finished sending messages.
+				break
 			}
-			if err := stream.Send(out); err != nil {
-				reportErr(err)
-				return
-			}
+			reportErr(err)
+			break
 		}
-	}()
+		if err := stream.Send(out); err != nil {
+			reportErr(err)
+			break
+		}
+	}
 
-	go func() {
-		defer logutil.LogPanic()
-		wg.Wait()
-		reportErr(nil)
-	}()
+	wg.Wait()
+	reportErr(nil)
 
 	return <-errCh
 }


server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *resourceGroupProxyServer) DeleteResourceGroup(ctx context.Context, req *resource_manager.DeleteResourceGroupRequest) (*resource_manager.DeleteResourceGroupResponse, error) {
+	client, err := s.getResourceManagerDelegateClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return client.DeleteResourceGroup(ctx, req)
+}
+
+// AcquireTokenBuckets implements the resource_manager.ResourceManagerServer interface.
+func (s *resourceGroupProxyServer) AcquireTokenBuckets(stream resource_manager.ResourceManager_AcquireTokenBucketsServer) error {
```
lhy1024, line 131: It will be blocked when two goroutines normally exit, because it wait by `return <-errCh`

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,146 @@
...
+				return
+			}
+			if err := stream.Send(out); err != nil {
+				errCh <- err
+				return
+			}
+		}
+	}()
+
+	return <-errCh
```
lhy1024, line 197:
```suggestion
   go func() {
		wg.Wait()
		close(errCh)
	}()

	for err := range errCh {
		if err != nil {
			return err
		}
	}

	return nil
```
we should use similarly way to receive.

server/resource_group_proxy_service.go
lhy1024: We can add a test to call `stream.CloseSend()`, we expect it should close two stream, but it will block in current code.

cmd/pd-server/main.go
```diff
@@ -281,6 +282,10 @@ func start(cmd *cobra.Command, args []string, services ...string) {
 		serviceBuilders = append(serviceBuilders, swaggerserver.NewHandler)
 	}
 	serviceBuilders = append(serviceBuilders, dashboard.GetServiceBuilders()...)
+	if len(services) > 0 && !cfg.Microservice.IsResourceManagerFallbackEnabled() {
```
JmPotato, line 285: Why enable the redirector when `IsResourceManagerFallbackEnabled` is false?

okJiang: This parameter name is a bit confusing. IsResourceManagerFallbackEnabled actually refers to DisableResourceManager. However, since serverless already uses this parameter, we'll refrain from modifying it.

Fallback means fallback from ResourceManager to PD

server/resource_group_proxy_service.go
okJiang: Fixed in https://github.com/tikv/pd/pull/9963/commits/9b482267aebdc8cb658f56947a729d7dc996df00

lhy1024: Good job. Do we need a test for it?

server/config/config.go
```diff
@@ -828,6 +829,10 @@ func (c *DRAutoSyncReplicationConfig) adjust(meta *configutil.ConfigMetaData) {
 type MicroserviceConfig struct {
 	EnableSchedulingFallback  bool `toml:"enable-scheduling-fallback" json:"enable-scheduling-fallback,string"`
 	EnableTSODynamicSwitching bool `toml:"enable-tso-dynamic-switching" json:"enable-tso-dynamic-switching,string"`
+	// If EnableResourceManagerFallback is false, resource manager service only
+	// works when api service is healthy. Please make sure api service is highly
+	// available before disabling this option.
+	EnableResourceManagerFallback bool `toml:"enable-resource-manager-fallback" json:"enable-resource-manager-fallback,string"`
```
lhy1024, line 836: Do we need to prevent users from setting `enable-resource-manager-fallback = false` when `s.IsKeyspaceGroupEnabled()` is false?

tests/cluster.go
```diff
@@ -113,6 +114,9 @@ func NewTestServer(ctx context.Context, cfg *config.Config, services []string) (
 		serviceBuilders = append(serviceBuilders, swaggerserver.NewHandler)
 	}
 	serviceBuilders = append(serviceBuilders, dashboard.GetServiceBuilders()...)
+	if len(services) > 0 && !cfg.Microservice.IsResourceManagerFallbackEnabled() {
```
lhy1024: Is len(services) > 0 strictly necessary?

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/pingcap/kvproto/pkg/resource_manager"
+
+	"github.com/tikv/pd/pkg/mcs/utils/constant"
+)
+
+type resourceGroupProxyServer struct {
```
rleungx, line 32: Shall we move it to mcs?

server/config/config.go
okJiang: `s.IsKeyspaceGroupEnabled()` doesn't represent deploying resource manager service.

server/resource_group_proxy_service.go
okJiang: This server is in PD server. Not belong to mcs

server/resource_group_proxy_service.go
okJiang: I think no need.

okJiang:
ptal @JmPotato @rleungx

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,162 @@
...
+	reportErr := func(err error) {
+		reportOnce.Do(func() {
+			errCh <- err
+		})
+	}
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	// client -> server
+	go func() {
```
rleungx, line 156: need defer logutil.LogPanic()

server/server.go
```diff
@@ -299,17 +301,24 @@ func CreateServer(ctx context.Context, cfg *config.Config, services []string, le
 	}
 	// New way to register services.
 	s.registry = registry.NewServerServiceRegistry()
+	runResourceManager := cfg.Microservice.IsResourceManagerFallbackEnabled()
```
rleungx, line 305: IsResourceManagerFallbackEnabled is dynamic?

okJiang: No, it is not dynamic

rleungx: The config inside Microservice is dynamic config

cmd/pd-server/main.go
rleungx: How about adding some comments here?

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,165 @@
...
+	*GrpcServer
+}
+
+func (s *resourceGroupProxyServer) getResourceManagerDelegateClient(ctx context.Context) (resource_manager.ResourceManagerClient, error) {
+	forwardedHost, ok := s.GetServicePrimaryAddr(ctx, constant.ResourceManagerServiceName)
+	if !ok || forwardedHost == "" {
+		return nil, status.Errorf(codes.Unavailable, "resource manager service is not available")
+	}
+	if s.isLocalRequest(forwardedHost) {
+		return nil, status.Errorf(codes.Unavailable, "API mode should not support resource manager requests")
```
rleungx:
```suggestion
		return nil, status.Errorf(codes.Unavailable, "Microservice mode should not support resource manager requests")
```

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	client, err := s.getDelegateClient(ctx, forwardedHost)
```
rleungx, line 44: What if the client is stale?

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+			if err := delegateStream.Send(in); err != nil {
+				reportErr(err)
+				return
+			}
+		}
+	}()
+
+	// server -> client
+	go func() {
```
rleungx: What if one of streams breaks?

server/resource_group_proxy_service.go
okJiang: The proxy service is a temporary transitional service during upgrade and should not be in service for a long time. I think this situation can be ignored.

server/resource_group_proxy_service.go
okJiang: PD client will retry to create a new stream

server/resource_group_proxy_service.go
rleungx: no need to delete?

server/resource_group_proxy_service.go
rleungx: I mean, just one of two breaks.

server/resource_group_proxy_service.go
okJiang: Sorry, I thought you were talking about rebuilding the client when client is stale. I think it should be deleted, thank you for your suggestion

server/resource_group_proxy_service.go
okJiang: I think it will return an error to PD client and re-build a new stream

server/resource_group_proxy_service.go
okJiang: Seem we can only open a new goroutine to clean the stale client connection. How about that?

server/resource_group_proxy_service.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+			if err := stream.Send(out); err != nil {
+				reportErr(err)
+				return
+			}
+		}
+	}()
+
+	go func() {
+		wg.Wait()
```
rleungx: need logutil.LogPanic()

PR #9967, opened 2025-12-01
mcs, etcd: avoid goroutine leak when watcher is failed

### LINKED ISSUE #9966: there maybe goroutine leak when watcher is failed

lhy1024:

Issue Number: Close #9966

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/integrations/mcs/scheduling/watcher_test.go
+++ b/tests/integrations/mcs/scheduling/watcher_test.go
@@ -12,38 +12,43 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package meta
+package scheduling
 
 import (
 	"context"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


pkg/mcs/router/server/meta/watcher.go
```diff
@@ -59,6 +59,7 @@ func NewWatcher(
 	}
 	err := w.initializeStoreWatcher()
 	if err != nil {
+		w.Close()
```
rleungx, line 62: The leak test doesn't cover this issue?

lhy1024: There is no related test. I added a simple test for it. But only meta watcher.
image

lhy1024:
The new tests will fail in master, and success in this pr
image
image

PR #9978, opened 2025-12-01
api: support to set store limit for tikv engine alone

### LINKED ISSUE #9970: api: support set store limit for tikv alone
## Development Task

Image

If we met engine tikv, we can not depends on the label of the stores.

bufferflies:

Issue Number: Close #9970

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -275,6 +275,16 @@ func (h *Handler) SetLabelStoresLimit(ratePerMin float64, limitType storelimit.T
 	}
 	for _, store := range c.GetStores() {
 		for _, label := range labels {
+			// set limit for tikv stores
+			if label.Key == core.EngineKey && label.Value == core.EngineTiKV {
+				if store.IsTiKV() {
+					err = c.SetStoreLimit(store.GetID(), limitType, ratePerMin)
+					if err != nil {
+						return err
+					}
+					continue
+				}
+			}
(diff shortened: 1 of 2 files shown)


server/handler.go
```diff
@@ -275,6 +275,16 @@ func (h *Handler) SetLabelStoresLimit(ratePerMin float64, limitType storelimit.T
 	}
 	for _, store := range c.GetStores() {
 		for _, label := range labels {
+			// set limit for tikv stores
+			if label.Key == core.EngineKey && label.Value == core.EngineTiKV {
```
okJiang, line 279: engine is an implicit label, it is not actually stored in the label. Users may not even notice what is `engine label`. There are two questions.
1. Do we really need to use label as a judgment instead of adding another parameter, such as `store limit all-tikv xxxxx`
2. This should be considered a feature and requires modification of the user documentation

bufferflies: Yes, pd only uses the engine label to check if this store belongs to Tiflash; if not, it means Tikv. So we can assume that all the TikV have this engine label for the store limit operator. We should update the pd-ctl usage docs for this operator .

lhy1024 (approved):

PR #9982, opened 2025-12-02
feat(active-to-active): impl unique tso index 

bufferflies:

Issue Number: ref #https://github.com/pingcap/tidb/issues/64281

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/tso/server/config.go
+++ b/pkg/mcs/tso/server/config.go
@@ -44,6 +44,7 @@ const (
 	defaultTSOUpdatePhysicalInterval = 50 * time.Millisecond
 	maxTSOUpdatePhysicalInterval     = 10 * time.Second
 	minTSOUpdatePhysicalInterval     = 1 * time.Millisecond
+	tsoMaxIndexUpperLimit            = 10
 )
 
 var _ tso.ServiceConfig = (*Config)(nil)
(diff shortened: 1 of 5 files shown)


server/config/config.go
```diff
@@ -440,6 +446,10 @@ func (c *Config) Adjust(meta *toml.MetaData, reloading bool) error {
 		return err
 	}
 
+	if c.TSOMaxIndex != 0 && c.TSOMaxIndex <= c.TSOUniqueIndex {
```
JmPotato, line 455: Should we check the`TSOMaxIndex` as well to ensure it won't be too big to occupy too many logical bits?

bufferflies: Yes, we should limit the max index can't exceed 10

bufferflies: done

pkg/mcs/tso/server/config.go
```diff
@@ -202,6 +216,15 @@ func (c *Config) Adjust(meta *toml.MetaData) error {
...
 			zap.Duration("update-physical-interval", c.TSOUpdatePhysicalInterval.Duration))
 	}
+	if c.TSOMaxIndex > tsoMaxIndexUpperLimit {
+		return fmt.Errorf("tso max index:%d should be less than %d", c.TSOMaxIndex, tsoMaxIndexUpperLimit)
+	}
+
+	if c.TSOMaxIndex != 0 && c.TSOMaxIndex <= c.TSOUniqueIndex {
+		log.Warn("tso max index is less than unique index", zap.Int64("max-index", c.TSOMaxIndex),
+			zap.Int64("unique-index", c.TSOUniqueIndex))
+		return errors.New("tso max index is less than unique index")
```
JmPotato: Better output the corresponding values within the error message as well.

pkg/tso/tso.go
```diff
@@ -368,6 +371,7 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
 		}
 		// Get a new TSO result with the given count
 		resp.Physical, resp.Logical = t.generateTSO(ctx, int64(count))
+		resp.SuffixBits = uint32(t.maxIndex)
```
JmPotato, line 374: The usage of `SuffixBits` indicates it's a number of bits, which does not match the current implementation. Is it possible to reuse the previous Local TSO suffix mechainism?

bufferflies: I think it's more straightforward, and the case where the maxIndex is 2 is more evident.

pkg/mcs/tso/server/config.go
bufferflies: done

PR #9992, opened 2025-12-03
fix(resourcemanager): introduce group-level RU tracking by client ID

JmPotato:

Issue Number: ref #9296.

```commit-message
In the previous implementation, we did not consider the load of `requiredToken` corresponding only to
its own client and mistakenly treated it as the demand of the whole resource group, so this PR refactors
RU tracking within keyspace resource groups to support per-client RU sampling and aggregation.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -419,6 +419,20 @@ func (grt *groupRUTracker) getRUPerSec() float64 {
 	return totalRUPerSec
 }
 
+func (grt *groupRUTracker) cleanupStaleRUTrackers() []uint64 {
+	grt.Lock()
+	defer grt.Unlock()
+	staleClientUniqueIDs := make([]uint64, 0, len(grt.ruTrackers))
+	for clientUniqueID, rt := range grt.ruTrackers {
+		if time.Since(rt.lastSampleTime) < slotExpireTimeout {
+			continue
+		}
+		delete(grt.ruTrackers, clientUniqueID)
+		staleClientUniqueIDs = append(staleClientUniqueIDs, clientUniqueID)
+	}
+	return staleClientUniqueIDs
+}
+
 // conciliateFillRates is used to conciliate the fill rate of each resource group.
 // Under the service limit, there might be multiple resource groups with different
 // priorities consuming the RU at the same time. In this case, we need to conciliate
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -532,6 +532,23 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 				m.metrics.cleanupAllMetrics(r, keyspaceName)
 				m.ruCollector.remove(keyspaceName)
 			}
+			// Clean up the stale RU trackers.
+			for _, krgm := range m.getKeyspaceResourceGroupManagers() {
+				for _, group := range krgm.getResourceGroupList(false, true) {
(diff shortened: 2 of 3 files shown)


pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -153,7 +154,7 @@ var (
 			Subsystem: ruSubsystem,
 			Name:      "sampled_request_unit_per_sec",
 			Help:      "Gauge of the sampled RU/s for all resource groups.",
-		}, []string{newResourceGroupNameLabel, keyspaceNameLabel})
+		}, []string{newResourceGroupNameLabel, keyspaceNameLabel, clientIDLabel})
```
bufferflies: If the cluster has 100 TIDB nodes, the dimensions of this metric are very high. Can we bring it into the client to reduce the dimensions?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -376,6 +375,60 @@ func (rt *ruTracker) getRUPerSec() float64 {
 	return rt.lastEMA
 }
 
+// groupRUTracker is used to track the RU consumption of a resource group.
+// It matains a map of RU trackers for each client unique ID.
+type groupRUTracker struct {
+	syncutil.RWMutex
+	ruTrackers map[uint64]*ruTracker
```
lhy1024, line 382: When will we clean up this map?

pkg/mcs/resourcemanager/server/metrics.go
lhy1024: +1

JmPotato: I’ve temporarily removed the changes to this monitoring metric; let’s implement it separately later.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -532,6 +532,23 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
...
 			}
+			// Clean up the stale RU trackers.
+			for _, krgm := range m.getKeyspaceResourceGroupManagers() {
+				for _, group := range krgm.getResourceGroupList(false, true) {
+					grt := krgm.getGroupRUTracker(group.Name)
+					if grt == nil {
+						continue
+					}
+					if staleClientUniqueIDs := grt.cleanupStaleRUTrackers(); len(staleClientUniqueIDs) > 0 {
+						log.Info("cleaned up stale ru trackers",
```
lhy1024, line 543: Will there be many logs when rolling restart

JmPotato: At most, N (the number of TiDB instances) logs will be generated. I don't think it will be a problem.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -376,6 +375,64 @@ func (rt *ruTracker) getRUPerSec() float64 {
...
+}
+
+func newGroupRUTracker() *groupRUTracker {
+	return &groupRUTracker{
+		ruTrackers: make(map[uint64]*ruTracker),
+	}
+}
+
+func (grt *groupRUTracker) getOrCreateRUTracker(clientUniqueID uint64) *ruTracker {
+	grt.RLock()
```
bufferflies, line 392: for this case, I think the singleflight https://pkg.go.dev/golang.org/x/sync/singleflight is very suitable.

bufferflies (approved):

Issue #9764, opened 2025-09-22
Support data affinity schedule for performance optimization


lhy1024:
## Background

As a distributed database, TiDB inherently incurs network overhead that is not present in single-node databases. For many workloads, data from the same table or partition is frequently accessed together. The distribution of this data across different TiKV nodes can lead to significant cross-node communication for query execution, increasing latency and consuming more resources.

In our real-world load testing, we have observed that by ensuring data and indexes for the same table or partition are consolidated within a single Region, we can achieve substantial performance gains. Specifically, this co-location leads to higher QPS, lower query latency, and a notable reduction in CPU and network I/O resource consumption.

This feature will unlock key optimization opportunities, such as:
- More Read and Write Push Down
- More 1PC txn requests

## Demo Test Result

Metric| index lookup pushdown | data affinity 1PC |  Both
-- | -- | -- | --
read queries | +7.59% | +7.12% | +18.41%
write queries | +7.59% | +7.12% | +18.41%
avg latency (ms) | -7.05% | -6.64% | -15.54%
95th percentile (ms) | -5.26% | -5.26% | -14.96%

![Image](https://github.com/user-attachments/assets/162f595f-eb43-457c-ad30-0460a1762cdd)

![Image](https://github.com/user-attachments/assets/9f5e8d3f-d222-4cba-850c-bd2c98af5b6b)

## Phase I plan
- TiDB
- [ ] https://github.com/pingcap/tidb/issues/64938
- [x] https://github.com/pingcap/tidb/pull/65044
- [x] https://github.com/pingcap/tidb/pull/64988
- [x] https://github.com/pingcap/tidb/pull/65049
- TiKV
- [x] https://github.com/tikv/tikv/pull/19013
- [x] https://github.com/tikv/tikv/pull/19014
- [x] https://github.com/tikv/tikv/pull/19221
- kvproto
- [x] https://github.com/pingcap/kvproto/pull/1351
- PD
- [x] #9993 
- [x] #9997 
- [x] #9998 
- [x] #9999
- [x] #10038
- [x] #10050
- [x] #10040
- [x] #10041
- [x] #10042
- [x] #10043
- [x] #10080
- [x] #10081
- [x] #10091
- [x] #10103
### Release 8.5.5
We don't put MCS in 8.5.5 and disable the affinity schedule by default.
- TiDB
- [x] https://github.com/pingcap/tidb/pull/65101
- TiKV
- [x] https://github.com/tikv/tikv/pull/19212
[...]

PR #9993, opened 2025-12-03
affinity: add affinity storage, new config, label plan and error code

lhy1024:

Issue Number:  Ref https://github.com/tikv/pd/issues/9764

1.  EnableAffinityScheduling
- control check and auto split
- api check it, return error when it is false
- default: false
2. MaxAffinityMergeRegionSize
- decouple with MaxMergeRegionSize
- only check MaxMergeRegionSize while affinity merge check
- default: 256

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/grpc_service.go
+++ b/pkg/mcs/resourcemanager/server/grpc_service.go
@@ -231,7 +231,7 @@ func (s *Service) AcquireTokenBuckets(stream rmpb.ResourceManager_AcquireTokenBu
 						tokens = rg.RequestRU(now, requiredToken, targetPeriodMs, clientUniqueID, krgm.getServiceLimiter())
 					}
 					// Sample the latest RU demand.
-					krgm.getOrCreateRUTracker(rg.Name).sample(now, requiredToken)
+					krgm.getOrCreateGroupRUTracker(rg.Name).sample(clientUniqueID, now, requiredToken)
 					if tokens == nil {
 						continue
 					}
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -65,9 +65,9 @@ type consumptionItem struct {
 
 type keyspaceResourceGroupManager struct {
 	syncutil.RWMutex
-	groups     map[string]*ResourceGroup
-	ruTrackers map[string]*ruTracker
-	sl         *serviceLimiter
+	groups          map[string]*ResourceGroup
+	groupRUTrackers map[string]*groupRUTracker
+	sl              *serviceLimiter
 
 	keyspaceID uint32
 	storage    endpoint.ResourceGroupStorage
@@ -75,11 +75,11 @@ type keyspaceResourceGroupManager struct {
 
 func newKeyspaceResourceGroupManager(keyspaceID uint32, storage endpoint.ResourceGroupStorage) *keyspaceResourceGroupManager {
 	return &keyspaceResourceGroupManager{
-		groups:     make(map[string]*ResourceGroup),
-		ruTrackers: make(map[string]*ruTracker),
-		keyspaceID: keyspaceID,
-		storage:    storage,
-		sl:         newServiceLimiter(keyspaceID, 0, storage),
+		groups:          make(map[string]*ResourceGroup),
+		groupRUTrackers: make(map[string]*groupRUTracker),
+		keyspaceID:      keyspaceID,
+		storage:         storage,
+		sl:              newServiceLimiter(keyspaceID, 0, storage),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


pkg/schedule/config/config.go
```diff
@@ -65,6 +65,7 @@ const (
 	defaultEnablePlacementRules            = true
 	defaultEnableWitness                   = false
 	defaultHaltScheduling                  = false
+	defaultEnableAffinityScheduling        = true
```
bufferflies: In most cases, the experiment feature is false.

server/config/persist_options.go
```diff
@@ -342,6 +357,13 @@ func (o *PersistOptions) SetMaxMergeRegionSize(maxMergeRegionSize uint64) {
 	o.SetScheduleConfig(v)
 }
 
+// SetMaxAffinityMergeRegionSize sets the max affinity merge region size.
+func (o *PersistOptions) SetMaxAffinityMergeRegionSize(maxAffinityMergeRegionSize uint64) {
```
bufferflies, line 356: Can we set this value smaller than the max-merge-size?

pkg/schedule/config/config.go
```diff
@@ -472,7 +492,17 @@ func (c *ScheduleConfig) GetMaxMergeRegionKeys() uint64 {
...
+}
+
+// GetMaxAffinityMergeRegionSize returns the max affinity merge region size.
+// It returns 0 if the MaxMergeRegionSize is 0 or the MaxAffinityMergeRegionSize is 0.
+// It returns the greater one between the MaxMergeRegionSize and the MaxAffinityMergeRegionSize.
+func (c *ScheduleConfig) GetMaxAffinityMergeRegionSize() uint64 {
+	if c.MaxMergeRegionSize == 0 || c.MaxAffinityMergeRegionSize == 0 {
+		return 0
+	}
+	return max(c.MaxAffinityMergeRegionSize, c.MaxMergeRegionSize)
```
bufferflies: I think it should be decoupled from the maximum merge region size. It can confuse the user about which configuration will work.

lhy1024: These configs are related.

pkg/schedule/config/config.go
HunDunDM: If no affinity group is set, nothing will be done, even if it is true.

server/config/persist_options.go
lhy1024: I think we still need `max(c.MaxAffinityMergeRegionSize, c.MaxMergeRegionSize)`. If set c.MaxMergeRegionSize to 100, then set MaxAffinityMergeRegionSize to 10, then set c.MaxMergeRegionSize to 1.

pkg/schedule/config/config.go
HunDunDM: How about keeping only the special case of 0?

lhy1024: @HunDunDM  only consider close merge checker by setting MaxMergeRegionSize to zero?

server/config/persist_options.go
bufferflies: We can forbid this illegal value by keeping the MaxAffinityMergeRegionSize larger than the MaxMergeRegionSize.

lhy1024:
I plan to adjust the config about affinity:

1.  EnableAffinityScheduling
- control check and auto split
- api check it, return error when it is false
- default: false
2. MaxAffinityMergeRegionSize
- decouple with MaxMergeRegionSize
- only check MaxMergeRegionSize while affinity merge check
- default: 256

What do you think? cc @bufferflies @HunDunDM @rleungx

pkg/mcs/scheduling/server/config/config.go
```diff
@@ -563,6 +568,24 @@ func (o *PersistConfig) GetMaxMergeRegionSize() uint64 {
...
+	}
+	affinitySize := o.GetScheduleConfig().GetMaxAffinityMergeRegionSize()
+	if affinitySize == 0 {
+		return 0
+	}
+	if size > affinitySize {
+		return size
+	}
+	return affinitySize
+}
```
okJiang: return math.Max()

pkg/schedule/config/config.go
```diff
@@ -80,6 +81,14 @@ const (
 	defaultHotRegionsWriteInterval = 10 * time.Minute
 	// It means we skip the preparing stage after the 48 hours no matter if the store has finished preparing stage.
 	defaultMaxStorePreparingTime = 48 * time.Hour
+
+	// defaultMaxAffinityMergeRegionSize is the default maximum size of affinity region when regions can be merged.
+	// It means 256 MB, according https://docs.pingcap.com/tidb/stable/tune-region-performance/#use-region-split-size-to-adjust-region-
+	// The maximum size of region maybe be larger than 256 MB, such as 512 MB or 1 GB.
+	defaultMaxAffinityMergeRegionSize = 256
```
okJiang, line 89: Why is defaultMaxAffinityMergeRegionSize 256 but defaultMaxMergeRegionSize 54? Can you add some comments?

pkg/schedule/config/config.go
```diff
@@ -314,6 +323,12 @@ type ScheduleConfig struct {
 
 	// PatrolRegionWorkerCount is the number of workers to patrol region.
 	PatrolRegionWorkerCount int `toml:"patrol-region-worker-count" json:"patrol-region-worker-count"`
+
+	// EnableAffinityScheduling is the option to enable affinity scheduling.
+	EnableAffinityScheduling bool `toml:"enable-affinity-scheduling" json:"enable-affinity-scheduling,string,omitempty"`
+	// If the size of region is smaller than MaxAffinityMergeRegionSize,
```
okJiang, line 330: Should the affinity's merge consider the key count?

pkg/slice/slice.go
```diff
@@ -71,3 +72,13 @@ func EqualWithoutOrder[T comparable](a, b []T) bool {
 	}
 	return true
 }
+
+// HasDupSorted takes a sorted slice and checks whether it contains any duplicate elements.
+func HasDupSorted[T comparable](slice []T) bool {
```
okJiang:
```suggestion
func HasDupInSortedSlice[T comparable](sortedSlice []T) bool {
```

pkg/schedule/labeler/plan.go
```diff
@@ -0,0 +1,97 @@
...
+
+package labeler
+
+import (
+	"github.com/tikv/pd/pkg/storage/kv"
+)
+
+// Plan is an execution plan for the RegionLabeler.
+// It is used to perform transactional commits together with other modules.
+type Plan struct {
```
okJiang, line 23: How about adding this structure in next pr using it.

pkg/mcs/scheduling/server/config/config.go
lhy1024: ok

pkg/schedule/config/config.go
lhy1024: Because we will merge affinity region as far as possible

pkg/schedule/config/config.go
lhy1024: Yes, it will calculate by MaxAffinityMergeRegionSize 
image

pkg/slice/slice.go
lhy1024: we will use `slice.HasDupSorted`

pkg/schedule/labeler/plan.go
lhy1024: next pr has been too huge

pkg/schedule/config/config.go
lhy1024: I added it.

pkg/schedule/config/config.go
okJiang: also can add a comment here

pkg/slice/slice.go
okJiang: I think only `Sorted` here is misunderstanding. For example, Has duplicated sorted?

pkg/schedule/config/config.go
lhy1024: done

pkg/slice/slice.go
lhy1024: How about `func HasDupInSorted[T comparable](sortedSlice []T) bool {`

okJiang (approved):

lhy1024:
PTAL @niubell

pkg/schedule/config/config.go
```diff
@@ -80,6 +81,15 @@ const (
 	defaultHotRegionsWriteInterval = 10 * time.Minute
 	// It means we skip the preparing stage after the 48 hours no matter if the store has finished preparing stage.
 	defaultMaxStorePreparingTime = 48 * time.Hour
+
+	// defaultMaxAffinityMergeRegionSize is the default maximum size of affinity region when regions can be merged.
+	// It means 256 MB, according https://docs.pingcap.com/tidb/stable/tune-region-performance/#use-region-split-size-to-adjust-region-
+	// The maximum size of region maybe be larger than 256 MB, such as 512 MB or 1 GB.
+	// Because we will merge affinity region as far as possible, so defaultMaxAffinityMergeRegionSize is larger than defaultMaxMergeRegionSize.
+	defaultMaxAffinityMergeRegionSize = 256
```
niubell, line 89: What's the merge behaviour of a table whose size larger than defaultMaxAffinityMergeRegionSize?

lhy1024: The affinity checker will schedule them to the related store and merge them as much as possible up to the limit.

For example, if there are four regions with sizes of 127, 128, 256, and 512 MB, the result will be three regions of sizes 255, 256, and 512 MB. (127 and 128 are merged; 256 and 512 remain separate as they reach or exceed the limit).

We set defaultMaxAffinityMergeRegionSize to 256 MB to avoid affecting TiKV performance. If in POC, we can try to adjust to a larger value.

PR #9997, opened 2025-12-04
affinity: add base models and manager skeleton

lhy1024:

  - Add base affinity models/state: group.go introduces Group/GroupState, availability/degraded/expired conditions, ID validation, and region affinity checks.
  - Manager skeleton: manager.go adds Manager with dual locking (metaMutex + RWMutex)

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/affinity/affinity_test.go
+++ b/pkg/schedule/affinity/affinity_test.go
@@ -34,7 +34,7 @@ func TestMain(m *testing.M) {
 }
 
 //nolint:unused
-func (m *Manager) getGroupForTest(re *require.Assertions, id string) *runtimeGroupInfo {
+func getGroupForTest(re *require.Assertions, m *Manager, id string) *runtimeGroupInfo {
 	m.RLock()
 	defer m.RUnlock()
 	group, ok := m.groups[id]
@@ -46,7 +46,7 @@ func (m *Manager) getGroupForTest(re *require.Assertions, id string) *runtimeGro
 }
 
 //nolint:unused
-func (m *Manager) createGroupForTest(re *require.Assertions, id string, rangeCount int) []keyutil.KeyRange {
+func createGroupForTest(re *require.Assertions, m *Manager, id string, rangeCount int) []keyutil.KeyRange {
 	gkr := GroupKeyRanges{
 		KeyRanges: make([]keyutil.KeyRange, rangeCount),
 		GroupID:   id,
@@ -62,7 +62,7 @@ func (m *Manager) createGroupForTest(re *require.Assertions, id string, rangeCou
 }
 
 //nolint:unused
-func (m *Manager) testCacheStale(re *require.Assertions, region *core.RegionInfo) {
+func testCacheStale(re *require.Assertions, m *Manager, region *core.RegionInfo) {
 	cache, group := m.getCache(region)
 	if cache != nil && group != nil {
 		re.NotEqual(cache.affinityVer, group.affinityVer)
--- a/pkg/schedule/affinity/group.go
+++ b/pkg/schedule/affinity/group.go
@@ -28,95 +28,66 @@ import (
 	"github.com/tikv/pd/pkg/utils/keyutil"
 )
 
-// condition is an enumeration that includes both Store state and Group state.
-type condition int
+// groupAvailability is an enum that represents the Group’s availability lifecycle.
+// groupAvailable, groupDegraded, and groupExpired define the Group’s availability lifecycle.
+// The groupDegraded status should have an expiration time.
+// Roughly:
+//
+//	groupAvailable ──degraded (e.g. store evict-leader)───────────────> groupDegraded
+//	groupDegraded  ──recovered────────────────────────────────────────> groupAvailable // expected to be temporary
+//	groupDegraded  ──expired──────────────────────────────────────────> groupExpired
+//	groupAvailable ──directly failed (e.g. store removed)─────────────> groupExpired
+//	groupExpired   ──reconfigured (e.g. peers moved to healthy stores)→ groupAvailable
+//
+// groupDegraded is intended to be a temporary status that may return to groupAvailable,
+// while groupExpired usually represents a terminal status under the current topology,
+// but can become groupAvailable again after the Group’s stores/peers are reconfigured.
+// groupDegraded has an expiration time (degradedExpiredAt). Once it expires, the Group is
+// automatically treated as groupExpired.
+type groupAvailability int
 
 const (
-	// groupAvailable indicates that the current Group allows affinity scheduling and disallows other balancing scheduling.
-	groupAvailable condition = iota
-
-	// groupDegraded indicates that the current Group does not generate affinity scheduling but still disallows other balancing scheduling.
-	// All values greater than groupAvailable and less than or equal to groupDegraded represent groupDegraded states.
-	// The groupDegraded state should have an expiration time. After it expires, it should be treated as groupExpired.
-	storeEvictLeader
-	storeDisconnected
-	storePreparing
-	storeLowSpace
+	groupAvailable groupAvailability = iota
 	groupDegraded
-
-	// groupExpired indicates that the current Group does not generate affinity scheduling and allows other balancing scheduling.
-	// All values greater than groupDegraded and less than or equal to groupExpired represent groupExpired states.
-	storeDown
-	storeRemovingOrRemoved
 	groupExpired
-
-	// groupAvailable, groupDegraded, and groupExpired define the Group’s availability lifecycle.
-	// Roughly:
-	//   groupAvailable ──degraded (e.g. store evict-leader)───────────────> groupDegraded
-	//   groupDegraded  ──recovered────────────────────────────────────────> groupAvailable // expected to be temporary
-	//   groupDegraded  ──expired──────────────────────────────────────────> groupExpired
-	//   groupAvailable ──directly failed (e.g. store removed)─────────────> groupExpired
-	//   groupExpired   ──reconfigured (e.g. peers moved to healthy stores)→ groupAvailable
-	// groupDegraded is intended to be a temporary state that may return to groupAvailable,
-	// while groupExpired usually represents a terminal state under the current topology,
-	// but can become groupAvailable again after the Group’s stores/peers are reconfigured.
-	// groupDegraded has an expiration time (degradedExpiredAt); once it expires, the Group is
-	// automatically treated as groupExpired.
 )
 
-// Phase is a status intended for API display
-type Phase string
-
-const (
-	// PhasePending indicates that the Group is still determining the StoreIDs.
-	// If the Group has no KeyRanges, it remains in PhasePending forever.
-	PhasePending = Phase("pending")
-	// PhasePreparing indicates that the Group is scheduling Regions according to the required Peers.
-	PhasePreparing = Phase("preparing")
-	// PhaseStable indicates that the Group has completed the required scheduling and is currently in a stable state.
-	PhaseStable = Phase("stable")
-)
-
-// idPattern is a regex that specifies acceptable characters of the id.
-// Valid id must be non-empty and 64 characters or fewer and consist only of letters (a-z, A-Z),
-// numbers (0-9), hyphens (-), and underscores (_).
-const idPattern = "^[-A-Za-z0-9_]{1,64}$"
-
-var idRegexp = regexp.MustCompile(idPattern)
-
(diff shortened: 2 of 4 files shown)


lhy1024:
@bufferflies @okJiang PTAL

pkg/schedule/affinity/stub_affinity.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+//nolint:revive
+func (m *Manager) DeleteAffinityGroups([]string, bool) error {
+	_ = m
+	return nil
+}
+
+//nolint:revive
+func (m *Manager) UpdateAffinityGroupPeers(string, uint64, []uint64) (*GroupState, error) {
```
bufferflies, line 69:
```suggestion
func (*Manager) UpdateAffinityGroupPeers(string, uint64, []uint64) (*GroupState, error) {
```

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,380 @@
...
+	if g.LeaderStoreID == 0 || len(g.VoterStoreIDs) == 0 {
+		return errs.ErrAffinityGroupContent.FastGenByArgs("leader store ID and voter store IDs must be provided together")
+	}
+
+	voterStoreIDs := slices.Clone(g.VoterStoreIDs)
+	slices.Sort(voterStoreIDs)
+	if slice.HasDupInSorted(voterStoreIDs) {
+		return errs.ErrAffinityGroupContent.FastGenByArgs("duplicate voter store ID")
+	}
+	if !slices.Contains(voterStoreIDs, g.LeaderStoreID) {
```
bufferflies, line 354: In generative speaking, the leader should belong to the voters. How about renaming voterStoreIDs as followerStoreIDs if you don't allow the leader to exist in this list.

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,380 @@
...
+		return errs.ErrAffinityGroupContent.FastGenByArgs("leader must be in voter stores")
+	}
+	for _, storeID := range voterStoreIDs {
+		if m.storeSetInformer.GetStore(storeID) == nil {
+			return errs.ErrAffinityGroupContent.FastGenByArgs("store does not exist")
```
bufferflies: How about informing the user that the store was not found?

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+
+	return nil
+}
+
+// GroupKeyRanges represents key ranges with group id.
+type GroupKeyRanges struct {
+	KeyRanges []keyutil.KeyRange
+	GroupID   string
```
bufferflies, line 369: How about adding more to the format of this groupID?

pkg/schedule/affinity/group_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	conf := mockconfig.NewTestOptions()
+	regionLabeler, err := labeler.NewRegionLabeler(ctx, store, time.Second*5)
+	re.NoError(err)
+	manager, err := NewManager(ctx, store, storeInfos, conf, regionLabeler)
+	re.NoError(err)
+
+	group := &Group{
+		ID:            "TEST-group_1",
```
bufferflies, line 89: affinity_group_id = _tidb_p_t_

pkg/schedule/affinity/group.go
lhy1024: Yes, we now only return error when LeaderStoreID is not in voterStoreIDs

pkg/schedule/affinity/stub_affinity.go
lhy1024: ok, but it will be used really in another pr later.

pkg/schedule/affinity/group.go
lhy1024: ok

pkg/schedule/affinity/group_test.go
lhy1024: it only for test

pkg/schedule/affinity/group.go
lhy1024: Or we add it in pd-ctl about group format, I have added it in pd-ctl. We don't limit the format in server, it should be a convention for pd-ctl and SQL.

lhy1024: The region label also does not conform to the server's specified format.

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,381 @@
...
+
+// idPattern is a regex that specifies acceptable characters of the id.
+// Valid id must be non-empty and 64 characters or fewer and consist only of letters (a-z, A-Z),
+// numbers (0-9), hyphens (-), and underscores (_).
+const idPattern = "^[-A-Za-z0-9_]{1,64}$"
+
+var idRegexp = regexp.MustCompile(idPattern)
+
+// toGroupState converts the condition into the corresponding Group state.
+func (s condition) toGroupState() condition {
```
okJiang:
```suggestion
func (s condition) groupState() condition {
```

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,381 @@
...
+	//   groupDegraded  ──recovered────────────────────────────────────────> groupAvailable // expected to be temporary
+	//   groupDegraded  ──expired──────────────────────────────────────────> groupExpired
+	//   groupAvailable ──directly failed (e.g. store removed)─────────────> groupExpired
+	//   groupExpired   ──reconfigured (e.g. peers moved to healthy stores)→ groupAvailable
+	// groupDegraded is intended to be a temporary state that may return to groupAvailable,
+	// while groupExpired usually represents a terminal state under the current topology,
+	// but can become groupAvailable again after the Group’s stores/peers are reconfigured.
+	// groupDegraded has an expiration time (degradedExpiredAt); once it expires, the Group is
+	// automatically treated as groupExpired.
+)
```
okJiang: Why not use a structure(contain storeState and groupState) instead of one type?

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,381 @@
...
+
+// Phase is a status intended for API display
+type Phase string
```
okJiang: Moving `Phase` and `condition` to a new file is better for me.

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func (g *Group) String() string {
+	b, _ := json.Marshal(g)
+	return string(b)
+}
+
+// GroupState defines the runtime state of an affinity group.
+// NOTE: This type is exported by HTTP API. Please pay more attention when modifying it.
+type GroupState struct {
```
okJiang, line 169: `GroupState` is duplicated with `toGroupState`. We need a better and different naming.

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,381 @@
...
+	// state should use the condition enum values whose names start with group.
+	// state must be used together with degradedExpiredAt. Therefore, GetState and SetState must always be used to
+	// read or modify them, and state must not be accessed directly.
+	state condition
+	// degradedExpiredAt indicates the expiration time of groupDegraded. After this time, it should be treated as groupExpired.
+	degradedExpiredAt uint64
+	// AffinityVer initializes at 1 and increments by 1 each time the Group changes.
+	AffinityVer uint64
+	// AffinityRegionCount indicates how many Regions have all Voter and Leader peers in the correct stores. (AffinityVer equals).
+	AffinityRegionCount int
```
okJiang, line 226: This is better.
```suggestion
	AffinitizedRegionCount int
```

okJiang, line 226: Why exist two same fields? L171-L173

pkg/schedule/affinity/affinity_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/tikv/pd/pkg/utils/keyutil"
+	"github.com/tikv/pd/pkg/utils/testutil"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m, testutil.LeakOptions...)
+}
+
+//nolint:unused
+func (m *Manager) getGroupForTest(re *require.Assertions, id string) *runtimeGroupInfo {
```
okJiang: This is better?

```suggestion
func getGroupForTest(m *Manager, re *require.Assertions, id string) *runtimeGroupInfo {
```

pkg/schedule/affinity/affinity_test.go
```diff
@@ -0,0 +1,96 @@
...
+	group, ok := m.groups[id]
+	re.True(ok)
+	// Return a cloned runtimeGroupInfo to ensure lock safety.
+	newGroup := *group
+	newGroup.Regions = maps.Clone(group.Regions)
+	return &newGroup
+}
+
+//nolint:unused
+func (m *Manager) createGroupForTest(re *require.Assertions, id string, rangeCount int) []keyutil.KeyRange {
```
okJiang: ditto

pkg/schedule/affinity/affinity_test.go
```diff
@@ -0,0 +1,96 @@
...
+			StartKey: []byte(fmt.Sprintf("test-%s-%04d", id, i)),
+			EndKey:   []byte(fmt.Sprintf("test-%s-%04d", id, i+1)),
+		}
+	}
+	re.NoError(m.CreateAffinityGroups([]GroupKeyRanges{gkr}))
+	return gkr.KeyRanges
+}
+
+//nolint:unused
+func (m *Manager) testCacheStale(re *require.Assertions, region *core.RegionInfo) {
```
okJiang: ditto

pkg/schedule/affinity/affinity_test.go
okJiang, line 1: How about renaming to `test_utils.go`

pkg/schedule/affinity/group.go
lhy1024: runtimeGroupInfo is struct for memory, GroupState is struct for api handler

pkg/schedule/affinity/affinity_test.go
HunDunDM: Files ending with `_test` will only be compiled during test.

pkg/schedule/affinity/group.go
lhy1024: affinity is also adj.

pkg/schedule/affinity/group.go
lhy1024: I think it is not necessary for now. It will cause problems for subsequent code merges. Once all code has been merged, we can consider making further adjustments if necessary.

pkg/schedule/affinity/group.go
lhy1024: Good

pkg/schedule/affinity/affinity_test.go
lhy1024: yes

pkg/schedule/affinity/group.go
okJiang: Why was it designed this way?

lhy1024:
@niubell PTAL

lhy1024:
> The first PR to add the basic structures, will add the corresponding tests in next PRs.

yes, they will be added in 9998 and 9999

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,378 @@
...
+	case storeDisconnected:
+		return "disconnected"
+	case storePreparing:
+		return "preparing"
+	case storeLowSpace:
+		return "low-space"
+	case storeDown:
+		return "down"
+	case storeRemovingOrRemoved:
+		return "removing-or-removed"
```
rleungx, line 102: prefer to split them

pkg/schedule/affinity/group.go
```diff
@@ -0,0 +1,378 @@
...
+		g.availability = newAvailability
+	}
+}
+
+func newDegradedExpiredAtFromNow() uint64 {
+	return uint64(time.Now().Unix()) + defaultDegradedExpirationSeconds
+}
+
+// IsAffinitySchedulingEnabled indicates whether affinity scheduling is allowed.
+func (g *runtimeGroupInfo) IsAffinitySchedulingEnabled() bool {
```
rleungx, line 326: The name is the same as IsAffinitySchedulingEnabled in the config, but the meaning is different.

PR #9998, opened 2025-12-04
affinity: add txn and key range operations

lhy1024:

Issue Number: Ref  https://github.com/tikv/pd/issues/9764 

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/affinity/txn.go
+++ b/pkg/schedule/affinity/txn.go
@@ -17,6 +17,8 @@ package affinity
 import (
 	"bytes"
 	"encoding/hex"
+	"fmt"
+	"maps"
 	"slices"
 	"strings"
 	"time"
@@ -115,7 +117,7 @@ func (m *Manager) CreateAffinityGroups(changes []GroupKeyRanges) error {
 	}
 
 	// Step 2: Convert and validate key ranges no overlaps
-	if err := m.validateNoKeyRangeOverlap(changes); err != nil {
+	if err := m.validateNoOverlapLocked(changes); err != nil {
 		return err
 	}
 
@@ -360,7 +362,6 @@ func (m *Manager) UpdateAffinityGroupKeyRanges(addOps, removeOps []GroupKeyRange
 	defer m.metaMutex.Unlock()
 
 	// Step 1: Validate the added KeyRanges.
-	var newRangesToValidate []GroupKeyRanges
 	for _, op := range addOps {
 		currentRanges, err := m.getCurrentRanges(op.GroupID)
 		if err != nil {
@@ -372,16 +373,10 @@ func (m *Manager) UpdateAffinityGroupKeyRanges(addOps, removeOps []GroupKeyRange
 			GroupID:   op.GroupID,
 			KeyRanges: slices.Concat(currentRanges.KeyRanges, op.KeyRanges),
 		}
-
-		// Collect new ranges for overlap validation
-		newRangesToValidate = append(newRangesToValidate, GroupKeyRanges{
-			KeyRanges: op.KeyRanges,
-			GroupID:   op.GroupID,
-		})
 	}
 	// Validate no overlaps with newly added ranges
-	if len(newRangesToValidate) > 0 {
-		if err := m.validateNoKeyRangeOverlap(newRangesToValidate); err != nil {
+	if len(addOps) > 0 {
+		if err := m.validateNoOverlapLocked(addOps); err != nil {
 			return err
 		}
 	}
@@ -488,12 +483,7 @@ func (m *Manager) getCurrentRanges(groupID string) (GroupKeyRanges, error) {
 		return GroupKeyRanges{}, errors.Errorf("label rule not found for group %s", groupID)
 	}
 
-	dataSlice, ok := labelRule.Data.([]*labeler.KeyRangeRule)
-	if !ok {
-		return GroupKeyRanges{}, errors.Errorf("invalid label rule data type for group %s, got type %T", groupID, labelRule.Data)
-	}
-
-	return parseKeyRangesFromData(dataSlice, groupID)
+	return extractKeyRangesFromLabelRule(labelRule)
 }
 
 // applyRemoveOps filters out ranges that match remove operations.
@@ -583,6 +573,8 @@ func decodeHexKey(hexStr, groupID, keyType string) ([]byte, error) {
 }
 
 // extractKeyRangesFromLabelRule extracts key ranges from a label rule data.
+// This function expects the rule to have already been processed by checkAndAdjust(),
+// which guarantees that rule.Data is of type []*labeler.KeyRangeRule.
 func extractKeyRangesFromLabelRule(rule *labeler.LabelRule) (GroupKeyRanges, error) {
 	if rule == nil || rule.Data == nil {
 		return GroupKeyRanges{}, nil
@@ -593,33 +585,13 @@ func extractKeyRangesFromLabelRule(rule *labeler.LabelRule) (GroupKeyRanges, err
 		return GroupKeyRanges{}, nil
 	}
 
-	switch data := rule.Data.(type) {
-	case []*labeler.KeyRangeRule:
-		return parseKeyRangesFromData(data, groupID)
-	case []any:
-		converted := make([]*labeler.KeyRangeRule, 0, len(data))
-		for _, item := range data {
-			entry, ok := item.(map[string]any)
-			if !ok {
-				return GroupKeyRanges{}, errs.ErrInvalidLabelRuleFormat.FastGenByArgs("is not a map")
-			}
-			startKey, ok := entry["start_key"].(string)
-			if !ok {
-				return GroupKeyRanges{}, errs.ErrInvalidLabelRuleFormat.FastGenByArgs("missing or invalid 'start_key'")
-			}
-			endKey, ok := entry["end_key"].(string)
-			if !ok {
-				return GroupKeyRanges{}, errs.ErrInvalidLabelRuleFormat.FastGenByArgs("missing or invalid 'end_key'")
-			}
-			converted = append(converted, &labeler.KeyRangeRule{
-				StartKeyHex: startKey,
-				EndKeyHex:   endKey,
-			})
-		}
-		return parseKeyRangesFromData(converted, groupID)
-	default:
-		return GroupKeyRanges{}, errs.ErrAffinityGroupContent.FastGenByArgs("invalid label rule data format")
+	data, ok := rule.Data.([]*labeler.KeyRangeRule)
+	if !ok {
+		return GroupKeyRanges{}, errs.ErrAffinityGroupContent.FastGenByArgs(
+			fmt.Sprintf("invalid label rule data type for group %s, got type %T", groupID, rule.Data))
 	}
+
+	return parseKeyRangesFromData(data, groupID)
 }
 
 // checkKeyRangesOverlap checks if two key ranges overlap.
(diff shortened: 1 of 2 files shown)


pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+	m.metaMutex.Lock()
+	defer m.metaMutex.Unlock()
+
+	// Step 1: Check whether the Group exists.
+	if err := m.noGroupsExist(groups); err != nil {
+		return err
+	}
+
+	// Step 2: Convert and validate key ranges no overlaps
+	if err := m.validateNoKeyRangeOverlap(changes); err != nil {
```
bufferflies: this function should rename xxxLocked.

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		})
+	}
+	return GroupKeyRanges{
+		KeyRanges: keyRanges,
+		GroupID:   groupID,
+	}, nil
+}
+
+// decodeHexKey decodes a hex string and returns an error if decoding fails.
+func decodeHexKey(hexStr, groupID, keyType string) ([]byte, error) {
```
bufferflies, line 564: how about move this into pkg/utils？

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+	}
+	decoded, err := hex.DecodeString(hexStr)
+	if err != nil {
+		return nil, errs.ErrInvalidKeyFormat.FastGenByArgs(keyType, hexStr, groupID)
+	}
+	return decoded, nil
+}
+
+// extractKeyRangesFromLabelRule extracts key ranges from a label rule data.
+func extractKeyRangesFromLabelRule(rule *labeler.LabelRule) (GroupKeyRanges, error) {
```
bufferflies, line 578: I think it belongs to the labeler.LabelRule function.

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+	if rule == nil || rule.Data == nil {
+		return GroupKeyRanges{}, nil
+	}
+
+	groupID, ok := parseAffinityGroupIDFromLabelRule(rule)
+	if !ok {
+		return GroupKeyRanges{}, nil
+	}
+
+	switch data := rule.Data.(type) {
```
bufferflies: I don't think the type of the rule.Data is any? Can it be the fixed type such as rule?

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		}
+		return parseKeyRangesFromData(converted, groupID)
+	default:
+		return GroupKeyRanges{}, errs.ErrAffinityGroupContent.FastGenByArgs("invalid label rule data format")
+	}
+}
+
+// checkKeyRangesOverlap checks if two key ranges overlap.
+// Returns true if [start1, end1) and [start2, end2) have any overlap.
+func checkKeyRangesOverlap(start1, end1, start2, end2 []byte) bool {
```
bufferflies, line 599: https://github.com/tikv/pd/blob/b88e47d916de884a59f8f063a3bd951e54834838/pkg/utils/keyutil/keyrange.go#L34-L46 Can it help?

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+				zap.String("rule-id", rule.ID),
+				zap.String("group-id", groupID),
+				zap.Error(err))
+			return true
+		}
+
+		if len(gkr.KeyRanges) > 0 {
+			allRanges = append(allRanges, gkr)
+			// Populate in-memory key ranges cache
+			m.keyRanges[groupID] = gkr
```
bufferflies, line 714: If the first update in-memory, the second not found  and returned, the m.keyRanges is stale

pkg/schedule/affinity/txn.go
HunDunDM: It only contains `DecodeString` and an `err` object with a `GroupID`, so it doesn't need to be used as a utils. Currently, other modules also use `hex.DecodeString` to generate errors.

pkg/schedule/affinity/txn.go
HunDunDM: This requires processing the keyrange while simultaneously checking which group it belongs to, making it inconvenient to place it on the labeler side.

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+				GroupID:   change.GroupID,
+			}
+		} else {
+			// No key ranges, remove from cache if exists
+			delete(m.keyRanges, change.GroupID)
+		}
+		log.Info("affinity group added", zap.String("group", groups[i].String()))
+	}
+
+	m.createGroups(groups, labelRules)
```
okJiang, line 175: if len(change.KeyRanges) == 0, len(groups) does not equal to len(labelRules)

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+// If force is true:
+//   - Skips non-existent groups
+//   - Deletes groups even if they have key ranges
+func (m *Manager) DeleteAffinityGroups(groupIDs []string, force bool) error {
+	if len(groupIDs) == 0 {
+		return errs.ErrAffinityGroupContent.FastGenByArgs("no group ids provided")
+	}
+
+	seen := make(map[string]struct{}, len(groupIDs))
+	toDelete := make([]string, 0, len(groupIDs))
```
okJiang, line 193: Please use meaningful name

pkg/schedule/affinity/txn.go
HunDunDM: It is consistent with the `MakeKeyRanges` design of keyspace.

pkg/schedule/affinity/txn.go
HunDunDM: OverLapped return true if the two KeyRanges are continuous

pkg/schedule/affinity/txn.go
lhy1024: L95-L107 means len(groups) is equal to len(changes)
L123 means len(labelRules) is equal to len(changes)

pkg/schedule/affinity/txn.go
lhy1024: replace them with `deduplicatedGroupIDs` and `groupIDsToDelete`?

pkg/schedule/affinity/txn.go
lhy1024: if len(change.KeyRanges) == 0, labelRules[i] will be nil

pkg/schedule/affinity/txn.go
lhy1024: ok

pkg/schedule/affinity/txn.go
HunDunDM: Both `labelRules` and `groups` have a length of `len(changes)`. `LabelRules` are accessed by index in the for loop.

okJiang: Got it

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		// RegularSchedulingEnabled == IsExpired
+		// The VoterStoreIDs are already sorted, so they can be compared directly
+		if !group.RegularSchedulingEnabled && !slices.Equal(voterStoreIDs, group.VoterStoreIDs) {
+			return group, nil
+		}
+	}
+
+	// Step 2: Save the Group in storage.
+	group.LeaderStoreID = leaderStoreID
+	group.VoterStoreIDs = slices.Clone(voterStoreIDs)
```
bufferflies, line 315: Does the voterStoreIDs must be sorted？

lhy1024: L295 sort group.VoterStoreIDs

pkg/schedule/affinity/txn.go
okJiang: It seems there is no need to modify it, let's keep it as it is. Sry for this comment😂

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		}
+		log.Info("affinity group added", zap.String("group", groups[i].String()))
+	}
+
+	m.createGroups(groups, labelRules)
+	return nil
+}
+
+// DeleteAffinityGroups deletes multiple affinity groups in a single transaction.
+// If force is false:
```
okJiang, line 180: In what situations might we use false?

pkg/schedule/affinity/txn.go
lhy1024: ok

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		}
+
+		// Merge current ranges with new ranges to add
+		toAdd[op.GroupID] = GroupKeyRanges{
+			GroupID:   op.GroupID,
+			KeyRanges: slices.Concat(currentRanges.KeyRanges, op.KeyRanges),
+		}
+
+		// Collect new ranges for overlap validation
+		newRangesToValidate = append(newRangesToValidate, GroupKeyRanges{
```
okJiang: Why we need a new structure? It is same as `addOps`

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		for _, kr := range gkr.KeyRanges {
+			allNewRanges = append(allNewRanges, flattenKeyRange{
+				KeyRange: kr,
+				groupID:  gkr.GroupID,
+			})
+		}
+	}
+
+	// Step 1: Check for overlaps within the new ranges themselves
+	// This is O(N²) where N is the total number of new ranges across all groups
```
okJiang: Why not sort the `flattenKeyRange` to reduce the complexity

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+				}
+				return errs.ErrAffinityGroupContent.FastGenByArgs(
+					"key range overlaps between groups: " +
+						allNewRanges[i].groupID + " and " + allNewRanges[j].groupID)
+			}
+		}
+	}
+
+	// Step 2: Check new ranges against ALL existing ranges
+	// This is O(N × G×R) where N is new ranges, G is existing groups, R is ranges per group
```
okJiang: ditto

pkg/schedule/affinity/txn.go
lhy1024: There is no key ranges in affinity group.

pkg/schedule/affinity/txn.go
lhy1024: yes, we plan to reduce time with binary search in the future.

pkg/schedule/affinity/txn.go
lhy1024: I will fix it.

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+}
+
+// decodeHexKey decodes a hex string and returns an error if decoding fails.
+func decodeHexKey(hexStr, groupID, keyType string) ([]byte, error) {
+	if hexStr == "" {
+		return nil, nil
+	}
+	decoded, err := hex.DecodeString(hexStr)
+	if err != nil {
+		return nil, errs.ErrInvalidKeyFormat.FastGenByArgs(keyType, hexStr, groupID)
```
okJiang, line 570: Prefer to wrap this error in upper level to aviod to delivery `groupID` and `keyType` here.

pkg/schedule/affinity/txn.go
```diff
@@ -0,0 +1,805 @@
...
+		return GroupKeyRanges{}, nil
+	}
+
+	groupID, ok := parseAffinityGroupIDFromLabelRule(rule)
+	if !ok {
+		return GroupKeyRanges{}, nil
+	}
+
+	switch data := rule.Data.(type) {
+	case []*labeler.KeyRangeRule:
```
okJiang: Where is this case's input?

bufferflies (approved):

okJiang (approved):

PR #9999, opened 2025-12-04
affinity: add policy, metrics collection and tests

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	github.com/pingcap/tidb-dashboard v0.0.0-20250928152019-e111aaee93f4
 	github.com/prometheus/client_golang v1.20.5
 	github.com/rleungx/leakcheck v1.0.1-0.20250721100113-2eb456ffd3e3
-	github.com/sasha-s/go-deadlock v0.3.5
+	github.com/sasha-s/go-deadlock v0.3.6
 	github.com/shirou/gopsutil/v3 v3.23.3
 	github.com/smallnest/chanx v1.2.1-0.20240521153536-01121e21ff99
 	github.com/soheilhy/cmux v0.1.5
@@ -171,7 +171,7 @@ require (
 	github.com/oleiade/reflections v1.0.1 // indirect
 	github.com/onsi/gomega v1.20.1 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.2 // indirect
-	github.com/petermattis/goid v0.0.0-20240813172612-4fcff4a6cae7 // indirect
+	github.com/petermattis/goid v0.0.0-20250813065127-a731cc31b4fe // indirect
 	github.com/pingcap/tipb v0.0.0-20220718022156-3e2483c20a9e // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -454,8 +454,8 @@ github.com/pascaldekloe/name v0.0.0-20180628100202-0fd16699aae1/go.mod h1:eD5Jxq
 github.com/pelletier/go-toml/v2 v2.0.1/go.mod h1:r9LEWfGN8R5k0VXJ+0BkIe7MYkRdwZOjgMj2KwnJFUo=
 github.com/pelletier/go-toml/v2 v2.2.2 h1:aYUidT7k73Pcl9nb2gScu7NSrKCSHIDE89b3+6Wq+LM=
 github.com/pelletier/go-toml/v2 v2.2.2/go.mod h1:1t835xjRzz80PqgE6HHgN2JOsmgYu/h4qDAS4n929Rs=
-github.com/petermattis/goid v0.0.0-20240813172612-4fcff4a6cae7 h1:Dx7Ovyv/SFnMFw3fD4oEoeorXc6saIiQ23LrGLth0Gw=
-github.com/petermattis/goid v0.0.0-20240813172612-4fcff4a6cae7/go.mod h1:pxMtw7cyUw6B2bRH0ZBANSPg+AoSud1I1iyJHI69jH4=
+github.com/petermattis/goid v0.0.0-20250813065127-a731cc31b4fe h1:vHpqOnPlnkba8iSxU4j/CvDSS9J4+F4473esQsYLGoE=
+github.com/petermattis/goid v0.0.0-20250813065127-a731cc31b4fe/go.mod h1:pxMtw7cyUw6B2bRH0ZBANSPg+AoSud1I1iyJHI69jH4=
 github.com/phf/go-queue v0.0.0-20170504031614-9abe38d0371d h1:U+PMnTlV2tu7RuMK5etusZG3Cf+rpow5hqQByeCzJ2g=
 github.com/phf/go-queue v0.0.0-20170504031614-9abe38d0371d/go.mod h1:lXfE4PvvTW5xOjO6Mba8zDPyw8M93B6AQ7frTGnMlA8=
 github.com/pingcap/check v0.0.0-20191216031241-8a5a85928f12 h1:rfD9v3+ppLPzoQBgZev0qYCpegrwyFx/BUpkApEiKdY=
@@ -515,8 +515,8 @@ github.com/russross/blackfriday/v2 v2.1.0 h1:JIOH55/0cWyOuilr9/qlrm0BSXldqnqwMsf
 github.com/russross/blackfriday/v2 v2.1.0/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
 github.com/samber/lo v1.37.0 h1:XjVcB8g6tgUp8rsPsJ2CvhClfImrpL04YpQHXeHPhRw=
 github.com/samber/lo v1.37.0/go.mod h1:9vaz2O4o8oOnK23pd2TrXufcbdbJIa3b6cstBWKpopA=
-github.com/sasha-s/go-deadlock v0.3.5 h1:tNCOEEDG6tBqrNDOX35j/7hL5FcFViG6awUGROb2NsU=
-github.com/sasha-s/go-deadlock v0.3.5/go.mod h1:bugP6EGbdGYObIlx7pUZtWqlvo8k9H6vCBBsiChJQ5U=
+github.com/sasha-s/go-deadlock v0.3.6 h1:TR7sfOnZ7x00tWPfD397Peodt57KzMDo+9Ae9rMiUmw=
+github.com/sasha-s/go-deadlock v0.3.6/go.mod h1:CUqNyyvMxTyjFqDT7MRg9mb4Dv/btmGTqSR+rky/UXo=
 github.com/sergi/go-diff v1.1.0 h1:we8PVUC3FE2uYfodKH/nBHMSetSfHDR6scGdBi+erh0=
 github.com/sergi/go-diff v1.1.0/go.mod h1:STckp+ISIX8hZLjrqAeVduY0gWCT9IjLuqbuNXdaHfM=
 github.com/shirou/gopsutil/v3 v3.21.12/go.mod h1:BToYZVTlSVlfazpDDYFnsVZLaoRG+g8ufT6fPQLdJzA=
--- a/pkg/gctuner/memory_limit_tuner.go
+++ b/pkg/gctuner/memory_limit_tuner.go
@@ -61,7 +61,7 @@ func (t *memoryLimitTuner) tuning() {
 		return
 	}
 	r := memory.ForceReadMemStats()
-	gogc := util.GetGOGC()
+	gogc := util.GetGCPercent()
 	ratio := float64(100+gogc) / 100
 	// This `if` checks whether the **last** GC was triggered by MemoryLimit as far as possible.
 	// If the **last** GC was triggered by MemoryLimit, we'll set MemoryLimit to MAXVALUE to return control back to GOGC
--- a/pkg/gctuner/tuner.go
+++ b/pkg/gctuner/tuner.go
@@ -60,40 +60,38 @@ func init() {
 	SetMaxGCPercent(defaultMaxGCPercent)
 }
 
-// SetDefaultGOGC is to set the default GOGC value.
-func SetDefaultGOGC() {
-	util.SetGOGC(int(defaultGCPercent))
-}
-
 // Tuning sets the threshold of heap which will be respect by gogc tuner.
 // When Tuning, the env GOGC will not be take effect.
 // threshold: disable tuning if threshold == 0
 func Tuning(threshold uint64) {
 	// disable gc tuner if percent is zero
-	if threshold <= 0 && globalTuner != nil {
-		globalTuner.stop()
-		globalTuner = nil
-		return
+	if t := globalTuner.Load(); t == nil {
+		t1 := newTuner(threshold)
+		globalTuner.CompareAndSwap(nil, &t1)
+	} else {
+		if threshold <= 0 {
+			(*t).stop()
+			globalTuner.CompareAndSwap(t, nil)
+		} else {
+			(*t).setThreshold(threshold)
+		}
 	}
-
-	if globalTuner == nil {
-		globalTuner = newTuner(threshold)
-		return
-	}
-	globalTuner.setThreshold(threshold)
 }
 
-// GetGOGC returns the current GCPercent.
-func GetGOGC() uint32 {
-	if globalTuner == nil {
-		return defaultGCPercent
+// GetGOGCPercent returns the current GCPercent.
+func GetGOGCPercent() (percent uint32) {
+	t := globalTuner.Load()
+	if t == nil {
+		percent = defaultGCPercent
+	} else {
+		percent = (*t).getGCPercent()
 	}
-	return globalTuner.getGCPercent()
+	return percent
 }
 
 // only allow one gc tuner in one process
 // It is not thread-safe. so it is a private, singleton pattern to avoid misuse.
-var globalTuner *tuner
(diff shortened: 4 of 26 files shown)


pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,231 @@
...
+
+var (
+	// availabilityCheckIntervalForTest can be set in tests to speed up availability checks.
+	// Default is 0, which means use defaultAvailabilityCheckInterval.
+	availabilityCheckIntervalForTest time.Duration
+)
+
+// logEntry is used to collect log messages to print after releasing lock
+type logEntry struct {
+	level            string // "info", "warn"
```
liyishuai: ~Consider avoiding magic raw string values. Prefer standardized level constants:
https://github.com/uber-go/zap/blob/v1.27.0/level.go#L34-L38~

After reading how the entries are handled, I strongly recommend avoiding the word `level`, and use `kind` instead.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, entry := range logEntries {
+		switch entry.level {
+		case "warn":
+			log.Warn("affinity group invalidated due to unavailable stores",
+				zap.String("group-id", entry.groupID),
+				zap.Uint64("unavailable-store", entry.unavailableStore),
+				zap.String("availability", entry.availability.String()))
+		case "info":
+			log.Info("affinity group become available", zap.String("group-id", entry.groupID))
+		}
```
liyishuai: ¿ Is it a standard practice to distinguish the semantics of logs by their levels ?

pkg/schedule/affinity/policy.go
lhy1024: done. And we also need to modify logutil

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+	availabilityCheckIntervalForTest time.Duration
+)
+
+// logEntry is used to collect log messages to print after releasing lock
+type logEntry struct {
+	level            zapcore.Level
+	groupID          string
+	unavailableStore uint64
+	availability     groupAvailability
+}
```
liyishuai:
```suggestion
type invalidatedLogType struct {
	groupID          string
	unavailableStore uint64
	availability     groupAvailability
}

type availableLogType struct {
	groupID string
}

type logKind int

const (
	logKindInvalidated logKind = iota
	logKindAvailable
)

// logEntry is used to collect log messages to print after releasing lock
type logEntry struct {
	kind               logKind
	invalidatedPayload invalidatedLogType
	availablePayload   availableLogType
}
```

pkg/schedule/affinity/policy.go
liyishuai:
```suggestion
		switch entry.kind {
		case logKindInvalidated:
			log.Warn("affinity group invalidated due to unavailable stores",
				zap.String("group-id", entry.invalidatedPayload.groupID),
				zap.Uint64("unavailable-store", entry.invalidatedPayload.unavailableStore),
				zap.String("availability", entry.invalidatedPayload.availability.String()))
		case logKindAvailable:
			log.Info("affinity group become available", zap.String("group-id", entry.availablePayload.groupID))
		}
```

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+					groupID:          groupInfo.ID,
+					unavailableStore: unavailableStore,
+					availability:     newAvailability,
+				})
+			} else {
+				logEntries = append(logEntries, logEntry{
+					level:   zapcore.InfoLevel,
+					groupID: groupInfo.ID,
+				})
+			}
```
liyishuai:
```suggestion
			if unavailableStore != 0 {
				logEntries = append(logEntries, logEntry{
					kind: logKindInvalidated,
					invalidatedPayload: invalidatedLogType{
						groupID:          groupInfo.ID,
						unavailableStore: unavailableStore,
						availability:     newAvailability,
					},
				})
			} else {
				logEntries = append(logEntries, logEntry{
					kind: logKindAvailable,
					availablePayload: availableLogType{
						groupID: groupInfo.ID,
					},
				})
			}
```

pkg/schedule/affinity/policy.go
lhy1024: I think `level` is enough.

lhy1024:
> Avoid overloading types. Clarify "kind" instead of "level".

After discussion with @liyishuai , this is the only place where `logEntry` is needed, and there are no immediate plans for expansion. If a series of similar entries at the same level are added in the future, we can follow this approach by adding a kind to distinguish them and then move them into logutil.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+// SetAvailabilityCheckIntervalForTest sets the availability check interval for testing. Only use this in tests.
+func SetAvailabilityCheckIntervalForTest(interval time.Duration) {
+	availabilityCheckIntervalForTest = interval
+}
+
+// startAvailabilityCheckLoop starts a goroutine to periodically check store availability and invalidate groups with unavailable stores.
+func (m *Manager) startAvailabilityCheckLoop() {
+	interval := getAvailabilityCheckInterval()
+	ticker := time.NewTicker(interval)
+	go func() {
```
bufferflies, line 135: Does it need a waitgroup to wait for this goroutine to exit?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+func (m *Manager) setGroupAvailabilityChanges(unavailableStores map[uint64]storeCondition, groupAvailabilityChanges map[string]groupAvailability) {
+	m.Lock()
+	defer m.Unlock()
+	m.unavailableStores = unavailableStores
+	for groupID, availability := range groupAvailabilityChanges {
+		m.updateGroupAvailabilityLocked(groupID, availability)
+	}
+}
+
+func (m *Manager) hasUnavailableStore(leaderStoreID uint64, voterStoreIDs []uint64) error {
```
bufferflies: In general, the function named hasxxxx should return the store ID rather than an error.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+	m.RLock()
+	defer m.RUnlock()
+	for _, storeID := range voterStoreIDs {
+		condition, ok := m.unavailableStores[storeID]
+		if ok && (!condition.affectsLeaderOnly() || storeID == leaderStoreID) {
```
bufferflies, line 255: If store 3 is not the leader but has the  evit leader label, it should return an error?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+// getAvailabilityCheckInterval returns the availability check interval, which can be overridden for testing.
+func getAvailabilityCheckInterval() time.Duration {
+	if availabilityCheckIntervalForTest > 0 {
+		return availabilityCheckIntervalForTest
+	}
+	return defaultAvailabilityCheckInterval
+}
+
+// SetAvailabilityCheckIntervalForTest sets the availability check interval for testing. Only use this in tests.
+func SetAvailabilityCheckIntervalForTest(interval time.Duration) {
```
bufferflies: where is it used?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+	leaderStoreID := region.GetLeader().GetStoreId()
+	voterStoreIDs := make([]uint64, len(region.GetVoters()))
+	for i, voter := range region.GetVoters() {
+		voterStoreIDs[i] = voter.GetStoreId()
+	}
+	_, _ = m.updateAffinityGroupPeersWithAffinityVer(group.ID, group.affinityVer, leaderStoreID, voterStoreIDs)
+}
+
+// getAvailabilityCheckInterval returns the availability check interval, which can be overridden for testing.
+func getAvailabilityCheckInterval() time.Duration {
```
bufferflies: Maybe using a fail point is better?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+					level:   zapcore.InfoLevel,
+					groupID: groupInfo.ID,
+				})
+			}
+		}
+	}
+	m.RUnlock()
+
+	// Log after releasing lock
+	for _, entry := range logEntries {
```
bufferflies: If we have 100 nodes that are unavailable, we need to log 100 logs every time.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+		return false, nil
+	}
+
+	// Analyze which Groups have changed availability
+	// Collect log messages to print after releasing lock
+	for _, groupInfo := range m.groups {
+		var unavailableStore uint64
+		var maxCondition storeCondition
+		for _, storeID := range groupInfo.VoterStoreIDs {
+			if condition, ok := unavailableStores[storeID]; ok && (!condition.affectsLeaderOnly() || storeID == groupInfo.LeaderStoreID) {
```
bufferflies: How about extracting one function to check the group is healthy?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+				logEntries = append(logEntries, logEntry{
+					level:            zapcore.WarnLevel,
+					groupID:          groupInfo.ID,
+					unavailableStore: unavailableStore,
+					availability:     newAvailability,
+				})
+			} else {
+				logEntries = append(logEntries, logEntry{
+					level:   zapcore.InfoLevel,
+					groupID: groupInfo.ID,
```
bufferflies: Do we need to log the newAvailability?

pkg/schedule/affinity/policy.go
lhy1024: it will be used in checker test in another PR

pkg/schedule/affinity/policy.go
HunDunDM: Logs will only be generated whenever the overall status changes.

pkg/schedule/affinity/policy.go
HunDunDM: No, that's reasonable. We do not check the evict leader for peers who are not the leader.

pkg/schedule/affinity/policy.go
lhy1024: ok, I will replace it. Previously, this was done to conveniently conduct local testing and avoid repeatedly disabling and enabling failpoints.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+		for _, storeID := range groupInfo.VoterStoreIDs {
+			if condition, ok := unavailableStores[storeID]; ok && (!condition.affectsLeaderOnly() || storeID == groupInfo.LeaderStoreID) {
+				if unavailableStore == 0 || condition > maxCondition {
+					unavailableStore = storeID
+					maxCondition = condition
+				}
+			}
+		}
+		newAvailability := maxCondition.groupAvailability()
+		if newAvailability != groupInfo.GetAvailability() {
```
okJiang, line 220: Please add some comments to explain why when availability changing, `unavailableStore != 0` represents that group has a new unavailableStore, otherwise represents group becomes health.

This contains a great deal of underlying logical relationships. Perhaps modifying the code could make this section simpler and easier to understand.

pkg/schedule/affinity/policy.go
HunDunDM: This function doesn't perform any critical operations, so could we add a TODO statement for now?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+			groupAvailabilityChanges[groupInfo.ID] = newAvailability
```
okJiang, line 221: `Changes` can contain oldAvailbility and newAvailability. And the log level can be confirmed by compare the oldAvailbility and newAvailability(compare can be wrapped to a function). This makes the code clearer

pkg/schedule/affinity/policy.go
HunDunDM: We will change the print quantity.

pkg/schedule/affinity/manager_test.go
```diff
@@ -0,0 +1,568 @@
...
+	// Create affinity group
+	ranges := createGroupForTest(re, manager, "test_group", 6)
+
+	// Test 0: voterStoreIDs will be sorted.
+	_, err = manager.UpdateAffinityGroupPeers("test_group", 1, []uint64{3, 2, 1})
+	re.NoError(err)
+	groupInfo := getGroupForTest(re, manager, "test_group")
+	re.True(slices.Equal([]uint64{1, 2, 3}, groupInfo.VoterStoreIDs))
+
+	// Test 1: Region not belonging to any affinity group should return false
```
okJiang: Could you modify it to the `for-testCases` format? This would make it easier to expand test cases in the future and improve maintainability. Please determine if similar modifications can be made elsewhere.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,232 @@
...
+func (m *Manager) getGroupAvailabilityChanges(unavailableStores map[uint64]storeCondition) (isUnavailableStoresChanged bool, groupAvailabilityChanges map[string]groupAvailability) {
+	var logEntries []logEntry
+	groupAvailabilityChanges = make(map[string]groupAvailability)
+
+	// Validate whether unavailableStores has changed.
+	m.RLock()
+	isUnavailableStoresChanged = !maps.Equal(unavailableStores, m.unavailableStores)
+	if !isUnavailableStoresChanged {
+		m.RUnlock()
+		return false, nil
```
okJiang, line 213: When the store remains unchanged, the group will persist in a degraded state without transitioning to expired.

pkg/schedule/affinity/manager_test.go
lhy1024: ok, I will do it

pkg/schedule/affinity/policy.go
HunDunDM: We always use `GetAvailability` to retrieve the availability, which contains the relevant processing logic. We don't update it using `GetAvailability` because it might be called within a read lock.

pkg/schedule/affinity/policy.go
bufferflies: Yes, pls add todolist avoid the running goroutine exist even if the server is closed .

pkg/schedule/affinity/policy.go
lhy1024: changed

pkg/schedule/affinity/policy.go
lhy1024: added

pkg/schedule/affinity/policy.go
lhy1024: refactor it

pkg/schedule/affinity/policy.go
lhy1024: reduce unnecessary log

bufferflies (approved):

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,260 @@
...
+const (
+	// defaultAvailabilityCheckInterval is the default interval for checking store availability.
+	defaultAvailabilityCheckInterval = 10 * time.Second
+)
+
+// storeCondition is an enum for store conditions. Valid values are the store-prefixed enum constants,
+// which are split into three groups separated by degradedBoundary.
+type storeCondition int
+
+const (
```
rleungx, line 42: It's better to add a comment about the sequence matter.

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,260 @@
...
+	case storeEvictLeader:
+		return true
+	default:
+		return false
+	}
+}
+
+// GetNewAvailability uses the given unavailableStores to compute a new groupAvailability.
+// Note that this function does not update runtimeGroupInfo.
+func (g *runtimeGroupInfo) GetNewAvailability(unavailableStores map[uint64]storeCondition) groupAvailability {
```
rleungx, line 100: Can we merge GetNewAvailability and GetAvailability?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,260 @@
...
+	maxCondition := storeAvailable
```
rleungx, line 101: maxCondition is not readable

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,260 @@
...
+			case <-ticker.C:
+				m.checkStoresAvailability()
+			}
+		}
+	}()
+	log.Info("affinity manager availability check loop started", zap.Duration("interval", interval))
+}
+
+// checkStoresAvailability checks the availability status of stores and invalidates groups with unavailable stores.
+func (m *Manager) checkStoresAvailability() {
```
rleungx, line 152: should be checkGroupAvailability?

pkg/schedule/affinity/policy.go
```diff
@@ -0,0 +1,260 @@
...
+	m.RLock()
+	defer m.RUnlock()
+
+	// Collect global metrics
+	groupCount.Set(float64(len(m.groups)))
+	regionCount.Set(float64(len(m.regions)))
+	affinityRegionCount.Set(float64(m.affinityRegionCount))
+}
+
+func (m *Manager) generateUnavailableStores() map[uint64]storeCondition {
```
rleungx, line 175:
```suggestion
func (m *Manager) collectUnavailableStores() map[uint64]storeCondition {
```

PR #10011, opened 2025-12-08
core: inherit region size and keys together for uninitialized statistics

### LINKED ISSUE #10014: core: use previous keys if heartbeat's keys is zero
## Enhancement Task
ref https://github.com/tikv/tikv/issues/19180

lhy1024:

Issue Number: Close #10014 Ref https://github.com/tikv/tikv/issues/19180

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -221,7 +221,7 @@ type RegionHeartbeatRequest interface {
 func RegionFromHeartbeat(heartbeat RegionHeartbeatRequest, flowRoundDivisor uint64) *RegionInfo {
 	// Convert unit to MB.
 	// If region isn't empty and less than 1MB, use 1MB instead.
-	// The size of empty region will be correct by the previous RegionInfo.
+	// The size and keys of empty region will be corrected by the previous RegionInfo.
 	regionSize := heartbeat.GetApproximateSize() / units.MiB
 	if heartbeat.GetApproximateSize() > 0 && regionSize < EmptyRegionApproximateSize {
 		regionSize = EmptyRegionApproximateSize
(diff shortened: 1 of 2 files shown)


pkg/core/region.go
```diff
@@ -281,6 +283,15 @@ func (r *RegionInfo) Inherit(origin *RegionInfo, bucketEnable bool) {
 			r.approximateSize = EmptyRegionApproximateSize
 		}
 	}
+	// regionKeys should not be zero if region is not empty.
+	// Inherit the previous keys to avoid incorrect scheduling decisions.
+	if r.GetApproximateKeys() == 0 {
```
rleungx: What if the key is actually 0?

bufferflies:
https://github.com/tikv/pd/blob/1335ff93a407012eaf37f32d6657ed5b63bc308a/pkg/schedule/filter/region_filters.go#L154-L156 Does it need to change?

pkg/core/region.go
lhy1024: I've confirmed with @glorv . Currently, approximate_keys reports 0 mainly in two scenarios:
1. After a leader transfer: When the new leader sends its first heartbeat, the statistics (keys/size) might not have been recalculated yet, so it returns 0.
2. After an unsafe destroy range.

In the first scenario, reporting 0 implies "unknown" rather than "empty." We need to inherit the previous value to prevent incorrect scheduling decisions (such as triggering unnecessary merges). Even in the second scenario, inheriting the previous value temporarily is acceptable, as it will be corrected by subsequent heartbeats once the actual value is finalized.

lhy1024:
> https://github.com/tikv/pd/blob/1335ff93a407012eaf37f32d6657ed5b63bc308a/pkg/schedule/filter/region_filters.go#L154-L156
> 
> Does it need to change?

It need be cherry-picked, I think we should reduce unnecessary change

pkg/core/region.go
glorv: Only `size = 0` means the size/keys are unknown, so both size/keys should inherit the old value. Otherwise, should always use the new value.

PR #10016, opened 2025-12-08
schedule: compare leader scores when building the target

### LINKED ISSUE #9759: preferOldPeerAsLeader for fix-demote-voter may cause leader imbalance
## Enhancement Task

Test Step: Configure a placement rule that promotes learner to voter on one data center and demotes voter to learner on the other data center. 
The demotion operation will generate a "fix-demote-voter" operator, which attempts to transfer the leadership away from the peer before demotion if it is currently the leader. The target leader is selected according to the following rules:
`
	leaderPreferFuncs := []func(uint64) int{
		b.preferLeaderRoleAsLeader,
		b.preferUpStoreAsLeader,
		b.preferCurrentLeader,
		b.preferKeepVoterAsLeader,
		b.preferOldPeerAsLeader,
	}
`
PreferOldPeerAsLeader gives higher priority to peers with smaller IDs, making them more likely to be chosen as the target leader. 
This behavior appears to correlate with store IDs, as smaller store IDs tend to host more leaders than larger store IDs.
Image

liyishuai:

Issue Number: Close #9759 

* When other factors are indistinguishable, prefer peers with smaller leader scores.

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
Fix the issue of leader imbalance, biased towards smaller ID's.
```

### DIFF
--- a/pkg/schedule/operator/builder.go
+++ b/pkg/schedule/operator/builder.go
@@ -675,14 +675,15 @@ func (b *Builder) setTargetLeaderIfNotExist() {
 			continue
 		}
 		indistinguishable := true
+	compare_loop:
 		for _, f := range leaderPreferFuncs {
 			if best, next := f(b.targetLeaderStoreID), f(candidateStoreID); best < next {
 				b.targetLeaderStoreID = candidateStoreID
 				indistinguishable = false
-				break
+				break compare_loop
 			} else if best > next {
 				indistinguishable = false
-				break
+				break compare_loop
 			}
 		}
 		// If all comparison functions consider these two stores are indistinguishable,


pkg/schedule/operator/builder.go
```diff
@@ -673,14 +674,37 @@ func (b *Builder) setTargetLeaderIfNotExist() {
...
 			} else if best > next {
+				indistinguishable = false
 				break
 			}
 		}
+		// If all comparison functions consider these two stores are indistinguishable,
+		// we use leader score as the final tie breaker.
+		if indistinguishable {
+			current := b.GetBasicCluster().GetStore(b.targetLeaderStoreID)
+			candidate := b.GetBasicCluster().GetStore(targetLeaderStoreID)
```
bufferflies: how about rename the targetLeaderStoreID as candidateStoreID, avoid to confuse this stores?

pkg/schedule/operator/builder.go
```diff
@@ -673,14 +674,37 @@ func (b *Builder) setTargetLeaderIfNotExist() {
...
+			currentCountScore := current.LeaderScore(constant.ByCount, 0)
+			candidateCountScore := candidate.LeaderScore(constant.ByCount, 0)
+
+			if currentCountScore > candidateCountScore {
+				b.targetLeaderStoreID = targetLeaderStoreID
+			} else if currentCountScore == candidateCountScore {
+				currentSizeScore := current.LeaderScore(constant.BySize, 0)
+				candidateSizeScore := candidate.LeaderScore(constant.BySize, 0)
+
+				if currentSizeScore > candidateSizeScore {
```
bufferflies, line 704: What's next if the size score is the same?I think Rand is better?

pkg/schedule/operator/builder.go
```diff
@@ -673,14 +674,37 @@ func (b *Builder) setTargetLeaderIfNotExist() {
...
+					b.targetLeaderStoreID = targetLeaderStoreID
+				}
+			}
+		}
```
JmPotato, line 708: What about making these logic as a `leaderPreferFunc` also?

pkg/schedule/operator/builder.go
```diff
@@ -673,14 +674,37 @@ func (b *Builder) setTargetLeaderIfNotExist() {
...
+				indistinguishable = false
 				break
 			} else if best > next {
+				indistinguishable = false
 				break
 			}
 		}
+		// If all comparison functions consider these two stores are indistinguishable,
+		// we use leader score as the final tie breaker.
+		if indistinguishable {
```
bufferflies, line 691: You can just return if the condition is false, and then handle the next pick steps to reduce the if-else.

pkg/schedule/operator/builder.go
liyishuai: The existing `leaderPreferFunc`s return integer, while the leader scores are `float64`. I am reluctant to mix the type interface, or casting between numeric datatypes.

pkg/schedule/operator/builder.go
liyishuai: The peer with smaller ID gets chosen: https://github.com/tikv/pd/issues/9759#issuecomment-3312900917

pkg/schedule/operator/builder.go
liyishuai: Renamed in ~f4650c22edb92e9b2d20a5c4d3eca58e60bc379d~ ca754ab91734975018f7f5fa1c755fdf81aee63d

pkg/schedule/operator/builder.go
liyishuai: ~Refactored in f4650c22edb92e9b2d20a5c4d3eca58e60bc379d~ reverted.

liyishuai: Negative! `break` says no need to consider other **functions**, but we should still consider other **candidates**.

pkg/schedule/operator/builder.go
liyishuai: My previous version was randomly shuffling the target peers, but discarded it for two reasons:
1. Provided the current leader-score-based comparison, more randomness is less likely to improve the results.
2. Nondeterminism makes the test framework much harder to write. (I wrote a Ph.D. dissertation on it.)

pkg/schedule/operator/builder.go
```diff
@@ -660,25 +661,49 @@ func (b *Builder) setTargetLeaderIfNotExist() {
...
+		indistinguishable := true
+	compare_loop:
 		for _, f := range leaderPreferFuncs {
-			if best, next := f(b.targetLeaderStoreID), f(targetLeaderStoreID); best < next {
-				b.targetLeaderStoreID = targetLeaderStoreID
-				break
+			if best, next := f(b.targetLeaderStoreID), f(candidateStoreID); best < next {
+				b.targetLeaderStoreID = candidateStoreID
+				indistinguishable = false
+				break compare_loop
```
JmPotato, line 683: What is the purpose of adding a label here? It seems that a direct break would achieve the same result.

liyishuai: to avoid https://github.com/tikv/pd/pull/10016#discussion_r2601515367

PR #10017, opened 2025-12-08
feat(resourcemanager/metering): report write bytes in RU metering data

JmPotato:

Issue Number: ref #9707.

```commit-message
Updated aggregation logic to include `write_bytes` in metering output.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mcs/resourcemanager/server/metering.go
+++ b/pkg/mcs/resourcemanager/server/metering.go
@@ -27,15 +27,16 @@ const (
 	ResourceManagerCategory = "resource-manager"
 	ruMeteringVersion       = "1"
 
-	meteringDataOltpRUField              = "oltp_ru"
-	meteringDataOlapRUField              = "olap_ru"
+	meteringDataOLTPRUField              = "oltp_ru"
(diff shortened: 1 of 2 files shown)


okJiang (approved):

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -27,15 +27,16 @@ const (
 	ResourceManagerCategory = "resource-manager"
 	ruMeteringVersion       = "1"
 
-	meteringDataOltpRUField              = "oltp_ru"
-	meteringDataOlapRUField              = "olap_ru"
+	meteringDataOLTPRUField              = "oltp_ru"
+	meteringDataOLAPRUField              = "olap_ru"
+	meteringDataWriteBytesField          = "write_bytes"
```
rleungx, line 32: Only the write bytes need to be considered?

JmPotato: Yes, currently we only need to add this field after confirming with PM.

rleungx (approved):

PR #10032, opened 2025-12-09
tests: fix TestTSOKeyspaceGroupSplitElection

### LINKED ISSUE #9842: `TestTSOKeyspaceGroupSplitElection` is flaky
### Which jobs are failing
```
keyspace_group_manager_test.go:377: 
        	Error Trace:	/home/runner/work/pd/pd/tests/integrations/mcs/tso/keyspace_group_manager_test.go:377
        	Error:      	Not equal: 
        	            	expected: []string{"http://127.0.0.1:39531/"}
        	            	actual  : []string{"http://127.0.0.1:43701/"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 ([]string) (len=1) {
        	            	- (string) (len=22) "http://127.0.0.1:39531/"
        	            	+ (string) (len=22) "http://127.0.0.1:43701/"
        	            	 }
        	Test:       	TestTSOKeyspaceGroupManagerSuite/TestTSOKeyspaceGroupSplitElection
```

### CI link
https://github.com/tikv/pd/actions/runs/18549657370/job/52874637660?pr=9841

liupan1111:

Issue Number: Close #9842

- pause split finish, and ensure two keyspace groups primary in the same tso server.
```commit-message
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/tso_keyspace_group.go
+++ b/pkg/keyspace/tso_keyspace_group.go
@@ -702,6 +702,9 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 	var splitTargetKg, splitSourceKg *endpoint.KeyspaceGroup
 	m.Lock()
 	defer m.Unlock()
+
+	failpoint.Inject("pauseFinishSplitBeforeTxn", nil)
+
 	if err := m.store.RunInTxn(m.ctx, func(txn kv.Txn) (err error) {
 		// Load the split target keyspace group first.
 		splitTargetKg, err = m.store.LoadKeyspaceGroup(txn, splitTargetID)
--- a/tests/integrations/mcs/tso/keyspace_group_manager_test.go
+++ b/tests/integrations/mcs/tso/keyspace_group_manager_test.go
@@ -462,6 +462,9 @@ func (suite *tsoKeyspaceGroupManagerTestSuite) requestTSO(
 
 func (suite *tsoKeyspaceGroupManagerTestSuite) TestTSOKeyspaceGroupSplitElection() {
 	re := suite.Require()
+
+	re.NoError(failpoint.Enable("github.com/tikv/pd/pkg/keyspace/pauseFinishSplitBeforeTxn", `pause`))
+
 	// Create the keyspace group `oldID` with keyspaces [444, 555, 666].
 	oldID := suite.allocID()
 	handlersutil.MustCreateKeyspaceGroup(re, suite.pdLeaderServer, &handlers.CreateKeyspaceGroupParams{
@@ -517,6 +520,7 @@ func (suite *tsoKeyspaceGroupManagerTestSuite) TestTSOKeyspaceGroupSplitElection
 	})
 	re.Equal(primary1.GetServingUrls(), primary2.GetServingUrls())
 	// Wait for the keyspace groups to finish the split.
+	re.NoError(failpoint.Disable("github.com/tikv/pd/pkg/keyspace/pauseFinishSplitBeforeTxn"))
 	waitFinishSplit(re, suite.pdLeaderServer, oldID, newID, []uint32{444}, []uint32{555, 666})
 }
 


pkg/keyspace/tso_keyspace_group.go
```diff
@@ -712,6 +702,11 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 	var splitTargetKg, splitSourceKg *endpoint.KeyspaceGroup
 	m.Lock()
 	defer m.Unlock()
+
+	failpoint.Inject("pauseTxn", func() {
+		time.Sleep(2 * time.Second)
+	})
+
```
liyishuai, line 707: If the test failed without the 2-second pause, will the production system fail as well?
* If yes, then only injecting the failpoint doesn't fix the production.
* If no, then what's the difference between the integration test and the production system that caused such diversion?

tests/integrations/mcs/tso/keyspace_group_manager_test.go
```diff
@@ -462,6 +462,9 @@ func (suite *tsoKeyspaceGroupManagerTestSuite) requestTSO(
 
 func (suite *tsoKeyspaceGroupManagerTestSuite) TestTSOKeyspaceGroupSplitElection() {
 	re := suite.Require()
+
+	re.NoError(failpoint.Enable("github.com/tikv/pd/pkg/tso/pauseTxn", `return(true)`))
```
ystaticy: Can we just use `return()` instead of `return(true)`?

liupan1111: I'm not very familiar with return yet—I just followed how others wrote it... 
Let me look into it and get back to you.

pkg/keyspace/tso_keyspace_group.go
liupan1111: it is not related about production system, it is just about this test case. This test case will check whether the primarys before splitting finish, but not well control the race condition

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -712,6 +702,11 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 	var splitTargetKg, splitSourceKg *endpoint.KeyspaceGroup
 	m.Lock()
 	defer m.Unlock()
+
+	failpoint.Inject("pauseTxn", func() {
+		time.Sleep(2 * time.Second)
```
rleungx: How about using `pause` instead of sleep?

liupan1111: Actually, I also considered "pause" last night, but @niubell and @ystaticy thought "sleep" was good enough, so I can't make up my mind... @niubell @ystaticy ping

tests/integrations/mcs/tso/keyspace_group_manager_test.go
liupan1111: @ystaticy After investigate, I agree.

pkg/keyspace/tso_keyspace_group.go
liupan1111: @rleungx @ystaticy i 'v replaced sleep with pause, please take a look

tests/integrations/mcs/tso/keyspace_group_manager_test.go
```diff
@@ -462,6 +462,9 @@ func (suite *tsoKeyspaceGroupManagerTestSuite) requestTSO(
 
 func (suite *tsoKeyspaceGroupManagerTestSuite) TestTSOKeyspaceGroupSplitElection() {
 	re := suite.Require()
+
+	re.NoError(failpoint.Enable("github.com/tikv/pd/pkg/tso/pauseTxn", `pause`))
```
rleungx: seem the path is incorrect

liupan1111: @rleungx thanks for correction, please take a look my change

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -712,6 +702,9 @@ func (m *GroupManager) FinishSplitKeyspaceByID(splitTargetID uint32) error {
 	var splitTargetKg, splitSourceKg *endpoint.KeyspaceGroup
 	m.Lock()
 	defer m.Unlock()
+
+	failpoint.Inject("pauseTxn", nil)
```
rleungx: Perhaps, we can use a more meaningful name like pauseFinishSplit or something else. pauseTxn sounds too general. The rest LGTM

rleungx (approved):

rleungx:
As for the PR title, usually we use the subsystem name or module name, see https://github.com/tikv/pd/blob/master/CONTRIBUTING.md

pkg/keyspace/tso_keyspace_group.go
liupan1111: Agree， let me change

liupan1111:
> As for the PR title, usually we use the subsystem name or module name, see https://github.com/tikv/pd/blob/master/CONTRIBUTING.md

ok, let me modify

pkg/keyspace/tso_keyspace_group.go
liupan1111: @rleungx done

PR #10038, opened 2025-12-10
schedule: add affinity filter and kvproto update and use affinity mananager

lhy1024:

Issue Number: Ref #9764

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/mock/mockcluster/mockcluster.go
+++ b/pkg/mock/mockcluster/mockcluster.go
@@ -216,7 +216,7 @@ func (mc *Cluster) initRuleManager() {
 		mc.RuleManager = placement.NewRuleManager(mc.ctx, mc.GetStorage(), mc, mc.GetSharedConfig())
 		err := mc.Initialize(int(mc.GetReplicationConfig().MaxReplicas), mc.GetReplicationConfig().LocationLabels, mc.GetReplicationConfig().IsolationLevel, false)
 		if err != nil {
-			log.Error("failed to initialize rule manager", errs.ZapError(err))
+			log.Info("failed to initialize rule manager", errs.ZapError(err))
 		}
 	}
 }
--- a/errors.toml
+++ b/errors.toml
@@ -8,7 +8,7 @@ KMS error
 
 ["PD:affinity:ErrAffinityDisabled"]
 error = '''
-affinity is disabled by config
+affinity is disabled
 '''
 
 ["PD:affinity:ErrAffinityGroupConflict"]
--- a/pkg/errs/errno.go
+++ b/pkg/errs/errno.go
@@ -561,7 +561,7 @@ var (
 	ErrAffinityGroupNotFound  = errors.Normalize("affinity group %s not found", errors.RFCCodeText("PD:affinity:ErrAffinityGroupNotFound"))
 	ErrAffinityGroupExist     = errors.Normalize("affinity group %s already exists", errors.RFCCodeText("PD:affinity:ErrAffinityGroupExist"))
 	ErrAffinityGroupConflict  = errors.Normalize("affinity group %s cannot appear in both add and remove operations in the same request", errors.RFCCodeText("PD:affinity:ErrAffinityGroupConflict"))
-	ErrAffinityDisabled       = errors.Normalize("affinity is disabled by config", errors.RFCCodeText("PD:affinity:ErrAffinityDisabled"))
+	ErrAffinityDisabled       = errors.Normalize("affinity is disabled", errors.RFCCodeText("PD:affinity:ErrAffinityDisabled"))
 	ErrAffinityInternal       = errors.Normalize("affinity internal error", errors.RFCCodeText("PD:affinity:ErrAffinityInternal"))
 	ErrAffinityGroupContent   = errors.Normalize("invalid affinity group content, %s", errors.RFCCodeText("PD:affinity:ErrAffinityGroupContent"))
 	ErrInvalidLabelRuleFormat = errors.Normalize("invalid label rule data format %s", errors.RFCCodeText("PD:affinity:ErrInvalidLabelRuleFormat"))


lhy1024:
@bufferflies @okJiang friendly ping

server/server.go
```diff
@@ -2165,3 +2166,16 @@ func (s *Server) GetMaxResetTSGap() time.Duration {
...
+
+// GetAffinityManager gets the affinity manager.
+func (s *Server) GetAffinityManager() (*affinity.Manager, error) {
+	rc := s.GetRaftCluster()
+	if rc == nil {
+		return nil, errs.ErrNotBootstrapped.GenWithStackByArgs()
+	}
+	manager := rc.GetAffinityManager()
+	if manager == nil {
+		return nil, errs.ErrAffinityDisabled
```
okJiang: Is this error accurate?

okJiang: Affinity Disabled is like a warning rather than an error

pkg/schedule/filter/region_filters.go
```diff
@@ -172,6 +172,16 @@ func (f *regionWitnessFilter) Select(region *core.RegionInfo) *plan.Status {
 	return statusOK
 }
 
+// HasWitnessPeers checks if the region has witness peers.
+func HasWitnessPeers(region *core.RegionInfo) bool {
```
okJiang, line 177: Where is this used?

okJiang:
rest lgtm

pkg/schedule/filter/region_filters.go
lhy1024: it will be used in 10040

rleungx: Do we need to consider the witness?

pkg/schedule/filter/affinity.go
```diff
@@ -0,0 +1,100 @@
...
+	sche "github.com/tikv/pd/pkg/schedule/core"
+	"github.com/tikv/pd/pkg/schedule/plan"
+)
+
+type affinityFilter struct {
+	affinityManager *affinity.Manager
+}
+
+// NewAffinityFilter creates a RegionFilter that filters all affinity regions.
+func NewAffinityFilter(cluster sche.SharedCluster) RegionFilter {
```
rleungx: Why not put it into the region filter?

pkg/schedule/filter/region_filters.go
lhy1024: Concerned about the impact of the witness, the region containing the witness peer will be skipped during the affinity check.

bufferflies: I suggest not considering the witness code

server/server.go
lhy1024: ok, I will add a new error for it

pkg/schedule/filter/affinity.go
```diff
@@ -0,0 +1,100 @@
...
+// Derived from default values:
+//   - Default MaxMergeRegionSize = 54MB
+//   - Default region-split-size = 256MB
+const (
+	affinityRegionSizeBufferMB   = 10
+	affinityRegionSizeMultiplier = 4
+)
+
+// AllowAutoSplit returns true if the region can be auto split
+func AllowAutoSplit(cluster sche.ClusterInformer, region *core.RegionInfo, reason pdpb.SplitReason) bool {
```
rleungx, line 37: How about:
```
func (c *RaftCluster) allowSplit(region *core.RegionInfo, reason pdpb.SplitReason) bool
```

pkg/mock/mockcluster/mockcluster.go
```diff
@@ -211,7 +214,7 @@ func (mc *Cluster) AllocPeer(storeID uint64) (*metapb.Peer, error) {
 func (mc *Cluster) initRuleManager() {
 	if mc.RuleManager == nil {
 		mc.RuleManager = placement.NewRuleManager(mc.ctx, mc.GetStorage(), mc, mc.GetSharedConfig())
-		mc.Initialize(int(mc.GetReplicationConfig().MaxReplicas), mc.GetReplicationConfig().LocationLabels, mc.GetReplicationConfig().IsolationLevel, false)
+		_ = mc.Initialize(int(mc.GetReplicationConfig().MaxReplicas), mc.GetReplicationConfig().LocationLabels, mc.GetReplicationConfig().IsolationLevel, false)
```
bufferflies: How about logging this error if it isn't nil for debugging certain cases?

pkg/schedule/filter/affinity.go
```diff
@@ -0,0 +1,100 @@
...
+	if region == nil {
+		// The default behavior is to allow it.
```
rleungx, line 39: What is the purpose of it?

pkg/schedule/filter/affinity.go
```diff
@@ -0,0 +1,100 @@
...
+		return true
+	}
+
+	if !cluster.GetCheckerConfig().IsAffinitySchedulingEnabled() {
+		return true
+	}
+
+	switch reason {
```
bufferflies: if reason == SplitReason_ADMIN{
    return true 
}
It just check the reason is admin, not needs to use switch

pkg/schedule/filter/affinity.go
HunDunDM: This ensures that the behavior of `region == nil` remains consistent with that before this code was added.

server/cluster/cluster_worker.go
```diff
@@ -135,6 +137,13 @@ func (c *RaftCluster) HandleAskBatchSplit(request *pdpb.AskBatchSplitRequest) (*
 	if repMode := c.GetReplicationMode(); repMode != nil && repMode.IsRegionSplitPaused() {
 		return nil, errors.New("region split is paused by replication mode")
 	}
+
+	region := c.GetRegion(reqRegion.GetId())
```
bufferflies, line 141: Do we need to add this check, HandleAskSplit, also?

HunDunDM: `AskSplit` and `ReportSplit` were marked as deprecated 8 years ago. Currently, no components use them, and their functionality is incomplete.

pkg/mock/mockcluster/mockcluster.go
lhy1024: ok

pkg/schedule/filter/affinity.go
lhy1024: ok

pkg/schedule/filter/affinity.go
lhy1024: ok

pkg/schedule/filter/region_filters.go
HunDunDM: Current implementation is incompatible with Witness.

pkg/schedule/filter/affinity.go
rleungx: BTW, we have already checked if the region == nil in the gRPC layer

pkg/schedule/affinity/manager.go
```diff
@@ -79,7 +79,7 @@ type Manager struct {
 // NewManager creates a new affinity Manager.
 func NewManager(ctx context.Context, storage endpoint.AffinityStorage, storeSetInformer core.StoreSetInformer, conf config.SharedConfigProvider, regionLabeler *labeler.RegionLabeler) (*Manager, error) {
```
bufferflies, line 80: Can we review this before the New Manager?

pkg/schedule/filter/region_filters.go
bufferflies: so we can remove all changeds related with witness

server/server.go
```diff
@@ -2165,3 +2166,16 @@ func (s *Server) GetMaxResetTSGap() time.Duration {
...
 }
+
+// GetAffinityManager gets the affinity manager.
+func (s *Server) GetAffinityManager() (*affinity.Manager, error) {
+	rc := s.GetRaftCluster()
+	if rc == nil {
+		return nil, errs.ErrNotBootstrapped.GenWithStackByArgs()
+	}
+	manager := rc.GetAffinityManager()
+	if manager == nil {
```
bufferflies, line 2177: when it happened?

pkg/schedule/affinity/manager.go
lhy1024: In code order, it is 
```
	c.regionLabeler, err = labeler.NewRegionLabeler(c.ctx, c.storage, regionLabelGCInterval)
	if err != nil {
		return err
	}

	// create affinity manager with region labeler for key range validation and rebuild
	c.affinityManager, err = affinity.NewManager(c.ctx, c.storage, c, c.GetOpts(), c.regionLabeler)
	if err != nil {
		return err
	}
```
So it only defensive programming

server/server.go
lhy1024: In fact, it is highly unlikely to occur, as we have all undergone the necessary checks.

PR #10040, opened 2025-12-10
schedule: add affinity checker

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/checker_controller.go
+++ b/pkg/schedule/checker/checker_controller.go
@@ -360,18 +360,15 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
 		return nil
 	}
 
-	// Check affinity schedule limit and affinityManager
-	if c.conf.GetAffinityScheduleLimit() > 0 && c.affinityChecker.affinityManager != nil {
-		if ops := measureChecker(c.metrics.checkRegionHistograms[affinityChecker], func() []*operator.Operator {
-			if opController.OperatorCount(operator.OpAffinity) < c.conf.GetAffinityScheduleLimit() {
-				// It makes sure that two affinity merge operators can be added successfully altogether.
-				return c.affinityChecker.Check(region)
-			}
-			operator.IncOperatorLimitCounter(c.affinityChecker.GetType(), operator.OpAffinity)
-			return nil
-		}); len(ops) > 0 {
-			return ops
+	if ops := measureChecker(c.metrics.checkRegionHistograms[affinityChecker], func() []*operator.Operator {
+		if opController.OperatorCount(operator.OpAffinity) < c.conf.GetAffinityScheduleLimit() {
+			// It makes sure that two affinity merge operators can be added successfully altogether.
+			return c.affinityChecker.Check(region)
 		}
+		operator.IncOperatorLimitCounter(c.affinityChecker.GetType(), operator.OpAffinity)
+		return nil
+	}); len(ops) > 0 {
+		return ops
 	}
 
 	if ops := measureChecker(c.metrics.checkRegionHistograms[mergeChecker], func() []*operator.Operator {


pkg/core/region_option.go
```diff
@@ -384,6 +384,17 @@ func WithReplacePeerStore(oldStoreID, newStoreID uint64) RegionCreateOption {
 	}
 }
 
+// WithLeaderStore sets the voter peer on leaderStoreID as the leader.
+func WithLeaderStore(leaderStoreID uint64) RegionCreateOption {
+	return func(region *RegionInfo) {
+		for _, p := range region.GetPeers() {
```
bufferflies, line 390: It should iterate over voters, not all peers.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	if !filter.IsRegionHealthy(region) {
+		affinityCheckerUnhealthyRegionCounter.Inc()
+		return nil
+	}
+	if !filter.IsRegionReplicated(c.cluster, region) {
+		affinityCheckerAbnormalReplicaCounter.Inc()
+		return nil
+	}
+	if filter.HasWitnessPeers(region) {
```
rleungx: We can remove it?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+	sourceVoters := region.GetVoters()
+
+	if len(sourceVoters) != len(voterStoreIDs) || !slices.Contains(voterStoreIDs, leaderStoreID) {
+		return nil
+	}
+
+	// presence records whether a storeID exists in the source and/or target:
+	//   - 1(0b01) if only in source
+	//   - 2(0b10) if only in target
+	//   - 3(0b11) if in both
```
rleungx: Why not directly use a readable variable instead of 1,2,3?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+		case 1:
+			sourceOnly = append(sourceOnly, storeID)
+		case 2:
+			targetOnly = append(targetOnly, storeID)
+		}
+	}
+
+	options := make([]core.RegionCreateOption, 0, len(sourceOnly)+1)
+	for i, sourceStoreID := range sourceOnly {
+		options = append(options, core.WithReplacePeerStore(sourceStoreID, targetOnly[i]))
```
rleungx: Does the targetOnly always have the same length with sourceOnly? BTW, targetOnly and sourceOnly are confused.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+	needRefetch := false
+	if group.AffinitySchedulingAllowed {
+		// If the Group is affinity scheduling allowed and the Region is not in the affinity state,
+		// we expect to schedule the Region to match the Group’s peers.
+		// Before doing so, check whether the Group’s peers are replicated.
+		// If not, the Group information should be expired (e.g. due to Placement Rules changes),
+		// so expire the group first, then provide the available Region information and fetch the Group state again.
+		if !isAffinity {
+			targetRegion := cloneRegionWithReplacePeerStores(region, group.LeaderStoreID, group.VoterStoreIDs...)
+			if targetRegion == nil || !filter.IsRegionReplicated(c.cluster, targetRegion) {
```
rleungx, line 105: When will the targetRegion not RegionReplicated?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+	peers := make(map[uint64]*metapb.Peer)
+	for storeID, role := range roles {
+		peers[storeID] = &metapb.Peer{
+			StoreId: storeID,
+			Role:    role.MetaPeerRole(),
+		}
+	}
+
+	// Use Builder to create the operator with preserved learners
+	builder := operator.NewBuilder(
```
rleungx: Should be moved to create_operator.go

pkg/schedule/checker/affinity_checker.go
lhy1024: If we confirm that no clusters are using witness, I believe it can be removed.
BTW, should we create an issue to mark when the witness is to be completely removed?

pkg/schedule/checker/affinity_checker.go
HunDunDM: placement rules changed

pkg/schedule/checker/affinity_checker.go
lhy1024: inSource, inTarget and inBoth?

pkg/schedule/checker/affinity_checker.go
lhy1024: How about storesToRemove and storesToAdd?

pkg/schedule/checker/affinity_checker.go
bufferflies: yes, we can

pkg/schedule/checker/affinity_checker.go
lhy1024: We have check len(sourceVoters) and len(voterStoreIDs), they are removed the same stores.

pkg/schedule/checker/affinity_checker.go
lhy1024: done

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+			roles[storeID] = placement.Leader
+		} else {
+			roles[storeID] = placement.Voter
+		}
+	}
+
+	// Preserve all existing learner peers
+	// Since regions that violate placement rules won't reach this function,
+	// we can safely assume all existing learners should be preserved
+	for _, learner := range region.GetLearners() {
```
bufferflies, line 194: Why do we need to change learner

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,450 @@
...
+
+	// Use Builder to create the operator with preserved learners
+	builder := operator.NewBuilder(
+		"affinity-move-region",
+		c.cluster,
+		region,
+	).SetPeers(peers).SetExpectedRoles(roles)
+
+	// Skip building if target leader store currently disallows leader in (e.g., evict-leader / reject-leader).
+	if targetLeader := c.cluster.GetStore(group.LeaderStoreID); targetLeader != nil {
```
bufferflies, line 212: this check should place before the build

lhy1024: done

pkg/schedule/checker/affinity_checker.go
lhy1024: remove it

pkg/core/region_option.go
lhy1024: No, RegionInfo.Clone will create &RegionInfo firstly, then apply opts(contains WithLeaderStore), then set voters(call classifyVoterAndLearner(region)). So voters is nil in opts.

So we iterate peers and check whether is learner.

pkg/schedule/checker/affinity_checker.go
lhy1024: We don't change learners. This code preserves all existing learner peers (e.g., TiFlash replicas) by including them in the target roles map.

If a learner's storeID conflicts with an affinity voter, we abort the operator creation (return nil) instead of modifying the learner

lhy1024: TestAffinityCheckerPreserveLearners keeps all learners
TestAffinityCheckerLearnerVoterConflict tests the case that learner and leader are conflict

pkg/schedule/checker/checker_controller.go
```diff
@@ -352,25 +354,48 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
...
+		denyCheckersByLabelerCounter.Inc()
+		return nil
 	}
-	if c.mergeChecker != nil {
-		if ops := measureChecker(c.metrics.checkRegionHistograms[mergeChecker], func() []*operator.Operator {
-			if opController.OperatorCount(operator.OpMerge) < c.conf.GetMergeScheduleLimit() {
-				// It makes sure that two operators can be added successfully altogether.
-				return c.mergeChecker.Check(region)
+
+	if c.conf.IsAffinitySchedulingEnabled() {
```
rleungx:
```suggestion
	if c.conf.IsAffinityEnabled() {
```

pkg/schedule/checker/checker_controller.go
```diff
@@ -352,25 +354,48 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
...
+		if ops := measureChecker(c.metrics.checkRegionHistograms[affinityChecker], func() []*operator.Operator {
```
rleungx: I prefer to use a separate function.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// Name returns AffinityChecker's name.
+func (*AffinityChecker) Name() string {
+	return types.AffinityChecker.String()
+}
+
+// Check verifies a region's replicas according to affinity group constraints, creating an Operator if needed.
+func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
+	if c.affinityManager == nil {
```
rleungx: When will it become nil?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,434 @@
...
+	needRefetch := false
+	if group.AffinitySchedulingAllowed {
+		// If the Group is affinity scheduling allowed and the Region is not in the affinity state,
+		// we expect to schedule the Region to match the Group’s peers.
+		// Before doing so, check whether the Group’s peers are replicated.
+		// If not, the Group information should be expired (e.g. due to Placement Rules changes),
+		// so expire the group first, then provide the available Region information and fetch the Group state again.
+		if !isAffinity {
+			targetRegion := cloneRegionWithReplacePeerStores(region, group.LeaderStoreID, group.VoterStoreIDs...)
+			if targetRegion == nil || !filter.IsRegionReplicated(c.cluster, targetRegion) {
```
rleungx, line 105: Only check replicated?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,434 @@
...
+
+	return op
+}
+
+// MergeCheck verifies if a region can be merged with its adjacent regions within the same affinity group.
+// It follows similar logic to merge_checker but with affinity-specific constraints:
+// - Does NOT skip recently split or recently started regions
+// - Does NOT skip hot spots regions
+// - Only merges regions within the same affinity group
+func (c *AffinityChecker) MergeCheck(region *core.RegionInfo, group *affinity.GroupState) []*operator.Operator {
```
rleungx:
```suggestion
func (c *AffinityChecker) mergeCheck(region *core.RegionInfo, group *affinity.GroupState) []*operator.Operator {
```

pkg/schedule/checker/checker_controller.go
lhy1024: This config is 
```
EnableAffinityScheduling bool `toml:"enable-affinity-scheduling"
```

pkg/schedule/checker/affinity_checker.go
lhy1024: ok

pkg/schedule/checker/affinity_checker.go
lhy1024: only defensive programming

pkg/schedule/checker/checker_controller.go
lhy1024: put this to a sparate function?
img_v3_02t0_94194bf1-18ce-4160-99b8-950252a7acag

pkg/schedule/checker/affinity_checker.go
lhy1024: yes

HunDunDM: The health check for the store is performed in the `checkStoresAvailability` function.

pkg/schedule/checker/checker_controller.go
lhy1024: Could we do it later?

pkg/schedule/checker/checker_controller.go
```diff
@@ -352,25 +354,48 @@ func (c *Controller) CheckRegion(region *core.RegionInfo) []*operator.Operator {
...
+			ops := c.affinityChecker.Check(region)
+			if len(ops) > 0 {
+				opKind := ops[0].Kind()
+				if (opKind & operator.OpMerge) == 0 {
```
rleungx: So there is no way to control the speed independently?

lhy1024: It is not neccessary to introduce a new limit for affinity scheduling

rleungx: What if it affects the performance?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	sourceVoters := region.GetVoters()
+
+	if len(sourceVoters) != len(voterStoreIDs) || !slices.Contains(voterStoreIDs, leaderStoreID) {
+		return nil
+	}
+
+	// presences records whether a storeID exists in the source and/or target
+	presences := make(map[uint64]int, len(sourceVoters)+len(voterStoreIDs))
```
rleungx, line 405: Can we remove it and directly put it into storesToRemove/storesToAdd?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -0,0 +1,434 @@
...
+		case inSource:
+			storesToRemove = append(storesToRemove, storeID)
+		case inTarget:
+			storesToAdd = append(storesToAdd, storeID)
+		}
+	}
+
+	options := make([]core.RegionCreateOption, 0, len(storesToRemove)+1)
+	for i, removeStoreID := range storesToRemove {
+		options = append(options, core.WithReplacePeerStore(removeStoreID, storesToAdd[i]))
```
rleungx, line 426: Should it keep the same naming as WithLeaderStore?

pkg/schedule/checker/checker_controller.go
lhy1024: Generally speaking, once it enters the stable state, no further scheduling occurs, and we can restrict it using region-schedule-limit and merge-schedule-limit.

Simultaneously, we can disable affinity scheduling to prevent it from affecting the entire cluster.

rleungx: I prefer to decouple these parameters.

pkg/schedule/checker/affinity_checker.go
rleungx: Ok, how about leaving it to the caller, like the merge checker?

pkg/schedule/checker/checker_controller.go
rleungx: sure

pkg/schedule/checker/affinity_checker.go
lhy1024: need to solve inBoth

pkg/schedule/checker/affinity_checker.go
lhy1024: rename to WithReplaceLeaderStore

pkg/schedule/checker/checker_controller.go
lhy1024: after discussion, we use a affinity-schedule-limit

rleungx (approved):

lhy1024:
@niubell PTAL

PR #10041, opened 2025-12-10
api: add affinity groups api and client support

lhy1024:

Issue Number: Ref #9764

```commit-message
```

### Check List

Tests

- Unit test
-

### Release note

```release-note
None.
```

### DIFF
--- a/server/apiv2/handlers/affinity.go
+++ b/server/apiv2/handlers/affinity.go
@@ -138,7 +138,7 @@ func createAffinityGroups(c *gin.Context) {
 		return
 	}
 
-	changes := make([]affinity.GroupKeyRanges, 0, len(req.AffinityGroups))
+	groups := make([]affinity.GroupKeyRanges, 0, len(req.AffinityGroups))
 	for groupID, input := range req.AffinityGroups {
 		if err := affinity.ValidateGroupID(groupID); err != nil {
 			c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
@@ -164,7 +164,7 @@ func createAffinityGroups(c *gin.Context) {
 			keyRanges = append(keyRanges, kr.toKeyutilKeyRange())
 		}
 
-		changes = append(changes, affinity.GroupKeyRanges{
+		groups = append(groups, affinity.GroupKeyRanges{
 			GroupID:   groupID,
 			KeyRanges: keyRanges,
 		})
@@ -172,12 +172,8 @@ func createAffinityGroups(c *gin.Context) {
 
 	// Create affinity groups with their key ranges
 	// The manager will handle storage persistence, label creation, and in-memory updates atomically
-	if err := manager.CreateAffinityGroups(changes); err != nil {
-		if errs.ErrAffinityGroupContent.Equal(err) {
-			c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
-			return
-		}
-		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+	err = manager.CreateAffinityGroups(groups)
+	if handleAffinityError(c, err) {
 		return
 	}
 
@@ -185,12 +181,12 @@ func createAffinityGroups(c *gin.Context) {
 	resp := AffinityGroupsResponse{
 		AffinityGroups: make(map[string]*affinity.GroupState, len(req.AffinityGroups)),
 	}
-	for _, change := range changes {
-		state := manager.GetAffinityGroupState(change.GroupID)
+	for _, group := range groups {
+		state := manager.GetAffinityGroupState(group.GroupID)
 		if state == nil {
 			state = &affinity.GroupState{}
 		}
-		resp.AffinityGroups[change.GroupID] = state
+		resp.AffinityGroups[group.GroupID] = state
 	}
 	c.IndentedJSON(http.StatusOK, resp)
 }
--- a/tests/server/apiv2/handlers/affinity_test.go
+++ b/tests/server/apiv2/handlers/affinity_test.go
@@ -1150,24 +1150,3 @@ func (suite *affinityHandlerTestSuite) TestAffinityForwardedHeader() {
 		re.Equal("true", resp.Header.Get(apiutil.XForwardedToMicroserviceHeader))
 	})
 }
-
-func TestAffinitySchedulingDisabled(t *testing.T) {
-	re := require.New(t)
-	env := tests.NewSchedulingTestEnvironment(t, func(conf *config.Config, _ string) {
-		conf.Schedule.EnableAffinityScheduling = false
-	})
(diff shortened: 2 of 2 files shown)


client/http/types.go
```diff
@@ -703,3 +703,82 @@ type Health struct {
...
+}
+
+// IsPreparing indicates that the Group is scheduling Regions according to the required Peers.
+func (s *AffinityGroupState) IsPreparing() bool {
+	return s.Phase == "preparing"
+}
+
+// IsStable indicates that the Group has completed the required scheduling and is currently in a stable state.
+func (s *AffinityGroupState) IsStable() bool {
+	return s.Phase == "stable"
```
rleungx, line 753: The phase name is strange, but we can rename it later.

server/apiv2/middlewares/affinity_enabled_checker.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/server"
+)
+
+// AffinitySchedulingEnabledChecker blocks requests when affinity scheduling is disabled.
+func AffinitySchedulingEnabledChecker() gin.HandlerFunc {
```
rleungx: I prefer to check it inside the handler

server/apiv2/handlers/affinity.go
```diff
@@ -0,0 +1,531 @@
...
+
+// RegisterAffinity registers affinity group related handlers to router paths.
+func RegisterAffinity(r *gin.RouterGroup) {
+	router := r.Group("affinity-groups")
+	router.Use(middlewares.BootstrapChecker(), middlewares.AffinitySchedulingEnabledChecker())
+	router.POST("", CreateAffinityGroups)
+	router.PATCH("", BatchModifyAffinityGroups)
+	router.PUT("/:group_id", UpdateAffinityGroupPeers)
+	router.DELETE("/:group_id", DeleteAffinityGroup)
+	router.POST("/batch-delete", BatchDeleteAffinityGroups)
```
rleungx: not meet the RESTful requirement

lhy1024: The primary reason is that we want to perform batch deletions within a single transaction, but the DELETE does not support carrying a body.

And I found that there is also "/config/rules/batch" and "/regions/accelerate-schedule/batch", so I chose a similar path.

server/apiv2/middlewares/affinity_enabled_checker.go
lhy1024: Check it in every affinity handler? I am worried about whether something is missing it

server/apiv2/handlers/affinity.go
lhy1024: I just find an example https://linter.aip.dev/235/http-method

lhy1024: After discussion, use POST /?delete to batch delete https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html

client/http/types.go
okJiang: It is public. Can we rename it in the future?

lhy1024: I think so

okJiang: I means **"Can"** we update it?

client/http/types.go
```diff
@@ -703,3 +703,82 @@ type Health struct {
...
+
+// CreateAffinityGroupInput defines the input for a single group in the creation request.
+type CreateAffinityGroupInput struct {
+	Ranges []AffinityGroupKeyRange `json:"ranges"`
+}
+
+// CreateAffinityGroupsRequest defines the body for the POST request to create affinity groups.
+type CreateAffinityGroupsRequest struct {
+	AffinityGroups map[string]CreateAffinityGroupInput `json:"affinity_groups"`
+	TableGroup     string                              `json:"table_group,omitempty"`
```
okJiang: Where does this input?

HunDunDM: It will be deleted.

server/apiv2/handlers/affinity.go
```diff
@@ -0,0 +1,541 @@
...
+		for _, kr := range input.Ranges {
+			// Validate key range
+			if err := validateKeyRange(kr.StartKey, kr.EndKey); err != nil {
+				c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
+				return
+			}
+			keyRanges = append(keyRanges, kr.toKeyutilKeyRange())
+		}
+
+		changes = append(changes, affinity.GroupKeyRanges{
```
okJiang: typo?

server/apiv2/handlers/affinity.go
```diff
@@ -0,0 +1,541 @@
...
+			GroupID:   groupID,
+			KeyRanges: keyRanges,
+		})
+	}
+
+	// Create affinity groups with their key ranges
+	// The manager will handle storage persistence, label creation, and in-memory updates atomically
+	if err := manager.CreateAffinityGroups(changes); err != nil {
+		if errs.ErrAffinityGroupContent.Equal(err) {
```
okJiang: How about renaming to `ErrAffinityGroupInput`

server/apiv2/handlers/affinity.go
```diff
@@ -0,0 +1,541 @@
...
+			c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
+			return
+		}
+		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	// Convert internal manager output to API output.
+	resp := AffinityGroupsResponse{
+		AffinityGroups: make(map[string]*affinity.GroupState, len(req.AffinityGroups)),
```
okJiang, line 182: Just a question. Is it necessary to return complete Group information?

okJiang: Will we use it?

server/apiv2/handlers/affinity.go
okJiang: use handleAffinityError

server/apiv2/handlers/affinity.go
lhy1024: In previous, this request also use another way. And I forget to update it also. I will change it.

server/apiv2/handlers/affinity.go
lhy1024: ok

server/apiv2/handlers/affinity.go
HunDunDM: In the future, we will be able to include the correct StoreIDs in the return value of Create, which will help tidb update the cache faster.

lhy1024:
@okJiang @rleungx PTAL

PR #10076, opened 2025-12-17
schedule: directly use kind when creating operators

### LINKED ISSUE #10077: Admin operator should not affect other schedule limits
## Enhancement Task
The admin operator is created through the API, which should not be counted and affect the other schedule limits. Currently, the merge and split have different behavior from other operators.

rleungx:

Issue Number: Close #10077

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/handler/handler.go
+++ b/pkg/schedule/handler/handler.go
@@ -608,7 +608,7 @@ func (h *Handler) AddMergeRegionOperator(regionID uint64, targetID uint64) error
 		return errs.ErrRegionNotAdjacent
 	}
 
-	ops, err := operator.CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin)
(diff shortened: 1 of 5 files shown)


lhy1024 (approved):

pkg/schedule/handler/handler.go
```diff
@@ -608,7 +608,7 @@ func (h *Handler) AddMergeRegionOperator(regionID uint64, targetID uint64) error
 		return errs.ErrRegionNotAdjacent
 	}
 
-	ops, err := operator.CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin)
+	ops, err := operator.CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin|operator.OpMerge)
```
lhy1024, line 611: BTW, after #10081, do we still need `operator.OpAdmin|operator.OpMerge`, how about only `operator.OpAdmin`

PR #10080, opened 2025-12-17
affinity: add scatter filter, add more evict check and avoid statistic miss

lhy1024:

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/affinity/manager.go
+++ b/pkg/schedule/affinity/manager.go
@@ -364,17 +364,17 @@ func (m *Manager) getCache(region *core.RegionInfo) (*regionCache, *GroupState) 
 // GetRegionAffinityGroupState returns the affinity group state and isAffinity for a region.
 // This is a read-only operation that does not modify cache and we use this for temporary check
 func (m *Manager) GetRegionAffinityGroupState(region *core.RegionInfo) (group *GroupState, isAffinity bool) {
-	return m.getRegionAffinityGroupState(region, false)
+	return m.calcRegionAffinityGroupState(region, false)
 }
 
 // GetAndCacheRegionAffinityGroupState returns the affinity group state and saves it to cache.
 // The caller must call InvalidCache() when the region is deleted or merged.
 // Currently only used by AffinityChecker.Check.
 func (m *Manager) GetAndCacheRegionAffinityGroupState(region *core.RegionInfo) (group *GroupState, isAffinity bool) {
-	return m.getRegionAffinityGroupState(region, true)
+	return m.calcRegionAffinityGroupState(region, true)
 }
 
-func (m *Manager) getRegionAffinityGroupState(region *core.RegionInfo, saveCache bool) (group *GroupState, isAffinity bool) {
+func (m *Manager) calcRegionAffinityGroupState(region *core.RegionInfo, saveCache bool) (group *GroupState, isAffinity bool) {
 	if region == nil || !m.IsAvailable() {
 		return nil, false
 	}


pkg/schedule/scatter/region_scatterer.go
```diff
@@ -297,6 +300,12 @@ func (r *RegionScatterer) Scatter(region *core.RegionInfo, group string, skipSto
 }
 
 func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, skipStoreLimit bool) (*operator.Operator, error) {
+	// Check if region is in an affinity group that doesn't allow regular scheduling
+	if !r.affinityFilter.Select(region).IsOK() {
+		scatterSkipAffinityCounter.Inc()
+		return nil, errors.Errorf("region %d is in affinity group", region.GetID())
```
rleungx: Will it trigger the retry?

pkg/schedule/filter/region_filters.go
```diff
@@ -200,22 +200,27 @@ func (f *SnapshotSenderFilter) Select(region *core.RegionInfo) *plan.Status {
 }
 
 type affinityFilter struct {
-	affinityManager *affinity.Manager
+	regionSetInformer core.RegionSetInformer
```
rleungx:
```suggestion
	cluster     *core.BasicCluster
```

lhy1024: ok

pkg/schedule/scatter/region_scatterer.go
lhy1024: it should be the same logic with hot region or no leader?
{93998DBC-280B-414B-A779-533C4C9BC672}

lhy1024:
@bufferflies @rleungx PTAL

pkg/schedule/affinity/manager.go
```diff
@@ -369,7 +362,8 @@ func (m *Manager) getCache(region *core.RegionInfo) (*regionCache, *GroupState)
 }
 
 // GetRegionAffinityGroupState returns the affinity group state and isAffinity for a region.
-func (m *Manager) GetRegionAffinityGroupState(region *core.RegionInfo) (group *GroupState, isAffinity bool) {
+// If skipSaveCache is not set to true, InvalidCache must be called at the appropriate time to prevent stale cache entries.
```
rleungx: How about adding comments about when we need to update the cache?

pkg/schedule/affinity/policy.go
```diff
@@ -60,8 +62,10 @@ func (c storeCondition) String() string {
 	switch c {
 	case storeAvailable:
 		return "available"
-	case storeEvictLeader:
+	case storeEvicted:
```
rleungx:
```suggestion
	case storeLeaderEvicted:
```

pkg/schedule/handler/handler.go
```diff
@@ -608,7 +608,7 @@ func (h *Handler) AddMergeRegionOperator(regionID uint64, targetID uint64) error
 		return errs.ErrRegionNotAdjacent
 	}
 
-	ops, err := operator.CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin|operator.OpMerge)
+	ops, err := operator.CreateMergeRegionOperator("admin-merge-region", c, region, target, operator.OpAdmin)
```
bufferflies, line 611: Why remove the OpAdmin type?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -115,8 +115,16 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 	// Recheck after refetching.
 	if needRefetch {
 		c.affinityManager.ObserveAvailableRegion(region, group)
-		group, isAffinity = c.affinityManager.GetRegionAffinityGroupState(region)
+		group, isAffinity = c.affinityManager.GetAndCacheRegionAffinityGroupState(region)
 	}
+
+	// A Region may no longer exist in the RegionTree due to a merge.
+	// In this case, clear the cache in affinity manager for that Region and skip processing it.
+	if c.cluster.GetRegion(region.GetID()) == nil {
```
bufferflies, line 123: Can it be put in line 87?

bufferflies (commented):
rest lgtm

pkg/schedule/handler/handler.go
lhy1024: https://github.com/tikv/pd/pull/10076#discussion_r2626860229

pkg/schedule/checker/affinity_checker.go
lhy1024: Cannot move the check before GetAndCacheRegionAffinityGroupState. 

  Problem with moving it earlier:
  - If check passes, then region is deleted, then HandleOverlaps calls InvalidCache (but cache doesn't 
  exist yet - no-op)
  - Then we save cache → stale cache leak (no cleanup path)

  Current approach guarantees:
  - Save cache first, then check
  - Either HandleOverlaps (when region deleted) OR AffinityChecker (when check fails) will clean up
  - At least one cleanup point always executes

lhy1024:
@bufferflies @rleungx PTAL

PR #10092, opened 2025-12-19
test: distinguish MustHandleStoreHeartbeat

lhy1024:

Issue Number: Ref #4399

```commit-message
```

When we need to test statistics, we need to forward store heartbeat to scheduling server.
If we only test store metadata, we only use store watcher and not need to forward store heartbeat.
### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/testutil.go
+++ b/tests/testutil.go
@@ -288,9 +288,10 @@ func MustPutStore(re *require.Assertions, tc *TestCluster, store *metapb.Store) 
 	)
 	raftCluster.GetBasicCluster().PutStore(newStore)
 	raftCluster.UpdateAllStoreStatus()
-	if tc.GetSchedulingPrimaryServer() != nil {
-		tc.GetSchedulingPrimaryServer().GetCluster().PutStore(newStore)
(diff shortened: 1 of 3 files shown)


bufferflies:
image
You can add more comments here to detail why removing this forwarding.

tests/testutil.go
```diff
@@ -327,16 +326,6 @@ func MustPutRegionInfo(re *require.Assertions, cluster *TestCluster, regionInfo
 func MustHandleStoreHeartbeat(re *require.Assertions, cluster *TestCluster, heartbeat *pdpb.StoreHeartbeatRequest) {
 	err := cluster.GetLeaderServer().GetRaftCluster().HandleStoreHeartbeat(heartbeat, &pdpb.StoreHeartbeatResponse{})
 	re.NoError(err)
-	if cluster.GetSchedulingPrimaryServer() != nil {
```
rleungx, line 330: Why remove it?

lhy1024: We have sync storeheartbeat between mcs and pd, we should not duplicate forward it in test.

After removing it, these tests still are successful.

PR #10102, opened 2025-12-22
tests: close gRPC connections after testing

rleungx:

Issue Number: ref #7969

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/cluster/cluster_test.go
+++ b/tests/server/cluster/cluster_test.go
@@ -974,9 +974,7 @@ func TestConcurrentHandleRegion(t *testing.T) {
 				default:
 					_, err = stream.Recv()
 					if err != nil {
-						isCanceled := strings.Contains(err.Error(), "context canceled")
-						isEOF := strings.Contains(err.Error(), "EOF")
-						re.True(isCanceled || isEOF)
+						re.ErrorContains(err, "code = Canceled")
 					}
 				}
 			}


pkg/utils/testutil/testutil.go
```diff
@@ -80,10 +80,10 @@ func NewRequestHeader(clusterID uint64) *pdpb.RequestHeader {
 }
 
 // MustNewGrpcClient must create a new PD grpc client.
-func MustNewGrpcClient(re *require.Assertions, addr string) pdpb.PDClient {
```
bufferflies, line 83: How about adding a new function as close for pdClient, it will close this conn?

rleungx: Here we just create a gRPC client instead of pd client.

tests/server/cluster/cluster_test.go
```diff
@@ -961,9 +974,7 @@ func TestConcurrentHandleRegion(t *testing.T) {
 				default:
 					_, err = stream.Recv()
 					if err != nil {
-						isCanceled := strings.Contains(err.Error(), "context canceled")
-						isEOF := strings.Contains(err.Error(), "EOF")
```
lhy1024, line 965: why remove EOF?

rleungx: no eof error anymore

PR #10103, opened 2025-12-22
affinity: add stale region cache to avoid cascading merges

### LINKED ISSUE #10104: affinity merge all group region to a region
### What did you do?

What did you do?

In a test environment with affinity groups configured:
1. Started with 60 regions (each ~240MB) that were already stable and aligned to the same affinity group
  (same 3 voter stores)
2. Increased max-affinity-merge-region-size from 256MB to 512MB

Environment characteristics:
- Very small test cluster with fast patrol interval (10ms)
- affinity-schedule-limit set very high
- All regions in the same affinity group with identical voter distribution
- Affinity split prevention enabled 

Image

Regions should merge gradually, two at a time, with proper size updates between merges.

For example: 60 regions → 30 regions

### What did you see instead?

All 60 regions cascaded into a single 14GB region in just 535ms.

Image

From the logs, region 79997 underwent 18 consecutive merges while its ApproximateSize remained stuck at
61MB throughout the entire cascade:

  [2025/12/22 10:59:55.648] First merge created
  [2025/12/22 10:59:55.937] Last merge completed (18 merges total)
  Total time: 289ms for 18 merge operations
  ApproximateSize: 61MB → 61MB → ... → 61MB (no updates during cascade)

Root Cause:
  The cascading merge is caused by stale ApproximateSize values being used for merge decisions:
  1. Timing mismatch:
    - Patrol interval: 10ms
    - Merge operator completion: 10-20ms
    - Region heartbeat (stats update): seconds
  2. Cascade mechanism: Because patrol and operator execution are much faster than heartbeat updates, each merge check uses outdated ApproximateSize (e.g., 61MB) even though the actual region has already grown much larger through previous merges.
  3. Lack of split protection: Unlike regular scenarios where TiKV's auto-split would break the cascade.

  Trigger Conditions:

  This requires all of the following:
  - Very small cluster with fast patrol speed (patrol completes before ApproximateSize updates)
  - Identical voter distribution (enabling very fast operator completion)
  - Many consecutive small regions eligible for merge
  - Affinity split prevention enabled for the key range
  - High affinity-schedule-limit (minor contributing factor)

  Why MergeChecker rarely has this issue:

  1. Split breaks the cascade: TiKV automatically splits at region-split-size (default 256MB), creating a "merge → split → merge" cycle that prevents cascading. Affinity's split prevention removes this safety net.
  2. Different cluster characteristics: Production clusters are typically larger with slower patrol speeds.
  3. Theoretically possible but self-healing: MergeChecker could theoretically cascade under similar extreme conditions, but without split prevention, TiKV's auto-split would quickly recover the situation.

 release-8.5

lhy1024:

Issue Number: Ref https://github.com/tikv/pd/issues/9764 Close #10104

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/checker/affinity_checker.go
+++ b/pkg/schedule/checker/affinity_checker.go
@@ -466,12 +466,12 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
 	return region.Clone(options...)
 }
 
-// RecordMergeOpSuccess is called when an operator completes successfully.
+// RecordOpSuccess is called when an operator completes successfully.
 // It caches merged regions to prevent immediate re-merging.
 //
 // Merge completes quickly (e.g., 21ms) but TiKV needs time to update approximate_size via split-check.
 // During this gap, PD may use stale size data to schedule more merges. This 1-minute cache prevents that.
-func (c *AffinityChecker) RecordMergeOpSuccess(op *operator.Operator) {
+func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
 	// if schedule limit is 0, disable schedule, so we don't need to cache it.
 	if c.conf.GetAffinityScheduleLimit() == 0 {
 		return
--- a/pkg/schedule/coordinator.go
+++ b/pkg/schedule/coordinator.go
@@ -91,7 +91,7 @@ func NewCoordinator(parentCtx context.Context, cluster sche.ClusterInformer, hbS
 	opController.SetSuccessCallbacks(
 		func(op *operator.Operator) {
 			if checker := checkers.GetAffinityChecker(); checker != nil {
-				checker.RecordMergeOpSuccess(op)
+				checker.RecordOpSuccess(op)
 			}
 		},
 	)
--- a/pkg/schedule/checker/affinity_checker_test.go
+++ b/pkg/schedule/checker/affinity_checker_test.go
@@ -2466,7 +2466,7 @@ func TestAffinityMergeCheckWithOperatorController(t *testing.T) {
 
 	// Phase 2: Manually invoke the callback to simulate first merge completion (region1+region2)
 	for _, op := range ops1 {
-		affinityChecker.RecordMergeOpSuccess(op)
+		affinityChecker.RecordOpSuccess(op)
 	}
 	re.True(affinityChecker.recentMergeCache.Exists(region1.GetID()),
 		"Region1 should be cached after callback")


pkg/schedule/checker/affinity_checker.go
```diff
@@ -437,3 +465,33 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
...
+	cached, ok := c.staleRegionsCache.Get(region.GetID())
+	if !ok {
+		return false
+	}
+	stats, ok := cached.(staleCheckStats)
+	if !ok {
+		c.staleRegionsCache.Remove(region.GetID())
+		return false
+	}
+	if region.GetApproximateSize() > stats.size || region.GetApproximateKeys() > stats.keys {
```
rleungx: What if we keep writing the region, but the merge is not finished?

lhy1024: If there is a merge op, we will not check this region.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -307,6 +326,15 @@ func (c *AffinityChecker) mergeCheck(region *core.RegionInfo, group *affinity.Gr
...
+	c.staleRegionsCache.PutWithTTL(region.GetID(), staleCheckStats{
+		size: region.GetApproximateSize(),
+		keys: region.GetApproximateKeys(),
+	}, affinityStaleCheckTTL)
+	c.staleRegionsCache.PutWithTTL(target.GetID(), staleCheckStats{
+		size: target.GetApproximateSize(),
+		keys: target.GetApproximateKeys(),
+	}, affinityStaleCheckTTL)
+
 	affinityMergeCheckerNewOpCounter.Inc()
```
bufferflies, line 338: We should update the cache after the operator has been put into the operator controller, to ensure the operator runs.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -34,21 +37,29 @@ import (
 	"github.com/tikv/pd/pkg/utils/logutil"
 )
 
+const (
+	// We use region heartbeat interval as TTL for stale regions
+	affinityStaleCheckTTL        = time.Minute
+	affinityStaleCheckGCInterval = time.Minute
```
bufferflies: 1 min may be too small if the region heartbeat pending is high?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -307,6 +326,15 @@ func (c *AffinityChecker) mergeCheck(region *core.RegionInfo, group *affinity.Gr
 		return nil
 	}
 
+	c.staleRegionsCache.PutWithTTL(region.GetID(), staleCheckStats{
+		size: region.GetApproximateSize(),
+		keys: region.GetApproximateKeys(),
+	}, affinityStaleCheckTTL)
+	c.staleRegionsCache.PutWithTTL(target.GetID(), staleCheckStats{
+		size: target.GetApproximateSize(),
```
bufferflies: Generally, the target region size should be the sum of the source and target region.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -437,3 +466,26 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
 
 	return region.Clone(options...)
 }
+
+// RecordOpComplete is called when an operator completes successfully.
+// It caches merged regions to prevent immediate re-merging.
+//
+// Merge completes quickly (e.g., 21ms) but TiKV needs time to update approximate_size via split-check.
+// During this gap, PD may use stale size data to schedule more merges. This 1-minute cache prevents that.
+func (c *AffinityChecker) RecordOpComplete(op *operator.Operator) {
```
HunDunDM:
```suggestion
func (c *AffinityChecker) RecordMergeOpComplete(op *operator.Operator) {
```

pkg/schedule/coordinator.go
```diff
@@ -86,6 +86,12 @@ func NewCoordinator(parentCtx context.Context, cluster sche.ClusterInformer, hbS
 	opController := operator.NewController(ctx, cluster.GetBasicCluster(), cluster.GetSharedConfig(), hbStreams)
 	schedulers := schedulers.NewController(ctx, cluster, cluster.GetStorage(), opController)
 	checkers := checker.NewController(ctx, cluster, cluster.GetCheckerConfig(), opController)
+
+	// Set the callback function for recording operator completion
+	opController.SetRecordOpComplete(func(op *operator.Operator) {
+		checkers.GetAffinityChecker().RecordOpComplete(op)
```
HunDunDM:
```suggestion
		if op.HasRelatedMergeRegion()
			checkers.GetAffinityChecker().RecordMergeOpComplete(op)
		}
```

lhy1024:
```
func (c *AffinityChecker) RecordOpComplete(op *operator.Operator) {
	// Process both merge and affinity merge operators
	if !op.HasRelatedMergeRegion() {
		return
	}
```

lhy1024:
Before this pr 
image
image

After this pr, it became slow 
image
image
image

pkg/schedule/checker/affinity_checker.go
```diff
@@ -437,3 +466,26 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
...
 }
+
+// RecordMergeOpComplete is called when an operator completes successfully.
+// It caches merged regions to prevent immediate re-merging.
+//
+// Merge completes quickly (e.g., 21ms) but TiKV needs time to update approximate_size via split-check.
+// During this gap, PD may use stale size data to schedule more merges. This 1-minute cache prevents that.
+func (c *AffinityChecker) RecordMergeOpComplete(op *operator.Operator) {
+	// Process both merge and affinity merge operators
+	if !op.HasRelatedMergeRegion() {
```
rleungx, line 480: Is the merge also recorded?

lhy1024: Yes,  the merge checker is also able to merge regions in affinity group.

pkg/schedule/operator/operator_controller.go
```diff
@@ -120,6 +125,11 @@ func NewController(ctx context.Context, cluster *core.BasicCluster, config confi
 	}
 }
 
+// SetRecordOpComplete sets the callback function for recording operator completion.
+func (oc *Controller) SetRecordOpComplete(fn func(op *Operator)) {
```
rleungx: How about changing to successCallbacks?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -34,21 +38,27 @@ import (
...
 type AffinityChecker struct {
 	PauseController
-	cluster         sche.CheckerCluster
-	affinityManager *affinity.Manager
-	conf            config.CheckerConfigProvider
+	cluster          sche.CheckerCluster
+	affinityManager  *affinity.Manager
+	conf             config.CheckerConfigProvider
+	recentMergeCache *cache.TTLUint64
+	startTime        time.Time
```
bufferflies, line 50: Does it only set this value in the construct?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -437,3 +466,30 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
...
+func (c *AffinityChecker) RecordMergeOpComplete(op *operator.Operator) {
+	// if schedule limit is 0, disable schedule, so we don't need to cache it.
+	if c.conf.GetAffinityScheduleLimit() == 0 {
+		return
+	}
+	// Process both merge and affinity merge operators
+	if !op.HasRelatedMergeRegion() {
+		return
+	}
+	relatedIDStr := op.GetAdditionalInfo(string(operator.RelatedMergeRegion))
```
bufferflies: How about extracting a new function?

pkg/schedule/operator/operator_controller.go
lhy1024: successCallbacks and AddSuccessCallback？

pkg/schedule/checker/affinity_checker.go
lhy1024: Yes, consider pd transfer leader, so we also pause schedule 1 min

lhy1024:
@bufferflies PTAL

pkg/schedule/checker/affinity_checker.go
```diff
@@ -34,21 +37,27 @@ import (
...
-func NewAffinityChecker(cluster sche.CheckerCluster, conf config.CheckerConfigProvider) *AffinityChecker {
+func NewAffinityChecker(ctx context.Context, cluster sche.CheckerCluster, conf config.CheckerConfigProvider) *AffinityChecker {
 	return &AffinityChecker{
-		cluster:         cluster,
-		affinityManager: cluster.GetAffinityManager(),
-		conf:            conf,
+		cluster:          cluster,
+		affinityManager:  cluster.GetAffinityManager(),
+		conf:             conf,
+		recentMergeCache: cache.NewIDTTL(ctx, recentMergeTTL, recentMergeTTL),
```
HunDunDM: how about `cache.NewIDTTL(ctx, gcInterval, recentMergeTTL)`

pkg/schedule/coordinator.go
```diff
@@ -86,6 +86,16 @@ func NewCoordinator(parentCtx context.Context, cluster sche.ClusterInformer, hbS
 	opController := operator.NewController(ctx, cluster.GetBasicCluster(), cluster.GetSharedConfig(), hbStreams)
 	schedulers := schedulers.NewController(ctx, cluster, cluster.GetStorage(), opController)
 	checkers := checker.NewController(ctx, cluster, cluster.GetCheckerConfig(), opController)
+
+	// Set the callbacks for operator success
+	opController.SetSuccessCallbacks(
```
HunDunDM, line 91: Can we check for nil values ​​during the adding process?

pkg/schedule/checker/affinity_checker.go
```diff
@@ -437,3 +465,25 @@ func cloneRegionWithReplacePeerStores(region *core.RegionInfo, leaderStoreID uin
 
 	return region.Clone(options...)
 }
+
+// RecordMergeOpSuccess is called when an operator completes successfully.
+// It caches merged regions to prevent immediate re-merging.
+//
+// Merge completes quickly (e.g., 21ms) but TiKV needs time to update approximate_size via split-check.
+// During this gap, PD may use stale size data to schedule more merges. This 1-minute cache prevents that.
+func (c *AffinityChecker) RecordMergeOpSuccess(op *operator.Operator) {
```
HunDunDM:
```suggestion
func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
```

HunDunDM (approved):

PR #10105, opened 2025-12-23
client: enable router client by default

### LINKED ISSUE #8690: Support batch processing with the GetRegion stream call

rleungx:

Issue Number: Close #8690

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/client/clients/router/client.go
+++ b/client/clients/router/client.go
@@ -258,8 +258,16 @@ func requestFinisher(resp *pdpb.QueryRegionResponse) batch.FinisherFunc[*Request
 		requestCtx := req.requestCtx
 		defer trace.StartRegion(requestCtx, "pdclient.regionReqDone").End()
 
+		// If there's an error, pass it to the request
 		if err != nil {
 			req.tryDone(err)
+			return
+		}
+
+		// If resp is nil but no error was provided, it means an abnormal situation occurred
+		// (e.g., timeout, connection issue). We should pass an error to indicate this.
+		if resp == nil {
+			req.tryDone(errs.ErrClientRouterConnectionTimeout)
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


bufferflies (approved):

client/clients/router/client.go
```diff
@@ -258,11 +258,19 @@ func requestFinisher(resp *pdpb.QueryRegionResponse) batch.FinisherFunc[*Request
...
+		// If there's an error, pass it to the request
 		if err != nil {
 			req.tryDone(err)
 			return
 		}
 
+		// If resp is nil but no error was provided, it means an abnormal situation occurred
+		// (e.g., timeout, connection issue). We should pass an error to indicate this.
+		if resp == nil {
+			req.tryDone(errs.ErrClientRouterConnectionTimeout)
```
JmPotato, line 270: Will this error be vague or misleading?

rleungx: Any better idea?

JmPotato (approved):

rleungx:
This feature is not compatible with the circuit breaker and forwarding under network isolation. /cc @JmPotato

PR #10113, opened 2025-12-24
tools: show the range of keyspace with a given id or name

rleungx:

Issue Number: ref #10122

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tools/pd-ctl/pdctl/command/keyspace_command.go
+++ b/tools/pd-ctl/pdctl/command/keyspace_command.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/tikv/pd/client/constants"
 	"github.com/tikv/pd/pkg/keyspace"
 	"github.com/tikv/pd/server/apiv2/handlers"
 )
@@ -350,10 +351,16 @@ func showKeyspaceRangeByIDCommandFunc(cmd *cobra.Command, args []string) {
 		return
 	}
 
-	keyspaceID := args[0]
-	id, err := strconv.ParseUint(keyspaceID, 10, 32)
+	id, err := strconv.ParseUint(args[0], 10, 32)
 	if err != nil {
 		cmd.PrintErrln("keyspace id should be a valid number: ", err)
+		return
+	}
+
+	keyspaceID := uint32(id)
+	if keyspaceID < constants.DefaultKeyspaceID || keyspaceID > constants.MaxKeyspaceID {
+		cmd.PrintErrf("invalid keyspace id %d. It must be in the range of [%d, %d]\n",
+			keyspaceID, constants.DefaultKeyspaceID, constants.MaxKeyspaceID)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -317,3 +321,129 @@ func listKeyspaceCommandFunc(cmd *cobra.Command, args []string) {
...
+		Run:   showKeyspaceRangeByIDCommandFunc,
+	}
+	rangeByID.Flags().Bool("raw", false, "show raw key range (default: txn key range)")
+
+	rangeByName := &cobra.Command{
+		Use:   "name <keyspace_name>",
+		Short: "show key ranges for the given keyspace name",
+		Run:   showKeyspaceRangeByNameCommandFunc,
+	}
+	rangeByName.Flags().Bool("raw", false, "show raw key range (default: txn key range)")
```
bufferflies: Can we put this flag into the r pointer? All the sub command needs this flag.

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -317,3 +321,129 @@ func listKeyspaceCommandFunc(cmd *cobra.Command, args []string) {
...
+	bound := keyspace.MakeRegionBound(uint32(id))
+	if raw {
+		ranges = []any{
+			map[string]any{
+				"start_key": hex.EncodeToString(bound.RawLeftBound),
+				"end_key":   hex.EncodeToString(bound.RawRightBound),
+			},
+		}
+	} else {
+		ranges = []any{
```
bufferflies: Why must ranges be an array?

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -317,3 +321,119 @@ func listKeyspaceCommandFunc(cmd *cobra.Command, args []string) {
...
+	return r
+}
+
+func showKeyspaceRangeByIDCommandFunc(cmd *cobra.Command, args []string) {
+	if len(args) != 1 {
+		cmd.Usage()
+		return
+	}
+
+	keyspaceID := args[0]
```
okJiang: Do we need check id exists?

bufferflies (approved):

tools/pd-ctl/pdctl/command/keyspace_command.go
rleungx: We can check if the id is valid, but no need to check if exists?

okJiang: I think it would be a good idea to check it out, it doesn't cost any resources

rleungx: No, it's just a kind of way to get the key range, like the tso command.

okJiang (approved):

PR #10115, opened 2025-12-24
config: add option to skip rawkv region bound

rleungx:

Issue Number: Close #xxx

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/keyspace/util.go
+++ b/pkg/keyspace/util.go
@@ -126,9 +126,17 @@ func MakeRegionBound(id uint32) *RegionBound {
 	}
 }
 
-// MakeKeyRanges encodes keyspace ID to correct LabelRule data.
-func MakeKeyRanges(id uint32) []any {
+// makeKeyRanges encodes keyspace ID to correct LabelRule data.
+func makeKeyRanges(id uint32, skipRaw bool) any {
 	regionBound := MakeRegionBound(id)
+	if skipRaw {
+		return []any{
+			map[string]any{
+				"start_key": hex.EncodeToString(regionBound.TxnLeftBound),
+				"end_key":   hex.EncodeToString(regionBound.TxnRightBound),
+			},
+		}
+	}
 	return []any{
 		map[string]any{
 			"start_key": hex.EncodeToString(regionBound.RawLeftBound),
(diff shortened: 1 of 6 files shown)


server/config/config.go
```diff
@@ -877,6 +878,8 @@ type KeyspaceConfig struct {
 	WaitRegionSplit bool `toml:"wait-region-split" json:"wait-region-split"`
 	// WaitRegionSplitTimeout indicates the max duration to wait region split.
 	WaitRegionSplitTimeout typeutil.Duration `toml:"wait-region-split-timeout" json:"wait-region-split-timeout"`
+	// DisableRawKVRegionSplit indicates whether to skip raw kv region split.
+	DisableRawKVRegionSplit bool `toml:"disable-raw-kv-region-split" json:"disable-raw-kv-region-split,string"`
```
bufferflies, line 882: pls add some comment about this configuration only work in keyspace,

server/config/config.go
```diff
@@ -245,6 +245,7 @@ const (
 	defaultGCTunerThreshold           = 0.6
 	minGCTunerThreshold               = 0
 	maxGCTunerThreshold               = 0.9
+	defaultDisableRawKVRegionSplit    = false
```
bufferflies, line 248: I think it should be true, most case user don't use raw kv in keyspace clustger

bufferflies: If the user cluster only uses the raw api, how to avoid splitting the txn key?

pkg/keyspace/util.go
```diff
@@ -126,9 +126,17 @@ func MakeRegionBound(id uint32) *RegionBound {
 	}
 }
 
-// MakeKeyRanges encodes keyspace ID to correct LabelRule data.
-func MakeKeyRanges(id uint32) []any {
+// makeKeyRanges encodes keyspace ID to correct LabelRule data.
+func makeKeyRanges(id uint32, skipRaw bool) any {
 	regionBound := MakeRegionBound(id)
+	if skipRaw {
```
lhy1024, line 132: But waitKeyspaceRegionSplit will call CheckKeyspaceRegionBound, and it always checks raw bound and txn bound
```
func (manager *Manager) CheckKeyspaceRegionBound(id uint32) bool {
	regionBound := MakeRegionBound(id)
	return manager.checkBound(regionBound.RawLeftBound) &&
		manager.checkBound(regionBound.RawRightBound) &&
		manager.checkBound(regionBound.TxnLeftBound) &&
		manager.checkBound(regionBound.TxnRightBound)
}
```

PR #10119, opened 2025-12-25
*: move initialization to coordinator

### LINKED ISSUE #10118: Load rules should not block leader election
## Enhancement Task
Before a PD becomes the leader, it will do some initialization work, see

https://github.com/tikv/pd/blob/9fe5653625628f54f7fe971b470a9969e63d8633/pkg/schedule/labeler/labeler.go#L54

https://github.com/tikv/pd/blob/9fe5653625628f54f7fe971b470a9969e63d8633/pkg/schedule/placement/rule_manager.go#L101-L106

If one of them has too many rules, the election will be blocked. It can significantly increase the unavailable time. One possible solution is moving these kinds of jobs to the coordinator. We can also guarantee that they are loaded before the scheduling job starts.

rleungx:

Issue Number: Close #10118.

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/schedule/coordinator.go
+++ b/pkg/schedule/coordinator.go
@@ -240,15 +240,15 @@ func (c *Coordinator) RunUntilStop(skipLoadRules bool) {
 		}
 		err := c.ruleManager.Initialize(c.config.GetMaxReplicas(), c.config.GetLocationLabels(), c.config.GetIsolationLevel(), skipLoadRules)
 		if err != nil {
-			log.Error("rule manager initialize failed", errs.ZapError(err))
+			log.Warn("rule manager initialize failed", errs.ZapError(err))
 			continue
 		}
 		if err := c.regionLabeler.Initialize(regionLabelGCInterval); err != nil {
-			log.Error("region labeler initialize failed", errs.ZapError(err))
+			log.Warn("region labeler initialize failed", errs.ZapError(err))
 			continue
 		}
 		if err := c.affinityManager.Initialize(); err != nil {
-			log.Error("affinity manager initialize failed", errs.ZapError(err))
+			log.Warn("affinity manager initialize failed", errs.ZapError(err))
 			continue
 		}
 		break
--- a/pkg/schedule/coordinator.go
+++ b/pkg/schedule/coordinator.go
@@ -28,12 +28,15 @@ import (
 
 	"github.com/tikv/pd/pkg/core"
 	"github.com/tikv/pd/pkg/errs"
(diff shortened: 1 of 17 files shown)


server/cluster/cluster.go
```diff
@@ -371,17 +369,8 @@ func (c *RaftCluster) Start(s Server, bootstrap bool) (err error) {
...
-	}
 
-	c.regionLabeler, err = labeler.NewRegionLabeler(c.ctx, c.storage, regionLabelGCInterval)
-	if err != nil {
-		return err
-	}
+	c.regionLabeler = labeler.NewRegionLabeler(c.ctx, c.storage)
 
 	// create affinity manager with region labeler for key range validation and rebuild
 	c.affinityManager, err = affinity.NewManager(c.ctx, c.storage, c, c.GetOpts(), c.regionLabeler)
```
bufferflies, line 376: Btw, how about adding a new log to show the time cost of this start?

pkg/schedule/coordinator.go
```diff
@@ -217,7 +230,30 @@ func (c *Coordinator) driveSlowNodeScheduler() {
...
-func (c *Coordinator) RunUntilStop() {
+func (c *Coordinator) RunUntilStop(skipLoadRules bool) {
+	for {
+		select {
+		case <-c.ctx.Done():
+			log.Info("coordinator initialization is stopped")
+			return
+		default:
+		}
+		err := c.ruleManager.Initialize(c.config.GetMaxReplicas(), c.config.GetLocationLabels(), c.config.GetIsolationLevel(), skipLoadRules)
```
lhy1024, line 241: For microservice, the PD server will not run the Coordinator, so these managers will not be initialized.

pkg/schedule/coordinator.go
```diff
@@ -217,7 +230,30 @@ func (c *Coordinator) driveSlowNodeScheduler() {
 }
 
 // RunUntilStop runs the coordinator until receiving the stop signal.
-func (c *Coordinator) RunUntilStop() {
+func (c *Coordinator) RunUntilStop(skipLoadRules bool) {
+	for {
```
lhy1024, line 234: We initialize three managers in a for loop, and there are two goroutines in regionLabeler.Initialize and affinityManager.Initialize. If we meet an error, will there be a goroutine leak?

PR #10120, opened 2025-12-25
core: region heartbeat with bucket meta

### LINKED ISSUE #10117: Region heartbeat should carry with the region bucket  meta
## Enhancement Task
Currently, the region bucket meta update depends on the bucket report stream, which incurs additional costs and introduces data race risks.

bufferflies:

Issue Number: Close #10117

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Regions now support bucket metadata handling, enabling improved synchronization of bucket information from cluster heartbeats. Bucket metadata is now preserved during region initialization and inheritance operations.

* **Dependencies**
  * Updated core protobuf dependencies across multiple modules to support enhanced region metadata capabilities.


pkg/syncer/client.go
```diff
@@ -246,6 +246,7 @@ func (s *RegionSyncer) StartSyncWithLeader(addr string) {
 					if hasBuckets {
 						if old := origin.GetBuckets(); buckets[i].GetVersion() > old.GetVersion() {
 							region.UpdateBuckets(buckets[i], old)
```
rleungx, line 248: If we use SetBucketMeta, do we still need UpdateBuckets?

bufferflies: No need, but currently, we must keep compatibility; we need to update the bucket also. We can remove this update in a higher version.

pkg/syncer/client.go
lhy1024, line 244: We firstly call putRegion, and then UpdateBuckets, will this order introduce data race?

pkg/core/region.go
```diff
@@ -85,11 +85,14 @@ type RegionInfo struct {
...
 	flowRoundDivisor          uint64
 	// buckets is not thread unsafe, it should be accessed by the request `report buckets` with greater version.
+	// todo: keep it compatible with previous design, we can remove it later.
 	buckets unsafe.Pointer
 	// source is used to indicate region's source, such as Storage/Sync/Heartbeat.
 	source RegionSource
 	// ref is used to indicate the reference count of the region in root-tree and sub-tree.
 	ref atomic.Int32
+	// bucketMeta is used to store the bucket meta reported by tikv.
+	bucketMeta *metapb.BucketMeta
```
lhy1024, line 95: If buckets and bucketMeta are both nil, which one will we use?

bufferflies: It will return nil.

pkg/syncer/client.go
bufferflies: No, UpdateBuckets will only update the point, not affect the bucket meta.

pkg/core/region.go
lhy1024: What if either buckets or bucketMeta is nil?

rleungx:
@coderabbitai full review

pkg/syncer/client.go
```diff
@@ -246,6 +246,7 @@ func (s *RegionSyncer) StartSyncWithLeader(addr string) {
...
+							region.SetBucketMeta(buckets[i])
 						}
```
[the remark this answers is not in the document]
bufferflies: ignore

lhy1024 (approved):

PR #10134, opened 2026-01-07
tests: fix some unstable tests

rleungx:

Issue Number: ref #7969

Fix the following tests:

- TestSet
- TestGetAllKeyspaceGCStates
- TestRemoveStoreLimit
- TestStoresList

```commit-message
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/api/store_test.go
+++ b/tests/server/api/store_test.go
@@ -72,9 +72,13 @@ func (suite *storeTestSuite) checkStoresList(cluster *tests.TestCluster) {
 		tests.MustPutStore(re, cluster, store)
 	}
 
-	// Prevent store 6 (offline store) from being auto-tombstoned by adding a region to it
-	// This ensures the store remains offline instead of being converted to tombstone
-	tests.MustPutRegion(re, cluster, 999, 6, []byte("a"), []byte("b"))
+	// Prevent store 6 (offline store) from being auto-tombstoned
+
+	// Enable failpoint to prevent auto-burying stores during the test
+	re.NoError(failpoint.Enable("github.com/tikv/pd/server/cluster/doNotBuryStore", "return(true)"))
+	defer func() {
+		re.NoError(failpoint.Disable("github.com/tikv/pd/server/cluster/doNotBuryStore"))
+	}()
 
 	leader := cluster.GetLeaderServer()
 	urlPrefix := leader.GetAddr() + "/pd/api/v1"


pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -862,31 +862,67 @@ func TestRemoveStoreLimit(t *testing.T) {
...
+		region = tc.GetRegion(i)
+		op = operator.NewTestOperator(region.GetID(), region.GetRegionEpoch(), operator.OpRegion)
+		op.SetPriorityLevel(constant.Low)
+		re.True(oc.AddOperator(op))
+	}
+	testutil.Eventually(re, func() bool {
+		op, err = scatterer.Scatter(tc.GetRegion(1), "", true)
+		re.NoError(err)
+		return op != nil
+	})
```
liyishuai: Here are two changes:
1. Adding 5 low-priority operators instead of 1.
2. Check the Scatter result eventually instead of immediately.

Is (1) an improvement, while (2) is the actual fix?

pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -862,31 +862,67 @@ func TestRemoveStoreLimit(t *testing.T) {
...
+	tc := mockcluster.NewCluster(ctx, opt)
+	stream := hbstream.NewTestHeartbeatStreams(ctx, tc, false)
+	oc := operator.NewController(ctx, tc.GetBasicCluster(), tc.GetSharedConfig(), stream)
+
+	// Add stores 1~5.
+	for i := uint64(1); i <= 5; i++ {
+		tc.AddRegionStore(i, 0)
+	}
+	// Add regions 1~5.
+	// Region 1 has the same distribution with the Region 2, which is used to test selectPeerToReplace.
```
liyishuai: All five regions seem to have the same distribution, not just Region 1 same as Region 2.

pkg/schedule/scatter/region_scatterer_test.go
rleungx: fixed

pkg/schedule/scatter/region_scatterer_test.go
rleungx: fixed

liyishuai (approved):

okJiang (commented):
can you post related issue? Or specific test case name here?

rleungx:
> can you post related issue? Or specific test case name here?

You can check the commit message.

rleungx:
> can you post related issue? Or specific test case name here?

I also added it to the PR description.

PR #10137, opened 2026-01-08
tso:update tso timestamp when met logic exhaust

### LINKED ISSUE #10138: tso: tso can't allocate if met goroutine scheduler busy
## Enhancement Task
Currently, tso allocater only uses one goroutine to advance the physical timestamp, but it's not reliable if the Go runtime doesn't wake up this goroutine. 
https://github.com/tikv/pd/blob/2a1151063e7674d6b79b1775201b60cdf07ac64b/pkg/tso/allocator.go#L126-L144

bufferflies:

Issue Number: Close #10138

```commit-message
1. Advance the physical timestamp by the GetTs request, not only by the ticker.
2. Only the ticker can save the timestamp into the storage when it reaches the save time to avoid frequent updating. 
```

### Check List

Tests

- Unit test
- manual test

```
tiup playground nightly --pd 1 --kv 1 --tiflash 0 --db 0 --pd.host 10.138.0.52 --pd.config pd.toml --pd.binpath ./pd/bin/pd-server
./bin/pd-tso-bench -pd 10.138.0.52:2379 -client 500 -duration 10m -c 10
```

Master(tso-update-physical-interval = "2s"):
image

This PR(tso-update-physical-interval = "2s")
image

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Added an observability metric to surface when saving timestamps is disallowed.

* **Bug Fixes**
  * Retry logging for timestamp updates now includes an overflow indicator and returns the final attempt outcome.
  * Timestamp persistence is gated to avoid unintended physical updates.
  * Improved synchronization to reduce unnecessary physical timestamp changes.

* **Tests**
  * Added tests for timestamp generation, guarded-save behavior, logical overflow, and concurrent access.


pkg/tso/tso.go
```diff
@@ -90,13 +91,18 @@ func (t *timestampOracle) saveTimestamp(ts time.Time) error {
...
 	// Do not update the zero physical time if the `force` flag is false.
 	if t.tsoMux.physical.Equal(typeutil.ZeroTime) && !force {
 		return
 	}
+	// If current physical time is bigger than the previous time, it indicates that the physical time has been updated.
+	if t.tsoMux.physical.After(pre) {
+		return
+	}
 	// make sure the ts won't fall back
 	if typeutil.SubTSOPhysicalByWallClock(next, t.tsoMux.physical) > 0 {
```
liyishuai: What's the difference between the two comparisons?

bufferflies: The first comparison ensures that the t.tsoMux.physical doesn't change when getting Tso .
The second comparison ensures that the logical physical can't rollback

pkg/tso/tso.go
```diff
@@ -378,7 +393,7 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
 			return pdpb.Timestamp{}, errs.ErrGenerateTimestamp.FastGenByArgs("timestamp in memory isn't initialized")
 		}
 		// Get a new TSO result with the given count
-		resp.Physical, resp.Logical = t.generateTSO(ctx, int64(count))
+		resp.Physical, resp.Logical, physical = t.generateTSO(ctx, int64(count))
```
liyishuai: `generateTSO` returns `physical, logical, current`, and here they become `resp.Physical, resp.Logical, physical`. Please standardize the namings.

pkg/tso/tso.go
liyishuai: I'm confused why they use different functions for comparison. Is it worth unifying the code style?

pkg/tso/tso.go
bufferflies: good, add comment in the function.

pkg/tso/tso.go
```diff
@@ -388,6 +402,9 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
+			if err := t.updateTimestamp(physical, false); err != nil && !errs.ErrUpdateTimestamp.Equal(err) {
```
rleungx: update but not persist?

bufferflies: No, update memory, physical memory, save storage.

rleungx: allowSaveStorage is false here

pkg/tso/tso.go
okJiang: Maybe
```suggestion
		resp.Physical, resp.Logical, current = t.generateTSO(ctx, int64(count))
```

pkg/tso/tso_test.go
```diff
@@ -0,0 +1,129 @@
...
+	timestampOracle.lastSavedTime.Store(current.Add(runDuration))
+
+	wg := sync.WaitGroup{}
+	concurrency := 20
+	wg.Add(concurrency)
+	changes := atomic.Int32{}
+	totalTso := atomic.Int32{}
+	for i := range concurrency {
+		go func(i int) {
+			pre, _ := timestampOracle.getTSO()
```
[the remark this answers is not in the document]
liyishuai: I'm confused as well. What is `pre` and what is `changes`?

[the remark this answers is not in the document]
bufferflies: getTSO returns time.Time, int64, not contains an error

pkg/tso/tso.go
bufferflies: done

liyishuai (approved):

pkg/tso/tso.go
```diff
@@ -90,13 +90,18 @@ func (t *timestampOracle) saveTimestamp(ts time.Time) error {
 	return t.storage.SaveTimestamp(ctx, t.keyspaceGroupID, ts, t.member.GetLeadership())
 }
 
-func (t *timestampOracle) setTSOPhysical(next time.Time, force bool) {
+// setTSOPhysical sets the TSO's physical part with the given time.
+func (t *timestampOracle) setTSOPhysical(old time.Time, next time.Time, force bool) {
```
JmPotato: Maybe this function needs to be renamed to emphasize “compare and set” more clearly, like `compareAndSetTSOPhysical`.

pkg/tso/tso.go
```diff
@@ -388,6 +404,9 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
+			if err := t.updateTimestamp(current, false); err != nil && !errs.ErrUpdateTimestamp.Equal(err) {
```
JmPotato: Although `ErrUpdateTimestamp` is currently generated only when saving is not allowed, I believe it is inappropriate to use the broad category of errors that `updateTimestamp` might return as the comparison condition.

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso.go
bufferflies: yes, I have added new metrics here, so it can return nil, not an error, to oberver the update failed

pkg/tso/tso.go
```diff
@@ -287,7 +297,7 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
 //
 // NOTICE: this function should be called after the TSO in memory has been initialized
 // and should not be called when the TSO in memory has been reset anymore.
-func (t *timestampOracle) updateTimestamp() error {
+func (t *timestampOracle) updateTimestamp(pre time.Time, allowSaveStorage bool) error {
```
JmPotato: Maybe we can change `pre` to a boolean value. When needed, we can get the current TSO once at the beginning inside the function; otherwise, we would have to manually get and pass it in before every call, which seems a bit redundant.

bufferflies: If we get the physical time in this function rather than in generateTso, the physical time may have been updated, triggering an update again.

JmPotato: As long as `updateTimestamp` is not whole protected by the lock, then there is no difference to pass in a `pre` or get one at the very first beginning.

bufferflies: done, only update when met logical overflow

pkg/tso/tso.go
```diff
@@ -90,17 +90,25 @@ func (t *timestampOracle) saveTimestamp(ts time.Time) error {
 	return t.storage.SaveTimestamp(ctx, t.keyspaceGroupID, ts, t.member.GetLeadership())
 }
 
-func (t *timestampOracle) setTSOPhysical(next time.Time, force bool) {
+// setTSOPhysical sets the TSO's physical part with the given time.
+// If the `allowSaveStorage` flag is true, it will always update the physical time.
+// else, it will only update when the logical part is used up.
+func (t *timestampOracle) setTSOPhysical(next time.Time, force bool, allowSaveStorage bool) {
```
JmPotato: From the perspective of `setTSOPhysical`, the name of the variable `allowSaveStorage` is inappropriate. The concept from the upper-level logic might be passed in under a different name.

bufferflies: Yes, I have refactored this setTSOPhysical function, it is clearer

```
// setTSOOption defines the option type for setTSOPhysical function.
// if it returns true, it will skip update physical time.
type setTSOOption func(t *timestampOracle) bool

func withInitialCheck() setTSOOption {
	return func(t *timestampOracle) bool {
		return t.tsoMux.physical.Equal(typeutil.ZeroTime)
	}
}

func withLogicalOverflowCheck() setTSOOption {
	return func(t *timestampOracle) bool {
		return t.tsoMux.logical < maxLogical
	}
}

// setTSOPhysical sets the TSO's physical part with the given time.
// If the `allowSaveStorage` flag is true, it will always update the physical time.
// else, it will only update when the logical part is used up.
func (t *timestampOracle) setTSOPhysical(next time.Time, opts ...setTSOOption) {
	t.tsoMux.Lock()
	defer t.tsoMux.Unlock()
	for _, opt := range opts {
		if opt(t) {
			return
		}
	}
	if typeutil.SubTSOPhysicalByWallClock(next, t.tsoMux.physical) > 0 {
		t.tsoMux.physical = next
		t.tsoMux.logical = 0
		t.metrics.saveEvent.Inc()
	}
}
```

pkg/tso/tso.go
```diff
@@ -388,6 +413,9 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
+			if err := t.updateTimestamp(false); err != nil {
+				log.Warn("update timestamp failed", logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0), zap.Error(err))
+			}
 			time.Sleep(t.updatePhysicalInterval)
```
JmPotato: If `updateTimestamp` returns successfully, the sleep here seems unnecessary.

JmPotato (commented):
The rest LGTM.

pkg/tso/tso.go
```diff
@@ -353,8 +379,7 @@ func (t *timestampOracle) updateTimestamp() error {
 		t.metrics.updateSaveDuration.Observe(time.Since(start).Seconds())
 	}
 	// save into memory
-	t.setTSOPhysical(next, false)
-
+	t.setTSOPhysical(next, withInitialCheck(), withLogicalOverflowCheck())
```
JmPotato: The logic here seems a bit problematic. If overflow hasn’t happened yet, then wouldn’t that mean the TSO won’t advance?

pkg/tso/tso.go
bufferflies: Yes, the updateTimestamp will return a bool indicating whether the current TSO logical is overflowed. If not, it can acquire tso right right not

liyishuai:
I don't quite get the point of `allowSaveStorage`. What happens if it is always `true`?

bufferflies:
> I don't quite get the point of `allowSaveStorage`. What happens if it is always `true`?

It avoids all goroutines that can save ts in etcd. Only allocatorUpdater can save tso in etcd

liyishuai:
> > I don't quite get the point of `allowSaveStorage`. What happens if it is always `true`?
> 
> It avoids all goroutines that can save ts in etcd. Only allocatorUpdater can save tso in etcd

If the Go runtime does not wake up `allocatorUpdater`, does the allocation halt as well?
That doesn't seem to address the problem. Consider allowing all goroutines to submit transactions to advance the allocation?

bufferflies:
> > > I don't quite get the point of `allowSaveStorage`. What happens if it is always `true`?
> > 
> > 
> > It avoids all goroutines that can save ts in etcd. Only allocatorUpdater can save tso in etcd
> 
> If the Go runtime does not wake up `allocatorUpdater`, does the allocation halt as well? That doesn't seem to address the problem. Consider allowing all goroutines to submit transactions to advance the allocation?

In the current user case, allocatorUpdater doesn't wake up for more than the lease duration (5s). The root cause is that the physical doesn't advance.

pkg/tso/tso.go
```diff
@@ -334,28 +355,37 @@ func (t *timestampOracle) updateTimestamp() error {
...
-	t.setTSOPhysical(next, false)
-
-	return nil
+	if allowSaveStorage {
+		overflowed = t.setTSOPhysical(next, withInitialCheck())
+	} else {
+		overflowed = t.setTSOPhysical(next, withInitialCheck(), withLogicalOverflowCheck())
+	}
+	return overflowed, nil
 }
```
JmPotato: The logic here is a bit confusing; perhaps more comments need to be added or the code needs to be revised. If `allowSaveStorage` is true, it means this is the regular TSO advancement logic that runs every 50ms, so as long as the `withInitialCheck` check passes, it can update the TSO in memory. If `allowSaveStorage` is false, it means this is an advancement attempt from the TSO allocation path, and reaching this point indicates that the previous etcd persistence logic was not triggered—why do we still need to ensure passing `withLogicalOverflowCheck` at this time? The reason we introduced this proactive advancement logic in the TSO path is precisely to allow advancement attempts when an overflow occurs, isn’t it?

bufferflies: Currently, I will add this comment to reduce the confusion.

pkg/tso/tso.go
```diff
@@ -287,9 +319,9 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
 //
 // NOTICE: this function should be called after the TSO in memory has been initialized
 // and should not be called when the TSO in memory has been reset anymore.
-func (t *timestampOracle) updateTimestamp() error {
+func (t *timestampOracle) updateTimestamp(updatePurpose updatePurpose) (bool, error) {
```
JmPotato: Better add a comment to explain what the returned bool means.

pkg/tso/tso.go
```diff
@@ -287,9 +319,9 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
...
 	if !t.isInitialized() {
-		return errs.ErrUpdateTimestamp.FastGenByArgs("timestamp in memory has not been initialized")
+		return true, errs.ErrUpdateTimestamp.FastGenByArgs("timestamp in memory has not been initialized")
```
JmPotato: If the return value indicates whether it has overflowed, this should not be true.

pkg/tso/tso.go
```diff
@@ -334,28 +365,39 @@ func (t *timestampOracle) updateTimestamp() error {
...
 		t.metrics.skipSaveEvent.Inc()
-		return nil
+		return false, nil
 	}
 
 	// It is not safe to increase the physical time to `next`.
 	// The time window needs to be updated and saved to etcd.
 	if typeutil.SubRealTimeByWallClock(t.getLastSavedTime(), next) <= updateTimestampGuard {
+		// Only IntervalUpdate is allowed to save timestamp into etcd.
+		// it would be dangerous to save timestamp into etcd when handling overflowUpdate.
```
JmPotato: It would be best to elaborate a bit in the comments on what “dangerous” actually means.

pkg/tso/tso.go
```diff
@@ -334,28 +365,39 @@ func (t *timestampOracle) updateTimestamp() error {
...
+		}
 		save := next.Add(t.saveInterval)
 		start := time.Now()
 		if err := t.saveTimestamp(save); err != nil {
 			log.Warn("save timestamp failed",
 				logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
 				zap.Error(err))
 			t.metrics.errSaveUpdateTSEvent.Inc()
-			return err
+			return true, err
```
JmPotato: Ditto.

pkg/tso/tso.go
```diff
@@ -334,28 +365,39 @@ func (t *timestampOracle) updateTimestamp() error {
...
 		t.lastSavedTime.Store(save)
 		t.metrics.updateSaveDuration.Observe(time.Since(start).Seconds())
 	}
-	// save into memory
-	t.setTSOPhysical(next, false)
-
-	return nil
+	var overflowed bool
+	// If it's an IntervalUpdate, we don't need to check logical overflow, just update physical time directly.
+	// otherwise, we need to check logical overflow to avoid unnecessary physical time update.
```
JmPotato: As with the previous comment, please elaborate on why, in the latter case, we need to check for overflow to avoid extra updates.

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso.go
bufferflies: good catch ,done

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso.go
```diff
@@ -285,11 +319,15 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
...
 // 3. The physical time is always less than the saved timestamp.
 //
+// The first returns a boolean indicates whether the logical time is overflowed, and the second returns an error if any error occurs.
+// When met logical overflow, the caller should wait for the next physical tick to retry allocating TSO, which may cause more clock offset issues and etcd load,
+// so we want to avoid it as much as possible by updating physical time in advance.
+//
 // NOTICE: this function should be called after the TSO in memory has been initialized
 // and should not be called when the TSO in memory has been reset anymore.
-func (t *timestampOracle) updateTimestamp() error {
+func (t *timestampOracle) updateTimestamp(updatePurpose updatePurpose) (bool, error) {
```
JmPotato: The parameter has the same name as the type `updatePurpose`. This compiles in Go but is confusing and shadows the type within the function body. Better consider renaming the parameter to `purpose`.

pkg/tso/tso.go
```diff
@@ -90,18 +90,41 @@ func (t *timestampOracle) saveTimestamp(ts time.Time) error {
...
+	return func(t *timestampOracle) bool {
+		return t.tsoMux.physical.Equal(typeutil.ZeroTime)
+	}
+}
+
+func withLogicalOverflowCheck() setTSOOption {
+	return func(t *timestampOracle) bool {
+		return !overflowedLogical(t.tsoMux.logical)
+	}
+}
```
JmPotato: The name `withLogicalOverflowCheck` suggests "check if there's an overflow", but it actually means "skip if there is NOT an overflow". A reader needs to mentally invert the logic. Consider renaming to something like `skipIfNotInitialized` / `skipIfNotOverflowed`, or inverting the convention so true means "proceed" instead of "skip".

pkg/tso/tso.go
```diff
@@ -382,13 +432,18 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
...
-		if resp.GetLogical() >= maxLogical {
+		if overflowedLogical(resp.GetLogical()) {
 			log.Warn("logical part outside of max logical interval, please check ntp time, or adjust config item `tso-update-physical-interval`",
 				logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
-			time.Sleep(t.updatePhysicalInterval)
+			if overflowed, err := t.updateTimestamp(overflowUpdate); err != nil {
+				log.Info("update timestamp failed", logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0), zap.Error(err))
```
JmPotato: This log should be more like a warnning.

pkg/tso/tso_test.go
```diff
@@ -0,0 +1,183 @@
...
+					_, err := timestampOracle.getTS(runCtx, 1)
+					totalTso.Add(1)
+					if err != nil {
+						select {
+						case errCh <- err:
+							runCancel()
+						default:
+						}
+					}
+					if i == 0 {
```
JmPotato: We actually can use a struct like `map[int64]struct{}` to count all physical timestamps generated by all goroutines to check if it doesn't exceed the expected number of timestamp advances.

JmPotato (commented):
😂 I’m “sorry” for my strict review—but it’s necessary. After all, this is extremely critical code logic like TSO, and it’s essential to keep the code clear in every sense.

pkg/tso/tso.go
bufferflies: yes, good suggestion

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso.go
bufferflies: done

pkg/tso/tso_test.go
bufferflies: You're right, I have collected all goroutines when the logical response time is equal to 1, which indicates the physical has been allocated.

pkg/tso/tso.go
```diff
@@ -90,18 +90,43 @@ func (t *timestampOracle) saveTimestamp(ts time.Time) error {
 	return t.storage.SaveTimestamp(ctx, t.keyspaceGroupID, ts, t.member.GetLeadership())
 }
 
-func (t *timestampOracle) setTSOPhysical(next time.Time, force bool) {
+// setTSOOption defines the option type for setTSOPhysical function.
+// if it returns true, it will skip update physical time.
+type setTSOOption func(t *timestampOracle) bool
```
okJiang: `setTSOOption` sounds like a 'configuration option', but this type is actually a function that determines 'whether to skip the update', more like a guard/predicate.
```suggestion
type setTSOPhysicalGuard func(t *timestampOracle) bool
```

okJiang: Or treat Initialized and Overflowed as a prerequisite that must be met before setting TSO.


```suggestion
type setTSOPrecond func(t *timestampOracle) bool
```

```go
    for _, precond := range preconds {
		if !precond(t) {
			return overflowedLogical(t.tsoMux.logical)
		}
	}
```

bufferflies: Yes, I have renamed this option to condition, and renamed mustInitialized and mustOverflowed to make the code more concise.

okJiang (approved):

pkg/tso/tso.go
```diff
@@ -334,28 +375,41 @@ func (t *timestampOracle) updateTimestamp() error {
...
+	var overflowed bool
+	// If it's an IntervalUpdate, we don't need to check logical overflow, just update physical time directly.
+	// Otherwise, the caller met logical overflow, so it will allocate physical time to alloc more timestamp in concurrent.
+	// So we need to check logical overflow before updating physical time to avoid allocating too much physical time due to logical overflow.
+	if purpose == intervalUpdate {
+		overflowed = t.setTSOPhysical(next, mustInitialized())
+	} else {
+		overflowed = t.setTSOPhysical(next, mustInitialized(), mustOverflowed())
+	}
+	return overflowed, nil
```
JmPotato: Can be simplified:

```suggestion
	// If it's an IntervalUpdate, we don't need to check logical overflow, just update physical time directly.
	// Otherwise, the caller met logical overflow, so it will allocate physical time to alloc more timestamp in concurrent.
	// So we need to check logical overflow before updating physical time to avoid allocating too much physical time due to logical overflow.
	conditions := []setTSOCondition{mustInitialized()}
	if purpose != intervalUpdate {
		conditions = append(conditions, mustOverflowed())
	}
	return t.setTSOPhysical(next, conditions...), nil
```

pkg/tso/tso.go
```diff
@@ -382,13 +436,18 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
...
-		if resp.GetLogical() >= maxLogical {
+		if overflowedLogical(resp.GetLogical()) {
 			log.Warn("logical part outside of max logical interval, please check ntp time, or adjust config item `tso-update-physical-interval`",
 				logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
-			time.Sleep(t.updatePhysicalInterval)
+			if overflowed, err := t.updateTimestamp(overflowUpdate); err != nil {
+				log.Warn("update timestamp failed", logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0), zap.Error(err))
```
JmPotato: Add the `overflowed` field as well?

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

pkg/tso/tso.go
```diff
@@ -382,13 +434,21 @@ func (t *timestampOracle) getTS(ctx context.Context, count uint32) (pdpb.Timesta
...
 		}
-		if resp.GetLogical() >= maxLogical {
+		if overflowedLogical(resp.GetLogical()) {
 			log.Warn("logical part outside of max logical interval, please check ntp time, or adjust config item `tso-update-physical-interval`",
 				logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
 				zap.Reflect("response", resp),
 				zap.Int("retry-count", i), errs.ZapError(errs.ErrLogicOverflow))
 			t.metrics.logicalOverflowEvent.Inc()
-			time.Sleep(t.updatePhysicalInterval)
+			if overflowed, err := t.updateTimestamp(overflowUpdate); err != nil {
```
rleungx, line 443: In high-QPS overflow cases, every concurrent request may enter this branch and call updateTimestamp/log warnings. Could we coalesce this path, for example with a singleflight or atomic flag, so only one goroutine advances the physical time while others retry or wait?

Issue #10140, opened 2026-01-08
Improve TSO code style


liyishuai:
## Enhancement Task
https://github.com/tikv/pd/blob/master/pkg/tso/tso.go has become a bowl of spaghetti. I'll attempt to standardize the code for readability.

PR #10142, opened 2026-01-09
tso: early return -> switch case

liyishuai:

Issue Number: Ref #10140 

Will change more after #10137 gets merged.

1. Replace early returns with switch cases
2. Unify timestamp comparisons
3. Remove redundant timestamp comparisons

```commit-message
```

### Check List

Tests

- Not required
### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization and timestamp validation to prevent incorrect skips, ensure monotonic timestamp updates during sync/reset, reduce erroneous writes, and cut noisy logging.

* **Refactor**
  * Consolidated internal gating and timeout handling for clearer control flow; no public APIs or external behaviors were changed.

### DIFF
--- a/pkg/tso/tso.go
+++ b/pkg/tso/tso.go
@@ -242,17 +242,19 @@ func (t *timestampOracle) isInitializedLocked() bool {
 func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUpperBoundCheck bool) error {
 	t.tsoMux.Lock()
 	defer t.tsoMux.Unlock()
-	var (
-		nextPhysical, nextLogical = tsoutil.ParseTS(tso)
-		logicalDifference         = int64(nextLogical) - t.tsoMux.logical
-		physicalDifference        = typeutil.SubTSOPhysicalByWallClock(nextPhysical, t.tsoMux.physical)
-	)
 	switch {
 	case !t.member.IsServing():
 		t.metrics.errLeaseResetTSEvent.Inc()
 		return errs.ErrResetUserTimestamp.FastGenByArgs(errs.NotLeaderErr)
 	case !t.isInitializedLocked():
 		return errs.ErrResetUserTimestamp.FastGenByArgs("timestamp in memory has not been initialized")
+	}
+	var (
+		nextPhysical, nextLogical = tsoutil.ParseTS(tso)
+		logicalDifference         = int64(nextLogical) - t.tsoMux.logical
+		physicalDifference        = typeutil.SubTSOPhysicalByWallClock(nextPhysical, t.tsoMux.physical)
+	)
+	switch {
 	case physicalDifference < 0:
 		t.metrics.errResetSmallPhysicalTSEvent.Inc()
 		switch {


liyishuai:
Interesting: Golang seems to favor early returns (while I do the opposite). Reverting.

pkg/tso/tso.go
```diff
@@ -94,9 +94,12 @@ func (t *timestampOracle) setTSOPhysical(next time.Time, force bool) {
 	t.tsoMux.Lock()
 	defer t.tsoMux.Unlock()
 	// Do not update the zero physical time if the `force` flag is false.
-	if t.tsoMux.physical.Equal(typeutil.ZeroTime) && !force {
+	if !t.isInitializedNoLock() && !force {
```
rleungx: Usually we use xxxLocked
```suggestion
	if !t.isInitializedLocked() && !force {
```

liyishuai: Renamed.

pkg/tso/tso.go
```diff
@@ -250,6 +265,9 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
 		}
 		return errs.ErrResetUserTimestamp.FastGenByArgs("the specified counter is smaller than now")
 	}
+	// else:
```
okJiang: How about using switch instead of if-else

liyishuai: Good idea. Updated correspondingly.

pkg/tso/tso.go
```diff
@@ -78,10 +78,12 @@ type timestampOracle struct {
 
 func (t *timestampOracle) getStorageTimeout() time.Duration {
 	timeout := t.saveInterval
-	if timeout < config.DefaultTSOSaveInterval {
+	switch {
+	case timeout >= config.DefaultTSOSaveInterval:
+		return timeout - time.Second
+	default:
 		return config.DefaultTSOSaveInterval - time.Second
```
JmPotato, line 89: Maybe add a comment explaining that this check ensures the storage timeout is at least `config.DefaultTSOSaveInterval - time.Second`.

liyishuai: Thanks! Done.

pkg/tso/tso.go
```diff
@@ -221,37 +242,34 @@ func (t *timestampOracle) isInitialized() bool {
...
+	switch {
+	case !t.member.IsServing():
+		t.metrics.errLeaseResetTSEvent.Inc()
+		return errs.ErrResetUserTimestamp.FastGenByArgs(errs.NotLeaderErr)
+	case !t.isInitializedLocked():
 		return errs.ErrResetUserTimestamp.FastGenByArgs("timestamp in memory has not been initialized")
-	}
-	// do not update if next physical time is less/before than prev
-	if physicalDifference < 0 {
+	case physicalDifference < 0:
```
okJiang, line 258: physicalDifference and `t.xxxxx()` should not mix in one switch.

liyishuai: Makes sense. Working on it.

okJiang:
> Interesting: Golang seems to favor early returns (while I do the opposite). Reverting.

@liyishuai  Recommend you to read this book https://100go.co/

> Unnecessary nested code (#2)

image

pkg/tso/tso.go
```diff
@@ -76,12 +76,18 @@ type timestampOracle struct {
...
 }
 
+// getStorageTimeout returns the timeout for storage operations.
+// The returned timeout ensures that the storage timeout is at least
+// `config.DefaultTSOSaveInterval - time.Second`, which gives enough buffer
+// for storage operations to complete before hitting the save deadline.
 func (t *timestampOracle) getStorageTimeout() time.Duration {
 	timeout := t.saveInterval
-	if timeout < config.DefaultTSOSaveInterval {
+	switch {
```
okJiang, line 85: Normally, it is only recommended to use switch in three situations or more

liyishuai: I see the difference: Golang is an imperative language, while my mind fell in the declarative world.
In Go, `return` means "no more computations from here", so early `return` simply means truncating the program.
In declarative languages, `if` means "the rest of this program is diverging", so every `if` must have its corresponding `else`.

I'll adjust the programming style accordingly. Expect another large edit.

PR #10150, opened 2026-01-14
config: support to update controller config by pd-ctl

### LINKED ISSUE #10148: pd-ctl: support update resource control configuration
## Enhancement Task
Currently, we can only call the HTTP request to change the resource config like this:

```

curl -X POST \
  http://<PD_ADDR>/resource-manager/api/v1/config/controller  \
  -H "Content-Type: application/json" \
  -d '{
    "ltb-max-wait-duration": "30m"
  }'
```

But this config is under the server config, should be changed by 

```

curl -X POST \
  http://<PD_ADDR>/pd/api/v1/config/ \
  -H "Content-Type: application/json" \
  -d '{
    "controller.ltb-max-wait-duration": "30m"
  }'
```

bufferflies:

Issue Number: Close #10148

```commit-message
support the config command `config set ltb-max-wait-duration 30m`
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Dynamic updates to controller configuration (cost params, max-wait-duration, trace logging) are supported and propagated; attempts to modify disallowed nested controller child items are rejected.
  * Updates validate leadership availability and surface errors when updates cannot be applied.

* **Tests**
  * Expanded end-to-end tests: CLI config flows, show validations, forbidden-update assertions, and cross-system CRUD checks for resource groups to ensure PD↔RM consistency.

### DIFF
--- a/tests/integrations/mcs/resourcemanager/redirector_test.go
+++ b/tests/integrations/mcs/resourcemanager/redirector_test.go
@@ -130,15 +130,15 @@ func (suite *resourceManagerRedirectorTestSuite) TestGRPCRedirectsResourceGroupR
 
 	pdClient := rmpb.NewResourceManagerClient(pdConn)
 	rmClient := rmpb.NewResourceManagerClient(rmConn)
-	req := &rmpb.GetResourceGroupRequest{
+	getReq := &rmpb.GetResourceGroupRequest{
 		ResourceGroupName: groupName,
 		KeyspaceId: &rmpb.KeyspaceIDValue{
 			Value: suite.keyspaceID,
 		},
 	}
-	pdResp, err := pdClient.GetResourceGroup(ctx, req)
+	pdResp, err := pdClient.GetResourceGroup(ctx, getReq)
 	re.NoError(err)
-	rmResp, err := rmClient.GetResourceGroup(ctx, req)
+	rmResp, err := rmClient.GetResourceGroup(ctx, getReq)
 	re.NoError(err)
 	re.Nil(pdResp.GetError())
 	re.Nil(rmResp.GetError())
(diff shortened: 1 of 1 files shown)


tools/pd-ctl/tests/resourcemanager/resource_manager_command_test.go
```diff
@@ -98,3 +98,48 @@ func (s *testResourceManagerSuite) TestConfigController() {
...
+		re.NoError(err)
+
+		actualCfg := server.ControllerConfig{}
+		err = json.Unmarshal(output, &actualCfg)
+		re.NoError(err, string(output))
+		re.Equal(expectCfg.Controller, actualCfg)
+	}
+
+	// Check default config
+	checkShow()
```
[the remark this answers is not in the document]
bufferflies: removed line 119

server/api/config.go
```diff
@@ -669,3 +675,48 @@ func (h *confHandler) getSchedulingServerConfig() (*config.Config, error) {
...
+		return errors.Errorf("can't update controller child item %s", k)
+	}
+	addrs := h.svr.GetMember().GetServingUrls()
+	if len(addrs) == 0 {
+		return errs.ErrLeaderNil.FastGenByArgs()
+	}
+	addr := addrs[0]
+	url := fmt.Sprintf("%s%s", addr, controllerConfigURL)
+
+	// 构造请求体
```
lhy1024: chinese comments?

server/api/config.go
```diff
@@ -669,3 +675,48 @@ func (h *confHandler) getSchedulingServerConfig() (*config.Config, error) {
...
+	// Don't allow updating child items, only allow updating top-level items.
+	if k != key {
+		return errors.Errorf("can't update controller child item %s", k)
+	}
+	addrs := h.svr.GetMember().GetServingUrls()
+	if len(addrs) == 0 {
+		return errs.ErrLeaderNil.FastGenByArgs()
+	}
+	addr := addrs[0]
+	url := fmt.Sprintf("%s%s", addr, controllerConfigURL)
```
lhy1024: If resource manager run as micro service, can it work?

bufferflies: yes, this request will be resent to the PD leader and redirected to the resource manager such like scheduling

server/api/config.go
```diff
@@ -669,3 +675,47 @@ func (h *confHandler) getSchedulingServerConfig() (*config.Config, error) {
...
 }
+
+func (h *confHandler) updateControllerConfig(key string, value any) error {
+	k := reflectutil.FindJSONFullTagByChildTag(reflect.TypeOf(rm.ControllerConfig{}), key)
+	if k == "" {
+		return errors.Errorf("controller config item %s not found", key)
+	}
+	// Don't allow updating child items, only allow updating top-level items.
+	if k != key {
+		return errors.Errorf("can't update controller child item %s", k)
```
AndreMouche: The error message uses `k` which might be the full path (e.g., "request-unit.write-base-cost"), but the user provided `key` (e.g., "write-base-cost"). This could be confusing. Consider using `key` in the error message instead:

```suggestion
		return errors.Errorf("can't update controller child item %s, only top-level items are allowed", key)
```

server/api/config.go
```diff
@@ -44,7 +45,10 @@ import (
 
 // This line is to ensure the package `sc` could always be imported so that
 // the swagger could generate the right definitions for the config structs.
-var _ *sc.ScheduleConfig = nil
+var (
+	_                   *sc.ScheduleConfig = nil
+	controllerConfigURL                    = "/resource-manager/api/v1/config/controller"
```
AndreMouche: how about
```suggestion
	resourceManagerControllerConfigURL                    = "/resource-manager/api/v1/config/controller"
```

server/api/config.go
```diff
@@ -669,3 +675,47 @@ func (h *confHandler) getSchedulingServerConfig() (*config.Config, error) {
...
+
+	body := map[string]any{
+		key: value,
+	}
+	data, err := json.Marshal(body)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, url, strings.NewReader(string(data)))
```
AndreMouche: Missing timeout for HTTP request, should we considered it? or Have we already considered it?

bufferflies: the h.svr.GetHTTPClient() will do it

server/api/config.go
bufferflies: done

server/api/config.go
bufferflies: done

server/api/config.go
bufferflies: removed

tools/pd-ctl/tests/resourcemanager/resource_manager_command_test.go
```diff
@@ -98,3 +98,45 @@ func (s *testResourceManagerSuite) TestConfigController() {
 	expectCfg.Controller.RequestUnit.WriteBaseCost = 2
 	checkShow()
 }
+
+func (s *testResourceManagerSuite) TestUpdateConfig() {
+	re := s.Require()
+	expectCfg := server.Config{}
```
AndreMouche: I suggest to Load the current configuration from the PD instead of initializing with default values:

The `expectCfg` is initialized with default values here, which can cause test failures due to state pollution from previous tests. 
If `TestConfigController` runs before `TestUpdateConfig`, it modifies the Resource Manager configuration (e.g., sets `WriteBaseCost` to 2). When `TestUpdateConfig` starts, it initializes `expectCfg` with default values (e.g., `WriteBaseCost = 1`), but the actual configuration in the Resource Manager still has the modified value (e.g., `WriteBaseCost = 2`). This causes the assertion at line 115 to fail.
The rootcause is `SetupSuite()` is only called once before all tests run, so the test cluster state persists across different test methods. Each test method shares the same cluster instance, leading to state pollution between tests.

bufferflies: yes, I have merge this two test into one avoid flaky test

Issue #10164, opened 2026-01-19
Fix flaky tests in Eventually assertions to prepare for testify v1.11.0 upgrade


JmPotato:
### Background

Following the introduction of https://github.com/stretchr/testify/pull/1427, we have observed a significant increase in the failure rate of flaky tests involving `Eventually` assertions within the project.

### Problem

The changes in testify's `Eventually` implementation have exposed underlying instabilities in our current test suite. To maintain CI stability and ensure a smooth transition to the upcoming testify v1.11.0, these flaky tests must be addressed.

### Goals

- Identify and fix all unstable tests triggered by the new `Eventually` behavior.
- Strengthen CI quality by reducing non-deterministic failures.
- Unblock the upgrade to testify v1.11.0 (and subsequent versions).

PR #10166, opened 2026-01-19
test(api/scheduler): improve pause/resume timing checks in scheduler tests

JmPotato:

Issue Number: ref #10164.

```commit-message
Refactors scheduler API tests to replace fixed sleep-based assertions with
eventual consistency checks using `testutil.Eventually`. This ensures the tests
reliably wait for all schedulers to reach the paused or resumed state within a
bounded time window rather than relying on static delays.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved scheduler verification mechanism in test suite with event-driven polling and eventual consistency checks instead of static delays.

### DIFF
--- a/tests/server/api/scheduler_test.go
+++ b/tests/server/api/scheduler_test.go
@@ -548,29 +548,40 @@ func (suite *scheduleTestSuite) checkAPI(cluster *tests.TestCluster) {
 	re.NoError(err)
 	err = testutil.CheckPostJSON(tests.TestDialClient, urlPrefix+"/all", pauseArgs, testutil.StatusOK(re))
 	re.NoError(err)
-
-	for _, testCase := range testCases {
-		createdName := testCase.createdName
-		if createdName == "" {
-			createdName = testCase.name
+	// Check if all schedulers are paused.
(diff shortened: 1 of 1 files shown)


tests/server/api/scheduler_test.go
```diff
@@ -548,49 +548,55 @@ func (suite *scheduleTestSuite) checkAPI(cluster *tests.TestCluster) {
...
+			}
+			if isSchedulerPaused(re, urlPrefix, createdName) {
+				return false
+			}
 		}
-		isPaused := isSchedulerPaused(re, urlPrefix, createdName)
-		re.False(isPaused)
+		return true
 	}
+	testutil.Eventually(re, checkAllSchedulersResumed) // Must be resumed eventually.
```
lhy1024, line 584: at least after 1 second?

JmPotato: It seems that we currently can’t implement an assertion like “wait at least 1s before the condition is met,” so here we’ll just use a simple eventually assertion. Do you have any ideas?

Issue #10149, opened 2026-01-14
checker, affinity: IsRegionReplicated does not enforce highest isolation level


lhy1024:
## Enhancement Task

Problem: 

With placement rules enabled, filter.IsRegionReplicated relies on RuleFit.IsSatisfied(), which only checks replica count/role/orphan peers and does not verify the highest isolation level (the deepest location-labels layer). As a result, a region can be considered replicated even when it does not reach the highest isolation level. This affects all callers of IsRegionReplicated, not only affinity-related logic.

Expected: 

When location-labels (or IsolationLevel) is configured, a “strict replicated” check should require the region to reach the highest isolation level before returning true, or provide an explicit strict variant for schedulers/checkers to opt into.

Repro: 

TestIsRegionReplicatedIgnoresHighestIsolationLevel sets labels=[zone,host] with two peers sharing the same host. GetRegionLabelIsolation is not host, yet IsRegionReplicated returns true.
```

func TestIsRegionReplicatedIgnoresHighestIsolationLevel(t *testing.T) {
	re := require.New(t)
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	peers := func(ids ...uint64) []*metapb.Peer {
		var peers []*metapb.Peer
		for _, id := range ids {
			peers = append(peers, &metapb.Peer{Id: id, StoreId: id})
		}
		return peers
	}

	opt := mockconfig.NewTestOptions()
	tc := mockcluster.NewCluster(ctx, opt)
	labels := []string{"zone", "host"}
	tc.SetLocationLabels(labels)
	tc.SetMaxReplicas(3)
	tc.SetEnablePlacementRules(true)

	tc.AddLabelsStore(1, 0, map[string]string{"zone": "z1", "host": "h1"})
	tc.AddLabelsStore(2, 0, map[string]string{"zone": "z1", "host": "h1"})
	tc.AddLabelsStore(3, 0, map[string]string{"zone": "z2", "host": "h2"})

	region := core.NewRegionInfo(&metapb.Region{Peers: peers(1, 2, 3)}, peers(1)[0])
	isolation := statistics.GetRegionLabelIsolation(tc.GetRegionStores(region), labels)
	re.NotEqual("host", isolation)
	re.True(IsRegionReplicated(tc, region))
}
```

<img width="1991" height="699" alt="Image" src="https://github.com/user-attachments/assets/aaa0ec8b-bbf5-4d64-b0cf-52f060e45670" />

[...]

PR #10192, opened 2026-01-28
affinity: use isRegionPlacementRuleSatisfiedWithBestLocation instead of IsRegionReplicated

HunDunDM:

Issue Number: Ref #10149

```commit-message
check location labels in AffinityChecker
```

### Check List

Tests

- Unit test

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Affinity checks now evaluate placement-rule satisfaction (including a cached and cacheless region fit path) and region label isolation reports a numeric level for level-based checks.

* **Bug Fixes**
  * Better detection and handling of superior replica placements, stricter enforcement of isolation constraints, preservation of learners, and early exit when placement rules are disabled.

* **Metrics**
  * Added a counter for placement-rules-disabled occurrences.

* **Tests**
  * Expanded and updated tests for placement-rule, isolation-level, normalization, and placement interactions.

### DIFF
--- a/pkg/schedule/checker/affinity_checker.go
+++ b/pkg/schedule/checker/affinity_checker.go
@@ -34,6 +34,7 @@ import (
 	"github.com/tikv/pd/pkg/schedule/operator"
 	"github.com/tikv/pd/pkg/schedule/placement"
 	"github.com/tikv/pd/pkg/schedule/types"
+	"github.com/tikv/pd/pkg/statistics"
 	"github.com/tikv/pd/pkg/utils/logutil"
 )
 
@@ -527,6 +528,10 @@ func (c *AffinityChecker) isRegionPlacementRuleSatisfiedWithBestLocation(region 
 		// filterByTempState being true means a better placement exists but is temporarily unschedulable.
 		// This is also considered not satisfied.
 		if newStoreID != 0 || filterByTempState {
+			return false
+		}
+		// If the isolation level does not meet the requirement, it is also considered not to be at the best location.
+		if !statistics.IsRegionLabelIsolationSatisfied(rf.Stores, rf.Rule.LocationLabels, rf.Rule.IsolationLevel) {
 			return false
 		}
 	}
--- a/pkg/schedule/checker/affinity_checker.go
+++ b/pkg/schedule/checker/affinity_checker.go
@@ -44,6 +44,7 @@ const recentMergeTTL = time.Minute
 type AffinityChecker struct {
 	PauseController
 	cluster          sche.CheckerCluster
+	ruleManager      *placement.RuleManager
 	affinityManager  *affinity.Manager
 	conf             config.CheckerConfigProvider
 	recentMergeCache *cache.TTLUint64
@@ -54,6 +55,7 @@ type AffinityChecker struct {
 func NewAffinityChecker(ctx context.Context, cluster sche.CheckerCluster, conf config.CheckerConfigProvider) *AffinityChecker {
 	return &AffinityChecker{
 		cluster:          cluster,
+		ruleManager:      cluster.GetRuleManager(),
 		affinityManager:  cluster.GetAffinityManager(),
 		conf:             conf,
 		recentMergeCache: cache.NewIDTTL(ctx, gcInterval, recentMergeTTL),
@@ -93,7 +95,7 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		affinityCheckerUnhealthyRegionCounter.Inc()
 		return nil
 	}
-	if !filter.IsRegionReplicated(c.cluster, region) {
+	if !c.isRegionPlacementRuleSatisfiedWithBestLocation(region, true /* isRealRegion */) {
 		affinityCheckerAbnormalReplicaCounter.Inc()
 		return nil
 	}
@@ -115,7 +117,7 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		// so expire the group first, then provide the available Region information and fetch the Group state again.
 		if !isAffinity {
 			targetRegion := cloneRegionWithReplacePeerStores(region, group.LeaderStoreID, group.VoterStoreIDs...)
-			if targetRegion == nil || !filter.IsRegionReplicated(c.cluster, targetRegion) {
+			if targetRegion == nil || !c.isRegionPlacementRuleSatisfiedWithBestLocation(targetRegion, false /* isRealRegion */) {
 				c.affinityManager.ExpireAffinityGroup(group.ID)
 				needRefetch = true
 			}
@@ -381,7 +383,7 @@ func (c *AffinityChecker) checkAffinityMergeTarget(region, adjacent *core.Region
 		return false
 	}
 
-	if !filter.IsRegionReplicated(c.cluster, adjacent) {
+	if !c.isRegionPlacementRuleSatisfiedWithBestLocation(adjacent, true /* isRealRegion */) {
 		affinityMergeCheckerAdjAbnormalReplicaCounter.Inc()
 		return false
 	}
@@ -490,4 +492,56 @@ func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
 	}
 	c.recentMergeCache.PutWithTTL(op.RegionID(), nil, recentMergeTTL)
 	c.recentMergeCache.PutWithTTL(relatedID, nil, recentMergeTTL)
+}
+
+func (c *AffinityChecker) isRegionPlacementRuleSatisfiedWithBestLocation(region *core.RegionInfo, isRealRegion bool) bool {
+	// Return false if Placement Rules are not enabled.
+	if c.cluster.GetSharedConfig().IsPlacementRulesEnabled() {
+		return false
+	}
+
+	// Get the RegionFit for the given Region. If the Region is not a real existing Region but a virtual target state,
+	// use FitRegionWithoutCache to bypass the cache.
+	var fit *placement.RegionFit
+	if isRealRegion {
+		fit = c.ruleManager.FitRegion(c.cluster, region)
+	} else {
+		fit = c.ruleManager.FitRegionWithoutCache(c.cluster, region)
+	}
+
+	// Check region is satisfied
+	if !fit.IsSatisfied() {
+		return false
+	}
+
+	// Check whether all peers covered by the rules are at the best isolation level.
+	// This logic is based on `RuleChecker.fixBetterLocation`.
+	for _, rf := range fit.RuleFits {
+		if len(rf.Rule.LocationLabels) == 0 {
+			continue
+		}
+		isWitness := rf.Rule.IsWitness && isWitnessEnabled(c.cluster)
+		// If the peer to be moved is a witness, since no snapshot is needed, we also reuse the fast failover logic.
+		strategy := c.strategy(region, rf.Rule, isWitness)
+		_, newStoreID, filterByTempState := strategy.getBetterLocation(c.cluster, region, fit, rf)
+		// filterByTempState being true means a better placement exists but is temporarily unschedulable.
+		// This is also considered not satisfied.
+		if newStoreID != 0 || filterByTempState {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (c *AffinityChecker) strategy(region *core.RegionInfo, rule *placement.Rule, fastFailover bool) *ReplicaStrategy {
+	return &ReplicaStrategy{
+		checkerName:    c.Name(),
+		cluster:        c.cluster,
+		isolationLevel: rule.IsolationLevel,
+		locationLabels: rule.LocationLabels,
+		region:         region,
+		extraFilters:   []filter.Filter{filter.NewLabelConstraintFilter(c.Name(), rule.LabelConstraints)},
+		fastFailover:   fastFailover,
+	}
 }
--- a/pkg/schedule/checker/affinity_checker.go
+++ b/pkg/schedule/checker/affinity_checker.go
@@ -100,7 +100,7 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		affinityCheckerUnhealthyRegionCounter.Inc()
 		return nil
 	}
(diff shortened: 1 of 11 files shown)


pkg/schedule/checker/affinity_checker.go
```diff
@@ -487,3 +493,50 @@ func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
...
+
+	// Check whether all peers covered by the rules are at the best isolation level.
+	// This logic is based on `RuleChecker.fixBetterLocation`.
+	for _, rf := range fit.RuleFits {
+		if len(rf.Rule.LocationLabels) == 0 {
+			continue
+		}
+		isWitness := rf.Rule.IsWitness && isWitnessEnabled(c.cluster)
+		// If the peer to be moved is a witness, since no snapshot is needed, we also reuse the fast failover logic.
+		strategy := c.strategy(region, rf.Rule, isWitness)
```
lhy1024, line 538: maybe we need not to consider witness

lhy1024 (commented):
I think the way is OK, but we need some tests for it.

pkg/schedule/checker/affinity_checker.go
HunDunDM: This is for compatibility reasons, as the original code was retained.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -487,3 +495,54 @@ func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
 	c.recentMergeCache.PutWithTTL(op.RegionID(), nil, recentMergeTTL)
 	c.recentMergeCache.PutWithTTL(relatedID, nil, recentMergeTTL)
 }
+
+func (c *AffinityChecker) isRegionPlacementRuleSatisfiedWithBestLocation(region *core.RegionInfo, isRealRegion bool) bool {
```
lhy1024: If the intended behavior here is that a region may still pass the affinity check even when it has not reached the deepest `location-labels` isolation level, as long as there is no strictly better store to move to in the current topology, then the logic itself seems reasonable to me. The part that feels unclear is that the code and PR wording can be read as if this were a general fix for `IsRegionReplicated` / `#10149`.

I suggest adding a more explicit function comment to clarify that this helper is an affinity-specific scheduling gate, not a general “strict replicated” check.

It would also help to state explicitly in the PR description that the change is intentionally scoped to affinity only.

lhy1024 (commented):
the rest LGTM

pkg/schedule/checker/affinity_checker.go
lhy1024: If this affinity-specific semantic split is intentional, please also add one test that explicitly locks it in, e.g. a case where filter.IsRegionReplicated(tc, region) == true but isRegionPlacementRuleSatisfiedWithBestLocation(region, true) == false, and document in the test that this difference is expected by design.

lhy1024:
```go
func TestAffinityStrictCheckDivergesFromIsRegionReplicated(t *testing.T) {
	re := require.New(t)
	defer cancel()

	opt := mockconfig.NewTestOptions()
	opt.SetLocationLabels([]string{"zone", "host"})
	tc := mockcluster.NewCluster(ctx, opt)
	tc.SetEnablePlacementRules(true)
	tc.SetMaxReplicas(3)

	tc.AddLabelsStore(1, 0, map[string]string{"zone": "z1", "host": "h1"})
	tc.AddLabelsStore(2, 0, map[string]string{"zone": "z1", "host": "h1"})
	tc.AddLabelsStore(3, 0, map[string]string{"zone": "z2", "host": "h2"})
	tc.AddLabelsStore(4, 0, map[string]string{"zone": "z3", "host": "h3"})

	checker := newTestAffinityChecker(ctx, tc, opt)

	tc.AddLeaderRegion(1, 1, 2, 3)
	region := tc.GetRegion(1)

	re.True(filter.IsRegionReplicated(tc, region))
	re.False(checker.isRegionPlacementRuleSatisfiedWithBestLocation(region, true))
}
```

lhy1024 (approved):

lhy1024:
@coderabbitai full review

pkg/schedule/placement/fit.go
```diff
@@ -138,8 +138,8 @@ type RuleFit struct {
 	// isolated. A larger value is better.
 	IsolationScore float64 `json:"isolation-score"`
 	WitnessScore   int     `json:"witness-score"`
-	// stores is the stores that the peers are placed in.
-	stores []*core.StoreInfo
+	// Stores is the stores that the peers are placed in.
+	Stores []*core.StoreInfo
 }
```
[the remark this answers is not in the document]
HunDunDM: `json:"-"` has been added.

HunDunDM:
@coderabbitai full review

pkg/schedule/checker/affinity_checker.go
```diff
@@ -487,3 +495,60 @@ func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
...
 }
+
+// isRegionPlacementRuleSatisfiedWithBestLocation is an affinity-specific
+// scheduling gate. Besides requiring the region to satisfy placement rules, it
+// also requires that affinity scheduling cannot find a strictly better location
+// under the current topology and that the matched rule's isolation level is
+// satisfied. This is intentionally stricter than filter.IsRegionReplicated and
+// should not be used as a general replicated-state check.
+func (c *AffinityChecker) isRegionPlacementRuleSatisfiedWithBestLocation(region *core.RegionInfo, isRealRegion bool) bool {
+	// Get the RegionFit for the given Region. If the Region is not a real existing Region but a virtual target state,
```
liyishuai: Consider renaming to `isExistingRegion`.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -487,3 +495,60 @@ func (c *AffinityChecker) RecordOpSuccess(op *operator.Operator) {
...
+		}
+		isWitness := rf.Rule.IsWitness && isWitnessEnabled(c.cluster)
+		// If the peer to be moved is a witness, since no snapshot is needed, we also reuse the fast failover logic.
+		strategy := c.strategy(region, rf.Rule, isWitness)
+		_, newStoreID, filterByTempState := strategy.getBetterLocation(c.cluster, region, fit, rf)
+		// filterByTempState being true means a better placement exists but is temporarily unschedulable.
+		// This is also considered not satisfied.
+		if newStoreID != 0 || filterByTempState {
+			return false
+		}
```
liyishuai, line 544: Is this an intentional bias? What is the drawback of dropping the criteria `filterByTempState`?

pkg/schedule/checker/replica_strategy.go
```diff
@@ -155,3 +157,48 @@ func (s *ReplicaStrategy) SelectStoreToRemove(coLocationStores []*core.StoreInfo
 	}
 	return source.GetID()
 }
+
+func (s *ReplicaStrategy) getBetterLocation(cluster sche.SharedCluster, region *core.RegionInfo, fit *placement.RegionFit, rf *placement.RuleFit) (oldStoreID, newStoreID uint64, filterByTempState bool) {
+	ruleStores := getRuleFitStores(cluster, rf)
+	oldStoreID = s.SelectStoreToRemove(ruleStores)
```
rleungx: This can miss a still-better placement when the source store is only temporarily unschedulable. `SelectStoreToRemove` uses the strict source filter, so if the peer that should be moved is currently busy, remove-limited, or snapshot-throttled, it returns `0` and `getBetterLocation` reports `(0, 0, false)`. The affinity gate then treats the region as already at the best location, even though a better target may exist. We should propagate a "filtered by temporary source state" signal here, similar to the target-store path, and let `isRegionPlacementRuleSatisfiedWithBestLocation` return false in that case.

pkg/schedule/checker/affinity_checker.go
HunDunDM: Yes, this is intentional. `filterByTempState` means a better placement exists under the current topology, but it is temporarily unschedulable because the selected source/target store is blocked by transient state such as busy, store limit, or snapshot throttling.

If we drop this criterion, the affinity checker may treat the region as already being at the best location and use that suboptimal state for affinity scheduling or group refresh. That can make the group converge on a placement that still needs to be improved once the temporary state clears. So the gate is deliberately conservative: a region is considered satisfied only when there is no better location, including no better location hidden behind temporary scheduling filters.

pkg/schedule/checker/affinity_checker.go
HunDunDM: Good point. I’ll rename it.

pkg/schedule/checker/replica_strategy.go
HunDunDM: Addressed. `SelectStoreToRemove` now uses a two-stage source selection: it first finds the best source with temporary states allowed, then applies the strict source filter. If that source is blocked only by temporary source state, `getBetterLocation` propagates `filterByTempState`, so the affinity gate returns false. I also added a regression test for the busy-source case.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -83,6 +86,10 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		affinityCheckerPausedCounter.Inc()
 		return nil
 	}
+	if !c.cluster.GetSharedConfig().IsPlacementRulesEnabled() {
+		affinityCheckerPlacementRulesDisabledCounter.Inc()
+		return nil
+	}
```
lhy1024, line 92: Question: is it intentional to skip the affinity checker entirely when placement rules are disabled?

  Previously this path relied on `filter.IsRegionReplicated`, which falls back to `MaxReplicas` + no learners when placement rules are disabled, so affinity groups could still be scheduled. With this early return, a cluster with `affinity-schedule-limit` enabled but placement rules disabled silently loses affinity scheduling, and I don't see a config/API-level check that forbids this combination.

  If this behavior change is not intentional, should we keep the old fallback when placement rules are disabled, and only use  `isRegionPlacementRuleSatisfiedWithBestLocation` when placement rules are enabled?

pkg/schedule/checker/replica_checker.go
```diff
@@ -191,8 +191,8 @@ func (c *ReplicaChecker) checkRemoveExtraReplica(region *core.RegionInfo) *opera
 	}
 	log.Debug("region has more than max replicas", zap.Uint64("region-id", region.GetID()), zap.Int("peers", len(region.GetPeers())))
 	regionStores := c.cluster.GetRegionStores(region)
-	old := c.strategy(region).SelectStoreToRemove(regionStores)
-	if old == 0 {
+	old, filterByTempState := c.strategy(region).SelectStoreToRemove(regionStores)
```
lhy1024: This `filterByTempState` branch leaks an affinity-specific temporary-state semantic into the regular replica checker, and may affect scheduling performance / convergence efficiency outside affinity.

  Previously `SelectStoreToRemove` selected a removable source after applying the strict source filter; if the worst source was temporarily blocked by busy state, remove limit, snapshot throttling, etc., the replica checker could still pick another schedulable, slightly less ideal source and continue remove-extra or location-replacement scheduling. Now `checkRemoveExtraReplica` / `checkLocationReplacement` skip entirely when `filterByTempState` is true, so these normal replica fixes wait for the temporary state to clear instead of using an available candidate peer. This can slow convergence for extra replica cleanup and location replacement, and may increase scheduling backlog.

  I suggest keeping this affinity-specific two-stage source selection out of `ReplicaChecker`, e.g. scope it to the affinity better-location check or add a dedicated affinity helper. `ReplicaChecker` should preserve the old behavior and skip only when `old == 0`. If this is an intentional global scheduling policy change, please add dedicated replica checker coverage and explain why the scheduling performance / convergence impact is acceptable.

pkg/schedule/checker/affinity_checker.go
```diff
@@ -93,7 +100,7 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
 		affinityCheckerUnhealthyRegionCounter.Inc()
 		return nil
 	}
-	if !filter.IsRegionReplicated(c.cluster, region) {
+	if !c.isRegionPlacementRuleSatisfiedWithBestLocation(region, true /* isExistingRegion */) {
```
lhy1024: Can we add an affinity-group fast path before running `isRegionPlacementRuleSatisfiedWithBestLocation`?

  `AffinityChecker.Check` can be called for regular regions whenever `affinity-schedule-limit > 0`. With the current order, we run `FitRegion` + `getBetterLocation` before knowing whether the region belongs to any affinity group. This is noticeably heavier than the previous `filter.IsRegionReplicated` check because it iterates RuleFits and runs source/target store filtering. Even when there is no affinity group, or when this region is unrelated to any group, we still pay that cost first.

  At minimum, please add a `!c.hasAffinityGroups()` fast path before this check. A stronger version would be to do a read-only group lookup first and only run the best-location gate for regions that actually belong to an affinity group.

rleungx (approved):

pkg/schedule/checker/affinity_checker.go
```diff
@@ -93,15 +100,21 @@ func (c *AffinityChecker) Check(region *core.RegionInfo) []*operator.Operator {
...
-		affinityCheckerAbnormalReplicaCounter.Inc()
+
+	if !c.hasAffinityGroups() {
 		return nil
 	}
 
-	// Get the affinity group for this region
+	// Check the affinity group before the heavier best-location gate. Most regions
+	// are not affinity regions.
 	group, isAffinity := c.affinityManager.GetAndCacheRegionAffinityGroupState(region)
```
lhy1024, line 110: GetAndCacheRegionAffinityGroupState itself is not new, but this patch moves it before isRegionPlacementRuleSatisfiedWithBestLocation. Since this call writes the affinity cache and updates RegionCount/AffinityRegionCount, a region can be recorded as affinity/stable even if the later placement/best-location gate returns false. Could we use the read-only GetRegionAffinityGroupState for the fast path first, and only write the cache after the gate
  passes?

PR #10194, opened 2026-01-28
feat(core): optimize region query efficiency with batched lookups

### LINKED ISSUE #10193: QueryRegion should handle by-ID lock acquisition in batches
## Enhancement Task

As the most common request type among `QueryRegion` gRPC stream requests, the PD server does not handle by-ID lock acquisition in batches, resulting in significant lock fetching overhead.

https://github.com/tikv/pd/blob/7a66bfb0214b7332f9d09524e39576076fcad600/pkg/core/region.go#L1602-L1628

The lock acquisition here should be optimized, and a suitable batch size should be found through benchmarking.

JmPotato:

Issue Number: close #10193.

```commit-message
- Add `getRegionsByIDs` for efficient batch retrieval from region map, reducing lock contention.
- Increase default `batchSearchSize` from 16 to 128 to improve efficiency.
- Add benchmark tests (`BenchmarkQueryRegions`) to evaluate performance across region and query sizes.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Refactor**
  * Improved region querying performance through enhanced batch processing and reduced synchronization overhead.

* **Tests**
  * Added comprehensive benchmark tests covering various region query scenarios and scales.

### DIFF
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -39,6 +39,7 @@ import (
 	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/slice"
 	"github.com/tikv/pd/pkg/utils/keyutil"
 	"github.com/tikv/pd/pkg/utils/logutil"
 	"github.com/tikv/pd/pkg/utils/syncutil"
@@ -49,7 +50,8 @@ const (
 	randomRegionMaxRetry = 10
 	// ScanRegionLimit is the default limit for the number of regions to scan in a region scan request.
 	ScanRegionLimit = 1000
-	batchSearchSize = 16
+	// batchSearchSize is the default size for the number of IDs/keys/prevKeys to search in a batch.
+	batchSearchSize = 128
 	// CollectFactor is the factor to collect the count of region.
 	CollectFactor = 0.9
 )
@@ -1576,10 +1578,6 @@ func (r *RegionsInfo) QueryRegions(
 		start = time.Now()
 		regions = r.getRegionsByKeys(keys)
 		queryRegionByKeysDuration.Observe(time.Since(start).Seconds())
-		// Assert the returned regions count matches the input keys.
-		if len(regions) != len(keys) {
-			panic("returned regions count mismatch with the input keys")
-		}
 	}
 
 	// Iterate the prevKeys to find the regions.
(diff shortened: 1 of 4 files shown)


pkg/core/region.go
```diff
@@ -1668,6 +1656,19 @@ func (r *RegionsInfo) getRegionsByPrevKeys(prevKeys [][]byte) []*RegionInfo {
 	return regions
 }
 
+// getRegionsByIDs searches RegionInfo from regionMap by IDs in batch.
+func (r *RegionsInfo) getRegionsByIDs(ids []uint64) []*RegionInfo {
+	regions := make([]*RegionInfo, 0, len(ids))
+	for _, batch := range slice.SplitIntoBatches(ids, batchSearchSize) {
```
rleungx, line 1662: Will it have a range overlap between IDs if we split into batches?

JmPotato: In theory, this is possible if a split-induced meta update occurs during the interval when the lock is released in a batch, and the left-side ID after the split appears in the next batch of queries. However, I believe this is unlikely in actual `QueryRegion` scenarios. Even if it does occur, PD’s region meta information only guarantees eventual consistency externally, which does not appear to violate our defined consistency primitives.

JmPotato: As for the merge case, it's similar. Though we can introduce another iteration to check the overlap, I don't think it is necessary since the old unary gRPC call has the same problem.

pkg/core/region.go
```diff
@@ -1588,10 +1586,6 @@ func (r *RegionsInfo) QueryRegions(
 		start = time.Now()
 		prevRegions = r.getRegionsByPrevKeys(prevKeys)
 		queryRegionByPrevKeysDuration.Observe(time.Since(start).Seconds())
-		// Assert the returned regions count matches the input keys.
-		if len(prevRegions) != len(prevKeys) {
```
bufferflies, line 1592: Why was this removed?

JmPotato: This assert logic was originally added to test the rigor of the code implementation. Now that the code logic is stable and will be used in production, it is no longer necessary nor appropriate to verify it through asserts.

pkg/core/region.go
```diff
@@ -49,7 +50,8 @@ const (
 	randomRegionMaxRetry = 10
 	// ScanRegionLimit is the default limit for the number of regions to scan in a region scan request.
 	ScanRegionLimit = 1000
-	batchSearchSize = 16
+	// batchSearchSize is the default size for the number of IDs/keys/prevKeys to search in a batch.
+	batchSearchSize = 128
```
okJiang, line 54: How about adding the bench result here.

PR #10199, opened 2026-02-02
keyspace: add preconditions for keyspace config update

### LINKED ISSUE #10214: PD: Support CAS-style preconditions for keyspace config updates
## Enhancement Task
PD keyspace config is planned to be used by the TiKV encryption key rotation workflow to store the current and target encryption config version. While the workflow is intended to have a single active actor, unexpected concurrent executions can cause config updates to overwrite each other and break the intended state transition.

PD currently lacks support for conditional (CAS-style) updates, making it impossible to safely express transitions such as setting next_encryption_file_id only if it is absent or advancing current_* only if it matches an expected value.

This issue proposes adding simple preconditions (for example, equal / absent checks) to keyspace config updates so state transitions can be performed safely.

hbisheng:

Issue Number: Close #10214

In the CMEK / encryption key rotation workflow, PD keyspace config is planned to be used to store the current and target encryption config version. Although the workflow is designed to have a single active actor, unexpected concurrent executions (for example, retries or overlapping operations) can lead to config updates overwriting each other and breaking the intended state transition.

This PR adds CAS-style guards to keyspace config updates so that encryption key rotation can advance safely and only when the system is in the expected state. With this, callers can, for example, set `next_encryption_file_id` only if it is absent, or advance `current_encryption_file_id` only if it matches `next_encryption_file_id`.

```commit-message
PD keyspace config updates are extended with optional preconditions that
must be satisfied for the update to succeed. These preconditions act as
lightweight CAS checks.

Changes:
- Extend /pd/api/v2/keyspaces/{name}/config API to accept preconditions
  for CAS-like config updates (equal / absent).
- Add pd-ctl support: keyspace update-config --expect and --expect-absent.
```

### Check List

Tests

- [x] Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Conditional keyspace configuration updates via preconditions in the API.
  * CLI flags --expect and --expect-absent to specify preconditions when updating configs.
  * New distinct error code for keyspace config precondition failures.

* **Bug Fixes**
  * Update flow returns HTTP 409 for precondition violations and transactional conflicts.

* **Tests**
  * Added extensive tests (including concurrent scenarios) and a helper to inspect non-OK responses.

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -227,6 +227,8 @@ func (manager *Manager) UpdateConfig(cfg Config) {
 
 // CreateKeyspace create a keyspace meta with given config and save it to storage.
 func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspacepb.KeyspaceMeta, error) {
+	tracer := &createKeyspaceTracer{}
+	tracer.Begin()
 	// Validate purposed name's legality.
 	if err := validateName(request.Name); err != nil {
 		return nil, err
@@ -251,6 +253,10 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	if err != nil {
 		return nil, err
 	}
+	tracer.SetKeyspace(newID, request.Name)
+	tracer.OnAllocateIDFinished()
+
+	// Get keyspace config.
 	userKind := endpoint.StringUserKind(request.Config[UserKindKey])
 	config, err := manager.kgm.GetKeyspaceConfigByKind(userKind)
 	if err != nil {
@@ -273,6 +279,8 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 			request.Config[GCManagementType] = KeyspaceLevelGC
 		}
 	}
+	tracer.OnGetConfigFinished()
+
 	// Create a disabled keyspace meta for tikv-server to get the config on keyspace split.
 	keyspace := &keyspacepb.KeyspaceMeta{
 		Id:             newID,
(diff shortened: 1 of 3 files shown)


hbisheng:
@coderabbitai review

rleungx:
@ystaticy PTAL

pkg/keyspace/keyspace.go
```diff
@@ -645,6 +645,68 @@ const (
...
+// - key -> "value" means the key must exist and equal "value".
+func (manager *Manager) UpdateKeyspaceConfigWithPreconditions(name string, mutations []*Mutation, preconditions map[string]*string) (*keyspacepb.KeyspaceMeta, error) {
+	if len(preconditions) == 0 {
+		return manager.UpdateKeyspaceConfig(name, mutations)
+	}
+	return manager.updateKeyspaceConfigTxn(name, func(meta *keyspacepb.KeyspaceMeta) error {
+		if err := checkKeyspaceConfigPreconditions(meta.GetConfig(), preconditions); err != nil {
+			return err
+		}
+		for _, mutation := range mutations {
```
rleungx: How about using a function?

hbisheng: Added a `applyKeyspaceConfigMutations` function

hbisheng: Added an `applyKeyspaceConfigMutations` function

hbisheng:
Would appreciate a review when you have time. Thanks! @ystaticy @lhy1024 @rleungx

server/apiv2/handlers/keyspace.go
```diff
@@ -350,8 +355,16 @@ func UpdateKeyspaceConfig(c *gin.Context) {
 		}
 	}
 
-	meta, err := manager.UpdateKeyspaceConfig(name, mutations)
+	meta, err := manager.UpdateKeyspaceConfigWithPreconditions(name, mutations, configParams.Preconditions)
 	if err != nil {
+		if goerrors.Is(err, errs.ErrKeyspaceConfigPreconditionFailed) {
+			c.AbortWithStatusJSON(http.StatusConflict, err.Error())
```
lhy1024, line 362: We also need to update swagger comments for 409.

hbisheng: Good catch - added a line of comment for it

PR #10212, opened 2026-02-04
mcs: scheduling mcs support gctuner

bufferflies:

Issue Number: Ref #10213

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Configurable GC tuner with enable flag, GC threshold, server memory limit, and memory-trigger settings.
  * Dynamic GC tuner initialization and live updates from persisted server config; exposed across watcher, server APIs, and cluster components.
* **Bug Fixes**
  * Improved thread-safety and stability for GC tuner initialization, updates, and stopping behavior.
* **Tests**
  * Added unit and integration tests for GC tuner init, config watch, and dynamic updates.

### DIFF
--- a/pkg/gctuner/tuner.go
+++ b/pkg/gctuner/tuner.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/log"
 
 	util "github.com/tikv/pd/pkg/gogc"
+	"github.com/tikv/pd/pkg/memory"
 )
 
 var (
@@ -190,4 +191,96 @@ func calcGCPercent(inuse, threshold uint64) uint32 {
 		return maxGCPercent.Load()
 	}
 	return gcPercent
+}
+
+// Config holds the configuration for GC tuner initialization/update.
+type Config struct {
+	EnableGOGCTuner            bool
+	GCTunerThreshold           float64
+	ServerMemoryLimit          float64
+	ServerMemoryLimitGCTrigger float64
+}
+
+// State holds the state for detecting config changes.
+type State struct {
+	totalMem                  uint64
+	EnableGCTuner             bool
+	GCThresholdBytes          uint64
+	MemoryLimitBytes          uint64
+	MemoryLimitGCTriggerRatio float64
+	MemoryLimitGCTriggerBytes uint64
+}
+
+// InitGCTuner initializes the GC tuner with the given config and total memory.
+// Returns the initial state that can be used for subsequent updates.
+func InitGCTuner(totalMem uint64, cfg *Config) *State {
+	state := &State{totalMem: totalMem}
+	state.EnableGCTuner = cfg.EnableGOGCTuner
+	state.MemoryLimitBytes = uint64(float64(totalMem) * cfg.ServerMemoryLimit)
+	state.GCThresholdBytes = uint64(float64(state.MemoryLimitBytes) * cfg.GCTunerThreshold)
+	if state.MemoryLimitBytes == 0 {
+		state.GCThresholdBytes = uint64(float64(totalMem) * cfg.GCTunerThreshold)
+	}
+	state.MemoryLimitGCTriggerRatio = cfg.ServerMemoryLimitGCTrigger
+	state.MemoryLimitGCTriggerBytes = uint64(float64(state.MemoryLimitBytes) * state.MemoryLimitGCTriggerRatio)
+
+	// Initialize GC tuner
+	Tuning(state.GCThresholdBytes)
+	EnableGOGCTuner.Store(state.EnableGCTuner)
+	log.Info("initialize gc tuner",
+		zap.Bool("enable-gc-tuner", state.EnableGCTuner),
+		zap.Uint64("gc-threshold-bytes", state.GCThresholdBytes))
+
+	// Initialize memory limit tuner
+	memory.ServerMemoryLimit.Store(state.MemoryLimitBytes)
+	GlobalMemoryLimitTuner.SetPercentage(state.MemoryLimitGCTriggerRatio)
+	GlobalMemoryLimitTuner.UpdateMemoryLimit()
+	log.Info("initialize gc memory limit",
+		zap.Uint64("memory-limit-bytes", state.MemoryLimitBytes),
+		zap.Float64("memory-limit-gc-trigger-ratio", state.MemoryLimitGCTriggerRatio))
+
+	return state
+}
+
+// UpdateIfNeeded checks if the config has changed and updates the GC tuner.
+// Returns true if any updates were made.
+func (s *State) UpdateIfNeeded(cfg *Config) bool {
+	updated := false
+
+	newEnableGCTuner := cfg.EnableGOGCTuner
+	newMemoryLimitBytes := uint64(float64(s.totalMem) * cfg.ServerMemoryLimit)
+	newGCThresholdBytes := uint64(float64(newMemoryLimitBytes) * cfg.GCTunerThreshold)
+	if newMemoryLimitBytes == 0 {
+		newGCThresholdBytes = uint64(float64(s.totalMem) * cfg.GCTunerThreshold)
+	}
+	newMemoryLimitGCTriggerRatio := cfg.ServerMemoryLimitGCTrigger
+	newMemoryLimitGCTriggerBytes := uint64(float64(newMemoryLimitBytes) * newMemoryLimitGCTriggerRatio)
+
+	// Check and update GC tuner
+	if newEnableGCTuner != s.EnableGCTuner || newGCThresholdBytes != s.GCThresholdBytes {
+		s.EnableGCTuner = newEnableGCTuner
+		s.GCThresholdBytes = newGCThresholdBytes
+		Tuning(s.GCThresholdBytes)
+		EnableGOGCTuner.Store(s.EnableGCTuner)
+		log.Info("update gc tuner",
+			zap.Bool("enable-gc-tuner", s.EnableGCTuner),
+			zap.Uint64("gc-threshold-bytes", s.GCThresholdBytes))
+		updated = true
+	}
+
+	// Check and update memory limit tuner
+	if newMemoryLimitBytes != s.MemoryLimitBytes || newMemoryLimitGCTriggerBytes != s.MemoryLimitGCTriggerBytes {
+		s.MemoryLimitBytes = newMemoryLimitBytes
+		s.MemoryLimitGCTriggerBytes = newMemoryLimitGCTriggerBytes
+		s.MemoryLimitGCTriggerRatio = newMemoryLimitGCTriggerRatio
+		memory.ServerMemoryLimit.Store(s.MemoryLimitBytes)
+		GlobalMemoryLimitTuner.SetPercentage(s.MemoryLimitGCTriggerRatio)
+		GlobalMemoryLimitTuner.UpdateMemoryLimit()
+		log.Info("update gc memory limit",
+			zap.Uint64("memory-limit-bytes", s.MemoryLimitBytes),
+			zap.Float64("memory-limit-gc-trigger-ratio", s.MemoryLimitGCTriggerRatio))
+		updated = true
+	}
+
+	return updated
 }
--- a/pkg/gctuner/tuner.go
+++ b/pkg/gctuner/tuner.go
@@ -269,10 +269,10 @@ func (s *State) UpdateIfNeeded(cfg *Config) bool {
 	}
 
 	// Check and update memory limit tuner
-	if newMemoryLimitBytes != s.MemoryLimitBytes || newMemoryLimitGCTriggerBytes != s.MemoryLimitGCTriggerBytes {
+	if newMemoryLimitBytes != s.MemoryLimitBytes || newMemoryLimitGCTriggerRatio != s.MemoryLimitGCTriggerRatio {
 		s.MemoryLimitBytes = newMemoryLimitBytes
-		s.MemoryLimitGCTriggerBytes = newMemoryLimitGCTriggerBytes
 		s.MemoryLimitGCTriggerRatio = newMemoryLimitGCTriggerRatio
+		s.MemoryLimitGCTriggerBytes = newMemoryLimitGCTriggerBytes
 		memory.ServerMemoryLimit.Store(s.MemoryLimitBytes)
 		GlobalMemoryLimitTuner.SetPercentage(s.MemoryLimitGCTriggerRatio)
 		GlobalMemoryLimitTuner.UpdateMemoryLimit()
--- a/pkg/gctuner/tuner.go
+++ b/pkg/gctuner/tuner.go
@@ -65,10 +65,12 @@ func init() {
 // When Tuning, the env GOGC will not be take effect.
 // threshold: disable tuning if threshold == 0
 func Tuning(threshold uint64) {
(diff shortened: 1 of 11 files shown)


okJiang (commented):
```go
// Tuning sets the threshold of heap which will be respect by gogc tuner.
// When Tuning, the env GOGC will not be take effect.
// threshold: disable tuning if threshold == 0
func Tuning(threshold uint64) {
	// disable gc tuner if percent is zero
	if t := globalTuner.Load(); t == nil {
		t1 := newTuner(threshold)
		globalTuner.CompareAndSwap(nil, &t1)
	} else {
		if threshold <= 0 {
			(*t).stop()
			globalTuner.CompareAndSwap(t, nil)
		} else {
			(*t).setThreshold(threshold)
		}
	}
}
```

The `Tuning` annotation says that `threshold==0` is disabled, but `newTuner(0)` still starts the finalizer loop when `globalTuner` is empty; And if `CompareAndSwap(nil, &t1)` fails during concurrent initialization, the tuner created by the losing party will not stop (it may survive for a long time and participate in parameter adjustment when enabled later).

okJiang:
Do other microservice components also need to support gctunner?

rleungx:
> Do other microservice components also need to support gctunner?

We should support it, and we can enable it for the specified case.

bufferflies:
> ```go
> // Tuning sets the threshold of heap which will be respect by gogc tuner.
> // When Tuning, the env GOGC will not be take effect.
> // threshold: disable tuning if threshold == 0
> func Tuning(threshold uint64) {
> 	// disable gc tuner if percent is zero
> 	if t := globalTuner.Load(); t == nil {
> 		t1 := newTuner(threshold)
> 		globalTuner.CompareAndSwap(nil, &t1)
> 	} else {
> 		if threshold <= 0 {
> 			(*t).stop()
> 			globalTuner.CompareAndSwap(t, nil)
> 		} else {
> 			(*t).setThreshold(threshold)
> 		}
> 	}
> }
> ```
> 
> The `Tuning` annotation says that `threshold==0` is disabled, but `newTuner(0)` still starts the finalizer loop when `globalTuner` is empty; And if `CompareAndSwap(nil, &t1)` fails during concurrent initialization, the tuner created by the losing party will not stop (it may survive for a long time and participate in parameter adjustment when enabled later).

Yes, the tuning should fix this bug

bufferflies:
> > ```go
> > // Tuning sets the threshold of heap which will be respect by gogc tuner.
> > // When Tuning, the env GOGC will not be take effect.
> > // threshold: disable tuning if threshold == 0
> > func Tuning(threshold uint64) {
> > 	// disable gc tuner if percent is zero
> > 	if t := globalTuner.Load(); t == nil {
> > 		t1 := newTuner(threshold)
> > 		globalTuner.CompareAndSwap(nil, &t1)
> > 	} else {
> > 		if threshold <= 0 {
> > 			(*t).stop()
> > 			globalTuner.CompareAndSwap(t, nil)
> > 		} else {
> > 			(*t).setThreshold(threshold)
> > 		}
> > 	}
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `Tuning` annotation says that `threshold==0` is disabled, but `newTuner(0)` still starts the finalizer loop when `globalTuner` is empty; And if `CompareAndSwap(nil, &t1)` fails during concurrent initialization, the tuner created by the losing party will not stop (it may survive for a long time and participate in parameter adjustment when enabled later).
> 
> Yes, have fixed this bug

okJiang (approved):

pkg/gctuner/tuner.go
```diff
@@ -64,10 +65,12 @@ func init() {
...
 func Tuning(threshold uint64) {
-	// disable gc tuner if percent is zero
 	if t := globalTuner.Load(); t == nil {
-		t1 := newTuner(threshold)
-		globalTuner.CompareAndSwap(nil, &t1)
+		// init gc tuner only when threshold > 0, otherwise do nothing
+		if threshold > 0 {
+			t1 := newTuner(threshold)
+			globalTuner.CompareAndSwap(nil, &t1)
+		}
```
lhy1024, line 76: `threshold == 0` is fixed now, but the concurrent first-time initialization case is still unsafe. If two goroutines call `Tuning(threshold > 0)` while `globalTuner` is nil, both of them will create `newTuner(threshold)` before the CAS. Since `newTuner()` already starts the finalizer loop, the loser of `tuning()` on later GCs once `EnableGOGCTuner` is enabled.

pkg/mcs/scheduling/server/config/config.go
```diff
@@ -80,6 +81,7 @@ type Config struct {
 
 	Schedule    sc.ScheduleConfig    `toml:"schedule" json:"schedule"`
 	Replication sc.ReplicationConfig `toml:"replication" json:"replication"`
+	Server      sc.ServerConfig      `toml:"pd-server" json:"pd-server"`
```
lhy1024: In this PR, `PersistConfig` now stores `serverConfig`, and the scheduling config watcher updates runtime GC tuner state from it. However, `Server.GetConfig()` still only refreshes `Schedule`, `Replication`, and `ClusterVersion` from `persistConfig`, while `cfg.Server` still comes from the startup local config (`s.cfg.Clone()`).

It may be worth also exposing/loading the persisted server config here for consistency.

bufferflies: it update when pd server config updated

bufferflies: done

pkg/mcs/scheduling/server/config/config.go
```diff
@@ -291,6 +299,16 @@ func (o *PersistConfig) SetStoreConfig(cfg *sc.StoreConfig) {
 	o.storeConfig.Store(cfg)
 }
 
+func (o *PersistConfig) setServerConfig(cfg *sc.ServerConfig) {
+	// Some of the fields won't be persisted and watched,
+	o.serverConfig.Store(cfg)
+}
```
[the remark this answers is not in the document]
bufferflies: it will cause panic

pkg/gctuner/tuner.go
```diff
@@ -65,18 +65,24 @@ func init() {
...
-		// init gc tuner only when threshold > 0, otherwise do nothing
-		if threshold > 0 {
-			t1 := newTuner(threshold)
-			globalTuner.CompareAndSwap(nil, &t1)
-		}
-	} else {
-		if threshold <= 0 {
-			(*t).stop()
-			globalTuner.CompareAndSwap(t, nil)
+	for range 3 {
```
lhy1024: `Tuning()` now publishes `&t1` into `globalTuner` before calling `t1.start()`. That means another goroutine concurrent `Tuning(0)` wins the CAS-to-nil path, it immediately calls `(*t).stop()`, which dereferences `t.finalizer` and panics.

 I reproduced this locally with a small unit test by placing a `newTunerWithoutStart()` instance into

Also, even if `stop()` were made nil-safe, the publish-before-start ordering would still be problematic: would leave behind a detached started tuner again. So the root issue is that the tuner is visible to other
 goroutines before it is fully initialized.

lhy1024:
```
  func TestTuningDisableHandlesPublishedButNotStartedTuner(t *testing.T) {
        re := require.New(t)

        old := globalTuner.Load()
        t.Cleanup(func() {
                globalTuner.Store(old)
        })

        tn := newTunerWithoutStart(1)
        globalTuner.Store(&tn)

        re.NotPanics(func() {
                Tuning(0)
        })
        re.Nil(globalTuner.Load())
  }
```

bufferflies: Yes, I will add a lock to keep it atomic.

lhy1024 (approved):

rleungx:
@coderabbitai full review

pkg/gctuner/tuner.go
```diff
@@ -191,3 +198,95 @@ func calcGCPercent(inuse, threshold uint64) uint32 {
...
+
+	// Initialize memory limit tuner
+	memory.ServerMemoryLimit.Store(state.MemoryLimitBytes)
+	GlobalMemoryLimitTuner.SetPercentage(state.MemoryLimitGCTriggerRatio)
+	GlobalMemoryLimitTuner.UpdateMemoryLimit()
+	log.Info("initialize gc memory limit",
+		zap.Uint64("memory-limit-bytes", state.MemoryLimitBytes),
+		zap.Float64("memory-limit-gc-trigger-ratio", state.MemoryLimitGCTriggerRatio))
+
+	return state
```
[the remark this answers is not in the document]
bufferflies: Yes, but we want keep the GC parameters still working even if the state has gc

bufferflies:
@coderabbitai full review

pkg/gctuner/tuner.go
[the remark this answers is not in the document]
rleungx: Will it cause a leak?

pkg/mcs/scheduling/server/server.go
```diff
@@ -585,6 +569,7 @@ func (s *Server) GetConfig() *config.Config {
 	cfg.Replication = *s.persistConfig.GetReplicationConfig().Clone()
 	cfg.ClusterVersion = *s.persistConfig.GetClusterVersion()
 	cfg.Schedule.MaxMergeRegionKeys = cfg.Schedule.GetMaxMergeRegionKeys()
+	cfg.Server = *s.persistConfig.GetServerConfig()
```
rleungx: Shall we also modify the getConfig function for /config?

bufferflies: no, it just show the config

bufferflies:
@coderabbitai full review

pkg/mcs/scheduling/server/config/watcher.go
```diff
@@ -90,8 +108,14 @@ func NewWatcher(
 		etcdClient:                etcdClient,
 		PersistConfig:             persistConfig,
 		storage:                   storage,
+		totalMem:                  totalMem,
 	}
-	err := cw.initializeConfigWatcher()
+
+	// Initialize GC tuner
+	// gcCfg := persistConfig.GetGCTunerConfig()
+	// cw.gcTunerState = gctuner.InitGCTuner(totalMem, gcCfg)
```
rleungx: no need anymore?

bufferflies: done

bufferflies:
@coderabbitai full review

pkg/gctuner/tuner.go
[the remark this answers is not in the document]
rleungx: @bufferflies still have a comment unresolved.

bufferflies:
@coderabbitai full review

server/cluster/cluster.go
```diff
@@ -558,75 +558,46 @@ func (c *RaftCluster) stopTSOJobsIfNeeded() {
...
+
+	// Initialize GC tuner using the helper function
+	state := gctuner.InitGCTuner(totalMem, c.getGCTunerConfig())
+
 	for {
 		select {
 		case <-c.ctx.Done():
 			log.Info("gc tuner is stopped")
+			state.Stop() // disable GC tuner when the cluster is stopping
 			return
```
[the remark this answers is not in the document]
bufferflies: It's my mistabke,  we should stop gc tuner if leader or primary changed

pkg/gctuner/tuner.go
```diff
@@ -191,3 +198,100 @@ func calcGCPercent(inuse, threshold uint64) uint32 {
...
+		updated = true
+	}
+
+	return updated
+}
+
+// Stop disables the GC tuner.
+func (s *State) Stop() {
+	Tuning(0) // disable GC tuner
+}
```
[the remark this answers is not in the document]
bufferflies: It must stop GC if the leader or primary changes. I want GC Tuner to just run when it's the leader or primary.

bufferflies:
@coderabbitai full review

pkg/gctuner/tuner.go
```diff
@@ -191,3 +198,100 @@ func calcGCPercent(inuse, threshold uint64) uint32 {
...
+			zap.Uint64("memory-limit-bytes", s.MemoryLimitBytes),
+			zap.Float64("memory-limit-gc-trigger-ratio", s.MemoryLimitGCTriggerRatio))
+		updated = true
+	}
+
+	return updated
+}
+
+// Stop disables the GC tuner.
+func (*State) Stop() {
```
rleungx, line 295: `InitGCTuner` / `UpdateIfNeeded` now also update the process-global memory-limit tuner, but `Stop()` only calls `Tuning(0)`. When the scheduling primary steps down (or the cluster stops), `memory.ServerMemoryLimit` / `debug.SetMemoryLimit` can stay at the last applied value even though this watcher lifecycle is primary-scoped. Should we also reset/stop the memory-limit side here so we don't leave stale process-wide GC state behind?

pkg/mcs/scheduling/server/config/watcher.go
```diff
@@ -136,6 +156,15 @@ func (cw *Watcher) initializeConfigWatcher() error {
 		cw.SetScheduleConfig(&cfg.Schedule)
 		cw.SetReplicationConfig(&cfg.Replication)
 		cw.SetStoreConfig(&cfg.Store)
+		if cfg.Server != nil {
```
rleungx: `cfg.Server` can be `nil` when the persisted config was written before the new `pd-server` field existed. In that case we skip the update here, but `PersistConfig` was initialized from `config.NewConfig()`, so `serverConfig` is still the zero value. That seems to change the effective behavior from PD defaults to `EnableGOGCTuner=false` / zero thresholds until someone persists a newer config. Could we initialize this from PD defaults or add an upgrade-path test for an old persisted config without `pd-server`?

rleungx (commented):
A couple of correctness concerns from the review.

pkg/gctuner/tuner.go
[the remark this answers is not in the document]
bufferflies: done, add a close function to disable the tuner.

pkg/gctuner/tuner.go
bufferflies: Yes, it will keep the mem limit setting, but will stop tuning.

tests/integrations/mcs/scheduling/config_test.go
```diff
@@ -131,6 +131,74 @@ func (suite *configTestSuite) TestConfigWatch() {
...
+		persistConfig(re, suite.pdLeaderServer)
+	})
+	pdServerCfg := origCfg.Clone()
+
+	// test0: config don't contain server-config field, the watcher should use the default value.
+	type FakeConfig struct {
+		FakeField string `json:"fake-field"`
+	}
+	cfg := &FakeConfig{}
+	suite.pdLeaderServer.GetServer().GetStorage().SaveConfig(cfg)
```
rleungx, line 163: CI is failing here because the `SaveConfig` error is ignored. Please check it with `re.NoError(...)`.

PR #10224, opened 2026-02-05
server: add server-side TSO batch size observability

okJiang:

`pd-tso-bench` can generate batched requests, but the existing Grafana panel relies on the client-side metric `pd_client_request_handle_tso_batch_size`. In stress scenarios this client metric can be missing or unrepresentative, making it hard to judge real server-side batching behavior from PD monitoring.

Issue Number: ref #10228

This PR changes TSO batch-size observability to use PD server metrics directly:

1. Add a server histogram metric `pd_server_handle_tso_batch_size` in `server/metrics.go`.
2. Record request batch size in PD TSO handling path (`server/grpc_service.go`).
3. Update Grafana panel `TSO request batch size` in `metrics/grafana/pd.json` to query `pd_server_handle_tso_batch_size` (`_bucket/_sum/_count`) instead of the client metric.
4. Avoid double-counting in microservices forwarding path by not recording batch size on the forwarding branch.

```commit-message
server: add server-side TSO batch size observability
```

### Check List

Tests

- Unit test

Manual test steps:

1. `make failpoint-enable`
2. `go test -tags without_dashboard ./tests/server -run TestUpdateAdvertiseUrls -count=1`
3. `make failpoint-disable`

Code changes

- Has the configuration change: No
- Has HTTP API interfaces changed: No
- Has persistent data change: No

Side effects

- Possible performance regression: No
- Increased code complexity: Low (metric instrumentation and dashboard query switch)
- Breaking backward compatibility: No

Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: N/A
- PR to update `pingcap/tiup`: N/A
- Need to cherry-pick to the release branch: TBD

### Release note

```release-note
Add a PD server-side TSO batch size metric and switch the Grafana TSO batch-size panel to use server-side data.
```

## Summary by CodeRabbit

* **New Features**
  * Added server-side metrics for TSO batch size performance monitoring to Grafana dashboard, including 99th percentile, 90th percentile, and average batch size visualizations alongside existing client-side metrics.

### DIFF
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -11003,7 +11003,7 @@
           "targets": [
             {
               "exemplar": true,
-              "expr": "histogram_quantile(0.99, sum(rate(pd_server_handle_tso_batch_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(rate(pd_client_request_handle_tso_batch_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (le))",
               "hide": false,
               "interval": "",
               "intervalFactor": 1,
@@ -11012,7 +11012,7 @@
             },
             {
               "exemplar": true,
-              "expr": "histogram_quantile(0.90, sum(rate(pd_server_handle_tso_batch_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (le))",
+              "expr": "histogram_quantile(0.90, sum(rate(pd_client_request_handle_tso_batch_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (le))",
               "hide": false,
               "interval": "",
               "intervalFactor": 1,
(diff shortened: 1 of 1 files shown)


server/grpc_service.go
```diff
@@ -569,6 +569,7 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 				err = errs.ErrGenerateTimestamp.FastGenByArgs("tso count should be positive")
 				return errs.ErrUnknown(err)
 			}
+			tsoBatchSize.Observe(float64(request.GetCount()))
```
okJiang: Assuming it's under the microservices mode, it seems that records will be duplicated here, being recorded twice in the PD server and the TSO server.

okJiang: Good catch. I removed `tsoBatchSize.Observe(...)` from the `IsServiceIndependent` forwarding path and now only record `handle_tso_batch_size` on the local `GenerateTSO` path, so PD no longer double-counts with TSO server metrics in microservices mode. Updated in commit f016eaed7.

tests/cluster.go
```diff
@@ -142,22 +143,35 @@ func NewTestServer(ctx context.Context, cfg *config.Config, services []string, h
...
 	s.Lock()
-	defer s.Unlock()
 	if s.state != Initial && s.state != Stop {
+		s.Unlock()
 		return errors.Errorf("server(state%d) cannot run", s.state)
 	}
+	s.state = Starting
+	s.Unlock()
+
 	if err := s.server.Run(); err != nil {
```
liyishuai, line 149: Is it safe to run `s.server.Run()` outside the lock?

okJiang: Good point. I tightened this path to make startup/stop safe without relying on `Close()` timing: (1) `Run()` now tracks a `runDone` channel and, after `server.Run()` returns, closes the server if a concurrent Stop/Destroy was requested; (2) `Stop()`/`Destroy()` set state first, call `Close()`, and wait for `runDone` when state was `Starting`; (3) `Destroy()` removes data dir only after startup completes. This avoids the window where `Close()` can be a no-op before `isRunning=1`. Updated in commit 4c7648b1a.

okJiang: The amount of changes for this unstable test exceeded my estimate, so I will open another issue and then fix it in a separate PR later.

metrics/grafana/pd.json
```diff
@@ -11003,7 +11003,7 @@
           "targets": [
             {
               "exemplar": true,
-              "expr": "histogram_quantile(0.99, sum(rate(pd_client_request_handle_tso_batch_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (le))",
```
rleungx, line 11006: no need `pd_client_request_handle_tso_batch_size_sum` anymore?

okJiang: The values they represent should be the same.

server/grpc_service.go
```diff
@@ -584,6 +584,7 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 			return errs.ErrMismatchClusterID(clusterID, request.GetHeader().GetClusterId())
 		}
 		count := request.GetCount()
+		tsoBatchSize.Observe(float64(count))
```
bufferflies, line 587: Do we need a keyspace group dimension?

okJiang: It's definitely worth considering, but it could also cause metrics to explode.

metrics/grafana/pd.json
okJiang: Are you asking whether we need to delete the client metrics? I tend to not delete it for now

PR #10227, opened 2026-02-06
rm: introduce manager write roles and gates

okJiang:

Issue Number: ref https://github.com/tikv/pd/issues/9737

To migrate RM metadata writes to PD in a controlled way, we need explicit and testable write ownership in the manager layer first. Without write-role gates, later cutover steps will mix metadata/state persistence responsibilities and are easy to regress.

- Add `ResourceGroupWriteRole` with three modes: `LegacyAll`, `PDMetaOnly`, `RMTokenOnly`.
- Add role gate helpers (`AllowsMetadataWrite`, `AllowsStateWrite`) and `errMetadataWriteDisabled`.
- Wire optional write-role provider into `NewManager`, defaulting to `LegacyAll` (no behavior change by default).
- Guard metadata write paths (`Add/Modify/DeleteResourceGroup`, `UpdateControllerConfigItem`, `SetKeyspaceServiceLimit`).
- Propagate role into keyspace manager/service limiter persistence behavior:
  - metadata persistence only when metadata writes are allowed
  - state persistence loop only when state writes are allowed
  - service-limit load path uses non-persisting apply
- Update RM REST service-limit setter to handle the new `SetKeyspaceServiceLimit` error return.
- Add unit tests for manager/keyspace write-role gates.

```commit-message
resourcemanager: add write roles and metadata gates
```

### Check List

Tests

- Unit test

Code changes

- Increased code complexity

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn): None
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup): None
- Need to cherry-pick to the release branch: No

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Write-role based access control for resource group and service-limit operations so persistence and state updates respect configured write permissions.
  * Service limit management now supports explicit persist vs. non-persist updates and cleans up or adjusts behavior when limits are changed (including zero/negative).

* **Bug Fixes**
  * Service limit API now surfaces errors properly (permission-denied vs. internal errors) instead of silently ignoring failures.

### DIFF
--- a/pkg/mcs/resourcemanager/server/keyspace_manager.go
+++ b/pkg/mcs/resourcemanager/server/keyspace_manager.go
@@ -67,7 +67,7 @@ type keyspaceResourceGroupManager struct {
 	syncutil.RWMutex
 	groups          map[string]*ResourceGroup
 	groupRUTrackers map[string]*groupRUTracker
-	sl              *serviceLimiter
+	serviceLimiter  *serviceLimiter
 
 	keyspaceID uint32
 	storage    endpoint.ResourceGroupStorage
@@ -89,7 +89,7 @@ func newKeyspaceResourceGroupManager(
 		keyspaceID:      keyspaceID,
 		storage:         storage,
 		writeRole:       writeRole,
-		sl:              newServiceLimiter(keyspaceID, 0, storage, writeRole),
+		serviceLimiter:  newServiceLimiter(keyspaceID, 0, storage, writeRole),
 	}
 }
 
@@ -181,13 +181,12 @@ func (krgm *keyspaceResourceGroupManager) modifyResourceGroup(group *rmpb.Resour
 	if !ok {
 		return errs.ErrResourceGroupNotExists.FastGenByArgs(group.Name)
 	}
-
+	if !krgm.writeRole.AllowsMetadataWrite() {
+		return errMetadataWriteDisabled
+	}
 	err := curGroup.PatchSettings(group)
 	if err != nil {
 		return err
-	}
-	if !krgm.writeRole.AllowsMetadataWrite() {
-		return nil
 	}
 	return curGroup.persistSettings(krgm.keyspaceID, krgm.storage)
 }
@@ -202,10 +201,11 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
 	if !ok {
 		return errs.ErrResourceGroupNotExists.FastGenByArgs(name)
 	}
-	if krgm.writeRole.AllowsMetadataWrite() {
-		if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
-			return err
-		}
+	if !krgm.writeRole.AllowsMetadataWrite() {
+		return errMetadataWriteDisabled
+	}
+	if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
+		return err
 	}
 	krgm.Lock()
 	delete(krgm.groups, name)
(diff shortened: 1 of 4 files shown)


pkg/mcs/resourcemanager/server/write_role.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2025 TiKV Project Authors.
```
okJiang:
```suggestion
// Copyright 2026 TiKV Project Authors.
```

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -172,6 +186,9 @@ func (krgm *keyspaceResourceGroupManager) modifyResourceGroup(group *rmpb.Resour
 	if err != nil {
 		return err
 	}
+	if !krgm.writeRole.AllowsMetadataWrite() {
+		return nil
+	}
```
lhy1024: Do we need to check AllowsMetadataWrite first and PatchSettings secondly?

okJiang: ~PatchSettings is an in-memory operation, I don't think it needs to be prohibited?~

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -71,15 +71,25 @@ type keyspaceResourceGroupManager struct {
...
+		writeRole = writeRoles[0]
+	}
 	return &keyspaceResourceGroupManager{
 		groups:          make(map[string]*ResourceGroup),
 		groupRUTrackers: make(map[string]*groupRUTracker),
 		keyspaceID:      keyspaceID,
 		storage:         storage,
-		sl:              newServiceLimiter(keyspaceID, 0, storage),
+		writeRole:       writeRole,
+		sl:              newServiceLimiter(keyspaceID, 0, storage, writeRole),
```
bufferflies: how about rename as serviceLimiter not use short name

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -185,8 +202,10 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
 	if !ok {
 		return errs.ErrResourceGroupNotExists.FastGenByArgs(name)
 	}
-	if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
-		return err
+	if krgm.writeRole.AllowsMetadataWrite() {
+		if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
```
bufferflies: How about checking the meta role in the DeleteResourceGroupSetting function?

bufferflies:
Can we prmote/demote the write role by dynamic ?

okJiang:
> Can we prmote/demote the write role by dynamic ?

Currently write role is not dynamically promoted/demoted at runtime.
It is read from config/provider during manager initialization and then treated as static for the process lifetime.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: Agreed. Renamed the short field name from `sl` to `serviceLimiter` in `keyspaceResourceGroupManager` and updated all references for clarity. Addressed in commit 1e8c73568.

okJiang:
Also addressed the recent CodeRabbit nitpicks in commit 1e8c73568:
- switched `IsMetadataWriteDisabledError` to standard `errors.Is` semantics
- updated `SetKeyspaceServiceLimit` swagger annotations with `@Failure 403` and `@Failure 500`

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: `DeleteResourceGroupSetting` is a function at the storage layer and should not introduce `manager` logic.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: Good catch. I have updated `modifyResourceGroup` to check `AllowsMetadataWrite()` before `PatchSettings`; when metadata write is disabled it now returns `errMetadataWriteDisabled` directly. This avoids in-memory-only mutation in this path. Addressed in commit 1e8c73568.

bufferflies (approved):

PR #10234, opened 2026-02-11
resourcemanager: add settings-only apply path

okJiang:

Issue Number: ref #9737

`PatchSettings` currently couples metadata settings update and token delta patch together.
For the upcoming PD-side metadata write path, we need a settings-only apply path to avoid
changing runtime tokens when only metadata is updated.

- Refactor `ResourceGroup.PatchSettings` to shared internal logic.
- Add `ResourceGroup.ApplySettings`, which applies settings without patching token delta.
- Deep copy token bucket settings in `NewGroupTokenBucket` and guarded settings copy in `patch`.
- Add unit test `TestApplySettingsWithoutTokenPatch` to assert fill/burst update but token unchanged.

```commit-message
resourcemanager: add settings-only apply path
```

### Check List

Tests

- Unit test

Code changes

- No configuration change
- No HTTP API interface change
- No persistent data schema change

Side effects

- Possible performance regression: no
- Increased code complexity: low
- Breaking backward compatibility: no

Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: N/A
- PR to update `pingcap/tiup`: N/A
- Need to cherry-pick to the release branch: no

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Add ability to apply resource group settings without modifying existing token counts.

* **Refactor**
  * Consolidated settings patching logic for clearer modes (apply vs. patch).
  * Improved token-bucket initialization to clone settings and avoid shared state.

* **Tests**
  * Added test verifying settings can update rate/limit while preserving existing token values.

### DIFF
--- a/pkg/mcs/resourcemanager/server/token_buckets.go
+++ b/pkg/mcs/resourcemanager/server/token_buckets.go
@@ -390,11 +390,7 @@ func (gtb *GroupTokenBucket) patch(tb *rmpb.TokenBucket) {
 	if tb == nil {
 		return
 	}
-	if settings := tb.GetSettings(); settings != nil {
-		setting := proto.Clone(settings).(*rmpb.TokenLimitSettings)
-		gtb.Settings = setting
-		gtb.settingChanged = true
-	}
+	gtb.applySettings(tb)
 
 	// The settings in token is delta of the last update and now.
 	gtb.Tokens += tb.GetTokens()


pkg/mcs/resourcemanager/server/token_buckets.go
JmPotato, line 394: This part can be simplified as `gtb.applySettings(tb)`.

JmPotato (approved):

PR #10246, opened 2026-02-12
rm: add RM metadata handler with redirector scaffold

okJiang:

Issue Number: ref #9737

Tracking issue: https://github.com/tikv/pd/issues/9737
Discussion baseline: https://github.com/tikv/pd/issues/9737#issuecomment-3859834603

This PR removes metadata handler logic drift between PD redirector fallback and
RM service APIs. Before this change, both sides handled the same
`/resource-manager/api/v1/config/**` metadata operations with duplicated
implementations and gradually diverging status-code semantics.

- Add shared package `pkg/mcs/resourcemanager/metadataapi`.
- Add `ConfigService` as the single metadata business-logic implementation.
- Add `Store` interface and `ManagerStore` adapter over `*rmserver.Manager`.
- Redirector side (`pkg/mcs/resourcemanager/redirector/pd_metadata_handler.go`) is now a thin route shell delegating to `ConfigService`.
- RM API side (`pkg/mcs/resourcemanager/server/apis/v1/api.go`) keeps existing route definitions and swagger-commented wrapper methods, but wrappers now delegate to `ConfigService`.
- Keep redirect behavior unchanged:
  - `shouldHandlePDMetadataLocally` remains `false`.
  - Metadata APIs are still redirected to RM primary.
  - Local fallback is still blocked when request carries `X-Forbidden-Forward-To-Microservice: true`.

Unified HTTP semantics in this PR:

- `400 Bad Request`
  - JSON bind failure
  - invalid controller field/type
  - `service_limit < 0`
- `404 Not Found`
  - resource group not found (`GET/PUT/DELETE`)
  - keyspace not found (`/config/keyspace/service-limit*`)
  - keyspace manager not found (service-limit `GET`)
- `403 Forbidden`
  - metadata write disabled (`IsMetadataWriteDisabledError`)
- `500 Internal Server Error`
  - other internal/runtime/storage errors

Keyspace resolution rule in service-limit APIs:

- path param `:keyspace_name` first
- fallback to query `keyspace_name` when path param is empty

In scope:

- merge metadata handler logic into one shared service
- normalize metadata HTTP error-code mapping
- keep redirector wiring consistent with current non-cutover strategy

Out of scope:

- enabling PD-local metadata cutover
- changing redirector global forwarding strategy
- gRPC write-path cutover
- client routing split

Behavior changes:

- `/config/keyspace/service-limit*`: keyspace-not-found now returns `404` (was partly `400` before)
- `PUT /config/group`: non-existing resource group now returns `404`

Risk:

- Main risk is compatibility for callers depending on old error codes.
- Request routing/cutover behavior itself is unchanged.

```commit-message
rm: unify metadata handlers via shared service

Introduce shared metadataapi.ConfigService and delegate both redirector fallback
and RM apis/v1 handlers to the same implementation. Keep metadata redirect
strategy unchanged, and normalize metadata HTTP error-code semantics.
```

### Check List

Tests

- Unit test
  - `go test ./pkg/mcs/resourcemanager/metadataapi ./pkg/mcs/resourcemanager/redirector ./pkg/mcs/resourcemanager/server/apis/v1 -count=1`
- Integration test
  - `make failpoint-enable && (cd tests/integrations && go test ./mcs/resourcemanager -run 'TestResourceManagerAPI|TestResourceManagerRedirector' -count=1); rc=$?; make failpoint-disable; exit $rc`
  - Not run in this PR
- No code
  - N/A

Code changes

  - No
  - Yes (status-code behavior only; no path added/removed)
- Has persistent data change
  - No

Side effects

  - No expected runtime-path regression
- Increased code complexity
  - Yes (new shared package and adapters)
  - Possibly, for clients depending on previous `400` responses in changed cases

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn): N/A
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup): N/A
- Need to cherry-pick to the release branch: No

### Release note

```release-note
Resource Manager: unify metadata handler logic shared by PD redirector fallback and RM APIs, and normalize metadata API error codes (including keyspace-not-found and PUT non-existing group to 404).
```

### DIFF
--- a/tests/integrations/mcs/resourcemanager/api_test.go
+++ b/tests/integrations/mcs/resourcemanager/api_test.go
@@ -30,8 +30,8 @@ import (
 	"github.com/pingcap/failpoint"
 	rmpb "github.com/pingcap/kvproto/pkg/resource_manager"
 
-	"github.com/tikv/pd/pkg/keyspace"
 	"github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/keyspace"
 	"github.com/tikv/pd/pkg/mcs/resourcemanager/server"
 	"github.com/tikv/pd/pkg/mcs/resourcemanager/server/apis/v1"
 	"github.com/tikv/pd/pkg/utils/testutil"


pkg/mcs/resourcemanager/redirector/pd_metadata_handler.go
```diff
@@ -0,0 +1,276 @@
...
+	return handler
+}
+
+func (h *pdMetadataHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.engine.ServeHTTP(w, r)
+}
+
+// TODO: Deduplicate metadata /config handlers with pkg/mcs/resourcemanager/server/apis/v1/api.go
+// once PD and RM handler error mapping can share one abstraction safely.
+func (h *pdMetadataHandler) registerRouter() {
```
rleungx: Why we need it?

okJiang: When deploying RM microservices, PD uses a redirector for forwarding. Implementing a simple handler in the redirector is more lightweight than directly reusing the entire ResourceManager service.

okJiang: `pdMetadataHandler` is used to handle write operations in PD side

pkg/mcs/resourcemanager/redirector/metadata_handler_test.go
```diff
@@ -0,0 +1,347 @@
...
+	bs "github.com/tikv/pd/pkg/basicserver"
+	pderrors "github.com/tikv/pd/pkg/errs"
+	"github.com/tikv/pd/pkg/keyspace/constant"
+	rmserver "github.com/tikv/pd/pkg/mcs/resourcemanager/server"
+	"github.com/tikv/pd/pkg/metering"
+	"github.com/tikv/pd/pkg/utils/apiutil"
+	"github.com/tikv/pd/pkg/utils/testutil"
+)
+
+func TestMain(m *testing.M) {
```
bufferflies, line 30: Do we need to add an integration test for this API in both the non-micro and micro scenarios?

pkg/mcs/resourcemanager/redirector/pd_metadata_handler.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	c.IndentedJSON(http.StatusOK, limiter)
+}
+
+func getServiceLimitTargetKeyspace(c *gin.Context) string {
+	keyspaceName := c.Param("keyspace_name")
+	if keyspaceName != "" {
+		return keyspaceName
+	}
+	return c.Query("keyspace_name")
```
bufferflies: Why support this query? And I think this function can be merged into the getKeyspaceIDByName.

pkg/mcs/resourcemanager/redirector/pd_metadata_handler.go
```diff
@@ -0,0 +1,276 @@
...
+		return
+	}
+	c.String(http.StatusOK, "Success!")
+}
+
+func (h *pdMetadataHandler) getKeyspaceServiceLimit(c *gin.Context) {
+	keyspaceName := getServiceLimitTargetKeyspace(c)
+	keyspaceID, err := h.getKeyspaceIDByName(c, keyspaceName)
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
```
bufferflies: If the keyspace name does not exist, getKeyspaceIDByName will return an error. I don't think this is a bad request; it should return 404.

pkg/mcs/resourcemanager/redirector/metadata_handler_test.go
okJiang: The redirector will not be used in non-microservice mode. Testing under the microservices pattern is in redirector_test.go

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
+// Copyright 2026 TiKV Project Authors.
```
rleungx, line 1: Shall we move it to server/apis/v1?

okJiang: No need, this is a shared library

pkg/mcs/resourcemanager/redirector/redirector.go
```diff
@@ -20,27 +20,66 @@ import (
...
+		Version:    "v1",
+		IsCore:     false,
+		PathPrefix: apis.APIPathPrefix,
+	}, nil
+}
+
+// TODO: Implement a real PD metadata handler.
+func shouldHandlePDMetadataLocally(r *http.Request) bool {
+	// Keep metadata APIs redirected to RM until PD<->RM metadata sync is in place.
+	_ = r
```
rleungx, line 74: Is it expetced?

pkg/mcs/resourcemanager/redirector/redirector.go
```diff
@@ -20,27 +20,66 @@ import (
...
+	return false
+}
+
+func newPDMetadataFallbackHandler(localHandler http.Handler) http.Handler {
```
rleungx: What do you mean pdmetadata?

pkg/mcs/resourcemanager/redirector/redirector.go
```diff
@@ -20,27 +20,66 @@ import (
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Prevent bypassing redirect rules via an injected "forbidden forward" header.
+		if r.Header.Get(apiutil.XForbiddenForwardToMicroserviceHeader) == "true" {
+			http.NotFound(w, r)
```
rleungx, line 90: NotFound?

pkg/mcs/resourcemanager/metadataapi/config_service_test.go
```diff
@@ -0,0 +1,354 @@
...
+	keyspaceIDs                  map[string]uint32
+	validKeyspaceIDs             map[uint32]struct{}
+	groups                       map[string]*rmserver.ResourceGroup
+	serviceLimits                map[uint32]float64
+	addErr                       error
+	setServiceLimitErr           error
+	updatedControllerConfigItems []string
+}
+
+func newTestStore() *testStore {
```
rleungx, line 221: Why do we need the test store?

pkg/mcs/resourcemanager/redirector/redirector.go
okJiang: Yes, this is expected for this PR. `shouldHandlePDMetadataLocally` is intentionally hardcoded to `false` to keep metadata traffic redirected to RM until PD<->RM metadata sync/cutover is ready. This PR only merges handler logic and error semantics, not traffic cutover behavior.

pkg/mcs/resourcemanager/redirector/redirector.go
okJiang: `pdMetadata` here means the PD-side metadata fallback path for `/resource-manager/api/v1/config/**`. It is only a fallback handler, and currently disabled by `shouldHandlePDMetadataLocally=false`.

pkg/mcs/resourcemanager/redirector/redirector.go
okJiang: `http.NotFound` is intentional here. When `X-Forbidden-Forward-To-Microservice=true` is set, we must neither forward nor execute local metadata logic. Returning 404 avoids a header-crafted local bypass route.

pkg/mcs/resourcemanager/metadataapi/config_service_test.go
okJiang: `testStore` is used for deterministic unit coverage of error/status mapping (400/403/404/500) and controller all-or-nothing behavior, where injecting specific branches through a real manager would be brittle. We still keep manager-backed tests for role behavior (for example, metadata-write-disabled -> 403).

pkg/mcs/resourcemanager/redirector/redirector.go
okJiang: Renamed to `rmMetadataHandler`

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
...
+	limiter := s.manager.GetKeyspaceServiceLimiter(keyspaceID)
+	if limiter == nil {
+		return nil, false
+	}
+	return limiter, true
+}
+
+// ConfigService serves resource-manager /config metadata APIs.
+type ConfigService struct {
+	store Store
```
bufferflies:
```suggestion
Store
```

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// Register mounts /config routes onto the provided router group.
+func (s *ConfigService) Register(configEndpoint *gin.RouterGroup) {
+	configEndpoint.POST("/group", s.PostResourceGroup)
+	configEndpoint.PUT("/group", s.PutResourceGroup)
+	configEndpoint.GET("/group/:name", s.GetResourceGroup)
+	configEndpoint.GET("/groups", s.GetResourceGroupList)
+	configEndpoint.DELETE("/group/:name", s.DeleteResourceGroup)
+	configEndpoint.GET("/controller", s.GetControllerConfig)
```
bufferflies, line 137:
```suggestion
	configEndpoint.GET("/controller-config", s.GetControllerConfig)
```

bufferflies:
rest lgtm

pkg/mcs/resourcemanager/metadataapi/config_service.go
okJiang: Good catch. Here `store` is an unexported internal field by design, so keeping it lowercase is intentional. Exporting it to `Store` would increase visibility without functional benefit. If you meant a different naming/style concern, I can adjust accordingly.

pkg/mcs/resourcemanager/metadataapi/config_service.go
okJiang: I prefer to keep `/controller` in this PR to preserve current API compatibility and scope (this PR does not change external paths). Changing to `/controller-config` would be an API behavior change; if needed, we can discuss it in a follow-up with compatibility handling.

bufferflies: How about using the `/controller/config ` path?

okJiang:
ping @rleungx

pkg/mcs/resourcemanager/redirector/redirector.go
```diff
@@ -20,27 +20,66 @@ import (
...
-					constant.ResourceManagerServiceName,
-					[]string{http.MethodGet, http.MethodPost, http.MethodPut, http.MethodDelete}),
-			),
-		), apiutil.APIServiceGroup{
-			Name:       "resource-manager",
-			Version:    "v1",
-			IsCore:     false,
-			PathPrefix: apis.APIPathPrefix,
-		}, nil
+	manager := rm_server.NewManager[*rmMetadataManagerProvider](&rmMetadataManagerProvider{Server: svr})
```
rleungx: It will add a callback in the leader election process, even if we use the standalone mode.

okJiang: Fixed in d1ce1a7d3. `redirector.NewHandler` no longer eagerly creates an RM manager during PD startup. The local RM metadata handler is now initialized lazily behind the fallback gate, and it uses `NewMetadataOnlyManager`, which loads metadata without registering start/service-ready callbacks. Verified with `make failpoint-enable && go test ./pkg/mcs/resourcemanager/server ./pkg/mcs/resourcemanager/redirector -count=1; make failpoint-disable` and `make static`.

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
...
+	for k, v := range conf {
+		key := reflectutil.FindJSONFullTagByChildTag(reflect.TypeOf(rmserver.ControllerConfig{}), k)
+		if key == "" {
+			c.String(http.StatusBadRequest, fmt.Sprintf("config item %s not found", k))
+			return
+		}
+		resolvedConf[key] = v
+	}
+	for key, v := range resolvedConf {
+		if err := s.store.UpdateControllerConfigItem(key, v); err != nil {
```
rleungx: This still is not fully all-or-nothing from the API perspective. We now validate all keys before applying, but value/type validation still happens inside `UpdateControllerConfigItem` one field at a time. If a request mixes one valid item with another item that fails there (for example an invalid duration value), an earlier field may already have been persisted before the later one returns `400`. That means callers can still get a partial config update even though `TestConfigServiceControllerAllOrNothing` suggests atomic behavior.

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -196,16 +185,7 @@ func (s *Service) postResourceGroup(c *gin.Context) {
...
-	if err := c.ShouldBindJSON(&group); err != nil {
-		c.String(http.StatusBadRequest, err.Error())
-		return
-	}
-	if err := s.manager.ModifyResourceGroup(&group); err != nil {
-		c.String(http.StatusInternalServerError, err.Error())
-		return
-	}
-	c.String(http.StatusOK, "Success!")
+	s.configService.PutResourceGroup(c)
```
rleungx, line 189: This PR changes `PUT /config/group` to return `404` for a missing resource group, but the swagger annotation above still only documents `400` and `500`. If we regenerate the spec from these comments, the published API contract will stay stale for the new behavior.

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -361,31 +266,7 @@ type KeyspaceServiceLimitRequest struct {
...
-	if err := s.manager.SetKeyspaceServiceLimit(keyspaceID, req.ServiceLimit); err != nil {
-		if rmserver.IsMetadataWriteDisabledError(err) {
-			c.String(http.StatusForbidden, err.Error())
-			return
-		}
-		c.String(http.StatusInternalServerError, err.Error())
-		return
-	}
-	c.String(http.StatusOK, "Success!")
+	s.configService.SetKeyspaceServiceLimit(c)
```
rleungx, line 271: Same documentation drift here: `ConfigService.SetKeyspaceServiceLimit` now maps keyspace-not-found to `404`, but this route annotation still only advertises `400`/`403`/`500`. The PR description calls out this status-code change explicitly, so the swagger comment should be updated as well.

pkg/mcs/resourcemanager/metadataapi/config_service.go
okJiang: This doesn’t seem to be an issue introduced by this PR. Do you think I should fix it in this PR or create a new issue to fix it in another PR?

okJiang: https://github.com/tikv/pd/issues/10335

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
...
+	if !ok {
+		c.String(http.StatusNotFound,
+			fmt.Sprintf("keyspace manager not found with keyspace name: %s, id: %d", keyspaceName, keyspaceID))
+		return
+	}
+	c.IndentedJSON(http.StatusOK, limiter)
+}
+
+func (*ConfigService) respondKeyspaceLookupError(c *gin.Context, err error) {
+	if errs.ErrKeyspaceNotExists.Equal(err) || strings.Contains(err.Error(), "keyspace not found with name:") {
```
rleungx: How about making `keyspace not found with name` a constant and adding a comment.

okJiang: Fixed in 4f18ba611. I moved the name-based lookup failure to a normalized RM error (`ErrKeyspaceNotExistsByName`) in `pkg/errs/errno.go`, switched `Manager.GetKeyspaceIDByName` to return that error, and removed the string-contains check from `respondKeyspaceLookupError`. `errors.toml` was regenerated as well. Verified with `make failpoint-enable && go test ./pkg/mcs/resourcemanager/server ./pkg/mcs/resourcemanager/metadataapi -count=1; make failpoint-disable`.

pkg/mcs/resourcemanager/redirector/metadata_handler.go
rleungx, line 1: The file name should also be changed.

pkg/mcs/resourcemanager/metadataapi/config_service.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/tikv/pd/pkg/utils/reflectutil"
+)
+
+// KeyspaceServiceLimitRequest is the request body for setting the service limit of a keyspace.
+type KeyspaceServiceLimitRequest struct {
+	ServiceLimit float64 `json:"service_limit"`
+}
+
+// Store abstracts metadata operations for config APIs.
+type Store interface {
```
rleungx: Is there a better name than store?

pkg/mcs/resourcemanager/redirector/metadata_handler.go
okJiang: Fixed in cb775ac32. The redirector file pair has been renamed to `metadata_handler.go` and `metadata_handler_test.go`, so the old `pd_metadata_handler*` naming is fully removed now.

pkg/mcs/resourcemanager/metadataapi/config_service.go
okJiang: Fixed in cb775ac32. I renamed the generic `Store` interface in `metadataapi` to `ConfigStore`, and updated the corresponding field/parameter names (`configStore`) to make the scope clearer while keeping `ManagerStore` unchanged.

rleungx: @coderabbitai Is there a better name?

niubell (approved):

PR #10248, opened 2026-02-13
tests: fix deadlock by introducing a starting state

### LINKED ISSUE #10241: POTENTIAL DEADLOCK in test
### Which jobs are failing

```
POTENTIAL DEADLOCK:
Previous place where the lock was grabbed
goroutine 1153 lock 0xc004e4d410
../cluster.go:144 tests.(*TestServer).Run { s.Lock() } <<<<<
../cluster.go:143 tests.(*TestServer).Run { func (s *TestServer) Run() error { }
../cluster.go:651 tests.RunServer.func1 { resC := make(chan error) }

Have been trying to lock it again for more than 30s
goroutine 281 lock 0xc004e4d410
../cluster.go:221 tests.(*TestServer).State { s.RLock() } <<<<<
../cluster.go:220 tests.(*TestServer).State { func (s *TestServer) State() int32 { }
../cluster.go:701 tests.(*TestCluster).runInitialServersWithRetry { for _, s := range servers { }
server_test.go:86 server_test.TestUpdateAdvertiseUrls { err = cluster.RunInitialServers() }
../cluster.go:670 tests.(*TestCluster).RunInitialServers { // RunInitialServers starts to run servers in InitialServers. }

Here is what goroutine 1153 doing now
goroutine 1153 [select]:
github.com/tikv/pd/server.(*Server).startEtcd(0xc004718f08, {0x7272848, 0xc00054dae0})
	/home/runner/work/pd/pd/server/server.go:356 +0x38f
github.com/tikv/pd/server.(*Server).Run(0xc004718f08)
	/home/runner/work/pd/pd/server/server.go:627 +0x1af
github.com/tikv/pd/tests.(*TestServer).Run(0xc004e4d410)
	/home/runner/work/pd/pd/tests/cluster.go:149 +0x1c7
github.com/tikv/pd/tests.RunServer.func1()
	/home/runner/work/pd/pd/tests/cluster.go:652 +0x45
created by github.com/tikv/pd/tests.RunServer in goroutine 281
	/home/runner/work/pd/pd/tests/cluster.go:652 +0x13a

```

### CI link

https://github.com/tikv/pd/actions/runs/21736731318/job/62703210572

awhvish:

Issue Number: Close #10241 

1. In `tests/cluster.go`, `TestServer.Run()` was holding the write lock (`s.Lock()`) for the entire duration of  `s.server.Run()` call, which can be slow (e.g., waiting for etcd).

2. If another thread tried to inspect the server state (e.g., calling `s.State()` -> s.RLock() in `runInitialServersWithRetry()` during error cleanup), it would block indefinitely waiting for the write lock to be released, causing a deadlock.

3.  Added a Starting state that releases the lock before calling `s.server.Run()`. This allows other operations to read the server state without blocking, preventing the deadlock.

4. Also changed a redundant `if err := os.RemoveAll(s.server.GetConfig().DataDir); err != nil` check caught by linter check in `Destroy()`

cc: @okJiang 

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Added an intermediate "starting" state and serialized startup to avoid races during test server startup.
  * Stop now cleanly supports stopping while starting; Destroy waits for any in-flight startup to finish before removing data, preventing data-race and cleanup issues.
  * Improved error handling for servers that stop during startup or shutdown, reducing flaky test failures.

### DIFF
--- a/tests/cluster.go
+++ b/tests/cluster.go
@@ -58,6 +58,7 @@ import (
 // TestServer states.
 const (
 	Initial int32 = iota
+	Starting
 	Running
 	Stop
 	Destroy
@@ -142,12 +143,32 @@ func NewTestServer(ctx context.Context, cfg *config.Config, services []string, h
 // Run starts to run a TestServer.
 func (s *TestServer) Run() error {
 	s.Lock()
-	defer s.Unlock()
 	if s.state != Initial && s.state != Stop {
+		s.Unlock()
 		return errors.Errorf("server(state%d) cannot run", s.state)
 	}
-	if err := s.server.Run(); err != nil {
+	// Immediately set state to Starting and unlock
+	s.state = Starting
+	s.Unlock()
+
+	// Run the server which is a slow operation
+	err := s.server.Run()
+
+	s.Lock()
+	defer s.Unlock()
+
+	if err != nil {
+		// If the server is destroyed or stopped while Run(), return nil
+		if s.state == Destroy || s.state == Stop {
+			return nil
+		}
+		s.state = Stop
 		return err
+	}
+	// Close and return nil if the server is destroyed or stopped
+	if s.state == Destroy || s.state == Stop {
+		s.server.Close()
+		return nil
 	}
 	s.state = Running
 	return nil
@@ -157,12 +178,16 @@ func (s *TestServer) Run() error {
 func (s *TestServer) Stop() error {
 	s.Lock()
 	defer s.Unlock()
-	if s.state != Running {
-		return errors.Errorf("server(state%d) cannot stop", s.state)
+	if s.state == Starting {
+		s.state = Stop
+		return nil
 	}
-	s.server.Close()
-	s.state = Stop
-	return nil
+	if s.state == Running {
+		s.server.Close()
+		s.state = Stop
+		return nil
+	}
+	return errors.Errorf("server(state%d) cannot stop", s.state)
 }
 
 // Destroy is used to destroy a TestServer.
@@ -172,11 +197,9 @@ func (s *TestServer) Destroy() error {
 	if s.state == Running {
 		s.server.Close()
 	}
-	if err := os.RemoveAll(s.server.GetConfig().DataDir); err != nil {
-		return err
-	}
+	// Set state to Destroy before RemoveAll
 	s.state = Destroy
-	return nil
+	return os.RemoveAll(s.server.GetConfig().DataDir) // removed redundant if err != nil check, linter fix
 }
 
 // ResetPDLeader resigns the leader of the server.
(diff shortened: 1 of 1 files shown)


tests/cluster.go
```diff
@@ -142,13 +143,33 @@ func NewTestServer(ctx context.Context, cfg *config.Config, services []string, h
...
+		s.Unlock()
 		return errors.Errorf("server(state%d) cannot run", s.state)
 	}
-	if err := s.server.Run(); err != nil {
+	// Immediately set state to Starting and unlock
+	s.state = Starting
+	s.Unlock()
+
+	// Run the server which is a slow operation
+	err := s.server.Run()
```
okJiang, line 160: In fact, adding a lock to `server.Run()` is appropriate because server's field is accessed during testing. The issue lies in the fact that `server.Run()` may take longer than 30 seconds, triggering a `potential deadlock` error.

After reviewing some test code, I found that, fundamentally, there doesn't seem to be any concurrent relationship between accessing server's field and `server.Run()`. Therefore, I believe this PR can be merged to prevent `potential deadlocks` error from occurring again. If data races emerge after merging, the PR should be reverted at that time.

okJiang (approved):

rleungx:
I think this still has a state-machine bug after introducing `Starting`.

`TestServer.Run()` now returns `nil` when startup was interrupted by `Stop()`/`Destroy()` while `s.server.Run()` was still in flight. That changes the meaning of a successful return: callers such as `RunServers()` / `RunInitialServers()` will treat a canceled startup as a successful start, even though the server never reached `Running`.

There is also a reentrancy window now. The first `Run()` sets `state = Starting` and unlocks before calling `s.server.Run()`. If another goroutine calls `Stop()`, it changes the state to `Stop` and returns immediately, but the first `Run()` is still executing. A second `Run()` can then pass the `state == Stop` check and enter `s.server.Run()` concurrently on the same `TestServer`. The old implementation did not allow that because the write lock was held for the whole `Run()`.

So while this avoids the original deadlock, it seems to replace it with two new problems:
1. interrupted startup can be reported as success;
2. overlapping `Run()` calls become possible on the same underlying server.

I think `Starting` needs an additional in-flight startup guard, and `Run()` should not return success unless the server actually transitions to `Running`.

tests/cluster.go
```diff
@@ -58,6 +58,7 @@ import (
 // TestServer states.
 const (
 	Initial int32 = iota
+	Starting
```
rleungx, line 61: This new `Starting` state is not handled by the retry cleanup paths, so a retry can leave an in-flight startup behind.

PR #10250, opened 2026-02-16
[tso]: Fix possible overflow in merging tso requests

### LINKED ISSUE #10249: There is a possible overflow in TSODispatcher
### What did you do?
 
The request count in pd/pkg/utils/tsoutil/tso_dispatcher.go (processRequests) is accumulated in a uint32, which could lead to overflow in extreme cases or be triggered by a malicious client.

```go
func (s *TSODispatcher) processRequests(forwardStream stream, requests []Request) error {                    
        // Merge the requests                                                                                
        count := uint32(0)
        for _, request := range requests {
                count += request.getCount()                                                                  
        }                                                                                                    
        
        start := time.Now()
        resp, err := requests[0].process(forwardStream, count)                                               
        if err != nil {
                return err                                                                                   
        }
        s.tsoProxyHandleDuration.Observe(time.Since(start).Seconds())                                        
        if s.tsoProxyBatchSize != nil {
                s.tsoProxyBatchSize.Observe(float64(count))                                                  
        }
        // Split the response
        ts := resp.GetTimestamp()
        physical, logical := ts.GetPhysical(), ts.GetLogical()
        // `logical` is the largest ts's logical part here, we need to do the subtracting before we finish each TSO request.  
        // This is different from the logic of client batch, for example, if we have a largest ts whose logical part is 10,   
        // count is 5, then the splitting results should be 5 and 10.                                        
        firstLogical := logical - int64(count)
        return s.finishRequest(requests, physical, firstLogical)                                             
} 
```

  - Merged count should not silently overflow.
  - If total count exceeds uint32, dispatcher should return an error (or use wider type with explicit bound check).
  - Proxy split logic should always match the exact number of timestamps requested upstream.

### What did you see instead?
  - count overflows silently due to uint32 wraparound.
  - Upstream process(...) is called with a wrapped (smaller) count.
  - Then finishRequest(...) still splits using original per-request counts, causing allocation/splitting mismatch.
  - tsoProxyBatchSize metric is also incorrect when overflow happens.

master branch

sunxiaoguang:

Issue Number: Close #10249

  - Fixed overflow in TSODispatcher.processRequests by switching merge accumulation to checked uint64 and rejecting totals above math.MaxUint32 before forwarding
  - Added explicit error on overflow: "merged tso request count overflow"
  - Added regression test TestProcessRequestsCountOverflow that reproduces math.MaxUint32 + 1 merged input and asserts no forward-stream call occurs

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents numeric overflow when combining request counts for timestamp generation, returning an error instead of proceeding with invalid totals to ensure stability.

* **Tests**
  * Adds a test that verifies an overflowed combined request count triggers the error and does not mark any requests as successful.

### DIFF
--- a/pkg/utils/tsoutil/tso_dispatcher.go
+++ b/pkg/utils/tsoutil/tso_dispatcher.go
@@ -188,7 +188,7 @@ func (s *TSODispatcher) processRequests(forwardStream stream, requests []Request
 	count := uint32(totalCount)
 
 	start := time.Now()
-	resp, err := requests[0].process(forwardStream, count)
+	resp, err := requests[0].process(forwardStream, count) // #nosec G602 -- requests is guaranteed non-empty by caller loop guard
 	if err != nil {
 		return err
 	}


sunxiaoguang:
> @sunxiaoguang: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-unit-test-next-gen-3	[c3ad631](https://github.com/tikv/pd/commit/c3ad63139304f59117d2bf5e157ce669a2ca68cf)	[link](https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10250/pull-unit-test-next-gen-3/2023395700674400256)	true	`/test pull-unit-test-next-gen-3`
> [Full PR test history](https://prow.tidb.net/pr-history?org=tikv&repo=pd&pr=10250). [Your PR dashboard](https://prow.tidb.net/pr?query=is%3Apr%20state%3Aopen%20author%3Asunxiaoguang).
> 
> Details
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes-sigs/prow](https://github.com/kubernetes-sigs/prow/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

The test is irrelevant to the changed code. Is it a flaky test?

rleungx (approved):

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -177,13 +178,17 @@ func (s *TSODispatcher) dispatch(
...
 func (s *TSODispatcher) processRequests(forwardStream stream, requests []Request) error {
 	// Merge the requests
-	count := uint32(0)
+	totalCount := uint64(0)
 	for _, request := range requests {
-		count += request.getCount()
+		totalCount += uint64(request.getCount())
 	}
+	if totalCount > math.MaxUint32 {
+		return errs.ErrGenerateTimestamp.FastGenByArgs("merged tso request count overflow")
```
okJiang, line 186: Why not directly change the `count` in kvproto to uint64?

okJiang: cc @rleungx @JmPotato

JmPotato: This means we need to introduce a new uint64 field to replace the original one, which may require more work but seems more reasonable.

okJiang: Changing the field from uint32 to uint64 should not introduce compatibility issues, and there is no need to add a new field?

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -177,13 +178,17 @@ func (s *TSODispatcher) dispatch(
 
 func (s *TSODispatcher) processRequests(forwardStream stream, requests []Request) error {
 	// Merge the requests
-	count := uint32(0)
+	totalCount := uint64(0)
 	for _, request := range requests {
-		count += request.getCount()
+		totalCount += uint64(request.getCount())
 	}
+	if totalCount > math.MaxUint32 {
```
JmPotato, line 185: Minor: The error message is a fixed string. It would be helpful for debugging to include the actual values:
```go
return errs.ErrGenerateTimestamp.FastGenByArgs(
    fmt.Sprintf("merged tso request count overflow: total=%d, requests=%d", totalCount, len(requests)))
```

pkg/utils/tsoutil/tso_dispatcher.go
```diff
@@ -177,13 +178,17 @@ func (s *TSODispatcher) dispatch(
...
+		return errs.ErrGenerateTimestamp.FastGenByArgs("merged tso request count overflow")
+	}
+	count := uint32(totalCount)
```
JmPotato, line 189: Nit: The `#nosec G602` comment is unrelated to this PR's overflow fix (the line's logic didn't change). Consider removing it from this PR or splitting it into a separate commit to keep the diff focused.

pkg/utils/tsoutil/tso_dispatcher_test.go
```diff
@@ -192,3 +193,27 @@ func (suite *tsoDispatcherTestSuite) TestGoroutineLeakOnStreamError() {
...
+	}
+
+	err := suite.dispatcher.processRequests(mockStream, requests)
+	re.Error(err)
+	re.Contains(err.Error(), "merged tso request count overflow")
+
+	mockStream.Lock()
+	defer mockStream.Unlock()
+	re.Equal(0, mockStream.succCount)
+}
```
JmPotato, line 219: Minor suggestions for the test:

1. Consider adding a boundary test for `totalCount == math.MaxUint32` (exactly at the limit) to verify it still succeeds — this would complement the overflow case.

2. Could also verify that the batched requests' `doneCh` channels are **not** closed on overflow, to confirm `postProcess` was never called:
```go
select {
case <-requests[0].(*mockRequest).doneCh:
    re.Fail("doneCh should not be closed on overflow")
default:
}
```

JmPotato (commented):
Overall the fix is correct and well-scoped. The approach — accumulate in `uint64`, check the bound, return a clear error — is sound. The regression test covers the critical overflow path effectively.

A few minor suggestions in the inline comments (none blocking).

PR #10257, opened 2026-02-27
fix(*): resolve gosec G602/G702-G706 and perfsprint issues

JmPotato:

Issue Number: ref #4322.

```commit-message
- Fix G602 array index out of bounds in MakeKeyRanges with bounds check
- Fix G702-G706 by using fmt.Fprintf instead of WriteString+Sprintf
- Upgrade golangci-lint v2.6.0 to v2.10.1
- Add corresponding gosec rules to .golangci.yml exclusions
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential out-of-bounds access when processing key ranges with an odd number of keys.

* **Chores**
  * Updated linting tool to v2.10.1 and configured additional security check exclusions.

* **Refactor**
  * Optimized memory allocation in string formatting operations.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -80,6 +80,12 @@ linters:
         - G402
         - G404
         - G115
+        - G602
+        - G702
+        - G703
+        - G704
+        - G705
+        - G706
     perfsprint:
       string-format: false
     revive:
--- a/pkg/schedule/labeler/labeler.go
+++ b/pkg/schedule/labeler/labeler.go
@@ -431,7 +431,7 @@ func (l *RegionLabeler) GetRegionLabels(region *core.RegionInfo) []*RegionLabel 
 // MakeKeyRanges is a helper function to make key ranges.
 func MakeKeyRanges(keys ...string) []any {
 	var res []any
-	for i := 0; i < len(keys); i += 2 {
+	for i := 0; i+1 < len(keys); i += 2 {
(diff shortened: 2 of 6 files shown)


pkg/schedule/labeler/labeler.go
```diff
@@ -431,7 +431,7 @@ func (l *RegionLabeler) GetRegionLabels(region *core.RegionInfo) []*RegionLabel
 // MakeKeyRanges is a helper function to make key ranges.
 func MakeKeyRanges(keys ...string) []any {
 	var res []any
-	for i := 0; i < len(keys); i += 2 {
+	for i := 0; i+1 < len(keys); i += 2 {
```
okJiang, line 434: Why🤔

JmPotato: In the loop, we’re always accessing i and i+1, so the constraint here should be aimed at the latter to be safest.

okJiang (approved):

.golangci.yml
```diff
@@ -80,6 +80,12 @@ linters:
         - G402
         - G404
         - G115
+        - G602
```
rleungx, line 83: How about fixing them instead of excluding them?

PR #10259, opened 2026-02-27
resourcemanager: add metadata watcher scaffold

okJiang:

The write-in-PD plan needs ResourceManager to consume resource-group metadata from etcd instead of owning metadata writes directly. This PR adds the metadata watcher bootstrap path so RM can stay in sync with PD-driven updates for resource-group settings, runtime states, controller config, and keyspace service limits.

Issue Number: ref #9737

This PR adds the ResourceManager metadata watcher scaffold in `pkg/mcs/resourcemanager/server` and wires it into `Manager.Init()` behind `EnableResourceGroupMetadataWatcher()`. The watcher now follows `resource_group/` changes for controller config, resource-group settings, runtime states, and keyspace service limits, while the manager keeps the legacy full-load path as the fallback bootstrap mode.

The rest of the branch hardens that watcher path so it is safe to enable: it fixes watcher lifecycle cleanup, publishes controller-config snapshots safely, keeps key validation and delete behavior consistent across watcher and legacy bootstrap paths, and makes service-limit replay and default-group restoration behave correctly in memory.

Validation:
- `make gotest GOTEST_ARGS='./pkg/mcs/resourcemanager/server -count=1'`
- `./.tools/bin/golangci-lint run --verbose ./pkg/mcs/resourcemanager/server/... --allow-parallel-runners`
- `make static`

```commit-message
Add the ResourceManager metadata watcher bootstrap path and watch
resource-group settings, states, controller config, and keyspace
service limits from etcd. Create the manager-owned cancelable context
before watcher startup, publish controller-config snapshots safely,
validate etcd key and payload-name consistency, and keep delete and
default-restore paths consistent with the in-memory cache and
service-limit state.
```

### Check List

Tests

- Unit test

Side effects

- Increased code complexity

### Release note

```release-note
Add ResourceManager metadata watcher bootstrap so RM can follow
resource-group settings, states, controller config, and keyspace
service limits from etcd with safer watcher lifecycle and cache
consistency handling.
```

## Summary by CodeRabbit

* **New Features**
  * Metadata watcher now synchronizes resource group configurations and service limits from centralized storage
  * Default resource groups can be restored and reinitialized

* **Bug Fixes**
  * Resource group deletion now uses transactional operations for improved consistency

* **Tests**
  * Added comprehensive test coverage for metadata watcher and resource group operations

### DIFF
--- a/errors.toml
+++ b/errors.toml
@@ -903,7 +903,7 @@ the %s resource group does not exist
 
 ["PD:resourcemanager:ErrInvalidGroup"]
 error = '''
-invalid group settings, please check the group name, priority and the number of resources
+invalid group settings, please check %s
 '''
 
 ["PD:resourcemanager:ErrKeyspaceNotExists"]


pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,203 @@
...
+			groupName:  name,
+		}, true
+	}
+	if strings.HasPrefix(trimmed, keyspaceResourceGroupSettingsPrefix) {
+		return parseKeyspaceWatchPath(strings.TrimPrefix(trimmed, keyspaceResourceGroupSettingsPrefix), resourceGroupWatchEntrySettings)
+	}
+	if strings.HasPrefix(trimmed, keyspaceResourceGroupStatesPrefix) {
+		return parseKeyspaceWatchPath(strings.TrimPrefix(trimmed, keyspaceResourceGroupStatesPrefix), resourceGroupWatchEntryStates)
+	}
+	return resourceGroupWatchTarget{}, false
```
okJiang: This parser currently drops every `resource_group/*` key except settings/states. In watcher mode that means `resource_group/controller` and `resource_group/keyspace/service_limits/*` are only loaded once during `Init()` and then never refreshed, so RM will serve stale config/limits after PD-side updates. Can we either watch those keys here as well or explicitly scope this PR to settings/states only?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,203 @@
...
+		m.srv.GetClient(),
+		"resource-manager-metadata-watcher",
+		resourceGroupWatchPrefix,
+		func([]*clientv3.Event) error { return nil },
+		putFn,
+		deleteFn,
+		func([]*clientv3.Event) error { return nil },
+		true, /* withPrefix */
+	)
+	watcher.StartWatchLoop()
```
okJiang, line 178: `LoopWatcher` is started on the parent callback `ctx`, but `Manager.close()` only cancels `m.cancel`, which is created later in `Init()`. Once this flag is enabled, `close()` and init-failure cleanup can't stop this goroutine cleanly, so watcher mode will leak/hang on shutdown. Please bind the watcher to a manager-owned cancelable context (or store a separate watcher cancel) before starting it.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -88,6 +89,11 @@ type writeRoleProvider interface {
 	GetResourceGroupWriteRole() ResourceGroupWriteRole
 }
 
+// metadataWatcherProvider is an optional provider used to toggle RM metadata watcher.
+type metadataWatcherProvider interface {
+	EnableResourceGroupMetadataWatcher() bool
+}
```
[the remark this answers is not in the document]
okJiang: Checked against the current branch: `pkg/mcs/resourcemanager/server/server.go` already implements `EnableResourceGroupMetadataWatcher()`, so the concrete `*Server` satisfies `metadataWatcherProvider` and `NewManager` can enable watcher mode outside unit tests.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Fixed in c76697114. `Manager.Init()` now creates the manager-owned cancelable context before watcher startup, passes that context into `initializeMetadataWatcher()`, cancels and waits on init failure, and the manager tests cover both init-failure cleanup and `Manager.close()` cancellation.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Fixed in c76697114. `parseResourceGroupWatchPath()` now recognizes `resource_group/controller` and `resource_group/keyspace/service_limits/*`, `handleMetadataWatchPut()` applies both paths, service-limit deletes reset the in-memory limiter, and the watcher tests cover the new controller/service-limit cases.

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+		key,
+		preEventsFn,
+		putFn,
+		deleteFn,
+		postEventsFn,
+		isWithPrefix,
+	)
+}
+
+func parseResourceGroupWatchPath(path string) (resourceGroupWatchTarget, bool) {
```
okJiang, line 84: Are there any existing functions that can reuse some of the logic within this function?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+type resourceGroupWatchEntryType uint8
+
+const (
+	resourceGroupWatchEntryUnknown resourceGroupWatchEntryType = iota
+	resourceGroupWatchEntryController
+	resourceGroupWatchEntrySettings
+	resourceGroupWatchEntryStates
+	resourceGroupWatchEntryServiceLimit
```
okJiang: Can these types be annotated with the actual path from the keypath pkg?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	krgm := m.getKeyspaceResourceGroupManager(target.keyspaceID)
+	if krgm == nil {
+		return nil
+	}
+	krgm.deleteResourceGroupFromCache(target.groupName)
+	return nil
+}
+
+func (m *Manager) applyControllerConfigFromRaw(rawValue string) error {
```
okJiang: Is there another more suitable place to put this function?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+			zap.Error(err))
+		return err
+	}
+	m.Lock()
+	m.controllerConfig = controllerConfig
+	m.Unlock()
+	return nil
+}
+
+func (m *Manager) applyResourceGroupSettingFromRaw(keyspaceID uint32, name, rawValue string) error {
```
okJiang: ditto

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+			zap.Uint32("keyspace-id", keyspaceID),
+			zap.String("group-name", name),
+			zap.String("raw-value", rawValue),
+			zap.Error(err))
+		return err
+	}
+	return nil
+}
+
+func (m *Manager) applyServiceLimitFromRaw(keyspaceID uint32, rawValue string) error {
```
okJiang: ditto

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,300 @@
...
+			zap.Uint32("keyspace-id", keyspaceID),
+			zap.String("raw-value", rawValue),
+			zap.Error(err))
+		return err
+	}
+	m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, false).setServiceLimitFromStorage(serviceLimit)
+	return nil
+}
+
+func (m *Manager) applyResourceGroupStatesFromRaw(keyspaceID uint32, name, rawValue string) error {
```
okJiang: ditto

pkg/mcs/resourcemanager/server/manager_test.go
```diff
@@ -67,13 +72,129 @@ func (*mockRoleConfigProvider) AddStartCallback(...func()) {}
...
+}
+
 func prepareManager() *Manager {
 	storage := storage.NewStorageWithMemoryBackend()
 	m := NewManager[*mockConfigProvider](&mockConfigProvider{})
 	m.storage = storage
 	return m
 }
 
+func TestInitMetadataWatcherUsesManagerOwnedContextAndCancelsOnError(t *testing.T) {
```
okJiang: This name is too long

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+		func([]*clientv3.Event) error { return nil },
+		true, /* withPrefix */
+	)
+	watcher.StartWatchLoop()
+	if err := watcher.WaitLoad(); err != nil {
+		return err
+	}
+	// Ensure reserved default groups exist even if settings were missing in storage.
+	m.initReserved()
+	return m.loadServiceLimits()
```
okJiang: Bootstrap still replays `service_limits/*` before `settings/*` because the initial `resource_group/` load is lexicographically ordered. Re-running `loadServiceLimits()` here does not fix that, because `serviceLimiter.setServiceLimitInternal()` returns immediately when the numeric limit is unchanged. A keyspace whose service limit is replayed before its groups will therefore keep `overrideBurstLimit == -1` until the limit changes again. Can we either defer applying service limits until after settings replay, or add a forced re-evaluation path that recomputes overrides even when the stored limit is the same?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -128,22 +172,30 @@ func (krgm *keyspaceResourceGroupManager) initDefaultResourceGroup() {
...
 	}
-	if err := krgm.addResourceGroup(defaultGroup.IntoProtoResourceGroup(krgm.keyspaceID)); err != nil {
-		log.Warn("init default group failed", zap.Uint32("keyspace-id", krgm.keyspaceID), zap.Error(err))
-	}
+}
+
+func (krgm *keyspaceResourceGroupManager) restoreDefaultResourceGroupFromReserved() {
+	krgm.Lock()
+	krgm.groups[DefaultResourceGroupName] = newDefaultResourceGroup()
+	krgm.Unlock()
```
okJiang, line 248: `restoreDefaultResourceGroupFromReserved()` now writes `newDefaultResourceGroup()` directly into the cache. That bypasses the normal `addResourceGroup()` / `FromProtoResourceGroup()` construction path, so the restored default group is missing runtime fields such as `RUConsumption`. After a watcher delete on the default group, the next state persist or RU-consumption update can therefore panic on the restored object. Can we restore it through the same normalized path we use for regular group loads, or at least initialize the runtime fields here before putting it back into `krgm.groups`?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -204,12 +256,13 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
...
+	if err := krgm.storage.DeleteResourceGroupStates(krgm.keyspaceID, name); err != nil {
+		return err
+	}
 	if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
 		return err
 	}
-	krgm.Lock()
-	delete(krgm.groups, name)
-	krgm.Unlock()
+	krgm.deleteResourceGroupFromCache(name)
```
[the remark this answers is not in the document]
okJiang: Fixed. `deleteResourceGroup()` now removes settings/states in a single `RunInTxn` when the storage supports transactions, falls back to deleting settings first otherwise, and only drops the cache after storage deletion completes. Added `TestDeleteResourceGroupUsesTransactionalDelete`.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -279,11 +302,73 @@ func (m *Manager) loadKeyspaceResourceGroups() error {
...
+	controllerConfig := &ControllerConfig{}
+	if err := json.Unmarshal([]byte(rawValue), controllerConfig); err != nil {
+		log.Error("failed to apply controller config from watcher",
+			zap.String("raw-value", rawValue),
+			zap.Error(err))
+		return err
+	}
+	m.Lock()
+	m.controllerConfig = controllerConfig
+	m.Unlock()
```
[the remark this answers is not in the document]
okJiang: Fixed. Controller config is now published/read through snapshots: `GetControllerConfig()` returns a clone, `UpdateControllerConfigItem()` mutates and publishes a copied config under the lock, and `backgroundMetricsFlush()` reads through the same snapshot path. Added `TestGetControllerConfigReturnsSnapshot` and `TestUpdateControllerConfigItemPublishesNewSnapshot`.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -212,21 +225,31 @@ func (m *Manager) Init(ctx context.Context) error {
...
+	if m.enableMetadataWatcher {
+		if err := m.initializeMetadataWatcher(ctx); err != nil {
+			m.cancel()
+			m.wg.Wait()
+			return err
+		}
+	} else {
+		if err := m.loadKeyspaceResourceGroups(); err != nil {
+			m.cancel()
+			return err
```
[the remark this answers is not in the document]
okJiang: Fixed. The legacy bootstrap path now shares the same key/payload-name validation as watcher mode via `parseResourceGroupFromRaw()`, so mismatched records are rejected instead of creating ghost groups. Added `TestLoadKeyspaceResourceGroupsRejectsMismatchedPayloadName`.

okJiang:
Also addressed the CodeRabbit review-body notes from the March 6 review:
- `handleMetadataWatchDelete()` now emits debug logs when delete events are skipped because the keyspace manager is missing.
- `initializeMetadataWatcher()` no longer replays `loadServiceLimits()` after `WaitLoad()`.
- Added `TestInitializeMetadataWatcherDoesNotReloadServiceLimits`.

Verification:
- `make gotest GOTEST_ARGS="./pkg/mcs/resourcemanager/server -count=1"`
- `./.tools/bin/golangci-lint run --verbose ./pkg/mcs/resourcemanager/server/... --allow-parallel-runners`

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Fixed in e7a2d93a3. Newly loaded and restored groups now re-apply the existing keyspace service limit when they enter the cache, so lexicographic replay order no longer leaves burstable groups with `overrideBurstLimit == -1` until the limit changes again. Added `TestHandleMetadataWatchPutAppliesExistingServiceLimitToNewGroup`.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: Fixed in e7a2d93a3. Restoring the reserved default group now initializes the runtime fields expected by the normal in-memory path, including `RUConsumption`, and applies the same service-limit sync as regular group loads. Added `TestHandleMetadataWatchDeleteRestoresDefaultGroupRuntimeFields`.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -279,11 +302,81 @@ func (m *Manager) loadKeyspaceResourceGroups() error {
...
+		return err
+	}
+	m.Lock()
+	m.controllerConfig = controllerConfig
+	m.Unlock()
+	return nil
+}
+
+func (m *Manager) applyResourceGroupSettingFromRaw(keyspaceID uint32, name, rawValue string) error {
+	krgm := m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, false)
```
lhy1024, line 410: it should be true?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -279,11 +302,81 @@ func (m *Manager) loadKeyspaceResourceGroups() error {
...
+func (m *Manager) applyServiceLimitFromRaw(keyspaceID uint32, rawValue string) error {
+	var serviceLimit float64
+	if err := json.Unmarshal([]byte(rawValue), &serviceLimit); err != nil {
+		log.Error("failed to apply service limit from watcher",
+			zap.Uint32("keyspace-id", keyspaceID),
+			zap.String("raw-value", rawValue),
+			zap.Error(err))
+		return err
+	}
+	m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, false).setServiceLimitFromStorage(serviceLimit)
```
lhy1024: it should be true, too.

pkg/mcs/resourcemanager/server/manager.go
okJiang: Changed this callsite to `true`.

The second argument is `initDefault`, so the watcher settings path should keep the same invariant as the bootstrap/API paths: a keyspace manager created here should immediately have the reserved default group. I also added a regression subtest under the existing `TestMetadataWatcherHandlePut` coverage for a keyspace first seen from watcher settings updates.

pkg/mcs/resourcemanager/server/manager.go
okJiang: Changed this one to `true` as well.

`SetKeyspaceServiceLimit()` already initializes the keyspace manager with the reserved default group, so the watcher service-limit path should behave the same way for a keyspace first created from watcher replay/update. The same regression coverage in `TestMetadataWatcherHandlePut` now checks that case too.

pkg/mcs/resourcemanager/server/manager.go
lhy1024: I'm sorry, my previous advice was incomplete. This fixes the original "missing reserved `default` group in memory" issue, but it also makes the watcher path call `initDefaultResourceGroup()`, which is not cache-only. 

`initDefaultResourceGroup()` goes through `addResourceGroup(...)`, and under the current RM write role (`ResourceGroupWriteRoleLegacyAll`) that persists metadata to storage.

That becomes incorrect if the first watcher event for a new keyspace is itself `resource_group/keyspace/settings/<ks>/default`. In that case the code first persists the built-in reserved default template (priority=8, fill_rate=UnlimitedRate, etc.), and only afterwards applies the real payload into memory via `upsertResourceGroupFromRaw("default", rawValue)`. The in-memory object may look correct, but the stored payload can already be overwritten.

I think the watcher should remain a metadata consumer in this path: it should synchronize local in-memory state from etcd, but it should not write metadata back to etcd/storage while handling the watch event.

lhy1024: I locally reproduced this by adding a targeted test in the same style as the new watcher test suite:

```go
t.Run("does_not_overwrite_storage_payload_when_default_group_arrives_first", func(t *testing.T) {
    re := require.New(t)

    memStorage := storage.NewStorageWithMemoryBackend()
    m := newMetadataWatcherTestManager(memStorage)
    defaultGroup := newMetadataWatcherResourceGroup(DefaultResourceGroupName, 1, 1000, -1)
    rawDefaultGroup := mustMarshalResourceGroup(t, defaultGroup)

    re.NoError(m.handleMetadataWatchPut("resource_group/keyspace/settings/10/default", rawDefaultGroup))

    stored, err := memStorage.Load(keypath.KeyspaceResourceGroupSettingPath(10, DefaultResourceGroupName))
    re.NoError(err)
    re.NotEmpty(stored)

    persisted := &rmpb.ResourceGroup{}
    re.NoError(proto.Unmarshal([]byte(stored), persisted))
    re.Equal(defaultGroup.Priority, persisted.Priority)
    re.Equal(
        defaultGroup.GetRUSettings().GetRU().GetSettings().GetFillRate(),
        persisted.GetRUSettings().GetRU().GetSettings().GetFillRate(),
    )
})
```

lhy1024: we should split this behavior, 
```
getOrCreateKeyspaceResourceGroupManager(keyspaceID, false)
ensureReservedDefaultGroupInCache()
```
Its behavior should be:

- if `default` already exists in `groups`, do nothing
- otherwise insert `newDefaultResourceGroup()` into the in-memory map
- call `syncBurstabilityWithServiceLimit(defaultGroup)` if needed
- do not call `addResourceGroup(...)`
- do not persist metadata

pkg/mcs/resourcemanager/server/manager.go
okJiang: Covered the service-limit watcher path in the same fix at a5f0a5b07. `applyServiceLimitFromRaw()` now also uses `getOrCreateKeyspaceResourceGroupManager(..., false)` plus the same cache-only `ensureReservedDefaultGroupInCache()` helper, so replaying a service limit for a new keyspace no longer persists the reserved default group. Added `TestMetadataWatcherHandlePut/service_limit_update_does_not_persist_reserved_default_group` for that path.

pkg/mcs/resourcemanager/server/manager.go
okJiang: Fixed in a5f0a5b07. The watcher settings path no longer goes through `getOrCreateKeyspaceResourceGroupManager(..., true)` -> `initDefaultResourceGroup()` -> `addResourceGroup(...)` on a new keyspace. It now creates the manager with `initDefault=false`, then uses a cache-only `ensureReservedDefaultGroupInCache()` helper that only inserts the reserved `default` group into `groups` and syncs burstability without persisting metadata/state. Added `TestMetadataWatcherHandlePut/does_not_overwrite_storage_payload_when_default_group_arrives_first` to cover the case you described.

JmPotato (commented):
## Review: resourcemanager: add metadata watcher scaffold

Overall the design is sound — wiring bootstrap behind `EnableResourceGroupMetadataWatcher()` keeps the PD path untouched, `newMetadataLoopWatcher` as an injectable variable is the right testing pattern, and the controller-config snapshot fix (`cloneControllerConfig` + `GetControllerConfig`) is a clear improvement. A few things worth addressing before merging:

---

### Medium

**[Correctness] `handleMetadataWatchDelete` control flow is non-obvious and fragile**

```go
switch target.entryType {
case resourceGroupWatchEntryServiceLimit:
    // ...
    return nil
case resourceGroupWatchEntrySettings:
default:
    return nil
}
// Keep the reserved group alive.
```

In Go, an empty `case` body does nothing and exits the switch — it does **not** fall through. So the post-switch code (default-group restoration + `deleteResourceGroupFromCache`) is only reached for `resourceGroupWatchEntrySettings`. This is correct, but the style is highly unexpected. Anyone unfamiliar with Go's switch semantics will read `case resourceGroupWatchEntrySettings:` directly above `default: return nil` and think both hit the early return.

**Suggestion**: make the intent explicit:
```go
if target.entryType != resourceGroupWatchEntrySettings {
    if target.entryType == resourceGroupWatchEntryServiceLimit {
        // ... handle
    }
    return nil
}
```
Or at minimum add a comment: `// Settings case falls through to group-deletion logic below.`

---

**[Consistency] States-delete events are silently no-ops; worth a comment**

When a group's settings key is deleted from etcd, the watcher also fires a delete event for the corresponding states key. The states-delete handler hits `default: return nil` and does nothing. This is correct because `deleteResourceGroupFromCache` already cleaned up memory when the settings delete was processed, but a future reader will be confused by the intentional no-op. A one-liner comment (`// States cleanup is handled when the settings key is deleted.`) would make this clear.

---

**[Correctness] `applyResourceGroupStatesFromRaw` silently drops states when the manager hasn't been created yet**

```go
func (m *Manager) applyResourceGroupStatesFromRaw(keyspaceID uint32, name, rawValue string) error {
    krgm := m.getKeyspaceResourceGroupManager(keyspaceID)
    if krgm == nil {
        log.Debug("skip applying resource group states without corresponding manager", ...)
        return nil
    }
```

During the initial watcher load (`WaitLoad`), etcd events replay in key order. If a states key for keyspace X arrives before the corresponding settings key (possible if the prefix-watch ordering happens to deliver them that way), the states are silently dropped. The next periodic state sync will recover, but there's a window where token bucket state is stale after bootstrap.

**Suggestion**: either document this as an accepted trade-off in a comment, or call `getOrCreateKeyspaceResourceGroupManager` here (mirroring the settings/service-limit handlers) so the manager is always created and states can be applied whenever they arrive.

---

**[Lock contention] `UpdateControllerConfigItem` holds the Manager write lock during `SaveControllerConfig`**

```go
m.Lock()
// ... build config snapshot
if updated {
    if err := m.storage.SaveControllerConfig(controllerConfig); err != nil { // etcd write under lock
        m.Unlock()
        return err
    }
    m.controllerConfig = controllerConfig
}
m.Unlock()
```

The etcd write holds the write lock, blocking all reads on `m.RLock()` (including `GetControllerConfig`, `backgroundMetricsFlush`, etc.) for the duration of the IO. This was also a problem pre-PR, but since this PR already introduces `cloneControllerConfig` to make the publish safe, it's a natural place to fix it: persist without the lock, then re-acquire to publish.

---

### Low

**[Style] `resourceGroupWatchPrefix` is a hardcoded string instead of a `keypath` reference**

```go
const resourceGroupWatchPrefix = "resource_group/"
```

The `keypath` package owns all etcd path constants. If the path prefix ever changes, this hardcoded constant will silently diverge. Consider deriving it from an existing `keypath` function or adding `keypath.ResourceGroupWatchPrefix()`.

---

### Highlights

- The lock refactor in `addResourceGroup` (persist outside lock, write inside) is the right direction.
- `newDefaultResourceGroup` helper unifies default-group construction — no more inline struct literals scattered across `initDefaultResourceGroup` and `ensureReservedDefaultGroupInCache`.
- Transactional delete of settings + states in `deleteResourceGroup` is a solid fix; `transactionalDeleteGuardStorage` in tests is a clever way to enforce the contract.
- Test coverage is thorough: lifecycle, error paths, ordering edge cases, and the non-reload assertion in `TestInitializeMetadataWatcher` are all valuable.

okJiang:
Addressed [this review](https://github.com/tikv/pd/pull/10259#pullrequestreview-3942343956) in `a5e47c1b8`, with a follow-up simplification in `93d4d6e8e`.

Changes made:
- Reworked `handleMetadataWatchDelete` into explicit entry-type branches, so the settings-delete path no longer depends on the empty-case switch pattern.
- Added an explicit comment for the states-delete no-op (`settings` deletion already handles the cache cleanup).
- Kept `applyResourceGroupStatesFromRaw` as a no-op when the manager does not exist, but documented the trade-off: `LoopWatcher` bootstrap is lexicographically ordered so settings load before states, and simply creating the manager on a state-only event would still not materialize the target group. For live watch reordering, the next persisted states update heals the gap.
- Replaced the hardcoded watcher prefix with `keypath.ResourceGroupPrefix()`.
- Kept `UpdateControllerConfigItem` on the original single-lock path, while preserving the earlier fix that only publishes the new snapshot after `SaveControllerConfig` succeeds. I tried splitting controller-config writes onto a second lock and then reverted it in `93d4d6e8e`: this is a low-frequency admin path, and the extra lock made the write invariants harder to follow without fixing a correctness bug.

Regression coverage:
- `TestManagerControllerConfigSnapshots`

Verification:
- failpoint-enabled `go test ./pkg/mcs/resourcemanager/server -count=1`
- `go test ./pkg/utils/keypath -count=1`
- `./.tools/bin/golangci-lint run --verbose ./pkg/mcs/resourcemanager/server/... --allow-parallel-runners`

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -93,19 +97,89 @@ func newKeyspaceResourceGroupManager(
...
 		log.Error("failed to parse the keyspace resource group meta info",
 			zap.Uint32("keyspace-id", krgm.keyspaceID), zap.String("name", name), zap.String("raw-value", rawValue), zap.Error(err))
+		return nil, err
+	}
+	if group.Name != name {
+		err := errors.Errorf("resource group key name %s does not match payload name %s", name, group.Name)
+		log.Error("resource group name mismatch in storage payload",
+			zap.Uint32("keyspace-id", krgm.keyspaceID),
+			zap.String("name", name),
+			zap.String("payload-name", group.Name),
```
bufferflies: The payload and name have been printed in the error.

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -93,19 +97,89 @@ func newKeyspaceResourceGroupManager(
...
+			zap.String("raw-value", rawValue),
+			zap.Error(err))
+		return nil, err
+	}
+	return group, nil
+}
+
+func validateResourceGroupProto(grouppb *rmpb.ResourceGroup) error {
+	if len(grouppb.Name) == 0 || len(grouppb.Name) > maxGroupNameLength {
+		return errs.ErrInvalidGroup
```
bufferflies: Do we need to distinguish them from name error or priority?

pkg/mcs/resourcemanager/server/server.go
```diff
@@ -307,6 +307,13 @@ func (*Server) GetResourceGroupWriteRole() ResourceGroupWriteRole {
 	return ResourceGroupWriteRoleLegacyAll
 }
 
+// EnableResourceGroupMetadataWatcher reports whether the independent RM service
```
bufferflies, line 310: Should this be implied later? If yes, pls add todo here?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return resourceGroupWatchTarget{
+		entryType:  entryType,
+		keyspaceID: constant.NullKeyspaceID,
+		groupName:  groupName,
+	}, true
+}
+
+func parseKeyspaceWatchPath(path string, entryType resourceGroupWatchEntryType) (resourceGroupWatchTarget, bool) {
+	keyspaceID, groupName, err := keypath.ParseKeyspaceResourceGroupPath(path)
```
bufferflies, line 117: maybe we loss this error?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,231 @@
...
+		groupName:  groupName,
+	}, true
+}
+
+func parseKeyspaceIDWatchPath(path string, entryType resourceGroupWatchEntryType) (resourceGroupWatchTarget, bool) {
+	if path == "" || strings.Contains(path, "/") {
+		return resourceGroupWatchTarget{}, false
+	}
+	keyspaceID, err := strconv.ParseUint(path, 10, 32)
+	if err != nil {
```
bufferflies, line 143: ditto

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -279,11 +302,87 @@ func (m *Manager) loadKeyspaceResourceGroups() error {
...
+	return m.loadServiceLimits()
+}
+
+func (m *Manager) loadServiceLimits() error {
 	return m.storage.LoadServiceLimits(func(keyspaceID uint32, serviceLimit float64) {
 		m.getOrCreateKeyspaceResourceGroupManager(keyspaceID, false).setServiceLimitFromStorage(serviceLimit)
 	})
 }
 
+func cloneControllerConfig(cfg *ControllerConfig) *ControllerConfig {
```
bufferflies, line 386: I don't this config have cloned, it make caller think the config is the new one

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: The group name is not sensitive information; we can see it in our monitoring.

pkg/mcs/resourcemanager/server/server.go
okJiang: A TODO is no longer needed here, as it has already been implemented.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: I kept `errs.ErrInvalidGroup` generic here on purpose.

`addResourceGroup()` already uses that same contract, and this helper was added to make the raw-load / watcher paths reuse the existing validation semantics instead of introducing a more specific error surface only for storage payloads. If we want to split invalid-name vs invalid-priority errors, I think that should be done consistently for both API and raw-load paths in a separate follow-up.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Fixed in `181086887`.

Malformed keyspace resource-group watch paths are no longer dropped completely silently: `parseKeyspaceWatchPath` now emits a debug log with the bad path and parse error before returning `false`. I also added `TestMetadataWatcherLogsMalformedWatchPaths` to cover this path.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Fixed in `181086887`.

Malformed keyspace service-limit watch paths now emit a debug log before returning `false`, so we keep the skip behavior but stop losing the parse failure entirely. The same `TestMetadataWatcherLogsMalformedWatchPaths` regression test covers this case too.

pkg/mcs/resourcemanager/server/manager.go
okJiang: This was introduced in `e7a2d93a3` for snapshot semantics.

The goal was to stop mutating the live `m.controllerConfig` in place: `UpdateControllerConfigItem()` edits a copied struct and only publishes it after the update succeeds, while `GetControllerConfig()` returns a snapshot instead of the shared pointer. That avoids readers or the save-failure path observing a partially updated config.

`ControllerConfig` is value-only today, so the shallow struct copy here is intentional and sufficient.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Follow-up in `52b345ffb`: I changed these messages from `debug` to `warn`.

I did not change this path to return an error, because `LoopWatcher` does not retry or abort on callback errors here: it only logs the error and keeps advancing both in the live watch path and in the initial load path. So returning an error would mostly create a second watcher-layer error log without changing recovery behavior. `warn` makes the malformed key visible while keeping the current skip semantics.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: Follow-up in `52b345ffb`: this is now `warn` rather than `debug` for the same reason.

I kept the callback return as `nil`, because in the current `LoopWatcher` implementation a callback error is logged and then the watcher continues anyway; it does not trigger a retry of that event or fail the load. Making the malformed key visible at `warn` level is the useful part here.

okJiang:
ping @bufferflies @JmPotato

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -93,19 +97,87 @@ func newKeyspaceResourceGroupManager(
...
+func (krgm *keyspaceResourceGroupManager) upsertResourceGroupFromRaw(name string, rawValue string) error {
+	group, err := krgm.parseResourceGroupFromRaw(name, rawValue)
+	if err != nil {
+		return err
+	}
+	if err := validateResourceGroupProto(group); err != nil {
 		return err
 	}
+
+	krgm.RLock()
```
bufferflies, line 153: It's better to wrap the curl for groups into two functions, such as groupExists and saveGroup? It makes the reader confused when to use the lock?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -204,12 +304,29 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
...
-	if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
-		return err
+	if txnStorage, ok := krgm.storage.(interface {
+		RunInTxn(context.Context, func(txn kv.Txn) error) error
+	}); ok {
+		if err := txnStorage.RunInTxn(context.Background(), func(txn kv.Txn) error {
+			if err := txn.Remove(keypath.KeyspaceResourceGroupSettingPath(krgm.keyspaceID, name)); err != nil {
+				return err
+			}
+			return txn.Remove(keypath.KeyspaceResourceGroupStatePath(krgm.keyspaceID, name))
```
bufferflies, line 314: If this delete failed, do we need to revert the KeyspaceResourceGroupSettingPath?

pkg/mcs/resourcemanager/server/keyspace_manager.go
```diff
@@ -204,12 +304,29 @@ func (krgm *keyspaceResourceGroupManager) deleteResourceGroup(name string) error
...
+		}); err != nil {
+			return err
+		}
+	} else {
+		if err := krgm.storage.DeleteResourceGroupSetting(krgm.keyspaceID, name); err != nil {
+			return err
+		}
+		if err := krgm.storage.DeleteResourceGroupStates(krgm.keyspaceID, name); err != nil {
```
bufferflies, line 322: If the error happened, don't return the error?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -114,6 +122,9 @@ func NewManager[T factoryProvider](srv bs.Server) *Manager {
 		writeRole = provider.GetResourceGroupWriteRole()
 	}
 	m := newManagerBase(fp.GetControllerConfig(), writeRole)
+	if provider, ok := any(fp).(metadataWatcherProvider); ok {
```
bufferflies, line 125: I think we should add an error if the ok is false to ensure the provider is not nil in the prod env.

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	return resourceGroupWatchTarget{
+		entryType:  entryType,
+		keyspaceID: uint32(keyspaceID),
+	}, true
+}
+
+func (m *Manager) initializeMetadataWatcher(ctx context.Context) error {
+	m.Lock()
+	m.krgms = make(map[uint32]*keyspaceResourceGroupManager)
```
bufferflies, line 157: Do we need to check if the krgms are nil to ensure they're init'd only once?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil
+	}
+}
+
+func (m *Manager) handleMetadataWatchDelete(key string) error {
+	target, ok := parseResourceGroupWatchPath(key)
+	if !ok {
+		return nil
+	}
+	if target.entryType == resourceGroupWatchEntryServiceLimit {
```
bufferflies, line 211: How about using switch target.entryType?

pkg/mcs/resourcemanager/server/metadata_watcher.go
```diff
@@ -0,0 +1,244 @@
...
+	switch target.entryType {
+	case resourceGroupWatchEntryController:
+		return m.applyControllerConfigFromRaw(rawValue)
+	case resourceGroupWatchEntrySettings:
+		return m.applyResourceGroupSettingFromRaw(target.keyspaceID, target.groupName, rawValue)
+	case resourceGroupWatchEntryStates:
+		return m.applyResourceGroupStatesFromRaw(target.keyspaceID, target.groupName, rawValue)
+	case resourceGroupWatchEntryServiceLimit:
+		return m.applyServiceLimitFromRaw(target.keyspaceID, rawValue)
+	default:
```
bufferflies, line 201: Does it meet our expectations if it meets an undefined type?

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: I kept this inline because the lock boundaries are the actual point here: we only need the manager map under `krgm.RLock()` to fetch the current pointer, and after that the update/insert paths diverge.

Extracting helpers like `groupExists` / `saveGroup` would still need to preserve the same read-lock vs write-lock split, so it would mostly hide the locking shape rather than simplify it.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: I don't think reverting `KeyspaceResourceGroupSettingPath` is the right fallback here.

In the transactional path this cannot happen because both deletes are done in one txn. In the non-transactional fallback, once the settings key delete has succeeded the group is already gone from the authoritative metadata view; trying to recreate that settings key would be another non-transactional write and could repersist stale metadata instead of restoring the original state cleanly.

pkg/mcs/resourcemanager/server/keyspace_manager.go
okJiang: I intentionally don't return that states-delete error in the fallback path.

At that point the settings key has already been deleted, so returning an error would make callers observe a failed delete even though the authoritative group metadata is already gone. The remaining states key is only orphaned cleanup data; logging a warning is useful, but surfacing it as the operation result would be misleading.

pkg/mcs/resourcemanager/server/manager.go
okJiang: I don't think adding an error on `!ok` is correct here.

`metadataWatcherProvider` is intentionally optional: the PD server path does not implement it and should stay on the legacy bootstrap path. So `ok == false` is a supported configuration, not a prod invariant violation.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: I don't think we should guard this on `krgms == nil`.

This initialization path is meant to rebuild manager state from metadata, so it should start from a fresh map each time, just like `loadKeyspaceResourceGroups()` does in the legacy path. Reusing an existing map here would keep stale managers across re-initialization instead of reconstructing from the watched metadata.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: I kept the explicit `if` chain here on purpose.

This function has three early-return cases with different preconditions, and the current structure makes each of them stand on its own. Converting it back to a `switch` would mostly be a style change, and we already had an earlier review round on keeping this delete control flow explicit rather than relying on more subtle switch behavior.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: The undefined-type path here is a defensive no-op.

`parseResourceGroupWatchPath` only returns the known entry types today. If a new enum value is added later but delete handling is not wired yet, treating it as a no-op is safer than accidentally falling into the settings-delete path for an unsupported type.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: I do not think we should guard this on `krgms == nil`.

This initialization path is meant to rebuild manager state from metadata, so it should start from a fresh map each time, just like `loadKeyspaceResourceGroups()` does in the legacy path. Reusing an existing map here would keep stale managers across re-initialization instead of reconstructing from the watched metadata.

pkg/mcs/resourcemanager/server/metadata_watcher.go
okJiang: I kept the explicit `if` chain here on purpose.

This function has three early-return cases with different preconditions, and the current structure makes each of them stand on its own. Converting it back to a `switch` would mostly be a style change, and we already had an earlier review round on keeping this delete control flow explicit rather than relying on more subtle switch behavior.

PR #10260, opened 2026-02-27
server,resourcemanager: route metadata writes via PD

okJiang:

After the metadata watcher scaffold landed, metadata write traffic should be handled by PD while RM focuses on token/read serving and tracks metadata via watcher updates.

Issue Number: ref #9737

This PR now carries the remaining routing cutover layer after `#10259`, plus the watcher-matrix follow-up that was previously split out as a stacked branch.

The branch was rebased onto the latest `master` on 2026-03-20 and then updated again on 2026-03-23 to absorb the remaining watcher-matrix coverage directly into `#10260`.

Behavior changes in this PR layer:

- PD fallback-disabled proxy handles metadata writes locally via a PD metadata manager.
- RM gRPC metadata write APIs reject writes with `FailedPrecondition` (must write via PD).
- RM standalone server startup enables metadata watcher bootstrap and metadata-write rejection together.
- Redirector integration test verifies:
  - direct RM metadata writes are rejected;
  - PD writes succeed;
  - RM eventually observes create/modify/delete via watcher.
- Watcher matrix integration test additionally verifies:
  - legacy and keyspace-scoped resource groups with the same name keep the expected precedence;
  - a new RM primary can bootstrap from watcher state after the previous primary exits;
  - the watcher recovers after etcd compaction once the blocked watch stream is released.

```commit-message
server,resourcemanager: route metadata writes via PD
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has HTTP API interfaces changed (behavioral routing change)

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * gRPC endpoints can reject or forward resource-group metadata write requests, with clear failure responses when local handling is required.

* **Bug Fixes**
  * Leader campaign aborts immediately on callback errors to avoid inconsistent promotion/state.

* **Improvements**
  * Resource-group handlers validate inputs, use a local metadata manager fallback, and eagerly initialize manager resources.

* **Tests**
  * Integration tests updated/added to assert gRPC rejection behavior, use polling for eventual consistency, and add watcher-matrix end-to-end scenarios.

### DIFF
--- a/tests/integrations/mcs/resourcemanager/resource_manager_test.go
+++ b/tests/integrations/mcs/resourcemanager/resource_manager_test.go
@@ -532,6 +532,9 @@ func (suite *resourceManagerClientTestSuite) cleanupResourceGroups(re *require.A
 			re.Contains(err.Error(), "cannot delete reserved group")
 			continue
 		}
+		if err != nil && strings.Contains(err.Error(), "resource group does not exist") {
+			continue
+		}
 		re.NoError(err)
 		re.Contains(deleteResp, "Success!")
 	}
@@ -576,12 +579,16 @@ func (suite *resourceManagerClientTestSuite) TestWatchResourceGroup() {
 		re.NoError(err)
 		re.Contains(resp, "Success!")
 
-		// Make sure the resource group active
-		meta, err = controller.GetResourceGroup(group.Name)
-		re.NotNil(meta)
-		re.NoError(err)
-		meta = controller.GetActiveResourceGroup(group.Name)
-		re.NotNil(meta)
+		// Under standalone RM discovery, metadata writes are applied on PD first and
+		// become visible to the RM controller after the watcher catches up.
+		testutil.Eventually(re, func() bool {
+			meta, err = controller.GetResourceGroup(group.Name)
+			if err != nil || meta == nil {
+				return false
+			}
+			meta = controller.GetActiveResourceGroup(group.Name)
+			return meta != nil
+		}, testutil.WithTickInterval(50*time.Millisecond))
 	}
 	// Mock modify resource groups
 	modifySettings := func(gs *rmpb.ResourceGroup, fillRate uint64) {


pkg/mcs/resourcemanager/server/server.go
```diff
@@ -215,6 +215,9 @@ func (s *Server) campaignLeader() {
 	for _, cb := range s.primaryCallbacks {
 		if err := cb(ctx); err != nil {
 			log.Error("failed to trigger the primary callback function", errs.ZapError(err))
+			// Do not promote a half-initialized primary: callback failures mean the
+			// service-ready state was not established, so this campaign must step down.
+			return
```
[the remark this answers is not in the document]
lhy1024: we need to solve it?

server/resource_group_proxy_service.go
```diff
@@ -87,44 +109,83 @@ func (s *resourceGroupProxyServer) GetResourceGroup(ctx context.Context, req *re
...
 	if err != nil {
 		return nil, err
 	}
-	resp, err := client.AddResourceGroup(ctx, req)
-	if err != nil {
-		s.closeClient(ctx)
+	if delegateClient != nil {
+		resp, err := delegateClient.AddResourceGroup(grpcutil.ResetForwardContext(ctx), req)
+		if err != nil {
+			s.closeDelegateClient(forwardedHost)
```
bufferflies: Do we need to check the error reason to decide whether to close the client?

okJiang: Fixed. We only need to drop the cached delegate connection for rebuild-worthy transport errors; application-level gRPC status errors should be returned directly without churning the connection.

I updated the resource-group proxy unary paths to gate `closeClient` / `closeDelegateClient` with `grpcutil.NeedRebuildConnection`, which matches the existing scheduling/TSO forwarding pattern.

pkg/mcs/resourcemanager/server/server.go
```diff
@@ -180,11 +185,21 @@ func (s *Server) primaryElectionLoop() {
...
+				timer.Stop()
+				return
+			case <-timer.C:
+			}
+		}
 	}
 }
 
-func (s *Server) campaignLeader() {
+func (s *Server) campaignLeader() bool {
```
bufferflies, line 205: What's the mean of the bool? Just mean callback failed?

okJiang: Fixed. I added a function comment on `campaignLeader()` to make the return value explicit: it returns `true` only when this node wins the campaign but fails service-ready callbacks, so the caller should back off before retrying.

bufferflies (approved):

Issue #10247, opened 2026-02-13
TestConfigTTLAfterTransferLeader timeout


okJiang:
### Which jobs are failing

```
panic: test timed out after 5m0s
	running tests:
		TestConfigTTLAfterTransferLeader (5m0s)
```

### CI link

https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10234/pull-unit-test-next-gen-3/2022158501463199744

okJiang:
Additional CI links for this flaky test:
- 2026-04-24 timeout: https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10618/pull-unit-test-next-gen-3/2047616001880625152
- 2026-04-25 assertion (Should not be: "pd2"): https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10622/pull-unit-test-next-gen-3/2048027091290558464

okJiang:
New CI link:
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10621/pull-unit-test-next-gen-3/2052280913881993216
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10645/pull-unit-test-next-gen-3/2052303272420577280

okJiang:
New CI links:
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10689/pull-unit-test-next-gen-3/2056996578069057536

okJiang:
New CI link:
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10689/pull-unit-test-next-gen-3/2057316979269701632

okJiang:
New CI links for TestConfigTTLAfterTransferLeader (ErrEtcdMoveLeader leader transfer timeout):
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10846/pull-unit-test-next-gen-3/2064252762685706240
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10798/pull-unit-test-next-gen-3/2064253109449789440
[...]

PR #10274, opened 2026-03-02
tests: stabilize TestConfigTTLAfterTransferLeader

okJiang:

Issue Number: ref #10247

`TestConfigTTLAfterTransferLeader` is flaky and can time out in CI.

### Root-cause evidence chain

- The flaky test currently drives leader switching by calling `leader.Destroy()` and relies on two fixed sleeps (`time.Sleep(2 * time.Second)`).
- `Destroy()` does `Close()` plus data-dir removal, which is a heavy teardown path and can block variably in CI.
- Fixed-delay sleeps make correctness depend on machine speed and scheduling jitter instead of explicit readiness.
- Issue #10247 reports `panic: test timed out after 5m0s` exactly on this test, which matches this non-deterministic orchestration shape.

### Historical analog

- Pattern match: leadership/watch synchronization + flaky stabilization from playbook.
- Related historical PR: #10134 (tests/integrations/client/client_test.go) changed brittle exact timing assumptions to bounded/condition-aware assertions in the same suite.

- Replace `Destroy()+sleep` based leader transfer with explicit `ResignLeader()`.
- Add explicit assertion that leader actually changes (`old != new`).
- Replace both fixed sleeps with `testutil.Eventually` checks:
  - Wait until TTL config is visible on current leader before transfer.
  - Wait until transferred leader observes expected config values.

This keeps test intent the same (TTL config survives leader transfer), but removes timing-based flakiness.

### Risk and impact

- Low risk: test-only change in `tests/integrations/client/client_test.go`.
- Main behavior impact: slower nodes now use condition-based waiting instead of fixed sleeps.

### Verification commands and results

- `make gotest GOTEST_ARGS='./tests/integrations/client -run TestConfigTTLAfterTransferLeader -count=1 -v'`
  - Failed (wrong module path for this repo layout).
- `cd tests/integrations && make gotest GOTEST_ARGS='-tags without_dashboard ./client -run TestConfigTTLAfterTransferLeader -count=1 -v'`
  - Passed.
- `cd tests/integrations && make gotest GOTEST_ARGS='-tags without_dashboard ./client -run TestConfigTTLAfterTransferLeader -count=10'`
  - Passed (`ok github.com/tikv/pd/tests/integrations/client 125.197s`).

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of cluster leadership transition tests by replacing fixed delays and leader destruction with explicit leader resignation and synchronized leader detection; switched to eventual-consistency checks that wait for persisted configuration (TTL-related snapshot limit and location-replacement flag) to converge after transfer.

### DIFF
--- a/tests/integrations/client/client_test.go
+++ b/tests/integrations/client/client_test.go
@@ -929,6 +929,7 @@ func TestConfigTTLAfterTransferLeader(t *testing.T) {
 	err = cluster.RunInitialServers()
 	re.NoError(err)
 	leaderName := cluster.WaitLeader()
+	re.NotEmpty(leaderName)
 	leader := cluster.GetServer(leaderName)
 	re.NoError(leader.BootstrapCluster())
 	addr := fmt.Sprintf("%s/pd/api/v1/config?ttlSecond=5", leader.GetAddr())


tests/integrations/client/client_test.go
```diff
@@ -928,7 +928,8 @@ func TestConfigTTLAfterTransferLeader(t *testing.T) {
 	defer cluster.Destroy()
 	err = cluster.RunInitialServers()
 	re.NoError(err)
-	leader := cluster.GetServer(cluster.WaitLeader())
+	leaderName := cluster.WaitLeader()
```
lhy1024, line 931: Do we need to check `re.NotEmpty(leaderName)`?

lhy1024 (commented):
the rest LGTM

tests/integrations/client/client_test.go
okJiang: Added `re.NotEmpty(leaderName)` before `GetServer` in `cfc6169ba`. `TestCluster.WaitLeader()` returns an empty string on timeout, so the test now fails explicitly before dereferencing the leader name.

Issue #9591, opened 2025-07-25
TestAPI/TestStores is unstable


lhy1024:
### Which jobs are failing
```
2025-07-25T02:34:16.5818943Z     api_check.go:30: 
2025-07-25T02:34:16.5819322Z         	Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/api_check.go:30
2025-07-25T02:34:16.5819889Z         	            				/home/runner/work/pd/pd/pkg/utils/testutil/api_check.go:166
2025-07-25T02:34:16.5820453Z         	            				/home/runner/work/pd/pd/pkg/utils/testutil/api_check.go:95
2025-07-25T02:34:16.5821144Z         	            				/home/runner/work/pd/pd/tests/integrations/mcs/scheduling/api_test.go:696
2025-07-25T02:34:16.5821601Z         	            				/home/runner/work/pd/pd/tests/testutil.go:416
2025-07-25T02:34:16.5822052Z         	            				/home/runner/work/pd/pd/tests/testutil.go:385
2025-07-25T02:34:16.5822733Z         	            				/home/runner/work/pd/pd/tests/integrations/mcs/scheduling/api_test.go:622
2025-07-25T02:34:16.5822880Z         	Error:      	Not equal: 
2025-07-25T02:34:16.5823071Z         	            	expected: 200
2025-07-25T02:34:16.5823254Z         	            	actual  : 404
2025-07-25T02:34:16.5823418Z         	Test:       	TestAPI/TestStores
2025-07-25T02:34:16.5823705Z         	Messages:   	resp: [PD:core:ErrStoreNotFound]store 7 not found
```

### CI link
https://github.com/tikv/pd/actions/runs/16512341188/job/46696484804

PR #10294, opened 2026-03-04
tests: fix flaky TestAPI/TestStores

bufferflies:

Issue Number: Ref #9591

`TestAPI/TestStores` is flaky because the scheduling server API may return 404 for store 7 (tombstone) immediately after `MustPutStore` returns. Although `MustPutStore` directly puts the store into the scheduling server's basic cluster, the store metadata may not be immediately visible via the API due to the asynchronous nature of the store watcher propagation through etcd.

Add a `testutil.Eventually` wait loop before querying individual stores on the scheduling server. The retry ensures all stores (including the tombstone store 7) are available via the scheduling server API before asserting on their metadata. The wait loop uses raw HTTP calls without `require` assertions to avoid the known anti-pattern of calling `t.FailNow()` inside a retry callback.

```commit-message
tests: fix flaky TestAPI/TestStores

Add a testutil.Eventually wait loop before querying individual stores
on the scheduling server. In the microservice environment, store
metadata (particularly for the tombstone store 7) may not be
immediately visible via the scheduling server API after MustPutStore
returns, because the store watcher propagates changes asynchronously
through etcd.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for the scheduling server by adding synchronization to ensure store metadata propagates before validation, reducing flaky test failures.

### DIFF
--- a/tests/integrations/mcs/scheduling/api_test.go
+++ b/tests/integrations/mcs/scheduling/api_test.go
@@ -680,6 +680,24 @@ func (suite *apiTestSuite) checkStores(cluster *tests.TestCluster) {
 	re.Equal(3, int(resp["count"].(float64)))
 	re.Len(resp["stores"].([]any), 3)
 	// Test /stores/{id}
+	// Wait until all stores (including the tombstone store 7) are available on
+	// the scheduling server before checking individual stores. The store metadata
+	// may not be immediately available via the scheduling server API due to the
+	// asynchronous nature of the store watcher propagation.
+	testutil.Eventually(re, func() bool {
+		for _, storeID := range []int{1, 6, 7} {
+			url := fmt.Sprintf("%s/scheduling/api/v1/stores/%d", scheServerAddr, storeID)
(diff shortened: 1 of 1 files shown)


tests/integrations/mcs/scheduling/api_test.go
```diff
@@ -680,6 +680,24 @@ func (suite *apiTestSuite) checkStores(cluster *tests.TestCluster) {
 	re.Equal(3, int(resp["count"].(float64)))
 	re.Len(resp["stores"].([]any), 3)
 	// Test /stores/{id}
+	// Wait until all stores (including the tombstone store 7) are available on
+	// the scheduling server before checking individual stores. The store metadata
+	// may not be immediately available via the scheduling server API due to the
+	// asynchronous nature of the store watcher propagation.
+	testutil.Eventually(re, func() bool {
+		for _, storeID := range []int{1, 6, 7} {
```
lhy1024, line 688: how about checking all stores?

PR #10295, opened 2026-03-04
*: upgrade testify from v1.10.0 to v1.11.1

### LINKED ISSUE #10164: Fix flaky tests in Eventually assertions to prepare for testify v1.11.0 upgrade

JmPotato:

Issue Number: Close #10164

```commit-message
Remove the `replace` directive that pinned testify to v1.10.0 and fix
all Eventually condition functions that used testify assertions
(re.NoError, re.Equal, ReadGetJSON, etc.) inside the condition, which
permanently marks tests as failed on first check with testify v1.11+.

testify v1.11 changed `Eventually` to check the condition immediately
before the first tick. This exposed an anti-pattern: using testify
assertions (which call t.Errorf()) inside Eventually condition functions.
When the condition fails on the first immediate check, the test is
permanently marked as failed even if the condition eventually succeeds.

The key changes are:
- Remove `replace` directives in root, tools, and tests/integrations
  go.mod files
- Upgrade client/go.mod from v1.9.0 to v1.11.1
- Add Try* utility functions to pkg/utils/testutil/api_check.go
  (TryReadGetJSON, TryCheckGetJSON, TryCheckPostJSON, etc.) that return
  errors instead of using testify assertions
- Replace all testify assertions inside Eventually condition functions
  with pure Go boolean expressions or Try* variants across ~50 test
  files
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved resilience by converting many immediate assertions into guarded checks inside retry/polling loops, reducing flakiness and deferring failures to eventual checks.
  * Added numerous error-returning "Try*" / "try*" helpers and non-panicking variants for HTTP, CLI and request helpers to enable graceful retries and clearer polling logic.
  * Tightened select verifications to ensure errors propagate through both synchronous returns and asynchronous callbacks.

* **Chores**
  * Upgraded testify to v1.11.1 and removed temporary version-pinning workarounds.

### DIFF
--- a/pkg/mcs/discovery/register_test.go
+++ b/pkg/mcs/discovery/register_test.go
@@ -19,7 +19,6 @@ import (
 	"os"
 	"regexp"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 	clientv3 "go.etcd.io/etcd/client/v3"
@@ -58,7 +57,10 @@ func TestRegister(t *testing.T) {
 	err = sr.Register()
 	re.NoError(err)
 	sr.cancel()
-	re.Empty(getKeyAfterLeaseExpired(ctx, re, client, sr.key))
+	testutil.Eventually(re, func() bool {
+		val, err := tryGetKey(ctx, client, sr.key)
+		return err == nil && val == ""
+	})
 
 	// Test the case that keepalive is failed when the etcd is restarted.
 	sr = NewServiceRegister(ctx, client, "test_service", "127.0.0.1:2", "127.0.0.1:2", DefaultLeaseInSeconds)
@@ -67,7 +69,10 @@ func TestRegister(t *testing.T) {
 	fname := testutil.InitTempFileLogger("info")
 	defer os.Remove(fname)
 	for i := range 3 {
-		re.Equal("127.0.0.1:2", getKeyAfterLeaseExpired(ctx, re, client, sr.key))
+		testutil.Eventually(re, func() bool {
+			val, err := tryGetKey(ctx, client, sr.key)
+			return err == nil && val == "127.0.0.1:2"
+		})
 		etcd.Server.HardStop() // close the etcd to make the keepalive failed
 		// ensure that the request is timeout
 		testutil.Eventually(re, func() bool {
@@ -94,17 +99,6 @@ func TestRegister(t *testing.T) {
 	// Close the last restarted etcd instance
 	etcd.Server.HardStop()
 	etcd.Close()
-}
-
-func getKeyAfterLeaseExpired(ctx context.Context, re *require.Assertions, client *clientv3.Client, key string) string {
-	time.Sleep(DefaultLeaseInSeconds * time.Second) // ensure that the lease is expired
-	time.Sleep(500 * time.Millisecond)              // wait for the etcd to clean up the expired keys
-	resp, err := client.Get(ctx, key)
-	re.NoError(err)
-	if len(resp.Kvs) == 0 {
-		return ""
-	}
-	return string(resp.Kvs[0].Value)
 }
 
 // tryGetKey retrieves a key from etcd, returning an error instead of using


JmPotato:
### Code review

Found 2 issues:

1. `TryReadGetJSON` and `TryCheckGetJSON` have identical implementations (same signature, same body), making one of them redundant. The original `ReadGetJSON` and `CheckGetJSON` had different semantics (`CheckGetJSON` accepted variadic `checkOpts`), but both `Try*` variants just call `tryCheckResp`. Either `TryCheckGetJSON` should be removed in favor of `TryReadGetJSON`, or `TryCheckGetJSON` should accept options to mirror the original `CheckGetJSON` behavior. (CLAUDE.md says "Simplicity First: Make every change as simple as possible. Impact minimal code.")

https://github.com/tikv/pd/blob/da9cbbbbb1e4b8fbdf024148146368798fea539a/pkg/utils/testutil/api_check.go#L183-L210

2. `tryRequestStatusBody` does not use `defer resp.Body.Close()`. If `io.ReadAll(resp.Body)` returns an error at line 537, the function returns without closing `resp.Body`, leaking the HTTP connection. The standard pattern is `defer resp.Body.Close()` immediately after the `client.Do` call succeeds. (CLAUDE.md says "Cancel timers/tickers; close resources with defer and error checks.")

https://github.com/tikv/pd/blob/da9cbbbbb1e4b8fbdf024148146368798fea539a/tests/server/api/store_test.go#L527-L544

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pkg/mcs/discovery/register_test.go
```diff
@@ -103,3 +106,18 @@ func getKeyAfterLeaseExpired(ctx context.Context, re *require.Assertions, client
 	}
 	return string(resp.Kvs[0].Value)
 }
+
+// tryGetKeyAfterLeaseExpired is like getKeyAfterLeaseExpired but returns an error
+// instead of using testify assertions, making it safe to use inside Eventually.
+func tryGetKeyAfterLeaseExpired(ctx context.Context, client *clientv3.Client, key string) (string, error) {
+	time.Sleep(DefaultLeaseInSeconds * time.Second) // ensure that the lease is expired
+	time.Sleep(500 * time.Millisecond)              // wait for the etcd to clean up the expired keys
```
JmPotato: Since it’s used inside an `eventually`, why do we still need to add sleep logic?

pkg/mcs/scheduling/server/rule/watcher.go
```diff
@@ -157,19 +158,22 @@ func (rw *Watcher) initializeRuleWatcher() error {
 		key := string(kv.Key)
 		if strings.HasPrefix(key, rw.rulesPathPrefix) {
 			log.Debug("delete placement rule", zap.String("key", key))
-			ruleJSON, err := rw.ruleStorage.LoadRule(strings.TrimPrefix(key, rw.rulesPathPrefix))
+			storeKey := strings.TrimPrefix(key, rw.rulesPathPrefix)
```
JmPotato, line 161: Is this consistent with the previous behavior?

PR #10302, opened 2026-03-05
core: fix some bug to improve the bucket compatibility

### LINKED ISSUE #10308: make report bucket work well
## Bug Report
Currently, bucket pointer CAS may use the new struct when bucket meta exists, which will cause CAS to always fail.

### What did you do?
notion

Cas success when region meta exist
### What did you see instead?
Cas failed
master and release-nextgen-8.5

bufferflies:

Issue Number: Close #10308

```commit-message
1. Rename bucket point as the report bucket to understand more easily 
2. Inherit will work even if the meta bucket is still working to avoid the cas report bucket always being successful.
3. Cas operator wrap into the region inner function avoid using bucket meta to do cas.
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Release Notes

* **New Features**
  * Added monitoring metrics for region cache misses and update operation success/failure tracking.

* **Bug Fixes**
  * Improved region bucket update mechanism for better handling of concurrent updates and metadata synchronization.
  * Enhanced bucket information handling during region metadata upgrades.

### DIFF
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -84,9 +84,9 @@ type RegionInfo struct {
 	replicationStatus         *replication_modepb.RegionReplicationStatus
 	queryStats                *pdpb.QueryStats
 	flowRoundDivisor          uint64
-	// buckets is not thread unsafe, it should be accessed by the request `report buckets` with greater version.
+	// reportBuckets is not thread unsafe, it should be accessed by the request `report reportBuckets` with greater version.
 	// todo: keep it compatible with previous design, we can remove it later.
-	buckets unsafe.Pointer
+	reportBuckets unsafe.Pointer
 	// source is used to indicate region's source, such as Storage/Sync/Heartbeat.
 	source RegionSource
 	// ref is used to indicate the reference count of the region in root-tree and sub-tree.
@@ -287,9 +287,9 @@ func (r *RegionInfo) Inherit(origin *RegionInfo, bucketEnable bool) {
 			r.approximateSize = EmptyRegionApproximateSize
 		}
 	}
-	if bucket := r.GetBuckets(); bucketEnable && bucket == nil && origin != nil {
-		inherited := atomic.LoadPointer(&origin.buckets)
-		atomic.StorePointer(&r.buckets, inherited)
+	if bucketEnable && origin != nil && atomic.LoadPointer(&r.reportBuckets) == nil {
+		inherited := atomic.LoadPointer(&origin.reportBuckets)
+		atomic.StorePointer(&r.reportBuckets, inherited)
 	}
 }
 
@@ -321,7 +321,7 @@ func (r *RegionInfo) Clone(opts ...RegionCreateOption) *RegionInfo {
 		approximateKeys:           r.approximateKeys,
 		interval:                  typeutil.DeepClone(r.interval, TimeIntervalFactory),
 		replicationStatus:         r.replicationStatus,
-		buckets:                   r.buckets,
+		reportBuckets:             atomic.LoadPointer(&r.reportBuckets),
 		queryStats:                typeutil.DeepClone(r.queryStats, QueryStatsFactory),
 		bucketMeta:                r.bucketMeta,
 	}
@@ -618,7 +618,7 @@ func (r *RegionInfo) GetStat() *pdpb.RegionStat {
 // UpdateBuckets sets the buckets of the region, only use by tests.
 func (r *RegionInfo) UpdateBuckets(buckets, old *metapb.Buckets) bool {
 	if buckets == nil {
-		atomic.StorePointer(&r.buckets, nil)
+		atomic.StorePointer(&r.reportBuckets, nil)
 		return true
 	}
 	// only need to update bucket keys, versions.
@@ -627,16 +627,17 @@ func (r *RegionInfo) UpdateBuckets(buckets, old *metapb.Buckets) bool {
 		Version:  buckets.GetVersion(),
 		Keys:     buckets.GetKeys(),
 	}
-	return atomic.CompareAndSwapPointer(&r.buckets, unsafe.Pointer(old), unsafe.Pointer(newBuckets))
+	return atomic.CompareAndSwapPointer(&r.reportBuckets, unsafe.Pointer(old), unsafe.Pointer(newBuckets))
 }
 
 // GetReportBuckets returns the buckets of the region, only use by tests.
 func (r *RegionInfo) GetReportBuckets() *metapb.Buckets {
-	return (*metapb.Buckets)(atomic.LoadPointer(&r.buckets))
+	return (*metapb.Buckets)(atomic.LoadPointer(&r.reportBuckets))
 }
 
-func (r *RegionInfo) CompareAndSetBuckets(buckets *metapb.Buckets) bool {
-	old := (*metapb.Buckets)(atomic.LoadPointer(&r.buckets))
+// CompareAndSetReportBuckets compares and sets the report buckets of the region.
+func (r *RegionInfo) CompareAndSetReportBuckets(buckets *metapb.Buckets) bool {
+	old := (*metapb.Buckets)(atomic.LoadPointer(&r.reportBuckets))
 	// region should not update if the version of the buckets is less than the old one.
 	if old != nil {
 		reportVersion := buckets.GetVersion()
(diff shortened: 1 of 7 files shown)


bufferflies:
@coderabbitai full review

pkg/core/region.go
```diff
@@ -627,7 +627,32 @@ func (r *RegionInfo) UpdateBuckets(buckets, old *metapb.Buckets) bool {
...
+			return true
+		} else if reportVersion == old.GetVersion() {
+			versionNotChangeCounter.Inc()
+			return true
+		}
+	}
+	failpoint.Inject("concurrentBucketHeartbeat", func() {
+		time.Sleep(500 * time.Millisecond)
+	})
+	return atomic.CompareAndSwapPointer(&r.reportBuckets, unsafe.Pointer(old), unsafe.Pointer(buckets))
```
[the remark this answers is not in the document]
bufferflies: No need to use clone, because the value will not change again.

pkg/core/region.go
```diff
@@ -627,7 +627,32 @@ func (r *RegionInfo) UpdateBuckets(buckets, old *metapb.Buckets) bool {
...
+	old := (*metapb.Buckets)(atomic.LoadPointer(&r.reportBuckets))
+	// region should not update if the version of the buckets is less than the old one.
+	if old != nil {
+		reportVersion := buckets.GetVersion()
+		if reportVersion < old.GetVersion() {
+			versionStaleCounter.Inc()
+			return true
+		} else if reportVersion == old.GetVersion() {
+			versionNotChangeCounter.Inc()
+			return true
```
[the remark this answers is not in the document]
bufferflies: yes, the return indicate that the caller should retry again.

tests/testutil.go
```diff
@@ -684,16 +684,12 @@ func InitRegions(regionLen int) []*core.RegionInfo {
...
-				Version: 1,
-			}
-			region := core.NewRegionInfo(r, r.Peers[0], core.SetSource(core.Heartbeat), core.WithBucketMeta(meta))
 			buckets := &metapb.Buckets{
 				RegionId: r.Id,
 				Keys:     [][]byte{r.StartKey, r.EndKey},
 				Version:  1,
 			}
+			region := core.NewRegionInfo(r, r.Peers[0], core.SetSource(core.Heartbeat), core.SetBuckets(buckets))
 			region.UpdateBuckets(buckets, region.GetBuckets())
```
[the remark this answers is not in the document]
bufferflies: Yes, I just hope the bucket meta is prioritized

pkg/core/region.go
```diff
@@ -287,9 +287,9 @@ func (r *RegionInfo) Inherit(origin *RegionInfo, bucketEnable bool) {
 			r.approximateSize = EmptyRegionApproximateSize
 		}
 	}
-	if bucket := r.GetBuckets(); bucketEnable && bucket == nil && origin != nil {
-		inherited := atomic.LoadPointer(&origin.buckets)
-		atomic.StorePointer(&r.buckets, inherited)
+	if bucketEnable && origin != nil && atomic.LoadPointer(&r.reportBuckets) == nil {
+		inherited := atomic.LoadPointer(&origin.reportBuckets)
```
rleungx, line 291: Will it cause the bucket roll back to nil?

pkg/core/region.go
```diff
@@ -627,7 +627,32 @@ func (r *RegionInfo) UpdateBuckets(buckets, old *metapb.Buckets) bool {
...
+	return atomic.CompareAndSwapPointer(&r.reportBuckets, unsafe.Pointer(old), unsafe.Pointer(newBuckets))
+}
+
+// GetReportBuckets returns the buckets of the region, only use by tests.
+func (r *RegionInfo) GetReportBuckets() *metapb.Buckets {
+	return (*metapb.Buckets)(atomic.LoadPointer(&r.reportBuckets))
+}
+
+// CompareAndSetReportBuckets compares and sets the report buckets of the region.
+func (r *RegionInfo) CompareAndSetReportBuckets(buckets *metapb.Buckets) bool {
```
rleungx, line 639: Same question for here, if tikv rollback, will bucket rollback too?

pkg/core/region.go
bufferflies: Do you mean the reportBuckets points? It only set value even if the report bucket is nil

pkg/core/region.go
bufferflies: The report bucket will not; it must increase. But the result of the GetBucket API may roll back because the bucket metadata may exceed the report bucket.

pkg/core/region.go
rleungx: assume that the tikv rollback, and the bucket meta suddenly become nil

bufferflies: yes, but the bucket info will be refresh after restart, so the report bucket should be greater than the previous.

pkg/core/region.go
lhy1024: I think there is still a rollback issue in the visible result of `GetBuckets()` for the same region.

  Case 1:
  - PD has an old report-bucket value `v1`
  - PD later receives a newer bucket meta from region heartbeat `v2`
  - `GetBuckets()` shows the newer bucket meta `v2`
  - when a later heartbeat does not carry bucket meta anymore, `GetBuckets()` falls back to the old report-bucket value `v1`

  Case 2:
  - PD has an old report-bucket value `v1`
  - PD later receives a newer bucket meta `v3`
  - before bucket meta disappears, PD receives a delayed report-bucket update `v2`, which is newer than the old report bucket but still older than the bucket meta
  - when a later heartbeat does not carry bucket meta anymore, `GetBuckets()` falls back to that delayed report-bucket value `v2` rather than `v3`

In both cases, the visible result of `GetBuckets()` rolls back to an older version than PD has already observed.

Could we add small unit tests in `pkg/core/region_test.go` for these two cases?

lhy1024: I tried two small unit tests locally, and both fail on the current patch.

  ```go
  func TestReportBucketsShouldNotFallBackBelowBucketMeta(t *testing.T) {
        re := require.New(t)
        origin := NewRegionInfo(&metapb.Region{Id: 1}, nil)
        re.True(origin.UpdateBuckets(&metapb.Buckets{
                RegionId: 1,
                Version:  1,
                Keys:     [][]byte{[]byte("a"), []byte("b")},
        }, nil))
        upgraded := origin.Clone(SetBuckets(&metapb.Buckets{
                RegionId: 1,
                Version:  3,
                Keys:     [][]byte{[]byte("a"), []byte("c")},
        }))
        upgraded.Inherit(origin, true)
        re.Equal(uint64(3), upgraded.GetBuckets().GetVersion())
        re.Equal(uint64(1), upgraded.GetReportBuckets().GetVersion())

        re.True(upgraded.CompareAndSetReportBuckets(&metapb.Buckets{
                RegionId: 1,
                Version:  2,
                Keys:     [][]byte{[]byte("a"), []byte("b")},
        }))
        re.Equal(uint64(2), upgraded.GetReportBuckets().GetVersion())
        re.Equal(uint64(3), upgraded.GetBuckets().GetVersion())

        downgraded := upgraded.Clone(SetBuckets(nil))
        downgraded.Inherit(upgraded, true)
        re.Equal(uint64(3), downgraded.GetBuckets().GetVersion())
  }

  func TestBucketMetaDisappearShouldNotRevealOldReportBuckets(t *testing.T) {
        re := require.New(t)

        origin := NewRegionInfo(&metapb.Region{Id: 1}, nil)
        re.True(origin.UpdateBuckets(&metapb.Buckets{
                RegionId: 1,
                Version:  1,
                Keys:     [][]byte{[]byte("a"), []byte("b")},
        }, nil))

        upgraded := origin.Clone(SetBuckets(&metapb.Buckets{
                RegionId: 1,
                Version:  3,
                Keys:     [][]byte{[]byte("a"), []byte("c")},
        }))
        re.Equal(uint64(3), upgraded.GetBuckets().GetVersion())

        downgraded := upgraded.Clone(SetBuckets(nil))
        downgraded.Inherit(upgraded, true)
        re.Equal(uint64(3), downgraded.GetBuckets().GetVersion())
  }

bufferflies: Yes, it may happen from a code perspective. But it will never happen because the source of the bucket meta and the report bucket is the same.

bufferflies: Only happen if the tikv rollback, but the bucket will refresh with bigger version

bufferflies: tikv code： https://github.com/tidbcloud/cloud-storage-engine/blob/285169685ecd9b0bd67b73e3a6ea1266af307104/components/rfstore/src/store/worker/pd.rs#L2113-L2141

lhy1024: Got it. This makes sense to me now.

On the normal TiKV path, both bucket meta and report bucket come from the same `bucket_stat`, so they should be produced together.

PR #10306, opened 2026-03-05
tests: fix flaky TestTransferLeaderForScheduler

### LINKED ISSUE #10305: Flaky test: TestTransferLeaderForScheduler
### Which jobs are failing

`TestTransferLeaderForScheduler` in `tests/server/cluster/cluster_test.go`

### CI link

https://github.com/tikv/pd/actions/runs/22724165515/job/65894471830?pr=10303

The test times out at line 1658 waiting for schedulers to be registered after a leader transfer:

```go
schedulersController = rc1.GetCoordinator().GetSchedulersController()
    return len(schedulersController.GetSchedulerNames()) == schedulersNum
})
```

`testutil.Eventually` reports "Condition never satisfied" — the scheduler count never reaches the expected `schedulersNum` within the timeout. This is a timing-sensitive operation where the new leader needs to fully load and start all schedulers after a leader transfer, which can be slow under CI resource pressure.

### Anything else

This failure is not related to any code changes in PR #10303. The test file `tests/server/cluster/cluster_test.go` was not modified.

JmPotato:

Issue Number: Close #10305

```commit-message
Replace `time.Sleep(time.Second)` followed by a direct `re.True(IsPrepared())`
assertion with `testutil.Eventually` to properly wait for the raft cluster to
become prepared after leader transfer. This matches the pattern already used
in the third leader transfer block of the same test.

Under CI resource pressure, a fixed 1-second sleep may not be enough for the
new leader to fully prepare, causing the subsequent scheduler count check to
time out.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Replaced fixed timing in cluster tests with a polling-based wait for scheduler readiness, improving reliability of leader-transfer scenarios and timing-sensitive assertions.

### DIFF
--- a/tests/server/cluster/cluster_test.go
+++ b/tests/server/cluster/cluster_test.go
@@ -1652,7 +1652,7 @@ func TestTransferLeaderForScheduler(t *testing.T) {
 	id = leaderServer.GetAllocator()
 	putRegionWithLeader(re, rc1, id, 1)
 	testutil.Eventually(re, func() bool {
-		return leaderServer.GetRaftCluster().IsPrepared()
+		return rc1.GetCoordinator().AreSchedulersInitialized()
 	})
 	// Check scheduler updated.
 	schedulersController = rc1.GetCoordinator().GetSchedulersController()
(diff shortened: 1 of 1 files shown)


YuhaoZhang00:
`Eventually(IsPrepared())` is a good change to have.

That said, it does not fully justify issue #10305. The reported failure is at the scheduler-count check on line 1658. If the new leader were not prepared yet, the old test should have already failed at the previous `re.True(leaderServer.GetRaftCluster().IsPrepared())` on line 1655. So the issue is not just that `IsPrepared()` is still `false`.

A more targeted fix would be to increase the wait at line 1658/1659:
```
		return len(schedulersController.GetSchedulerNames()) == schedulersNum
	}, testutil.WithWaitFor(60*time.Second))
```

rleungx (approved):

PR #10312, opened 2026-03-07
*: fix potential bug by ai in pkg/schedule/schedulers

### LINKED ISSUE #10449: pkg/schedule/schedulers: avoid panics on malformed scheduler config input
## Development Task

The scheduler config and scheduler-related API paths under `pkg/schedule/schedulers` may panic or return inaccurate HTTP errors when receiving malformed or unexpected input.

Typical risky cases include:
- unsafe type assertions on scheduler config fields parsed from JSON
- unexpected scheduler type casts in transfer witness handling
- incorrect HTTP status codes for invalid user input
- nil-response handling gaps in scheduler API error paths

This task tracks hardening those code paths so that invalid input is rejected safely with clear client-facing errors instead of causing runtime panics or misleading 5xx responses.

Related work:
- ref #10326
- implemented in #10312

bufferflies:

Issue Number: Close #10449

```commit-message
- It removes several unsafe type assertions on JSON input (like input["x"].(string)), replacing them with checked parsing and clear 400 Bad Request responses, so malformed payloads no longer panic PD.  
  - Key places: pkg/schedule/schedulers/balance_range.go, pkg/schedule/schedulers/evict_leader.go, pkg/schedule/schedulers/grant_hot_region.go, pkg/schedule/schedulers/grant_leader.go
- It fixes a risky type cast in transfer witness handling by making RecvRegionInfo safe (nil on wrong scheduler type) and guarding before channel send.  
  - pkg/schedule/schedulers/transfer_witness_leader.go, pkg/schedule/schedulers/scheduler_controller.go
- It corrects HTTP error behavior in scheduler API paths:
  - invalid user options now return 400 instead of 500
  - avoids nil dereference when delete redirect fails and response is nil
  - handles “handler unavailable” cases explicitly  
  - server/api/scheduler.go
- It adds regression tests specifically asserting no panic and expected status codes for bad/odd inputs.  
  - pkg/schedule/schedulers/balance_range_test.go, pkg/schedule/schedulers/evict_leader_test.go, pkg/schedule/schedulers/grant_hot_region_test.go, pkg/schedule/schedulers/grant_leader_test.go, pkg/schedule/schedulers/transfer_witness_leader_test.go
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter input validation and clearer client-facing errors for scheduler config updates (engine, rule, alias, timeouts, store IDs, ranges), safer control flow to avoid runtime errors and unintended pauses/resumes, and more accurate HTTP status codes for invalid input.

* **Tests**
  * Added tests for invalid field types, malformed ranges, wrong scheduler types, and config update error cases to prevent regressions.

### DIFF
--- a/errors.toml
+++ b/errors.toml
@@ -56,6 +56,11 @@ error = '''
 the option %s does not exist
 '''
 
+["PD:apiutil:ErrOptionTypeInvalid"]
+error = '''
+the type of option %s is invalid
+'''
+
 ["PD:apiutil:ErrRedirect"]
 error = '''
 redirect failed
--- a/pkg/errs/errno.go
+++ b/pkg/errs/errno.go
@@ -311,6 +311,7 @@ var (
 var (
 	ErrRedirect             = errors.Normalize("redirect failed", errors.RFCCodeText("PD:apiutil:ErrRedirect"))
 	ErrOptionNotExist       = errors.Normalize("the option %s does not exist", errors.RFCCodeText("PD:apiutil:ErrOptionNotExist"))
+	ErrOptionTypeInvalid    = errors.Normalize("the type of option %s is invalid", errors.RFCCodeText("PD:apiutil:ErrOptionTypeInvalid"))
 	ErrRedirectNoLeader     = errors.Normalize("redirect finds no leader", errors.RFCCodeText("PD:apiutil:ErrRedirectNoLeader"))
 	ErrRedirectToNotLeader  = errors.Normalize("redirect to not leader", errors.RFCCodeText("PD:apiutil:ErrRedirectToNotLeader"))
 	ErrRedirectToNotPrimary = errors.Normalize("redirect to not primary", errors.RFCCodeText("PD:apiutil:ErrRedirectToNotPrimary"))
--- a/pkg/utils/apiutil/apiutil.go
+++ b/pkg/utils/apiutil/apiutil.go
@@ -335,13 +335,18 @@ func CollectEscapeStringOption(option string, input map[string]any, collectors .
 
 // CollectStringOption is used to collect string using from input map for given option
 func CollectStringOption(option string, input map[string]any, collectors ...func(v string)) error {
-	if v, ok := input[option].(string); ok {
-		for _, c := range collectors {
-			c(v)
-		}
-		return nil
+	v, exist := input[option]
+	if !exist {
+		return errs.ErrOptionNotExist.FastGenByArgs(option)
 	}
-	return errs.ErrOptionNotExist.FastGenByArgs(option)
+	str, ok := v.(string)
+	if !ok {
+		return errs.ErrOptionTypeInvalid.FastGenByArgs(option)
+	}
+	for _, c := range collectors {
+		c(str)
+	}
+	return nil
 }
 
 // ParseKey is used to parse interface into []byte and string


liyishuai (approved):
Tesed on my side. The AI review is worth considering. Otherwise LGTM.

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -427,8 +428,28 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
...
+	if hasRanges {
+		switch val := rangesVal.(type) {
+		case []string:
+			ranges = val
+		case []any:
+			ranges = make([]string, 0, len(val))
+			for _, item := range val {
+				s, ok := item.(string)
+				if !ok {
+					handler.config.resumeLeaderTransferIfExist(id)
```
rleungx: If there is a wrong request, e.g., an invalid range type, will it resume incorrectly?

bufferflies: Good catch. I think the original code is pretty good; we should ignore the other type.

rleungx:
There is still one missing case here: when creating a new balance-range scheduler, `timeout` is still read via `apiutil.CollectStringOption()`, which treats a wrong-typed field the same as a missing one. So a request like `{"timeout":123}` will still silently fall back to the default `30m` instead of returning `400 Bad Request`. Could we distinguish between “not provided” and “wrong type” here, so this path stays consistent with the stricter validation added elsewhere in this PR?

bufferflies:
> CollectStringOption

```
// CollectStringOption is used to collect string using from input map for given option
func CollectStringOption(option string, input map[string]any, collectors ...func(v string)) error {
	if v, ok := input[option].(string); ok {
		for _, c := range collectors {
			c(v)
		}
		return nil
	}
	return errs.ErrOptionNotExist.FastGenByArgs(option)
}
```

It will return ErrOptionNotExist if the key does not exist or is not a string. Do you suggest disguising them?

pkg/schedule/schedulers/transfer_witness_leader_test.go
```diff
@@ -89,3 +89,11 @@ func TestTransferWitnessLeaderWithUnhealthyPeer(t *testing.T) {
 	ops, _ = sl.Schedule(tc, false)
 	re.Empty(ops)
 }
+
+func TestRecvRegionInfoWithWrongSchedulerType(t *testing.T) {
```
okJiang: Is this test unnecessary?

bufferflies: removed, done

server/api/scheduler.go
```diff
@@ -142,13 +146,13 @@ func (h *schedulerHandler) CreateScheduler(w http.ResponseWriter, r *http.Reques
 			if errors.ErrorEqual(err, errs.ErrOptionNotExist) {
 				collector(defaultTimeout)
 			} else {
-				h.r.JSON(w, http.StatusInternalServerError, err.Error())
+				h.r.JSON(w, http.StatusBadRequest, err.Error())
 				return
 			}
 		}
```
lhy1024, line 152: CollectStringOption returns `ErrOptionNotExist` for non-string type and option not exist.

The core issue is that in the create path, {"timeout":123} is still treated as “timeout not provided”, so it silently falls back to the default value instead of returning 400.

lhy1024:
```
func TestCollectStringOptionTypeMismatchMatchesErrOptionNotExist(t *testing.T) {
        input := map[string]any{"timeout": 123}

        err := CollectStringOption("timeout", input)
        require.Error(t, err)
        require.True(t, perrors.ErrorEqual(err, errs.ErrOptionNotExist))
  }
```
```
go test ./pkg/utils/apiutil -run '^TestCollectStringOptionTypeMismatchMatchesErrOptionNotExist$' -count=1
ok    github.com/tikv/pd/pkg/utils/apiutil    0.007s
```

pkg/schedule/schedulers/balance_range.go
```diff
@@ -90,27 +90,52 @@ func (handler *balanceRangeSchedulerHandler) addJob(w http.ResponseWriter, r *ht
...
+			if err != nil {
+				handler.rd.JSON(w, http.StatusBadRequest, fmt.Sprintf("timeout:%s is invalid", timeoutStr))
+				return
+			}
+			if timeout <= 0 {
+				handler.rd.JSON(w, http.StatusBadRequest, "timeout must be positive")
+				return
+			}
+			job.Timeout = timeout
+		}
```
lhy1024, line 138: For the initial HTTP create path, timeout: "0s" and negative values are still accepted, because the create decoder only calls time.ParseDuration. It is not consistency with here.

lhy1024:
```
func (suite *scheduleTestSuite) TestBalanceRangeCreateRejectsZeroTimeout() {
	suite.te.RunTest(func(cluster *tests.TestCluster) {
		suite.checkBalanceRangeCreateRejectsNonPositiveTimeout(cluster, "0s")
	})
}

func (suite *scheduleTestSuite) TestBalanceRangeCreateRejectsNegativeTimeout() {
	suite.te.RunTest(func(cluster *tests.TestCluster) {
		suite.checkBalanceRangeCreateRejectsNonPositiveTimeout(cluster, "-1s")
	})
}

func (suite *scheduleTestSuite) checkBalanceRangeCreateRejectsNonPositiveTimeout(cluster *tests.TestCluster, timeout string) {
	re := suite.Require()
	input := map[string]any{
		"alias":     "test.sysbench.partition(p1)",
		"engine":    "tikv",
		"rule":      "leader-scatter",
		"name":      types.BalanceRangeScheduler.String(),
		"timeout":   timeout,
		"start-key": url.QueryEscape("100"),
		"end-key":   url.QueryEscape("200"),
	}
	data, err := json.Marshal(input)
	re.NoError(err)
	leaderAddr := cluster.GetLeaderServer().GetAddr()
	urlPrefix := fmt.Sprintf("%s/pd/api/v1", leaderAddr)
	re.NoError(testutil.CheckPostJSON(
		tests.TestDialClient,
		urlPrefix+"/schedulers",
		data,
		testutil.Status(re, http.StatusBadRequest),
	))
}
```

lhy1024:
```
go test -tags without_dashboard ./tests/server/api -run 'Test(NonMicroservice|Microservice)SchedulingTestSuite/
  TestBalanceRangeCreateRejectsZeroTimeout$' -count=1
  go test -tags without_dashboard ./tests/server/api -run 'Test(NonMicroservice|Microservice)SchedulingTestSuite/
  TestBalanceRangeCreateRejectsNegativeTimeout$' -count=1
```
```
  0s：

  --- FAIL: TestMicroserviceSchedulingTestSuite (18.20s)
      testutil.go:425: start test TestBalanceRangeCreateRejectsZeroTimeout in microservice environment
      --- FAIL: TestMicroserviceSchedulingTestSuite/TestBalanceRangeCreateRejectsZeroTimeout (12.34s)
          api_check.go:30:
               Error:       Not equal:
                            expected: 400
                            actual  : 200
               Messages:    resp: "The scheduler is created."
  FAIL
  FAIL  github.com/tikv/pd/tests/server/api     21.550s
  FAIL

  -1s：

  --- FAIL: TestMicroserviceSchedulingTestSuite (18.19s)
      testutil.go:425: start test TestBalanceRangeCreateRejectsNegativeTimeout in microservice environment
      --- FAIL: TestMicroserviceSchedulingTestSuite/TestBalanceRangeCreateRejectsNegativeTimeout (12.34s)
          api_check.go:30:
               Error:       Not equal:
                            expected: 400
                            actual  : 200
               Messages:    resp: "The scheduler is created."
  FAIL
  FAIL  github.com/tikv/pd/tests/server/api     21.485s
  FAIL
```

server/api/scheduler.go
rleungx, line 251: If there is a bad rule, it still returns 500, instead of 400.

server/api/scheduler.go
bufferflies: It must be string lit, can't be number lit

server/api/scheduler.go
bufferflies: good  catch, done

pkg/schedule/schedulers/balance_range.go
bufferflies: done,

server/api/scheduler.go
lhy1024: Should we distinguish between non-string type and option not exist for timeout?

server/api/scheduler.go
rleungx: Actually, AddScheduler can also return 500 sometimes.

PR #10203, opened 2026-02-03
test: fix flaky test TestForwardTestSuite in next-gen

okJiang:
### What problem does this PR solve?

Issue Number: Close https://github.com/tikv/pd/issues/10200

### What is changed and how does it work?

```commit-message
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Code changes

- Has the configuration change
- Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- Has persistent data change

Side effects

- Possible performance regression
- Increased code complexity
- Breaking backward compatibility

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup cleanup to reliably stop embedded services, close clients and network connections on startup failures, reducing resource leaks.

* **Tests**
  * Hardened cluster tests to treat cluster ID mismatches like port conflicts, automatically recreate servers with new ports and retry to improve test reliability.

Issue #10154, opened 2026-01-15
TestSwitchModeDuringWorkload is unstable


lhy1024:
### Which jobs are failing
```
2026-01-15T02:56:02.6464737Z     testutil.go:68: 
2026-01-15T02:56:02.6465109Z         	Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:68
2026-01-15T02:56:02.6466117Z         	            				/home/runner/work/pd/pd/tests/integrations/mcs/resourcemanager/resource_manager_test.go:492
2026-01-15T02:56:02.6466312Z         	Error:      	Condition never satisfied
2026-01-15T02:56:02.6466591Z         	Test:       	TestSwitchModeDuringWorkload/pd-to-standalone
```

### CI link
https://github.com/tikv/pd/actions/runs/20988510834/job/60426623138?pr=10042

okJiang:
## Flaky Test
### Which jobs are failing
```text
=== RUN   TestSwitchModeDuringWorkload/pd-to-standalone
Error Trace:	/home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:68
            				/home/runner/work/pd/pd/tests/integrations/mcs/resourcemanager/resource_manager_test.go:492
Error:	Condition never satisfied
Test:	TestSwitchModeDuringWorkload/pd-to-standalone
```
### New CI link
https://github.com/tikv/pd/actions/runs/22842015542/job/66249971803
https://github.com/tikv/pd/actions/runs/22856930709/job/66299693231
### Reason for failure (if possible)

### Anything else

okJiang:
## Flaky Test
### Which jobs are failing
```text
=== RUN   TestSwitchModeDuringWorkload/pd-to-standalone
Error Trace: /home/runner/work/pd/pd/pkg/utils/testutil/testutil.go:68
             /home/runner/work/pd/pd/tests/integrations/mcs/resourcemanager/resource_manager_test.go:492
Error:       Condition never satisfied
Test:        TestSwitchModeDuringWorkload/pd-to-standalone
```
### New CI link
https://github.com/tikv/pd/actions/runs/22888276891/job/66405802988
### Reason for failure (if possible)

### Anything else

okJiang:
## Flaky Test
### Which jobs are failing
```text
    testutil.go:68:
        	Error Trace:	/home/prow/go/src/github.com/tikv/pd/pkg/utils/testutil/testutil.go:68
[...]

PR #10316, opened 2026-03-09
tests: stabilize TestSwitchModeDuringWorkload

okJiang:

Issue Number: ref #10154

`TestSwitchModeDuringWorkload/pd-to-standalone` is flaky.

Root-cause evidence chain

- Issue #10154 and recent rerun logs fail at:
  - `pkg/utils/testutil/testutil.go:68`
  - `tests/integrations/mcs/resourcemanager/resource_manager_test.go:492`
  - `Condition never satisfied`.
- CI job logs (`22856930709` / `66299693231`) show transient RM discovery/connectivity errors around the switch window, e.g. repeated `resource manager error` with `connection refused`, while the test already starts counting post-switch success.
- Current test only gates on leader/service-url checks before setting `switched=true`, but the controller may still be in transient degraded mode until the first successful token-bucket response after endpoint switch.
- Therefore `okAfter` can stay below threshold in the bounded wait even though switch eventually recovers.

Historical analog

- Pattern: `flaky_stabilization` + `test_harness_alignment` from flaky fix playbook.
- Closest corpus analog: #10203 (test readiness synchronization during transient service state) and #10134 (unstable-test stabilization with explicit readiness conditions).

- In `TestSwitchModeDuringWorkload`, after switch + leader/service routing checks and before enabling post-switch counting, add an explicit wait:
  - `testutil.Eventually(... !rgController.IsDegraded() ...)`
- This keeps test intent unchanged (must make post-switch progress), but removes the race where counting starts while controller is still transiently degraded.

Risk

- Low risk; test-only change.
- Slightly longer wait in rare slow environments (bounded to 30s) to avoid false negatives.

Verification

- Focused verification (pass):
  - `cd tests/integrations && make gotest GOTEST_ARGS='-tags without_dashboard ./mcs/resourcemanager -run TestSwitchModeDuringWorkload -count=3'`
  - Result: `ok github.com/tikv/pd/tests/integrations/mcs/resourcemanager 431.983s`
- Baseline verification (has unrelated pre-existing failures in this run):
  - `make basic-test`
  - Result: failed with non-touched-scope flakes/noise in this environment:
    - `pkg/gctuner`: goleak (unexpected goroutine in `memory_limit_tuner`)
    - `pkg/storage/endpoint`: `TestDataPhysicalRepresentation` expected path mismatch (`/pd/0/...` vs `/pd/<cluster-id>/...`)

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved test stability by adding synchronization checks when switching resource manager deployment modes, ensuring proper controller recovery before test continuation.

### DIFF
--- a/tests/integrations/mcs/resourcemanager/resource_manager_test.go
+++ b/tests/integrations/mcs/resourcemanager/resource_manager_test.go
@@ -487,6 +487,11 @@ func TestSwitchModeDuringWorkload(t *testing.T) {
 				waitResourceManagerServiceURL(re, cli, false)
 			}
 			waitLeaderServingClient(re, cli, leader.GetAddr())
+			testutil.Eventually(re, func() bool {
+				// Switching endpoint can transiently leave the controller in degraded mode
+				// until the first successful token-bucket response arrives.
+				return !rgController.IsDegraded()
+			}, testutil.WithWaitFor(30*time.Second), testutil.WithTickInterval(50*time.Millisecond))
 
 			switched.Store(true)
 			testutil.Eventually(re, func() bool {


tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -487,6 +487,11 @@ func TestSwitchModeDuringWorkload(t *testing.T) {
 				waitResourceManagerServiceURL(re, cli, false)
 			}
 			waitLeaderServingClient(re, cli, leader.GetAddr())
+			testutil.Eventually(re, func() bool {
```
rleungx, line 490: This test still needs "restart it", do we need to refactor it directly?

okJiang: How do you want to refactor it? This is an integration test.

tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -487,6 +487,11 @@ func TestSwitchModeDuringWorkload(t *testing.T) {
...
+				// Switching endpoint can transiently leave the controller in degraded mode
+				// until the first successful token-bucket response arrives.
+				return !rgController.IsDegraded()
```
rleungx, line 493: I don't think `!rgController.IsDegraded()` fully proves recovery here. It can still be `false` before the controller has actually seen a successful post-switch token-bucket response. Could we gate this on a stronger post-switch success signal?

PR #10334, opened 2026-03-12
resource_group: add per-keyspace RU version config via ControllerConfig

### LINKED ISSUE #10338: resource_group: add keyspace RU version support
## Description

Add per-keyspace RU version configuration to the resource group manager, allowing the TiDB Cloud management plane to configure different RU calculation versions for each keyspace via the existing service limit HTTP API.

## Tasks

- [x] Add etcd key path `resource_group/keyspace/ru_versions/{keyspace_id}` for RU version storage
- [x] Add `SaveRuVersion`/`LoadRuVersion`/`LoadRuVersions` to `ResourceGroupStorage` interface
- [x] Add `RuVersion` field to `serviceLimiter` with persistence support
- [x] Load RU versions from etcd during `Manager.Init`
- [x] Extend `KeyspaceServiceLimitRequest` with optional `ru_version` field
- [x] Include `ru_version` in GET service limit response via `serviceLimiter.Clone()`
- [x] Add input validation (reject negative `ru_version`)
- [x] Add unit tests

sdojjy:

Issue Number: close #10338

### Problem Summary

Add per-keyspace RU version configuration to the resource group manager, consolidated into the existing `ControllerConfig` stored at the single etcd key `resource_group/controller`. This allows the TiDB Cloud management plane to configure different RU calculation versions for each keyspace, which is needed for the RU v2 migration.

**Design: Merge RU Version Policy into ControllerConfig**

Instead of using separate etcd keys (`resource_group/keyspace/ruconfig/{keyspace_id}`) and a separate watch channel, this PR consolidates everything into the existing `ControllerConfig`. This:
- Eliminates new etcd keys and watches
- Reuses the existing `watchConfigChannel` for propagating RU version changes to clients
- Ensures atomic updates (single etcd key)
- Simplifies both server and client code

**RUVersionPolicy Struct:**
```go
type RUVersionPolicy struct {
    Default   int32            `json:"default"`
    Overrides map[string]int32 `json:"overrides,omitempty"`
}
```
- `Default`: the default RU version for all keyspaces (fallback)
- `Overrides`: per-keyspace overrides keyed by keyspace ID string (JSON only allows string keys)
- Embedded as `*RUVersionPolicy` pointer in `ControllerConfig` with `omitempty` — existing configs without this field unmarshal cleanly (nil = no policy = default v1)

**Server-side (`ControllerConfig`):**
- Added `RUVersionPolicy` struct and `*RUVersionPolicy` field to `ControllerConfig` on both server (`pkg/mcs/resourcemanager/server/config.go`) and client (`client/resource_group/controller/config.go`)
- `SetKeyspaceRuVersion(keyspaceID, ruVersion)` on Manager: initializes policy if nil, sets override (or deletes if ruVersion=0), saves entire `ControllerConfig` to etcd
- `GetRUVersionPolicy()` on Manager: returns current policy
- `UpdateControllerConfigItem` supports `"ru-version-policy"` key

**API Layer:**
- New `POST /config/controller/ru_version/:keyspace_name` endpoint
- Resolves keyspace name → ID, validates `ru_version >= 0`, updates the `ControllerConfig`'s `RUVersionPolicy`
- Removed old separate ruconfig endpoints (`/config/keyspace/ruconfig/`)

**Client-side:**
- `updateRuVersionFromConfig(config)` method extracts ruVersion from `ControllerConfig.RUVersionPolicy`:
  1. If override exists for this keyspace ID → use it
  2. Else if default > 0 → use default
  3. Else → store 0 (GetRuVersion returns 1)
- Called during initial config load in `NewResourceGroupController` and on every `watchConfigChannel` event
- Removed separate `watchRuConfigChannel`, `loadRuVersion()`, `ruConfigJSON` struct
- Removed `RuConfigKeyBytes()` and `keyspaceRuConfigPathPrefix` from client

**Removed Code:**
- Storage layer: `KeyspaceRuConfig` struct, `SaveRuConfig`/`LoadRuConfig`/`LoadRuConfigs` methods
- Keypath: `KeyspaceRuConfigPath()`, `KeyspaceRuConfigPrefix()`, path format constants
- Server: `keyspaceResourceGroupManager.ruConfig` field and methods, old ruconfig API handlers
- Client: separate watch channel setup, retry logic, and handler

### Check List

Tests

- [x] Unit test: `TestRUVersionPolicyJSONSerialization`, `TestRUVersionPolicyInControllerConfig`, `TestManagerSetKeyspaceRuVersion`, `TestManagerSetKeyspaceRuVersionWriteDisabled`, `TestManagerSetKeyspaceRuVersionMultipleKeyspaces`, `TestManagerSetKeyspaceRuVersionResetToDefault`
- [x] Client test: `TestRuVersionFromControllerConfig`, `TestRuVersionOverrideFromControllerConfig`, `TestRuVersionDefaultFallback`, `TestRuVersionNilPolicy`, `TestRuVersionFromInitialControllerConfig`, `TestRuVersionWatchViaControllerConfig`
- [x] Integration test: RU version API endpoint tests in `api_test.go`

Side effects

- [x] No side effects

### Release note

```release-note
Add per-keyspace RU version config support via ControllerConfig's RUVersionPolicy. New API endpoint `POST /config/controller/ru_version/:keyspace_name` sets RU version per keyspace. Consolidated into existing ControllerConfig etcd key, eliminating separate etcd keys and watch channels.
```

### DIFF
--- a/client/resource_group/controller/global_controller.go
+++ b/client/resource_group/controller/global_controller.go
@@ -79,7 +79,7 @@ type ResourceGroupKVInterceptor interface {
 	// IsBackgroundRequest If the resource group has background jobs, we should not record consumption and wait for it.
 	IsBackgroundRequest(ctx context.Context, resourceGroupName, requestResource string) bool
 	// GetRUVersion returns the current RU calculation version for this keyspace.
-	GetRUVersion() int32
+	GetRUVersion() RUVersion
 }
 
 // ResourceGroupProvider provides some api to interact with resource manager server.
@@ -252,7 +252,7 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
 // GetRUVersion returns the current RU calculation version for this keyspace.
 // Returns DefaultRUVersion (v1) if not configured or not loaded yet.
 // This is a pure memory read (atomic load), no network call.
-func (c *ResourceGroupsController) GetRUVersion() int32 {
+func (c *ResourceGroupsController) GetRUVersion() RUVersion {
 	v := c.ruVersion.Load()
 	if v <= 0 {
 		return DefaultRUVersion
--- a/pkg/mcs/resourcemanager/server/ru_version_policy_test.go
+++ b/pkg/mcs/resourcemanager/server/ru_version_policy_test.go
@@ -1,4 +1,4 @@
-// Copyright 2026 TiKV Project Authors.
+// Copyright 2025 TiKV Project Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


pkg/utils/keypath/absolute_key_path.go
```diff
@@ -122,6 +122,9 @@ const (
 	// service limit path
 	keyspaceServiceLimitsPathPrefixFormat = "resource_group/keyspace/service_limits/"   // "resource_group/keyspace/service_limits/"
 	keyspaceServiceLimitsPathFormat       = "resource_group/keyspace/service_limits/%d" // "resource_group/keyspace/service_limits/{keyspace_id}"
+	// ru version path
+	keyspaceRuVersionsPathPrefixFormat = "resource_group/keyspace/ru_versions/"   // "resource_group/keyspace/ru_versions/"
+	keyspaceRuVersionsPathFormat       = "resource_group/keyspace/ru_versions/%d" // "resource_group/keyspace/ru_versions/{keyspace_id}"
```
disksing: maybe better to store together with service_limits ? cc @JmPotato

pkg/mcs/resourcemanager/server/service_limit.go
```diff
@@ -40,6 +40,9 @@ type serviceLimiter struct {
 	AvailableTokens float64 `json:"available_tokens"`
 	// LastUpdate records the last time the limiter was updated.
 	LastUpdate time.Time `json:"last_update"`
+	// RuVersion is the RU calculation version for the keyspace.
+	// 0 means not configured.
+	RuVersion int32 `json:"ru_version"`
 	// KeyspaceID is the keyspace ID of the keyspace that this limiter belongs to.
 	keyspaceID uint32
 	// storage is used to persist the service limit.
```
JmPotato, line 45: `RuVersion` on `serviceLimiter` is a responsibility mismatch — this struct is about rate limiting (token bucket, trickle time, fill rate), while RU version is a calculation strategy selector. Consider whether a separate `KeyspaceConfig` or at minimum a dedicated storage/accessor outside the limiter would be cleaner long-term.

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -385,6 +390,18 @@ func (s *Service) setKeyspaceServiceLimit(c *gin.Context) {
 		c.String(http.StatusInternalServerError, err.Error())
 		return
 	}
```
JmPotato, line 394: Non-atomic dual update: `SetKeyspaceServiceLimit` is called unconditionally above, then `SetKeyspaceRuVersion` is called here conditionally. If `SetKeyspaceRuVersion` fails, the response returns an error but `service_limit` has already been persisted — partial state.

Also, callers who only want to update `ru_version` must still provide the correct current `service_limit`, otherwise it gets overwritten with zero.

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -347,6 +347,7 @@ func (s *Service) setControllerConfig(c *gin.Context) {
 // KeyspaceServiceLimitRequest is the request body for setting the service limit of the keyspace.
 type KeyspaceServiceLimitRequest struct {
 	ServiceLimit float64 `json:"service_limit"`
+	RuVersion    int32   `json:"ru_version,omitempty"`
```
JmPotato: Zero-value ambiguity: `ru_version = 0` with `omitempty` means both "not configured" (storage) and "no update" (request). If you ever need to explicitly reset a keyspace's ru_version to "unconfigured", the API can't express that. A pointer type (`*int32`) or a separate boolean would disambiguate.

client/resource_group/controller/global_controller.go
```diff
@@ -298,6 +347,16 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 		if err != nil {
 			log.Warn("watch resource group config failed", zap.Error(err))
 		}
+
+		// Load initial ru_config and set up watch.
+		c.loadRuVersion(ctx)
+		ruConfigKey := pd.RuConfigKeyBytes(c.keyspaceID)
+		var watchRuConfigChannel chan []*meta_storagepb.Event
+		watchRuConfigChannel, err = c.provider.Watch(ctx, ruConfigKey)
```
JmPotato: No `WithRev(...)`?

client/resource_group/controller/global_controller.go
```diff
@@ -440,6 +506,36 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
...
+				}
+				for _, item := range resp {
+					if item.Type == meta_storagepb.Event_PUT {
+						var config ruConfigJSON
+						if err := json.Unmarshal(item.Kv.Value, &config); err != nil {
+							log.Warn("[resource group controller] failed to parse ru_config from watch event",
+								zap.String("value", string(item.Kv.Value)), zap.Error(err))
+							continue
+						}
+						if config.RuVersion != nil {
```
JmPotato: If it's nil, should we reset it?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -307,6 +338,8 @@ func (m *Manager) UpdateControllerConfigItem(key string, value any) error {
 	switch kp[0] {
 	case "request-unit":
 		config = &m.controllerConfig.RequestUnit
+	case "ru-version-policy":
+		config = &m.controllerConfig.RUVersionPolicy
```
JmPotato: Minor: This allows the generic `POST /config/controller` endpoint to overwrite the entire `RUVersionPolicy` with arbitrary JSON, bypassing the validation logic in `SetKeyspaceRuVersion` (write role check, negative version check, etc.).

Also note that `SaveControllerConfig` after `UpdateControllerConfigItem` is called **outside** the lock (pre-existing pattern), while `SetKeyspaceRuVersion` persists **inside** the lock. Concurrent calls through these two paths could race.

Not blocking — this is a pre-existing design pattern in the codebase — but worth noting for future consideration.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -165,6 +166,36 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
...
+		m.controllerConfig.RUVersionPolicy.Overrides[keyspaceIDStr] = ruVersion
+	}
+	return m.storage.SaveControllerConfig(m.controllerConfig)
+}
+
+// GetRUVersionPolicy returns the current RU version policy from the controller config.
+func (m *Manager) GetRUVersionPolicy() *RUVersionPolicy {
+	m.RLock()
+	defer m.RUnlock()
+	return m.controllerConfig.RUVersionPolicy
```
JmPotato: Nit: `GetRUVersionPolicy` returns the internal pointer directly. Callers outside the `RLock` scope can mutate the `Overrides` map concurrently with server-side writes (e.g. `SetKeyspaceRuVersion`), which would be a data race.

Currently only tests call this, so no practical concern. But if future API handlers iterate over the result, consider returning a deep copy.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -165,6 +166,36 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
...
 
+// SetKeyspaceRuVersion sets the RU version for a specific keyspace in the controller config.
+func (m *Manager) SetKeyspaceRuVersion(keyspaceID uint32, ruVersion int32) error {
+	if !m.writeRole.AllowsMetadataWrite() {
+		return errMetadataWriteDisabled
+	}
+	m.Lock()
+	defer m.Unlock()
+	if m.controllerConfig.RUVersionPolicy == nil {
+		m.controllerConfig.RUVersionPolicy = &RUVersionPolicy{Default: 1}
```
JmPotato: Observation: `Default` is hardcoded to `1` here and there is no API to modify the global default — only per-keyspace overrides can be set via `SetKeyspaceRuVersion`. The only way to change `Default` is through the generic `POST /config/controller` endpoint.

If this is intentional (default should always be v1 for backward compatibility), it might be worth adding a comment to make that explicit. If a global migration to v2 is anticipated in the future, a dedicated API endpoint may be needed.

pkg/mcs/resourcemanager/server/config.go
```diff
@@ -93,6 +93,12 @@ type Config struct {
 	Metering config.MeteringConfig `toml:"metering" json:"metering"`
 }
 
+// RUVersionPolicy configures which RU calculation version to use per keyspace.
```
JmPotato, line 109: Nit: `RUVersionPolicy` is defined identically in both `client/resource_group/controller/config.go` and here. This follows the existing codebase pattern (e.g. `ControllerConfig` is also duplicated), but creates a maintenance risk if the struct evolves differently on either side. No action needed for this PR.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -192,6 +166,40 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
 	return nil
 }
 
+// SetKeyspaceRuVersion sets the RU version for a specific keyspace in the controller config.
+func (m *Manager) SetKeyspaceRuVersion(keyspaceID uint32, ruVersion int32) error {
```
rleungx: What if we set ruVersion to 0?

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -307,6 +342,8 @@ func (m *Manager) UpdateControllerConfigItem(key string, value any) error {
 	switch kp[0] {
 	case "request-unit":
 		config = &m.controllerConfig.RequestUnit
+	case "ru-version-policy":
+		config = &m.controllerConfig.RUVersionPolicy
```
rleungx: The new `ru-version-policy` special case looks structurally incompatible with `jsonutil.AddKeyValue`. That helper always wraps updates as `{key: value}` before merging, so here it will try to merge `{"ru-version-policy": ...}` into `&m.controllerConfig.RUVersionPolicy`. That shape matches `ControllerConfig`, but not `*RUVersionPolicy`, whose JSON shape is just `{"default": ..., "overrides": ...}`. In other words, this helper expects `root object + child key`, but this branch switches the root object to the child itself while still passing the same outer key. I think this should either keep merging against `m.controllerConfig` directly, or handle `RUVersionPolicy` with a dedicated assignment path instead of `AddKeyValue`.

pkg/mcs/resourcemanager/server/manager.go
sdojjy: In the PD client, 0 will be treated as 1

pkg/mcs/resourcemanager/server/manager.go
sdojjy: I will remove it

pkg/mcs/resourcemanager/server/manager.go
rleungx: If the default version later becomes 2 after setting it to 0, will it be 1?

sdojjy: Yes, it's hard-coded here
```
func (c *ResourceGroupsController) GetRuVersion() int32 {
	v := c.ruVersion.Load()
	if v <= 0 {
		return 1
	}
	return v
}
```

client/resource_group/controller/global_controller.go
```diff
@@ -241,6 +250,46 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
 	return c.safeRuConfig.Load()
 }
 
+// GetRuVersion returns the current RU calculation version for this keyspace.
+// Returns 1 (default v1) if not configured or not loaded yet.
+// This is a pure memory read (atomic load), no network call.
+func (c *ResourceGroupsController) GetRuVersion() int32 {
+	v := c.ruVersion.Load()
+	if v <= 0 {
+		return 1
```
rleungx: Another suggestion: how about not hardcoding the default RU version?

client/resource_group/controller/global_controller.go
```diff
@@ -241,6 +250,46 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
...
+}
+
+// updateRuVersionFromConfig extracts the RU version for this keyspace from the controller config.
+func (c *ResourceGroupsController) updateRuVersionFromConfig(config *Config) {
+	if config.RUVersionPolicy != nil {
+		keyspaceIDStr := strconv.FormatUint(uint64(c.keyspaceID), 10)
+		if v, ok := config.RUVersionPolicy.Overrides[keyspaceIDStr]; ok {
+			old := c.ruVersion.Swap(v)
+			if old != v {
+				log.Info("ru_version updated from controller config",
```
rleungx:
```suggestion
				log.Info("ru version updated from controller config",
```

client/resource_group/controller/global_controller.go
```diff
@@ -241,6 +250,46 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
...
+					zap.Int32("old", old), zap.Int32("new", v),
+					zap.Uint32("keyspace-id", c.keyspaceID))
+			}
+		} else if config.RUVersionPolicy.Default > 0 {
+			old := c.ruVersion.Swap(config.RUVersionPolicy.Default)
+			if old != config.RUVersionPolicy.Default {
+				log.Info("ru_version updated to default from controller config",
```
rleungx:
```suggestion
				log.Info("ru version updated to default from controller config",
```

pkg/mcs/resourcemanager/server/ru_version_policy_test.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2025 TiKV Project Authors.
```
rleungx: 2026

client/resource_group/controller/config.go
```diff
@@ -102,6 +102,15 @@ type BaseConfig struct {
...
 	// EnableControllerTraceLog is to control whether resource control client enable trace.
 	EnableControllerTraceLog bool `toml:"enable-controller-trace-log" json:"enable-controller-trace-log,string"`
+
+	// RUVersionPolicy configures which RU calculation version to use per keyspace.
+	RUVersionPolicy *RUVersionPolicy `toml:"ru-version-policy" json:"ru-version-policy,omitempty"`
+}
+
+// RUVersionPolicy configures RU calculation version per keyspace.
+type RUVersionPolicy struct {
+	Default   int32            `json:"default"`
```
JmPotato: Better define a specific enum type to describe the version policy.

client/resource_group/controller/config.go
```diff
@@ -102,6 +102,15 @@ type BaseConfig struct {
...
+	Overrides map[string]int32 `json:"overrides,omitempty"`
```
JmPotato: Why using string keysapce ID rather than uint32?

client/resource_group/controller/global_controller.go
JmPotato: So what's the default value of RU version? 0 or 1?

client/resource_group/controller/global_controller.go
```diff
@@ -298,6 +347,7 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 		if err != nil {
 			log.Warn("watch resource group config failed", zap.Error(err))
 		}
+
```
JmPotato: Why add an empty line?

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -130,6 +132,8 @@ func (s *Service) RegisterRouter() {
 	// With keyspace name, it will get/set the service limit of the given keyspace.
 	configEndpoint.POST("/keyspace/service-limit/:keyspace_name", s.setKeyspaceServiceLimit)
 	configEndpoint.GET("/keyspace/service-limit/:keyspace_name", s.getKeyspaceServiceLimit)
+	// RU version endpoint — sets per-keyspace RU version in controller config.
+	configEndpoint.POST("/controller/ru_version/:keyspace_name", s.setKeyspaceRuVersion)
```
JmPotato:
```suggestion
	configEndpoint.POST("/controller/ru-version/:keyspace_name", s.setKeyspaceRuVersion)
```

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -281,7 +405,58 @@ func (s *Service) setKeyspaceServiceLimit(c *gin.Context) {
...
+//
+//	@Tags		ResourceManager
+//	@Summary	Set the RU version of the keyspace.
+//	@Param		keyspace_name	path	string	true	"Keyspace name"
+//	@Param		config			body	object	true	"json params, SetKeyspaceRuVersionRequest"
+//	@Success	200				{string}	string	"Success!"
+//	@Failure	400				{string}	error
+//	@Failure	403				{string}	error
+//	@Failure	500				{string}	error
+//	@Router		/config/controller/ru_version/{keyspace_name} [post]
```
JmPotato: Ditto.

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -192,6 +166,40 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
...
+		m.controllerConfig.RUVersionPolicy.Overrides = make(map[string]int32)
+	}
+	keyspaceIDStr := strconv.FormatUint(uint64(keyspaceID), 10)
+	defaultVersion := m.controllerConfig.RUVersionPolicy.Default
+	if ruVersion == defaultVersion {
+		delete(m.controllerConfig.RUVersionPolicy.Overrides, keyspaceIDStr)
+	} else {
+		m.controllerConfig.RUVersionPolicy.Overrides[keyspaceIDStr] = ruVersion
+	}
+	return m.storage.SaveControllerConfig(m.controllerConfig)
```
JmPotato, line 190: Better not saving the etcd storage while holding the write lock.

pkg/mcs/resourcemanager/server/manager.go
YuhaoZhang00, line 346: Better add validation of json format and values here for RUVersionPolicy

client/resource_group/controller/global_controller.go
YuhaoZhang00: +1. If hardcode cannot be avoided, should make it a global variable instead of magic number

sdojjy: fixed

client/resource_group/controller/config.go
sdojjy: fixed

client/resource_group/controller/global_controller.go
```diff
@@ -78,6 +78,8 @@ type ResourceGroupKVInterceptor interface {
 	OnResponseWait(ctx context.Context, resourceGroupName string, req RequestInfo, resp ResponseInfo) (*rmpb.Consumption, time.Duration, error)
 	// IsBackgroundRequest If the resource group has background jobs, we should not record consumption and wait for it.
 	IsBackgroundRequest(ctx context.Context, resourceGroupName, requestResource string) bool
+	// GetRuVersion returns the current RU calculation version for this keyspace.
+	GetRuVersion() int32
```
JmPotato: Use `GetRUVersion()`.

pkg/mcs/resourcemanager/server/config.go
```diff
@@ -93,6 +93,27 @@ type Config struct {
...
+type RUVersionPolicy struct {
+	Default   int32            `json:"default"`
```
JmPotato: Please keep consistent with the client enum definition.

pkg/mcs/resourcemanager/server/manager.go
JmPotato: +1

client/resource_group/controller/global_controller.go
sdojjy: fixed

pkg/mcs/resourcemanager/server/config.go
sdojjy: fixed

pkg/mcs/resourcemanager/server/manager.go
sdojjy: fixed

client/resource_group/controller/global_controller.go
```diff
@@ -241,6 +249,45 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
 	return c.safeRuConfig.Load()
 }
 
+// GetRUVersion returns the current RU calculation version for this keyspace.
+// Returns DefaultRUVersion (v1) if not configured or not loaded yet.
```
YuhaoZhang00, line 253:
```suggestion
// Returns DefaultRUVersion if not configured or not loaded yet.
```

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -192,6 +165,39 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
...
 }
 
+// SetKeyspaceRuVersion sets the RU version for a specific keyspace in the controller config.
+func (m *Manager) SetKeyspaceRuVersion(keyspaceID uint32, ruVersion int32) error {
+	if !m.writeRole.AllowsMetadataWrite() {
+		return errMetadataWriteDisabled
+	}
+	m.Lock()
+	if m.controllerConfig.RUVersionPolicy == nil {
+		// DefaultRUVersion (v1) means no RU model change.
```
YuhaoZhang00, line 175:
```suggestion
		// DefaultRUVersion means no RU model change.
```

client/resource_group/controller/global_controller.go
```diff
@@ -438,6 +485,8 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 					if enableControllerTraceLog.Load() != config.EnableControllerTraceLog {
 						enableControllerTraceLog.Store(config.EnableControllerTraceLog)
 					}
+					// Update ru version from the controller config RUVersionPolicy.
+					c.updateRuVersionFromConfig(config)
```
YuhaoZhang00:
```suggestion
					c.updateRUVersionFromConfig(config)
```

client/resource_group/controller/global_controller.go
```diff
@@ -241,6 +249,45 @@ func (c *ResourceGroupsController) GetConfig() *RUConfig {
...
+func (c *ResourceGroupsController) GetRUVersion() int32 {
+	v := c.ruVersion.Load()
+	if v <= 0 {
+		return DefaultRUVersion
+	}
+	return v
+}
+
+// updateRuVersionFromConfig extracts the RU version for this keyspace from the controller config.
+func (c *ResourceGroupsController) updateRuVersionFromConfig(config *Config) {
```
YuhaoZhang00:
```suggestion
func (c *ResourceGroupsController) updateRUVersionFromConfig(config *Config) {
```

client/resource_group/controller/global_controller.go
```diff
@@ -214,6 +220,8 @@ func NewResourceGroupController(
 	controller.calculators = []ResourceCalculator{newKVCalculator(controller.ruConfig), newSQLCalculator(controller.ruConfig)}
 	controller.safeRuConfig.Store(controller.ruConfig)
 	enableControllerTraceLog.Store(config.EnableControllerTraceLog)
+	// Extract initial ruVersion from the controller config's RUVersionPolicy.
+	controller.updateRuVersionFromConfig(config)
```
YuhaoZhang00:
```suggestion
	controller.updateRUVersionFromConfig(config)
```

client/resource_group/controller/global_controller.go
YuhaoZhang00: ditto

client/resource_group/controller/global_controller.go
YuhaoZhang00: ditto

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -249,11 +327,34 @@ func (s *Service) getControllerConfig(c *gin.Context) {
...
 }
 
 // KeyspaceServiceLimitRequest is the request body for setting the service limit of the keyspace.
-type KeyspaceServiceLimitRequest = metadataapi.KeyspaceServiceLimitRequest
+type KeyspaceServiceLimitRequest struct {
+	ServiceLimit float64 `json:"service_limit"`
+}
+
+// SetKeyspaceRuVersionRequest is the request body for setting the RU version of a keyspace.
+type SetKeyspaceRuVersionRequest struct {
```
YuhaoZhang00:
```suggestion
type SetKeyspaceRUVersionRequest struct {
```

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -249,11 +327,34 @@ func (s *Service) getControllerConfig(c *gin.Context) {
...
+	RuVersion int32 `json:"ru_version"`
```
YuhaoZhang00: RU

pkg/mcs/resourcemanager/server/apis/v1/api.go
YuhaoZhang00: ditto

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -130,6 +132,8 @@ func (s *Service) RegisterRouter() {
 	// With keyspace name, it will get/set the service limit of the given keyspace.
 	configEndpoint.POST("/keyspace/service-limit/:keyspace_name", s.setKeyspaceServiceLimit)
 	configEndpoint.GET("/keyspace/service-limit/:keyspace_name", s.getKeyspaceServiceLimit)
+	// RU version endpoint — sets per-keyspace RU version in controller config.
+	configEndpoint.POST("/controller/ru-version/:keyspace_name", s.setKeyspaceRuVersion)
```
YuhaoZhang00:
```suggestion
	configEndpoint.POST("/controller/ru-version/:keyspace_name", s.setKeyspaceRUVersion)
```

pkg/mcs/resourcemanager/server/apis/v1/api.go
```diff
@@ -281,7 +405,58 @@ func (s *Service) setKeyspaceServiceLimit(c *gin.Context) {
...
+//	@Tags		ResourceManager
+//	@Summary	Set the RU version of the keyspace.
+//	@Param		keyspace_name	path	string	true	"Keyspace name"
+//	@Param		config			body	object	true	"json params, SetKeyspaceRuVersionRequest"
+//	@Success	200				{string}	string	"Success!"
+//	@Failure	400				{string}	error
+//	@Failure	403				{string}	error
+//	@Failure	500				{string}	error
+//	@Router		/config/controller/ru-version/{keyspace_name} [post]
+func (s *Service) setKeyspaceRuVersion(c *gin.Context) {
```
YuhaoZhang00: ditto

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -192,6 +165,39 @@ func (m *Manager) SetKeyspaceServiceLimit(keyspaceID uint32, serviceLimit float6
 	return nil
 }
 
+// SetKeyspaceRuVersion sets the RU version for a specific keyspace in the controller config.
+func (m *Manager) SetKeyspaceRuVersion(keyspaceID uint32, ruVersion int32) error {
```
YuhaoZhang00: ditto

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -337,6 +340,13 @@ func (m *Manager) UpdateControllerConfigItem(key string, value any) error {
...
 		config = &m.controllerConfig.RequestUnit
+	case "ru-version-policy":
+		// Validate before applying: parse the value as RUVersionPolicy and check it.
+		if err := validateRUVersionPolicyValue(value); err != nil {
+			m.Unlock()
+			return err
+		}
+		config = m.controllerConfig
 	default:
 		config = m.controllerConfig
```
YuhaoZhang00, line 346: If the default path (full controller config update) is also exposed to outside PD, same validation logic of RUVersionPolicy should also be applied here.

sdojjy: fixed

pkg/mcs/resourcemanager/server/ru_version_policy_test.go
```diff
@@ -0,0 +1,281 @@
...
```
YuhaoZhang00: 2026

client/resource_group/controller/global_controller.go
```diff
@@ -78,6 +78,8 @@ type ResourceGroupKVInterceptor interface {
 	OnResponseWait(ctx context.Context, resourceGroupName string, req RequestInfo, resp ResponseInfo) (*rmpb.Consumption, time.Duration, error)
 	// IsBackgroundRequest If the resource group has background jobs, we should not record consumption and wait for it.
 	IsBackgroundRequest(ctx context.Context, resourceGroupName, requestResource string) bool
+	// GetRUVersion returns the current RU calculation version for this keyspace.
+	GetRUVersion() int32
```
JmPotato: Can we return `RUVersion` here?

sdojjy: fixed

JmPotato (approved):

PR #10349, opened 2026-03-13
cluster: fix TiFlash store removal being blocked by TiKV replica check

### LINKED ISSUE #10350: TiFlash store removal incorrectly blocked by TiKV replica check
### What did you do?

Attempted to remove a TiFlash node from the cluster (scale-in operation):

1. Cluster setup: 3 TiKV nodes (max-replicas=3) + 4 TiFlash nodes
2. Delete TiFlash store using pd-ctl:
   ```bash
   pd-ctl -u http://127.0.0.1:2379 store delete 1540
   ```
3. Or remove TiFlash node using TiUP scale-in

Expected to successfully remove the TiFlash node because:
- TiFlash uses learner nodes for columnar storage
- TiFlash does not participate in Raft voting
- TiFlash removal should not affect Raft replica count check
- Cluster has 3 TiKV nodes, which satisfies max-replicas=3 requirement

### What did you see instead?

PD rejected the removal operation with error:

```
Failed to delete store 1540: [400] "[PD:core:ErrStoresNotEnough]can not remove store 1540 since the number of up stores would be 2 while need 3"
```

The TiFlash node removal failed even though the cluster has 3 TiKV nodes (satisfying max-replicas=3).

**Root Cause**:

The `checkReplicaBeforeOfflineStore` function at `server/cluster/cluster.go:1569`:
- Uses `getUpStores()` which only returns **TiKV stores**
- Incorrectly assumes ANY store deletion reduces TiKV count by 1
- Applies TiKV replica count validation logic to TiFlash stores

`v9.0.0-beta.2.pre-315-gb2ac11253` (master branch)

**Environment**:
- TCMS test execution: 8056627
- Related issue: [pingcap/tiup#2449](https://github.com/pingcap/tiup/issues/2449)

liyishuai:

Issue Number: close #10350

```commit-message
When attempting to remove a TiFlash store, PD incorrectly applied
TiKV replica count validation, causing removal failures with
"ErrStoresNotEnough" even when TiKV stores were sufficient.

Root cause:
- checkReplicaBeforeOfflineStore() used getUpStores() which only
  returns TiKV stores
- The function assumed ANY store deletion reduces TiKV count by 1
- This is incorrect for non-TiKV stores which don't affect Raft replicas

Fix:
- Add store type check at the beginning of checkReplicaBeforeOfflineStore()
- Skip TiKV replica validation for non-TiKV stores using !store.IsTiKV()
- Only TiKV stores participate in Raft voting; non-TiKV stores (TiFlash, etc.)
  use learners and can be removed without affecting Raft replica count
```

### Check List

Tests

- Unit test (TestRemoveTiFlashStore)

Code changes

- Modified server/cluster/cluster.go to add non-TiKV store check in checkReplicaBeforeOfflineStore()

Side effects

- No

Related changes

- No

### Release note

```release-note
Fix TiFlash store removal being incorrectly blocked by TiKV replica check
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added null-checks for store references to prevent errors during offline operations.
  * Replica validation now applies only to TiKV stores; other store types can bypass TiKV-specific replica checks during removal.

* **Tests**
  * Added test coverage validating removal of TiFlash stores while ensuring TiKV replication requirements and proper state transitions.

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -1538,7 +1538,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 		return errs.ErrStoreDestroyed.FastGenByArgs(storeID)
 	}
 	if (store.IsPreparing() || store.IsServing()) && !physicallyDestroyed {
-		if err := c.checkReplicaBeforeOfflineStore(storeID); err != nil {
+		if err := c.checkTikvReplicaBeforeOfflineStore(storeID); err != nil {
 			return err
 		}
 	}
@@ -1566,7 +1566,7 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 	return nil
 }
 
-func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
+func (c *RaftCluster) checkTikvReplicaBeforeOfflineStore(storeID uint64) error {
 	store := c.GetStore(storeID)
 	if store == nil {
 		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
@@ -1579,7 +1579,7 @@ func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
 		return nil
 	}
 
-	upStores := c.getUpStores()
+	upStores := c.getUpTikvStores()
 	expectUpStoresNum := len(upStores) - 1
 	if expectUpStoresNum < c.opt.GetMaxReplicas() {
 		return errs.ErrStoresNotEnough.FastGenByArgs(storeID, expectUpStoresNum, c.opt.GetMaxReplicas())
@@ -1613,8 +1613,8 @@ func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
 	return nil
 }
 
-// getUpStores gets all up TiKV stores.
-func (c *RaftCluster) getUpStores() []uint64 {
+// getUpTikvStores gets all up TiKV stores.
+func (c *RaftCluster) getUpTikvStores() []uint64 {
 	upStores := make([]uint64, 0)
 	for _, store := range c.GetStores() {
 		if !store.IsTiKV() {
--- a/server/cluster/cluster_test.go
+++ b/server/cluster/cluster_test.go
@@ -1584,7 +1584,7 @@ func TestStoreConfigUpdate(t *testing.T) {
 	for _, s := range stores {
 		re.NoError(tc.setStore(s))
 	}
-	re.Len(tc.getUpStores(), 5)
+	re.Len(tc.getUpTikvStores(), 5)
 	// Case1: big region.
 	{
 		body := `{ "coprocessor": {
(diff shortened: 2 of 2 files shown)


liyishuai:
@coderabbitai help

liyishuai:
@coderabbitai full review

liyishuai:
@coderabbitai full review

server/cluster/cluster.go
```diff
@@ -1567,6 +1567,17 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
 
 func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
+	store := c.GetStore(storeID)
+	if store == nil {
+		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
+	}
+
+	// TiFlash stores don't affect Raft replica count, skip the check.
+	// TiFlash stores are for columnar storage and use learners, not Raft voters.
+	if store.IsTiFlashWrite() || store.IsTiFlashCompute() {
```
rleungx: Shall we use isTiKV?

liyishuai: Fixed.

server/cluster/cluster.go
```diff
@@ -1567,6 +1567,18 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
+	store := c.GetStore(storeID)
+	if store == nil {
+		return errs.ErrStoreNotFound.FastGenByArgs(storeID)
+	}
+
+	// Only TiKV stores participate in Raft voting and affect replica count.
+	// Non-TiKV stores (e.g., TiFlash) use learners and can be removed without
+	// affecting Raft replica validation.
+	if !store.IsTiKV() {
+		return nil
```
AndreMouche, line 1579: Should we consider the number of learners here? For example, if we create a table that need at least 3 learners, and we only have 3 TiFlash nodes, when we try to offline one TiFlash node , should we return an ErrTiflashsNotEnough error?

server/cluster/cluster.go
```diff
@@ -1567,6 +1567,18 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
...
+	}
+
 	upStores := c.getUpStores()
```
bufferflies: getUpStores should rename getUpTikvStores?

server/cluster/cluster.go
```diff
@@ -1567,6 +1567,18 @@ func (c *RaftCluster) RemoveStore(storeID uint64, physicallyDestroyed bool) erro
 }
 
 func (c *RaftCluster) checkReplicaBeforeOfflineStore(storeID uint64) error {
```
bufferflies: ditto, the function should tell the caller that it only carries about the tikv engine?

server/cluster/cluster.go
liyishuai: @bufferflies any ideas here?

server/cluster/cluster.go
liyishuai: Both renamed in 3e57fa7cf8d17f4b9671f6eb54dfee95c3d566fe

server/cluster/cluster.go
bufferflies: We can't consider all cases when using placement rules; it's too difficult. If not match the rules after scale-in this node, it will be stuck and wait for the user to scale out another node.

rleungx (approved):

Issue #10206, opened 2026-02-03
Make PD Great Again—with AI


JmPotato:
## Enhancement Task

This issue is used to link various sub-issues and PRs that help AI participate in, assist with, or even fully take over our development process.

- [x] https://github.com/tikv/pd/pull/10147
- [ ] https://github.com/tikv/pd/pull/10207
- [ ] https://github.com/tikv/pd/issues/10159

PR #10431, opened 2026-03-16
agents: add create-issue skill

JmPotato:

Issue Number: ref #10206

PD already has a `create-pr` skill, but issue creation is still ad hoc. That makes duplicate search, template selection, and pre-submit confirmation easy to miss when filing routine follow-up work.

Add a `create-issue` skill under `.agents/skills/create-issue/` that automates GitHub issue creation on tikv/pd.

The skill covers:
- Verifying repo checkout and `gh` auth
- Classifying the issue into the correct template (`bug-report`, `flaky-test`, `development-task`, `enhancement-task`, `feature-request`)
- Searching for duplicates before drafting
- Reading the template's headings and `labels:` frontmatter
- Drafting the title and body from concrete evidence
- Requiring user approval before submission
- Creating the issue via `gh issue create --label --title --body-file` (not `--template`, which is incompatible with both `--body-file` and `--label` in the current `gh` CLI)

Also registers `create-issue` in the `AGENTS.md` skill table.

```commit-message
Add a create-issue skill under .agents/skills/ for tikv/pd.
Classifies issue types, searches for duplicates, reads template
headings and labels from frontmatter, drafts issue content, and
creates via gh issue create with explicit --label flags after
user approval. Syncs AGENTS.md skill index.
```

### Check List

Tests

- No code

### Release note

```release-note
None.
```

### DIFF
--- a/.agents/skills/create-issue/SKILL.md
+++ b/.agents/skills/create-issue/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: create-issue
-description: Draft and create GitHub issues on tikv/pd using the repository's issue templates. Searches for duplicates, picks the right template, drafts the title/body from the current problem statement, shows the draft to the user for approval, and submits via `gh issue create`.
+description: Use when a problem, task, or idea needs to be tracked as a GitHub issue on tikv/pd. Searches for duplicates, picks the right issue template, drafts the title and body, shows the draft for user approval, and submits via `gh issue create`.
 compatibility: Requires `gh` CLI authenticated with tikv/pd repo access. Works best from a local tikv/pd checkout so issue templates can be read directly.
 ---
 


.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+   - `bug-report` -> `Bug Report`
+   - `flaky-test` -> `Flaky Test`
+   - `development-task` -> `Development Task`
+   - `enhancement-task` -> `Enhancement Task`
+   - `feature-request` -> `Feature Request`
+
+2. Create the issue with `gh`. Preserve formatting with a temporary file or heredoc:
+   ```bash
+   gh issue create --repo tikv/pd --template "<template-name>" --title "<title>" --body-file <body-file>
+   ```
```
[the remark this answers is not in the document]
JmPotato: Addressed in `28904844f`.

- `--body-file` is now quoted in the example command.
- The skill now reads the selected issue template frontmatter and uses its `name:` for `gh issue create --template`, so we no longer keep hard-coded template/label mappings in the workflow.
- The repo skill index in `AGENTS.md` is synced, and the extra `openai.yaml` metadata file has been removed.

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+### Phase 3: Create the Issue
+
+After user approval:
+
+1. Use the selected template's frontmatter `name:` field as the `gh --template` argument so GitHub can apply the template's own defaults.
+   If the template metadata is missing or ambiguous, stop and ask the user before creation.
+
+2. Create the issue with `gh`. Preserve formatting with a temporary file or heredoc:
+   ```bash
+   gh issue create --repo tikv/pd --template "<template-name>" --title "<title>" --body-file "<body-file>"
```
rleungx: Will it add the default label?

JmPotato: Good catch — it won't. `gh issue create --template` only uses the template for body text in non-interactive mode; it does **not** apply the template's frontmatter labels (this is a known `gh` CLI limitation: [cli/cli#3633](https://github.com/cli/cli/issues/3633)).

Fixed in `a57f1581b`: the skill now extracts `labels:` from the template frontmatter and passes them explicitly via `--label` flags.

JmPotato: Deeper investigation revealed `--template` is fundamentally broken for non-interactive use:

1. `--template` + `--body-file` → explicitly rejected by `gh` CLI
2. `--template` + `--label` → GraphQL API error ([cli/cli#5017](https://github.com/cli/cli/issues/5017))

Fixed in `1f08fa8e6`: dropped `--template` entirely. The agent already reads the template file and drafts the body from its headings, so `--template` was redundant. The command now uses `--body-file` + `--label` + `--title` only, with labels extracted from the template's frontmatter.

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
+---
+name: create-issue
+description: Draft and create GitHub issues on tikv/pd using the repository's issue templates. Searches for duplicates, picks the right template, drafts the title/body from the current problem statement, shows the draft to the user for approval, and submits via `gh issue create`.
```
okJiang: description should explain when to call this skill

AGENTS.md
okJiang, line 140: Just a question. Not sure if skills need to be specified in AGENTS.md?🤔

JmPotato: Given that the definition paths, filenames, and standards for skills are not currently universal across different coding agents, I think it is reasonable to mention the locations of these skills in a frequently read file.

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+## Common Pitfalls
+
+- Do not skip the duplicate search just because the issue seems routine.
+- Do not mix templates: the body headings and `--label` values must come from the same template.
+- Do not use `--template` with `gh issue create` — it conflicts with both `--body-file` and `--label` in the current `gh` CLI.
+- Do not file flaky-test issues without the failing job name and a concrete CI link.
+
+## Agent Constraints
+
+- Always search for duplicates first.
```
bufferflies, line 94: How does it search include all issues?

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+3. Classify the issue into exactly one template:
+   - `bug-report`: broken behavior, regression, or correctness issue
+   - `flaky-test`: unstable test or CI job
+   - `development-task`: routine engineering task, maintenance item, or scoped follow-up
+   - `enhancement-task`: engineering improvement with a fairly clear implementation direction
+   - `feature-request`: user-facing/API-facing new capability request
+
+4. Search for duplicates before drafting. Start narrow, then broaden:
+   ```bash
+   gh issue list --repo tikv/pd --state open --search "<keywords>" --limit 10
```
bufferflies, line 42: What are the keywords, is it generated by ai summary?

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+   - development or enhancement: imperative action or scoped task summary
+   - feature request: concise statement of the requested capability
+
+3. Fill every section of the template from concrete evidence. Use `N/A` only when a section truly does not apply. Do not invent repro steps, versions, CI links, or impact claims.
+
+4. If the issue comes from code review or a PR follow-up, link the triggering PR, commit, or discussion in the body.
+
+5. Show the drafted title and full body to the user before creating the issue. Ask for corrections at this stage.
+
+### Phase 3: Create the Issue
```
bufferflies, line 67: Do we need to add a label for bug issues, such as affect and severity

.agents/skills/create-issue/SKILL.md
JmPotato: It will use the previously mentioned gh command to perform the search.

.agents/skills/create-issue/SKILL.md
JmPotato: Agent will generate keywords based on the specific context of the issue; I think we don’t need to specify anything for it here.

.agents/skills/create-issue/SKILL.md
JmPotato: This skill is only for creating the action itself. For the specific triage work, I think we still need manual labor to complete it, or perhaps we can add a new skill in the future.

Issue #10329, opened 2026-03-11
tidb abnormally restart with panic when change pd ms mode


Lily2025:
### What did you do?
1、change pd ms mode

### What did you expect to see?
tidb are normal

### What did you see instead?
tidb abnormally restart with panic
[2026/03/11 07:18:15.190 +08:00] [PANIC] [dispatcher.go:519] ["[tso] timestamp fallback"] [keyspace=4294967295] [last-ts="(1773184695175, 4)"] [cur-ts="(1773184695149, 1)"] [last-tso-server=tc-pd-0.tc-pd-peer.pdms-oltp-tpcc-ops-tps-8081031-1-371.svc.cluster.local:2379] [cur-tso-server=tc-pd-2.tc-pd-peer.pdms-oltp-tpcc-ops-tps-8081031-1-371.svc.cluster.local:2379] [last-keyspace-group-in-request=0] [cur-keyspace-group-in-request=0] [last-keyspace-group-in-response=0] [cur-keyspace-group-in-response=0] [last-response-received-at=2026/03/11 07:18:15.177 +08:00] [cur-response-received-at=2026/03/11 07:18:15.190 +08:00] [last-stream-id=tc-pd-0.tc-pd-peer.pdms-oltp-tpcc-ops-tps-8081031-1-371.svc.cluster.local:2379-12] [cur-stream-id=tc-pd-2.tc-pd-peer.pdms-oltp-tpcc-ops-tps-8081031-1-371.svc.cluster.local:2379-14]
[...]

PR #10439, opened 2026-03-17
tso, server: add debug logs for TSO sync, closure, and forwarding paths

YuhaoZhang00:

Issue Number: ref #10329

When investigating TSO fallback and dual-writer incidents such as #10329, the existing logs and
  metrics are not enough to reconstruct:
  1. which persisted TSO window was written to etcd,
  2. when PD/embedded TSO state changed during switching, and
  3. how many TSO streams were forwarded by PD to the independent TSO service.

This PR improves observability for TSO persistence and forwarding paths.

  1. Add persistence-focused logs for successful TSO window writes to etcd:
     - `syncTimestamp`
     - `updateTimestamp`
     - `resetUserTimestamp`

  2. Improve TSO switching related logs:
     - clarify allocator reset warning when `UpdateTSO` fails and the allocator resigns leadership
     - rename the shutdown log to `closing the embedded TSO allocator`
     - replace the noisy per-check skip log with transition-based logs when TSO dynamic switching
  is disabled or re-enabled

  3. Replace the forwarded TSO stream debug log with a Prometheus metric:
     - add `pd_server_tso_forward_stream_total`
     - increment it when PD forwards a TSO stream to the independent TSO service

  4. Add a Grafana panel for the new metric in the PD dashboard:
     - `Forward TSO streams count`

All new logs on hot paths use debug level to avoid overwhelming logs under burst requests.

### Check List

Tests

- Manual test
  - Grafana panel working with new metric locally

fwd-tso-streams-count

Code changes

N/A

Side effects

- Increased observability/log volume in TSO persistence paths

Related changes

N/A

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Chores**
  * Enhanced TSO logging: clearer allocator-reset notice on leadership resignation and new success logs when TSO windows are persisted (including reason, member and time details).
  * Improved dynamic TSO mode visibility with explicit resume/skip messages when mode state changes.
* **New Features**
  * Added a Prometheus counter for forwarded TSO streams and incremented it when forwarding occurs.
  * Updated Grafana dashboard: resized panels and added a “Forward TSO streams count” graph.

### DIFF
--- a/pkg/tso/tso.go
+++ b/pkg/tso/tso.go
@@ -223,7 +223,6 @@ func (t *timestampOracle) syncTimestamp() error {
 	t.metrics.syncSaveDuration.Observe(time.Since(start).Seconds())
 
 	t.metrics.syncOKEvent.Inc()
-	// "persisted tso window to etcd (sync)"
 	// "last" is the etcd value, "last-saved" is the in-memory window, "save" is the
 	// newly persisted window, and "next" is the physical time loaded into memory
 	log.Info("sync and save timestamp",


pkg/tso/tso.go
```diff
@@ -398,7 +398,12 @@ func (t *timestampOracle) updateTimestamp(purpose updatePurpose) (bool, error) {
 			return false, err
 		}
 		t.lastSavedTime.Store(save)
-		t.metrics.updateSaveDuration.Observe(time.Since(start).Seconds())
+		saveDuration := time.Since(start)
+		t.metrics.updateSaveDuration.Observe(saveDuration.Seconds())
+		log.Debug("saved timestamp to etcd",
```
JmPotato: What do you think about only logging this and setting it to the INFO level when the `saveDuration` exceeds a certain threshold? 

We don't really need to know every time it saves; we are much more concerned with how long the saving process actually takes.

server/cluster/cluster.go
```diff
@@ -461,23 +461,25 @@ func (c *RaftCluster) checkSchedulingService() {
...
-					c.UnsetServiceIndependent(constant.TSOServiceName)
-				}
-			} else {
-				c.stopTSOJobsIfNeeded()
-				if !c.IsServiceIndependent(constant.TSOServiceName) {
-					log.Info("TSO is provided by TSO server")
-					c.SetServiceIndependent(constant.TSOServiceName)
-				}
+		if !c.opt.GetMicroserviceConfig().IsTSODynamicSwitchingEnabled() {
+			log.Debug("TSO dynamic switching is disabled, skipping TSO service check")
```
JmPotato: It isn’t on a hot path here, or it won’t be printed that frequently. So I think using INFO-level logging is fine, and it can give us some information about the timing.

server/grpc_service.go
```diff
@@ -485,6 +485,7 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		defer done()
 	}
 	if s.IsServiceIndependent(constant.TSOServiceName) {
+		log.Debug("forwarding TSO stream to independent TSO service")
```
JmPotato: I feel this indicator would be better suited as a Grafana metric. This way, we can observe the count on a monitoring dashboard.

server/cluster/cluster.go
YuhaoZhang00: Under `isKeyspaceGroupEnabled = true` and `isTSODynamicSwitchingEnabled = false` config, the log will be printed once per 100ms. Maybe worth keeping it in `Debug` level, but would be better eliminating it. For this method, logs on successful/failed state transition is just enough.

pkg/tso/tso.go
```diff
@@ -223,10 +223,14 @@ func (t *timestampOracle) syncTimestamp() error {
 	t.metrics.syncSaveDuration.Observe(time.Since(start).Seconds())
 
 	t.metrics.syncOKEvent.Inc()
-	log.Info("sync and save timestamp",
+	log.Info("persisted tso window to etcd",
+		zap.String("reason", "sync"),
 		logutil.CondUint32("keyspace-group-id", t.keyspaceGroupID, t.keyspaceGroupID > 0),
```
bufferflies, line 229: I think group 0 shouldn't be ignore

pkg/tso/tso.go
```diff
@@ -292,6 +296,12 @@ func (t *timestampOracle) resetUserTimestamp(tso uint64, ignoreSmaller, skipUppe
 		}
 		t.lastSavedTime.Store(save)
 		t.metrics.resetSaveDuration.Observe(time.Since(start).Seconds())
+		log.Info("persisted tso window to etcd",
+			zap.String("reason", "user-reset"),
```
bufferflies: how about put this reason into the "persisted tso window to etcd"

pkg/tso/tso.go
```diff
@@ -399,6 +409,12 @@ func (t *timestampOracle) updateTimestamp(purpose updatePurpose) (bool, error) {
 		}
 		t.lastSavedTime.Store(save)
 		t.metrics.updateSaveDuration.Observe(time.Since(start).Seconds())
+		log.Debug("persisted tso window to etcd",
```
bufferflies: Does it t.metrics.updateSaveDuration enough?

server/cluster/cluster.go
```diff
@@ -462,6 +469,10 @@ func (c *RaftCluster) checkSchedulingService() {
 func (c *RaftCluster) checkTSOService() {
 	if c.isKeyspaceGroupEnabled {
 		if c.opt.GetMicroserviceConfig().IsTSODynamicSwitchingEnabled() {
+			prev := c.tsoDynamicSwitchingState.Swap(tsoDynamicSwitchingStateEnabled)
```
bufferflies, line 472: Why not use CAS?

YuhaoZhang00: `checkTSOService()` is effectively serialized (only called from `Start()` and from the single `runServiceCheckJob()` goroutine), and I don't feel this path is intended to become concurrent in the near future. Maybe keep `Swap` here is the simplest yet working way.

pkg/tso/tso.go
YuhaoZhang00: It's not enough. I think the metric and the debug log serve different purposes here:
- `t.metrics.updateSaveDuration` gives the aggregate **latency** view (which is more for observation and optimization purpose)
- the debug log captures per-event context (reason, member, **saved time**), enabling correlation with other “persisted TSO window to etcd” logs to trace individual save operations. This granularity is valuable for diagnosing issues such as TSO fallback in the future.

pkg/tso/tso.go
YuhaoZhang00: To clarify: this does not suppress the log itself. It only suppresses the`keyspace-group-id` field when the group id is `0`. The log itself still exists.

I kept that behavior because it matches the existing TSO logging convention:
- in the legacy PD/default path, the allocator is always serving keyspace group `0`, so printing `keyspace-group-id: 0` on every such log is redundant.
- For non-default keyspace groups (`> 0`), we still include the field so the log remains distinguishable.

It matches all other existing log style in this file.

pkg/tso/tso.go
bufferflies: OK， mostly the debug level can't be set in the product environment. It's better to use Info level

pkg/tso/tso.go
```diff
@@ -223,10 +223,13 @@ func (t *timestampOracle) syncTimestamp() error {
 	t.metrics.syncSaveDuration.Observe(time.Since(start).Seconds())
 
 	t.metrics.syncOKEvent.Inc()
-	log.Info("sync and save timestamp",
+	log.Info("persisted tso window to etcd (sync)",
```
JmPotato: This is a critical path log that has existed for a long time, and I tend to avoid changing all the fields, as that might affect some alerting systems.

server/grpc_service.go
```diff
@@ -485,6 +485,7 @@ func (s *GrpcServer) Tso(stream pdpb.PD_TsoServer) error {
 		defer done()
 	}
 	if s.IsServiceIndependent(constant.TSOServiceName) {
+		tsoForwardStreamCounter.Inc()
```
JmPotato, line 488: Does this counter only increase and never decrease?

JmPotato (approved):

PR #10441, opened 2026-03-17
tests: retry canceled etcd startup

### LINKED ISSUE #10331: TestTSOFollowerProxy is flaky
### Which jobs are failing
```text
panic: test timed out after 5m0s
        	            	github.com/tikv/pd/tests/integrations/client_test.TestTSOFollowerProxy
        	            				/home/prow/go/src/github.com/tikv/pd/tests/integrations/client/client_test.go:276
        	Error:      	Received unexpected error:
```
### CI link
- https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/tikv_pd/10275/pull-unit-test-next-gen-3/2031622997105709056
### Anything else
- triage-key: `testtsofollowerproxy`
- test: `TestTSOFollowerProxy`

okJiang:

Issue Number: Close #10331

`TestTSOFollowerProxy` flakes when a follower etcd start gets canceled during cluster bootstrap. The CI log for #10331 shows `pd2` / `pd3` exiting during startup, followed by repeated connection-refused probes and a final `[PD:server:ErrCancelStartEtcd]etcd start canceled` from `tests/integrations/client/client_test.go`.

This PR restores the retry behavior that `RunInitialServers()` used to have for canceled etcd startup:

- factor startup error handling into `classifyStartServersError`
- retry the current server set for both `ErrCancelStartEtcd` and `ErrStartEtcd`
- keep recreating the server set only for port-conflict-style errors
- add a regression unit test for the classification logic

This also realigns `RunInitialServers()` with `RunServersWithRetry()`, which still treated canceled etcd startup as retryable.

```commit-message
tests: retry canceled etcd startup
```

### Check List

Tests

- Unit test
- Integration test

Code changes

Side effects

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Smarter retry classification for etcd startup/stop to decide when to retry, recreate, or stop, improving handling of port conflicts, cluster ID mismatches, and cancellation scenarios.

* **Tests**
  * Added tests validating error classification and retry behavior, plus test harness improvements to detect goroutine leaks.

### DIFF
--- a/tests/cluster.go
+++ b/tests/cluster.go
@@ -88,7 +88,15 @@ const (
 	startServersRetryRecreate
 )
 
-func classifyStartServersError(err error) startServersRetryAction {
+func shouldRetryCurrentServers(err error) bool {
+	if err == nil {
+		return false
+	}
+	errMsg := err.Error()
+	return strings.Contains(errMsg, "ErrCancelStartEtcd") || strings.Contains(errMsg, "ErrStartEtcd")
+}
+
+func classifyInitialServersError(err error) startServersRetryAction {
 	if err == nil {
 		return startServersNoRetry
 	}
@@ -96,7 +104,7 @@ func classifyStartServersError(err error) startServersRetryAction {
 	switch {
 	case strings.Contains(errMsg, "address already in use") || strings.Contains(errMsg, "Etcd cluster ID mismatch"):
 		return startServersRetryRecreate
-	case strings.Contains(errMsg, "ErrCancelStartEtcd") || strings.Contains(errMsg, "ErrStartEtcd"):
+	case shouldRetryCurrentServers(err):
 		return startServersRetryCurrent
 	default:
 		return startServersNoRetry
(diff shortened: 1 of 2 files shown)


tests/cluster.go
```diff
@@ -80,6 +80,29 @@ var (
...
 
+type startServersRetryAction int
+
+const (
+	startServersNoRetry startServersRetryAction = iota
+	startServersRetryCurrent
+	startServersRetryRecreate
+)
+
+func classifyStartServersError(err error) startServersRetryAction {
```
lhy1024: From the PR description, it sounds like the intended change is to make `RunInitialServers()` retry  `ErrCancelStartEtcd` while keeping port-conflict handling separate. However, reusing  `classifyStartServersError()` here also changes `RunServersWithRetry()`: bind failures like `"address
 already in use"` used to be retried here after being wrapped as `ErrStartEtcd`, but now they are classified as `startServersRetryRecreate` and this path returns immediately. Is that extra behavior change intended?

okJiang: Good catch. That extra behavior change was not intended. `RunServersWithRetry()` does not have enough context to recreate the server set, so it should keep the old behavior and retry wrapped `ErrStartEtcd` / `ErrCancelStartEtcd` failures on the current servers, even when the wrapped cause is `address already in use`.

I split the logic so the recreate classification stays local to `RunInitialServers()`, while `RunServersWithRetry()` now only checks the retryable etcd-start errors. I also added coverage for the wrapped bind-failure case and pushed the fix in `8dd5c6406`.

okJiang: Follow-up for bridge audit: this thread has already been addressed, and the fix is already pushed on the PR head at `8dd5c6406`. No further code change is needed for this comment.

PR #10442, opened 2026-03-17
tests: stabilize TestUpdateMemberWhenRecovery

### LINKED ISSUE #9994: TestUpdateMemberWhenRecovery is flaky
### Which jobs are failing
```shell
make failpoint-enable
cd tests/integrations/mcs/keyspace
go test -v -run TestKeyspaceGroupTestSuite/TestUpdateMemberWhenRecovery
```
https://github.com/tikv/pd/blob/33a67caac489ef9430778357b0aa929dd97044b9/tests/integrations/mcs/keyspace/tso_keyspace_group_test.go#L661-L673

### CI link
https://github.com/tikv/pd/actions/runs/20056411798/job/57522640945?pr=10016

* [linux.log](https://github.com/user-attachments/files/23904762/linux.log)
* [macos.log](https://github.com/user-attachments/files/23904753/macos.log)

investigation pending

### Anything else
Tested on 33a67caac489ef9430778357b0aa929dd97044b9

okJiang:

Issue Number: Close #9994

`TestUpdateMemberWhenRecovery` is flaky when all TSO nodes are stopped and one node is restarted.
The blocked `GetTS` request can stay attached to stale discovery or keyspace-group metadata long enough to hit its timeout even though the recovered TSO service is already able to serve fresh requests.

Instead of waiting only on the original in-flight `GetTS` result, the test now treats recovery as successful when either:
- the blocked request finally returns a newer TS, or
- a fresh short-timeout `GetTS` retry returns a newer TS after the service recovers.

This keeps the original no-legacy-fallback assertion, but removes the timing dependency on when the first blocked request rebinds to refreshed TSO discovery.

Verification:
- `cd tests/integrations && GOCACHE=/tmp/pd-gocache GOTMPDIR=/tmp/pd-gotmp make gotest GOTEST_ARGS='-tags=without_dashboard ./mcs/keyspace -run TestKeyspaceGroupTestSuite/TestUpdateMemberWhenRecovery -count=5 -v'` -> PASS (`5/5`, `ok ... 66.772s`)
- `GOCACHE=/tmp/pd-gocache GOTMPDIR=/tmp/pd-gotmp make basic-test` -> FAIL in unrelated packages `pkg/gctuner`, `pkg/schedule/operator`, and `pkg/storage/endpoint`

```commit-message
tests: stabilize TestUpdateMemberWhenRecovery
```

### Check List

Tests

- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Improved recovery validation by asserting a clear, monotonic timestamp increase after a service node restart, using asynchronous request handling and explicit cancellation to reduce flakiness.

* **Refactor**
  * Consolidated test retry logic into a single, reliable eventual-observation flow for recovery checks, simplifying control flow and improving stability.

### DIFF
--- a/tests/integrations/mcs/keyspace/tso_keyspace_group_test.go
+++ b/tests/integrations/mcs/keyspace/tso_keyspace_group_test.go
@@ -660,16 +660,16 @@ func (suite *keyspaceGroupTestSuite) setupTSONodesAndClient(re *require.Assertio
 
 // TestUpdateMemberWhenRecovery verifies that in TSO microservice mode (API_SVC_MODE), when all TSO nodes
 // become temporarily unavailable and then recover, the client should NOT fallback to
-// the legacy path (group 0), but should wait and successfully get TSO after nodes restart.
+// the legacy path (group 0), but should eventually get a newer TSO once service recovers.
 //
 // Test scenario:
 // 1. Setup: Start 2 TSO nodes and create keyspace group 1, client gets initial TSO
 // 2. Close all TSO nodes to simulate total TSO microservice failure
 // 3. Wait until all TSO nodes are deregistered (getTSOServerURLs returns empty)
 // 4. Enable failpoints: assertNotReachLegacyPath (panic if fallback) and extend timeout
-// 5. Start async GetTS call (will wait for TSO service to recover)
-// 6. Restart one TSO node while GetTS is waiting
-// 7. Verify GetTS succeeds after node restart (assertNotReachLegacyPath ensures no fallback)
+// 5. Start an async GetTS call while the TSO service is down
+// 6. Restart one TSO node to recover the TSO service
+// 7. Verify eventual recovery: either the in-flight GetTS or a fresh retry gets a newer TS
 func (suite *keyspaceGroupTestSuite) TestUpdateMemberWhenRecovery() {
 	re := suite.Require()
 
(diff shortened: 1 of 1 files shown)


tests/integrations/mcs/keyspace/tso_keyspace_group_test.go
```diff
@@ -738,18 +738,32 @@ func (suite *keyspaceGroupTestSuite) TestUpdateMemberWhenRecovery() {
...
-	result := <-resultCh
-	if result.err != nil {
-		testutil.Eventually(re, func() bool {
-			retryCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-			defer cancel()
-			_, _, err := client.GetTS(retryCtx)
-			return err == nil
-		}, testutil.WithWaitFor(60*time.Second), testutil.WithTickInterval(500*time.Millisecond))
-	}
+	// The in-flight GetTS may stay attached to stale discovery/metadata during
```
lhy1024, line 741: The test goal has changed here: the original in-flight GetTS sent through resultCh no longer has to recover by itself; the test now passes as long as either that request or a fresh retry gets a newer TS after recovery.

I think the comments should be updated to match. Phrases like “will wait for TSO service to recover” and “Verify GetTS succeeds after node restart” still read more like they describe the original blocked request, which no longer matches the actual assertion.

okJiang: Updated the function header and the Step 5/6/7 comments to match the current assertion: we now describe eventual recovery, where either the in-flight request or a fresh retry can observe the recovered TSO service and return a newer TS. The wording no longer implies the original blocked request must recover by itself.

okJiang: Rechecked this thread: commit a48b7a219 updates the function header and the Step 5/6/7 comments to describe eventual recovery correctly, so the wording no longer implies the original in-flight request must recover by itself. The PR head is already at that commit.\n\n

PR #10455, opened 2026-03-19
client, tso: support multi-node TSO discovery

### LINKED ISSUE #10454: Support multi-node TSO keyspace group discovery
## Problem

When a keyspace group has multiple TSO nodes(>2), client and server behavior is not robust if a request reaches a node that has watched the keyspace group metadata but is not currently serving that group's allocator.

This can lead to:
- `FindGroupByKeyspaceID` returning an error instead of usable keyspace group metadata
- TSO client discovery failing to retry through another valid node
- Health checks reporting internal errors when the allocator is absent on the current node
- split-transition paths touching allocator state without guarding nil allocators

## What is expected?

For multi-node TSO deployments:
- non-serving nodes should still return enough keyspace group metadata for clients to continue discovery
- clients should retry against another TSO node when the first node is not serving the allocator
- health checks should distinguish "allocator not found" from internal failures
- membership update logic should be safe when allocators are temporarily absent

## Reproduction

1. Start a TSO cluster with 3 nodes.
2. Create a keyspace group with only a subset of nodes as members.
3. Send `FindGroupByKeyspaceID` or client discovery traffic to a non-member TSO node.
4. Observe discovery and health behavior.

## Proposal

- treat `ErrGetAllocator` as a non-fatal "not served here" signal
- keep returning keyspace group metadata to the client
- retry client discovery through another TSO node
- return 404 from health API when allocator is absent
- add integration coverage for multi-node behavior and split flow

bufferflies:

Issue Number: Close #10454

```commit-message
tso client and server discovery now tolerate requests landing on a TSO node
that has keyspace group metadata but does not serve the allocator for that
group.

The server keeps returning keyspace group metadata for ErrGetAllocator cases,
the client retries another TSO node when no primary is returned, the health API
returns 404 when an allocator is absent, and allocator access during split-state
transitions is guarded for nil cases. Integration coverage is added for
multi-node TSO discovery and split scenarios.
```

### Check List

Tests

- Integration test

Code changes

Side effects

- Increased code complexity

### Release note

```release-note
Improve multi-node TSO keyspace group discovery so clients can retry through
other TSO nodes when a requested node is not serving the allocator.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Health endpoint returns 404 when allocator is missing to avoid a runtime crash.
  * Stale metadata updates no longer block applying URL/member changes; stale mod-revision is only treated as an error when no meta change occurred.

* **New Features**
  * Service discovery will retry using cached secondary servers when no primary is found.
  * Update outcomes now distinguish primary switch vs. meta-change for clearer behavior.

* **Tests**
  * Added unit and multi-node integration tests covering discovery, elections, and stale-revision scenarios.


tests/integrations/mcs/tso/keyspace_group_manager_test.go
```diff
@@ -126,9 +127,101 @@ func cleanupKeyspaceGroups(re *require.Assertions, server *tests.TestServer) {
...
+
+	req := tsopb.FindGroupByKeyspaceIDRequest{
+		Header: &tsopb.RequestHeader{
+			ClusterId:       clusterID,
+			KeyspaceId:      keyspaceIDs[0],
+			KeyspaceGroupId: keyspaceGroupID,
+		},
+		KeyspaceId:  keyspaceIDs[0],
+		ModRevision: 0,
+	}
```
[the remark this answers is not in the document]
bufferflies: No, it just ensure to send request to the same endpoint after group changed

lhy1024 (approved):

client/servicediscovery/tso_service_discovery.go
```diff
@@ -78,24 +79,27 @@ func (k *keyspaceGroupSvcDiscovery) getModRevision() uint64 {
...
 	// If the new primary URL is empty, we don't switch the primary URL.
 	oldPrimaryURL = k.primaryURL
 	if len(newPrimaryURL) > 0 {
 		primarySwitched = !strings.EqualFold(oldPrimaryURL, newPrimaryURL)
 		k.primaryURL = newPrimaryURL
 	}
+	// the mod revision just ensures the member use is the latest one , but primary can't be protected by the mod revision.
+	if k.getModRevision() > modRevision {
+		return oldPrimaryURL, primarySwitched, false
+	}
```
JmPotato: `k.primaryURL` was modified on line 96. This check should be moved earlier.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -78,24 +79,27 @@ func (k *keyspaceGroupSvcDiscovery) getModRevision() uint64 {
 	return k.modRevision.Load()
 }
 
+// metaChanged indicates whether the keyspace group meta has been updated,
+// it don't include the change of primary URLs.
```
JmPotato:
```suggestion
// it doesn't include the change of primary URLs.
```

client/servicediscovery/tso_service_discovery.go
```diff
@@ -482,9 +486,31 @@ func (c *tsoServiceDiscovery) updateMember() error {
...
+
+	err = c.updateMemberInner(tsoServerURL)
+	// the url is not one of the group members, so it can't get the primary URL but can get the keyspace meta info.
+	// If met errNoPrimaryURL error, it should try to get primary url from the member url.
+	if err == errNoPrimaryURL {
+		urls := c.getSecondaryURLs()
+		if len(urls) == 0 {
+			return err
+		}
+		url := urls[rand.IntN(len(urls))]
```
JmPotato, line 506: If there are 3 TSO nodes and the node you connect to first is not the allocator, randomly selecting a secondary node has only a 50% chance of reaching a serving node.

client/servicediscovery/tso_service_discovery.go
bufferflies: No, ModRevision can just guard the member info; the primary URL is dynamic.

client/servicediscovery/tso_service_discovery.go
bufferflies: Yes, but it's different. The first try will set the secondary URLs to the allocator URL, so the second must access the allocator URL.

client/servicediscovery/tso_service_discovery.go
bufferflies: done

client/servicediscovery/tso_service_discovery.go
rleungx: What if there are far more than 3 nodes?

bufferflies: Most likely, the TSO node accessed on the first attempt is not a member of the TSO group, but it can return the corresponding TSO members. On the second attempt, the request will be routed to the correct primary via those TSO members.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -78,24 +79,27 @@ func (k *keyspaceGroupSvcDiscovery) getModRevision() uint64 {
...
-	success = true
 	// If the new primary URL is empty, we don't switch the primary URL.
 	oldPrimaryURL = k.primaryURL
 	if len(newPrimaryURL) > 0 {
 		primarySwitched = !strings.EqualFold(oldPrimaryURL, newPrimaryURL)
 		k.primaryURL = newPrimaryURL
 	}
+	// the mod revision just ensures the member use is the latest one, and the primary can't be protected by the mod revision.
+	if k.getModRevision() > modRevision {
+		return oldPrimaryURL, primarySwitched, false
```
rleungx, line 100: k.primaryURL has been modified.

bufferflies: Yes, primary is another watch key, so it can't be protected by the modRevision.

bufferflies: done， revert this changes

pkg/mcs/tso/server/grpc_service.go
```diff
@@ -140,7 +140,9 @@ func (s *Service) FindGroupByKeyspaceID(
 			Header: wrapErrorToHeader(tsopb.ErrorType_INVALID_VALUE, errs.ErrKeyspaceGroupModRevisionStale.Error(), respKeyspaceGroup),
 		}, nil
 	}
-	if err != nil {
+	// If the error is the ErrGetAllocator, it indicates the server have watched the keyspace group meta.
+	// But this node's server is not one of the members.
+	if err != nil && !errs.ErrGetAllocator.Equal(err) {
```
rleungx, line 145: This lets ErrGetAllocator continue with allocator == nil, but the code below still calls allocator.GetPrimaryAddr() and can panic. Please handle the nil allocator case when building members.

pkg/mcs/tso/server/apis/v1/api.go
```diff
@@ -237,6 +237,10 @@ func getHealth(c *gin.Context) {
 		c.String(http.StatusInternalServerError, err.Error())
 		return
 	}
+	if allocator == nil {
```
rleungx: This branch is unreachable because GetAllocator returns ErrGetAllocator when the allocator is absent. Please map that error to 404 before returning 500.

pkg/tso/keyspace_group_manager.go
```diff
@@ -926,7 +926,9 @@ func (kgm *KeyspaceGroupManager) updateKeyspaceGroupMembership(
 	if oldGroup != nil {
 		// SplitTarget -> !Splitting
 		if oldGroup.IsSplitTarget() && !newGroup.IsSplitting() {
-			kgm.allocators[groupID].GetMember().(*member.Participant).SetCampaignChecker(nil)
+			if allocator := kgm.allocators[groupID]; allocator != nil {
```
rleungx, line 929: This nil guard only covers the split-complete path. `groupSplitPatroller` can still see the group in `splittingGroups` after its allocator is removed and then panic on `allocator.isServing()`, so that path needs the same guard.

pkg/mcs/tso/server/apis/v1/api.go
bufferflies: Good catch, fixed in f28c5a32c. `GetAllocator` returns `ErrGetAllocator` (non-nil error) when the allocator is absent, so the old `if allocator == nil` branch was unreachable. Now the handler maps `ErrGetAllocator` to 404 inside the `err != nil` check and falls back to 500 for other errors, and the dead branch is removed.

pkg/mcs/tso/server/grpc_service.go
bufferflies: `Allocator.GetPrimaryAddr()` is nil-receiver safe — it returns `""` when `a == nil || a.member == nil` (allocator.go), so `member.IsAddressEquivalent("")` is false for every member and no panic occurs; the node just returns the group meta with no primary, which is exactly what the client retry path expects. So this is safe as-is, but I can add an explicit `if allocator != nil` guard when building members if you prefer it to be explicit.

pkg/tso/keyspace_group_manager.go
bufferflies: `groupSplitPatroller` calls `allocator.isServing()`, which is also nil-receiver safe — it returns `false` when `a == nil || a.member == nil` (allocator.go), so the patroller hits `continue` instead of panicking when the allocator was already removed. The explicit guard I added in `updateKeyspaceGroupMembership` is for the `SetCampaignChecker(nil)` call, which dereferences through `GetMember()` and is not nil-safe. So the patroller path is already covered, but happy to add a guard there too for symmetry if you want.

tests/integrations/mcs/tso/keyspace_group_manager_test.go
```diff
@@ -126,9 +127,98 @@ func cleanupKeyspaceGroups(re *require.Assertions, server *tests.TestServer) {
...
+
+	// split keyspaceID-1 and keyspaceID-2 into new keyspace group
+	newGroupID := suite.allocID()
+	handlersutil.MustSplitKeyspaceGroup(re, suite.pdLeaderServer, keyspaceGroupID, &handlers.SplitKeyspaceGroupByIDParams{
+		NewID:     newGroupID,
+		Keyspaces: []uint32{keyspaceIDs[1], keyspaceIDs[2]},
+	})
+	suite.waitKeyspaceReady([]uint32{newGroupID}, []uint32{keyspaceIDs[1], keyspaceIDs[2]})
+	handlersutil.MustFinishSplitKeyspaceGroup(re, suite.pdLeaderServer, newGroupID)
+	checkFn()
```
rleungx: This still reuses the pre-split request, so the post-split check does not verify discovery for `newGroupID`.

bufferflies: Good catch, fixed in 677d1c134. `checkFn` is now parameterized by `(groupID, keyspaceID)` and builds the request inside, so after the split it probes both the source group (`keyspaceGroupID` / `keyspaceIDs[0]`) and the split-out `newGroupID` (`keyspaceIDs[1]`). I also added `re.Equal(groupID, resp.KeyspaceGroup.Id)` and `re.Equal(groupID, discovery.GetKeyspaceGroupID())` so a wrong-group response is caught. Verified locally: `TestMultiNodes` passes.

rleungx (approved):

PR #10303, opened 2026-03-05
tests: fix multiple flaky tests, panics, deadlocks, and goroutine leaks

JmPotato:
### What problem does this PR solve?

Issue Number: close #9533, close #8103, close #9316, close #8512, close #8035, close #8739, close #9584, close #8792, close #9800, close #9583, close #5304

This PR addresses 11 open flaky test / panic / deadlock / goroutine leak issues in a single batch fix.

### What is changed and how does it work?

**Fix flaky tests (5 issues):**
- Replace testify assertions (`re.NoError`, `re.Equal`, `re.EqualError`, etc.) inside `testutil.Eventually()` callbacks with plain Go comparisons. The testify assertions call `t.Errorf()` which permanently marks the test as failed, even if the Eventually condition eventually becomes true on retry.
- `TestConfigController` (#8035): wrap config verification in `testutil.Eventually()` to handle async config propagation.
- `TestResourceGroupRUConsumption` (#8739): replace hard-coded sleep with `testutil.Eventually()` for RU stats verification.
- `TestMergeKeyspaceGroup` (#9316): replace `re.NoError` inside `testutil.Eventually()` with plain Go checks.
- `TestResourceGroupController` (#8512): fix `re.EqualError` inside Eventually and add nil checks.

**Fix test bugs (1 issue):**
- `TestStoreStateFilterReason` (#5304): the `check` function called `Source()` twice instead of `Source()` then `Target()`. Fixed the call and updated expected values.

**Fix panics (2 issues):**
- `cleanupResourceGroups` (#8792): add `re.Error()` check before `err.Error()` to prevent nil pointer dereference.
- pd-ctl scheduler test (#9584): use safe type assertions with comma-ok pattern instead of bare assertions that panic.

**Fix deadlock (1 issue):**
- `TestServer.Run()` (#9800): release mutex before calling `s.server.Run()` which can block indefinitely on etcd initialization, preventing deadlock with `Stop()`/`Destroy()`.

**Fix goroutine leak (1 issue):**
- `ResourceGroupsController` (#9583): track goroutines spawned for `handleTokenBucketUpdateEvent` with the controller's WaitGroup so they are properly waited on during `Stop()`.

### Check List

Tests 

- [x] Existing unit/integration tests cover the changes
- [ ] Manual test (e.g. if the change is about determine the version of a binary)

Code changes

- [x] Has no Go public API or behavior changes
- [ ] Contains Go public API or behavior changes

Side effects

- [x] No increase in test time

PR #10462, opened 2026-03-19
tests: stabilize TestResourceGroupRUConsumption

### LINKED ISSUE #8739: TestResourceGroupRUConsumption is unstable
### CI link
https://github.com/tikv/pd/actions/runs/11452910057/job/31870117021

okJiang:

Issue Number: close #8739

`TestResourceGroupRUConsumption` is unstable because it assumes RU consumption
statistics are visible immediately after `AcquireTokenBuckets` returns.

The test currently sleeps for 10ms and then asserts `GetResourceGroup(...,
pd.WithRUStats)` already reflects the reported consumption. That assumption is
not guaranteed: `AcquireTokenBuckets` only enqueues consumption into
`Manager.consumptionDispatcher`, and the actual `RUStats` update happens later
in `backgroundMetricsFlush`.

To make that race deterministic, this patch adds a test-only failpoint
`delayConsume` in `backgroundMetricsFlush` and enables it in the flaky test.
With `delayConsume=100ms`, the old fixed-sleep assertion fails reproducibly:
`g.RUStats` is still zero while `testConsumption` already contains the expected
values.

- add a test-only `delayConsume` failpoint before the background consumption
  worker applies RU stats
- update `TestResourceGroupRUConsumption` to wait on the observable condition
  instead of sleeping for a fixed 10ms
- reuse the same condition-based check after leader transfer so the test waits
  for persisted stats to become visible again

This keeps production behavior unchanged. The new failpoint only activates when
a test explicitly enables it.

Root-cause evidence

- `pkg/mcs/resourcemanager/server/manager.go:730+`: consumption is processed
  asynchronously from `consumptionDispatcher`
- `tests/integrations/mcs/resourcemanager/resource_manager_test.go` previously
  slept 10ms before asserting RU stats had already updated
- red test with `delayConsume=100ms`:
  - `GOCACHE=/tmp/pd-go-cache make gotest GOTEST_ARGS='-tags=without_dashboard ./mcs/resourcemanager -run TestResourceManagerClientTestSuite/pd-resource-manager/TestResourceGroupRUConsumption -count=1 -vet=off'`
  - failed before the fix because `g.RUStats` remained zero

Historical analog

- Playbook Pattern 5: Flaky and Unstable Test Stabilization
- Representative merged PRs:
  - #10134 `tests: fix some unstable tests`
  - #9465 `*: fix flaky test TestPreparingProgress and TestRemovingProgress`
- Prior issue-specific attempt:
  - #10303 `tests: fix multiple flaky tests, panics, deadlocks, and goroutine leaks`

Risk

- scoped to one flaky integration test plus a failpoint-only hook in the
  asynchronous RU stats path
- no production behavior, API, or config changes unless the failpoint is
  explicitly enabled in tests

Verification

- red before the fix:
  - `GOCACHE=/tmp/pd-go-cache make gotest GOTEST_ARGS='-tags=without_dashboard ./mcs/resourcemanager -run TestResourceManagerClientTestSuite/pd-resource-manager/TestResourceGroupRUConsumption -count=1 -vet=off'`
  - failed deterministically with `delayConsume=100ms`
- green after the fix:
  - `GOCACHE=/tmp/pd-go-cache make gotest GOTEST_ARGS='-tags=without_dashboard ./mcs/resourcemanager -run TestResourceManagerClientTestSuite/pd-resource-manager/TestResourceGroupRUConsumption -count=1 -vet=off'`
  - passed
- green in standalone RM discovery mode:
  - `GOCACHE=/tmp/pd-go-cache make gotest GOTEST_ARGS='-tags=without_dashboard ./mcs/resourcemanager -run TestResourceManagerClientTestSuite/standalone-resource-manager-with-client-discovery/TestResourceGroupRUConsumption -count=1 -vet=off'`
  - passed
- touched manager path:
  - `GOCACHE=/tmp/pd-go-cache make gotest GOTEST_ARGS='./pkg/mcs/resourcemanager/server -run TestCleanUpTicker -count=1'`
  - passed
- repo baseline:
  - `GOCACHE=/tmp/pd-go-cache make basic-test`
  - not clean in this worktree due an unrelated existing failure in
    `pkg/mcs/resourcemanager/server TestSetServiceLimit`

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Tests**
  * Enhanced test robustness for resource consumption metrics by implementing polling-based validation instead of fixed delays, and adding test infrastructure for simulating consumption processing delays.

### DIFF
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -732,7 +732,7 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 				continue
 			}
 			failpoint.Inject("github.com/tikv/pd/pkg/mcs/resourcemanager/server/delayConsume", func(val failpoint.Value) {
-				time.Sleep(time.Duration(val.(int)) * time.Millisecond)
+				time.Sleep(time.Duration(val.(int) * int(time.Millisecond)))
 			})
 			keyspaceID := consumptionInfo.keyspaceID
 			keyspaceName, err := m.getKeyspaceNameByID(ctx, keyspaceID)
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -731,7 +731,7 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 			if consumptionInfo == nil || consumptionInfo.Consumption == nil {
 				continue
 			}
-			if val, _err_ := failpoint.Eval(_curpkg_("delayConsume")); _err_ == nil {
+			if val, _err_ := failpoint.Eval("github.com/tikv/pd/pkg/mcs/resourcemanager/server/delayConsume"); _err_ == nil {
 				time.Sleep(time.Duration(val.(int)) * time.Millisecond)
 			}
 			keyspaceID := consumptionInfo.keyspaceID
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -731,9 +731,9 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 			if consumptionInfo == nil || consumptionInfo.Consumption == nil {
 				continue
 			}
-			if val, _err_ := failpoint.Eval("github.com/tikv/pd/pkg/mcs/resourcemanager/server/delayConsume"); _err_ == nil {
+			failpoint.Inject("github.com/tikv/pd/pkg/mcs/resourcemanager/server/delayConsume", func(val failpoint.Value) {
 				time.Sleep(time.Duration(val.(int)) * time.Millisecond)
-			}
+			})
 			keyspaceID := consumptionInfo.keyspaceID
 			keyspaceName, err := m.getKeyspaceNameByID(ctx, keyspaceID)
 			if err != nil {
--- a/pkg/mcs/resourcemanager/server/manager.go
+++ b/pkg/mcs/resourcemanager/server/manager.go
@@ -731,6 +731,9 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 			if consumptionInfo == nil || consumptionInfo.Consumption == nil {
 				continue
 			}
+			if val, _err_ := failpoint.Eval(_curpkg_("delayConsume")); _err_ == nil {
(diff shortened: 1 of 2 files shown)


pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -731,6 +731,9 @@ func (m *Manager) backgroundMetricsFlush(ctx context.Context) {
 			if consumptionInfo == nil || consumptionInfo.Consumption == nil {
 				continue
 			}
+			failpoint.Inject("github.com/tikv/pd/pkg/mcs/resourcemanager/server/delayConsume", func(val failpoint.Value) {
```
okJiang, line 734: This is a test-only failpoint used to reproduce unstable tests. After the review, this failpoint can be removed.

tests/integrations/mcs/resourcemanager/resource_manager_test.go
```diff
@@ -1532,10 +1536,23 @@ func (suite *resourceManagerClientTestSuite) TestResourceGroupRUConsumption() {
...
+			}
+			return g.RUStats.RRU == testConsumption.RRU &&
+				g.RUStats.WRU == testConsumption.WRU &&
+				g.RUStats.ReadBytes == testConsumption.ReadBytes &&
+				g.RUStats.WriteBytes == testConsumption.WriteBytes &&
+				g.RUStats.TotalCpuTimeMs == testConsumption.TotalCpuTimeMs &&
+				g.RUStats.SqlLayerCpuTimeMs == testConsumption.SqlLayerCpuTimeMs &&
+				g.RUStats.KvReadRpcCount == testConsumption.KvReadRpcCount &&
+				g.RUStats.KvWriteRpcCount == testConsumption.KvWriteRpcCount
+		})
```
lhy1024, line 1553: Will we miss some fields? Or will we be unable to cover new fields added in the future?

okJiang: If you add a field and want to test it, it should be added to `testConsumption` and set to a specific value. At that point, the modifier should add the corresponding field again.

pkg/mcs/resourcemanager/server/manager.go
rleungx, line 734: This should use the short failpoint name (`delayConsume`). `failpoint-ctl enable` wraps `Inject` names with `_curpkg_`, so the full path here becomes double-prefixed and the test Enable call will not trigger it.

PR #10467, opened 2026-03-20
keyspace, schedule: make keyspace creation atomic

### LINKED ISSUE #10461: keyspace: ensure cache updates are atomic with txn commit
### Problem

`CreateKeyspace()` performs keyspace creation through multiple independent steps instead of one atomic storage transaction. If an earlier transaction succeeds but a later transaction fails, persistent storage can be left in a partially committed and inconsistent state.

### Affected flow

In `pkg/keyspace/keyspace.go:229`, `CreateKeyspace()` roughly does:

1. `saveNewKeyspace()`
   - stores keyspace ID mapping
   - stores keyspace metadata in `DISABLED` state
2. `splitKeyspaceRegion()`
3. `UpdateKeyspaceStateByID()`
   - updates keyspace state to `ENABLED` in another txn
4. `UpdateKeyspaceForGroup()`
   - updates keyspace group membership separately

These steps are not atomic as a whole.

### Current behavior

If `txn1` succeeds but a later step fails, the earlier storage writes remain committed.

Examples:
- keyspace ID and metadata are persisted, but state is never updated to `ENABLED`
- keyspace metadata exists, but keyspace group membership is missing
- region split side effects and storage metadata can observe different progress

This leaves storage in an intermediate state that does not represent either:
- a fully created keyspace, or
- a fully rolled-back keyspace

### Expected behavior

Keyspace creation should provide atomic semantics at the workflow level:
- either all required storage-visible creation steps succeed
- or no partial keyspace state remains persisted

At minimum, failure in later steps should trigger reliable compensation so storage remains consistent.

### Why this matters

A partially created keyspace can confuse later reads and management operations because different subsystems observe different creation progress.

This can cause:
- keyspace metadata existing in storage but not being usable
- missing or incomplete keyspace group membership
- follow-up operations acting on half-created keyspaces
- harder recovery and operational debugging

### Suggested fix

Ensure the create flow has workflow-level atomicity.

Possible directions:
- reduce the number of separate storage transactions where possible
- make intermediate states explicit and safely recoverable
- add compensation/rollback for any already-committed storage updates when later steps fail
- guarantee that keyspace metadata, state transition, and group membership cannot be left inconsistent after failure

### Reproduction idea

Inject a failure after `saveNewKeyspace()` succeeds but before `UpdateKeyspaceStateByID()` or `UpdateKeyspaceForGroup()` completes.

Then verify storage contains partial keyspace data, such as:
- keyspace ID mapping exists
- keyspace metadata exists in `DISABLED` state or incomplete state
- keyspace group assignment is missing

This demonstrates that one transaction can succeed while a later one fails, leaving inconsistent persistent state.

bufferflies:

Issue Number: Close #10461

```commit-message
Make keyspace creation persist keyspace metadata, keyspace-group membership,
and region label rules in one transaction so create failures no longer leave
partial storage state behind. Move in-memory keyspace and keyspace-group cache
updates to post-commit callbacks so failed transactions do not leak committed
state into memory.
```

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

### Release note

```release-note
Fix partial keyspace creation state when later create steps fail after the
initial metadata transaction succeeds.
```

## Summary by CodeRabbit

* **Refactor**
  * Restructured keyspace creation logic to use transactional operations for improved atomicity and reliability.
  * Enhanced region labeler with improved transaction handling and method visibility updates.

### DIFF
--- a/pkg/keyspace/keyspace.go
+++ b/pkg/keyspace/keyspace.go
@@ -37,7 +37,6 @@ import (
 	"github.com/tikv/pd/pkg/storage/endpoint"
 	"github.com/tikv/pd/pkg/storage/kv"
 	"github.com/tikv/pd/pkg/utils/etcdutil"
-	"github.com/tikv/pd/pkg/utils/keypath"
 	"github.com/tikv/pd/pkg/utils/logutil"
 	"github.com/tikv/pd/pkg/utils/syncutil"
 	"github.com/tikv/pd/pkg/versioninfo/kerneltype"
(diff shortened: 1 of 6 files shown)


bufferflies:
#10492

PR #10473, opened 2026-03-20
tso: prevent dual writers when upgrading from PD service mode

YuhaoZhang00:

Issue Number: ref #10329

When upgrading from PD service mode (`isKeyspaceGroupEnabled=false`) to API service mode with a TSO microservice, there is a window where both the PD leader and the TSO microservice could write to the same TSO timestamp key, causing dual writers and potential timestamp regression.

In PD service mode, `SaveTimestamp` now atomically checks for an existing TSO microservice primary before writing the timestamp window. If a TSO primary is detected, the write is rejected, causing the allocator to reset and resign PD leadership. This allows a PD in API service mode to take over and proxy TSO requests, preventing dual writers during rolling upgrades.

Also remove `TestMixedTSODeployment` which tested seamless coexistence of PD service mode and TSO microservice — a scenario not supported by the official upgrade procedure (PD restart is required).

Key changes:

- Add `checkTSOPrimary` parameter to `SaveTimestamp` that atomically loads the TSO primary election key (`/ms/{cluster_id}/tso/00000/primary`) within the same etcd transaction before writing the timestamp window. When a TSO primary is detected, the write is rejected.
- `checkTSOPrimary` is only true in PD service mode (`!isKeyspaceGroupEnabled`). In API service mode, PD uses the existing `checkTSOService` stop/start mechanism instead.
- Remove `TestMixedTSODeployment` - seamless coexistence of PD service mode and TSO microservice is not a supported upgrade path.

### Check List

Tests

- Unit test
- Integration test

Code changes

None.

Side effects

None.

Related changes

None.

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents dual timestamp writers by adding an optional primary-election check that blocks timestamp writes when a TSO microservice primary is present.
  * Logging variable naming corrected.

* **Tests**
  * Added tests covering the primary-election guard (success/failure cases).
  * Removed an obsolete mixed-deployment integration test and added a new integration verifying dual-writer prevention.
* **Other**
  * Behavior is now configurable to enable/disable the primary check.

### DIFF
--- a/pkg/storage/endpoint/tso.go
+++ b/pkg/storage/endpoint/tso.go
@@ -85,6 +85,7 @@ func (se *StorageEndpoint) SaveTimestamp(
 		zap.Time("ts", ts),
 		zap.String("leader-key", leadership.GetLeaderKey()),
 		zap.String("expected-leader-value", leadership.GetLeaderValue()),
+		zap.Bool("check-tso-primary", checkTSOPrimary),
 	}
 	log.Debug("saving timestamp to the storage", logFields...)
 	// The PD leadership or TSO primary will always be granted first before the TSO timestamp window is saved.


pkg/utils/keypath/absolute_key_path.go
```diff
@@ -269,6 +269,11 @@ func TimestampPath(groupID uint32) string {
 	return fmt.Sprintf(msTimestampPathFormat, ClusterID(), groupID)
 }
 
+// TSODefaultPrimaryPath returns the primary election path for the default keyspace group's TSO service.
+func TSODefaultPrimaryPath() string {
```
JmPotato: Maybe resue `ElectionPath` rather than adding a new one?

pkg/storage/endpoint/tso.go
```diff
@@ -88,9 +98,22 @@ func (se *StorageEndpoint) SaveTimestamp(ctx context.Context, groupID uint32, ts
...
+		// In PD service mode, check if a TSO microservice primary has been elected.
+		// If so, the TSO microservice owns the timeline and this PD must stop
+		// writing to avoid dual writers during a rolling upgrade.
+		if checkTSOPrimary {
+			tsoPrimaryPath := keypath.TSODefaultPrimaryPath()
+			tsoPrimary, err := txn.Load(tsoPrimaryPath)
+			if err != nil {
+				return err
+			}
+			if tsoPrimary != "" {
```
JmPotato: Prefer using `len` cause it also checks the nil case.

```suggestion
			if len(tsoPrimary) == 0 {
```

tests/integrations/tso/client_test.go
```diff
@@ -596,55 +595,6 @@ func (suite *tsoClientTestSuite) TestTSOServiceDiscovery() {
 	checkServiceDiscovery(re, pdClient, 3)
 }
 
-// When we upgrade the PD cluster, there may be a period of time that the old and new PDs are running at the same time.
-// So we need another cluster to run this test.
-func TestMixedTSODeployment(t *testing.T) {
```
JmPotato, line 601: Why remove this test?

YuhaoZhang00: This test asserts that PD (in PD service mode) and the TSO microservice can serve TSO **concurrently** as dual writers - which is exactly the unsafe behavior this PR prevents.

Anyway, I should not have removed this test case; it should be FIXED rather than removed. Will address this in follow-up commits.

YuhaoZhang00: Please check the fixed test case :)

pkg/storage/endpoint/tso.go
```diff
@@ -68,14 +69,24 @@ func (se *StorageEndpoint) LoadTimestamp(groupID uint32) (time.Time, error) {
...
+	groupID uint32,
+	ts time.Time,
+	leadership *election.Leadership,
+	checkTSOPrimary bool,
+) error {
+	logFields := []zap.Field{
 		zap.Uint32("group-id", groupID),
 		zap.Time("ts", ts),
 		zap.String("leader-key", leadership.GetLeaderKey()),
 		zap.String("expected-leader-value", leadership.GetLeaderValue()),
```
JmPotato, line 87: You can add the `checkTSOPrimary` to the `logFields` as well.

pkg/storage/endpoint/tso.go
```diff
@@ -68,14 +69,24 @@ func (se *StorageEndpoint) LoadTimestamp(groupID uint32) (time.Time, error) {
...
+//
+// When checkTSOPrimary is true (PD service mode), SaveTimestamp atomically checks whether a TSO microservice
+// primary exists. If a TSO primary is detected, the write is rejected to prevent dual writers during a
+// rolling upgrade from PD service mode to API service mode.
+func (se *StorageEndpoint) SaveTimestamp(
+	ctx context.Context,
+	groupID uint32,
+	ts time.Time,
+	leadership *election.Leadership,
+	checkTSOPrimary bool,
```
JmPotato, line 81: I noticed that, to pass this variable in, we've gone through many layers and modified several structures to carry additional fields.

Is there a more elegant or direct way for the `storageEndpoint` function to detect this variable, rather than passing it all the way through the call stack?

YuhaoZhang00: Good point about the layering. I explored alternatives but think the current approach is the best trade-off here.

1. The check must live inside `SaveTimestamp`'s etcd transaction for atomicity guarantee. Plus, `SaveTimestamp` already does the same kind of thing: leadership checking
  2. The decision to enable it can only come from `server.go`. That's the only place that knows whether we're in PD service mode (`!isKeyspaceGroupEnabled`)
  3. Wiring it differently doesn't really help. For example, storing the flag on `StorageEndpoint` avoids threading it through the call stack, but `StorageEndpoint` is a generic base shared   by all storage backends, and putting a TSO-specific config there is a different kind of layering issue.

pkg/storage/endpoint/tso.go
```diff
@@ -88,9 +100,22 @@ func (se *StorageEndpoint) SaveTimestamp(ctx context.Context, groupID uint32, ts
...
-			log.Error("leader value does not match", append(logFilds, zap.String("current-leader-value", leaderValue))...)
+			log.Error("leader value does not match", append(logFields, zap.String("current-leader-value", leaderValue))...)
 			return errors.Errorf("%s due to leader value does not match, current: %s, expected: %s", errs.NotLeaderErr, leaderValue, expected)
 		}
+		// In PD service mode, check if a TSO microservice primary has been elected.
+		// If so, the TSO microservice owns the timeline and this PD must stop
+		// writing to avoid dual writers during a rolling upgrade.
+		if checkTSOPrimary {
+			tsoPrimaryPath := keypath.ElectionPath(&keypath.MsParam{ServiceName: constant.TSOServiceName})
+			tsoPrimary, err := txn.Load(tsoPrimaryPath)
```
liyishuai, line 111: What is the performance impact of this operation?

YuhaoZhang00: For 1 single `SaveTimestamp()` method call, this query adds 1 extra single-key read in the same etcd transaction that already does the leadership check and timestamp-window persistence. So the incremental cost here is just one more key lookup before commit.

`SaveTimestamp()`  only runs roughly once per `TSOSaveInterval` (5s) in steady state. This is not on the `GetTS` hot path.

bufferflies: Only the first save time needs to check the TSO primary. Can the latter ignore  this if the first save is successful?

Issue #10481, opened 2026-03-24
resource-manager: add billing RU metric by keyspace RU version


disksing:
## Enhancement Task

Add a Prometheus-only billing RU metric in resource manager.

The metric should:
- expose billing RU totals without sending them to metering
- switch the billing formula dynamically by each keyspace's `ru_version`
- use `RRU + WRU + SQL RU` for v1 keyspaces
- use `TiKV RU v2 + TiDB RU v2 + TiFlash RU v2` for v2 keyspaces
- follow the existing RU v1/v2 metric naming and collection flow

PR #10482, opened 2026-03-24
resource-manager: add active request unit metric by keyspace version

disksing:

Issue Number: ref #10481

```commit-message
add a Prometheus-only active request unit metric for resource manager and
switch the active RU calculation dynamically by keyspace RU version

for v1 keyspaces, the metric accumulates RRU + WRU + SQL RU
for v2 keyspaces, the metric accumulates TiKV RU v2 + TiDB RU v2 +
TiFlash RU v2

add helpers to resolve effective RU version from the controller config and
cover both v1 default and v2 override cases with unit tests
```

### Check List

Tests

- Unit test

### Release note

```release-note
Added a Prometheus active request unit metric for resource manager that
switches between RU v1 and RU v2 accounting per keyspace configuration.
```

## Release Notes

* **New Features**
  * Added new Prometheus metric to track active request units with per-keyspace granularity.
  * Enabled keyspace-specific Request Unit version overrides, supporting both V1 and V2 calculation methods with configurable defaults.

* **Tests**
  * Added comprehensive test coverage for active request unit metrics and version-specific calculations.

### DIFF
--- a/pkg/mcs/resourcemanager/server/metrics.go
+++ b/pkg/mcs/resourcemanager/server/metrics.go
@@ -67,12 +67,12 @@ var (
 			Name:      "write_request_unit_sum",
 			Help:      "Counter of the write request unit cost for all resource groups.",
 		}, []string{resourceGroupNameLabel, newResourceGroupNameLabel, typeLabel, keyspaceNameLabel})
-	activeResourceRequestUnitCost = prometheus.NewCounterVec(
+	activeRequestUnitCost = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: namespace,
 			Subsystem: ruSubsystem,
-			Name:      "active_resource_request_unit_sum",
-			Help:      "Counter of the active resource request unit cost for all resource groups.",
+			Name:      "active_request_unit_sum",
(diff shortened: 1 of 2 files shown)


pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -67,6 +67,13 @@ var (
 			Name:      "write_request_unit_sum",
 			Help:      "Counter of the write request unit cost for all resource groups.",
 		}, []string{resourceGroupNameLabel, newResourceGroupNameLabel, typeLabel, keyspaceNameLabel})
+	billingRequestUnitCost = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: ruSubsystem,
+			Name:      "billing_request_unit_sum",
```
JmPotato: Should we expose its actual purpose in the code? Would it be better to keep a more generic or less direct name?

disksing: name was given by PM, but I think we can change it to a better one. do you have any thought? maybe active_request_unit_sum?

PR #10498, opened 2026-03-26
gc, server: Add WatchGCStates API

### LINKED ISSUE #10497: Implement WatchGCStates API
## Development Task

Previously in GC API refactor project (#8978), the watching API was considered in low priority and not supported right away. Now as starter/essential is going to change its PD branch from `pd-cse` to this upstream repo, and these clusters can have too many keyspaces in a single physical cluster, TiKV polling `GetAllKeyspacesGCStates` is ineffective. Therefore a watching API is needed.

In pd-cse, there's a `WatchGCSafePointV2` API, but was removed in the refactor project, as it's not used by versions other than serverless and cannot be easily implemented based on the current architector of GC's code. Now as there's such a migrating plan, it becomes necessary too for smooth upgrading.

So we need:

* Implement `WatchGCStates` API
* Implement a compatible wrapper over `WatchGCStates` to provide a `WatchGCSafePointsV2` API.

MyonKeminta:

Issue Number: Close #10497, Ref #8978

```commit-message
This PR Adds `WatchGCStates` API, and makes `WatchGCSafePointV2` compatible.
```

Requires:
* https://github.com/pingcap/kvproto/pull/1440

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Added streaming API to watch garbage collection state changes in real-time with optional barrier filtering.
  * Implemented functional GC safe point streaming endpoints with batched delivery and duplicate suppression.

* **Tests**
  * Added comprehensive test coverage for GC state watching and leader/follower transitions.

* **Chores**
  * Updated Go module dependencies to latest stable versions.


pkg/core/region.go
```diff
@@ -257,11 +257,7 @@ func RegionFromHeartbeat(heartbeat RegionHeartbeatRequest, flowRoundDivisor uint
 		region.approximateKvSize = int64(h.GetApproximateKvSize() / units.MiB)
 		region.approximateColumnarKvSize = int64(h.GetApproximateColumnarKvSize() / units.MiB)
 		region.replicationStatus = h.GetReplicationStatus()
-		if cpuStats := h.GetCpuStats(); cpuStats != nil {
-			region.cpuUsage = cpuStats.GetUnifiedRead()
-		} else {
-			region.cpuUsage = h.CpuUsage
-		}
+		region.cpuUsage = h.CpuUsage
```
rleungx, line 264: why?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +169,320 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+	m.nodeLeadership.Add(1)
+}
+
+// OnNodeBecomesFollower marks the current PD node as follower and closes all existing GC state watches.
+func (m *GCStateManager) OnNodeBecomesFollower() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.nodeLeadership.Add(-1)
+	m.cancelAllGCStateListenersLocked()
```
rleungx: I don't think the leadership counter is buying us much yet. If lease A's follower callback runs after lease B has already called `OnNodeBecomesLeader()`, we still decrement here and then close every watcher unconditionally. That means a stale callback from the old lease can tear down watches that belong to the current lease. Should this only cancel listeners when the counter drops to 0?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +169,320 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+	defer m.deregisterGCStateListener(listener)
+
+	if !skipLoadingInitial {
+		m.sendInitialGCStates(ctx, listener)
+	}
+
+	// In the gRPC path, leadership loss closes listener.ch first. The consume loop then returns, the
+	// handler's deferred cancel() fires, and this wait is released so the watch goroutine can exit and
+	// run the deferred deregistration.
+	<-ctx.Done()
```
rleungx: One lifecycle gap here: closing `listener.ch` does not stop this goroutine. `runGCStateWatch()` only waits on `ctx.Done()`, while `removeGCStateListenerLocked()` and `cancelAllGCStateListenersLocked()` close the channel without canceling the context. The new gRPC path happens to `defer cancel()`, so it eventually unwinds there, but the `GCStateManager` API itself still leaks until the caller cancels. I think this loop should also exit when the listener is removed or the channel is closed.

pkg/gc/gc_state_manager.go
MyonKeminta: > I don't think the leadership counter is buying us much yet.

The counter is for solving the problem:

1. Becomes leader of lease 1 (marked as leader)
2. Becomes follower
3. Becomes leader of lease 3 (marked as leader)
4. Receives the message of losing leadership of lease 1 (marked as follower but it's acutally leader of lease 3)

And the incorrect state may be kept for a long time.

> Should this only cancel listeners when the counter drops to 0?

This may cause streams that were created during leadership of lease 1 still be kept alive in lease 3. If there were any update during lease 2 in which the other PD node is the leader, the stream may lose that event and the cliet will never know it. Considering the possibility of low GC run interval, a stale GC state may be hold for a long time, causing ineffective GC.

> That means a stale callback from the old lease can tear down watches that belong to the current lease.

To solve the problem, the proper solution is to distinguish listeners that were created in different leases. But I think it's questionable that how likely the case may happen and how unacceptable the overhead of recreating these new stream may be.

pkg/gc/gc_state_manager.go
MyonKeminta: `consumeGCStateWatch` should exit when the channel is closed, and the deferred `cancel` should be able to ble called. Does it make sense?
I don't like to write it this way personnally, but just accepted codex's suggestion. It seems it's not robust enough to rely on the external cancellation.

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +220,336 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+		m.mu.Unlock()
+		watch.setCloseErr(errs.ErrNotLeader)
+		watchCancel()
+		close(watch.ch)
+		return nil, errs.ErrNotLeader
+	}
+	m.gcStateListeners = append(m.gcStateListeners, listener)
+	m.mu.Unlock()
+
+	go m.runGCStateWatch(listener, skipLoadingInitial)
```
rleungx: Will the GC state fallback if the state changed before watch?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +220,336 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+	return watch, nil
+}
+
+func (m *GCStateManager) runGCStateWatch(listener gcStateListener, skipLoadingInitial bool) {
```
rleungx: Do we need to log panic or use a wait group/context to manage the lifetime?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +220,336 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+	defer m.deregisterGCStateListener(listener, listener.watch.Err())
+
+	if !skipLoadingInitial {
```
rleungx, line 322: Please also add a comment for skipLoadingInitial about when it will become true and false.

MyonKeminta: Added to comments of `WatchGCStates`

pkg/gc/gc_state_manager.go
MyonKeminta: Sorry I didn't get it. What problem does this solve?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +238,190 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+
+func (m *GCStateManager) sendGCStateToListenerLocked(listener gcStateListener, gcState GCState, barrierOnly bool) bool {
+	if barrierOnly && listener.flags&gcStateListenerExcludeGCBarriers != 0 {
+		return true
+	}
+
+	timer := time.NewTimer(gcStateListenerSendTimeout)
+	defer timer.Stop()
+
+	select {
```
rleungx, line 434: If the request is canceled, we need to wait for a timeout?

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +238,190 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
...
+		case <-listener.watcher.Done():
+			// Do nothing. The `iterateAllKeyspacesGCStates` call should be able to exit when it checks the context the
+			// next time.
+			return
+		default:
+		}
+		m.sendGCStateToListenerLocked(listener, gcState, false)
+	})
+	if err != nil {
+		log.Warn("failed to load initial GC states for GC state watcher", zap.Error(err))
```
rleungx: If the error happens, will it keep waiting until <-listener.watcher.Done()?

MyonKeminta: It's a mistake to ignore the error. Changed to closing the watcher.

pkg/gc/gc_state_manager.go
MyonKeminta: listener.watcher.ctx is missing in the select block. Added.

pkg/gc/gc_state_manager.go
rleungx: We need a log if a goroutine panics.

MyonKeminta: Added recover and a log here.

rleungx:
CI failed

rleungx (commented):
The rest LGTM

MyonKeminta:
@rleungx kvproto should be merged first

MyonKeminta:
Mark as draft as the required kvproto PR is not merge yet

wfxr (changes requested):
After reviewing the current implementation together with the CSE consumer, I think the following changes are needed before merging.

### 1. Model the stream as GC state changes

Please update pingcap/kvproto#1528 so that the stream carries changes rather than bare states:

```proto
message GCStateChange {
  oneof change {
    // Insert or replace the complete GC state of this keyspace.
    GCState upsert = 1;

    // Remove this keyspace from the client's materialized GC-state view.
    KeyspaceScope removed_keyspace = 2;
  }
}

message WatchGCStatesResponse {
  ResponseHeader header = 1;
  repeated GCStateChange changes = 2;
}
```

`removed_keyspace` means removing the entry from the client's materialized view; it does not necessarily mean that the keyspace metadata was physically deleted.

The first PD implementation may emit only `upsert`, but kvproto and the CSE adapter should support both variants now. Changing field 2 from `repeated GCState` to an event wrapper after the interface is released would not be a safe wire-compatible evolution.

The interface should document that:

- `upsert` carries a complete state rather than a partial patch.
- Changes are applied in order, both within a response and across stream responses.
- A client must not observe an older state for a keyspace after a newer update.
- After the stream disconnects, the client reconnects and reloads the initial state.

### 2. Prevent initial/live update reordering

`sendInitialGCStates` can currently produce the following sequence:

```text
The initial loader reads K=v1 into a local batch.
A concurrent Advance updates K to v2 and enqueues v2.
The initial loader later enqueues the stale v1 batch.
The CSE client observes v2 followed by v1.
```

The CSE cache currently performs an unconditional upsert, so its final value really can regress to v1.

One possible solution is to maintain a per-watcher `dirtyDuringInitial` set. Live updates mark the keyspace dirty under `GCStateManager.mu`; before an initial batch is enqueued, initial entries for dirty keyspaces are filtered under the same mutex. This guarantees that either the initial value is enqueued before the live value, or the stale initial value is omitted.

Please add a deterministic failpoint test that pauses after reading v1, advances the state to v2, resumes the initial loader, and verifies that the client never ends with or observes v2 followed by v1.

### 3. Isolate slow watchers from the GC write path

Each watcher already has an independent queue, but `sendGCStatesToListenerLocked` may wait for up to 10 seconds while holding the global `GCStateManager.mu`.

When a watcher's queue is full, please cancel and remove only that watcher instead of waiting while holding the manager mutex:

```go
select {
case watcher.ch <- changes:
    return true
case <-watcher.Done():
    removeWatcher()
    return false
default:
    removeSlowWatcher()
    return false
}
```

The CSE client can reconnect with exponential backoff and jitter and recover from the new initial state.

Please test that filling watcher A's queue:

- Does not delay a GC safe-point update.
- Cancels only watcher A.
- Does not prevent watcher B from receiving the update.
- Allows watcher A to recover after reconnecting.

### 4. Keep the RPC interface within the current scope

The current scope does not include implementing `WatchGCSafePointV2` through `WatchGCStates`. Please remove the compatibility wrapper and its associated tests from this PR.

Please also remove `skip_loading_initial`. There is no current caller after removing the compatibility wrapper, and without a revision or another continuity mechanism, callers cannot safely compose `GetAllKeyspacesGCStates` with `WatchGCStates(skip_loading_initial=true)` without an update-loss window.

### 5. Fix the gRPC rate-limit label

`rateLimitCheck()` is called from `openGCStateWatch`, so stack-based caller detection resolves the label as `openGCStateWatch` rather than the registered gRPC method name `WatchGCStates`. As a result, the configured limiter for `WatchGCStates` is not applied.

After removing the compatibility wrapper, this check can be moved directly into the public `WatchGCStates` handler. Alternatively, pass the service label explicitly rather than deriving it from the call stack.

Please add a test covering the concurrency limit for `WatchGCStates`.

### 6. Account for the complete protobuf wire size

The response batching logic currently adds only `proto.Size(item)`. It omits the repeated-field tag and length prefix. After introducing `GCStateChange`, it must also account for the wrapper and its nested message envelope.

Please calculate each item's complete wire contribution before comparing it with `gcStateWatchMaxResponseBytes`.

PR #10507, opened 2026-03-30
resourcegroup: replace blind sleep with signal-aware wait in acquireTokens (#10252)

### LINKED ISSUE #10251: Resource Group: 1s latency spike when FillRate far exceeds actual RU consumption
## Bug Report

When a resource group's `FillRate` greatly exceeds actual RU consumption (e.g., FillRate=1,000,000 vs ~200 RU/s), requests intermittently experience **~1s blocking** due to the client-side token limiter entering a starved state.

### Reproduction

1. Create a resource group with a high `FillRate` (e.g., 1,000,000 RU/s)
2. Run a low-traffic workload (~200 RU/s)
3. Observe periodic ~1s latency spikes on SQL requests

### Root Cause

In non-trickle mode (RU abundant), the server intentionally returns `FillRate=0` — the client consumes from a granted token pool without maintaining a local rate limiter. When tokens deplete and the limiter transitions from non-trickle (`rate=0`) to needing more tokens:

1. `reserveN()` with `limit=0` → `durationFromTokens()` returns `InfDuration` → reservation fails immediately
2. `notify()` triggers an async RPC to the server (~50-100ms round trip)
3. `acquireTokens()` enters a retry loop: **20 retries × 50ms blind `time.Sleep`** = up to **1000ms**
4. The async token refresh completes in ~50-100ms via `Reconfigure()`, but the retry loop has no way to observe this — it keeps sleeping

```go
for range gc.mainCfg.WaitRetryTimes { // 20
    res = counter.limiter.Reserve(...)
    if d, err = WaitReservations(...); err == nil {
        break
    }
    time.Sleep(gc.mainCfg.WaitRetryInterval) // 50ms, blind
}
```

### Impact

- **P95/P99 latency spikes** of ~1s under low-traffic workloads with high FillRate configurations
- More pronounced with single-client setups where slot deletion happens every period

### Expected Behavior

When tokens are refreshed via `Reconfigure()` (~50-100ms), the retry loop should wake up immediately and proceed, not continue blind-sleeping for up to 1s.

JmPotato:

Issue Number: close #10251.

Cherry-pick of #10252 to release-8.5.

```commit-message
Add a close-and-recreate channel on the `Limiter` that is closed on every `Reconfigure()` call.
The retry loop in `acquireTokens` now selects on this channel instead of blind-sleeping,
waking up immediately when new tokens arrive. The timer serves as a fallback to preserve original
behavior when no `Reconfigure` occurs.
```

### Check List

Tests

- Unit test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Performance Improvements**
  * Token acquisition retry mechanism now responds immediately to configuration changes instead of waiting for full intervals, reducing latency during system reconfigurations.
  * Enhanced recovery from resource throttling with faster adaptation to configuration updates.

### DIFF
--- a/client/resource_group/controller/controller.go
+++ b/client/resource_group/controller/controller.go
@@ -1343,8 +1343,9 @@ func (gc *groupCostController) acquireTokens(ctx context.Context, delta *rmpb.Co
 	gc.metrics.runningKVRequestCounter.Inc()
 	defer gc.metrics.runningKVRequestCounter.Dec()
 	var (
-		err error
-		d   time.Duration
+		err            error
+		d              time.Duration
+		reconfiguredCh <-chan struct{}
 	)
 retryLoop:
 	for range gc.mainCfg.WaitRetryTimes {
@@ -1363,6 +1364,7 @@ retryLoop:
 		case rmpb.GroupMode_RUMode:
 			res := make([]*Reservation, 0, len(requestUnitLimitTypeList))
 			for typ, counter := range gc.run.requestUnitTokens {
+				reconfiguredCh = counter.limiter.GetReconfiguredCh()
 				if v := getRUValueFromConsumption(delta, typ); v > 0 {
 					// record the consume token histogram if enable controller debug mode.
 					if enableControllerTraceLog.Load() {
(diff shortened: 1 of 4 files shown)


okJiang (approved):

client/resource_group/controller/controller.go
```diff
@@ -1363,6 +1364,7 @@ retryLoop:
 		case rmpb.GroupMode_RUMode:
 			res := make([]*Reservation, 0, len(requestUnitLimitTypeList))
 			for typ, counter := range gc.run.requestUnitTokens {
+				reconfiguredCh = counter.limiter.GetReconfiguredCh()
```
bufferflies, line 1367: Should we move it to line 1365 if no only one type ?

client/resource_group/controller/controller.go
```diff
@@ -1381,8 +1383,28 @@ retryLoop:
...
+			*waitDuration += time.Since(waitStart)
+			return d, ctx.Err()
+		case <-reconfiguredCh:
+			if !waitTimer.Stop() {
+				select {
+				case <-waitTimer.C:
+				default:
+				}
+			}
+		case <-waitTimer.C:
```
bufferflies, line 1405: Do we need to stop the timer here?

client/resource_group/controller/controller.go
JmPotato: Although there is technically only one type at the moment, I kept it within a for loop to avoid code conflicts or confusing semantics.

client/resource_group/controller/controller.go
JmPotato: For a `time.Timer` that has already triggered, there is no need to stop it.

bufferflies:
@coderabbitai full review

rleungx (approved):

PR #10510, opened 2026-03-30
[cherry-pick] pd: report hot read cpu in heartbeat

### LINKED ISSUE #5718: scheduler: introduce read cpu dimension
## Enhancement Task

After 6.4, the hot region scheduler already can solve qps and byte conflict and make them balanced as possible.

But there are still some cases with unbalanced tikv cpu, although qps and byte are balanced enough.

Case1:
The performance of the machines in the cluster is different, for example #4612

Case2:
There is a huge cpu gap with different query, for example #4513

lhy1024:

Issue Number: Close #5718

image

#### Simple description

This pr introduces cpu as a new dimension for hot scheduler, it only serve hot read scheduler

From store heartbeat cpu_usages, we can get unfied read pool cpu for schdule. Read priorities become cpu→byte when supported, otherwise fall back to query→byte (or byte→key if query isn’t supported).

```commit-message
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * CPU-based metrics for hot-region scheduling; API now exposes per-store CPU read rates (`cpu-read-rate`)
  * Scheduler config adds CPU thresholds and tuning (`min-hot-cpu-rate`, `cpu-rate-rank-step-ratio`)

* **Enhancements**
  * Hot-region history and statistics now include CPU flow, per-store and total CPU rates
  * Grafana dashboard updated with a "Store read cpu" panel

* **Chores**
  * Version gating for CPU support (cluster version >= 8.5.7)

### DIFF
--- a/pkg/schedule/schedulers/hot_region.go
+++ b/pkg/schedule/schedulers/hot_region.go
@@ -50,6 +50,7 @@ const (
 	splitProgressiveRank    = 5
 	minHotScheduleInterval  = time.Second
 	maxHotScheduleInterval  = 20 * time.Second
+	defaultPendingWeight    = 1.0
 	defaultPendingAmpFactor = 2.0
 	defaultStddevThreshold  = 0.1
 	defaultTopnPosition     = 10
@@ -153,10 +154,25 @@ func (s *baseHotScheduler) updateHistoryLoadConfig(sampleDuration, sampleInterva
 	s.stHistoryLoads = s.stHistoryLoads.UpdateConfig(sampleDuration, sampleInterval)
 }
 
+func (s *baseHotScheduler) getEffectivePendingWeight() float64 {
+	if conf, ok := s.conf.(*hotRegionSchedulerConfig); ok {
+		pendingWeight := conf.getPendingWeight()
+		if pendingWeight < defaultPendingWeight {
+			log.Warn("pending-weight is less than default, fallback to default",
+				zap.Float64("pendingWeight", pendingWeight),
+				zap.Float64("defaultPendingWeight", defaultPendingWeight))
+			return defaultPendingWeight
+		}
+		return pendingWeight
+	}
+	return defaultPendingWeight
+}
+
 // summaryPendingInfluence calculate the summary of pending Influence for each store
 // and clean the region from regionInfluence if they have ended operator.
 // It makes each dim rate or count become `weight` times to the origin value.
 func (s *baseHotScheduler) summaryPendingInfluence(storeInfos map[uint64]*statistics.StoreSummaryInfo) {
+	pendingWeight := s.getEffectivePendingWeight()
 	for id, p := range s.regionPendings {
 		for _, from := range p.froms {
 			from := storeInfos[from]
@@ -169,6 +185,7 @@ func (s *baseHotScheduler) summaryPendingInfluence(storeInfos map[uint64]*statis
 				continue
 			}
 
+			weight *= pendingWeight
 			if from != nil && weight > 0 {
 				from.AddInfluence(&p.origin, -weight)
 			}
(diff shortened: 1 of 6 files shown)


okJiang (approved):

client/http/types.go
```diff
@@ -217,6 +217,7 @@ type HistoryHotRegion struct {
 	HotRegionType string  `json:"hot_region_type"`
 	HotDegree     int64   `json:"hot_degree"`
 	FlowBytes     float64 `json:"flow_bytes"`
+	FlowCPU       float64 `json:"flow_cpu"`
```
rleungx, line 223: This updates the history hot-region type, but the typed hot-region API structs in this file still do not expose the new CPU fields returned by the server (`store_cpu`, `total_flow_cpu`, and per-peer `flow_cpu` in the hot read/write responses). As a result, `client/http` callers will silently lose the new CPU data even though the API returns it. Could we keep the typed client structs in sync here as well?

pkg/schedule/schedulers/hot_region.go
```diff
@@ -1317,15 +1327,16 @@ func (bs *balanceSolver) buildOperators() (ops []*operator.Operator) {
...
 func (bs *balanceSolver) bucketFirstStat() utils.RegionStatKind {
 	base := utils.RegionReadBytes
 	if bs.rwTy == utils.Write {
 		base = utils.RegionWriteBytes
 	}
 	offset := bs.firstPriority
-	// todo: remove it if bucket's qps has been supported.
-	if bs.firstPriority == utils.QueryDim {
+	// TODO: remove it if buckets report these dimensions.
+	if bs.firstPriority == utils.QueryDim || bs.firstPriority == utils.CPUDim {
```
rleungx: This fallback can still produce an invalid bucket dimension when read priorities are configured as `["query", "cpu"]`. In that case `bucketFirstStat()` falls back from `query` to `cpu`, but bucket loads only contain byte/key/query dimensions, so `splitBucketsByLoad()` can index past the end of `stat.Loads`. It looks like we should either forbid `cpu` here as a bucket fallback target, or fall back to a bucket-supported dimension only.

pkg/statistics/store_hot_peers_infos.go
```diff
@@ -257,6 +260,8 @@ func summaryStoresLoadByEngine(
 		hotPeerSummary.WithLabelValues(ty, engine).Set(stddevLoads[utils.KeyDim])
 		ty = "stddev-query-rate-" + rwTy.String() + "-" + kind.String()
 		hotPeerSummary.WithLabelValues(ty, engine).Set(stddevLoads[utils.QueryDim])
+		ty = "stddev-cpu-rate-" + rwTy.String() + "-" + kind.String()
```
rleungx, line 268: `stddevLoads[i]` is normalized by `expectLoads[i]` above, so when the expected CPU load is `0` this can become `NaN` or `Inf`. That value then feeds both the exported metric and the scheduler uniform-store checks. Since CPU is optional and can legitimately be all-zero here, should we guard the division and define a stable zero-stddev behavior for the zero-mean case?

pkg/schedule/schedulers/hot_region_config.go
```diff
@@ -524,20 +546,23 @@ func getWritePeerPriorities(c *prioritiesConfig) []string {
...
-	if !querySupport && withQuery {
+	if !querySupport && (withQuery || withCPU) {
 		return compatibles
 	}
 
 	defaults := getPriorities(&defaultPrioritiesConfig)
 	isLegal := slice.AllOf(origins, func(i int) bool {
-		return origins[i] == utils.BytePriority || origins[i] == utils.KeyPriority || origins[i] == utils.QueryPriority
+		return origins[i] == utils.BytePriority || origins[i] == utils.KeyPriority || origins[i] == utils.QueryPriority || origins[i] == utils.CPUPriority
 	})
```
lhy1024, line 578: I remove adjust dim by version here. Because we cannot distinguish the difference v8.5.5 and v8.5.5-patch.

Issue #10516, opened 2026-03-31
refactor(pd): merge cse-pd into pd repository


bufferflies:
## Summary

Track the `release-8.5-keyspace` commit batch that has been split into 41 subtasks by functional area.

## Branch

- target branch: `release-8.5-keyspace`
- origin repo: `bufferflies/pd`

### mTLS / pd-ctl
- [ ] #10531
- [x] #10561
- [x] #10532

### Keyspace / GC
- [ ] #10534
- [x] #10566
- [ ] [`78f39c6d`](https://github.com/bufferflies/pd/commit/78f39c6d19acd31e338adf6d4a363db25bbb9977) support new keyspace enable gc v2
- [ ] [`6326f2d5`](https://github.com/bufferflies/pd/commit/6326f2d511aa011c5b44ca217380ebfc2d237212) fix global service safe point
- [ ] [`f88a4fd9`](https://github.com/bufferflies/pd/commit/f88a4fd923fe5e3724e0d1bea9f03ab4a0b250b8) fix `enable-global-safe-point-v2` init
- [ ] [`76277a6b`](https://github.com/bufferflies/pd/commit/76277a6b45d404313da08a71a9d131f341e287e2) create keyspace metrics and logs

### Meta Service Group
- [ ] #10596
- [ ] [`0b70d5de`](https://github.com/bufferflies/pd/commit/0b70d5de2949142d3071e066321eeee0907fa6dd) change add method to patch
- [ ] [`e1d53f36`](https://github.com/bufferflies/pd/commit/e1d53f365379c3057d5507a06e10b44662f4c374) add switch for auto assign keyspace to meta service group
- [ ] [`5be7eae7`](https://github.com/bufferflies/pd/commit/5be7eae7a2868f10ef8a2c4921b841d5cbbe06a4) add meta-service-group status
- [ ] [`d473bf9d`](https://github.com/bufferflies/pd/commit/d473bf9dbbdd1bfe2a096fc908ce0c4cf926bd85) add meta service group fallback ratio
- [ ] [`60279efd`](https://github.com/bufferflies/pd/commit/60279efd68c61fc61ff470e3b4f902f45dfaf304) unassign keyspace from meta service group when setting tombstone
- [ ] [`2baf2a50`](https://github.com/bufferflies/pd/commit/2baf2a502b864633cd6198a90094033fd28caff4) cache meta-service group status

### API / Observability
[...]

PR #10535, opened 2026-04-01
[release-8.5-keyspace] add option to skip rawkv region bound

### LINKED ISSUE #10534: 0c88ce89 add option to skip rawkv region bound

bufferflies:

Issue Number: ref #10516,close #10534

author: @zeminzhou

cp [`0c88ce89`](https://github.com/bufferflies/pd/commit/0c88ce8921caabfa532e4a2a259bf39f7a49bdd4)

Cherry-pick `0c88ce8921caabfa532e4a2a259bf39f7a49bdd4` onto `master` and adapt the keyspace region-bound path to the current code structure.

This follow-up removes the temporary `disable-raw-kv-region-split` switch and derives the bound type directly from `key-type`:
- `key-type=table` -> use txn keyspace bounds
- `key-type=raw` -> use raw keyspace bounds
- `key-type=txn` -> use raw keyspace bounds

The exported helper surface stays compatible, while the keyspace manager now routes label-rule generation and bound checks through the internal key-type mapping.

### Check List

Tests

- Unit test

### Release note

```release-note
Use `key-type` to derive keyspace region bounds when creating keyspaces. `table` uses txn bounds, while `raw` and `txn` use raw bounds.
```

## Summary by CodeRabbit

* **Refactor**
  * Enhanced internal region bound handling to support multiple bound type configurations.
  * Improved region-split operation flow with refined completion tracking.
  * Expanded test coverage for region bound type mappings and label rule generation across different scenarios.


bufferflies:
Review note:

The new `DisableRawKVRegionSplit` field is tagged as `json:"disable-raw-kv-region-split,string"` in `server/config/config.go`. `KeyspaceConfig` is exported via the config HTTP API, so this makes the field serialize as a JSON string (`"true"` / `"false"`) instead of a JSON boolean, unlike the sibling `wait-region-split` field. That changes the API shape for config consumers and is likely to break scripts or clients that expect a boolean.

Please change the tag to `json:"disable-raw-kv-region-split"` unless there is a specific compatibility requirement for string-encoded booleans here.

bufferflies:
Re-review update:

I rechecked the latest delta and the previous finding is resolved.

- `DisableRawKVRegionSplit` now uses `json:"disable-raw-kv-region-split"`, so the config HTTP API keeps this field as a JSON boolean instead of a string.
- The new regression test in `server/config/config_test.go` exercises the JSON encoding path and verifies that `keyspace.disable-raw-kv-region-split` is emitted as a bool.

I do not have additional findings on this update. The remaining merge gate is the required `make check` pass mentioned in the task thread.

bufferflies:
Review note:

`makeKeyRanges` should stay public and keep returning `[]any`. In the current patch it was changed from `MakeKeyRanges(id uint32) []any` to `makeKeyRanges(id uint32, skipRaw bool) any`, which unnecessarily regresses both API visibility and type precision.

Even if current call sites are internal, this helper was previously exported and had a concrete return type. Please keep the exported helper and preserve the `[]any` return type, then layer the new `skipRaw` behavior without widening the signature to `any`.

bufferflies:
Review note:

This is only a partial compatibility fix. `MakeKeyRanges` is public again and the return type is back to `[]any`, but its exported signature still changed from `MakeKeyRanges(id uint32) []any` to `MakeKeyRanges(id uint32, skipRaw bool) []any`.

That is still a breaking API change for any existing caller that uses the old one-argument form. If compatibility is the goal, the old exported signature should remain available. A safer shape would be to keep `MakeKeyRanges(id uint32) []any` with the original behavior and add a new helper for the skip-raw variant (or keep the skip-raw helper internal).

bufferflies:
Review note:

There is still another compatibility regression in the same area: `MakeLabelRule` was also changed from an exported helper to the private `makeLabelRule`.

Even though the current discussion focused on `MakeKeyRanges`, this has the same compatibility risk pattern: a previously exported helper became private as part of the refactor. If compatibility is the goal, `MakeLabelRule` should also keep its exported surface, or there needs to be a very explicit decision that this API break is acceptable.

bufferflies:
Re-review update:

I rechecked the latest delta and the `MakeLabelRule` compatibility issue is resolved.

- `MakeLabelRule(id uint32)` is exported again.
- The skip-raw behavior stays behind internal helpers instead of changing the exported helper signature.
- This is aligned with the compatibility expectation discussed in the thread.

I do not have an additional compatibility finding on this specific follow-up change.

pkg/keyspace/keyspace.go
```diff
@@ -576,7 +577,15 @@ func (manager *Manager) waitKeyspaceRegionSplit(id uint32) error {
 
 // CheckKeyspaceRegionBound checks whether the keyspace region has been split.
 func (manager *Manager) CheckKeyspaceRegionBound(id uint32) bool {
```
lhy1024: If a keyspace created with `disable-raw-kv-region-split=true` only has txn bounds. If the config is later changed to `false`, it may never pass `CheckKeyspaceRegionBound` again and can become effectively invisible to `LoadKeyspace`.

bufferflies: Yes, the existing keyspace isn't affected by the configuration by design, and later add notes on the user docs.

AmoebaProtozoa (approved):

pkg/keyspace/keyspace.go
lhy1024: If this is intended to be immutable after keyspace creation, can we enforce that in code (for example, reject dynamic updates for this field)?

bufferflies: good catch, I will create new issue to  follow it https://github.com/tikv/pd/issues/10572

lhy1024: Do we need to add TODO here?

lhy1024 (approved):
the rest LGTM

server/config/config.go
```diff
@@ -245,6 +245,7 @@ const (
 	defaultGCTunerThreshold           = 0.6
 	minGCTunerThreshold               = 0
 	maxGCTunerThreshold               = 0.9
+	defaultDisableRawKVRegionSplit    = false
```
rleungx: Should we change to enablexxx?

bufferflies: It may introduce configuration compatibility issues.

rleungx: But usually we use enablexxx for a switch.

pkg/keyspace/keyspace.go
```diff
@@ -576,11 +577,31 @@ func (manager *Manager) waitKeyspaceRegionSplit(id uint32) error {
...
+	default:
+		return manager.checkBound(regionBound.RawLeftBound) &&
+			manager.checkBound(regionBound.RawRightBound) &&
+			manager.checkBound(regionBound.TxnLeftBound) &&
+			manager.checkBound(regionBound.TxnRightBound)
+	}
+}
+
+func (manager *Manager) getRegionBoundType() regionBoundType {
+	if manager.cluster == nil || manager.cluster.GetSharedConfig() == nil {
```
bufferflies, line 601: This fallback looks risky. If `cluster` exists but `GetSharedConfig()` is temporarily nil / unavailable, we now silently fall back to `table -> txn bound`, which changes the old behavior from "check all bounds" to `txn-only`. That can misclassify non-table deployments and generate the wrong label rule / split check before shared config is ready. Please either keep an explicit "unknown -> allRegionBound" fallback here, or make the caller fail closed until the key type is available instead of defaulting to `Table`.

bufferflies:
Re-review update:\n\nI rechecked the latest follow-up at `68f7c909355f4731a06c43067e524be9daebc172`.\n\n- `getRegionBoundType()` now falls back to `allRegionBound` when shared config is unavailable, so the previous `table -> txn-only` misclassification risk is removed.\n- The new regression test in `server/config/config_test.go` also covers legacy `disable-raw-kv-region-split` input for both JSON and TOML unmarshal paths and confirms it does not panic after removing the field from the config struct.\n\nI do not have a new finding on this follow-up delta.

server/config/config.go
bufferflies: done, use the exist configuration key-type, removed this configuration

AmoebaProtozoa (approved):

pkg/core/constant/kind.go
```diff
@@ -151,6 +152,8 @@ func StringToKeyType(input string) KeyType {
 		return Raw
 	case Txn.String():
 		return Txn
+	case "":
```
rleungx, line 155: Why do we need it?

pkg/keyspace/keyspace.go
```diff
@@ -506,7 +505,8 @@ func (manager *Manager) splitKeyspaceRegion(id uint32, waitRegionSplit bool) (er
 	})
 
 	start := time.Now()
-	keyspaceRule := MakeLabelRule(id)
+	boundType := manager.getRegionBoundType()
```
rleungx, line 508: Do we need to prevent dynamically changing the keytype?

pkg/core/constant/kind.go
bufferflies: In the unit test.

pkg/keyspace/keyspace.go
bufferflies: Yes, later we should persist the key type when we create the keyspace.

pkg/core/constant/kind.go
rleungx: Maybe we shouldn't support an empty string.

pkg/keyspace/keyspace.go
rleungx: Currently, we can change it through pd-ctl.

bufferflies: Yes, it only affects the later new keyspace.

pkg/keyspace/util_test.go
```diff
@@ -194,12 +198,82 @@ func TestMakeLabelRule(t *testing.T) {
...
-		re.Equal(testCase.expectedLabelRule, MakeLabelRule(testCase.id))
+		re.Equal(testCase.expectedLabelRule, buildLabelRule(testCase.id, testCase.boundType))
 	}
 }
 
+func TestKeyTypeToRegionBoundType(t *testing.T) {
+	re := require.New(t)
+	re.Equal(txnRegionBound, keyTypeToRegionBoundType(coreconstant.Table))
+	re.Equal(rawRegionBound, keyTypeToRegionBoundType(coreconstant.Raw))
+	re.Equal(rawRegionBound, keyTypeToRegionBoundType(coreconstant.Txn))
```
rleungx, line 274: why?

pkg/keyspace/util.go
```diff
@@ -112,6 +113,37 @@ type RegionBound struct {
...
+	allRegionBound regionBoundType = iota
+	rawRegionBound
+	txnRegionBound
+)
+
+// String returns the string representation of the regionBoundType.
+func (t regionBoundType) String() string {
+	switch t {
+	case allRegionBound:
+		return "all"
```
rleungx: What does "all" here mean? mixed?

bufferflies: it means keyspace region bound should include raw and txn range

pkg/core/constant/kind.go
bufferflies: Yes, other string should be treated as a table type.

pkg/keyspace/keyspace.go
bufferflies: Not currently, pd-ctl can only set the txn region labeler. If you want to set a raw region labeler, I can implement it later .

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -560,16 +560,6 @@ func setPlacementCommandFunc(cmd *cobra.Command, args []string) {
 		Index:    100,
 		Override: true,
 		Rules: []*pd.Rule{
-			{
-				GroupID: groupID,
-				ID:      fmt.Sprintf("keyspace-%d-rule", keyspaceID),
-				Role:    pd.Voter,
-				// TODO: make replica count configurable
-				Count:            3,
-				StartKeyHex:      keyRanges[0].(map[string]any)["start_key"].(string),
```
rleungx, line 569: Removing this rule leaves MakeKeyRanges with only one returned range, but the remaining txn rule still reads keyRanges[1]. This currently panics in Tools Test with index out of range.

pkg/keyspace/keyspace.go
```diff
@@ -576,11 +577,23 @@ func (manager *Manager) waitKeyspaceRegionSplit(id uint32) error {
 
 // CheckKeyspaceRegionBound checks whether the keyspace region has been split.
 func (manager *Manager) CheckKeyspaceRegionBound(id uint32) bool {
+	return manager.hasKeyspaceRegionBound(id, manager.getRegionBoundType())
```
rleungx: This uses the current key-type, while the label rule was created with the key-type at keyspace creation time. If pd-server.key-type is changed later, an existing keyspace can fail this check and LoadKeyspace will return not found.

pkg/core/constant/kind.go
```diff
@@ -143,17 +143,15 @@ func (k KeyType) String() string {
...
+	if input == Raw.String() {
 		return Raw
-	case Txn.String():
+	}
+	if input == Txn.String() {
 		return Txn
-	default:
-		panic("invalid key type: " + input)
 	}
+	return Table
```
rleungx: This silently maps any invalid non-empty key type to table. Please keep the empty-string default if needed, but unknown values should still be rejected instead of being treated as table.

pkg/keyspace/util.go
```diff
@@ -166,17 +191,26 @@ func MakeRegionBound(id uint32) *RegionBound {
 }
 
 // MakeKeyRanges encodes keyspace ID to correct LabelRule data.
+// only for test use, as it always generates txn region bound.
 func MakeKeyRanges(id uint32) []any {
+	return buildKeyRanges(id, txnRegionBound)
```
rleungx, line 209: This helper is still used by pd-ctl set-placement, so hardcoding txn here means that command cannot generate a raw keyspace placement range.

pkg/core/constant/kind.go
bufferflies: done

pkg/keyspace/keyspace.go
bufferflies: done， persist key-type value  in keyspace config, so check the region bound will base on the keyspace configuration not on the current server config.

pkg/keyspace/keyspace.go
```diff
@@ -575,12 +580,31 @@ func (manager *Manager) waitKeyspaceRegionSplit(id uint32) error {
 }
 
 // CheckKeyspaceRegionBound checks whether the keyspace region has been split.
-func (manager *Manager) CheckKeyspaceRegionBound(id uint32) bool {
+func (manager *Manager) CheckKeyspaceRegionBound(meta *keyspacepb.KeyspaceMeta) bool {
+	config := meta.GetConfig()
+	val, ok := config[RegionBoundType]
+	// if config does not contain region bound type, we use the default one from manager.
```
rleungx, line 665: This fallback still depends on the current key-type for metas without region_bound_type. CreateKeyspaceByID and reserved keyspaces do not persist this field, so a later key-type change can make LoadKeyspace check the wrong bounds.

pkg/keyspace/keyspace.go
```diff
@@ -68,6 +67,9 @@ const (
 	// UnifiedGC is a type of gc_management_type used to indicate that the GC states of this keyspace is managed
 	// in a unified way (managed by the NullKeyspace).
 	UnifiedGC = "unified"
+	// RegionBoundType is the key for region bound type in keyspace config,
+	// which is used to indicate the region bound type of the keyspace.
+	RegionBoundType = "region_bound_type"
 )
```
rleungx, line 77: Since this value is used for correctness, it should not be mutable through the normal keyspace config API. Please protect region_bound_type from update/delete mutations, or store it outside the user-editable Config map.

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -562,24 +562,14 @@ func setPlacementCommandFunc(cmd *cobra.Command, args []string) {
 		Rules: []*pd.Rule{
 			{
 				GroupID: groupID,
-				ID:      fmt.Sprintf("keyspace-%d-rule", keyspaceID),
+				ID:      fmt.Sprintf("keyspace-%d-rule-txn", keyspaceID),
```
rleungx: set-placement still uses MakeKeyRanges, which always returns a txn range now. For raw keyspaces this placement rule targets the wrong range; it should use the keyspace's persisted bound type or expose a raw option.

bufferflies: Currently， we just need to support the txn scenario. I added more description about this command just for the txn region. If we want to support the raw scenario, I will implement it later.

pkg/keyspace/keyspace.go
bufferflies: Yes, simply forbid updating this configuration such as keyspace level gc.

pkg/keyspace/keyspace.go
bufferflies: Good catch, add this configuration  in the bootstrap step.

pkg/keyspace/util.go
bufferflies: rename as MakeTxnKeyRanges

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -551,8 +551,7 @@ func setPlacementCommandFunc(cmd *cobra.Command, args []string) {
 	}
 
 	// Generate key ranges for the keyspace
-	keyRanges := keyspace.MakeKeyRanges(keyspaceID32)
-
+	keyRanges := keyspace.MakeTxnKeyRanges(keyspaceID32)
```
rleungx: For raw keyspaces this always creates a txn range rule, but the keyspace is split on raw bounds. Please derive the persisted bound type or add a raw mode so placement covers the actual range.

tools/pd-ctl/pdctl/command/keyspace_command_test.go
```diff
@@ -28,68 +28,47 @@ func TestMakeKeyRanges(t *testing.T) {
...
 			keyspaceID:       constant.MaxValidKeyspaceID,
-			expectedRawStart: "72ffffff00000000fb",
-			expectedRawEnd:   "7300000000000000fb",
 			expectedTxnStart: "78ffffff00000000fb",
 			expectedTxnEnd:   "7900000000000000fb",
 		},
 	}
 
 	for _, tc := range testCases {
 		ranges := keyspace.MakeKeyRanges(tc.keyspaceID)
```
rleungx: This helper was renamed to `MakeTxnKeyRanges`, so this package no longer compiles. Please update this call or keep a compatibility wrapper.

tools/pd-ctl/pdctl/command/keyspace_command.go
rleungx: This regresses raw-key clusters. `set-placement` now always calls `MakeTxnKeyRanges`, but keyspace creation in this PR can persist/use `region_bound_type=raw` when PD is running with `key-type=raw`. In that mode the command succeeds but installs a rule for the `x...` txn range, while the actual keyspace regions are split and labeled under the `r...` raw range, so the placement rule will not match those regions. Please derive the range from the keyspace metadata persisted at creation time, or expose/use a raw-aware option here instead of always generating txn bounds.

pkg/keyspace/util.go
```diff
@@ -187,6 +228,10 @@ func getRegionLabelID(id uint32) string {
 
 // MakeLabelRule makes the label rule for the given keyspace id.
 func MakeLabelRule(id uint32) *labeler.LabelRule {
+	return buildLabelRule(id, txnRegionBound)
```
rleungx, line 237: This changes the exported `MakeLabelRule` behavior from raw+txn bounds to txn-only. Please keep the old exported behavior for compatibility and use the bound-specific builder only internally.

bufferflies: done， it's used only for test and renamed as MakeTxnLabelRule to tell the caller.

tools/pd-ctl/pdctl/command/keyspace_command.go
bufferflies: fixed , it will get current keytype from pd server and then pass this param to construct key range rule

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -550,9 +551,16 @@ func setPlacementCommandFunc(cmd *cobra.Command, args []string) {
 		})
 	}
 
-	// Generate key ranges for the keyspace
-	keyRanges := keyspace.MakeKeyRanges(keyspaceID32)
+	// Fetch the key-type config from PD to decide which region bound to use:
+	// raw key-type uses the raw region bound, while table/txn use the txn bound.
+	keyType, err := getKeyType(cmd)
```
rleungx: This still uses the current cluster `key-type`, but existing keyspace ranges are determined by persisted `region_bound_type`, so changing `key-type` later can make this rule target the wrong bound.

bufferflies: Yes, the configuration for the existing keyspace is wrong, but for the new keyspace is right. So I chose to add one more flag to choose the key type, not based on the keyspace meta and pd config.

tools/pd-ctl/pdctl/command/keyspace_command.go
```diff
@@ -550,9 +553,20 @@ func setPlacementCommandFunc(cmd *cobra.Command, args []string) {
...
-	keyRanges := keyspace.MakeKeyRanges(keyspaceID32)
+	raw, err := cmd.Flags().GetBool("raw")
+	if err != nil {
+		cmd.PrintErrf("Failed to get raw flag: %v\n", err)
+		return
+	}
 
+	// Generate key ranges for the keyspace. Use the raw region bound when --raw
+	// is set, otherwise the txn region bound.
+	keyRanges := keyspace.MakeTxnKeyRanges(keyspaceID32)
```
rleungx: This changes the default `set-placement` behavior from covering both raw and txn ranges to covering only the txn range. Existing raw-keyspace usages can still report success while the installed rule does not match the raw regions.

pkg/keyspace/util.go
```diff
@@ -168,18 +200,33 @@ func MakeRegionBound(id uint32) *RegionBound {
 	}
 }
 
-// MakeKeyRanges encodes keyspace ID to correct LabelRule data.
-func MakeKeyRanges(id uint32) []any {
+// MakeTxnKeyRanges encodes keyspace ID to correct LabelRule data with the txn
+// region bound. Used by tests and pd-ctl.
+func MakeTxnKeyRanges(id uint32) []any {
```
rleungx: This removes the previously exported `MakeKeyRanges` / `MakeLabelRule` helper surface from `pkg/keyspace`. Out-of-package callers using the old signatures will fail to compile.

tools/pd-ctl/pdctl/command/keyspace_command.go
bufferflies: fixed, add one more flag to create raw region labeler

pkg/keyspace/util.go
bufferflies: It's used for tools and no any out of package use it.

rleungx (approved):

PR #10540, opened 2026-04-01
client: add backoff mechanism for service URL updates

### LINKED ISSUE #10539: 6e2a4d20 add backoff mechanism for service URL updates

bufferflies:
## Summary
- add a backoff interval before repeated RM service URL refreshes triggered by error paths
- keep the wait context-aware so shutdown can still exit the update loop promptly

## Issue Number
Issue Number: ref #10516, close #10539

## Validation
- `cd client && go test . -run 'Test.*ResourceManager.*' -count=1`\n- `cd client && go test . -run 'TestTryResourceManagerConnectUsesRMForTokenAndFallbackToPD' -count=1`\n- `make check`\n\n## Release Note\n```release-note\nNone\n```

## Summary by CodeRabbit

* **Bug Fixes**
  * Introduced a backoff to service discovery updates so repeated update signals respect a minimum interval, reducing rapid repeat updates and improving stability and logging around wait durations.

* **Tests**
  * Added a test that validates the update backoff behavior, ensures a delayed second update, and confirms the update loop exits promptly when canceled.

### DIFF
--- a/client/servicediscovery/resource_manager_service_discovery.go
+++ b/client/servicediscovery/resource_manager_service_discovery.go
@@ -263,7 +263,7 @@ func (r *ResourceManagerDiscovery) updateServiceURLLoop(revision int64) {
 				}
 			}
 			lastUpdateTime = time.Now()
-			log.Info("[resource-manager] updating service URL", zap.String("old-url", r.serviceURL))
+			log.Info("[resource-manager] updating service URL", zap.String("old-url", r.GetServiceURL()))
 			discoverAndUpdate()
 		}
 	}


bufferflies:
Re-reviewed the change on top of commit `31c1608aa15de85a7ca2a1784e1b5407c8c22c31`.

I do not have a new code finding on this delta. The backoff is confined to the error-triggered refresh path, and the wait stays context-aware so shutdown can still break out promptly.

Residual risk: I do not see a targeted regression test for the new backoff behavior itself in this PR, so the current validation still relies on the existing RM discovery test surface plus `make check`.

client/servicediscovery/resource_manager_service_discovery_test.go
```diff
@@ -0,0 +1,111 @@
...
+		defer close(done)
+		discovery.updateServiceURLLoop(0)
+	}()
+
+	discovery.ScheduleUpdateServiceURL()
+	require.Eventually(t, func() bool {
+		return len(metaCli.snapshotCallTimes()) >= 1
+	}, time.Second, 10*time.Millisecond)
+
+	discovery.ScheduleUpdateServiceURL()
```
bufferflies, line 102: This test now depends on the production backoff interval (`3s`) and waits for a real wall-clock window before it can pass. That makes the unit test suite slower and also more timing-sensitive in CI. Please consider making the backoff interval overrideable in tests (for example via a package-level variable or injected option), then set it to a much smaller duration here so the regression stays fast and stable.

bufferflies:
Re-reviewed the latest follow-up on top of commit `dbc6db0aa6372194908783f7012dfe341af814cc`.

The previous test finding is now addressed:
- the production default backoff remains unchanged
- the test overrides the interval locally, so it no longer waits on the real `3s` window

I do not have a new finding on this follow-up delta.

bufferflies:
Re-reviewed the latest follow-up on top of commit `dc612b5dc3811ccff02cf1c43e772a1f7d9f9e67`.

The accepted `r.serviceURL` suggestion is now addressed: the log path reads the old URL through `GetServiceURL()` instead of reaching into the protected field directly.

I do not have a new finding on this follow-up delta.

lhy1024 (approved):

PR #10543, opened 2026-04-01
client: remove meta revision

### LINKED ISSUE #10542: c3cd07c6 remove meta revision

bufferflies:
Issue Number: ref #10516, close #10542

author: @disksing

cp [`c3cd07c6`](https://github.com/bufferflies/pd/commit/c3cd07c686ba71bc6cf7ee7e70f95103274605b9)

This removes the resource-group meta revision cursor from the client watch path.

- stop loading the initial resource-group revision only for watch startup
- stop passing `WithRev(metaRevision)` when creating or recreating the watch
- stop updating `metaRevision` from each watched event

### Check List

Tests

- [x] `go test . -run "Test.*ResourceManager.*" -count=1`
- [x] `go test ./resource_group/controller -run TestDoesNotExist -count=1`
- [x] `make check`

Side effects

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Refactor**
  * Simplified resource-group metadata watcher and retry behavior by removing startup pre-load and local revision tracking. The watcher is now created without anchoring to a specific revision, streamlining initialization and retry flows.


client/resource_group/controller/global_controller.go
```diff
@@ -363,7 +359,7 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 				if !c.ruConfig.isSingleGroupByKeyspace && watchMetaChannel == nil {
 					// Use WithPrevKV() to get the previous key-value pair when get Delete Event.
 					prefix := pd.GroupSettingsPathPrefixBytes(c.keyspaceID)
-					watchMetaChannel, err = c.provider.Watch(ctx, prefix, opt.WithRev(metaRevision), opt.WithPrefix(), opt.WithPrevKV())
+					watchMetaChannel, err = c.provider.Watch(ctx, prefix, opt.WithPrefix(), opt.WithPrevKV())
```
bufferflies, line 362: This changes the reconnect semantics of the resource-group meta watch. Before this PR, the controller resumed from the last processed `metaRevision`, so updates that happened while the watch stream was broken could still be replayed. After removing `WithRev(metaRevision)`, a re-created watch starts from "now", which can silently skip PUT/DELETE events that landed during the disconnect window. That means the local controller cache can diverge from RM state after a transient watch failure. Please keep a resume revision (or reload a fresh snapshot before recreating the watch) so reconnects do not lose intermediate resource-group changes.

bufferflies: Good catch. I updated the retry path to reload a fresh resource-group snapshot, resync the cached controllers, and recreate the watch from snapshot revision + 1 so reconnects do not skip the disconnect window. I also added `TestReloadResourceGroupMetaWatch` to cover the retry behavior.

client/resource_group/controller/global_controller.go
```diff
@@ -496,6 +489,57 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
...
+	// keep the cache in sync without replaying the snapshot itself as watch events.
+	return c.provider.Watch(
+		ctx,
+		pd.GroupSettingsPathPrefixBytes(c.keyspaceID),
+		opt.WithRev(revision+1),
+		opt.WithPrefix(),
+		opt.WithPrevKV(),
+	)
+}
+
```
bufferflies: This snapshot sync still misses one reconnect case: groups that were created while the watch stream was down. `syncResourceGroupSnapshot` only walks `c.groupsController.Range(...)`, so it can update existing entries and tombstone deleted ones, but it never creates controllers for names that exist in the freshly loaded snapshot and are absent from the local cache. Because the recreated watch starts from `revision + 1`, those groups will not be replayed as historical PUT events either. The result is that reconnect can still leave the local cache missing newly created resource groups. Please also create controllers for snapshot-only groups before resuming the watch.

bufferflies: Fixed. `syncResourceGroupSnapshot()` now also creates local controllers for groups that only exist in the reloaded snapshot, so reconnect can recover both updated existing groups and groups created during the disconnect window. I extended `TestReloadResourceGroupMetaWatch` to cover the new-group case and reran the targeted tests plus `make check`.

bufferflies:
Re-reviewed the latest follow-up on top of commit `64409c56374adbb6092ceb6e0e16052da1093706`.

The previous reconnect/snapshot finding is now addressed:
- snapshot sync still updates existing groups
- snapshot sync still tombstones groups that disappeared from the fresh snapshot
- snapshot-only groups are now also created in the local controller cache before the watch resumes from `revision + 1`

I do not have a new finding on this follow-up delta.

client/resource_group/controller/global_controller.go
```diff
@@ -335,7 +331,7 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 		if !c.ruConfig.isSingleGroupByKeyspace {
 			// Use WithPrevKV() to get the previous key-value pair when get Delete Event.
 			prefix := pd.GroupSettingsPathPrefixBytes(c.keyspaceID)
-			watchMetaChannel, err = c.provider.Watch(ctx, prefix, opt.WithRev(metaRevision), opt.WithPrefix(), opt.WithPrevKV())
+			watchMetaChannel, err = c.provider.Watch(ctx, prefix, opt.WithPrefix(), opt.WithPrevKV())
```
lhy1024, line 334: The initial watch here no longer uses a revision, which leaves a startup window. If a resource group changes after the controller starts but before the watch is actually established, the local cache can miss that update. The reconnect path already rebuilds from a snapshot/revision, but the initial startup path still needs the same barrier.

client/resource_group/controller/global_controller.go
```diff
@@ -496,6 +489,72 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
...
+			return true
+		}
+		if c.groupsController.CompareAndSwap(name, gc, newGC) {
+			log.Info("[resource group controller] re-create resource group cost controller from snapshot",
+				zap.String("name", name))
+		}
+		return true
+	})
+
+	for name, group := range groupMap {
```
lhy1024: This creates controllers for all groups that only appear in the reloaded snapshot. That means even groups this client has never used may later be treated as reportable and send empty token requests. It seems safer to keep the lazy-loading behavior here and only reconcile groups that already exist locally, or groups that are being restored from tombstone.

client/resource_group/controller/global_controller.go
bufferflies: Yes, but we just need the latest event and can ignore all the mvvc events

client/resource_group/controller/global_controller.go
bufferflies, line 362: Reverting `reloadResourceGroupMetaWatch`/`syncResourceGroupSnapshot` brings back the reconnect correctness gap here. After `watchMetaChannel` breaks, this path recreates the watch from "now" again, so PUT/DELETE events that landed during the disconnect window can be skipped. That means the local resource-group cache can diverge from RM state after a transient watch failure. If we want to keep the strict-cp shape, we still need some equivalent resume barrier on retry; otherwise this is the same bug the previous follow-up was fixing.

bufferflies:
@coderabbitai full review

client/resource_group/controller/global_controller.go
bufferflies: good catch， revert this changes

client/resource_group/controller/global_controller.go
lhy1024: I agree we do not need the whole MVCC history. The issue is that without a startup barrier, we may miss the latest update itself if it happens before the watch is established and no later event arrives. In that case the local cache can stay stale indefinitely.

disksing: @lhy1024 In the previous implementation, the returned ResourceGroup was actually discarded and never stored in the local cache. Therefore, using the latest version should result in the same behavior as before.

disksing (approved):

client/resource_group/controller/global_controller.go
okJiang: It is indeed the same behavior.

Does this issue need to be fixed? Alternatively, what was the original reasoning or consideration regarding this matter? @disksing

okJiang (approved):

AmoebaProtozoa (approved):

PR #10556, opened 2026-04-02
tso:Remove archive/tombstone keyspace from group (407ad06f)

### LINKED ISSUE #10555: 407ad06f support removing archive and tombstone keyspaces from keyspace group

bufferflies:
## Summary
- add a DELETE keyspace-group API to remove archived or tombstoned keyspaces from a group
- validate keyspace state before mutating group membership and skip non-existent or already-absent keyspaces
- add API tests covering skipped, missing, and archived/tombstone removal cases

## Issue Number
Issue Number: ref #10516, close #10555

author: @ystaticy

cp [`407ad06f`](https://github.com/bufferflies/pd/commit/407ad06f1620fc06b2ccaa3315c39c76124f454a)

## Checklist
- [x] cherry-picked from upstream commit
- [x] conflict resolved against current `master`
- [x] `make check`

## Release note
```
Add a DELETE API to remove archived or tombstoned keyspaces from a keyspace group.
```


bufferflies:
Re-reviewed the latest follow-up on head `53bf48c2997a0d1e9a3664a946947c87aef93d07`.

This delta looks clean to me. The unresolved review items in the keyspace-group removal path and the related test cleanup are addressed, and I do not see a new behavior regression introduced by this follow-up.

I do not have a new finding on this delta.

bufferflies:
Re-reviewed the latest cleanup follow-up on head `14d91a39ab7241d0161f0a698062fe99f3acbe08`.

This delta looks clean to me. The failpoint cleanup in `TestRemoveKeyspacesFromGroup` is now deferred via closures, so the disable calls execute at function exit instead of at defer-registration time.

I do not have a new finding on this delta.

bufferflies:
Re-reviewed the latest follow-up on head `6b28c8c86824b1680458b9d45f0f02a20bac58fd`.

This delta looks clean to me. The group-removal path now protects bootstrap/system keyspaces via `isProtectedKeyspaceID(ksID)` instead of only skipping `constant.DefaultKeyspaceID`, which matches the requested review comment.

I do not have a new finding on this delta.

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -558,3 +560,89 @@ func parseNodeAddress(c *gin.Context) (string, error) {
...
+
+	// If no valid keyspaces to remove, load and return the group without modification
+	if len(validKeyspaces) == 0 {
+		kg, err := groupManager.GetKeyspaceGroupByID(groupID)
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+			return
+		}
+		if kg == nil {
+			c.AbortWithStatusJSON(http.StatusInternalServerError, errs.ErrKeyspaceGroupNotExists.FastGenByArgs(groupID).Error())
```
rleungx: 500？

bufferflies: Yes, it means the keyspace meta exists, but the keyspace group doesn't contain this keyspace. So the problem is caused by inconsistent caching.

rleungx: But the group ID comes from the request?

bufferflies: yes，good catch，I will return the 404 http code

rleungx (approved):

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -558,3 +560,93 @@ func parseNodeAddress(c *gin.Context) (string, error) {
...
 }
+
+// RemoveKeyspacesFromGroupParams defines the params for removing keyspaces from a keyspace group.
+type RemoveKeyspacesFromGroupParams struct {
+	Keyspaces []uint32 `json:"keyspaces"`
+}
+
+// RemoveKeyspacesFromGroup removes the specified keyspaces from the given keyspace group.
+// Keyspaces in archived or tombstone state will be removed. Keyspaces not in the group will be skipped.
+func RemoveKeyspacesFromGroup(c *gin.Context) {
```
lhy1024, line 571: `force_refresh_group_id` still assumes every keyspace belongs to some keyspace group. After this change, an archived/tombstone keyspace can legitimately belong to no group, so `GetGroupByKeyspaceID` can now return `ErrKeyspaceNotInAnyKeyspaceGroup` and this path turns it into HTTP 500. Could we handle that case explicitly instead of surfacing it as a server error?

bufferflies: LoadKeyspace will return ErrKeyspaceNotFound err not  ErrKeyspaceNotInAnyKeyspaceGroup after this keyspace has been removed. 
https://github.com/tikv/pd/blob/d87251e8c3d9ab1fed83d689363f5de67fb23a37/server/apiv2/handlers/keyspace.go#L155-L159

lhy1024: `RemoveKeyspacesFromGroup` only removes the keyspace ID from `kg.Keyspaces`; it does not delete the keyspace name mapping or the keyspace meta. Also, `LoadKeyspace(name)` only checks whether the name/id/meta exists, not whether the keyspace still belongs to any group.

So after removal, `LoadKeyspace(name)` still succeeds, and the later `force_refresh_group_id` branch calls `GetGroupByKeyspaceID(meta.GetId())`, which can now return `ErrKeyspaceNotInAnyKeyspaceGroup`. That path is still surfaced as HTTP 500.

lhy1024: I reproduced this on the PR branch locally.
```
  func (suite *keyspaceTestSuite) TestLoadKeyspaceForceRefreshGroupIDAfterRemoval() {
        re := suite.Require()

        created := MustCreateKeyspace(re, suite.server, &handlers.CreateKeyspaceParams{
                Name: "test_keyspace_force_refresh_after_removal",
        })

        km := suite.server.GetKeyspaceManager()
        re.NotNil(km)

        _, err := km.UpdateKeyspaceStateByID(created.GetId(), keyspacepb.KeyspaceState_DISABLED, 0)
        re.NoError(err)
        _, err = km.UpdateKeyspaceStateByID(created.GetId(), keyspacepb.KeyspaceState_ARCHIVED, 0)
        re.NoError(err)

        kgm := suite.server.GetServer().GetKeyspaceGroupManager()
        re.NotNil(kgm)

        _, err = kgm.RemoveKeyspacesFromGroup(constant.DefaultKeyspaceGroupID, []uint32{created.GetId()})
        re.NoError(err)

        httpReq, err := http.NewRequest(
                http.MethodGet,
                suite.server.GetAddr()+keyspacesPrefix+"/"+created.GetName(),
                http.NoBody,
        )
        re.NoError(err)
        query := httpReq.URL.Query()
        query.Set("force_refresh_group_id", "true")
        httpReq.URL.RawQuery = query.Encode()

        resp, err := tests.TestDialClient.Do(httpReq)
        re.NoError(err)
        defer resp.Body.Close()

        re.Equal(http.StatusOK, resp.StatusCode)

        loaded := &handlers.KeyspaceMeta{}
        re.NoError(json.NewDecoder(resp.Body).Decode(loaded))
        re.Equal(created.GetId(), loaded.GetId())
        re.Equal(keyspacepb.KeyspaceState_ARCHIVED, loaded.GetState())
  }
```

ystaticy: @lhy1024 Thanks for the feedback. I'll fix this.

bufferflies: I have fixed it by removing the keyspace meta together.

pkg/keyspace/keyspace.go
```diff
@@ -843,6 +843,35 @@ func (manager *Manager) UpdateKeyspaceState(name string, newState keyspacepb.Key
...
+		}
+		if meta == nil {
+			return errs.ErrKeyspaceNotFound
+		}
+		if meta.GetState() == keyspacepb.KeyspaceState_ENABLED || meta.GetState() == keyspacepb.KeyspaceState_DISABLED {
+			return errors.Errorf("cannot remove keyspace in state %s", meta.GetState().String())
+		}
+		return manager.store.RemoveKeyspace(txn, id, meta.GetName())
+	})
+	if err != nil {
```
lhy1024, line 854:
```suggestion
	if err == nil {
```
The cache should be cleared after successful deletion, not on failure.

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -558,3 +562,107 @@ func parseNodeAddress(c *gin.Context) (string, error) {
...
+	if err != nil {
+		if errs.ErrKeyspaceGroupNotExists.Equal(err) {
+			c.AbortWithStatusJSON(http.StatusNotFound, err.Error())
+			return
+		}
+		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+		return
+	}
+	failedKeyspaces := make([]uint32, 0)
+	for _, keyspaceID := range validKeyspaces {
```
lhy1024: Please have `RemoveKeyspacesFromGroup` return the actually removed IDs, and only call `RemoveKeyspace` for those IDs. `validKeyspaces` is only state-filtered from the request body; it can still include keyspaces that are not members of the target group. Since `RemoveKeyspacesFromGroup` silently skips non-members, deleting meta for every `validKeyspaces` entry can leave another group with a dangling keyspace ID.

bufferflies: Good catch, using the RemoveKeyspacesFromGroup result as the deleted keyspace avoids deleting the mismatched keyspace.

pkg/keyspace/keyspace.go
bufferflies: done

server/apiv2/handlers/tso_keyspace_group.go
```diff
@@ -558,3 +562,107 @@ func parseNodeAddress(c *gin.Context) (string, error) {
...
+	// Remove the keyspaces from the keyspace group
+	kg, removedIDs, err := groupManager.RemoveKeyspacesFromGroup(groupID, validKeyspaces)
+	if err != nil {
+		if errs.ErrKeyspaceGroupNotExists.Equal(err) {
+			c.AbortWithStatusJSON(http.StatusNotFound, err.Error())
+			return
+		}
+		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+		return
+	}
```
lhy1024: This removal flow is not atomic. `RemoveKeyspacesFromGroup` first commits the keyspace-group membership update, and only then the handler removes keyspace meta/name mappings one by one via `RemoveKeyspace`. If the second step fails due to an etcd transient error, context cancellation, or a PD crash, the API returns 500 but the group membership has already been persisted.

Retrying cannot fully recover either: on retry, those keyspaces are no longer in the group, so `removedIDs` becomes empty and `RemoveKeyspace` will ot be called again. This can leave orphaned keyspace metadata that no longer belongs to any keyspace group.

Could we put the group membership update and keyspace meta/name mapping deletion into the same etcd transaction, and update the in-memory caches only after the transaction commits successfully?

bufferflies: yes, make all txn operators into one txn to make all updates atomic.

pkg/keyspace/tso_keyspace_group.go
```diff
@@ -424,6 +424,81 @@ func (m *GroupManager) GetGroupByKeyspaceID(id uint32) (uint32, error) {
...
+// It returns the updated keyspace group and any error encountered.
+func (m *GroupManager) RemoveKeyspacesFromGroup(groupID uint32, km *Manager, keyspaceIDs []uint32) (*endpoint.KeyspaceGroup, error) {
+	m.Lock()
+	defer m.Unlock()
+
+	var (
+		kg  *endpoint.KeyspaceGroup
+		err error
+	)
+	removedIDs := make([]uint32, 0, len(keyspaceIDs))
```
lhy1024: `removedIDs` is now populated but never used. Could we remove it to avoid misleading future readers? The actual deletion is already performed while iterating `kg.Keyspaces`.

pkg/keyspace/keyspace.go
```diff
@@ -843,6 +843,28 @@ func (manager *Manager) UpdateKeyspaceState(name string, newState keyspacepb.Key
 	return meta, nil
 }
 
+// RemoveKeyspace removes the keyspace specified by id if it's in proper state and not protected.
+func (manager *Manager) RemoveKeyspace(txn kv.Txn, id uint32) error {
```
lhy1024, line 847: The current caller filters protected keyspaces before calling this method, but `RemoveKeyspace` is exported and its comment says protected keyspaces should not be removed. Could we keep an `isProtectedKeyspaceID` guard here as defensive protection for future callers?

lhy1024 (approved):
the rest LGTM

lhy1024 (approved):

pkg/keyspace/keyspace.go
```diff
@@ -843,6 +843,31 @@ func (manager *Manager) UpdateKeyspaceState(name string, newState keyspacepb.Key
...
+	}
+	if meta == nil {
+		return errs.ErrKeyspaceNotFound
+	}
+	if meta.GetState() == keyspacepb.KeyspaceState_ENABLED || meta.GetState() == keyspacepb.KeyspaceState_DISABLED {
+		return errors.Errorf("cannot remove keyspace in state %s", meta.GetState().String())
+	}
+	err = manager.store.RemoveKeyspace(txn, id, meta.GetName())
+	if err == nil {
+		manager.keyspaceNameLookup.Delete(id)
```
rleungx, line 865: This clears the in-memory cache before the enclosing txn commits. A concurrent cache miss can reload the still-existing meta and repopulate stale cache, so please invalidate these entries only after RunInTxn returns nil.

PR #10582, opened 2026-04-08
client: add pre-throttling demand RU/s metric

### LINKED ISSUE #10581: resource control: add client-side pre-throttling demand RU/s metric
### Problem

Currently, there is no metric that accurately reflects the **real-time RU/s demand** from clients **before** Resource Control throttling takes effect:

- **Client-side `avgRUPerSec`** (`group_controller.go`) is computed from `getRUValueFromConsumption()` — actual post-throttling consumption. When a resource group is throttled, requests wait in `Reserve()`, consumption slows, and `avgRUPerSec` only reflects the throttled rate.
- **Server-side `read_request_unit_max_per_sec` / `write_request_unit_max_per_sec`** are derived from `Consumption.RRU/WRU` reported by clients — also post-throttling values.
- **Server-side `sampled_request_unit_per_sec`** is based on `requiredToken` in `AcquireTokenBuckets`, which is `avgRUPerSec * targetPeriod * amplification - availableTokens` — not a clean demand rate, and lacks per-instance granularity.

This makes it impossible for operators to determine the true workload demand when Resource Control is actively throttling.

### Proposal

Add a new **client-side** Prometheus Gauge that tracks the EMA of demanded RU/s, sampled at the `acquireTokens()` entry point (before `Reserve()` throttling):

- **Metric**: `resource_manager_client_resource_group_demand_ru_per_sec{resource_group="..."}`
- **Data source**: the RU cost (`v`) passed to `acquireTokens()` in `group_controller.go`, which represents the true per-request demand before any token bucket throttling
- **Smoothing**: time-aware EMA (reuse the existing `movingAvgFactor` logic)

### Expected Usage

```promql
# Per-instance demand
resource_manager_client_resource_group_demand_ru_per_sec{instance="tidb-0", resource_group="default"}

# Cluster-wide demand for a resource group
sum(resource_manager_client_resource_group_demand_ru_per_sec) by (resource_group)

# Peak demand over time
max_over_time(sum(resource_manager_client_resource_group_demand_ru_per_sec) by (resource_group)[1h])
```

### Benefits

- **Accurate**: samples RU cost before throttling, reflects true workload demand
- **Per-instance**: client-side metric naturally carries `instance` label
- **Aggregatable**: `sum by` in Grafana for cluster-wide view
- **Rolling-upgrade friendly**: pure client-side change, no proto or PD server changes required

### Related

- Part of the observability improvements tracked in #10488

JmPotato:

Issue Number: Close #10581

```commit-message
Add a new client-side Prometheus Gauge
`resource_manager_client_resource_group_demand_ru_per_sec` that tracks the
EMA of demanded RU/s before Resource Control throttling takes effect.

The existing `avgRUPerSec` is based on post-throttling consumption: when a
request is rejected by the token bucket, its RU cost is subtracted from the
consumption counter (`onRequestWaitImpl`). This means the consumption-based
EMA underreports the true workload demand when the resource group is
actively throttled.

The new demand metric introduces a monotonically increasing `demandRUTotal`
counter that accumulates RU cost at every request entry point
(`onRequestWaitImpl`, `onResponseImpl`, `onResponseWaitImpl`,
`addRUConsumption`) and is never subtracted on throttle failure. A demand
EMA is then computed using the same `movingAvgFactor` as the consumption
EMA and flushed to the Gauge on each `updateAvgRequestResourcePerSec` tick.

This enables operators to:
- See per-instance RU demand in Grafana (natural `instance` label).
- Aggregate cluster-wide demand via `sum by (resource_group)`.
- Identify the true workload peak via `max_over_time(...)`.

Pure client-side change — no proto or PD server changes required, making it
rolling-upgrade friendly.
```

### Check List

Tests

- Unit test

### Release note

```release-note
Add a new client-side Prometheus metric `resource_manager_client_resource_group_demand_ru_per_sec` that tracks the EMA of demanded RU/s before Resource Control throttling, enabling operators to observe true workload demand per instance and across the cluster.
```

## Summary by CodeRabbit

* **New Features**
  * Added a demand RU/sec metric (EMA) to report requested resource usage before throttling.
  * Tracked total demanded RU separately from consumed RU so demand is visible even when requests are throttled; demand is recorded at request/response entry points.

* **Chores**
  * Remove per-resource-group demand metric series when a group is deleted.

* **Tests**
  * Added tests validating demand accounting, EMA updates, and capture on throttled responses.

### DIFF
--- a/client/resource_group/controller/global_controller.go
+++ b/client/resource_group/controller/global_controller.go
@@ -637,8 +637,9 @@ func (c *ResourceGroupsController) cleanUpResourceGroup() {
 				metrics.DemandRUPerSecGauge.DeleteLabelValues(resourceGroupName)
 				// TODO: clean up the remaining per-group metrics (e.g. TokenConsumedHistogram,
 				// GroupRunningKVRequestCounter, SuccessfulRequestDuration, FailedRequestCounter,
-				// ResourceGroupTokenRequestCounter, RequestRetryCounter, FailedLimitReserveDuration)
-				// which currently leak label series on resource group deletion.
+				// ResourceGroupTokenRequestCounter, RequestRetryCounter, FailedLimitReserveDuration,
+				// LowTokenRequestNotifyCounter) which currently leak label series on resource
+				// group deletion.
 				return true
 			}
 			gc.inactive = true
--- a/client/resource_group/controller/global_controller.go
+++ b/client/resource_group/controller/global_controller.go
@@ -634,6 +634,11 @@ func (c *ResourceGroupsController) cleanUpResourceGroup() {
 			if gc.inactive || gc.tombstone.Load() {
 				c.groupsController.Delete(resourceGroupName)
 				metrics.ResourceGroupStatusGauge.DeleteLabelValues(resourceGroupName, resourceGroupName)
+				metrics.DemandRUPerSecGauge.DeleteLabelValues(resourceGroupName)
+				// TODO: clean up the remaining per-group metrics (e.g. TokenConsumedHistogram,
+				// GroupRunningKVRequestCounter, SuccessfulRequestDuration, FailedRequestCounter,
+				// ResourceGroupTokenRequestCounter, RequestRetryCounter, FailedLimitReserveDuration)
+				// which currently leak label series on resource group deletion.
 				return true
 			}
 			gc.inactive = true
--- a/client/resource_group/controller/group_controller.go
+++ b/client/resource_group/controller/group_controller.go
@@ -51,6 +51,9 @@ type groupCostController struct {
 		consumption   *rmpb.Consumption
 		storeCounter  map[uint64]*rmpb.Consumption
 		globalCounter *rmpb.Consumption
+		// demandRUTotal accumulates total demanded RU (pre-throttling).
+		// Unlike consumption, this is never subtracted on throttle failure.
+		demandRUTotal float64
 	}
 
 	// fast path to make once token limit with un-limit burst.
@@ -74,6 +77,9 @@ type groupCostController struct {
 		// targetPeriod stores the value of the TargetPeriodSetting setting at the
 		// last update.
 		targetPeriod time.Duration
+
+		// demandRUTotal is a snapshot of mu.demandRUTotal, copied in updateRunState.
+		demandRUTotal float64
 
 		// consumptions stores the last value of mu.consumption.
 		// requestUnitConsumptions []*rmpb.RequestUnitItem
@@ -106,6 +112,7 @@ type groupMetricsCollection struct {
 	tokenRequestCounter               prometheus.Counter
 	runningKVRequestCounter           prometheus.Gauge
 	consumeTokenHistogram             prometheus.Observer
+	demandRUPerSecGauge               prometheus.Gauge
 }
 
 func initMetrics(oldName, name string) *groupMetricsCollection {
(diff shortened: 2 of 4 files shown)


client/resource_group/controller/group_controller.go
```diff
@@ -122,6 +129,7 @@ func initMetrics(oldName, name string) *groupMetricsCollection {
 		tokenRequestCounter:               metrics.ResourceGroupTokenRequestCounter.WithLabelValues(oldName, name),
 		runningKVRequestCounter:           metrics.GroupRunningKVRequestCounter.WithLabelValues(name),
 		consumeTokenHistogram:             metrics.TokenConsumedHistogram.WithLabelValues(name),
+		demandRUPerSecGauge:               metrics.DemandRUPerSecGauge.WithLabelValues(name),
```
rleungx, line 132: Will these metrics being cleaned?

client/resource_group/controller/metrics/metrics.go
```diff
@@ -31,6 +31,8 @@ const (
 var (
 	// ResourceGroupStatusGauge comments placeholder
 	ResourceGroupStatusGauge *prometheus.GaugeVec
+	// DemandRUPerSecGauge is the EMA of demanded RU/s before throttling per resource group.
+	DemandRUPerSecGauge *prometheus.GaugeVec
```
rleungx, line 36: If it is throttled, will the demand be recorded?

client/resource_group/controller/group_controller.go
JmPotato: Good catch. On the current master only `ResourceGroupStatusGauge` is cleaned in `cleanUpResourceGroup`, so this new gauge would leak label series the same way most other per-group metrics already do. The leak impact is small in practice (resource groups are typically long-lived, client-side only, bounded per process), but it's still a real leak and not a design decision. I'll add `DeleteLabelValues` for `DemandRUPerSecGauge` in the cleanup path to keep this PR self-consistent. Cleaning up the other pre-existing per-group metrics is out of scope here — I'm tracking that in a separate branch.

client/resource_group/controller/metrics/metrics.go
JmPotato: Yes — that's exactly the design intent of this metric. Demand is accumulated at request entry (`onRequestWaitImpl` etc.) **before** `acquireTokens` is called, and on throttle error only `consumption` is rolled back via `sub(gc.mu.consumption, delta)`; `demandRUTotal` is never subtracted. So throttled requests still count toward demand, which is the whole point of exposing this separately from the existing consumption-based `avgRUPerSec`. I'll also tighten the Help text / doc comment to make this invariant explicit ("including requests rejected by the token bucket").

client/resource_group/controller/global_controller.go
rleungx, line 633: Will the metrics be deleted if the request RU is large?

client/resource_group/controller/group_controller.go
```diff
@@ -257,11 +272,29 @@ func (gc *groupCostController) updateRunState() {
...
 	logControllerTrace("[resource group controller] update run state", zap.String("name", gc.name), zap.Any("request-unit-consumption", gc.run.consumption), zap.Bool("is-throttled", gc.isThrottled.Load()))
 	gc.run.now = newTime
 }
 
+// recordDemand accumulates a delta into the pre-throttling demand counter.
+//
+// Call sites MUST invoke this before any token-bucket wait/acquire so that
+// demand is captured even when the request is ultimately rejected by the
+// limiter; that is the entire reason `demandRUTotal` is tracked separately
+// from `consumption`. Demand is monotonically increasing and is never rolled
```
YuhaoZhang00, line 286: "monotonically increasing" - Can delta RU be negative? (especially for `onResponseImpl`/`onResponseWaitImpl` path).

Since we'll be introducing the RU pre-charge functionality in the near future, RU over-draft in pre-charge period is likely, and in this case, the `..Response..` path delta RU would be negative.

PR #10588, opened 2026-04-09
client/resource_group: expose RU consumption by request source

### LINKED ISSUE #10634: client/resource_group: cache request source RU metrics in pd/client
## Enhancement Task

Move RU-by-request-source metric ownership from `client-go` into `pd/client`, and cache `(resource_group, request_source)` counter handles to avoid `WithLabelValues()` on the hot path.

Tracking issue (TiDB side): pingcap/tidb#64339

Companion PR (client-go side): tikv/client-go#1929

YuhaoZhang00:

Issue Number: close #10634

Tracking issue: pingcap/tidb#64339  
Related PR: tikv/client-go#1929

`pd/client` accounts RU consumption in the resource-group controller, but it does not expose a breakdown by request source. This makes it difficult to attribute RU usage to user queries, DDL, stats, and other internal activities.

This PR exposes RRU and WRU consumption by request source through the existing `pd/client` accounting flow.

- Optionally read `RequestSource() string` from `controller.RequestInfo`. Requests without this provider use an empty `request_source`.
- Add `resource_manager_client_request_ru_total{resource_group, request_source, type, direction}`:
  - `type`: `rru` or `wru`
  - `direction`: `consume` or `refund`
- Record deltas at the same request and response boundaries as the controller's RU accounting. Negative deltas are recorded as positive `refund` values, so net RU is `consume - refund`.
- Lazily cache metric handles on the request path and clean them up with the resource group or controller.

This metric may create up to the following number of Prometheus time series within one client process if every request source occurs in every active resource group:

`active resource groups * distinct request_source values * 2 RU types * 2 directions`

Metric handles are created lazily, so only observed non-zero label combinations produce series. However, this PR does not cap the number of `request_source` values; deployments with many resource groups or high-cardinality request sources may therefore create a large number of series.

### Dependency on client-go

This PR works independently because `RequestSource()` is optional. However, attributing TiDB/client-go traffic to its actual source requires tikv/client-go#1929. Without that plumbing, requests are recorded with an empty `request_source`.

The metric becomes effective for client-go after it consumes a `pd/client` version containing this PR.

### Check List

Tests:

- Unit tests cover request/response accounting, consume/refund deltas, failed-write payback, concurrent metric creation, and lifecycle cleanup.
- `cd client && make static`

Manual end-to-end test:

Real `INSERT` and `SELECT` statements produced:

```text
request_source="external_Insert", type="wru", direction="consume"
request_source="leader_external_Select", type="rru", direction="consume"
```

After repeating the workload, the counters increased as expected:

```text
external_Insert:        1.8484375 → 5.44765625
leader_external_Select: 0.5092324625651041 → 0.9982272949218749
```

Requests that bypass RU accounting are intentionally absent from this metric.

### Release note

```release-note
None
```


client/resource_group/controller/group_controller.go
```diff
@@ -122,6 +129,52 @@ func initMetrics(oldName, name string) *groupMetricsCollection {
...
+	}
+	if consumption.WRU > 0 {
+		sourceMetrics.wru.Add(consumption.WRU)
+	}
+}
+
+func (mc *groupMetricsCollection) cleanupRequestSourceMetrics(resourceGroupName string) {
+	mc.sourceMetricsMu.Lock()
+	defer mc.sourceMetricsMu.Unlock()
+	for requestSource := range mc.sourceMetrics {
```
rleungx: Will it leak if the getOrCreateRequestSourceMetrics create a new one?

YuhaoZhang00: ~~No extra leak from this cache:~~

~~1. These cached request-source metrics are cleaned up when the resource group is cleaned up (`cleanupRequestSourceMetrics()` called), so they do not stay around forever.~~

~~2. The `request_source` cardinality is also bounded in practice. In TiDB/client-go it currently comes from a small set of hardcoded values (< 100), so we do not expect it to grow uncontrollably.~~

~~If the concern is about concurrency issue, all `sourceMetrics` operations are wrapped by mutex locks.~~

client/resource_group/controller/global_controller.go
```diff
@@ -521,6 +480,22 @@ func (c *ResourceGroupsController) loadOrStoreGroupController(name string, gc *g
 	return tmp.(*groupCostController), loaded
 }
 
+func (c *ResourceGroupsController) getOrCreateRequestSourceMetricsState(name string) *requestSourceMetricsState {
```
rleungx, line 483: Will there be a race between create and cleanup

YuhaoZhang00: Good catch. Fixed by switching to `LoadAndDelete` in `cleanupRequestSourceMetricsState`.  Once `LoadAndDelete` succeeds, the entry is gone from the map, so `getOrCreateRequestSourceMetricsState` will never observe a closed state, and the next `Load`/`LoadOrStore` creates a fresh one. PTAL

rleungx (approved):

client/resource_group/controller/metrics/metrics.go
```diff
@@ -153,6 +155,15 @@ func initMetrics(constLabels prometheus.Labels) {
...
 		}, []string{newResourceGroupNameLabel})
 
+	RequestSourceRUCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace:   namespace,
+			Subsystem:   requestSubsystem,
+			Name:        "ru_total",
+			Help:        "Counter of request RU consumption grouped by resource group and request source.",
+			ConstLabels: constLabels,
+		}, []string{newResourceGroupNameLabel, "request_source", errType})
```
JmPotato: Although `errType` is also a "type," the semantics here are just a plain type and not related to err, right? It would be better to give `errType` a more generic name.

client/resource_group/controller/group_controller.go
```diff
@@ -165,7 +249,10 @@ func newGroupCostController(
 	default:
 		return nil, errs.ErrClientResourceGroupConfigUnavailable.FastGenByArgs("not supports the resource type")
 	}
-	ms := initMetrics(group.Name, group.Name)
+	if sourceState == nil {
+		sourceState = newRequestSourceMetricsState(group.Name)
```
JmPotato: The `sourceState` created here is not registered in the `requestSourceStates` map, which means it will never be cleaned up by `cleanupRequestSourceMetricsState`. Currently, only tests pass `nil`; in production, the path always passes a non-`nil` value through `getOrCreateRequestSourceMetricsState`. I suggest adding a comment clarifying that this fallback is for test scenarios only, to avoid future misuse that could lead to metrics leaks.

client/resource_group/controller/global_controller.go
```diff
@@ -384,6 +338,7 @@ func (c *ResourceGroupsController) Start(ctx context.Context) {
 			/* channels */
 			case <-c.loopCtx.Done():
 				metrics.ResourceGroupStatusGauge.Reset()
+				metrics.RequestSourceRUCounter.Reset()
```
okJiang, line 341: This PR still resets `RequestSourceRUCounter` when one controller shuts down. Because the counter is process-global, this can wipe request-source RU metrics for other active controllers in the same process and make the new observability misleading. Please remove the global reset or scope the cleanup so one controller cannot clear another controller's series.

client/resource_group/controller/group_controller.go
```diff
@@ -122,6 +156,55 @@ func initMetrics(oldName, name string) *groupMetricsCollection {
...
+		return
+	}
+	sourceMetrics := mc.getOrCreateRequestSourceMetrics(requestSource)
+	if sourceMetrics == nil {
+		return
+	}
+	if consumption.RRU > 0 {
+		sourceMetrics.rru.Add(consumption.RRU)
+	}
+	if consumption.WRU > 0 {
```
okJiang: Does this need to account for failed write paybacks before exporting WRU? `onRequestWaitImpl` records the write-side delta to this counter before the response is known, but `KVCalculator.AfterKVRequest` calls `payBackWriteCost` when `ResponseInfo.Succeed()` is false, producing a negative WRU delta that this counter drops because it only adds positive values. That means failed writes still permanently increase `resource_manager_client_request_ru_total{type="wru"}`, while the controller consumption is later reduced. If this metric is meant to report RU consumption by request source, failed writes will be overcounted.

YuhaoZhang00: Good catch — fixed in 043413d. The per-source export is now deferred to response time and uses the round-trip `count` (delta + BeforeKVRequest), so the failed-write payback is reflected and the counter track `gc.mu.consumption`. Kept the `> 0` clamp in addRequestSourceRU as defense in depth

okJiang (approved):

okJiang:
@YuhaoZhang00 please fix the ci

YuhaoZhang00:
Cross-repository dependency: for this metric to take effect in TiDB/client-go, client-go must propagate the request source and implement `RequestSource()` on its `resourcecontrol.RequestInfo` (see tikv/client-go#1929). A consumer that upgrades to this `pd/client` version without that method will no longer satisfy `controller.RequestInfo`; without a meaningful propagated source, the metric cannot provide the intended per-source attribution.

YuhaoZhang00:
/hold

hold temporarily for local (re)test.

client/resource_group/controller/group_controller.go
```diff
@@ -115,7 +116,53 @@ type groupMetricsCollection struct {
...
+
+func (s *requestSourceMetricsState) cleanup() {
+	if s == nil {
+		return
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.closed = true
+	for key := range s.items {
+		metrics.RequestSourceRUCounter.DeleteLabelValues(
```
JmPotato, line 159: `RequestSourceRUCounter` is process-global, so different controllers can cache the same series. Deleting it here leaves other controllers with orphaned handles whose later `Add` calls are no longer exported. Please make deletion ownership-aware (for example, delete only after the last owner) and add a two-controller test.

JmPotato (commented):
Please address the inline lifecycle issue.

client/resource_group/controller/group_controller.go
YuhaoZhang00: I believe next-gen is designed to have only one active `ResourceGroupsController` per process in steady state (correct me if I'm wrong!).

However, a very small overlap window may still exist during bootstrap, upgrade, or controller handoff while the old controller is shutting down and the new one is starting, and I'll try to fix this.

JmPotato: I understand your point: in the current way TiDB is used, a process has only a single `ResourceGroupsController`. However, this is not a hard constraint, and it may change in future development, which is why I have some concerns. Still, the PR can be merged as is. For this issue, I think we should leave a TODO or at least keep it in mind, as we may encounter potential pitfalls down the road.

YuhaoZhang00: Will open a new issue to note this down -> https://github.com/tikv/pd/issues/11080

PR #10589, opened 2026-04-09
*: fix request body not rewindable when forwarding HTTP requests

### LINKED ISSUE #10590: *: forwarding HTTP requests can fail with "net/http: cannot rewind body after connection loss"
### What did you do?

With audit or rate-limit middleware enabled, send HTTP requests to a PD
follower so that the follower forwards them to the PD leader via
`customReverseProxies` (e.g. `GET /pd/api/v1/cluster`).

When the forwarding `http.Client.Do` needs to retry on a new connection —
typically due to a stale keep-alive connection or HTTP/2 GOAWAY — the retry
fails.

The forwarded request is transparently retried on a new connection and
succeeds.

### What did you see instead?

The forward fails with:

```
[ERROR] [apiutil.go:465] ["request failed"] [error="[PD:http:ErrSendRequest]send HTTP request failed: Get \"https://.../pd/api/v1/cluster\": net/http: cannot rewind body after connection loss"]
```

Root cause: when audit/rate-limit middleware is enabled,
`requestutil.getBodyParam` replaces `r.Body` with
`io.NopCloser(bytes.NewBuffer(buf))` without setting `r.GetBody`. This turns
what was originally `http.NoBody` into a non-nil body reader with no rewind
function. On the forwarding path, `net/http` Transport probes the body
(`didRead=true`), and on connection loss cannot rewind because `GetBody` is
nil, so the retry path errors out with the message above.

master (reproduces on current master and on the release branches that carry
the same audit middleware code path).

iosmanthus:

Issue Number: close #10590

When audit or rate-limit middleware is enabled, `requestutil.getBodyParam`
replaces `r.Body` with `io.NopCloser(bytes.NewBuffer(buf))` without wiring
up `r.GetBody`. A request that originally had `http.NoBody` ends up with a
non-nil body reader and no rewind function.

When such a request is forwarded by `customReverseProxies` via
`http.Client.Do`, Go's Transport may need to retry on a new connection
(stale keep-alive, HTTP/2 GOAWAY). The Transport probes the body
(`didRead=true`) and on retry cannot rewind because `GetBody` is nil,
failing the forward with:

```
net/http: cannot rewind body after connection loss
```

- Add `apiutil.EnsureRewindableBody(r)` helper. It is a no-op when the body
  is `nil`, `http.NoBody`, or `r.GetBody` is already set. Otherwise it
  drains the body once, restores `http.NoBody` semantics for empty payloads
  (so `Transport.outgoingLength` returns 0 and no body probing happens),
  wires `r.GetBody` to return a fresh reader over the buffered bytes, and
  sets `r.ContentLength` to the exact length.
- Use the helper in `customReverseProxies.ServeHTTP` before calling
  `http.Client.Do`, so any forwarded request is rewindable.
- Refactor `requestutil.getBodyParam` to reuse the helper, then read the
  cached body via `r.GetBody()` for audit logging. This fixes the root
  cause while keeping existing audit behavior.
- Add `TestEnsureRewindableBody` covering nil/`NoBody`/existing-`GetBody`
  no-op paths, empty-body `NoBody` restoration, non-empty rewindability
  across multiple `GetBody()` invocations, and read-error propagation.

```commit-message
*: fix request body not rewindable when forwarding HTTP requests

When audit or rate-limit middleware is enabled, getBodyParam replaces
r.Body without setting r.GetBody. Forwarding such a request through
customReverseProxies via http.Client.Do fails on retry with
"net/http: cannot rewind body after connection loss". Introduce
apiutil.EnsureRewindableBody and use it on the forwarding path and in
getBodyParam to make the body safely rewindable.
```

### Check List

Tests

- Unit test

Side effects

- Possible performance regression (body is buffered into memory once per
  forwarded request that has a body — same as the pre-existing audit path)

Related changes

### Release note

```release-note
Fix "net/http: cannot rewind body after connection loss" when a PD
follower forwards HTTP requests to the leader with audit or rate-limit
middleware enabled.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP request body handling to ensure request bodies can be reliably read and re-read throughout the request lifecycle.

* **Tests**
  * Added comprehensive test coverage for request body rewindability across various scenarios.

### DIFF
--- a/pkg/utils/apiutil/apiutil.go
+++ b/pkg/utils/apiutil/apiutil.go
@@ -484,7 +484,50 @@ func NewCustomReverseProxies(dialClient *http.Client, urls []url.URL) http.Handl
 	return p
 }
 
+// EnsureRewindableBody makes r's body safe to be retried by net/http
+// Transport on connection loss. It is a no-op when the body is nil,
+// http.NoBody, or r.GetBody is already set. Otherwise it drains the body
+// into memory and wires up GetBody so the transport can rewind.
+//
+// Callers should ensure the body fits in memory; this helper buffers the
+// entire payload. It guards against the
+// "net/http: cannot rewind body after connection loss" error that can
+// occur when a server-side request (GetBody == nil) is forwarded via
+// http.Client.Do and the underlying keep-alive connection goes stale.
+func EnsureRewindableBody(r *http.Request) error {
+	if r.Body == nil || r.Body == http.NoBody || r.GetBody != nil {
+		return nil
+	}
+	buf, err := io.ReadAll(r.Body)
+	_ = r.Body.Close()
+	if err != nil {
+		return err
+	}
+	// Restore NoBody semantics for empty payloads so that Transport's
+	// outgoingLength returns 0 and no body probing happens.
+	if len(buf) == 0 {
+		r.Body = http.NoBody
+		r.GetBody = func() (io.ReadCloser, error) { return http.NoBody, nil }
+		r.ContentLength = 0
+		return nil
+	}
+	r.Body = io.NopCloser(bytes.NewReader(buf))
+	r.GetBody = func() (io.ReadCloser, error) {
+		return io.NopCloser(bytes.NewReader(buf)), nil
+	}
+	// We now know the exact length; set it so the transport can pick
+	// Content-Length framing over chunked encoding when forwarding.
+	r.ContentLength = int64(len(buf))
+	return nil
+}
+
 func (p *customReverseProxies) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if err := EnsureRewindableBody(r); err != nil {
+		log.Error("failed to read request body", zap.Error(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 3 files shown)


JmPotato (commented):
Could you add a corresponding test?

iosmanthus:
> Could you add a corresponding test?

Sure!

pkg/utils/apiutil/apiutil.go
```diff
@@ -484,7 +484,50 @@ func NewCustomReverseProxies(dialClient *http.Client, urls []url.URL) http.Handl
...
+		return io.NopCloser(bytes.NewReader(buf)), nil
+	}
+	// We now know the exact length; set it so the transport can pick
+	// Content-Length framing over chunked encoding when forwarding.
+	r.ContentLength = int64(len(buf))
+	return nil
+}
+
 func (p *customReverseProxies) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if err := EnsureRewindableBody(r); err != nil {
```
rleungx, line 525: Do we need to move it before `Do`?

PR #10492, opened 2026-03-26
keyspace, schedule: make keyspace creation atomic 

bufferflies:
### What problem does this PR solve?

Issue Number: Close #10461

### What is changed and how does it work?

```commit-message
Make keyspace creation persist keyspace metadata, keyspace-group membership,
and region label rules in one transaction so create failures no longer leave
partial storage state behind. Move in-memory keyspace and keyspace-group cache
updates to post-commit callbacks so failed transactions do not leak committed
state into memory.
```

### Known limitations

- `rollbackCreateKeyspace` (the compensating cleanup that runs when a
  keyspace is created but the subsequent wait-for-region-split step
  fails) identifies the keyspace it is cleaning up by id, name, and
  state (DISABLED). If the original keyspace is manually archived and
  removed while its own create call is still waiting on the split, and
  a `CreateKeyspaceByID` call then reuses the same id *and* the same
  name before this stale rollback runs, rollback will adopt and delete
  that replacement instead of leaving it alone - it is otherwise
  indistinguishable from the keyspace this rollback actually created.
  This requires two independent, unlikely events to line up and is
  accepted rather than guarded against: closing it fully would need a
  durable, PD-generated marker with no existing storage-layer support
  to piggyback on cheaply (a purpose-built one was prototyped and
  dropped as not worth a new persisted key for a case this narrow).
  The far more common id-reuse case - a different name - is not
  affected; the existing name check already rejects it.
- If `rollbackCreateKeyspace`'s compensating transaction cannot commit
  after retrying (e.g. the keyspace's TSO group is persistently mid
  split/merge, or storage is having trouble), it gives up rather than
  applying a partial cleanup. The keyspace is left exactly as it was
  after creation: DISABLED, with its meta, TSO group membership, and
  label rule all intact. This is the same explicit, recoverable state
  a wait-split failure already leaves on master (see #10461's own
  "make intermediate states explicit and safely recoverable" as an
  accepted alternative to compensation); it does mean the keyspace's
  name and id stay occupied until an operator manually cleans it up or
  re-enables it, rather than becoming immediately reusable.

### Check List

Tests

- Unit test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

### Release note

```release-note
Fix partial keyspace creation state when later create steps fail after the
initial metadata transaction succeeds.
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced keyspace bootstrap and creation with retrying pre-allocation and more robust handling of already-existing keyspaces.
[...]

PR #10597, opened 2026-04-14
support etcd groups

### LINKED ISSUE #10596: 45bc455d support etcd groups

bufferflies:

Issue Number: Close #10596, ref #10516

Author: @AmoebaProtozoa

Cherry-pick: [`45bc455d`](https://github.com/bufferflies/pd/commit/45bc455d98137fae552bec4dd9b27e88505f2299)

```commit-message
Support meta service groups for keyspace metadata.

Add storage helpers and keyspace manager logic to assign keyspaces to configured meta service groups and track assignment counts.

Add HTTP API v2 handlers and pd-ctl commands to list meta service groups and add new groups.
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has HTTP APIs changed
- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Support configuring and managing meta service groups for keyspace metadata through PD HTTP API v2 and pd-ctl.
```


AmoebaProtozoa (approved):

pkg/keyspace/keyspace.go
```diff
@@ -256,6 +271,22 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	tracer.SetKeyspace(newID, request.Name)
 	tracer.OnAllocateIDFinished()
 
+	if request.Config != nil {
+		delete(request.Config, MetaServiceGroupAddressesKey)
+	}
+	// assign meta-service group for the new keyspace if meta-service groups exist.
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.SelectGroup()
```
lhy1024: This splits meta-service group selection and assignment-count update into two phases: SelectGroup() first, then updateAssignmentTxn() later in saveNewKeyspace(). With concurrent keyspace creation, multiple requests can select the same group based on the same stale counts, causing skewed assignment; overlapping etcd transactions may also surface more txn conflicts.

Please make “select group, persist keyspace config, and increment assignment count” one storage transaction, with retry on ErrEtcdTxnConflict. The same pattern in CreateKeyspaceByID should be fixed as well.

pkg/keyspace/keyspace.go
```diff
@@ -494,6 +562,13 @@ func (manager *Manager) saveNewKeyspace(keyspace *keyspacepb.KeyspaceMeta) error
 		if loadedMeta != nil {
 			return errs.ErrKeyspaceExists
 		}
+		if manager.mgm != nil {
+			if groupID := keyspace.GetConfig()[MetaServiceGroupIDKey]; groupID != "" {
+				if err := manager.mgm.updateAssignmentTxn(txn, "", groupID); err != nil {
```
lhy1024: The CreateKeyspace/CreateKeyspaceByID paths do not validate a user-provided meta_service_group_id. If no meta-service groups are configured, or if the request contains an unknown group ID, this still increments the assignment count for that arbitrary group and persists it in the keyspace config. UpdateKeyspaceConfig rejects unknown groups, so create and update behave inconsistently.

Please validate meta_service_group_id on create as well, or treat meta-service-group fields as system-managed and always clear/overwrite user input during creation.

server/config/config.go
lhy1024: MetaServiceGroups is a new map field, but Clone() only deep-copies PreAlloc. The cloned KeyspaceConfig still shares the same MetaServiceGroups map with the original config. As a result, callers of Server.GetConfig()/GetKeyspaceConfig() can mutate live config without going through SetKeyspaceConfig, and this may also introduce concurrent map read/write risks.

Please deep-copy MetaServiceGroups in Clone() as well.

pkg/keyspace/meta_service_group.go
```diff
@@ -0,0 +1,173 @@
...
+	)
+	err = m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
+		count, err = m.store.GetAssignmentCount(txn, m.metaServiceGroups)
+		return err
+	})
+	return count, err
+}
+
+func (m *MetaServiceGroupManager) selectGroupTxn(txn kv.Txn) (string, error) {
+	countMap, err := m.store.GetAssignmentCount(txn, m.metaServiceGroups)
```
okJiang, line 63: Prefer to moving this line out of function and renaming `selectGroupTxn` to `filterProperGroup`/`findGroupTxn`. Then it is easier to write an unit test for it.

pkg/keyspace/meta_service_group.go
```diff
@@ -0,0 +1,173 @@
...
+
+// GetAssignmentCounts returns the count of each meta-service group.
+func (m *MetaServiceGroupManager) GetAssignmentCounts() (map[string]int, error) {
+	m.RLock()
+	defer m.RUnlock()
+	var (
+		err   error
+		count map[string]int
+	)
+	err = m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
```
okJiang: We don't need to access the disk every time to get the ETCD group with the smallest number. 

We can directly maintain a variable to record the number of each group when inserting data. This way, we can easily calculate which groups we need and which groups are appropriate, thereby avoiding multiple accesses to the disk.

pkg/utils/keypath/meta_service_group.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package keypath
+
+import "fmt"
+
+// MetaServiceGroupAssignmentCountPath returns the path for the meta-service
+// group assignment count.
+func MetaServiceGroupAssignmentCountPath(id string) string {
+	return fmt.Sprintf("/pd/%d/meta_service_groups/%s/assignment_count", ClusterID(), id)
```
okJiang: Please create a constant in the keypath.go file.

okJiang (approved):

okJiang (changes requested):
revert approve(misoperation

ystaticy (approved):

pkg/keyspace/meta_service_group.go
bufferflies: Rename selectGroupTxn as findMinMetaGroup

pkg/utils/keypath/meta_service_group.go
bufferflies: good suggestion

pkg/keyspace/keyspace.go
bufferflies: yes, it will be fixed in https://github.com/tikv/pd/pull/10492

pkg/keyspace/keyspace.go
bufferflies: count, err := loadAssignmentCount(txn, id) will return an error if the meta group ID does not exist.

pkg/keyspace/meta_service_group.go
bufferflies: add todo optimize here

pkg/keyspace/meta_service_group.go
```diff
@@ -0,0 +1,173 @@
...
+	defer m.RUnlock()
+	var assignedGroup string
+	if err := m.store.RunInTxn(m.ctx, func(txn kv.Txn) error {
+		var err error
+		assignedGroup, err = m.selectGroupTxn(txn)
+		if err != nil {
+			return err
+		}
+		return m.store.IncrementAssignmentCount(txn, assignedGroup, count)
+	}); err != nil {
```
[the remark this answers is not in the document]
bufferflies: Ignore this for the caller can't give the negative number

pkg/keyspace/keyspace.go
lhy1024:
```
func loadAssignmentCount(txn kv.Txn, id string) (int, error) {
	countPath := keypath.MetaServiceGroupAssignmentCountPath(id)
	countVal, err := txn.Load(countPath)
	if err != nil {
		return 0, err
	}
	if countVal == "" {
		return 0, nil
	}
	return strconv.Atoi(countVal)
}
```
 if the meta group ID does not exist, countVal will be empty？

server/config/config.go
lhy1024, line 920: Clone() still does not deep-copy MetaServiceGroups. The struct copy cfg := *c shares the same underlying map between the original and the clone — callers of Server.GetConfig() / GetKeyspaceConfig() can mutate live config without going through SetKeyspaceConfig, and this also carries a  concurrent map read/write risk.

Please deep-copy the map in Clone():
```
   metaServiceGroups := make(map[string]string, len(c.MetaServiceGroups))
   for k, v := range c.MetaServiceGroups {
       metaServiceGroups[k] = v
   }
   cfg.MetaServiceGroups = metaServiceGroups
```

The GetPreAlloc defensive copy is a nice addition but doesn't fix this issue.

pkg/keyspace/keyspace.go
```diff
@@ -256,6 +271,22 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	tracer.SetKeyspace(newID, request.Name)
 	tracer.OnAllocateIDFinished()
 
+	if request.Config != nil {
+		delete(request.Config, MetaServiceGroupAddressesKey)
+	}
+	// assign meta-service group for the new keyspace if meta-service groups exist.
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.SelectGroup(manager.ctx)
```
rleungx: This select is outside the transaction that increments the assignment count, so concurrent creates can choose the same stale minimum. Please combine select and increment in one RunInTxn, and apply the same fix to CreateKeyspaceByID.

server/apiv2/handlers/meta_service_group.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	var requests []AddMetaServiceGroupRequest
+	err := c.ShouldBindJSON(&requests)
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, errs.ErrBindJSON.Wrap(err).GenWithStackByCause().Error())
+		return
+	}
+	// Constructs new meta-service groups.
+	currentGroups := manager.GetGroups()
```
rleungx: This read-modify-write is not protected against concurrent POSTs. Two successful requests can build from the same currentGroups and the later persist can drop the earlier addition; please serialize this path or use a CAS-style config update.

server/config/config.go
```diff
@@ -879,6 +879,9 @@ type KeyspaceConfig struct {
 	WaitRegionSplitTimeout typeutil.Duration `toml:"wait-region-split-timeout" json:"wait-region-split-timeout"`
 	// CheckRegionSplitInterval indicates the interval to check whether the region split is complete
 	CheckRegionSplitInterval typeutil.Duration `toml:"check-region-split-interval" json:"check-region-split-interval"`
+	// MetaServiceGroups is the available external meta-service groups.
+	// The key is the meta-service group name, and the value is the corresponding endpoint.
+	MetaServiceGroups map[string]string `toml:"meta-service-groups" json:"meta-service-groups"`
```
rleungx, line 884: Because this new field is a map, KeyspaceConfig.Clone must also deep-copy it. Otherwise callers of cloned config can mutate MetaServiceGroups without validation, persistence, or manager refresh.

bufferflies: done

server/apiv2/handlers/meta_service_group.go
bufferflies: good suggestion

pkg/keyspace/keyspace.go
bufferflies: done, pick include increment

server/config/config.go
bufferflies: done

pkg/keyspace/keyspace.go
```diff
@@ -279,6 +299,18 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 			request.Config[GCManagementType] = KeyspaceLevelGC
 		}
 	}
+	// assign meta-service group for the new keyspace if meta-service groups exist.
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.PickGroup(manager.ctx)
```
rleungx, line 308: PickGroup updates the persisted assignment count before the keyspace is saved. If saveNewKeyspace fails, the count is leaked; please make this atomic or roll it back on all failure paths.

server/api/config.go
```diff
@@ -230,7 +232,7 @@ func (h *confHandler) updateKeyspaceConfig(config *config.Config, key string, va
 	}
 
 	if updated {
-		err = h.svr.SetKeyspaceConfig(config.Keyspace)
+		err = h.svr.SetKeyspaceConfig(&oldCfg, newCfg)
```
rleungx: This old value comes from a cloned config, so CAS will not match the persisted pointer. Please use the persisted keyspace config as the CAS old value.

server/config/config.go
```diff
@@ -932,3 +947,17 @@ func (c *KeyspaceConfig) GetWaitRegionSplitTimeout() time.Duration {
...
 	return c.CheckRegionSplitInterval.Duration
 }
+
+// GetMetaServiceGroups returns the current meta-service-group configuration.
+func (c *KeyspaceConfig) GetMetaServiceGroups() map[string]string {
+	ret := make(map[string]string, len(c.MetaServiceGroups))
+	for name, endpoint := range c.MetaServiceGroups {
+		ret[name] = endpoint
+	}
+	return c.MetaServiceGroups
```
rleungx: This builds a defensive copy but returns the original map. Please return ret to avoid exposing mutable config state.

pkg/keyspace/keyspace.go
bufferflies: The  meta group ID only pick from the meta group manager, so it must exist.

server/config/config.go
bufferflies: done

pkg/keyspace/keyspace.go
bufferflies: Yes, I will fix atomic txn in pr https://github.com/tikv/pd/pull/10492

server/config/config.go
```diff
@@ -890,6 +893,11 @@ func (c *KeyspaceConfig) Validate() error {
 	if c.CheckRegionSplitInterval.Duration >= c.WaitRegionSplitTimeout.Duration {
 		return errors.New("[keyspace] check-region-split-interval should be less than wait-region-split-timeout")
 	}
+	for _, endpoint := range c.MetaServiceGroups {
```
rleungx: Please also reject empty meta-service group IDs, and trim IDs and addresses consistently. An empty ID is treated as no group selected elsewhere and can break keyspace creation.

pkg/keyspace/keyspace.go
```diff
@@ -409,6 +453,17 @@ func (manager *Manager) CreateKeyspaceByID(request *CreateKeyspaceByIDRequest) (
 			request.Config[GCManagementType] = KeyspaceLevelGC
 		}
 	}
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.PickGroup(manager.ctx)
```
rleungx, line 464: Same assignment-count leak risk here: PickGroup reserves a count before this keyspace is persisted, so later creation failures can leave the group count too high.

server/apiv2/handlers/meta_service_group.go
```diff
@@ -0,0 +1,164 @@
...
```
rleungx: Please build currentGroups from oldCfg.GetMetaServiceGroups() after reading the CAS base config. Reading the manager first can drop a group added by a concurrent POST.

pkg/keyspace/keyspace.go
rleungx, line 308: PickGroup increments the persisted assignment count before the keyspace is saved. If creation fails after this point, the count can leak; please make the assignment update atomic with saving the keyspace or compensate all later failures.

server/config/config.go
bufferflies: good suggestion, done

server/apiv2/handlers/meta_service_group.go
bufferflies: Good catch, fixed.

server/config/config.go
bufferflies: done

server/api/config.go
bufferflies: done

pkg/keyspace/keyspace.go
```diff
@@ -387,6 +430,9 @@ func (manager *Manager) CreateKeyspaceByID(request *CreateKeyspaceByIDRequest) (
 	if err != nil {
 		return nil, err
 	}
+	if request.Config != nil {
+		delete(request.Config, MetaServiceGroupAddressesKey)
```
lhy1024, line 436: This is separate from the atomicity concern above. The create path still allows users to provide meta_service_group_id. The current code only removes meta_service_group_addrs, but it does not reject or clear meta_service_group_id; when no meta-service groups are configured, this internal field can still be persisted into the keyspace config.

  If meta-service-group fields are expected to be PD-managed, please handle both fields explicitly on create: either clear both meta_service_group_id / meta_service_group_addrs and let PD assign them, or reject requests containing these fields. CreateKeyspaceByID should behave consistently as well.

pkg/keyspace/keyspace.go
```diff
@@ -256,6 +272,10 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	tracer.SetKeyspace(newID, request.Name)
 	tracer.OnAllocateIDFinished()
 
+	if request.Config != nil {
+		delete(request.Config, MetaServiceGroupAddressesKey)
```
lhy1024, line 276: ditto

pkg/keyspace/keyspace.go
bufferflies: Good catch, user can't pass the meta group ID and address by creating keyspace, it must be set in the PD server. If we want to migrate the meta group, we call the update keyspace config to modify the meta group information.

pkg/keyspace/keyspace.go
bufferflies: done

pkg/keyspace/keyspace.go
```diff
@@ -279,6 +301,18 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 			request.Config[GCManagementType] = KeyspaceLevelGC
 		}
 	}
+	// assign meta-service group for the new keyspace if meta-service groups exist.
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.PickGroup(manager.ctx)
```
rleungx, line 308: PickGroup increments the persisted count before saveNewKeyspace succeeds. If saving later fails, the count is leaked. Please move the assignment into the save transaction or roll it back on all later failures.

bufferflies: It will be fixed in https://github.com/tikv/pd/pull/10492

rleungx (commented):
I left a few inline review comments.

server/api/config.go
```diff
@@ -219,8 +219,10 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value any) er
 	return errors.Errorf("config prefix %s not found", kp[0])
 }
 
-func (h *confHandler) updateKeyspaceConfig(config *config.Config, key string, value any) error {
-	updated, found, err := jsonutil.AddKeyValue(&config.Keyspace, key, value)
+func (h *confHandler) updateKeyspaceConfig(key string, value any) error {
+	oldCfg := h.svr.GetPersistOptions().GetKeyspaceConfig()
+	newCfg := oldCfg.Clone()
+	updated, found, err := jsonutil.AddKeyValue(&newCfg, key, value)
```
rleungx: This passes a **KeyspaceConfig into AddKeyValue. No-op or unknown updates can hit the reflection path and panic; please pass newCfg instead and add a no-op/unknown-key test.

pkg/keyspace/keyspace.go
rleungx, line 308: PickGroup increments the assignment count before the keyspace is saved. If saveNewKeyspace later fails, the count is leaked; please make the assignment and keyspace save atomic or roll back every failure path.

pkg/keyspace/keyspace.go
```diff
@@ -256,6 +272,11 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 	tracer.SetKeyspace(newID, request.Name)
 	tracer.OnAllocateIDFinished()
 
+	if request.Config != nil {
+		delete(request.Config, MetaServiceGroupAddressesKey)
```
rleungx, line 276: Please also delete MetaServiceGroupIDKey from user input. When no groups are configured, a client can persist an arbitrary meta_service_group_id with no assignment count.

pkg/keyspace/meta_service_group.go
```diff
@@ -0,0 +1,162 @@
...
+	return assignedGroup, nil
+}
+
+// PickGroup returns the meta-service group with the least assigned keyspaces
+// without updating the persisted assignment count.
+func (m *MetaServiceGroupManager) PickGroup(ctx context.Context) (string, error) {
+	m.RLock()
+	defer m.RUnlock()
+	var assignedGroup string
+	if err := m.store.RunInTxn(ctx, func(txn kv.Txn) error {
```
rleungx, line 87: On etcd, this transaction reads all group counts and writes one count, so concurrent creates can conflict and fail without retry. Please add retry or avoid reading all count keys in the create path.

pkg/keyspace/keyspace.go
lhy1024: +1, the same input-sanitization issue also exists in CreateKeyspaceByID. It would be better to fix both create paths together and add a regression test for user-provided meta_service_group_id.

pkg/keyspace/meta_service_group.go
bufferflies: No concurrent, one RLock is here.

pkg/keyspace/keyspace.go
bufferflies: All txn rollback about creating keyspace  will be fixed in this pr https://github.com/tikv/pd/pull/10492

server/api/config.go
bufferflies: good catch, fixed and add unit test for no-op/unknown-key

pkg/keyspace/keyspace.go
bufferflies: done

rleungx (commented):
I left two follow-up inline comments.

pkg/keyspace/keyspace.go
```diff
@@ -279,6 +302,18 @@ func (manager *Manager) CreateKeyspace(request *CreateKeyspaceRequest) (*keyspac
 			request.Config[GCManagementType] = KeyspaceLevelGC
 		}
 	}
+	// assign meta-service group for the new keyspace if meta-service groups exist.
+	assignToMetaServiceGroup := manager.mgm != nil && len(manager.mgm.GetGroups()) > 0
+	if assignToMetaServiceGroup {
+		metaServiceGroup, err := manager.mgm.PickGroup(manager.ctx)
```
rleungx, line 308: PickGroup increments the assignment count before the keyspace is saved. If saveNewKeyspace later fails, the count is leaked; please make them atomic or roll back this update.

pkg/keyspace/meta_service_group.go
rleungx, line 87: This can hit etcd txn conflicts under concurrent keyspace creation because it reads all group counts and writes one count without retry. Please add retry or reduce the conflict scope.

server/api/config.go
```diff
@@ -210,7 +210,7 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value any) er
 		return h.updateClusterVersion(value)
```
lhy1024, line 210: I think most of the previous concerns are addressed, but I still have one concern around updating `keyspace.meta-service-groups` through the generic config API.

  Currently `/pd/api/v1/config` can still replace the whole `meta-service-groups` map. This can bypass the append-only semantics in the new v2 API, and may leave existing keyspaces pointing to a meta-service group that no longer exists. It also seems better to share the validation logic here so invalid group IDs, such as empty/whitespace IDs, are rejected consistently.

  Could we either disallow updating `keyspace.meta-service-groups` through the generic config API, or make this path follow the same append-only and validation rules as the v2 API?

lhy1024 (commented):
the rest LGTM

pkg/keyspace/keyspace.go
```diff
@@ -785,6 +876,9 @@ func (manager *Manager) updateKeyspaceConfigTxn(name string, update func(meta *k
 		)
 		return nil, err
 	}
+	if manager.mgm != nil {
+		manager.mgm.AttachEndpoints(meta.GetConfig())
```
rleungx, line 880: Please apply the same AttachEndpoints handling in UpdateKeyspaceState and UpdateKeyspaceStateByID; otherwise state-update responses miss meta_service_group_addrs while load/config-update responses include it.

server/apiv2/handlers/meta_service_group.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, errs.ErrBindJSON.Wrap(err).GenWithStackByCause().Error())
+		return
+	}
+	oldCfg := svr.GetPersistOptions().GetKeyspaceConfig()
+	newCfg := oldCfg.Clone()
+	currentGroups := newCfg.GetMetaServiceGroups()
+	newGroups := make(map[string]string, len(currentGroups)+len(requests))
+	for _, request := range requests {
+		if request.ID == "" || request.Addresses == "" {
```
rleungx, line 89: Please trim ID and addresses here, or return validation failures as 400. Whitespace-only values currently reach SetKeyspaceConfig and come back as 500, but they are bad requests.

server/api/config.go
bufferflies: yes, add todo to remove it.

pkg/keyspace/keyspace.go
bufferflies: done

server/apiv2/handlers/meta_service_group.go
bufferflies: done

server/api/config.go
```diff
@@ -209,8 +209,11 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value any) er
 	case "cluster-version":
 		return h.updateClusterVersion(value)
 	case "label-property": // TODO: support changing label-property
+		// todo:  use apiv2 handlers instead to update keyspace related config,
+		// since the logic is more complex and may involve multiple config items update in one request,
+		// e.g. add/remove meta-service group.
 	case "keyspace":
-		return h.updateKeyspaceConfig(cfg, kp[len(kp)-1], value)
+		return h.updateKeyspaceConfig(kp[len(kp)-1], value)
```
rleungx, line 216: This still allows /config to update keyspace.meta-service-groups directly, bypassing the new v2 API validation and add-only semantics. Should we reject meta-service-groups here and require the v2 API?

bufferflies: yes,  I have add todo action and fix it later after discuss with users

lhy1024 (approved):

rleungx (approved):

niubell (approved):

PR #10599, opened 2026-04-14
resource_control: add PredictedReadBytes hint for RC paging pre-charge

YuhaoZhang00:
## Summary

Add an optional `predictedReadBytesProvider` interface on `RequestInfo`.
When a caller (e.g. TiDB maintaining a per-logical-scan EMA across paging
cop RPCs) supplies a non-zero `PredictedReadBytes`, `BeforeKVRequest` and
`AfterKVRequest` use that value as the byte basis for the paging pre-charge
instead of `PagingSizeBytes`. `PagingSizeBytes` remains the fallback and
worst-case cap.

Existing `RequestInfo` implementations compile unchanged — the hint is an
optional interface, not a method on `RequestInfo`.

## Why

The fixed 4 MiB paging pre-charge introduced in #10548 typically overshoots
actual scanned bytes and stalls concurrent workers at Phase 1 under tight
resource tiers. A learned per-scan estimate from the TiDB side eliminates
the over-estimation without changing kvproto or TiKV behavior.

## Status

**Draft.** Part of a stacked change:
- `tikv/client-go`: adds `tikvrpc.Request.PredictedReadBytes` + `RequestInfo` getter
- `pingcap/tidb`: maintains per-logical-scan TAEMA on `copTask`, feeds prediction on every paging RPC

Not yet ready for review; e2e validation on the simulation cluster is pending.

## Test plan

- [x] Unit tests in `client/resource_group/controller/` cover hint-present and hint-absent paths, plus an excess-pre-charge refund case
- [ ] e2e: FullScan + JOIN paging sweep on simulation cluster (pending)

### DIFF
--- a/client/go.mod
+++ b/client/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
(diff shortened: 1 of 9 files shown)


YuhaoZhang00:
Added 4 observability metrics on the PD client side for upcoming e2e validation of the paging pre-charge path:

- `resource_manager_client_request_paging_precharge_source_total{source=predicted|fallback}` — does the `PredictedReadBytes` hint actually drive `BeforeKVRequest`
- `resource_manager_client_request_paging_precharge_bytes_total{source}` — pre-charged byte volume per path
- `resource_manager_client_request_paging_actual_bytes_total{source}` — observed `ReadBytes` per path (ratio with above ≈ over-charge factor; historical ≈4×, target ≈1×)
- `resource_manager_client_request_paging_prediction_residual_bytes` histogram — signed `actual − predicted` distribution for EMA accuracy

No public interface change; instrumentation is at the `onRequestWait` / `onResponse` call sites in `group_controller.go`. Draft status unchanged; keeping this open pending Step E validation results.

PR #10600, opened 2026-04-14
tso: validate callee id for tso requests

### LINKED ISSUE #10552: 88e4f7d1 validate callee ID for TSO requests
cp [`88e4f7d1`](https://github.com/bufferflies/pd/commit/88e4f7d152f2ead7cc116c8b0ae36078ee7c50b1)

bufferflies:

Issue Number: Close #10552, ref #10516

Author: @iosmanthus

Cherry-pick: [`88e4f7d1`](https://github.com/bufferflies/pd/commit/88e4f7d152f2ead7cc116c8b0ae36078ee7c50b1)

The TSO client can keep using a stale gRPC connection after the service endpoint changes. This PR carries the expected callee ID in TSO requests so the server can reject requests that reach the wrong endpoint, allowing the client to drop stale connections and reconnect.

```commit-message
Add callee ID validation for TSO requests.

Attach the callee ID to TSO request headers and validate it in the TSO gRPC service against the advertised listen address.

Treat callee ID mismatch as a reconnect signal in the TSO client, and add service discovery support for removing stale gRPC connections.

Update affected tests and kvproto-dependent modules for the new request field.
```

### Check List

Tests

- Unit test

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Fix the issue that the TSO client may keep using a stale gRPC connection after the TSO service endpoint changes.
```


iosmanthus (approved):

client/clients/tso/dispatcher.go
```diff
@@ -375,13 +375,17 @@ func (td *tsoDispatcher) handleProcessRequestError(
 	conCtxMgr.Release(streamURL)
 	// Update the member list to ensure the latest topology is used before the next batch.
 	svcDiscovery := td.provider.getServiceDiscovery()
-	if errs.IsLeaderChange(err) {
+	switch {
+	case errs.IsCalleeMismatch(err):
```
rleungx, line 378: We need a test to cover this change.

pkg/mcs/tso/server/grpc_service.go
```diff
@@ -60,7 +63,8 @@ func NewService(svr bs.Server) registry.RegistrableService {
 		log.Fatal("create tso server failed")
 	}
 	return &Service{
-		Server: server,
+		Server:              server,
+		advertiseListenHost: getAdvertiseListenHost(server.GetAdvertiseListenAddr()),
```
rleungx: Will the startServer init advertiseListenHost?

bufferflies: yes，reduce the net parser cost

client/clients/tso/dispatcher.go
bufferflies: done

client/clients/tso/client.go
```diff
@@ -198,11 +198,15 @@ func (c *Cli) getTSOLeaderClientConn() (*grpc.ClientConn, string) {
 	if len(url) == 0 {
 		log.Fatal("[tso] the tso leader should exist")
 	}
-	cc, ok := c.svcDiscovery.GetClientConns().Load(url)
-	if !ok {
+	cc, err := c.svcDiscovery.GetOrCreateGRPCConn(url)
```
lhy1024, line 201: Do we need to cover discovery or unary? They also reuse the same gRPC connection, such as FindGroupByKeyspaceID / GetMinTS.

client/servicediscovery/tso_service_discovery.go
```diff
@@ -318,6 +318,17 @@ func (c *tsoServiceDiscovery) GetOrCreateGRPCConn(url string) (*grpc.ClientConn,
 	return grpcutil.GetOrCreateGRPCConn(c.ctx, &c.clientConns, url, c.tlsCfg, c.option.GRPCDialOptions...)
 }
 
+// RemoveClientConn removes and closes the gRPC connection of the given URL.
+func (c *tsoServiceDiscovery) RemoveClientConn(url string) {
```
lhy1024, line 322: Do we also need to clear the cached `*grpc.ClientConn` in serviceClient?

lhy1024 (commented):
the rest LGTM

client/clients/tso/client.go
bufferflies: FindGroupByKeyspaceID / GetMinTS -> all the tso services can handle this, it doesn't matter if the node mismatches.

pkg/mcs/tso/server/server.go
```diff
@@ -91,14 +91,23 @@ type Server struct {
...
+
+	// advertiseListenHost is the host part of the advertise listen address,
+	advertiseListenHost atomic.Pointer[string]
 }
 
-func getAdvertiseListenHost(advertiseListenAddr string) string {
-	parsed, err := url.Parse(advertiseListenAddr)
-	if err != nil || parsed.Host == "" {
-		return ""
+func (s *Server) getAdvertiseListenHost() string {
```
rleungx, line 99: Why do we need to use cas here?

bufferflies: This function is called by many threads, so it has a concurrent risk. I think the store is also ok for the last win.

rleungx: why?The address won't be changed after the server is started.

bufferflies: done

pkg/mcs/tso/server/server.go
```diff
@@ -403,10 +389,16 @@ func (s *Server) startServer() (err error) {
 
 // CreateServer creates the Server
 func CreateServer(ctx context.Context, cfg *Config) *Server {
+	addr := cfg.GetAdvertiseListenAddr()
+	parsed, err := url.Parse(addr)
+	if err != nil {
+		panic(fmt.Sprintf("invalid advertise listen address: %s", addr))
```
rleungx, line 395: How about returning an error?

client/servicediscovery/tso_service_discovery.go
bufferflies: Good catch, GetClientConn will check the connection state. If it's shundown, it will return nil and then update conn later

pkg/mcs/tso/server/server.go
bufferflies: No, it must tell the user the PD server address is illegal in time.

client/servicediscovery/service_discovery.go
```diff
@@ -253,6 +256,10 @@ func (c *serviceClient) GetClientConn() *grpc.ClientConn {
 	if c == nil {
 		return nil
 	}
+	// If the connection is in Shutdown state, it means the connection is closed and we should not reuse it.
+	if c.conn.GetState() == connectivity.Shutdown {
```
lhy1024, line 260: This can introduce a nil pointer panic. newPDServiceClient explicitly allows conn == nil and marks the client as unavailable via networkFailure; for example, when creating a follower connection fails, a serviceClient with a nil conn can still be stored in followers. GetClientConn() now only checks  c == nil and then calls c.conn.GetState(), so the next member update or caller-side GetClientConn() check can dereference nil and crash the client.

Please keep the nil-conn guard before checking the Shutdown state:
```
  if c == nil || c.conn == nil {
      return nil
  }
  if c.conn.GetState() == connectivity.Shutdown {
      return nil
  }
```
It would also be good to add a unit test covering newPDServiceClient(..., nil, ...).GetClientConn() returning nil.

rleungx, line 260: Please keep the nil connection guard here; `newPDServiceClient` can store a nil conn after dial failure.

client/clients/tso/dispatcher.go
```diff
@@ -375,13 +374,17 @@ func (td *tsoDispatcher) handleProcessRequestError(
...
```
rleungx, line 378: After removing the stale connection, `GetTS` should retry this error too; otherwise the request that detects the mismatch is still returned to the caller as a failure.

client/servicediscovery/service_discovery.go
bufferflies: Yes, if c.conn.GetState() == connectivity.Shutdown is true, I will set conn nil.

bufferflies: fixed

client/clients/tso/dispatcher.go
bufferflies: func isRetryableGetTSError(err error) bool {
	return errs.IsLeaderChange(err) || errs.IsCalleeMismatch(err)
} 

fixed in isRetryableGetTSError

client/servicediscovery/service_discovery.go
```diff
@@ -253,6 +256,10 @@ func (c *serviceClient) GetClientConn() *grpc.ClientConn {
 	if c == nil {
 		return nil
 	}
+	// If the connection is in Shutdown state, it means the connection is closed and we should not reuse it.
+	if c.conn != nil && c.conn.GetState() == connectivity.Shutdown {
+		c.conn = nil
```
rleungx: `GetClientConn` can be called concurrently, so mutating `c.conn` here introduces a data race. Please either return nil without changing the field or protect `conn` with synchronization.

bufferflies: yes, revert this write path

client/servicediscovery/service_discovery.go
```diff
@@ -250,7 +253,11 @@ func (c *serviceClient) checkNetworkAvailable(ctx context.Context) {
 
 // GetClientConn implements ServiceClient.
 func (c *serviceClient) GetClientConn() *grpc.ClientConn {
-	if c == nil {
+	if c == nil || c.conn != nil {
```
rleungx: This condition looks inverted. It should probably be `c.conn == nil`; otherwise a valid connection returns nil, and a nil connection will panic below.

bufferflies: done

client/clients/tso/client.go
rleungx, line 201: This only handles stale connections on the streaming TSO path. `FindGroupByKeyspaceID` still reuses the cached TSO connection without sending `CalleeId`, so if that unary request hits the stale connection first, it can keep failing. Could we add the same callee-id check and eviction there too?

bufferflies: No,  the caller ID is just the target URL, so in the create connection step,  the server doesn't reject the connection if the IP mismatches. Currently, the response doesn't carry the server caller ID.

In the other way, FindGroupByKeyspaceID request can be sent to a mismatched tso group server, it can return the correct tso infos.

rleungx: I think this still matters before stream creation. `FindGroupByKeyspaceID` does not always get correct info from any TSO node: if the contacted node has the KG metadata but no local allocator, `getKeyspaceGroupMetaWithCheck` returns NotServed and the response has only an error header. Since PD exposes all TSO service URLs and KG replicas default to 2, a stale cached connection can keep hitting a non-member without eviction. Could we add `CalleeId` to this unary request too, or make the server return the KG metadata even when the local allocator is nil?

pkg/mcs/tso/server/server.go
```diff
@@ -385,10 +389,16 @@ func (s *Server) startServer() (err error) {
 
 // CreateServer creates the Server
 func CreateServer(ctx context.Context, cfg *Config) *Server {
+	addr := cfg.GetAdvertiseListenAddr()
+	parsed, err := url.Parse(addr)
```
lhy1024, line 394: Checking only the error from url.Parse is not enough here. An address like localhost:3379 can be parsed successfully, but parsed.Host is empty, so advertiseListenHost also becomes empty. The Tso handler then skips callee-id validation because host == "", which silently disables this protection for a malformed address.

  If the intent is to fail fast on an invalid advertise listen address, could we also check parsed.Host != "" when addr != "", and add a test case such as localhost:3379?

client/clients/tso/dispatcher.go
```diff
@@ -375,13 +374,17 @@ func (td *tsoDispatcher) handleProcessRequestError(
...
+		// If the callee ID mismatches, the existing gRPC connection is stale and must be recreated.
+		svcDiscovery.RemoveClientConn(streamURL)
```
lhy1024, line 380: This eviction only covers the streaming TSO path. Before the stream is created, FindGroupByKeyspaceID also reuses the same cached connection, but it does not carry or validate CalleeId. If a stale conn first hits the wrong endpoint in this unary discovery path, updateMember() returns the error directly and RemoveClientConn is not called.

  Could we also make the FindGroupByKeyspaceID path clear the cached connection and retry when it reaches a mismatched/stale endpoint?

pkg/mcs/tso/server/server.go
bufferflies: Yes,fixed and added unit test for localhost:3379 and 127.0.0.1:2379

client/clients/tso/dispatcher.go
bufferflies: done, fixed

client/clients/tso/client.go
bufferflies: done, fixed

rleungx (approved):

lhy1024 (approved):

ystaticy (approved):

Issue #10488, opened 2026-03-25
resource control: improve observability for Controller and Server modules


JmPotato:
## Enhancement Task

Improve the observability of Resource Control related modules on both the client-side Controller (`client/resource_group/controller`) and the server-side Resource Manager (`pkg/mcs/resourcemanager/server`). The current metrics and logging cover aggregate consumption well but lack visibility into allocation decisions, per-client behavior, and internal token state — all of which are critical for debugging resource contention and fairness in multi-tenant deployments.

### Motivation

- **Operators** cannot easily diagnose why a specific resource group or client is being throttled.
- **Developers** lack insight into how the demand-aware slot allocation algorithm distributes fill rates across clients.
- Several important internal states (limiter token balance, per-slot fill rate, RU demand samples, token depletion rate) are only logged at debug level or not exposed at all.

#### 1. Client-Side Controller Improvements

| Area | Current State | Proposed Improvement |
|---|---|---|
| **Token consumption per request** | `TokenConsumedHistogram` only observed in trace mode | Always observe token consumption; break down by RU type (read/write/SQL) |
| **Token depletion rate** | Only "low token notify" counter exists | Add metrics for token consumption rate vs. refill rate to detect imbalance before throttling |
| **Limiter state visibility** | Only active/tombstone status gauge | Expose current token balance, effective fill rate, and burst limit per resource group |
| **RU calculation breakdown** | Consumption sent to server as aggregate | Add per-component (KV CPU, read/write bytes, SQL CPU) consumption metrics on the client side |
| **Request wait time** | Only failed-request wait time histogram | Add wait time histogram for successful requests to understand latency impact of rate limiting |

#### 2. Server-Side Resource Manager Improvements

| Area | Current State | Proposed Improvement |
|---|---|---|
| **Per-slot fill rate allocation** | Calculated in memory, only debug-level logging | Add metrics exposing allocated fill rate per client/slot within a resource group |
| **Per-slot token grants** | `"assign slot tokens"` debug log only | Add histogram for token quantities granted per request to each client |
| **RU demand sampling** | Tracked internally in `ruTracker.sample()`, not exposed | Add metrics for sampled RU/sec per client to show which clients drive allocation decisions |
| **Slot lifecycle** | Logged at debug level only | Add gauge for active slot count per resource group and counter for slot creation/deletion |
[...]

PR #10604, opened 2026-04-20
client/resource_group: add client-side demand observability

okJiang:
Issue Number: ref #10488, ref #10581

Resource control observability on the client side does not show the difference between real demand and post-throttling consumption. It is also hard to tell whether pressure is local to one TiDB instance or spread across the whole resource group.

This PR adds client-side metrics and slow-path logs so we can explain demand, limiter state, and per-instance pressure before requests are throttled.

```commit-message
client/resource_group: add client-side demand observability
```

It includes:
- a pre-throttling `demand_ru_per_sec` metric collected before `Reserve()`
- client-side balance, fill rate, burst limit, average RU/s, throttled state, and consumption breakdown metrics
- slow-path structured logs for degraded mode, slow token requests, long waits, and throttling
- cleanup coverage so stale resource-group labels are removed correctly

### Check List

Tests

- Unit test

Manual test:
- Verified in a local `tiup playground` setup with 3 PD, 2 TiDB, and 3 TiKV.
- Created a low-fill-rate resource group and confirmed Grafana shows `demand_ru_per_sec > fill_rate`, higher client wait, and throttling on the hot TiDB instance.
- Confirmed stale metrics are cleaned when the resource group is removed.

Code changes

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Add client-side resource control metrics for pre-throttling demand, limiter state, and per-instance RU consumption breakdowns.
```

## Summary by CodeRabbit

* **New Features**
  * Expanded resource-group telemetry with RU consumption by type, token balance/fill rate/burst limits, average and demand RU/s, throttled-state gauge, and read/write byte + KV/SQL CPU cost counters.

* **Improvements**
  * Enhanced structured logging for token and throttling events, including client identity and involved resource groups.
  * Improved concurrent safety for limiter read-only inspections.
  * Improved per-group metric cleanup to remove additional stale/inactive metric label values.

* **Tests**
  * Added/updated unit tests to validate new metric emissions, throttling/demand behavior, concurrent logging behavior, and metric cleanup.

### DIFF
--- a/client/resource_group/controller/group_controller.go
+++ b/client/resource_group/controller/group_controller.go
@@ -412,7 +412,7 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 }
 
 func (gc *groupCostController) calcAvg(counter *tokenCounter, new float64) bool {
-	return gc.calcMovingAvgAt(gc.run.now, &counter.avgRUPerSec, &counter.avgRUPerSecLastRU, &counter.avgLastTime, new)
+	return calcMovingAvgAt(gc.run.now, &counter.avgRUPerSec, &counter.avgRUPerSecLastRU, &counter.avgLastTime, new)
 }
 
 func (gc *groupCostController) calcDemandAvg(counter *tokenCounter, delta float64) bool {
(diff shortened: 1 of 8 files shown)


client/resource_group/controller/group_controller.go
```diff
@@ -301,21 +351,40 @@ func (gc *groupCostController) handleTokenBucketUpdateEvent(ctx context.Context)
 }
 
 func (gc *groupCostController) calcAvg(counter *tokenCounter, new float64) bool {
-	deltaDuration := gc.run.now.Sub(counter.avgLastTime)
+	return calcMovingAvgAt(gc.run.now, &counter.avgRUPerSec, &counter.avgRUPerSecLastRU, &counter.avgLastTime, new)
+}
+
+func (gc *groupCostController) calcDemandAvg(counter *tokenCounter, delta float64) bool {
```
YuhaoZhang00, line 423: `calcDemandAvg` is called from `acquireTokens` request goroutine so data race might happen.

Suggestion: accumulate demandTotalRU under gc.mu in the request path, compute the EMA in the main loop — same pattern as avgRUPerSec.

client/resource_group/controller/group_controller.go
```diff
@@ -571,11 +698,19 @@ func (gc *groupCostController) onRequestWaitImpl(
 			failpoint.Inject("triggerUpdate", func() {
 				gc.lowRUNotifyChan <- notifyMsg{}
 			})
+			if waitDuration > slowNotifyFilterDuration {
```
YuhaoZhang00, line 796: slowNotifyFilterDuration == 10 ms might be too short for token wait scenario? (this also applies to other log.Warn added in this PR)

lhy1024:
Dependency update: #10997 fixed this logging issue on `release-nextgen-202603` and merged in 19ee324ee. The equivalent change has now been applied to this PR in b8a835cd3 and 916fcbad5. Successful-wait WARNs are removed while failed-wait WARNs and the existing `SuccessfulRequestDuration` semantics are preserved.

client/resource_group/controller/group_controller.go
lhy1024: Agreed. A successful token wait is expected throttling behavior, so logging every successful wait over 10ms at WARN is too noisy under high QPS. #10997 removed these successful-wait WARNs on `release-nextgen-202603` and has merged. I applied the equivalent change here in b8a835cd3 and 916fcbad5: successful request-side and response-side wait WARNs are removed, failed-wait WARNs remain, and `SuccessfulRequestDuration` preserves its existing semantics.

PR #10606, opened 2026-04-20
gc_state_manager: Make methods of GCStateManager able to timeout or cancel

### LINKED ISSUE #10607: GCStateManager calls may be blocked and unable to cancel or timeout when there's IO issue
## Enhancement Task

`GCStateManager` needs to acquire a mutex and perform IO on the etcd while holding the mutex. This actually has risks: if some requests encounters slow IO, other incoming requests may be blocked on the mutex and unable to be canceled or timeout before finally successfully acquiring the mutex.

There's some way we can do to optimize the case:

* Replace the mutex with some other synchronizing mechanism that can be canceled or timeout while waiting.
* Use read lock instead of exclusive lock in GetGCState request (using exclusive lock here is actually a bug)
* When multiple GetGCState requests ont the same keyspace are executed concurrently, try to share the results if possible (using the ordered singleflight, like how `GetAllkeyspacesGCStates` does)

Further, it might also be a problem that `StorageEndpoint` (the abstract layer of etcd) doesn't provide APIs with contexts, making outer etcd IO unable to cancel.

MyonKeminta:

Issue Number: Close #10607

In this PR:

* Implemented a simple cancellable `RWMutex` by using semaphore internally
* Added `OrderedSingleFlightGroup`
* In `GCStateManager`, Replace the original `RWLock` with the `CancellableRWMutex`, and set timeout for its APIs
* Use read lock instead of write lock in GetGCState

We've encountered a case that some GC APIs are accumulating waiting for the mutex. By making these changes, the case should be able to be *alleviated*.

```commit-message
gc_state_manager: Make methods of GCStateManager able to timeout or cancel

* Implemented a simple cancellable `RWMutex` by using semaphore internally
* Added `OrderedSingleFlightGroup`
* In `GCStateManager`, Replace the original `RWLock` with the `CancellableRWMutex`, and set timeout for its APIs
* Use read lock instead of write lock in GetGCState
```

### Check List

Tests

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -113,7 +113,7 @@ type GCStateManager struct {
 	// concurrent operations happen on a single PD leader) it doesn't need to cause conflicts in etcd transactions
 	// layer. It can be more efficient and avoid failures due to transaction conflict in most cases.
 	// The etcd transactions is still necessary considering the possibility of rare cases like PD leader changes.
-	mu              syncutil.RWMutex
+	mu              syncutil.CancellableRWMutex
 	gcMetaStorage   endpoint.GCStateProvider
 	cfg             config.PDServerConfig
 	keyspaceManager *keyspace.Manager
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -118,6 +118,8 @@ type GCStateManager struct {
 	cfg             config.PDServerConfig
 	keyspaceManager *keyspace.Manager
 
+	gcStateSingleFlightGroup *syncutil.OrderedSingleFlightGroup[uint32, GCState]
+
 	allKeyspacesGCStatesSingleFlight *syncutil.OrderedSingleFlight[map[uint32]GCState]
 }
 
@@ -127,6 +129,7 @@ func NewGCStateManager(store endpoint.GCStateProvider, cfg config.PDServerConfig
 		gcMetaStorage:                    store,
 		cfg:                              cfg,
 		keyspaceManager:                  keyspaceManager,
+		gcStateSingleFlightGroup:         syncutil.NewOrderedSingleFlightGroup[uint32, GCState](),
 		allKeyspacesGCStatesSingleFlight: syncutil.NewOrderedSingleFlight[map[uint32]GCState](),
 	}
 }
@@ -179,7 +182,7 @@ func (m *GCStateManager) redirectKeyspace(keyspaceID uint32, isUserAPI bool) (ui
 }
 
 // CompatibleLoadGCSafePoint loads current GC safe point from storage for the legacy GC API `GetGCSafePoint`.
-func (m *GCStateManager) CompatibleLoadGCSafePoint(keyspaceID uint32) (uint64, error) {
+func (m *GCStateManager) CompatibleLoadGCSafePoint(_ context.Context, keyspaceID uint32) (uint64, error) {
 	keyspaceID, _, err := m.redirectKeyspace(keyspaceID, false)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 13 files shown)


pkg/gc/gc_state_manager.go
```diff
@@ -113,11 +113,13 @@ type GCStateManager struct {
 	// concurrent operations happen on a single PD leader) it doesn't need to cause conflicts in etcd transactions
 	// layer. It can be more efficient and avoid failures due to transaction conflict in most cases.
 	// The etcd transactions is still necessary considering the possibility of rare cases like PD leader changes.
-	mu              syncutil.RWMutex
+	mu              syncutil.CancellableRWMutex
```
rleungx, line 116: Why do we need a cancellable mutex?

pkg/gc/gc_state_manager.go
```diff
@@ -671,13 +689,28 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
...
+	ctx = addKeyspaceNameToCtx(ctx, keyspaceNameForDiag)
+
+	return m.gcStateSingleFlightGroup.Do(ctx, keyspaceID, func(execCtx context.Context) (GCState, error) {
+		result, err := m.getGCStateImpl(execCtx, keyspaceID)
+		failpoint.Inject("onGetGCStateFinish", func() {})
+		return result, err
+	})
+}
+
+func (m *GCStateManager) getGCStateImpl(ctx context.Context, keyspaceID uint32) (GCState, error) {
```
rleungx, line 706: When the context is cancelled, it cannot cancel the transaction.

pkg/utils/syncutil/cancellable_rw_mutex.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (m *CancellableRWMutex) init() {
+	m.initOnce.Do(func() {
+		m.inner = semaphore.NewWeighted(maxReaderCount)
+	})
+}
+
+// RLock acquires a read lock or returns an error if ctx is canceled before the
+// lock is acquired.
+func (m *CancellableRWMutex) RLock(ctx context.Context) error {
```
rleungx, line 59: The deadlock testing does not work anymore.

pkg/utils/syncutil/cancellable_rw_mutex.go
```diff
@@ -0,0 +1,149 @@
...
+	// -1 when a writer holds the lock, 0 when unlocked, and >0 for the
+	// number of active readers.
+	// This is not part of the implementation of the locking mechanism, but only for detecting misuses like
+	// unpaired locking and unlocking calls.
+	state atomic.Int64
+}
+
+func (m *CancellableRWMutex) init() {
+	m.initOnce.Do(func() {
+		m.inner = semaphore.NewWeighted(maxReaderCount)
```
rleungx, line 53: Will the read be blocked more easily?

pkg/gc/gc_state_manager.go
MyonKeminta: As it may involve etcd IO in its critical section, and it may need a long time to successfully acquire the lock. Without the ability to cancel or timeout, the goroutines waiting for it would be unable to exit, and the memory used by them can't be released either.

pkg/gc/gc_state_manager.go
MyonKeminta: This is a problem but it's unable to pass ctx to the StorageEndpoint API now. Seems it requires a large refactor.

pkg/utils/syncutil/cancellable_rw_mutex.go
MyonKeminta: The overhead of the lock itself is larger than the builtin `RWMutex`, but  in the current use case I think it won't

PR #10611, opened 2026-04-21
client/resource_group: add RC paging pre-charge with PredictedReadBytes hint

### LINKED ISSUE #10612: client/resource_group: add RC paging pre-charge controller consuming client-supplied read-bytes hint
### Problem

Under RC (resource control), paging coprocessor requests are today billed only at response settlement. A misbehaving resource group can therefore burst large reads before its token bucket reacts, because no RU is reserved up front.

The controller has no pre-charge primitive today — every cost path ends at `RemoveTokens` / `Reserve` driven by *completed* work. There is no way to reserve tokens against a prediction, and no inverse primitive to refund when the prediction overshoots.

### Proposal

Add an opt-in pre-charge mechanism to `client/resource_group/controller` that consumes a caller-supplied read-bytes prediction:

- **Hint intake.** Read an optional `PredictedReadBytes` from `RequestInfo` via a duck-typed `predictedReadBytesProvider` interface — older client-go implementations keep working unchanged.
- **Pre-charge on request.** In `BeforeKVRequest`, when the hint is > 0, add `ReadBytesCost * predicted` to `delta.RRU` so the existing `acquireTokens` path reserves tokens via the limiter.
- **Settlement with refund.** In `AfterKVRequest`, subtract the predicted basis and add `ReadBytesCost * actual`. The resulting delta can be negative (over-estimate), so introduce a `Limiter.RefundTokens` primitive (inverse of `RemoveTokens`) and branch refund vs. consume by sign.
- **Observability.** Per-resource-group counters and histograms grouped by sampling unit (count → bytes → RU): pre-charged vs. cold (unhinted) request counts, predicted/actual byte sums, prediction residual histogram (`actual − predicted`), pre-charge RU vs. settlement RU, and per-RPC settlement-RU delta histogram.

Correctness: total RU billed for a pre-charged request is exactly `ReadBytesCost * actual` — identical to today.

### Related issues

- client-go transport: tikv/client-go#1953
- TiDB caller: pingcap/tidb#68085

### Non-Goals

- No change to the wire protocol between TiDB/TiKV or TiDB/PD.
- No change to the RU cost model. Pre-charge only shifts *when* tokens are reserved, not how many are billed overall.
- No change to behavior when a client does not implement the optional hint or reports zero — existing billing is preserved exactly.
- NextGen keyspace billing basis (where `max(TotalVersionsSize, ProcessedVersionsSize)` is correct) is out of scope; tracked as a future follow-up.

YuhaoZhang00:

Issue Number: close #10612

Under RC (resource control), paging coprocessor requests are today billed only at response settlement. A misbehaving resource group can therefore burst large reads before its token bucket reacts, because no RU is reserved up front. This PR introduces a pre-charge mechanism driven by a caller-supplied read-bytes prediction so the token bucket is pressured before the KV RPC is sent, while keeping the final bill accurate at settlement.

The main logic is in `client/resource_group/controller`, with matching resource-manager server-side metrics/metering handling.

1. **Pre-charge on request** - `BeforeKVRequest` reads an optional `PredictedReadBytes()` provider from `RequestInfo`. The hint is used only when the request is a read coprocessor request (`IsCop() == true` and `IsWrite() == false`). When the hint is positive, `KVCalculator` adds `ReadBytesCost * predicted` to `delta.RRU`, and the existing token acquisition path reserves that many tokens before the KV RPC is sent. Writes, non-coprocessor reads, and coprocessor reads without a positive hint keep the existing settlement-only behavior.

2. **Refund on settlement** - `AfterKVRequest` settles the pre-charge by subtracting the predicted basis and adding `ReadBytesCost * actual`. The resulting token delta can be negative when the estimate was too high, so `Limiter.RefundTokens` is used as the inverse of `RemoveTokens`. Reported consumption strips the synthetic pre-charge so pre-charged reads still report the same actual request consumption as before: request-side read base cost, then response-side actual read bytes and CPU.

3. **Consumption reporting** - The synthetic paging pre-charge/refund delta is used only for local token accounting. It is stripped from `ConsumptionSinceLastRequest`, so the resource-manager server still receives the same actual consumption as before. Empty consumption deltas are still reported as empty `Consumption` messages, preserving the existing metric activity refresh behavior; nil consumption is ignored defensively on the server side.

4. **Observability and metering safeguards** - five per-resource-group paging metrics are added, grouped by sampling unit. Server-side metering now also clamps negative RU values to zero before converting them to unsigned metering values, so malformed or defensive negative RU input cannot become an extremely large unsigned value.

   Count:
   - `cop_read_precharge_total` - read coprocessor RPCs that carried a positive `PredictedReadBytes` hint and triggered request-side read-byte pre-charge
   - `cop_read_no_precharge_total` - read coprocessor RPCs that reached the controller without a positive `PredictedReadBytes` hint, so request-side read-byte pre-charge was skipped

   Bytes:
   - `paging_precharge_bytes_total` - sum of predicted hints used as the pre-charge basis
   - `paging_actual_bytes_total` - actual bytes read by pre-charged RPCs
   - `paging_prediction_residual_bytes` - distribution of signed `actual - predicted` for pre-charged RPCs

Pre-charge is strictly opt-in: a client must expose `PredictedReadBytes()` and pass a non-zero value on a read coprocessor request. Passing 0, omitting the optional provider, or using any write/non-coprocessor request preserves the existing settlement-only path. The wire protocol is unchanged.

### Related PRs

- client-go side: tikv/client-go#1947
- TiDB caller side: pingcap/tidb#67941
- failed-write accounting split: tikv/pd#10986 (refs tikv/pd#10985)

### Check List

Tests

- Unit test
  - `TestPredictedReadBytesPreCharge` / `TestPredictedReadBytesRequiresCopRead` / `TestRequestInfoPagingProvidersAreOptional` / `TestRequestInfoMissingPredictionProviderReturnsZeroHint` / `TestNoPreChargeWithoutPredictedReadBytes` - request-side pre-charge gating
  - `TestPagingPreChargeTokenRefund` / `TestPagingPreChargeNoRefundWhenActualExceedsEstimate` / `TestOnResponseImplPagingRefund` / `TestOnResponseWaitPrechargedPositiveSettlementDebitsImmediately` / `TestPagingPreChargeRefundOnFailedRead` - response-side settlement and refund
  - `TestPagingPrechargeDoesNotEnterReportedRequestConsumption` / `TestPagingPrechargeRefundDoesNotEnterReportedResponseConsumption` / `TestReportedConsumptionStripsPagingPrecharge` / `TestReportedConsumptionRestoresPagingActualUsage` - reported consumption excludes synthetic pre-charge
  - `TestDeletePagingLabelsResetsSeries` / `TestCopReadNoPrechargeGatedByIsCop` / `TestCopReadNoPrechargeCounterObservedOnRequest` / `TestPagingPrechargeNotObservedOnThrottle` - paging metric behavior
  - `TestUpdateDeltaConsumptionReportsEmptyConsumption` / `TestRecordConsumptionKeepsRecordForEmptyConsumption` / `TestRecordConsumptionIgnoresNilConsumption` / `TestRUCollectorUsesConsumption` / `TestNewRUValueClampsNegativeValues` - consumption reporting, server metrics, and metering handling
  - `TestFailedWriteDoesNotUsePagingRefundPath` / `TestFailedWriteOnResponseDoesNotUsePagingRefundPath` - split boundary for the failed-write accounting fix
  - `TestRefundTokens` - limiter primitive

Side effects

- Increased code complexity (optional paging hint provider + coprocessor gating + refund path).
- A paging over-estimate can refund local limiter tokens through `RefundTokens`; this affects only the local token bucket and does not enter reported consumption.
- Negative RU values passed into metering are clamped to zero before unsigned conversion.

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Added paging-aware request accounting, including pre-charge, settlement, and token refund handling for estimated reads.
  * Expanded metrics to show paging pre-charge counts, actual bytes, and prediction accuracy.

* **Bug Fixes**
  * Improved RU and token accounting so reported consumption reflects actual usage more accurately.
  * Prevented nil consumption records from affecting metering.
  * Clamped negative RU values to zero for safer metering output.

* **Tests**
  * Added broad coverage for paging behavior, limiter refunds, metric cleanup, and consumption reporting.


client/resource_group/controller/model.go
```diff
@@ -138,6 +161,10 @@ func (kc *KVCalculator) AfterKVRequest(consumption *rmpb.Consumption, req Reques
 	if !req.IsWrite() {
 		// For now, we can only collect the KV CPU cost for a read request.
 		kc.calculateCPUCost(consumption, res)
+		// Paging settlement
+		if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+			consumption.RRU -= float64(kc.ReadBytesCost) * float64(bytesForEst)
+		}
```
JmPotato, line 201: If I remember correctly, the logic here is only reached when the response is non-empty (meaning no error occurred).

If a request fails, shouldn't the previously deducted RU be refunded?

client/resource_group/controller/group_controller.go
```diff
@@ -601,10 +638,20 @@ func (gc *groupCostController) onResponseImpl(
 	for _, calc := range gc.calculators {
 		calc.AfterKVRequest(delta, req, resp)
 	}
+	if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+		gc.metrics.observePagingActual(bytesForEst, resp.ReadBytes())
+	} else if !req.IsWrite() {
+		if _, ok := req.(predictedReadBytesProvider); ok {
```
JmPotato: Looking at the code, all requests implement this interface, meaning they will all enter this branch under any path.

Given that, what is the purpose of this non-precharge metric?

client/resource_group/controller/model.go
YuhaoZhang00: The refund doesn't happen here. The actual refund lives in `onResponse{Wait}Impl`, where `v < 0` -> RefundTokens. So on failure, the pre-charge is correctly returned via that path.

Added `TestPagingPreChargeRefundOnFailedRead` unit test to verify.

client/resource_group/controller/group_controller.go
```diff
@@ -601,23 +651,30 @@ func (gc *groupCostController) onResponseImpl(
...
+	// `count` is the full per-request consumption (BeforeKVRequest + AfterKVRequest).
+	count := &rmpb.Consumption{}
+	*count = *delta
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(count, req)
+	}
+	if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+		gc.metrics.observePagingActual(bytesForEst, resp.ReadBytes(),
+			getRUValueFromConsumption(count), getRUValueFromConsumption(delta))
+	} else if !req.IsWrite() {
```
rleungx: After `PredictedReadBytes` became part of every `RequestInfo`, this branch counts every read with a zero prediction, not only paging reads. Can we add an explicit paging/hint-present signal before recording `paging_nonprecharge_*`?

client/resource_group/controller/limiter.go
```diff
@@ -325,6 +325,20 @@ func (lim *Limiter) RemoveTokens(now time.Time, amount float64) {
...
+// No burst cap is applied here
+func (lim *Limiter) RefundTokens(now time.Time, amount float64) {
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+	if lim.burst < 0 || lim.fillRate == Inf {
+		return
+	}
+	_, tokens := lim.getTokens(now)
+	lim.last = now
+	lim.tokens = tokens + amount
```
rleungx, line 344: Refunding tokens updates the bucket, but existing waiters still sleep on their original timer in `WaitReservations`. Can `RefundTokens` notify or wake waiters so over-estimated precharges release blocked requests promptly?

client/resource_group/controller/group_controller.go
```diff
@@ -122,9 +134,44 @@ func initMetrics(oldName, name string) *groupMetricsCollection {
 		tokenRequestCounter:               metrics.ResourceGroupTokenRequestCounter.WithLabelValues(oldName, name),
 		runningKVRequestCounter:           metrics.GroupRunningKVRequestCounter.WithLabelValues(name),
 		consumeTokenHistogram:             metrics.TokenConsumedHistogram.WithLabelValues(name),
+
+		prechargeCounter:        metrics.PagingPrechargeCounter.WithLabelValues(name),
```
rleungx: These new per-resource-group metric series also need cleanup on group deletion/tombstone cleanup. Otherwise deleted groups can leave stale `Paging*` series behind.

client/resource_group/controller/group_controller.go
```diff
@@ -632,40 +710,76 @@ func (gc *groupCostController) onResponseWaitImpl(
...
 		calc.AfterKVRequest(delta, req, resp)
 	}
+	// `count` is the full per-request consumption (BeforeKVRequest + AfterKVRequest).
+	count := &rmpb.Consumption{}
+	*count = *delta
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(count, req)
+	}
+	if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+		gc.metrics.observePagingActual(bytesForEst, resp.ReadBytes(),
```
rleungx: Can we move this paging settlement metric observation until after the positive-delta `acquireTokens` path succeeds? In `onResponseWaitImpl`, `acquireTokens` below can still return an error and the function returns before updating `gc.mu.consumption`, `storeCounter`, or `globalCounter`. In that case `paging_actual_bytes_total` / `paging_settlement_ru_total` would record a request that was not actually settled, which makes the “finally consumed” metrics drift from controller accounting. The `OnResponse` path is fine because it cannot fail after observation; this risk is specific to `OnResponseWait`.

client/resource_group/controller/group_controller.go
```diff
@@ -601,23 +672,30 @@ func (gc *groupCostController) onResponseImpl(
...
+	} else if !req.IsWrite() && req.IsCop() {
+		gc.metrics.observePagingNonprecharge(resp.ReadBytes())
+	}
 	if !gc.burstable.Load() {
 		counter := gc.run.requestUnitTokens
 		if v := getRUValueFromConsumption(delta); v > 0 {
 			counter.limiter.RemoveTokens(time.Now(), v)
+		} else if v < 0 {
+			// Paging over-estimate: refund the excess pre-charge.
+			counter.limiter.RefundTokens(time.Now(), -v)
```
YuhaoZhang00, line 697: This refund path handles paging over-estimation correctly for the token bucket, but the server-side RU counter still needs matching net/refund semantics.

Today PD server `counterMetrics.add()` only adds `consumption.RRU` to `resource_manager_resource_unit_read_request_unit_sum` when `RRU > 0`. A negative settlement delta from this path can refund tokens/internal consumption, but it is ignored by the Prometheus counter because counters cannot go down. That makes Grafana `RU Per Group` report gross precharge RU instead of net settled RU in over-precharge cases.

We should add a refund/negative-RU metric or explicit net-RU metric on the server side, update the dashboard query to use net RU, and add a regression test for `+precharge` followed by `-refund`.

client/resource_group/controller/limiter.go
```diff
@@ -325,6 +325,36 @@ func (lim *Limiter) RemoveTokens(now time.Time, amount float64) {
...
+// against the at-reserve-time token count); reservation re-evaluation
+// is out of scope here.
+func (lim *Limiter) RefundTokens(now time.Time, amount float64) {
+	lim.mu.Lock()
+	defer lim.mu.Unlock()
+	if lim.burst < 0 || lim.fillRate == Inf {
+		return
+	}
+	_, tokens := lim.getTokens(now)
+	lim.last = now
```
JmPotato: Should use `updateLast`?

client/resource_group/controller/model.go
```diff
@@ -52,6 +52,26 @@ type RequestInfo interface {
...
+	// bounded-size reads bypass the paging metrics even though they may
+	// reach this controller through the same RC interceptor.
+	IsCop() bool
+}
+
+// estimatedReadBytes returns the predicted read-bytes hint for read requests.
+// Writes always return 0 so paging pre-charge / settlement / metrics stay
+// gated to reads.
+func estimatedReadBytes(req RequestInfo) uint64 {
+	if req.IsWrite() {
```
JmPotato: Do we need to ensure `IsCop()` as well?

client/resource_group/controller/group_controller.go
```diff
@@ -601,23 +672,30 @@ func (gc *groupCostController) onResponseImpl(
...
+	// `count` is the full per-request consumption (BeforeKVRequest + AfterKVRequest).
+	count := &rmpb.Consumption{}
+	*count = *delta
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(count, req)
+	}
+	if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+		gc.metrics.observePagingActual(bytesForEst, resp.ReadBytes(),
+			getRUValueFromConsumption(count), getRUValueFromConsumption(delta))
+	} else if !req.IsWrite() && req.IsCop() {
```
JmPotato: Maybe we can merge this into a new gate check, as the previous comment said.

client/resource_group/controller/group_controller.go
```diff
@@ -632,40 +710,85 @@ func (gc *groupCostController) onResponseWaitImpl(
...
 	if !gc.burstable.Load() {
-		allowDebt := delta.ReadBytes+delta.WriteBytes < bigRequestThreshold || !gc.isThrottled.Load()
-		d, err := gc.acquireTokens(ctx, delta, &waitDuration, allowDebt)
-		if err != nil {
-			if errs.ErrClientResourceGroupThrottled.Equal(err) {
-				gc.metrics.failedRequestCounterWithThrottled.Inc()
-				gc.metrics.failedLimitReserveDuration.Observe(d.Seconds())
+		v := getRUValueFromConsumption(delta)
+		if v > 0 {
+			if bytesForEst > 0 {
```
JmPotato: Can we flatten these nested if statements?

client/resource_group/controller/model.go
```diff
@@ -52,6 +52,26 @@ type RequestInfo interface {
...
+	// paging_* accounting; point gets, batch gets, scans and other
+	// bounded-size reads bypass the paging metrics even though they may
+	// reach this controller through the same RC interceptor.
+	IsCop() bool
+}
+
+// estimatedReadBytes returns the predicted read-bytes hint for read requests.
+// Writes always return 0 so paging pre-charge / settlement / metrics stay
+// gated to reads.
+func estimatedReadBytes(req RequestInfo) uint64 {
```
JmPotato, line 102: `estimatedReadBytes` only gates on `!req.IsWrite()`, but the pre-charge and all `paging_*` metrics are documented as coprocessor RPCs. A non-coprocessor read that returns a non-zero `PredictedReadBytes()` would be pre-charged and counted into `paging_precharge_*` / residual / settlement histograms, diverging from the metric help text. Suggest enforcing the cop-only contract here rather than relying on the caller:

```go
if req.IsWrite() || !req.IsCop() {
	return 0
}
```

client/resource_group/controller/group_controller.go
```diff
@@ -601,23 +672,30 @@ func (gc *groupCostController) onResponseImpl(
...
+	count := &rmpb.Consumption{}
+	*count = *delta
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(count, req)
+	}
+	if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+		gc.metrics.observePagingActual(bytesForEst, resp.ReadBytes(),
+			getRUValueFromConsumption(count), getRUValueFromConsumption(delta))
+	} else if !req.IsWrite() && req.IsCop() {
+		gc.metrics.observePagingNonprecharge(resp.ReadBytes())
```
JmPotato: `paging_nonprecharge_total` is incremented here on the response path, while `paging_precharge_total` is incremented in `onRequestWaitImpl` after a successful token acquire (request path). The two counters then have different event boundaries: a pre-charged RPC that is later canceled / fails transport still counts toward `precharge_total`, but a non-pre-charged RPC that is canceled does not count toward `nonprecharge_total`, so a `precharge / (precharge + nonprecharge)` coverage ratio is systematically skewed.

Suggest counting `nonprecharge_total` at the same boundary — in `onRequestWaitImpl` after the successful acquire when `IsCop() && !IsWrite() && estimated == 0` — and keeping only `paging_nonprecharge_actual_bytes_total` on the response path.

client/resource_group/controller/metrics/metrics.go
```diff
@@ -156,6 +189,97 @@ func initMetrics(constLabels prometheus.Labels) {
...
+			ConstLabels: constLabels,
+		}, []string{newResourceGroupNameLabel})
+
+	PagingPredictionResidualBytes = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: requestSubsystem,
+			Name:      "paging_prediction_residual_bytes",
+			// Signed residual = actual - predicted. Buckets span ±64MB to
+			// absorb large first-page responses when the predictor has no
```
JmPotato: This comment says the buckets absorb large first-page responses when the predictor has no prior observation (`predicted=0`). But `predicted=0` RPCs take the non-pre-charge path and are never observed into this histogram (it is only observed when `bytesForEst > 0`), so the `predicted=0` rationale does not apply here — suggest removing that sentence.

Separately, please confirm the ±64MB range matches the real scale of `PredictedReadBytes()`. If it is a per-paging-batch estimate, ±64MB is fine; if it can be a query/region-level estimate (hundreds of MB), residuals will saturate the end buckets and lose tail resolution — consider ±256MB in that case.

client/resource_group/controller/metrics/metrics.go
```diff
@@ -156,6 +189,97 @@ func initMetrics(constLabels prometheus.Labels) {
...
+			// prior observation (predicted=0) and workload shifts that
+			// leave the prediction above actual. Factor-4 spacing keeps
+			// resolution near zero.
+			Buckets: []float64{-67108864, -16777216, -4194304, -1048576, -262144, -65536, -16384, -4096, 0, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864},
+			Help:        "Histogram of (actual_read_bytes - predicted_read_bytes) for pre-charged coprocessor RPCs. Shows predictor accuracy.",
```
JmPotato: For this histogram and `paging_settlement_ru_delta`: both observe signed values (residual / settlement delta can be negative), so the `_sum` series is not monotonic. Dashboards must not use `rate(..._sum)` or `..._sum / ..._count` (the resulting average is meaningless); only the `_bucket` series are valid (e.g. `histogram_quantile`). Worth stating this in the help text / dashboard description so the signed `_sum` is not misused.

client/resource_group/controller/group_controller_test.go
```diff
@@ -208,6 +217,390 @@ func TestOnResponseWaitConsumption(t *testing.T) {
...
+	// ReadBytesCost*predicted and calculateReadCost adds 0, yielding a negative
+	// delta that flows through RefundTokens. ReadBaseCost is not refunded,
+	// matching existing behavior for non-paging read failures.
+	cfg := DefaultRUConfig()
+	expectedRefund := float64(cfg.ReadBytesCost) * float64(predictedReadBytes)
+	re.InDelta(tokensAfterPreCharge+expectedRefund, tokensAfterSettlement, 1.0,
+		"failed read with paging hint should refund ReadBytesCost*predicted")
+}
+
+func TestDeletePagingLabelsResetsSeries(t *testing.T) {
```
JmPotato, line 782: `TestDeletePagingLabelsResetsSeries` asserts deletion for the counter vecs, but does not observe a sample into `PagingPredictionResidualBytes` / `PagingSettlementRUDelta` and assert those series are removed by `deletePagingLabels`. Suggest adding histogram-vec observe + post-delete assertions so the cleanup path is fully covered — a histogram omitted from `deletePagingLabels` would otherwise leak series silently.

JmPotato (commented):
A few observability notes from reviewing the paging pre-charge metrics (metric semantics, signed-histogram usage, a comment fix, and a test-coverage gap). The metric design overall looks good — the symmetric buckets for the signed histograms and the paired debit/refund counter pattern are the right choices. Comments inline.

client/resource_group/controller/limiter.go
YuhaoZhang00: This belongs to the generic future-reservation reflow issue tracked in #10762. I’ll keep this PR scoped to paging pre-charge/refund and leave in-flight `WaitReservations` rescheduling to #10762.

YuhaoZhang00:
~~One follow-up question to revisit before merging: how should paging pre-charge account for failed requests that still consumed partial TiKV work?~~

~~Current behavior appears to distinguish two cases:~~

~~1. If a response is available, even with `Succeed() == false`, settlement still runs. The precharged predicted-read portion is subtracted, and any `ResponseInfo.ReadBytes()` / `KVCPU()` reported by the failed response is still charged. In effect, a failed response with partial read bytes is settled as `base + actual_read_bytes + kv_cpu`, not fully refunded.~~

~~2. If the request fails before producing any response, `OnRequestCancel` rolls back the request-path precharge completely, because there is no response-side accounting signal for partial read bytes or CPU.~~

~~This seems reasonable if failed responses reliably carry partial `ReadBytes()` / `KVCPU()` when TiKV has already scanned data. If those signals are zero or unavailable for some failed-after-work cases, we may undercount by refunding the predicted-read portion without charging the partial work. Let's revisit this near the end after the main review comments are handled.~~

client/resource_group/controller/metrics/metrics.go
YuhaoZhang00: Checked: PredictedReadBytes is a per-RPC read-response estimate, so ±64MB is intended.

client/resource_group/controller/metrics/metrics.go
YuhaoZhang00: Checked: dashboards did not use `_sum/_count` metrics. Fixed helped text

client/resource_group/controller/group_controller.go
```diff
@@ -654,40 +731,81 @@ func (gc *groupCostController) onResponseWaitImpl(
...
+// that delta from per-group consumption and refunds the corresponding
+// tokens to the limiter so speculatively reserved RU is not lost.
+//
+// The per-store snapshot maintained by onRequestWaitImpl is intentionally
+// not rolled back — it is bookkeeping for penalty distribution and will be
+// overwritten by the next OnRequestWait against the same store.
+func (gc *groupCostController) onRequestCancelImpl(info RequestInfo) {
+	delta := &rmpb.Consumption{}
+	for _, calc := range gc.calculators {
+		calc.BeforeKVRequest(delta, info)
```
rleungx: resp == nil does not prove the RPC never reached TiKV, so refunding the full BeforeKVRequest delta can roll back normal read base cost or write WRU for work that may have already executed.

client/resource_group/controller/group_controller.go
```diff
@@ -106,6 +106,18 @@ type groupMetricsCollection struct {
 	tokenRequestCounter               prometheus.Counter
 	runningKVRequestCounter           prometheus.Gauge
 	consumeTokenHistogram             prometheus.Observer
+
+	// Paging pre-charge observers, cached per-RG to avoid WithLabelValues
+	// on the hot path.
+	prechargeCounter        prometheus.Counter
```
rleungx, line 111: Do we really need these metrics?

client/resource_group/controller/model.go
```diff
@@ -138,6 +174,10 @@ func (kc *KVCalculator) AfterKVRequest(consumption *rmpb.Consumption, req Reques
 	if !req.IsWrite() {
 		// For now, we can only collect the KV CPU cost for a read request.
 		kc.calculateCPUCost(consumption, res)
+		// Paging settlement
+		if bytesForEst := estimatedReadBytes(req); bytesForEst > 0 {
+			consumption.RRU -= float64(kc.ReadBytesCost) * float64(bytesForEst)
```
rleungx, line 200: This can make `gc.mu.consumption.RRU` decrease after a precharge was already reported, but `updateDeltaConsumption` only emits positive deltas. The refund will not be sent to PD, and `lastRequestConsumption.RRU` can stay at the old high-water mark and hide later consumption. Please make RRU/WRU reporting signed while keeping byte/count fields monotonic.

client/resource_group/controller/group_controller.go
YuhaoZhang00: removed `paging_nonprecharge_actual_bytes_total` metric. Kept all 8 remaining metrics cuz they cover three different aspects:

  - Coverage: `paging_precharge_total` and `paging_nonprecharge_total` show how many cop requests are handled with/without a prediction hint.
  - Prediction quality: `paging_precharge_bytes_total`, `paging_actual_bytes_total`, and `paging_prediction_residual_bytes` show predicted volume, actual volume, and per-request prediction error distribution.
  - RU accounting impact: `paging_precharge_ru_total`, `paging_settlement_ru_total`, and `paging_settlement_ru_delta` show how much RU is precharged, how much is finally settled, and whether response-side settlement usually refunds or charges extra.

They are all gated to cop requests and cached per resource group to avoid WithLabelValues on the hot path.

client/resource_group/controller/metrics/metrics.go
```diff
@@ -156,6 +185,87 @@ func initMetrics(constLabels prometheus.Labels) {
...
+	PagingSettlementRU = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace:   namespace,
+			Subsystem:   requestSubsystem,
+			Name:        "paging_settlement_ru_total",
+			Help:        "Sum of total RU finally consumed by pre-charged coprocessor RPCs (read base cost + CPUMsCost * kv_cpu_ms + ReadBytesCost * actual_bytes). Equals precharge_ru + settlement_ru_delta.",
+			ConstLabels: constLabels,
+		}, []string{newResourceGroupNameLabel})
+
+	PagingSettlementRUDelta = prometheus.NewHistogramVec(
```
rleungx: This adds another per-resource-group histogram, so each resource group gets one series per bucket plus `_sum`/`_count` here and in `paging_prediction_residual_bytes`. Could we confirm `paging_settlement_ru_delta` is needed as a long-term metric? If the byte residual histogram is enough for rollout/debugging, dropping this RU delta histogram would reduce the added cardinality.

rleungx (approved):
The rest LGTM except for these metrics

client/resource_group/controller/metrics/metrics.go
YuhaoZhang00: Good point. Currently, local RU consumption is indeed dominated by byte consumption. I think it makes sense to remove the RU-related statistics.

```
prechargeRU = 0.475 + predicted_bytes / 65536
settlementDeltaRU = (actual_bytes - predicted_bytes) / 65536 + cpu_ms / 3
```

YuhaoZhang00: Removed paging RU metrics: 934f8129666c566ddd0ae50874b011c8b84e002f

client/resource_group/controller/model.go
```diff
@@ -52,6 +52,38 @@ type RequestInfo interface {
 	StoreID() uint64
 	RequestSize() uint64
 	AccessLocationType() AccessLocationType
+	// PredictedReadBytes returns the caller-supplied read-bytes hint. The
```
JmPotato, line 60: [blocker] This widens the public `RequestInfo` interface in `github.com/tikv/pd/client`, so external mocks/implementors will fail to compile on upgrade. Please keep these as optional provider interfaces instead: missing hint => 0, missing `IsCop` => no paging accounting.

client/resource_group/controller/model.go
```diff
@@ -242,11 +282,11 @@ func add(custom1 *rmpb.Consumption, custom2 *rmpb.Consumption) {
 
 func updateDeltaConsumption(last *rmpb.Consumption, now *rmpb.Consumption) *rmpb.Consumption {
 	delta := &rmpb.Consumption{}
-	if now.RRU > last.RRU {
+	if now.RRU != last.RRU {
```
JmPotato: [warning] Allowing RRU/WRU to go backwards also feeds `calcAvg` through `getRUValueFromConsumption(gc.run.consumption)`. A negative settlement can make `avgRUPerSec` negative and suppress the next token request; please clamp only the rate-estimation delta to zero while still reporting signed RU.

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -131,10 +156,24 @@ func (c *ruCollector) Aggregate() []map[string]any {
...
 	if len(keyspaceRUMetering) == 0 {
+		c.Unlock()
 		return nil
 	}
+	for keyspaceName, ruMetering := range keyspaceRUMetering {
+		carry := c.keyspaceRUCarry[keyspaceName]
+		if carry == nil {
+			carry = &RUCarry{}
+		}
+		ruMetering.oltpRU = applyRUCarry(ruMetering.oltpRU, &carry.oltpRU)
```
JmPotato: [blocker] This carry is in-memory only. If the process/keyspace is removed or there is no later positive RU to offset it, a previously flushed pre-charge is never corrected in metering. The refund path needs durable/final settlement semantics, or metering should record actual usage instead of reservation.

JmPotato (commented):
Inline review comments from implementation review.

pkg/mcs/resourcemanager/server/metering.go
YuhaoZhang00: Thanks for the feedback. I agree that the previous iterations around paging RU refund reporting became awkward, especially after trying to make the same reported value satisfy both settlement and metering semantics.

I changed the design to split the RU information reported from the client to PD into two separate meanings.

Previously, PD used the same `consumption_since_last_request` field for multiple purposes:

- **Metering / billing**: this needs the final actual resource usage and should not consume negative refund values.
- **Resource group stats / active RU / RUv2**: these should also be based on actual usage.
- **RRU / WRU observability metrics**: these need to represent token settlement debit/refund, so the reported RRU/WRU values need signed semantics.

After this change:

- `consumption_since_last_request` is kept, but its meaning is narrowed to actual resource usage. PD uses it for metering, resource group stats, active RU, RUv2, bytes, CPU, and related actual-usage accounting.
- A new `settlement_since_last_request` field is added for signed token settlement. Its RRU/WRU values may be negative and are used by PD-side RRU/WRU debit/refund metrics. It is not used for metering.

The kvproto change is in https://github.com/pingcap/kvproto/pull/1494, which adds:

```proto
Consumption settlement_since_last_request = 8;
```

pkg/mcs/resourcemanager/server/manager.go
```diff
@@ -87,6 +88,14 @@ func (cfg *pushMetricsConfig) syncPushMetricsTicker(
 	return ticker
 }
 
+func resourceGroupPushCollectors() []prometheus.Collector {
```
JmPotato: With the refund counters gone in the latest revision, this helper returns exactly the same three collectors the pusher used to chain directly, and `TestResourceGroupPushCollectors` now just asserts the function body against itself. Suggest reverting both back to the original chained `.Collector(...)` calls to keep the diff minimal.

pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -223,8 +223,9 @@ var (
 )
 
 type metrics struct {
-	// record update time of each resource group
-	consumptionRecordMap map[consumptionRecordKey]time.Time
+	// Records the latest activity time for metric series cleanup. This is
+	// independent from metering/actual consumption.
+	metricsActivityRecordMap map[metricsActivityRecordKey]time.Time
```
JmPotato: The `consumptionRecordMap` → `metricsActivityRecordMap` rename (~60 lines across manager/metrics/tests) was motivated by the earlier design where settlement-only reports also refreshed activity. With the settlement stream removed, every record here is once again driven by actual consumption, so the original name is accurate again. Suggest reverting the rename to shrink the diff.

pkg/mcs/resourcemanager/server/metrics.go
```diff
@@ -350,13 +350,16 @@ func (m *metrics) recordConsumption(
 		ruLabelType = tiflashTypeLabel
 	}
 	consumption := consumptionInfo.Consumption
+	if isZeroConsumption(consumption) {
```
JmPotato: This early return changes cleanup behavior for idle groups: previously an all-zero consumption report still refreshed the activity record, keeping the group's metric series alive indefinitely; now an idle-but-connected group's series get deleted after `metricsCleanupTimeout` and recreated on the next activity (a counter reset, which PromQL `rate()` tolerates). Probably fine — but is this intended? If kept, please call it out in the PR description. Same question for the guard in `counterMetrics.add` (L425).

pkg/mcs/resourcemanager/server/metering.go
```diff
@@ -47,6 +47,9 @@ type ruMetering struct {
 }
 
 func (rm *ruMetering) add(consumption *consumptionItem) {
+	if consumption == nil || consumption.Consumption == nil {
```
JmPotato: This guard is unreachable: the only call path is `backgroundMetricsFlush` → `ruCollector.Collect` → `rm.add`, and the flush loop already skips items with `consumptionInfo.Consumption == nil` before calling `Collect`. Suggest dropping it — `TestRUCollectorUsesConsumption` still works as a plain behavior test.

client/resource_group/controller/model.go
```diff
@@ -185,6 +246,85 @@ func (kc *KVCalculator) payBackWriteCost(consumption *rmpb.Consumption, req Requ
...
+		}
+		kvCalc.payBackWriteCost(refund, req)
+	}
+	return refund
+}
+
+func reportedRequestConsumption(calculators []ResourceCalculator, req RequestInfo, tokenDelta *rmpb.Consumption) *rmpb.Consumption {
+	reported := cloneConsumption(tokenDelta)
+	adjustPagingPrechargeRRU(calculators, reported, req, -1)
+	sub(reported, writeReservationConsumption(calculators, req))
```
JmPotato: Stripping the write reservation from the reported request consumption (and re-adding it at response time) is severable from paging pre-charge — the feature itself only needs the `adjustPagingPrechargeRRU` strip/restore pair. This line changes when WRU is reported for **all** writes (success: request time → response time), and it also fixes a real pre-existing quirk: a failed write straddling a report boundary used to over-count WRU, because the monotonic `>` guards in `updateDeltaConsumption` never subtract the payback. That fix is legitimate, but bundling it here widens the blast radius of an already-XXL PR. Could it move to a separate PR, or at least be called out explicitly in the description as an independent behavior change? If minimizing this PR is preferred, keeping the write path as-is would also let `writeReservationConsumption`/`writeRefundConsumption` (~40 lines) drop out.

client/resource_group/controller/model.go
```diff
@@ -294,9 +434,16 @@ func updateDeltaConsumption(last *rmpb.Consumption, now *rmpb.Consumption) *rmpb
 		delta.TiflashRUV2 = now.TiflashRUV2 - last.TiflashRUV2
 		last.TiflashRUV2 = now.TiflashRUV2
 	}
+	if isZeroConsumption(delta) {
```
JmPotato: Returning nil for an all-zero delta means new clients stop sending empty periodic reports. Old servers handle a nil `ConsumptionSinceLastRequest` fine (the flush loop already skips it), but combined with the server-side zero guard this is what changes the idle-group metric cleanup semantics — see my comment in `metrics.go`. Intended? Worth a line in the PR description either way.

client/resource_group/controller/group_controller.go
```diff
@@ -315,7 +366,11 @@ func (gc *groupCostController) calcAvg(counter *tokenCounter, new float64) bool
 	failpoint.Inject("acceleratedReportingPeriod", func() {
 		deltaDuration = 100 * time.Millisecond
 	})
-	delta := (new - counter.avgRUPerSecLastRU) / deltaDuration.Seconds()
+	ruDelta := new - counter.avgRUPerSecLastRU
+	if ruDelta < 0 {
```
JmPotato: This clamp also affects non-paging flows, and in the current revision I'm not sure any steady-state path still produces a negative delta: pre-charge refunds no longer enter `gc.mu.consumption` (the reported deltas are all non-negative), and the failed-write payback no longer decreases it either. The remaining source seems to be the transient add-then-sub window in `onRequestWaitImpl`'s failure path, which pre-dates this PR. Is the clamp load-bearing for pre-charge, or defensive? A short code comment naming the path that produces negative deltas would help future readers.

JmPotato (commented):
Did another pass after the `SettlementSinceLastRequest` removal — the wire-protocol concern is fully resolved, and I verified the reported consumption for pre-charged reads now matches master exactly (request = base cost, response = actual + CPU; the pre-charge never reaches billing/stats). Remaining items, in priority order:

**Blocker — go.sum is not tidy.** The `go.mod` files of the root, `tests/integrations`, and `tools` modules are untouched, yet each of their `go.sum` files still carries ~800 added lines of stale checksums left over from the fork-replace era. A module whose `go.mod` didn't change should have zero `go.sum` diff, so `make tidy` will prune all of them and the CI clean-diff check is likely to fail as-is. Please restore the three `go.sum` files and re-run `go mod tidy` in each module. (Promoting `github.com/prometheus/client_model` to a direct dependency in `client/go.mod` is fine — the new tests use `dto.Metric` directly.)

**PR description drift.** The description still says `PredictedReadBytes` is "a required method on the interface" (it is an optional type assertion now, together with the `IsCop()` gate which isn't mentioned at all), still scopes the change to `client/resource_group/controller` (server-side files are touched), and the test list references `TestPagingPreChargeZeroDelta`, which no longer exists under that name. Please refresh it to match the current implementation.

The inline comments below cover settlement-era leftovers that can be reverted now, plus a few behavior changes that deserve either a split or an explicit call-out.

client/resource_group/controller/group_controller.go
```diff
@@ -622,23 +677,28 @@ func (gc *groupCostController) onResponseImpl(
...
+		calc.BeforeKVRequest(count, req)
+	}
+	if bytesForEst, ok := pagingReadEstimate(req); ok {
+		gc.metrics.observePagingResponse(bytesForEst, resp.ReadBytes())
+	}
 	if !gc.burstable.Load() {
 		counter := gc.run.requestUnitTokens
 		if v := getRUValueFromConsumption(delta); v > 0 {
 			counter.limiter.RemoveTokens(time.Now(), v)
+		} else if v < 0 {
```
JmPotato: This branch catches more than the paging over-estimate the comment describes. A failed write's settlement delta is `readBytesCost − writeCost`, which is usually negative, so failed writes now refund their reserved write tokens to the local limiter through this path. On master, `acquireTokens` only acts on `v > 0` and this payback was silently dropped, so this is a real behavior change for non-paging traffic: the net local debit for a failed write becomes just the read-bytes cost instead of the full write reservation. I believe the new behavior is the more correct one, but please: (1) update this comment (and the twin branch in `onResponseWaitImpl`) to say it covers both paging over-estimates and failed-write paybacks; (2) add a limiter-level assertion for the failed-write refund — `TestFailedWriteReservationDoesNotEnterReportedConsumption` only checks the reporting side; (3) document this together with the write-reservation reporting shift (WRU moving from request time to response time) in the PR description, which currently only describes the read path.

client/resource_group/controller/model.go
```diff
@@ -185,6 +246,85 @@ func (kc *KVCalculator) payBackWriteCost(consumption *rmpb.Consumption, req Requ
...
+	for _, calc := range calculators {
+		kvCalc, ok := calc.(*KVCalculator)
+		if !ok {
+			continue
+		}
+		consumption.RRU += sign * kvCalc.pagingPrechargeRRU(req)
+	}
+}
+
+func writeReservationConsumption(calculators []ResourceCalculator, req RequestInfo) *rmpb.Consumption {
```
JmPotato: Hot-path allocations: `reportedRequestConsumption`/`reportedResponseConsumption` run for every KV request, and this helper allocates the empty `&rmpb.Consumption{}` before the `IsWrite()` check, so pure-read workloads pay for it on both the request and response sides (same in `writeRefundConsumption`), on top of the unconditional `cloneConsumption`. `add`/`sub` are already nil-safe, so returning nil for non-writes removes most of it — roughly 4 extra allocations per read request back to master's level. Optionally, when the request is neither a write nor carries a paging hint, the clone can be skipped entirely and `tokenDelta` returned as-is (neither value is mutated afterwards), but the nil-for-non-writes fix alone is the important part.

JmPotato (commented):
The rest LGTM — once the two inline comments below are addressed, this is good to merge from my side.

client/resource_group/controller/model.go
YuhaoZhang00: Agree. Extracted the bug and its fix to a separate issue / PR.
  - Issue: https://github.com/tikv/pd/issues/10985
  - PR: https://github.com/tikv/pd/pull/10986

JmPotato (approved):

PR #10615, opened 2026-04-22
gc_state_manager: Workaround mutex contention in high-concurrency scenario

MyonKeminta:

Issue Number: Ref #10607

```commit-message
This PR provides a simpler workaround to the issue that when PD receives too many GC-related requests, it may suffer bad performace on the mutex of GCStateManager's mutex.

In the follwing cases, the GCState will be fully loaded and inserted to a memory cache:
* When GetGCState/GetAllKeyspacesGCState encounters a cache miss and reads the actual GCState from etcd
* When AdvanceTxnSafePoint is called

In the following cases, the GCState cache will be incrementally updated if it already exists in the cache:
* When AdvanceGCSafePoint/SetGCBarrier/DeleteGCBarrier is called

Also note that UpdateServiceGCSafePoint is internally redirected to one of AdvanceTxnSafePoint SetGCBarrier or DeleteGCBarrier.
```

### Check List

Tests

- Unit test (Covered by existing tests)
- Integration test (Covered by existing tests)

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -120,6 +120,7 @@ type GCStateManager struct {
 	cfg             config.PDServerConfig
 	keyspaceManager *keyspace.Manager
 
+	// Maps from keyspaceID to the cached GCState of that keyspace.
 	gcStatesCache map[uint32]GCState
 
 	allKeyspacesGCStatesSingleFlight *syncutil.OrderedSingleFlight[map[uint32]GCState]
@@ -794,7 +795,7 @@ func (m *GCStateManager) getGCStateImpl(keyspaceID uint32) (GCState, error) {
 	m.mu.RLock()
 	if cachedGCState, ok := m.tryGetGCStateFromCacheLocked(keyspaceID); ok {
 		m.mu.RUnlock()
-		gcStateCacheAccessCounter.WithLabelValues("hit").Inc()
+		gcStateCacheAccessHitCounter.Inc()
 		return cachedGCState, nil
 	}
 	m.mu.RUnlock()
@@ -805,11 +806,11 @@ func (m *GCStateManager) getGCStateImpl(keyspaceID uint32) (GCState, error) {
 
 	// Check cache again: it may be updated by other concurrent invocations when the lock is not held.
 	if cachedGCState, ok := m.tryGetGCStateFromCacheLocked(keyspaceID); ok {
-		gcStateCacheAccessCounter.WithLabelValues("slow_hit").Inc()
+		gcStateCacheAccessSlowHitCounter.Inc()
 		return cachedGCState, nil
 	}
 
-	gcStateCacheAccessCounter.WithLabelValues("cache_miss").Inc()
+	gcStateCacheAccessMissCounter.Inc()
 
 	var result GCState
 	err := m.gcMetaStorage.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -790,6 +790,14 @@ func (m *GCStateManager) tryGetGCStateFromCacheLocked(keyspaceID uint32) (GCStat
 	return res, ok
 }
 
+// ForceInvalidateGCStateCache invalidates all the GCState cache.
+func (m *GCStateManager) ForceInvalidateGCStateCache() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.gcStatesCache = make(map[uint32]GCState)
+}
+
 func (m *GCStateManager) getGCStateImpl(keyspaceID uint32) (GCState, error) {
 	// Fast path
 	m.mu.RLock()
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -633,7 +633,7 @@ func (m *GCStateManager) setGCBarrierImpl(ctx context.Context, keyspaceID uint32
 	}
 
 	{
-		// For simplicity, only update if the cache already exists. In this
+		// For simplicity, only update if the cache already exists.
 		cachedGCState, ok := m.gcStatesCache[keyspaceID]
 		if ok {
 			replaced := false
@@ -770,10 +770,7 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
 	return result, nil
 }
 
-func (m *GCStateManager) tryGetGCStateFromCache(keyspaceID uint32) (GCState, bool) {
-	m.mu.RLock()
-	defer m.mu.RUnlock()
-
+func (m *GCStateManager) tryGetGCStateFromCacheLocked(keyspaceID uint32) (GCState, bool) {
 	if !m.nodeIsLeader() {
 		return GCState{}, false
 	}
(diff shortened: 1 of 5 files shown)


pkg/gc/gc_state_manager.go
```diff
@@ -667,28 +770,72 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
...
+
 	var result GCState
-	err = m.gcMetaStorage.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
+	err := m.gcMetaStorage.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
 		var err1 error
 		result, err1 = m.getGCStateInTransaction(keyspaceID, wb)
 		return err1
 	})
 
+	m.gcStatesCache[keyspaceID] = result
```
rleungx, line 833: If an error happens, do we still need to update gcStatesCache?

pkg/gc/gc_state_manager.go
```diff
@@ -667,28 +770,72 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
...
+		gcStateCacheAccessCounter.WithLabelValues("hit").Inc()
+		return cachedGCState, nil
 	}
+
+	// Slow path
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	// Check cache again: it may be updated by other concurrent invocations when the lock is not held.
+	if cachedGCState, ok := m.gcStatesCache[keyspaceID]; ok {
```
rleungx: Do we also need the check leader?

pkg/gc/gc_state_manager.go
```diff
@@ -667,28 +770,72 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
...
+	defer m.mu.RUnlock()
+
+	if !m.nodeIsLeader() {
+		return GCState{}, false
+	}
+
+	res, ok := m.gcStatesCache[keyspaceID]
+
+	// Deep copy the GCBarriers field to avoid unintentional mutation.
+	if ok && res.GCBarriers != nil {
```
rleungx, line 782: Will the cache be stale if we call deleteGCBarrier?

pkg/gc/gc_state_manager.go
rleungx: no need to clone GCBarriers?

pkg/gc/gc_state_manager.go
```diff
@@ -561,6 +632,28 @@ func (m *GCStateManager) setGCBarrierImpl(ctx context.Context, keyspaceID uint32
 		return nil, err
 	}
 
+	{
+		// For simplicity, only update if the cache already exists. In this
```
wfxr: incomplete commment?

pkg/gc/gc_state_manager.go
MyonKeminta: it should be updated in `deleteGCBarrierImpl`?

pkg/gc/metrics.go
```diff
@@ -24,8 +24,16 @@ var (
...
+			Namespace: "pd",
+			Subsystem: "gc",
+			Name:      "gc_state_cache_access_total",
+			Help:      "Counter of GC state cache accesses by result.",
+		}, []string{"result"})
 )
 
 func init() {
 	prometheus.MustRegister(gcSafePointGauge)
+	prometheus.MustRegister(gcStateCacheAccessCounter)
```
bufferflies, line 42: you can pre define sub metrics such as slow_hit,cache_miss

pkg/gc/gc_state_manager.go
MyonKeminta: Fixed

pkg/gc/gc_state_manager.go
MyonKeminta: Fixed

pkg/gc/gc_state_manager.go
```diff
@@ -118,7 +120,14 @@ type GCStateManager struct {
 	cfg             config.PDServerConfig
 	keyspaceManager *keyspace.Manager
 
+	gcStatesCache map[uint32]GCState
```
bufferflies, line 124: How about adding a new comment to describe the map key value pair?

pkg/gc/gc_state_manager.go
MyonKeminta: Sorry it's copilot murmuring nonsense

pkg/gc/gc_state_manager.go
rleungx: deleteGCBarrierImpl won't update the cache here

pkg/gc/gc_state_manager.go
```diff
@@ -149,6 +159,33 @@ func getKeyspaceNameFromCtx(ctx context.Context) string {
 	return "<unknown>"
 }
 
+// OnNodeBecomesLeader marks the current PD node as leader for GC state watches.
+func (m *GCStateManager) OnNodeBecomesLeader() {
```
bufferflies, line 164: Where is it used?

pkg/gc/gc_state_manager.go
```diff
@@ -667,28 +770,75 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, _ *endpoint.
...
-// GetGCState returns the GC state of the given keyspace.
-//
-// When this method is called on a keyspace without keyspace-level GC enabled, it will be equivalent to calling it on
-// the NullKeyspace.
-func (m *GCStateManager) GetGCState(keyspaceID uint32) (GCState, error) {
-	keyspaceID, _, err := m.redirectKeyspace(keyspaceID, true)
-	if err != nil {
-		return GCState{}, err
+func (m *GCStateManager) tryGetGCStateFromCacheLocked(keyspaceID uint32) (GCState, bool) {
+	if !m.nodeIsLeader() {
```
wfxr, line 775: Does this check necessary since already checked at the rpc entry?

https://github.com/MyonKeminta/pd/blob/1e32df85908b5599d0ff9fafe318b91121c3a5f0/server/gc_service.go#L702-L705

pkg/gc/gc_state_manager.go
```diff
@@ -264,6 +301,16 @@ func (m *GCStateManager) advanceGCSafePointImpl(ctx context.Context, keyspaceID
 		return 0, 0, err
 	}
 
+	if newGCSafePoint != oldGCSafePoint {
```
bufferflies, line 305: How about moving this to line 314?

pkg/gc/gc_state_manager.go
MyonKeminta: server/cluster/cluster.go

pkg/gc/gc_state_manager.go
MyonKeminta: Not that necessary, but for now I think it doesn't harm

pkg/gc/gc_state_manager.go
MyonKeminta: I actually prefer the current style that distinguishes pure logging and important logic,

rleungx (approved):

wfxr (approved):
LGTM

bufferflies (approved):

PR #10617, opened 2026-04-23
gc,server: add GC state cache test coverage

wfxr:

Issue Number: Ref #10607

```commit-message
gc,server: add GC state cache test coverage

Add focused unit and integration coverage for the GC state cache behavior
introduced by GetGCState caching.

The added tests cover cache hit/miss behavior, cache update and invalidation
paths, and leader transition semantics for both cached and uncached requests.
```

This PR supplements the GC state cache change with focused test coverage.

Checklist of covered scenarios:

- [x] The target keyspace has no cache beforehand
  - [x] Concurrent `GetGCState` on the same uncached keyspace
  - [x] Request old leader after leader switch
  - [x] No-cache slow path loses leadership and does not recover
  - [x] No-cache slow path loses and then regains leadership
  - [x] Storage read failure does not populate cache
  - [x] Delete barrier before cache exists
- [x] The target keyspace already has cache
  - [x] Concurrent `GetGCState` on the same cached keyspace
  - [x] Request old leader after leader switch
  - [x] Cache-hit request loses leadership before response
  - [x] Cache-hit request loses and then regains leadership
  - [x] Delete barrier after cache exists
  - [x] Advance GC safepoint after cache exists
  - [x] Add or update barrier after cache exists
  - [x] Delete legacy service GC safepoint

### Check List

Tests

- [x] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Code changes

- [ ] Has the configuration change
- [ ] Has HTTP API interfaces changed (Don't forget to [add the declarative for the new API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- [ ] Has persistent data change

Side effects

- [ ] Possible performance regression
- [x] Increased code complexity
- [ ] Breaking backward compatibility

Related changes

- [ ] PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn):
- [ ] PR to update [`pingcap/tiup`](https://github.com/pingcap/tiup):
- [ ] Need to cherry-pick to the release branch

### Release note

```release-note
None.
```

### DIFF
--- a/tests/server/gc/gc_test.go
+++ b/tests/server/gc/gc_test.go
@@ -18,12 +18,14 @@ import (
 	"context"
 	"math"
 	"slices"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.uber.org/goleak"
 
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 
 	"github.com/tikv/pd/pkg/keyspace"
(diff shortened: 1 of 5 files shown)


MyonKeminta (approved):

tests/server/gc/gc_test.go
```diff
@@ -386,3 +474,270 @@ func TestGCOperations(t *testing.T) {
...
+	re.NoError(err)
+
+	re.NoError(cluster.GetServer(newLeader).ResignLeaderWithRetry())
+	re.Equal(oldLeader, cluster.WaitLeader())
+
+	point.releaseAndDisable(re)
+	res := <-resultCh
+	re.NoError(res.err)
+	re.Nil(res.resp.GetHeader().GetError())
+	re.Equal(uint64(10), res.resp.GetGcState().GetTxnSafePoint())
```
rleungx, line 617: Does it meet the expectation?

wfxr: This is mainly meant to pin the current cache-hit behavior rather than define the ideal long-term contract.

In this scenario, the request has already copied the cached GC state before the leadership churn. The current handler only rechecks whether this PD is leader again before replying, so if the same PD regains leadership before the final response check, it can return that earlier cached value.

I updated the test comment to make this explicit and to avoid implying that this stale snapshot is a permanent/ideal API contract. If we decide to tighten `GetGCState` semantics in the future, the implementation, docs, and this test should be updated together.

rleungx (approved):

okJiang (approved):

PR #10618, opened 2026-04-24
pkg/election: use native lease keepalive stream

JmPotato:

Issue Number: ref #10653.

```commit-message
Use etcd native lease KeepAlive streams instead of spawning a KeepAliveOnce goroutine on each keepalive interval.
Lease.KeepAlive now runs in a single goroutine whose select multiplexes ctx cancellation, the etcd keepalive channel, and the leaseTimeout watchdog. There is no longer a separate worker goroutine or intermediate 1-slot output channel; the loop consumes etcd's keepalive channel directly and applies the existing maxExpire monotonicity and watchdog double-check logic inline.
Exit promptly when the keepalive channel closes or the response carries a non-positive TTL, and double-check the local expire time before treating the watchdog timer as a lease keepalive timeout.
Recreate the underlying etcd lessor when a Lease is granted again after Close, preserving the existing Lease reuse behavior because lessor.Close permanently halts native KeepAlive on the same instance.
Bake purpose and lease-id into a scoped logger so every keepalive log line carries them; all pre-existing log strings and field names are preserved for operator alerts.
After rebasing over the upstream lease metrics change, keep the stream-compatible lease metrics for keepalive response interval, local TTL remaining, and renewal termination reason. Remove the KeepAliveOnce request duration metric and its Grafana panel because the native stream model no longer has one request duration per renewal.
```

### Check List

Tests

- Unit test

### Release note

```release-note
Improve PD lease keepalive reliability by using etcd native keepalive streams.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lease renewal reliability and responsiveness via a streaming keep-alive worker
  * Enhanced lease state tracking and closure management to allow safe reuse after close
  * Strengthened keep-alive loop robustness to avoid spurious timeouts and handle invalid TTLs and closed streams
  * Improved timer reset and coalescing behavior during rapid renewal cycles

* **Tests**
  * Added tests to validate keep-alive behavior, expiry updates, coalescing, and shutdown scenarios


pkg/election/lease.go
```diff
@@ -94,151 +102,214 @@ func (l *Lease) Close() error {
...
+				zap.Duration("expected-interval", expectedInterval),
+				zap.Int64("ttl", ttl),
+				zap.Time("last-response-time", lastResponseTime))
 		}
-	}()
+	}
+	// KeepAlive responses do not carry a per-request start time in this
+	// streaming model. Using arrival time makes the local estimate reflect
+	// when PD actually observed the renewal, not when etcd generated it.
+	return now.Add(time.Duration(ttl) * time.Second), true
```
lhy1024, line 293: This can revive an already expired lease. `now` is the time PD consumes the keepalive channel, not when the etcd client actually received the response. Since native `KeepAlive` returns a buffered channel, if the local lease has already expired according to `expireTime` but there is still a buffered/delayed response in the channel, this line recomputes a future expireTime from the current consumer time and makes the expired lease valid again. This changes `lease.IsExpired()` to false.

Please check the current local deadline before applying a response; if the lease has already expired locally, the loop should exit or reject the response instead of extending expireTime.

lhy1024: I can reproduce it with a unit test:
```
func TestReviewReproStaleKeepAliveResponseDoesNotReviveExpiredLease(t *testing.T) {
        re := require.New(t)
        keepAliveCh := make(chan *clientv3.LeaseKeepAliveResponse, 1)
        lease := newTestLeaseWithKeepAliveCh(keepAliveCh, time.Hour)
        lease.expireTime.Store(time.Now().Add(-time.Second))
        keepAliveCh <- &clientv3.LeaseKeepAliveResponse{ID: clientv3.LeaseID(1), TTL: 30}

        ctx, cancel := context.WithCancel(t.Context())
        defer cancel()
        go lease.KeepAlive(ctx)

        time.Sleep(100 * time.Millisecond)
        re.True(lease.IsExpired())
  }
```

pkg/election/lease.go
```diff
@@ -94,151 +102,214 @@ func (l *Lease) Close() error {
...
-	timer := time.NewTimer(l.leaseTimeout)
-	defer timer.Stop()
 	for {
 		select {
-		case t := <-timeCh:
+		case <-ctx.Done():
+			l.metrics.contextCanceled.Inc()
+			return
+		case resp, ok := <-keepAliveCh:
+			if !ok {
```
lhy1024, line 196: On a non-cancel keepalive channel close, this branch only logs and returns, but it does not update `expireTime`. That means `KeepAlive` has stopped, while callers can still observe the lease as valid through `Leadership.Check()` / `Participant.IsServing()` until the old local TTL expires.

I can reproduce it with a unit test: set `expireTime` to the future, close the keepalive channel, and `KeepAlive` returns while `lease.IsExpired()` is still false.

Please set `expireTime` to `time.Now()` or zero on non-cancel channel close, and record an explicit termination reason/metric. Otherwise upper layers may continue serving after the keepalive stream is already unavailable.

lhy1024:
```
func TestReviewReproChannelCloseMarksLeaseExpired(t *testing.T) {
        re := require.New(t)
        keepAliveCh := make(chan *clientv3.LeaseKeepAliveResponse)
        lease := newTestLeaseWithKeepAliveCh(keepAliveCh, time.Hour)
        lease.expireTime.Store(time.Now().Add(time.Hour))
        close(keepAliveCh)

        done := make(chan struct{})
        go func() {
                lease.KeepAlive(t.Context())
                close(done)
        }()

        select {
        case <-done:
        case <-time.After(time.Second):
                t.Fatal("timed out waiting for keepalive to stop")
        }
        re.True(lease.IsExpired())
  }
```

JmPotato:
## Analysis Summary: Not recommended for merge

A full end-to-end code comparison of the `KeepAliveOnce` → native `KeepAlive` stream migration was performed (etcd client v3.5.15 internals + PD lease.go old/new versions). Key findings:

### 1. Shared stream creates a single point of failure for all leases (most severe)

This PR uses `l.lease.KeepAlive(ctx, leaseID)`, which means all leases on the same client share **one bidirectional gRPC stream**.

- **old**: Each `KeepAliveOnce` opens an independent short-lived stream → one lease's failure does not affect others.
- **new**: One stream breaks → **all lease keepalive channels on that client close simultaneously**.

In PD's microservice architecture, a single PD server may hold leases for multiple keyspace groups. A single etcd connection hiccup triggers a cascading step-down + re-campaign for all of them.

### 2. Loss of renewal frequency upper bound

- **old** (`keepAliveWorker`): interval = `leaseTimeout/3`, capped at `maxLeaseKeepAliveInterval = 5s/3 ≈ 1.67s`.
- **new** (`sendKeepAliveLoop`): interval = `now + TTL/3`, **no upper bound**. For a 60s TTL election lease, the renewal gap reaches 20s.

The old capped heartbeat (≤1.67s) ensured fast detection of etcd anomalies. The new 20s sparse heartbeat significantly degrades fault detection speed. While the `timeoutTimer` with `leaseTimeout` provides a fallback, it waits for expiration rather than detecting anomalies early.

### 3. etcd client retry window is too narrow

- **old** (`KeepAliveOnce`): etcd client's `KeepAliveOnce` internally has a `for {}` retry loop that keeps retrying throughout the entire `leaseTimeout` window.
- **new** (native stream): etcd client's `recvKeepAliveLoop` waits only `retryConnWait = 500ms` before attempting `resetRecv()`. If reconnection fails, channel close immediately causes PD's KeepAlive to exit.

etcd leader election transitions typically take 1-3s. The 500ms retry window may be insufficient, causing PD to spuriously step down during normal etcd leader changes.

### 4. Response queue overflow causes false-positive step-down

The etcd client's internal channel has capacity `LeaseResponseChSize = 16`. When PD's KeepAlive goroutine is slow to consume (CPU starvation, GC), the 17th response is dropped. Dropped responses do not update PD's local `expireTime`, so even though the etcd-side lease renewal succeeded, PD's `timeoutTimer` fires and triggers step-down. The old unbuffered channel provided natural backpressure and did not have this problem.

### Recommendation

None of the above risks exist in the current `KeepAliveOnce + ticker` architecture. Given the high stability requirements of the election path, the stream-based single point of failure, loss of renewal frequency upper bound, and false-positive step-down risks introduced by this approach should not be merged without thorough validation.

Suggested follow-up directions:
- Maintain per-lease independence (avoid shared streams).
- Preserve the renewal frequency upper bound.
- At minimum, run large-scale fault injection tests (etcd leader kill, network partition, high load) to validate these edge cases before and after migration.

JmPotato:
Closing per the analysis above — the native keepalive stream approach introduces shared-stream single point of failure, loss of renewal frequency upper bound, too-narrow retry window, and false-positive step-down risks that the current KeepAliveOnce + ticker architecture does not have.

PR #10620, opened 2026-04-24
server: fix silent store heartbeat forwarding failure

### LINKED ISSUE #10619: mcs: silent store heartbeat forwarding failure causes false down-store alerts
### What did you do?

Deployed the scheduling service independently. During a PD leader switch, the `servicePrimaryMap` in PD temporarily loses the primary address of the scheduling service until the etcd watcher catches up. During this window, `GetServicePrimaryAddr` returns an empty string, causing the scheduling gRPC client to evaluate to `nil` in `updateSchedulingClient`.

When `StoreHeartbeat` requests arrive at PD, they are handled locally and then forwarded. However, when the scheduling client is `nil`, the forwarded heartbeat is silently dropped at `server/grpc_service.go:954-972`.

Forwarding failures should be observable via metrics, just like `RegionHeartbeat` forwarding failures, so that the silent dropping of heartbeats doesn't go unnoticed.

### What did you see instead?

The failure is entirely silent. No metrics are incremented, and no errors are logged or returned. Because the scheduling service starves for `StoreHeartbeat`s, the stores' `LastHeartbeatTS` age out locally on the scheduling server, eventually causing the stores to transition from `disconnected` -> `unhealthy` -> `down`. This triggers a false `SchedulingServerDiscoverDownStore` alert, even though the stores are perfectly healthy and successfully heartbeating to PD.

v8.5.4-nextgen.202510.7

liyishuai:

Issue Number: Close #10619

When PD is deployed in microservice mode, `StoreHeartbeat` requests arriving at the monolithic PD server are handled locally and then forwarded to the scheduling server. 

However, during a PD leader switch, the `servicePrimaryMap` in PD temporarily loses the primary address of the scheduling service until the etcd watcher catches up. During this window, `GetServicePrimaryAddr` returns an empty string, causing the scheduling gRPC client to evaluate to `nil` in `updateSchedulingClient`. 

When the scheduling client was `nil`, the forwarded heartbeat was silently dropped. Because the scheduling service was starved for `StoreHeartbeat`s, the stores' `LastHeartbeatTS` aged out locally on the scheduling server, eventually causing the stores to transition from `disconnected` -> `unhealthy` -> `down`. This triggered a false `SchedulingServerDiscoverDownStore` alert, even though the stores were perfectly healthy and successfully heartbeating to PD.

This PR fixes the silent failure by adding an explicit check for a `nil` client when forwarding store heartbeats. If the client is `nil`, it increments a new Prometheus metric `pd_server_forward_fail_total` with labels `("store_heartbeat", "client")`. This makes the forwarding failure observable and aligns the error handling with the existing `RegionHeartbeat` forwarding logic.

```commit-message
server: fix silent store heartbeat forwarding failure

When PD is in microservice mode but the scheduling primary's address
is not yet available (e.g. during a leader switch where the primary
watcher hasn't caught up), the scheduling client evaluates to nil.

Previously, this resulted in the StoreHeartbeat forward being silently
dropped, which starved the scheduling server of heartbeat information
and caused false 'SchedulingServerDiscoverDownStore' alerts as the
stores aged out locally on the scheduling server.

This commit adds an explicit check and increments a new Prometheus
metric 'pd_server_forward_fail_total' with labels
('store_heartbeat', 'client') when the client is nil, making the
forwarding failure observable and consistent with the region heartbeat
forwarding error handling.
```

### Check List

Tests

- Integration test

Code changes

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
Fix an issue where store heartbeats were silently dropped when forwarded from PD to the scheduling service during a leader switch, which could cause false down-store alerts.
```

## Summary by CodeRabbit

* **New Features**
  * Added a metric counter that distinguishes heartbeat-forwarding failures when the scheduling service client is unavailable.

* **Tests**
  * Added an integration test that verifies the new metric increases when the scheduling primary is unavailable, making heartbeat-forwarding failures observable.

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -71,6 +71,7 @@ var (
 	errScatterRegionSend     = forwardFailCounter.WithLabelValues("scatter_region", "send")
 	errSplitRegionsSend      = forwardFailCounter.WithLabelValues("split_regions", "send")
 	errStoreHeartbeatSend    = forwardFailCounter.WithLabelValues("store_heartbeat", "send")
+	errStoreHeartbeatClient  = forwardFailCounter.WithLabelValues("store_heartbeat", "client")
 	errGetOperatorSend       = forwardFailCounter.WithLabelValues("get_operator", "send")
 )
 
@@ -968,6 +969,8 @@ func (s *GrpcServer) StoreHeartbeat(ctx context.Context, request *pdpb.StoreHear
 					// reset to let it be updated in the next request
 					s.schedulingClient.CompareAndSwap(forwardCli, &schedulingClient{})
 				}
+			} else {
+				errStoreHeartbeatClient.Inc()
 			}
 		}
 	}
--- a/tests/integrations/mcs/scheduling/server_test.go
+++ b/tests/integrations/mcs/scheduling/server_test.go
@@ -395,6 +395,29 @@ func (suite *serverTestSuite) TestStoreHeartbeatForwardFailureWithNoSchedulingPr
 		return resp1Err == nil
 	})
 
+	// Snapshot the baseline counter value
+	getCounter := func() float64 {
+		mfs, gatherErr := prometheus.DefaultGatherer.Gather()
+		re.NoError(gatherErr)
+		for _, mf := range mfs {
+			if mf.GetName() == "pd_server_forward_fail_total" {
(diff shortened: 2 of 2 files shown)


liyishuai:
@coderabbitai full review

PR #10621, opened 2026-04-24
checker, scatter, server: add load-based split-scatter with range-aware group baseline

### LINKED ISSUE #10592: Scatter recently split regions in PD
### Background

After a region is split, the new regions may still stay concentrated on the same set of stores for a while.

For load-based split, this means the split can happen, but the load is not scattered soon enough.

We want PD to scatter newly split regions shortly after split, instead of relying on manual scatter or waiting for later balancing.

### Proposal

Add a simplified split-scatter flow in PD:

- When PD handles `AskBatchSplit`, record the source region and newly allocated region IDs in a short-lived pending cache.
- After the new regions report heartbeat to PD, use their latest region stats to calculate a simple priority score based on CPU and byte traffic.
- Maintain a bounded priority queue in PD for these pending split regions.
- Pop regions from the queue with a configurable limit and create scatter operators one by one.
- Regions from the same split batch should use the same scatter group so they can be scattered together.

lhy1024:

Issue Number: close #10592

This PR adds an internal split-scatter flow for newly split regions on the monolithic PD path.

Scope in this PR:
- Only splits with `SplitReason_LOAD` enter the new split-scatter flow.
- `HandleAskBatchSplit` records the source region and newly allocated region IDs into a short-lived pending state.
- The checker waits until the split result is observable in PD (heartbeat / source version advance), then dispatches internal scatter for the pending regions.
- Internal scatter uses a range-aware live baseline plus running-operator delta, so regions from the same table/index range can be scattered with shared group context across cascading split batches.
- Dispatch is limited by `schedule.split-scatter-schedule-limit`, default `4`. Setting it to `0` disables the internal split-scatter dispatcher. This limit only applies to internal split-scatter and does not affect admin scatter.
- The existing admin scatter model remains aligned with `pingcap/master`.

Out of scope in this PR:
- scheduling-service / NEXT_GEN forwarding path
- priority-score ordering and bounded priority queue proposed in #10592

The scheduling-service / NEXT_GEN gap is left as explicit TODOs in the forwarding code.

```commit-message
checker, scatter, server: add load-based split-scatter on monolithic PD path

Record load-based split batches in PD, wait until the split result is visible,
and dispatch internal scatter with range-aware grouping.

Keep admin scatter behavior aligned with pingcap/master. Leave
scheduling-service / NEXT_GEN support and the fuller priority-queue proposal
out of scope for follow-up work.
```

### Check List

Tests
- Unit test
- Manual validation on testbed for load-based split -> internal split-scatter flow
  - TiKV `9.0.0-beta.2` (`28890b9736b1553a162c63373fdc052bb60c9172`) on testbed `8069489`
  - PD image hash `523a77e2a0f5656c9d6cc78b1dc6b45649147dad`
  - TiDB copr cache disabled (`tikv-client.copr-cache.capacity-mb = 0`)
  - Background schedulers removed during the run (`balance-leader`, `balance-region`, `balance-hot-region`, `evict-*`) and `merge-schedule-limit` set to `0`
  - `UNIFIED_THRESHOLD=0.3`, single-region baseline: `1 -> 5` regions, PD created and finished internal `split-scatter-region` operators
  - `UNIFIED_THRESHOLD=0.2`, single-region baseline: `1 -> 33` regions, split-scatter activity concentrated in the first ~3 minutes and then converged
image
image
image

### Release note

```release-note
Load-based split-scatter currently applies only to monolithic PD. scheduling-service/NEXT_GEN forwarding does not propagate SplitReason for this flow yet.
```


lhy1024:
@coderabbitai full review

pkg/schedule/checker/split_scatter_group.go
```diff
@@ -0,0 +1,89 @@
...
+	endKey := region.GetEndKey()
+
+	// We intentionally over-approximate ambiguous table-key ranges. If PD can
+	// no longer prove the region stays within a single index prefix, it falls
+	// back to the table-scoped group instead of dropping back to the family
+	// group, so table-boundary splits and merged ranges still participate in
+	// the broader scatter continuity/baseline.
+	if len(endKey) == 0 || len(indexRange.startKey) == 0 || len(indexRange.endKey) == 0 || bytes.Compare(endKey, indexRange.endKey) > 0 {
+		return splitScatterPrefixRange(tablePrefix)
+	}
```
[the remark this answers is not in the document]
lhy1024: Addressed. We added a clarifying MemComparable comment there, explicitly documenting why bytes.Compare is valid for the key ordering.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,212 @@
...
+				continue
+			}
+			if !pending.expireAt.IsZero() && !now.Before(pending.expireAt) {
+				delete(c.pending, regionID)
+			}
+		}
+		c.pendingMu.Unlock()
+	}
+	return candidates
+}
```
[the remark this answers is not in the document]
lhy1024: Addressed. We added a TODO in collectTopPendingSplitScatter() to make it explicit that priority ordering is intentionally deferred for later work.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,212 @@
...
+					zap.String("group", pending.group),
+					zap.String("operator-desc", op.Desc()))
+				continue
+			}
+		}
+		c.pendingMu.Lock()
+		delete(c.pending, pending.regionID)
+		c.pendingMu.Unlock()
+	}
+}
```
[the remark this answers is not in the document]
lhy1024: The pending entries are now bounded by expireAt, and the dispatch limit is small, so we accept this residual risk in the current PR and keep the retry semantics simple. If we later observe stable log spam / CPU waste on these paths, we can split them out and refine the backoff policy in a dedicated follow-up.

pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -1014,3 +1293,238 @@ func TestScatterWithAffinity(t *testing.T) {
...
+	op, err := scatterer.Scatter(region, "", true)
+	re.ErrorContains(err, "is hot")
+	re.Nil(op)
+
+	op, err = scatterer.ScatterInternal(region, "", region.GetStartKey(), region.GetEndKey())
+	re.NoError(err)
+	if op != nil {
+		re.Equal(InternalScatterOperatorDesc, op.Desc())
+	}
+}
```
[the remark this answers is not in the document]
lhy1024: The purpose of this test is to verify that ScatterInternal is not rejected by the hot-region gate the same way admin Scatter is, not to guarantee operator creation in this exact fixture.

server/cluster/split_scatter_test.go
```diff
@@ -0,0 +1,225 @@
...
+	re.NoError(cluster.processRegionHeartbeat(
+		core.ContextTODO(),
+		newSplitScatterRegion(splitRegionID, []byte("m"), []byte(""), 120),
+	))
+
+	dispatchSplitScatterInPatrol(t, cluster, cancelPatrol, nil)
+
+	re.Nil(cluster.GetOperatorController().GetOperator(splitRegionID))
+	re.Nil(cluster.GetOperatorController().GetOperator(100))
+}
```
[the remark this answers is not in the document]
lhy1024: Partially addressed. We removed the original blind-sleep branch from the shared helper and made each test case provide its own explicit wait condition. The size-reason case still effectively waits for one ticker window rather than a stronger observable progress signal, so this is an intentional partial convergence rather than a full implementation of the suggested deterministic-wait approach.

server/cluster/split_scatter_test.go
```diff
@@ -0,0 +1,224 @@
...
+		opGroup, ok := op.GetAdditionalInfo("group")
+		re.True(ok)
+		if group == "" {
+			group = opGroup
+			re.True(strings.HasPrefix(group, "split-scatter-100-"))
+			continue
+		}
+		re.Equal(group, opGroup)
+	}
+	re.NotEmpty(group)
```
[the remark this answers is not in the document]
lhy1024: Addressed in 15549f045. The test now waits until both split children have internal scatter operators before stopping patrol, and it fails if either operator is missing.

server/grpc_service.go
```diff
@@ -1743,6 +1743,8 @@ func (s *GrpcServer) AskBatchSplit(ctx context.Context, request *pdpb.AskBatchSp
 		}
 		cli := forwardCli.getClient()
 		if cli != nil {
+			// TODO(split-scatter): propagate split reason and support load-based split-scatter
```
rleungx, line 1746: This makes load-split scatter behave differently between monolithic PD and scheduling-service/NEXT_GEN deployments. If that gap is intentional for this PR, can we call it out with a feature gate or release note so operators are not surprised?

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *splitScatterController) collectTopPendingSplitScatter(limit int) []splitScatterPendingItem {
+	if limit <= 0 {
+		return nil
+	}
+	// TODO: currently iterating over the pending map in random order and
+	// truncating to limit; consider ordering by region priority score.
+	now := time.Now()
+	c.pendingMu.RLock()
+
+	candidates := make([]splitScatterPendingItem, 0, len(c.pending))
```
rleungx: This runs on every patrol tick, but it scans the whole pending map before taking only 4 items. Under a load-split burst, that can become steady O(pending) work for up to the 3m TTL. Can we use a bounded queue keyed by retry/expire/priority, or at least cap the pending size?

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -407,48 +662,52 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
...
-		r.Put(targetPeers, targetLeader, group)
+		if state != nil {
+			r.applyScatterStateDelta(state, region, targetPeers, targetLeader, group, true)
+		} else {
+			r.Put(targetPeers, targetLeader, group)
+		}
 		return nil, nil
 	}
-	op, err := operator.CreateScatterRegionOperator(scatterOperatorDesc, r.cluster, region, targetPeers, targetLeader, skipStoreLimit)
+	op, err := operator.CreateScatterRegionOperator(desc, r.cluster, region, targetPeers, targetLeader, skipStoreLimit)
```
rleungx: This internal path still creates an OpAdmin scatter operator with high priority. Since it is now triggered automatically by load splits, it may compete with regular hot/balance scheduling more aggressively than expected. Should this use a lower priority or be guarded by a config switch?

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,224 @@
...
+	group := makeSplitScatterGroup(sourceRegionID, newRegionIDs[0])
+	expireAt := time.Now().Add(splitScatterPendingTTL)
+	sourceWaitVersion := uint64(1)
+	if sourceRegion := c.cluster.GetRegion(sourceRegionID); sourceRegion != nil && sourceRegion.GetRegionEpoch() != nil {
+		sourceWaitVersion = sourceRegion.GetRegionEpoch().GetVersion() + 1
+	}
+	c.pendingMu.Lock()
+	defer c.pendingMu.Unlock()
+	for _, regionID := range newRegionIDs {
+		c.pending[regionID] = splitScatterPendingItem{regionID: regionID, group: group, expireAt: expireAt}
```
rleungx: Only the source region waits for the version bump here. The new regions can be dispatched first if their heartbeats arrive earlier, so the range baseline may be built from a partially visible split. I think the whole batch should wait until the source version has advanced.

server/grpc_service.go
lhy1024: I think it's okay to add it. The reason I didn't add it before was because I was worried that this PR would be too big since it was intended as a patch.

pkg/schedule/checker/split_scatter.go
lhy1024: In the initial version, I added a priority queue sorted by CPU, but I removed it for reasons similar to avoiding excessively high pull requests. I will adopt a simpler implementation method.

pkg/schedule/checker/split_scatter.go
lhy1024: ok, I will fix it.

pkg/schedule/scatter/region_scatterer.go
lhy1024: Internal scatter is a transient, one-off operation, so I think it's reasonable to give it higher priority?

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,269 @@
...
+			if c.opController.ExceedStoreLimit(op) {
+				c.delayPendingSplitScatter(pending)
+				log.Info("dispatch internal split scatter delayed",
+					zap.Uint64("region-id", pending.regionID),
+					zap.String("group", pending.group),
+					zap.String("reason", "exceed-store-limit"),
+					zap.String("operator-desc", op.Desc()))
+				continue
+			}
+			if !c.opController.AddOperator(op) {
```
liyishuai, line 390: Will the next loop body get the operator added here? If yes, we should explicitly mention that the loop is sequential.

server/grpc_service.go
JmPotato, line 1746: This TODO is not enough to cover the user-visible deployment-mode difference. Please also state in the PR body / release note that load-split auto scatter only works on monolithic PD, and NEXT_GEN/scheduling-service is not supported yet.

pkg/schedule/scatter/region_scatterer.go
JmPotato: This still creates an `OpAdmin` operator, but the trigger is an automatic background split-scatter flow. Please either use a non-admin kind, or add tests that pin down the kind / priority / store-limit / metrics semantics.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,271 @@
...
+		if sourceRegion.GetRegionEpoch() != nil {
+			sourceVersion = sourceRegion.GetRegionEpoch().GetVersion()
+		}
+		if pending.sourceWaitVersion > 0 && sourceVersion < pending.sourceWaitVersion {
+			continue
+		}
+		candidates = append(candidates, pending)
+	}
+	if len(candidates) > limit {
+		candidates = candidates[:limit]
```
JmPotato, line 146: `pending` comes from a map, so truncating before sorting makes dispatch random. Please sort by `(group, regionID)` before truncation to make it deterministic and group-friendly.

server/cluster/split_scatter_test.go
```diff
@@ -0,0 +1,227 @@
...
+	splitRegionIDs := []uint64{
+		resp.GetIds()[0].GetNewRegionId(),
+		resp.GetIds()[1].GetNewRegionId(),
+	}
+	re.NoError(cluster.processRegionHeartbeat(core.ContextTODO(), newSplitScatterRegion(100, []byte(""), []byte("m"), 0).Clone(core.WithIncVersion())))
+	re.NoError(cluster.processRegionHeartbeat(core.ContextTODO(), newSplitScatterRegion(splitRegionIDs[0], []byte("m"), []byte("t"), 120)))
+	re.NoError(cluster.processRegionHeartbeat(core.ContextTODO(), newSplitScatterRegion(splitRegionIDs[1], []byte("t"), []byte(""), 80)))
+
+	dispatchSplitScatterInPatrol(t, cluster, cancelPatrol, func() bool {
+		return cluster.GetOperatorController().GetOperator(splitRegionIDs[0]) != nil ||
```
JmPotato: This stops once either child has an operator, and the later loop skips nil operators. The test can pass even if the other child is never scattered. Please wait for both children, or lower the test claim.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,271 @@
...
+	c.pendingMu.RUnlock()
+
+	if len(expiredRegionIDs) > 0 {
+		c.pendingMu.Lock()
+		for _, regionID := range expiredRegionIDs {
+			pending, ok := c.pending[regionID]
+			if !ok {
+				continue
+			}
+			if !pending.expireAt.IsZero() && !now.Before(pending.expireAt) {
```
JmPotato: TTL expiration is a silent skip here. Please add `split_scatter_pending_expired_total` so TTL/capacity-related missed scatters are observable.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,271 @@
...
+	if _, ok := c.pending[sourceRegionID]; !ok {
+		newPendingCount++
+	}
+	for _, regionID := range newRegionIDs {
+		if _, ok := c.pending[regionID]; !ok {
+			newPendingCount++
+		}
+	}
+	if len(c.pending)+newPendingCount > splitScatterPendingLimit {
+		log.Info("skip recording split scatter batch due to pending limit",
```
JmPotato, line 295: Dropping pending entries because of the cap should not be log-only. Please add `split_scatter_pending_dropped_total`, together with the expired counter, to expose the v1 capacity limit.

server/grpc_service.go
lhy1024: update

pkg/schedule/scatter/region_scatterer.go
lhy1024: add test

lhy1024: internal scatter operator: SchedulerKind=OpAdmin priority=High

pkg/schedule/checker/split_scatter.go
lhy1024: add sorting

pkg/schedule/checker/split_scatter.go
lhy1024: add counter

pkg/schedule/checker/split_scatter.go
lhy1024: done

pkg/schedule/scatter/region_scatterer.go
rleungx: OpAdmin is usually triggered by calling the API manually.

lhy1024: > OpAdmin is usually triggered by calling the API manually.

I changed the internal split-scatter path to build a non-admin scatter operator whose kind is derived from the generated steps. The public/admin scatter path still uses OpAdmin.

pkg/schedule/checker/split_scatter_test.go
```diff
@@ -0,0 +1,549 @@
...
+
+func (c *Controller) collectTopPendingSplitScatter(limit int) []splitScatterPendingItem {
+	return c.splitScatter.collectTopPendingSplitScatter(limit)
+}
+
+func (c *Controller) dispatchSplitScatterRegions() {
+	c.splitScatter.dispatchSplitScatterRegions()
+}
+
+func TestRecordSplitScatterBatchCollectsPendingRegions(t *testing.T) {
```
rleungx, line 57: Is there a test to cover that one region keeps splitting?

lhy1024: added

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -407,48 +662,56 @@ func (r *RegionScatterer) scatterRegion(region *core.RegionInfo, group string, s
 		}
 	}
 
-	scatterWithSameEngine(ordinaryPeers, r.ordinaryEngine)
+	scatterWithSameEngine(ordinaryPeers, ordinaryContext)
```
rleungx: Looks like this changes the old leader selection behavior a bit. `scatterWithSameEngine` appends to `leaderCandidateStores`, so after moving special-engine peers before leader selection, special-engine stores may also be considered as leader targets. The comment below still says leader selection only considers ordinary stores. Could we keep ordinary leader candidates separate, or select the leader before handling special peers?

pkg/schedule/scatter/region_scatterer_test.go
```diff
@@ -554,14 +571,22 @@ func TestScattersGroup(t *testing.T) {
 		} else {
 			re.Empty(failures)
 		}
+		stream.Close()
```
rleungx: Small cleanup concern: after enabling `scatterFail`, we only disable it near the end of the case. If any `require` fails before that, the failpoint may stay enabled and affect later tests. Could we defer the disable right after enabling it, and also defer `stream.Close()` / `cancel()` per case?

rleungx (approved):

pkg/schedule/checker/split_scatter_group.go
```diff
@@ -0,0 +1,107 @@
...
+func splitScatterPrefixRange(rawPrefix []byte) splitScatterRangeHint {
+	return splitScatterPrefixRangeWithGroup(rawPrefix, "")
+}
+
+func splitScatterPrefixRangeWithGroup(rawPrefix []byte, scatterGroup string) splitScatterRangeHint {
+	startKey := append([]byte(nil), codec.EncodeBytes(rawPrefix)...)
+	endRawPrefix := splitScatterNextPrefix(rawPrefix)
+	if len(endRawPrefix) == 0 {
+		return splitScatterRangeHint{startKey: startKey, scatterGroup: scatterGroup}
+	}
```
liyishuai, line 84: Is this branch reachable?

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,307 @@
...
+	sort.Slice(candidates, func(i, j int) bool {
+		if candidates[i].group != candidates[j].group {
+			return candidates[i].group < candidates[j].group
+		}
+		return candidates[i].regionID < candidates[j].regionID
+	})
+	if len(candidates) > limit {
+		candidates = candidates[:limit]
+	}
+	c.pendingMu.RUnlock()
```
liyishuai: This lock is long. Can we take a snapshot under a short lock, and read from the snapshot after releasing the lock?

pkg/schedule/checker/split_scatter_group.go
lhy1024: add more comments

pkg/schedule/checker/split_scatter.go
lhy1024: done

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/tikv/pd/pkg/schedule/filter"
+	"github.com/tikv/pd/pkg/schedule/operator"
+	"github.com/tikv/pd/pkg/schedule/scatter"
+	"github.com/tikv/pd/pkg/schedule/types"
+	"github.com/tikv/pd/pkg/utils/syncutil"
+)
+
+const (
+	splitScatterPendingLimit = 1024
+	splitScatterRetryBackoff = time.Second
```
bufferflies, line 40: how about using Exponential Backoff?

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+	expireAt          time.Time
+}
+
+type splitScatterController struct {
+	cluster         sche.CheckerCluster
+	opController    *operator.Controller
+	regionScatterer *scatter.RegionScatterer
+
+	pendingMu syncutil.RWMutex
+	pending   map[uint64]splitScatterPendingItem
```
bufferflies: Add one comment for the map key.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+	running := c.opController.OperatorCount(operator.OpSplitScatter)
+	if running >= limit {
+		operator.IncOperatorLimitCounter(types.SplitScatterChecker, operator.OpSplitScatter)
+		return
+	}
+	dispatchLimit := int(limit - running)
+	// Dispatch sequentially so operators added for earlier pending items in this pass
+	// are visible to later ScatterInternal calls through the running-operator delta.
+	for _, pending := range c.collectTopPendingSplitScatter(dispatchLimit) {
+		region := c.cluster.GetRegion(pending.regionID)
```
bufferflies, line 341: It may panic if the dispatchLimit<0.

pkg/schedule/checker/split_scatter.go
bufferflies: Should the sort consider the expiration time to prioritize items that are close to expiration?

pkg/schedule/checker/split_scatter.go
bufferflies: I think this check has been done in lines 94~100.

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	group := makeSplitScatterGroup(sourceRegionID, newRegionIDs[0])
+	expireAt := time.Now().Add(splitScatterPendingTTL)
+	sourceWaitVersion := uint64(1)
+	if sourceRegion := c.cluster.GetRegion(sourceRegionID); sourceRegion != nil && sourceRegion.GetRegionEpoch() != nil {
+		sourceWaitVersion = sourceRegion.GetRegionEpoch().GetVersion() + 1
+	}
+	c.pendingMu.Lock()
+	defer c.pendingMu.Unlock()
+	if expiredCount := c.removeExpiredPendingSplitScatterLocked(time.Now()); expiredCount > 0 {
```
bufferflies: I think the function doesn't need the input(time.now())

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+	newPendingCount := 0
+	if _, ok := c.pending[sourceRegionID]; !ok {
+		newPendingCount++
+	}
+	for _, regionID := range newRegionIDs {
+		if _, ok := c.pending[regionID]; !ok {
+			newPendingCount++
+		}
+	}
+	if len(c.pending)+newPendingCount > splitScatterPendingLimit {
```
bufferflies, line 290: It will reject all new pending items if the pending < splitScatterPendingLimit

pkg/schedule/scatter/region_scatterer.go
```diff
@@ -67,9 +66,16 @@ const (
 	maxSleepDuration     = time.Minute
 	initialSleepDuration = 100 * time.Millisecond
 	maxRetryLimit        = 30
-	scatterOperatorDesc  = "scatter-region"
+	// AdminScatterOperatorDesc is used by external admin/API scatter requests.
+	AdminScatterOperatorDesc = "scatter-region"
+	// InternalScatterOperatorDesc is used by PD-internal split-scatter dispatch.
+	InternalScatterOperatorDesc = "internal-scatter-region"
```
bufferflies: how about rename direct as split-scatter-region

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,316 @@
...
+			c.delayPendingSplitScatter(pending)
+			log.Info("dispatch internal split scatter failed",
+				zap.Uint64("region-id", pending.regionID),
+				zap.String("batch-group", pending.group),
+				zap.String("scatter-group", scatterGroup),
+				zap.Error(err))
+			continue
+		}
+		if op != nil {
+			op.SetAdditionalInfo("batch-group", pending.group)
```
bufferflies, line 378: batch-group is confused.

pkg/schedule/checker/split_scatter.go
lhy1024: The 1s backoff is only the minimum retry interval. The actual retry happens on the next checker dispatch after retryAt, so it is also bounded by the checker tick/batch. I prefer keeping it fixed and short because these failures are usually transient and the pending TTL is limited.

pkg/schedule/checker/split_scatter.go
lhy1024: ok,  near-expiration first

pkg/schedule/checker/split_scatter.go
lhy1024: Do you have a good idea for it? pending-group?

pkg/schedule/checker/split_scatter.go
lhy1024: yes

pkg/schedule/checker/split_scatter.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/tikv/pd/pkg/utils/syncutil"
+)
+
+const (
+	splitScatterPendingLimit = 1024
+	// The actual retry cadence is also bounded by the checker dispatch loop. This
+	// is only the minimum interval to avoid retrying the same pending item too
+	// frequently when checker ticks are fast.
+	splitScatterRetryBackoff = time.Second
+	splitScatterPendingTTL   = 3 * time.Minute
```
bufferflies: If the patrol region tree interval exceeds 3 minutes, the region will expire and will not scatter.

lhy1024: I extended the pending TTL to 10 minutes, aligned with PD's slow operator step threshold, so pending split-scatter items have enough time to be retried before expiring.

lhy1024: I also added pending observability: current pending count, dropped pending entries, and expired pending entries split by whether they had ever been selected for dispatch. This should make it clear whether entries are dropped by capacity, expire before the first dispatch attempt, or expire after repeated blocked attempts.

pkg/schedule/checker/split_scatter.go
bufferflies: load-split-scatter

bufferflies (approved):

niubell (approved):
LGTM

PR #10622, opened 2026-04-25
pkg/election: add lease keepalive metrics

JmPotato:

Issue Number: ref #10653.

```commit-message
pkg/election: add lease keepalive metrics
```

Add Prometheus metrics for the current lease keepalive loop:
- `pd_lease_keepalive_response_interval_seconds{purpose}` — histogram of intervals between consecutive valid `KeepAliveOnce` responses.
- `pd_lease_keepalive_request_duration_seconds{purpose,result}` — histogram of `Lease.KeepAliveOnce` request latency, broken down by `result` (`success` / `error` / `canceled`).
- `pd_lease_local_ttl_remaining_seconds{purpose}` — gauge of the PD local estimate of remaining lease TTL, updated on every keepalive response and on keepalive timeout firing; reset to `0` on `Lease.Close()` so a retired lease does not show stale TTL.
- `pd_lease_renewal_termination_total{purpose,reason}` — counter of keepalive loop terminations, with reasons `invalid_ttl`, `lease_expired` (abnormal) and `context_canceled` (benign, caller-initiated shutdown).

Metric children are cached on `Lease` at construction time so the keepalive path does not call `.WithLabelValues`. Streaming-only reasons such as stream setup or channel-close failures are intentionally not introduced in this PR; they can be added when #10618 lands.

Additional review follow-ups:
- Harden `loadExpireTime()` for nil leases, empty values, and unexpected value types, returning `typeutil.ZeroTime` so missing/invalid expire time is treated as expired instead of panicking.
- Add a collapsed `Leader` row at the end of the PD Grafana dashboard, showing `Leader/Primary`, `Raft term`, lease local TTL remaining, lease keepalive response interval, lease keepalive request duration and lease renewal terminations.
- Add `reason="context_canceled"` and skip `lease_expired` accounting when the keepalive timeout races with caller cancellation. Each cancellation path emits a descriptive log line (with `actual-expire`, `last-response-time`, `max-expire` where relevant) so operators can identify which branch fired. Operators alerting on actual failures should filter `reason!~"context_canceled"`.
- Keep lease metric children for the lifetime of the process. They are created on first use via `newLeaseMetrics` and are never deleted on `Lease.Close()` or keyspace group retire, so per-purpose series remain continuous across re-elections, group splits / merges and primary handovers. Cardinality is bounded by the live + historical purpose set (PD leader election, microservice primary / expected-primary leases, per-keyspace-group primary elections), so the cost is negligible.
- Prefix the scheduling and resource_manager microservices' primary election lease purpose with the service name (`scheduling primary election`, `resource_manager primary election`) to match the existing `<service> expected primary` purpose used by `KeepExpectedPrimaryAlive`, so each microservice's primary election is distinguishable on shared lease panels without leaning on `instance` / `job` labels.
- Filter out `<service> expected primary` purposes on every lease panel; the housekeeping lease tracks the same lease timeout as the corresponding primary election lease and is mostly noise.
- Drop `instance` from aggregation and legends on every lease panel so each `(job, purpose)` series stays continuous across primary handover instead of breaking into per-instance segments. The TTL panel uses `max() by (job, purpose)` to pick the live primary's value over zero from non-primary instances.

### Check List

Tests

- Unit test
- Manual test (dashboard JSON validation, local TiUP playground microservices mode with multiple keyspace groups, transfer leader and `kill -9` leader)

- `make gotest GOTEST_ARGS='./pkg/election -count=1'`
- `make gotest GOTEST_ARGS='./pkg/tso -run TestKeyspaceGroupPrimaryElectionPurposeIncludesGroupID -count=1'`
- `make static PACKAGE_DIRECTORIES='./pkg/election ./pkg/tso ./pkg/mcs/scheduling/server ./pkg/mcs/resourcemanager/server' SUBMODULES=`
- `jq empty metrics/grafana/pd.json`
- `git diff --check`

image

### Release note

```release-note
None.
```


pkg/election/lease.go
```diff
@@ -78,9 +79,14 @@ func (l *Lease) Grant(leaseTimeout int64) error {
 	l.ID.Store(leaseResp.ID)
 	l.leaseTimeout = time.Duration(leaseTimeout) * time.Second
 	l.expireTime.Store(start.Add(time.Duration(leaseResp.TTL) * time.Second))
+	l.initMetrics()
```
rleungx: Since this initializes a per-purpose TTL gauge, please also clear it when the lease is closed or reset. Otherwise `local_ttl_remaining_seconds` can keep the last positive sample after stepdown/resign and make the dashboard look stale.

JmPotato: Nice catch. Fixed.

pkg/election/lease.go
```diff
@@ -138,9 +165,13 @@ func (l *Lease) KeepAlive(ctx context.Context) {
 					l.expireTime.Store(t)
 				}
 			}
+			l.metrics.ttlRemaining.Set(maxExpire.Sub(now).Seconds())
 			timer.Reset(l.leaseTimeout)
 		case <-timer.C:
-			log.Info("keep alive lease too slow", zap.Duration("timeout-duration", l.leaseTimeout), zap.Time("actual-expire", l.expireTime.Load().(time.Time)), zap.String("purpose", l.Purpose))
+			actualExpire := l.loadExpireTime()
+			l.metrics.ttlRemaining.Set(time.Until(actualExpire).Seconds())
+			l.metrics.leaseExpired.Inc()
```
rleungx, line 175: This can still count a clean shutdown as `lease_expired` if `ctx.Done()` and `timer.C` are ready at the same time. Please check `ctx.Err()` before incrementing the failure counter, and return without recording a failure when the keepalive loop is already canceled.

pkg/election/lease.go
```diff
@@ -78,16 +79,26 @@ func (l *Lease) Grant(leaseTimeout int64) error {
 	l.ID.Store(leaseResp.ID)
 	l.leaseTimeout = time.Duration(leaseTimeout) * time.Second
 	l.expireTime.Store(start.Add(time.Duration(leaseResp.TTL) * time.Second))
+	l.initMetrics()
```
bufferflies: How about moving this into NewLease?

bufferflies (commented):
rest lgtm

bufferflies (approved):

pkg/election/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	)
+
+	localTTLRemaining = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "local_ttl_remaining_seconds",
+			Help:      "Event-sampled PD local estimate of remaining lease TTL based on observed keepalive responses or local expiration checks; not etcd authoritative TTL.",
+		},
+		[]string{metricsLabelPurpose},
```
lhy1024, line 72: local_ttl_remaining_seconds is labeled only by purpose, but one process can have multiple Lease instances with the same purpose, for example TSO keyspaces using "keyspace group primary election". Those leases share the same Gauge child, making the value last-writer-wins; with the new Close() reset, closing one lease can also clear the TTL for another still-active lease. Please add a bounded lease identity label or avoid exposing this gauge for purposes that can have multiple concurrent lease instances.

JmPotato: There will indeed be such a problem; let me find a way to solve it.

JmPotato: Added a TODO comment explaining the current trade-off.

https://github.com/tikv/pd/pull/10622/changes#diff-1a7c3205b8ee906fd57f7b3c82f97c1626942b8d54959612c0c023fcad85e141R30-R38

pkg/election/lease.go
```diff
@@ -88,6 +90,12 @@ func (l *Lease) Close() error {
 	}
 	// Reset expire time.
 	l.expireTime.Store(typeutil.ZeroTime)
+	// Drop both keepalive histogram series so the dashboard does not keep
+	// displaying stale percentiles for up to the rate() window after the
+	// lease stops renewing. The series are automatically recreated on the
+	// next successful keepalive observation when the lease is re-granted.
+	keepAliveResponseInterval.DeleteLabelValues(l.Purpose)
+	localTTLRemaining.DeleteLabelValues(l.Purpose)
```
lhy1024: We should not delete histogram children only by purpose here. In the TSO keyspace group case, one process can have multiple active Lease instances with the same purpose, e.g. "keyspace group primary election". When one Lease.Close() deletes the shared series for that purpose, other still-active leases keep using the observers cached in NewLease(), so later Observe() calls write to the deleted child and may no longer be exported by Prometheus. Please avoid DeleteLabelValues in Close(), or make the cleanup aware of active same-purpose leases, e.g. via ref counting/lifecycle tracking.

pkg/election/lease.go
```diff
@@ -88,6 +90,14 @@ func (l *Lease) Close() error {
...
 	// Reset expire time.
 	l.expireTime.Store(typeutil.ZeroTime)
+	// Reset the local TTL remaining gauge so the dashboard does not show a
+	// stale positive sample after stepdown/resign.
+	l.metrics.ttlRemaining.Set(0)
+	// Drop the keepalive response interval histogram series so the dashboard
+	// does not keep displaying stale percentiles for up to the rate() window
+	// after the lease stops renewing. The series is automatically recreated on
+	// the next successful keepalive observation when the lease is re-granted.
+	keepAliveResponseInterval.DeleteLabelValues(l.Purpose)
```
rleungx: Could we avoid deleting this HistogramVec child here, or refresh the cached observer before the next KeepAlive? `Lease.metrics.responseInterval` caches the child created in `NewLease`, and client_golang keeps a deleted child non-exported even if the cached handle is observed later. If the same `Lease` is reused via `Close -> Grant -> KeepAlive`, this histogram may stop being exported.

metrics/grafana/pd.json
```diff
@@ -16367,6 +16367,483 @@
...
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "expr": "histogram_quantile(0.99, sum(rate(pd_lease_local_ttl_remaining_seconds_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}[1m])) by (job, purpose, instance, le))",
```
lhy1024: P99?P1?

pkg/election/lease.go
```diff
@@ -178,14 +212,16 @@ func (l *Lease) keepAliveWorker(ctx context.Context, interval time.Duration) <-c
...
-					select {
-					case ch <- expire:
-					// Here we don't use `ctx1.Done()` because we want to make sure if the keep alive success, we can update the expire time.
-					case <-ctx.Done():
-						log.Info("lease keep alive once exit", zap.String("purpose", l.Purpose), zap.Time("start", start), zap.Time("expire", expire))
-					}
+				if res.TTL <= 0 {
+					l.metrics.invalidTTL.Inc()
+					return
+				}
```
lhy1024, line 224: This res.TTL <= 0 branch seems unreachable: clientv3.KeepAliveOnce turns non-positive TTL into ErrLeaseNotFound, and the err != nil branch above returns immediately. As a result, renewal_termination_total{reason="invalid_ttl"} may never be emitted.

YuhaoZhang00 (approved):

pkg/election/lease.go
JmPotato: Yes, `KeepAliveOnce` currently converts non-positive TTL to `ErrLeaseNotFound`, so this branch is not expected to be hit with the current etcd client. I kept it intentionally as a defensive guard for the successful-response path in case a mocked/custom lease implementation or future client behavior returns nil error with a non-positive TTL. The label handles are cached during lease metric initialization, and the runtime cost is only this cheap guard. I also added a code comment to make the intent explicit.

metrics/grafana/pd.json
JmPotato: Changed the remaining-TTL panel from P99 to P1. For TTL remaining, lower values are the risk signal, so the lower percentile is more useful. I kept `keepalive_response_interval_seconds` at P99 because larger intervals are the risky direction there.

lhy1024 (approved):

PR #9076, opened 2025-02-12
*: add QueryRegion related metrics

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

### What problem does this PR solve?

Issue Number: ref #8690.

### What is changed and how does it work?

```commit-message
Add QueryRegion related metrics.
```

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

图片

### Release note

```release-note
None.
```

PR #10628, opened 2026-04-28
*: backport QueryRegion gRPC stream to release-8.5

JmPotato:

Issue Number: ref #8690

Backports the `QueryRegion` gRPC stream to `release-8.5`, including both:

- the PD server-side stream handler and batched region lookup logic; and
- the Go client-side opt-in stream path with unary fallback.

This lets release-8.5 PD serve and consume batched region queries (key / prev-key / ID lookups) through the same QueryRegion stream semantics used on master, while keeping the feature disabled by default on the client side.

### Dependency

Requires the kvproto QueryRegion proto on the `release-8.5` line:
**pingcap/kvproto#1467** (commit `f96c651`) must merge first. This PR pins
kvproto to `v0.0.0-20260518092652-f96c651c7702` (kvproto release-8.5 tip
with the QueryRegion proto cherry-picked and `pdpb.pb.go` regenerated using
the CI-pinned protoc 3.8.0 + gogo v1.3.2 toolchain, so release-8.5-only
protos such as `StoreStats.NetworkSlowScores` are retained).

This PR contains eight commits:

1. `chore: bump kvproto for QueryRegion backport` — pin kvproto across all four modules.
2. `feat: backport QueryRegion gRPC server to release-8.5` — backport the server-side QueryRegion path:
   - **#8979** server, core: implement the query region gRPC server.
     Adds `RegionsInfo.QueryRegions`, `regionTree.searchByKeys/searchByPrevKeys`,
     the `GrpcServer.QueryRegion` stream handler and server metric. The
     `rateLimitCheck` / `errs.ErrMismatchClusterID` helpers (added later
     upstream by #8995) are adapted to release-8.5 inline idioms
     (`currentFunction` limiter + `status.Errorf`).
   - **#9055** size `regionsByID` with `len(regions)+len(prevRegions)+len(ids)`.
   - **#9076** add QueryRegion related metrics (`pkg/core` `query_region_*`
     vecs, server `query_region_duration_seconds`, per-step metrics and guards).
   - **#9196** allow the follower to handle QueryRegion (leader/follower
     branch using `s.cluster` + region syncer, mirroring the `GetRegion`
     follower path).
   - **#10194** optimize region query with batched lookups
     (`slice.SplitIntoBatches`, `getRegionsByIDs`, `batchSearchSize` 16→128,
     drop panic assertions).
3. `refactor(client): genericize TSO batch controller` — backport master-aligned
   `client/pkg/batch.Controller[T]` and migrate the 8.5 TSO dispatcher onto it.
   The generic controller is kept aligned with master; 8.5-specific deltas are
   limited to the testutil import path and preserving 8.5 TSO `suffixBits` in the
   TSO request finisher.
4. `refactor(client): genericize TSO connection context manager` — backport
   master-aligned `client/pkg/connectionctx.Manager[T]` and migrate the 8.5 TSO
   stream connection map onto it. The generic manager is kept aligned with master;
   8.5 keeps the existing per-dispatcher manager shape to preserve dc-local TSO behavior.
5. `feat(client): backport QueryRegion stream client` — backport the client-side
   QueryRegion stream path on top of the generic batch and connectionctx helpers:
   - `GetRegion`, `GetPrevRegion`, and `GetRegionByID` use the stream path when
     `EnableQueryRegion` is enabled, and fall back to unary RPC on stream errors.
   - `EnableQueryRegion` is dynamic and defaults to `false`.
   - the client request finisher preserves master semantics for mixed key / prev-key / ID batches.
   - returned region objects are deep-copied, preserving the #9080 behavior.
   - router/microservice client code is intentionally not backported because the
     release-8.5 target only needs the PD leader/follower path.
6. `fix(client): update QueryRegion follower streams on option change` — keep
   the release-8.5 stream client aligned with master follower-handle behavior by
   notifying the QueryRegion connection daemon when `EnableFollowerHandle` changes.
7. `test(client): backport QueryRegion client coverage` — backport/adapt the
   master QueryRegion client unit and integration coverage to release-8.5. The
   router/microservice integration tests are intentionally not included because
   release-8.5 does not backport `routerpb` or the router microservice path.
8. `fix(client): stop QueryRegion client on disable` — mirror the master router
   client initializer lifecycle (#9082) for the release-8.5 QueryRegion client:
   enabling creates the stream client through the initializer, disabling closes
   it after releasing the client lock, and repeated same-value updates do not
   emit duplicate notifications. Release-8.5-specific deltas are limited to
   keeping the local option setter style (get-then-store instead of master CAS)
   and explicitly draining the QueryRegion client during `Close()` after the
   parent context is canceled; this is deterministic shutdown for the 8.5
   in-package client shape and does not change request behavior.

This PR does not flip `defaultEnableQueryRegion`.

### Check List

Tests

- Unit test (`pkg/core` `TestQueryRegions`, `pkg/slice` `TestSliceSplitIntoBatches`, plus regenerated benchmark `BenchmarkQueryRegions`)
- Unit test (`client/pkg/batch`, `client/pkg/connectionctx`, QueryRegion client request/finisher/fallback tests, including the master no-data-race finisher case)
- Integration test (`tests/integrations/client` QueryRegion client enabled/disabled suites and `TestGetRegionByQueryRegionStream`)

Release note

```release-note
None.
```


client/client.go
```diff
@@ -903,6 +998,19 @@ func (c *client) GetRegion(ctx context.Context, key []byte, opts ...GetRegionOpt
...
 		opt(options)
 	}
+	if c.option.getEnableRouterClient() {
+		resp, fallback, err := c.queryRegion(ctx, &pdpb.QueryRegionRequest{
+			Header:      c.requestHeader(),
+			NeedBuckets: options.needBuckets,
+			Keys:        [][]byte{key},
+		}, options.allowFollowerHandle && c.option.getEnableFollowerHandle())
+		if !fallback {
+			if err = c.respForErr(metrics.CmdFailDurationGetRegion, start, err, resp.GetHeader()); err != nil {
```
rleungx: Could we guard resp before calling resp.GetHeader() here? queryRegion can return resp == nil with fallback == false on non-fallback stream errors, which would panic instead of returning the original error.

server/grpc_service.go
```diff
@@ -1636,6 +1637,73 @@ func (s *GrpcServer) GetRegionByID(ctx context.Context, request *pdpb.GetRegionB
...
+	if s.GetServiceMiddlewarePersistOptions().IsGRPCRateLimitEnabled() {
+		fName := currentFunction()
+		limiter := s.GetGRPCRateLimiter()
+		if done, err := limiter.Allow(fName); err == nil {
+			defer done()
+		} else {
+			return err
+		}
+	}
+	for {
```
rleungx, line 1651: Could we handle forwarding metadata for this stream before serving requests locally? With WithForwardingOption enabled, the client may connect to a follower and expect the request to be forwarded to the leader, but this path reads the follower cache directly.

tests/integrations/client/router_client_test.go
```diff
@@ -0,0 +1,218 @@
...
+		EndKey:   []byte("zz"),
+		Peers:    peers,
+	}
+	re.NoError(regionHeartbeat.Send(&pdpb.RegionHeartbeatRequest{
+		Header: newHeader(),
+		Region: region2,
+		Leader: peers[0],
+	}))
+
+	testutil.Eventually(re, func() bool {
```
rleungx: Could we assert that the success path actually opens `/pdpb.PD/QueryRegion`? Otherwise this test would still pass if the client silently fell back to the unary GetRegion path.

server/grpc_service.go
```diff
@@ -1636,6 +1637,73 @@ func (s *GrpcServer) GetRegionByID(ctx context.Context, request *pdpb.GetRegionB
...
+		request, err := stream.Recv()
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		if clusterID := keypath.ClusterID(); request.GetHeader().GetClusterId() != clusterID {
```
rleungx, line 1662: Could we also run the same role validation as the unary path here, not just the cluster ID check? Otherwise a follower may serve QueryRegion locally even when follower handling was not requested.

client/client.go
```diff
@@ -671,6 +684,12 @@ func (c *client) UpdateOption(option DynamicOption, value any) error {
 			return errors.New("[pd] invalid value type for TSOClientRPCConcurrency option, it should be int")
 		}
 		c.option.setTSOClientRPCConcurrency(value)
+	case EnableQueryRegion:
+		enable, ok := value.(bool)
+		if !ok {
+			return errors.New("[pd] invalid value type for EnableQueryRegion option, it should be bool")
+		}
+		c.option.setEnableQueryRegion(enable)
```
rleungx, line 744: When EnableQueryRegion is turned off dynamically, the existing regionClient keeps its daemon and streams alive until Close. Should we close/release it on disable to actually stop the stream path?

client/pkg/connectionctx/manager.go
```diff
@@ -0,0 +1,160 @@
...
+func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url string, stream T, overwrite ...bool) {
+	c.Lock()
+	defer c.Unlock()
+	overwriteFlag := false
+	if len(overwrite) > 0 {
+		overwriteFlag = overwrite[0]
+	}
+	_, ok := c.connectionCtxs[url]
+	if !overwriteFlag && ok {
+		return
```
rleungx, line 65: If Store returns early because this URL already exists, the newly-created stream/context from the caller is left uncanceled. Could Store return whether it stored the context, or should callers cancel on the false path?

JmPotato:
This single PR has too many changes, temporarily closed. Will resubmit the PR after splitting it later.

Issue #10630, opened 2026-04-29
Support updating PD leader lease timeout online


JmPotato:
Part of #10653.

## Enhancement Task

Make the PD leader lease timeout configurable through the existing runtime config API.

Scope:
- Support `POST /pd/api/v1/config` with the top-level `lease` field.
- Persist the effective leader lease with the existing PD config blob.
- Make the next PD leader campaign use the persisted runtime value.
- Keep existing toml startup behavior unchanged.
- Do not implement in-place lease rotation; operators can use leader transfer or resign for immediate effect.

PR #10631, opened 2026-04-29
server, config: support updating leader lease online

JmPotato:

Issue Number: ref #10630

Production clusters can see PD leader lease keepalive p99 latency approach the
5s default lease timeout, leaving little safety margin before leader churn.
Operators need a way to tune the PD leader lease online between releases without
editing the local config file and restarting PD.

```commit-message
Support updating the PD leader lease timeout through POST /pd/api/v1/config with the top-level lease field.
Store the effective leader lease in PersistOptions and persist it with the existing config blob.
Before campaigning, load only the persisted leader lease so the next PD leader campaign uses the runtime lease value.
Keep existing toml startup behavior unchanged for `lease = 0`, cap the effective lease at 60 seconds, and reject invalid API values.
```

Implementation notes:

- `GET /pd/api/v1/config` returns the effective top-level `lease` value (the in-memory value of the queried member).
- `POST /pd/api/v1/config` accepts `{"lease": <seconds>}` and persists it through the existing config blob. The accepted online range is `1..60` seconds; `0`, negative, and `> 60` are rejected.
- The current leader's etcd lease TTL is fixed at grant time, so an online update takes effect on the **next** PD leader campaign. Use leader transfer or resign to apply it immediately.
- Before campaigning, the server loads only the persisted leader lease (a cheap election-path read). A **positive** persisted value is adopted; otherwise (absent, non-positive, or a load error) the current in-memory value is used, preserving the previous campaign behavior.
- The full config reload (run on becoming leader, and on followers when the leader changes) likewise adopts a positive persisted lease and otherwise keeps the in-memory value, so `GET /config` converges to the effective lease.
- Once a positive lease has been persisted it is authoritative: it overrides the local `lease` in the config file on reload/campaign. To change the lease on a running cluster, use the online API — editing the config file and restarting has **no effect** while a positive lease is persisted. This matches how PD already treats other persisted config.
- Older blobs with an absent or non-positive `lease` (e.g. clusters upgraded from a version without this feature) keep the startup toml value.
- File config `lease = 0` keeps the existing startup behavior and defaults to `5` through `AdjustInt64`. File config or persisted values above `60` are clamped to `60` with a warning, so existing deployments do not fail to start after upgrade while the effective leader lease stays within the supported range.
- The `60s` upper bound is a failover availability guard. Leader lease does not affect TSO correctness; TSO save-window protection is independent of this runtime lease value.
- Use the online lease API after all PD members are upgraded. In a mixed-version rolling upgrade, an old PD member may persist the config blob without the online lease value, reverting the setting. Re-apply the online lease after all PD members are upgraded.

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

- PR to update [`pingcap/docs`](https://github.com/pingcap/docs)/[`pingcap/docs-cn`](https://github.com/pingcap/docs-cn): TBD

Manual test

Unit / integration:

- `go test ./server/config -run 'TestReloadLeaderLease|TestValidateLeaderLease|TestAdjustClampsLeaderLeaseAboveMax|TestLoadPersistedLeaderLease' -count=1`
- `go test -tags without_dashboard ./tests/server/api -run 'TestLeaderLeaseConfigAPI|TestLeaderLeaseTakesEffectAfterTransfer|TestLeaderLeaseCampaignSurvivesLoadFailure' -count=1`
- `golangci-lint run ./server/config`
- `gofmt -l server/config/config.go server/config/config_test.go server/config/persist_options.go server/server.go tests/server/api/api_test.go`
- `git diff --check`

End-to-end (3-node PD cluster; the leadership lease TTL was read directly from etcd to confirm the lease each campaign actually used):

| Scenario | Expectation | Observed |
|---|---|---|
| Fresh cluster, file `lease = 7`, blob absent | campaign uses the local value | leadership TTL = `7`; all `GET /config` = `7` |
| `POST {"lease":15}` on the leader | in-memory + persisted update is immediate; current leader unchanged | `GET /config` = `15`, blob = `15`, current leadership TTL still `7` |
| Resign the leader after the update | next campaign adopts the persisted value | new leader leadership TTL = `15`; all nodes converge to `15` |
| Restart a node with file `lease = 30` while blob = `15` | persisted overrides the local file | node `GET /config` = `15`; after transfer its campaign leadership TTL = `15` (not `30`) |
| Persisted `lease = 0`, restart all with file `lease = 33` | non-positive persisted value ignored → local fallback | leadership TTL = `33` |
| `POST` `lease` = `0` / `-1` / `61` | rejected, value unchanged | HTTP `400`; `GET /config` unchanged |
| Persisted `lease = 200`, next campaign | clamped to the `60s` max | leadership TTL = `60`; clamp warning logged |

### Release note

```release-note
Support updating the PD leader lease timeout online through `POST /pd/api/v1/config` with the `lease` field. The new value takes effect on the next PD leader campaign; use leader transfer or resign to apply it immediately. The API accepts `1..60` seconds and rejects values outside that range. Once set online the lease is persisted and authoritative across restarts, so to change it on a running cluster use this API rather than editing the local config file. File config `lease = 0` keeps the existing default-to-5 startup behavior, while file config or persisted values above `60` are clamped to `60` with a warning. Use this API after all PD members are upgraded; during a mixed-version rolling upgrade, an old PD member may persist the config blob without the online lease value, reverting the setting. Re-apply the online lease after all PD members are upgraded.
```

## Summary by CodeRabbit

* **New Features**
  * Leader lease is now persisted, surfaced in the config API (GET/POST), and applied at runtime; server provides a validated API to update the lease.

* **Bug Fixes**
  * Reload now ignores non‑positive persisted lease values and preserves other persisted options when updating the lease.

* **Tests**
  * Added unit and end‑to‑end tests for API behavior, persistence, reload semantics, and validation.

### DIFF
--- a/server/api/config.go
+++ b/server/api/config.go
@@ -80,6 +80,7 @@ func (h *confHandler) GetConfig(w http.ResponseWriter, r *http.Request) {
 		mergedCfg := localCfg
 		mergedCfg.Replication = leaderCfg.Replication
 		mergedCfg.Schedule = leaderCfg.Schedule
+		mergedCfg.LeaderLease = leaderCfg.LeaderLease
 		h.rd.JSON(w, http.StatusOK, mergedCfg)
 		return
 	}
@@ -215,6 +216,8 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value any) er
 		return h.updateMicroserviceConfig(cfg, kp[len(kp)-1], value)
 	case "controller":
 		return h.updateControllerConfig(kp[len(kp)-1], value)
+	case "lease":
+		return h.updateLeaderLease(cfg, value)
 	}
 	return errors.Errorf("config prefix %s not found", kp[0])
 }
@@ -252,6 +255,20 @@ func (h *confHandler) updateMicroserviceConfig(config *config.Config, key string
 		err = h.svr.SetMicroserviceConfig(config.Microservice)
 	}
 	return err
+}
+
+func (h *confHandler) updateLeaderLease(config *config.Config, value any) error {
+	updated, found, err := jsonutil.AddKeyValue(config, "lease", value)
+	if err != nil {
+		return err
+	}
+	if !found {
+		return errors.New("config item lease not found")
+	}
+	if updated {
+		return h.svr.SetLeaderLease(config.LeaderLease)
+	}
+	return nil
 }
 
 func (h *confHandler) updateSchedule(config *config.Config, key string, value any) error {
--- a/server/config/persist_options.go
+++ b/server/config/persist_options.go
@@ -859,9 +859,44 @@ func (o *PersistOptions) Reload(storage endpoint.ConfigStorage) error {
 	return nil
 }
 
+// LoadPersistedLeaderLease returns the leader lease that should be used for the
+// next campaign. It reads only the persisted leader lease instead of reloading
+// the whole configuration, so it is cheap enough to run on the election path.
+// When the persisted blob carries no valid lease (e.g. a blob written before
+// online lease update was supported), the current in-memory value is returned,
+// preserving the previous campaign behavior. Storage and decode errors are
+// returned to the caller, which decides whether to fall back.
+func (o *PersistOptions) LoadPersistedLeaderLease(storage endpoint.ConfigStorage) (int64, error) {
+	failpoint.Inject("loadLeaderLeaseFail", func() {
+		failpoint.Return(int64(0), errors.New("failpoint: fail to load persisted leader lease"))
+	})
+	cfg := &persistedConfig{Config: &Config{}}
+	isExist, err := storage.LoadConfig(cfg)
+	if err != nil {
+		return 0, err
+	}
+	if isExist && cfg.hasValidLeaderLease() {
+		return cfg.LeaderLease, nil
+	}
+	return o.GetLeaderLease(), nil
+}
+
 // IsValidLeaderLease returns whether the given PD leader lease timeout is valid.
+// It only checks positivity, because it also gates whether a persisted lease
+// should be honored on reload.
 func IsValidLeaderLease(lease int64) bool {
 	return lease > 0
+}
+
+// ValidateLeaderLease validates a leader lease (in seconds) supplied through
+// the online update API. Keep the accepted range aligned with file config
+// behavior: reject explicit non-positive updates, but do not add an online-only
+// upper bound.
+func ValidateLeaderLease(lease int64) error {
+	if !IsValidLeaderLease(lease) {
+		return errors.Errorf("leader lease must be positive, got %d", lease)
+	}
+	return nil
 }
 
 func adjustScheduleCfg(scheduleCfg *sc.ScheduleConfig) {
--- a/server/config/persist_options.go
+++ b/server/config/persist_options.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 	"strconv"
 	"sync/atomic"
@@ -55,6 +56,7 @@ type PersistOptions struct {
 	keyspace        atomic.Value
 	microservice    atomic.Value
 	storeConfig     atomic.Value
+	leaderLease     atomic.Int64
 	clusterVersion  unsafe.Pointer
 }
 
@@ -68,6 +70,7 @@ func NewPersistOptions(cfg *Config) *PersistOptions {
 	o.labelProperty.Store(cfg.LabelProperty)
 	o.keyspace.Store(&cfg.Keyspace)
 	o.microservice.Store(&cfg.Microservice)
+	o.leaderLease.Store(cfg.LeaderLease)
 	// storeConfig will be fetched from TiKV later,
 	// set it to an empty config here first.
 	o.storeConfig.Store(&sc.StoreConfig{})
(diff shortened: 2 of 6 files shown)


server/server.go
```diff
@@ -1013,6 +1013,7 @@ func (s *Server) GetServiceMiddlewareConfig() *config.ServiceMiddlewareConfig {
 // GetConfig gets the config information.
 func (s *Server) GetConfig() *config.Config {
 	cfg := s.cfg.Clone()
+	cfg.LeaderLease = s.persistOptions.GetLeaderLease()
```
rleungx, line 1016: This keeps the leader response fresh, but follower `GET /config` still only merges schedule/replication from the leader. Can we also copy `LeaderLease` there so followers don't return the old lease after a POST?

server/config/persist_options.go
```diff
@@ -813,12 +820,42 @@ func (o *PersistOptions) Reload(storage endpoint.ConfigStorage) error {
 		o.labelProperty.Store(cfg.LabelProperty)
 		o.keyspace.Store(&cfg.Keyspace)
 		o.microservice.Store(&cfg.Microservice)
+		if IsValidLeaderLease(cfg.LeaderLease) {
```
rleungx: This still can't tell "missing lease field" from "lease was persisted". For old config blobs without `lease`, `Adjust` fills the default 5 and this can override a custom startup lease.

server/config/persist_options.go
```diff
@@ -813,12 +820,42 @@ func (o *PersistOptions) Reload(storage endpoint.ConfigStorage) error {
...
 	return nil
 }
 
+// ReloadLeaderLease reloads only the PD leader lease timeout from storage.
+func (o *PersistOptions) ReloadLeaderLease(storage endpoint.ConfigStorage) error {
+	cfg, isExist, err := loadPersistedConfig(storage)
+	if err != nil {
+		return err
+	}
+	if isExist && IsValidLeaderLease(cfg.LeaderLease) {
```
rleungx: Same here: before applying this value, we should first know the `lease` key was actually present in storage, not just filled by defaults.

server/server.go
JmPotato: Fixed in c334f2830 by copying LeaderLease from the leader config in the follower GET /config merge path.

server/config/persist_options.go
JmPotato: Fixed in c334f2830. persistedConfig now records whether the top-level lease key was actually present during JSON unmarshal, and Reload only applies a present positive lease.

server/config/persist_options.go
JmPotato: Fixed in c334f2830 with the same presence check in ReloadLeaderLease, so narrow reload keeps the startup lease when old config blobs do not contain lease.

YuhaoZhang00 (approved):

server/api/config.go
```diff
@@ -215,6 +216,8 @@ func (h *confHandler) updateConfig(cfg *config.Config, key string, value any) er
 		return h.updateMicroserviceConfig(cfg, kp[len(kp)-1], value)
 	case "controller":
 		return h.updateControllerConfig(kp[len(kp)-1], value)
+	case "lease":
```
rleungx, line 219: Please consider version-gating this API until all PD members understand the persisted lease. During rolling upgrade, an old PD can still persist `lease: 0` in the config blob, which makes the new code ignore the online value and fall back to local toml on the next campaign.

server/config/persist_options.go
```diff
@@ -813,12 +850,55 @@ func (o *PersistOptions) Reload(storage endpoint.ConfigStorage) error {
...
+// should be honored on reload.
+func IsValidLeaderLease(lease int64) bool {
+	return lease > 0
+}
+
+// ValidateLeaderLease validates a leader lease (in seconds) supplied through
+// the online update API. Keep the accepted range aligned with file config
+// behavior: reject explicit non-positive updates, but do not add an online-only
+// upper bound.
+func ValidateLeaderLease(lease int64) error {
```
rleungx, line 890: Could we add an upper bound here? Values larger than `math.MaxInt64 / int64(time.Second)` can be persisted successfully, but `time.Duration(lease) * time.Second` overflows later in election keepalive and may break the next campaign.

server/config/persist_options.go
```diff
@@ -813,12 +850,55 @@ func (o *PersistOptions) Reload(storage endpoint.ConfigStorage) error {
...
+// the whole configuration, so it is cheap enough to run on the election path.
+// When the persisted blob carries no valid lease (e.g. a blob written before
+// online lease update was supported), the current in-memory value is returned,
+// preserving the previous campaign behavior. Storage and decode errors are
+// returned to the caller, which decides whether to fall back.
+func (o *PersistOptions) LoadPersistedLeaderLease(storage endpoint.ConfigStorage) (int64, error) {
+	failpoint.Inject("loadLeaderLeaseFail", func() {
+		failpoint.Return(int64(0), errors.New("failpoint: fail to load persisted leader lease"))
+	})
+	cfg := &persistedConfig{Config: &Config{}}
```
rleungx: This still unmarshals the full persisted config on the election path. Could we decode only a small `{lease}` struct here? That avoids extra work and prevents unrelated config decode issues from changing campaign behavior.

PR #10636, opened 2026-04-30
election: Protect leader key cleanup from stale deletes

### LINKED ISSUE #10635: Prevent stale election cleanup from deleting a new leader key
## Problem

Election recovery can call `DeleteLeaderKey` after a participant observes that the persisted leader/primary is itself. The current delete removes the election key unconditionally. If a different participant has already written a newer leader/primary key, the stale cleanup can delete that newer key.

This can affect PD leader election and microservice primary election, including scheduling service primary discovery. Once the primary key is removed, PD watchers can clear the cached service primary even though service instances are still registered.

## Root cause

`pkg/election/leadership.go` deletes `leaderKey` without comparing the observed key revision or leader value. `pkg/member/member.go` and `pkg/member/participant.go` already read the leader/primary `mod_revision`, but the revision is not used when cleaning up the key.

## Expected behavior

Election cleanup should only delete the exact key version that was observed. If the key has already been replaced, cleanup should preserve the new key and return a conflict so the election loop can retry.

## Fix direction

- Add a revision-protected delete helper in the election layer.
- Use the observed `mod_revision` when member/participant recovery deletes its own stale key.
- Keep `DeleteLeaderKey` guarded by the current leader value for existing callers.
- Add a regression test proving an old revision cannot delete a newly campaigned leader key.

okJiang:

Issue Number: close #10635

This PR prevents stale election cleanup from deleting a newer leader/primary key:

- `DeleteLeaderKey` now deletes only when the stored value still matches the local leader value.
- `DeleteLeaderKeyByRevision` deletes only when the leader key still has the mod revision observed by the caller.
- PD member and microservice participant recovery pass the observed revision when deleting their own stale key.
- A regression test covers the case where an old revision must not delete a newly campaigned leader key.

### Check List

Tests

- Unit test

### Release note

```release-note
Fix stale election cleanup that could delete a newly written leader or primary key.
```

### Local validation

- `make failpoint-enable`
- `go test ./pkg/election -run 'TestLeadership|TestDeleteLeaderKeyDoesNotDeleteChangedLeader' -count=1`
- `go test ./pkg/member -run TestNonExisting -count=1`
- `(cd tests/integrations && go test ./mcs/tso -run 'TestTSOKeyspaceGroupManagerSuite/TestWatchFailed' -count=1 -v)`
- `make failpoint-disable`
- `make static`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added a revision-scoped leader-key deletion option to make leadership changes safer.

* **Bug Fixes**
  * Deletion now verifies stored leader value/revision before removing the key, preventing accidental removal of a new leader and enabling idempotent cleanup when the key is already absent.
  * Leader-election flows updated to use revision-guarded deletion during re-election.

* **Tests**
  * Added tests covering revision-guarded deletion and leader-change scenarios.

### DIFF
--- a/pkg/election/leadership.go
+++ b/pkg/election/leadership.go
@@ -250,11 +250,32 @@ func (ls *Leadership) leaderCmp() clientv3.Cmp {
 
 // DeleteLeaderKey deletes the corresponding leader from etcd by the leaderPath as the key.
 func (ls *Leadership) DeleteLeaderKey() error {
-	resp, err := kv.NewSlowLogTxn(ls.client).Then(clientv3.OpDelete(ls.leaderKey)).Commit()
+	return ls.removeLeaderKey(clientv3.Compare(clientv3.Value(ls.leaderKey), "=", ls.GetLeaderValue()))
+}
+
+// DeleteLeaderKeyByRevision deletes the leader key only when its mod revision
+// is the same as the observed revision.
+func (ls *Leadership) DeleteLeaderKeyByRevision(revision int64) error {
+	return ls.removeLeaderKey(clientv3.Compare(clientv3.ModRevision(ls.leaderKey), "=", revision))
+}
+
+func (ls *Leadership) removeLeaderKey(cmp clientv3.Cmp) error {
+	resp, err := kv.NewSlowLogTxn(ls.client).
+		If(cmp).
+		Then(clientv3.OpDelete(ls.leaderKey)).
+		Else(clientv3.OpGet(ls.leaderKey)).
+		Commit()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


lhy1024 (approved):

pkg/election/leadership.go
```diff
@@ -250,11 +250,32 @@ func (ls *Leadership) leaderCmp() clientv3.Cmp {
 
 // DeleteLeaderKey deletes the corresponding leader from etcd by the leaderPath as the key.
 func (ls *Leadership) DeleteLeaderKey() error {
```
rleungx: Do we still need it?

rleungx (approved):

PR #10639, opened 2026-05-05
pkg/unsaferecovery: optimize empty region overlap checks

### LINKED ISSUE #10638: Unsafe recovery create-empty-region plan generation can be O(holes * peer reports)
### What did you do?

Run online unsafe recovery (`/pd/api/v1/admin/unsafe/remove-failed-stores`) on a large cluster where the recovery process needs to enter the create-empty-region stage.

The current `generateCreateEmptyRegionPlan` implementation performs a paranoid overlap check for every detected empty range:

https://github.com/tikv/pd/blob/0fc8673576f31c5fe25c15514a3496f59ed7a3df/pkg/unsaferecovery/unsafe_recovery_controller.go#L1235-L1251

The check scans all peer reports in `peersMap` for each newly created empty region candidate:

```go
for _, peers := range peersMap {
    for _, peer := range peers {
        if !peer.isInitialized() {
            continue
        }
        if ... overlap ... {
            return false
        }
    }
}
```

This makes the create-empty-region plan generation cost roughly:

```text
O(number_of_holes * number_of_peer_reports)
```

For a large cluster, this can become very expensive. For example, with about 1.5M regions and 5K holes, the paranoid check can require billions of range checks. A local benchmark using the original package-level data structures (`regionTree`, `map[uint64][]*regionItem`, `PeerReport -> RegionLocalState -> Region`, and the original overlap condition) showed:

```text
1.5M regions, 1 hole, 1 peer report/region: ~435 ms
1.5M regions, 1 hole, 3 peer reports/region: ~1.05 s
```

Linearly extrapolated to 5K holes:

```text
1 peer report/region:  ~36 min
3 peer reports/region: ~1.5 h
```

A CPU profile captured during unsafe recovery also showed `generateCreateEmptyRegionPlan.func3` as the main PD unsafe-recovery business hotspot, with samples concentrated in:

```text
github.com/tikv/pd/pkg/unsaferecovery.(*Controller).generateCreateEmptyRegionPlan.func3
  github.com/tikv/pd/pkg/unsaferecovery.(*regionItem).isInitialized
  github.com/tikv/pd/pkg/unsaferecovery.(*regionItem).region
  bytes.Compare
```

There is also a broader impact from where this work runs:

* `HandleStoreHeartbeat` holds the unsafe recovery controller mutex while it calls `buildUpFromReports()` and `generatePlan()`:
  https://github.com/tikv/pd/blob/0fc8673576f31c5fe25c15514a3496f59ed7a3df/pkg/unsaferecovery/unsafe_recovery_controller.go#L330-L355
* `IsRunning()` documents that, while unsafe recovery is running, further unsafe recovery requests, schedulers, checkers, `AskSplit`, and `AskBatchSplit` requests are blocked:
  https://github.com/tikv/pd/blob/0fc8673576f31c5fe25c15514a3496f59ed7a3df/pkg/unsaferecovery/unsafe_recovery_controller.go#L183-L188
* Stage transitions set the scheduling allowance status with the `"online-unsafe-recovery"` source:
  https://github.com/tikv/pd/blob/0fc8673576f31c5fe25c15514a3496f59ed7a3df/pkg/unsaferecovery/unsafe_recovery_controller.go#L506-L510

As a result, this is not just CPU overhead. A long create-empty-region calculation can also keep the unsafe recovery controller lock held for a long time, delaying subsequent unsafe recovery store-heartbeat handling/status checks. More importantly, extending the recovery from minutes to tens of minutes or hours extends the period during which scheduling/checker/split-related work is halted by unsafe recovery.

The create-empty-region plan generation should scale well enough for large clusters and finish within the configured unsafe recovery timeout.

The paranoid overlap check should keep the same safety semantics, but avoid scanning all peer reports for every hole.

One possible approach:

1. Flatten initialized peer reports from `peersMap` once.
2. Cache each peer's region/start/end so the check does not repeatedly call `peer.region()`.
3. Build a range index sorted by `StartKey`.
4. Since empty ranges are generated in key order, use a linear two-pointer scan, prefix max-end index, or a region-tree/range-tree lookup to find only possible overlaps.
5. Treat ranges as half-open intervals: `[start, end)`, with empty `end` as positive infinity.

This could reduce the check from:

```text
O(H * N)
```

to roughly:

```text
O(N log N + H log N)
```

or even:

```text
O(N log N + N + H)
```

depending on the chosen index.

### What did you see instead?

The current implementation can spend a long time in `generateCreateEmptyRegionPlan` before dispatching the generated recovery plan.

If this calculation takes longer than the user-specified unsafe recovery timeout (default 600s), the next timeout check can move unsafe recovery into the exit-force-leader path before all create-empty-region plans are dispatched/executed.

This makes large-cluster unsafe recovery sensitive to the number of empty ranges, not just the number of regions. It can also prolong the period where the unsafe recovery controller lock is held and where scheduling/checker/split-related paths are halted due to unsafe recovery being in progress.

The issue is based on the current code path in `pkg/unsaferecovery/unsafe_recovery_controller.go`. The observed profile came from a production PD during unsafe recovery; exact `pd-server -V` was not captured in this issue.

Connor1996:

Issue Number: Close #10638

Unsafe recovery creates empty regions to fill range holes. The paranoid overlap check currently scans every collected peer for each generated empty region, which can make plan generation O(number of holes * number of peer reports). On large clusters this work happens while the unsafe recovery controller lock is held, delaying schedulers/checkers and related unsafe-recovery operations.

PD also re-dispatches a store recovery plan after a fixed 40s execution window. When TiKV spends longer executing create-empty-region plans, PD may repeatedly log `unsafe recovery store recovery plan execution timeout, retry` and send duplicate plans.

```commit-message
pkg/unsaferecovery: optimize empty region overlap checks
```

Collect newly created empty regions while scanning the newest region tree. These empty regions are generated in key order and do not overlap with each other, so they can be used as a small sorted index. Then scan collected peer reports once and binary-search the empty-region index to detect any overlap.

This changes the paranoid overlap check from O(empty regions * peer reports) to O(newest regions + empty regions + peer reports * log(empty regions)). When there is no generated empty region, the extra check returns immediately. For multiple peer reports of the same region, duplicate same-version same-range reports reuse the first no-overlap result and skip another binary search.

The check uses half-open region range semantics and keeps the existing error path when an overlap is found.

### Before/after comparison

For a recovery round with about 1.5M existing regions and 5k generated empty regions:

| Item | Before this PR | After this PR |
| --- | --- | --- |
| Paranoid check algorithm | For every generated empty region, scan every collected initialized peer report. | Build an index from generated empty regions, then scan peer reports once and binary-search the empty-region index. |
| Complexity | O(empty regions * peer reports). | O(newest regions + empty regions + peer reports * log(empty regions)). |
| Work for 1.5M regions / 5k holes | Estimated from the original data-structure benchmark: about 36 minutes to more than 1 hour locally, and could become several hours in production while holding the unsafe recovery controller lock. | Expected paranoid-check cost is sub-second at this scale; remaining wall time mostly depends on the existing newest-region tree scan and ID allocation for generated regions. |
| Lock impact | The O(empty regions * peer reports) scan runs while the unsafe recovery controller lock is held, so large recoveries can delay schedulers, checkers, and related unsafe-recovery handling. | The lock is still held during plan generation, but the expensive paranoid-check part is reduced from a several-hour worst-case risk to sub-second scale in the estimate above. |

The exact wall time still depends on key length, CPU, report fanout, the existing newest-region tree scan, and ID allocation for generated regions, so the numbers above are rough estimates rather than a strict production SLA.

This PR also extends the default plan execution timeout from 40s to 60s, and adds request-level options for large unsafe recovery operations:

```json
{
  "stores": [1, 2, 3],
  "timeout": 600,
  "plan-execution-timeout": 600,
  "disable-paranoid-check": true
}
```

The same options are exposed in `pd-ctl`:

```bash
pd-ctl unsafe remove-failed-stores 1,2,3 --timeout 600 --plan-execution-timeout 600 --disable-paranoid-check
```

- `plan-execution-timeout` controls how long PD waits for a store to execute a dispatched recovery plan before re-dispatching it. The default is 60s.
- `disable-paranoid-check` skips the empty-region overlap paranoid check. The default remains enabled.

When either option is set, the recovery status output records it, for example `plan execution timeout 10m0s` and `paranoid check disabled`.

### Check List

Tests

- Unit test

### Release note

```release-note
Optimize unsafe recovery empty-region plan generation when many regions and gaps exist
```

## Summary by CodeRabbit

* **New Features**
  * Configurable plan execution timeout for unsafe recovery operations.
  * Option to disable the paranoid empty-region overlap check.
  * API and CLI support to pass these options for removal operations.

* **Improvements**
  * Batched empty-region creation with overlap guarding; tracks created and affected regions.
  * Status output now reports paranoid-check state and configured plan execution timeout.

* **Tests**
  * Added validation and parsing tests for timeouts and paranoid-check flags; CLI validation tests.

### DIFF
--- a/pkg/unsaferecovery/unsafe_recovery_controller.go
+++ b/pkg/unsaferecovery/unsafe_recovery_controller.go
@@ -44,7 +44,7 @@ import (
 type stage int
 
 const (
-	storeRequestInterval = time.Second * 40
+	defaultPlanExecutionTimeout = time.Second * 60
 )
 
 // Stage transition graph: for more details, please check `Controller.HandleStoreHeartbeat()`
@@ -126,6 +126,12 @@ type Controller struct {
 	failedStores map[uint64]struct{}
 	timeout      time.Time
 	autoDetect   bool
+	// planExecutionTimeout is the duration PD waits for a store to execute the recovery plan
+	// before dispatching the same plan again.
+	planExecutionTimeout time.Duration
+	// disableParanoidCheck skips the sanity check that newly created empty regions
+	// don't overlap with any collected peer report.
+	disableParanoidCheck bool
 
 	// collected reports from store, if not reported yet, it would be nil
 	storeReports      map[uint64]*pdpb.StoreReport
@@ -180,6 +186,8 @@ func (u *Controller) reset() {
 	u.newlyCreatedRegions = make(map[uint64]struct{}, 0)
 	u.orphanedPeers = make(map[uint64][]*metapb.Peer)
 	u.err = nil
+	u.planExecutionTimeout = defaultPlanExecutionTimeout
+	u.disableParanoidCheck = false
 }
 
 // IsRunning returns whether there is ongoing unsafe recovery process. If yes, further unsafe
@@ -194,8 +202,24 @@ func isRunning(s stage) bool {
 	return s != Idle && s != Finished && s != Failed
 }
 
+// RemoveFailedStoresOptions controls optional unsafe recovery behavior.
+type RemoveFailedStoresOptions struct {
+	PlanExecutionTimeout time.Duration
+	DisableParanoidCheck bool
+}
+
 // RemoveFailedStores removes Failed stores from the cluster.
 func (u *Controller) RemoveFailedStores(failedStores map[uint64]struct{}, timeout uint64, autoDetect bool) error {
+	return u.RemoveFailedStoresWithOptions(failedStores, timeout, autoDetect, RemoveFailedStoresOptions{})
+}
+
+// RemoveFailedStoresWithOptions removes Failed stores from the cluster with options.
+func (u *Controller) RemoveFailedStoresWithOptions(
+	failedStores map[uint64]struct{},
+	timeout uint64,
+	autoDetect bool,
+	options RemoveFailedStoresOptions,
+) error {
 	u.Lock()
 	defer u.Unlock()
 
@@ -239,6 +263,10 @@ func (u *Controller) RemoveFailedStores(failedStores map[uint64]struct{}, timeou
 	u.timeout = time.Now().Add(time.Duration(timeout) * time.Second)
 	u.failedStores = failedStores
 	u.autoDetect = autoDetect
+	if options.PlanExecutionTimeout > 0 {
+		u.planExecutionTimeout = options.PlanExecutionTimeout
+	}
+	u.disableParanoidCheck = options.DisableParanoidCheck
 	u.changeStage(CollectReport)
 	return nil
 }
@@ -321,7 +349,7 @@ func (u *Controller) handleErr(err error) bool {
 	// blocks reads and writes.
 	u.storePlanExpires = make(map[uint64]time.Time)
 	u.storeRecoveryPlans = make(map[uint64]*pdpb.RecoveryPlan)
-	u.timeout = time.Now().Add(storeRequestInterval * 2)
+	u.timeout = time.Now().Add(u.planExecutionTimeout * 2)
 	// empty recovery plan would trigger exit force leader
 	u.changeStage(ExitForceLeader)
 	return false
@@ -463,7 +491,7 @@ func (u *Controller) dispatchPlan(heartbeat *pdpb.StoreHeartbeatRequest, resp *p
 		// Dispatch the recovery plan to the store, and the plan may be empty.
 		resp.RecoveryPlan = u.getRecoveryPlan(storeID)
 		resp.RecoveryPlan.Step = u.step
-		u.storePlanExpires[storeID] = now.Add(storeRequestInterval)
+		u.storePlanExpires[storeID] = now.Add(u.planExecutionTimeout)
 	}
 }
 
@@ -527,6 +555,12 @@ func (u *Controller) changeStage(stage stage) {
 				ids = append(ids, strconv.FormatUint(store, 10))
 			}
 			output.Details = append(output.Details, fmt.Sprintf("Failed stores %s", strings.Join(ids, ", ")))
+		}
+		if u.disableParanoidCheck {
+			output.Details = append(output.Details, "paranoid check disabled")
+		}
+		if u.planExecutionTimeout != defaultPlanExecutionTimeout {
+			output.Details = append(output.Details, fmt.Sprintf("plan execution timeout %s", u.planExecutionTimeout))
 		}
 
 	case TombstoneTiFlashLearner:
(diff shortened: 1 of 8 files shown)


pkg/unsaferecovery/unsafe_recovery_controller.go
```diff
@@ -321,7 +349,7 @@ func (u *Controller) handleErr(err error) bool {
 	// blocks reads and writes.
 	u.storePlanExpires = make(map[uint64]time.Time)
 	u.storeRecoveryPlans = make(map[uint64]*pdpb.RecoveryPlan)
-	u.timeout = time.Now().Add(storeRequestInterval * 2)
+	u.timeout = time.Now().Add(u.planExecutionTimeout * 2)
```
lhy1024, line 352: Can we cap `plan-execution-timeout` at `MaxInt64/2` duration instead of `MaxInt64`? The controller later does `u.planExecutionTimeout * 2` in the error cleanup path, so values larger than half of max duration can overflow and make the ExitForceLeader timeout incorrect. The API/pd-ctl validators and boundary tests should be adjusted together.

Connor1996: Capped both the HTTP and `pd-ctl` validators at `MaxInt64/2` duration so the later `u.planExecutionTimeout * 2` path cannot overflow, and updated the boundary coverage alongside the validator changes.

Connor1996:
Addressed the remaining unsafe-recovery review follow-up in `728009d10`: tail/full-range empty-region creates now go through the same create-plan path before the paranoid overlap check, so they are checked consistently and still fall back away from TiFlash stores when needed.

Connor1996:
Addressed the latest API validation nit: `remove-failed-stores` now rejects requests that provide both the dashed and snake_case forms of `plan-execution-timeout` or `disable-paranoid-check`, instead of silently letting one overwrite the other. Added unit coverage in `server/api` and request-level coverage in `tests/server/api` for the duplicate-key `400` path.

Connor1996:
Pushed 57dfc7f to address the latest API review nits: getAliasedOption now treats the first key as canonical instead of passing the canonical name twice, and TestParseDisableParanoidCheck now covers explicit true/false happy paths.

Connor1996:
Checked the latest review feedback against current HEAD `57dfc7f`.

- The two `server/api` nitpicks are already covered by the current helper/test updates in this branch.
- The earlier tail/full-range paranoid-check concern is already handled by the current `appendCreatePlan` flow, and covered by `TestCreateEmptyRegionTailParanoidCheck` plus `TestCreateEmptyRegionFullRangeParanoidCheck`.

No additional code change was needed, so I did not add a follow-up commit.

Connor1996:
Addressed the still-valid review item on `timeout` parsing in `server/api`: the handler now rejects non-integer, non-positive, and oversized `timeout` values with `400` instead of casting directly to `uint64`, and I added unit plus API coverage for those invalid inputs. The two smaller parsing-helper nitpicks from the same review were already present on the PR head before this follow-up.

Connor1996:
Checked the latest review feedback and current HEAD already contains the timeout validation / alias parsing fixes under review, so I did not add another code change here.

I reran the failed GitHub Actions jobs once to classify the remaining red CI. The failures are still in the unrelated `chunks (9, Microservice Integration(!TSO))` resource-manager suite, and the symptom changed between runs:
- First run: `tests/integrations/mcs/resourcemanager/resource_manager_test.go:1832` `TestCheckBackgroundJobs` -> `Condition never satisfied`
- Rerun: `tests/integrations/mcs/resourcemanager/resource_manager_test.go:1587` `standalone-resource-manager-with-client-discovery/TestResourceManagerClientFailover` -> `ErrGroupNotExists` for `failover_test`

`report-coverage` is only cascading because `covprofile_9` is missing after `chunks (9)` fails.

lhy1024 (approved):

v01dstar (approved):

nolouch (approved):

PR #10645, opened 2026-05-07
mcs/scheduling: clean primary resources on exit

### LINKED ISSUE #10644: mcs/scheduling: clean primary-scoped resources when scheduling primary exits
### What did you do?

Ran the independent scheduling service and triggered scheduling primary exit or re-campaign.

In the current implementation, `startCluster` creates primary-scoped resources such as the in-memory cluster state, heartbeat streams, and metadata/config watchers on the server-level context. When the scheduling primary steps down, `stopCluster` stops the background jobs but does not fully release these resources.

This can also leave partially initialized resources behind when `startCluster` fails after creating some of them.

When the scheduling primary exits, all primary-scoped state should be cleaned up immediately:
- close heartbeat stream workers
- stop and release watchers
- clear the in-memory cluster/storage references
- roll back partially initialized resources if primary startup fails

### What did you see instead?

Some scheduling primary state can survive after primary exit until the whole server process stops, including:
- `hbStreams` background goroutines
- watcher instances and their retained references
- in-memory `basicCluster` / storage references

This makes stale scheduling state linger across primary changes and can lead to unnecessary memory retention.

master

bufferflies:

Issue Number: Close #10644

```commit-message
bind scheduling primary resources to the primary campaign context so they stop when
the primary exits. add cleanup for heartbeat streams, watchers, and in-memory
cluster references, and roll back partially initialized resources on startup
failure.
```

### Check List

Tests

- Unit test

1. create 30w regions  
2.  transfer scheduler primary every 10 min utils 1 hour
image

### Release note

```release-note
Fix stale in-memory scheduling state retention after scheduling primary exits.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale suspect ranges from persisting after watcher shutdown.
  * Heartbeat endpoints now return "not bootstrapped" when required stream/monitoring components or cluster state are unavailable.

* **Improvements**
  * Refined cluster start/stop flow with explicit, idempotent cleanup to avoid partial state and ensure watchers/streams are closed.
  * API handlers now validate cluster readiness and return an error when uninitialized.
  * Affinity manager starts availability checks after initialization.

* **Tests**
  * Added tests validating cleanup idempotency and safe heartbeat behavior during teardown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10645)


pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,104 @@ func (s *Server) startServer() (err error) {
...
 	}
-	// Start the affinity watcher after the cluster is created.
-	s.affinityWatcher, err = affinity.NewWatcher(s.Context(), s.GetClient(), cluster.GetAffinityManager())
+	affinityWatcher, err := affinity.NewWatcher(ctx, s.GetClient(), cluster.GetAffinityManager())
 	if err != nil {
+		ruleWatcher.Close()
+		hbStreams.Close()
+		configWatcher.Close()
+		metaWatcher.Close()
 		return err
```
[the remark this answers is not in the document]
bufferflies: done

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,110 @@ func (s *Server) startServer() (err error) {
...
+func (s *Server) cleanupClusterResources() {
+	s.stopWatcher()
+	if s.hbStreams != nil {
+		s.hbStreams.Close()
+		s.hbStreams = nil
 	}
+	s.cluster.Store((*Cluster)(nil))
+	s.basicCluster = nil
+	s.storage = nil
 }
```
lhy1024: This cleanup order creates a real reproducible panic window: hbStreams can already be closed and set to nil while cluster is still non-nil. In that state, RegionHeartbeat passes the s.GetCluster() check and then calls s.hbStreams.BindStream(...), which panics with a nil pointer.

Please clear cluster first so handlers observe the service as unavailable before hbStreams, watchers, basicCluster, and storage are torn down. It would also be safer to add nil guards in RegionHeartbeat, StoreHeartbeat, and APIs that directly call GetBasicCluster() to avoid panics from in-flight requests during stepdown.

lhy1024: I verified this with a minimal unit test against the current PR:
```
func TestRegionHeartbeatDuringCleanupDoesNotPanic(t *testing.T) {
      re := require.New(t)

      basicCluster := core.NewBasicCluster()
      basicCluster.PutStore(core.NewStoreInfo(&metapb.Store{Id: 1, Address: "store-1"}))

      s := &Server{}
      s.cluster.Store(&Cluster{BasicCluster: basicCluster})
      s.hbStreams = nil // cleanupClusterResources has nilled hbStreams before clearing cluster.

      stream := &regionHeartbeatStream{
          ctx: context.Background(),
          requests: []*schedulingpb.RegionHeartbeatRequest{
              {
                  Region: &metapb.Region{
                      Id:    1,
                      Peers: []*metapb.Peer{{Id: 1, StoreId: 1}},
                  },
                  Leader: &metapb.Peer{Id: 1, StoreId: 1},
              },
          },
      }

      re.NotPanics(func() {
          _ = (&Service{Server: s}).RegionHeartbeat(stream)
      })
  }
```

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,110 @@ func (s *Server) startServer() (err error) {
...
+		configWatcher.Close()
+		metaWatcher.Close()
 		return err
 	}
-	s.cluster.Store(cluster)
-	// Inject the cluster components into the config watcher after the scheduler controller is created.
-	s.configWatcher.SetSchedulersController(cluster.GetCoordinator().GetSchedulersController())
-	// Start the rule watcher after the cluster is created.
-	s.ruleWatcher, err = rule.NewWatcher(s.Context(), s.GetClient(), s.storage,
+	defer func() {
```
rleungx, line 528: StopBackgroundJobs returns immediately when StartBackgroundJobs has not run, so this rollback does not cancel the cluster context on partial startup failure. NewCluster already starts context-bound goroutines such as the labeler GC / affinity check loop; please use a cleanup path that always cancels the cluster context.

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,110 @@ func (s *Server) startServer() (err error) {
...
+	}
+}
+
+func (s *Server) cleanupClusterResources() {
+	s.stopWatcher()
+	if s.hbStreams != nil {
+		s.hbStreams.Close()
+		s.hbStreams = nil
 	}
+	s.cluster.Store((*Cluster)(nil))
```
rleungx: This clears the public cluster/basicCluster references while the participant can still report serving until Resign runs. During that window redirected HTTP handlers may handle locally and dereference a nil BasicCluster, and gRPC may still observe a stopping cluster. Please unpublish serving state or these refs before resource teardown, or move cleanup after resigning.

pkg/mcs/scheduling/server/server.go
bufferflies: yes, fixed

pkg/mcs/scheduling/server/server.go
bufferflies: Yes, fixed it by manually canceling if the stop failed.

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,114 @@ func (s *Server) startServer() (err error) {
...
 	cluster.StartBackgroundJobs()
+	cluster = nil // defer cleanup no longer needed
 	return nil
 }
 
 func (s *Server) stopCluster() {
 	cluster := s.GetCluster()
 	if cluster != nil {
-		s.cluster.Store((*Cluster)(nil))
 		cluster.StopBackgroundJobs()
```
lhy1024, line 590: This changes the shutdown visibility order in a risky way. The old cluster remains published via `s.GetCluster()` while `StopBackgroundJobs()` is stopping the coordinator/runners. During this window, RPC/API handlers can still enter the old cluster and operate on resources that are already stopping. The previous code unpublished the cluster before stopping watchers. Could we first atomically unpublish the serving resources, then stop the captured old cluster/resources?

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +506,114 @@ func (s *Server) startServer() (err error) {
...
+	affinityWatcher, err := affinity.NewWatcher(ctx, s.GetClient(), cluster.GetAffinityManager())
 	if err != nil {
+		ruleWatcher.Close()
+		hbStreams.Close()
+		configWatcher.Close()
+		metaWatcher.Close()
 		return err
 	}
+
+	s.basicCluster = basicCluster
```
lhy1024: These fields are now part of the primary lifecycle and are reassigned/nilled on primary transitions, but RPC/API handlers read them concurrently  (`GetBasicCluster`, `RegionHeartbeat`, `StoreHeartbeat`). Since `basicCluster`, `hbStreams`, `metaWatcher`, etc. are plain pointer fields, this introduces data races and inconsistent snapshots. Please consider publishing all per-primary resources as one immutable resource bundle via `atomic.Value` or protecting both reads and writes with a mutex.

pkg/mcs/scheduling/server/apis/v1/api.go
```diff
@@ -1633,6 +1654,15 @@ func getAffinityManager(c *gin.Context) (*affinity.Manager, bool) {
 	return manager, true
 }
 
+func getBasicCluster(c *gin.Context, svr *scheserver.Server) (*pdcore.BasicCluster, bool) {
+	basicCluster := svr.GetBasicCluster()
```
lhy1024, line 1664: Checking only `s.basicCluster != nil` is not enough after this PR. During cleanup, `cluster` may already be unpublished while `basicCluster` is still non-nil, so `/stores` and `/regions` can return stale cache from a server that is no longer serving as primary. This should probably derive the basic cluster from the same serving snapshot as `GetCluster()`, or at least check `GetCluster()` first

pkg/mcs/scheduling/server/server_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestRegionHeartbeatDuringCleanupDoesNotPanic(t *testing.T) {
+	re := require.New(t)
+
+	basicCluster := core.NewBasicCluster()
+	basicCluster.PutStore(core.NewStoreInfo(&metapb.Store{Id: 1, Address: "store-1"}))
+
+	s := &Server{}
+	s.cluster.Store(&Cluster{BasicCluster: basicCluster})
+	s.hbStreams = nil // cleanupClusterResources has nilled hbStreams before clearing cluster.
```
lhy1024: This test state does not match the current cleanup order. `cleanupClusterResources()` clears `cluster` before niling `hbStreams`, so `cluster != nil && hbStreams == nil` is not the real shutdown window. The real risky window is in `stopCluster()`: `cluster` is still published while `StopBackgroundJobs()` is running. Could we add a test that exercises the actual `stopCluster()` order, preferably with a concurrent heartbeat/API read?

pkg/schedule/affinity/manager.go
```diff
@@ -132,7 +132,6 @@ func (m *Manager) initialize() error {
 		log.Error("failed to rebuild group-label mapping", zap.Error(err))
 		return err
 	}
-
```
lhy1024, line 135: unnecessary empty line remove

pkg/mcs/scheduling/server/apis/v1/api.go
lhy1024: My concern is the inconsistent visibility during cleanup. There is a window after `cluster` has been set to nil but before `basicCluster` is set to nil,where new requests can still read the old cache through `GetBasicCluster()`. For example, `/stores` and `/regions` use `GetBasicCluster()` directly instead of deriving it from `GetCluster()`.

This is not about the old BasicCluster pointer becoming invalid; the pointer is still usable. The issue is that the server state exposed to handlers is inconsistent: `GetCluster()` already says there is no serving cluster, but some APIs can still bypass that and return stale store/region cache. Could we make these APIs derive BasicCluster from `GetCluster().GetBasicCluster()`, or publish/cleanup the primary resources as a single atomic snapshot?

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -138,7 +137,7 @@ func (s *Service) RegionHeartbeat(stream schedulingpb.Scheduling_RegionHeartbeat
 		}
 
 		c := s.GetCluster()
-		if c == nil {
+		if c == nil || s.hbStreams == nil {
```
lhy1024: `basicCluster` is now derived from `GetCluster()`, which fixes the stale cache visibility issue for `/stores` and `/regions`. But I think the same lifecycle problem still exists for `hbStreams` and `metaWatcher`.

Both fields are plain pointers on `Server`, while request handlers read them concurrently:
  - `RegionHeartbeat` checks `s.hbStreams == nil` and later calls `s.hbStreams.BindStream(...)`.
  - `StoreHeartbeat` checks `s.metaWatcher == nil` and later calls `s.metaWatcher.GetStoreWatcher().ForceLoad()`.

During cleanup, these fields can be nilled between the nil check and the later method call, so this can race and even panic. It can also pair a `Cluster` snapshot from `GetCluster()` with a different/nil per-primary resource. Could we either capture these resources from the same atomic per-primary snapshot as `cluster`, or move the needed resource references under `Cluster` and access them through the captured `c`?

pkg/mcs/scheduling/server/server_test.go
lhy1024: This thread is outdated, but I don't think the underlying test concern is fully resolved yet. The current test still manually constructs `cluster != nil &&  hbStreams == nil` instead of exercising the actual `stopCluster()` order. The real window is still that `stopCluster()` keeps the old cluster published while `StopBackgroundJobs()` is running.

Also, after removing `Server.basicCluster`, this test file currently does not compile because `TestCleanupClusterResources` still references `basicCluster` in the `Server` literal and assertion.

pkg/mcs/scheduling/server/server.go
bufferflies: Yes, remove the basicCluster from the service struct; all callers should get the basic cluster from the Cluster field.

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -138,7 +138,8 @@ func (s *Service) RegionHeartbeat(stream schedulingpb.Scheduling_RegionHeartbeat
 		}
 
 		c := s.GetCluster()
-		if c == nil {
+		streams := s.hbStreams
```
rleungx: 这里读成局部变量可以避免后面二次读到 nil，但 `s.hbStreams` 仍然会和 cleanup 里的 close/set nil 并发读写。建议把 cluster/hbStreams/metaWatcher 放进同一个 atomic/mutex 保护的 primary runtime snapshot，RPC 只 load 一次 snapshot。

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -244,13 +245,14 @@ func (s *Service) RegionBuckets(stream schedulingpb.Scheduling_RegionBucketsServ
 // StoreHeartbeat implements gRPC SchedulingServer.
 func (s *Service) StoreHeartbeat(_ context.Context, request *schedulingpb.StoreHeartbeatRequest) (*schedulingpb.StoreHeartbeatResponse, error) {
 	c := s.GetCluster()
-	if c == nil {
+	metaWatcher := s.metaWatcher
```
rleungx: 这里也有同样的并发问题：RPC 读 `s.metaWatcher` 时，primary exit 可能正在 cleanup 里 close 并置 nil。建议和 cluster 一起通过同一个 runtime snapshot 发布/撤下。

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +508,112 @@ func (s *Server) startServer() (err error) {
...
 }
 
 func (s *Server) stopCluster() {
 	cluster := s.GetCluster()
 	if cluster != nil {
-		s.cluster.Store((*Cluster)(nil))
 		cluster.StopBackgroundJobs()
 	}
-	s.stopWatcher()
+	s.cleanupClusterResources()
```
rleungx: 这里 cleanup 发生在 `StopBackgroundJobs` 之后，teardown 窗口里新 RPC 仍可能拿到正在停止的 cluster。建议先原子撤下对外可见的 runtime，再 stop/close 内部资源。

pkg/mcs/scheduling/server/server.go
lhy1024: Thanks for moving `hbStreams` and watchers into `Cluster`; that fixes the separately mutable `Server` field issue.

I think the shutdown visibility issue in this thread is still valid, though. `stopCluster()` still calls `cluster.StopBackgroundJobs()` before  `cleanupClusterResources(cluster)`, and `cleanupClusterResources` is where `s.cluster.Store(nil)` happens. So while `StopBackgroundJobs()` is stopping coordinator/runners and canceling the cluster context, the old cluster is still visible through `GetCluster()` and handlers can still enter it.

Could we unpublish the cluster first, then stop and clean the captured old cluster? For example: capture `old := s.GetCluster()`, store nil into `s.cluster`, then call `old.StopBackgroundJobs()` and `old.cleanupRuntimeResources()`.

lhy1024: I added a regression test to verify this window.
```
  func TestStopClusterUnpublishesClusterBeforeStoppingBackgroundJobs(t *testing.T) {
        re := require.New(t)
        s, cluster := newServerWithClusterForStopTest(t)

        cluster.heartbeatRunner = syncRunner
        cluster.miscRunner = syncRunner
        cluster.logRunner = syncRunner
        cluster.running.Store(true)
        // Keep StopBackgroundJobs blocked so the test can observe the shutdown window.
        cluster.wg.Add(1)

        done := make(chan struct{})
        released := false
        defer func() {
                if !released {
                        cluster.wg.Done()
                }
                select {
                case <-done:
                case <-time.After(5 * time.Second):
                        t.Log("stopCluster did not return after releasing the test gate")
                }
        }()

        go func() {
                s.stopCluster()
                close(done)
        }()

        re.Eventually(func() bool {
                return !cluster.IsBackgroundJobsRunning()
        }, 5*time.Second, 10*time.Millisecond)
        select {
        case <-done:
                t.Fatal("stopCluster returned before the test observed the shutdown window")
        default:
        }

        if got := s.GetCluster(); got != nil {
                t.Errorf("expected stopCluster to make cluster invisible before waiting for background jobs to stop, got %p", got)
        }

        cluster.wg.Done()
        released = true
        select {
        case <-done:
        case <-time.After(5 * time.Second):
                t.Fatal("stopCluster did not return")
        }
  }
```
```
--- FAIL: TestStopClusterUnpublishesClusterBeforeStoppingBackgroundJobs
      server_test.go:100: expected stopCluster to make cluster invisible before waiting for background jobs to stop, got 0xc0002dedc0
```

bufferflies: good catch, fixed

pkg/mcs/scheduling/server/server.go
lhy1024: This still does not fully address @rleungx's comment, and the latest defer also introduces a successful-path cleanup issue.

`StopBackgroundJobs()` still returns immediately when `running=false`, so if `NewCluster()` succeeds but a later watcher creation fails, the cluster context is still not cancelled.

Also, the successful path only sets `cluster = nil`, while `hbStreams` / watchers remain non-nil. The defer will close the runtime resources that were just transferred to `Cluster` before `startCluster()` returns successfully.

Could we either disarm those local variables after ownership is transferred, or centralize cleanup in a method that always cancels the cluster context?

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -653,7 +729,7 @@ func (c *Cluster) StartBackgroundJobs() {
 	c.running.Store(true)
 }
 
-// StopBackgroundJobs stops background jobs.
+// StopBackgroundJobs stops background jobs, these jobs is created by NewCluster.
 func (c *Cluster) StopBackgroundJobs() {
 	if !c.running.Load() {
```
rleungx: On startup rollback, `NewCluster` may have already started goroutines while `StartBackgroundJobs` was never called, so `running` is false and this returns without canceling `c.ctx`. Please add a cleanup path that cancels the cluster context even before background jobs start.

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +500,95 @@ func (s *Server) startServer() (err error) {
...
-	s.affinityWatcher, err = affinity.NewWatcher(s.Context(), s.GetClient(), cluster.GetAffinityManager())
+	affinityWatcher, err = affinity.NewWatcher(ctx, s.GetClient(), cluster.GetAffinityManager())
 	if err != nil {
 		return err
 	}
+
+	cluster.SetRuntimeResources(metaWatcher, configWatcher, ruleWatcher, affinityWatcher)
+	s.cluster.Store(cluster)
 	cluster.StartBackgroundJobs()
+	cluster = nil // defer cleanup no longer needed
```
rleungx: This only skips `cluster.StopBackgroundJobs`; the defer will still close `hbStreams` and all watchers after a successful start because those locals remain non-nil. Please run the defer only on error, or nil out the resources after ownership is transferred to `cluster`.

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +500,95 @@ func (s *Server) startServer() (err error) {
...
+		hbStreams       *hbstream.HeartbeatStreams
+		configWatcher   *config.Watcher
+		metaWatcher     *meta.Watcher
+		ruleWatcher     *rule.Watcher
+		affinityWatcher *affinity.Watcher
+		err             error
+	)
+	metaWatcher, configWatcher, err = s.startMetaConfWatcher(ctx, basicCluster, storage)
 	if err != nil {
+		configWatcher.Close()
```
rleungx: `startMetaConfWatcher` can return nil watchers with an error, so this can panic and hide the real startup error. Please remove these closes here or guard them; the helper already cleans up partial watcher state.

pkg/mcs/scheduling/server/rule/watcher.go
```diff
@@ -280,5 +280,8 @@ func (rw *Watcher) initializeRegionLabelWatcher() error {
 // Close closes the watcher.
 func (rw *Watcher) Close() {
 	rw.cancel()
+	if rw.checkerController != nil {
+		rw.checkerController.ClearSuspectKeyRanges()
+	}
```
lhy1024, line 286: This cleanup order can still leave stale suspect key ranges. Close cancels the watcher and immediately calls ClearSuspectKeyRanges, but an already-running watcher callback is not preempted by cancellation and may still run postEventsFn/AddSuspectKeyRange afterward, re-adding suspect ranges after they were cleared.
  

Could we move ClearSuspectKeyRanges after rw.wg.Wait(), so cleanup happens after all watcher callbacks have exited?

lhy1024:
```
func TestCloseClearsSuspectKeyRangesAfterInFlightCallback(t *testing.T) {
        re := require.New(t)

        checkerCtx, checkerCancel := context.WithCancel(context.Background())
        defer checkerCancel()
        cluster := mockcluster.NewCluster(checkerCtx, mockconfig.NewTestOptions())
        checkerController := checker.NewController(checkerCtx, cluster, cluster.GetCheckerConfig(), nil)

        watcherCtx, watcherCancel := context.WithCancel(context.Background())
        rw := &Watcher{
                ctx:               watcherCtx,
                cancel:            watcherCancel,
                checkerController: checkerController,
        }

        callbackStarted := make(chan struct{})
        releaseCallback := make(chan struct{})
        rw.wg.Add(1)
        go func() {
                defer rw.wg.Done()
                <-watcherCtx.Done()
                close(callbackStarted)
                <-releaseCallback
                checkerController.AddSuspectKeyRange([]byte("a"), []byte("z"))
        }()

        closeDone := make(chan struct{})
        go func() {
                rw.Close()
                close(closeDone)
        }()

        select {
        case <-callbackStarted:
        case <-time.After(time.Second):
                re.FailNow("expected in-flight watcher callback to observe cancellation")
        }
        select {
        case <-closeDone:
                re.FailNow("Close returned before the in-flight callback finished")
        case <-time.After(50 * time.Millisecond):
        }

        close(releaseCallback)
        select {
        case <-closeDone:
        case <-time.After(time.Second):
                re.FailNow("Close did not return after the in-flight callback finished")
        }

        _, ok := checkerController.PopOneSuspectKeyRange()
        re.False(ok, "Close should not leave suspect key ranges added by in-flight callbacks")
  }
```

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -262,9 +272,75 @@ func (c *Cluster) BucketsStats(degree int, regionIDs ...uint64) map[uint64][]*bu
...
+	if c.metaWatcher != nil {
+		c.metaWatcher.Close()
+		c.metaWatcher = nil
+	}
+	if c.configWatcher != nil {
+		c.configWatcher.Close()
+		c.configWatcher = nil
+	}
+	if c.hbStreams != nil {
+		c.hbStreams.Close()
```
lhy1024: Closing hbStreams synchronously here does wait for an in-flight stream.Send to return. Close only cancels the context and then waits on wg; if hbstream.run has already entered the send/keepalive branch, it will not observe ctx.Done until Send returns. In production heartbeatServer.Send has a 5s timeout, so this is usually not a permanent block, but it can delay primary stepdown/transfer on slow streams.

Could we check context cancellation inside the send/keepalive path, or make the close path bounded, so primary exit is not amplified by slow streams?

lhy1024:
```
type blockingHeartbeatStream struct {
        once    sync.Once
        started chan struct{}
        release chan struct{}
  }

  func (s *blockingHeartbeatStream) Send(core.RegionHeartbeatResponse) error {
        s.once.Do(func() { close(s.started) })
        <-s.release
        return nil
  }

  func TestCloseDoesNotBlockOnInFlightSend(t *testing.T) {
        re := require.New(t)
        defer cancel()

        basicCluster := core.NewBasicCluster()
        basicCluster.PutStore(core.NewStoreInfo(&metapb.Store{Id: 1, Address: "store-1"}))
        streams := NewHeartbeatStreams(ctx, "", basicCluster)
        stream := &blockingHeartbeatStream{started: make(chan struct{}), release: make(chan struct{})}
        streams.BindStream(1, stream)

        leader := &metapb.Peer{Id: 1, StoreId: 1}
        region := core.NewRegionInfo(&metapb.Region{Id: 1, Peers: []*metapb.Peer{leader}}, leader)

        for {
                streams.SendMsg(region, &Operation{})
                select {
                case <-stream.started:
                        goto closeStream
                case <-time.After(time.Second):
                        re.FailNow("expected heartbeat stream send to start")
                default:
                        time.Sleep(time.Millisecond)
                }
        }

  closeStream:
        closeDone := make(chan struct{})
        go func() {
                streams.Close()
                close(closeDone)
        }()
select {
        case <-closeDone:
        case <-time.After(50 * time.Millisecond):
                close(stream.release)
                <-closeDone
                re.FailNow("HeartbeatStreams.Close should not wait for an in-flight stream Send after cancellation")
        }
  }
```

pkg/mcs/scheduling/server/rule/watcher.go
lhy1024: how about this
```
 func (rw *Watcher) Close() {
      rw.cancel()
      rw.wg.Wait()
      if rw.checkerController != nil {
          rw.checkerController.ClearSuspectKeyRanges()
      }
  }
```

bufferflies: good idea

pkg/mcs/scheduling/server/cluster.go
bufferflies: Yes, put the context check in the loop head.

pkg/mcs/scheduling/server/cluster.go
bufferflies: done, return the bool indicate the cancel behavior.

pkg/mcs/scheduling/server/cluster.go
lhy1024: I rechecked the latest commit. The new ctx check at the top of the loop avoids processing the next item after Close(), but it does not cover an already in-flight stream.Send; Close() still waits for that Send to return.

This is not a correctness blocker, because the production heartbeatServer.Send has a 5s timeout and should not block forever in normal cases.

However, it can still delay primary exit/transfer on slow connections, especially in the keepalive path where streams are sent one by one and thedelay can be amplified.

Could we confirm whether this stepdown latency risk is acceptable? If we want to avoid it, the send path may need to be interruptible, or Close may need a bounded wait.

lhy1024 (approved):

pkg/mcs/scheduling/server/cluster.go
bufferflies: yes, it's acceptable if the stream send block

pkg/mcs/scheduling/server/server.go
```diff
@@ -506,70 +500,113 @@ func (s *Server) startServer() (err error) {
...
 	cluster.StartBackgroundJobs()
+	initSucceeded = true
 	return nil
 }
 
 func (s *Server) stopCluster() {
-	cluster := s.GetCluster()
-	if cluster != nil {
+	if cluster := s.GetCluster(); cluster != nil {
 		s.cluster.Store((*Cluster)(nil))
```
rleungx, line 589: Could we resign/unpublish serving before clearing the cluster? `hbStreams.Close()` can wait on in-flight sends, so this node may still be seen as serving while handlers already return not bootstrapped.

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -653,10 +729,10 @@ func (c *Cluster) StartBackgroundJobs() {
 	c.running.Store(true)
 }
 
-// StopBackgroundJobs stops background jobs.
-func (c *Cluster) StopBackgroundJobs() {
+// StopBackgroundJobs stops background jobs, these jobs is created by NewCluster.
+func (c *Cluster) StopBackgroundJobs() bool {
 	if !c.running.Load() {
```
rleungx: Please make this an atomic test-and-set, for example `CompareAndSwap(true, false)`, so concurrent `StopBackgroundJobs` calls cannot both run the shutdown sequence.

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -76,7 +79,13 @@ type Cluster struct {
 	regionStats       *statistics.RegionStatistics
 	labelStats        *statistics.LabelStatistics
 	hotStat           *statistics.HotStat
+	resourceMu        sync.RWMutex
```
rleungx: How about using a better name instead of resourceMu

pkg/mcs/scheduling/server/server.go
bufferflies: Theoretically, hbstream can indeed get stuck, since send does not support context cancellation. To fully fix this, we would need to spawn a new goroutine for the send operation, which is somewhat wasteful. For now, I’ll just add some logging here, and if issues actually show up later, we can address it properly then.

pkg/mcs/scheduling/server/cluster.go
bufferflies: good， done

pkg/mcs/scheduling/server/cluster.go
bufferflies: Rename as runtimeMu and add a comment to describe the lock content.

pkg/mcs/scheduling/server/cluster.go
```diff
@@ -653,18 +741,18 @@ func (c *Cluster) StartBackgroundJobs() {
 	c.running.Store(true)
 }
 
-// StopBackgroundJobs stops background jobs.
-func (c *Cluster) StopBackgroundJobs() {
-	if !c.running.Load() {
-		return
+// StopBackgroundJobs stops background jobs, these jobs is created by NewCluster.
+func (c *Cluster) StopBackgroundJobs() bool {
+	if c.running.CompareAndSwap(true, false) {
```
rleungx: The CAS guard is inverted here. When `CompareAndSwap(true, false)` succeeds, this is the goroutine that should run the shutdown sequence, but the current code returns immediately, so a running cluster skips `coordinator.Stop()`, runner stops, `cancel()`, and `wg.Wait()`. Conversely, when `running` is already false, it falls through and can panic on a partially initialized cluster. The guard should be `if !c.running.CompareAndSwap(true, false) { return false }`.

I reproduced this locally with:

```sh
go test ./pkg/mcs/scheduling/server -run 'TestStopCluster|TestMultipleLeaderTermsNoHbStreamGoroutineLeak' -count=1
```

bufferflies: fixed ,done

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -138,7 +139,11 @@ func (s *Service) RegionHeartbeat(stream schedulingpb.Scheduling_RegionHeartbeat
 		}
 
 		c := s.GetCluster()
-		if c == nil {
+		var streams *hbstream.HeartbeatStreams
+		if c != nil {
+			streams = c.GetHeartbeatStreams()
```
rleungx: Is it possible to avoid the lock here?

bufferflies: yes,  only bindStream should acquice this lock.

bufferflies: done

pkg/mcs/scheduling/server/grpc_service.go
```diff
@@ -154,7 +155,13 @@ func (s *Service) RegionHeartbeat(stream schedulingpb.Scheduling_RegionHeartbeat
 		storeLabel := strconv.FormatUint(storeID, 10)
 
 		if time.Since(lastBind) > time.Minute {
```
rleungx, line 157: Could we keep a cheap readiness check on every heartbeat? With this check only inside the rebind path, an already-bound stream can skip it for up to one minute and still call `HandleRegionHeartbeat` on a stale cluster after `stopCluster` has cleaned up runtime resources.

bufferflies: No， the cluster set nil first if we stop the raft cluster, the new stream can't bind the stale stream

bufferflies: func (s *Server) stopCluster() {
    if cluster := s.GetCluster(); cluster != nil {
        s.cluster.Store((*Cluster)(nil))
        cluster.StopBackgroundJobs()
        cluster.cleanupRuntimeResources()
    }
}

rleungx (approved):

PR #10646, opened 2026-05-08
grafana: render stacked nulls as zero

nizheming:
## Summary
- set stacked Grafana graph panels to render null points as zero
- avoid misleading stacked visualizations when a series has data gaps

## Test
- parsed dashboard JSON files
- verified every `stack: true` graph panel has `nullPointMode: "null as zero"`

## Summary by CodeRabbit

* **Style**
  * Updated Grafana dashboard visualization settings to display missing data points as zeros in performance monitoring panels.

### DIFF
--- a/metrics/grafana/pd.json
+++ b/metrics/grafana/pd.json
@@ -2467,7 +2467,7 @@
           "lines": false,
           "linewidth": 1,
           "links": [],
-          "nullPointMode": "null",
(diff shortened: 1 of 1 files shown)


lhy1024 (approved):

lhy1024:
We need to solve DCO and add link for issues.

PR #10649, opened 2026-05-09
pkg/election: add keepalive tick interval metric, live TTL gauge

JmPotato:

Issue Number: ref #10653.

The lease keepalive worker emits a `the interval between keeping alive lease is too long` warning when its main-loop tick interval exceeds `2 × interval` (see `pkg/election/lease.go`). This signal is currently only observable via log scraping (e.g. Loki), which makes it hard to alert on, hard to compare across keyspace groups, and impossible to surface in Grafana.

The existing `pd_lease_local_ttl_remaining_seconds` gauge is also misleading: it is only `Set()` on each keepalive response, so every Prometheus scrape sees a value pinned at ~`leaseTimeout`. Real lease decay between renewals is invisible, defeating the original intent of the metric.

```commit-message
- Add `pd_lease_keepalive_tick_interval_seconds` histogram in
  `keepAliveWorker`, recording `start.Sub(lastTime)` on every loop
  iteration. Same source signal as the `the interval between keeping
  alive lease is too long` warning, but exposed as a quantitative,
  per-`purpose` distribution.
- Replace the `pd_lease_local_ttl_remaining_seconds` GaugeVec with a
  custom Prometheus collector that computes `time.Until(expireTime)`
  at scrape time. Active leases register on `Grant` and unregister on
  `Close`, so the metric reflects live lease decay (sawtooth between
  renewals) instead of being pinned at `leaseTimeout`. Closed leases
  no longer emit phantom zero series.
- Grafana: add `Lease keepalive tick interval` panel to the Leader row
  (P99 by job/purpose); shuffle `Lease renewal terminations` to keep
  the layout grid clean.
```

### Check List

Tests

- Unit test

Manual test: started a `tiup playground` cluster with PD microservices mode (`--pd.mode ms --tso 2 --scheduling 1 --resource-manager 1`) and 3 keyspace groups. Verified that `pd_lease_local_ttl_remaining_seconds` samples vary in `[3.33s, 5s]` across consecutive 1s scrapes (sawtooth of live lease decay), and `pd_lease_keepalive_tick_interval_seconds` buckets are populated on every PD/TSO instance.

image

Side effects

- None. Metric name `pd_lease_local_ttl_remaining_seconds` is preserved for dashboard backward compatibility; the value semantics shift from "frozen at TTL" to "live remaining TTL", which strictly improves observability without affecting any existing `< threshold` alerts (steady-state minimum is still `leaseTimeout - leaseTimeout/3`).

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Added a "Lease keepalive tick interval" panel to the Grafana dashboard.
  * Added histograms for keepalive request durations and tick intervals.

* **Improvements**
  * Switched to scrape-time TTL reporting for active leases for more accurate remaining-TTL metrics.
  * Dashboard queries updated to consistently exclude "expected primary" series.

* **Tests**
  * Added tests validating the TTL reporting collector behavior.

* **Refactor**
  * Internal lease handling updated to support the new metrics and instrumentation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10649)


pkg/election/metrics.go
```diff
@@ -62,41 +70,45 @@ var (
...
-	keepAliveRequestDuration = prometheus.NewHistogramVec(
+	keepAliveTickInterval = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Namespace: metricsNamespace,
 			Subsystem: metricsSubsystem,
-			Name:      "keepalive_request_duration_seconds",
-			Help:      "Duration of etcd Lease.KeepAliveOnce requests observed by PD, by purpose and result.",
+			Name:      "keepalive_tick_interval_seconds",
+			Help:      "Interval between consecutive iterations of the lease keepalive worker loop, by purpose. Spikes correlate with the `the interval between keeping alive lease is too long` warning.",
 			Buckets:   prometheus.ExponentialBuckets(0.001, 2, 16),
```
lhy1024: The tick interval histogram is capped at ~32.768s. Is it enough?

pkg/election/metrics.go
```diff
@@ -105,24 +117,74 @@ type leaseMetrics struct {
...
-	if seconds < 0 {
-		seconds = 0
+// localTTLRemainingCollector is a custom Prometheus collector that emits the
+// remaining lease TTL for every active Lease at scrape time. Computing the
+// value lazily means the curve naturally decays between renewals instead of
+// being frozen at the configured TTL.
+type localTTLRemainingCollector struct {
+	desc   *prometheus.Desc
+	mu     sync.Mutex
+	leases map[string]*Lease // purpose -> active Lease
```
lhy1024, line 144: This collector keys active leases only by `purpose`, but one process can hold multiple active leases with the same purpose, for example TSO expected-primary leases across keyspace groups use `"<service> expected primary"` without the group ID. Later registrations overwrite earlier ones, so the collector does not actually emit every active lease.

pkg/election/metrics.go
JmPotato: I think this is enough, because we can‘t make a lease renewal interval reach such a long time.

pkg/election/metrics.go
JmPotato: Currently, all purpose values that have observational significance have been changed to be unique.

Regarding the "expect primary" value you mentioned, it has already been filtered out from the Grafana and lacks any practical observational significance. Therefore, we can leave it as is for now.

lhy1024 (approved):

YuhaoZhang00 (approved):

pkg/election/metrics.go
```diff
@@ -62,41 +73,45 @@ var (
...
-	keepAliveRequestDuration = prometheus.NewHistogramVec(
+	keepAliveTickInterval = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Namespace: metricsNamespace,
 			Subsystem: metricsSubsystem,
-			Name:      "keepalive_request_duration_seconds",
-			Help:      "Duration of etcd Lease.KeepAliveOnce requests observed by PD, by purpose and result.",
+			Name:      "keepalive_tick_interval_seconds",
+			Help:      "Interval between consecutive iterations of the lease keepalive worker loop, by purpose. Spikes correlate with the `the interval between keeping alive lease is too long` warning.",
 			Buckets:   prometheus.ExponentialBuckets(0.001, 2, 16),
```
okJiang: It seems that we do not need millisecond-level information; maybe we can start from 10ms.

pkg/election/metrics.go
```diff
@@ -62,41 +73,45 @@ var (
...
-		prometheus.GaugeOpts{
+	keepAliveRequestDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
 			Namespace: metricsNamespace,
 			Subsystem: metricsSubsystem,
-			Name:      "local_ttl_remaining_seconds",
-			Help:      "PD local estimate of remaining lease TTL, updated on every keepalive response and on keepalive timeout firing; not etcd authoritative TTL.",
+			Name:      "keepalive_request_duration_seconds",
+			Help:      "Duration of etcd Lease.KeepAliveOnce requests observed by PD, by purpose and result.",
+			Buckets:   prometheus.ExponentialBuckets(0.001, 2, 16),
```
okJiang: ditto

okJiang:
image

“Lease keepalive response interval” and “Lease keepalive tick interval” look exactly the same. Are they duplicates?

JmPotato:
> image
> “Lease keepalive response interval” and “Lease keepalive tick interval” look exactly the same. Are they duplicates?

These two metrics measure different points in the keepalive pipeline:

1. **Tick interval** — the interval between consecutive `KeepAliveOnce` requests being dispatched to etcd (recorded at the top of the `keepAliveWorker` loop).
2. **Response interval** — the interval between receiving successful renewal responses in the main keepalive loop.

Under normal conditions they are nearly identical. Under runtime pressure (scheduler delay, etcd latency, GC pauses), the gap widens — tick interval captures how often we attempt renewal, while response interval captures how often renewal actually completes.

okJiang:
> > image
> > “Lease keepalive response interval” and “Lease keepalive tick interval” look exactly the same. Are they duplicates?
> 
> These two metrics measure different points in the keepalive pipeline:
> 
> 1. **Tick interval** — the interval between consecutive `KeepAliveOnce` requests being dispatched to etcd (recorded at the top of the `keepAliveWorker` loop).
> 2. **Response interval** — the interval between receiving successful renewal responses in the main keepalive loop.
> 
> Under normal conditions they are nearly identical. Under runtime pressure (scheduler delay, etcd latency, GC pauses), the gap widens — tick interval captures how often we attempt renewal, while response interval captures how often renewal actually completes.


> Tick interval — the interval between consecutive KeepAliveOnce requests being dispatched to etcd (recorded at the top of the keepAliveWorker loop).


"the interval between consecutive KeepAliveOnce requests being dispatched to etcd" Should it be placed before KeepAliveOnce to avoid the impact of goroutine scheduling?

> Under normal conditions they are nearly identical. Under runtime pressure (scheduler delay, etcd latency, GC pauses), the gap widens — tick interval captures how often we attempt renewal, while response interval captures how often renewal actually completes.

Perhaps you could add some descriptions suggesting when each one should be used, rather than just explaining what they are, so that a oncaller can understand them without having to pore over the code. Their names are too easy to confuse.

JmPotato:
> > > image
> > > “Lease keepalive response interval” and “Lease keepalive tick interval” look exactly the same. Are they duplicates?
> > 
> > 
> > These two metrics measure different points in the keepalive pipeline:
> > 
> > 1. **Tick interval** — the interval between consecutive `KeepAliveOnce` requests being dispatched to etcd (recorded at the top of the `keepAliveWorker` loop).
> > 2. **Response interval** — the interval between receiving successful renewal responses in the main keepalive loop.
> > 
> > Under normal conditions they are nearly identical. Under runtime pressure (scheduler delay, etcd latency, GC pauses), the gap widens — tick interval captures how often we attempt renewal, while response interval captures how often renewal actually completes.
> 
> > Tick interval — the interval between consecutive KeepAliveOnce requests being dispatched to etcd (recorded at the top of the keepAliveWorker loop).
> 
> "the interval between consecutive KeepAliveOnce requests being dispatched to etcd" Should it be placed before KeepAliveOnce to avoid the impact of goroutine scheduling?
> 
> > Under normal conditions they are nearly identical. Under runtime pressure (scheduler delay, etcd latency, GC pauses), the gap widens — tick interval captures how often we attempt renewal, while response interval captures how often renewal actually completes.
> 
> Perhaps you could add some descriptions suggesting when each one should be used, rather than just explaining what they are, so that a oncaller can understand them without having to pore over the code. Their names are too easy to confuse.

My latest commit includes two changes:

1. Improved the descriptions for each lease-related panel so that even those unfamiliar with the codebase can understand the meaning of each monitor.
2. Fixed the statistical scope to accurately record the timestamp between each `KeepAliveOnce` call.

okJiang (approved):

Issue #10653, opened 2026-05-09
election: improve lease renewal observability, stability, and performance


JmPotato:
## Enhancement Task

PD lease renewal is currently tied too closely to the configured lease TTL in some paths. When the lease TTL is increased, the `KeepAliveOnce` cadence can also become slower, which works against the goal of keeping leadership leases from expiring under runtime stalls, slow scheduling, or transient etcd latency.

Track lease renewal improvements around:

- Observability: expose and refine metrics/logging for renewal latency, response interval, remaining TTL, and termination reasons.
- Stability: make renewal cadence less sensitive to large lease TTL values and reduce lease-expiration risk.
- Performance: keep the renewal loop lightweight and avoid unnecessary load while preserving a safe renewal frequency.

Initial scoped PR:

- Use a fixed upper bound for PD election lease `KeepAliveOnce` intervals so increasing the lease TTL does not proportionally slow renewal.

PR #10654, opened 2026-05-09
election: use fixed lease keepalive interval

JmPotato:

Issue Number: ref #10653

```commit-message
Cap the election lease keepalive interval at 500ms.

PD's election lease renewal previously derived its `KeepAliveOnce` cadence
from `leaseTimeout/3`, so raising the lease TTL also slowed renewal. The
keepalive ticker interval is now `min(leaseTimeout/3, 500ms)`:

- Small leases keep the existing `leaseTimeout/3` behavior, matching etcd's
  built-in `clientv3.KeepAlive` cadence.
- Large leases are capped at 500ms so PD's leader-failover reaction time
  stays sub-second regardless of the configured TTL.

The per-call `KeepAliveOnce` context timeout continues to use the full
`leaseTimeout`, leaving enough RPC budget to absorb etcd tail latency above
the keepalive interval without prematurely cancelling renewals.
```

### Check List

Tests

- Unit test
- Manual test

image

### Release note

```release-note
Fixed an issue where increasing PD leader lease could also slow down lease keepalive renewal.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Capped lease keep-alive renewal cadence at 500ms for more consistent and reliable lease renewals.

* **Tests**
  * Updated keep-alive tests to wait for renewal signals with improved timeout handling to validate the new cadence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10654)

### DIFF
--- a/pkg/election/lease.go
+++ b/pkg/election/lease.go
@@ -158,7 +158,7 @@ func (l *Lease) KeepAlive(ctx context.Context) {
 	}
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	timeCh := l.keepAliveWorker(ctx, leaseKeepAliveInterval, leaseKeepAliveInterval)
+	timeCh := l.keepAliveWorker(ctx)
 	defer log.Info("lease keep alive stopped", zap.String("purpose", l.purpose))
 
 	var (
@@ -223,17 +223,17 @@ func (l *Lease) KeepAlive(ctx context.Context) {
 }
 
 // Periodically call `lease.KeepAliveOnce` and post back latest received expire time into the channel.
-func (l *Lease) keepAliveWorker(ctx context.Context, interval, timeout time.Duration) <-chan time.Time {
+func (l *Lease) keepAliveWorker(ctx context.Context) <-chan time.Time {
 	ch := make(chan time.Time)
 
 	go func() {
 		defer logutil.LogPanic()
-		ticker := time.NewTicker(interval)
+		ticker := time.NewTicker(leaseKeepAliveInterval)
 		defer ticker.Stop()
 
 		logger := log.With(zap.String("purpose", l.purpose),
 			zap.Int64("lease-id", int64(l.GetID())),
-			zap.Duration("interval", interval))
+			zap.Duration("interval", leaseKeepAliveInterval))
 		logger.Info("start lease keep alive worker")
 		defer logger.Info("stop lease keep alive worker")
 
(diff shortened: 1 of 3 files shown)


pkg/election/lease.go
```diff
@@ -34,6 +34,8 @@ const (
 	revokeLeaseTimeout = time.Second
 	requestTimeout     = etcdutil.DefaultRequestTimeout
 	slowRequestTime    = etcdutil.DefaultSlowRequestTime
+	// leaseKeepAliveInterval is fixed to renew leases frequently regardless of the configured lease timeout.
+	leaseKeepAliveInterval = 500 * time.Millisecond
```
lhy1024: why choose 500ms?

pkg/election/lease.go
```diff
@@ -184,25 +186,25 @@ func (l *Lease) KeepAlive(ctx context.Context) {
...
 		for {
 			start := time.Now()
-			if start.Sub(lastTime) > interval*2 {
+			if start.Sub(lastTime) > leaseKeepAliveInterval*2 {
 				log.Warn("the interval between keeping alive lease is too long", zap.Time("last-time", lastTime))
 			}
 			go func(start time.Time) {
 				defer logutil.LogPanic()
-				ctx1, cancel := context.WithTimeout(ctx, l.leaseTimeout)
+				ctx1, cancel := context.WithTimeout(ctx, leaseKeepAliveInterval)
```
rleungx: Timeout shouldn't be set to 500ms

pkg/election/lease.go
rleungx: BTW, shall we also consider if the lease is less than 2s?

pkg/election/lease_test.go
```diff
@@ -112,8 +112,11 @@ func TestLeaseKeepAlive(t *testing.T) {
 	lease := NewLease(client, "test_lease")
 
 	re.NoError(lease.Grant(defaultLeaseTimeout))
-	ch := lease.keepAliveWorker(ctx, 2*time.Second)
-	time.Sleep(2 * time.Second)
-	<-ch
+	ch := lease.keepAliveWorker(ctx)
+	select {
+	case <-ch:
+	case <-time.After(2 * lease.getKeepAliveInterval()):
```
rleungx, line 118: This does not really verify the capped cadence: the first keepalive is sent immediately, and the timeout is derived from the implementation under test. Please add a direct interval test, or assert timing on the second keepalive.

pkg/election/lease.go
```diff
@@ -221,9 +239,10 @@ func (l *Lease) KeepAlive(ctx context.Context) {
...
 
 // Periodically call `lease.KeepAliveOnce` and post back latest received expire time into the channel.
-func (l *Lease) keepAliveWorker(ctx context.Context, interval time.Duration) <-chan time.Time {
+func (l *Lease) keepAliveWorker(ctx context.Context) <-chan time.Time {
 	ch := make(chan time.Time)
 
+	interval := l.getKeepAliveInterval()
 	go func() {
 		defer logutil.LogPanic()
 		ticker := time.NewTicker(interval)
```
rleungx, line 248: This changes the ticker to 500ms for large leases, but each `KeepAliveOnce` below still uses `l.leaseTimeout` as its context timeout. With a large lease and slow etcd, that can accumulate many overlapping RPCs. Please cap the request timeout separately, or limit in-flight attempts.

PR #10665, opened 2026-05-12
keyspace: fix max valid keyspace region bound

### LINKED ISSUE #10664: keyspace: max valid keyspace region bound wraps around
### What did you do?

While splitting the keyspace bound fix out of #10648, we found that `keyspace.MakeRegionBound(constant.MaxValidKeyspaceID)` builds the upper bound by doing `id + 1` and then taking the low 24 bits.

For `MaxValidKeyspaceID` (`0xffffff`), this wraps the keyspace ID suffix to `00 00 00`.

The upper bound of the max valid keyspace should be the next lexicographical prefix after the mode-prefixed keyspace ID:

- raw mode: `r ff ff ff` -> `s 00 00 00`
- txn mode: `x ff ff ff` -> `y 00 00 00`

Encoded ranges should therefore end with:

- raw end key: `7300000000000000fb`
- txn end key: `7900000000000000fb`

### What did you see instead?

The upper bound wraps to keyspace ID zero under the same mode prefix:

- raw mode: `r ff ff ff` -> `r 00 00 00`
- txn mode: `x ff ff ff` -> `x 00 00 00`

Encoded ranges become:

- raw end key: `7200000000000000fb`
- txn end key: `7800000000000000fb`

This can make keyspace region bounds and the corresponding pd-ctl/keyspace label-rule ranges incorrect for the max valid keyspace ID.

Current master before the fix. The issue was found by code inspection and unit-test coverage around `MakeRegionBound`.

lhy1024:

Issue Number: Close #10664 ref #10592

`MakeRegionBound` used `id + 1` to build the upper bound. For the max valid keyspace ID, the low 24 bits wrap to zero, so the generated upper bound becomes `r/x 00 00 00` instead of the next prefix after `r/x ff ff ff`.

```commit-message
Fix keyspace region-bound generation to use the next prefix of the current mode-prefixed keyspace ID.

This keeps ordinary keyspace bounds unchanged and gives the max valid keyspace ID a non-wrapping upper bound for both raw and txn modes.
```

### Check List

Tests

- Unit test

Test commands:

```bash
go test ./pkg/keyspace -run TestMakeRegionBound -count=1
(cd tools && go test ./pd-ctl/pdctl/command -run TestMakeKeyRanges -count=1)
```

### Release note

```release-note
Fix incorrect keyspace range encoding for the max valid keyspace ID.
```

## Summary by CodeRabbit

* **Refactor**
  * Improved region-bound computation to ensure correct exclusive upper-bound semantics across edge cases, including the maximum keyspace value.

* **Tests**
  * Added unit tests validating region/range boundary encodings, including carry/overflow edge cases and the maximum keyspace value.
  * Converted an existing test to table-driven golden cases for raw and transactional ranges.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10665)

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10665)

### DIFF
--- a/pkg/keyspace/util.go
+++ b/pkg/keyspace/util.go
@@ -16,6 +16,7 @@ package keyspace
 
 import (
 	"container/heap"
+	"encoding/binary"
 	"encoding/hex"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


bufferflies:
The max valid keyspace ID may be too large; maybe OxFFFE is enough.

bufferflies (approved):

okJiang (approved):

Issue #10659, opened 2026-05-11
Optimize GCStateManager for capacity in large keyspace number scenarios (100k - 1M keyspaces)


MyonKeminta:
## Enhancement Task

 The `GCStateManager`, the core part of the GC refactory project (#8978), was not yet optimized for scenarios for massive keyspaces. However, as the situation and the product roadmap changes, the related optimization becomes a strong requirement. Also, we've encountered some incidents where high concurrency GC-related calls causes problem on PD or etcd, which should be avoided in the future.

Basically, we expect that the `GCStateManager` should be capable to serve at most 1M keyspaces, so that `GCStateManager` won't be the performance bottleneck for a long time.

### Subtasks

- [ ] Support excluding GC barriers for GC state retrieving APIs (`GetGCState`, `GetAllKeyspacesGCStates`)
- [ ] GC state cache (excluding GC barriers)
  - Ref: #10607
- [ ] WatchGCState
  - Ref: #10497
- [ ] Replace mutex (whose critical section contains IO operations) with cancellable synchronization methods
- [ ] Batch GC state writes
- [ ] Keyspace level mutexes and transactions (instead of global; requires special synchronization algorithm for global GC barriers)

PR #10669, opened 2026-05-12
gc: Support excluding GC barriers and global GC barriers when retrieving GC states

MyonKeminta:

Issue Number: Ref #10659

```commit-message
Makes GC states API `GetGCState` and `GetAllKeyspacesGCStates` supportexcluding GC barriers and global GC barriers.

As in most cases the caller don't really need these information, ignoring them could reduce the etcd read
pressure. This is also the prerequisite of supporting GC state caching -- we plan to cache all GC safe points
and txn safe points, but not including GC barriers to keep the memory consumption under control.

The protocol doesn't enable these options by default (otherwise old clients may lost GC barriers information
silently). But the new golang PD client enables them by default, and the updated PD client API can detect the
error.
```

* Requires: https://github.com/pingcap/kvproto/pull/1461

### Check List

Tests

- Unit test
- Integration test

Code changes

\-

Side effects

- Increased code complexity
  - Code calling `GetGCState` or `GetAllKeyspacesGCStates` with the Golang PD client and uses GC barriers info in it will need to modify their code to adapt to the changes.

Related changes

\-

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * GC State API now accepts options to fetch GC state with or without barrier data (exclusion-by-default).

* **Refactor**
  * GC barrier fields made internal; accessors added to indicate presence or return an error when absent.
  * Manager and server APIs updated to support barrier-exclusion views and separate result handling.

* **Tests**
  * Added/updated tests for barrier inclusion/exclusion, accessor behavior, and concurrency isolation of option variants.

* **Chores**
  * Updated module dependency versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10669?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pkg/gc/gc_state_manager.go
```diff
@@ -698,15 +700,15 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32) (GCState, error) {
 // the caller should NEVER change the content of the returned result and error. It's guaranteed that the returned result
 // must be fetched AFTER the beginning of the current invocation, and it never reuses the result of invocations that
 // started earlier than the current one.
-func (m *GCStateManager) GetAllKeyspacesGCStates(ctx context.Context) (map[uint32]GCState, error) {
+func (m *GCStateManager) GetAllKeyspacesGCStates(ctx context.Context, excludeGCBarriers bool) (map[uint32]GCState, error) {
 	return m.allKeyspacesGCStatesSingleFlight.Do(ctx, func(execCtx context.Context) (map[uint32]GCState, error) {
-		result, err := m.getAllKeyspacesGCStatesImpl(execCtx)
+		result, err := m.getAllKeyspacesGCStatesImpl(execCtx, excludeGCBarriers)
```
JmPotato, line 715: `singleflight.Do` uses `ctx` as the dedup key, but `excludeGCBarriers` is not included in the key. When two concurrent calls pass different `excludeGCBarriers` values, only one lambda will execute, and the other caller will get a result of the wrong shape (for example, caller A passes `excludeGCBarriers=false` expecting to get GC barriers, but after dedup, it gets the result from caller B who passed `excludeGCBarriers=true` — no barriers).

MyonKeminta: It's not the builtin singleflight but the hand-written linearizable version `OrderedSingleFlight`, and it doesn't support key yet.
I added a separated singleflight fir it.

client/clients/gc/client.go
```diff
@@ -32,6 +34,45 @@ type Client interface {
...
+// GCStatesAPIOptions represents all options for GC states API.
+//
+//nolint:revive
+type GCStatesAPIOptions struct {
+	ExcludeGCBarriers       bool
+	ExcludeGlobalGCBarriers bool
+}
+
+// DefaultGCStatesAPIOptions returns the default options for GC states API.
+func DefaultGCStatesAPIOptions() GCStatesAPIOptions {
```
rleungx, line 46: This changes the Go client default behavior: existing callers of GetGCState/GetAllKeyspacesGCStates will silently stop receiving barriers. Can we keep the old default and make exclusion opt-in, or call this out as a breaking change?

client/clients/gc/client.go
```diff
@@ -245,16 +286,89 @@ func (b *GlobalGCBarrierInfo) isExpiredImpl(now time.Time) bool {
...
 type GCState struct {
 	// The ID of the keyspace this GC state belongs to.
-	KeyspaceID   uint32
-	TxnSafePoint uint64
-	GCSafePoint  uint64
-	GCBarriers   []*GCBarrierInfo
+	KeyspaceID    uint32
+	TxnSafePoint  uint64
+	GCSafePoint   uint64
+	hasGCBarriers bool
```
rleungx, line 292: Removing the exported barrier fields breaks source compatibility for existing users. Could we keep exported fields for compatibility and add Has/Get helpers to distinguish not-fetched from an empty result?

client/gc_client.go
```diff
@@ -331,12 +337,10 @@ func pbToGCState(pb *pdpb.GCState, reqStartTime time.Time) gc.GCState {
 	for _, b := range pb.GetGcBarriers() {
 		gcBarriers = append(gcBarriers, pbToGCBarrierInfo(b, reqStartTime))
 	}
-	return gc.GCState{
-		KeyspaceID:   keyspaceID,
-		TxnSafePoint: pb.GetTxnSafePoint(),
-		GCSafePoint:  pb.GetGcSafePoint(),
-		GCBarriers:   gcBarriers,
+	if !excludeGCBarriers {
```
rleungx: Before this branch, we already allocate and convert all returned barriers. During rolling upgrades, an old PD may still return barriers even when the new client requested exclusion, so we can return early before the conversion.

client/clients/gc/client.go
MyonKeminta: As it's hard to push all callers to update their usage, I suggest let this be a breaking change. Callers that really needs the GC barriers will receive compile error, and is forced to switch to the safe retriver function; then if they missed the new options, they will get an error.
Actually, I expect that there's no such usage for now. If so, nothing will happen after upgrading the PD client.

client/clients/gc/client.go
MyonKeminta: Ditto. And this prevents misuse.

rleungx (approved):

JmPotato (approved):

PR #10677, opened 2026-05-14
gc: Implement GC state cache 

### LINKED ISSUE #10607: GCStateManager calls may be blocked and unable to cancel or timeout when there's IO issue

MyonKeminta:

Issue Number: Ref #10659 , Close #10607

```commit-message
This PR implements GC state cache. It now caches GC safe points and txn safe points of all keyspaces in memory, and can be used by GetGCState and GetAllKeyspacesGCStates.
This is a formal version of #10615 , and includes tests from #10617 with proper adaptions
```

**Requires**:

* https://github.com/pingcap/kvproto/pull/1461
* https://github.com/tikv/pd/pull/10669

**Note**:

This PR has some different behavior from #10615 :

* The second leader check in `GrpcServer` after invoking `GCStateManager` is removed.
* `GCStateManager` has built-in leader check. Cache-read is disallowed when it's not leader, but it can still try etcd IO read in this case, as etcd should be able to guarantee the consistency.

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  - Option to exclude GC barriers from GC state views; new metric reporting cache access outcomes.

* **Performance Improvements**
  - Leader-aware, sharded in-memory per-keyspace GC state cache with automatic invalidation on leadership changes to reduce safe-point read latency.

* **Tests**
  - Expanded unit/integration tests and benchmarks for cache semantics, concurrency, leadership churn, exclusion behavior, and failure recovery.

* **Chores**
  - Cluster lifecycle integration and client/health-checker test harness configurability.

### DIFF
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -355,6 +355,10 @@ func (m *GCStateManager) advanceGCSafePointImpl(ctx context.Context, keyspaceID 
 		if err1 != nil {
 			return err1
 		}
+		txnSafePoint, err1 = m.gcMetaStorage.LoadTxnSafePoint(keyspaceID)
+		if err1 != nil {
+			return err1
+		}
 		if target < oldGCSafePoint {
 			if compatible {
 				// When in compatible mode, trying to update the safe point to a smaller value fails silently, returning
@@ -367,10 +371,6 @@ func (m *GCStateManager) advanceGCSafePointImpl(ctx context.Context, keyspaceID 
 			}
 			// Otherwise, return error to reject the operation explicitly.
 			return errs.ErrDecreasingGCSafePoint.GenWithStackByArgs(oldGCSafePoint, target)
-		}
-		txnSafePoint, err1 = m.gcMetaStorage.LoadTxnSafePoint(keyspaceID)
-		if err1 != nil {
-			return err1
 		}
 		if target > txnSafePoint {
 			return errs.ErrGCSafePointExceedsTxnSafePoint.GenWithStackByArgs(oldGCSafePoint, target, txnSafePoint)
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -1028,7 +1028,7 @@ func (m *GCStateManager) iterateAllKeyspacesGCStates(
 		}
 
 		// Workaround: Check unified GC before checking `keyspacePred`. This breaks the semantics of
-		// the function, but it makes sure keyspaces that chagned from keyspace-level GC to unified GC, the invalidated
+		// the function, but it makes sure keyspaces that changed from keyspace-level GC to unified GC, the invalidated
 		// cached GC states (if any) are always overwritten by the unified GC ones.
 
 		if keyspaceMeta.Config[keyspace.GCManagementType] != keyspace.KeyspaceLevelGC {
--- a/pkg/gc/gc_state_manager.go
+++ b/pkg/gc/gc_state_manager.go
@@ -922,7 +922,8 @@ func (m *GCStateManager) GetAllKeyspacesGCStates(ctx context.Context, excludeGCB
 				func(_keyspaceID uint32) bool { return true },
 				func(gcState GCState) {
 					result[gcState.KeyspaceID] = gcState
-				})
+				},
+				nil)
 			failpoint.Inject("onGetAllKeyspacesGCStatesFinish", func() {})
 			return result, err
 		})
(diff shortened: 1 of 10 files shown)


pkg/gc/gc_state_manager.go
```diff
@@ -705,101 +917,99 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32, excludeGCBarriers bool) (
...
 	}
 
+	// If excludeGCBarriers is set, most required information should be available in the cache.
+	// Note that the invocation with and without `excludeGCBarriers` should go to different singleflights, otherwise
+	// invocation with different parameters may share their results incorrectly.
 	return m.allKeyspacesGCStatesExcludeGCBarriersSingleFlight.Do(ctx, func(execCtx context.Context) (map[uint32]GCState, error) {
-		result, err := m.getAllKeyspacesGCStatesImpl(execCtx, excludeGCBarriers)
+		cachedGCStates := make(map[uint32]GCState)
+		if m.nodeIsLeader() {
+			cachedGCStates = m.gcStateCache.cloneAllAsGCStates()
```
rleungx: Cloning the whole cache into the result can return stale keyspaces. Cached entries skip the iteration below, so current keyspace state and GC-management filters are not re-applied. If a cached keyspace is disabled/archived or switched to unified GC, `GetAllKeyspacesGCStates(..., true)` will still return it. Please iterate current keyspaces first and use the cache only after the keyspace passes the filters.

pkg/gc/gc_state_manager.go
```diff
@@ -263,9 +382,17 @@ func (m *GCStateManager) advanceGCSafePointImpl(ctx context.Context, keyspaceID
 		log.Error("failed to advance GC safe point",
 			zap.Uint32("keyspace-id", keyspaceID), zap.String("keyspace-name", getKeyspaceNameFromCtx(ctx)),
 			zap.Uint64("target", target), zap.Bool("compatible-mode", compatible), zap.Error(err))
+		// Invalidate cache on error.
+		m.gcStateCache.remove(keyspaceID)
 		return 0, 0, err
 	}
 
+	// Update cache.
+	m.gcStateCache.store(keyspaceID, gcStateCacheEntry{
```
rleungx, line 393: This can cache an invalid txn safe point in the compatible no-op path. When `compatible` is true and `target < oldGCSafePoint`, the transaction returns before `txnSafePoint` is loaded, so this stores `TxnSafePoint: 0` with the current GC safe point. A later `GetGCState(..., true)` can then serve an impossible cached state. Please load the txn safe point before that early return or skip updating the cache for that path.

pkg/gc/gc_state_manager.go
```diff
@@ -187,6 +301,10 @@ func (m *GCStateManager) CompatibleLoadGCSafePoint(keyspaceID uint32) (uint64, e
 		return 0, err
 	}
 
+	if cachedGCState, ok := m.gcStateCache.load(keyspaceID); ok {
```
rleungx: Should this legacy read also gate cache usage on `nodeIsLeader()`? `GetGCState` avoids serving cached state on followers, but this path can return a cached GC safe point unconditionally.

pkg/gc/gc_state_manager.go
MyonKeminta: I was considering to handle the case that keyspace being disabled or archived in later PRs, which may requires listeneng to events from `KeyspaceManager` to invalidate the cache in time.
Ok, for now I'll try to remove the items in the result by checking the result of scanning keyspaces.
However, I guess we can ignore the possibility that a keyspace is switched from keyspace-level GC to unified GC. It should never actually happen, and we can consider disable its code path.

pkg/gc/gc_state_manager.go
```diff
@@ -705,101 +919,137 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32, excludeGCBarriers bool) (
...
+			})
+
+		// Filter out invalid GC states that may be stale in the cache.
+		keyspacesToRemove := make([]uint32, 0)
+		for keyspaceID := range gcStates {
+			if _, ok := actualKeyspaces[keyspaceID]; !ok {
+				keyspacesToRemove = append(keyspacesToRemove, keyspaceID)
+			}
+		}
+		for _, keyspaceID := range keyspacesToRemove {
```
rleungx, line 969: Can we also remove these invalid keyspaces from gcStateCache? Otherwise archived keyspaces stay in memory and are cloned on every all-keyspaces cache read.

pkg/gc/gc_state_manager.go
```diff
@@ -629,6 +775,76 @@ func (m *GCStateManager) deleteGCBarrierImpl(ctx context.Context, keyspaceID uin
 	return deletedBarrier, nil
 }
 
+func (m *GCStateManager) getGCStateImpl(keyspaceID uint32, excludeGCBarriers bool) (GCState, error) {
+	// Try getting from cache if possible.
+	if excludeGCBarriers && m.nodeIsLeader() {
```
rleungx, line 780: Could this fast path use the same leadership epoch/lock as stepdown? nodeIsLeader() can become false before OnNodeBecomesFollower finishes clearing the cache, so an old leader may still serve a cached value.

pkg/gc/gc_state_manager.go
```diff
@@ -705,101 +919,137 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32, excludeGCBarriers bool) (
...
 		if !ok {
 			break
 		}
 
-		// Just handle the active keyspace, leave the others up to keyspace management.
 		if keyspaceMeta.State != keyspacepb.KeyspaceState_ENABLED {
 			continue
 		}
 
+		if !keyspacePred(keyspaceMeta.GetId()) {
```
rleungx: Cached entries skip the later GC-management filter. If a warm keyspace is changed from keyspace-level GC to unified GC, this can return stale per-keyspace safe points.

pkg/gc/gc_state_manager_test.go
```diff
@@ -2368,52 +2857,230 @@ func benchmarkGetAllKeyspacesGCStatesImpl(b *testing.B, keyspacesCount int, para
...
+		go func() {
+			defer wg.Done()
+			for i := uint64(1); ; i++ {
+				select {
+				case <-stopWriteCh:
+					return
+				default:
+				}
+				keyspaceID := rand.Uint32N(uint32(keyspacesCount)) + 1
+				_, err := gcStateManager.AdvanceTxnSafePoint(keyspaceID, i, time.Now())
```
rleungx: These writer goroutines use independent counters, so two writers can pick the same keyspace and hit ErrDecreasingTxnSafePoint. The benchmark should use a shared monotonic counter or ignore only that expected error.

pkg/gc/gc_state_manager.go
MyonKeminta: It may become harder to keep the cache consistent with keyspaces if there are concurrent enabling/disabling behavior.
But it should be ok to do so ignoring the potential problem. I'll add it.

pkg/gc/gc_state_manager.go
MyonKeminta: Sorry I didn't understand what you mean

pkg/gc/gc_state_manager.go
MyonKeminta: Can we just disable switching GC mode?

pkg/gc/gc_state_manager.go
```diff
@@ -705,101 +919,138 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32, excludeGCBarriers bool) (
...
+		if m.nodeIsLeader() {
+			gcStates = m.gcStateCache.cloneAllAsGCStates()
+		}
+
+		actualKeyspaces := make(map[uint32]struct{}, len(gcStates))
+
+		err := m.iterateAllKeyspacesGCStates(execCtx, excludeGCBarriers,
+			func(keyspaceID uint32) bool {
+				_, ok := gcStates[keyspaceID]
+				return !ok
```
rleungx, line 948: This skips cached keyspaces before checking their current GC mode. If a keyspace is changed away from keyspace_level, the stale cached state can still be returned here; please validate the current config or evict the cache entry before reusing it.

pkg/gc/gc_state_manager.go
```diff
@@ -705,101 +919,142 @@ func (m *GCStateManager) GetGCState(keyspaceID uint32, excludeGCBarriers bool) (
...
 			break
 		}
 
-		// Just handle the active keyspace, leave the others up to keyspace management.
 		if keyspaceMeta.State != keyspacepb.KeyspaceState_ENABLED {
 			continue
 		}
 
+		// Workaround: Check unified GC before checking `keyspacePred`. This breaks the semantics of
+		// the function, but it makes sure keyspaces that chagned from keyspace-level GC to unified GC, the invalidated
```
rleungx:
```suggestion
		// the function, but it makes sure keyspaces that changed from keyspace-level GC to unified GC, the invalidated
```

rleungx (approved):

pkg/gc/gc_state_manager.go
```diff
@@ -422,9 +556,21 @@ func (m *GCStateManager) advanceTxnSafePointImpl(ctx context.Context, keyspaceID
 		return wb.SetTxnSafePoint(keyspaceID, newTxnSafePoint)
 	})
 	if err != nil {
+		log.Error("failed to advance GC safe point",
```
JmPotato:
```suggestion
		log.Error("failed to advance txn safe point",
```

pkg/gc/gc_state_manager.go
```diff
@@ -676,21 +892,19 @@ func (m *GCStateManager) getGCStateInTransaction(keyspaceID uint32, excludeGCBar
...
-	var result GCState
-	err = m.gcMetaStorage.RunInGCStateTransaction(func(wb *endpoint.GCStateWriteBatch) error {
-		var err1 error
-		result, err1 = m.getGCStateInTransaction(keyspaceID, excludeGCBarriers, wb)
-		return err1
-	})
+	gcState, err := m.getGCStateImpl(keyspaceID, excludeGCBarriers)
+
+	if err != nil {
+		log.Error("failed to get GC satete", zap.Uint32("keyspace-id", keyspaceID),
```
JmPotato:
```suggestion
		log.Error("failed to get GC state", zap.Uint32("keyspace-id", keyspaceID),
```

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

niubell (approved):

PR #10682, opened 2026-05-15
server/api: allow follower local region reads

### LINKED ISSUE #10681: api: allow PD followers to serve local region read APIs
## Enhancement Task

Tracking sub-issue of #10666.

### Background

PD followers maintain a local region view through the region syncer. Today this view can be read through the gRPC Active Follower path, but the HTTP region metadata APIs still go through `clusterRouter` and are rejected on followers with:

`TiKV cluster not bootstrapped, please start TiKV first`

This makes production debugging harder because operators cannot directly inspect a follower's local region view through HTTP.

This issue only tracks follower-local read APIs. Resetting or deleting local region cache should be tracked separately by #10667.

### Goal

Allow PD followers to serve safe, read-only HTTP region metadata APIs from their local synced region cache when the request explicitly opts in with `PD-Allow-Follower-Handle: true`.

Expected APIs include read-only region query paths such as:

- `/pd/api/v1/regions`
- `/pd/api/v1/region/id/{id}`
- `/pd/api/v1/region/key/{key}`
- `/pd/api/v1/regions/key`
- `/pd/api/v1/regions/count`
- `/pd/api/v1/regions/store/{id}`
- related read-only region inspection endpoints

### Non-goals

- Do not reset, delete, or mutate follower local region cache in this issue.
- Do not make follower-local handling the default behavior.
- Do not serve APIs that depend on leader-only scheduler state unless they are proven safe.

### Suggested direction

Use the follower's local synced region cache for safe region read APIs when `PD-Allow-Follower-Handle: true` is set and the region syncer is running. Keep existing redirect behavior for normal requests and keep unsafe or write APIs protected.

okJiang:

Issue Number: Close #10681

```commit-message
Allow safe read-only region HTTP APIs to use the follower's local synced region cache when `PD-Allow-Follower-Handle: true` is set and the region syncer is running.

Keep normal redirect behavior unchanged for requests without the opt-in header, and keep follower-local handling limited to read-only region inspection paths.
```

### Check List

Tests

- Unit test

Code changes

### Release note

```release-note
PD followers can now serve read-only region HTTP APIs from their local synced region cache when explicitly requested.
```

## Summary by CodeRabbit

* **New Features**
  * Follower-synced nodes can directly serve many region read endpoints (lookups, lists/counts, topology and read stats) when a specific GET header is present, improving read availability when the leader is not serving. Region-read routes now use the follower-capable middleware.

* **Bug Fixes / Behavior**
  * Certain region check endpoints still respond as unbootstrapped even with the allow header.

* **Tests**
  * Added test verifying follower-served region responses and unchanged behavior for some endpoints.

* **Documentation**
  * Clarified comment on a helper used for follower-local access.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10682?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/api/middleware.go
```diff
@@ -102,6 +107,54 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
...
+	rc := m.s.DirectlyGetRaftCluster()
+	if rc == nil || !rc.GetRegionSyncer().IsRunning() {
+		return nil
+	}
+	return rc
+}
+
+func isFollowerSyncedRegionPath(path string) bool {
+	switch {
+	case path == "/pd/api/v1/regions",
```
okJiang: Why requests under these paths need to be skipped, please add comments

server/api/middleware.go
```diff
@@ -93,6 +95,9 @@ func newClusterMiddleware(s *server.Server) clusterMiddleware {
 func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		rc := m.s.GetRaftCluster()
```
rleungx, line 109: Shall we check the header first?

server/api/middleware.go
```diff
@@ -102,6 +107,61 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
...
 
+func (m clusterMiddleware) getFollowerSyncedCluster(r *http.Request) *cluster.RaftCluster {
+	if r.Method != http.MethodGet ||
+		r.Header.Get(apiutil.PDAllowFollowerHandleHeader) == "" ||
+		!isFollowerSyncedRegionPath(r.URL.Path) ||
+		m.s.GetMember().IsServing() {
+		return nil
+	}
+	rc := m.s.DirectlyGetRaftCluster()
+	if rc == nil || !rc.GetRegionSyncer().IsRunning() {
```
rleungx, line 130: IsRunning() may be too weak as a readiness gate here. The syncer can become running after the first sync batch, so large clusters may serve partial region cache results before the initial full sync finishes.

server/api/middleware.go
```diff
@@ -102,6 +107,61 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
...
+		return nil
+	}
+	return rc
+}
+
+func isFollowerSyncedRegionPath(path string) bool {
```
rleungx: Is there a better way instead of path match?

server/api/middleware.go
okJiang: For debugging interfaces like this, is it acceptable for them to provide results ahead of synced?

server/api/middleware.go
okJiang: This is a middleware that both the leader and follower will use. I don't think it's strictly necessary to check the header.

server/api/middleware.go
okJiang: Fixed by moving the follower-synced region allowlist to route registration. The cluster middleware now only enables follower local region reads for routes registered on the follower-synced region router, so it no longer matches request paths manually.

server/api/middleware.go
```diff
@@ -79,20 +80,36 @@ func (rm *requestInfoMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
+	for _, opt := range opts {
+		opt(&m)
+	}
+	return m
 }
 
 func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		rc := m.s.GetRaftCluster()
+		if rc == nil {
```
bufferflies, line 110: Does it need to check allowFollowerSyncedRegion and set it to true?

server/api/middleware.go
```diff
@@ -102,6 +119,20 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
 	})
 }
 
+func (m clusterMiddleware) getFollowerSyncedCluster(r *http.Request) *cluster.RaftCluster {
+	if r.Method != http.MethodGet ||
+		r.Header.Get(apiutil.PDAllowFollowerHandleHeader) == "" ||
```
bufferflies: r.Header.Get(apiutil.PDAllowFollowerHandleHeader) == "". This condition is better to put the latest check.

server/api/router.go
```diff
@@ -245,27 +248,27 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
...
+	registerFunc(followerSyncedRegionRouter, "/regions/readquery", regionsHandler.GetTopReadQueryRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(followerSyncedRegionRouter, "/regions/confver", regionsHandler.GetTopConfVerRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(followerSyncedRegionRouter, "/regions/version", regionsHandler.GetTopVersionRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(followerSyncedRegionRouter, "/regions/size", regionsHandler.GetTopSizeRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(followerSyncedRegionRouter, "/regions/keys", regionsHandler.GetTopKeysRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
+	registerFunc(followerSyncedRegionRouter, "/regions/cpu", regionsHandler.GetTopCPURegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
 	registerFunc(clusterRouter, "/regions/check/miss-peer", regionsHandler.GetMissPeerRegions, setMethods(http.MethodGet), setAuditBackend(prometheus))
```
bufferflies, line 272: Maybe the checker api can be allowed to access in the follower server?

server/api/middleware.go
rleungx: Partial results might be acceptable, but then we should document that these endpoints are eventually consistent and may be incomplete during the initial full sync. Otherwise this needs a stronger readiness signal, because streamingRunning is set after each received batch while the leader may still be sending the rest of the full sync.

server/api/router.go
rleungx: I would keep these checker APIs out for now. They call Handler.GetRegionsByType(), which reads s.GetRaftCluster() instead of the request cluster from getCluster(r), so a follower still returns ErrNotBootstrapped unless we refactor the handler/statistics path.

server/api/router.go
```diff
@@ -130,8 +130,11 @@ func createRouter(prefix string, svr *server.Server) *mux.Router {
 
 	clusterRouter := apiRouter.NewRoute().Subrouter()
 	clusterRouter.Use(newClusterMiddleware(svr).middleware)
+	followerSyncedRegionRouter := apiRouter.NewRoute().Subrouter()
+	followerSyncedRegionRouter.Use(newClusterMiddleware(svr, withFollowerSyncedRegion()).middleware)
 
 	escapeRouter := clusterRouter.NewRoute().Subrouter().UseEncodedPath()
+	followerSyncedRegionEscapeRouter := followerSyncedRegionRouter.NewRoute().Subrouter().UseEncodedPath()
```
rleungx: BTW, can we use another name? Otherwise, it sounds like it's only used by followers

server/api/middleware.go
okJiang: Tracked this follow-up in #10690. The client-side behavior should avoid sending follower region requests to a PD follower until that follower has completed full region sync; `IsRunning()` alone is too weak if it can become true before the full region view is ready.

server/api/router.go
okJiang: done

server/api/middleware.go
```diff
@@ -102,6 +119,20 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
 	})
 }
 
+func (m clusterMiddleware) getFollowerSyncedCluster(r *http.Request) *cluster.RaftCluster {
+	if r.Method != http.MethodGet ||
+		!m.allowFollowerSyncedRegion ||
+		m.s.GetMember().IsServing() ||
+		r.Header.Get(apiutil.PDAllowFollowerHandleHeader) == "" {
```
rleungx: What if PDAllowFollowerHandleHeader is another value?

server/api/middleware.go
```diff
@@ -102,6 +119,20 @@ func (m clusterMiddleware) middleware(h http.Handler) http.Handler {
...
+		return nil
+	}
+	rc := m.s.DirectlyGetRaftCluster()
```
rleungx, line 129: I think we need to change the "Only used for test" comment in DirectlyGetRaftCluster.

rleungx (approved):
The rest LGTM

server/api/router.go
bufferflies: OK

bufferflies (approved):

Issue #10666, opened 2026-05-12
region syncer: follower may serve stale region cache forever after history buffer overflow


JmPotato:
### What did you do?

Run a PD cluster where a follower's region sync stream falls behind the leader by more than `defaultHistoryBufferSize` (10000) records, then recovers.

### What did you expect to see?

After the sync stream recovers, the follower's region cache (and local LevelDB region storage) should converge with the leader's view in bounded time.

### What did you see instead?

The follower keeps serving stale region cache indefinitely for any region that does not receive a new update after the gap.

Code paths confirming the bug:

1. After leader election, the follower loads regions exclusively from its **local** LevelDB region storage (`pkg/storage/storage.go` `TryLoadRegionsOnce`, `pkg/syncer/client.go` `StartSyncWithLeader`). It never pulls a fresh snapshot from the leader on startup.

2. The follower opens the sync stream with `StartIndex: s.history.getNextIndex()` (`pkg/syncer/client.go` `syncRegion`). The index is persisted to LevelDB under `historyKey` and reloaded after restart (`pkg/syncer/history_buffer.go`).

3. The leader keeps a fixed-size ring buffer of the last `defaultHistoryBufferSize = 10000` records (`pkg/syncer/server.go`).

4. In `syncHistoryRegion` (`pkg/syncer/server.go`), when `recordsFrom(startIndex)` returns nothing:
   - If `startIndex == 0`, the leader pushes a full snapshot.
   - Else if `getNextIndex() == startIndex`, the leader replies with an empty in-sync response.
   - **Else (the buffer-overflow case), the leader only logs `"no history regions from index, the leader may be restarted"` and returns `nil`.** No full sync, no error to the follower.

5. The follower's stale cache is only refreshed when the leader pushes a newer version of the same region. Updates that happened during the gap are silently lost; `bc.PreCheckPutRegion` (`pkg/syncer/client.go`) drops anything older than the in-cache copy, so stale entries are never overwritten by replays.

This becomes user-visible in at least the following two scenarios:

1. The PD leader is transferred to a follower that holds a stale view of some region, and that region has not reported a heartbeat for a long time — the new leader will keep serving / scheduling on the stale region info.
2. With Active Follower enabled, region information served by the affected follower may be stale.

Suggested directions to track in this issue:

- Detect the overflow in `syncHistoryRegion` (e.g. `startIndex < history.firstIndex()`) and force the follower into a full snapshot path instead of returning `nil`.
[...]

PR #10685, opened 2026-05-15
syncer: close failed follower streams

### LINKED ISSUE #10684: region syncer: leader drops failed follower stream without closing it
### What did you do?

Run a PD cluster with region syncer enabled. A follower keeps a `SyncRegions` stream open to the PD leader, and the leader later fails to send a region sync response to that follower in `broadcast()`.

Relevant current code path:

1. `Sync()` binds the follower stream in `s.mu.streams`.
2. `broadcast()` sends region updates through each bound stream.
3. If `sender.Send(regions)` returns an error, `broadcast()` deletes `s.mu.streams[name]`.
4. Deleting the stream from the map does not make the corresponding `Sync()` handler return an error to the follower. The handler can remain waiting on `stream.Recv()`, while the follower is also waiting on `stream.Recv()` for leader data.

When the PD leader drops a failed follower stream, the server-side `Sync()` handler should return an error for that stream.

The follower should observe the stream error and enter the existing reconnect path.

### What did you see instead?

The leader removes the stream from memory and no longer sends data to that follower, but the follower may still consider the stream normal and keep waiting for data.

From that point, the affected follower can stop receiving region updates until the transport eventually reports an error by itself.

Related issues:

- #10666 tracks stale follower region cache and history buffer overflow after region sync falls behind.
- https://github.com/tikv/pd/issues/10666#issuecomment-4428929411 describes this leader-side stream deletion path.
- #5221 tracked the same follower-side symptom and is now closed in favor of this issue.

Reproduced by code inspection on current `master`.

okJiang:

Issue Number: Close #10684, ref #10666

```commit-message
Make region syncer close a failed follower stream when broadcast drops it.

Each bound follower stream now carries a close signal owned by the leader-side region syncer. When broadcast fails to send data to a follower and removes that stream from memory, it also closes the signal so the matching Sync handler returns Unavailable. The follower then observes a stream error and uses the existing reconnect path instead of waiting forever on a stream that the leader no longer tracks.

The same close signal is also used when a follower replaces an old stream and when the leader-side region syncer stops.
```

### Test result comparison

Red test on base `master` without this PR fix:

- Branch: `codex/repro-region-syncer-send-failure-master`
- Base: `upstream/master` `8cabcf154`
- Local repro commit: `fb8921ab9` only adds the integration test and a failpoint hook.
- Command: `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncerReconnectsAfterLeaderSendFailure -count=1 -v'`
- Result: failed at `tests/server/region_syncer/region_syncer_test.go:251`, waiting for `cluster.WaitRegionSyncerClientsReady(2)`.
- Observed logs: the leader injected a send failure, logged `region syncer send data meet error`, then logged `region syncer delete the stream [stream=pd2]`. The follower did not observe a stream error and did not reconnect.

Green test on this PR branch:

- Branch: `codex/region-syncer-close-streams`
- Command: `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncerReconnectsAfterLeaderSendFailure -count=1'`
- Result: passed, `ok github.com/tikv/pd/tests/server/region_syncer 11.232s`.
- Also passed: `make gotest GOTEST_ARGS='./pkg/syncer -count=1'` and targeted `make static` for `./pkg/syncer ./tests/server/region_syncer`.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Fix a region syncer issue that could leave a follower waiting on a stream after the PD leader failed to send updates and removed the stream from memory.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stream lifecycle and shutdown: stream replacements now close cleanly, stale streams are reliably removed, shutdown responses are delivered even when sends are blocked, and failures or canceled contexts correctly terminate syncs.

* **Tests**
  * Added tests validating sync shutdown, blocked-send shutdown behavior, send-failure handling, and context-canceled-before-request scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10685?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -486,14 +486,13 @@ func (s *RegionSyncer) sendRegionSyncResponse(
 	sender *regionSyncStream,
 	regions *pdpb.SyncRegionResponse,
 ) bool {
-	failpoint.Inject("regionSyncerSendFail", func(val failpoint.Value) {
-		if targetName, ok := val.(string); ok && targetName == name {
-			err := errors.New("injected region sync send failure")
-			log.Warn("region syncer send data meet error", zap.String("name", name),
-				errs.ZapError(errs.ErrGRPCSend, err))
-			failpoint.Return(false)
-		}
-	})
+	var sendErr error
+	failpoint.InjectCall("regionSyncerSendFail", name, &sendErr)
+	if sendErr != nil {
+		log.Warn("region syncer send data meet error", zap.String("name", name),
+			errs.ZapError(errs.ErrGRPCSend, sendErr))
+		return false
+	}
 
 	resultCh := make(chan error, 1)
 	go func() {
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -486,6 +486,15 @@ func (s *RegionSyncer) sendRegionSyncResponse(
 	sender *regionSyncStream,
 	regions *pdpb.SyncRegionResponse,
 ) bool {
+	failpoint.Inject("regionSyncerSendFail", func(val failpoint.Value) {
+		if targetName, ok := val.(string); ok && targetName == name {
+			err := errors.New("injected region sync send failure")
+			log.Warn("region syncer send data meet error", zap.String("name", name),
+				errs.ZapError(errs.ErrGRPCSend, err))
+			failpoint.Return(false)
+		}
+	})
+
 	resultCh := make(chan error, 1)
 	go func() {
 		resultCh <- sender.stream.Send(regions)
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/docker/go-units"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
@@ -221,6 +223,13 @@ func (s *RegionSyncer) Sync(ctx context.Context, stream pdpb.PD_SyncRegionsServe
 			return err
 		}
 		s.bindStream(request.GetMember().GetName(), stream)
+		defer s.unbindStream(request.GetMember().GetName(), stream)
+		select {
+		case <-ctx.Done():
+			return status.Error(codes.Unavailable, "region syncer stopped")
+		case <-stream.Context().Done():
+			return nil
+		}
 	}
 }
 
@@ -354,6 +363,14 @@ func (s *RegionSyncer) bindStream(name string, stream ServerStream) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.mu.streams[name] = stream
+}
+
+func (s *RegionSyncer) unbindStream(name string, stream ServerStream) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.mu.streams[name] == stream {
+		delete(s.mu.streams, name)
+	}
 }
 
 func (s *RegionSyncer) broadcast(ctx context.Context, regions *pdpb.SyncRegionResponse) {
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -62,6 +62,25 @@ type ServerStream interface {
 	Send(regions *pdpb.SyncRegionResponse) error
 }
 
+type regionSyncStream struct {
+	stream ServerStream
+	done   chan struct{}
+	once   sync.Once
+}
+
+func newRegionSyncStream(stream ServerStream) *regionSyncStream {
+	return &regionSyncStream{
+		stream: stream,
+		done:   make(chan struct{}),
+	}
+}
+
+func (s *regionSyncStream) close() {
+	s.once.Do(func() {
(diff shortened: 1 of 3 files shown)


pkg/syncer/server.go
rleungx: What if before Recv, the ctx is canceled?

okJiang:
Addressed the review comments in 84827760e:

- Wrapped RegionStorage.Close in deferred closures in the new tests.
- Made Sync return when the region syncer context is canceled while waiting for the first Recv.
- Made closeAllClient close local streams before the best-effort Send, and avoid holding the stream map lock while sending.

Validated with:
- make gotest GOTEST_ARGS='./pkg/syncer -run "TestSyncExitsWhenContextCanceledBeforeRequest|TestSyncExitsWhenRegionSyncerStops|TestSyncExitsWhenBroadcastSendFails|TestCloseAllClientClosesStreamsBeforeSend" -count=1'
- make gotest GOTEST_ARGS='./pkg/syncer -count=1'
- make static PACKAGE_DIRECTORIES=./pkg/syncer

pkg/syncer/server.go
```diff
@@ -391,13 +421,13 @@ func (s *RegionSyncer) broadcast(ctx context.Context, regions *pdpb.SyncRegionRe
 		}
 
 		wg.Add(1)
-		go func(name string, sender ServerStream) {
+		go func(name string, sender *regionSyncStream) {
 			defer wg.Done()
-			err := sender.Send(regions)
+			err := sender.stream.Send(regions)
```
rleungx: What if the send is blocked?

okJiang:
Addressed the blocked Send comment in dbbdaa7b3. broadcast now times out a blocked stream send, closes the follower stream, and removes it from the active stream map. Added TestBroadcastClosesStreamWhenSendBlocks and tightened TestCloseAllClientClosesStreamsBeforeSend to wait until Send is actually blocked.

Validated with:
- make gotest GOTEST_ARGS='./pkg/syncer -run "TestBroadcastClosesStreamWhenSendBlocks|TestCloseAllClientClosesStreamsBeforeSend|TestSyncExitsWhenContextCanceledBeforeRequest|TestSyncExitsWhenRegionSyncerStops|TestSyncExitsWhenBroadcastSendFails" -count=1'
- make gotest GOTEST_ARGS='./pkg/syncer -count=1'
- make static PACKAGE_DIRECTORIES=./pkg/syncer
- (cd tests/integrations && make gotest GOTEST_ARGS='./mcs/router -run TestServerTestSuite/TestBasicSync -count=1')

rleungx (approved):

pkg/syncer/server.go
```diff
@@ -220,7 +246,41 @@ func (s *RegionSyncer) Sync(ctx context.Context, stream pdpb.PD_SyncRegionsServe
...
+}
+
+type syncRegionRequestResult struct {
+	request *pdpb.SyncRegionRequest
+	err     error
+}
+
+func recvSyncRegionRequest(ctx context.Context, stream pdpb.PD_SyncRegionsServer) (*pdpb.SyncRegionRequest, error) {
+	resultCh := make(chan syncRegionRequestResult, 1)
+	go func() {
```
bufferflies, line 272: How about using a loop to check the stream result to not create one goroutine?

pkg/syncer/server.go
```diff
@@ -405,10 +480,60 @@ func (s *RegionSyncer) broadcast(ctx context.Context, regions *pdpb.SyncRegionRe
...
+	var sendErr error
+	failpoint.InjectCall("regionSyncerSendFail", name, &sendErr)
+	if sendErr != nil {
+		log.Warn("region syncer send data meet error", zap.String("name", name),
+			errs.ZapError(errs.ErrGRPCSend, sendErr))
+		return false
+	}
+
+	resultCh := make(chan error, 1)
+	go func() {
```
bufferflies, line 498: ditto

bufferflies (commented):
rest lgtm

pkg/syncer/server.go
okJiang: There is no way to do it. stream.Recv() is a synchronous interface.

okJiang: Both `Recv()` and `Send()` are synchronous blocking APIs, so a loop cannot observe `ctx.Done()` while the call itself is blocked.

For this `Recv()` path, the goroutine is needed so `Sync()` can return when the region syncer stops before receiving the next request. `TestSyncExitsWhenContextCanceledBeforeRequest` covers this case.

pkg/syncer/server.go
okJiang: For the `Send()` path, the goroutine is also intentional. `sender.stream.Send()` can block, and the goroutine lets us enforce `sendTimeout` and close the stream instead of letting `broadcast()` wait forever.

`TestBroadcastClosesStreamWhenSendBlocks` covers this behavior. We could reduce the outer broadcast goroutine, but then one slow follower could delay sending to other followers until timeout, so I prefer to keep the current fan-out behavior.

bufferflies (approved):

PR #10689, opened 2026-05-19
server: resync follower region cache reset

### LINKED ISSUE #10667: admin: allow PD followers to serve and reset their local region cache via HTTP APIs
## Enhancement Task

Tracking sub-issue of #10666.

### Background

Today the HTTP region metadata APIs (`/pd/api/v1/region/...`, `/pd/api/v1/regions/...`) and the region cache / storage admin APIs (`DELETE /pd/api/v1/admin/cache/region/{id}`, `DELETE /pd/api/v1/admin/cache/regions`, `DELETE /pd/api/v1/admin/storage/region/{id}`) are all registered on `clusterRouter` in `server/api/router.go`. `clusterRouter` is wrapped by `clusterMiddleware` (`server/api/middleware.go`), which calls `s.GetRaftCluster()`. That returns `nil` unless `s.cluster.IsRunning()`, and `RaftCluster.Start` is only invoked from `campaignLeader` (`server/server.go`). So even when a client sets `PD-Allow-Follower-Handle: true` to bypass the redirector, the request is rejected with `ErrNotBootstrapped` on a PD follower.

The follower's in-memory region cache (`BasicCluster`, populated by the region syncer) is reachable today only through the gRPC Active Follower path (`GetRegion`, `GetPrevRegion`, `GetRegionByID`, `ScanRegions`, `BatchScanRegions`, `QueryRegion` in `server/grpc_service.go`).

Concretely this means:

- We have no HTTP API to inspect a follower's region cache, so there is no way to verify whether a given follower's view is consistent with the leader's.
- `DELETE /admin/cache/region/{id}` and `DELETE /admin/cache/regions` only act on the leader. There is no operator-facing way to drop a stale entry on a follower; even restarting the follower reloads the same data from the local LevelDB region storage.

This gap is exposed by #10666: when a follower's region syncer falls behind the leader's history buffer and recovers, stale entries can persist on the follower indefinitely and leak through Active Follower gRPC reads, with no API path to detect or remediate.

### Goal

Provide HTTP API access to a follower's local region cache so operators can:

1. Inspect a follower's view of one or more regions (read its `BasicCluster` directly, not the leader's view).
2. Drop one or all stale entries from a follower's cache without restarting the process.

### Suggested direction

- Decouple the follower-servable region read paths from `clusterMiddleware` / `RaftCluster.IsRunning()`, e.g. by serving them from `BasicCluster` directly (mirroring the gRPC Active Follower path) when `PD-Allow-Follower-Handle: true` is set.
- Make the region cache admin endpoints (`/admin/cache/region/{id}`, `/admin/cache/regions`) addressable on followers under the same opt-in header, acting on the local cache only.
- Ensure these follower-local paths skip the redirector and skip `clusterMiddleware`'s leader-only check, while still being safe (read-only or local-only mutations).
- Update `pd-ctl` / docs to expose the new behavior.

okJiang:

Issue Number: Close #10667

Stacked on #10682. Please review the new commits on this branch until #10682 is merged.

```commit-message
Allow PD followers to handle the existing region cache reset admin APIs when PD-Allow-Follower-Handle is set.

For follower reset requests, the follower stops its region syncer, removes stale local region cache and region storage entries, resets the sync index to 0, and reconnects to the leader. This forces a full region sync so the follower rebuilds local state from the leader.

The leader-side reset cache behavior is unchanged.

The region syncer now treats startIndex == 0 as full sync, falls back to full sync when requested history is missing, sends history generated during the snapshot phase before binding the stream, retries full sync if the history buffer overflows during catch-up, and marks the follower syncer running only after the full sync completion response.
```

### Check List

Tests

- Unit test
- Integration test

Code changes

- Has persistent data change

### Release note

```release-note
PD followers can reset stale local region cache through region cache admin APIs and automatically resync regions from the leader.
```


server/server.go
```diff
@@ -931,6 +931,85 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+	leader := s.GetLeader()
+	if leader == nil {
+		return errs.ErrLeaderNil.FastGenByArgs()
+	}
+	leaderURLs := leader.GetClientUrls()
+	if len(leaderURLs) == 0 {
+		return errors.New("pd leader has no listen url")
+	}
+
+	syncer := s.cluster.GetRegionSyncer()
```
rleungx, line 954: This reset path needs to be serialized. Two concurrent follower reset requests can both call StopSyncWithLeader and then defer StartSyncWithLeader; StartSyncWithLeader overwrites clientCancel each time, so one of the newly started syncer goroutines can become unreachable by future stops and later resets or leader changes may hang in wg.Wait(). Please guard the whole stop/delete/reset/start sequence with a mutex, or provide an atomic RegionSyncer restart method, and add a concurrent DELETE test.

pkg/syncer/server.go
```diff
@@ -243,10 +246,6 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
 			}
 			return stream.Send(resp)
 		}
-		// do full synchronization
-		if startIndex == 0 {
-			return s.syncFullRegions(ctx, name, stream)
-		}
 		log.Warn("no history regions from index, the leader may be restarted", zap.Uint64("index", startIndex))
 		return nil
```
rleungx, line 413: This branch should trigger a full sync instead of binding the stream and continuing. If the leader has lost the requested history range after restart, the follower can remain permanently incomplete for regions that do not change again, while later keepalives/broadcasts can still make the follower look synced to the local-read API. Please call syncFullRegions here, or otherwise send an explicit full-sync signal, so the follower rebuilds its cache.

server/server.go
```diff
@@ -931,6 +931,85 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+	syncer.StopSyncWithLeader()
+	defer syncer.StartSyncWithLeader(leaderURLs[0])
+
```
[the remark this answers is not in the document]
okJiang: Checked this path against the current code. `ResetFollowerRegionCache` gets a `*pdpb.Member`, which only exposes `ClientUrls`; the `member.Leader.GetListenUrls()` used by `leaderLoop` is a wrapper that returns the same client URLs. Switching this call to `GetListenUrls()` does not compile here. I kept `GetClientUrls()` and aligned the error text to `client url`.

server/server.go
okJiang: Fixed in the latest push. `ResetFollowerRegionCache` now serializes the stop/delete/reset/start sequence with a mutex, and `TestFollowerRegionResetCacheWithNoForward` covers concurrent DELETE requests for `/pd/api/v1/admin/cache/regions`.

pkg/syncer/server.go
okJiang: Fixed in the latest push. When the requested history range is missing and the follower is not already at the leader next index, the leader now falls back to full region sync. Added `TestSyncFallsBackToFullSyncWhenHistoryMissing` for this case.

pkg/syncer/server.go
```diff
@@ -350,6 +349,13 @@ func (*RegionSyncer) syncHistoryRecords(startIndex uint64, records []*core.Regio
...
 func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream pdpb.PD_SyncRegionsServer) error {
 	regions := s.server.GetRegions()
+	if len(regions) == 0 {
+		resp := &pdpb.SyncRegionResponse{
+			Header:     &pdpb.ResponseHeader{ClusterId: keypath.ClusterID()},
+			StartIndex: 0,
+		}
+		return stream.Send(resp)
+	}
 	lastIndex := 0
```
okJiang, line 476: When the FollowerRegion begins a reset and the server calls Sync4Regions, the LastIndex becomes inconsistent with the actual index in the server's history buffer.

How should we handle this?

okJiang: This is the broader recovery problem covered by #10670. To make `LastIndex` fully consistent after a full snapshot, the server needs to catch up history generated during full sync and then send a final empty response with the catch-up index; the client also needs to keep `streamingRunning` false while it is still in full-sync mode. I would keep that full index-recovery flow in #10670 rather than expanding this reset-cache PR further.

server/server.go
```diff
@@ -931,6 +931,85 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+			resetErr = err
+		}
+		s.basicCluster.ResetRegionCache()
+	} else {
+		for _, regionID := range regionIDs {
+			if err := s.deleteFollowerRegion(regionID); resetErr == nil && err != nil {
+				resetErr = err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
okJiang: Fixed in the latest push. The all-regions reset path now scans the local region storage instead of using only `BasicCluster`, and the single-region reset path loads and deletes a storage-only region when it is missing from memory. `TestFollowerRegionResetCacheWithNoForward` now covers storage-only stale regions for both single-region and all-regions reset. Verified with `go test ./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1`, `go test ./pkg/syncer -count=1`, and `git diff --check`.

pkg/syncer/server_test.go
```diff
@@ -333,6 +334,71 @@ func TestSyncExitsWhenContextCanceledBeforeRequest(t *testing.T) {
...
+			Name:       "pd-follower",
+			ClientUrls: []string{"http://127.0.0.1:2379"},
+		},
+	}
+	var resp *pdpb.SyncRegionResponse
+	select {
+	case resp = <-stream.sendCh:
+	case <-time.After(3 * time.Second):
+		re.FailNow("expected full sync response")
+	}
```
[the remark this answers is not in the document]
okJiang: Fixed in the latest push. Added `defer cancel()` immediately after creating the context so the sync goroutine is signaled on failure paths too. Verified with `go test ./pkg/syncer -run TestSyncFallsBackToFullSyncWhenHistoryMissing -count=1` and `git diff --check`.

server/server.go
```diff
@@ -931,6 +933,167 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+	syncer.ResetHistoryIndex(0)
+
```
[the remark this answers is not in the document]
okJiang: Fixed by keeping the pre-delete flush for pending batched saves and adding a final flush after storage deletions before `ResetHistoryIndex(0)` and the syncer restart. That prevents stale deleted regions from being reloaded from storage after reset. Verified with `go test ./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1`, `go test ./pkg/syncer -count=1`, `make swagger-spec`, and `git diff --check`.

server/server.go
```diff
@@ -931,6 +933,167 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+		return errs.ErrLeaderNil.FastGenByArgs()
+	}
+	leaderURLs := leader.GetClientUrls()
+	if len(leaderURLs) == 0 {
+		return errors.New("pd leader has no client url")
+	}
+
+	syncer := s.cluster.GetRegionSyncer()
+	syncer.StopSyncWithLeader()
+	defer syncer.StartSyncWithLeader(leaderURLs[0])
```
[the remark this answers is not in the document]
okJiang: Fixed by guarding the follower syncer Start/Stop path in `leaderLoop` with `followerRegionResetMu`, so a manual reset and leader handoff cannot interleave syncer lifecycle changes. Verified with `go test ./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1`, `go test ./pkg/syncer -count=1`, `make swagger-spec`, and `git diff --check`.

server/api/admin.go
```diff
@@ -69,6 +69,14 @@ func (h *adminHandler) DeleteRegionCache(w http.ResponseWriter, r *http.Request)
...
 		return
 	}
+	if isFollowerSyncedClusterRequest(r) {
+		if err = h.svr.ResetFollowerRegionCache(regionID); err != nil {
+			h.rd.JSON(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+		h.rd.JSON(w, http.StatusOK, "The region is removed from follower cache and the follower starts to resync regions from leader.")
+		return
+	}
```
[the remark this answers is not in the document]
okJiang: Fixed by updating the follower reset API swagger annotations to mention server/follower cache and the new 500 response. Verified with `make swagger-spec`, `go test ./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1`, `go test ./pkg/syncer -count=1`, and `git diff --check`.

okJiang:
Addressed the CodeRabbit outside-diff comment in https://github.com/tikv/pd/pull/10689#pullrequestreview-4326532775.

DELETE follower reset requests no longer require `RegionSyncer.IsRunning()`, while GET follower reads still do. I also added coverage in `TestFollowerRegionResetCacheWithNoForward` for reset while the follower syncer is stopped.

Verified with `go test ./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1` and `git diff --check`.

pkg/syncer/server.go
```diff
@@ -303,12 +306,8 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
 			return stream.Send(resp)
 		}
-		// do full synchronization
-		if startIndex == 0 {
-			return s.syncFullRegions(ctx, name, stream)
-		}
-		log.Warn("no history regions from index, the leader may be restarted", zap.Uint64("index", startIndex))
-		return nil
+		log.Warn("no history regions from index, fall back to full sync", zap.Uint64("index", startIndex))
+		return s.syncFullRegions(ctx, name, stream)
```
rleungx: When this falls back to full sync, the follower can mark the syncer as running after the first response batch. For large clusters, follower reads may see a partial region cache before full sync finishes. Can we add an explicit completion signal or keep follower local reads disabled until the full sync is done?

server/api/middleware.go
```diff
@@ -107,38 +114,60 @@ func newClusterMiddleware(s *server.Server, opts ...clusterMiddlewareOption) clu
...
+	if r.Header.Get(apiutil.PDAllowFollowerHandleHeader) == "" ||
+		m.s.GetMember().IsServing() {
+		return nil
+	}
+	switch r.Method {
+	case http.MethodGet:
+		if !m.allowFollowerSyncedRegion {
+			return nil
+		}
+	case http.MethodDelete:
```
rleungx, line 142: This reuses PD-Allow-Follower-Handle for a DELETE operation, but the existing header is read-oriented. Existing clients or proxies that already attach it could unexpectedly reset follower region cache. Could we use a dedicated opt-in signal or add compatibility coverage for this behavior?

server/server.go
```diff
@@ -931,6 +933,170 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+
+	var resetErr error
+	if err := s.storage.Flush(); err != nil {
+		resetErr = errors.Wrap(err, "flush follower region storage")
+	}
+	if len(regionIDs) == 0 {
+		regions, err := s.loadFollowerRegionStorage()
```
rleungx: This loads all local regions into memory before deleting them. On large clusters, reset may use a lot of memory or take too long. Can we delete page by page while scanning LoadRange instead?

okJiang: Fixed in b3c130ecc. The all-region follower reset now scans local region storage page by page with `LoadRange` and deletes each loaded page immediately, instead of loading all local regions into memory first.

Verified with:
- `make gotest GOTEST_ARGS="./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1"`

pkg/syncer/server.go
okJiang: Fixed in b3c130ecc. Full sync now sends an empty completion response after all full-sync batches, and the follower client keeps local follower reads disabled until that completion response is received.

Coverage added in `TestClientWaitsForFullSyncCompletionBeforeRunning` and the existing fallback full sync test. Verified with:
- `make gotest GOTEST_ARGS="./pkg/syncer -run 'TestSyncFallsBackToFullSyncWhenHistoryMissing|TestClientWaitsForFullSyncCompletionBeforeRunning' -count=1"`
- `make gotest GOTEST_ARGS="./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1"`

server/api/middleware.go
okJiang: Fixed in b3c130ecc. The follower reset path now requires an explicit `PD-Follower-Region-Reset` header in addition to `PD-Allow-Follower-Handle`, so existing read-oriented callers/proxies that only carry the old header will not trigger a follower reset.

Coverage was added in the existing follower reset test. Verified with:
- `make gotest GOTEST_ARGS="./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1"`

pkg/utils/apiutil/apiutil.go
```diff
@@ -57,6 +57,8 @@ const (
 	PDRedirectorHeader = "PD-Redirector"
 	// PDAllowFollowerHandleHeader is used to mark whether this request is allowed to be handled by the follower PD.
 	PDAllowFollowerHandleHeader = "PD-Allow-follower-handle" // #nosec G101
+	// PDFollowerRegionResetHeader is used to explicitly allow a follower PD to reset its local region cache.
+	PDFollowerRegionResetHeader = "PD-Follower-Region-Reset" // #nosec G101
```
rleungx: I think we can avoid introducing a second header here. `PD-Allow-Follower-Handle` can be the unified opt-in meaning that the client allows the current follower PD to handle the external request locally instead of forwarding it to the leader. As long as only explicitly whitelisted endpoints honor that header and we document the broader semantics, follower region cache reset can reuse the existing header.

okJiang:
Fixed `pull-error-log-review` in ac66d235e by changing the new completion-send log from error to warning. Verified with `make gotest GOTEST_ARGS="./pkg/syncer -run 'TestSyncFallsBackToFullSyncWhenHistoryMissing|TestClientWaitsForFullSyncCompletionBeforeRunning' -count=1"` and `git diff --check`.

The next-gen failures are in unrelated suites (`TestConfigTestSuite/TestTTLConflict`, `TestAffinityHandlerTestSuite/TestAffinityListWithEmptyID`, `TestHealthTestSuite/TestHealthSlice`, `TestResourceManagerWatcherMatrix/TestWatcherRecoversAfterCompaction`, and `TestKeyspaceGroupTestSuite/TestUpdateMemberWhenRecovery`), so rerunning the failed jobs.

/test pull-error-log-review
/test pull-unit-test-next-gen-2
/test pull-unit-test-next-gen-3

pkg/utils/apiutil/apiutil.go
okJiang: Fixed in 7e424559c. Removed the extra `PD-Follower-Region-Reset` header and kept `PD-Allow-Follower-Handle` as the unified follower-local opt-in. The DELETE reset path is still limited by the dedicated `regionResetRouter` / `withFollowerRegionReset()` whitelist, and the header comment now documents the broader local-handle semantics.

Updated the existing follower reset test to cover the no-header path: it only removes the server cache and does not restart the follower syncer or delete local region storage; the opt-in header path still performs the follower reset/resync.

Verified with:
- `make gotest GOTEST_ARGS="./tests/server/api -tags without_dashboard -run TestFollowerRegionResetCacheWithNoForward -count=1"`
- `git diff --check`

okJiang:
Addressed the Codecov patch coverage failure in a366025c by adding focused `server` package tests for follower region reset storage cleanup and error branches.

Verified with:
- `make gotest GOTEST_ARGS="./server -run 'TestResetFollowerRegionCacheRequiresRegionStorage|TestDeleteFollowerRegion|TestDeleteFollowerRegionReturnsStorageErrors|TestDeleteFollowerRegionStorage|TestDeleteFollowerRegionStorageReturnsStorageErrors|TestParseRegionIDFromStorageKey' -count=1"`
- `make gotest GOTEST_ARGS="./server -run 'TestResetFollowerRegionCacheRequiresRegionStorage|TestDeleteFollowerRegion|TestDeleteFollowerRegionReturnsStorageErrors|TestDeleteFollowerRegionStorage|TestDeleteFollowerRegionStorageReturnsStorageErrors|TestParseRegionIDFromStorageKey' -count=1 -covermode=atomic -coverprofile=/tmp/pd-pr10689-server.cover"`
- `git diff --check`

okJiang:
Checked the latest failures. `report-coverage` failed only because chunk 9 did not produce `covprofile_9`. Chunk 9 failed in `mcs/resourcemanager TestWatchWithSingleGroupByKeyspace`, which is unrelated to this follower region reset change. The realcluster failure is `TestRebootPD/TestReloadLabel`, where setting the reserved `engine` label returned 400, also unrelated to this PR.

Rerunning the failed jobs.

/test pull-integration-realcluster-test

okJiang:
Correction: `/test pull-integration-realcluster-test` has been requested for the Prow/Jenkins failure. GitHub Actions failed-job rerun is not available from this account (`gh run rerun --failed` requires repository admin rights), so chunk 9 / report-coverage will need a maintainer rerun or another push if GitHub Actions does not rerun automatically.

okJiang:
Fixed the repeated `pull-integration-realcluster-test` failure in d54a432d2. `TestReloadLabel` now skips the reserved `engine` label when it rebuilds the label map for `SetStoreLabels`; that label is managed by TiKV/TiFlash and the store label API correctly rejects user updates to it.

Verified with:
- `go test ./realcluster -run TestRebootPD/TestReloadLabel -count=0 -tags deadlock` from `tests/integrations`
- `git diff --check`

/test pull-integration-realcluster-test

okJiang:
Checked the latest GitHub Actions failure. `chunks (9, Microservice Integration(!TSO))` failed in `mcs/resourcemanager TestResourceManagerClientFailover` with `failover_test resource group does not exist`; this is outside the follower region reset path changed by this PR. `report-coverage` failed only because chunk 9 did not produce its coverage profile.

I cannot rerun the failed GitHub Actions jobs from this account because GitHub requires repository admin rights for `gh run rerun --failed`.

okJiang:
Checked the latest Prow/Jenkins failures from https://github.com/tikv/pd/pull/10689#issuecomment-4499002754.

- `pull-integration-realcluster-test`: the previous `TestReloadLabel` failure is fixed and now passes. The new failure is `TestScheduler/TestRegionLabelDenyScheduler` timing out while waiting for scheduler state, which is outside this PR's follower region reset path.
- `pull-unit-test-next-gen-3`: failed in `mcs/tso TestTSOKeyspaceGroupManagerSuite/TestWatchFailed`, also outside the follower region reset path.

Rerunning both required jobs.

/test pull-unit-test-next-gen-3
/test pull-integration-realcluster-test

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499237949. It failed in `tests/integrations/tso TestLegacyTSOConsistencySuite` with a 5-minute timeout while `RunInitialServers` was starting the test PD cluster. This is outside the follower region reset path changed by this PR, and it is a different failure from the previous next-gen-3 run.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499329805. It failed in `client/resource_group/controller TestQPS` with an observed QPS overshoot (`23600` greater than expected `1000`) in the `concurrency=1000,reserveN=10,limit=400000` case. This is outside the follower region reset path changed by this PR, and it is a different failure from the previous next-gen-3 runs.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499437563. It failed in `tests/integrations/tso TestMicroserviceTSOClientSuite` with a 5-minute timeout while `RunInitialServers` was starting the test PD cluster. This is outside the follower region reset path changed by this PR, and it is a different failure from the previous next-gen-3 runs.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499600838. It failed in `tests/integrations/mcs/resourcemanager TestServiceLimitTestSuite/TestKeyspaceServiceLimit` because the measured RU was `111.22605850000001` while the expected value was `100 ± 10`. This suite is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/mcs/resourcemanager`.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499684009. It failed in `tests/integrations/client TestTSOFollowerProxy`: the test timed out waiting for a `server not started` error, then the failpoint callback was blocked on an unbuffered channel during cleanup. This is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/client/client_test.go` or the `delayStartServer` failpoint path.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499792551. It failed in `client/resource_group/controller TestQPS`, with `1310` greater than the expected upper bound `1000` in the `concurrency=1000,reserveN=10,limit=400000` case. This is outside the follower region reset path changed by this PR, and the PR diff does not touch `client/resource_group/controller`.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4499920965. It failed in `tests/integrations/tso TestLegacyTSOClientSuite/TestTSONotLeaderWhenRebaseErr`: `ResignLeaderWithRetry` returned `[PD:etcd:ErrEtcdMoveLeader] etcdserver: request timed out, leader transfer took too long`. The PR diff does not touch `tests/integrations/tso/client_test.go` or the TSO client/rebase path; the only leader-loop change here adds an uncontended lock around region syncer start/stop, and this test does not exercise follower region reset.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4500184812. It failed in `tests/integrations/mcs/resourcemanager TestResourceManagerClientTestSuite`, specifically `standalone-resource-manager-with-client-discovery/TestResourceGroupRUConsumption`. This is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/mcs/resourcemanager` or the resource manager client/controller code.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4500421042. It failed in `tests/integrations/client TestHTTPClientTestSuite/TestRedirectWithMetrics` with a data race between `client/metrics.InitAndRegisterMetrics()` and `client/servicediscovery.getClusterInfo()`. This is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/client`, `client/metrics`, or `client/servicediscovery`.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4500592465. It failed in `tests/integrations/client TestClientLeaderChange`: the test timed out after 5 minutes while the embedded etcd member repeatedly failed to publish member attributes with `etcdserver: request timed out`. This is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/client` or the client leader-change test path.

Also checked the outside-diff CodeRabbit comment in https://github.com/tikv/pd/pull/10689#pullrequestreview-4326532775. The current code already allows DELETE follower reset requests through without requiring `RegionSyncer.IsRunning()`; the `IsRunning()` check is only applied to GET follower local-read requests, so no code change is needed for that comment.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4500734666. It failed in `tests/integrations/mcs/keyspace TestKeyspaceGroupTestSuite/TestUpdateMemberWhenRecovery` because the wait condition in `tso_keyspace_group_test.go:745` was never satisfied, with repeated `ErrKeyspaceGroupModRevisionStale` / TSO service discovery errors.

This is outside the follower region reset path changed by this PR, and the PR diff does not touch `tests/integrations/mcs/keyspace`, TSO keyspace group manager, or TSO service discovery code.

Rerunning the failed required job.

/test pull-unit-test-next-gen-3

pkg/syncer/server.go
okJiang: If we don't fix this issue, won't there be a problem once this PR is merged? 

Even after the region cache is reset, subsequent region synchronization will still have issues. Isn't this effectively a bug?

okJiang: Fixed in 2a3f13d5c. The leader now records the history index before full sync, sends any region history generated while the snapshot is being sent, and only completes the full sync after the catch-up records are sent and the downstream stream is bound.

Verified with:
- `git diff --check`
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`

okJiang:
Checked the latest `pull-error-log-review` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4504360928. It failed because 2a3f13d5c added a new `log.Error("failed to send sync region response", ...)`, which requires log approver approval.

Fixed in b778b7abf by lowering that new full-sync send-failure log to `log.Warn`, matching the existing non-critical send-failure handling in this path.

Verified with:
- `git diff --check`
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4504437335. It failed in `pkg/mcs/tso/server TestTSOKeyspaceGroupManagerSuite/TestWatchFailed` with repeated etcd context-canceled errors during test shutdown.

Also checked the current GitHub Actions failure in `chunks (9, Microservice Integration(!TSO))`: it failed in `tests/integrations/mcs/scheduling TestMeta/TestStoreWatch` because the wait condition at `tests/integrations/mcs/scheduling/meta_test.go:106` was never satisfied. The `report-coverage` failure is caused by the missing `covprofile_9` artifact from that failed chunk.

Both failures are outside the follower region reset / syncer path changed by this PR; the PR diff does not touch `pkg/mcs/tso/server`, `tests/integrations/mcs/scheduling`, or the scheduling metadata watch path.

Rerunning the failed required Prow job and the failed GitHub Actions jobs.

/test pull-unit-test-next-gen-3

server/server.go
```diff
@@ -931,6 +933,152 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+	syncer := s.cluster.GetRegionSyncer()
+	syncer.StopSyncWithLeader()
+	defer syncer.StartSyncWithLeader(leaderURLs[0])
+
+	var resetErr error
+	if err := s.storage.Flush(); err != nil {
+		resetErr = errors.Wrap(err, "flush follower region storage")
+	}
+	if len(regionIDs) == 0 {
+		if err := s.deleteFollowerRegionStorage(); resetErr == nil && err != nil {
```
rleungx, line 964: Do we need to delete it? If we have 10M region, how long will it take?

okJiang:
Correction on the GitHub Actions rerun: the Prow rerun command above was posted, but `gh run rerun --failed` for the failed GitHub Actions run was rejected because this token does not have repository admin rights. No code change is needed for that GHA failure based on the current logs; it is the same unrelated `tests/integrations/mcs/scheduling TestMeta/TestStoreWatch` wait-condition failure described above.

server/server.go
okJiang: Fixed in d095867e8.

The all-region reset still needs to delete local region storage so storage-only stale regions cannot be reloaded later. To reduce the 10M-region cost, it now scans region keys and deletes by parsed region ID, without loading and decoding every region value before deletion.

Verified with:
- `git diff --check`
- `make gotest GOTEST_ARGS='./server -run "TestDeleteFollowerRegionStorage|TestDeleteFollowerRegionStorageReturnsStorageErrors|TestDeleteFollowerRegion|TestResetFollowerRegionCacheRequiresRegionStorage|TestParseRegionIDFromStorageKey" -count=1'`

okJiang:
Checked the latest Prow failures in https://github.com/tikv/pd/pull/10689#issuecomment-4504526981.

- `pull-unit-test-next-gen-2` failed in `tests/server/apiv2/handlers TestAffinityHandlerTestSuite/TestAffinityListWithEmptyID`: the microservice run returned an empty affinity group map for `ids=&ids=group-2`.
- `pull-unit-test-next-gen-3` failed in `tests/integrations/client TestFollowerForwardAndHandleTestSuite/TestGetTsoByFollowerForwarding2`: `ResignLeader` failed because etcd leader transfer timed out.

The latest PR commit only changed `server/server.go` and `server/server_test.go` to avoid loading region values during follower storage reset. These failures are outside that path, so rerunning the two failed Prow jobs.

/test pull-unit-test-next-gen-2
/test pull-unit-test-next-gen-3

server/server.go
okJiang: @okJiang can we test it locally and report some test results here

okJiang: Local scale test result with the in-memory storage backend:

- 100,000 regions: old path delete cost 74.67 ms, new path delete cost 41.45 ms, about 1.80x faster.
- 1,000,000 regions: old path delete cost 811.82 ms, new path delete cost 466.39 ms, about 1.74x faster.

Command used with a temporary local-only test, not committed:
- `PD_FOLLOWER_REGION_DELETE_SCALE=100000 go test ./server -run '^TestFollowerRegionStorageDeleteScaleLocalOnly$' -count=1 -v`
- `PD_FOLLOWER_REGION_DELETE_SCALE=1000000 go test ./server -run '^TestFollowerRegionStorageDeleteScaleLocalOnly$' -count=1 -v`

This is not a real disk/etcd latency benchmark. It only verifies the changed path: the reset is still O(N), but it no longer does one `LoadRegion` read and protobuf unmarshal per region before deletion. For 10M regions it is still a linear admin recovery operation, but the avoidable per-region value load is removed.

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4504597799. It failed in `pkg/mcs/tso/server TestTSOKeyspaceGroupManagerSuite/TestWatchFailed`: the assertion at `tests/integrations/mcs/tso/keyspace_group_manager_test.go:181` expected keyspace group `0x9d` but got `0x0`.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This TSO keyspace group watch failure is outside that path, so rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

server/server.go
```diff
@@ -931,6 +933,145 @@ func (s *Server) GetBasicCluster() *core.BasicCluster {
...
+		if err != nil {
+			return errors.Wrap(err, "load follower regions from local storage")
+		}
+		for _, key := range keys {
+			regionID, err := parseRegionIDFromStorageKey(key)
+			if err != nil {
+				return err
+			}
+			meta := &metapb.Region{Id: regionID}
+			if err := s.deleteFollowerRegionMeta(meta); err != nil && firstErr == nil {
```
okJiang: Deleting regions one by one feels a bit slow. Is there a way to delete everything at once, or perhaps just recreate a new one from scratch?

okJiang: Fixed in 776055e12.

The full reset path still scans local region storage page by page, but each page is now deleted in one storage transaction instead of issuing one `DeleteRegion` write per region. This keeps memory bounded and reduces the write cost to one batch per `LoadRange` page.

I did not recreate the storage from scratch because current region storage is shared by the running server/syncer and does not expose a safe swap/recreate path. There is also no generic range-delete API on the storage interface, so batching the existing page scan is the minimal safe change here.

Verified with:
- `git diff --check`
- `make gotest` for the follower-region reset/delete server tests

okJiang:
Checked the latest Prow failures in https://github.com/tikv/pd/pull/10689#issuecomment-4504862267.

- `pull-unit-test-next-gen-2` failed in `tests/server/api TestFollowerRegionAPIWithNoForward` by timing out while starting the test cluster. The test did not reach the follower-region API assertions. I reran the same test locally with `go test -tags nextgen,without_dashboard ./tests/server/api -run TestFollowerRegionAPIWithNoForward -count=1`, and it passed.
- `pull-unit-test-next-gen-3` failed in `tests/integrations/client TestConfigTTLAfterTransferLeader` by timing out while starting the test cluster / transferring leadership. I reran the same test locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./client -run TestConfigTTLAfterTransferLeader -count=1`, and it passed.

The latest PR commit only changed the follower region storage reset path in `server/server.go` and its unit tests. These two failures are startup/leader-transfer timeouts outside that changed path, so no code change is needed here. Rerunning the failed Prow jobs.

/test pull-unit-test-next-gen-2
/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4504939078.

It failed in `tests/integrations/tso TestMicroserviceTSOServer` because the test timed out while starting the microservice TSO test cluster. The test did not reach follower region reset or syncer code. I reran the same test locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./tso -run TestMicroserviceTSOServer -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This TSO microservice startup timeout is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505008810.

It failed in `tests/integrations/mcs/resourcemanager TestResourceManagerClientTestSuite/standalone-resource-manager-with-client-discovery/TestResourceGroupController`: one timing assertion observed `492.485574ms`, above the `300ms` limit. I reran the same subtest locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./mcs/resourcemanager -run 'TestResourceManagerClientTestSuite/standalone-resource-manager-with-client-discovery/TestResourceGroupController$' -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This resource manager timing failure is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505083875.

It failed in `tests/integrations/client TestUnavailableTimeAfterLeaderIsReady` because the test timed out while starting the test cluster. The test did not reach follower region reset or syncer code. I reran the same test locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./client -run TestUnavailableTimeAfterLeaderIsReady -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This client cluster-startup timeout is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505217744.

It failed again in `tests/integrations/mcs/tso TestTSOKeyspaceGroupManagerSuite/TestWatchFailed`: the assertion at `tests/integrations/mcs/tso/keyspace_group_manager_test.go:181` expected keyspace group `0x90` but got `0x0`, with `ErrKeyspaceGroupModRevisionStale` in the log. I reran the same subtest locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./mcs/tso -run 'TestTSOKeyspaceGroupManagerSuite/TestWatchFailed$' -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This TSO keyspace-group watch failure is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505332510.

It failed in `tests/integrations/mcs/resourcemanager TestSwitchModeDuringWorkload/pd-to-standalone`: the wait condition at `tests/integrations/mcs/resourcemanager/resource_manager_test.go:492` was never satisfied after switching resource-manager mode. I reran the same subtest locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./mcs/resourcemanager -run 'TestSwitchModeDuringWorkload/pd-to-standalone$' -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This resource-manager mode-switch workload failure is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505431721.

It failed in `tests/integrations/client TestRouterClientEnabledSuite` because the test timed out while starting the initial PD test cluster; the log shows repeated etcd peer connection failures and `etcdserver: request timed out` before the suite setup finished. I reran the same suite locally under `tests/integrations` with `go test -tags nextgen,without_dashboard ./client -run TestRouterClientEnabledSuite -count=1`, and it passed.

The latest PR commit only changed `server/server.go` and `server/server_test.go` for follower region storage reset deletion. This client test-cluster startup timeout is outside that changed path, so no code change is needed here. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

pkg/syncer/server.go
okJiang: What happens if the local History Buffer overflows during a full data synchronization? Specifically, what should be done if the number of changes during this period exceeds the length of the local buffer?

okJiang: Covered in b47a9ebe0.

If the leader loses the catch-up history that was recorded at the start of a full sync, it must not bind the follower stream with a partial cache. The current path detects `catchUpIndex < history.firstIndex()` and restarts full sync from a fresh snapshot, so the follower only marks local reads as running after it receives the final completion response from a complete sync attempt.

I added `TestFullSyncRestartsWhenHistoryBufferOverflowsDuringCatchUp` with a one-entry history buffer. The test forces history overflow while the first full-sync response is blocked, then verifies that the leader sends a restarted full snapshot and only then sends the completion response.

Verified with:
- `git diff --check`
- `make gotest GOTEST_ARGS='./pkg/syncer -run "TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncRestartsWhenHistoryBufferOverflowsDuringCatchUp|TestClientWaitsForFullSyncCompletionBeforeRunning" -count=1'`

okJiang:
Checked the latest `pull-unit-test-next-gen-3` failure from https://github.com/tikv/pd/pull/10689#issuecomment-4505526050.

It failed in `client/resource_group/controller TestQPS`: the `concurrency=1000,reserveN=10,limit=400000` case observed RU `450430`, above the test tolerance. I reran the same test locally under `client` with `go test ./resource_group/controller -run TestQPS -count=1`, and it passed.

This QPS limiter test is outside the follower region reset / syncer full-sync path changed by this PR. The new commit `b47a9ebe0` only adds syncer test coverage for the history-buffer overflow review question. Rerunning the failed Prow job.

/test pull-unit-test-next-gen-3

rleungx:
Is it ready for review?

okJiang:
> Is it ready for review?

I am preparing more fixes related to the region syncer. I think we can focus on the region syncer first, and once it can provide sufficiently safe sync capabilities, we can then move forward with the reset region cache PR.

pkg/syncer/client.go
```diff
@@ -107,6 +112,9 @@ func (s *RegionSyncer) handleRegionSyncResponse(ctx context.Context, resp *pdpb.
 	regions := resp.GetRegions()
 	buckets := resp.GetBuckets()
 	regionLeaders := resp.GetRegionLeaders()
+	if !s.IsRunning() && resp.GetStartIndex() == 0 && len(regions) > 0 {
+		*fullSyncing = true
+	}
```
JmPotato:
```suggestion
	*fullSyncing = !s.IsRunning() && resp.GetStartIndex() == 0 && len(regions) > 0
```

pkg/syncer/client.go
```diff
@@ -85,7 +90,7 @@ func (s *RegionSyncer) syncRegion(ctx context.Context, conn *grpc.ClientConn) (C
 
 var regionGuide = core.GenerateRegionGuideFunc(false)
 
-func (s *RegionSyncer) handleRegionSyncResponse(ctx context.Context, resp *pdpb.SyncRegionResponse, bc *core.BasicCluster, regionStorage storage.Storage) bool {
+func (s *RegionSyncer) handleRegionSyncResponse(ctx context.Context, resp *pdpb.SyncRegionResponse, bc *core.BasicCluster, regionStorage storage.Storage, fullSyncing *bool) bool {
```
JmPotato: What about returning two named boolen varibales rather than using a pointer?

pkg/syncer/client.go
```diff
@@ -242,6 +257,7 @@ func (s *RegionSyncer) StartSyncWithLeader(addr string) {
 				continue
 			}
 			log.Info("server starts to synchronize with leader", zap.String("server", s.server.Name()), zap.String("leader", s.server.GetLeader().GetName()), zap.Uint64("request-index", s.history.getNextIndex()))
+			fullSyncing := false
```
rleungx, line 265: If the stream fails after snapshot batches were applied but before the full-sync completion response, the next retry can send the snapshot row count as the history index and continue incremental replay from an unrelated leader index.

pkg/syncer/client.go
okJiang: Changed in 7b0c7f9d7. `handleRegionSyncResponse` now returns `(handled, nextFullSyncing)` instead of mutating a bool pointer, and the caller updates the stream-local full-sync state explicitly after each response.

Verified with:
- `go test ./pkg/syncer -run 'TestHandleRegionSyncResponseSkipsErrorResponse|TestClientWaitsForFullSyncCompletionBeforeRunning|TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncDisconnectsWhenHistoryBufferOverflowsDuringCatchUp' -count=1`
- `go test ./pkg/syncer -count=1`
- `git diff --check`

pkg/syncer/client.go
okJiang: Fixed in 7b0c7f9d7. The client now keeps the retry history index at 0 while a full sync is in progress and only resets it to the leader history index when it receives the empty completion response. Snapshot batches are still applied to cache/storage, but they no longer advance local history, so a stream failure before completion retries full sync instead of requesting incremental replay from a snapshot row count.

Verified with:
- `go test ./pkg/syncer -run 'TestHandleRegionSyncResponseSkipsErrorResponse|TestClientWaitsForFullSyncCompletionBeforeRunning|TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncDisconnectsWhenHistoryBufferOverflowsDuringCatchUp' -count=1`
- `go test ./pkg/syncer -count=1`
- `git diff --check`

pkg/syncer/client.go
okJiang: Addressed in 7b0c7f9d7, but not with the direct assignment form. A direct assignment would set `fullSyncing` back to false on later full-sync responses whose `StartIndex` is not 0, before the empty completion response arrives. The updated code only detects the start of full sync once, carries that state through return values, and clears it only on the empty completion response.

Verified with:
- `go test ./pkg/syncer -run 'TestHandleRegionSyncResponseSkipsErrorResponse|TestClientWaitsForFullSyncCompletionBeforeRunning|TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncDisconnectsWhenHistoryBufferOverflowsDuringCatchUp' -count=1`
- `go test ./pkg/syncer -count=1`
- `git diff --check`

pkg/syncer/client.go
```diff
@@ -147,16 +163,23 @@ func (s *RegionSyncer) handleRegionSyncResponse(ctx context.Context, resp *pdpb.
...
-		if err == nil {
+		if err == nil && !nextFullSyncing {
 			s.history.record(region)
 		}
 		for _, old := range overlaps {
 			_ = regionStorage.DeleteRegion(old.GetMeta())
 		}
 	}
+	if nextFullSyncing {
+		if len(regions) == 0 {
```
JmPotato: **Rolling-upgrade / mixed-version concern.** A follower only leaves the `fullSyncing` state through this empty-completion branch (`len(regions) == 0`), which relies on the leader emitting the trailing `StartIndex: nextIndex` completion frame that this PR adds to `syncFullRegionsLocked` (server.go).

During a rolling PD upgrade, a new follower may sync from a not-yet-upgraded leader that doesn't send that completion frame. The follower would then enter `fullSyncing == true` on the first `StartIndex == 0 && len(regions) > 0` response and never flip `streamingRunning` back to `true` here — staying not-ready until the stream errors/EOFs and it retries against an upgraded leader.

Could you confirm the mixed-version behavior? Specifically: (1) the response sequence a previous-release leader sends for `startIndex == 0`, and (2) that the follower reliably recovers via retry rather than getting stuck not-ready for an extended window. If it's only reachable transiently during the upgrade it may be acceptable, but a regression/integration test for the new-follower ↔ old-leader path would be reassuring.

pkg/syncer/client.go
```diff
@@ -85,16 +90,31 @@ func (s *RegionSyncer) syncRegion(ctx context.Context, conn *grpc.ClientConn) (C
...
 			zap.String("error-type", syncErr.GetType().String()),
 			zap.String("error-message", syncErr.GetMessage()))
-		return false
+		return false, nextFullSyncing
+	}
+	stats := resp.GetRegionStats()
+	regions := resp.GetRegions()
+	buckets := resp.GetBuckets()
+	regionLeaders := resp.GetRegionLeaders()
+	startFullSync := !nextFullSyncing && !s.IsRunning() && resp.GetStartIndex() == 0 && len(regions) > 0
```
JmPotato: **Readability.** The `fullSyncing` / `nextFullSyncing` / `startFullSync` interplay is hard to follow, mainly because `nextFullSyncing` is initialized to the *current* state, mutated mid-function, and then returned as the *next* state — so the same name means different things on different lines. The index-reset guard `(startFullSync || !nextFullSyncing || len(regions) == 0)` below is also unnamed; it really only asks "is this frame's StartIndex a real history index?".

A behavior-equivalent refactor that I found easier to follow:

- Use an immutable `inFullSync := fullSyncing || startFullSync` instead of mutating `nextFullSyncing`.
- Give the untrusted case a name and guard on it:

```go
// During a full sync, intermediate batches carry a positional offset, not a history index.
isPositionalBatch := inFullSync && !startFullSync && len(regions) > 0
if !isPositionalBatch && s.history.getNextIndex() != resp.GetStartIndex() {
    ...
    s.history.resetWithIndex(resp.GetStartIndex())
}
```

- Record history with `!inFullSync`.
- Compute the result once at the end, which drops the nested trailing `if nextFullSyncing { ... }` branch and the duplicated `streamingRunning.Store(true)` / `return`:

```go
nextFullSyncing = inFullSync && len(regions) > 0
if !nextFullSyncing {
    // full sync completed, or incremental sync caught up
    s.streamingRunning.Store(true)
}
return true, nextFullSyncing
```

The variables then flow in one direction with fixed meanings (`fullSyncing` → `startFullSync` → `inFullSync` → `nextFullSyncing`), and the loop body only swaps `!nextFullSyncing` → `!inFullSync`, so the diff stays small. Non-blocking, just a suggestion.

JmPotato (commented):
Two follow-up notes from reviewing the follower full-sync path: a possible mixed-version (rolling-upgrade) edge case, and a readability refactor for `handleRegionSyncResponse`. Both are non-blocking.

pkg/syncer/client.go
okJiang: Refactored in fa978196d. `handleRegionSyncResponse` now uses `startFullSync`, `inFullSync`, and `isPositionalBatch` so the state flow is explicit: full-sync intermediate data frames are treated as positional batches, not reusable history-index frames. The result state is computed once at the end as `nextFullSyncing = inFullSync && len(regions) > 0`.

Verified with:
- `go test ./pkg/syncer -run 'TestClientWaitsForFullSyncCompletionBeforeRunning|TestClientFinishesFullSyncWithOldLeaderKeepAlive|TestHandleRegionSyncResponseSkipsErrorResponse|TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncDisconnectsWhenHistoryBufferOverflowsDuringCatchUp' -count=1`
- `go test ./pkg/syncer -count=1`
- `git diff --check`

pkg/syncer/client.go
okJiang: Covered in fa978196d. I checked the previous-release server path: for `startIndex == 0`, it sends full-snapshot batches with positional `StartIndex` values and does not send the new trailing completion frame. After the full sync returns, the stream is bound and an old leader can still send the next empty keepalive frame with the stream send index.

I added `TestClientFinishesFullSyncWithOldLeaderKeepAlive` to cover that mixed-version sequence. The test verifies that the follower stays in full-sync mode while receiving non-empty positional frames, does not advance local history from those frames, and exits full sync when it receives the old-leader empty keepalive with the final index.

Verified with:
- `go test ./pkg/syncer -run 'TestClientWaitsForFullSyncCompletionBeforeRunning|TestClientFinishesFullSyncWithOldLeaderKeepAlive|TestHandleRegionSyncResponseSkipsErrorResponse|TestSyncFallsBackToFullSyncWhenHistoryMissing|TestFullSyncDisconnectsWhenHistoryBufferOverflowsDuringCatchUp' -count=1`
- `go test ./pkg/syncer -count=1`
- `git diff --check`

pkg/syncer/server.go
```diff
@@ -469,8 +472,18 @@ func (s *RegionSyncer) syncFullRegionsLocked(ctx context.Context, name string, s
 	releaseRetain := s.history.retainFrom(syncStartIndex)
 	defer releaseRetain()
 	regions := s.server.GetRegions()
-	lastIndex := 0
 	start := time.Now()
+	if len(regions) == 0 {
```
lhy1024: The empty-snapshot full-sync response is ambiguous with the empty completion/keepalive response. In `syncFullRegionsLocked`, when the leader snapshot is empty, the server sends an empty `StartIndex=0` response before sending catch-up history and the final completion response. However, the client only enters `fullSyncing` when `StartIndex == 0 && len(regions) > 0`, so this first empty frame is treated as a normal sync frame: it resets the local history index to 0 and immediately sets `streamingRunning` to true.

  This can enable follower local reads before the catch-up records are applied, exposing a partial/stale cache. I verified the sequence with a temporary UT: the server sends `StartIndex=0, Regions=[]`, then `StartIndex=42, Regions=[catch-up]`, then `StartIndex=43, Regions=[]`; after handling the first frame, the client is already running instead of staying in full-sync mode. Please avoid sending this extra empty `StartIndex=0` snapshot frame, or add an explicit distinguishable full-sync start/completion signal. At minimum, please add coverage for empty snapshot + catch-up.

lhy1024:
```
func TestEmptyFullSyncSequenceWithCatchUp(t *testing.T) {
        re := require.New(t)
        syncer, _ := newTestRegionSyncer(t)
        syncer.history.resetWithIndex(42)

        stream := newMockSyncRegionsServer()
        syncStream, startIndex := syncer.bindStreamForSync("pd-follower", stream)
        defer syncer.unbindStream("pd-follower", syncStream)
        unblockSend := stream.blockSend()
        done := make(chan error, 1)
        go func() {
                done <- syncFullRegionsForTest(context.Background(), syncer, syncStream, startIndex)
        }()
        testutil.Eventually(re, stream.isSendBlocked)

        catchUpRegion := newTestSyncRegion(2, 12)
        syncer.history.record(catchUpRegion)
        close(unblockSend)
        re.NoError(<-done)

        first := <-stream.sendCh
        re.Equal(uint64(0), first.GetStartIndex())
        re.Empty(first.GetRegions())
        second := <-stream.sendCh
        re.Equal(uint64(42), second.GetStartIndex())
        re.Equal([]*metapb.Region{catchUpRegion.GetMeta()}, second.GetRegions())
        third := <-stream.sendCh
        re.Equal(uint64(43), third.GetStartIndex())
        re.Empty(third.GetRegions())

        regionStorage := storage.NewStorageWithMemoryBackend()
        clientServer := mockserver.NewMockServer(context.Background(), nil, nil, regionStorage, core.NewBasicCluster())
        clientSyncer := NewRegionSyncer(clientServer)
        clientSyncer.history.resetWithIndex(42)
        bc := core.NewBasicCluster()

        handled, fullSyncing := clientSyncer.handleRegionSyncResponse(context.Background(), first, bc, regionStorage, false)
        re.True(handled)
        re.True(fullSyncing)
        re.False(clientSyncer.IsRunning())
  }
```

okJiang: Fixed in a1e53a034 by removing the special empty-snapshot `StartIndex=0` response. Empty snapshot full sync now sends catch-up records and the final completion frame only, so the ambiguous first empty frame in the temporary UT no longer exists.

I did not add a dedicated new UT because this was introduced by the extra response added in this PR; the fix removes that extra behavior instead of adding another path. I updated the existing stream-exit/blocking tests that had been waiting for the extra frame.

Verified with:
- `make gotest GOTEST_ARGS='./pkg/syncer -run TestSyncExitsWhenRegionSyncerStops\|TestSyncExitsWhenBroadcastSendFails\|TestCloseAllClientTimesOutBlockedSend\|TestBroadcastClosesStreamWhenSendBlocks\|TestSyncFallsBackToFullSyncWhenHistoryMissing\|TestClientWaitsForFullSyncCompletionBeforeRunning\|TestClientFinishesFullSyncWithOldLeaderKeepAlive -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -count=1'`
- `git diff --check`

okJiang:
ping @JmPotato @lhy1024

pkg/syncer/server.go
lhy1024: One minor follow-up: in the empty-snapshot path, the client still does not enter `fullSyncing`. If the retained catch-up records are split into multiple batches, it looks like `streamingRunning` may become true after the first batch, before the later batches and the final completion frame are applied. Could we add a small regression test for empty snapshot + multi-batch catch-up, or clarify why this cannot happen in practice?

lhy1024:
```
func TestTemporaryEmptyFullSyncMultipleCatchUpBatches(t *testing.T) {
        re := require.New(t)
        syncer, _ := newTestRegionSyncer(t)
        syncer.history.resetWithIndex(42)

        stream := newMockSyncRegionsServer()
        syncStream, startIndex := syncer.bindStreamForSync("pd-follower", stream)
        defer syncer.unbindStream("pd-follower", syncStream)

        for i := range maxSyncRegionBatchSize + 1 {
                syncer.history.record(newTestSyncRegion(uint64(10000+i), uint64(20000+i)))
        }

        re.NoError(syncFullRegionsForTest(context.Background(), syncer, syncStream, startIndex))

        first := <-stream.sendCh
        re.Equal(uint64(42), first.GetStartIndex())
        re.Len(first.GetRegions(), maxSyncRegionBatchSize)

        regionStorage := storage.NewStorageWithMemoryBackend()
        server := mockserver.NewMockServer(context.Background(), nil, nil, regionStorage, core.NewBasicCluster())
        clientSyncer := NewRegionSyncer(server)
        clientSyncer.history.resetWithIndex(42)

        handled, fullSyncing := clientSyncer.handleRegionSyncResponse(
                context.Background(), first, core.NewBasicCluster(), regionStorage, false,
        )
        re.True(handled)
        re.True(fullSyncing)
        re.False(clientSyncer.IsRunning())
  }
```

lhy1024 (approved):

okJiang:
ping @JmPotato

JmPotato (approved):

PR #10678, opened 2026-05-14
schedule: add split-scatter for release-8.5-20251204-v8.5.4

lhy1024:
### What problem does this PR solve?

Issue Number: ref #10621

### What is changed and how does it work?

```commit-message
Cherry-pick #10621 to support load-based split-scatter on release-8.5-20251204-v8.5.4.

PD records split batches reported with SplitReason_LOAD after AskBatchSplit allocates split IDs, keeps pending split-scatter regions, and dispatches internal scatter operators with range-aware scatter groups.

This PR also adds the split-scatter schedule limit config and updates kvproto to pingcap/kvproto v0.0.0-20260518035033-ca8835cfa721, which includes pingcap/kvproto#1464.
```

### Cherry-pick / dependency handling

Source PR: #10621, merge commit `38a0f9ab0a3e373c2da23fa7c556423b4a67a6db`.
Target branch: `release-8.5-20251204-v8.5.4`.
Current head: `44949721a3483de54eec810979bbc5cba1532e38`.

- The source PR #10621 does not change Go module files; this backport changes them only because the release branch needs the split-reason kvproto backport.
- Updated root, `tools/`, and `tests/integrations/` modules to `github.com/pingcap/kvproto v0.0.0-20260518035033-ca8835cfa721`.
- Removed the temporary `github.com/lhy1024/kvproto@07f3062a` replace and stale checksums after pingcap/kvproto#1464 was merged.
- Restored the split-scatter metric tests to the source PR's `prometheus/testutil` style; root `github.com/prometheus/client_model v0.6.1` remains indirect after `go mod tidy`.
- No master-only affinity checker/operator code or checker metrics refactor was brought into this release backport.
- MCS/NEXT_GEN split-scatter behavior remains out of scope; the existing TODOs are preserved and only the shared config getter is added for interface compatibility.

### Check List

Tests

- Unit test
- Manual test

Manual test commands:

```sh
PATH="$PWD/.tools/bin:$PATH" GOTOOLCHAIN=go1.23.12 make check
.tools/bin/failpoint-ctl enable . && go test ./pkg/schedule/checker -run SplitScatter -count=1; rc=$?; .tools/bin/failpoint-ctl disable .; exit $rc
[...]

PR #10691, opened 2026-05-19
checker: address split scatter pending follow-ups

lhy1024:

Issue Number: ref #10592 pick from #10678

```commit-message
Backport the split-scatter pending follow-up fixes from the release-8.5 cherry-pick back to master.

Reset the process-global pending gauge when creating a split-scatter controller, avoid recording or retaining pending split-scatter entries while split-scatter scheduling is disabled, add dispatch backoff for blocked pending work, and count missing pending regions when delaying retries.

Keep the master test style by using prometheus/testutil for metric assertions.
```

### Check List

Tests

- Unit test

### Release note

```release-note
PD now avoids stale pending load-based split-scatter entries and excessive retry loops when split-scatter scheduling is disabled or pending split regions are not ready.
```

## Summary by CodeRabbit

* **New Features**
  * Added dispatch throttling and explicit retry delays for split-scatter scheduling; pending items with missing regions are delayed and retried later. Pending counters are initialized/reset correctly.

* **Bug Fixes**
  * Prevented recording/dispatch when scheduling is disabled and ensured pending state is cleared.
  * Improved backoff when operator limits are reached or no candidates are available; missing-region occurrences now increment a metric.

* **Tests**
  * Added coverage for pending gauge reset, disabled scheduling, missing-region retries, and dispatch backoff.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10691?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/schedule/checker/split_scatter.go
+++ b/pkg/schedule/checker/split_scatter.go
@@ -63,7 +63,8 @@ type splitScatterController struct {
 	// pending maps a pending region ID to its latest split-scatter batch item.
 	// The item keeps its batch group so stale snapshots cannot mutate a newer
 	// pending entry for the same region.
-	pending map[uint64]splitScatterPendingItem
+	pending        map[uint64]splitScatterPendingItem
+	nextDispatchAt time.Time
 }
 
 func newSplitScatterController(
@@ -72,6 +73,7 @@ func newSplitScatterController(
 	opController *operator.Controller,
 	addPendingProcessedRegions func(needCheckLen bool, ids ...uint64),
 ) *splitScatterController {
+	splitScatterPendingGauge.Set(0)
 	return &splitScatterController{
 		cluster:         cluster,
 		opController:    opController,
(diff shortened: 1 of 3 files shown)


liyishuai (approved):

lhy1024:
@rleungx  @bufferflies @okJiang PTAL

pkg/schedule/checker/split_scatter.go
```diff
@@ -250,6 +280,29 @@ func (c *splitScatterController) cleanupExpiredPendingSplitScatter() int {
 	return pendingCount
 }
 
+func (c *splitScatterController) clearPendingSplitScatter() {
+	c.pendingMu.Lock()
+	defer c.pendingMu.Unlock()
+	pendingCount := len(c.pending)
+	if pendingCount > 0 {
+		c.pending = make(map[uint64]splitScatterPendingItem)
+		c.updatePendingGaugeLocked()
```
bufferflies: Will the metrics  reset zero if pending count is zero，

pkg/schedule/checker/split_scatter.go
```diff
@@ -317,27 +373,39 @@ func (c *splitScatterController) recordSplitScatterBatch(sourceRegionID, sourceW
...
 	if c.cleanupExpiredPendingSplitScatter() == 0 {
 		return
 	}
 	limit := c.cluster.GetCheckerConfig().GetSplitScatterScheduleLimit()
 	if limit == 0 {
 		splitScatterDispatchDisabledCounter.Inc()
+		c.clearPendingSplitScatter()
+		return
+	}
+	if c.skipDispatchUntil(now) {
```
bufferflies, line 386: Do we needs to add new metrics recording  why the split controller  doesn't work?

pkg/schedule/checker/split_scatter.go
lhy1024: Updated in 64529c343. `clearPendingSplitScatter` now always refreshes `splitScatterPendingGauge` after clearing pending, so the metric is reset to 0 even when the pending count is already 0. `PatrolRegions` also calls the same cleanup path when it exits.

pkg/schedule/checker/split_scatter.go
lhy1024: I think the existing checker event metrics already cover the main reasons why split-scatter dispatch does not make progress. They are reported through `pd_checker_event_count{type="split_scatter_checker", name=...}`, including `dispatch-disabled`, `dispatch-schedule-limit`, `dispatch-region-missing`, `dispatch-schedule-disabled`, `dispatch-not-fully-replicated`, `dispatch-scatter-failed`, `dispatch-store-limit`, and `dispatch-add-operator-failed`. The `skipDispatchUntil` path is only the retry backoff after one of those reasons has already been recorded, so I do not add another metric for it here.

bufferflies (approved):

rleungx (approved):

PR #10703, opened 2026-05-21
client/tso: fix race in stream setup that cancels in-flight requests

### LINKED ISSUE #10655: TSO stream setup can race between dispatcher and background connection updater
## Problem

During initial TSO client setup, the background connection context updater and the request dispatcher can both call `updateConnectionCtxs` concurrently.

If both paths create a stream for the same TSO primary, the later store can cancel and replace the earlier stored stream. A pending TSO request that has already been attached to the replaced stream can then fail with `context canceled`, even though the caller's context is still valid.

## Possible sequence

1. A fresh TSO client starts `connectionCtxsUpdater`.
2. The first updater iteration enters `updateConnectionCtxs`, gets the TSO primary connection, and builds stream A, but has not stored it in `conCtxMgr` yet.
3. Before stream A is stored, the first `GetTS` request reaches the TSO dispatcher.
4. The dispatcher calls `conCtxMgr.RandomlyPick()` and sees no ready stream.
5. The dispatcher enters the no-stream path and calls `provider.updateConnectionCtxs(ctx)`.
6. This second `updateConnectionCtxs` call builds and stores stream B for the same TSO primary.
7. The dispatcher sends or attaches the pending request to stream B.
8. The background updater resumes and stores stream A with `CleanAllAndStore`.
9. `CleanAllAndStore` releases the existing stream for the same URL, canceling stream B.
10. The pending request attached to stream B observes `context canceled`, so the initial `GetTS` fails.

## Expected behavior

Initial stream setup should not allow two concurrent `updateConnectionCtxs` calls to create competing streams for the same TSO primary.

The dispatcher could wait for an in-flight initial update, or the connection update path could be protected by singleflight/serialization, so a request already attached to a valid stream is not canceled by another initial stream creation.

## Notes

This race is limited to cold TSO client setup, before any ready stream has been stored. That makes it hard to observe without controlled scheduling, but the interleaving is possible from the current dispatcher no-stream fallback plus the background first update.

liyishuai:

Issue Number: Close #10655

On a freshly-created TSO client, the background `connectionCtxsUpdater` goroutine and the dispatcher's no-stream fallback both call `updateConnectionCtxs` concurrently. Both can observe an empty `conCtxMgr`, each build a gRPC stream to the TSO primary, and each call `CleanAllAndStore`. The second call's `storeLocked` cancels the first stream's context via `releaseLocked`, propagating `context.Canceled` to any request already attached to that stream.

```commit-message
client/tso: fix race in stream setup that cancels in-flight requests

`CleanAllAndStore` now aborts instead of overwriting when the URL is
already registered: it cancels the new (redundant) context and returns
without touching the existing entry or any in-flight requests on it.

The early-return path in `tryConnectToTSO` that previously called
`CleanAllAndStore` with no stream now calls `GC` directly, removing
the unnecessary context allocation.

The backup-stream store in the forwarding path is moved before the
`checkLeader` goroutine spawn so the entry is visible in the manager
before the goroutine can interact with it.

Four `failpoint.InjectCall` sites are added at the key scheduling
points of the race, and `TestTSOStreamSetupRace` uses them to
deterministically reproduce the interleaving: pause the background
updater before its first store, let the dispatcher build and attach a
request to its own stream, then release the background updater.
Without the fix the test fails with `context.Canceled`; with the fix
it passes.
```

### Check List

Tests

- Integration test

Side effects

- None

Related changes

### Release note

```release-note
Fix a race on TSO client cold startup where the background connection updater and the dispatcher's no-stream fallback could both build streams to the same TSO primary; the second `CleanAllAndStore` would cancel the first stream, failing any in-flight request with `context canceled`.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved race conditions during TSO connection and stream setup; improved context/store checks to avoid stale or duplicate connections and ensure reliable cancellation on failure.
  * Improved robustness of leader/follower failover and forwarded-request handling.

* **Tests**
  * Added an integration test to validate TSO stream setup under concurrent pauses and recoveries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10703?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/clients/router/client.go
+++ b/client/clients/router/client.go
@@ -458,7 +458,11 @@ func (c *Cli) updateMsConnection(ctx context.Context) {
 		}
 		// Store the stream connection context if it is successfully created.
 		if stream != nil {
-			c.msConCtxMgr.Store(cctx, cancel, url, stream)
+			if !c.msConCtxMgr.Store(cctx, cancel, url, stream) {
+				log.Info("[router] already exists router service stream connection", zap.String("url", url))
+				cancel()
+				continue
+			}
 			log.Info("[router] successfully established the router service stream connection", zap.String("url", url))
 		} else {
 			log.Warn("[router] failed to create the router service stream connection")
@@ -490,7 +494,11 @@ func (c *Cli) updateConnection(ctx context.Context) {
 		}
 		// Store the stream connection context if it is successfully created.
 		if stream != nil {
-			c.conCtxMgr.Store(cctx, cancel, url, stream)
+			if !c.conCtxMgr.Store(cctx, cancel, url, stream) {
+				log.Info("[router] already exists leader router stream connection", zap.String("url", url))
+				cancel()
+				return
+			}
 			log.Info("[router] successfully established the leader router stream connection", zap.String("url", url))
 		} else {
 			log.Warn("[router] failed to create the leader router stream connection")
@@ -519,7 +527,11 @@ func (c *Cli) updateConnection(ctx context.Context) {
 			}
 			// Store the stream connection context if it is successfully created.
 			if stream != nil {
-				c.conCtxMgr.Store(cctx, cancel, url, stream)
+				if !c.conCtxMgr.Store(cctx, cancel, url, stream) {
+					log.Info("[router] already exists router stream connection", zap.String("url", url))
+					cancel()
+					continue
+				}
 				log.Info("[router] successfully established the router stream connection", zap.String("url", url))
 			} else {
 				log.Warn("[router] failed to create the router stream connection")
--- a/client/clients/tso/client.go
+++ b/client/clients/tso/client.go
@@ -331,9 +331,8 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 		c.svcDiscovery.ScheduleCheckMemberChanged()
 		cc, url = c.getTSOLeaderClientConn()
 		if c.conCtxMgr.Exist(url) {
-			cctx, cancel := context.WithCancel(ctx)
 			// Just trigger the clean up of the stale connection contexts.
-			c.conCtxMgr.CleanAllAndStore(cctx, cancel, url)
+			c.conCtxMgr.GC(func(curURL string) bool { return curURL != url })
 			return nil
 		}
 		if cc != nil {
@@ -345,8 +344,11 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 			})
 			if stream != nil && err == nil {
 				failpoint.InjectCall("pauseBeforeBackgroundStoreTSOLeaderStream")
-				c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream)
+				stored := c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream)
 				failpoint.InjectCall("notifyAfterBackgroundStoreTSOLeaderStream")
+				if !stored {
+					cancel()
+				}
 				return nil
 			}
 
@@ -388,10 +390,13 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 			if err == nil {
 				forwardedHostTrim := tlsutil.TrimHTTPPrefix(forwardedHost)
 				addr := tlsutil.TrimHTTPPrefix(backupURL)
+				if !c.conCtxMgr.CleanAllAndStore(cctx, cancel, backupURL, stream) {
+					cancel()
+					return nil
+				}
 				// the goroutine is used to check the network and change back to the original stream
 				go c.checkLeader(ctx, cancel, forwardedHostTrim, addr, url)
 				metrics.RequestForwarded.WithLabelValues(forwardedHostTrim, addr).Set(1)
-				c.conCtxMgr.CleanAllAndStore(cctx, cancel, backupURL, stream)
 				return nil
 			}
 			cancel()
@@ -436,7 +441,9 @@ func (c *Cli) checkLeader(
 				stream, err := c.tsoStreamBuilderFactory.makeBuilder(cc).build(cctx, cancel, c.option.Timeout)
 				if err == nil && stream != nil {
 					log.Info("[tso] recover the original tso stream since the network has become normal", zap.String("url", url))
-					c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream)
+					if !c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream) {
+						cancel()
+					}
 					return
 				}
 			}
@@ -490,7 +497,10 @@ func (c *Cli) tryConnectToTSOWithProxy(ctx context.Context) error {
 				addrTrim := tlsutil.TrimHTTPPrefix(addr)
 				metrics.RequestForwarded.WithLabelValues(forwardedHostTrim, addrTrim).Set(1)
 			}
-			c.conCtxMgr.Store(cctx, cancel, addr, stream)
+			if !c.conCtxMgr.Store(cctx, cancel, addr, stream) {
+				log.Info("[tso] already exists tso stream", zap.String("addr", addr))
+				cancel()
+			}
 			continue
 		}
 		log.Error("[tso] create the tso stream failed",
--- a/client/clients/tso/client.go
+++ b/client/clients/tso/client.go
@@ -344,7 +344,9 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 				err = status.New(codes.Unavailable, "unavailable").Err()
 			})
 			if stream != nil && err == nil {
+				failpoint.InjectCall("pauseBeforeBackgroundStoreTSOLeaderStream")
 				c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream)
+				failpoint.InjectCall("notifyAfterBackgroundStoreTSOLeaderStream")
 				return nil
 			}
 
--- a/client/pkg/connectionctx/manager.go
+++ b/client/pkg/connectionctx/manager.go
@@ -53,7 +53,9 @@ func (c *Manager[T]) Exist(url string) bool {
 
 // Store is used to store the connection context, `overwrite` is used to force the store operation
 // no matter whether the connection context exists before, which is false by default.
-func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url string, stream T, overwrite ...bool) {
+// It returns whether the manager took ownership of ctx/cancel/stream. If false, the caller
+// must cancel ctx to avoid leaking the stream and its associated resources.
+func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url string, stream T, overwrite ...bool) bool {
 	c.Lock()
 	defer c.Unlock()
 	overwriteFlag := false
@@ -62,9 +64,10 @@ func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url s
 	}
 	_, ok := c.connectionCtxs[url]
 	if !overwriteFlag && ok {
-		return
+		return false
 	}
 	c.storeLocked(ctx, cancel, url, stream)
+	return true
 }
 
 func (c *Manager[T]) storeLocked(ctx context.Context, cancel context.CancelFunc, url string, stream T) {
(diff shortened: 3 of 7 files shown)


client/clients/tso/client.go
```diff
@@ -266,7 +269,13 @@ func (c *Cli) connectionCtxsUpdater() {
 
 	ctx, cancel := context.WithCancel(c.ctx)
 	defer cancel()
+	ctx = context.WithValue(ctx, backgroundConnectionCtxsUpdaterKey{}, struct{}{})
```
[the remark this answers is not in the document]
liyishuai: Disclaimer: AIGC.
> The context value is used here as the idiomatic Go mechanism for passing data through an existing call chain ('connectionCtxsUpdater' → 'updateConnectionCtxs' → 'tryConnectToTSO') without altering any function signatures. The alternative — adding a `fromBackground bool` parameter — would require changing `updateConnectionCtxs`, `tryConnectToTSO`, and the `createTSOConnection` function-value type (currently `func(context.Context) error`) used in `updateConnectionCtxs`. The context key type is private and unexported, so it cannot leak or be misused outside this package. The usage is also test-only (it gates a `failpoint.InjectCall` that is a no-op in production). Happy to refactor if there is a strong preference, but I'd push back that this is the lowest-complexity option given the call chain.

client/clients/tso/client.go
```diff
@@ -344,7 +353,16 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 				err = status.New(codes.Unavailable, "unavailable").Err()
 			})
 			if stream != nil && err == nil {
-				c.conCtxMgr.CleanAllAndStore(cctx, cancel, url, stream)
+				if _, ok := ctx.Value(backgroundConnectionCtxsUpdaterKey{}).(struct{}); ok {
+					failpoint.InjectCall("pauseBeforeBackgroundStoreTSOLeaderStream")
+				}
```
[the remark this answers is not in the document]
liyishuai: Disclaimer: AIGC.
> Same reasoning as above: the context value passes a caller-identity flag through the existing `updateConnectionCtxs` → `tryConnectToTSO` call chain without touching any function signatures or the `createTSOConnection` function-value type. The key type is unexported so it can only be read by this package, and it gates a test-only `failpoint.InjectCall` that is a no-op in production builds. I agree it is unconventional but it is the lowest-change option given the current architecture.

client/clients/tso/client.go
```diff
@@ -344,7 +353,16 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
...
+				c.connectMu.Lock()
```
lhy1024: This fixes the double-store race for leader streams, but the fallback store and `checkLeader` recovery store in the same non-proxy single-stream path still call `CleanAllAndStore(..., stream)` without this guard. I confirmed with a minimal test that a second `CleanAllAndStore` on the same URL cancels the ctx stored by the first one, so with `EnableForwarding` this can still cancel an in-flight request.

Could we extract a conditional store helper and use it consistently for leader/fallback/recover? For fallback, `checkLeader` should ideally start only after the stream is confirmed to be stored.

tests/integrations/tso/client_test.go
```diff
@@ -743,6 +743,117 @@ func checkTSO(
...
+	waitFor(backgroundUpdateStarted, "background updater first tick")
+	close(releaseBackgroundUpdate)
+	waitFor(backgroundBeforeStore, "background goroutine reaching CleanAllAndStore")
+	waitFor(dispatcherSawNoStream, "dispatcher seeing no stream")
+	close(releaseDispatcherUpdate)
+	waitFor(requestAttachedToStream, "request attached to dispatcher's stream")
+	close(releaseBackgroundStore)
+	close(releaseRequest)
+
+	err = <-errCh
```
lhy1024: non-blocking: The happy path looks fine, but failure paths can still leave goroutines blocked inside failpoint callbacks. A `waitFor` timeout does not release callbacks already blocked on `<-release...`, and `failpoint.Disable` will not unblock callbacks that have already entered. The final `<-errCh` also has no timeout.

  Could we add a cleanup defer to safely close all release channels and also bound the final `errCh` wait?

client/clients/tso/client.go
```diff
@@ -386,10 +396,10 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
 			if err == nil {
 				forwardedHostTrim := tlsutil.TrimHTTPPrefix(forwardedHost)
 				addr := tlsutil.TrimHTTPPrefix(backupURL)
+				c.conCtxMgr.CleanAllAndStore(cctx, cancel, backupURL, stream)
```
rleungx: When `backupURL` is already registered, `CleanAllAndStore` cancels this new fallback stream, but the following `checkLeader` goroutine and forwarding metric update still belong to that discarded stream.

liyishuai: Added boolean return value to indicate whether the url was empty, and called the cancellation explicitly outside CleanAllAndStore.

client/clients/tso/client.go
```diff
@@ -386,10 +398,10 @@ func (c *Cli) tryConnectToTSO(ctx context.Context) error {
...
```
rleungx: This return value is still ignored here. If `backupURL` is already registered, the newly built fallback stream is not stored, its context remains uncanceled, and the following `checkLeader` goroutine and forwarding metric update still belong to that loser stream.

client/pkg/connectionctx/manager_test.go
```diff
@@ -89,16 +89,17 @@ func TestManager(t *testing.T) {
 	re.False(manager.Exist("test-url"))
 	re.Nil(manager.GetConnectionCtx("test-url"))
 	re.True(manager.Exist("test-another-url"))
+	re.False(manager.CleanAllAndStore(ctx, cancel, "test-another-url", 4))
```
rleungx: This covers the false return and the preserved entry, but it reuses the existing `ctx/cancel`, so it does not cover the caller-visible loser-context behavior that matters for stream leaks.

client/pkg/connectionctx/manager.go
```diff
@@ -73,19 +73,21 @@ func (c *Manager[T]) storeLocked(ctx context.Context, cancel context.CancelFunc,
 }
 
 // CleanAllAndStore is used to store the connection context exclusively. It will release
-// all other connection contexts. `stream` is optional, if it is not provided, all
-// connection contexts other than the given `url` will be released.
-func (c *Manager[T]) CleanAllAndStore(ctx context.Context, cancel context.CancelFunc, url string, stream ...T) {
+// all other connection contexts.
+// If a connection is already registered with the `url`, then return false.
+// Otherwise, return true to indicate that `stream` is registered to the `url`.
+func (c *Manager[T]) CleanAllAndStore(ctx context.Context, cancel context.CancelFunc, url string, stream T) bool {
```
rleungx, line 84: This conflict contract now exists for `CleanAllAndStore`, but `Store` still has a duplicate-ignore path that drops the newly created ctx/stream without cancellation. The proxy update path can hit that case after its separate `Exist(addr)` check.

liyishuai: Updated usages of Store accordingly.

rleungx (approved):

client/clients/router/client.go
```diff
@@ -458,7 +458,11 @@ func (c *Cli) updateMsConnection(ctx context.Context) {
 		}
 		// Store the stream connection context if it is successfully created.
 		if stream != nil {
-			c.msConCtxMgr.Store(cctx, cancel, url, stream)
+			if !c.msConCtxMgr.Store(cctx, cancel, url, stream) {
```
bufferflies, line 461: How about warp this into the store function?

liyishuai: I tend to have the store function only do the store or fail, and not touch the context, so as to keep its semantics simple and readable.

tests/integrations/tso/client_test.go
```diff
@@ -743,6 +743,132 @@ func checkTSO(
...
+		default:
+			close(ch)
+		}
+	}
+	defer func() {
+		safeClose(releaseBackgroundUpdate)
+		safeClose(releaseBackgroundStore)
+		safeClose(releaseDispatcherUpdate)
+		safeClose(releaseRequest)
+	}()
```
lhy1024: The cleanup defer is registered too early. Go defers run in LIFO order, so the later `defer pdClient.Close()` runs before this release-channel after `pdClient.Close()` returns, the test can still hang.

Please move this release-channel cleanup defer after `defer pdClient.Close()`, or combine cleanup into one defer that closes the release channels before closing the client/canceling/disabling failpoints.

tests/integrations/tso/client_test.go
```diff
@@ -836,17 +813,31 @@ func (suite *tsoClientTestSuite) TestTSOStreamSetupRace() {
...
+		safeClose(releaseBackgroundUpdate)
+		safeClose(releaseBackgroundStore)
+		safeClose(releaseDispatcherUpdate)
+		safeClose(releaseRequest)
+		pdClient.Close()
+		cancel()
+		re.NoError(failpoint.Disable(tsoFailpointPrefix + "pauseAfterTSORequestAttachedToStream"))
+		re.NoError(failpoint.Disable(tsoFailpointPrefix + "pauseDispatcherBeforeUpdateConnectionCtxsWhenNoStream"))
+		re.NoError(failpoint.Disable(tsoFailpointPrefix + "pauseBeforeBackgroundStoreTSOLeaderStream"))
+		re.NoError(failpoint.Disable(tsoFailpointPrefix + "pauseBeforeBackgroundStoreTSOLeaderStream"))
```
lhy1024: This cleanup now fixes the defer-order issue, but the failpoint disable list has a copy-paste mistake: `pauseBeforeBackgroundStoreTSOLeaderStream` is disabled twice, while `pauseBeforeBackgroundFirstUpdateConnectionCtxs` is never disabled. This can still leak the first failpoint into later tests.

Please replace one of the duplicated disables with:
`failpoint.Disable(tsoFailpointPrefix + "pauseBeforeBackgroundFirstUpdateConnectionCtxs")`

client/clients/tso/client.go
```diff
@@ -311,11 +318,26 @@ func (c *Cli) updateConnectionCtxs(ctx context.Context) bool {
...
+	}
+	return true
+}
+
 // tryConnectToTSO will try to connect to the TSO leader. If the connection becomes unreachable
 // and enableForwarding is true, it will create a new connection to a follower to do the forwarding,
 // while a new daemon will be created also to switch back to a normal leader connection ASAP the
 // connection comes back to normal.
-func (c *Cli) tryConnectToTSO(ctx context.Context) error {
+func (c *Cli) tryConnectToTSO(ctx context.Context, fromBackground bool) error {
```
lhy1024: non-blocking: The ctx key is gone now, but `fromBackground` is still a test-only signal in the production function signature. The normal path does not seem to need different semantics for the background updater and the dispatcher. Could we make this a generic “before storing TSO leader stream” failpoint and let the test ordering plus `sync.Once` guarantee that the first hit is from the background updater? Then `tryConnectToTSO` would not need a test-only parameter, and `backgroundUpdateConnectionCtxs` could be removed.

client/pkg/connectionctx/manager.go
```diff
@@ -53,7 +53,7 @@ func (c *Manager[T]) Exist(url string) bool {
 
 // Store is used to store the connection context, `overwrite` is used to force the store operation
 // no matter whether the connection context exists before, which is false by default.
-func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url string, stream T, overwrite ...bool) {
+func (c *Manager[T]) Store(ctx context.Context, cancel context.CancelFunc, url string, stream T, overwrite ...bool) bool {
```
lhy1024, line 58: non-blocking: The bool return value is now also an ownership signal: when it returns false, the manager did not take ownership of the newly passed ctx/cancel/stream, so the caller must cancel it. Could we document this on `Store` and `CleanAllAndStore` to prevent future call sites from ignoring the return value and leaking a stream/context?

client/clients/tso/client.go
bufferflies: You can define fromBackground variable and then modify this fromBackground  in a failpoint.

lhy1024 (approved):

Issue #10711, opened 2026-05-26
syncer: make region history buffer adaptively resize


okJiang:
## Enhancement Task

The region syncer history buffer can still overflow during a full region sync when the PD leader continues receiving region heartbeats and the history index advances faster than the follower can finish the full snapshot and catch-up phase.

#10692 improves the initial history buffer capacity by scaling it with visible memory, but the buffer capacity is still fixed after startup. A fixed capacity can reduce the overflow probability, but it cannot prevent repeated full-sync retries when the actual required history window becomes larger than the current buffer.

This issue tracks an adaptive history buffer implementation for the region syncer.

### Goal

Make the region syncer history buffer resize according to the actual required history window.

Expected behavior:

- Keep the default/base history buffer capacity memory-aware, following the direction in #10692.
- Keep the runtime capacity in units of `2^n * 10000`.
- Grow the buffer when the required replay window stays above the healthy range.
- Shrink the buffer when the required replay window stays below the healthy range for a while.
- Keep the healthy usage range around `(capacity / 4, capacity / 2]`.
- Use 2x growth and 1/2 shrink steps.
- Avoid triggering growth merely because the ring buffer is full; use the required replay window instead.
- During full region sync, retain the history range needed for catch-up so the leader does not overwrite the change logs required by the follower before the full sync completes.
- Keep a memory-based maximum capacity to avoid unbounded memory usage.
- If the required retained range exceeds the maximum capacity, fail the current full sync attempt explicitly and keep follower local reads disabled.

### Suggested implementation direction

- Add adaptive resize support to `historyBuffer`.
- Track the required history window from:
  - full sync retain start index,
  - follower requested `StartIndex`,
  - replay window observed during normal sync,
  - history miss / full-sync fallback cases.
- Resize capacity to `2^n * 10000`.
- Use the memory-based sizing from #10692 as the base capacity source.
- Add a memory-based maximum capacity higher than the base capacity.
- Only shrink when there is no active full-sync retain.
- Keep resize logic encapsulated inside `historyBuffer` as much as possible.

### Validation

Add tests for:

- Normal ring behavior when there is no retain.
- Growth when the required replay window exceeds the healthy range.
- Shrink when the required replay window remains below the healthy range.
- Capacity is always `2^n * 10000`.
- Full sync retains the required catch-up history.
- Multiple concurrent retain windows preserve the earliest required index.
- Full sync succeeds when history grows beyond the base capacity but stays below the maximum capacity.
[...]

PR #10712, opened 2026-05-26
syncer: adapt region history buffer size

okJiang:

Issue Number: ref #10711

This PR is based on `master` and is independent from #10689. It only includes the adaptive history buffer core logic and tests. Metrics and dashboard changes are left to a follow-up PR.

```commit-message
Make the region syncer history buffer adaptive instead of fixed at startup.

The buffer now uses a memory-aware base capacity in 2^n * 10000 units, grows from the observed replay window, retains full-sync catch-up history, and shrinks after the required window stays low with no active full-sync retain.

If full sync produces more change logs than the maximum retained window can hold, the leader returns an explicit sync error instead of binding a follower that can no longer catch up incrementally.
```

### Check List

Tests

- Unit test

Side effects

- Increased code complexity

### Release note

```release-note
Improve PD region synchronization by adaptively resizing the history buffer to reduce repeated full sync caused by history buffer overflow.
```

## Summary by CodeRabbit

* **New Features**
  * History buffer now dynamically resizes with memory-aware max sizing, automatic grow/shrink, periodic shrink ticker, and retention support to preserve replay windows during sync; outbound sync sends are serialized to reduce concurrency issues.

* **Bug Fixes**
  * Improved retention and overflow handling to avoid losing needed records, stabilize catch-up behavior, and reduce sync failures after restarts.

* **Tests**
  * Added/extended unit tests for memory-based sizing, resize/growth/shrink behavior, persistence across restart, and sync catch-up scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10712?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -19,7 +19,6 @@ import (
 
 	"go.uber.org/zap"
 
-	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/core"
@@ -32,7 +31,6 @@ const (
 	historyKey                = "historyIndex"
 	defaultFlushCount         = 100
 	historyBufferCapacityUnit = 10000
-	historyBufferGrowFactor   = 8
 	historyBufferShrinkRounds = 3
 )
 
@@ -42,8 +40,6 @@ var (
 	firstIndexGauge = regionSyncerStatus.WithLabelValues("first_index")
 	lastIndexGauge  = regionSyncerStatus.WithLabelValues("last_index")
 )
-
-var errHistoryBufferRetainOverflow = errors.New("history buffer retained range exceeds maximum capacity")
 
 type historyBuffer struct {
 	syncutil.RWMutex
@@ -55,22 +51,14 @@ type historyBuffer struct {
 	baseCapacity           int
 	maxCapacity            int
 	capacityUnit           int
-	retains                map[uint64]uint64
-	nextRetainID           uint64
 	observedRequiredWindow uint64
 	lowWindowRounds        int
 	kv                     kv.Base
 	flushCount             int
 }
 
-type historyRetainer struct {
-	h  *historyBuffer
-	id uint64
-}
-
-func newHistoryBuffer(size int, kv kv.Base) *historyBuffer {
-	baseCapacity := normalizeHistoryBufferCapacity(size, historyBufferCapacityUnit)
-	return newHistoryBufferWithConfig(baseCapacity, baseCapacity*historyBufferGrowFactor, historyBufferCapacityUnit, kv)
+func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
+	return newHistoryBufferWithConfig(baseCapacity, maxCapacity, historyBufferCapacityUnit, kv)
 }
 
 func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv kv.Base) *historyBuffer {
@@ -91,7 +79,6 @@ func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv 
 		baseCapacity: baseCapacity,
 		maxCapacity:  maxCapacity,
 		capacityUnit: capacityUnit,
-		retains:      make(map[uint64]uint64),
 		kv:           kv,
 		flushCount:   defaultFlushCount,
 	}
@@ -136,13 +123,6 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
-	if retainIndex, ok := h.minRetainIndexLocked(); ok {
-		window := uint64(0)
-		if h.index+1 > retainIndex {
-			window = h.index + 1 - retainIndex
-		}
-		h.ensureWindowLocked(window)
-	}
 	syncIndexGauge.Set(float64(h.index))
 	h.records[h.tail] = r
 	h.tail = (h.tail + 1) % h.size
@@ -173,50 +153,6 @@ func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
 	return records
 }
 
-func (h *historyBuffer) retainFrom(index uint64) *historyRetainer {
-	h.Lock()
-	defer h.Unlock()
-	h.nextRetainID++
-	id := h.nextRetainID
-	h.retains[id] = index
-	window := uint64(0)
-	if h.index > index {
-		window = h.index - index
-	}
-	h.ensureWindowLocked(window)
-	return &historyRetainer{h: h, id: id}
-}
-
-func (r *historyRetainer) release() {
-	if r == nil || r.h == nil {
-		return
-	}
-	r.h.releaseRetain(r.id)
-}
-
-func (r *historyRetainer) overflowed() bool {
-	if r == nil || r.h == nil {
-		return false
-	}
-	return r.h.retainOverflowed(r.id)
-}
-
-func (h *historyBuffer) releaseRetain(id uint64) {
-	h.Lock()
-	defer h.Unlock()
-	delete(h.retains, id)
-}
-
-func (h *historyBuffer) retainOverflowed(id uint64) bool {
-	h.RLock()
-	defer h.RUnlock()
-	index, ok := h.retains[id]
-	if !ok {
-		return false
-	}
-	return index < h.firstIndex()
-}
-
 func (h *historyBuffer) observeRequiredWindow(window uint64) {
 	h.Lock()
 	defer h.Unlock()
@@ -235,7 +171,7 @@ func (h *historyBuffer) observeRequiredWindowLocked(window uint64) {
 func (h *historyBuffer) maybeShrink() {
 	h.Lock()
 	defer h.Unlock()
-	if len(h.retains) > 0 || h.capacity() <= h.baseCapacity {
+	if h.capacity() <= h.baseCapacity {
 		h.observedRequiredWindow = 0
 		h.lowWindowRounds = 0
 		return
@@ -250,12 +186,15 @@ func (h *historyBuffer) maybeShrink() {
 		h.observedRequiredWindow = 0
 		return
 	}
-	target := h.baseCapacity
+	target := h.capacity() / 2
 	if h.observedRequiredWindow > 0 {
-		target = normalizeHistoryBufferCapacity(int(h.observedRequiredWindow*2), h.capacityUnit)
-		if target < h.baseCapacity {
-			target = h.baseCapacity
+		required := normalizeHistoryBufferCapacity(int(h.observedRequiredWindow*2), h.capacityUnit)
+		if required > target {
+			target = required
 		}
+	}
+	if target < h.baseCapacity {
+		target = h.baseCapacity
 	}
 	if target < h.capacity() {
 		h.resizeLocked(target)
@@ -275,7 +214,6 @@ func (h *historyBuffer) resetWithIndexLocked(index uint64) {
 	h.head = 0
 	h.tail = 0
 	h.flushCount = defaultFlushCount
-	h.retains = make(map[uint64]uint64)
 	h.observedRequiredWindow = 0
 	h.lowWindowRounds = 0
 	if h.capacity() > h.baseCapacity {
@@ -316,16 +254,6 @@ func (h *historyBuffer) persist() {
 	if err != nil {
 		log.Warn("persist history index failed", zap.Uint64("persist-index", h.nextIndex()), errs.ZapError(err))
 	}
-}
-
-func (h *historyBuffer) ensureWindowLocked(window uint64) {
-	if window > h.observedRequiredWindow {
-		h.observedRequiredWindow = window
-	}
-	if window <= uint64(h.capacity()) {
-		return
-	}
-	h.growForWindowLocked(window * 2)
 }
 
 func (h *historyBuffer) growForWindowLocked(window uint64) {
@@ -370,18 +298,4 @@ func (h *historyBuffer) getLocked(index uint64) *core.RegionInfo {
 		return h.records[pos]
 	}
 	return nil
-}
-
-func (h *historyBuffer) minRetainIndexLocked() (uint64, bool) {
-	var (
-		minIndex uint64
-		ok       bool
-	)
-	for _, index := range h.retains {
-		if !ok || index < minIndex {
-			minIndex = index
-			ok = true
-		}
-	}
-	return minIndex, ok
 }
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -56,8 +56,10 @@ type historyBuffer struct {
 	observedRequiredWindow uint64
 	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
 	lowWindowRounds int
-	kv              kv.Base
-	flushCount      int
+	// retains tracks full-sync start indexes that must stay replayable until catch-up finishes.
+	retains    map[uint64]int
+	kv         kv.Base
+	flushCount int
 }
 
 func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
@@ -123,6 +125,9 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
+	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
+		h.growForWindowLocked(h.index - retainIndex + 1)
+	}
 	syncIndexGauge.Set(float64(h.index))
 	h.records[h.tail] = r
 	h.tail = (h.tail + 1) % h.size
@@ -140,6 +145,10 @@ func (h *historyBuffer) record(r *core.RegionInfo) {
 func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
 	h.RLock()
 	defer h.RUnlock()
+	return h.recordsFromLocked(index)
+}
+
+func (h *historyBuffer) recordsFromLocked(index uint64) []*core.RegionInfo {
 	var pos int
 	if index < h.nextIndex() && index >= h.firstIndex() {
 		pos = (h.head + int(index-h.firstIndex())) % h.size
@@ -151,6 +160,63 @@ func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
 		records = append(records, h.records[i])
 	}
 	return records
+}
+
+func (h *historyBuffer) observeAndRecordsFrom(index uint64) ([]*core.RegionInfo, uint64) {
+	h.Lock()
+	defer h.Unlock()
+	nextIndex := h.nextIndex()
+	if index != 0 && index < nextIndex {
+		h.observeRequiredWindowLocked(nextIndex - index)
+	}
+	return h.recordsFromLocked(index), nextIndex
+}
+
+func (h *historyBuffer) retainedRecordsFrom(index uint64) ([]*core.RegionInfo, uint64, bool) {
+	h.RLock()
+	defer h.RUnlock()
+	nextIndex := h.nextIndex()
+	if index == nextIndex {
+		return nil, nextIndex, true
+	}
+	records := h.recordsFromLocked(index)
+	return records, nextIndex, len(records) == int(nextIndex-index)
+}
+
+func (h *historyBuffer) retainFromCurrent() (uint64, func()) {
+	h.Lock()
+	defer h.Unlock()
+	index := h.nextIndex()
+	if h.retains == nil {
+		h.retains = make(map[uint64]int)
+	}
+	h.retains[index]++
+	return index, func() {
+		h.releaseRetain(index)
+	}
+}
+
+func (h *historyBuffer) releaseRetain(index uint64) {
+	h.Lock()
+	defer h.Unlock()
+	count := h.retains[index]
+	if count <= 1 {
+		delete(h.retains, index)
+		return
+	}
+	h.retains[index] = count - 1
+}
+
+func (h *historyBuffer) minRetainIndexLocked() (uint64, bool) {
+	var min uint64
+	var ok bool
+	for index := range h.retains {
+		if !ok || index < min {
+			min = index
+			ok = true
+		}
+	}
+	return min, ok
 }
 
 func (h *historyBuffer) observeRequiredWindow(window uint64) {
@@ -174,6 +240,11 @@ func (h *historyBuffer) observeRequiredWindowLocked(window uint64) {
 func (h *historyBuffer) maybeShrink() {
 	h.Lock()
 	defer h.Unlock()
+	if len(h.retains) > 0 {
+		h.observedRequiredWindow = 0
+		h.lowWindowRounds = 0
+		return
+	}
 	if h.capacity() <= h.baseCapacity {
 		h.observedRequiredWindow = 0
 		h.lowWindowRounds = 0
@@ -190,12 +261,6 @@ func (h *historyBuffer) maybeShrink() {
 		return
 	}
 	target := h.capacity() / 2
-	if h.observedRequiredWindow > 0 {
-		required := normalizeHistoryBufferCapacity(int(h.observedRequiredWindow*2), h.capacityUnit)
-		if required > target {
-			target = required
-		}
-	}
 	if target < h.baseCapacity {
 		target = h.baseCapacity
 	}
@@ -219,6 +284,7 @@ func (h *historyBuffer) resetWithIndexLocked(index uint64) {
 	h.flushCount = defaultFlushCount
 	h.observedRequiredWindow = 0
 	h.lowWindowRounds = 0
+	h.retains = nil
 	if h.capacity() > h.baseCapacity {
 		h.resizeLocked(h.baseCapacity)
 	}
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -43,18 +43,21 @@ var (
 
 type historyBuffer struct {
 	syncutil.RWMutex
-	index                  uint64
-	records                []*core.RegionInfo
-	head                   int
-	tail                   int
-	size                   int
-	baseCapacity           int
-	maxCapacity            int
-	capacityUnit           int
+	// index is the next index to assign to a recorded region change.
+	index        uint64
+	records      []*core.RegionInfo
+	head         int // Position of the first valid record in the ring.
+	tail         int // Position where the next record will be written.
+	size         int // Physical ring size; one slot is reserved to distinguish full and empty states.
+	baseCapacity int
+	maxCapacity  int
+	capacityUnit int
+	// observedRequiredWindow is the largest replay window requested since the last shrink check.
 	observedRequiredWindow uint64
-	lowWindowRounds        int
-	kv                     kv.Base
-	flushCount             int
+	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
+	lowWindowRounds int
+	kv              kv.Base
+	flushCount      int
 }
 
 func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
@@ -168,6 +171,9 @@ func (h *historyBuffer) observeRequiredWindowLocked(window uint64) {
 	}
 }
 
+// maybeShrink gradually reduces capacity after the required replay window
+// remains low for several checks. It shrinks by at most half each time and
+// never below the base capacity.
 func (h *historyBuffer) maybeShrink() {
 	h.Lock()
 	defer h.Unlock()
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -72,9 +72,6 @@ func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv 
 	}
 	// use an empty space to simplify operation
 	size := baseCapacity + 1
-	if size < 2 {
-		size = 2
-	}
 	records := make([]*core.RegionInfo, size)
 	h := &historyBuffer{
 		records:      records,
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -19,6 +19,7 @@ import (
 
 	"go.uber.org/zap"
 
+	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 
 	"github.com/tikv/pd/pkg/core"
@@ -28,8 +29,11 @@ import (
 )
 
 const (
-	historyKey        = "historyIndex"
-	defaultFlushCount = 100
+	historyKey                = "historyIndex"
+	defaultFlushCount         = 100
+	historyBufferCapacityUnit = 10000
+	historyBufferGrowFactor   = 8
+	historyBufferShrinkRounds = 3
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/syncer/history_buffer.go
```diff
@@ -28,8 +29,11 @@ import (
 )
 
 const (
-	historyKey        = "historyIndex"
-	defaultFlushCount = 100
+	historyKey                = "historyIndex"
+	defaultFlushCount         = 100
+	historyBufferCapacityUnit = 10000
+	historyBufferGrowFactor   = 8
```
okJiang: Where does this parameter come from, and why is it set to 8?

pkg/syncer/history_buffer.go
```diff
@@ -41,32 +43,60 @@ var (
...
-	head       int
-	tail       int
-	size       int
-	kv         kv.Base
-	flushCount int
+	index                  uint64
+	records                []*core.RegionInfo
+	head                   int
+	tail                   int
+	size                   int
```
okJiang: Please provide some comments for `index`, `head`, `tail`, and `size`.

pkg/syncer/history_buffer.go
```diff
@@ -119,13 +153,72 @@ func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
...
+func (h *historyBuffer) observeRequiredWindowLocked(window uint64) {
+	if window > h.observedRequiredWindow {
+		h.observedRequiredWindow = window
+	}
+	if window > uint64(h.capacity()/2) {
+		h.growForWindowLocked(window * 2)
+	}
+}
+
+func (h *historyBuffer) maybeShrink() {
```
okJiang, line 240: Need to add some necessary comments to this function

pkg/syncer/history_buffer.go
```diff
@@ -41,32 +43,60 @@ var (
...
+	index                  uint64
+	records                []*core.RegionInfo
+	head                   int
+	tail                   int
+	size                   int
+	baseCapacity           int
+	maxCapacity            int
+	capacityUnit           int
+	observedRequiredWindow uint64
+	lowWindowRounds        int
```
okJiang: Please add comments to these two fields.

pkg/syncer/server.go
```diff
@@ -192,6 +211,7 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 			}
 			s.broadcast(ctx, regions)
 		case <-ticker.C:
+			s.history.maybeShrink()
```
okJiang, line 224: I think we could use a longer time limit for the call, like 5 minutes.

pkg/syncer/history_buffer.go
```diff
@@ -41,32 +43,60 @@ var (
...
+func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv kv.Base) *historyBuffer {
+	baseCapacity = normalizeHistoryBufferCapacity(baseCapacity, capacityUnit)
+	maxCapacity = normalizeHistoryBufferCapacity(maxCapacity, capacityUnit)
+	if maxCapacity < baseCapacity {
+		maxCapacity = baseCapacity
+	}
 	// use an empty space to simplify operation
-	size++
+	size := baseCapacity + 1
 	if size < 2 {
```
okJiang: please remove this

pkg/syncer/history_buffer.go
okJiang: Done in daef0aac3. Added comments for the observed replay window and low-window round counter.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/server.go
okJiang: Done in daef0aac3. Split shrink checks from the 10s keep-alive tick and run `maybeShrink` every 5 minutes instead. Keep-alive broadcasts still use the existing 10s interval.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/history_buffer.go
okJiang: Done in daef0aac3. Added field comments for the ring index bookkeeping (`index`, `head`, `tail`, and `size`) so the reserved-slot invariant is explicit.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/history_buffer.go
okJiang: Done in daef0aac3. Added a function comment explaining that shrink is gradual, requires several low-window checks, shrinks by at most half each time, and never goes below the base capacity.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

okJiang:
Checked the failed CI on 5a45d0d10:

- `pull-unit-test-next-gen-2` failed in `TestAffinityHandlerTestSuite/TestAffinityListWithEmptyID`.
- `pull-unit-test-next-gen-3` failed during test cluster startup with `bind: address already in use`, then timed out.
- GitHub `chunks (9, Microservice Integration(!TSO))` failed in `TestResourceManagerClientTestSuite/standalone-resource-manager-with-client-discovery/TestCheckBackgroundJobs`.

These failures do not involve the syncer history buffer changes. I pushed daef0aac3 for the review comments, so fresh checks should start automatically; no manual `/test` rerun was posted.

Verified locally:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/history_buffer.go
okJiang: Done in e564fffce. Removed the redundant `size < 2` guard; normalized base capacity already guarantees the ring size is at least 2.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/history_buffer.go
```diff
@@ -28,8 +28,10 @@ import (
 )
 
 const (
-	historyKey        = "historyIndex"
-	defaultFlushCount = 100
+	historyKey                = "historyIndex"
+	defaultFlushCount         = 100
+	historyBufferCapacityUnit = 10000
```
rleungx, line 33: What does the historyBufferCapacityUnit mean?

okJiang: The growth base is calculated in units of 10,000.

For example, it scales as 10,000, 20,000, 40,000, and 80,000, rather than starting from 1 and increasing by powers of 2 (2ⁿ). In other words, the current formula is 2ⁿ × 10,000.

rleungx: Will it cause OOM?

okJiang: It should not grow without bound.

`historyBufferCapacityUnit` only controls capacity rounding: capacities grow as 10,000, 20,000, 40,000, 80,000. The actual upper bound is still `maxCapacity`, which is derived from machine memory by `historyBufferMaxSizeFromMemory` and capped by `maxHistoryBufferSize` (currently 80,000). If the required window exceeds that cap, the buffer stops growing and the follower falls back to full region sync.

So this can retain more history than before on larger machines, but it is bounded by the memory-based max size and should not cause unbounded OOM growth.

pkg/syncer/server.go
```diff
@@ -287,6 +311,10 @@ func recvSyncRegionRequest(ctx context.Context, stream pdpb.PD_SyncRegionsServer
 func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.SyncRegionRequest, stream pdpb.PD_SyncRegionsServer) error {
 	startIndex := request.GetStartIndex()
 	name := request.GetMember().GetName()
+	nextIndex := s.history.getNextIndex()
+	if startIndex < nextIndex {
+		s.history.observeRequiredWindow(nextIndex - startIndex)
```
lhy1024: This treats `startIndex == 0` full-sync requests as a replay window and may trigger buffer growth. However, `startIndex=0` means the follower needs a full sync, not incremental replay from history index 0. When the leader `nextIndex` is already large, this grows the buffer to max for the wrong reason.

lhy1024:
```
func TestSyncHistoryRegionStartIndexZeroDoesNotGrowHistoryWindow(t *testing.T) {
        re := require.New(t)
        syncer := newTestRegionSyncer(t, core.NewBasicCluster())
        syncer.history = newHistoryBufferWithConfig(2, 8, 1, storage.NewStorageWithMemoryBackend())
        syncer.history.resetWithIndex(10)

        err := syncer.syncHistoryRegion(context.Background(), &pdpb.SyncRegionRequest{
                Header:     &pdpb.RequestHeader{ClusterId: keypath.ClusterID()},
                Member:     &pdpb.Member{Name: "pd-follower", ClientUrls: []string{"http://127.0.0.1:2379"}},
                StartIndex: 0,
        }, newMockSyncRegionsServer())

        re.NoError(err)
        re.Equal(2, syncer.history.capacity())
  }

func newTestRegionSyncer(t *testing.T, bc *core.BasicCluster) *RegionSyncer {
        t.Helper()
        re := require.New(t)
        tempDir := t.TempDir()
        regionStorage, err := storage.NewRegionStorageWithLevelDBBackend(context.Background(), tempDir, nil)
        re.NoError(err)
        t.Cleanup(func() {
                re.NoError(regionStorage.Close())
        })
        server := mockserver.NewMockServer(
                context.Background(),
                nil,
                nil,
                storage.NewCoreStorage(storage.NewStorageWithMemoryBackend(), regionStorage),
                bc,
        )
        return NewRegionSyncer(server)
  }
```

pkg/syncer/server.go
lhy1024, line 340: The full-sync path currently does not retain the history index captured at the beginning of full sync. While `syncFullRegions` is sending the full region snapshot, the leader can still keep calling `history.record` for region changes. If those writes exceed the current ring capacity, the change logs generated after full sync started can be overwritten. `Sync` then still binds the stream, so the follower appears to enter streaming mode but can no longer catch up from the full-sync snapshot to the leader's latest state.

I verified this with two tests:
  - `TestSyncFullRegionsRetainsCatchUpHistory`: while full-sync send is blocked, 5 history records are written. Expected 5 records replayable from the full-sync start index; actual is 0.
  - `TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax`: when the required catch-up window exceeds max capacity, full sync should fail explicitly; actual `syncFullRegions` returns nil.

  Please retain `history.nextIndex()` when full sync starts, grow/protect the ring based on the earliest active retain index during recording, and return an explicit error instead of binding the follower if the required window exceeds max capacity.

lhy1024:
```
  func TestSyncFullRegionsRetainsCatchUpHistory(t *testing.T) {
        re := require.New(t)
        bc := core.NewBasicCluster()
        bc.PutRegion(newHistoryBufferTestRegion(1))
        syncer := newTestRegionSyncer(t, bc)
        syncer.history = newHistoryBufferWithConfig(2, 8, 1, storage.NewStorageWithMemoryBackend())
        syncer.history.resetWithIndex(10)

        stream := newMockSyncRegionsServer()
        unblockSend := stream.blockSend()
        startIndex := syncer.history.getNextIndex()
        done := make(chan error, 1)
        go func() {
                done <- syncer.syncFullRegions(context.Background(), "pd-follower", stream)
        }()
        testutil.Eventually(re, stream.isSendBlocked)

        for i := range 5 {
                syncer.history.record(newHistoryBufferTestRegion(uint64(100 + i)))
        }
        records := syncer.history.recordsFrom(startIndex)

        close(unblockSend)
        re.NoError(<-done)
        re.Len(records, 5)
  }

  func TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax(t *testing.T) {
        re := require.New(t)
        bc := core.NewBasicCluster()
        bc.PutRegion(newHistoryBufferTestRegion(1))
        syncer := newTestRegionSyncer(t, bc)
        syncer.history = newHistoryBufferWithConfig(2, 2, 1, storage.NewStorageWithMemoryBackend())
        syncer.history.resetWithIndex(10)

        stream := newMockSyncRegionsServer()
        unblockSend := stream.blockSend()
        done := make(chan error, 1)
        go func() {
                done <- syncer.syncFullRegions(context.Background(), "pd-follower", stream)
        }()
        testutil.Eventually(re, stream.isSendBlocked)

        for i := range 3 {
                syncer.history.record(newHistoryBufferTestRegion(uint64(200 + i)))
        }

        close(unblockSend)
        re.Error(<-done)
  }

```

pkg/syncer/history_buffer.go
```diff
@@ -119,13 +153,75 @@ func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
...
+		h.lowWindowRounds = 0
+		return
+	}
+	h.lowWindowRounds++
+	if h.lowWindowRounds < historyBufferShrinkRounds {
+		h.observedRequiredWindow = 0
+		return
+	}
+	target := h.capacity() / 2
+	if h.observedRequiredWindow > 0 {
```
lhy1024: This block looks redundant. The earlier check already guarantees `observedRequiredWindow <= capacity/4`, so `normalize(observedRequiredWindow*2)` should not exceed the current `target := capacity/2`. It does not effectively change the target and can be simplified to avoid misleading future maintainers.

pkg/syncer/server.go
```diff
@@ -287,6 +311,10 @@ func recvSyncRegionRequest(ctx context.Context, stream pdpb.PD_SyncRegionsServer
...
+	}
 	records := s.history.recordsFrom(startIndex)
```
lhy1024: `getNextIndex`, `observeRequiredWindow`, and `recordsFrom` are separate locked operations. `RunServer` can still record new changes and advance/overwrite the ring between them, so the state observed for growth may no longer match the state used by `recordsFrom`.

  Please consider encapsulating “check current first/next, grow if needed, and copy records” inside historyBuffer under one lock.

pkg/syncer/server.go
lhy1024, line 342: I want to clarify the intended semantics here. When the follower requests a non-zero `startIndex` that is no longer available in the history buffer, the current logic only returns nil, and `Sync` still binds the stream afterwards. The follower then waits for later broadcasts and, when it receives a higher `StartIndex`, resets its local index on the client side. This effectively skips the missing history window.

  If this is the existing eventual-consistency behavior, keeping it unchanged may be fine. However, this PR / #10711 mentions handling history misses / full-sync fallback and explicitly failing when the required window cannot be retained. Should this path also trigger full sync or return a clear error instead of binding the stream directly?

pkg/syncer/server.go
okJiang: Fixed in ba54314b5. Added `historyBuffer.observeAndRecordsFrom` so sync-history handling observes the required window, grows if needed, and copies records under one history-buffer lock. `syncHistoryRegion` now uses that single snapshot instead of separate locked calls.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/server.go
okJiang: Fixed in ba54314b5. `startIndex == 0` no longer feeds the adaptive replay-window observation, so a full-sync request does not grow the history buffer for index 0. Added `TestSyncHistoryRegionStartIndexZeroDoesNotGrowHistoryWindow`.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `make gotest GOTEST_ARGS="./pkg/syncer -run "TestSyncHistoryRegionStartIndexZeroDoesNotGrowHistoryWindow|TestSyncFullRegionsRetainsCatchUpHistory|TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax" -count=1"`

pkg/syncer/history_buffer.go
okJiang: Fixed in ba54314b5. Removed the redundant `observedRequiredWindow` target adjustment from `maybeShrink`; after the low-window checks pass, shrink now simply halves capacity and clamps to the base capacity.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/server.go
okJiang: Fixed in ba54314b5. If a non-zero `startIndex` is behind the leader `nextIndex` but is no longer available in the history buffer, the leader now falls back to full sync instead of binding the stream and skipping the missing window. If full-sync catch-up history cannot be retained, it returns `ResourceExhausted`.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

okJiang: Fixed in ba54314b5. Full sync now retains the history index captured at the start, grows the buffer while records are appended under that active retain, binds the stream before catch-up with sends serialized, replays retained catch-up records, and returns `ResourceExhausted` if the required catch-up window is no longer available.

Added coverage:
- `TestSyncFullRegionsRetainsCatchUpHistory`
- `TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax`

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `make gotest GOTEST_ARGS="./pkg/syncer -run "TestSyncHistoryRegionStartIndexZeroDoesNotGrowHistoryWindow|TestSyncFullRegionsRetainsCatchUpHistory|TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax" -count=1"`

okJiang:
Addressed CodeRabbit's `closeAllClient` send-serialization comment in e83d12646. `closeAllClient` now sends the close response through `sender.send`, so all writes to the same gRPC stream go through the same `sendMu`.

Checked the `pull-unit-test-next-gen-3` failure on ba54314b5. It failed in `TestSwitchModeDuringWorkload/pd-to-standalone` under `tests/integrations/mcs/resourcemanager`, which is outside the syncer history buffer change. I pushed e83d12646 for the CodeRabbit comment, so fresh checks should start automatically; no manual retest was posted.

Verified locally:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

pkg/syncer/server.go
```diff
@@ -348,7 +380,9 @@ func (*RegionSyncer) syncHistoryRecords(startIndex uint64, records []*core.Regio
 	return stream.Send(resp)
```
rleungx: This can now send a very large history replay in one response. Please split it into smaller batches, otherwise it can exceed the client 8 MiB receive limit and keep retrying.

pkg/syncer/server.go
```diff
@@ -406,12 +440,35 @@ func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream
...
 		zap.String("requested-server", name), zap.String("server", s.server.Name()), zap.Duration("cost", time.Since(start)))
-	return nil
+	syncStream := newRegionSyncStream(stream)
+	syncStream.sendMu.Lock()
+	s.bindRegionSyncStream(name, syncStream)
+	records, nextIndex, ok := s.history.retainedRecordsFrom(retainIndex)
+	if !ok {
+		s.unbindStream(name, syncStream)
+		syncStream.sendMu.Unlock()
+		return nil, status.Errorf(codes.ResourceExhausted,
```
rleungx, line 460: This error is returned after the full snapshot has already been sent. The follower may mark sync as running before catch-up fails, which can enable follower reads with incomplete data.

pkg/syncer/server.go
```diff
@@ -301,21 +334,20 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
+		return nil, nil
 	}
 	log.Info("sync the history regions with server",
 		zap.String("server", name),
 		zap.Uint64("from-index", startIndex),
-		zap.Uint64("last-index", s.history.getNextIndex()),
+		zap.Uint64("last-index", nextIndex),
 		zap.Int("records-length", len(records)))
-	return s.syncHistoryRecords(startIndex, records, stream)
+	return nil, s.syncHistoryRecords(startIndex, records, stream)
```
rleungx, line 354: Please bind the stream before sending the initial history response. Broadcasts during this send are still dropped for this follower, so the next StartIndex can jump and force another reset.

pkg/syncer/server.go
okJiang: Please explain the size of the gRPC message request when using the current maximum batch size and whether it exceeds the 8MiB limit.

pkg/syncer/server.go
okJiang: This should be outside the scope of this PR.

pkg/syncer/server.go
okJiang: Fixed in e512d015e. `syncHistoryRecords` now splits history replay into `maxSyncRegionBatchSize` chunks, so the initial replay and retained catch-up path no longer build one large response.

Added `TestSyncHistoryRecordsSplitBatches`.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

pkg/syncer/server.go
okJiang: Fixed in e512d015e. The leader now sends an empty completion response only after retained catch-up succeeds, and the client marks the syncer as running only on that completion response. If catch-up fails after a full snapshot, the stream returns the error and the follower does not enter running state with incomplete data.

Added `TestHandleRegionSyncResponseMarksRunningAfterCompletion` and extended `TestSyncFullRegionsRetainsCatchUpHistory` to check the completion marker.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

pkg/syncer/server.go
okJiang: Fixed in e512d015e by closing the same gap through retained catch-up. The initial history replay now retains its end index while it is being sent. After that replay, the stream is bound under `sendMu`, changes retained from that index are replayed, an empty completion marker is sent, and stale queued broadcasts are trimmed by `minStartIndex`.

This avoids dropped changes during the initial replay without binding a long initial replay into the normal broadcast path.

Added `TestRegionSyncStreamTrimsStaleBroadcast` and covered the completion marker in the full-sync test.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

okJiang:
Addressed the latest CodeRabbit review in e512d015e.

- The incremental replay gap is handled by retaining the replay end index, binding under `sendMu` for retained catch-up, sending an empty completion marker, and trimming stale queued broadcasts.
- I did not add a per-stream FIFO queue in this PR. `RunServer` calls `broadcast` synchronously, and `broadcast` waits for each per-stream send attempt before returning, so successive normal broadcasts are not emitted concurrently for the same stream. The overlap introduced by bind-and-catch-up is serialized with `sendMu` and stale trimming.

The Prow failures reported for e83d12646 are for an older head. Current head is e512d015e, and fresh checks should start automatically; not retriggering manually.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

pkg/syncer/server.go
okJiang: Correction: I reverted the incremental replay catch-up change in 8002488cd and left a TODO in `syncHistoryRegion` instead.

This is existing behavior: incremental history replay was already sent before binding the stream before this PR. The adaptive history buffer change does not introduce that ordering gap, so I would prefer to keep this PR focused and handle stream binding / send-ordering semantics in a follow-up PR.

The current PR still keeps the batching fix and the full-sync completion/catch-up fix.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

okJiang:
Correction to my previous CodeRabbit note: 8002488cd reverts the incremental history replay ordering change and leaves it as a TODO/follow-up instead. That ordering gap existed before this PR and should be handled separately with the stream binding / send-ordering semantics.

This PR still keeps the response batching fix and the full-sync completion/catch-up fix.

Fresh checks should start from 8002488cd automatically; not retriggering manually.

pkg/syncer/server.go
okJiang: I checked the size as requested.

The important bound is not `maxSyncRegionBatchSize` before this fix. Before splitting, `syncHistoryRecords` could send the whole retained history in one response, up to `maxHistoryBufferSize` = 80,000 records.

Measured with `SyncRegionResponse.Size()`:
- 1,000 minimal records: 10,875 bytes
- 80,000 minimal records: 943,492 bytes
- 1,000 records with epoch, 3 peers, and 32B start/end keys: 105,748 bytes
- 80,000 records with epoch, 3 peers, and 32B start/end keys: 8,766,982 bytes

The client recv limit is 8 MiB = 8,388,608 bytes, so a full 80k history replay can exceed it with normal-looking region metadata. Splitting by the existing `maxSyncRegionBatchSize` keeps each response well below the limit and is a small change localized to `syncHistoryRecords`.

No additional code change in this pass; the existing `TestSyncHistoryRecordsSplitBatches` covers the split behavior.

okJiang:
Checked the current-head `pull-unit-test-next-gen-2` failure on 8002488cd.

The failed package is `tests/server/api`:
- `TestStoreTestSuite/TestStoreSetState` failed at `store_test.go:468` with store state expected `Offline`, actual `Tombstone`.
- `TestRedirectorTestSuite/TestRedirect` later hit etcd no-leader / ReadIndex timeout errors.

These failures do not involve the syncer history-buffer code touched by the latest changes. I also reran the focused cases locally and they passed:

`NEXT_GEN=1 make gotest GOTEST_ARGS="./tests/server/api -run \"TestStoreTestSuite/TestStoreSetState|TestRedirectorTestSuite/TestRedirect\" -count=1"`

No manual `/test` or `/retest` was posted.

pkg/syncer/server.go
okJiang: After narrowing the PR back to e83d12646 as the base, I kept only the minimal fix that is directly caused by this PR.

The size concern is valid: with the adaptive max capacity, `syncHistoryRecords` can replay up to 80,000 retained records. In the local size check, 80,000 records with epoch, 3 peers, and 32B start/end keys measured 8,766,982 bytes, which is above the 8 MiB client receive limit.

Fixed in 46e7b3c9a by splitting history replay responses by the existing `maxSyncRegionBatchSize`. This keeps the replay response size bounded without bringing back the later completion-marker or stream-ordering changes.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make gotest GOTEST_ARGS="./pkg/syncer -run "TestSyncHistoryRecordsSplitBatches|TestSyncFullRegionsRetainsCatchUpHistory|TestSyncFullRegionsFailsWhenCatchUpHistoryExceedsMax" -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`
- `git diff --check`

pkg/syncer/server.go
okJiang: Correction after narrowing the PR scope: the previous e512d015e completion-marker change is no longer in this PR. The current head is based on e83d12646 plus only the small replay-batching fix in 46e7b3c9a.

With this boundary, I am not changing the running-state protocol in this PR. In the current code, the follower marks `streamingRunning` only when it receives an empty success response. Full snapshot and retained catch-up responses carry region records, so a catch-up error after the full snapshot does not send that empty success response.

Making the success transition explicit with a completion marker is reasonable, but it changes the stream protocol semantics and should be handled in a follow-up PR rather than expanding this one.

pkg/syncer/server.go
okJiang: Correction after narrowing the PR scope: the previous e512d015e / 8002488cd replies are obsolete. The current head is based on e83d12646, plus only the small replay-batching fix in 46e7b3c9a.

This requested change is still outside the current PR boundary. It is the existing incremental history replay ordering gap: `syncHistoryRegion` sends the initial history response before binding the stream, so broadcasts produced during that send are not queued for this follower. The adaptive buffer change does not introduce that ordering gap.

I left a TODO in `syncHistoryRegion` and would handle the stream binding / send-ordering semantics in a follow-up PR instead of mixing that design change into this one.

okJiang:
Checked the current-head CI failures on 46e7b3c9a.

The PR diff is limited to `pkg/syncer/*`, while the failed jobs are outside that area:

- `pull-unit-test-next-gen-2`: `tests/server/id` / `TestMonotonicID` failed while starting a test cluster with `Etcd cluster ID mismatch` after port-conflict retries.
- `pull-unit-test-next-gen-3`: failures are in MCS keyspace/resource-manager tests, including `TestKeyspaceGroupTestSuite/TestUpdateMemberWhenRecovery` and `TestResourceManagerClientTestSuite/standalone-resource-manager-with-client-discovery/TestResourceManagerClientFailover`.
- GitHub `chunks (9, Microservice Integration(!TSO))`: MCS resource-manager/keyspace failures, including a timing assertion in `TestResourceGroupController` and `TestUpdateMemberWhenRecovery`.

These do not involve the syncer history-buffer changes in this PR. Local focused verification for the touched package passed on this head:

- `make gotest GOTEST_ARGS="./pkg/syncer -count=1"`
- `make static PACKAGE_DIRECTORIES="./pkg/syncer"`

No manual `/test` or `/retest` was posted.

pkg/syncer/server.go
```diff
@@ -406,12 +450,35 @@ func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream
 	}
 	log.Info("requested server has completed full synchronization with server",
 		zap.String("requested-server", name), zap.String("server", s.server.Name()), zap.Duration("cost", time.Since(start)))
-	return nil
+	syncStream := newRegionSyncStream(stream)
+	syncStream.sendMu.Lock()
```
lhy1024, line 454: There is one more timeout semantic issue here. After full sync finishes, the stream is bound first, and then `syncStream.sendMu` is held while replaying the retained catch-up history. This correctly prevents broadcasts from interleaving with catch-up responses. However, any new broadcast will enter `sendRegionSyncResponse` and block on the same `sendMu`, while the `sendTimeout` timer has already started.

If the retained history is large, or the follower/network is slow, catch-up replay may take longer than `sendTimeout`. In that case, the broadcast path closes and removes the stream even though the follower is still actively receiving full-sync catch-up data. `Sync` will then return Unavailable via `syncStream.done`, the client reconnects, and it may trigger full sync again. Under high churn this can reintroduce repeated full-sync/reconnect loops, which is what this PR is trying to avoid.

I verified this with a small test: block the catch-up send, trigger a broadcast with a short `sendTimeout`, and the broadcast removes the stream because it times out waiting on `sendMu`.

Please avoid counting time spent waiting for the internal catch-up send lock as broadcast send timeout, or make broadcasts skip/queue/not close streams that are still in the full-sync catch-up phase. The key point is not to treat an actively catching-up follower as a timed-out broadcast receiver.

pkg/syncer/server.go
```diff
@@ -301,54 +334,65 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
+		return nil, nil
 	}
 	log.Info("sync the history regions with server",
 		zap.String("server", name),
 		zap.Uint64("from-index", startIndex),
-		zap.Uint64("last-index", s.history.getNextIndex()),
+		zap.Uint64("last-index", nextIndex),
 		zap.Int("records-length", len(records)))
-	return s.syncHistoryRecords(startIndex, records, stream)
+	// TODO: Incremental history replay is sent before binding the stream, so
```
rleungx, line 350: This gap should be fixed in this PR instead of left as a follow-up. During incremental replay the stream is not bound yet, so broadcasts produced while syncHistoryRecords is sending are dropped for this follower. With the larger adaptive history window and multi-batch replay, this can skip updates and leave stale regions on the follower.

okJiang: I agree this ordering gap is real, and batching can make the replay window more visible.

I would still keep it out of this PR. The stream was already bound only after the initial history replay before this PR; fixing it correctly requires changing the stream binding / send-ordering semantics, not just the adaptive buffer implementation. That design also needs to define how replay, normal broadcasts, close responses, and stale queued responses are ordered on the same stream.

For this PR, I would like to keep the scope limited to the adaptive history buffer and the small replay-batching fix needed to keep responses below the client receive limit. I left the TODO here so the ordering gap can be handled in a focused follow-up PR.

pkg/syncer/server.go
okJiang: Yes, this is the issue covered by #10716.

I agree the timeout behavior is a valid concern, but I would keep it out of this adaptive-buffer PR. #10716 changes the stream model so each follower stream has its own buffer and next index. While a follower is still doing full/history sync and its send lock is held, broadcasts append to that follower's stream buffer instead of blocking on the lock and starting a send timeout against the lock wait. The stream then drains its own buffered records after the initial sync.

So #10712 can stay scoped to the adaptive history buffer and replay response batching, while this timeout/catch-up behavior is handled in #10716.

lhy1024 (approved):

rleungx (approved):

PR #10713, opened 2026-05-26
schedulers: increase leader batch limit

rleungx:

Issue Number: ref #8265

```commit-message
Increase the maximum configurable batch size for evict-leader-scheduler
and evict-slow-store-scheduler from 10 to 100.

Add tests to cover the new evict leader batch limit and the interaction
with scheduler-max-waiting-operator, so the larger batch size does not
imply changes to the existing waiting operator admission limit.
```

### Check List

Tests

- Unit test

Code changes

Side effects

### Release note

```release-note
Increase the maximum configurable batch size of evict-leader-scheduler and evict-slow-store-scheduler from 10 to 100.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Raised evict-leader and evict-slow-store batch cap to 100.
  * Improved batch validation with clearer 400 responses for non-float, non-integer, or out-of-range inputs.
  * Fixed pause/resume so leader-transfer resumes only when it was actually paused.

* **Tests**
  * Added and expanded tests and a benchmark covering batch limits, invalid inputs, operator caps, and scheduling behavior.

### DIFF
--- a/pkg/schedule/schedulers/evict_leader.go
+++ b/pkg/schedule/schedulers/evict_leader.go
@@ -41,8 +41,8 @@ import (
 const (
 	// EvictLeaderBatchSize is the number of operators to transfer
 	// leaders by one scheduling
-	EvictLeaderBatchSize = 3
-	lastStoreDeleteInfo  = "The last store has been deleted"
+	EvictLeaderBatchSize    = 3
+	maxEvictLeaderBatchSize = 100
 )
 
 type evictLeaderSchedulerConfig struct {
@@ -228,7 +228,7 @@ func (conf *evictLeaderSchedulerConfig) delete(id uint64) (any, error) {
 		}
 		return resp, err
 	}
-	resp = lastStoreDeleteInfo
+	resp = "The last store has been deleted"
 	return resp, nil
 }
 
@@ -419,9 +419,9 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	batch := handler.config.getBatch()
 	batchFloat, ok := input["batch"].(float64)
 	if ok {
-		if batchFloat < 1 || batchFloat > 10 {
+		if batchFloat < 1 || batchFloat > maxEvictLeaderBatchSize {
 			handler.config.resumeLeaderTransferIfExist(id)
-			handler.rd.JSON(w, http.StatusBadRequest, "batch is invalid, it should be in [1, 10]")
+			handler.rd.JSON(w, http.StatusBadRequest, "batch is invalid, it should be in [1, 100]")
 			return
 		}
 		batch = (int)(batchFloat)
--- a/pkg/schedule/schedulers/evict_leader.go
+++ b/pkg/schedule/schedulers/evict_leader.go
@@ -45,6 +45,12 @@ const (
 	maxEvictLeaderBatchSize = 100
 )
 
+func isValidEvictLeaderBatchSize(batchFloat float64) bool {
+	return batchFloat >= 1 &&
+		batchFloat <= maxEvictLeaderBatchSize &&
+		batchFloat == float64(int(batchFloat))
+}
+
 type evictLeaderSchedulerConfig struct {
 	syncutil.RWMutex
 	schedulerConfig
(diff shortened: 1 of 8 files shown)


pkg/schedule/schedulers/evict_leader.go
```diff
@@ -41,10 +41,16 @@ import (
 const (
 	// EvictLeaderBatchSize is the number of operators to transfer
 	// leaders by one scheduling
-	EvictLeaderBatchSize = 3
-	lastStoreDeleteInfo  = "The last store has been deleted"
```
lhy1024, line 45: why remove it?

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -419,9 +425,9 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	batch := handler.config.getBatch()
 	batchFloat, ok := input["batch"].(float64)
 	if ok {
```
lhy1024: Non-numeric `batch` values are still silently accepted here. For example, with `{"batch":"abc"}`, `ok` is false, validation is skipped, the old batch value is reused, and the request returns 200. This is inconsistent with the evict-slow-store-scheduler validation and with the new error semantics.

lhy1024:
dco failed

pkg/schedule/schedulers/evict_leader.go
```diff
@@ -417,11 +426,16 @@ func (handler *evictLeaderHandler) updateConfig(w http.ResponseWriter, r *http.R
 	}
 
 	batch := handler.config.getBatch()
-	batchFloat, ok := input["batch"].(float64)
-	if ok {
-		if batchFloat < 1 || batchFloat > 10 {
+	batchFloat, inputBatch := input["batch"].(float64)
+	if input["batch"] != nil && !inputBatch {
+		handler.config.resumeLeaderTransferIfExist(id)
```
lhy1024: Could we confirm whether this behavior is expected? This new non-numeric batch error path calls `resumeLeaderTransferIfExist(id)`. If the request includes an already existing evict store, for example `{"store_id":1,"batch":"abc"}`, this request does not seem to perform a new `PauseLeaderTransfer`, but this branch still resumes the existing pause state once.

  I did a quick temporary unit-test check: after `PrepareConfig` pauses store 1, sending the request above returns 400 and `AllowLeaderTransferIn()` becomes true. In that state, the store is still in the evict config, but leader transfer-in is allowed again.

  Is that intended? If not, we may need to distinguish a store newly paused by this request from a store that was already in the evict config, and only resume the former on the error path.

pkg/schedule/schedulers/evict_leader.go
rleungx: Handled earlier: non-numeric batch now returns 400, and TestEvictLeaderBatchLimit covers {"batch":"abc"}.

pkg/schedule/schedulers/evict_leader.go
rleungx: It is not removed. The diff is just gofmt alignment after adding maxEvictLeaderBatchSize; EvictLeaderBatchSize still remains 3 as the default batch size.

pkg/schedule/schedulers/evict_leader.go
rleungx: Fixed in 7976a08ee: the handler now tracks whether the current request newly paused leader transfer and only resumes that case on validation errors. TestEvictLeaderInvalidBatchKeepsLeaderTransferState covers both existing and newly added stores.

pkg/schedule/schedulers/evict_slow_store.go
```diff
@@ -310,8 +310,8 @@ func (handler *evictSlowStoreHandler) updateConfig(w http.ResponseWriter, r *htt
 		return
 	}
 	if inputBatch {
-		if batchFloat < 1 || batchFloat > 10 {
-			handler.rd.JSON(w, http.StatusBadRequest, "batch is invalid, it should be in [1, 10]")
+		if !isValidEvictLeaderBatchSize(batchFloat) {
```
JmPotato: Can be merged into one line if.

rleungx: Done in 7e5081341: merged the nested batch validation into a single if without changing behavior.

lhy1024 (approved):

JmPotato (commented):
**`pkg/schedule/schedulers/evict_leader.go:436`** — 错误信息缺少诊断值

当用户传 `"batch": "abc"` 这类非法类型时，返回 `"invalid argument for 'batch'"` 但不包含实际传入的类型/值。建议包含诊断信息以便排查，例如 `fmt.Sprintf("invalid argument for 'batch': expected a number, got %T", input["batch"])`。

JmPotato (commented):
**`pkg/schedule/schedulers/evict_leader.go:436`** — error message lacks diagnostic value

When a non-numeric value like `"batch": "abc"` is passed, the response is "invalid argument for 'batch'" without indicating what was actually received. Suggest including the type or value for debuggability, e.g. `fmt.Sprintf("invalid argument for 'batch': expected a number, got %T", input["batch"])`.

JmPotato (approved):

Issue #10690, opened 2026-05-19
client: only send follower region requests after follower full region sync


okJiang:
## Development Task

Tracking follow-up for #10681 / #10682 and #10668.

### Background

PD followers can serve region requests from their local region cache when follower handle is enabled. This is only safe after the follower has completed its initial full region sync with the leader.

If PD client sends region requests to a PD follower before the follower region cache is fully synced, the follower may return empty, incomplete, or stale region data. This can happen during follower startup, after follower local cache reset, or during recovery that requires a full region sync.

### Goal

PD client should only send follower region requests to a PD follower after that follower has completed full region sync.

Expected behavior:

- Before follower full region sync completes, PD client should not route region requests to that follower.
- Requests should fall back to the PD leader or another ready endpoint.
- After follower cache reset, the follower should become unavailable for follower region requests until full sync completes again.
- “syncer running” should not be treated as enough unless it also means the full region view is ready.

### Non-goals

- This issue does not track the full-sync fallback itself; that is tracked by #10668.
- This issue does not track follower-local HTTP region read APIs; that is tracked by #10681 / #10682.
- This issue does not track follower reset cache APIs; that is tracked by #10667.

### Validation

- Add coverage where PD client enables follower handle before follower full region sync completes.
- Verify region requests are not sent to the not-ready follower.
- Verify requests can be sent to the follower after full region sync completes.
- Verify resetting follower region cache makes that follower temporarily unavailable for follower region requests until it finishes full sync again.

Issue #10668, opened 2026-05-12
region syncer: trigger full region sync when follower index gap exceeds history buffer size


okJiang:
## Development Task

Tracking implementation task for #10666.

### Background

When Active Follower is enabled, PD followers serve region reads from their local region cache. The follower cache is maintained by the region syncer: the follower loads existing region metadata from local region storage, then keeps catching up from the leader through incremental sync.

The current region syncer depends on the leader-side history buffer. If the follower falls behind the leader by more than the history buffer size, some updates can no longer be replayed. After the sync stream recovers, the follower may still keep stale region metadata indefinitely, unless those regions receive newer updates later.

### Goal

Make region syncer recover safely when the follower's index gap is larger than the leader-side history buffer.

Expected behavior:

- The PD leader should detect that the follower's requested index is older than the available history buffer.
- The PD leader should actively fall back to sending a full region snapshot.
- The PD follower should clear stale local region cache/state before accepting the full snapshot.
- The PD follower should persist the rebuilt region data back to local region storage.
- After recovery, follower region cache should converge with the leader's region view.

### Suggested implementation direction

- Add explicit gap detection in the leader-side region sync path.
- Treat unrecoverable history gaps as a full-sync-required condition instead of silently returning no records.
- Add a follower-side reset path before applying the full snapshot.
- Ensure local region storage and in-memory region cache are rebuilt consistently.
- Add metrics/logs for history buffer overflow and full sync fallback.

### Validation

- Add a test where follower `startIndex` is older than the leader history buffer and verify full sync is triggered.
- Verify stale regions are removed from follower cache after full sync.
- Verify follower local region storage is rebuilt with the leader's latest region metadata.
- Verify normal incremental sync behavior is unchanged when the gap is within the history buffer.
- Verify Active Follower reads from the recovered follower return the same region view as the leader.

okJiang:
Resolved by #10689, built on the stream state isolation from #10716.

The merged fix now falls back to full region sync when the follower's requested index cannot be served from the leader history window, including StartIndex == 0 and StartIndex > endIndex. It also sends a full-sync completion response and keeps the follower region syncer from being marked running until full sync is complete.

Closing this as completed.

PR #10716, opened 2026-05-26
syncer: isolate downstream state by stream

### LINKED ISSUE #10715: syncer: isolate RegionSyncer follower state by stream
## Enhancement Task

RegionSyncer currently sends sync data from the leader through a shared broadcast path. This issue is intended to be a prerequisite for #10668 and #10690, so the foundation needs to make each follower stream own its send progress instead of sharing one leader-side broadcast cursor.

### Background

When one follower is doing history sync or full region sync, other healthy followers can continue receiving live region updates from the leader. The syncing follower must not inherit the send progress of those healthy followers. Otherwise, after its history/full sync finishes, its next live sync index can be ahead of the data it actually received.

This matters for follow-up work:

- #10668 needs safe full-sync fallback when a follower is too far behind the leader history buffer.
- #10690 needs a clear point at which a follower has completed full sync and is safe for follower region reads.

### Goal

Keep a single leader-side history buffer, and make each downstream stream maintain and advance its own send index.

Expected behavior:

- Keep the existing global history buffer for reconnect history sync and downstream live replay.
- Do not add a separate per-follower history buffer.
- Each follower stream has its own next send index.
- Live broadcast records region changes once in the global history and triggers each follower stream to drain from its own index.
- A follower doing history/full sync uses the same stream send lock, so live records for that follower cannot interleave with the initial sync.
- Other healthy followers continue sending and advancing independently while one follower is still catching up.
- Send success advances only that follower stream's index.
- Send failure removes only that follower stream.
- If a follower falls behind the retained global history, close that follower stream and let follow-up work handle the full fallback path.

### Non-goals

- Do not change the SyncRegions protobuf protocol.
- Do not add ack handling.
- Do not implement the full history-gap fallback from #10668.
- Do not reset or rebuild follower local region cache in this task.
- Do not change PD client follower-read routing readiness from #10690.

### Relationship

- Prerequisite for #10668.
- Prerequisite for #10690.
- Related to #10666.

### Validation

- Verify two followers maintain independent send indexes while sharing one leader history buffer.
- Verify one follower send failure does not affect another follower.
- Verify a busy follower stream does not block healthy followers.
- Verify the busy follower later drains from its own index.
- Verify full/history sync only replays records up to the stream's bind-time leader index.

okJiang:

Issue Number: Close #10715

Related to #10666.

Prerequisite for #10668 and #10690.

This PR prepares RegionSyncer for safe follower recovery. The important case is when one follower is doing history sync or full region sync while other healthy followers continue receiving live region updates. The syncing follower must not inherit another follower's send progress.

```commit-message
Split RegionSyncer leader-side downstream send progress by follower stream.

The leader keeps one shared RegionSyncer history buffer. Each downstream stream keeps only its own next send index and serial send lock. Live region updates are recorded once in the shared history, and each stream drains pending records from that shared history according to its own index.

A successful send advances only that stream. A failed send removes only that stream. History/full sync and live sync on the same follower stream are serialized by the same send lock, so live records cannot interleave with the initial sync. Other followers continue sending and advancing independently while one follower is still catching up.

This preserves the current SyncRegions protocol and does not add ack handling. The full history-gap fallback, follower cache reset/rebuild, and client follower-read readiness changes remain in follow-up issues #10668 and #10690.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

### DIFF
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -57,10 +57,10 @@ type historyBuffer struct {
 	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
 	lowWindowRounds int
 	// retains tracks full-sync start indexes that must stay replayable until catch-up finishes.
-	retains      map[uint64]int
-	kv           kv.Base
-	persistIndex bool
-	flushCount   int
+	retains             map[uint64]int
+	kv                  kv.Base
+	persistHistoryIndex bool
+	flushCount          int
 }
 
 func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
@@ -73,11 +73,13 @@ func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv 
 	return h
 }
 
+// newMemoryHistoryBuffer creates a process-local buffer with an explicit next index.
+// It does not reload from KV because downstream buffers are not persisted.
 func newMemoryHistoryBuffer(size int, index uint64) *historyBuffer {
 	return newHistoryBufferWithConfigAndIndex(size, size, historyBufferCapacityUnit, index, nil, false)
 }
 
-func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit int, index uint64, kv kv.Base, persistIndex bool) *historyBuffer {
+func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit int, index uint64, kv kv.Base, persistHistoryIndex bool) *historyBuffer {
 	baseCapacity = normalizeHistoryBufferCapacity(baseCapacity, capacityUnit)
 	maxCapacity = normalizeHistoryBufferCapacity(maxCapacity, capacityUnit)
 	if maxCapacity < baseCapacity {
@@ -87,15 +89,15 @@ func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit 
 	size := baseCapacity + 1
 	records := make([]*core.RegionInfo, size)
 	return &historyBuffer{
-		index:        index,
-		records:      records,
-		size:         size,
-		baseCapacity: baseCapacity,
-		maxCapacity:  maxCapacity,
-		capacityUnit: capacityUnit,
-		kv:           kv,
-		persistIndex: persistIndex,
-		flushCount:   defaultFlushCount,
+		index:               index,
+		records:             records,
+		size:                size,
+		baseCapacity:        baseCapacity,
+		maxCapacity:         maxCapacity,
+		capacityUnit:        capacityUnit,
+		kv:                  kv,
+		persistHistoryIndex: persistHistoryIndex,
+		flushCount:          defaultFlushCount,
 	}
 }
 
@@ -139,7 +141,7 @@ func (h *historyBuffer) record(r *core.RegionInfo) {
 	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
 		h.growForWindowLocked(h.index - retainIndex + 1)
 	}
-	if h.persistIndex {
+	if h.persistHistoryIndex {
 		syncIndexGauge.Set(float64(h.index))
 	}
 	h.records[h.tail] = r
@@ -148,7 +150,7 @@ func (h *historyBuffer) record(r *core.RegionInfo) {
 		h.head = (h.head + 1) % h.size
 	}
 	h.index++
-	if !h.persistIndex {
+	if !h.persistHistoryIndex {
 		return
 	}
 	h.flushCount--
@@ -298,6 +300,14 @@ func (h *historyBuffer) resetWithIndex(index uint64) {
 	h.resetWithIndexLocked(index)
 }
 
+func (h *historyBuffer) advanceToIndex(index uint64) {
+	h.Lock()
+	defer h.Unlock()
+	if h.index <= index {
+		h.resetWithIndexLocked(index)
+	}
+}
+
 func (h *historyBuffer) resetWithIndexLocked(index uint64) {
 	h.index = index
 	h.head = 0
@@ -330,7 +340,7 @@ func (h *historyBuffer) get(index uint64) *core.RegionInfo {
 }
 
 func (h *historyBuffer) reload() {
-	if !h.persistIndex {
+	if !h.persistHistoryIndex {
 		return
 	}
 	v, err := h.kv.Load(historyKey)
@@ -347,7 +357,7 @@ func (h *historyBuffer) reload() {
 }
 
 func (h *historyBuffer) persist() {
-	if !h.persistIndex {
+	if !h.persistHistoryIndex {
 		return
 	}
 	firstIndexGauge.Set(float64(h.firstIndex()))
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -57,9 +57,10 @@ type historyBuffer struct {
 	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
 	lowWindowRounds int
 	// retains tracks full-sync start indexes that must stay replayable until catch-up finishes.
-	retains    map[uint64]int
-	kv         kv.Base
-	flushCount int
+	retains      map[uint64]int
+	kv           kv.Base
+	persistIndex bool
+	flushCount   int
 }
 
 func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
@@ -67,6 +68,16 @@ func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer 
 }
 
 func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv kv.Base) *historyBuffer {
+	h := newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit, 0, kv, true)
+	h.reload()
+	return h
+}
+
+func newMemoryHistoryBuffer(size int, index uint64) *historyBuffer {
+	return newHistoryBufferWithConfigAndIndex(size, size, historyBufferCapacityUnit, index, nil, false)
+}
+
+func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit int, index uint64, kv kv.Base, persistIndex bool) *historyBuffer {
 	baseCapacity = normalizeHistoryBufferCapacity(baseCapacity, capacityUnit)
 	maxCapacity = normalizeHistoryBufferCapacity(maxCapacity, capacityUnit)
 	if maxCapacity < baseCapacity {
@@ -75,17 +86,17 @@ func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv 
 	// use an empty space to simplify operation
 	size := baseCapacity + 1
 	records := make([]*core.RegionInfo, size)
-	h := &historyBuffer{
+	return &historyBuffer{
+		index:        index,
 		records:      records,
 		size:         size,
 		baseCapacity: baseCapacity,
 		maxCapacity:  maxCapacity,
 		capacityUnit: capacityUnit,
 		kv:           kv,
+		persistIndex: persistIndex,
 		flushCount:   defaultFlushCount,
 	}
-	h.reload()
-	return h
 }
 
 func normalizeHistoryBufferCapacity(size, capacityUnit int) int {
@@ -128,13 +139,18 @@ func (h *historyBuffer) record(r *core.RegionInfo) {
 	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
 		h.growForWindowLocked(h.index - retainIndex + 1)
 	}
-	syncIndexGauge.Set(float64(h.index))
+	if h.persistIndex {
+		syncIndexGauge.Set(float64(h.index))
+	}
 	h.records[h.tail] = r
 	h.tail = (h.tail + 1) % h.size
 	if h.tail == h.head {
 		h.head = (h.head + 1) % h.size
 	}
 	h.index++
+	if !h.persistIndex {
+		return
+	}
 	h.flushCount--
 	if h.flushCount <= 0 {
 		h.persist()
@@ -149,27 +165,29 @@ func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
 }
 
 func (h *historyBuffer) recordsFromLocked(index uint64) []*core.RegionInfo {
-	var pos int
-	if index < h.nextIndex() && index >= h.firstIndex() {
-		pos = (h.head + int(index-h.firstIndex())) % h.size
-	} else {
-		return nil
-	}
-	records := make([]*core.RegionInfo, 0, h.distanceToTail(pos))
-	for i := pos; i != h.tail; i = (i + 1) % h.size {
-		records = append(records, h.records[i])
-	}
-	return records
+	return h.recordsBetweenLocked(index, h.nextIndex())
 }
 
-func (h *historyBuffer) observeAndRecordsFrom(index uint64) ([]*core.RegionInfo, uint64) {
-	h.Lock()
-	defer h.Unlock()
-	nextIndex := h.nextIndex()
-	if index != 0 && index < nextIndex {
-		h.observeRequiredWindowLocked(nextIndex - index)
+func (h *historyBuffer) recordsBetween(index, endIndex uint64) []*core.RegionInfo {
+	h.RLock()
+	defer h.RUnlock()
+	return h.recordsBetweenLocked(index, endIndex)
+}
+
+func (h *historyBuffer) recordsBetweenLocked(index, endIndex uint64) []*core.RegionInfo {
+	if endIndex > h.nextIndex() {
+		endIndex = h.nextIndex()
 	}
-	return h.recordsFromLocked(index), nextIndex
+	if index >= endIndex || index < h.firstIndex() || index > h.nextIndex() {
+		return nil
+	}
+	pos := (h.head + int(index-h.firstIndex())) % h.size
+	records := make([]*core.RegionInfo, 0, int(endIndex-index))
+	for i := pos; index < endIndex; i = (i + 1) % h.size {
+		records = append(records, h.records[i])
+		index++
+	}
+	return records
 }
 
 func (h *historyBuffer) retainedRecordsFrom(index uint64) ([]*core.RegionInfo, uint64, bool) {
@@ -183,15 +201,18 @@ func (h *historyBuffer) retainedRecordsFrom(index uint64) ([]*core.RegionInfo, u
 	return records, nextIndex, len(records) == int(nextIndex-index)
 }
 
-func (h *historyBuffer) retainFromCurrent() (uint64, func()) {
+func (h *historyBuffer) retainFrom(index uint64) func() {
 	h.Lock()
 	defer h.Unlock()
-	index := h.nextIndex()
+	return h.retainLocked(index)
+}
+
+func (h *historyBuffer) retainLocked(index uint64) func() {
 	if h.retains == nil {
 		h.retains = make(map[uint64]int)
 	}
 	h.retains[index]++
-	return index, func() {
+	return func() {
 		h.releaseRetain(index)
 	}
 }
@@ -296,6 +317,12 @@ func (h *historyBuffer) getNextIndex() uint64 {
 	return h.index
 }
 
+func (h *historyBuffer) getFirstIndex() uint64 {
+	h.RLock()
+	defer h.RUnlock()
+	return h.firstIndex()
+}
+
 func (h *historyBuffer) get(index uint64) *core.RegionInfo {
 	h.RLock()
 	defer h.RUnlock()
@@ -303,6 +330,9 @@ func (h *historyBuffer) get(index uint64) *core.RegionInfo {
 }
 
 func (h *historyBuffer) reload() {
+	if !h.persistIndex {
+		return
+	}
 	v, err := h.kv.Load(historyKey)
 	if err != nil {
 		log.Warn("load history index failed", zap.String("error", err.Error()))
@@ -317,6 +347,9 @@ func (h *historyBuffer) reload() {
 }
 
 func (h *historyBuffer) persist() {
+	if !h.persistIndex {
+		return
+	}
 	firstIndexGauge.Set(float64(h.firstIndex()))
 	lastIndexGauge.Set(float64(h.nextIndex()))
 	err := h.kv.Save(historyKey, strconv.FormatUint(h.nextIndex(), 10))
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -57,10 +57,9 @@ type historyBuffer struct {
 	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
 	lowWindowRounds int
 	// retains tracks full-sync start indexes that must stay replayable until catch-up finishes.
-	retains             map[uint64]int
-	kv                  kv.Base
-	persistHistoryIndex bool
-	flushCount          int
+	retains    map[uint64]int
+	kv         kv.Base
+	flushCount int
 }
 
 func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer {
@@ -68,18 +67,6 @@ func newHistoryBuffer(baseCapacity, maxCapacity int, kv kv.Base) *historyBuffer 
 }
 
 func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv kv.Base) *historyBuffer {
-	h := newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit, 0, kv, true)
-	h.reload()
-	return h
-}
-
-// newMemoryHistoryBuffer creates a process-local buffer with an explicit next index.
-// It does not reload from KV because downstream buffers are not persisted.
-func newMemoryHistoryBuffer(size int, index uint64) *historyBuffer {
-	return newHistoryBufferWithConfigAndIndex(size, size, historyBufferCapacityUnit, index, nil, false)
-}
-
-func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit int, index uint64, kv kv.Base, persistHistoryIndex bool) *historyBuffer {
 	baseCapacity = normalizeHistoryBufferCapacity(baseCapacity, capacityUnit)
 	maxCapacity = normalizeHistoryBufferCapacity(maxCapacity, capacityUnit)
 	if maxCapacity < baseCapacity {
@@ -88,17 +75,17 @@ func newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit 
 	// use an empty space to simplify operation
 	size := baseCapacity + 1
 	records := make([]*core.RegionInfo, size)
-	return &historyBuffer{
-		index:               index,
-		records:             records,
-		size:                size,
-		baseCapacity:        baseCapacity,
-		maxCapacity:         maxCapacity,
-		capacityUnit:        capacityUnit,
-		kv:                  kv,
-		persistHistoryIndex: persistHistoryIndex,
-		flushCount:          defaultFlushCount,
+	h := &historyBuffer{
+		records:      records,
+		size:         size,
+		baseCapacity: baseCapacity,
+		maxCapacity:  maxCapacity,
+		capacityUnit: capacityUnit,
+		kv:           kv,
+		flushCount:   defaultFlushCount,
 	}
+	h.reload()
+	return h
 }
 
 func normalizeHistoryBufferCapacity(size, capacityUnit int) int {
@@ -141,18 +128,13 @@ func (h *historyBuffer) record(r *core.RegionInfo) {
 	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
 		h.growForWindowLocked(h.index - retainIndex + 1)
 	}
-	if h.persistHistoryIndex {
-		syncIndexGauge.Set(float64(h.index))
-	}
+	syncIndexGauge.Set(float64(h.index))
 	h.records[h.tail] = r
 	h.tail = (h.tail + 1) % h.size
 	if h.tail == h.head {
 		h.head = (h.head + 1) % h.size
 	}
 	h.index++
-	if !h.persistHistoryIndex {
-		return
-	}
 	h.flushCount--
 	if h.flushCount <= 0 {
 		h.persist()
@@ -300,14 +282,6 @@ func (h *historyBuffer) resetWithIndex(index uint64) {
 	h.resetWithIndexLocked(index)
 }
 
-func (h *historyBuffer) advanceToIndex(index uint64) {
-	h.Lock()
-	defer h.Unlock()
-	if h.index <= index {
-		h.resetWithIndexLocked(index)
-	}
-}
-
 func (h *historyBuffer) resetWithIndexLocked(index uint64) {
 	h.index = index
 	h.head = 0
@@ -340,9 +314,6 @@ func (h *historyBuffer) get(index uint64) *core.RegionInfo {
 }
 
 func (h *historyBuffer) reload() {
-	if !h.persistHistoryIndex {
-		return
-	}
 	v, err := h.kv.Load(historyKey)
 	if err != nil {
 		log.Warn("load history index failed", zap.String("error", err.Error()))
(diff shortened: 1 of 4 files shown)


pkg/syncer/server.go
```diff
@@ -430,7 +435,13 @@ func (s *RegionSyncer) unbindStream(name string, stream *regionSyncStream) {
...
 }
 
-func (s *RegionSyncer) broadcast(ctx context.Context, regions *pdpb.SyncRegionResponse) {
+func cloneSyncRegionResponse(regions *pdpb.SyncRegionResponse, startIndex uint64) *pdpb.SyncRegionResponse {
+	resp := *regions
+	resp.StartIndex = startIndex
+	return &resp
+}
+
+func (s *RegionSyncer) broadcast(ctx context.Context, regions *pdpb.SyncRegionResponse, records []*core.RegionInfo) {
```
okJiang: What is the difference if the records passed in via broadcast are the same, even if the start indices are different? It seems that the issue with fragmented regions (shredded edges) still hasn't been resolved.

okJiang: Yes, this PR does not claim to resolve the shredded-edge case yet.

In the current normal broadcast path, the records passed to each follower are still the same batch. The intended change here is only to stop sharing the server-side send state across followers: each stream now has its own history index and buffer, so a later resend/replay path can select records according to that specific follower state instead of being tied to a single shared broadcast state.

So the difference in this PR is state isolation, not data selection. The actual gap/shredded-edge recovery still needs a follow-up on top of this foundation, and I kept that out of scope intentionally to avoid mixing the state refactor with resend semantics.

okJiang: I would like to clarify this point carefully.

Suppose the start indexes of certain followers are already different (meaning the indexes of two followers do not match). In that case, my understanding is that the region information they need to send must change according to their respective indexes.

Sending the same region information across different indexes would definitely be incorrect. Do you agree with this?

okJiang: Yes, I agree.

If a follower is truly at a different missing index, sending the same records with only a different StartIndex would be incorrect for recovery. The records selected for that follower must match its requested or missing index.

This is why this PR does not claim to fix the gap or shredded-edge recovery case. It only separates each downstream send state so the server no longer has one shared send cursor. The follow-up resend logic still needs to select the records per follower based on that follower index.

pkg/syncer/server.go
```diff
@@ -409,16 +413,17 @@ func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream
 	return nil
 }
 
-// bindStream binds the established server stream.
-func (s *RegionSyncer) bindStream(name string, stream ServerStream) *regionSyncStream {
-	syncStream := newRegionSyncStream(stream)
+func (s *RegionSyncer) bindStreamForSync(name string, stream ServerStream) (*regionSyncStream, uint64) {
+	startIndex := s.history.getNextIndex()
+	syncStream := newRegionSyncStream(stream, startIndex)
+	syncStream.sendMu.Lock()
```
okJiang: Why aren't the `lock` and `unlock` operations placed together?

This doesn't seem to comply with standard coding conventions.

pkg/syncer/server.go
```diff
@@ -242,12 +232,20 @@ func (s *RegionSyncer) Sync(ctx context.Context, stream pdpb.PD_SyncRegionsServe
 			zap.String("requested-server", request.GetMember().GetName()),
 			zap.String("url", request.GetMember().GetClientUrls()[0]))
 
-		err = s.syncHistoryRegion(ctx, request, stream)
+		name := request.GetMember().GetName()
+		syncStream, syncStartIndex := s.bindStreamForSync(name, stream)
+		err = s.syncHistoryRegion(ctx, request, stream, syncStartIndex)
 		if err != nil {
+			syncStream.sendMu.Unlock()
```
okJiang: ditto

pkg/syncer/server.go
```diff
@@ -242,12 +232,20 @@ func (s *RegionSyncer) Sync(ctx context.Context, stream pdpb.PD_SyncRegionsServe
...
+			s.unbindStream(name, syncStream)
+			return err
+		}
+		err = s.drainDownstreamLocked(ctx, name, syncStream)
+		syncStream.sendMu.Unlock()
```
okJiang: ditto

pkg/syncer/server.go
```diff
@@ -430,54 +435,117 @@ func (s *RegionSyncer) unbindStream(name string, stream *regionSyncStream) {
...
+			return errors.Errorf("region syncer buffered records from index %d overflow, first available index is %d", nextIndex, firstIndex)
+		}
+		bufferNextIndex := stream.history.getNextIndex()
+		if nextIndex > bufferNextIndex {
+			return errors.Errorf("region syncer buffered records index %d exceeds next index %d", nextIndex, bufferNextIndex)
+		}
+		if nextIndex == bufferNextIndex {
+			return nil
+		}
+		records := stream.history.recordsBetween(nextIndex, bufferNextIndex)
```
okJiang: The `maxSyncRegionBatchSize` should also be taken into account during `drainDownstreamLocked`.

pkg/syncer/server.go
```diff
@@ -409,16 +413,17 @@ func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream
 	return nil
 }
 
-// bindStream binds the established server stream.
-func (s *RegionSyncer) bindStream(name string, stream ServerStream) *regionSyncStream {
-	syncStream := newRegionSyncStream(stream)
+func (s *RegionSyncer) bindStreamForSync(name string, stream ServerStream) (*regionSyncStream, uint64) {
+	startIndex := s.history.getNextIndex()
```
lhy1024, line 549: There is a bind-time race here. `startIndex` is read before taking `s.mu`, but the new stream is inserted into `s.mu.streams` only later. If a region update is recorded into `s.history` and broadcast in between, that record is not replayed by `syncHistoryRegion(..., endIndex=startIndex)`, and it is also not appended to the new stream's per-stream history because the stream was not in the map during broadcast. `drainDownstreamLocked` will not see it either, so the follower can miss that update.

  I verified the interleaving with a small unit test: read `syncStartIndex=10`, then record+broadcast region 1 while `streams` is empty, then insert the stream and run history sync/drain with `endIndex=10`. The stream only receives the empty sync response; region 1 is never sent.

  Please make reading `s.history.getNextIndex()` and inserting into `s.mu.streams` atomic under the same `s.mu.Lock`, or otherwise make the bind boundary mutually exclusive with broadcast. After the bind start index is sampled, live records must either be included in history replay or appended to that stream's per-stream buffer.

pkg/syncer/server.go
okJiang: Fixed in 55abdc600. The manual lock/unlock path was removed; stream sends now go through the per-stream sender path with scoped locking.

pkg/syncer/server.go
okJiang: Handled together with the lock cleanup in 55abdc600.

pkg/syncer/server.go
okJiang: Handled together with the lock cleanup in 55abdc600.

pkg/syncer/server.go
okJiang: Fixed in 55abdc600. drainDownstreamLocked now caps each response to maxSyncRegionBatchSize before sending.

pkg/syncer/server.go
okJiang: Fixed in 55abdc600. bindStreamForSync now samples history.getNextIndex() and inserts the stream while holding s.mu.Lock. broadcast appends records by the broadcast start index and skips records already covered by a stream, so a boundary update is either included in history sync or buffered on the stream. Verified with pkg/syncer tests, region_syncer tests, and make check.

okJiang:
Rebuilt this branch on the latest `upstream/master` with `--force-with-lease` because the previous conflict resolution accidentally dropped part of #10712 full-sync catch-up history behavior.

Old head: `55abdc600`
New head: `f7fb6ced9`

What is preserved/changed:
- Preserved #10712 adaptive history buffer and full-sync catch-up retention behavior.
- Added memory-only history buffers for per-stream downstream state.
- During full sync, the leader now retains history from the stream bind index and sends catch-up records before the downstream sender starts.
- Broadcast records are still split by downstream stream with independent send indexes.

Verified:
- `make gotest GOTEST_ARGS="./pkg/syncer -run Test -count=1"`
- `make gotest GOTEST_ARGS="./tests/server/region_syncer -run TestRegionSyncer -count=1"`
- `make check`

pkg/syncer/history_buffer.go
```diff
@@ -57,16 +57,27 @@ type historyBuffer struct {
...
 	return newHistoryBufferWithConfig(baseCapacity, maxCapacity, historyBufferCapacityUnit, kv)
 }
 
 func newHistoryBufferWithConfig(baseCapacity, maxCapacity, capacityUnit int, kv kv.Base) *historyBuffer {
+	h := newHistoryBufferWithConfigAndIndex(baseCapacity, maxCapacity, capacityUnit, 0, kv, true)
+	h.reload()
+	return h
+}
+
+func newMemoryHistoryBuffer(size int, index uint64) *historyBuffer {
```
rleungx: No need to reload?

pkg/syncer/history_buffer.go
```diff
@@ -57,16 +57,27 @@ type historyBuffer struct {
 	// lowWindowRounds counts consecutive shrink checks where the required window stays low.
 	lowWindowRounds int
 	// retains tracks full-sync start indexes that must stay replayable until catch-up finishes.
-	retains    map[uint64]int
-	kv         kv.Base
-	flushCount int
+	retains      map[uint64]int
+	kv           kv.Base
+	persistIndex bool
```
rleungx: How about using another name? This variable name is more like an index instead of a bool.

pkg/syncer/server.go
```diff
@@ -334,64 +349,66 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
+		RegionStats:   stats,
+		RegionLeaders: leaders,
+		Buckets:       buckets,
+	}
+}
+
+func (*RegionSyncer) syncHistoryRecords(startIndex uint64, records []*core.RegionInfo, stream ServerStream) error {
+	for start := 0; start < len(records); start += maxSyncRegionBatchSize {
+		end := min(start+maxSyncRegionBatchSize, len(records))
+		if err := stream.Send(buildSyncRegionResponse(startIndex+uint64(start), records[start:end])); err != nil {
```
rleungx: This send bypasses regionSyncStream.sendMu even though the stream has already been bound. Initial history/full sync sends should use the same per-stream lock so they cannot race with closeAllClient or downstream sends.

pkg/syncer/server.go
```diff
@@ -450,42 +467,33 @@ func (s *RegionSyncer) syncFullRegions(ctx context.Context, name string, stream
...
 	}
-	syncStream.sendMu.Unlock()
-	return syncStream, nil
-}
-
-// bindStream binds the established server stream.
-func (s *RegionSyncer) bindStream(name string, stream ServerStream) *regionSyncStream {
-	syncStream := newRegionSyncStream(stream)
-	return s.bindRegionSyncStream(name, syncStream)
+	syncStream.history.resetWithIndex(nextIndex)
```
rleungx: Live records can be appended to this stream while the catch-up send above is still in progress. Resetting to this earlier nextIndex can drop those buffered records; please drain or reset under the same stream lock.

pkg/syncer/history_buffer.go
okJiang: Addressed in 54ff781d0. Renamed `persistIndex` to `persistHistoryIndex` so the bool meaning is clear.

pkg/syncer/server.go
okJiang: Addressed in 54ff781d0. Initial history/full sync now holds the stream send lock and sends through the bound `regionSyncStream`, so it is serialized with close and downstream sends.

pkg/syncer/server.go
okJiang: Addressed in 54ff781d0. The catch-up path now advances the per-stream buffer only when it is not already ahead, so live records appended during catch-up are preserved. Added a unit test for this race.

pkg/syncer/server.go
```diff
@@ -497,54 +526,135 @@ func (s *RegionSyncer) unbindStream(name string, stream *regionSyncStream) {
...
+	endIndex := startIndex + uint64(len(records))
+	nextIndex := s.history.getNextIndex()
+	if nextIndex >= endIndex {
+		return true
 	}
+	if nextIndex < startIndex {
+		return false
+	}
+	for _, record := range records[nextIndex-startIndex:] {
+		s.history.record(record)
```
okJiang: Is it necessary to have two history buffers for `RegionSyncer` and `regionSyncStream`? I think it's sufficient to share one; there's no need to record multiple times between different history buffers.

okJiang: Updated in 1f6bab79d. The server now keeps a single RegionSyncer history buffer; regionSyncStream only keeps its own sendIndex and reads pending records from the shared history. This removes the duplicate per-stream history buffer while preserving independent follower progress.

pkg/syncer/server.go
okJiang: Follow-up update in 1f6bab79d: the design no longer keeps a per-stream history buffer. The PR now keeps one shared RegionSyncer history buffer and only splits each stream's sendIndex. The scope is still state/progress isolation; gap and shredded-edge recovery remain follow-up work.

pkg/syncer/server.go
okJiang: Follow-up update in 1f6bab79d: after the stream is bound at the sampled history index, live updates are recorded once in the shared RegionSyncer history. The stream drains from that shared history by its own sendIndex, so the bind boundary no longer depends on a per-stream buffer.

pkg/syncer/history_buffer.go
okJiang: Follow-up update in 1f6bab79d: `newMemoryHistoryBuffer` and the per-stream history buffer were removed, so this reload path is no longer part of the PR.

pkg/syncer/server.go
okJiang: Follow-up update in 1f6bab79d: the per-stream buffer reset was removed. Full-sync catch-up now sends retained records from the shared history and advances only the stream sendIndex; live records appended during catch-up remain in the shared history and are drained afterward.

pkg/syncer/server.go
```diff
@@ -66,17 +66,45 @@ type ServerStream interface {
...
+func (s *regionSyncStream) getSendIndex() uint64 {
+	s.indexMu.RLock()
+	defer s.indexMu.RUnlock()
+	return s.sendIndex
+}
+
+func (s *regionSyncStream) advanceSendIndex(count int) {
+	s.indexMu.Lock()
+	defer s.indexMu.Unlock()
+	s.sendIndex += uint64(count)
```
okJiang: `advanceSendIndex` will only advance `sendIndex`, while the `head` in the `history buffer` seems to never advance. Could this cause the history buffer to only expand and never shrink?

okJiang: No, this should not make the history buffer only expand.

`sendIndex` is only the per-stream read cursor. It intentionally does not move the shared history buffer head. The shared history head advances in `historyBuffer.record` when the ring overwrites old entries, and `maybeShrink` can shrink capacity once there is no full-sync retain and the observed replay window stays low.

If a stream's `sendIndex` falls behind `history.firstIndex()`, `drainDownstreamLocked` treats that stream as overflowed and removes only that stream. We should not keep expanding the shared history just because one follower is too slow; the follow-up full-sync fallback should handle that case.

pkg/syncer/server.go
```diff
@@ -317,12 +320,31 @@ func recvSyncRegionRequest(ctx context.Context, stream pdpb.PD_SyncRegionsServer
...
+	ctx context.Context,
+	request *pdpb.SyncRegionRequest,
+	syncStream *regionSyncStream,
+	endIndex uint64,
+) error {
 	startIndex := request.GetStartIndex()
 	name := request.GetMember().GetName()
-	records, nextIndex := s.history.observeAndRecordsFrom(startIndex)
+	if startIndex != 0 && startIndex < endIndex {
+		s.history.observeRequiredWindow(endIndex - startIndex)
```
okJiang, line 388: Currently, `observeRequiredWindow` is only called during `syncHistoryRegion`. That doesn't seem right, does it? `syncHistoryRegion` is only called when the stream is created, which means it will only be called once and won't observe the window during subsequent sync operations. This likely isn't the intended behavior.

pkg/syncer/server.go
okJiang: However, `history.firstIndex()` is actually only related to `history.head`. You can take a closer look at the code for `firstIndex()` and `h.len()`, as well as how and when the index, head, and tail within the history are updated. You'll see that `history.head` is only updated when it's caught up by the tail. This means that after `len()` expands, it won't shrink again unless the window shrinks during stream creation (see another comment). This deviates somewhat from the design philosophy of dynamic history buffers. We should monitor the window during sync and determine whether to expand or shrink (as mentioned in another comment). Furthermore, `retainFrom` should be treated as a routine operation, used in all sync logic, not just during `syncFullRegions`/`sycnHistoryRegion`. Perhaps it could be renamed to `advance`, directly advancing the `head` field in the history buffer, using the `head` field as the basis for `retain`. When adding a record or advancing the tail, instead of directly overwriting the head, a forced expansion should be performed (if the upper limit hasn't been reached). Then, we can directly use `len()` instead of `window` to determine the scaling. Doesn't this approach seem more logical, cover more scenarios, and reduce bugs? What are your thoughts?

By the way, `history.firstIndex()` and `h.len()` seem a bit redundant and complex. If there's better code to replace them, we can try improving it. If not, it's not a big deal.

pkg/syncer/server.go
okJiang: Moved this follow-up fix to #10735. That PR observes the live sync window before appending records and advances the shared history retain boundary as downstream streams make progress.

pkg/syncer/server.go
okJiang: Moved this history-buffer retain/window change to a separate follow-up PR: #10735.

#10716 will stay focused on splitting downstream send progress by stream. #10735 handles advancing the shared history retain boundary by downstream progress while preserving the base replay window and full-sync retains.

pkg/syncer/server.go
```diff
@@ -561,9 +631,15 @@ func (s *RegionSyncer) sendRegionSyncResponse(
...
+	case <-sender.done:
+		return false
+	default:
+	}
+
 	resultCh := make(chan error, 1)
 	go func() {
-		resultCh <- sender.send(regions)
+		resultCh <- sender.stream.Send(regions)
 	}()
```
lhy1024, line 709: One small send-lock issue remains here. `sendRegionSyncResponse` is called while the caller holds `sendMu`, but the actual `stream.Send` runs in this goroutine. If `sender.done`, `ctx.Done`, or the timer wins first, the caller returns and releases `sendMu` while this goroutine may still be blocked inside `stream.Send`.

  Then `closeAllClient` can call `sender.send(resp)` and enter another `Send` on the same gRPC stream concurrently. The previous code used `sender.send` inside the goroutine, so `sendMu` stayed held until the underlying `Send` returned.

  Could we keep the per-stream send serialization until the actual `Send` goroutine exits, or otherwise avoid sending the close response on a stream that may still have an in-flight `Send`?

lhy1024 (commented):
the rest LGTM

pkg/syncer/server.go
```diff
@@ -561,9 +631,15 @@ func (s *RegionSyncer) sendRegionSyncResponse(
...
+	select {
+	case <-sender.done:
+		return false
+	default:
+	}
+
 	resultCh := make(chan error, 1)
 	go func() {
-		resultCh <- sender.send(regions)
+		resultCh <- sender.stream.Send(regions)
```
rleungx: Could this leave a Send goroutine running after this function returns? On timeout, context cancellation, or sender.done, the outer call releases sendMu while this goroutine may still be blocked in sender.stream.Send. A later closeAllClient or sendDownstream can then call Send on the same gRPC stream concurrently, which is not safe for server streams. It would be safer to keep serialization until the Send returns, or avoid issuing another Send after a timed-out send has been abandoned.

pkg/syncer/server.go
```diff
@@ -497,56 +522,101 @@ func (s *RegionSyncer) unbindStream(name string, stream *regionSyncStream) {
...
+			return sentRecords, errors.Errorf("region syncer buffered records from index %d overflow, first available index is %d", sendIndex, firstIndex)
+		}
+		bufferNextIndex := s.history.getNextIndex()
+		if sendIndex > bufferNextIndex {
+			return sentRecords, errors.Errorf("region syncer buffered records index %d exceeds next index %d", sendIndex, bufferNextIndex)
+		}
+		if sendIndex == bufferNextIndex {
+			return sentRecords, nil
+		}
+		records := s.history.recordsBetween(sendIndex, bufferNextIndex)
```
rleungx: This copies the entire backlog from sendIndex to bufferNextIndex even though the next block only sends maxSyncRegionBatchSize records. If a downstream is far behind, each drain iteration allocates and copies a large slice, then discards most of it, and holds the history RLock while doing so. Consider capping endIndex before recordsBetween, for example min(bufferNextIndex, sendIndex+uint64(maxSyncRegionBatchSize)).

pkg/syncer/server.go
okJiang: Addressed in 46fe66980. The actual underlying stream.Send is now guarded by a separate per-stream send mutex, so a timed-out send goroutine still serializes later close/downstream sends until the underlying Send returns. Added TestSendRegionSyncResponseSerializesStreamSendAfterTimeout. Verification: pkg/syncer Test suite, tests/server/region_syncer TestRegionSyncer suite, and make check.

pkg/syncer/server.go
okJiang: Addressed in 46fe66980. drainDownstreamLocked now caps endIndex before recordsBetween, so each drain iteration reads at most maxSyncRegionBatchSize records instead of copying the whole backlog first. Verification: pkg/syncer Test suite, tests/server/region_syncer TestRegionSyncer suite, and make check.

okJiang:
cc @lhy1024 @rleungx

pkg/syncer/server.go
```diff
@@ -66,17 +66,46 @@ type ServerStream interface {
...
+
 type regionSyncStream struct {
-	stream ServerStream
-	done   chan struct{}
-	once   sync.Once
-	sendMu sync.Mutex
+	stream       ServerStream
+	sendMu       sync.Mutex
+	streamSendMu sync.Mutex
+	indexMu      syncutil.RWMutex
```
rleungx: Do we really need these locks?

lhy1024 (approved):

pkg/syncer/server.go
okJiang: Yes, both locks are still needed but they cover different levels.

`sendMu` serializes one stream's logical send flow: initial/history/full sync, downstream drain, keepalive, close response, and sendIndex advancement. It prevents those flows from interleaving with each other.

`streamSendMu` only serializes the underlying gRPC `stream.Send`. This is needed because `sendRegionSyncResponse` runs the actual Send in a goroutine so it can timeout. After timeout the caller releases `sendMu`, but that goroutine may still be blocked in `stream.Send`; `streamSendMu` prevents a later close/downstream send from entering the same gRPC stream concurrently.

So `sendMu` protects RegionSyncer state sequencing, while `streamSendMu` protects the gRPC stream after timeout.

rleungx: How about using syncutil? BTW, if we do need them, can u wrap them with the protected field?

okJiang: Addressed in 6c7a38c4b.

I changed the new stream locks to `syncutil`, wrapped the gRPC stream with its send mutex, and wrapped `sendIndex` with its RW mutex. `sendMu` remains a flow-level mutex because it serializes the whole logical send/drain sequence rather than a single field. I also routed the empty history response through `sendStream` so all actual gRPC sends use the same stream lock.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `make check`

pkg/syncer/server.go
```diff
@@ -66,17 +66,60 @@ type ServerStream interface {
...
+	stream struct {
+		syncutil.Mutex
+		ServerStream
+	}
+	sendMu syncutil.Mutex
+	index  struct {
+		syncutil.RWMutex
+		sendIndex uint64
+	}
+	notifyCh chan downstreamSendSignal
```
bufferflies: why not using chan bool?

pkg/syncer/server.go
```diff
@@ -66,17 +66,60 @@ type ServerStream interface {
 	Send(regions *pdpb.SyncRegionResponse) error
 }
 
+type downstreamSendSignal struct {
+	keepAlive bool
+}
+
 type regionSyncStream struct {
```
bufferflies, line 69: I'm afraid it will cause a deadlock if we add four locks to one struct. And it doesn't needs high performance here.

pkg/syncer/server.go
```diff
@@ -89,6 +132,12 @@ func (s *regionSyncStream) close() {
 func (s *regionSyncStream) send(regions *pdpb.SyncRegionResponse) error {
 	s.sendMu.Lock()
 	defer s.sendMu.Unlock()
+	return s.sendStream(regions)
+}
+
+func (s *regionSyncStream) sendStream(regions *pdpb.SyncRegionResponse) error {
+	s.stream.Lock()
```
bufferflies, line 163: why need another locker here?

pkg/syncer/server.go
okJiang: Addressed in 364ac8b5e. `notifyCh` now uses `chan bool` directly since it only carries the keepalive flag.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `make check`

pkg/syncer/server.go
okJiang: Addressed in 364ac8b5e. I kept this lock because it protects the actual gRPC `stream.Send`, not the logical send sequence. When `sendRegionSyncResponse` times out, its Send goroutine may still be blocked after `sendMu` is released; this lock keeps later sends serialized at the gRPC stream level.

I also reduced the other lock by merging `sendIndex` under `sendMu`.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `make check`

pkg/syncer/server.go
okJiang: Addressed in 364ac8b5e. I removed the separate `sendIndex` RWMutex and let `sendMu` protect both the logical send flow and `sendIndex`, so the stream state now has fewer locks.

The only separate lock left is around the underlying gRPC `Send`, because the timeout path can return while its Send goroutine is still blocked. That lock prevents a later close/downstream send from entering the same gRPC stream concurrently.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `make check`

pkg/syncer/server.go
```diff
@@ -89,6 +138,12 @@ func (s *regionSyncStream) close() {
 func (s *regionSyncStream) send(regions *pdpb.SyncRegionResponse) error {
 	s.sendMu.Lock()
 	defer s.sendMu.Unlock()
```
rleungx, line 149: Could we also keep `closeAllClient` bounded now that `send()` goes through the serialized `sendStream` path? `closeAllClient` still calls `sender.close()` and then `sender.send(resp)` synchronously. If an earlier `sendRegionSyncResponse` timed out, its goroutine may still be blocked in `sendStream` holding the stream mutex, so this close notification can wait forever and block `RunServer` shutdown. Using `sendRegionSyncResponse` with the syncer timeout there, or making the close notification best-effort, would avoid turning one stuck downstream stream into a shutdown hang.

okJiang: Addressed in d9ef526d3. `closeAllClient` now sends the close notification through the timeout-bounded `sendRegionSyncResponse` path and closes the stream afterward, so a blocked downstream `Send` cannot hang `RunServer` shutdown forever. The test now covers a blocked close send returning by timeout.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run TestCloseAllClientTimesOutBlockedSend -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `git diff --check`
- `make check`

okJiang: Follow-up after merging the latest master in b8bc717bb: master already changed `closeAllClient` to close all streams first and send close notifications in parallel. I kept that behavior, but added a close-notification-specific send path that still uses the syncer timeout and does not skip the send just because `sender.done` is already closed.

So the shutdown path now preserves master's close-all-first behavior while still keeping the close notification bounded when an earlier `Send` is stuck.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run TestCloseAllClientTimesOutBlockedSend -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `git diff --check`
- `make check`

pkg/syncer/server.go
```diff
@@ -334,64 +383,78 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
 		zap.Uint64("from-index", startIndex),
-		zap.Uint64("last-index", nextIndex),
+		zap.Uint64("last-index", endIndex),
 		zap.Int("records-length", len(records)))
-	// TODO: Incremental history replay is sent before binding the stream, so
-	// broadcasts produced during this replay are not delivered to this follower.
-	// Handle this existing ordering gap in a follow-up PR together with stream
-	// binding and send-ordering semantics.
-	return nil, s.syncHistoryRecords(startIndex, records, stream)
+	return s.syncHistoryRecordsLocked(startIndex, records, syncStream)
```
rleungx, line 420: During incremental history replay, this stream is already bound but its downstream sender is not started until syncHistoryRegion returns. If live records after endIndex overflow the shared history buffer while this Send is blocked, the later drain will close the stream. Should we retain from the bind-time send index here like full sync does, or add a test documenting that overflow/disconnect is intentional?

okJiang: Addressed in b5777ed25 by adding `TestIncrementalHistoryReplayOverflowDisconnectsStream`.

This documents the intended boundary: the shared history buffer is already adaptive, so if live records produced after the bind-time send index still overflow the buffer while incremental history replay is blocked, the stream is disconnected and the follower should reconnect/resync. I kept this different from full sync retain to avoid letting a stuck incremental replay pin the leader-side history buffer indefinitely.

Verification:
- `make gotest GOTEST_ARGS='./pkg/syncer -run TestIncrementalHistoryReplayOverflowDisconnectsStream -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`
- `make gotest GOTEST_ARGS='./tests/server/region_syncer -run TestRegionSyncer -count=1'`
- `git diff --check`

pkg/syncer/server.go
```diff
@@ -334,64 +383,78 @@ func (s *RegionSyncer) syncHistoryRegion(ctx context.Context, request *pdpb.Sync
...
+func (s *RegionSyncer) syncHistoryRecords(startIndex uint64, records []*core.RegionInfo, stream *regionSyncStream) error {
+	stream.sendMu.Lock()
+	defer stream.sendMu.Unlock()
+	return s.syncHistoryRecordsLocked(startIndex, records, stream)
+}
+
+func (*RegionSyncer) syncHistoryRecordsLocked(startIndex uint64, records []*core.RegionInfo, stream *regionSyncStream) error {
+	for start := 0; start < len(records); start += maxSyncRegionBatchSize {
+		end := min(start+maxSyncRegionBatchSize, len(records))
+		if err := stream.sendStream(buildSyncRegionResponse(startIndex+uint64(start), records[start:end])); err != nil {
```
rleungx: Could we check `stream.done` before each initial replay/full-sync batch? When the stream is replaced, `oldStream.close()` only closes `done`, but these direct `sendStream` calls may keep sending on the old RPC.

okJiang: Fixed. Initial replay and full sync now check whether the downstream stream has been closed before sending each batch, so a replaced old stream stops before continuing with more batches. Added coverage for both incremental replay and full sync closing between batches.

Verified with `make gotest GOTEST_ARGS="./pkg/syncer -run Test -count=1"`.

rleungx (approved):

PR #10723, opened 2026-05-27
core: keep RegionInfo ref count consistent on flow-only updates

### LINKED ISSUE #10722: core: flow-only region updates can leave RegionInfo ref count inconsistent
### What did you do?

While reviewing the region heartbeat cache update path on master, I found that flow-only region updates can make the `RegionInfo` reference count inconsistent.

Relevant path on `upstream/master` (`9500cf3c7`):

1. `GenerateRegionGuideFunc` marks a heartbeat with changed `RoundBytesWritten` / `RoundBytesRead` as `saveCache=true`.
2. `processRegionHeartbeat` calls `CheckAndPutRootTree`, then schedules `CheckAndPutSubTree`.
3. For same-range root tree updates, `setRegionLocked` calls `r.tree.updateStat(origin, region)`.
4. Since the root tree is `countRef=true`, `updateStat` runs `origin.DecRef()` and `region.IncRef()`.
5. For same-range and same-peer subtree updates, `preUpdateSubTreeLocked` takes the fast path:

```go
if tree, ok := r.subRegions[region.GetID()]; ok {
    r.updateSubTreeStat(origin, region)
    tree.RegionInfo = region
}
return true
```

This replaces the `RegionInfo` in the subtree / overlapTree shared `regionItem`, but it does not call `r.overlapTree.update(...)`. The full subtree insertion path is the path that increments the overlapTree reference.

A focused unit test can reproduce this with:

1. Insert an initial region through `CheckAndPutRegion`.
2. Update the same region with only flow fields changed through `CheckAndPutRootTree` + `CheckAndPutSubTree`.
3. Check the new root region's ref count.

A live `RegionInfo` referenced by both the root tree and overlapTree should have ref count `2`.

Flow-only subtree updates should preserve the same root tree / overlapTree ref-count symmetry as the full subtree update path.

### What did you see instead?

After a flow-only update, the new `RegionInfo` can receive only the root tree reference, so `GetRef()` stays at `1`, while the subtree / overlapTree `regionItem` has already been switched to that same object.

Later heartbeats with no cache changes can hit:

```go
if origin.GetRef() < 2 {
    ctx.TaskRunner.RunTask(
        regionID,
        ratelimit.UpdateSubTree,
        func(context.Context) {
            c.CheckAndPutSubTree(region)
        },
        ratelimit.WithRetained(true),
    )
}
```

This can enqueue retained `UpdateSubTree` work even though the region is already present in the subtree. Running `CheckAndPutSubTree` again can take the same fast path and still not repair the missing overlapTree ref.

In large clusters, this may cause continuous redundant `UpdateSubTree` scheduling and retained task buildup after heartbeat bursts, increasing heap pressure.

master, `upstream/master` at `9500cf3c7`.

JmPotato:

Issue Number: Close #10722

Flow-only region heartbeats (only `RoundBytesWritten` / `RoundBytesRead` changed) update the root tree via `r.tree.updateStat`, which moves the ref count from the previous `RegionInfo` to the new one. However, the same-range / same-peer fast path in `preUpdateSubTreeLocked` only reassigns the shared `regionItem` (`tree.RegionInfo = region`) and never moves the ref-counted `overlapTree` reference.

As a result, the live region is left at `ref == 1` even though it is still present in the subtree. Later no-change heartbeats then keep hitting:

```go
if origin.GetRef() < 2 {
    ctx.TaskRunner.RunTask(
        regionID,
        ratelimit.UpdateSubTree,
        func(context.Context) { c.CheckAndPutSubTree(region) },
        ratelimit.WithRetained(true),
    )
}
```

which enqueues retained `UpdateSubTree` tasks that can never repair the count (the repair re-runs the same fast path), causing redundant subtree scheduling and retained task buildup after heartbeat bursts (e.g. a full re-report following a PD leader switch).

`overlapTree` is the only ref-counted subtree, and the fast path repoints its shared item to the new `RegionInfo`. So its single reference must be transferred from `origin` to `region`, mirroring what the root tree's `updateStat` already does for the in-place replacement.

- Add `regionTree.updateRef(origin, region)`, which transfers the reference for an in-place item replacement and is a no-op for trees that do not count references (keyed off the tree's `countRef` property).
- Call it on `overlapTree` in the flow-only fast path so a fully tracked region keeps `ref == 2`.
- Reuse the same helper inside `regionTree.updateStat`, since it performs the identical `countRef`-gated transfer.

No extra tree operations or stat recomputation are introduced; only the reference is transferred.

### Check List

Tests

- Unit test

Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed region reference handling so shared region references remain correct when only flow/statistics change, preventing incorrect live-state drift after updates.

* **Tests**
  * Added a test ensuring reference counts and leadership remain stable across flow-only updates and redundant subtree updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/pd/pull/10723?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -1239,6 +1239,8 @@ func (r *RegionsInfo) preUpdateSubTreeLocked(
 		// t4: thread-A: update region subtree
 		// to keep region tree consistent with subtree, we need to drop this update.
 		if tree, ok := r.subRegions[region.GetID()]; ok {
+			// Fetch the origin region info from the subtree again to ensure it is up-to-date.
+			origin := tree.RegionInfo
 			r.updateSubTreeStat(origin, region)
 			// overlapTree is the only ref-counted subtree and the shared item is
 			// repointed to region below, so transfer its reference here. Otherwise
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -1240,6 +1240,12 @@ func (r *RegionsInfo) preUpdateSubTreeLocked(
 		// to keep region tree consistent with subtree, we need to drop this update.
 		if tree, ok := r.subRegions[region.GetID()]; ok {
 			r.updateSubTreeStat(origin, region)
+			// overlapTree is the only ref-counted subtree and we repoint its
+			// shared item to region below, so move that one reference from origin
+			// to region. Otherwise a flow-only update leaves the live region stuck
+			// at ref 1 while it is still present in the subtree.
+			origin.DecRef()
+			region.IncRef()
 			tree.RegionInfo = region
 		}
 		return true
--- a/pkg/core/region.go
+++ b/pkg/core/region.go
@@ -1240,12 +1240,11 @@ func (r *RegionsInfo) preUpdateSubTreeLocked(
 		// to keep region tree consistent with subtree, we need to drop this update.
 		if tree, ok := r.subRegions[region.GetID()]; ok {
 			r.updateSubTreeStat(origin, region)
-			// overlapTree is the only ref-counted subtree and we repoint its
-			// shared item to region below, so move that one reference from origin
-			// to region. Otherwise a flow-only update leaves the live region stuck
-			// at ref 1 while it is still present in the subtree.
-			origin.DecRef()
-			region.IncRef()
+			// overlapTree is the only ref-counted subtree and the shared item is
+			// repointed to region below, so transfer its reference here. Otherwise
+			// a flow-only update leaves the live region stuck at ref 1 while it is
+			// still present in the subtree.
+			r.overlapTree.updateRef(origin, region)
 			tree.RegionInfo = region
 		}
 		return true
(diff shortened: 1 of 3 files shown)


rleungx:
I think this still has an out-of-order subtree update issue. In the fast path, `origin` comes from the root-tree update, but the subtree may already point to a newer `tree.RegionInfo` if another async subtree update ran first.

In that case, `r.overlapTree.updateRef(origin, region)` transfers the ref from the wrong object: the newer root region can keep `ref == 2` even after the stale update overwrites the subtree, so the later `origin.GetRef() < 2` repair path will not trigger.

Could we transfer from the current subtree item instead? For example:

```go
current := tree.RegionInfo
r.updateSubTreeStat(current, region)
r.overlapTree.updateRef(current, region)
tree.RegionInfo = region
```

This keeps the ref count aligned with the actual subtree owner, and stale subtree updates can still be repaired later.

lhy1024 (approved):

rleungx (approved):

PR #10730, opened 2026-05-28
server, core: backport QueryRegion support to release-8.5

JmPotato:

Issue Number: ref #8690

QueryRegion (tracked by #8690) is a gRPC streaming service that lets clients resolve regions by keys, previous keys, and region IDs over one bidirectional stream. Its server/core implementation landed on `master` across #8979, #9055, #9076, #9196, and #10194.

This PR backports only the server/core side of QueryRegion support to `release-8.5`. It also includes the related core cleanup from #9079 to avoid redundant region lookups while reading adjacent or key-matched regions. It intentionally does not include the PD client QueryRegion stream path, router client changes, router microservice, `routerpb`, or the master inner-client package layout.

```commit-message
Backport QueryRegion server/core support to release-8.5:

- Add RegionsInfo.QueryRegions and the GrpcServer.QueryRegion stream handler, resolving regions by keys, previous keys, and IDs via batched region-tree lookups.
- Avoid redundant region-map lookups in GetRegionByKey, GetPrevRegionByKey, and GetAdjacentRegions (#9079).
- Allow followers to serve QueryRegion from the region syncer, mirroring the existing follower handling of GetRegion.
- Add per-query-type metrics (pd_core_query_region_duration_seconds, pd_core_query_region_count) and the server query_region_duration_seconds metric.
- Optimize batched ID lookups with getRegionsByIDs and a generic slice.SplitIntoBatches helper.
- Raise batchSearchSize to 128.

Adaptations for release-8.5: rateLimitCheck and errs.ErrMismatchClusterID do not exist on this branch, so the QueryRegion handler inlines the equivalent rate-limit check and cluster-id mismatch error. The branch keeps the current release-8.5 kvproto version and avoids all client/router-service/routerpb changes.
```

### Check List

Tests

- Unit test (`go test ./pkg/core ./pkg/slice ./server`)

### Release note

```release-note
Improve PD server region lookup performance by supporting batched region queries.
```

### DIFF
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1663,17 +1663,34 @@ func (s *GrpcServer) QueryRegion(stream pdpb.PD_QueryRegionServer) error {
 			return status.Errorf(codes.FailedPrecondition,
 				"mismatch cluster id, need %d but got %d", clusterID, request.GetHeader().GetClusterId())
 		}
-		rc := s.GetRaftCluster()
-		if rc == nil {
-			resp := &pdpb.QueryRegionResponse{
-				Header: notBootstrappedHeader(),
+		var (
+			rc          *cluster.RaftCluster
+			needBuckets bool
+		)
+		if s.member.IsLeader() {
+			rc = s.GetRaftCluster()
+			if rc == nil {
+				resp := &pdpb.QueryRegionResponse{
+					Header: notBootstrappedHeader(),
+				}
+				if err = stream.Send(resp); err != nil {
+					return errors.WithStack(err)
+				}
+				continue
 			}
-			if err = stream.Send(resp); err != nil {
-				return errors.WithStack(err)
+			needBuckets = rc.GetStoreConfig().IsEnableRegionBucket() && request.GetNeedBuckets()
+		} else {
+			rc = s.cluster
+			if !rc.GetRegionSyncer().IsRunning() {
+				resp := &pdpb.QueryRegionResponse{
+					Header: regionNotFound(),
+				}
+				if err = stream.Send(resp); err != nil {
+					return errors.WithStack(err)
+				}
+				continue
 			}
-			continue
 		}
-		needBuckets := rc.GetStoreConfig().IsEnableRegionBucket() && request.GetNeedBuckets()
 
 		start := time.Now()
 		keyIDMap, prevKeyIDMap, regionsByID := rc.QueryRegions(
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1682,7 +1682,7 @@ func (s *GrpcServer) QueryRegion(stream pdpb.PD_QueryRegionServer) error {
 			request.GetIds(),
 			needBuckets,
 		)
-		regionQueryDuration.Observe(time.Since(start).Seconds())
+		queryRegionDuration.Observe(time.Since(start).Seconds())
 		// Build the response and send it to the client.
 		response := &pdpb.QueryRegionResponse{
 			Header:       wrapHeader(),
--- a/server/grpc_service.go
+++ b/server/grpc_service.go
@@ -1636,12 +1636,64 @@ func (s *GrpcServer) GetRegionByID(ctx context.Context, request *pdpb.GetRegionB
 	}, nil
 }
 
-// QueryRegion implements gRPC PDServer.
-//
-// release-8.5 has not backported the QueryRegion feature. Keep the new
-// kvproto RPC explicitly unimplemented instead of pulling in unrelated logic.
-func (*GrpcServer) QueryRegion(_ pdpb.PD_QueryRegionServer) error {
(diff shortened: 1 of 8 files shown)


okJiang:
Thanks for the backport. A few things need to be fixed before this can merge:

1. The PR is currently based on an older `release-8.5` and conflicts with the current branch. `release-8.5` already bumped `kvproto` to `v0.0.0-20260526084754-39498d6b17fc` and added an explicit `QueryRegion` unimplemented stub. When resolving the conflict, please keep the newer `kvproto` / `go.sum` entries from `release-8.5` and replace only the stub with this server implementation; otherwise the backport would regress the protocol version that just landed.

2. The release note is still `None`, but this changes `release-8.5` from returning `Unimplemented` for `QueryRegion` to serving the gRPC stream on the server side. Since that is user-visible behavior, please add a release note or explain why it should remain hidden.

3. The added tests cover `RegionsInfo.QueryRegions` and slice batching, but not the new gRPC stream handler. Please add a minimal server-level test for `QueryRegion`, so this backport verifies that the RPC no longer returns `Unimplemented` and can return a normal response for a basic query.

Checked locally:
- `go test ./pkg/core -run TestQueryRegions -count=1`
- `go test ./pkg/slice -run TestSliceSplitIntoBatches -count=1`
- `go test ./server -run '^$' -count=0`

server/grpc_service.go
```diff
@@ -1636,6 +1636,83 @@ func (s *GrpcServer) GetRegionByID(ctx context.Context, request *pdpb.GetRegionB
...
+
+		if clusterID := keypath.ClusterID(); request.GetHeader().GetClusterId() != clusterID {
+			return status.Errorf(codes.FailedPrecondition,
+				"mismatch cluster id, need %d but got %d", clusterID, request.GetHeader().GetClusterId())
+		}
+		var (
+			rc          *cluster.RaftCluster
+			needBuckets bool
+		)
+		if s.member.IsLeader() {
```
rleungx, line 1670: This stream handler bypasses the existing follower-handle / forwarding validation used by `GetRegion`, `GetPrevRegion`, and `GetRegionByID`. Those RPCs go through `unaryFollowerMiddleware` / `validateRoleInRequest`, so a request hitting a follower is only served locally when follower handling is explicitly allowed; otherwise it follows the normal not-leader / follower-handle-not-allowed path. Here any non-leader directly reads from `s.cluster` and returns `regionNotFound` when the syncer is not running.

If `QueryRegion` is expected to mirror the existing region RPCs, please preserve the same follower validation semantics for this stream RPC, or add tests/documentation showing that `QueryRegion` intentionally allows follower serving without the follower-handle opt-in.

JmPotato: I understand that the `QueryRegion` feature has inherently supported follower processing since its design and inception. As a result, it does not require the unary follower middleware used by its predecessors, such as the `GetRegion` request, to determine whether follower processing is permitted.

This implementation is currently the same on the master branch. If we need to add relevant comments, I should submit a new PR on master with the necessary explanation. However, this can be done independently of the current cherry-pick PR. You can go ahead and review the other parts of this PR first.

JmPotato:
> Thanks for the backport. A few things need to be fixed before this can merge:
> 
> 1. The PR is currently based on an older `release-8.5` and conflicts with the current branch. `release-8.5` already bumped `kvproto` to `v0.0.0-20260526084754-39498d6b17fc` and added an explicit `QueryRegion` unimplemented stub. When resolving the conflict, please keep the newer `kvproto` / `go.sum` entries from `release-8.5` and replace only the stub with this server implementation; otherwise the backport would regress the protocol version that just landed.
> 2. The release note is still `None`, but this changes `release-8.5` from returning `Unimplemented` for `QueryRegion` to serving the gRPC stream on the server side. Since that is user-visible behavior, please add a release note or explain why it should remain hidden.
> 3. The added tests cover `RegionsInfo.QueryRegions` and slice batching, but not the new gRPC stream handler. Please add a minimal server-level test for `QueryRegion`, so this backport verifies that the RPC no longer returns `Unimplemented` and can return a normal response for a basic query.
> 
> Checked locally:
> 
> * `go test ./pkg/core -run TestQueryRegions -count=1`
> * `go test ./pkg/slice -run TestSliceSplitIntoBatches -count=1`
> * `go test ./server -run '^$' -count=0`

1. The conflicts have been addressed.
2. This PR only involves changes related to the cherry-pick for server-side `QueryRegion` implementation.  We can update the release notes in the final cherry-pick PR once the entire `QueryRegion` feature is completed, rather than including them in every single PR.
3. As mentioned by 2, when we cherry-pick the client-side changes later, the relevant integration tests will be added. This will cover the code path you mentioned, so I don't believe we need to introduce additional specialized tests in the current PR, especially since they don't exist on the master branch either.

rleungx (approved):

okJiang (approved):

PR #10733, opened 2026-05-28
syncer: handle region sync close responses safely

### LINKED ISSUE #10731: syncer: handle region sync close responses safely
## Problem

Region syncer has two shutdown-path issues on master:

1. The follower handles a `SyncRegionResponse` with `Header.Error` as normal sync data. `closeAllClient` sends an error response without `StartIndex`, so the follower can compare its current history index with `0` and reset the local history index to `0`.
2. `closeAllClient` closes and sends the shutdown response stream by stream. If one `Send` blocks, the remaining streams are not closed until that send returns.

## Expected behavior

- A follower should check `resp.GetHeader().GetError()` before normal region sync handling and should not mutate local history state for an error response.
- Leader shutdown should close all tracked streams first, then send shutdown responses through the existing timeout-safe send path so one blocked send cannot delay closing the rest.

## Context

These were found while reviewing the release-8.5 cherry-pick PR #10709, but the same risks exist on master.

okJiang:

Issue Number: Close #10731

Region syncer shutdown can currently hit two unsafe paths:

- A follower treats a `SyncRegionResponse` with `Header.Error` as normal sync data, so a close response without `StartIndex` can reset the local history index to `0`.
- `closeAllClient` closes and sends shutdown responses one stream at a time, so one blocked send can delay closing the remaining streams.

### What changed?

- Check `resp.GetHeader().GetError()` before normal region sync response handling and skip local history mutation for error responses.
- Close all tracked streams first in `closeAllClient`, then send shutdown responses through the existing timeout-safe send path in parallel.
- Add regression tests for both behaviors.

### Tests

- `go test ./pkg/syncer -run 'TestHandleRegionSyncResponseSkipsErrorResponse|TestCloseAllClientClosesStreamsBeforeSend|TestBroadcastClosesStreamWhenSendBlocks|TestSyncExitsWhenBroadcastSendFails' -count=1 -v -timeout=2m`\n- `go test ./pkg/syncer -count=1 -timeout=2m`

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved region synchronization to detect leader errors, stop broken streams, and automatically retry with a controlled wait interval.
  * Sync loops now reliably close and retry streams when the leader steps down or context is canceled.
* **Refactor**
  * Server shutdown now snapshots and closes client streams promptly, sending shutdown notices concurrently for more reliable termination.
* **Tests**
  * Added coverage for handling error responses in sync flows.
  * Reduced flakiness in shutdown tests with tighter timing control.

### DIFF
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -621,7 +621,6 @@ func (s *RegionSyncer) closeAllClient() {
 				},
 			},
 		}
-		stream := stream
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
--- a/pkg/syncer/server.go
+++ b/pkg/syncer/server.go
@@ -594,24 +594,19 @@ func (s *RegionSyncer) sendRegionSyncResponse(
 }
 
 func (s *RegionSyncer) closeAllClient() {
-	type streamWithName struct {
-		name   string
-		sender *regionSyncStream
-	}
-
 	s.mu.RLock()
-	streams := make([]streamWithName, 0, len(s.mu.streams))
+	streams := make(map[string]*regionSyncStream, len(s.mu.streams))
 	for name, sender := range s.mu.streams {
-		streams = append(streams, streamWithName{name: name, sender: sender})
+		streams[name] = sender
 	}
 	s.mu.RUnlock()
 
-	for _, stream := range streams {
-		stream.sender.close()
+	for _, sender := range streams {
+		sender.close()
 	}
 
-	var wg sync.WaitGroup
-	for _, stream := range streams {
+	wg := &sync.WaitGroup{}
+	for name, sender := range streams {
 		resp := &pdpb.SyncRegionResponse{
 			Header: &pdpb.ResponseHeader{
 				ClusterId: keypath.ClusterID(),
@@ -622,10 +617,10 @@ func (s *RegionSyncer) closeAllClient() {
 			},
 		}
 		wg.Add(1)
-		go func() {
+		go func(name string, sender *regionSyncStream, resp *pdpb.SyncRegionResponse) {
 			defer wg.Done()
-			s.sendRegionSyncResponse(context.Background(), stream.name, stream.sender, resp)
-		}()
+			s.sendRegionSyncResponse(context.Background(), name, sender, resp)
+		}(name, sender, resp)
 	}
 	wg.Wait()
 }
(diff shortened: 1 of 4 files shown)


pkg/syncer/server.go
```diff
@@ -594,13 +594,24 @@ func (s *RegionSyncer) sendRegionSyncResponse(
...
+		sender *regionSyncStream
+	}
+
 	s.mu.RLock()
-	streams := make([]*regionSyncStream, 0, len(s.mu.streams))
-	for _, sender := range s.mu.streams {
-		streams = append(streams, sender)
+	streams := make([]streamWithName, 0, len(s.mu.streams))
+	for name, sender := range s.mu.streams {
+		streams = append(streams, streamWithName{name: name, sender: sender})
```
okJiang: Isn't it better to use a map instead of a new struct?

pkg/syncer/server.go
```diff
@@ -610,9 +616,11 @@ func (s *RegionSyncer) closeAllClient() {
 				},
 			},
 		}
-		sender.close()
-		if err := sender.send(resp); err != nil {
-			log.Warn("region syncer send close message meet error", errs.ZapError(errs.ErrGRPCSend, err))
-		}
+		wg.Add(1)
+		go func(name string, sender *regionSyncStream, resp *pdpb.SyncRegionResponse) {
+			defer wg.Done()
```
rleungx, line 622: log panic

rleungx (approved):
The rest LGTM

pkg/syncer/server.go
```diff
@@ -610,9 +616,12 @@ func (s *RegionSyncer) closeAllClient() {
...
 		}
-		sender.close()
-		if err := sender.send(resp); err != nil {
-			log.Warn("region syncer send close message meet error", errs.ZapError(errs.ErrGRPCSend, err))
-		}
+		wg.Add(1)
+		go func(name string, sender *regionSyncStream, resp *pdpb.SyncRegionResponse) {
+			defer logutil.LogPanic()
+			defer wg.Done()
+			s.sendRegionSyncResponse(context.Background(), name, sender, resp)
```
bufferflies: How about using syncer context rather than background?

bufferflies (commented):
rest lgtm

bufferflies (approved):

PR #10735, opened 2026-05-28
syncer: retain history for slow downstreams

### LINKED ISSUE #10734: syncer: retain RegionSyncer history by downstream progress
## Enhancement Task

RegionSyncer is moving toward a shared leader-side history buffer with per-stream send progress in #10716. After that change, the history buffer needs to retain records according to active downstream progress, not only according to the initial history-sync window.

### Background

A review discussion in #10716 pointed out that advancing a follower stream's send index does not move the shared history buffer head. The existing dynamic buffer logic mostly observes the required window when a stream is created, so live sync can leave the buffer retaining more history than the active streams need, or fail to reserve enough window for a slow stream before new records are appended.

Discussion: https://github.com/tikv/pd/pull/10716#discussion_r3315915614

### Goal

Make the shared RegionSyncer history buffer account for downstream send progress during live sync.

Expected behavior:

- Before appending live records, reserve enough history window for the slowest active downstream stream.
- After a stream advances or is removed, advance the shared history retain boundary using the slowest remaining downstream stream.
- Preserve the base recent-history window for reconnect/history sync.
- Preserve active full-sync retain points.
- Keep the SyncRegions protocol unchanged.

### Non-goals

- Do not add ack handling.
- Do not implement full missing-index resend logic.
- Do not change follower cache rebuild or follower-read readiness behavior.

### Validation

- Verify the history buffer grows before overwriting records needed by a slow downstream.
- Verify the history buffer head can advance after downstream progress moves forward.
- Verify active full-sync retain points still prevent old records from being dropped.
- Verify existing RegionSyncer behavior tests still pass.

okJiang:

Issue Number: Close #10734

Follow-up to #10716.

This PR builds on the RegionSyncer shared-history design merged in #10716. After each downstream stream only keeps its own send index, the shared leader history buffer needs to avoid overwriting records that a slow downstream still has not sent during live sync, not only during the one-time history-sync window.

```commit-message
Make the RegionSyncer history buffer account for downstream send progress.

Before appending live region records, the leader reserves enough shared history window for the slowest active downstream stream. The send index itself does not move the buffer first index; it only describes the earliest record still needed by active downstreams. Before shrinking, the leader observes the current downstream window again so shrink will not drop records still needed by a slow downstream.

The history buffer still keeps the base recent replay window for reconnect/history sync and still respects active full-sync retain points. The SyncRegions protocol is unchanged.
```

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
None.
```

## Summary by CodeRabbit

* **New Features**
  * Improved history retention and replay-window management so retained records aren’t overwritten; append capacity is reserved before recording and downstream windows are observed on shrink ticks.
  * Region batching now reserves history capacity for slow downstreams before recording and notifies downstreams after batching; downstream progress drives retention and windowing decisions.

* **Tests**
  * Added and revised tests validating reserve/observe window behavior, downstream-driven retention, and downstream drain/observe interactions.

### DIFF
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -122,14 +122,7 @@ func (h *historyBuffer) capacity() int {
 	return h.size - 1
 }
 
-func (h *historyBuffer) record(r *core.RegionInfo) {
-	h.Lock()
-	defer h.Unlock()
-	h.prepareRequiredWindowLocked(1)
-	h.recordLocked(r)
-}
-
-func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
+func (h *historyBuffer) record(records ...*core.RegionInfo) {
 	if len(records) == 0 {
 		return
 	}
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -125,21 +125,17 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
-	h.prepareRequiredWindowLocked(0, 1)
+	h.prepareRequiredWindowLocked(1)
 	h.recordLocked(r)
 }
 
 func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
-	h.recordBatchWithRequiredWindow(records, 0)
-}
-
-func (h *historyBuffer) recordBatchWithRequiredWindow(records []*core.RegionInfo, requiredWindow uint64) {
 	if len(records) == 0 {
 		return
 	}
 	h.Lock()
 	defer h.Unlock()
-	h.prepareRequiredWindowLocked(requiredWindow, uint64(len(records)))
+	h.prepareRequiredWindowLocked(uint64(len(records)))
 	for _, r := range records {
 		h.recordLocked(r)
 	}
@@ -160,14 +156,13 @@ func (h *historyBuffer) recordLocked(r *core.RegionInfo) {
 	}
 }
 
-func (h *historyBuffer) prepareRequiredWindowLocked(requiredWindow uint64, appendCount uint64) {
+func (h *historyBuffer) prepareRequiredWindowLocked(appendCount uint64) {
 	if retainIndex, ok := h.minRetainIndexLocked(); ok {
 		endIndex := h.index + appendCount
 		if retainIndex < endIndex {
-			requiredWindow = max(requiredWindow, endIndex-retainIndex)
+			h.growForWindowLocked(endIndex - retainIndex)
 		}
 	}
-	h.growForWindowLocked(requiredWindow)
 }
 
 func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -125,29 +125,21 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
-	h.prepareRequiredWindowLocked(0, false, 1)
+	h.prepareRequiredWindowLocked(0, 1)
 	h.recordLocked(r)
 }
 
 func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
-	if len(records) == 0 {
-		return
-	}
-	h.Lock()
-	defer h.Unlock()
-	h.prepareRequiredWindowLocked(0, false, uint64(len(records)))
-	for _, r := range records {
-		h.recordLocked(r)
-	}
+	h.recordBatchWithRequiredWindow(records, 0)
 }
 
-func (h *historyBuffer) recordBatchFromReplay(replayFrom uint64, records []*core.RegionInfo) {
+func (h *historyBuffer) recordBatchWithRequiredWindow(records []*core.RegionInfo, requiredWindow uint64) {
 	if len(records) == 0 {
 		return
 	}
 	h.Lock()
 	defer h.Unlock()
-	h.prepareRequiredWindowLocked(replayFrom, true, uint64(len(records)))
+	h.prepareRequiredWindowLocked(requiredWindow, uint64(len(records)))
 	for _, r := range records {
 		h.recordLocked(r)
 	}
@@ -168,16 +160,14 @@ func (h *historyBuffer) recordLocked(r *core.RegionInfo) {
 	}
 }
 
-func (h *historyBuffer) prepareRequiredWindowLocked(replayFrom uint64, hasReplayFrom bool, appendCount uint64) {
-	retainIndex, ok := h.minRequiredIndexLocked(replayFrom, hasReplayFrom)
-	if !ok {
-		return
+func (h *historyBuffer) prepareRequiredWindowLocked(requiredWindow uint64, appendCount uint64) {
+	if retainIndex, ok := h.minRetainIndexLocked(); ok {
+		endIndex := h.index + appendCount
+		if retainIndex < endIndex {
+			requiredWindow = max(requiredWindow, endIndex-retainIndex)
+		}
 	}
-	endIndex := h.index + appendCount
-	if retainIndex >= endIndex {
-		return
-	}
-	h.growForWindowLocked(endIndex - retainIndex)
+	h.growForWindowLocked(requiredWindow)
 }
 
 func (h *historyBuffer) recordsFrom(index uint64) []*core.RegionInfo {
@@ -262,27 +252,16 @@ func (h *historyBuffer) minRetainIndexLocked() (uint64, bool) {
 	return min, ok
 }
 
-func (h *historyBuffer) minRequiredIndexLocked(replayFrom uint64, hasReplayFrom bool) (uint64, bool) {
-	min, ok := h.minRetainIndexLocked()
-	if hasReplayFrom && (!ok || replayFrom < min) {
-		min = replayFrom
-		ok = true
-	}
-	return min, ok
-}
-
-func (h *historyBuffer) observeReplayWindow(replayFrom, endIndex uint64) {
+func (h *historyBuffer) observeRequiredWindow(window uint64) {
 	h.Lock()
 	defer h.Unlock()
-	h.observeReplayWindowLocked(replayFrom, true, endIndex)
+	h.observeRequiredWindowLocked(window)
 }
 
-func (h *historyBuffer) observeReplayWindowLocked(replayFrom uint64, hasReplayFrom bool, endIndex uint64) {
-	retainIndex, ok := h.minRequiredIndexLocked(replayFrom, hasReplayFrom)
-	if !ok || retainIndex >= endIndex {
+func (h *historyBuffer) observeRequiredWindowLocked(window uint64) {
+	if window == 0 {
 		return
 	}
-	window := endIndex - retainIndex
 	if window > h.observedRequiredWindow {
 		h.observedRequiredWindow = window
 	}
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -136,7 +136,9 @@ func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
 	h.prepareRequiredWindowLocked(0, false, uint64(len(records)))
-	h.recordBatchLocked(records)
+	for _, r := range records {
+		h.recordLocked(r)
+	}
 }
 
 func (h *historyBuffer) recordBatchFromReplay(replayFrom uint64, records []*core.RegionInfo) {
@@ -146,10 +148,6 @@ func (h *historyBuffer) recordBatchFromReplay(replayFrom uint64, records []*core
 	h.Lock()
 	defer h.Unlock()
 	h.prepareRequiredWindowLocked(replayFrom, true, uint64(len(records)))
-	h.recordBatchLocked(records)
-}
-
-func (h *historyBuffer) recordBatchLocked(records []*core.RegionInfo) {
 	for _, r := range records {
 		h.recordLocked(r)
 	}
--- a/pkg/syncer/history_buffer.go
+++ b/pkg/syncer/history_buffer.go
@@ -125,9 +125,37 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
-	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
-		h.growForWindowLocked(h.index - retainIndex + 1)
+	h.prepareRequiredWindowLocked(0, false, 1)
+	h.recordLocked(r)
+}
+
+func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
+	if len(records) == 0 {
+		return
 	}
+	h.Lock()
+	defer h.Unlock()
+	h.prepareRequiredWindowLocked(0, false, uint64(len(records)))
+	h.recordBatchLocked(records)
+}
+
+func (h *historyBuffer) recordBatchFromReplay(replayFrom uint64, records []*core.RegionInfo) {
+	if len(records) == 0 {
+		return
+	}
+	h.Lock()
+	defer h.Unlock()
+	h.prepareRequiredWindowLocked(replayFrom, true, uint64(len(records)))
+	h.recordBatchLocked(records)
+}
+
+func (h *historyBuffer) recordBatchLocked(records []*core.RegionInfo) {
+	for _, r := range records {
+		h.recordLocked(r)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/syncer/server.go
```diff
@@ -236,9 +236,8 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 	keepAliveTicker := time.NewTicker(syncerKeepAliveInterval)
 	shrinkTicker := time.NewTicker(historyBufferShrinkInterval)
 
-	processRegion := func(region *core.RegionInfo) {
+	appendRegion := func(region *core.RegionInfo) {
```
okJiang: This function is no longer needed.

pkg/syncer/history_buffer.go
```diff
@@ -224,6 +224,56 @@ func (h *historyBuffer) minRetainIndexLocked() (uint64, bool) {
...
+// reserveAppendWindowFrom grows the buffer before appending records so that
+// records still needed from index are not overwritten by the append.
+func (h *historyBuffer) reserveAppendWindowFrom(index uint64, appendCount int) {
+	if appendCount <= 0 {
+		return
+	}
+	h.Lock()
+	defer h.Unlock()
+	nextIndex := h.nextIndex() + uint64(appendCount)
+	index = h.oldestRequiredIndexLocked(index, nextIndex)
```
okJiang: There are too many concepts about index, and I find it hard to distinguish them. Either add the necessary comments, or reduce the number of concepts to decrease code complexity and increase readability.

pkg/syncer/server.go
okJiang: Fixed in f003407aa by removing the unnecessary appendRegion helper and appending records directly in RunServer.

Verified with:
- `make gotest GOTEST_ARGS='./pkg/syncer -run "Test(HistoryBuffer|ReserveHistoryAppend|ObserveHistoryWindow)" -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`

pkg/syncer/history_buffer.go
okJiang: Fixed in f003407aa by simplifying the replay-index naming and comments: the code now uses oldestReplayIndexLocked to describe the earliest index that must stay replayable, derives it from downstream progress, full-sync starts, and the base recent window, and removes the extra baseRequiredIndexLocked helper.

Verified with:
- `make gotest GOTEST_ARGS='./pkg/syncer -run "Test(HistoryBuffer|ReserveHistoryAppend|ObserveHistoryWindow)" -count=1'`
- `make gotest GOTEST_ARGS='./pkg/syncer -run Test -count=1'`

pkg/syncer/server.go
```diff
@@ -73,43 +73,37 @@ type regionSyncStream struct {
...
+	return syncStream
 }
 
 func (s *regionSyncStream) getSendIndex() uint64 {
-	s.sendMu.Lock()
-	defer s.sendMu.Unlock()
 	return s.getSendIndexLocked()
 }
 
 func (s *regionSyncStream) getSendIndexLocked() uint64 {
```
okJiang: This function is no longer needed.

pkg/syncer/server.go
```diff
@@ -238,7 +232,6 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 
 	processRegion := func(region *core.RegionInfo) {
```
okJiang: This function is no longer needed.

pkg/syncer/server.go
```diff
@@ -281,6 +276,43 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 	}
 }
 
+func (s *RegionSyncer) appendHistoryRecords(records []*core.RegionInfo) {
+	replayFrom, ok := s.minDownstreamSendIndex()
+	if !ok {
+		s.history.recordBatch(records)
+		return
+	}
+	s.history.recordBatchFromReplay(replayFrom, records)
```
okJiang: Doesn't the history buffer already store the minimum index? Why is it necessary to extract it from the stream and then pass it back into the history buffer?

pkg/syncer/history_buffer.go
```diff
@@ -125,9 +125,37 @@ func (h *historyBuffer) capacity() int {
...
+	defer h.Unlock()
+	h.prepareRequiredWindowLocked(replayFrom, true, uint64(len(records)))
+	h.recordBatchLocked(records)
+}
+
+func (h *historyBuffer) recordBatchLocked(records []*core.RegionInfo) {
+	for _, r := range records {
+		h.recordLocked(r)
 	}
+}
```
okJiang: Does this need to be wrapped in a function?

pkg/syncer/server.go
okJiang: The history buffer stores the shared record window and full-sync retains, but it does not own active downstream stream state. The slowest active downstream is computed in `RegionSyncer` from each stream atomic `sendIndex`, then passed as `replayFrom` so `historyBuffer` only sees an abstract replay boundary.

Keeping that direction avoids putting per-stream state or stream lifecycle handling into `historyBuffer`, and stream removal naturally stops affecting the retained window.

pkg/syncer/server.go
okJiang: Removed the one-line `processRegion` closure in cc1a7543b and append records directly in `RunServer`.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`

pkg/syncer/history_buffer.go
okJiang: Removed the `recordBatchLocked` wrapper in cc1a7543b and kept the loop inline in the two batch append paths.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`

pkg/syncer/server.go
okJiang: Removed the extra locked getter in cc1a7543b. `getSendIndex` now reads the atomic send index directly.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`

pkg/syncer/server.go
okJiang: The update of sendIndex should be completely consistent with the retains field in historyBuffer, right? (I also think they should remain consistent)

okJiang: They should be consistent in effect, but not stored in the same place. Both are replay boundaries that `historyBuffer` must respect: `retains` covers active full sync start indexes, and `sendIndex` covers active downstream live replay progress.

The consistency point is `minRequiredIndexLocked`: before append/shrink, the buffer uses the earlier boundary from full-sync retains and downstream replay progress. The state itself stays separate because their lifecycle is different: retains are acquired/released by full sync, while `sendIndex` moves per stream and disappears when the stream is unbound.

okJiang: Yes, since they are completely identical, I think we don't need to first extract from the stream and then pass it to the history buffer; we can directly retrieve it from the history buffer. Furthermore, I think we might not even need to save a copy in the stream, because they are completely identical.

okJiang: They are not completely identical. `sendIndex` is per downstream and means the next history index that this stream must send; it is needed on every live drain to build the response `StartIndex` and advance only that stream.

`retains` is different: it tracks temporary full-sync start indexes so history is not overwritten while full sync catches up, and it is released when that full sync finishes. The history buffer should combine both as retention inputs, but if `sendIndex` is moved into `historyBuffer`, the buffer would need to own per-stream progress and stream removal, effectively adding a downstream progress map there. I would keep the current split.

okJiang: I think retains are not directly released after the full synchronization is completed; after the full synchronization is completed, they just increment along with sendIndex.

okJiang: In the current PR, `retains` does not increment with `sendIndex`. It is a fixed full-sync start index retained by `retainFrom(syncStartIndex)`, and `defer releaseRetain()` releases it when `syncFullRegionsLocked` returns.

That return happens after the snapshot full sync and the retained history replay from `syncStartIndex` are both sent. After that point, the live sender uses only the stream `sendIndex` for downstream progress; the history buffer no longer has a retain entry for that stream. So the two are related replay boundaries, but their state and lifecycle are different.

pkg/syncer/server.go
```diff
@@ -281,6 +268,43 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
...
```
okJiang: I see that many places are passing around the replayFrom variable. Is there a more elegant way to handle this variable?

okJiang: Simplified this in d62753622 by removing the `replayFrom` plumbing from `historyBuffer`. `RegionSyncer` now converts downstream progress into a required replay window, and `historyBuffer` only records/observes that window size together with its own full-sync retains.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `git diff --check`

okJiang: I didn't mean to wrap it in a function. My idea was: can it just be saved directly into the history every time?

okJiang: Changed in 71ade2557. Before appending, `RegionSyncer` now observes the downstream replay window into `historyBuffer`, then calls the normal `recordBatch` path. This removes the separate required-window append wrapper and keeps the replay window stored in history for the next append/shrink decision.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `git diff --check`

pkg/syncer/history_buffer.go
```diff
@@ -125,9 +125,23 @@ func (h *historyBuffer) capacity() int {
 func (h *historyBuffer) record(r *core.RegionInfo) {
 	h.Lock()
 	defer h.Unlock()
-	if retainIndex, ok := h.minRetainIndexLocked(); ok && retainIndex <= h.index {
-		h.growForWindowLocked(h.index - retainIndex + 1)
+	h.prepareRequiredWindowLocked(1)
+	h.recordLocked(r)
+}
+
+func (h *historyBuffer) recordBatch(records []*core.RegionInfo) {
```
okJiang: `record` and `recordBatch` can be combined into a single `record` function that takes a variable-length argument list.

okJiang: Changed in 5ceb4b490. `historyBuffer.record` now accepts a variable-length argument list, and the separate `recordBatch` helper is removed. Existing single-record callers keep using `record(region)`, while batched callers pass `records...` or multiple arguments directly.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `git diff --check`

pkg/syncer/server.go
```diff
@@ -119,7 +109,7 @@ func (s *regionSyncStream) advanceSendIndex(count int) {
 }
 
 func (s *regionSyncStream) advanceSendIndexLocked(count int) {
```
bufferflies: Why not define count as uint64 directly?

pkg/syncer/server.go
```diff
@@ -281,6 +268,43 @@ func (s *RegionSyncer) RunServer(ctx context.Context, regionNotifier <-chan *cor
 	}
 }
 
+func (s *RegionSyncer) appendHistoryRecords(records []*core.RegionInfo) {
+	s.observeDownstreamReplayWindow(uint64(len(records)))
+	s.history.record(records...)
+}
```
lhy1024, line 260: There is a bind/append boundary race here. `appendHistoryRecords` first observes the downstream replay window from the current streams, then appends records to history, but these two operations are not mutually exclusive with `bindStreamForSync`.

  If a new downstream binds after `observeDownstreamReplayWindow` but before `history.record`, it receives the pre-append `startIndex`, while this append did not reserve history for it. When the batch exceeds the current capacity, the required records can be overwritten immediately, and the later `drainDownstreamLocked` will overflow and close the stream.

  I suggest making “read slowest downstream sendIndex + append history” atomic with the bind boundary, for example by holding `s.mu.RLock()` across both window observation and `history.record(records...)`, while `bindStreamForSync` keeps using `s.mu.Lock()`. Then a bind either happens before the append and participates in retention, or happens after the append and gets the post-append nextIndex.

lhy1024:
```
func TestAppendHistoryRecordsRetainsStreamBoundBetweenObserveAndRecord(t *testing.T) {
        re := require.New(t)
        stream := &testServerStream{}
        syncer := &RegionSyncer{
                history:     newTestHistoryBuffer(8),
                sendTimeout: syncerKeepAliveInterval,
        }
        syncer.mu.streams = make(map[string]*regionSyncStream)
        syncer.history.resetWithIndex(10)

        records := []*core.RegionInfo{
                newTestRegion(1),
                newTestRegion(2),
                newTestRegion(3),
        }

        // Simulate appendHistoryRecords being preempted after observing the
        // downstream window but before recording the batch.
        syncer.observeDownstreamReplayWindow(uint64(len(records)))
        syncStream, syncStartIndex := syncer.bindStreamForSync("pd2", stream)
        defer syncer.unbindStream("pd2", syncStream)
        re.Equal(uint64(10), syncStartIndex)

        syncer.history.record(records...)

        re.NoError(syncer.sendDownstream(context.Background(), "pd2", syncStream, false))
        responses := stream.sentResponses()
        re.Len(responses, 1)
        re.Equal(uint64(10), responses[0].GetStartIndex())
        re.Equal([]*metapb.Region{{Id: 1}, {Id: 2}, {Id: 3}}, responses[0].GetRegions())
        re.Equal(uint64(13), syncStream.getSendIndex())
  }
```

pkg/syncer/server.go
okJiang: Changed in f7218f94b. `advanceSendIndex` and `advanceSendIndexLocked` now take `uint64` directly, so the atomic add no longer needs to convert inside the helper.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `git diff --check`

pkg/syncer/server.go
okJiang: Fixed in f7218f94b. `appendHistoryRecords` now holds `s.mu.RLock()` while it observes the downstream window and records the batch into history, while `bindStreamForSync` still uses `s.mu.Lock()`. With that boundary, a bind either happens before the append and is included in the retained window, or happens after the append and starts from the post-append index.

Added `TestBindStreamWaitsForAppendHistoryRecordsBoundary` for this case.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run "Test(HistoryBuffer|AppendHistoryRecords|BindStreamWaits|RemovedStream|KeepAlive)" -count=1`
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `git diff --check`

pkg/syncer/server_test.go
```diff
@@ -759,6 +759,177 @@ func TestBroadcastRecordsBusyDownstreamAndDrainsLater(t *testing.T) {
...
+		newTestRegion(1),
+		newTestRegion(2),
+		newTestRegion(3),
+	}
+
+	syncer.mu.RLock()
+	bindResultCh := make(chan uint64, 1)
+	go func() {
+		syncStream, syncStartIndex := syncer.bindStreamForSync("pd2", stream)
+		defer syncer.unbindStream("pd2", syncStream)
```
rleungx, line 811: This goroutine can still be running when test cleanup starts. Because `bindResultCh` is buffered, the deferred `unbindStream` may race with cleanup iterating `streams`; please wait for the goroutine to finish before the test returns.

pkg/syncer/server.go
```diff
@@ -73,53 +73,43 @@ type regionSyncStream struct {
 		syncutil.Mutex
 		ServerStream
 	}
-	// sendMu serializes the logical send flow and protects sendIndex.
+	// sendMu serializes the logical send flow.
 	sendMu struct {
```
bufferflies, line 77: Is it necessary for syncutil.Mutex?

lhy1024 (approved):

pkg/syncer/server_test.go
okJiang: Fixed in 93c375eae. The test now waits for the bind goroutine to finish after receiving the start index, so the deferred unbind completes before test cleanup can iterate streams.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run TestBindStreamWaitsForAppendHistoryRecordsBoundary -count=1`
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `make failpoint-disable`
- `git diff --check`

pkg/syncer/server.go
okJiang: Changed in 93c375eae. `sendMu` now uses `sync.Mutex`; it only serializes this stream logical send flow and does not need the project `syncutil.Mutex` wrapper.

Verified with:
- `make gotest GOTEST_ARGS=./pkg/syncer -run TestBindStreamWaitsForAppendHistoryRecordsBoundary -count=1`
- `make gotest GOTEST_ARGS=./pkg/syncer -run Test -count=1`
- `make failpoint-disable`
- `git diff --check`

bufferflies (approved):

